PR #825, opened 2023-05-08
Feat: Log exception stack traces from executor tasks to API console

vchan:
This PR has the API send stack traces in SSEs so exceptions from executor tasks, e.g. plan/apply, don't get lost and the app can react to the errors.

### DIFF
--- a/web/server/console.py
+++ b/web/server/console.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import traceback
 import typing as t
 import unittest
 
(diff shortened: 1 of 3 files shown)


web/server/console.py
```diff
@@ -100,3 +100,9 @@ def log_test_results(
     def log_success(self, msg: str) -> None:
         self.queue.put_nowait(self._make_event(msg))
         self.stop_snapshot_progress()
+
+    def log_exception(self, tb: str) -> None:
+        self.queue.put_nowait(self._make_event(tb, event="errors", ok=False))
```
mykalmax, line 105: can we have traceback as property in payload in case we need to add more details to describe error:
```
{
    "message": "Plan Failed",
    "traceback": "..."
}
```

PR #826, opened 2023-05-08
Feat: Cache snapshot start to improve load time of large projects

crericha:


crericha:
@tobymao here is the caching approach for improving the performance of snapshot start calculations. I had to include to expose the cache in a few different places for the performance improvement to matter. Let me know what you think.

crericha:
@tobymao reverted to original version

sqlmesh/core/scheduler.py
```diff
@@ -343,8 +343,8 @@ def start_date(
     Returns:
         Start datetime object.
     """
-    if snapshot.model.start:
-        return to_datetime(snapshot.model.start)
+    if snapshot.start:
+        return to_datetime(snapshot.start)
```
tobymao, line 347: you shouldn't need to call to_datetime here because you're setting a datetime on line 367

sqlmesh/core/snapshot/definition.py
```diff
@@ -340,6 +340,7 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
     change_category: t.Optional[SnapshotChangeCategory] = None
     unpaused_ts: t.Optional[int] = None
     effective_from: t.Optional[TimeLike] = None
+    _start: t.Optional[TimeLike] = None
```
tobymao, line 343: make this a datetime

tobymao: ah i guess it's due to self.model_start

sqlmesh/core/snapshot/definition.py
```diff
@@ -738,6 +739,10 @@ def normalized_effective_from_ts(self) -> t.Optional[int]:
             else None
         )
 
+    @property
+    def start(self) -> t.Optional[TimeLike]:
+        return self.model.start or self._start
```
tobymao, line 744: return to_datetime(self.model.start or self._start)

sqlmesh/core/scheduler.py
```diff
@@ -362,6 +362,10 @@ def start_date(
         elif start_dt:
             earliest = min(earliest, start_dt)
 
+    if earliest:
+        earliest = to_datetime(earliest)
```
tobymao, line 366: snapshot._start = to_datetime(earliest)

crericha: Looks like earliest is already a datetime, so removing conversion

tobymao (approved):

PR #828, opened 2023-05-09
Fix: JSON encoding for models

vchan:
An API model used to be simpler so `json.dumps()` was good enough. However, a model now has datetime fields and other fields the standard Python `json` library can't encode out of the box.

### DIFF
--- a/web/server/watcher.py
+++ b/web/server/watcher.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import logging
 
+from fastapi.encoders import jsonable_encoder
 from watchfiles import DefaultFilter, awatch
 
 from sqlmesh.core import constants as c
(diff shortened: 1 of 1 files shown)


tobymao (approved):

web/server/watcher.py
```diff
@@ -27,8 +27,4 @@ async def watch_project(queue: asyncio.Queue) -> None:
         except Exception:
             logger.exception("Error loading context")
         else:
-            queue.put_nowait(
-                Event(
-                    event="models", data=json.dumps([model.dict() for model in get_models(context)])
-                )
-            )
+            queue.put_nowait(Event(event="models", data=jsonable_encoder(get_models(context))))
```
mykalmax, line 30: it seems we still need to wrap in `json.dumps` to send valid json
`json.dumps(jsonable_encoder(get_models(context))))`

PR #832, opened 2023-05-09
Fix: various bugs from file tree, editor, lineage, docs

mykalmax:


web/client/src/workers/sqlglot/worker.ts
```diff
@@ -27,20 +27,18 @@ const pyodideReadyPromise = loadPyodideAndPackages()
 scope.onmessage = async (e: MessageEvent) => {
   const [parse, get_dialect, dialects] = await pyodideReadyPromise
 
-  if (e.data.topic === 'parse') {
+  if (e.data.topic === 'validate') {
```
mykalmax, line 30: `validate` is still a placeholder

web/client/src/models/sqlmesh-model.ts
```diff
@@ -22,7 +16,6 @@ export class ModelSQLMeshModel<
   dialect: string
   type: string
   columns: Column[]
-  lineage: Record<string, Lineage>
```
mykalmax, line 25: `lineage` should belong to `context` and not particular `model`

vchan (approved):

PR #833, opened 2023-05-09
Fix: optimizing without the schema is inconsistent

tobymao:


GeorgeSittas:
What's the motivation behind these changes?

tobymao:
the optimizer sometimes makes wrong decisions when there's no schema. and the incorrectly optimized query gets cached. so even when we get the schema, we've already cached a bad result

GeorgeSittas:
I see, thanks for clarifying.

> so even when we get the schema, we've already cached a bad result

What would be a scenario where we don't have schema at the time of optimization, but we learn it later on? Is this possible right now somehow?

PR #840, opened 2023-05-11
Feat[breaking]: don't qualify columns on external tables.

tobymao:


tests/core/test_audit.py
```diff
@@ -144,7 +144,7 @@ def test_no_query():
 
 
 def test_macro(model: Model):
-    expected_query = "SELECT * FROM (SELECT * FROM db.test_model AS test_model WHERE test_model.ds <= '1970-01-01' AND test_model.ds >= '1970-01-01') AS _q_0 WHERE _q_0.a IS NULL"
+    expected_query = "SELECT * FROM (SELECT * FROM db.test_model AS test_model WHERE ds <= '1970-01-01' AND ds >= '1970-01-01') AS _q_0 WHERE a IS NULL"
```
GeorgeSittas, line 147: Did these tests change because we don't optimize anymore (e.g. due to not having schema or something)?

sqlmesh/core/renderer.py
```diff
@@ -269,20 +262,9 @@ def time_column_filter(self, start: TimeLike, end: TimeLike) -> exp.Between:
...
     def update_schema(self, schema: MappingSchema) -> None:
-        old_schema = self._schema
         self._schema = schema
-
-        if self.contains_star_query and old_schema != schema:
-            new_cache = {}
-            for cached_key, cached_query in self._query_cache.items():
-                new_cache[cached_key] = self._optimize_query(cached_query) or cached_query
-            self._query_cache = new_cache
+        self._query_cache.clear()
```
izeigerman, line 267: This kills the benefit of having a cache in the first place, doesn't it?

izeigerman: Basically we'll never use cached rendered query since all of them will be immediately wiped out once we set the schema.

GeorgeSittas: The schema is set once though, no? I think that for any subsequent `render_query` we should be hitting the cache?

izeigerman: It's set after we load from the cache though.

izeigerman: I'm talking about the file system cache, not the in-memory one.

GeorgeSittas: Ah, I thought you meant the latter (`_query_cache`), so I'm probably missing context regarding that. Thanks for clarifying.

tobymao: @izeigerman well then isn't the current cache wrong?

sqlmesh/core/loader.py
```diff
@@ -41,14 +42,17 @@ def update_model_schemas(dag: DAG[str], models: UniqueKeyDict[str, Model]) -> No
...
         model.update_schema(schema)
         schema.add_table(name, model.columns_to_types)
 
+        if any(dep not in models for dep in model.depends_on):
+            if "*" in model.columns_to_types:
+                raise ConfigError(
+                    f"Can't expand SELECT * expression for model '{name}'. Projections for models that use external sources must be specified explicitly at '{model._path}'"
+                )
+        else:
+            validate_qualify_columns(model.render_query())
```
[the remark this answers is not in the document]
tobymao: it checks if you're selecting a column that doesn't exist

sqlmesh/core/model/definition.py
```diff
@@ -103,11 +102,13 @@ class _Model(ModelMeta, frozen=True):
         post: Post-hooks to run after the model executes.
         expressions: All of the expressions between the model definition and final query, used for setting certain variables or environments.
         python_env: Dictionary containing all global variables needed to render the model's macros.
+        mapping_schema: The schema of table names to column and types.
     """
 
     expressions_: t.Optional[t.List[exp.Expression]] = Field(default=None, alias="expressions")
     python_env_: t.Optional[t.Dict[str, Executable]] = Field(default=None, alias="python_env")
     jinja_macros: JinjaMacroRegistry = JinjaMacroRegistry()
+    mapping_schema: t.Dict[str, t.Any] = {}
```
izeigerman, line 111: Can we do better than `t.Any` here? We can't even tell whether this can be serialized or not.

sqlmesh/core/renderer.py
```diff
@@ -293,11 +272,16 @@ def update_cache(
...
-            return optimize(query, schema=self._schema, rules=RENDER_OPTIMIZER_RULES)
-        except (SchemaError, OptimizeError):
-            return None
+            if not schema.empty:
+                qualify_columns(query, schema=schema, infer_schema=False)
         except SqlglotError as ex:
-            raise_config_error(f"Invalid model query. {ex}", self._path)
-            raise
+            raise_config_error(
+                f"Error qualifying columns, it may not exist or be ambiguous. {ex}", self._path
```
izeigerman, line 285: what's "it" here?

PR #843, opened 2023-05-11
Fix: Use pd.DataFrame to insert snapshots and environments in state sync

izeigerman:


tobymao (approved):
cc @vchan this may change your implementation

PR #845, opened 2023-05-11
Fix: Use the public SQLGlot diff API when categorizing changes

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -677,9 +677,7 @@ def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
         previous_query = previous.render_query()
         this_query = self.render_query()
 
-        edits = ChangeDistiller(t=0.5).diff(
-            previous_query, this_query, matchings=[(previous_query, this_query)]
-        )
```
GeorgeSittas, line 682: Out of curiosity: how sensitive is the distiller with respect to `t`? I think the default is `0.6` in SQLGlot?

GeorgeSittas (approved):

tobymao (approved):

sqlmesh/core/model/definition.py
izeigerman: More sensitive with smaller (sub)trees and less sensitive with the large ones. It basically represents the ratio of leaf nodes the 2 nodes should have in common in order to be matched.

PR #848, opened 2023-05-12
Fix: Improve lineage performance

mykalmax:
Improves lineage performance with when many models present. Also added filter for columns in case model has large number of columns

### DIFF
--- a/web/client/src/library/components/graph/help.ts
+++ b/web/client/src/library/components/graph/help.ts
@@ -6,7 +6,6 @@ import {
 } from '@api/client'
 import { Position, type Edge, type Node, type XYPosition } from 'reactflow'
 import { type Lineage } from '@context/editor'
-import { type ActiveColumns } from './context'
 import { type ModelSQLMeshModel } from '@models/sqlmesh-model'
 
 export interface GraphNodeData {
@@ -69,7 +68,6 @@ function getNodesAndEdges({
   nodesMap: Record<string, Node>
   edges: Edge[]
   nodes: Node[]
-  activeColumns: ActiveColumns
 } {
   const currentEdges = edges.reduce(
     (acc: Record<string, Edge>, edge) =>
@@ -93,7 +91,6 @@ function getNodesAndEdges({
     withColumns,
   })
   const outputEdges: Edge[] = []
-  const activeColumns: ActiveColumns = new Map()
 
   for (const modelSource of modelNames) {
     const modelLineage = lineage[modelSource]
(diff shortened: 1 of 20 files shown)


web/client/src/main.tsx
```diff
@@ -22,17 +22,17 @@ const client = new QueryClient({
...
-      <ThemeProvider>
-        <Header />
-        <Divider />
-        <RouterProvider router={router} />
-        <Divider />
-        <Footer />
-      </ThemeProvider>
-    </QueryClientProvider>
-  </React.StrictMode>,
+  // <React.StrictMode>
```
vchan, line 25: Can we just delete these?

web/client/src/routes.tsx
```diff
@@ -40,7 +42,18 @@ export const router = createBrowserRouter([
       },
       {
         path: 'docs',
-        element: <Docs />,
+        element: (
+          <Suspense
+            fallback={
+              <Loading className="inline-block ">
```
vchan, line 48:
```suggestion
              <Loading className="inline-block">
```

web/client/src/routes.tsx
```diff
@@ -69,11 +82,22 @@ export const router = createBrowserRouter([
...
-                element: <Docs.Welcome />,
+                element: <DocsWelcome />,
               },
               {
                 path: ':modelName',
-                element: <Docs.Content />,
+                element: (
+                  <Suspense
+                    fallback={
+                      <Loading className="inline-block ">
```
vchan, line 92:
```suggestion
                      <Loading className="inline-block">
```

web/client/src/library/pages/docs/Search.tsx
```diff
@@ -91,57 +93,28 @@ function highlightMatch(source: string, match: string): JSX.Element {
...
-            .map(column => String(Object.values(column).join(' ')))
-            .join(' ')
-        )
+): Array<[ModelSQLMeshModel, string]> {
+  return indices.reduce(
+    (acc: Array<[ModelSQLMeshModel, string]>, [model, index]) => {
+      const idx = index.indexOf(search.toLocaleLowerCase())
+
+      if (idx > -1) {
+        const SIZE = 40
```
vchan, line 104: Should we move this out of the reduce? Maybe use a more descriptive name and make it a global variable?

web/client/src/library/components/graph/help.ts
```diff
@@ -346,30 +313,85 @@ function toNodeOrEdgeId(...args: Array<string | undefined>): string {
...
-  const edges = [
-    connections.ins.map(id => toNodeOrEdgeId('source', id)),
-    connections.outs.map(id => toNodeOrEdgeId('target', id)),
-  ]
-    .flat()
-    .concat([sourceId, targetId])
-
-  if (action === 'remove') {
-    removeActiveEdges(edges)
+// TODO: use better merge
```
vchan, line 316: What's the plan for this function? You mentioned you tested lineage with like 1k models, each with hundreds of columns. Is this a bottleneck with its triply nested for loops?

vchan (approved):

PR #850, opened 2023-05-12
Fix: Use Snowflake's API to write a dataframe into a table

izeigerman:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -263,17 +263,14 @@ def _create_table_from_df(
         replace: bool = False,
         **kwargs: t.Any,
     ) -> None:
-        table = exp.to_table(table_name)
         if not isinstance(df, pd.DataFrame):
             raise ValueError("df must be a pandas DataFrame")
-        expression = next(
```
izeigerman, line 269: This logic was wrong. If he number of rows exceeded the `DEFAULT_BATCH_SIZE` of the adapter we'd write partial data.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -64,3 +65,40 @@ def _get_data_objects(
...
+
+    def _insert_append_pandas_df(
+        self,
+        table_name: TableName,
+        df: pd.DataFrame,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+    ) -> None:
+        from snowflake.connector.pandas_tools import write_pandas
+
+        table = exp.to_table(table_name)
```
tobymao, line 77: this, *db = tuple(reverse(part.name if part.quoted part.name.upper() for part in table.parts))

maybe?

tobymao (approved):

sqlmesh/core/engine_adapter/snowflake.py
izeigerman: nice!

eakmanrq (approved):
Nice catch on the bug in base. Love the more native implementations of handling Dataframes.

PR #851, opened 2023-05-13
Fix: Return no missing intervals for seed models if evaluated at least once

izeigerman:


tobymao (approved):

eakmanrq (commented):
Do we allow hooks on seeds? If we do, wouldn't that then mean that hooks only run once? I thought we tracked intervals for everything, even if they are effectively no-ops, in order to ensure that hooks are run.

tobymao:
we probably don’t need hooks for seeds

izeigerman:
@eakmanrq @tobymao  Even if we do allow hooks for seeds I think the contract is that hooks should run on each **evaluation**. I think it's pretty clear that seeds are evaluated once and only once (unlike full model kind), so there should be no problem with that.

eakmanrq:
@izeigerman What is our reasoning then why views would then be evaluated on each cadence? I remember when we discussed this we decided to that because someone might have hooks on the views that they would want to be executed. I just don't see why this is any different than seeds? I think consistency here would be great.

izeigerman:
@eakmanrq 

> because someone might have hooks on the views that they would want to be executed

That's not the only reason though. View might be dependent on something like `@latest_ds`, so the evaluation won't be the same every time.

eakmanrq:
Fair but someone might also do other things with hooks on views that have a static definition (just like their seeds). It seems non-intuitive to allow that but not allow that on seeds.

tobymao:
fwiw i think views of seeds are different so i'm fine with it

PR #853, opened 2023-05-16
Feat: Store seed content separately from snapshot payload

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -751,31 +762,61 @@ def seed_path(self) -> Path:
...
+            return self
+        return self.copy(
+            update={
+                "seed": Seed(content=""),
+                "is_hydrated": False,
+                "column_hashes_": self.column_hashes,
+            }
+        )
+
+    def hydrated(self, content: str) -> SeedModel:
```
tobymao, line 786: these should be verbs, 

def hyrdate() and def dehydrate

sqlmesh/core/model/definition.py
```diff
@@ -704,6 +701,8 @@ class SeedModel(_Model):
 
     kind: SeedKind
     seed: Seed
+    column_hashes_: t.Optional[t.Dict[str, str]] = Field(default=None, alias="column_hashes")
+    is_hydrated: bool = True
```
tobymao, line 705: does this get serialized? should this be private and only exposed in a property?

sqlmesh/core/model/seed.py
```diff
@@ -24,6 +25,16 @@ class Seed(PydanticModel):
     def columns_to_types(self) -> t.Dict[str, exp.DataType]:
         return columns_to_types_from_df(self._get_df())
 
+    @property
+    def column_hashes(self) -> t.Dict[str, str]:
+        df = self._get_df()
+        return {
+            column_name: hashlib.md5(  # type: ignore
```
tobymao, line 32: let's use crc, it's faster, non cryptographic, and avoids collisions on close text

tobymao (approved):

sqlmesh/core/model/definition.py
izeigerman: Not in this case since we don't mutate the instance, but return a new *hydrated* and *dehydrated* instance instead. So the full name of these methods should've been `create_hydrated` and `create_dehydrated` but I decided to shorten it in the best tradition of our naming strategy.

sqlmesh/core/model/definition.py
izeigerman: It does get serialized, why shouldn't it?

sqlmesh/core/model/definition.py
tobymao: it doesn't have to be a mutation to be a verb, like get_x() is not mutation

sqlmesh/core/model/definition.py
tobymao: i was thinking in the db, but i guess it makes sense over the wire

sqlmesh/core/model/seed.py
izeigerman: md5 has long been considered to be non-cryptographic lol. TBH I don't like our choice of crc even for snapshot fingerprinting. We unnecessarily expose ourselves to the risk of collision. I'll make a change and will wait for one

sqlmesh/core/model/definition.py
izeigerman: Please also note that `dehydrated` is a property and will now become a method.

tobymao: sounds good

PR #858, opened 2023-05-17
Feat: Convert dbt tests into SQLMesh audits

crericha:
Support for model column tests and tests that reference a single node.

### DIFF
--- a/sqlmesh/dbt/basemodel.py
+++ b/sqlmesh/dbt/basemodel.py
@@ -17,51 +17,22 @@ from sqlmesh.dbt.column import (
     column_descriptions_to_sqlmesh,
     column_types_to_sqlmesh,
 )
-from sqlmesh.dbt.common import DbtConfig, GeneralConfig, QuotingConfig, SqlStr
+from sqlmesh.dbt.common import (
+    DbtConfig,
+    Dependencies,
+    GeneralConfig,
+    QuotingConfig,
+    SqlStr,
+)
 from sqlmesh.utils import AttributeDict
 from sqlmesh.utils.conversions import ensure_bool
 from sqlmesh.utils.errors import ConfigError
-from sqlmesh.utils.jinja import MacroReference
-from sqlmesh.utils.pydantic import PydanticModel
 
 if t.TYPE_CHECKING:
     from sqlmesh.dbt.context import DbtContext
 
 
 BMC = t.TypeVar("BMC", bound="BaseModelConfig")
-
-
-class Dependencies(PydanticModel):
-    """
-    DBT dependencies for a model, macro, etc.
-
-    Args:
-        macros: The references to macros
-        sources: The "source_name.table_name" for source tables used
-        refs: The table_name for models used
-    """
-
-    macros: t.Set[MacroReference] = set()
-    sources: t.Set[str] = set()
-    refs: t.Set[str] = set()
-
-    def union(self, other: Dependencies) -> Dependencies:
-        dependencies = Dependencies()
-        dependencies.macros = self.macros | other.macros
-        dependencies.sources = self.sources | other.sources
-        dependencies.refs = self.refs | other.refs
-
-        return dependencies
-
-    def dict(self, *args: t.Any, **kwargs: t.Any) -> t.Dict[str, t.Any]:
-        # See https://github.com/pydantic/pydantic/issues/1090
-        exclude = kwargs.pop("exclude", None) or set()
-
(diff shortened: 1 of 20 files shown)


sqlmesh/dbt/common.py
```diff
@@ -137,3 +144,93 @@ def replace(self, other: T) -> None:
...
+    sources: t.Set[str] = set()
+    refs: t.Set[str] = set()
+
+    def union(self, other: Dependencies) -> Dependencies:
+        dependencies = Dependencies()
+        dependencies.macros = self.macros | other.macros
+        dependencies.sources = self.sources | other.sources
+        dependencies.refs = self.refs | other.refs
+
+        return dependencies
```
vchan, line 169:
```suggestion
        return Dependencies(
            macros=self.macros | other.macros,
            sources=self.sources | other.sources,
            refs=self.refs | other.refs,
        )
```

sqlmesh/dbt/loader.py
```diff
@@ -36,6 +39,14 @@ def sqlmesh_config(project_root: t.Optional[Path] = None, **kwargs: t.Any) -> Co
 
 
 class DbtLoader(Loader):
+    def __init__(self) -> None:
+        self._project: t.Optional[Project] = None
+        super().__init__()
+
+    def load(self, context: Context) -> LoadedProject:
+        self._Project = None
```
vchan, line 47: Should this be `self._project`?

sqlmesh/dbt/basemodel.py
```diff
@@ -253,11 +226,17 @@ def sqlmesh_model_kwargs(self, model_context: DbtContext) -> t.Dict[str, t.Any]:
             if field_val:
                 optional_kwargs[field] = field_val
 
+        audits: t.List[AuditReference] = [(test.name, {}) for test in self.tests]
+        dependencies = self.dependencies
+        for test in self.tests:
+            dependencies = dependencies.union(test.dependencies)
```
izeigerman, line 232: So are we inheriting audit dependencies in the model? I thought we decided not to do this

izeigerman: As far as I remember we decided to extend the Snapshot's `parents` but not model's `depends_on`.

crericha: Correct. That is the next step. In this PR, all column-level tests are implemented by inheriting the dependencies, which matches dbt behavior. In the follow-up PR, this can be decoupled where makes sense.

izeigerman: OK, this is not correct. Let's come up with a follow-up ASAP then.

sqlmesh/dbt/loader.py
```diff
@@ -105,7 +111,31 @@ def _load_models(
...
+        audits: UniqueKeyDict = UniqueKeyDict("audits")
+
+        project = self._load_project()
+        context = project.context
+
+        logger.info("Converting audits to sqlmesh")
+        for package in project.packages.values():
+            context.variables = package.variables
+            for test in package.tests.values():
+                logger.info(f"Converting {test.name} to sqlmesh format")
```
izeigerman, line 123: [Nit] Add quotes for `{test.name}`. Also use `%s` format for logging instead of string interpolation.

sqlmesh/dbt/manifest.py
```diff
@@ -112,17 +125,54 @@ def _load_macros(self) -> None:
...
+                if not dependencies.refs and not dependencies.sources:
+                    raise ConfigError(
+                        f"Audit '{node.name}' is not associated with any models or sources"
+                    )
+
+                if len(dependencies.refs) == 1:
+                    test_owner = list(dependencies.refs)[0]
+                else:
+                    logger.info(
+                        f"Skipping audit '{node.name}'. Multi-owner audits not supported yet."
```
izeigerman, line 153: Ditto. Let's not use string interpolation for logging.

sqlmesh/dbt/manifest.py
```diff
@@ -112,17 +125,54 @@ def _load_macros(self) -> None:
...
+                    )
+                    continue
+
+            if test_owner in dependencies.sources or not test_owner:
+                logger.info(f"Skipping audit '{node.name}'. Source audits not supported yet")
```
izeigerman, line 158: Ditto

sqlmesh/dbt/basemodel.py
crericha: Removed multi-model audits for the time being.

PR #860, opened 2023-05-18
Feat!: Add Databricks Connect Support

eakmanrq:
Breaking Change: Previously we had multiple databricks connection types while now there is a single `databricks` type. I could create aliases for these old names and avoid the breaking change but I think since we are early enough in SQLMesh Databricks adoption that we can just make this breaking change.


sqlmesh/core/config/connection.py
```diff
@@ -148,199 +150,139 @@ def _connection_factory(self) -> t.Callable:
...
-                for k, v in self.spark_config.items():
-                    spark_config.set(k, v)
+        if runtime_env.is_databricks:
+            from pyspark.sql import SparkSession
 
             return dict(
-                spark=SparkSession.builder.config(conf=spark_config)
-                .enableHiveSupport()
-                .getOrCreate()
+                spark=SparkSession.builder.enableHiveSupport().getOrCreate(),
```
izeigerman, line 272: Why are we building a spark session instead of getting an active one (`getActiveSession`)? If we do need to build it why do we no longer support custom spark config?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -366,22 +398,20 @@ def create_view(
...
-
-        if isinstance(query_or_df, DF_TYPES):
-            if PySparkDataFrame is not None and isinstance(query_or_df, PySparkDataFrame):
-                query_or_df = query_or_df.toPandas()
-
-            if not isinstance(query_or_df, pd.DataFrame):
-                raise SQLMeshError("Can only create views with pandas dataframes.")
-
+        df = self.try_get_pandas_df(query_or_df)
+        pyspark_df = self.try_get_pyspark_df(query_or_df)
```
izeigerman, line 402: I wonder whether this logic should even be in the base class. Can this be moved into the spark version of the class?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -366,22 +398,20 @@ def create_view(
...
+        if df is not None or pyspark_df:
```
izeigerman, line 403: [Nit] We use `is not None` for `df` but not for `pyspark_df`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -366,22 +398,20 @@ def create_view(
...
+            if pyspark_df:
+                df = pyspark_df.toPandas()
```
izeigerman, line 405:
```
df = df or pyspark_df.toPandas()
```
maybe

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -1,13 +1,112 @@
...
+        }.issubset(self._extra_config)
+
+    @property
+    def is_spark_session_cursor(self) -> bool:
+        from sqlmesh.engines.spark.db_api.spark_session import SparkSessionCursor
+
+        return isinstance(self.cursor, SparkSessionCursor)
+
+    @property
+    def spark(self) -> PySparkSession:
```
izeigerman, line 76: Can we raise here if `_use_spark_session` is False? I think that would be the best place to raise a clear error message if our expectations about using the spark session hasn't been met. Additionally I think this method is something user can access in a python model.

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -1,13 +1,112 @@
...
+            catalog = self._extra_config.get("catalog")
+            if catalog:
+                self._spark.sql(f"USE CATALOG {catalog}")
+        return self._spark
+
+    def _fetch_native_df(self, query: t.Union[exp.Expression, str]) -> DF:
+        """Fetches a DataFrame that can be either Pandas or PySpark from the cursor"""
+        if self.is_spark_session_cursor:
+            return super()._fetch_native_df(query)
+        if self.can_access_spark_session:
```
izeigerman, line 97: Why do we sometimes use this and sometimes we use `_use_spark_session ` instead? Having both seem to be redundant. For example in which case do I have access to the spark session but don't want to use it?

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -46,7 +45,7 @@ def create_view(
         underlying table without dropping the view first. This is a problem for us since we want to be able to
         swap tables out from under views. Therefore we create the view as non-binding.
         """
-        if isinstance(query_or_df, DF_TYPES):
+        if self.try_get_df(query_or_df) is not None:
```
izeigerman, line 48: Can `is_df` be used here instead?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -40,12 +61,29 @@ def _insert_overwrite_by_condition(
...
+        df = self.try_get_pandas_df(query_or_df)
+        pyspark_df = self.try_get_pyspark_df(query_or_df)
+        if self._use_spark_session and (df is not None or pyspark_df):
+            if df is not None:
+                pyspark_df = self._ensure_pyspark_df(df)
+            assert pyspark_df
+            self._insert_pyspark_df(table_name, pyspark_df, overwrite=True)
         else:
-            super()._insert_overwrite_by_condition(table_name, query_or_df, where, columns_to_types)
+            if df is not None:
```
izeigerman, line 73: Why are we not falling back to parent class here?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -40,12 +61,29 @@ def _insert_overwrite_by_condition(
...
+                query_or_df = next(
```
izeigerman, line 74: What about the batch size?

sqlmesh/core/engine_adapter/base.py
eakmanrq: This is because of a nuance with pandas dataframes. They don't allow you to directly do boolean checks like that. It errors and provides options of types of checks you might want to do. So I do `is not None` just for pandas and then not for PySpark in order to make PySpark consistent with the rest of the codebase.

sqlmesh/core/config/connection.py
eakmanrq: Good point we don't need to build. Will change to `getActiveSession`. 

I removed the config because I'm not even sure if that worked within Databricks since the session already exists. Also I'm sure Databricks has constraints on what type of config you can set. So I decided to remove this and we can revisit if needed.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Base class assumes you can't do an `INSERT OVERWRITE` so it does a transactional delete + insert append.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Without additional complexity, we can't really batch here. We are doing and overwrite so if we batched then we need to make sure that each batch overwrites a complete partition. 

One option would be to remove this and use the parent and do a non-transactional delete/insert append. My thinking though was that if someone had so much data that this is an issue then they should use PySpark DataFrames.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Great idea!

sqlmesh/core/engine_adapter/databricks.py
eakmanrq: So the scenario is that someone has Databricks Connect installed but doesn't want to use it. So they want everything routed through the SQL connector which means converting pandas to SQL. 

At one point I had `disable_databricks_connect` as a property but removed it since I was curious if this was really needed. I see though that without this then this property doesn't make sense. So I think I'm going to add it back in.

sqlmesh/core/engine_adapter/base.py
eakmanrq: So we talked and the downside would be that the try_get would then have a cost in the case that the value was a pyspark df. Therefore you should consider that in implementation. With the current approach these try_gets are free so we decided to stick with the current approach.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yeah I felt mixed about this too. Technically we can do this since we duck type now (don't have to worry about import errors) so this is fairly simple approach but yeah it does mean spark stuff is making it's way into the base. Will change.

sqlmesh/core/engine_adapter/spark.py
tobymao: doesn't the first one do an overwrite and the others append?

sqlmesh/utils/__init__.py
```diff
@@ -160,3 +160,8 @@ def double_escape(s: str) -> str:
 
 def nullsafe_join(join_char: str, *args: t.Optional[str]) -> str:
     return join_char.join(filter(None, args))
+
+
+class classproperty(property):
```
tobymao, line 165: what is this for can you document this?

eakmanrq: It is used for Databrick's `can_access_spark_session` class property. Basically it is a property but at the class level. Will add a docstring.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: The first one might not overwrite all the partitions. If that was the case then you would end up with duplication.

tobymao: maybe let’s chat about this one monday?

tobymao (approved):

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Addressed in this PR: https://github.com/TobikoData/sqlmesh/pull/890

PR #865, opened 2023-05-18
Feat!: Store snapshot intervals separately from snapshot payloads

izeigerman:
Breaking change: Previously intervals were stored as part of the Snapshot record. Every time when we added or removed an interval we had to deal with race conditions when multiple concurrent writers were involved. To address this we used to rely on row-level locks, which are only supported by relational storage backends. 

This PR introduces a new way of storing intervals. Interval records are now stored in a separate table. Every time a new interval is added or an existing one is removed we append a new record to indicate a change instead of modifying / deleting existing ones. Upon reading intervals we merge all records into a single set of intervals for each snapshot. We also compact these records as part of the janitor process to avoid negative impact on the reading performance as a result of having too many rows to consume.

This approach allows us not to worry about synchronizing concurrent writers when updating intervals. This significantly improves experience for users who use storages without transaction support (eg. DeltaLake) to store state. This also improves the overall performance when evaluating snapshots concurrently, since there is no longer a single point of contention.


sqlmesh/core/model/seed.py
```diff
@@ -29,7 +29,7 @@ def columns_to_types(self) -> t.Dict[str, exp.DataType]:
     def column_hashes(self) -> t.Dict[str, str]:
         df = self._get_df()
         return {
-            column_name: str(zlib.crc32(df[column_name].to_json().encode("utf-8")))
+            column_name: str(zlib.crc32(df[column_name].to_json().encode("utf-8")))  # type: ignore
```
tobymao, line 32: this didn't raise for me, why is it for you?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -402,6 +425,109 @@ def _get_environment(
...
+    def remove_interval(
+        self,
+        snapshots: t.Iterable[SnapshotInfoLike],
+        start: TimeLike,
+        end: TimeLike,
+        all_snapshots: t.Optional[t.Iterable[Snapshot]] = None,
+    ) -> None:
+        all_snapshots = all_snapshots or self._get_snapshots_with_same_version(snapshots)
+
+        if logger.isEnabledFor(logging.INFO):
```
tobymao, line 454: why do you need to check this?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -402,6 +425,109 @@ def _get_environment(
...
+            snapshot_ids = ", ".join([str(s.snapshot_id) for s in snapshots])
```
tobymao, line 455: the [] is unnecessary, join takes a generator

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -402,6 +425,109 @@ def _get_environment(
...
+        )
+
+    @transactional()
+    def compact_intervals(self) -> None:
+        query = (
+            exp.select(
+                "id", "name", "identifier", "version", "start_ts", "end_ts", "is_dev", "is_removed"
+            )
+            .from_(self.intervals_table)
+            .join(
```
tobymao, line 471: what's the point of this join? why not just do a where statement on the from?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -402,6 +425,109 @@ def _get_environment(
...
+                exp.select("name", "identifier")
+                .from_(self.intervals_table)
+                .where(exp.column("is_compacted").not_())
+                .distinct(),
+                using=["name", "identifier"],
+            )
+            .order_by(
```
tobymao, line 478: why do you need to order by?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -402,6 +425,109 @@ def _get_environment(
...
+        snapshot_intervals = _rows_to_snapshot_intervals(rows)
+
+        self._push_intervals(snapshot_intervals)
+
+        if interval_ids:
+            self.engine_adapter.delete_from(
+                self.intervals_table, exp.column("id").isin(*interval_ids)
+            )
+
+    def get_snapshot_intervals(
```
tobymao, line 494: maybe this takes in where filter and we can dry up previous method

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -456,7 +592,6 @@ def _migrate_rows(self) -> None:
             model = snapshot.model
             models: t.Dict[str, Model] = {}
             audits: t.Dict[str, Audit] = {}
-            env: t.Dict[str, t.Dict] = {"models": models, "audits": audits}
```
tobymao, line 459: thanks, forgot to delete this

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -548,8 +683,14 @@ def map_data_versions(
...
 
+        delete_filter = self._snapshot_id_filter(snapshot_mapping)
+        self.engine_adapter.delete_from(self.seeds_table, delete_filter)
+        self.engine_adapter.delete_from(self.intervals_table, delete_filter)
         self.delete_snapshots(snapshot_mapping)
-        self._push_snapshots(set(snapshot_mapping.values()), overwrite=True)
+
+        new_snapshots = set(snapshot_mapping.values())
+        self._push_snapshots(new_snapshots, overwrite=True)
+        self._push_intervals(new_snapshots)
```
tobymao, line 693: should this also have orwrite?

sqlmesh/core/model/seed.py
izeigerman: have no idea TBH. Gonna find out

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: To avoid constructing a string that used only for logging purposes if the logging is not even enabled.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -597,14 +738,87 @@ def _transaction(self, transaction_type: TransactionType) -> t.Generator[None, N
...
+def _interval_to_df(
+    snapshot: t.Union[Snapshot, SnapshotIntervals],
+    start_ts: int,
+    end_ts: int,
+    is_dev: bool = False,
+    is_removed: bool = False,
+    is_compacted: bool = False,
+) -> t.Dict[str, t.Any]:
+    return {
+        "id": random_id(),
```
tobymao, line 801: why do we need this?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -597,14 +738,87 @@ def _transaction(self, transaction_type: TransactionType) -> t.Generator[None, N
...
+        "created_ts": now_timestamp(),
+        "name": snapshot.name,
+        "identifier": snapshot.identifier,
+        "version": snapshot.version,
+        "start_ts": start_ts,
+        "end_ts": end_ts,
+        "is_dev": is_dev,
+        "is_removed": is_removed,
+        "is_compacted": is_compacted,
```
tobymao, line 810: what is is_compacted for?

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Oops, this looks like a leftover artifact.

izeigerman: ah, no, still relevant

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Because I don't like the syntax around constructing an IN condition with a tuple.

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: not sure why would we want it here

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: To know which interval records to delete after compaction is done.

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Whether the record was produced as a result of compaction.

sqlmesh/migrations/v0007_create_intervals_table.py
```diff
@@ -0,0 +1,26 @@
...
+            "name": exp.DataType.build("text"),
+            "identifier": exp.DataType.build("text"),
+            "version": exp.DataType.build("text"),
+            "start_ts": exp.DataType.build("bigint"),
+            "end_ts": exp.DataType.build("bigint"),
+            "is_dev": exp.DataType.build("boolean"),
+            "is_removed": exp.DataType.build("boolean"),
+            "is_compacted": exp.DataType.build("boolean"),
+        },
+        primary_key=("id",),
```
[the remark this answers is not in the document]
izeigerman: We absolutely do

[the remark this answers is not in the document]
izeigerman: See this comment: https://github.com/TobikoData/sqlmesh/pull/865#discussion_r1198300602

sqlmesh/core/state_sync/engine_adapter.py
tobymao: meh i think it's ok to just always generate it

sqlmesh/core/model/seed.py
izeigerman: Should've upgraded `pandas-stubs`.

tobymao (approved):

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Refactored the code to reuse the query.

PR #866, opened 2023-05-18
Chore: Update packages

mykalmax:


setup.py
```diff
@@ -108,9 +108,9 @@
             "snowflake-connector-python[pandas]",
         ],
         "web": [
-            "fastapi==0.95.0",
+            "fastapi==0.95.2",
             "watchfiles>=0.19.0",
-            "pyarrow==11.0.0",
+            "pyarrow==12.0.0",
```
vchan, line 113: This conflicts with `snowflake-connector-python`, which wants `pyarrow<10.1.0,>=10.0.1`. Maybe we should change this to ">=10.0.1".

vchan (approved):

PR #867, opened 2023-05-18
Feat: Improve pre and post hook impl

z3z1ma:
## Context

From slack

> Something occurred to me on post hooks.
I love the fact you can have pre hooks by simply having valid SQL statements above the SELECT statement. What is to say you can't have post-hooks be valid SQL statements below the SELECT?
I know, you may immediately think oh then we need some token or way to parse which of the array of statements is the actual model contents; but that is not so.
That's because you can simply enforce there is only 1 SELECT statement per model. Running a SELECT statement in a hook is completely useless. Therefore all non-select statements can be ran with the model and their position above or below dictates pre or post.
What triggered this thought was I have a MERGE statement used to update an activity_next_occurrence column containing the surrogate key of the next occurrence of the same event performed by the same actor. Which essentially reduces the compute required for a self join down the line. I figured it'd be pretty ugly to stuff that in the MODEL's post_hook as some janky string

## Details

Keeping the surface area as light as possible, this should work to support hooks more naturally through the existing interfaces where they are ran, `pre` and `post` respectively. Because we now parse pre/post and put it in the expected model metadata -- we no longer need the other code path that just looped through expressions excluding macro defs and ran them. That is basically a pre hook but not going through the prehook code pathway.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -929,7 +929,7 @@ def load_model(
         statements = prehooks + posthooks
     else:
         # Assume this is a different model type (seed, python, etc.)
-        query = None
+        query = expressions[-1] if isinstance(expressions[-1], d.MacroVar) else None
         prehooks, posthooks = [], []
         statements = expressions[1:-1]
 
@@ -971,9 +971,9 @@ def load_model(
         meta_prehooks = meta_fields.setdefault("pre", [])
         meta_posthooks = meta_fields.setdefault("post", [])
         if not isinstance(meta_prehooks, list):
-            meta_fields["pre"] = meta_prehooks = list(meta_prehooks)
+            meta_fields["pre"] = meta_prehooks = [meta_prehooks]
         if not isinstance(meta_posthooks, list):
-            meta_fields["post"] = meta_posthooks = list(meta_posthooks)
+            meta_fields["post"] = meta_posthooks = [meta_posthooks]
         meta_prehooks.extend([s for s in prehooks if not isinstance(s, d.MacroDef)])
         meta_posthooks.extend([s for s in posthooks if not isinstance(s, d.MacroDef)])
         return create_sql_model(
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -919,7 +919,7 @@ def load_model(
 
     dialect = dialect or ""
     meta = expressions[0]
-    query, pre, post = extract_sql_model_select(expressions)
+    query, pre, post = extract_sql_model_select(expressions[1:])
     statements = pre + post
 
     if not isinstance(meta, d.Model):
@@ -936,8 +936,6 @@ def load_model(
         **{prop.name.lower(): prop.args.get("value") for prop in meta.expressions},
         **kwargs,
     }
-    meta_fields.setdefault("pre", []).extend([s for s in pre if not isinstance(s, d.MacroDef)])
-    meta_fields.setdefault("post", []).extend([s for s in post if not isinstance(s, d.MacroDef)])
 
     name = meta_fields.pop("name", "")
     if not name:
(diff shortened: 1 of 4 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -992,6 +994,32 @@ def load_model(
...
+
+    Raises:
+        ConfigError: If the model definition contains more than one SELECT query.
+    """
+    query_positions = [(e, ix) for ix, e in enumerate(expressions) if isinstance(e, exp.Select)]
+
+    if not query_positions:
+        return None, [], []
+    elif len(query_positions) > 1:
+        raise_config_error("Only one SELECT query is allowed per model")
```
tobymao, line 1017: should specify which model this is for easier debugging

tobymao:
this looks great, can we add a test case?

z3z1ma:
Yeah gotta fix something first, then will add a test or two

sqlmesh/core/model/definition.py
```diff
@@ -919,8 +919,19 @@ def load_model(
...
+    if any(isinstance(e, (exp.Subqueryable, d.Jinja)) for e in expressions):
+        try:
+            # Extract the query and any pre/post statements
+            query, pre, post = extract_sql_model_select(expressions[1:])
+        except Exception as e:
+            raise_config_error(f"Invalid model definition: {e}", path)
+        statements = pre + post
+    else:
+        # Assume this is a different model type (seed, python, etc.)
+        query = expressions[-1] if len(expressions) > 1 else None
```
tobymao, line 932: seed and python models don't have queries, why did you need this block?

sqlmesh/core/model/definition.py
```diff
@@ -956,6 +967,25 @@ def load_model(
             **meta_fields,
         )
     elif query is not None:
+        # Inject preceding and following statements into the model's meta fields.
+        meta_pre: HookCall = meta_fields.setdefault("pre", exp.Tuple())
+        if not pre:
+            pass  # Nothing parsed to inject
+        elif hasattr(meta_pre, "expressions"):
+            meta_pre.expressions.extend([s for s in pre if not isinstance(s, (d.MacroDef))])
```
tobymao, line 975: this section is a bit sus, how was this handeled before and why do you need it now

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -136,9 +136,6 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
                 else:
                     raise SQLMeshError(f"Unexpected SnapshotKind: {snapshot.model.kind}")
 
-        for sql_statement in model.sql_statements:
```
tobymao, line 139: let's delete sql_statements in model definition if it is not used

sqlmesh/core/model/definition.py
tobymao: i feel like this could probably be cleaned up

sqlmesh/core/model/definition.py
z3z1ma: I fallback to the codepath that already existed because tests were getting messed up and it was really hard to figure out. The worst offender was the `test_no_query` or something.

sqlmesh/core/model/definition.py
z3z1ma: I think the issue is that macros are a union type and are already not solidly defined in a single interface

z3z1ma: So the pre/post could already be just an `exp.Anonymous` instead of a `exp.Tuple`/`exp.Array` and I need to wrap it basically in order for the downstream parsing to pick up that it needs to iterate through expressions. At least that was what I observed and what made tests happy

z3z1ma: Needs a real interface, probably. But maybe a way to get by for now thats cleaner

sqlmesh/core/snapshot/evaluator.py
z3z1ma: Ok I'll take a look, will do 👍

sqlmesh/core/model/definition.py
z3z1ma: Let me take a second look rq, I think I panicked when I saw the union type. 😆

z3z1ma: Maybe take a second look now, I adjusted the code to what I think is a more ideal state although it might make a few tests angry.

tobymao:
looks better, let's get this in

PR #868, opened 2023-05-19
Feat!: Add Github Action Bot Deploy Command

eakmanrq:
Breaking Change: The bot now has a behavior change where before if you had a SQLMesh project where no required approvers were defined in your config then a PR would automatically be deployed to prod if tests passed and PR environment was updated. Now the bot skips the prod deploy in this case. To do the deploy you need to issue the `/deploy` command.

### DIFF
--- a/docs/integrations/github.md
+++ b/docs/integrations/github.md
@@ -82,7 +82,7 @@ users:
 Commands can be issued to the bot to have it run specific actions. Below are the currently supported commands:
 * `/deploy` - Deploys the changes in the PR to Production. It will run all the same checks as if you just pushed a commit except for the required approvals check since we assume we can to skip that check if you are manually deploying.
 
-See [Run All Configuration](#run-all-configuration) if you want SQLMesh's commands to have a different prefix to avoid them from clashing with other bots. 
+See [Run All Configuration](#run-all-configuration) for the `--command_namespace` option if you want SQLMesh's commands to have a different prefix to avoid them from clashing with other bots. 
 
 ## Configuration
 ### SQLMesh Configuration
@@ -102,7 +102,7 @@ It has two boolean flags that can be passed in to enable additional functionalit
 * `--delete` - This will delete the PR environment after deploying to production. 
     * Note: If using `--delete` then the runner will need a connection to the engine even if you are using Airflow. This is because the delete is done outside of Airflow.
     * Eventually want the SQLMesh Janitor to automatically do this cleanup which would remove the need for this flag.
-* `--command_namespace` - Namespace to use for SQLMesh commands. For example if you provide `SQLMesh` as a value then commands will be expected in the format of `#SQLMesh/<command>`. The `#` prefix is automatically added.
+* `--command_namespace` - Namespace to use for SQLMesh commands. For example if you provide `#SQLMesh` as a value then commands will be expected in the format of `#SQLMesh/<command>`. 
 
 ### Custom Workflow Configuration
 You can configure each individual action to run as a separate step. This can allow for more complex workflows or integrating specific steps with other actions you want to trigger. Run `sqlmesh_cicd github` to see a list of commands that can be supplied and their potential options.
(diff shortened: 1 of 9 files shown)


docs/integrations/github.md
```diff
@@ -91,6 +102,7 @@ It has two boolean flags that can be passed in to enable additional functionalit
 * `--delete` - This will delete the PR environment after deploying to production. 
     * Note: If using `--delete` then the runner will need a connection to the engine even if you are using Airflow. This is because the delete is done outside of Airflow.
     * Eventually want the SQLMesh Janitor to automatically do this cleanup which would remove the need for this flag.
+* `--command_namespace` - Namespace to use for SQLMesh commands. For example if you provide `SQLMesh` as a value then commands will be expected in the format of `#SQLMesh/<command>`. The `#` prefix is automatically added.
```
izeigerman, line 105: I'm a little confused by the 

> The `#` prefix is automatically added.

and 

> will be expected in the format of `#SQLMesh/<command>`

aren't these statements contradicting?

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -115,6 +115,35 @@ def is_skipped(self) -> bool:
...
 
+class BotCommand(Enum):
+    INVALID = 1
+    DEPLOY_PROD = 2
+
+    @classmethod
+    def from_comment_body(cls, body: str, namespace: t.Optional[str] = None) -> BotCommand:
+        body = body.strip()
+        command_aliases = {
+            cls.DEPLOY_PROD: {"/deploy_prod", "/deploy-prod", "/deployprod", "/deploy"},
```
izeigerman, line 126: Do we need all these aliases? I remember multiple aliases is something we decided against for engine operators in airflow.

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -115,6 +115,35 @@ def is_skipped(self) -> bool:
...
+        }
+        if namespace:
+            command_aliases = {
+                command: {f"#{namespace}{alias}" for alias in alias_values}
```
izeigerman, line 130: So it seems like `#` is not actually **required** and is indeed added automatically

docs/integrations/github.md
eakmanrq: I will just remove the pound sign and let the user set whatever namespace they want to avoid this confusion.

sqlmesh/integrations/github/cicd/controller.py
eakmanrq: This is why I think this is different: Config is something the user generally sets once and that is it. Therefore asking the user to check for specific syntax is fine since it is a low cost operation. Also if they do it incorrectly they will quickly see an error. 

In this case the user is providing an input potentially many times. It is in an interface we don't control and can't provide hints in. Also it takes some time to even know if the input was processed (the actions have to kick off and parse). So in this context I think it is good to be forgiving on syntax.

tobymao: what do you mean we don't control? there's no feedback for unregistered commands? 

fwiw i don't think we need all these aliases, probably just /deploy is good enough

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -139,7 +172,13 @@ def is_review(self) -> bool:
 
     @property
     def is_comment(self) -> bool:
-        return bool(self.payload.get("comment"))
+        comment = self.payload.get("comment")
+        if not comment:
+            return False
+        body = comment.get("body")
+        if not body:
```
tobymao, line 179: doesn't look like you use body here

if not comment.get("body"):

sqlmesh/integrations/github/cicd/controller.py
eakmanrq: I removed the aliases

tobymao (approved):

PR #875, opened 2023-05-20
Fix: Reduce or Fix Databricks Notebook Freezing

eakmanrq:
Sometimes Databricks freezes up while applying a plan. Removing the rich terminal progress updating removed all freezing while testing locally. There could still be some issues remaining but this at least fixes one of the sources of freezing. Unfortunately this makes the notebook UI even less appealing but in the short term stability is certainly worth it over presentation.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -912,6 +912,11 @@ class DatabricksMagicConsole(CaptureTerminalConsole):
     and it makes it difficult to see the progress of the plan.
     """
 
+    def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.evaluation_batch_progress: t.Dict[str, t.Tuple[str, int, int]] = {}
+        self.promotion_status: t.Tuple[int, int] = (0, 0)
+
     def _print(self, value: t.Any, **kwargs: t.Any) -> None:
         super()._print(value, **kwargs)
         for captured_output in self._captured_outputs:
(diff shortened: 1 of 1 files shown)


sqlmesh/core/console.py
```diff
@@ -927,6 +932,52 @@ def _confirm(self, message: str, **kwargs: t.Any) -> bool:
...
+        """Update snapshot progress."""
+        view_name, loaded_batches, total_batches = self.evaluation_batch_progress[snapshot_name]
+        loaded_batches += 1
+        self.evaluation_batch_progress[snapshot_name] = (view_name, loaded_batches, total_batches)
+        finished_loading = loaded_batches == total_batches
+        status = "Loaded" if finished_loading else "Loading"
+        print(f"{status} '{view_name}', Completed Batches: {loaded_batches}/{total_batches}")
+        if finished_loading:
+            total_finished_loading = len(
+                [x for x in self.evaluation_batch_progress.values() if x[1] == x[2]]
```
tobymao, line 954: don't need the brackets, leave it a gen

sqlmesh/core/console.py
```diff
@@ -927,6 +932,52 @@ def _confirm(self, message: str, **kwargs: t.Any) -> bool:
...
+        """Indicates that a new promotion progress has begun."""
+        self.promotion_status = (0, total_tasks)
+        print(f"Virtually Updating '{environment}'")
+
+    def update_promotion_progress(self, num_tasks: int) -> None:
+        """Update promotion progress."""
+        num_promotions, total_promotions = self.promotion_status
+        num_promotions += num_tasks
+        self.promotion_status = (num_promotions, total_promotions)
+        if num_promotions % 5 == 0:
```
tobymao, line 974: if you just did this limiter on rich does it still freeze?

sqlmesh/core/console.py
eakmanrq: Is this an example of what you mean?
```
len(x for x in [(1,2), (1,4), (5,6)])
```

If so I get this:
```
TypeError: object of type 'generator' has no len()
```

tobymao: my bad

tobymao (approved):

PR #879, opened 2023-05-22
Fix: Render model / audit query when computing a snapshot fingerprint when using the native SQLMesh format

izeigerman:


sqlmesh/core/model/meta.py
```diff
@@ -62,12 +63,17 @@ class ModelMeta(PydanticModel):
     column_descriptions_: t.Optional[t.Dict[str, str]]
     audits: t.List[AuditReference] = []
     tags: t.List[str] = []
+    loader_type: str = c.SQLMESH
```
crericha, line 66: If you don't want to store the loader with the model, I have an alternative implementation idea. `Snapshot.from_model()` could take in the loader_type (context is the one that calls it and already has that info) and then `fingerprint_from_model()` could have an optional kwarg like render: bool. Then we don't have to propagate the loader knowledge.

crericha: I haven't made up my mind whether it is a good suggestion yet, but I thought I'd throw it out there.

izeigerman: The downside of this approach is that if the `loader_type` is not inherent to a model every other potential caller of the `fingerprint_from_model()` method will need to be aware of whether the rendering should take place or not. I'd rather make it so that the method always returns the right result for the given model.

sqlmesh/core/model/meta.py
```diff
@@ -62,6 +62,7 @@ class ModelMeta(PydanticModel):
     column_descriptions_: t.Optional[t.Dict[str, str]]
     audits: t.List[AuditReference] = []
     tags: t.List[str] = []
+    hash_raw_query: bool = Field(default=False)
```
tobymao, line 65: why do you need field? why can't you make it bool = False?

tobymao (approved):

sqlmesh/core/model/meta.py
izeigerman: I thought Python 3.8 / 3.9 failures were my fault so I tried to address them. Will remove now

crericha (approved):

PR #880, opened 2023-05-22
Fix: Allow create_external_models to run for projects with star projections

crericha:
Also support external models in dbt formatted projects.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -848,12 +848,13 @@ class Context(BaseContext):
         The schema file contains all columns and types of external models, allowing for more robust
         lineage, validation, and optimizations.
         """
+        models = self._models or self._loader.load(self, update_schemas=False).models
         for path, config in self.configs.items():
             create_schema_file(
                 path=path / c.SCHEMA_YAML,
                 models={
                     name: model
-                    for name, model in self._models.items()
+                    for name, model in models.items()
                     if self.config_for_model(model) is config
                 },
                 adapter=self._engine_adapter,
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -69,7 +69,9 @@ def update_model_schemas(dag: DAG[str], models: UniqueKeyDict[str, Model]) -> No
         if external:
             if "*" in model.columns_to_types:
                 raise ConfigError(
-                    f"Can't expand SELECT * expression for model '{name}'. Projections for models that use external sources must be specified explicitly at '{model._path}'"
+                    f"Can't expand SELECT * expression for model '{name}'."
+                    " Projections for models that use external sources must be specified explicitly"
+                    " or use external models (https://sqlmesh.readthedocs.io/en/stable/concepts/models/external_models)."
                 )
         elif model.mapping_schema:
             try:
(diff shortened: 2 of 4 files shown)


sqlmesh/dbt/loader.py
```diff
@@ -102,6 +102,8 @@ def _load_models(
                 }
             )
 
+        models.update(self._load_external_models())
```
izeigerman, line 105: Shouldn't `_load_external_models` be called in the `load` directly and then we wouldn't need to call it explicitly here?

sqlmesh/core/loader.py
```diff
@@ -97,7 +99,7 @@ def __init__(self) -> None:
         self._path_mtimes: t.Dict[Path, float] = {}
         self._dag: DAG[str] = DAG()
 
-    def load(self, context: Context) -> LoadedProject:
+    def load(self, context: Context, update_schemas: bool = True) -> LoadedProject:
```
izeigerman, line 102: [Nit] Missing docstring for the new argument.

sqlmesh/core/loader.py
```diff
@@ -69,7 +69,9 @@ def update_model_schemas(dag: DAG[str], models: UniqueKeyDict[str, Model]) -> No
         if external:
             if "*" in model.columns_to_types:
                 raise ConfigError(
-                    f"Can't expand SELECT * expression for model '{name}'. Projections for models that use external sources must be specified explicitly at '{model._path}'"
+                    f"Can't expand SELECT * expression for model '{name}'."
+                    " Projections for models that use external sources must be specified explicitly"
```
izeigerman, line 73: Perhaps include which command to run to fix the issue.

sqlmesh/core/context.py
```diff
@@ -848,12 +848,13 @@ def create_external_models(self) -> None:
         The schema file contains all columns and types of external models, allowing for more robust
         lineage, validation, and optimizations.
         """
+        models = self._models or self._loader.load(self, update_schemas=False).models
```
izeigerman, line 851: So it seems like in this case we only need to load models but we'd load everything instead.  

If we made `_load_models` public and kept everything else unchanged we'd only load models and we wouldn't require the `update_schemas` flag.

crericha: I looked at that first too. _load_models requires macros, etc., So it felt unnecessarily complicated to load each bit manually to save time on a one-off command.

sqlmesh/core/loader.py
crericha: I link to the external models docs page, which includes the CLI command. Do you think more is needed?

sqlmesh/dbt/loader.py
crericha: Yeah, I can do it that way. It'd still look like `models.update(super()._load_models(args))`, so I thought the code was simpler by duplicating the one line, but I can use inheritance if you prefer. Let me know.

sqlmesh/core/loader.py
izeigerman: Yeah, I think they should be able to try something immediately without having to read the lengthy doc page. If the quick way doesn't work for them, then they can check out the doc page.

PR #885, opened 2023-05-23
Fix!: Derive the name of a snapshot's physical schema from its model's schema

izeigerman:


sqlmesh/core/snapshot/definition.py
```diff
@@ -716,6 +730,15 @@ def version_get_or_generate(self) -> str:
...
         return self.version or self.fingerprint.to_version()
 
+    @property
+    def physical_schema(self) -> str:
+        if self.physical_schema_ is not None:
+            return self.physical_schema_
+        _, schema, _ = parse_model_name(self.name)
+        if schema is None:
+            schema = c.DEFAULT_SCHEMA
+        return f"{schema}__{c.SQLMESH}__physical_layer"
```
tobymao, line 740: do you really need do you really need physical_layer

izeigerman: trying to make it as distinct as possible. Also indicate its purpose as clearly as possible. Otherwise it looks just like any other environment.

sqlmesh/core/snapshot/definition.py
```diff
@@ -672,6 +685,7 @@ def categorize_as(self, category: SnapshotChangeCategory) -> None:
         )
         if is_forward_only and self.previous_version:
             self.version = self.previous_version.data_version.version
+            self.physical_schema_ = self.previous_version.physical_schema
```
izeigerman, line 688: @tobymao the fact that we didn't inherit the previous physical schema on forward-only / non-breaking was a bug which is now fixed.

sqlmesh/core/snapshot/definition.py
tobymao: i think the SQLMESH prefix is enough, adding physical_layer is so noisy

izeigerman: I was thinking about clustering it by user's schema name instead. I think I need another `_` between the user's schema and `sqlmesh` so that the physical schema always comes out on top.

tobymao (approved):

PR #889, opened 2023-05-24
Fix: use the models' dialects in schema methods

GeorgeSittas:


sqlmesh/core/dialect.py
```diff
@@ -423,7 +423,7 @@ def parse(sql: str, default_dialect: str | None = None) -> t.List[exp.Expression
     for chunk, is_jinja in chunks:
         if is_jinja:
             start, *_, end = chunk
-            segment = sql[start.start : end.end + 1]
+            segment = sql[start.start : end.end + 2]
```
GeorgeSittas, line 426: There was an off-by-one error here, I assume due to this line: https://github.com/tobymao/sqlglot/commit/17be003fef644d44f2b912517e0d2f42c3af818f#diff-9b1dac9e0bc983d20fca22326adf2c9d74deb06cc2c2a486139f332c6b492dc6R905

The error was in the following test: https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_dialect.py#L74

```python
E       AssertionError: assert 'MODEL (\n  n...E y = {{ 1 }}' == 'MODEL (\n  n... y = {{ 1 }};'
E         Skipping 58 identical leading characters in diff, use -v to show
E         - y = {{ 1 }};
E         ?            -
E         + y = {{ 1 }}
```

tobymao (approved):

PR #890, opened 2023-05-24
Fix: Concat Dataframes if insert overwrite

eakmanrq:
A follow up PR will make use of the `SUPPORTS_INSERT_OVERWRITE` property within the engine adapter to cleanup implementation. This PR is just focused on resolving yielding dataframes into an insert/overwrite.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import logging
 import typing as t
 from contextlib import contextmanager
+from functools import reduce
 
 import pandas as pd
 from sqlglot import exp, select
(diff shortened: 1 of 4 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -175,7 +177,18 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
                                 execute(exp.select(existing_limit.expression)).rows[0][0],
                             )
                     return query_or_df.head(limit) if hasattr(query_or_df, "head") else self.adapter._fetch_native_df(query_or_df.limit(limit))  # type: ignore
-
+                elif (
+                    self.adapter.SUPPORTS_INSERT_OVERWRITE
```
tobymao, line 181: just to confirm, engines that don't support insert overwrite are not affected because they do a DELET FROM statement with start and end. wheras support insert overwrite uses dynamic partitioning.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -175,7 +177,18 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
...
+                    and snapshot.is_incremental_by_time_range
+                    and df is not None
+                ):
```
tobymao, line 184: put a comment here to explain why this is needed

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -175,7 +177,18 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
...
+                    for next_df in queries_or_dfs:
```
tobymao, line 185: can we do instead 

reduce(lambda a, b: a.unionAll(b) if self.adapter.pyspark(a) else pd.concat([a, b]) queries_or_dfs

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Yes the DELETE FROM combined with the append on future dataframes is why it works.

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Yeah I initial wrote that but the challenge is that `queries_or_dfs` has already yielded the initial DataFrame so you are just unioning the remaining Dataframes. After that you need to then union again the current dataframe into those unioned dataframes. Therefore I thought the current solution was better since we start with the dataframe we have already yielded.

tobymao: maybe refactor it a bit

if limit and limit > 0:
   return next(...).limit(...)
if  supports insert overwrite:
   apply(reduce(...),)
else:
  for enumerate ...

eakmanrq: This part is the problem: `if supports insert overwrite:`. It requires that we check the initial yielded result if it is a DataFrame since we only want to do the reduce concat logic if it is a DataFrame.

eakmanrq: @tobymao Updated: https://github.com/TobikoData/sqlmesh/pull/890/commits/916e0e11f6630545ec6848e49aab7a1c3fc9556a

tobymao (approved):

PR #893, opened 2023-05-24
Fix: Use SUPPORTS_INSERT_OVERWRITE and cleanup

eakmanrq:
Depends on: https://github.com/TobikoData/sqlmesh/pull/890

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -596,13 +596,38 @@ class EngineAdapter:
         where: t.Optional[exp.Condition] = None,
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
     ) -> None:
-        if where is None:
-            raise SQLMeshError(
-                "Where condition is required when doing a delete/insert for insert/overwrite"
+        table = exp.to_table(table_name)
+        if self.SUPPORTS_INSERT_OVERWRITE:
+            df = self.try_get_pandas_df(query_or_df)
+            if df is not None:
+                query_or_df = next(
+                    pandas_to_sql(
+                        df,
+                        alias=table.alias_or_name,
+                        columns_to_types=columns_to_types,
+                    )
+                )
+            query = t.cast("Query", query_or_df)
+            if where is not None:
+                query = (
+                    exp.select("*")
+                    .from_(
+                        t.cast(exp.Subqueryable, query).subquery("_subquery", copy=False),
+                        copy=False,
+                    )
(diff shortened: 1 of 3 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -596,13 +596,38 @@ def _insert_overwrite_by_condition(
...
+            query = t.cast("Query", query_or_df)
+            if where is not None:
+                query = (
+                    exp.select("*")
+                    .from_(
+                        t.cast(exp.Subqueryable, query).subquery("_subquery", copy=False),
+                        copy=False,
+                    )
+                    .where(where)
+                )
```
eakmanrq, line 619: @tobymao Is this the right way to add a where clause?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -596,13 +596,38 @@ def _insert_overwrite_by_condition(
...
+                    pandas_to_sql(
+                        df,
+                        alias=table.alias_or_name,
+                        columns_to_types=columns_to_types,
+                    )
+                )
+            query = t.cast("Query", query_or_df)
+            if where is not None:
+                query = (
+                    exp.select("*")
```
GeorgeSittas, line 613:
```suggestion
                    exp.select("*", copy=False)
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -596,13 +596,38 @@ def _insert_overwrite_by_condition(
...
+                    .from_(
+                        t.cast(exp.Subqueryable, query).subquery("_subquery", copy=False),
```
GeorgeSittas, line 615: Let's be explicit about setting the `alias` kwarg here:
```suggestion
                        t.cast(exp.Subqueryable, query).subquery(alias="_subquery", copy=False),
```

sqlmesh/core/engine_adapter/base.py
GeorgeSittas: Looks good: this will _append_ the `where` filter and copy it (by default).

GeorgeSittas:
Seems like we're using SQLGlot builders without providing `dialect = self.dialect`. Is this safe?

sqlmesh/core/engine_adapter/base.py
tobymao: the where is just used for deletes, so this is unnecessary, also building the query in this way won't work due to CTEs and other complexity

tobymao (approved):

PR #898, opened 2023-05-25
Chore: add external models to sushi project

tobymao:


tobymao:
@izeigerman this doesn't work yet because hooks don't run on creation

crericha (approved):

izeigerman (approved):
Nice 👍 My other PR addresses this, so we should be good to go here soon.

PR #903, opened 2023-05-25
Chore!: fix mypy issues, use _parse_table_parts instead of _parse_table

GeorgeSittas:
This will resolve mypy issues created due to the latest sqlglot [changes](https://github.com/tobymao/sqlglot/pull/1689/).

When we bump sqlglot to v14.1.0, I'll bump its version here too and the CI will be fixed.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -168,13 +168,12 @@ class EngineAdapter:
         """
         if not self.SUPPORTS_INDEXES:
             return
+
         expression = exp.Create(
             this=exp.Index(
                 this=exp.to_identifier(index_name),
                 table=exp.to_table(table_name),
-                columns=exp.Tuple(
-                    expressions=[exp.to_column(c) for c in columns],
-                ),
+                columns=[exp.to_column(c) for c in columns],
             ),
             kind="INDEX",
             exists=exists,
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
         "requests",
         "rich",
         "ruamel.yaml",
-        "sqlglot~=14.0.0",
+        "sqlglot~=14.1.0",
         "fsspec",
     ],
     extras_require={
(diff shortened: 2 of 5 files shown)


tobymao (approved):

sqlmesh/core/engine_adapter/base.py
```diff
@@ -168,13 +168,12 @@ def create_index(
...
             return
+
         expression = exp.Create(
             this=exp.Index(
                 this=exp.to_identifier(index_name),
                 table=exp.to_table(table_name),
-                columns=exp.Tuple(
-                    expressions=[exp.to_column(c) for c in columns],
-                ),
+                columns=[exp.to_column(c) for c in columns],
```
GeorgeSittas, line 176: @tobymao can you take a look here? Seems like this became invalid due to https://github.com/tobymao/sqlglot/commit/fbf5f472fffa79538d1a58bfd1c1562796bafc4d.

Note that by assigning a list of `Column`s we'll generate a different AST here than what `parse_one` would give:
```python
>>> import sqlglot
>>> sqlglot.parse_one("CREATE INDEX my_idx ON tbl (a, b)")
(CREATE this:
  (INDEX this:
    (IDENTIFIER this: my_idx, quoted: False), table:
    (TABLE this:
      (IDENTIFIER this: tbl, quoted: False)), columns:
    (ORDERED this:
      (COLUMN this:
        (IDENTIFIER this: a, quoted: False)), desc: False, nulls_first: True),
    (ORDERED this:
      (COLUMN this:
        (IDENTIFIER this: b, quoted: False)), desc: False, nulls_first: True)), kind: INDEX)
```

tobymao: yea that was wrong

tobymao: columns is fine

GeorgeSittas: Cool, thanks for verifying. Doing a final pass on the codebase to see if there's any other place where we could have a regression, but it's probably good to go.

tobymao (approved):

PR #905, opened 2023-05-25
Chore[docs]: Restructure macro docs

treysp:
Change structure from single page to separate pages for predefined variables, sqlmesh macros, and jinja macros.

/macros
-overview
-macro_variables
-sqlmesh_macros
-jinja_macros

### DIFF
--- a/docs/concepts/macros/jinja_macros.md
+++ b/docs/concepts/macros/jinja_macros.md
@@ -182,7 +182,7 @@ FROM table
 
 The double quotes around `"'item_id'"` signal to SQLMesh that it is not a column name.
 
-Some SQL dialects interpret double and single quotes different. We could replace single quoted `'item_id'` with double quoted `"item_id"` in the previous example by switching the placement of quotes in the macro function call. Instead of `alias("'item_id'", 'item_id2')` we would use `alias('"item_id"', 'item_id2')`.
+Some SQL dialects interpret double and single quotes differently. We could replace the rendered single quoted `'item_id'` with double quoted `"item_id"` in the previous example by switching the placement of quotes in the macro function call. Instead of `alias("'item_id'", 'item_id2')` we would use `alias('"item_id"', 'item_id2')`.
 
 ## Mixing macro systems
 


docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,123 @@
+# Jinja macros
+
+SQLMesh supports macros from the [Jinja](https://jinja.palletsprojects.com/en/3.1.x/) templating system. 
+
+Jinja's macro approach is pure string substitution. Unlike SQLMesh macros, they assemble SQL query text without building a semantic representation, so users must ensure SQL queries are syntactically valid (e.g., have no misplaced commas or quotes). 
+
+**NOTE:** SQLMesh supports the standard Jinja function library only - it does **not** support dbt-specific jinja functions like `{{ ref() }}`.
```
tobymao, line 7: unless you're using the dbt adapter

tobymao (approved):

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+FROM table
+```
+
+The double quotes around `"'item_id'"` signal to SQLMesh that it is not a column name.
+
+## Mixing macro systems
+
+SQLMesh supports both Jinja and [SQLMesh](./sqlmesh_macros.md) macro systems. We strongly recommend using only one system in a single model - if both are present, they may fail or behave in unintuitive ways. 
+
+[Predefined SQLMesh macro variables](./macro_variables.md) can be used alongside Jinja macros, but they cannot be passed as arguments to a user-defined Jinja macro function. [User-defined SQLMesh macro variables](./sqlmesh_macros.md#user-defined-variables) cannot be used in a model containing Jinja functions.
```
eakmanrq, line 189: It seems like not being able to pass predefined SQLMesh macro variables to Jinja macro functions is a pretty big negative. @tobymao is this something easily fixable?

eakmanrq (approved):
Great improvement!

docs/concepts/macros/jinja_macros.md
tobymao: i think it means you can’t pass @ into jinja, but you can still access all the variables

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+FROM table
+```
+
+The rendered query would be the same as before.
+
+#### if 
+
+if statements allow you to take an action (or not) based on some condition. 
+
+Jinja if statements begin with `{% if ... %}` and ends with `{% endif %}`. The starting `if` statement must contain code that evaluates to `True` or `False`. For example, all of `True`, `1 + 1 == 2`, and `'a' in ['a', 'b']` evaluate to `True`.
```
GeorgeSittas, line 95: Nit:
```suggestion
Jinja if statements begin with `{% if ... %}` and end with `{% endif %}`. The starting `if` statement must contain code that evaluates to `True` or `False`. For example, all of `True`, `1 + 1 == 2`, and `'a' in ['a', 'b']` evaluate to `True`.
```

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+```sql linenums="1"
+SELECT
+    item_id,
+    item_id as item_id2
+FROM table
+```
+
+Note that both argument values are quoted in the call `alias('item_id', 'item_id2')` but are not quoted in the rendered query. During the rendering process, SQLMesh uses its semantic understanding of the query to build the rendered text - it recognizes that the first argument is a column name and that column aliases are unquoted by default.
+
+In that example, the SQL query selects the column `item_id` with the alias `item_id2`. If instead we wanted to select the *string* `'item_id'` with the name `item_id2`, we would pass the `expression` argument with double quotes around it: `"'item_id'"`:
```
GeorgeSittas, line 165: Perhaps it would be interesting to mention that one can also create a quoted _identifier_ by inverting the quotes:

```sql
SELECT {{ alias('"item_id"', 'item_id2') }} FROM bla  -- results in SELECT "item_id" AS item_id2
```

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+
+Also note that a comma is present at the end of the `CASE WHEN` line. Trailing commas are not valid SQL and would normally require special handling, but SQLMesh's semantic understanding of the query allows it to identify and remove the offending comma.
+
+The example renders to this after SQLMesh processing:
+
+```sql linenums="1"
+SELECT
+    CASE WHEN user_vehicle = 'car' THEN 1 ELSE 0 END as vehicle_car,
+    CASE WHEN user_vehicle = 'truck' THEN 1 ELSE 0 END as vehicle_truck,
+    CASE WHEN user_vehicle = 'bus' THEN 1 ELSE 0 END as vehicle_bus
```
GeorgeSittas, line 73: Nit:
```suggestion
    CASE WHEN user_vehicle = 'car' THEN 1 ELSE 0 END AS vehicle_car,
    CASE WHEN user_vehicle = 'truck' THEN 1 ELSE 0 END AS vehicle_truck,
    CASE WHEN user_vehicle = 'bus' THEN 1 ELSE 0 END AS vehicle_bus
```

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+    {{ alias('item_id', 'item_id2')}}
+FROM table
+```
+
+After processing, it would render to this:
+
+```sql linenums="1"
+SELECT
+    item_id,
+    item_id as item_id2
```
GeorgeSittas, line 159: Ditto:
```suggestion
    item_id AS item_id2
```

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+    {{ alias("'item_id'", 'item_id2')}}
+FROM table
+```
+
+After processing, it would render to this:
+
+```sql linenums="1"
+SELECT
+    item_id,
+    'item_id' as item_id2
```
GeorgeSittas, line 179: Ditto:
```suggestion
    'item_id' AS item_id2
```

docs/concepts/macros/jinja_macros.md
```diff
@@ -0,0 +1,189 @@
...
+FROM table
+```
+
+The double quotes around `"'item_id'"` signal to SQLMesh that it is not a column name.
+
+## Mixing macro systems
+
+SQLMesh supports both Jinja and [SQLMesh](./sqlmesh_macros.md) macro systems. We strongly recommend using only one system in a single model - if both are present, they may fail or behave in unintuitive ways.
```
GeorgeSittas, line 187: Nit: I think we need to use "the" here, otherwise I'd expect the sentence to end in _[...] SQLMesh **macros**._
```suggestion
SQLMesh supports both the Jinja and [SQLMesh](./sqlmesh_macros.md) macro systems. We strongly recommend using only one system in a single model - if both are present, they may fail or behave in unintuitive ways. 
```

docs/concepts/macros/macro_variables.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+This example used one of SQLMesh's predefined variables, but you can also define your own macro variables.
+
+We describe SQLMesh's predefined variables below; user-defined macro variables are discussed in the [SQLMesh macros](./sqlmesh_macros.md#user-defined-variables) and [Jinja macros](./jinja_macros.md#user-defined-variables) pages.
+
+## Predefined Variables
+SQLMesh comes with predefined variables that can be used in your queries. They are automatically set by the SQLMesh runtime. 
+
+These variables are related to time and are comprised of a combination of prefixes (start, end, latest) and postfixes (date, ds, ts, epoch, millis).
```
GeorgeSittas, line 40: Can also rephrase as: _[...] and consist of a prefix (...) and a suffix (...)._
```suggestion
These variables are related to time and comprise a combination of prefixes (start, end, latest) and postfixes (date, ds, ts, epoch, millis).
```

docs/concepts/macros/overview.md
```diff
@@ -0,0 +1,13 @@
+# Overview
+
+SQL is a static language. It does not have features like variables or control flow logic (if-then) that allow SQL commands to behave differently in different situations.
```
GeorgeSittas, line 3: I think I'd personally use the word _declarative_ instead of _static_ here, but I don't feel strongly about it. Maybe users will be able to understand the latter easier (?).

GeorgeSittas (approved):
Awesome work Trey! Left some comments

PR #906, opened 2023-05-26
Fix: Support columns for late-binding views in Redshift.

crericha:


sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -54,6 +54,18 @@ def create_view(
             view_name, query_or_df, columns_to_types, replace, no_schema_binding=True
         )
 
+    def columns(self, table_name: TableName) -> t.Dict[str, exp.DataType]:
+        """Fetches column names and types for the target table."""
+        table = exp.to_table(table_name)
+        sql = (
+            exp.select("column_name", "data_type")
+            .from_("SVV_COLUMNS")
+            .where(f"table_name = '{table.alias_or_name}' AND table_schema = '{table.args['db']}'")
```
GeorgeSittas, line 63:
```suggestion
            .where(f"table_name = '{table.alias_or_name}' AND table_schema = '{table.db}'")
```

GeorgeSittas: Actually, there's a difference: what you currently have will implicitly invoke `table.args['db'].sql()` without a dialect because it's interpolated in the string, whereas in my suggestion you just get the raw text of the `db` field. These two are essentially different when `db` is quoted -- not sure if this can happen though.

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -54,6 +54,18 @@ def create_view(
             view_name, query_or_df, columns_to_types, replace, no_schema_binding=True
         )
 
+    def columns(self, table_name: TableName) -> t.Dict[str, exp.DataType]:
+        """Fetches column names and types for the target table."""
```
eakmanrq, line 58: Can you make a comment explaining why we did this? I could see us later wondering why we needed this instead of using the base postgres implementation.

eakmanrq (approved):

sqlmesh/core/engine_adapter/redshift.py
crericha: Yes, it can be quoted. Which is the correct syntax to maintain quoting?

crericha: I guess I don't want to maintain quoting because in the where clause the quotes are added. So raw string is better?

GeorgeSittas: Sounds reasonable

GeorgeSittas: (FWIW if you wanted to preserve the quotes you could just do `table.args['db'].sql(dialect=self.dialect)`)

tobymao (approved):

PR #907, opened 2023-05-26
Feat: table-diff, easily diff two tables.

tobymao:


tobymao:
@z3z1ma

sqlmesh/core/context.py
```diff
@@ -374,9 +387,76 @@ def run(
...
-        """Returns a model with the given name or None if a model with such name doesn't exist."""
-        return self._models.get(name)
+    @t.overload
+    def get_model(
+        self, model_or_snapshot: ModelOrSnapshot, raise_if_missing: Literal[True] = True
+    ) -> Model:
+        ...
+
+    @t.overload
+    def get_model(
```
izeigerman, line 397: Why do we need these overloads?

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+class SummaryDiff(PydanticModel, frozen=True):
+    """Summary statistics and a sample dataframe."""
+
+    source: str
+    target: str
+    stats: t.Dict[str, float]
+    sample: pd.DataFrame
+
+    @property
+    def s_count(self) -> int:
```
izeigerman, line 56: Let's use full names please

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+from sqlglot.dialects.dialect import DialectType
+
+from sqlmesh.utils.pydantic import PydanticModel
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core._typing import TableName
+    from sqlmesh.core.engine_adapter import EngineAdapter
+
+
+class SchemaDiff(PydanticModel, frozen=True):
```
izeigerman, line 16: Please note that we already have a concept of a schema diff. Perhaps worth disambiguating

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def schema_diff(self) -> SchemaDiff:
+        return SchemaDiff(
+            source=self.source,
+            target=self.target,
+            source_schema=self.source_schema,
+            target_schema=self.target_schema,
+        )
+
+    def summary_diff(self) -> SummaryDiff:
```
izeigerman, line 125: Can we break this function down more?

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+                for c, t in self.source_schema.items()
+                if t == self.target_schema.get(c)
+            ]
+
+            query = (
+                exp.select(
+                    *s_selects.values(),
+                    *t_selects.values(),
+                    exp.func(
+                        "IF", exp.and_(*(c.not_().is_(exp.Null()) for c in s_index)), 1, 0
```
izeigerman, line 151: Isn't it enough to have at least one non-null key to prove existence?

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+    stats: t.Dict[str, float]
+    sample: pd.DataFrame
+
+    @property
+    def s_count(self) -> int:
+        """Count of the source."""
+        return int(self.stats["s_count"])
+
+    @property
+    def t_count(self) -> int:
```
izeigerman, line 61: Do we need extraction methods for `*_matches` interface too? I don't think that accessing the `stats` dict directly is a good interface.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -778,6 +778,28 @@ def execute(
...
+
+        Args:
+            query_or_df: The query or df to create a temp table for.
+            name: The base name of the temp table.
+
+        Yields:
+            The table expression
+        """
+
+        table = self._get_temp_table(name)
```
izeigerman, line 795: Should this be wrapped into a transaction?

sqlmesh/core/table_diff.py
```diff
@@ -0,0 +1,195 @@
...
+            t_index = []
+
+            for col in self.on.find_all(exp.Column):
+                col.not_().is_(exp.Null())
+                if col.table == "s":
+                    s_index.append(col)
+                elif col.table == "t":
+                    t_index.append(col)
+
+            comparisons = [
```
izeigerman, line 140: It doesn't look like we report this anywhere.

sqlmesh/core/table_diff.py
tobymao: no, you can have a null in a key

sqlmesh/core/table_diff.py
tobymao: although there's only a SchemaDiffer not a schema diff, what do you want to rename it to?

sqlmesh/core/table_diff.py
tobymao: don't have a use case for it quite yet but we can add those

sqlmesh/core/table_diff.py
tobymao: it's not reported, it's used for filtering

PR #910, opened 2023-05-26
Feat!: Use pre / post SQL statements instead of hooks

izeigerman:
With this update with move away from using pre- / post- hooks in favor to a much more SQL native approach: plan SQL statements. 

Instead of having a dedicated metadata attribute, users can now express their pre- / post- evaluation logic as part of the model body:

```sql
MODEL (
  ...
);

-- This block is evaluated before the model's query
CREATE TABLE IF NOT EXISTS temp_table (a INT);

ADD JAR s3://special_udf.jar;
CREATE TEMPORARY FUNCTION UDF AS 'my.jar.udf';

-- Model's query
SELECT ...


-- This block is evaluated after the model's query
DROP TABLE temp_table;
```


sqlmesh/core/engine_adapter/base.py
```diff
@@ -766,17 +766,20 @@ def supports_transactions(self, transaction_type: TransactionType) -> bool:
...
     ) -> None:
         """Execute a sql query."""
         to_sql_kwargs = (
             {"unsupported_level": ErrorLevel.IGNORE} if ignore_unsupported_errors else {}
         )
-        sql = self._to_sql(sql, **to_sql_kwargs) if isinstance(sql, exp.Expression) else sql
-        logger.debug(f"Executing SQL:\n{sql}")
-        self.cursor.execute(sql, **kwargs)
+
+        expressions_or_strs = expressions if isinstance(expressions, list) else [expressions]
```
GeorgeSittas, line 778: Nit: we can leverage sqlglot's `ensure_list` here

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -393,14 +393,17 @@ def execute(
...
-        sql = self._to_sql(sql, **to_sql_kwargs) if isinstance(sql, exp.Expression) else sql
-        logger.debug(f"Executing SQL:\n{sql}")
-        retry.retry_target(
-            target=functools.partial(self._retryable_execute, sql=sql),
-            predicate=_ErrorCounter(self._extra_config["job_retries"]).should_retry,
-            sleep_generator=retry.exponential_sleep_generator(initial=1.0, maximum=3.0),
-            deadline=self._extra_config.get("job_retry_deadline_seconds"),
-        )
+
+        expressions_or_strs = expressions if isinstance(expressions, list) else [expressions]
```
GeorgeSittas, line 397: Ditto

sqlmesh/core/model/definition.py
```diff
@@ -967,33 +1021,24 @@ def load_model(
...
-            path=path,
-            time_column_format=time_column_format,
-            **meta_fields,
-        )
-    elif query is not None:
+    if query is not None:
+        macro_references = [
+            extract_macro_references(query.sql()),
+            *[extract_macro_references(e.sql()) for e in pre_statements],
+            *[extract_macro_references(e.sql()) for e in post_statements],
```
GeorgeSittas, line 1028: Should we provide a dialect here?

sqlmesh/core/model/definition.py
```diff
@@ -1270,19 +1351,11 @@ def _capture_expression_macros(expression: exp.Expression) -> None:
                     name = macro_func.this.name.lower()
                     used_macros[name] = macros[name]
 
-    _capture_expression_macros(query)
+    if not isinstance(expressions, list):
+        expressions = [expressions]
```
GeorgeSittas, line 1355: Ditto

izeigerman:
CC: @z3z1ma

sqlmesh/core/model/definition.py
izeigerman: shouldn't matter I think

tobymao (approved):
incredible pr very nice

z3z1ma:
I can get some credit (maybe 1 of 100) for the idea but dang nice PR. I was just trying to shove those statements into the existing constructs. This implementation rocks. 🤩

PR #913, opened 2023-05-26
Fix: use dialect in column_to_types_ validator

### LINKED ISSUE #911: Model explicit columns not respecting dialect
https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/meta.py#L202

needs to get passed the model's dialect or the default config dialect

GeorgeSittas:
Fixes #911

### DIFF
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -194,12 +194,14 @@ class ModelMeta(PydanticModel):
         return cron
 
     @validator("columns_to_types_", pre=True)
-    def _columns_validator(cls, v: t.Any) -> t.Optional[t.Dict[str, exp.DataType]]:
+    def _columns_validator(
+        cls, v: t.Any, values: t.Dict[str, t.Any]
+    ) -> t.Optional[t.Dict[str, exp.DataType]]:
         if isinstance(v, exp.Schema):
             return {column.name: column.args["kind"] for column in v.expressions}
         if isinstance(v, dict):
             return {
-                k: maybe_parse(data_type, into=exp.DataType)  # type: ignore
+                k: maybe_parse(data_type, into=exp.DataType, dialect=values["dialect"])
                 for k, data_type in v.items()
             }
         return v
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -6,7 +6,7 @@ from enum import Enum
 
 from croniter import croniter
 from pydantic import Field, root_validator, validator
-from sqlglot import exp, maybe_parse
+from sqlglot import exp
 
 from sqlmesh.core import dialect as d
 from sqlmesh.core.model.kind import (
@@ -201,7 +201,7 @@ class ModelMeta(PydanticModel):
             return {column.name: column.args["kind"] for column in v.expressions}
         if isinstance(v, dict):
             return {
-                k: maybe_parse(data_type, into=exp.DataType, dialect=values["dialect"])
+                k: exp.DataType.build(data_type, dialect=values["dialect"])
                 for k, data_type in v.items()
             }
         return v


sqlmesh/core/model/meta.py
```diff
@@ -194,12 +194,14 @@ def _cron_validator(cls, v: t.Any) -> t.Optional[str]:
...
-    def _columns_validator(cls, v: t.Any) -> t.Optional[t.Dict[str, exp.DataType]]:
+    def _columns_validator(
+        cls, v: t.Any, values: t.Dict[str, t.Any]
+    ) -> t.Optional[t.Dict[str, exp.DataType]]:
         if isinstance(v, exp.Schema):
             return {column.name: column.args["kind"] for column in v.expressions}
         if isinstance(v, dict):
             return {
-                k: maybe_parse(data_type, into=exp.DataType)  # type: ignore
+                k: maybe_parse(data_type, into=exp.DataType, dialect=values["dialect"])
```
crericha, line 204: Is dialect guaranteed to be in values? It isn't a required field.

crericha: Might need values.get("dialect", "")

GeorgeSittas: Thanks for catching this, that's a good point. I think `.get("dialect")` will suffice in this case. Making the fix.

tobymao: can you just do exp.DataType.build(data_type, dialect=...)

GeorgeSittas: Actually I see that we have `dialect: str = ""` in `ModelMeta`'s attribute list, so not sure if the lookup will fail.

crericha: Because pre=True is used in the validator, my guess is dialect is not included. You can test though and see what happens when dialect is omitted from the constructor.

tobymao: can you make sure that dialect is passed in with either the config default dialect or model dialect?

GeorgeSittas: I tried the following example to test whether it'd fail or not:
```python
>>> from sqlmesh.utils.pydantic import PydanticModel
>>> from pydantic import validator
>>>
>>> class Test(PydanticModel):
...     foo: str = ""
...     bar: str = ""
...     @validator("foo", pre=True)
...     def _foo_validator(cls, v):
...             return str(v)
...     @validator("bar", pre=True)
...     def _bar_validator(cls, v, values):
...         return values["foo"]
...
>>> t = Test()
>>> t
Test(foo='', bar='')
>>> t = Test(bar=None)  # This would fail if there was a problem, because validator looks up values["foo"] 
>>> t
Test(foo='', bar='')
```

GeorgeSittas: > can you just do exp.DataType.build(data_type, dialect=...)

> can you make sure that dialect is passed in with either the config default dialect or model dialect?

Will do both.

GeorgeSittas: @tobymao verified that the correct dialect is used in all cases, e.g. it's `""` when no dialect is specified, it's equal to the default config dialect if the model doesn't specify its dialect and it's equal to the model's dialect when it specifies one.

crericha (approved):

sqlmesh/core/model/meta.py
GeorgeSittas: Leaving this here for reference too: seems like `values` is a dict that contains every property and its value (e.g. initial value or validated one). I used the `sqlmesh init` dummy project to test this out:

Code used for `columns_to_types_` validator:

```python
    @validator("columns_to_types_", pre=True)
    def _columns_validator(
        cls, v: t.Any, values: t.Dict[str, t.Any]
    ) -> t.Optional[t.Dict[str, exp.DataType]]:
        if isinstance(v, exp.Schema):
            return {column.name: column.args["kind"] for column in v.expressions}
        if isinstance(v, dict):
            breakpoint()
            return {
                k: exp.DataType.build(data_type, dialect=values["dialect"])
                for k, data_type in v.items()
            }
        return v
```

In the run that's shown below, no dialect was specified (either in the config defaults or in any model in the project)
```python
$ sqlmesh plan dev
> sqlmesh/core/model/meta.py(204)_columns_validator()
-> return {
(Pdb) values
{'name': 'external_db.external_table', 'kind': ModelKind(name=<ModelKindName.EXTERNAL: 'EXTERNAL'>), 'dialect': '', 'cron': '@daily', 'owner': None, 'description': 'An external table', 'stamp': None, 'start': None, 'retention': None, 'storage_format': None, 'partitioned_by_': [], 'pre': [], 'post': [], 'depends_on_': None}
```

tobymao (approved):

PR #915, opened 2023-05-26
Feat: display node for CTE

mykalmax:
When requesting column lineage display CTE nodes + source in case model has CTEs

Screenshot 2023-05-30 at 12 14 19 PM

Screenshot 2023-05-30 at 12 13 47 PM

Screenshot 2023-05-30 at 12 15 43 PM

### DIFF
--- a/tests/web/test_main.py
+++ b/tests/web/test_main.py
@@ -484,32 +484,20 @@ def test_get_environments(project_context: Context) -> None:
 
 def test_get_lineage(web_sushi_context: Context) -> None:
     response = client.get("/api/lineage/sushi.waiters/ds")
-    assert response.status_code == 200
 
+    assert response.status_code == 200
     assert response.json() == {
         "sushi.waiters": {
             "ds": {
-                "source": """SELECT DISTINCT
-  <b>CAST(o.ds AS TEXT) AS ds</b>
-FROM (
-  SELECT
-    CAST(NULL AS INT) AS id,
-    CAST(NULL AS INT) AS customer_id,
-    CAST(NULL AS INT) AS waiter_id,
-    CAST(NULL AS INT) AS start_ts,
-    CAST(NULL AS INT) AS end_ts,
-    CAST(NULL AS TEXT) AS ds
-  FROM (VALUES
-    (1)) AS t(dummy)
-) AS o /* source: sushi.orders */
-WHERE
-  o.ds <= '1970-01-01' AND o.ds >= '1970-01-01'""",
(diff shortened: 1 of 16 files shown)


web/server/api/endpoints/lineage.py
```diff
@@ -66,7 +56,6 @@ async def column_lineage(
             sources={
                 model: context.models[model].render_query()
                 for model in context.dag.upstream(model_name)
-                if model in context.models
```
vchan, line 69: This was added to deal with external tables that will appear in the upstream DAG but are not models so their queries cannot be rendered. Is it causing an issue?

web/server/api/endpoints/lineage.py
```diff
@@ -81,19 +70,16 @@ async def column_lineage(
 
     for i, node in enumerate(node.walk()):
         if i > 0:
-            table = _get_table(node) or model_name
+            table = _get_table(node) or node.name
             column_name = exp.to_column(node.name).name
         if column_name in cache_column_names:
             column_name = cache_column_names[column_name]
```
vchan, line 76: Should `cache_column_names` be removed entirely?

tests/web/test_main.py
```diff
@@ -484,32 +484,20 @@ def test_get_environments(project_context: Context) -> None:
...
-    CAST(NULL AS INT) AS waiter_id,
-    CAST(NULL AS INT) AS start_ts,
-    CAST(NULL AS INT) AS end_ts,
-    CAST(NULL AS TEXT) AS ds
-  FROM (VALUES
-    (1)) AS t(dummy)
-) AS o /* source: sushi.orders */
-WHERE
-  o.ds <= '1970-01-01' AND o.ds >= '1970-01-01'""",
+                "source": "SELECT DISTINCT\n  CAST(o.ds AS TEXT) AS ds\nFROM (\n  SELECT\n    CAST(NULL AS INT) AS id,\n    CAST(NULL AS INT) AS customer_id,\n    CAST(NULL AS INT) AS waiter_id,\n    CAST(NULL AS INT) AS start_ts,\n    CAST(NULL AS INT) AS end_ts,\n    CAST(NULL AS TEXT) AS ds\n  FROM (VALUES\n    (1)) AS t(dummy)\n) AS o /* source: sushi.orders */\nWHERE\n  o.ds <= '1970-01-01' AND o.ds >= '1970-01-01'",
```
vchan, line 492: Is it possible to leave the expected source as a triple quoted string like it was before and remove the <b> tags? It was way more readable.

web/client/src/library/components/graph/Graph.tsx
```diff
@@ -444,6 +545,10 @@ const ModelColumns = memo(function ModelColumns({
     function updateColumnLineage(
       columnLineage: Record<string, Record<string, LineageColumn>> = {},
     ): void {
+      console.log()
```
vchan, line 548: Debug statement?

web/server/api/endpoints/lineage.py
mykalmax: If something is not a model, it should default to be a CTE node 

Screenshot 2023-05-30 at 2 28 39 PM

mykalmax: oh, i see, yeah need to bring it back,

vchan: A CTE wouldn't show up in in a model's upstream DAG though. This only handles external models.

vchan (approved):

PR #917, opened 2023-05-28
Feat!: Add depends_on_past through self-reference check support (Built-in only)

eakmanrq:
Breaking Change: Bugfix is included in this PR that correctly includes `batch_size` in the model metadata fingerprint for seed models. 

This adds support for the built-in scheduler but without proper restatement support.

TODO:
* Add restatement support for all `depends_on_past` models (so that means incremental by unique key also)
  * PR: https://github.com/TobikoData/sqlmesh/pull/921
* Add Airflow support

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -454,8 +454,8 @@ class _Model(ModelMeta, frozen=True):
     def has_self_reference_query(self) -> bool:
         if self._has_self_reference_query is None:
             self._has_self_reference_query = (
-                self.name in _find_tables([self.render_query()])
-                or self.kind.is_incremental_by_unique_key
+                self.kind.is_incremental_by_unique_key
+                or self.name in _find_tables([self.render_query()])
             )
         return self._has_self_reference_query
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -406,7 +406,7 @@ class _Model(ModelMeta, frozen=True):
             return self.depends_on_ - {self.name}
 
         if self._depends_on is None:
-            self._depends_on = _find_tables(self._render_all_sql()) - {self.name}
+            self._depends_on = self._all_model_references - {self.name}
         return self._depends_on
 
     @property
@@ -448,6 +448,14 @@ class _Model(ModelMeta, frozen=True):
     @property
     def is_seed(self) -> bool:
         return False
+
+    @property
+    def has_self_reference(self) -> bool:
+        return self.name in self._all_model_references
+
+    @property
+    def _all_model_references(self) -> t.Set[str]:
+        return _find_tables(self._render_all_sql())
 
     def validate_definition(self) -> None:
         """Validates the model's definition.
@@ -816,6 +824,10 @@ class SeedModel(_Model):
         if not seed_path.is_absolute():
             return self._path.parent / seed_path
         return seed_path
+
+    @property
+    def has_self_reference(self) -> bool:
+        return False
 
     def to_dehydrated(self) -> SeedModel:
         """Creates a dehydrated copy of this model.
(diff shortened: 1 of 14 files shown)


tests/core/test_snapshot.py
```diff
@@ -399,7 +399,7 @@ def test_fingerprint_seed_model():
 
     expected_fingerprint = SnapshotFingerprint(
         data_hash="3834815287",
-        metadata_hash="1120323454",
+        metadata_hash="1624292983",
```
eakmanrq, line 402: This is due to correctly including the batch_size now in the metadata hash for seeds.

sqlmesh/core/model/definition.py
```diff
@@ -445,6 +445,14 @@ def is_python(self) -> bool:
     def is_seed(self) -> bool:
         return False
 
+    @property
+    def has_self_reference(self) -> bool:
+        return self.name in self._all_model_references
+
+    @property
+    def _all_model_references(self) -> t.Set[str]:
+        return _find_tables(self.render_query())
```
tobymao, line 454: this is no longer cached can you remove the _depends_on cache and make a new one for this?

sqlmesh/core/model/kind.py
```diff
@@ -12,6 +13,11 @@
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import PydanticModel
 
+if sys.version_info >= (3, 9):
```
tobymao, line 16: what is this for?

sqlmesh/core/model/kind.py
```diff
@@ -59,7 +65,12 @@ def is_materialized(self) -> bool:
     @property
     def only_latest(self) -> bool:
         """Whether or not this model only cares about latest date to render."""
-        return self.model_kind_name in (ModelKindName.VIEW, ModelKindName.FULL)
+        return self.is_view or self.is_full
+
+    @property
+    def is_time_based_load(self) -> bool:
```
tobymao, line 71: def time_based

tobymao: or has_time_column ?

sqlmesh/core/model/kind.py
```diff
@@ -81,18 +92,80 @@ def model_kind_name(self) -> ModelKindName:
 class ModelKind(PydanticModel, ModelKindMixin):
     name: ModelKindName
 
-    def to_expression(self, **kwargs: t.Any) -> d.ModelKind:
-        return d.ModelKind(this=self.name.value.upper(), **kwargs)
+    @classmethod
+    def field_validator(cls) -> classmethod:
```
tobymao, line 96: why was this change needed?

eakmanrq: Initially I was adding more field validators for the new FULL_WITH_HISTORY and it was clumsy adding them to the bottom of the file when they might reference a class above it. If that happened then you had to put the definition before that class. This way we couple the validators with the thing they are validating so it is easy to get the appropriate validator for the given field you are referencing. So the need is reduced but I think it is a better pattern.

sqlmesh/core/snapshot/definition.py
```diff
@@ -804,6 +804,11 @@ def start(self) -> t.Optional[datetime]:
     def model_kind_name(self) -> ModelKindName:
         return self.model.kind.name
 
+    @property
+    def depends_on_past(self) -> bool:
+        """Whether or not this models depends on past intervals to be accurate before loading following intervals."""
```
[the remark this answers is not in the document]
eakmanrq: Sure either way works for me. Updated.

tests/core/test_integration.py
```diff
@@ -644,6 +649,24 @@ def test_multi(mocker):
     validate_apply_basics(context, c.PROD, plan.snapshots)
 
 
+@pytest.mark.integration
+@pytest.mark.core_integration
+def test_full_with_history(sushi_context: Context):
```
tobymao, line 654: rename the test

tobymao (approved):

PR #918, opened 2023-05-28
Fix: wrap router in suspense

mykalmax:


web/client/src/routes.tsx
```diff
@@ -5,14 +5,10 @@ import Loading from '@components/loading/Loading'
...
 
-const Editor = lazy(async () => await import('./library/pages/editor/Editor'))
-const Docs = lazy(async () => await import('./library/pages/docs/Docs'))
-const DocsContent = lazy(
-  async () => await import('./library/pages/docs/Content'),
-)
-const DocsWelcome = lazy(
-  async () => await import('./library/pages/docs/Welcome'),
-)
+const Editor = lazy(() => import('./library/pages/editor/Editor'))
```
vchan, line 8: Curious why the async was removed.

vchan (approved):

web/client/src/routes.tsx
mykalmax: we had TS lint rule that forces to await all async functions. lazy already expects a Promise-like object and import returns promise, so no need to await for import because lazy should do it anyway. (https://react.dev/reference/react/lazy#reference)

PR #920, opened 2023-05-29
Include a model's CTEs when testing one of its CTEs

### LINKED ISSUE #916: Testing CTEs which reference other CTEs is not supported
When testing individual CTEs we should not only evaluate the target CTE but also CTEs it depends on.

vchan:
Fixes #916

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -18,6 +18,7 @@ class TestError(SQLMeshError):
 
 
 class ModelTest(unittest.TestCase):
+    __test__ = False
     view_names: list[str] = []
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -18,6 +18,7 @@ class TestError(SQLMeshError):
 
 
 class ModelTest(unittest.TestCase):
+    __test__ = False
```
vchan, line 21: Pytest discovery was picking up SqlModelTest as a test case below and breaking our unit tests.

tobymao (approved):

PR #921, opened 2023-05-29
Feat: Add Restatement Support for depends_on_past (Built-in only)

eakmanrq:
Depends on: https://github.com/TobikoData/sqlmesh/pull/917

TODO:
- Add Airflow support
- Update docs to document behavior


sqlmesh/core/model/definition.py
```diff
@@ -761,6 +769,10 @@ def seed_path(self) -> Path:
             return self._path.parent / seed_path
         return seed_path
 
+    @property
+    def has_self_reference(self) -> bool:
```
izeigerman, line 773: Do we want the same override for `ExternalModel`?

sqlmesh/core/model/kind.py
```diff
@@ -59,7 +65,12 @@ def is_materialized(self) -> bool:
     @property
     def only_latest(self) -> bool:
         """Whether or not this model only cares about latest date to render."""
-        return self.model_kind_name in (ModelKindName.VIEW, ModelKindName.FULL)
+        return self.is_view or self.is_full
+
+    @property
+    def is_time_based_load(self) -> bool:
+        """Whether or not this model has a time column which influences the load pattern used."""
+        return self.is_incremental_by_time_range
```
izeigerman, line 73: Do we expect other combinations of flags to fall into the category of "is_time_based_load"? If not, do we need this?

tobymao: i think this is removed already

sqlmesh/core/model/kind.py
```diff
@@ -81,18 +92,80 @@ def model_kind_name(self) -> ModelKindName:
...
+                return klass(**v)
+
+            name = (v.name if isinstance(v, exp.Expression) else str(v)).upper()
+
+            try:
+                return ModelKind(name=ModelKindName(name))
+            except ValueError:
+                raise ConfigError(f"Invalid model kind '{name}'")
+
+        return validator("kind", pre=True, allow_reuse=True)(_model_kind_validator)
```
izeigerman, line 133: Why are we returning the `validator` itself? Can't we just do something like:
```
_kind_validator = validator("kind", pre=True, allow_reuse=True)(_model_kind_validator)
```

izeigerman: We already follow a similar pattern here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/common.py#L47

izeigerman: Ah, I see now that you moved away from it. Why?

sqlmesh/core/snapshot/definition.py
```diff
@@ -614,14 +626,19 @@ def missing_intervals(
...
         Returns:
             A list of all the missing intervals as epoch timestamps.
         """
+        restatements = restatements or set()
         if self.is_symbolic or (self.is_seed and self.intervals):
             return []
 
         latest = make_inclusive_end(latest or now())
+        end = (
+            make_inclusive_end(now()) if self.depends_on_past and self.name in restatements else end
```
izeigerman, line 640: should this be derived from `latest`? or otherwise should `latest` and `end` share the same value of `now()`?

izeigerman: I think the fact that `end` can be greater than `latest` may lead to the kind of edge cases I don't even want to think about lol

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -782,7 +782,11 @@ def _intervals_to_df(
         [
             _interval_to_df(
                 snapshot,
-                *snapshot.inclusive_exclusive(start, end),
+                *(
+                    snapshot.get_remove_interval(start, end)
+                    if is_removed
+                    else snapshot.inclusive_exclusive(start, end)
```
izeigerman, line 788: Can we somehow name this or `get_remove_interval` to highlight their relation? Eg. `get_add_interval` / `get_remove_interval`. Or `inclusive_exclusive` and `inclusive_exclusive_for_remove`. Or maybe 
```
def inclusive_exclusive(start, end, for_remove: bool = False)
```
Do we even need a separate method btw?

sqlmesh/core/model/kind.py
izeigerman: what do you mean? Am I looking at the wrong PR?

eakmanrq: This PR is based off of my other PR. So some fixes are in my other one that haven't been merged into this one yet. Will be updating soon.

sqlmesh/core/model/definition.py
eakmanrq: It is not needed. Seed needed it because of an error I got when it does a columns_to_type thing.

sqlmesh/core/model/kind.py
eakmanrq: Well I kept the pattern in that we are still sharing these validators. I just now have them inside the class that they are used to validate again so it is easier to reference the validator you need. At one point I was adding more validators and this felt like a cleaner way to organize them.

sqlmesh/core/snapshot/definition.py
eakmanrq: I think that is fine. Honestly I have a hard time understanding the use case for latest. Does this make sense to you @tobymao?

sqlmesh/core/model/kind.py
izeigerman: That's ok. Do we need to create a new `validator` every time, though, or can can it just be a class member?

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: I went with the `for_remove` property. Let me know what you think: https://github.com/TobikoData/sqlmesh/pull/921/commits/3cb6541f9044bdf45e6430295f5c365ec6544572

sqlmesh/core/model/kind.py
```diff
@@ -12,6 +13,11 @@
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import PydanticModel
 
+if sys.version_info >= (3, 9):
```
tobymao, line 16: delete

sqlmesh/core/snapshot/definition.py
```diff
@@ -551,23 +551,25 @@ def remove_interval(self, start: TimeLike, end: TimeLike) -> None:
...
         Args:
             start: The start date/time of the interval (inclusive)
             end: The end date/time of the interval (inclusive)
             strict: Whether to fail when the inclusive start is the same as the exclusive end.
+            for_removal: Indicates if the interval is being calculated as a removal instead of an add.
 
         Returns:
             A [start, end) pair.
         """
+        end = now() if for_removal and self.depends_on_past else end
```
tobymao, line 572: should this be latest?

PR #926, opened 2023-05-30
Fix!: Make sure internal models are not accidentally converted to external ones

izeigerman:


sqlmesh/core/plan/evaluator.py
```diff
@@ -159,17 +159,20 @@ def _restate(self, plan: Plan) -> None:
...
-            if plan.is_dev
-            else self.state_sync.get_snapshots_by_models(*plan.restatements)
-        )
-        self.state_sync.remove_interval(
-            [],
-            start=plan.start,
-            end=plan.end,
-            all_snapshots=all_snapshots,
-        )
+        target_snapshots = [s for s in plan.snapshots if s.name in plan.restatements]
```
tobymao, line 162: what's going on here?

sqlmesh/core/schema_loader.py
```diff
@@ -24,13 +29,23 @@ def create_schema_file(
...
     """
     external_tables = {
         dep for model in models.values() for dep in model.depends_on if dep not in models
     }
 
+    # Make sure we don't convert internal models into external ones.
+    existing_models = state_reader.models_exist(external_tables, exclude_external=True)
+    if existing_models:
+        logger.warning(
+            "The following models already exist and can't be converted to external: %s",
```
tobymao, line 44: i think a helpful hint would be nice. Did you forget to rename a reference...

tobymao (approved):

sqlmesh/core/plan/evaluator.py
izeigerman: Discussed in person. This is to make sure that when a user triggers restatement in the development environment, it only impacts snapshots in that target environment, without impacting other environments if possible.

PR #927, opened 2023-05-30
Fix: Fixing some minor styling bugs

mykalmax:


web/client/src/library/pages/docs/Docs.tsx
```diff
@@ -45,59 +39,52 @@ export default function PageDocs(): JSX.Element {
...
+        <div className="p-4 flex flex-col w-full h-full overflow-hidden">
+          <div className="flex justify-center items-center w-full h-full">
+            <div className="center">
+              <h3 className="text-lg mb-4">
+                Still Waiting For Models To Load...
+              </h3>
+              <p>It should take few more seconds.</p>
+              <p>
+                If it takes too long, probably, there is a problem with
+                connection.
```
vchan, line 52: The UI and API both run locally so the issue wouldn't be with their connection, right?

web/client/src/library/pages/docs/Docs.tsx
```diff
@@ -45,59 +39,52 @@ export default function PageDocs(): JSX.Element {
...
+              </p>
+              <p>Check if server is running and refresh.</p>
```
vchan, line 54:
```suggestion
              <p>Check if the server is running and refresh.</p>
```

vchan (approved):

web/client/src/library/pages/docs/Docs.tsx
mykalmax: meant with connection between UI and API, need to rephrase

PR #930, opened 2023-05-31
Fix!: normalize identifiers and model names

GeorgeSittas:
This PR (WIP) aims to improve how we handle case-sensitive identifiers across different dialects, especially in Snowflake. The problem is that Snowflake [resolves identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax) as uppercase, which [used to be](https://github.com/tobymao/sqlglot/commit/dcfe67f5aff6a1f8739c704dfc4f71ad2f2b293e) inconsistent with SQLGlot's optimizer, hence causing problems when expanding star projections (case-sensitivity was not preserved).

There are several outstanding items:

- [x] Make a new SQLGlot release and bump its version here.
- [x] Ensure that model names are consistently normalized, e.g. when received from a CLI / Notebook command.
- [x] Same goes for `depends_on`: if a user specifies this as part of the model's meta, every name in it needs to be normalized as well, according to the given dialect. When it's missing from the model's meta, `depends_on` returns normalized names because `_optimize_query` is called under the hood.
- [x] Make sure to normalize & quote as needed in places where we create new identifiers. One such example is when we're expanding model names into subqueries: the aliases attached to these subqueries aren't currently quoted, because they're created post-optimization. EDIT: this is now handled towards the very end of `render_query`.

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -109,13 +109,8 @@ def render(
     dialect: t.Optional[str] = None,
 ) -> None:
     """Renders a model's query, optionally expanding referenced models."""
-    snapshot = ctx.obj.snapshots.get(model)
-
-    if not snapshot:
-        raise click.ClickException(f"Model `{model}` not found.")
-
     rendered = ctx.obj.render(
-        snapshot,
+        model,
         start=start,
         end=end,
         latest=latest,
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -51,7 +51,7 @@ from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context_diff import ContextDiff
 from sqlmesh.core.dialect import (
     format_model_expressions,
-    normalize_table,
+    normalize_model_name,
     pandas_to_sql,
     parse,
 )
@@ -282,7 +282,7 @@ class Context(BaseContext):
         Returns:
             A new instance of the updated or inserted model.
         """
-        model = self.get_model(model)
+        model = self.get_model(model, raise_if_missing=True)
         path = model._path
 
         # model.copy() can't be used here due to a cached state that can be a part of a model instance.
@@ -414,7 +414,7 @@ class Context(BaseContext):
             The expected model.
         """
         if isinstance(model_or_snapshot, str):
-            normalized_name = normalize_table(model_or_snapshot, dialect=self.config.dialect)
+            normalized_name = normalize_model_name(model_or_snapshot, dialect=self.config.dialect)
             model = self._models.get(normalized_name)
         elif isinstance(model_or_snapshot, Snapshot):
             model = model_or_snapshot.model
@@ -451,7 +451,7 @@ class Context(BaseContext):
             The expected snapshot.
         """
         if isinstance(model_or_snapshot, str):
-            normalized_name = normalize_table(model_or_snapshot, dialect=self.config.dialect)
+            normalized_name = normalize_model_name(model_or_snapshot, dialect=self.config.dialect)
             snapshot = self.snapshots.get(normalized_name)
         elif isinstance(model_or_snapshot, Snapshot):
             snapshot = model_or_snapshot
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -282,13 +282,13 @@ class Context(BaseContext):
         Returns:
             A new instance of the updated or inserted model.
         """
-        if isinstance(model, str):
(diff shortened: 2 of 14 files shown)


sqlmesh/core/model/meta.py
```diff
@@ -41,9 +42,9 @@ class IntervalUnit(str, Enum):
 class ModelMeta(PydanticModel):
     """Metadata for models which can be defined in SQL."""
 
+    dialect: str = ""
     name: str
     kind: ModelKind = ModelKind(name=ModelKindName.VIEW)
-    dialect: str = ""
```
GeorgeSittas, line 46: The `dialect` was moved before `name` in order to reference it in `name`'s validator (see below).

sqlmesh/core/loader.py
```diff
@@ -52,9 +53,7 @@ def update_model_schemas(dag: DAG[str], models: UniqueKeyDict[str, Model]) -> No
 
         for dep in model.depends_on:
             external = external or dep not in models
-            table = schema._normalize_table(
-                schema._ensure_table(dep, dialect=model.dialect), dialect=model.dialect
-            )
+            table = exp.maybe_parse(dep, into=exp.Table, dialect=model.dialect)
```
tobymao, line 56: can also just do to_table(...)

sqlmesh/cli/main.py
```diff
@@ -109,13 +109,8 @@ def render(
...
 ) -> None:
     """Renders a model's query, optionally expanding referenced models."""
-    snapshot = ctx.obj.snapshots.get(model)
-
-    if not snapshot:
-        raise click.ClickException(f"Model `{model}` not found.")
-
     rendered = ctx.obj.render(
-        snapshot,
+        model,
```
GeorgeSittas, line 113: This check was moved in `Context.render` (fyi: there wasn't a similar check in the `evaluate` function).

tests/core/test_model.py
```diff
@@ -1045,6 +1046,54 @@ def test_star_expansion(assert_exp_eq) -> None:
     )
 
 
+def test_case_sensitivity(assert_exp_eq):
+    config = Config(model_defaults=ModelDefaultsConfig(dialect="snowflake"))
```
GeorgeSittas, line 1050: Important note: if the config's default dialect is different from what's specified in the model, there might be an issue when doing lookup on the model's name. For example, if we had:

```sql
MODEL (
    name example.source,
    kind EMBEDDED,
    dialect snowflake,
);

SELECT
    "id",
    "name",
    "payload"
FROM db.schema."table"
```

And the config had no default dialect, doing `sqlmesh render example.source` would likely fail because the model's name is normalized into `EXAMPLE.SOURCE` and so we'll be unable to find `example.source`.

This is an issue that will probably only come up in Snowflake codebases, so maybe it's a reasonable tradeoff. The reason we can't use the model's dialect and rely on the config's default one, is that we _need_ to know the corresponding `Model` object in `Context.render` to refer to its `dialect` property, so it becomes a chicken-and-egg problem.

tests/dbt/test_transformation.py
```diff
@@ -104,6 +104,7 @@ def test_model_kind():
 
 def test_model_columns():
     model = ModelConfig(
+        alias="test",
```
GeorgeSittas, line 107: This was added because we're hitting a parse error due to the model's name being an empty string => we're trying to parse an empty string into a `Table` inside the model's `"name"` validator and it raises a parse error.

sqlmesh/core/context.py
```diff
@@ -277,13 +282,13 @@ def upsert_model(self, model: t.Union[str, Model], **kwargs: t.Any) -> Model:
         Returns:
             A new instance of the updated or inserted model.
         """
-        if isinstance(model, str):
-            model = self._models[model]
+        model = self.get_model(model)
```
tobymao, line 285: do you need to set raise_on_missing?

sqlmesh/core/dialect.py
```diff
@@ -544,3 +546,9 @@ def pandas_to_sql(
         batch_size=batch_size,
         alias=alias,
     )
+
+
+def normalize_table(table: str | exp.Table, dialect: DialectType = None) -> str:
```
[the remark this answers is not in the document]
tobymao: you normalize_model_name

sqlmesh/core/context.py
GeorgeSittas: Yeah, good catch thanks

tobymao (approved):

PR #931, opened 2023-05-31
Chore: Output rendered queries to debug log.

crericha:


sqlmesh/utils/formatting.py
```diff
@@ -0,0 +1,3 @@
+def indent(string: str, num: int = 1) -> str:
```
tobymao, line 1: why do you need this?

crericha: Its difficult to scan the log when the query is inline with all the other text. Indenting made it easy to parse.

tobymao: this only indents line breaks though, can you just do

query.sql(indent=4)

crericha: Didn't realize sql() had an indent option. Will look into that.

crericha: @tobymao what I'm trying to do is indent level 0. `indent` only applies to level 1+.

crericha: Screenshot 2023-05-31 at 9 51 26 PM

tobymao: ok, use textwrap.indent

tobymao (approved):

PR #933, opened 2023-05-31
Feat: Add primary key capture to model to seed future interfaces

z3z1ma:
We should start capturing primary key early as it could seed future interfaces. What is a model without a primary key anyway. 

Given this interface, we can simplify the data-diff interface, use this to do automatic non_null & unique audits, and to communicate outwards in the UI and metadata clearly the key for the model, be it a single or composite value.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1468,7 +1468,7 @@ META_FIELD_CONVERTER: t.Dict[str, t.Callable] = {
     "tags": lambda value: (
         exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
     ),
-    "primary_key": lambda value: (
+    "grain": lambda value: (
         exp.to_column(value[0]) if len(value) == 1 else exp.Tuple(expressions=exp.to_column(value))
-    )
+    ),
 }
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1469,6 +1469,6 @@ META_FIELD_CONVERTER: t.Dict[str, t.Callable] = {
         exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
     ),
     "primary_key": lambda value: (
-        exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
+        exp.to_column(value[0]) if len(value) == 1 else exp.Tuple(expressions=exp.to_column(value))
     )
 }
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1468,4 +1468,7 @@ META_FIELD_CONVERTER: t.Dict[str, t.Callable] = {
     "tags": lambda value: (
         exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
     ),
+    "primary_key": lambda value: (
+        exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
+    )
 }
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -57,6 +57,7 @@ class ModelMeta(PydanticModel):
     column_descriptions_: t.Optional[t.Dict[str, str]]
     audits: t.List[AuditReference] = []
     tags: t.List[str] = []
+    primary_key: t.List[str] = []
     hash_raw_query: bool = False
 
     _croniter: t.Optional[croniter] = None
@@ -105,7 +106,7 @@ class ModelMeta(PydanticModel):
             ]
         return v
 
-    @validator("partitioned_by_", "tags", pre=True)
+    @validator("partitioned_by_", "tags", "primary_key", pre=True)
     def _value_or_tuple_validator(cls, v: t.Any) -> t.Any:
         if isinstance(v, (exp.Tuple, exp.Array)):
             return [e.name for e in v.expressions]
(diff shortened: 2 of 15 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -941,6 +941,7 @@ def _model_metadata_hash(model: Model, audits: t.Dict[str, Audit]) -> str:
         str(model.batch_size) if model.batch_size is not None else None,
         json.dumps(model.mapping_schema, sort_keys=True),
         *model.tags,
+        *model.primary_key
```
izeigerman, line 944: I wonder whether should be a part of the data hash instead. For instance we can use this to set the actual primary key constraint for databases that support this (eg. postgres), in which case we'd like want a new table version.

izeigerman: @tobymao wdyt?

sqlmesh/core/model/definition.py
```diff
@@ -1468,4 +1468,7 @@ def _is_udtf(expr: exp.Expression) -> bool:
     "tags": lambda value: (
         exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
     ),
+    "primary_key": lambda value: (
+        exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
```
izeigerman, line 1472: This looks identical to line 1469. Should we have a function for this?

izeigerman:
@z3z1ma thanks for contributing this, Alex!

sqlmesh/core/model/definition.py
GeorgeSittas: I wonder whether doing `exp.Tuple(expressions=value)` is correct in this context. Why are we not parsing each expression in `value` into an identifier as well?

z3z1ma: It seems all that this is used for is converting meta fields into AST nodes for the purposes of rendering a definition. So having a Union output of Tuple or Identifier based on the input being multiple values or one value seems secondary. At least at a glance, though I am all for less union types, since its simple later on.

z3z1ma: Unless you mean:
`exp.Tuple(expressions=exp.to_identifier(value))` 

In which case, makes sense actually 🙃

tobymao:
i had a discussion with @eakmanrq on this and we decided to use the terminology "grain" because "primary key" evokes the notion of a constraint which is not applied.

sqlmesh/core/model/definition.py
tobymao: i think the ast here is wrong, it should be exp.to_column()

sqlmesh/core/model/meta.py
```diff
@@ -57,6 +57,7 @@ class ModelMeta(PydanticModel):
     column_descriptions_: t.Optional[t.Dict[str, str]]
     audits: t.List[AuditReference] = []
     tags: t.List[str] = []
+    primary_key: t.List[str] = []
```
tobymao, line 60: let's rename this to grain

sqlmesh/core/snapshot/definition.py
tobymao: it is explicitly not an actual primary key, which is why i want it to be called grain and only used for metadata

z3z1ma:
> i had a discussion with @eakmanrq on this and we decided to use the terminology "grain" because "primary key" evokes the notion of a constraint which is not applied.

Primary key as an idea or concept is so ubiquitous. Its **baked** into our thinking, especially as data engineers working  on OLAP systems without constraints constantly. I still tell people the field is the **pk** or these fields constitute the **pk**. so by definition alone, I like `primary_key`. But I am not THAT invested, I understand the pushback wrt the conflation with constraints. :) 

> In the [relational model](https://en.wikipedia.org/wiki/Relational_model) of [databases](https://en.wikipedia.org/wiki/Database), a primary key is a specific choice of a minimal set of attributes ([columns](https://en.wikipedia.org/wiki/Column_(database))) that uniquely specify a tuple ([row](https://en.wikipedia.org/wiki/Row_(database))) in a [relation](https://en.wikipedia.org/wiki/Relation_(database)) ([table](https://en.wikipedia.org/wiki/Table_(database))).[[a]](https://en.wikipedia.org/wiki/Primary_key#cite_note-1)[[1]](https://en.wikipedia.org/wiki/Primary_key#cite_note-2) Informally, a primary key is "which attributes identify a record," and in simple cases constitute a single attribute: a unique ID. More formally, a primary key is a choice of [candidate key](https://en.wikipedia.org/wiki/Candidate_key) (a minimal [superkey](https://en.wikipedia.org/wiki/Superkey)); any other candidate key is an alternate key.

z3z1ma:
> can you add some primary_keys / grains to our existing sushi models

Yes

tobymao (approved):

sqlmesh/core/model/definition.py
```diff
@@ -1468,4 +1468,7 @@ def _is_udtf(expr: exp.Expression) -> bool:
     "tags": lambda value: (
         exp.to_identifier(value[0]) if len(value) == 1 else exp.Tuple(expressions=value)
     ),
+    "grain": lambda value: (
+        exp.to_column(value[0]) if len(value) == 1 else exp.Tuple(expressions=exp.to_column(value))
+    ),
```
GeorgeSittas, line 1473: Sorry for the late response, I just saw the discussion in the previous comment.

I'm not sure about the else clause here. If `value`'s supposed to be a list (which is consistent with the `len(value) == 1` check), then the `expressions=exp.to_column(value)` part doesn't really make sense in this context, because we can't convert a list to a column.

I think we might want to do something like:

```python
exp.Tuple(expressions=[exp.to_column(v) for v in value]
```

Same goes for the `"tags"` convert right above this.

cc: @tobymao

GeorgeSittas: Yep

PR #935, opened 2023-06-01
Fix: Use dbt source_name for source method and add support package arg for ref method.

crericha:


sqlmesh/dbt/model.py
```diff
@@ -20,6 +21,11 @@
 from sqlmesh.dbt.target import TargetConfig
 from sqlmesh.utils.errors import ConfigError
 
+if sys.version_info >= (3, 9):
```
izeigerman, line 24: Should this be removed?

sqlmesh/dbt/seed.py
```diff
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
+import sys
 import typing as t
 
 from sqlmesh.core.model import Model, SeedKind, create_seed_model
 from sqlmesh.dbt.basemodel import BaseModelConfig
 
+if sys.version_info >= (3, 9):
```
izeigerman, line 9: Ditto

sqlmesh/dbt/source.py
```diff
@@ -28,11 +28,9 @@ class SourceConfig(GeneralConfig):
         columns: Columns within the source
     """
 
-    # sqlmesh fields
-    config_name: str = ""
-
     # DBT configuration fields
-    name: t.Optional[str] = None
+    name: str = ""
+    source_name_: str = Field("", alias="source_name")
```
izeigerman, line 33: Do we even need the `source_name` property name now? Overloading this property seems weird, given the method returns something entirely different.

sqlmesh/dbt/model.py
crericha: Surprised these weren't auto-removed when the dependencies were removed.

sqlmesh/dbt/source.py
crericha: The source_name property is meant to represent the name within the db. I thought about changing model_name and source_name to model_sql_name and source_sql_name. What are your thoughts?

sqlmesh/dbt/context.py
```diff
@@ -173,3 +184,35 @@ def jinja_globals(self) -> t.Dict[str, JinjaGlobalAttribute]:
...
+        from sqlmesh.dbt.seed import SeedConfig
+
+        dependency_context = self.copy()
+
+        models = {}
+        seeds = {}
+        sources = {}
+
+        for ref in dependencies.refs:
+            model = self.model_for_ref(ref)
```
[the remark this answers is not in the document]
crericha: We could cache both ways like I do for ref.

sqlmesh/dbt/source.py
izeigerman: Can we just do sql_name? That should help I think

sqlmesh/dbt/context.py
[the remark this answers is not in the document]
izeigerman: Can we just use this as a key in existing dicts? Otherwise - yes, let’s cache this indexing too.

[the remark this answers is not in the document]
crericha: There are two viable options for how a ref can be called -- ref(package_name, model_name) or ref(model_name). I'm using package_name.model_name for the former and model_name for the latter.

The dependency could be in either form, so I need to be able to find the corresponding model from either method.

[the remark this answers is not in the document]
izeigerman: Got it, makes sense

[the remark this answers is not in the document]
crericha: Refactored and simplified by changing values held by refs dict to the model itself.

sqlmesh/dbt/builtin.py
```diff
@@ -152,11 +152,9 @@ def var(name: str, default: t.Optional[str] = None) -> str:
 
 
 def generate_ref(refs: t.Dict[str, t.Any]) -> t.Callable:
-
-    # TODO suport package name
     def ref(package: str, name: t.Optional[str] = None) -> t.Optional[BaseRelation]:
-        name = name or package
-        relation_info = refs.get(name)
+        ref_name = f"{package}.{name}" if name else package
```
izeigerman, line 156: Can there be a freak case when the `package` is the same as the `name` in a different package?

izeigerman: Never mind, it shouldn't matter.

sqlmesh/dbt/common.py
```diff
@@ -19,7 +19,7 @@
 T = t.TypeVar("T", bound="GeneralConfig")
 
 if t.TYPE_CHECKING:
-    from sqlmesh.dbt.context import DbtContext
+    pass
```
izeigerman, line 22: Should be removed

izeigerman (approved):
Few more comments, LGTM otherwise

sqlmesh/dbt/manifest.py
```diff
@@ -310,7 +310,11 @@ def _node_base_config(node: ManifestNode) -> t.Dict[str, t.Any]:
 
 
 def _tests_for_node(node: ManifestNode, tests: t.Dict[str, TestConfig]) -> t.List[TestConfig]:
-    return [test for test in tests.values() if test.owner == node.name]
+    return [
```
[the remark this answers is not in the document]
crericha: Okay, I can cache it.

PR #936, opened 2023-06-01
Chore: Fix typo and rephrase unresolved projections error.

crericha:


sqlmesh/core/loader.py
```diff
@@ -70,8 +70,9 @@ def update_model_schemas(dag: DAG[str], models: UniqueKeyDict[str, Model]) -> No
             if "*" in model.columns_to_types:
                 raise ConfigError(
                     f"Can't expand SELECT * expression for model '{name}'."
-                    " Projects from an external source must either be explicitly specified within models or added as"
-                    " external models by using the command 'sqlmesh create_external_models'."
+                    " Projections from an external source must be explicitly specified within"
+                    " models or add source tables as external models by using the command"
+                    " 'sqlmesh create_external_models'."
```
GeorgeSittas, line 75: This reads a bit awkwardly to me, especially the "or add ..." part. What about:

```
Can't expand SELECT * expression for model '{name}'. Consider specifying the external source
projections explicitly or adding the source tables as "external models" by using the command ...
```

crericha: I don't think `consider` is the appropriate word because you *have* to do something to resolve the problem.

I can quote external models.

GeorgeSittas: I see, agreed. I was actually aiming to rephrase the `must be explicitly specified.... or add` part, because the first option is phrased using passive voice and the second using active.

GeorgeSittas (approved):

PR #938, opened 2023-06-01
Fix: ensure AST is correct in meta field converter

GeorgeSittas:
If we don't mind converting the grains to `Identifier`s, the implementation can be simplified (fwiw it can also be simplified if I just do it all inline like it used to be).


tobymao (approved):

tobymao:
sure we can change it to identifier for all

PR #941, opened 2023-06-02
Docs: Add depends_on_past

eakmanrq:


tobymao (approved):
i think the language here should be a bit more opionated and slightly discourage this pattern

eakmanrq:
> @eakmanrq are we waiting on anything here?

Yep waiting on finalizing how `depends_on_past` works (https://github.com/TobikoData/sqlmesh/pull/1018). For example we need to update documentation to talk about ignored snapshots.

PR #943, opened 2023-06-05
Fix: select * from external models for dbt, skip private variables when deep copying JinjaMacroRegistry, and properly parse jinja pre and post expressions

crericha:


izeigerman (approved):

tests/core/test_model.py
```diff
@@ -1185,3 +1186,8 @@ def test_is_breaking_change():
         )
         is None
     )
+
+
+def test_parse_expression_list_with_jinja():
```
izeigerman, line 1191: thank you! 🙏

PR #944, opened 2023-06-05
Feat: Enhance notification target functionality

vchan:
This PR adds Slack and SMTP notification targets as well as the notion of events that should trigger notifications.

- Notification targets defined in a config but missing fields will be silently ignored. Most common case is probably fields that are taken from environment variables but the environment variables are not available.
- If no username is specified in a config, the config's notification targets are used. Most common use case is probably production environments.
- If a username is specified in a config, only that user gets notified. Most common use case is probably users developing locally. There is currently one exception. See below:
- Audit model owners are notified their audits failed if there is a user with a notification target configured for an `audit_failure` event and the environment is prod.

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -11,7 +11,6 @@ from sqlmesh.core.config import (
 )
 from sqlmesh.core.notification_target import (
     BasicSMTPNotificationTarget,
-    NotificationStatus,
     SlackApiNotificationTarget,
     SlackWebhookNotificationTarget,
 )
@@ -19,14 +18,6 @@ from sqlmesh.core.user import User, UserRole
 
 CURRENT_FILE_PATH = os.path.abspath(__file__)
 DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
-
-
-class CustomSMTPNotificationTarget(BasicSMTPNotificationTarget):
-    def notify_run_failure(self, exc: Exception) -> None:
-        with open("/home/sqlmesh/sqlmesh.log", "r", encoding="utf-8") as f:
-            msg = f"{exc}\n\n{f.read()}"
-        subject = "SQLMesh Run Failed!"
-        self.send(notification_status=NotificationStatus.FAILURE, msg=msg, subject=subject)
 
 
 # An in memory DuckDB config.
@@ -85,7 +76,7 @@ required_approvers_config = Config(
             notify_on=["apply_start", "run_start"],
             url=os.getenv("SLACK_WEBHOOK_URL"),
         ),
-        CustomSMTPNotificationTarget(
+        BasicSMTPNotificationTarget(
             notify_on=["run_failure"],
             host=os.getenv("SMTP_HOST"),
             user=os.getenv("SMTP_USER"),
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -9,10 +9,24 @@ from sqlmesh.core.config import (
     GatewayConfig,
     SparkConnectionConfig,
 )
+from sqlmesh.core.notification_target import (
+    BasicSMTPNotificationTarget,
+    NotificationStatus,
+    SlackApiNotificationTarget,
+    SlackWebhookNotificationTarget,
+)
 from sqlmesh.core.user import User, UserRole
 
 CURRENT_FILE_PATH = os.path.abspath(__file__)
 DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
+
+
+class CustomSMTPNotificationTarget(BasicSMTPNotificationTarget):
+    def notify_run_failure(self, exc: Exception) -> None:
+        with open("/home/sqlmesh/sqlmesh.log", "r", encoding="utf-8") as f:
+            msg = f"{exc}\n\n{f.read()}"
+        subject = "SQLMesh Run Failed!"
+        self.send(notification_status=NotificationStatus.FAILURE, msg=msg, subject=subject)
 
 
 # An in memory DuckDB config.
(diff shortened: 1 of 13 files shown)


examples/sushi/config.py
```diff
@@ -9,12 +9,26 @@
...
+)
 from sqlmesh.core.user import User, UserRole
 
 CURRENT_FILE_PATH = os.path.abspath(__file__)
 DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
 
 
+class CustomSMTPNotificationTarget(BasicSMTPNotificationTarget):
+    def notify_run_failure(self, exc: Exception) -> None:
+        with open("/home/sqlmesh/sqlmesh.log", "r", encoding="utf-8") as f:
```
tobymao, line 26: why is this hardcoded?

tobymao:
does this work? looks fine to me

examples/sushi/config.py
vchan: Oops, I'll remove it, although it'd be nice to showcase how to customize notifications to handle more advance use cases.

examples/sushi/config.py
```diff
@@ -53,5 +67,33 @@
...
 required_approvers_config = Config(
     default_connection=DuckDBConnectionConfig(),
-    users=[User(username="test", roles=[UserRole.REQUIRED_APPROVER])],
+    users=[
+        User(
+            username="test",
+            roles=[UserRole.REQUIRED_APPROVER],
+            notification_targets=[
+                SlackApiNotificationTarget(
+                    notify_on=["apply_start", "audit_failure"],
```
eakmanrq, line 76: From what I can tell, it seems like if I subscribe to these events then I will get all of them notified to me instead of just those that are relevant to me? For example in prod I want audits to notify owners of the model upon failure but when in dev I want apply and audits to notify the plan author of failure. From what I can tell in the current implementation I will get notified for everything. Is that right?

tobymao:
@vchan what are we waiting on for this one?

vchan:
I'm working on the config environment variables, but the actual notification functionality is good to go. I was trying to see if I could disable Pydantic field validations during initialization and trigger them manually after the context loads the configs, but it doesn't seem possible at this point.

sqlmesh/core/notification_target.py
```diff
@@ -1,19 +1,28 @@
...
 
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
 
+if t.TYPE_CHECKING:
+    from slack_sdk import WebClient, WebhookClient
+
+NOTIFICATION_FUNCTIONS: dict[NotificationEvent, str] = {}
```
eakmanrq, line 24: t.Dict[NotificationEvent, str] instead of `dict[NotificationEvent, str]`

sqlmesh/core/notification_target.py
```diff
@@ -43,20 +52,116 @@ def is_progress(self) -> bool:
...
+class NotificationTargetManager:
+    """Wrapper around a list of notification targets.
+
+    Calling a notification target's "notify_" method on this object will call it
+    on all registered notification targets.
+    """
+
+    def __init__(
+        self,
+        notification_targets: dict[NotificationEvent, set[BaseNotificationTarget]] | None = None,
```
eakmanrq, line 131: t.Dict and t.Set

sqlmesh/core/notification_target.py
```diff
@@ -43,20 +52,116 @@ def is_progress(self) -> bool:
...
+        user_notification_targets: dict[str, set[BaseNotificationTarget]] | None = None,
+        username: str = "",
```
eakmanrq, line 133: [Nit] I think a more standard pattern in the codebase is to make it optional and default to None instead of empty string?

sqlmesh/core/notification_target.py
```diff
@@ -79,3 +184,97 @@ def send(self, notification_status: NotificationStatus, msg: str, **kwargs: t.An
...
+    _client: t.Optional[WebhookClient] = None
+
+    @property
+    def client(self) -> WebhookClient:
+        if not self._client:
+            try:
+                from slack_sdk import WebhookClient
+            except ModuleNotFoundError as e:
+                raise MissingDependencyError(
+                    "Missing Slack dependencies. Run `pip install 'sqlmesh[slack]'` to install them."
```
eakmanrq, line 201: Nice handling of this.

eakmanrq (commented):
A few small comments. Excited to get this in!

examples/sushi/config.py
```diff
@@ -53,5 +58,33 @@
 
 required_approvers_config = Config(
     default_connection=DuckDBConnectionConfig(),
-    users=[User(username="test", roles=[UserRole.REQUIRED_APPROVER])],
+    users=[
+        User(
+            username="test",
```
tobymao, line 63: "admin"

examples/sushi/config.py
```diff
@@ -53,5 +58,33 @@
...
+            roles=[UserRole.REQUIRED_APPROVER],
+            notification_targets=[
+                SlackApiNotificationTarget(
+                    notify_on=["apply_start", "audit_failure"],
+                    token=os.getenv("SLACK_API_TOKEN"),
```
tobymao, line 68: "ADMIN_SLACK_API_TOKEN"

sqlmesh/core/context.py
```diff
@@ -749,7 +764,17 @@ def apply(self, plan: Plan) -> None:
...
         if plan.uncategorized:
             raise PlanError("Can't apply a plan with uncategorized changes.")
-        self._scheduler.create_plan_evaluator(self).evaluate(plan)
+        self.notification_target_manager.notify(
+            NotificationEvent.APPLY_START, environment=plan.environment.name
+        )
+        try:
+            self._scheduler.create_plan_evaluator(self).evaluate(plan)
+        except Exception as e:
+            self.notification_target_manager.notify(NotificationEvent.APPLY_FAILURE, e)
```
tobymao, line 773: e is not useful right, can you notify the traceback? import tb

tobymao (approved):
let's do a follow up pr and add docs / guides

maybe vincent you can do a first pass ?

cc @treysp

eakmanrq (approved):

PR #945, opened 2023-06-05
Feat: Add spark connection config

eakmanrq:
Resolves: https://github.com/TobikoData/sqlmesh/issues/576

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -6,10 +6,13 @@ from sqlmesh.core.config import (
     CategorizerConfig,
     Config,
     DuckDBConnectionConfig,
+    GatewayConfig,
+    SparkConnectionConfig,
 )
 from sqlmesh.core.user import User, UserRole
 
-DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
+CURRENT_FILE_PATH = os.path.abspath(__file__)
+DATA_DIR = os.path.join(CURRENT_FILE_PATH, "data")
 
 
 # An in memory DuckDB config.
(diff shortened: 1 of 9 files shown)


examples/sushi/config.py
```diff
@@ -28,12 +28,28 @@
 )
 
 # Due to a 3.7 mypy bug we ignore. Can remove once 3.7 support is dropped.
-airflow_config = Config(default_scheduler=AirflowSchedulerConfig())  # type: ignore
+airflow_config = Config(  # type: ignore
+    default_scheduler=AirflowSchedulerConfig(),
+    gateways={
+        "spark": {
```
izeigerman, line 34: Should we use models instead of dicts?

sqlmesh/utils/__init__.py
```diff
@@ -162,6 +162,19 @@ def nullsafe_join(join_char: str, *args: t.Optional[str]) -> str:
     return join_char.join(filter(None, args))
 
 
+def str_to_bool(s: t.Optional[str] = None) -> bool:
```
izeigerman, line 165: Can we also use this here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/__init__.py#L116

examples/sushi/config.py
```diff
@@ -28,12 +28,28 @@
...
+            "connection": {
+                "type": "spark",
```
izeigerman, line 36: This config is used when pytest runs on the host (as opposed to container). Shouldn't we provide a path to the config folder here?

tests/schedulers/airflow/test_end_to_end.py
```diff
@@ -38,7 +38,9 @@ def test_sushi(mocker: MockerFixture, is_docker: bool):
         auto_apply=True,
     )
 
-    data_validator.validate("sushi.customer_revenue_lifetime", start, end)
+    data_validator.validate(
+        "sushi.customer_revenue_lifetime", start, yesterday(), env_name="test_dev"
```
eakmanrq, line 42: So it appears that when we pass in `now()` as end to plan then it gets converted into an inclusive yesterday. This function wants an inclusive start and end so I pass yesterday into it. The reason why it is inclusive is because it currently just handles dates and therefore you call it like you would a between.

sqlmesh/utils/__init__.py
izeigerman, line 165: [Nit] Why have a default value at all?

izeigerman (approved):

sqlmesh/utils/__init__.py
eakmanrq: Yeah good point. Will remove the default.

PR #949, opened 2023-06-06
Fix: Use fixed dates for table diff test

vchan:


tests/core/test_table_diff.py
```diff
@@ -28,6 +46,6 @@ def test_data_diff(sushi_context):
     assert schema_diff.removed == []
 
     row_diff = diff.row_diff()
-    assert row_diff.source_count == 123
-    assert row_diff.target_count == 1685
+    assert row_diff.source_count == 546
```
tobymao, line 49: why do the numbers match?

PR #951, opened 2023-06-06
Feat: Add support for materialzied views

### LINKED ISSUE #904: Add Materialized View Support
Snowflake is an initial example to target for support. 

https://tobiko-data.slack.com/archives/C044BRE5W4S/p1685051232200519

izeigerman:
This PR introduces support for materialized views. 

The `VIEW` kind has been extended to accommodate the new `materialized` flag. The model definition looks as follows:
```sql
MODEL (
    name test_schema.test_model,
    kind VIEW (
        materialized true
    )
);
```

The support for materialized views has been enabled for the following engines:
* Snowflake
* BigQuery
* Redshift
* Postgres

The `materialized` flag is ignored when applied to engines that don't support it. 

During evaluation of a model of this kind, the view will be replaced (or recreated) only if the model's query rendered during evaluation doesn't match the query used during the previous view creation for this model, or if the target view doesn't exist. Thus, views are recreated only when needed in order to realize all the benefits provided by materialized views.

Closes #904

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -127,7 +127,7 @@ class SnowflakeConnectionConfig(_ConnectionConfig):
     def _validate_authenticator(
         cls, values: t.Dict[str, t.Optional[str]]
     ) -> t.Dict[str, t.Optional[str]]:
-        if values["type"] != "snowflake":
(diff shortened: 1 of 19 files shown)


eakmanrq (approved):
Nice!

tobymao (approved):

PR #952, opened 2023-06-07
Fix: Support sql_header config and config.set() jinja method for dbt projects.

crericha:


sqlmesh/dbt/model.py
```diff
@@ -189,6 +189,7 @@ def _extract_sql_config(self) -> None:
     def to_sqlmesh(self, context: DbtContext) -> Model:
         """Converts the dbt model into a SQLMesh model."""
         dialect = self.model_dialect or context.dialect
+        print(self.sql_no_config)
```
tobymao, line 192: debug

tobymao (approved):

sqlmesh/dbt/model.py
crericha: Good catch

PR #955, opened 2023-06-07
Feat: Introduce the table diff API endpoint

vchan:


tests/web/test_main.py
```diff
@@ -516,3 +516,24 @@ def test_get_lineage(web_sushi_context: Context) -> None:
...
+        no_prompts=True,
+        auto_apply=True,
+        skip_tests=True,
+    )
+    response = client.get(
+        "/api/table_diff",
+        params={
+            "source": "prod",
+            "target": "dev",
+            "on": "s.id = t.id",
```
tobymao, line 533: while you're add it, can you add the automatic usage of "grain" to the table diff? so there's now a concept of grain which will populate the on field auto, so you won't need to set this manually

vchan: Yea, I will add it.

web/server/models.py
```diff
@@ -259,3 +260,37 @@ class ApiExceptionPayload(BaseModel):
...
+    modified: t.Dict[str, str]
+
+    @validator("source_schema", "target_schema", "added", "removed", "modified", pre=True)
+    def validate_schema(
+        cls, v: t.Union[t.Dict[str, exp.DataType], t.List[t.Tuple[str, exp.DataType]]]
+    ) -> t.Dict[str, str]:
+        if isinstance(v, dict):
+            return {k: str(v) for k, v in v.items()}
+        if isinstance(v, list):
+            return {k: str(v) for k, v in v}
```
tobymao, line 281: what if it doesn't go through this block?

tobymao (approved):

web/server/models.py
vchan: Good catch.

PR #956, opened 2023-06-08
Feat: Introduce support for pre- and post- statements to seeds

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -1289,20 +1327,23 @@ def _create_model(
     return t.cast(Model, model)
 
 
-def _extract_sql_model_select(
+SEED_INSERT_MACRO_CALL = d.parse("@SEED_INSERT()")[0]
```
izeigerman, line 1330: @tobymao @crericha open to suggestions around naming this macro call

sqlmesh/core/model/definition.py
```diff
@@ -596,7 +581,64 @@ def render_pre_statements(
...
+
+    @property
+    def post_statements(self) -> t.List[exp.Expression]:
+        return self.post_statements_ or []
+
+    @property
+    def macro_definitions(self) -> t.List[d.MacroDef]:
+        """All macro definitions from the list of expressions."""
+        return [
+            s for s in [*self.pre_statements, *self.post_statements] if isinstance(s, d.MacroDef)
```
tobymao, line 596: itertools.chain() or self.pre_statements + self.post_statements

sqlmesh/core/model/definition.py
```diff
@@ -596,7 +581,64 @@ def render_pre_statements(
...
+        ]
+
+    def _render_statements(
+        self,
+        statements: t.Iterable[exp.Expression],
+        **kwargs: t.Any,
+    ) -> t.List[exp.Expression]:
+        rendered = [
```
tobymao, line 604: (
) to make it a gen so two arrays aren't created

sqlmesh/core/model/definition.py
tobymao: what about @INSERT_SEED()

tobymao (approved):

sqlmesh/dbt/basemodel.py
crericha, line 1: In BaseModelConfig's docstring, remove "(not supported for seeds)" from pre_hook and post_hook.

sqlmesh/dbt/basemodel.py
```diff
@@ -236,6 +237,16 @@ def sqlmesh_model_kwargs(self, context: DbtContext) -> t.Dict[str, t.Any]:
             "jinja_macros": jinja_macros,
             "path": self.path,
             "hash_raw_query": True,
+            "pre_statements": [
+                exp
+                for hook in self.pre_hook
+                for exp in d.parse(hook.sql, default_dialect=self.model_dialect or context.dialect)
```
crericha, line 243: Include sql_header as the first pre_hook if sql_header is not None. I merged changes today that had this behavior in Model.py, which should be here now instead.

crericha: Note that the removed code in Model.py used `pre_hooks` var and not `self.pre_hook`

crericha:
Do plan context diffs include the pre/post statements? If so, is there any work that needs to be done there?

sqlmesh/dbt/basemodel.py
izeigerman: good catch, thanks

izeigerman:
> Do plan context diffs include the pre/post statements? If so, is there any work that needs to be done there?

They do

PR #957, opened 2023-06-08
Docs: Update engines.md

serkef:
`default_engine_operator_args` is actually: `engine_operator_args`
Also the key `sqlmesh_gcp_conn_id` should be: `bigquery_conn_id`


GeorgeSittas (approved):
Good catch, thanks!

I think we also need to fix line 40 (`default_engine_operator_args` -> `engine_operator_args`), as well as the corresponding Redshift section in line ~230.

EDIT: addressed in #958

eakmanrq (approved):
Thank you!

PR #970, opened 2023-06-12
Update quickstart

treysp:


docs/quick_start.md
```diff
@@ -3,7 +3,54 @@ In this quick start guide, you'll get up and running with SQLMesh's scaffold gen
...
+
+Although the pipeline is simple, the project touches on all the primary concepts needed to use SQLMesh productively. 
+
+### Plans
+
+SQLMesh's key actions are creating and applying *plans* to *environments*. 
+
+A [SQLMesh environment](./concepts/environments.md) is an isolated namespace containing models and the data they generated. The most important environment is `prod` ("production"), which consists of the databases behind the applications your business uses to operate each day. Environments other than `prod` provide a place where you can test and preview changes to model code before they go live and affect business operations. 
+
+A [SQLMesh plan](./concepts/plans.md) contains a comparison of one environment to another and the set of changes needed to bring them into alignment. For example, if a new SQL model was added, tested, and run in the `dev` environment, it would need to be added and run in the `prod` environment to bring them into alignment. SQLMesh identifies all such changes and classifies them as either breaking and non-breaking.
```
GeorgeSittas, line 38:
```suggestion
A [SQLMesh plan](./concepts/plans.md) contains a comparison of one environment to another and the set of changes needed to bring them into alignment. For example, if a new SQL model was added, tested, and run in the `dev` environment, it would need to be added and run in the `prod` environment to bring them into alignment. SQLMesh identifies all such changes and classifies them as either breaking or non-breaking. 
```

docs/quick_start.md
```diff
@@ -36,32 +83,129 @@ This will create the directories and files that you can use to organize your SQL
...
+- ./audits
+    - assert_positive_order_ids.sql
+- ./tests
+    - test_full_model.yaml
+
+### Project data
+
+The data used in this example project is contained in the `seed_data.csv` file in the `/seeds` project directory. The data reflects sales of 3 items over 7 days in January 2020.
+
+The data contains three columns, `id`, `item_id`, and `ds`, which correspond to each row's unique ID, the sold item's ID number, and the date the item was sold, respectively.
```
GeorgeSittas, line 103: "The data contains" seems a bit odd. What about "The file contains"?

docs/quick_start.md
```diff
@@ -87,19 +245,27 @@ Virtually Updating 'prod' ━━━━━━━━━━━━━━━━━━
...
 ======================================================================
 Successfully Ran 1 tests against duckdb
 ----------------------------------------------------------------------
 New environment `dev` will be created from `prod`
+Apply - Virtual Update [y/n]: 
+```
+
+The output does not list any added or modified models because `dev` is being created from the existing `prod` environment without modification. 
+
+Line 6 shows that when you apply the plan creating the `dev` environment will only involve a Virtual Update. This is because SQLMesh is able to safely reuse the tables you've already backfilled in the `prod` environment. Enter `y` and press `Enter` to perform the Virtual Update:
```
GeorgeSittas, line 266:
```suggestion
Line 6 shows that when you apply the plan creating the `dev` environment, it will only involve a Virtual Update. This is because SQLMesh is able to safely reuse the tables you've already backfilled in the `prod` environment. Enter `y` and press `Enter` to perform the Virtual Update:
```

docs/quick_start.md
```diff
@@ -143,28 +318,37 @@ Summary of differences against `dev`:
...
 Models needing backfill (missing dates):
-└── sqlmesh_example__dev.incremental_model: (2020-01-01, 2023-04-27)
-Enter the backfill start date (eg. '1 year', '2020-01-01') or blank for the beginning of history: 
+└── sqlmesh_example__dev.incremental_model: 2020-01-01 - 2023-05-31
+Enter the backfill start date (eg. '1 year', '2020-01-01') or blank to backfill from the beginning of history:
+```
+
+Lines 5-9 of the output summarize the differences between the modified pipeline and the existing `dev` environment, detecting that we directly modified `incremental_model` and that `full_model` was indirectly modified because it selects from the incremental model. 
+
+On line 23, we see that SQLMesh semantically understood that the change was additive (added a column not used by `full_model`) and was automatically classified as a non-breaking change.
```
GeorgeSittas, line 344: I'm not sure if "semantically understood" is a valid phrase here.

An alternative to consider: "[..] the change was _semantically additive_ [..]"

GeorgeSittas (commented):
Looks good! I've left a few comments.

Should we consider replacing the term "pipeline" with "DAG"? I have a feeling that we've discussed this in the past and we concluded that the latter is more correct in this context, albeit a bit more technical / possibly less intuitive (?).

cc: @izeigerman

docs/quick_start.md
```diff
@@ -3,7 +3,56 @@ In this quick start guide, you'll get up and running with SQLMesh's scaffold gen
 
 Before beginning, ensure that you meet all the [prerequisites](prerequisites.md) for using SQLMesh.
 
-## 1. Create a SQLMesh project
+## Project structure
+
+This project demonstrates key SQLMesh features by walking through the SQLMesh workflow on a simple data pipeline. This section describes the project structure and the SQLMesh concepts you will encounter as you work through it.
+
+The term "pipeline" is generic and usually refers to more than a collection of models and the relationships among them. We use instead the more specific term "directed acyclic graph" (or "DAG") to refer to the models and their relationships in a SQLMesh project. In this context, "graph" is synonymous with "network" - each model is a node in the network, and two models are linked if one of them `SELECT`s from the other. A SQLMesh project's model graph is "acyclic" because it cannot contain cycles where two models both `SELECT` from one another, either directly or indirectly via other models.
```
izeigerman, line 10: > In this context, "graph" is synonymous with "network"

I'm not sure that by introducing yet another term (network) we help explain what other terms are. 

Can we just get rid of pipelines and networks altogether? I'd rather explain model as a transformation step and then mention that models / transformation depend on each other and, thus, form a graph of dependencies. I think that's all we need:
* Transformation
* Model as expression of transformation
* Graph to capture dependencies between models

PR #974, opened 2023-06-12
Fix: make loading faster

tobymao:


sqlmesh/utils/dag.py
```diff
@@ -80,23 +83,25 @@ def graph(self) -> t.Dict[T, t.Set[T]]:
...
-        result: t.List[T] = []
+        if self._sorted is None:
+            self._sorted = []
 
-        unprocessed_nodes = self.graph
-        while unprocessed_nodes:
-            next_nodes = {node for node, deps in unprocessed_nodes.items() if not deps}
+            unprocessed_nodes = self.graph
+            while unprocessed_nodes:
+                next_nodes = {node for node, deps in unprocessed_nodes.items() if not deps}
```
crericha, line 94: We should raise an exception if next_nodes is empty. It means we have a circular dependency between the remaining unprocessed nodes.

sqlmesh/core/context.py
```diff
@@ -829,7 +828,7 @@ def get_dag(self, format: str = "svg") -> graphviz.Digraph:
 
         graph = graphviz.Digraph(node_attr={"shape": "box"}, format=format)
 
-        for name, upstream in self.dag.graph.items():
+        for name, upstream in self.dag._graph.items():
```
izeigerman, line 831: Why did we make `_graph` private and are now accessing it from the outside?

sqlmesh/core/loader.py
```diff
@@ -41,21 +41,22 @@
...
-    for name in dag.sorted():
+    for name in dag.sorted:
         model = models.get(name)
 
         # External models don't exist in the context, so we need to skip them
         if not model:
             continue
 
         model.update_schema(schema)
+        # don't add columns to types if empty?
```
izeigerman, line 52: Is this a FIXME / TODO?

sqlmesh/utils/cron.py
```diff
@@ -0,0 +1,32 @@
...
+
+import typing as t
+
+from croniter import croniter
+
+from sqlmesh.utils import ttl_cache
+from sqlmesh.utils.date import TimeLike, now, to_datetime
+
+
+@ttl_cache(maxsize=1280000)
```
izeigerman, line 11: what does this magic number mean? Do we need a named constant?

sqlmesh/core/context.py
tobymao: because by default graph does a copy. should i make it not copy by default?

sqlmesh/core/model/meta.py
```diff
@@ -265,10 +267,10 @@ def normalized_cron(self) -> str:
             return "0 0 * * *"
         return ""
 
-    def croniter(self, value: TimeLike) -> croniter:
+    def croniter(self, value: TimeLike) -> CroniterCache:
         if self._croniter is None:
-            self._croniter = croniter(self.normalized_cron())
-        self._croniter.set_current(to_datetime(value))
+            self._croniter = CroniterCache(self.normalized_cron())
+        self._croniter.curr = value
```
izeigerman, line 273: Why are we not passing this in constructor? We basically waste another `now` call.

tobymao: what do you mean?

tobymao: ah you mean value?

sqlmesh/utils/cron.py
```diff
@@ -0,0 +1,32 @@
...
+def cron_next(cron: str, time: TimeLike) -> float:
+    return croniter(cron, to_datetime(time)).get_next()
+
+
+@ttl_cache(maxsize=1280000)
+def cron_prev(cron: str, time: TimeLike) -> float:
+    return croniter(cron, to_datetime(time)).get_prev()
```
izeigerman, line 18: why aren't we better off caching the `croniter` instance itself?

sqlmesh/utils/date.py
```diff
@@ -22,24 +24,33 @@
     from sqlmesh.core.scheduler import Interval
 
 
-def now() -> datetime:
+def now(floor: bool = True) -> datetime:
```
izeigerman, line 27: bad argument name. Suggestions: `round_to_minute`.

sqlmesh/utils/date.py
```diff
@@ -22,24 +24,33 @@
...
     """
-    return datetime.utcnow().replace(tzinfo=UTC)
+    now = datetime.utcnow()
+    if floor:
+        return now.replace(second=0, microsecond=0, tzinfo=UTC)
+    return now.replace(tzinfo=UTC)
 
 
-def now_timestamp() -> int:
+def now_timestamp(floor: bool = False) -> int:
```
izeigerman, line 43: ditto

sqlmesh/utils/date.py
```diff
@@ -96,6 +107,7 @@ def to_timestamp(value: TimeLike, relative_base: t.Optional[datetime] = None) ->
     return int(to_datetime(value, relative_base=relative_base).timestamp() * 1000)
 
 
+@ttl_cache(maxsize=1280000)
```
izeigerman, line 110: ditto: magic number. Should this be default in the `ttl_cache`?

sqlmesh/utils/cron.py
tobymao: cache works on methods, also ceo inter is stateful

tests/utils/test_yaml.py
```diff
@@ -13,7 +13,6 @@ def test_yaml() -> None:
     prod:
       type: postgres
       host: 127.0.0.1
-      # IMPORTANT: Make sure to quote the entire Jinja string here
```
izeigerman, line 16: why was this removed?

sqlmesh/utils/date.py
tobymao: https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.floor.html

sqlmesh/utils/date.py
tobymao: there already is one, ttl_cache is time based, maxsize is pass through to the python cache method

tests/utils/test_yaml.py
tobymao: because the C yaml parser doesn't preserve comments

sqlmesh/utils/date.py
izeigerman: well in that case, there's an additional argument which specifies resolution. We don't have it there. The core of my concern here is that it's unclear to which granularity it will be floored.

sqlmesh/utils/date.py
izeigerman: the current default is 60. Why have it if we override it everywhere. Should the default just be 1280000?

sqlmesh/core/model/meta.py
izeigerman: yes, exactly

izeigerman (approved):

PR #975, opened 2023-06-13
Feat: Cache the optimized model query

izeigerman:


sqlmesh/core/renderer.py
```diff
@@ -279,42 +319,54 @@ def time_column_filter(self, start: TimeLike, end: TimeLike) -> exp.Between:
...
+                            "*",
+                            "",
+                        ):
+                            select.replace(exp.alias_(select, select.output_name))
 
         return annotate_types(simplify(query), schema=schema)
 
 
+@contextmanager
+def _normalize_and_quote(query: E, dialect: str) -> t.Iterator[E]:
```
izeigerman, line 355: wow!

izeigerman:
Maybe add test for the new cache

sqlmesh/core/renderer.py
GeorgeSittas: +1!

PR #977, opened 2023-06-13
Feat: Greatly simplify table diff CLI invocation

z3z1ma:
We need dead-simple diffs. Certain parameters are required so they should be positional arguments, not options. We should leverage the new `grain` metadata (it already leverages it). The invocation should be as concise and self evident as possible without having to remember flags and help text. Please see the examples below to get a feel for the utility and practical approach.

**Notes:**

1. Use pos args, reduce flags
2. Use a new easy-to-grasp syntax for cross env comparison, `<from-env>:<to-env>` as pos arg 1
3. We are diffing a model against itself in another env, so just take the name as pos arg 2
4. We use the `grain` metadata if supplied reducing invocation to just 2 args, total, for full functionality
5. Keep `--on` flag so user has escape hatch to override grain if they want
6. `--where` and `--limit` left as is

🥇 Please see improved ergonomics below.

```sh
sqlmesh table_diff dev:prod sushi.top_waiters

sqlmesh table_diff pr_123:prod sushi.top_waiters

sqlmesh table_diff prod:alex sushi.top_waiters

sqlmesh table_diff alex:toby sushi.top_waiters
```

Now we should also take quick look at the CLI output and improve that in this pass too. `show_row_diff` needs some love. The row count/percentage change numbers are part of the highlight alongside the more obvious schema changes. Let's make sure there are line breaks and negative space to emphasize these.

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -391,7 +391,7 @@ def create_external_models(obj: Context) -> None:
 @click.argument(
     "from_to",
     required=True,
-    help="The source and target environment to diff. Format: source:target",
+    metavar="ENV:ENV",
 )
 @click.argument("model_name", required=True)
 @click.option(
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -387,12 +387,14 @@ def create_external_models(obj: Context) -> None:
 
 
 @cli.command("table_diff")
-@click.argument("from_to", required=True, metavar="FROM:TO",)
-@click.argument("model_name", required=False)
+@click.argument("source_to_target", required=True, metavar="SOURCE:TARGET")
+@click.argument("model", required=False)
 @click.option(
+    "-O",
     "--on",
     type=str,
-    help='The SQL join condition or list of columns to use as keys. Table aliases must be "s" and "t" for source and target.',
+    multiple=True,
+    help="The column to join on. Can be specified multiple times. The model grain will be used if not specified.",
 )
 @click.option(
     "--where",
@@ -408,14 +410,14 @@ def create_external_models(obj: Context) -> None:
 @click.pass_obj
 @error_handler
 def table_diff(
-    obj: Context, from_to: str, model_name: t.Optional[str], **kwargs: t.Any
+    obj: Context, source_to_target: str, model: t.Optional[str], **kwargs: t.Any
 ) -> None:
     """Show the diff between two tables."""
-    source, target = from_to.split(":")
+    source, target = source_to_target.split(":")
     obj.table_diff(
         source=source,
         target=target,
-        model_or_snapshot=model_name,
+        model_or_snapshot=model,
         **kwargs,
     )
 
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -6,7 +6,6 @@ import sys
 import typing as t
 
 import click
-from sqlglot import exp
 
 from sqlmesh import debug_mode_enabled, enable_logging
 from sqlmesh.cli import error_handler
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -388,35 +388,16 @@ def create_external_models(obj: Context) -> None:
 
 
 @cli.command("table_diff")
-@click.option(
-    "--source",
-    "-s",
-    type=str,
+@click.argument(
+    "from_to",
     required=True,
-    help="The source environment or table.",
+    help="The source and target environment to diff. Format: source:target",
 )
-@click.option(
-    "--target",
-    "-t",
-    type=str,
-    required=True,
-    help="The target environment or table.",
-)
-@click.option(
-    "--grain",
-    type=str,
-    multiple=True,
-    help="The list of columns to use as keys.",
-)
+@click.argument("model_name", required=True)
 @click.option(
     "--on",
     type=str,
     help='The SQL join condition or list of columns to use as keys. Table aliases must be "s" and "t" for source and target.',
-)
-@click.option(
-    "--model",
-    type=str,
-    help="The model to diff against when source and target are environments and not tables.",
 )
 @click.option(
     "--where",
@@ -430,16 +411,19 @@ def create_external_models(obj: Context) -> None:
 )
 @click.pass_obj
 @error_handler
-def table_diff(obj: Context, **kwargs: t.Any) -> None:
-    """Show the diff between two tables.
-
-    Can either be two tables or two environments and a model.
-    """
-    kwargs["model_or_snapshot"] = kwargs.pop("model", None)
-    on = kwargs.pop("on", None)
-    grain = ensure_list(kwargs.pop("grain", None))
-    kwargs["on"] = exp.condition(on) if on else grain
-    obj.table_diff(**kwargs)
+def table_diff(
+    obj: Context, from_to: str, model_name: str, on: t.Optional[str], **kwargs: t.Any
+) -> None:
+    """Show the diff between two tables."""
+    model = obj.get_model(model_name, raise_if_missing=True)
+    source, target = from_to.split(":")
+    obj.table_diff(
+        source=source,
+        target=target,
+        model_or_snapshot=model,
+        on=exp.condition(on or " and ".join([f"s.{col} = t.{col}" for col in model.grain])),
+        **kwargs,
+    )
 
 
 @cli.command("prompt")
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -387,12 +387,8 @@ def create_external_models(obj: Context) -> None:
 
 
 @cli.command("table_diff")
-@click.argument(
-    "from_to",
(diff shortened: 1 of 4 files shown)


z3z1ma:
image

Example output improvement. Notice the color correlation from what is initially a resolved table name which is actually non-descript to what is the actual environment name during row diffs. This connects the dots and draws someone very quickly to value realization whilst still providing the actual evaluated snapshot id.

sqlmesh/cli/main.py
```diff
@@ -426,20 +406,24 @@ def create_external_models(obj: Context) -> None:
...
+    obj: Context, from_to: str, model_name: str, on: t.Optional[str], **kwargs: t.Any
+) -> None:
+    """Show the diff between two tables."""
+    model = obj.get_model(model_name, raise_if_missing=True)
+    source, target = from_to.split(":")
+    obj.table_diff(
+        source=source,
+        target=target,
+        model_or_snapshot=model,
+        on=exp.condition(on or " and ".join([f"s.{col} = t.{col}" for col in model.grain])),
```
z3z1ma, line 424: Lets evaluate assert either `on` or `model.grain` is truthy before this line so we can provide a nicer exception message to the user to easily resolve.

sqlmesh/core/console.py
```diff
@@ -482,10 +484,16 @@ def show_schema_diff(self, schema_diff: SchemaDiff) -> None:
...
+        self.console.print("\n[b]Row Count:[/b]")
+        self.console.print(
+            f" [yellow]{row_diff.source_env.upper()}[/yellow]: {row_diff.source_count} rows"
+        )
         self.console.print(
-            f"[bold]Row Count:[/bold] {row_diff.source}: {row_diff.source_count}, {row_diff.target}: {row_diff.target_count} -- {row_diff.count_pct_change}%"
+            f" [green]{row_diff.target_env.upper()}[/green]: {row_diff.target_count} rows"
         )
-        self.console.print(row_diff.sample.to_string(index=False))
+        self.console.print(f"\n[b]Rows Added/Removed[b]: {row_diff.count_pct_change:.1f}%")
```
z3z1ma, line 494: `Row Diff: N%` is more concise and accurate

sqlmesh/core/context.py
```diff
@@ -756,52 +756,58 @@ def table_diff(
         self,
         source: str,
         target: str,
+        model_or_snapshot: ModelOrSnapshot,
```
z3z1ma, line 759: The API simplification is predicated on tightening the input. We expect source and target to be environment names only, _not_ snapshot names. `model_or_snapshot` should really just be a model name.

While it might be less flexible, it seems far more likely a user will diff a model across environments than a user will diff 2 snapshots. They will not be able to easily obtain these snapshots ids. Comparing an orphaned snapshot (IE a snapshot not connected to an environment) probably would not work, therefore you get the same utility this way since it should always be relative to an environment. 
The difference is now the input/API is uniform.

sqlmesh/cli/main.py
```diff
@@ -388,36 +387,17 @@ def create_external_models(obj: Context) -> None:
...
-    help="The target environment or table.",
-)
-@click.option(
-    "--grain",
-    type=str,
-    multiple=True,
-    help="The list of columns to use as keys.",
+    metavar="ENV:ENV",
 )
+@click.argument("model_name", required=True)
```
tobymao, line 395: model_name shouldn't be required because you can diff two arbitrary tables

sqlmesh/cli/main.py
tobymao: you shouldn't have to do this for model.grain, it's already built into base table diff

sqlmesh/core/context.py
z3z1ma: Reverted, we can managed both so long as the simplified interface is intact
`sqlmesh table_diff dev:prod some.table`

will work just as well as 
`sqlmesh table_diff sushi__alex.top_waiters:sushi__toby.top_waiters -O waiter_id`

will work just as well as 
`sqlmesh table_diff sqlmesh__sushi.sushi__top_waiters__1307081719:sqlmesh__sushi.sushi__top_waiters__477763947 -O waiter_id`


Only in the first case is `grain` identifiable and should be the preferred pattern.

z3z1ma:
This syntax is ideal and now supported
`sqlmesh table_diff dev:prod sushi.top_waiters`

And this supports arbitrary tables
`sqlmesh table_diff sushi__dev.top_waiters:sushi__prod.top_waiters --on=waiter_id`

Can do this too
`sqlmesh table_diff sqlmesh__sushi.sushi__top_waiters__1307081719:sqlmesh__sushi.sushi__top_waiters__477763947 -O waiter_id`


LGTM now, all the benefits of the first syntax with the flexibility of the others preserved 
cc: @tobymao

sqlmesh/core/table_diff.py
```diff
@@ -83,14 +85,18 @@ def __init__(
...
         self.adapter = adapter
         self.source = source
         self.target = target
         self.where = exp.condition(where, dialect=dialect) if where else None
         self.limit = limit
+        self.source_env = source_env
+        self.target_env = target_env
 
-        if isinstance(on, list):
+        if isinstance(on, (list, tuple)):
```
z3z1ma, line 99: FYI `click.option` with `multiple=True` returns a **tuple**, hence this addition. The `multiple=True` is a nice, call it idiomatic, way to support passing a param one or more times without ambiguity

z3z1ma: Without this, the CLI -- even rebased on latest main, was still fundamentally broken for me because `--on` was just a string in the `kwargs` and it would never hit this code path. There was no way for it to be a list without `multiple` or custom parsing.
cc: @tobymao

sqlmesh/core/table_diff.py
```diff
@@ -52,6 +52,8 @@ class RowDiff(PydanticModel, frozen=True):
     target: str
     stats: t.Dict[str, float]
     sample: pd.DataFrame
+    source_env: str = "source"
```
izeigerman, line 55: Do we even need default values for source and env names? Should they just be optional?

sqlmesh/core/context.py
```diff
@@ -777,6 +777,7 @@ def table_diff(
         Returns:
             The TableDiff object containing schema and summary differences.
         """
+        source_env_name, target_env_name = "SOURCE", "TARGET"
```
izeigerman, line 780: ditto: why do we need these default values? It might be a little confusing to a user. It might be even more confusing even there are real environments named "source" and / or "target".

izeigerman (approved):
This change looks really really good! Not sure about default environment names.

sqlmesh/core/context.py
z3z1ma: good point

sqlmesh/core/table_diff.py
z3z1ma: same, resolved

z3z1ma:
> This change looks really really good! Not sure about default environment names.

Updated to make the intent more obvious and to drop the defaults @izeigerman

sqlmesh/cli/main.py
```diff
@@ -388,35 +386,14 @@ def create_external_models(obj: Context) -> None:
...
-    help="The target environment or table.",
-)
-@click.option(
-    "--grain",
-    type=str,
-    multiple=True,
-    help="The list of columns to use as keys.",
-)
-@click.option(
+    "-O",
```
tobymao, line 392: maybe just -o

izeigerman:
Thanks @z3z1ma! LGTM 🚢

sqlmesh/core/context.py
```diff
@@ -793,6 +794,8 @@ def table_diff(
             target = next(
                 snapshot for snapshot in target_env.snapshots if snapshot.name == model.name
             ).table_name()
+            source_alias = source_env.name
```
tobymao, line 797: what's the point of source_alias? i'm not sure i understand what this is

z3z1ma: because without it, you get ugly snapshot names using the syntax I prescribed. 
it is better to capture the environment name to connect the dots or its extremely non-obvious

sqlmesh/cli/main.py
z3z1ma: kk

sqlmesh/core/context.py
z3z1ma: see the user experience here thanks to the alias, otherwise those snapshot names mean nothing other than the assumption the left/top is source and right/bot is target

image

tobymao (approved):

sqlmesh/core/context.py
z3z1ma: Following the exact same logic,

image

I _could_ possibly make the case that the `s__{{column}}` could be better represented to the end user as `{{env-name}}__{{column}}` provided the env name was available and its not a direct diff of 2 tables/snapshots. (see above, not committed to this PR). 

Not a big deal really but something to consider. Its all about UX in this case

PR #979, opened 2023-06-13
Fix: Backfilling batch size for seeds should always be None

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -872,6 +872,10 @@ def seed_path(self) -> Path:
     def depends_on_past(self) -> bool:
         return False
 
+    @property
+    def batch_size(self) -> t.Optional[int]:
```
eakmanrq, line 876: Can you add a comment on this explaining why it must be none even though in other models we get the batch_size from the kind?

eakmanrq (approved):

PR #982, opened 2023-06-13
Chore: Refactor the snapshot evaluator

izeigerman:
Refactored the flag-driven implementation of the Snapshot Evaluator so that it relies on polymorphism instead.

This should help make subsequent iterations on this module a lot less error-prone. 

While refactoring I noticed a bug which impacted `INCREMENTAL_BY_UNIQUE_KEY` models in case when more than one DataFrame is returned. The first DataFrame was inserted using `merge` strategy while all subsequent inserts where appended. The new implementation made it very clear.


z3z1ma:
Very nice 💯 
I feel it greatly simplifies contribution as well.

PR #983, opened 2023-06-14
Fix: Convert seed's dataframe columns to the datetime object when their type is explicitly set to date / time

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -814,7 +816,26 @@ def render(
         **kwargs: t.Any,
     ) -> t.Generator[QueryOrDF, None, None]:
         self._ensure_hydrated()
-        yield from self.seed.read(batch_size=self.kind.batch_size)
+
+        date_or_time_columns = []
+        bool_columns = []
+        string_columns = []
+        for name, tpe in (self.columns_to_types_ or {}).items():
+            if tpe in DATE_TIME_TYPES:
```
tobymao, line 824: tpe in exp.DataType.TEMPORAL_TYPES

sqlmesh/core/model/definition.py
```diff
@@ -814,7 +816,26 @@ def render(
...
+                date_or_time_columns.append(name)
+            elif tpe == exp.DataType.build("boolean"):
```
tobymao, line 826: tpe.is_type("boolean")

sqlmesh/core/model/definition.py
```diff
@@ -814,7 +816,26 @@ def render(
...
+                bool_columns.append(name)
+            elif tpe in STRING_TYPES:
```
tobymao, line 828: tpe in exp.DataType.TEXT_TYPES

sqlmesh/core/model/definition.py
```diff
@@ -814,7 +816,26 @@ def render(
...
+                string_columns.append(name)
+
+        for df in self.seed.read(batch_size=self.kind.batch_size):
+            for column in date_or_time_columns:
+                df[column] = pd.to_datetime(df[column])
```
tobymao, line 833: this will only create date times right, what if you want to distinguish between date and datetime?

izeigerman: All time-like types (date, timestamp, etc.) are convertible from the pandas datetime type. Tested this behavior with PyArrow.

tobymao (approved):

PR #992, opened 2023-06-15
Fix: Render values stored in dbt variables before using them

izeigerman:


sqlmesh/dbt/builtin.py
```diff
@@ -144,9 +145,9 @@ def no_log(msg: str, info: bool = False) -> str:
     return ""
 
 
-def generate_var(variables: t.Dict[str, t.Any]) -> t.Callable:
+def generate_var(variables: t.Dict[str, t.Any], jinja_env: Environment) -> t.Callable:
     def var(name: str, default: t.Optional[str] = None) -> str:
-        return variables.get(name, default)
+        return jinja_env.from_string(str(variables.get(name, default))).render()
```
crericha, line 150: I don't think we can render vars this late. They may be using env_var and need the local environment. Can we render during loading?

izeigerman: Ah, excellent point! Where do you think would be a good place to render these variables?

crericha: How about here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/dbt/loader.py#L97?

When building the jinja environment, make sure macros in the same package can be accessed without pre-pending the package name. `package.macro_name` should also work as `macro_name`.

sqlmesh/dbt/context.py
```diff
@@ -85,6 +85,19 @@ def add_variables(self, variables: t.Dict[str, t.Any]) -> None:
         self._variables.update(variables)
         self._jinja_environment = None
 
+    def set_and_render_variables(self, variables: t.Dict[str, t.Any], package: str) -> None:
+        self.variables = variables
+
+        jinja_registry = JinjaMacroRegistry(
+            create_builtins_module="sqlmesh.dbt", top_level_packages=["dbt", package]
```
crericha, line 92: Maybe a constant here?

crericha: Also, should package precede "dbt" in the top_level_packages order?

izeigerman: Why does it matter?

crericha: If the macro is in both dbt and the top-level-package, which version of the macro is used?

izeigerman: actually if we want package macros to take precedence then the existing order is correct.

crericha: So the last item in the list gets precedence?

izeigerman: yes

crericha: Can we document that somewhere or reverse the list within JinjaMacroRegistry, so that its more intuitive?

sqlmesh/dbt/context.py
```diff
@@ -85,6 +85,19 @@ def add_variables(self, variables: t.Dict[str, t.Any]) -> None:
...
+        )
+        jinja_environment = jinja_registry.build_environment(**self.jinja_globals)
```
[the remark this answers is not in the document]
crericha: Discussed offline and going to stick with this approach for consistency.

PR #996, opened 2023-06-15
Feat: Optional rendered query and columns-to-types mapping

izeigerman:
To support model queries with jinja expressions that can only be rendered at runtime (as opposed to parse time) I refactored the model definition and made the rendered query and the columns-to-types mapping optional for sql models.

For context, the jinja expression can't be rendered at parse time if it requires connection to the data warehouse.

### DIFF
--- a/sqlmesh/core/model/cache.py
+++ b/sqlmesh/core/model/cache.py
@@ -13,7 +13,7 @@ from sqlmesh.utils.pydantic import PydanticModel
 
 class SqlModelCacheEntry(PydanticModel):
     model: SqlModel
-    rendered_query: exp.Expression
+    rendered_query: t.Optional[exp.Expression]
 
 
 class ModelCache:
@@ -45,7 +45,8 @@ class ModelCache:
         cache_entry = self._file_cache.get(name, entry_id)
         if cache_entry:
             model = cache_entry.model
-            model._query_renderer.update_cache(cache_entry.rendered_query, optimized=False)
+            if cache_entry.rendered_query is not None:
+                model._query_renderer.update_cache(cache_entry.rendered_query, optimized=False)
             return model
 
         loaded_model = loader()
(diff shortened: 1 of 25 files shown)


sqlmesh/core/model/cache.py
```diff
@@ -84,22 +85,27 @@ def with_optimized_query(self, model: Model) -> bool:
         if not isinstance(model, SqlModel):
             return True
 
-        entry_id = self._entry_id(model)
+        unoptimized_query = model.render_query(optimize=False)
+        if unoptimized_query is None:
+            return False
```
tobymao, line 90: i think you want return True here so that it doesn't go into validation, that's how i use this return value

izeigerman: It won't go into validation because I check whether the query is not none there. I think we shouldn't use this flag to bypass validation. It should instead indicate whether there was a cache hit.

tobymao: then you should fix the previous one that returns True when it shouldn't

tobymao (commented):
z

sqlmesh/dbt/builtin.py
```diff
@@ -306,45 +305,35 @@ def create_builtin_globals(
         {k: builtin_globals.get(k) for k in ("ref", "source", "config")}
     )
 
+    adapter: BaseAdapter
```
tobymao, line 308: just move this type def to line 310

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -364,15 +365,15 @@ def _create_snapshot(self, snapshot: Snapshot, snapshots: t.Dict[SnapshotId, Sna
 
         render_kwargs: t.Dict[str, t.Any] = dict(
             engine_adapter=self.adapter,
-            snapshots=snapshots,
+            snapshots=parent_snapshots_by_name,
```
crericha, line 368: For model validation, should we raise an error if we cannot render the model and `depends_on` isn't set? I think it makes sense to require `depends_on` to be set for models that can only be rendered at runtime.

crericha: Not related to my above comment, for runtime-only rendering models, we'll need to interrogate for parents once we can render, to pick-up any tables manually joined at runtime.

sqlmesh/dbt/adapter.py
```diff
@@ -96,35 +96,45 @@ def dispatch(self, name: str, package: t.Optional[str] = None) -> t.Callable:
 
 class ParsetimeAdapter(BaseAdapter):
     def get_relation(self, database: str, schema: str, identifier: str) -> t.Optional[BaseRelation]:
-        return None
+        self._raise_parstime_adapter_call_error("get relation")
```
crericha, line 99: typo--missing the e in parsetime

sqlmesh/core/snapshot/evaluator.py
izeigerman: > Not related to my above comment, for runtime-only rendering models, we'll need to interrogate for parents once we can render, to pick-up any tables manually joined at runtime.

It's too late at that point. We may not have necessary parents in the mapping. This is especially relevant in airflow.

tobymao (approved):

crericha (approved):

PR #1004, opened 2023-06-16
Feat: Support dbt partition_by and month time intervals.

crericha:


sqlmesh/dbt/model.py
```diff
@@ -98,6 +102,34 @@ def _validate_list(cls, v: t.Union[str, t.List[str]]) -> t.List[str]:
...
+    @root_validator
+    def _cron_interval_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+        # partition_by ignored if partitioned_by set
+        if values.get("partitioned_by") or not values.get("partition_by"):
+            return values
+
+        granularity = values["partition_by"]["granularity"]
+        if not values.get("cron"):
+            values["cron"] = timespan_to_cron_expr(granularity)
+            print(values)
```
GeorgeSittas, line 122: Forgot a print here

sqlmesh/core/model/meta.py
```diff
@@ -241,7 +244,9 @@ def interval_unit(self, sample_size: int = 10) -> IntervalUnit:
             croniter = CroniterCache(self.cron)
             samples = [croniter.get_next() for _ in range(sample_size)]
             min_interval = min(b - a for a, b in zip(samples, samples[1:]))
-            if min_interval >= 86400:
+            if min_interval >= 2419200:
+                self._interval_unit = IntervalUnit.MONTH
+            elif min_interval >= 86400:
                 self._interval_unit = IntervalUnit.DAY
             elif min_interval >= 3600:
                 self._interval_unit = IntervalUnit.HOUR
```
GeorgeSittas, line 252: Should we use constants here? E.g. `SECONDS_IN_MONTH`, etc.

sqlmesh/core/plan/definition.py
```diff
@@ -181,7 +181,7 @@ def _get_end_date(self, end_and_units: t.List[t.Tuple[int, IntervalUnit]]) -> Ti
         if end_and_units:
             end, unit = max(end_and_units)
 
-            if unit == IntervalUnit.DAY:
+            if unit in (IntervalUnit.DAY, IntervalUnit.MONTH):
```
GeorgeSittas, line 184: Should we factor this out in a variable? It's also used in `_format_date_time`.

sqlmesh/core/model/meta.py
crericha: It isn't used elsewhere, so I'm not sure how much value the constant would provide. I also didn't want to deviate from how the rest of the method worked.

GeorgeSittas: It's probably fine either way, just thought it might make it easier to read, compared to decoding that these numbers are the # of seconds in the corresponding unit. 🤷

PR #1007, opened 2023-06-17
Fix: Allow SELECT * in models that reference external tables

izeigerman:
Since the column-to-types mapping is now optional there doesn't seem to be much reason to prohibit `SELECT *` queries in models that reference external tables.

With this update we assume that we don't have a mapping available if we identify a `*` projection in the rendered / optimized query.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -652,7 +652,8 @@ def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Se
     """Find all tables referenced in a query.
 
     Args:
-        expressions: The list of expressions to find tables for.
+        expressions: The query to find the tables in.
+        dialect: The dialect to use for normalization of table names.
 
     Returns:
         A Set of all the table names.
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -10,6 +10,7 @@ import pandas as pd
 from sqlglot import Dialect, Generator, Parser, Tokenizer, TokenType, exp
 from sqlglot.dialects.dialect import DialectType
 from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
+from sqlglot.optimizer.scope import traverse_scope
 from sqlglot.tokens import Token
 
 from sqlmesh.core.constants import MAX_MODEL_DEFINITION_SIZE
@@ -644,4 +645,21 @@ def extract_columns_to_types(query: exp.Subqueryable) -> t.Dict[str, exp.DataTyp
     return {
         expression.output_name: expression.type or exp.DataType.build("unknown")
         for expression in query.selects
+    }
+
+
+def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Set[str]:
+    """Find all tables referenced in a query.
+
+    Args:
+        expressions: The list of expressions to find tables for.
+
+    Returns:
+        A Set of all the table names.
+    """
+    return {
+        normalize_model_name(table, dialect=dialect)
+        for scope in traverse_scope(expression)
+        for table in scope.tables
+        if isinstance(table.this, exp.Identifier) and exp.table_name(table) not in scope.cte_sources
     }
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -54,8 +54,12 @@ def update_model_schemas(
             continue
 
         model.update_schema(schema)
+
         cache_hit = optimized_query_cache.with_optimized_query(model)
-        schema.add_table(name, model.columns_to_types, dialect=model.dialect)
+
+        columns_to_types = model.columns_to_types
+        if columns_to_types is not None:
+            schema.add_table(name, columns_to_types, dialect=model.dialect)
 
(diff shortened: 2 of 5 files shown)


sqlmesh/core/loader.py
```diff
@@ -57,15 +57,7 @@ def update_model_schemas(
         cache_hit = optimized_query_cache.with_optimized_query(model)
         schema.add_table(name, model.columns_to_types, dialect=model.dialect)
 
-        if any(dep not in models for dep in model.depends_on):
```
tobymao, line 60: do you think this is a good opportunity to educate users about external models?

izeigerman: We should probably look for a different way, or just print a message instead of failing. Otherwise the platform's behavior is inconsistent: we fail if the target model can be rendered at parse time, but happily proceed if it can't. What do you think?

tobymao: yes i meant print a message

sqlmesh/core/model/definition.py
```diff
@@ -460,6 +460,18 @@ def update_schema(self, schema: MappingSchema) -> None:
                     tuple(str(part) for part in table.parts),
                     {k: str(v) for k, v in mapping_schema.items()},
                 )
+            else:
+                # Reset the entire mapping if at least one upstream dependency is missing from the mapping
+                # to prevent partial mappings from being used.
+                logger.warning(
```
izeigerman, line 466: @tobymao added a log message here.

izeigerman: Btw, this will be printed out even if `SQLMESH_DEBUG` is not set.

sqlmesh/core/renderer.py
```diff
@@ -325,10 +324,21 @@ def time_column_filter(self, start: TimeLike, end: TimeLike) -> exp.Between:
...
-        schema = ensure_schema(self.schema, dialect=self._dialect, normalize=False)
+        schema = MappingSchema(self.schema, dialect=self._dialect, normalize=False)
         original = query
         failure = False
 
+        if not schema.empty:
+            for dependency in d.find_tables(query, dialect=self._dialect):
+                if schema.find(exp.to_table(dependency, dialect=self._dialect)) is None:
+                    logger.warning(
+                        "Query cannot be optimized due to missing schema for model '%s'. "
```
eakmanrq, line 335: Can we explain more what we mean by "cannot be optimized"? I could see someone reading this and thinking "Oh that is ok my engine does optimizations anyways" and missing the real negative of this. Also I wonder if having a dedicated doc talking about this could be useful (but probably outside the scope of this PR).

izeigerman: I wonder what the real negative of this at runtime anyway?

izeigerman: At compile time we catch this earlier when we set the mapping schema as part of the `_Model.update_schema`.

sqlmesh/core/dialect.py
```diff
@@ -645,3 +646,20 @@ def extract_columns_to_types(query: exp.Subqueryable) -> t.Dict[str, exp.DataTyp
         expression.output_name: expression.type or exp.DataType.build("unknown")
         for expression in query.selects
     }
+
+
+def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Set[str]:
+    """Find all tables referenced in a query.
+
+    Args:
+        expressions: The list of expressions to find tables for.
```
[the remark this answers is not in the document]
GeorgeSittas: Haha @izeigerman sorry for nitpicking on this, there's still a trailing `s` in the docstring, i.e. `expression*s*` instead of `expression` (it doesn't match the actual parameter name).

tests/core/test_model.py
```diff
@@ -1227,7 +1229,7 @@ def test_case_sensitivity(assert_exp_eq):
             """
             MODEL (name example.source, kind EMBEDDED);
 
-            SELECT "id", "name", "payload" FROM db.schema."table"
+            SELECT 'id' AS "id", 'name' AS "name", 'payload' AS "payload"
```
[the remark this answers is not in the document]
izeigerman: In this test the `SOURCE` model gets expanded. Now we have a check in the renderer which skips optimization if at least one table is missing in the mapping. The `db.schema."table"` table was ending up in the rendered query as a result of the expansion but was missing in the mapping.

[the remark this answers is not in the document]
GeorgeSittas: Sounds good, thanks for clarifying 👍

sqlmesh/core/dialect.py
```diff
@@ -645,3 +646,21 @@ def extract_columns_to_types(query: exp.Subqueryable) -> t.Dict[str, exp.DataTyp
...
+
+    Args:
+        expressions: The query to find the tables in.
+        dialect: The dialect to use for normalization of table names.
+
+    Returns:
+        A Set of all the table names.
+    """
+    return {
+        normalize_model_name(table, dialect=dialect)
```
GeorgeSittas, line 662: Was there a bug here previously? How come we didn't normalize this table?

izeigerman: yep, it was a bug

PR #1011, opened 2023-06-19
Add success messages to context, init, and model magics

treysp:


izeigerman (approved):
@eakmanrq can you please take a look as well

sqlmesh/magics.py
```diff
@@ -53,6 +53,9 @@ def _context(self) -> Context:
             f"Context must be defined and initialized with one of these names: {', '.join(CONTEXT_VARIABLE_NAMES)}"
         )
 
+    def success_message(self, message: str) -> str:
+        return f'<span style="color:green; font-weight:bold">{message}</span>'
```
eakmanrq, line 57: Do you want to return an HTML object instead of a string? It seems like in all current usages you pass this returned value into `HTML` which made me wonder if you could skip that step.

eakmanrq (approved):
Thanks for improving this! One small suggestion.

sqlmesh/magics.py
treysp: That's how I initially had it, but I wasn't sure how to modify the HTML object contents to add paths for %context. 

This works, but Is it the pythonic way?

```
message = self.success_message("SQLMesh project context set to:")
message.data = message.data + "<br>&emsp;" + "<br>&emsp;".join(args.paths)
```

tobymao: do you mean using hyperscript? look at core/console.py for usage of hyperscript

treysp: Ok, updated it to use hyperscript

PR #1018, opened 2023-06-21
Fix: Improvements to depends_on_past behavior

eakmanrq:
This is a big PR. Might be best to pair with someone first on it to review and then after that session open the full PR. 

Some highlights:
* Add a cache for interval rows. 
* Add a cache for snapshots.
* When removing intervals for `depends_on_past` we only remove the intervals requested. `missing_intervals` was updated to understand that if a past interval is removed then future ones need to be considered "missing" and restated.
* Added a concept of "ignored models". Currently these are snapshots that are part of a plan and the plan's start date would result in the data to be invalid. The example we have today is a `depends_on_past` snapshot and the start date is before the start of the model and therefore the data is invalid. These snapshots are not included in the plan and the user is told when they exist.
* Updated missing_snapshots to use dev_intervals if appropriate and therefore when we merge intervals for a snapshot in the snapshot cache then it doesn't require doing checks to see if a merge should happen based on if it is a dev representation.

TODO:
* Update UI/API to represent invalid snapshots
* Update documentation to explain invalid snapshots
* Add tests for the cache


examples/sushi/models/items.py
```diff
@@ -50,7 +50,7 @@
 @model(
     "sushi.items",
     kind=IncrementalByTimeRangeKind(time_column="ds", batch_size=30),
-    start="3 months ago",
+    start="1 week ago",
```
tobymao, line 53: why change this? maybe just change the life time agg one to be less

sqlmesh/core/console.py
```diff
@@ -300,19 +314,27 @@ def plan(self, plan: Plan, auto_apply: bool) -> None:
         if auto_apply:
             plan.apply()
 
+    def _get_invalid_tree(self, invalid_snapshot_names: t.Set[str]) -> Tree:
+        invalid = Tree(f"[bold][invalid]Invalid Snapshots:")
```
tobymao, line 318: hm re invalid, perhaps ignored? they're not really invalid entirely, the usage may be concerning

sqlmesh/core/plan/definition.py
```diff
@@ -398,28 +432,61 @@ def _missing_intervals(self) -> t.Dict[str, Intervals]:
...
+        # We know that all downstream snapshots must exist since they can't be based on an external model since those
+        # only exist at the root level. Therefore we don't have to check for None here
+        return [self._get_snapshot(x) for x in self._dag.downstream(snapshot if isinstance(snapshot, str) else snapshot.name)]  # type: ignore
+
+    def _get_upstream_snapshots(self, snapshot: t.Union[str, Snapshot]) -> t.List[Snapshot]:
+        return [
+            snapshot
+            for snapshot in [
+                self._get_snapshot(snapshot_name)
+                for snapshot_name in self._dag.upstream(
```
tobymao, line 461: i think upstream is very inefficient, we'll need to make this method more efficient if you're going to use this in a forloop here

sqlmesh/core/plan/definition.py
```diff
@@ -398,28 +432,61 @@ def _missing_intervals(self) -> t.Dict[str, Intervals]:
...
             ]
 
-            if not downstream:
+            if not downstream_snapshots:
                 raise PlanError(
                     f"Cannot restate from '{table}'. Either such model doesn't exist or no other model references it."
                 )
-            self._restatements.update(downstream)
+            self._restatements.update(
+                [downstream_snapshot.name for downstream_snapshot in downstream_snapshots]
```
tobymao, line 488: is the array [ necessary here?

sqlmesh/core/snapshot/definition.py
```diff
@@ -127,6 +127,23 @@ class SnapshotIntervals(PydanticModel, frozen=True):
     def snapshot_id(self) -> SnapshotId:
         return SnapshotId(name=self.name, identifier=self.identifier)
 
+    @classmethod
+    def from_snapshot(cls, snapshot: Snapshot, **kwargs: t.Any) -> SnapshotIntervals:
+        return cls(
+            **{
+                **{
```
tobymao, line 134: is the double nesting necessary?

sqlmesh/core/snapshot/definition.py
```diff
@@ -375,6 +390,10 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
     change_category: t.Optional[SnapshotChangeCategory] = None
     unpaused_ts: t.Optional[int] = None
     effective_from: t.Optional[TimeLike] = None
+    # The alias also helps with backwards compatibility when reading snapshots from prior releases of
+    # SQLMesh where the intervals were included in the snapshot model itself.
+    intervals_: t.Optional[Intervals] = Field(default=None, alias="intervals")
```
tobymao, line 395: why do you need backwards compat? just run a migration?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -58,6 +58,108 @@
...
+        return all(snapshot.seed_is_hydrated for snapshot in self.values())
+
+    @property
+    def snapshot_intervals(self) -> t.List[SnapshotIntervals]:
+        return [snapshot.snapshot_intervals for snapshot in self.values()]
+
+    def get_cached_snapshots(self, snapshots: t.List[Snapshot]) -> t.List[Snapshot]:
+        cached_snapshots = []
+        for snapshot in snapshots:
+            self[snapshot.snapshot_id] = snapshot
```
tobymao, line 102: self.add_snapshot(snapshot)

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -58,6 +58,108 @@
...
+        for snapshot in self.snapshots:
+            snapshot.init_intervals()
+            assert snapshot.version
+            snapshot_intervals = interval_mapping.get((snapshot.name, snapshot.version), [])
+            for interval in snapshot_intervals:
+                snapshot.merge_intervals(interval)
+
+    def _set_snapshot(self, key: SnapshotId, value: Snapshot) -> None:
+        # Set all slots to None slots are private attributed and they are considered "ephemeral" and should not be
+        # persisted in cache. Currently this is just `_start` and ideally this is removed entirely
```
tobymao, line 156: let's do a fast follow up to get rid of this after this pr is merged

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,53 @@ def _environments_query(
...
         Args:
             snapshot_ids: The collection of snapshot like objects to fetch.
             lock_for_update: Lock the snapshot rows for future update
             hydrate_seeds: Whether to hydrate seed snapshots with the content.
-
+            hydrate_intervals: Whether to hydrate interval snapshots.
         Returns:
             A dictionary of snapshot ids to snapshots for ones that could be found.
         """
+        if snapshot_ids is not None and len(list(snapshot_ids)) == 0:
```
tobymao, line 461: snapshot_ids = list(snapshot_ids)

if not snapshot_ids:
return {}

sqlmesh/migrations/v0012_ensure_change_category.py
```diff
@@ -0,0 +1,24 @@
...
+
+    for name, identifier, snapshot in state_sync.engine_adapter.fetchall(
+        f"SELECT name, identifier, snapshot FROM {snapshots_table}"
+    ):
+        snapshot = json.loads(snapshot)
+        if snapshot.get("change_category"):
+            continue
+        # Just set it as breaking since we don't know
+        snapshot["change_category"] = 1
+        # this is not efficient, I'm doing this because I'm lazy and there shouldn't be a lot of snapshots out there
```
tobymao, line 17: lol you can only do that once

sqlmesh/core/plan/definition.py
```diff
@@ -238,13 +227,48 @@ def missing_intervals(self) -> t.List[SnapshotIntervals]:
...
         """Gets only new snapshots in the plan/environment."""
-        return list(self.context_diff.new_snapshots.values())
+        return [
+            snapshot
+            for snapshot in self.snapshots
+            if snapshot.snapshot_id in self.context_diff.new_snapshots
+        ]
+
+    @property
+    def has_changes(self) -> bool:
```
izeigerman, line 251: Seems a bit confusing that plan and context_diff have different notions of whether there are changes or not. Can we consolidate? Eg. can invalid snapshot names be moved into ContextDiff?

sqlmesh/core/plan/definition.py
```diff
@@ -238,13 +227,48 @@ def missing_intervals(self) -> t.List[SnapshotIntervals]:
...
+        return (
+            self.context_diff.is_new_environment
+            or self.context_diff.is_unfinalized_environment
+            or bool(modified_names)
+        )
+
+    def is_valid_snapshot(self, snapshot_name: t.Union[str, Snapshot]) -> bool:
+        """Checks if a snapshot is valid. Valid meaning that it is not being ignored in the plan"""
+        snapshot_name = snapshot_name if isinstance(snapshot_name, str) else snapshot_name.name
+        return snapshot_name in self._dag
```
izeigerman, line 271: The docstring claims:

> Valid meaning that it is not being ignored in the plan

But I don't see invalid snapshots being excluded from `self._dag` anywhere.

izeigerman: Ah, I take this back. It happens below.

sqlmesh/migrations/v0012_ensure_change_category.py
eakmanrq: I thought you would like this

sqlmesh/core/plan/definition.py
```diff
@@ -279,14 +303,26 @@ def loaded_snapshot_intervals(self) -> t.List[LoadedSnapshotIntervals]:
...
                 if downstream_snapshot.is_indirect_non_breaking:
                     continue
                 loaded_snapshots.append(LoadedSnapshotIntervals.from_snapshot(downstream_snapshot))
         return loaded_snapshots
 
+    def _refresh_dag_and_ignored_snapshots(self) -> None:
+        self.ignored_snapshot_names: t.Set[str] = set()
+        self._dag: DAG[str] = DAG()
+        snapshots = self.context_diff.snapshots.values()
+        for name, snapshot in self.context_diff.snapshots.items():
```
izeigerman, line 317: Seems like this iteration must happen in DAG order to ensure the correct behavior. Does `self.context_diff.snapshots` guarantee that order? Even if it does, I'm concerned that it's somewhat dangerous to assume that the dict is in correct order. I'd prefer this to be explicit.

sqlmesh/utils/date.py
```diff
@@ -21,6 +21,7 @@
 DATE_INT_FMT = "%Y%m%d"
 
 if t.TYPE_CHECKING:
+    from sqlmesh.core.model.meta import IntervalUnit
```
izeigerman, line 24: Should IntervalUnit just be moved into this module? Seems like a better place for it anyway + prevents utility package from depending on the core one. Btw, we should do the same to the `Interval` below.

sqlmesh/core/plan/definition.py
```diff
@@ -398,28 +432,61 @@ def _missing_intervals(self) -> t.Dict[str, Intervals]:
...
+        return self.context_diff.snapshots.get(snapshot_name)
+
+    def _get_downstream_snapshots(self, snapshot: t.Union[str, Snapshot]) -> t.List[Snapshot]:
+        # We know that all downstream snapshots must exist since they can't be based on an external model since those
+        # only exist at the root level. Therefore we don't have to check for None here
+        return [self._get_snapshot(x) for x in self._dag.downstream(snapshot if isinstance(snapshot, str) else snapshot.name)]  # type: ignore
+
+    def _get_upstream_snapshots(self, snapshot: t.Union[str, Snapshot]) -> t.List[Snapshot]:
+        return [
+            snapshot
```
izeigerman, line 458: [Nit] Seems like we overload `snapshot` variable quite a bit here. Somewhat hard to read.

sqlmesh/core/plan/definition.py
```diff
@@ -398,28 +432,61 @@ def _missing_intervals(self) -> t.Dict[str, Intervals]:
...
+            for snapshot in [
+                self._get_snapshot(snapshot_name)
+                for snapshot_name in self._dag.upstream(
+                    snapshot if isinstance(snapshot, str) else snapshot.name
```
izeigerman, line 462: This expression comes up a lot. Should there be a function? `_snapshot_name` or something

sqlmesh/core/plan/definition.py
```diff
@@ -398,28 +432,61 @@ def _missing_intervals(self) -> t.Dict[str, Intervals]:
...
-            downstream = self._dag.downstream(table)
-            if table in self.context_diff.snapshots:
-                downstream.append(table)
+            snapshot = self._get_snapshot(table)
+            if snapshot is None:
+                raise PlanError(
+                    f"Cannot restate from '{table}'. Either the model is an external model or doesn't exist"
+                )
+            downstream_snapshots = self._get_downstream_snapshots(table)
+            if self.is_valid_snapshot(table):
```
izeigerman, line 476: We check the target table explicitly here, but the `_get_downstream_snapshots` method doesn't seem to have this check. Is this expected?

izeigerman: Ah, it won't be in the DAG if it's invalid. Ignore this.

sqlmesh/core/snapshot/definition.py
```diff
@@ -127,6 +127,23 @@ class SnapshotIntervals(PydanticModel, frozen=True):
     def snapshot_id(self) -> SnapshotId:
         return SnapshotId(name=self.name, identifier=self.identifier)
 
+    @classmethod
+    def from_snapshot(cls, snapshot: Snapshot, **kwargs: t.Any) -> SnapshotIntervals:
+        return cls(
```
izeigerman, line 132: Why not use `snapshot.copy(update=kwargs)`? https://docs.pydantic.dev/latest/usage/exporting_models/#modelcopy

sqlmesh/core/snapshot/definition.py
```diff
@@ -482,36 +436,40 @@ def from_model(
...
-                audits=audits,
-                cache=cache,
-            ),
-            model=model,
-            parents=tuple(
-                SnapshotId(
-                    name=name,
-                    identifier=fingerprint_from_model(
-                        models[name],
+            **{
```
izeigerman, line 440: Why 2 levels?

sqlmesh/core/snapshot/definition.py
```diff
@@ -751,6 +727,61 @@ def version_get_or_generate(self) -> str:
...
+        """Checks if the given start and end are valid for this snapshot.
+
+        Args:
+            start: The start date/time of the interval (inclusive)
+            snapshot_default_start: The default start date/time of the snapshot if one is not defined (inclusive)
+        """
+        # The snapshot may not have a start defined. If so we use the provided default start.
+        snapshot_start = self.start or snapshot_default_start
+        if self.depends_on_past and start:
+            assert snapshot_start, "Snapshot must have a start defined if it depends on past"
```
izeigerman, line 742: Should this be an exception instead?

sqlmesh/core/snapshot/definition.py
```diff
@@ -751,6 +727,61 @@ def version_get_or_generate(self) -> str:
...
+            start_ts = to_timestamp(self.model.cron_floor(start))
+            if not self.intervals and start_ts > to_timestamp(snapshot_start):
```
izeigerman, line 744: Should this be `>=` instead? Start is inclusive

sqlmesh/core/snapshot/definition.py
```diff
@@ -751,6 +727,61 @@ def version_get_or_generate(self) -> str:
...
+                return False
+            # Make sure that if there are missing intervals for this snapshot that they all occur at or after the
+            # provided start_ts. Otherwise we know that we are doing a non-contiguous load and therefore this is not
+            # a valid start.
+            missing_intervals = self.missing_intervals(snapshot_start, make_inclusive_end(now()))
+            earliest_interval = missing_intervals[0][0] if missing_intervals else None
+            if earliest_interval:
+                return earliest_interval >= start_ts
```
izeigerman, line 752: This one is correct.

sqlmesh/core/snapshot/definition.py
izeigerman: Ah, if it's `==` then we should return True. Makes sense.

izeigerman, line 744: Should this be:
```
if not self.intervals:
    return to_timestamp(snapshot_start) >= start_ts
```
What's the point of proceeding to missing intervals otherwise.

sqlmesh/core/snapshot/definition.py
```diff
@@ -127,6 +127,23 @@ class SnapshotIntervals(PydanticModel, frozen=True):
...
 
+    @classmethod
+    def from_snapshot(cls, snapshot: Snapshot, **kwargs: t.Any) -> SnapshotIntervals:
+        return cls(
+            **{
+                **{
+                    "name": snapshot.name,
+                    "identifier": snapshot.identifier,
+                    "version": snapshot.version,
+                    "intervals": snapshot.intervals.copy() if snapshot.intervals_ else [],
```
izeigerman, line 138: If `snapshot.intervals_ ` is None that means we are copying a snapshot that's not hydrated and make it hydrated instead. Is this logic correct?

sqlmesh/core/snapshot/definition.py
```diff
@@ -751,6 +727,61 @@ def version_get_or_generate(self) -> str:
...
+        return self.intervals_
+
+    @property
+    def dev_intervals(self) -> Intervals:
+        """Gets the dev intervals for a Snapshot. Asserts that intervals have been hydrated."""
+        assert self.dev_intervals_ is not None, "Intervals have not been hydrated"
+        return self.dev_intervals_
+
+    @property
+    def seed_is_hydrated(self) -> bool:
```
izeigerman, line 779: [Nit] Maybe `is_seed_hydrated`?

sqlmesh/core/snapshot/definition.py
```diff
@@ -683,7 +650,16 @@ def missing_intervals(
             )
             compare_ts = seq_get(dates, i + lookback) or dates[-1]
 
-            for low, high in self.intervals:
+            #  in missing intervals check if is and forward only and paused and if so use dev intervals
```
izeigerman, line 653: >  if is and forward only and paused

not sure I fully understand the sentence

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -58,6 +58,108 @@
...
+
+    @property
+    def is_hydrated(self) -> bool:
+        return all(snapshot.seed_is_hydrated for snapshot in self.values())
+
+    @property
+    def snapshot_intervals(self) -> t.List[SnapshotIntervals]:
+        return [snapshot.snapshot_intervals for snapshot in self.values()]
+
+    def get_cached_snapshots(self, snapshots: t.List[Snapshot]) -> t.List[Snapshot]:
```
izeigerman, line 99: It seems counterintuitive that we mutate state in what appears to be a getter method. Should this be called `update_and_get...`?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -58,6 +58,108 @@
...
+
+    def merge_intervals_by_name_version(
+        self, intervals: t.Optional[t.Iterable[SnapshotIntervals]] = None
+    ) -> None:
+        interval_mapping: t.Dict[t.Tuple[str, str], t.List[SnapshotIntervals]] = defaultdict(list)
+        intervals = intervals or self.snapshot_intervals
+        for interval in intervals:
+            interval_mapping[(interval.name, interval.version)].append(interval)
+        for snapshot in self.snapshots:
+            snapshot.init_intervals()
```
izeigerman, line 148: Should the init happen lazily as part of `merge_intervals`? Not sure why do we need a dedicated method for this

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -58,6 +58,108 @@
...
+
+    def __getitem__(self, key: SnapshotId) -> Snapshot:
+        return self._cache[key]
+
+    def __setitem__(self, key: SnapshotId, value: Snapshot) -> None:
+        existing_snapshot = self.get(key)
+        if existing_snapshot:
+            value = value if value.updated_ts > existing_snapshot.updated_ts else existing_snapshot
+        self._set_snapshot(key, value)
+        self.merge_intervals_by_name_version()
```
izeigerman, line 75: Why do we merge intervals for **all** snapshots as opposed to only for those that share the version with the target one?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
             hydrate_seeds: Whether to hydrate seed snapshots with the content.
-
+            hydrate_intervals: Whether to hydrate interval snapshots.
         Returns:
             A dictionary of snapshot ids to snapshots for ones that could be found.
         """
+        snapshot_ids = list(snapshot_ids) if snapshot_ids else []
+        if not snapshot_ids:
+            return {}
+        if self._snapshot_cache and self._snapshot_cache.contains(snapshot_ids or set()):
```
izeigerman, line 464: Why is it all or nothing? Can some IDs be present in the cache but not the others?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
             lock_for_update: Lock the snapshot rows for future update
             hydrate_seeds: Whether to hydrate seed snapshots with the content.
-
+            hydrate_intervals: Whether to hydrate interval snapshots.
         Returns:
             A dictionary of snapshot ids to snapshots for ones that could be found.
         """
+        snapshot_ids = list(snapshot_ids) if snapshot_ids else []
+        if not snapshot_ids:
+            return {}
```
izeigerman, line 463: What if I'm fetching **all** available snapshots? Like in case of migration and Airflow

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
+        if self._snapshot_cache and self._snapshot_cache.contains(snapshot_ids or set()):
+            if hydrate_seeds:
+                self._hydrate_seeds(snapshot_ids)
+            if hydrate_intervals:
+                self._snapshot_cache.merge_intervals_by_name_version(
```
izeigerman, line 468: Why do we ignore `intervals_by_id` here?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -354,14 +487,22 @@ def _get_snapshots(
...
-            if hydrate_seeds and isinstance(snapshot.model, SeedModel) and row[1]:
-                snapshot.model = snapshot.model.to_hydrated(row[1])
+        self._snapshot_cache = SnapshotCache(snapshots)
+        if hydrate_seeds:
+            self._hydrate_seeds(snapshots)
+        if hydrate_intervals and not intervals_by_id:
+            self._snapshot_cache.merge_intervals_by_name_version(
+                self._get_snapshot_intervals(snapshot_ids)[1]
+            )
+        if hydrate_intervals and intervals_by_id:
```
izeigerman, line 502: To ensure that we don't hydrate twice I'd make the branching more explicit:

```
if hydrate_intervals:
    if intervals_by_id:
    else:
```

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -397,7 +538,9 @@ def _get_snapshots_with_same_version(
             query = query.lock(copy=False)
 
         snapshot_rows = self.engine_adapter.fetchall(query, ignore_unsupported_errors=True)
-        return [Snapshot(**json.loads(row[0])) for row in snapshot_rows]
+        return self._snapshot_cache.get_cached_snapshots(
```
izeigerman, line 541: The name is very misleading here.

izeigerman: Also it seems like if snapshots were deleted from the state we'd still return them here if they happened to be in the cache.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
         """
+        snapshot_ids = list(snapshot_ids) if snapshot_ids else []
+        if not snapshot_ids:
+            return {}
+        if self._snapshot_cache and self._snapshot_cache.contains(snapshot_ids or set()):
+            if hydrate_seeds:
+                self._hydrate_seeds(snapshot_ids)
+            if hydrate_intervals:
+                self._snapshot_cache.merge_intervals_by_name_version(
+                    self._get_snapshot_intervals(snapshot_ids)[1]
```
izeigerman, line 469: Why do we fetch intervals on cache hit?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
+        query = (
+            exp.select("name", "identifier", "content")
+            .from_(self.seeds_table)
+            .where(None if not snapshot_ids else self._snapshot_id_filter(snapshot_ids))
+        )
+        for name, identifier, content in self.engine_adapter.fetchall(query):
+            snapshot = self._snapshot_cache.get(SnapshotId(name=name, identifier=identifier))
+            if snapshot and not snapshot.seed_is_hydrated:
+                assert isinstance(snapshot.model, SeedModel), "Must be seed model to hydrate"
+                snapshot.model = snapshot.model.to_hydrated(content)
```
izeigerman, line 441: But we cache the hydrated version.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -317,32 +428,54 @@ def _environments_query(
...
 
+    def _hydrate_seeds(self, snapshot_ids: t.Optional[t.Iterable[SnapshotIdLike]] = None) -> None:
+        query = (
+            exp.select("name", "identifier", "content")
+            .from_(self.seeds_table)
+            .where(None if not snapshot_ids else self._snapshot_id_filter(snapshot_ids))
+        )
+        for name, identifier, content in self.engine_adapter.fetchall(query):
+            snapshot = self._snapshot_cache.get(SnapshotId(name=name, identifier=identifier))
+            if snapshot and not snapshot.seed_is_hydrated:
```
izeigerman, line 439: Should we check whether the cached version is hydrated prior to fetching its contents?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -511,16 +658,19 @@ def query_modifier(query: exp.Select) -> exp.Select:
...
+        self,
+        snapshot_ids: t.Optional[t.Iterable[SnapshotIdLike]] = None,
+        query_modifier: t.Optional[t.Callable[[exp.Select], exp.Select]] = None,
     ) -> t.Tuple[t.Set[str], t.List[SnapshotIntervals]]:
         query = (
             exp.select(
                 "id", "name", "identifier", "version", "start_ts", "end_ts", "is_dev", "is_removed"
             )
             .from_(self.intervals_table)
+            .where(None if not snapshot_ids else self._snapshot_id_filter(snapshot_ids))
```
izeigerman, line 673: I think we shouldn't fetch intervals if the `snapshot_ids` list is empty. That might be an error that we'd mask.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -564,11 +714,18 @@ def _push_snapshot_intervals(
         new_intervals = []
         for snapshot in snapshots:
             logger.info("Pushing intervals for snapshot %s", snapshot.snapshot_id)
+            cached_snapshot = self._snapshot_cache.get(snapshot.snapshot_id)
+            if not cached_snapshot and isinstance(snapshot, Snapshot):
+                cached_snapshot = self._snapshot_cache[snapshot.snapshot_id] = snapshot
             for start_ts, end_ts in snapshot.intervals:
+                if cached_snapshot:
+                    cached_snapshot.add_interval(start_ts, end_ts, is_dev=False)
```
izeigerman, line 722: So if the given instance is `Snapshot` we'd add its own intervals to it? And cause merge again?

sqlmesh/migrations/v0012_ensure_change_category.py
```diff
@@ -0,0 +1,24 @@
+import json
```
izeigerman, line 1: There's already `v0012` migration in the repo.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -837,7 +991,7 @@ def _snapshots_to_df(snapshots: t.Iterable[Snapshot]) -> pd.DataFrame:
                 "name": snapshot.name,
                 "identifier": snapshot.identifier,
                 "version": snapshot.version,
-                "snapshot": snapshot.json(exclude={"intervals", "dev_intervals"}),
+                "snapshot": snapshot.json(exclude={"intervals_", "dev_intervals_"}),
```
izeigerman, line 994: Are you sure that we shouldn't be referring to these by their aliases?

izeigerman: Just checked: your version is correct.

sqlmesh/migrations/v0012_ensure_change_category.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def migrate(state_sync):  # type: ignore
```
izeigerman, line 4: Why do you need this script? Aren't we doing this already as part of the `_migrate_rows`?
https://github.com/TobikoData/sqlmesh/blob/2e648f641637d3188eb6e571857d6fb3ab2b7342/sqlmesh/core/state_sync/engine_adapter.py#L842-L849

sqlmesh/migrations/v0012_ensure_change_category.py
```diff
@@ -0,0 +1,24 @@
...
+    snapshots_table = f"{state_sync.schema}._snapshots"
+
+    for name, identifier, snapshot in state_sync.engine_adapter.fetchall(
+        f"SELECT name, identifier, snapshot FROM {snapshots_table}"
+    ):
+        snapshot = json.loads(snapshot)
+        if snapshot.get("change_category"):
+            continue
+        # Just set it as breaking since we don't know
+        snapshot["change_category"] = 1
```
izeigerman, line 16: I don't think this logic is at all correct. See the snippet I provided above.

sqlmesh/utils/date.py
```diff
@@ -273,3 +274,15 @@ def time_like_to_str(time_like: TimeLike) -> str:
...
         return to_ds(time_like)
     return to_datetime(time_like).isoformat()
+
+
+def to_end_date(
+    end_and_units: t.Union[t.Tuple[int, IntervalUnit], t.List[t.Tuple[int, IntervalUnit]]]
+) -> TimeLike:
+    from sqlmesh.core.model.meta import IntervalUnit
+
+    end_and_units = [end_and_units] if isinstance(end_and_units, tuple) else end_and_units
```
izeigerman, line 284: I thought there was `ensure_list` function somewhere.

sqlmesh/utils/date.py
```diff
@@ -273,3 +274,15 @@ def time_like_to_str(time_like: TimeLike) -> str:
...
+    end, unit = max(end_and_units)
+    if unit == IntervalUnit.DAY:
```
izeigerman, line 286: CC: @crericha I assume we'd need the same thing for `MONTH`, `WEEK` and `YEAR` units.

sqlmesh/utils/rich.py
```diff
@@ -11,6 +11,7 @@
         "direct": "magenta",  # directly modified
         "indirect": "yellow",  # indirectly modified
         "metadata": "cyan",  # metadata updated
+        "ignored": "grey37",  # invalid snapshot
```
izeigerman, line 14: [Nit] `invalid` in the comment

sqlmesh/core/console.py
```diff
@@ -252,20 +260,24 @@ def show_model_difference_summary(
         if context_diff.added:
             added_tree = Tree(f"[bold][added]Added Models:")
             for model in context_diff.added:
-                added_tree.add(f"[added]{model}")
+                if model not in ignored_snapshot_names:
```
izeigerman, line 263: [Nit] So if all added models happened to be ignored we'd still show the `Added Models` header.

sqlmesh/core/console.py
```diff
@@ -252,20 +260,24 @@ def show_model_difference_summary(
...
+                    added_tree.add(f"[added]{model}")
             tree.add(added_tree)
 
         if context_diff.removed:
             removed_tree = Tree(f"[bold][removed]Removed Models:")
             for model in context_diff.removed:
-                removed_tree.add(f"[removed]{model}")
+                if model not in ignored_snapshot_names:
```
izeigerman, line 270: Ditto

sqlmesh/core/console.py
```diff
@@ -252,20 +260,24 @@ def show_model_difference_summary(
...
+                    removed_tree.add(f"[removed]{model}")
             tree.add(removed_tree)
 
         if context_diff.modified_snapshots:
             direct = Tree(f"[bold][direct]Directly Modified:")
             indirect = Tree(f"[bold][indirect]Indirectly Modified:")
             metadata = Tree(f"[bold][metadata]Metadata Updated:")
             for model in context_diff.modified_snapshots:
+                if model in ignored_snapshot_names:
```
izeigerman, line 279: Ditto

sqlmesh/core/console.py
```diff
@@ -903,6 +945,12 @@ def show_model_difference_summary(
                 for model in metadata_modified:
                     self._print(f"- `{model}`\n")
                 self._print("\n")
+        if ignored_snapshot_names:
+            self._print(f"**Ignored Models (Valid Plan Start):**\n")
```
izeigerman, line 949: I was thinking about this one: how about `Expected` instead of `Valid`?

sqlmesh/core/plan/definition.py
eakmanrq: Yeah great catch I completely agree. I will enforce the order.

sqlmesh/core/snapshot/definition.py
```diff
@@ -375,6 +390,10 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
     change_category: t.Optional[SnapshotChangeCategory] = None
     unpaused_ts: t.Optional[int] = None
     effective_from: t.Optional[TimeLike] = None
+    # The alias also helps with backwards compatibility when reading snapshots from prior releases of
+    # SQLMesh where the intervals were included in the snapshot model itself.
+    intervals_: t.Optional[Intervals] = Field(default=None, alias="intervals")
+    dev_intervals_: t.Optional[Intervals] = Field(default=None, alias="dev_intervals")
```
izeigerman, line 396: I don't see `def merge_intervals` being updated to include `dev_intervals`.

izeigerman: Ah, it seems like I did add it after all

sqlmesh/core/plan/definition.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/1018/commits/b4bdfffb254d51b42e82e504c92b61b9372acb51

sqlmesh/core/snapshot/definition.py
eakmanrq: that was a note to myself. thanks for catching.

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: This concern is purely performance related correct? There is not a functional issue with the current approach? I agree it is wasteful to merge everything when we know what was updated.

sqlmesh/core/plan/definition.py
eakmanrq: Yes very confusing. Thanks for catching this.

sqlmesh/utils/date.py
eakmanrq: This is a confusing nuance of that function. If you pass it a tuple then it will actually return a list of the tuple values instead of a list with a single value being the tuple. I want a list with the tuple being a value so I do this here.

sqlmesh/core/console.py
eakmanrq: Thanks for catching these

sqlmesh/core/snapshot/definition.py
```diff
@@ -127,6 +127,23 @@ class SnapshotIntervals(PydanticModel, frozen=True):
...
         return SnapshotId(name=self.name, identifier=self.identifier)
 
+    @classmethod
+    def from_snapshot(cls, snapshot: Snapshot, **kwargs: t.Any) -> SnapshotIntervals:
+        return cls(
+            **{
+                **{
+                    "name": snapshot.name,
+                    "identifier": snapshot.identifier,
+                    "version": snapshot.version,
```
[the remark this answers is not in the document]
eakmanrq: We now ensure categorized with this change: https://github.com/TobikoData/sqlmesh/pull/1018/commits/3323361b95251e980a2b3fea7244a3c23840e779

sqlmesh/core/snapshot/definition.py
eakmanrq: Intervals are now empty list by default which removes this if/else logic: https://github.com/TobikoData/sqlmesh/pull/1018/commits/3323361b95251e980a2b3fea7244a3c23840e779

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: Removed since intervals are now empty list by default: https://github.com/TobikoData/sqlmesh/pull/1018/commits/3323361b95251e980a2b3fea7244a3c23840e779

sqlmesh/core/snapshot/definition.py
```diff
@@ -751,6 +727,61 @@ def version_get_or_generate(self) -> str:
...
+                (to_timestamp(max(x[1] for x in self.intervals)), self.model.interval_unit())
+            )
+            if self.intervals
+            else self.start
+        )
+
+    @property
+    def intervals(self) -> Intervals:
+        """Gets the intervals for a Snapshot. Asserts that intervals have been hydrated."""
+        assert self.intervals_ is not None, "Intervals have not been hydrated"
```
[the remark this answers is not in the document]
eakmanrq: Intervals are now empty list by default which removes this assert: https://github.com/TobikoData/sqlmesh/pull/1018/commits/3323361b95251e980a2b3fea7244a3c23840e779

eakmanrq:
Deprecated by: https://github.com/TobikoData/sqlmesh/pull/1186

PR #1019, opened 2023-06-21
Feat: Code editor enhancement

mykalmax:


web/client/src/library/components/editor/hooks.ts
```diff
@@ -0,0 +1,383 @@
...
+        debouncedSaveChange()
+
+        return true
+      },
+    },
+  ]
+}
+
+const cache = new Map<string, (e: MessageEvent) => void>()
+const WHITE_SPACE = ' '
```
vchan, line 193: Should these be closer to the top of the file?

vchan (approved):

PR #1028, opened 2023-06-23
Fix: force quoting on unsafe table identifiers

GeorgeSittas:


sqlmesh/core/dialect.py
```diff
@@ -635,9 +635,10 @@ def pandas_to_sql(
 
 
 def normalize_model_name(table: str | exp.Table, dialect: DialectType = None) -> str:
-    return exp.table_name(
-        normalize_identifiers(exp.to_table(table, dialect=dialect), dialect=dialect)
-    )
+    table = exp.to_table(table, dialect=dialect)
+    name = exp.table_name(normalize_identifiers(table, dialect=dialect))
+    quoted = any(not exp.SAFE_IDENTIFIER_RE.match(p.name) for p in table.parts if p) or None
```
GeorgeSittas, line 640: Should we move this logic in SQLGlot to ensure `table_name` will return a quoted identifier if there's an unsafe "part" in it? This would require adding a `dialect` argument to it, in order to generate the correct identifer delimiters.

tobymao: ok

GeorgeSittas: xref: https://github.com/tobymao/sqlglot/pull/1820

eakmanrq (approved):
This fixed the issue I was seeing on Databricks. Thanks!

PR #1030, opened 2023-06-24
Fix: Replace tables after optimizing the query

izeigerman:


tests/core/test_model.py
```diff
@@ -1174,9 +1174,9 @@ def test_star_expansion(assert_exp_eq) -> None:
           "model1"."ds" AS "ds"
         FROM (
           SELECT
-            CAST("t"."id" AS INT) AS "id",
-            CAST("t"."item_id" AS INT) AS "item_id",
-            CAST("t"."ds" AS TEXT) AS "ds"
+            CAST("id" AS INT) AS "id",
```
izeigerman, line 1177: @tobymao I have no idea why this had to change.

tobymao: that means the optimizer broke

izeigerman: But the outter qualified columns are ok

tobymao (approved):

PR #1031, opened 2023-06-25
Fix: Disregard column order when comparing DataFrames in tests

### LINKED ISSUE #1023: Unit test expected vs actual shouldn't be dependent on column ordering
Unit tests will fail if the order of how columns are defined differs from the contents of the expected result .

vchan:
Fixes #1023


crericha (approved):
Thanks, @vchan!

PR #1032, opened 2023-06-25
Feat: Add schema map config

z3z1ma:
Not sure what method to use to render a query where this is applied. Once I know this, we can extend the test to validate the mapping is used, and is otherwise a very simple change.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -60,7 +60,7 @@ from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.environment import Environment
 from sqlmesh.core.loader import Loader, SqlMeshLoader, update_model_schemas
 from sqlmesh.core.macros import ExecutableOrMacro
-from sqlmesh.core.model import Model
+from sqlmesh.core.model import Model, parse_model_name
 from sqlmesh.core.model.definition import _Model
 from sqlmesh.core.plan import Plan
 from sqlmesh.core.scheduler import Scheduler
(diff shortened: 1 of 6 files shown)


tobymao:
you need to add this field to the data hash since changing the mapping should trigger a backfill

eakmanrq:
@z3z1ma Can you add this feature to the documentation? My preference is that it is framed as more of an advanced feature and give some examples of where it would be useful. Open to feedback on that though.

izeigerman:
@z3z1ma sorry, I'm not sure I understand your concern here. There is no concept of a "new" model but there is a concept of a "new" snapshot. So whatever you put into a snapshot constitutes a change. 

The way we detect changes is by computing the fingerprint of a **snapshot** which of course includes a model among other things but also covers audits, macros, etc. So I believe putting a physical schema in there should work perfectly fine unless I'm missing something.

sqlmesh/core/context.py
```diff
@@ -525,6 +526,7 @@ def snapshots(self) -> t.Dict[str, Snapshot]:
                 config = self.config_for_model(model)
                 ttl = config.snapshot_ttl
                 project = config.project
+                physical_schema = config.physical_schema_map.get(parse_model_name(model.name)[1])
```
izeigerman, line 529: Shouldn't we inherit the `physical_schema` from the remote snapshot if one was found in the other branch?

izeigerman: So I think you should set `physical_schema` to `config.physical_schema_map.get(parse_model_name(model.name)[1])` regardless of whether the `model.name` is in `remote_snapshots` or not. Which also means you can get rid of `physical_schema = None`.

tobymao:
closing this for now, we may pick this up later, or you're free to @alex-harness

PR #1033, opened 2023-06-25
Fix: add pseudo columns for ingest time partitioned tables

z3z1ma:
Bigquery has pseudo columns for ingest time partitioned tables. We should check and add these. Its quite easy since we have the Table API response already. This helps the optimizer not fail on valid queries and also allows users not to have to manually mess with the schema yaml file to add these columns otherwise.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -101,11 +101,18 @@ class BigQueryEngineAdapter(EngineAdapter):
 
     def columns(self, table_name: TableName) -> t.Dict[str, exp.DataType]:
         """Fetches column names and types for the target table."""
+        from google.cloud.bigquery import TimePartitioningType
+
         table = self._get_table(table_name)
-        return {
+        columns = {
             field.name: exp.DataType.build(field.field_type, dialect=self.dialect)
             for field in table.schema
         }
+        if table.time_partitioning and not table.time_partitioning.field:
+            columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP")
+            if table.time_partitioning.type_ == TimePartitioningType.DAY:
+                columns["_PARTITIONDATE"] = exp.DataType.build("DATE")
+        return columns
 
     def fetchone(
         self,


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -101,11 +101,18 @@ def create_schema(self, schema_name: str, ignore_if_exists: bool = True) -> None
...
+
         table = self._get_table(table_name)
-        return {
+        columns = {
             field.name: exp.DataType.build(field.field_type, dialect=self.dialect)
             for field in table.schema
         }
+        if table.time_partitioning and not table.time_partitioning.field:
+            columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP")
+            if table.time_partitioning.type_ == TimePartitioningType.DAY:
```
eakmanrq, line 113: Is it the case that all pseudo column time partitioned tables have a `_PARTITIONTIME` column but possibly not a `_PARTITIONDATE` field? So there is no way to have a pseduo column that is date only?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -101,11 +101,18 @@ def create_schema(self, schema_name: str, ignore_if_exists: bool = True) -> None
 
     def columns(self, table_name: TableName) -> t.Dict[str, exp.DataType]:
```
eakmanrq, line 102: Can you add a boolean param to this method called `include_pseudo_columns` (including the base class and all those who override it) and default it to `True`? I wonder if we have some cases where returning pseudo columns is not what we want/expect and if we add this now it will be easy to update those calls to exclude it as needed. It also makes it clear from the function signature that we are returning columns like this if they are defined. The one negative that I see is that other engines may have pseudo columns that we just haven't implemented retrieving them but we can follow up on that later if needed.m

eakmanrq (commented):
Nice fix! Just two small things to follow up on.

sqlmesh/core/engine_adapter/bigquery.py
z3z1ma: Yep, that is correct!

https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table

sqlmesh/core/engine_adapter/bigquery.py
z3z1ma: Makes sense. Done ✅

z3z1ma:
@eakmanrq Doing a scan of the codebase, I am going to suggest we inverse this relationship and specifically ask pseudocolumns to be included. Because there is only one place right now we want them.

z3z1ma:
This is peripherally related
https://github.com/tobymao/sqlglot/pull/1826

If we put `_PARTITIONTIME/DATE` columns in the schema for users with partitioned tables, we need to make sure sqlglot doesn't expand them in a `*` expansion. This issue _could_ affect me today but I don't do `select *` on any of my partitioned tables, hence why the optimizer has not tripped me up. 

These 2 PRs in concert make a better/safer BigQuery experience for ingestion time partitioning, which is pretty popular; even in the event of a `select *`.

eakmanrq (approved):
Thanks!

PR #1034, opened 2023-06-26
Feat: Add ability to invalidate the environment

izeigerman:
This update allows user to invalidate the target environment by setting its expiration timestamp to now, forcing it to be garbage collected during a subsequent run of the janitor process.

CC: @z3z1ma

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -18,7 +18,6 @@ from sqlmesh.core import constants as c
 from sqlmesh.core.console import MarkdownConsole
 from sqlmesh.core.context import Context
 from sqlmesh.core.environment import Environment
-from sqlmesh.core.model import parse_model_name
 from sqlmesh.core.plan import LoadedSnapshotIntervals, Plan
 from sqlmesh.core.user import User
 from sqlmesh.utils.date import now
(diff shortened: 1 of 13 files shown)


sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -407,15 +406,7 @@ def delete_pr_environment(self) -> None:
...
-        schemas = {parse_model_name(model.name)[1] for model in self._context.models.values()}
-        for schema in schemas:
-            assert schema
-            self._context.engine_adapter.drop_schema(
-                schema_name="__".join([schema, self.pr_environment_name]),
-                ignore_if_not_exists=True,
-                cascade=True,
-            )
-        return
+        self._context.invalidate_environment(self.pr_environment_name)
```
izeigerman, line 409: CC: @eakmanrq please take a look

tobymao (approved):

z3z1ma:
Would `teardown` make sense here as being more intuitive to an end user? That way users can "teardown" a non-prod environment in a single command which is sort of the "button" I can imagine people looking for. 

`teardown` == `invalidate` + `janitor` execution

Thoughts?

izeigerman:
@z3z1ma the problem with both `destroy` and `teardown` is that I think it would be misleading in this case. The environment doesn't get "destroyed" upon running the command. It's merely being marked for deletion at some point in time later. 

Additionally, we invoke the janitor process directly only when using the built-in scheduler. With Airflow the janitor runs based on a cadence configured by a user. The janitor process can be pretty expensive and long since it collects all expired environments and not just the one specified in the command, so I think it's better to keep it async.

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -407,15 +406,7 @@ def delete_pr_environment(self) -> None:
         """
         Deletes all the schemas for a given environment by checking all the schemas used by models.
```
eakmanrq, line 407: Can you update this docstring?

eakmanrq (approved):

eakmanrq:
At least form a CI/CD bot perspective, async deletion of an environment is fine IMO. It is ok if the environments sticks around for a bit until the next janitor run. I get what you are saying though @z3z1ma if there was a need for an environment to be immediately removed but I don't see that need currently.

PR #1035, opened 2023-06-26
Fix: Validate model definition only after the schema for it is set

izeigerman:
Validating definition requires the columns-to-types mapping which in turn relies on the rendered optimized query. The latter cannot be produced correctly without setting the schema first.


crericha:
Should `context.upsert_model` also call `validate_definition`?

PR #1036, opened 2023-06-26
Fix: Various Databricks related fixes

eakmanrq:
Note: This PR has Databricks do a non-transactional delete + insert due to issues with dynamic insert/overwrite on Unity catalog tables. Once a `REPLACE WHERE` bug is fixed we will be switching to that which will make it atomic again. 

The `USE CATALOG` approach from before only worked if SQL was being executed. So if you had a Dataframe then that catalog would not be used. The `setCurrentCatalog` works across both SQL and Dataframe API.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -65,10 +65,6 @@ class InsertOverwriteStrategy(Enum):
     def is_replace_where(self) -> bool:
         return self == InsertOverwriteStrategy.REPLACE_WHERE
 
-    @property
-    def is_not_delete_insert(self) -> bool:
-        return not self.is_delete_insert
-
 
 class EngineAdapter:
     """Base class wrapping a Database API compliant connection.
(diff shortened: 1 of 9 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -612,7 +631,19 @@ def _insert_overwrite_by_condition(
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
     ) -> None:
         table = exp.to_table(table_name)
-        if self.SUPPORTS_INSERT_OVERWRITE:
+        if (
+            self.INSERT_OVERWRITE_STRATEGY.is_delete_insert
```
izeigerman, line 635: [Nit] `self.INSERT_OVERWRITE_STRATEGY in (InsertOverwriteStrategy.DELETE_INSERT, InsertOverwriteStrategy.REPLACE_WHERE)` maybe

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -151,13 +151,16 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
...
+            # if the engine supports INSERT OVERWRITE or REPLACE WHERE and the snapshot is incremental by time range, we risk
             # having a partial result since each dataframe write can re-truncate partitions. To avoid this, we
             # union all the dataframes together before writing. For pandas this could result in OOM and a potential
             # workaround for that would be to serialize pandas to disk and then read it back with Spark.
             # Note: We assume that if multiple things are yielded from `queries_or_dfs` that they are dataframes
             # and not SQL expressions.
-            elif self.adapter.SUPPORTS_INSERT_OVERWRITE and snapshot.is_incremental_by_time_range:
+            elif (
+                self.adapter.INSERT_OVERWRITE_STRATEGY.is_insert_overwrite
+                or self.adapter.INSERT_OVERWRITE_STRATEGY.is_replace_where
```
izeigerman, line 162: Ditto here. Maybe a helper function in `base.py`.

izeigerman (approved):

PR #1037, opened 2023-06-26
Fix: Skip dbt tests that reference disabled models

izeigerman:


sqlmesh/dbt/manifest.py
```diff
@@ -136,9 +136,24 @@ def _load_tests(self) -> None:
             if node.resource_type != "test":
                 continue
 
+            skip_test = False
+            package_name = node.package_name
+            refs = _refs(node)
+            for ref in refs:
+                ref_node_name = f"model.{ref}" if "." in ref else f"model.{package_name}.{ref}"
```
crericha, line 143: We need to check for model, seed, and snapshot.

crericha: Maybe check all nodes that don't start with test?

izeigerman: ah, good catch, thanks!

crericha (approved):

PR #1041, opened 2023-06-27
Feat: Support partitioned_by expressions for bigquery

crericha:


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -247,7 +247,10 @@ def _create_table_properties(
         if partitioned_by:
             partition_columns_property = exp.PartitionedByProperty(
                 this=exp.Schema(
-                    expressions=[exp.to_identifier(column) for column in partitioned_by]
+                    expressions=[
+                        column.this if isinstance(column, exp.Column) else column
```
izeigerman, line 251: Should we fail if we detect a non-column expression?

sqlmesh/core/model/definition.py
```diff
@@ -1644,12 +1648,16 @@ def _single_value_or_tuple(values: t.Sequence) -> exp.Identifier | exp.Tuple:
     )
 
 
+def _single_expr_or_tuple(values: t.Sequence) -> exp.Expression | exp.Tuple:
```
izeigerman, line 1651: [Nit] `t.Sequence[exp.Expression]`?

sqlmesh/core/model/meta.py
```diff
@@ -25,15 +24,21 @@
...
     """
 
+    YEAR = "year"
+    MONTH = "month"
     DAY = "day"
     HOUR = "hour"
     MINUTE = "minute"
 
+    @property
+    def is_date_type(self) -> bool:
```
izeigerman, line 39: [Nit] To me it's more of a "granularity" than type.

sqlmesh/core/model/definition.py
```diff
@@ -557,7 +557,11 @@ def validate_definition(self) -> None:
             ConfigError
         """
         if self.partitioned_by:
-            unique_partition_keys = {k.strip().lower() for k in self.partitioned_by}
+            unique_partition_keys = {
+                col.name.strip().lower()
+                for expr in self.partitioned_by
+                for col in expr.find_all(exp.Column)
```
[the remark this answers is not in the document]
izeigerman: Found this logic in the column validator.

sqlmesh/dbt/model.py
```diff
@@ -98,6 +101,14 @@ def _validate_list(cls, v: t.Union[str, t.List[str]]) -> t.List[str]:
     def _validate_sql(cls, v: t.Union[str, SqlStr]) -> SqlStr:
         return SqlStr(v)
 
+    @validator("partition_by", pre=True)
+    def _validate_partition_by(cls, v: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+        if not v.get("field"):
+            raise ConfigError("'field' key required for partition_by.")
```
izeigerman, line 107: This only true for BigQuery. For example in Spark you can just specify a list of strings
https://docs.getdbt.com/reference/resource-configs/spark-configs
```
{{ config(
    materialized='incremental',
    partition_by=['date_day'],
    file_format='parquet'
) }}
```

tests/core/test_model.py
```diff
@@ -194,14 +193,22 @@ def test_model_validation_union_query():
...
 
-def test_partitioned_by():
-    expressions = parse(
-        """
+@pytest.mark.parametrize(
+    "partition_by_input, partition_by_output",
+    [
+        ("a", ["a"]),
+        ("(a, b)", ["a", "b"]),
+        ("TIMESTAMP_TRUNC(a, DAY)", ["TIMESTAMP_TRUNC(a, DAY)"]),
```
izeigerman, line 201: Do we have a test for expressions that don't reference a column?

sqlmesh/core/engine_adapter/spark.py
crericha: If only column expressions are valid, then yes. Please confirm.

sqlmesh/core/model/meta.py
```diff
@@ -136,6 +141,39 @@ def _cron_validator(cls, v: t.Any) -> t.Optional[str]:
...
+        elif isinstance(v, exp.Expression):
+            partitions = [v]
+        else:
+            dialect = values.get("dialect")
+            partitions = [
+                d.parse_one(entry, dialect=dialect) if isinstance(entry, str) else entry
+                for entry in ensure_list(v)
+            ]
+        partitions = [
+            exp.to_column(expr.name) if isinstance(expr, exp.Identifier) else expr
```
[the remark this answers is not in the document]
crericha: The parser returns an identifier if a single value. I took a more cautious approach here. I could potentially only convert if a single value.

sqlmesh/dbt/model.py
crericha: Thanks for pointing out. I didn't realize dbt spark had the same field with a different format.

tests/core/test_model.py
crericha: No, but we should. Good catch.

sqlmesh/core/engine_adapter/spark.py
izeigerman: For vanilla spark - yes, I think so

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -237,18 +237,21 @@ def create_view(
...
+            for expr in partitioned_by:
+                if not isinstance(expr, exp.Column):
+                    raise SQLMeshError(
+                        f"PARTITIONED BY contains non-column value '{expr.sql(dialect='spark')}'."
+                    )
             partition_columns_property = exp.PartitionedByProperty(
-                this=exp.Schema(
-                    expressions=[exp.to_identifier(column) for column in partitioned_by]
-                ),
+                this=exp.Schema(expressions=partitioned_by),
```
izeigerman, line 254: Is it ok that that this is a list of `Column` vs list of identifiers like it used to be?

izeigerman (approved):

sqlmesh/core/engine_adapter/spark.py
crericha: yes, verified via unit tests.

PR #1046, opened 2023-06-27
Feat: cluster by

tobymao:


sqlmesh/core/snapshot/definition.py
```diff
@@ -915,8 +915,11 @@ def _model_data_hash(model: Model) -> str:
         model.storage_format,
         str(model.lookback),
         *(model.partitioned_by or []),
+        *(model.clustered_by or []),
         model.stamp,
     ]
+    if model.clustered_by:
+        raise
```
izeigerman, line 922: Not sure I understand what we do here.

tests/core/test_model.py
```diff
@@ -1231,23 +1233,23 @@ def test_star_expansion(assert_exp_eq) -> None:
...
     assert_exp_eq(
         context.models["db.model2"].render_query(snapshots=snapshots),
-        """
+        f"""
         SELECT
           "model1"."id" AS "id",
           "model1"."item_id" AS "item_id",
           "model1"."ds" AS "ds"
-        FROM "sqlmesh__db"."db__model1__3784582009" AS "model1"
+        FROM "sqlmesh__db"."db__model1__{snapshots['db.model1'].version}" AS "model1"
```
izeigerman, line 1241: Did fingerprint change here because of your changes? Why? These models don't have clustered_by

izeigerman:
Don't you need to update this: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/definition.py#L1647
And also update model def validation logic here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/definition.py#L559 to bring it on par with partitioned_by.

tests/core/test_snapshot.py
```diff
@@ -78,11 +78,12 @@ def test_json(snapshot: Snapshot):
     assert json.loads(snapshot.json()) == {
         "created_ts": 1663891973000,
         "ttl": "in 1 week",
-        "fingerprint": snapshot.fingerprint,
+        "fingerprint": snapshot.fingerprint.dict(),
```
izeigerman, line 81: wow, how did this work before?

tests/core/test_model.py
tobymao: local sqlglot version triggered this, i decided to change this for future stuff

tests/core/test_snapshot.py
tobymao: pydantic overrides equality checks

izeigerman (approved):

crericha:
Snowflake supports `CLUSTER BY`. I think we should add that now or throw an exception if `clustered_by` is set for an engine that doesn't support it.

crericha:
Looks like the spark dbt adapter uses a field called `clustered_by` instead of `cluster_by`. We'll need to support that field name too. I can do this after the PR is merged if you like.

tobymao:
@crericha we'll need to add the bucket flag for spark with clustered_by

PR #1052, opened 2023-06-28
Fix!: tables with dots and more conservative optimization

tobymao:


setup.py
```diff
@@ -45,7 +45,7 @@
         "requests",
         "rich",
         "ruamel.yaml",
-        "sqlglot~=16.4.2",
+        "sqlglot~=16.5.0",
```
izeigerman, line 48: Shouldn't this be `16.6.0`?

sqlmesh/core/loader.py
```diff
@@ -54,23 +52,12 @@ def update_model_schemas(
...
         model.update_schema(schema)
-
-        cache_hit = optimized_query_cache.with_optimized_query(model)
+        optimized_query_cache.with_optimized_query(model)
 
         columns_to_types = model.columns_to_types
         if columns_to_types is not None:
             schema.add_table(name, columns_to_types, dialect=model.dialect)
 
-        if isinstance(model, SqlModel) and model.mapping_schema and not cache_hit:
```
izeigerman, line 64: why remove it though?

sqlmesh/core/renderer.py
```diff
@@ -148,7 +148,9 @@ def render(
             except MacroEvalError as ex:
                 raise_config_error(f"Failed to resolve macro for expression. {ex}", self._path)
 
-            _normalize_and_quote(expression, self._dialect)
+            if expression:
+                with _normalize_and_quote(expression, self._dialect) as expression:
```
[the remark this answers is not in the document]
izeigerman: ah, no, never mind.

sqlmesh/core/renderer.py
```diff
@@ -383,12 +385,13 @@ def _optimize_query(self, query: exp.Subqueryable) -> exp.Subqueryable:
...
                 query = query.copy()
 
-                qualify(
-                    query,
-                    dialect=self._dialect,
-                    schema=schema,
-                    infer_schema=False,
-                    validate_qualify_columns=False,
+                simplify(
+                    qualify(
```
izeigerman, line 389: Do we validate columns here now?

izeigerman (approved):

PR #1054, opened 2023-06-28
Fix!: preserve yaml order

tobymao:


tobymao:
@vchan this should fix the original issue ouf tests output order not matching

PR #1056, opened 2023-06-28
Fix!: Use target dbt adapter relation and column classes.

crericha:


sqlmesh/dbt/builtin.py
```diff
@@ -31,12 +32,15 @@ def warn(self, msg: str) -> str:
...
 
 class Api:
-    def __init__(self) -> None:
-        from dbt.adapters.base.column import Column
-        from dbt.adapters.base.relation import BaseRelation
+    def __init__(self, target: t.Optional[AttributeDict] = None) -> None:
+        if not target:
+            self.Relation = BaseRelation
+            self.Column = Column
+            return
```
izeigerman, line 39: I'm not sure it's a good practice to return in a constructor. Can we do `else` instead?

izeigerman (approved):

PR #1057, opened 2023-06-28
Feat: add tranpile option to format

z3z1ma:


tobymao:
as discussed, we need to edit the model dialect if it exists

sqlmesh/core/context.py
```diff
@@ -628,8 +628,16 @@ def format(self) -> None:
                 expressions = parse(
                     file.read(), default_dialect=self.config_for_model(model).dialect
                 )
+                for prop in expressions[0].expressions:
+                    p = t.cast(exp.Property, prop)
```
tobymao, line 632: i don't think you need this cast here do you?

sqlmesh/core/context.py
```diff
@@ -628,8 +628,16 @@ def format(self) -> None:
                 expressions = parse(
                     file.read(), default_dialect=self.config_for_model(model).dialect
                 )
+                for prop in expressions[0].expressions:
```
tobymao, line 631: this should only run if transpile right? otherwise don't touch it

sqlmesh/core/context.py
tobymao: yea let's remove it, it's not that useful in this context

z3z1ma: @tobymao done

tobymao (approved):

PR #1058, opened 2023-06-28
Fix: Use correct quote policy in dbt projects. Create correct relation instances in the adapter

izeigerman:


sqlmesh/dbt/builtin.py
```diff
@@ -351,3 +351,17 @@ def create_builtin_globals(
...
+    relation_info: t.Dict[str, t.Any],
+    relation_type: t.Type[BaseRelation],
+    target_quote_policy: Policy,
+) -> BaseRelation:
+    merged_quote_policy_dict = {}
+    for k, v in (relation_info.pop("quote_policy", {})).items():
+        v = getattr(target_quote_policy, k, None) if v is None else v
+        if v is not None:
+            merged_quote_policy_dict[k] = v
+    quote_policy = Policy(**merged_quote_policy_dict)
```
crericha, line 366: Could this be simplified to something like
{**target_quote_policy.dict(), **{k: v for k, v in relation_info.pop("quote_policy", {}) if v}}

crericha:
Can you confirm that yaml `null` get mapped to `None`.

crericha (approved):

sqlmesh/dbt/builtin.py
izeigerman: nice one 👍

PR #1060, opened 2023-06-29
Fix: Parse cluster_by in dbt models before converting to SQLMesh models

izeigerman:


sqlmesh/dbt/model.py
```diff
@@ -226,7 +226,9 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
             )
 
         if self.cluster_by:
-            optional_kwargs["clustered_by"] = self.cluster_by
+            optional_kwargs["clustered_by"] = [
+                d.parse_one(c, dialect=dialect).name for c in self.cluster_by
+            ]
```
crericha, line 231: Do we need the parser or can we instead do `exp.to_column(c).name`?

crericha: Also, do we properly quote CLUSTER BY cols in the generated sql?

izeigerman: Nope, need to parse it to get rid of quotes:
```
In [5]: exp.to_column('`col`').name
Out[5]: '`col`'
```

crericha: There is probably a sqlglot way to get the non-quoted text. @tobymao would know.

tobymao: you need to pass in the dialect, to_column

PR #1064, opened 2023-06-29
Fix: Map relations to snapshot physical tables in the dbt adapter

izeigerman:
Sometimes `ref` are passed into the dbt adapter while constructing a model query. This means that the production table name will be used since jinja rendering happens before we get a chance to resolve table references. 

This update ensures that dbt's `ref` returns resolved table references when using adapter.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -61,6 +61,7 @@ class ExpressionRenderer:
         jinja_macro_registry: t.Optional[JinjaMacroRegistry] = None,
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         only_latest: bool = False,
+        resolve_tables: bool = True,
     ):
         self._expression = expression
         self._dialect = dialect
@@ -69,6 +70,7 @@ class ExpressionRenderer:
         self._jinja_macro_registry = jinja_macro_registry or JinjaMacroRegistry()
         self._python_env = python_env or {}
         self._only_latest = only_latest
+        self._should_resolve_tables = resolve_tables
 
         self._cache: t.Dict[t.Tuple[datetime, datetime, datetime], t.Optional[exp.Expression]] = {}
 
@@ -111,7 +113,9 @@ class ExpressionRenderer:
             }
 
             env = prepare_env(self._python_env)
-            jinja_env = self._jinja_macro_registry.build_environment(**{**render_kwargs, **env})
+            jinja_env = self._jinja_macro_registry.build_environment(
+                **{**render_kwargs, **env}, snapshots=(snapshots or {}), is_dev=is_dev
+            )
 
             if isinstance(expression, d.Jinja):
                 try:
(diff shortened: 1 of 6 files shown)


sqlmesh/core/renderer.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         jinja_macro_registry: t.Optional[JinjaMacroRegistry] = None,
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         only_latest: bool = False,
+        resolve_tables: bool = True,
```
crericha, line 64: Can we push resolution of tables to later all the time? I'm wondering if that will simplify the code (assuming its possible)

sqlmesh/dbt/builtin.py
```diff
@@ -162,13 +171,33 @@ def var(name: str, default: t.Optional[t.Any] = None) -> t.Any:
...
+            logger.warning("Could not resolve ref '%s'", ref_name)
             return None
 
+        relation_info = relation_info.copy()
+        model_name = ".".join(
+            relation_info[p] for p in ("database", "schema", "identifier") if relation_info.get(p)
+        )
+        table_mapping = to_table_mapping(snapshots.values(), is_dev)
+        if model_name in table_mapping:
+            physical_table_name = table_mapping[model_name]
```
crericha, line 194: Is it possible that the mapping isn't the dict? If so, should we catch and raise a better error message?

sqlmesh/dbt/builtin.py
```diff
@@ -162,13 +171,33 @@ def var(name: str, default: t.Optional[t.Any] = None) -> t.Any:
...
+            logger.info("Resolved ref '%s' to snapshot table '%s'", ref_name, physical_table_name)
```
crericha, line 195: Maybe debug instead of info?

sqlmesh/dbt/builtin.py
```diff
@@ -178,6 +207,7 @@ def generate_source(sources: t.Dict[str, t.Any], api: Api) -> t.Callable:
     def source(package: str, name: str) -> t.Optional[BaseRelation]:
         relation_info = sources.get(f"{package}.{name}")
         if relation_info is None:
+            logger.warning("Could not resolve source '%s.%s'", package, name)
```
crericha, line 210: Package isn't schema, so "package.name" doesn't really make sense. Recommend package=, name=.

sqlmesh/core/renderer.py
izeigerman: Not sure I understand what you're suggesting exactly

sqlmesh/dbt/builtin.py
izeigerman: When mapping can not be a dict?

crericha: I meant the mapping isn't *in* the dict. Key word missing there!

sqlmesh/core/renderer.py
crericha: My understanding is this boolean denotes whether we should resolve tables here or not. I'm assuming that means we resolve them later? And if so, why wouldn't we always do it there.

izeigerman: Actually, let me refactor

izeigerman: refactored, please take a look

crericha (approved):
I like this approach much more. About as clean as we can get I think.

izeigerman:
@crericha thanks for suggesting it in the first place!

Issue #1065, opened 2023-06-29
sqlmesh requires a password for snowflake connections in dbt mode even when using a different authentication method


Qwertronix:
sqlmesh seems like it should support authenticator=externalbrowser for Snowflake, and dbt does as well... but sqlmesh using the dbt integration expects a non-optional password and immediately fails:

```
PS C:\WIP\my_project> sqlmesh plan       
Error: 1 validation error for SnowflakeConfig
password
  field required (type=value_error.missing)
```

PR #1066, opened 2023-06-29
Feat: Add Snowflake authenticator connection config for dbt projects

crericha:
Fixes: #1065

### DIFF
--- a/sqlmesh/dbt/target.py
+++ b/sqlmesh/dbt/target.py
@@ -153,12 +153,20 @@ class SnowflakeConfig(TargetConfig):
         retry_all: A boolean flag to retry on all Snowflake connector errors
     """
 
-    # TODO add other forms of authentication
     type: Literal["snowflake"] = "snowflake"
     account: str
-    warehouse: str
     user: str
-    password: str
+
+    # User and password authentication
+    password: t.Optional[str]
+
+    # SSO authentication
+    authenticator: t.Optional[str]
+
+    # TODO add other forms of authentication
+
+    # Optional
+    warehouse: t.Optional[str]
     role: t.Optional[str]
     client_session_keep_alive: bool = False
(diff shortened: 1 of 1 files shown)


sqlmesh/dbt/target.py
```diff
@@ -167,6 +175,14 @@ class SnowflakeConfig(TargetConfig):
     retry_on_database_errors: bool = False
     retry_all: bool = False
 
+    @root_validator(pre=True)
+    def validate_authentication(
+        cls, values: t.Dict[str, t.Union[t.Tuple[str, ...], t.Optional[str], t.Dict[str, t.Any]]]
+    ) -> t.Dict[str, t.Union[t.Tuple[str, ...], t.Optional[str], t.Dict[str, t.Any]]]:
+        if values.get("password") or values.get("authenticator"):
```
GeorgeSittas, line 182: Should we make sure that both `password` and `user` are present, in case `authenticator` isn't used?

GeorgeSittas: Note that this is similar to connection.py:

```python
    @root_validator(pre=True)
    def _validate_authenticator(
        cls, values: t.Dict[str, t.Optional[str]]
    ) -> t.Dict[str, t.Optional[str]]:
        if "type" in values and values["type"] != "snowflake":
            return values
        auth = values.get("authenticator")
        user = values.get("user")
        password = values.get("password")
        if not auth and (not user or not password):
            raise ConfigError("User and password must be provided if using default authentication")
        return values
```

crericha: `user` is a required field in `dbt`. If not set, pydantic will throw an error.

GeorgeSittas: Gotcha, thanks for clarifying.

GeorgeSittas (approved):

PR #1067, opened 2023-06-29
Refactor: macro reference extraction doesn't need dialect specific sq…

tobymao:
…l. using generic for performance


GeorgeSittas (approved):
Looks good.

Does this actually produce a noticeable performance improvement?

tobymao:
yes becsuse dialects use transform which does a copy

GeorgeSittas:
Ahh good point!

PR #1069, opened 2023-06-30
Fix: Always set global variables for dbt packages. Manually extract macro references in additional to the ones fetched from the manifest

izeigerman:


tobymao (approved):

crericha (approved):
Extract from macros too and gtg.

PR #1071, opened 2023-06-30
Fix: Handle optional model and audit paths in the API

vchan:


sqlmesh/dbt/test.py
```diff
@@ -108,14 +108,16 @@ def to_sqlmesh(self, context: DbtContext) -> Audit:
...
-        return Audit(
+        audit = Audit(
             name=self.name,
             dialect=self.dialect,
             skip=skip,
             blocking=blocking,
             query=query,
             jinja_macros=jinja_macros,
         )
+        audit._path = self.path
```
vchan, line 119: @crericha I noticed path wasn't set here so I added it. Please let me know if that will cause problems.

tobymao (approved):

web/server/settings.py
```diff
@@ -37,10 +37,11 @@ def _get_context(path: str | Path, config: str) -> Context:
...
     mapping: dict[Path, FileType] = {}
     for audit in context._audits.values():
-        path = audit._path.relative_to(context.path)
-        mapping[path] = FileType.audit
+        if audit._path:
+            path = audit._path.relative_to(context.path)
+            mapping[path] = FileType.audit
     for model in context.models.values():
-        if model.source_type == "external":
+        if model.source_type == "external" or not model._path:
```
izeigerman, line 44: [Nit] `model.kind.is_external` instead of `model.source_type == "external"`

PR #1072, opened 2023-06-30
Fix: Allow multiple statements after rendering jinja for pre- / post-statements

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -686,12 +686,12 @@ def _render_statements(
...
         **kwargs: t.Any,
     ) -> t.List[exp.Expression]:
-        rendered = (
+        rendered = [
             self._statement_renderer(statement).render(**kwargs)
             for statement in statements
             if not isinstance(statement, d.MacroDef)
-        )
-        return [r for r in rendered if r is not None]
+        ]
```
tobymao, line 693: why'd you get rid of the generator?

izeigerman: no good reason. Thanks for catching

tobymao (approved):

PR #1073, opened 2023-06-30
Chore: polish up chi square test and allow auditing for dependence OR independence

z3z1ma:
Make chi test a nice example of the power of sqlmesh tests and polish it to make it more stats-ey and less long-form relational


izeigerman (approved):
We gotta add all these new awesome audits to documentation some time soon.

PR #1075, opened 2023-06-30
Feature: auto rollback migration if it fails

tobymao:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -637,6 +645,7 @@ def rollback(self) -> None:
 
         if self.engine_adapter.table_exists(_backup_table_name(self.intervals_table)):
             self._restore_table(self.seeds_table, _backup_table_name(self.intervals_table))
+        logger.info("Migration rollback successfull.")
```
vchan, line 648:
```suggestion
        logger.info("Migration rollback successful.")
```

vchan (approved):

PR #1077, opened 2023-06-30
Chore: update table diff styling and expand tests

treysp:

PR #1078, opened 2023-06-30
Docs: Add table diff guide

treysp:
Uses console styling from #1077


docs/guides/tablediff.md
```diff
@@ -0,0 +1,105 @@
...
+Compare a SQLMesh model across environments with the SQLMesh CLI interface by using the command `sqlmesh table_diff [source environment]:[target environment] [model name]`. 
+
+For example, we could add a `WHERE` clause to the [SQLMesh quickstart](../quick_start.md) model `sqlmesh_example.full_model`:
+
+```sql linenums="1"
+MODEL (
+  name sqlmesh_example.full_model,
+  kind FULL,
+  cron '@daily',
+  audits [assert_positive_order_ids],
```
tobymao, line 30: set the grain here, and then you don't need to define -o

sqlmesh/cli/example_project.py
```diff
@@ -8,7 +8,7 @@
     local:
         connection:
             type: duckdb
-            database: db.db
+            database: ./db.db
```
tobymao, line 11: why did you change this

tobymao (approved):

sqlmesh/cli/example_project.py
treysp: The quickstart db currently writes to the __file__ location, which I found unintuitive as a user (and is unstable depending on python config).

I noticed it while repeatedly deleting and re-initializing the project.

Should I revert?

tobymao: so it writes to the same location as the project?

treysp: yes, in the project directory - db.db and the wal

treysp: logic being that we can be sure the directory where `init` is called is writable and easily located by the user

tobymao: to me, db.db should be colocated with the example project, not where you run it, is that the current behavior?

treysp: By colocated, do you mean in the same directory as the example project directory, or do you mean inside the example project directory where config.yaml is located?

treysp: Either way, i'm having trouble consistently reproducing the behavior in my dev environment in a fresh dir, venv, and pip. We haven't heard any complaints about this, so i'll go ahead and revert.

tobymao (approved):

PR #1081, opened 2023-07-03
Fix: Fallback to model dialect when rendering an audit query

izeigerman:


sqlmesh/core/audit/definition.py
```diff
@@ -240,7 +240,7 @@ def macro_definitions(self) -> t.List[d.MacroDef]:
     def _create_query_renderer(self, model: Model) -> QueryRenderer:
         return QueryRenderer(
             self.query,
-            self.dialect,
+            self.dialect or model.dialect,
```
GeorgeSittas, line 243: Does this affect migration v0013? Do we also want to use the model's dialect there as a fallback?

izeigerman: This change only impacts runtime and doesn't impact the state.

GeorgeSittas: Ah makes sense, since it's just the renderer's dialect that's changed.

PR #1083, opened 2023-07-03
Fix: Fetch columns-to-types mapping from the target DW when determining a type of the time column

izeigerman:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -657,6 +662,10 @@ def _insert_overwrite_by_condition(
                     )
                 )
             query = t.cast("Query", query_or_df)
+
+            if where and isinstance(query, exp.Subqueryable):
```
tobymao, line 666: put this check into the add_where_to_query method directly

sqlmesh/core/model/definition.py
```diff
@@ -38,7 +38,6 @@
     print_exception,
     serialize_env,
 )
-from sqlmesh.utils.pandas import filter_df_by_timelike
```
tobymao, line 41: can you delete this method now?

izeigerman: seems like I can

tobymao: how do you filter pandas dataframes then?

izeigerman: when it's translated to sql

izeigerman: removed the function

tobymao (approved):

PR #1084, opened 2023-07-03
Fix: Make all BigQuery db calls retryable

eakmanrq:
Prior to this change whenever we did client direct calls (which is fairly common in BQ engine adapter) they were not retried and only the SQL execution calls were retried. This retries all db calls.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,6 +78,9 @@ ignore_missing_imports = True
 [mypy-dbt.adapters.*]
 ignore_missing_imports = True
 
+[mypy-slack_sdk.*]
+ignore_missing_imports = True
+
 [autoflake]
 in-place = True
 expand-star-imports = True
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -78,6 +78,9 @@ ignore_missing_imports = True
 [mypy-dbt.adapters.*]
 ignore_missing_imports = True
 
+[mypy-slack_sdk.*]
```
eakmanrq, line 81: I think we were missing this? I get a mypy error without it.

eakmanrq: @vchan

eakmanrq: Oh I see you made it a dependency of `make install-dev` but not `dev` in setup.py.

tobymao (approved):

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -193,11 +190,24 @@ def __load_pandas_to_table(
...
             raise SQLMeshError(result.errors)
         return result
 
+    def __db_load_table_from_dataframe(
+        self, df: pd.DataFrame, table: bigquery.Table, job_config: bigquery.LoadJobConfig
+    ) -> None:
+        job = self.client.load_table_from_dataframe(
+            dataframe=df, destination=table, job_config=job_config
+        )
+        result = self._db_call(job.result)
```
tobymao, line 208: return self._db_call

setup.cfg
vchan: Yes, it should be an optional dependency.

tobymao (approved):

PR #1087, opened 2023-07-03
Feat!: Add tags support for dbt projects.

crericha:


GeorgeSittas (approved):

sqlmesh/dbt/model.py
```diff
@@ -70,7 +69,6 @@ class ModelConfig(BaseModelConfig):
     lookback: t.Optional[int] = None
 
     # DBT configuration fields
-    start: t.Optional[str] = None
```
tobymao, line 73: why did this go away?

crericha: It was already being inherited [here](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/dbt/common.py#L87), so unnecessary.

tobymao (approved):

PR #1088, opened 2023-07-03
Fix: Deterministic order when computing fingerprint for jinja macro packages

izeigerman:


crericha (approved):

sqlmesh/core/snapshot/definition.py
```diff
@@ -942,7 +942,7 @@ def _model_data_hash(model: Model) -> str:
             data.append(macro_name)
             data.append(macro.definition)
 
-        for package in model.jinja_macros.packages.values():
+        for _, package in sorted(model.jinja_macros.packages.items(), key=lambda x: x[0]):
```
vchan, line 945: Nit: The key argument isn't really needed here.

tobymao (approved):

sqlmesh/core/snapshot/definition.py
izeigerman: what do you mean?

izeigerman: we need it for sorting

vchan: I don't think it's a big deal. I meant `sorted` will use the first element of x automatically for sorting.

tobymao: big brain

```python
>>> sorted({"b": 1, "a": 2}.items())
[('a', 2), ('b', 1)]
```

PR #1089, opened 2023-07-03
Fix!: fix bigquery nested types

tobymao:


eakmanrq (commented):
What do you think about writing some tests against `dtype_to_sql`? Cover some common primitives, array, struct, and then nested struct/array.

tobymao:
> What do you think about writing some tests against `dtype_to_sql`? Cover some common primitives, array, struct, and then nested struct/array.

problem is that it takes in bigquery types

PR #1095, opened 2023-07-05
Fix: SQL syntax in editor

mykalmax:
refactored code editor syntax highlighting


web/client/src/workers/sqlglot/sqlglot.py
```diff
@@ -34,4 +34,30 @@ def get_dialect(name: str = "") -> str:
...
+def validate(sql: str = "", read: DialectType = None) -> str:
+    is_valid = True
+    try:
+        sqlglot.transpile(
+            sql, read=read, pretty=False, unsupported_level=sqlglot.errors.ErrorLevel.IMMEDIATE
+        )
+    except sqlglot.errors.ParseError as e:
+        is_valid = False
+
+    return json.dumps(is_valid)
```
vchan, line 54:
```suggestion
    try:
        sqlglot.transpile(
            sql, read=read, pretty=False, unsupported_level=sqlglot.errors.ErrorLevel.IMMEDIATE
        )
    except sqlglot.errors.ParseError as e:
        return json.dumps(False)  # or just return 'false'
    return json.dumps(True)
```

web/client/src/library/components/editor/extensions/SQLMeshDialect.ts
```diff
@@ -0,0 +1,192 @@
...
+  options?: { types: string; keywords: string },
+  dialects?: string[],
+) => LanguageSupport
+
+export const SQLMeshDialect: ExtensionSQLMeshDialect = function SQLMeshDialect(
+  models,
+  options = { types: '', keywords: '' },
+  dialects,
+): LanguageSupport {
+  const SQLKeywords = options.keywords + WHITE_SPACE
```
vchan, line 26: Can you please remind me again why all these strings need a trailing white space?

web/client/src/library/components/editor/extensions/SQLMeshDialect.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  sqlglotWorker.removeEventListener('message', handler)
+}
+
+export function getSQLMeshModelKeywords(
+  dialects: string[] = [],
+): Map<string, Completion[]> {
+  return new Map([
+    [
+      'kind',
```
vchan, line 128: Should we include EMBEDDED here?

web/client/src/utils/index.ts
```diff
@@ -48,11 +48,11 @@ export function isObjectLike(value: unknown): boolean {
   return isNotNil(value) && typeof value === 'object'
 }
 
-export function isNil(value: unknown): boolean {
+export function isNil(value: unknown): value is undefined | null {
   return value == null
 }
 
-export function isNotNil(value: unknown): boolean {
+export function isNotNil<T>(value: T | null | undefined): value is T {
```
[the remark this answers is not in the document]
mykalmax: It means that value is not null / undefined but actually some value of type T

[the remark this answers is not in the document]
vchan: Okay, but isn't that where the return type goes?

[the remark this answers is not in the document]
mykalmax: Oh, yes it is return type.
this should let eslint/typescript know that value we checking exists for code block below the check

web/client/src/library/components/editor/extensions/SQLMeshDialect.ts
mykalmax: in case `options.keywords` coming without trailing space

vchan: Is it for concatenating with `SQLMeshKeywords` below for `lang`? Does codemirror expect a trailing whitespace? What would happen if `SQLKeywords` and `SQLMeshKeywords` didn't have trailing whitespaces, but instead below is `keywords: (SQLKeywords + WHITE_SPACE + SQLMeshKeywords).toLowerCase()`?

vchan (approved):

PR #1101, opened 2023-07-05
Fix: Include database in dbt source fqns when it differs from the project's database

crericha:


tobymao (approved):

izeigerman (approved):
Great! I assume models work as expected

PR #1102, opened 2023-07-05
Feat: add progress bar for migrations

### LINKED ISSUE #1050: Feature: progress bars for migration + snapshot creation

GeorgeSittas:
Fixes #1050 (partially -- I'm planning to make a followup PR for the snapshot creation progress)

Demo:

Screenshot 2023-07-06 at 2 05 52 AM

----

Note: renamed the `SchedulerBatchColumn` class to just `BatchColumn`, because the "Scheduler" part seemed unnecessary / confusing to me. Is there additional context I'm missing regarding this?

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -176,7 +176,7 @@ class TerminalConsole(Console):
                 BarColumn(bar_width=40),
                 "[progress.percentage]{task.percentage:>3.1f}%",
                 "•",
-                srich.SchedulerBatchColumn(),
+                srich.BatchColumn(),
                 "•",
                 TimeElapsedColumn(),
                 console=self.console,
@@ -249,6 +249,8 @@ class TerminalConsole(Console):
                 BarColumn(bar_width=40),
                 "[progress.percentage]{task.percentage:>3.1f}%",
                 "•",
+                srich.BatchColumn(),
+                "•",
                 TimeElapsedColumn(),
                 console=self.console,
             )
@@ -261,7 +263,6 @@ class TerminalConsole(Console):
 
     def update_migration_progress(self, num_tasks: int) -> None:
         """Update migration progress."""
-        breakpoint()
         if self.migration_progress is not None and self.migration_task is not None:
             self.migration_progress.update(self.migration_task, refresh=True, advance=num_tasks)
 
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -245,7 +245,7 @@ class TerminalConsole(Console):
         """Indicates that a new migration progress has begun."""
         if self.migration_progress is None:
             self.migration_progress = Progress(
-                TextColumn(f"[bold blue]Migrating", justify="right"),
+                TextColumn(f"[bold blue]Migrating snapshots", justify="right"),
                 BarColumn(bar_width=40),
                 "[progress.percentage]{task.percentage:>3.1f}%",
                 "•",
@@ -257,7 +257,7 @@ class TerminalConsole(Console):
 
             self.migration_progress.start()
             self.migration_task = self.migration_progress.add_task(
-                f"Migrating...",
+                f"Migrating snapshots...",
                 total=total_tasks,
             )
 
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -41,7 +41,7 @@ SNAPSHOT_CHANGE_CATEGORY_STR = {
 
 class Console(abc.ABC):
     """Abstract base class for defining classes used for displaying information to the user and also interact
-    with them when their input is needed"""
+    with them when their input is needed."""
 
     @abc.abstractmethod
     def start_snapshot_progress(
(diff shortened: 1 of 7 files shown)


tobymao:
the main progress is in migrate_rows, not the migrations. it takes a long time to migrate every snapshot, so that's what actually needs to be tracked

GeorgeSittas:
> the main progress is in migrate_rows, not the migrations. it takes a long time to migrate every snapshot, so that's what actually needs to be tracked

Thanks for clarifying, I'll update accordingly. Do we want to keep the current progress bar or is it irrelevant?

tobymao:
not super important, they're represented in the logs already

GeorgeSittas:
Updated demo:

Screenshot 2023-07-06 at 2 59 36 AM

GeorgeSittas:
Maybe just saying "Migrating" next to the progress bar doesn't provide enough context. I thought about "Migrating snapshots", but then I remembered that we generally don't want to expose the concept of the "snapshot" to the end user.

Any ideas?

tobymao:
Migrate snapshots is fine

sqlmesh/core/console.py
```diff
@@ -162,13 +176,14 @@ def start_snapshot_progress(
                 BarColumn(bar_width=40),
                 "[progress.percentage]{task.percentage:>3.1f}%",
                 "•",
-                srich.SchedulerBatchColumn(),
+                srich.BatchColumn(),
```
tobymao, line 179: what's this change?

sqlmesh/core/context.py
```diff
@@ -1031,7 +1031,7 @@ def migrate(self) -> None:
 
         Please contact your SQLMesh administrator before doing this.
         """
-        self._new_state_sync().migrate()
+        self._new_state_sync().migrate(console=self.console)
```
tobymao, line 1034: why don't you add the console to the state_sync in init?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -729,6 +736,7 @@ def _migrate_rows(self) -> None:
                     for name in _parents_from_model(model, models)
                 )
             except Exception:
+                completed = False
```
tobymao, line 739: this is not correct, we can still migrate successfull, we just didn't auto convert everything

sqlmesh/core/console.py
GeorgeSittas: This class is not related to "Scheduler" in any way, it's just a generic template to create the `x/total` column in rich.

sqlmesh/core/state_sync/engine_adapter.py
GeorgeSittas: So this progress should always be successful? I assumed this was the only "failure" state, but I can remove `completed` completely if needed.

sqlmesh/core/context.py
GeorgeSittas: Will update in a sec

GeorgeSittas: Done

GeorgeSittas:
The updated progress bar title is now "Migrating snapshots".

sqlmesh/core/state_sync/engine_adapter.py
tobymao: yea it's not necessary, the only failure state is if we rollback

GeorgeSittas: Another related detail here: do we want to bump the snapshot tracker in the progress bar if we get an exception? I was thinking that in the end of the migration it might be weird to see "Migration completed successfully" but have something like "2/3" in the batch column.

GeorgeSittas: > yea it's not necessary, the only failure state is if we rollback

I see, thanks for the explanation.

tobymao: well that could indicate how many snapshots were auto converted

GeorgeSittas: Cool, making the change now so that the migration is not marked as a success if we rollback.

GeorgeSittas: I added a `raise SQLMeshError(...)` after I ran `sqlmesh plan` the first time, so this is what it looks like when a migration fails (the 100% • 3/3 is expected here since I created the error here deliberately).

Screenshot 2023-07-06 at 10 27 48 PM

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -73,9 +74,11 @@ def __init__(
         self,
         engine_adapter: EngineAdapter,
         schema: str = c.SQLMESH,
+        console: t.Optional[Console] = None,
     ):
         self.schema = schema
         self.engine_adapter = engine_adapter
+        self.console = console or get_console()
```
izeigerman, line 81: TBH, I think if console is None we should just not report.

GeorgeSittas: I thought the `console or get_console()` is a common pattern in classes that be instantiated with a `Console` argument. For example, both `Scheduler` and `BuiltInEvaluator` do the same. Fwiw I'm fine with not reporting too.

izeigerman: if that's common, feel free to leave this as is

tobymao (approved):

PR #1116, opened 2023-07-07
Refactor!: remove hydrating and missing_intervals from state_sync

tobymao:


sqlmesh/core/state_sync/cache.py
```diff
@@ -31,8 +28,19 @@ def __init__(self, state_sync: StateSync):
         # False means that the snapshot does not exist in the state sync but has been requested before
         self.snapshot_cache: t.Dict[SnapshotId, Snapshot | Literal[False]] = {}
 
+        #    def _set_snapshot(self, snapshot: Snapshot) -> None:
```
izeigerman, line 31: Delete this?

sqlmesh/core/state_sync/base.py
```diff
@@ -52,14 +48,18 @@ class StateReader(abc.ABC):
 
     @abc.abstractmethod
     def get_snapshots(
-        self, snapshot_ids: t.Optional[t.Iterable[SnapshotIdLike]], hydrate_seeds: bool = False
+        self,
+        snapshot_ids: t.Optional[t.Iterable[SnapshotIdLike]],
+        hydrate_seeds: bool = False,
+        refresh: bool = False,
```
izeigerman, line 54: This seems like very specific to cache as opposed to just any state sync.

izeigerman: I don't think we even need this. See my other comments.

sqlmesh/core/state_sync/common.py
```diff
@@ -215,25 +219,13 @@ def _ensure_no_gaps(
...
-            self.get_snapshots(changed_version_prev_snapshots_by_name.values()).values(),
-            all_snapshot_intervals,
-        )
-        hydrated_target_snapshots = Snapshot.hydrate_with_intervals_by_version(
-            changed_version_target_snapshots, all_snapshot_intervals
-        )
-        hydrated_target_snapshots_by_name = {s.name: s for s in hydrated_target_snapshots}
+        prev_snapshots = self.get_snapshots(
+            changed_version_prev_snapshots_by_name.values(),
+            refresh=True,
```
izeigerman, line 224: Can we just call `self._get_snapshots()` here instead of `refresh=True` and bypass cache altogether?

izeigerman: Perhaps do the same here: https://github.com/TobikoData/sqlmesh/blob/1584aa5c5d1a15a2dec80eb8ade1e61188f13ab6/sqlmesh/core/state_sync/common.py#L93

sqlmesh/schedulers/airflow/operators/hwm_sensor.py
```diff
@@ -35,7 +35,10 @@ def __init__(
     def poke(self, context: Context) -> bool:
         dag_run = context["dag_run"]
 
-        target_snapshot = self._get_target_snapshot()
+        with util.scoped_state_sync() as state_sync:
+            target_snapshot = state_sync.get_snapshots([self.target_snapshot_info], refresh=True)[
```
izeigerman, line 39: I think `refresh=True` is meaningless here since `util.scoped_state_sync()` always returns a new state sync instance, so there is no cache anyway.

sqlmesh/core/state_sync/common.py
```diff
@@ -206,6 +209,7 @@ def _ensure_no_gaps(
     ) -> None:
         from sqlmesh.core.scheduler import start_date
 
+        # check if w eneed to refresh here
```
izeigerman, line 212: typo

izeigerman: Just always fetch from state when ensuring no gaps. Bypass the cache altogether.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -527,18 +521,10 @@ def recycle(self) -> None:
...
-                return query
-            return query.where(self._snapshot_name_version_filter(snapshots))
-
-        return self._get_snapshot_intervals(query_modifier)[1]
-
     def _get_snapshot_intervals(
-        self, query_modifier: t.Optional[t.Callable[[exp.Select], exp.Select]]
+        self,
+        snapshots: t.Optional[t.Collection[SnapshotNameVersionLike]] = None,
+        uncompacted: bool = False,
```
izeigerman, line 527: I guess we need to clarify that this means uncompacted **only**.

sqlmesh/core/snapshot/definition.py
```diff
@@ -1127,3 +1131,40 @@ def has_paused_forward_only(
...
+def missing_intervals(
+    snapshots: t.Collection[Snapshot],
+    start: t.Optional[TimeLike] = None,
+    end: t.Optional[TimeLike] = None,
+    latest: t.Optional[TimeLike] = None,
+    restatements: t.Optional[t.Iterable[str]] = None,
+    is_dev: bool = False,
+) -> t.Dict[Snapshot, Intervals]:
+    """Returns all missing intervals given a collection of snapshots."""
+    from sqlmesh.core import scheduler
```
izeigerman, line 1145: This is not very clean. Should `earliest_start_date` belong here as well?

izeigerman:
Can we get rid of this now https://github.com/TobikoData/sqlmesh/blob/1584aa5c5d1a15a2dec80eb8ade1e61188f13ab6/sqlmesh/core/snapshot/definition.py#L412-L417 ?

tobymao:
> Can we get rid of this now
> 
> https://github.com/TobikoData/sqlmesh/blob/1584aa5c5d1a15a2dec80eb8ade1e61188f13ab6/sqlmesh/core/snapshot/definition.py#L412-L417
> ?

@izeigerman i'm not familiar with this section do we?

sqlmesh/core/state_sync/cache.py
tobymao: need to still research this

izeigerman: Can this code be removed now?

sqlmesh/schedulers/airflow/operators/hwm_sensor.py
```diff
@@ -35,7 +35,10 @@ def __init__(
     def poke(self, context: Context) -> bool:
         dag_run = context["dag_run"]
 
-        target_snapshot = self._get_target_snapshot()
+        with util.scoped_state_sync() as state_sync:
+            target_snapshot = state_sync._get_snapshots([self.target_snapshot_info])[
```
izeigerman, line 39: Why use the private method? We should just use the public one

izeigerman (approved):
Couple more comments, LGTM otherwise

sqlmesh/schedulers/airflow/operators/hwm_sensor.py
izeigerman: Yeah, but `scoped_state_sync()` returns a new instance every time, so there will be no cache.

sqlmesh/schedulers/airflow/util.py
```diff
@@ -35,7 +35,7 @@
 
 
 @contextlib.contextmanager
-def scoped_state_sync() -> t.Generator[StateSync, None, None]:
+def scoped_state_sync() -> t.Generator[EngineAdapterStateSync, None, None]:
```
izeigerman, line 38: Not needed either.

tests/core/test_scheduler.py
```diff
@@ -68,62 +75,15 @@ def test_interval_params_missing(scheduler: Scheduler, sushi_context_fixed_date:
...
+    assert compute_interval_params(
+        [waiters], start=start_ds, end=end_ds, latest=end_ds, is_dev=False
+    ) == {
         waiters: [
             (to_datetime(start_ds), to_datetime("2022-03-02")),
         ]
     }
 
 
-def test_multi_version_snapshots(
```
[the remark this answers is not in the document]
tobymao: it's testing snapshot merging behavior based on version

PR #1121, opened 2023-07-10
Fix: Populate unittest inputs before rendering query, which may access the input tables

crericha:
- dbt models may try to hit the db during rendering, so make sure the db tables are populated before rendering the model.
- Allow unit tests to override table mappings when rendering.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -437,7 +437,7 @@ def _resolve_tables(
 
     snapshots = snapshots or {}
     table_mapping = table_mapping or {}
-    mapping = table_mapping | to_table_mapping(snapshots.values(), is_dev)
+    mapping = {**to_table_mapping(snapshots.values(), is_dev), **table_mapping}
     # if a snapshot is provided but not mapped, we need to expand it or the query
     # won't be valid
     expand = set(expand) | {name for name in snapshots if name not in mapping}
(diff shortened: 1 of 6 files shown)


sqlmesh/core/renderer.py
```diff
@@ -291,7 +302,7 @@ def render(
             The rendered expression.
         """
         cache_key = _dates(start, end, latest)
-        skip_cache = bool(snapshots or expand)
+        skip_cache = bool(snapshots or table_mapping or expand)
```
izeigerman, line 305: @tobymao I don't get why we skip cache here on snapshots, etc. We don't cache after table resolution anyway.

sqlmesh/core/test/definition.py
```diff
@@ -152,67 +153,44 @@ def _get_rows(self, table: list[Row] | dict[str, list[Row]]) -> list[Row]:
...
 
     def runTest(self) -> None:
-        self.test_ctes()
+        # For tests we just use the model name for the table reference and we don't want to expand
+        mapping = {
+            name: _test_fixture_name(name)
+            for name in self.models.keys() | self.body.get("inputs", {}).keys()
+        }
+        query = self.model.render_query_or_raise(
+            add_incremental_filter=True,
```
izeigerman, line 182: This flag is no long relevant in the renderer. Let's remove it.

izeigerman (approved):

vchan (approved):

sqlmesh/core/renderer.py
tobymao: if these are not needed, can you remove them @crericha

crericha: Confirmed not needed anymore. Removed in the latest commit to this PR.

PR #1126, opened 2023-07-11
Fix!: normalize model names and dependencies as sqlglot dialect

tobymao:


GeorgeSittas (approved):
Looks good I think.

Just FMI: you scanned all `to_table` calls and got rid of the `dialect` arg?

tobymao:
yes

PR #1127, opened 2023-07-11
Feat!: Require dialect arg in sqlmesh config

treysp:


sqlmesh/cli/main.py
```diff
@@ -94,21 +99,28 @@ def cli(
...
 @error_handler
-def init(ctx: click.Context, template: t.Optional[str] = None) -> None:
+def init(ctx: click.Context, dialect: str, template: t.Optional[str] = None) -> None:
     """Create a new SQLMesh repository."""
     try:
         project_template = ProjectTemplate(template.lower() if template else "default")
     except ValueError:
         raise click.ClickException(f"Invalid project template '{template}'")
-    init_example_project(ctx.obj, template=project_template)
+    _ = Dialects(dialect.lower())
```
tobymao, line 122: what is this? why do you have _ ?

sqlmesh/core/context.py
```diff
@@ -236,6 +237,11 @@ def __init__(
         self._jinja_macros = JinjaMacroRegistry()
 
         self.path, self.config = t.cast(t.Tuple[Path, Config], next(iter(self.configs.items())))
+        if self.config.model_defaults.dialect is None:
+            raise ConfigError("Default model SQL dialect is a required configuration parameter.")
+        else:
+            _ = Dialects(self.config.model_defaults.dialect.lower())
```
tobymao, line 243: delete this?

sqlmesh/core/context.py
```diff
@@ -236,6 +237,11 @@ def __init__(
         self._jinja_macros = JinjaMacroRegistry()
 
         self.path, self.config = t.cast(t.Tuple[Path, Config], next(iter(self.configs.items())))
+        if self.config.model_defaults.dialect is None:
```
tobymao, line 240: this should just be a valiadotr inside of model_defaults

sqlmesh/magics.py
```diff
@@ -31,6 +32,10 @@
 ]
 
 
+def _sqlglot_dialects() -> str:
```
tobymao, line 35: turn this into a helper function, don't copy and paste code

sqlmesh/magics.py
```diff
@@ -107,7 +117,8 @@ def init(self, line: str) -> None:
             )
         except ValueError:
             raise MagicError(f"Invalid project template '{args.template}'")
-        init_example_project(args.path, project_template)
+        _ = Dialects(args.dialect.lower())
```
tobymao, line 120: delete this

sqlmesh/core/context.py
treysp: That's what i originally had, but it errored because `Config.parse_obj()` is called on an incomplete Config dict in [core.config.loader load_config_from_env()](https://github.com/TobikoData/sqlmesh/blob/168fc6b06403277cb7895198dec711ccc3741850/sqlmesh/core/config/loader.py#L108).

No one `Config` must have a default dialect, so my logic was that the validation had to occur after all of them had been collected and we could see the entire set to determine whether any of them had one.

Am I thinking about this wrong?

sqlmesh/cli/main.py
tobymao: you don't need to do it at all i think, doesn't it already crash when you pass an invalid one in?

vchan: Are you able to use click's choice option?
> https://click.palletsprojects.com/en/8.1.x/options/#choice-options

sqlmesh/core/context.py
tobymao: @vchan @izeigerman what do you think?

sqlmesh/cli/main.py
treysp: yes, the checks i added are redundant - i deleted them all

sqlmesh/core/context.py
vchan: This doesn't check if any `Config` has a default dialect, it only checks the first config of `self.configs` has a default dialect, right?

tobymao: @vchan the goal is just to make model.default_dialect non null

treysp: > This doesn't check if any `Config` has a default dialect, it only checks the first config of `self.configs` has a default dialect, right?

Good point

vchan: I agree with @tobymao and this should be a pydantic validator. Maybe `core.config.loader.load_config_from_env` shouldn't return a `Config` object since env vars will probably only supplement/override config fields.

sqlmesh/cli/main.py
```diff
@@ -26,6 +27,10 @@ def _sqlmesh_version() -> str:
         return "0.0.0"
 
 
+def _sqlglot_dialects() -> str:
```
tobymao, line 30: put this in helpers and make it public since you import it in other places

sqlmesh/core/context.py
vchan: Looks like #1136 addressed the incomplete Config issue.

sqlmesh/core/config/loader.py
```diff
@@ -13,53 +13,73 @@
...
-            raise ConfigError(
-                f"Unsupported config file extension '{extension}' in config file '{path}'."
-            )
+
+        return {"python_config": python_config, "non_python_configs": non_python_configs}
+
+    project_config_dicts = _load_configs(project_paths)
+    personal_config_dicts = _load_configs(personal_paths)
+
+    no_dialect_err_msg = "Default model SQL dialect is a required configuration parameter - set it in the `model_defaults` key."
```
eakmanrq, line 57: Can this be changed to a validator now on Config and then remove this concept of project config and personal config?

treysp: I was hoping so, but both the Config [`update_with` method](https://github.com/TobikoData/sqlmesh/blob/eda2a7649827fe395908fb940a456fd232dcff58/sqlmesh/core/config/base.py#L112) and the [`update_field` method](https://github.com/TobikoData/sqlmesh/blob/eda2a7649827fe395908fb940a456fd232dcff58/sqlmesh/core/config/base.py#L81) it calls depend on the ability to instantiate Config objects without all the fields being specified (as would be required with a validator).

My sense is that modifying those will be more complex (in the sense of having side/downstream effects) than this project/non-project approach.

Alternatively, if we're willing to accept some edge cases (like someone hand-constructing project configs instead of using `init`), we can go back to the very first approach of just loading/collating all the configs and checking for a default dialect after that in Context.

sqlmesh/core/config/model.py
```diff
@@ -27,7 +27,7 @@ class ModelDefaultsConfig(BaseConfig):
     """
 
     kind: t.Optional[ModelKind]
-    dialect: t.Optional[str]
+    dialect: t.Optional[str | None] = None
```
tobymao, line 30: this is redundant, t.Optional[str] = None

tobymao: why are you changing this? why not make it dialect: str

sqlmesh/core/config/loader.py
```diff
@@ -13,53 +13,73 @@
...
-            non_python_configs.append(load_config_from_env())
-
-    if not non_python_configs and not python_config:
-        raise ConfigError(
-            "SQLMesh config could not be found. Point the cli to the right path with `sqlmesh -p`. If you haven't set up SQLMesh, run `sqlmesh init`."
-        )
+            non_python_config.update_with(load_config_from_env())
 
-    non_python_config = Config.parse_obj(merge_dicts(*non_python_configs))
+    python_config: t.Union[None, Config] = project_config_dicts["python_config"]
```
tobymao, line 76: t.Optional[Config]

sqlmesh/core/config/loader.py
```diff
@@ -13,53 +13,73 @@
...
     if python_config:
+        python_defaults = getattr(python_config, "model_defaults")
```
tobymao, line 78: why are you doing getattr? i think probably because of your bad type above? whenever you have to do these kinds of things you should question yourself and find a better way

sqlmesh/core/config/loader.py
```diff
@@ -13,15 +13,22 @@
...
     config_name: str = "config",
     load_from_env: bool = True,
 ) -> Config:
 
     visited_folders: t.Set[Path] = set()
     python_config: t.Optional[Config] = None
     non_python_configs = []
-    for path in paths:
+
+    if not project_paths or not any([path.exists() for path in project_paths]):
```
tobymao, line 26: the [] are not necessary

tobymao (approved):

PR #1128, opened 2023-07-11
Fix!: quote all identifiers in execute except those coming from fetchdf

GeorgeSittas:


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -47,11 +47,17 @@ def _ensure_pyspark_df(self, generic_df: DF) -> PySparkDataFrame:
             raise SQLMeshError("Ensure PySpark DF can only be run on a PySpark or Pandas DataFrame")
         return self.spark.createDataFrame(df)
 
-    def fetchdf(self, query: t.Union[exp.Expression, str]) -> pd.DataFrame:
-        return self.fetch_pyspark_df(query).toPandas()
+    def fetchdf(
+        self, query: t.Union[exp.Expression, str], normalize_identifiers: bool = True
```
eakmanrq, line 51: Should this actually default to `False` since this is a user facing API?

GeorgeSittas: Good point, will update. Thanks!

sqlmesh/core/context.py
```diff
@@ -130,16 +130,20 @@ def table(self, model_name: str) -> str:
         """
         return self._model_tables[model_name]
 
-    def fetchdf(self, query: t.Union[exp.Expression, str]) -> pd.DataFrame:
+    def fetchdf(
+        self, query: t.Union[exp.Expression, str], normalize_identifiers: bool = True
```
tobymao, line 134: normalize_identifiers should be False by default

tobymao: i don't think you even need this flag, it should always be false

sqlmesh/core/engine_adapter/base.py
```diff
@@ -790,14 +792,18 @@ def fetchall(
...
+        self, query: t.Union[exp.Expression, str], normalize_identifiers: bool = True
+    ) -> DF:
         """Fetches a DataFrame that can be either Pandas or PySpark from the cursor"""
-        self.execute(query)
+        self.execute(query, normalize_identifiers=normalize_identifiers)
         return self.cursor.fetchdf()
 
-    def fetchdf(self, query: t.Union[exp.Expression, str]) -> pd.DataFrame:
+    def fetchdf(
+        self, query: t.Union[exp.Expression, str], normalize_identifiers: bool = True
```
tobymao, line 803: i think this should by default be false

GeorgeSittas: Ditto

GeorgeSittas: Do we need to keep this flag at all?

tobymao: well, how do do you set it to true then if the class uses it internally?

GeorgeSittas: yea that's a good point, I assumed we'd just use `execute` directly but that's not necessarily the case

GeorgeSittas: converted to false

sqlmesh/core/engine_adapter/spark.py
GeorgeSittas: done

GeorgeSittas:
TODO for this PR:

- Update https://github.com/TobikoData/sqlmesh/blob/3508a6d29381d9ffbdf80db9e8e6da4c94581eb9/sqlmesh/core/engine_adapter/snowflake.py#L57

sqlmesh/core/context.py
```diff
@@ -139,7 +139,7 @@ def fetchdf(self, query: t.Union[exp.Expression, str]) -> pd.DataFrame:
         Returns:
             The default dataframe is Pandas, but for Spark a PySpark dataframe is returned.
         """
-        return self.engine_adapter.fetchdf(query)
+        return self.engine_adapter.fetchdf(query, normalize_identifiers=False)
```
tobymao, line 142: why do you need false here isn't the default already false?

tobymao (approved):

examples/wursthall/models/src/order_item_details.py
```diff
@@ -50,22 +51,15 @@ def execute(
...
-            id AS customer_id,
-            register_ds
-        FROM {customer_details_table_name}
-        WHERE
-            register_ds <= '{to_ds(end)}'
-        """
+        exp.select("id AS customer_id", "register_ds", copy=False)
+        .from_(customer_details_table_name, copy=False)
+        .where(f"register_ds <= '{to_ds(end)}'", copy=False),
+        quote_identifiers=True,
```
eakmanrq, line 57: Can we keep these as raw SQL? I don't want to leave the impression on new users that they have to write/learn sqlglot expressions.

tobymao: i think the tricky part is that this is a multi dialect project, so george can't just write raw sql because of differences between dialects, what he could do is do

parse_one(raw_sql, dialect="") which washes away the differences

GeorgeSittas: Exactly what Toby said -- happy to use parse_one instead if that makes it more clear.

tobymao: yea change it to parse_one, and leave a comment saying why you did it

GeorgeSittas: Changed.

PR #1130, opened 2023-07-11
Fix: Allow forward-only on added external models

eakmanrq:
Prior to this change if you added an external model then it would block being able to apply that change as forward-only. Now we only block if the model is an internal model.

### DIFF
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -185,6 +185,11 @@ class ContextDiff(PydanticModel):
     def has_snapshot_changes(self) -> bool:
         return bool(self.added or self.removed or self.modified_snapshots)
 
+    @property
+    def added_internal_models(self) -> t.Set[str]:
+        """Returns the set of added internal models."""
+        return {name for name in self.added if not self.snapshots[name].model_kind_name.is_external}
+
     def directly_modified(self, model_name: str) -> bool:
         """Returns whether or not a model was directly modified in this context.
 
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -186,9 +186,9 @@ class ContextDiff(PydanticModel):
         return bool(self.added or self.removed or self.modified_snapshots)
 
     @property
-    def added_internal_models(self) -> t.Set[str]:
+    def added_materialized_models(self) -> t.Set[str]:
         """Returns the set of added internal models."""
-        return {name for name in self.added if not self.snapshots[name].model_kind_name.is_external}
+        return {name for name in self.added if self.snapshots[name].model_kind_name.is_materialized}
 
     def directly_modified(self, model_name: str) -> bool:
         """Returns whether or not a model was directly modified in this context.
--- a/sqlmesh/core/model/common.py
+++ b/sqlmesh/core/model/common.py
@@ -4,7 +4,7 @@ import typing as t
 
 from pydantic import validator
 from sqlglot import exp
-from sqlglot.helper import seq_get, split_num_words
+from sqlglot.helper import seq_get
 
 from sqlmesh.core.dialect import parse
 from sqlmesh.utils.errors import ConfigError
@@ -19,7 +19,8 @@ def parse_model_name(name: str) -> t.Tuple[t.Optional[str], t.Optional[str], str
     Returns:
         A tuple consisting of catalog, schema, table name.
     """
-    return split_num_words(name, ".", 3)  # type: ignore
+    table = exp.to_table(name)
+    return table.catalog or None, table.db or None, table.name
 
 
 def parse_expression(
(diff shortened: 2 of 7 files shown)


sqlmesh/core/context_diff.py
```diff
@@ -185,6 +185,11 @@ def has_changes(self) -> bool:
     def has_snapshot_changes(self) -> bool:
         return bool(self.added or self.removed or self.modified_snapshots)
 
+    @property
+    def added_internal_models(self) -> t.Set[str]:
+        """Returns the set of added internal models."""
+        return {name for name in self.added if not self.snapshots[name].model_kind_name.is_external}
```
tobymao, line 191: maybe you should do is_symbolic, or even is_materialized?

sqlmesh/core/context_diff.py
```diff
@@ -185,6 +185,11 @@ def has_changes(self) -> bool:
     def has_snapshot_changes(self) -> bool:
         return bool(self.added or self.removed or self.modified_snapshots)
 
+    @property
+    def added_internal_models(self) -> t.Set[str]:
```
tobymao, line 189: change the name to represent a change to symbolic or materialized

sqlmesh/core/context_diff.py
izeigerman: Why? I thought we only needed to treat external models differently.

sqlmesh/core/model/common.py
```diff
@@ -19,7 +19,8 @@ def parse_model_name(name: str) -> t.Tuple[t.Optional[str], t.Optional[str], str
     Returns:
         A tuple consisting of catalog, schema, table name.
     """
-    return split_num_words(name, ".", 3)  # type: ignore
+    table = exp.to_table(name)
+    return table.catalog or None, table.db or None, table.name
```
tobymao, line 23: just table.name

eakmanrq, line 23: I was having an issue with this after fixing the forward-only for external models. Seems like it is time for us to move away from string splitting to get this information. I wasn't sure about dialect here with recent changes related to dialect.

tobymao, line 23: why do you need or None

tobymao: do we use this anywhere else?

sqlmesh/core/snapshot/definition.py
```diff
@@ -756,7 +756,7 @@ def version_get_or_generate(self) -> str:
     def physical_schema(self) -> str:
         if self.physical_schema_ is not None:
             return self.physical_schema_
-        _, schema, _ = parse_model_name(self.name)
+        schema = parse_model_name(self.name)[1]
```
tobymao, line 759: i wonder if we just do to_table.schema

sqlmesh/core/context_diff.py
tobymao: isn’t it more flexible? we have more options with forward only

sqlmesh/core/model/common.py
tobymao: right i forgot to remove this with my refactor

sqlmesh/core/snapshot/definition.py
eakmanrq: Good idea going to just replace this with `to_table`

eakmanrq: https://github.com/TobikoData/sqlmesh/pull/1130/commits/79db74c91162a57081296be3f2f2707bf90b4fb2

sqlmesh/core/model/common.py
eakmanrq: Removed: https://github.com/TobikoData/sqlmesh/pull/1130/commits/79db74c91162a57081296be3f2f2707bf90b4fb2

sqlmesh/core/context_diff.py
eakmanrq: Yeah it seems like the debate is if embedded or views should be treated like external models in terms of being allowed to be added in forward-only plans.

sqlmesh/core/context_diff.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/1130/commits/1e8e9505bef66caef0d870612578ea848a844646

sqlmesh/core/context_diff.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/1130/commits/1e8e9505bef66caef0d870612578ea848a844646

sqlmesh/core/snapshot/definition.py
```diff
@@ -194,8 +194,12 @@ def snapshot_id(self) -> SnapshotId:
 
     @property
     def qualified_view_name(self) -> QualifiedViewName:
-        catalog, schema, table = parse_model_name(self.name)
-        return QualifiedViewName(catalog=catalog, schema_name=schema, table=table)
+        view_name = exp.to_table(self.name)
+        return QualifiedViewName(
+            catalog=view_name.catalog or None,
```
eakmanrq, line 199: If there isn't a catalog it seems that sqlglot returns an empty string but we are expecting None

tobymao (approved):

PR #1131, opened 2023-07-11
Feat!: Promote only changed snapshots in dev environments

izeigerman:
To reduce the number of views created during creation of a new environment I propose to promote only new snapshots when targeting a dev environment.

If this is not a desired behavior, a new flag (`--promote-all`) has been added to enforce promotion of the entire graph.


sqlmesh/cli/main.py
```diff
@@ -264,6 +264,12 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     help="Disable automatic change categorization.",
     default=None,
 )
+@click.option(
+    "--promote-all",
```
tobymao, line 268: what is a concrete use case of wanting to promote all ?

sqlmesh/core/plan/definition.py
```diff
@@ -253,14 +257,26 @@ def environment(self) -> Environment:
...
         )
+
+        snapshots = [s.table_info for s in self.snapshots]
+        promoted_snapshot_ids = None
+        if self.is_dev and not self.promote_all:
+            names_to_promote = {
+                *self.context_diff.promoted_model_names,
+                *self.context_diff.added,
+                *self.context_diff.modified_snapshots,
+            } - self.context_diff.removed
```
eakmanrq, line 268: Want to make this into a property on context_diff? `promotable_snapshots` or something?

eakmanrq (commented):
Shouldn't this be breaking since the default behavior has now changed?

sqlmesh/cli/main.py
eakmanrq: Wanting to point dashboard(s) to an environment using a global find/replace instead of adjusting each reference that is included in the environment.

eakmanrq: Basically any use case where you want a complete representation of prod to validate against instead of just seeing a subset.

sqlmesh/core/plan/definition.py
```diff
@@ -253,14 +257,26 @@ def environment(self) -> Environment:
             if self.is_dev and self.environment_ttl is not None
             else None
         )
+
+        snapshots = [s.table_info for s in self.snapshots]
+        promoted_snapshot_ids = None
+        if self.is_dev and not self.promote_all:
+            names_to_promote = {
+                *self.context_diff.promoted_model_names,
```
tobymao, line 265: i'm not sure i understand this, what is promoted_model_names used for, why do we need it? why do we need to store in it the state sync?

sqlmesh/cli/main.py
tobymao: i see, i was confused on the purpose of this. this is more important in the blank environment case

sqlmesh/core/plan/definition.py
izeigerman: We need to track which models were promoted in the target environment during the previous plan application, so that we don't wipe them out while applying subsequent changes to the same environment.

tobymao: i see now. promoted_model_names ensures that you replace model names that were previously created

izeigerman: Previously promoted, yes!

sqlmesh/cli/main.py
izeigerman: > this is more important in the blank environment case

this is exactly it.

> once you have an environment and make a change, do we only ever promote things that have changed? eg there's no option for this one?

We only promote things that have changed + the things that were promoted in that environment previously. This is relevant to our other thread.


Additionally users can change their minds and apply a plan with `--promote-all` to an existing dev environment, in which case everything will be promoted.

sqlmesh/core/context_diff.py
```diff
@@ -53,6 +53,8 @@ class ContextDiff(PydanticModel):
     """New snapshots."""
     previous_plan_id: t.Optional[str]
     """Previous plan id."""
+    promoted_model_names: t.Set[str]
```
tobymao, line 56: previously_promoted

eakmanrq:
@izeigerman What do you think about having a config property that users can set to change this default behavior?

sqlmesh/core/context_diff.py
izeigerman: renamed

izeigerman:
@eakmanrq added a config property

eakmanrq (approved):
Great change! 💯

tobymao (approved):

PR #1133, opened 2023-07-11
Feat: Store magic command responses in variables

eakmanrq:
Currently we store context into a variable called `context`. For other commands that also returned values we did not store them. This updated the magic commands to store their responses (if there is one) in a variable that matches their command name to fit the pattern established by `context`. This means also updating `fetchdf` to store in `fetchdf` and let users assign it to another variable name if they want to reduce the complexity of that function and have it fit the pattern of other commands.

### DIFF
--- a/sqlmesh/magics.py
+++ b/sqlmesh/magics.py
@@ -362,14 +362,14 @@ class SQLMeshMagics(Magics):
         self._context.refresh()
         args = parse_argstring(self.evaluate, line)
 
-        df = self._context.evaluate(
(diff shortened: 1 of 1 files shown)


tobymao:
i don't think we should do this. setting variables should be an explicit action. mutating the name space is not really a good pattern. 

you can simply call context.x() if you need the result of the payload

eakmanrq:
Hmm. We do this by default though with context. Why is it allowed there but not others? I'm looking for consistency. It does help with debugging by knowing if a response will be stored or not and storing by default makes it so you know for sure that it will be available.

PR #1136, opened 2023-07-12
Fix!: Allow partials when loading configs + Personal Config Overrides Project

eakmanrq:
Breaking Change: Personal config files in your home directory now override config defined in a project. 

Prior to this change whenever a config was loaded it was validated as is and therefore it was expected any source of configs be complete and not partial. This made doing things like defining a password in a personal config not possible. This change combines all the config as dicts and then after that validates that combined representation. 

Python config cannot be partial and all the other configs override it if they are defined. I wonder though if we should just have it be that users either use Python or use YAML + Env vars. Open to feedback. 

Also included is some cleanup of tests where I didn't realize we could get `tmp_path` which is a `pathlib.Path` object instead of using `tmpdir`. 

Follow up work:
* Improve validation messages so if a config a incorrectly supplied we get a clean error instead of the current confusing error from Pydantic

### DIFF
--- a/sqlmesh/utils/__init__.py
+++ b/sqlmesh/utils/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import importlib
 import os
 import re
@@ -11,7 +12,7 @@ import typing as t
 import uuid
 from contextlib import contextmanager
 from copy import deepcopy
-from functools import lru_cache, wraps
+from functools import lru_cache, reduce, wraps
 from pathlib import Path
 
 T = t.TypeVar("T")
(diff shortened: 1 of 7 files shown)


sqlmesh/utils/__init__.py
```diff
@@ -236,3 +237,22 @@ def env_vars(environ: dict[str, str]) -> t.Iterator[None]:
...
+
+def merge_dicts(*args: t.Dict) -> t.Dict:
+    """
+    Merges dicts. Just does key collision replacement
+    """
+
+    def merge(a: t.Dict, b: t.Dict) -> t.Dict:
+        for b_key, b_value in b.items():
+            a_value = a.get(b_key)
+            if b_key in a and isinstance(a_value, dict) and isinstance(b_value, dict):
```
tobymao, line 250: b_key in a is unnecessary because you check isinstance(a_value, dict)

sqlmesh/utils/__init__.py
```diff
@@ -236,3 +237,22 @@ def env_vars(environ: dict[str, str]) -> t.Iterator[None]:
...
+                merge(a_value, b_value)
+            elif isinstance(b_value, dict):
+                a[b_key] = copy.deepcopy(b_value)
```
tobymao, line 253: do you need the deepcopy here?

eakmanrq: Yes or else we end up with mutations on the input dicts.

izeigerman (approved):

PR #1137, opened 2023-07-12
Refactor!: Replace snapshot model with snapshot node

crericha:
Setting the stage for moving audits to their own nodes.


sqlmesh/core/node.py
```diff
@@ -0,0 +1,169 @@
...
+            from croniter import CroniterBadCronError, croniter
+
+            try:
+                croniter(cron)
+            except CroniterBadCronError:
+                raise ConfigError(f"Invalid cron expression '{cron}'")
+        return cron
+
+    @property
+    def batch_size(self) -> t.Optional[int]:
```
tobymao, line 69: what do you envision this for?

sqlmesh/core/node.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def is_date_granularity(self) -> bool:
+        return self in (IntervalUnit.YEAR, IntervalUnit.MONTH, IntervalUnit.DAY)
+
+
+class Node(PydanticModel):
+    """
+    Args:
+        start: The earliest date that the model will be backfilled for. If this is None,
```
tobymao, line 37: model?

sqlmesh/core/node.py
```diff
@@ -0,0 +1,169 @@
...
+    DAY = "day"
+    HOUR = "hour"
+    MINUTE = "minute"
+
+    @property
+    def is_date_granularity(self) -> bool:
+        return self in (IntervalUnit.YEAR, IntervalUnit.MONTH, IntervalUnit.DAY)
+
+
+class Node(PydanticModel):
```
eakmanrq, line 34: Move node specific fingerprinting logic into this node class.

sqlmesh/core/node.py
crericha: The reason I added batch_size to node is so the scheduler and engine adapters wouldn't need model specific logic to execute nodes. I envision it as a no-op (default value of None) for non-model nodes. Model will override to model specific configuration.

sqlmesh/core/model/definition.py
```diff
@@ -628,6 +636,104 @@ def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
...
+        if isinstance(self.kind, IncrementalByTimeRangeKind):
+            data.append(self.kind.time_column.column)
+            data.append(self.kind.time_column.format)
+        elif isinstance(self.kind, IncrementalByUniqueKeyKind):
+            data.extend(self.kind.unique_key)
+
+        return hash_data(data)
+
+    @property
+    def _additional_data_for_hash(self) -> t.List[str]:
```
izeigerman, line 672: [Nit] I guess the name doesn't make it clear that this is for data hash specifically

sqlmesh/core/node.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    Node is the core abstraction for entity that can be executed within the orchestrator.
```
izeigerman, line 39: [Nit] I don't think we refer to `orchestrator` anywhere in the code. Maybe `scheduler`?

sqlmesh/migrations/v0018_rename_snapshot_model_to_node.py
```diff
@@ -0,0 +1,46 @@
...
+    snapshots_table = f"{schema}._snapshots"
+
+    new_snapshots = []
+
+    for name, identifier, version, snapshot, kind_name in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name").from_(snapshots_table)
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        parsed_snapshot["node"] = parsed_snapshot["model"]
+        parsed_snapshot.pop("model")
```
izeigerman, line 20: Doesn't have to be 2 lines:
```
parsed_snapshot["node"] = parsed_snapshot.pop("model")
```

izeigerman (approved):
🚢 🤞

sqlmesh/core/model/definition.py
```diff
@@ -926,6 +1043,22 @@ def _query_renderer(self) -> QueryRenderer:
             )
         return self.__query_renderer
 
+    @property
+    def _additional_data_for_hash(self) -> t.List[str]:
+        query = self.query if self.hash_raw_query else self.render_query() or self.query
+        data = [query.sql(comments=False), *super()._additional_data_for_hash]
+
+        for macro_name, macro in sorted(self.jinja_macros.root_macros.items(), key=lambda x: x[0]):
```
tobymao, line 1051: i don't know if you need key here since by default it will sort the tuple by the first  element

tobymao (approved):

sqlmesh/core/model/definition.py
crericha: Refactored

PR #1144, opened 2023-07-12
Feat!: Add support for forward_only and disable_restatement flags to incremental models

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -409,12 +411,16 @@ def _add_restatements(self, restate_models: t.Iterable[str]) -> None:
...
-                d for d in downstream if snapshots[d].is_materialized and not snapshots[d].is_seed
+                d
+                for d in downstream
+                if snapshots[d].is_materialized
+                and (not snapshots[d].model.disable_restatement or self.is_dev)
+                and not snapshots[d].is_seed
             ]
 
             if not downstream:
                 raise PlanError(
```
tobymao, line 422: so just double checking. if you choose to restate something, and a disabled_restatement is in the graph, it will simply be excluded, but it won't be restated is that correct? in that case the down stream of the disabled restatement will still retrigger, but it won't get the new data from this branch

sqlmesh/core/model/definition.py
```diff
@@ -564,6 +564,14 @@ def depends_on_past(self) -> bool:
 
         return self._depends_on_past
 
+    @property
+    def forward_only(self) -> bool:
+        return isinstance(self.kind, _Incremental) and self.kind.forward_only
```
eakmanrq, line 569: add an `is_incremental` property to `ModelKindMixin` which is either `is_incremental_by_time_range` or `is_incremental_by_unique_key`?

sqlmesh/core/snapshot/definition.py
```diff
@@ -986,6 +986,8 @@ def _model_metadata_hash(model: Model, audits: t.Dict[str, Audit]) -> str:
         json.dumps(model.mapping_schema, sort_keys=True),
         *sorted(model.tags),
         *sorted(model.grain),
+        str(model.forward_only),
+        str(model.disable_restatement),
```
tobymao, line 990: you'll need an empty migration file for this

sqlmesh/core/model/definition.py
izeigerman: I think I still need the `isinstance` check to make mypy happy, right?

sqlmesh/core/plan/definition.py
izeigerman: > it will simply be excluded, but it won't be restated is that correct? in that case the down stream of the disabled restatement will still retrigger, but it won't get the new data from this branch

All of these statements are correct

izeigerman: > disabled restatement will still retrigger, but it won't get the new data from this branch

those downstream might also depend on other upstream that is being restated in addition to the target disabled model.

PR #1146, opened 2023-07-13
Feature: Add connection config for GCP managed Postgres

crericha:
GCP has a special connector that integrates their gcloud auth with their postgres managed service.


eakmanrq (approved):
Look good. Some things to consider:
1. Updating the docs to include this connector. It seems like there are two ways to connect which are with a password or using iam
2. Adding a validator to make sure that either password or iam are set and make sure they can't both be set

crericha:
Will add those in a follow-on PR. Thanks for the review @eakmanrq !

PR #1148, opened 2023-07-13
Feat: add promote_all option to UI

mykalmax:
When apply to `new` env without changes

Screenshot 2023-07-13 at 8 05 16 AM

Screenshot 2023-07-13 at 8 05 22 AM

but if set to FALSE 

Screenshot 2023-07-13 at 8 05 05 AM

### DIFF
--- a/web/client/openapi.json
+++ b/web/client/openapi.json
@@ -701,7 +701,8 @@
               "skip_backfill": false,
               "no_gaps": false,
               "forward_only": false,
-              "no_auto_categorization": false
+              "no_auto_categorization": false,
+              "promote_all": false
             }
           },
           "categories": {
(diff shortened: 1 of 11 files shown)


izeigerman:
@eakmanrq @mykalmax should we then set `promote_all` to true by default in config to avoid breaking UI like this?

mykalmax:
> How hard do you think the more user friendly flow will be? If it is not too bad then we can leave it as is but if we think it will take a while to ship then changing the default might be best.

It should not be hard, but i've not thought about it yet. I think, I can have something ready some time tomorrow. It just in case of a demo current UI state is not handling `promote_all = False` correctly

mykalmax:
What behavior would be expected in case `promote_all == False`, just `No Change` ?

Screenshot 2023-07-13 at 10 04 09 AM

vchan (approved):

PR #1151, opened 2023-07-13
Feat: Improve consistent styling for inputs + use radix/select for selectors

mykalmax:


web/client/src/library/components/input/Textfield.tsx
```diff
@@ -0,0 +1,49 @@
...
+      onKeyDown,
+    }: PropsTextfield,
+    ref?: React.Ref<HTMLInputElement>,
+  ): JSX.Element {
+    return (
+      <input
+        ref={ref}
+        className={clsx(
+          'placeholder:text-neutral-300 dark:placeholder:text-neutral-700',
+          isTrue(disabled) && 'opacity-50 cursor-not-allowed',
```
vchan, line 35: Is `isTrue()` needed here? It's not present at line 184 in Selector.tsx: https://github.com/TobikoData/sqlmesh/pull/1151/files#diff-0c4605cd8c3c53a54b557d9f959536740216efb4fd2a3ecbe2445d597e41be0dR184

web/client/src/library/pages/docs/SourceList.tsx
```diff
@@ -24,12 +24,18 @@ export default function SourceList({
...
-            setFilter(e.target.value)
-          }}
-        />
+        >
+          {({ className }) => (
+            <Input.Textfield
+              className={clsx(className, 'w-full')}
+              value={filter}
+              placeholder="Filter models"
+              onInput={e => {
```
vchan, line 33: Missing type? Also, when and why do we need `e.stopPropagation()`?

web/client/src/library/components/editor/EditorInspector.tsx
```diff
@@ -377,68 +381,88 @@ function FormActionsModel({
...
-                  ...form,
-                  start: e.target.value ?? '',
-                })
-              }}
-            />
+            >
+              {({ className }) => (
+                <Input.Textfield
+                  className={clsx(className, 'w-full')}
+                  placeholder="02/11/2023"
```
vchan, line 392: Probably outside the scope of this PR, but should these date fields use YYYY-MM-DD format?

web/client/src/library/components/input/Selector.tsx
```diff
@@ -0,0 +1,194 @@
...
+                className="w-4"
+                aria-hidden="true"
+              />
+            </Select.ScrollDownButton>
+          </Select.Content>
+        </Select.Portal>
+      </Select.Root>
+    )
+
+    // return (
```
vchan, line 99: Let's delete this. We can always restore it with git if we need to.

vchan (approved):

web/client/src/library/components/input/Textfield.tsx
mykalmax: yes, it is not needed

web/client/src/library/pages/docs/SourceList.tsx
mykalmax: it stops event from being despatch up the DOM. So if you press `ArrowDown` on a page it should scroll page a bit , but in some cases you may not want that default behavior , like when you trying to open Menu Items list by pressing `ArrowDown`. In such as cases (or any cases we know that event listener is final destination) we can `stopPropagation` to not trigger accidentally handler on other DOM element

PR #1153, opened 2023-07-13
Feat: Determine time_column from bigquery partition_by config

crericha:
We can discern the time column for incremental by time models automatically using the dbt bigquery partition_by config, which specifies the column.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from pydantic import validator
+from pydantic import Field, validator
 from sqlglot import exp
 from sqlglot.helper import ensure_list
 
@@ -66,7 +66,7 @@ class ModelConfig(BaseModelConfig):
 
     # sqlmesh fields
     sql: SqlStr = SqlStr("")
-    time_column: t.Optional[str] = None
+    time_column_: t.Optional[str] = Field(None, alias="time_column")
     cron: t.Optional[str] = None
     dialect: t.Optional[str] = None
     batch_size: t.Optional[int] = None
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/model.py
```diff
@@ -119,6 +119,14 @@ def _validate_partition_by(cls, v: t.Any) -> t.Union[t.List[str], t.Dict[str, t.
         },
     }
 
+    @property
+    def time_column(self) -> t.Optional[str]:
+        if self.time_column_:
+            return self.time_column_
+        if isinstance(self.partition_by, dict) and self.partition_by["data_type"] != "int64":
+            return self.partition_by["field"]
```
izeigerman, line 127: I think we should also make sure that the incremental strategy is `insert_overwrite`.

sqlmesh/dbt/model.py
```diff
@@ -119,6 +119,14 @@ def _validate_partition_by(cls, v: t.Any) -> t.Union[t.List[str], t.Dict[str, t.
         },
     }
 
+    @property
+    def time_column(self) -> t.Optional[str]:
+        if self.time_column_:
+            return self.time_column_
+        if isinstance(self.partition_by, dict) and self.partition_by["data_type"] != "int64":
```
izeigerman, line 126: Wouldn't this fail if an engine other than BQ supports partition_by input as dictionary?

crericha: No other engine does.

sqlmesh/dbt/model.py
crericha: Yes, good idea.

izeigerman (approved):

PR #1156, opened 2023-07-14
Fix: Fix windows paths for executables

eakmanrq:
Prior to this change if a path was generated using a Windows computer it would generate a different fingerprint than a linux/macos computer. This normalizes them.

### DIFF
--- a/sqlmesh/migrations/v0016_fix_windows_path.py
+++ b/sqlmesh/migrations/v0016_fix_windows_path.py
@@ -18,13 +18,11 @@ def migrate(state_sync):  # type: ignore
         parsed_snapshot = json.loads(snapshot)
         model = parsed_snapshot["model"]
         python_env = model.get("python_env")
-        if not python_env:
-            continue
-        for py_definition in python_env.values():
-            path = py_definition.get("path")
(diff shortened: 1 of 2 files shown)


tobymao:
this is a breaking change and requires a migration trigger

sqlmesh/migrations/v0016_fix_windows_path.py
```diff
@@ -0,0 +1,53 @@
...
+        parsed_snapshot = json.loads(snapshot)
+        model = parsed_snapshot["model"]
+        python_env = model.get("python_env")
+        if not python_env:
+            continue
+        for py_definition in python_env.values():
+            path = py_definition.get("path")
+            if not path:
+                continue
+            py_definition["path"] = path.replace("\\", "/")
```
eakmanrq, line 27: @tobymao This assumes that any backslash in the path is part of a windows path.

tobymao (approved):

tobymao:
do we have this problem anywhere else?

eakmanrq:
> do we have this problem anywhere else?

Not that I can see

PR #1162, opened 2023-07-15
Fix: Fix Windows faulthandler and seed path

eakmanrq:
Two fixes:
1. Windows does not support `faulthandler.register` so I just check if the attribute exists and if so then we error. Docs: https://docs.python.org/3/library/faulthandler.html#faulthandler.register
2. If a user wrote a seed path in windows format then that is how it would be stored in state. Now we ensure that windows paths are converted to linux.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -344,12 +344,6 @@ def _model_sql(self: Generator, expression: Model) -> str:
 
 
 def _model_kind_sql(self: Generator, expression: ModelKind) -> str:
-    # props = {
-    #     prop.this: self.sql(prop, "value") for prop in expression.expressions
-    # }
-    # if "path" in props:
-    #     # Replace windows paths with unix paths
-    #     props["path"] = props["path"].replace("\\", "/")
     props = ",\n".join(
         self.indent(f"{prop.this} {self.sql(prop, 'value')}") for prop in expression.expressions
     )
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -267,8 +267,11 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
         )
     else:
         value = self._parse_bracket(self._parse_field(any_token=True))
-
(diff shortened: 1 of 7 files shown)


sqlmesh/core/dialect.py
```diff
@@ -267,8 +267,11 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
         )
     else:
         value = self._parse_bracket(self._parse_field(any_token=True))
-
-    return self.expression(exp.Property, this=key.name.lower(), value=value)
+    name = key.name.lower()
+    if name == "path" and value:
+        # Make sure if we get a windows path that it is converted to posix
+        value = exp.Literal.string(value.this.replace("\\", "/"))
```
tobymao, line 273: pathlib.path(...).as_posix() or something rather than regex

sqlmesh/migrations/v0017_fix_windows_seed_path.py
```diff
@@ -0,0 +1,47 @@
...
+
+    new_snapshots = []
+
+    for name, identifier, version, snapshot, kind_name in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name").from_(snapshots_table)
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        model_kind = parsed_snapshot["model"]["kind"]
+        if "path" in model_kind:
+            model_kind["path"] = model_kind["path"].replace("\\", "/")
```
tobymao, line 21: let's use posix here as it's more robust

sqlmesh/core/dialect.py
eakmanrq: It doesn't work like that. 
```
pathlib.Path("..\..\test.py").as_posix()
> '..\\..\test.py'
```

That approach only works if you are on Windows.

sqlmesh/migrations/v0017_fix_windows_seed_path.py
eakmanrq: see above reply

tobymao (approved):

PR #1163, opened 2023-07-15
Fix: Allow reverting to previous versions

eakmanrq:
Prior to this change if someone makes a change and the fingerprint is for a snapshot that already exists then we would throw an error saying that the snapshot already has intervals and that is unexpected. This enables us to do this again.

### DIFF
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -337,9 +337,9 @@ class Plan:
 
         for child in self.indirectly_modified[snapshot.name]:
             child_snapshot = self.context_diff.snapshots[child]
-            # If the child already has a version then the snapshot must have previously existed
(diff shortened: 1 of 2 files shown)


sqlmesh/core/plan/definition.py
```diff
@@ -337,6 +337,10 @@ def set_choice(self, snapshot: Snapshot, choice: SnapshotChangeCategory) -> None
 
         for child in self.indirectly_modified[snapshot.name]:
             child_snapshot = self.context_diff.snapshots[child]
+            # If the child already has a version then the snapshot must have previously existed
+            # and therefore we don't want to change its categorization.
+            if child_snapshot.version:
```
tobymao, line 342: change it to self.is_new_snapshot like 331, that's probably more robust

tobymao (approved):

PR #1164, opened 2023-07-16
Fix: Tab not showing after refresh + store unsaved changes

mykalmax:


web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -228,15 +237,30 @@ export default function PageIDE(): JSX.Element {
...
-
-        if (storedTabsId === file.id) {
-          selectTab(tab)
-        } else {
-          addTab(tab)
-        }
+    console.log('restoreEditorTabsFromSaved', {
+      storedTabsIds,
+      storedTabsId,
+    })
```
vchan, line 243:
```suggestion
```

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -228,15 +237,30 @@ export default function PageIDE(): JSX.Element {
...
+        storedTab.file.content = content ?? ''
+      } else {
+        storedTab = createTab(file)
+
+        storedTab.file.content = content ?? storedTab.file.content
+      }
+
+      addTab(storedTab)
+
+      if (storedTabsId === id || (isNil(id) && storedTabsId === content)) {
```
vchan, line 262: So `storedTabsId` can equal id or content? Shouldn't it only ever be able to equal id?

web/client/src/context/editor.ts
```diff
@@ -101,19 +107,49 @@ export const useStoreEditor = create<EditorStore>((set, get) => ({
...
+
+        tabs.push(output)
+      }
+    }
+
+    let id: ID | undefined
+
+    if (isNotNil(s.tab)) {
+      if (s.tab.file.isLocal && s.tab.file.isChanged) {
+        id = s.tab.file.content
```
vchan, line 137: It's a bit confusing when id can sometimes be content and sometimes be id.

web/client/src/context/editor.ts
```diff
@@ -101,19 +107,49 @@ export const useStoreEditor = create<EditorStore>((set, get) => ({
...
+          content: tab.file.content,
+        })
+      } else {
+        const output: StoredTab = {
+          id: tab.file.id,
+        }
+
+        if (tab.file.isChanged) {
+          output.content = tab.file.content
+        }
```
vchan, line 127: So this new output tab won't always have content set?

mykalmax:
We have 2 types of tabs, from selecting a `File` or `Custom SQL`, if `File` has no unsaved changes , no need to store `content` cause we still calling `API` to get `File` content, but we need to match tab to file by `id`. So for `File` - `id` should always be there but `content` is optional. For `Custom SQL` we don't need `id` to match to anything so the important part is `content`. I was trying being clever but, you are right it is confusing and not necessary so I removed all that and just storing ids

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -228,17 +236,23 @@ export default function PageIDE(): JSX.Element {
...
+
+      return storedTab
     })
+    const tab = tabs.find(tab => tab.file.id === storedTabId)
+
+    addTabs(tabs)
+
+    if (isNil(tab)) return
+
+    selectTab(tab)
```
vchan, line 255: Nit: I think this is slightly easier to follow but I don't feel strongly about it so feel free to ignore.
```suggestion
    if (isNotNil(tab)) {
      selectTab(tab)
    }
```

vchan (approved):
Thank you for explaining the different types of tabs.

PR #1171, opened 2023-07-18
Feat: Support custom schema name for state metadata

eakmanrq:
Allow users to customize the schema name for state metadata. If None or empty string is provided then no schema is used and therefore the default schema for the connection/user is used instead.

### DIFF
--- a/sqlmesh/migrations/v0001_init.py
+++ b/sqlmesh/migrations/v0001_init.py
@@ -9,12 +9,11 @@ from sqlglot import exp
 def migrate(state_sync):  # type: ignore
     engine_adapter = state_sync.engine_adapter
     schema = state_sync.schema
-    if schema:
-        engine_adapter.create_schema(schema)
     snapshots_table = f"_snapshots"
     environments_table = f"._environments"
     versions_table = f"._versions"
     if schema:
+        engine_adapter.create_schema(schema)
         snapshots_table = f"{schema}.{snapshots_table}"
         environments_table = f"{schema}.{environments_table}"
         versions_table = f"{schema}.{versions_table}"
--- a/sqlmesh/migrations/v0001_init.py
+++ b/sqlmesh/migrations/v0001_init.py
@@ -9,9 +9,9 @@ from sqlglot import exp
 def migrate(state_sync):  # type: ignore
     engine_adapter = state_sync.engine_adapter
     schema = state_sync.schema
-    snapshots_table = f"_snapshots"
-    environments_table = f"._environments"
-    versions_table = f"._versions"
+    snapshots_table = "_snapshots"
+    environments_table = "_environments"
+    versions_table = "_versions"
     if schema:
         engine_adapter.create_schema(schema)
         snapshots_table = f"{schema}.{snapshots_table}"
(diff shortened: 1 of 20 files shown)


sqlmesh/migrations/v0001_init.py
```diff
@@ -9,11 +9,14 @@
 def migrate(state_sync):  # type: ignore
     engine_adapter = state_sync.engine_adapter
     schema = state_sync.schema
-    engine_adapter.create_schema(schema)
-
-    snapshots_table = f"{schema}._snapshots"
-    environments_table = f"{schema}._environments"
-    versions_table = f"{schema}._versions"
+    snapshots_table = f"_snapshots"
+    environments_table = f"._environments"
```
izeigerman, line 13: `f` is no longer required here. Also how come does this work with trailing `.`?

sqlmesh/migrations/v0003_move_batch_size.py
```diff
@@ -6,7 +6,9 @@
 def migrate(state_sync):  # type: ignore
     """Move batch_size from the model and into the kind."""
 
-    snapshots_table = f"{state_sync.schema}._snapshots"
+    snapshots_table = f"_snapshots"
```
izeigerman, line 9: Ditto on `f`.

sqlmesh/migrations/v0004_environmnent_add_finalized_at.py
```diff
@@ -4,7 +4,9 @@
 
 def migrate(state_sync):  # type: ignore
     engine_adapter = state_sync.engine_adapter
-    environments_table = f"{state_sync.schema}._environments"
+    environments_table = f"_environments"
```
izeigerman, line 7: Ditto here and further

sqlmesh/migrations/v0005_create_seed_table.py
```diff
@@ -4,7 +4,9 @@
 
 def migrate(state_sync):  # type: ignore
     engine_adapter = state_sync.engine_adapter
-    seeds_table = f"{state_sync.schema}._seeds"
+    seeds_table = f"._seeds"
```
izeigerman, line 7: Trailing dot

sqlmesh/schedulers/airflow/util.py
```diff
@@ -36,13 +38,23 @@
...
         connection = Connection.get_connection_from_secrets(SQLMESH_STATE_CONN_ID)
 
+        connection_config_dict = json.loads(connection.extra)
+        state_schema = connection_config_dict.get("state_schema", state_schema)
+        if "type" not in connection_config_dict:
+            logger.info(
+                "SQLMesh connection in Airflow did not have type defined. "
+                "Therefore using Airflow database connection"
+            )
+            raise AirflowException
```
izeigerman, line 52: Are we raising a type here? Shouldn't it be `AirflowException()`?

eakmanrq: I think this works as expected:
```
def main():
    try:
        print("here")
        raise AirflowException
        print("here2")
    except AirflowException:
        print("exception")
```
```
here
exception
```

izeigerman (approved):

PR #1173, opened 2023-07-18
Feat: Introduce an additional 'incremental_unmanaged' model kind to support existing incremental mode in dbt projects

izeigerman:


sqlmesh/core/model/__init__.py
```diff
@@ -14,6 +14,7 @@
 from sqlmesh.core.model.kind import (
     IncrementalByTimeRangeKind,
     IncrementalByUniqueKeyKind,
+    IncrementalUnsafeKind,
```
tobymao, line 17: is this the best name for this?

izeigerman: Open to suggestions. I was definitely aiming at discouraging people from using it

tobymao: IncrementalManual
IncrementalUnmanaged

izeigerman: oh, Unmanaged is pretty good

tobymao (approved):

eakmanrq (approved):

sqlmesh/core/plan/definition.py
```diff
@@ -441,13 +441,14 @@ def _add_restatements(self, restate_models: t.Iterable[str]) -> None:
...
-                if snapshots[d].is_materialized
-                and (not snapshots[d].model.disable_restatement or self.is_dev)
-                and not snapshots[d].is_seed
+                d for d in downstream if snapshots[d].is_materialized and not snapshots[d].is_seed
             ]
 
+            if not self.is_dev:
+                for d in downstream:
+                    if snapshots[d].model.disable_restatement:
+                        raise PlanError(f"Restatement is disabled for model '{d}'.")
```
crericha, line 450: From a user perspective, do we want to raise on the first model with this condition or do we want to return all models with the condition? My personal opinion is that the latter would be more informative.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -130,6 +130,7 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
             start=start,
             end=end,
             latest=latest,
+            dbt_is_incremental=snapshot.is_incremental_unmanaged and bool(snapshot.intervals),
```
crericha, line 133: Do you think we need to call it dbt_is_incremental here? We've jumped through a lot of hoops to abstract dbt away from core code. Do you feel its okay to let it into the evaluator?

sqlmesh/dbt/model.py
```diff
@@ -115,22 +116,10 @@ def _validate_partition_by(cls, v: t.Any) -> t.Union[t.List[str], t.Dict[str, t.
         **BaseModelConfig._FIELD_UPDATE_STRATEGY,
         **{
             "sql": UpdateStrategy.IMMUTABLE,
-            "time_column_": UpdateStrategy.IMMUTABLE,
+            "time_column": UpdateStrategy.IMMUTABLE,
         },
     }
 
-    @property
-    def time_column(self) -> t.Optional[str]:
```
crericha, line 123: Is the rationale for removing this that if someone is switching to `sqlmesh_incremental`, we'll want them to add the `time_column` param too?

sqlmesh/dbt/model.py
```diff
@@ -187,16 +177,21 @@ def model_kind(self, target: TargetConfig) -> ModelKind:
...
-            raise ConfigError(
-                f"{self.sql_name}: Incremental materialization requires either a "
-                f"time_column ({collection_to_str(INCREMENTAL_BY_TIME_STRATEGIES)}) or a "
-                f"unique_key ({collection_to_str(INCREMENTAL_BY_UNIQUE_KEY_STRATEGIES.union(['none']))}) configuration."
+            logger.warning(
+                "Incremental materialization in model '%s' requires either a time_column (%s) or a unique_key (%s) configuration",
+                self.sql_name,
+                collection_to_str(INCREMENTAL_BY_TIME_STRATEGIES),
+                collection_to_str(INCREMENTAL_BY_UNIQUE_KEY_STRATEGIES.union(["none"])),
+            )
```
crericha, line 190: The warning reads to me as something is required and I need to take action for my model to run. What do you want the user to take away from this warning? Can we be more explicit?

sqlmesh/dbt/model.py
```diff
@@ -187,16 +177,21 @@ def model_kind(self, target: TargetConfig) -> ModelKind:
...
+            strategy = self.incremental_strategy or target.default_incremental_strategy(
+                IncrementalUnmanagedKind
             )
+            return IncrementalUnmanagedKind(insert_overwrite=strategy == "insert_overwrite")
```
crericha, line 194: How do we know whether or not the incremental by time or incremental by unique key strategies will work if a time_column or unique_key is specified? If the only where clause is an `is_incremental` block, I could be wrong, but I think we need to fall back to `IncrementalUnmanagedKind`.

sqlmesh/core/snapshot/evaluator.py
izeigerman: Yeah, at this point I don't feel like this is something we should support outside dbt. This might change in future

sqlmesh/dbt/model.py
izeigerman: Yes, that would make their intent explicit

sqlmesh/dbt/model.py
izeigerman: I don't understand this comment. If there is `unique_key` or `time_column` available then the user's intent is pretty clear.

sqlmesh/core/snapshot/evaluator.py
crericha: I agree that this is a dbt compatability mode feature. I don't understand why the kind was named in a generic way, but the param wasn't. More of a nit.

sqlmesh/dbt/model.py
crericha: unique_key is a dbt config field. It wasn't necessarily added as part of a sqlmesh migration.

crericha: unique_key is a dbt config field. It wasn't necessarily added as part of a sqlmesh migration.

sqlmesh/core/snapshot/evaluator.py
tobymao: i'd rather name it something generic, i could see us supporting this for id based workflows and you don't want to use merge unique

sqlmesh/dbt/model.py
crericha: Please update the dbt integration docs to reflect this. It's a bit dangerous because the user will not get an error if they have a mismatch (time_column, but not {% sqlmesh_incremental is defined %}. If they get it wrong, they are basically doing a FULL kind.

izeigerman: Going to update the documentation once all relevant changes are in.

Issue #1175, opened 2023-07-18
sqlmesh invalidate should also remove dataset of the invalidated environment


zzztimbo:
Given
```
sqlmesh plan pr123
```

When
```
sqlmesh invalidate pr123
```

Then the next time I run 
```
sqlmesh run
```

I would expect the dataset `{project}__pr123` to no longer be there. 

Running
```
bq ls --max_results=100000 | rg _pr123
```
should yield no results.

PR #1177, opened 2023-07-18
Feat: Drop schemas for expired environments

eakmanrq:
Resolves: https://github.com/TobikoData/sqlmesh/issues/1175

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1164,8 +1164,19 @@ class Context(BaseContext):
 
     def _run_janitor(self) -> None:
         expired_environments = self.state_sync.delete_expired_environments()
+        expired_schemas: t.Set[str] = set()
         for expired_environment in expired_environments:
             self.snapshot_evaluator.demote(expired_environment.snapshots, expired_environment.name)
+            expired_schemas.update(
+                {
+                    exp.to_table(
+                        snapshot.qualified_view_name.for_environment(expired_environment.name)
+                    ).db
+                    for snapshot in expired_environment.snapshots
+                }
+            )
+        for expired_schema in expired_schemas:
+            self.engine_adapter.drop_schema(expired_schema, ignore_if_not_exists=True, cascade=True)
 
         expired_snapshots = self.state_sync.delete_expired_snapshots()
         self.snapshot_evaluator.cleanup(expired_snapshots)
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1169,9 +1169,7 @@ class Context(BaseContext):
             self.snapshot_evaluator.demote(expired_environment.snapshots, expired_environment.name)
             expired_schemas.update(
                 {
-                    exp.to_table(
-                        snapshot.qualified_view_name.for_environment(expired_environment.name)
-                    ).db
+                    snapshot.qualified_view_name.schema_for_environment(expired_environment.name)
                     for snapshot in expired_environment.snapshots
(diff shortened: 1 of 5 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -239,6 +239,16 @@ def delete_expired_environments(self) -> t.List[Environment]:
...
+        schema_names: t.Set[str] = set()
+        for environment in environments:
+            schema_names.update(
+                {
+                    exp.to_table(snapshot.qualified_view_name.for_environment(environment.name)).db
+                    for snapshot in environment.snapshots
+                }
+            )
+        for schema_name in schema_names:
+            self.engine_adapter.drop_schema(schema_name, ignore_if_not_exists=True, cascade=True)
```
izeigerman, line 251: I don't think this logic belongs here. Our goal is to remove schemas in the DW, not in the state.

izeigerman: This is where we delete individual views right now:
https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/context.py#L1167-L1168
Airflow: https://github.com/TobikoData/sqlmesh/blob/37d6a7fa3792a2fb641a279ca82c088aff4897cb/sqlmesh/engines/commands.py#L117-L118

eakmanrq: Man I keep getting this mixed up. I think it is because the SQLMesh installs I work with are storing state in the DW so I don't differentiate. Will need to take another look at this.

sqlmesh/core/context.py
```diff
@@ -1164,8 +1164,19 @@ def _load_configs(
 
     def _run_janitor(self) -> None:
         expired_environments = self.state_sync.delete_expired_environments()
+        expired_schemas: t.Set[str] = set()
         for expired_environment in expired_environments:
             self.snapshot_evaluator.demote(expired_environment.snapshots, expired_environment.name)
+            expired_schemas.update(
+                {
+                    exp.to_table(
+                        snapshot.qualified_view_name.for_environment(expired_environment.name)
```
tobymao, line 1173: can't this just be schema_for_environment?

eakmanrq: It is a `SnapshotTableInfo` and not a `Snapshot`.

tobymao: snapshot.qualified_view_name.schema_for_environment(env)

instead of to_table(...).db ?

eakmanrq: Oh nm I see now on qualified view

sqlmesh/engines/commands.py
```diff
@@ -114,8 +114,18 @@ def cleanup(
 ) -> None:
     if isinstance(command_payload, str):
         command_payload = CleanupCommandPayload.parse_raw(command_payload)
+
+    expired_schemas: t.Set[str] = set()
     for environment in command_payload.environments:
         evaluator.demote(environment.snapshots, environment.name)
+        expired_schemas.update(
```
izeigerman, line 121: Do we even need to demote individual snapshots now?

eakmanrq: Yeah good point. All current engines support CASCADE and I guess if we end up adding one that doesn't support it then we can just add the ability to drop the individual objects within the engine adapter method.

izeigerman (approved):

PR #1178, opened 2023-07-19
Feat: add lookforward to incremental models for lambda live views

z3z1ma:
Spike as a POC


z3z1ma:
The idea here is likely to be implemented in another iteration. It needs more discussion to ensure a correct and clean implementation.

PR #1179, opened 2023-07-19
Fix: Skip directories in dbt project the user is not permitted to access

crericha:


tobymao (approved):

tobymao:
why would we do this? why wouldn’t they be able to access

crericha:
They have folders created by docker that can't be moved

PR #1185, opened 2023-07-19
Feat: Support sessions in the BigQuery adapter

izeigerman:


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -87,6 +87,18 @@ def _job_params(self) -> t.Dict[str, t.Any]:
             params["maximum_bytes_billed"] = self._extra_config.get("maximum_bytes_billed")
         return params
 
+    def _begin_session(self) -> None:
+        from google.cloud.bigquery import QueryJobConfig
+
+        job = self.client.query("SELECT 1;", job_config=QueryJobConfig(create_session=True))
```
eakmanrq, line 93: Ah so you can't actually just pass in a session ID then? I was thinking you had to get it from BQ.

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -364,3 +364,18 @@ def test_merge(mocker: MockerFixture):
         "job_config",
         "table",
     ]
+
+
+def test_begin_end_session(mocker: MockerFixture):
```
eakmanrq, line 369: Does it also make sense to ensure that a series of engine adapter calls within a single session share a common session ID on their jobs?

sqlmesh/core/engine_adapter/bigquery.py
izeigerman: yep, you were right after all.

tests/core/engine_adapter/test_bigquery.py
izeigerman: Updated the test

sqlmesh/core/engine_adapter/bigquery.py
eakmanrq: Cool making sure we still explored passing it in. That would have been nice. 😃

eakmanrq (approved):

PR #1187, opened 2023-07-19
Refactor!: serialize the source SQL code for each expression

GeorgeSittas:
The motivation behind this PR is to use the raw SQL code as the source of truth, instead of relying on sqlglot's `transpile` output, which may change (into semantically equivalent forms) over time even though the model's code is unchanged.

### DIFF
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -115,7 +115,6 @@ class ModelMeta(Node):
         ]
 
         for partition in partitions:
-            partition.meta["dialect"] = dialect
             num_cols = len(list(partition.find_all(exp.Column)))
             error_msg: t.Optional[str] = None
             if num_cols == 0:
@@ -132,20 +131,13 @@ class ModelMeta(Node):
     def _columns_validator(
         cls, v: t.Any, values: t.Dict[str, t.Any]
     ) -> t.Optional[t.Dict[str, exp.DataType]]:
-        dialect = values.get("dialect")
-        columns_to_types = {}
         if isinstance(v, exp.Schema):
-            for column in v.expressions:
-                expr = column.args["kind"]
-                expr.meta["dialect"] = dialect
-                columns_to_types[column.name] = expr
-            return columns_to_types
+            return {column.name: column.args["kind"] for column in v.expressions}
         if isinstance(v, dict):
-            for k, data_type in v.items():
-                expr = exp.DataType.build(data_type, dialect=dialect)
-                expr.meta["dialect"] = dialect
-                columns_to_types[k] = expr
-            return columns_to_types
+            return {
+                k: exp.DataType.build(data_type, dialect=values.get("dialect"))
+                for k, data_type in v.items()
+            }
         return v
 
     @validator("depends_on_", pre=True)
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -109,12 +109,17 @@ class ModelMeta(Node):
                 d.parse_one(entry, dialect=dialect) if isinstance(entry, str) else entry
                 for entry in ensure_list(v)
             ]
-        partitions = [
-            exp.to_column(expr.name) if isinstance(expr, exp.Identifier) else expr
-            for expr in partitions
-        ]
 
-        for partition in partitions:
+        validated_partitions = []
+        for expr in partitions:
+            if isinstance(expr, exp.Identifier):
+                column = exp.to_column(expr.name)
+                column.meta["sql"] = expr.meta["sql"]
+                expr = column
+
+            validated_partitions.append(expr)
+
+        for partition in validated_partitions:
             num_cols = len(list(partition.find_all(exp.Column)))
             error_msg: t.Optional[str] = None
             if num_cols == 0:
@@ -125,7 +130,7 @@ class ModelMeta(Node):
             if error_msg:
                 raise ConfigError(f"partitioned_by field '{partition}' {error_msg}")
 
-        return partitions
+        return validated_partitions
 
     @validator("columns_to_types_", pre=True)
     def _columns_validator(
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -110,16 +110,15 @@ class ModelMeta(Node):
                 for entry in ensure_list(v)
             ]
 
-        validated_partitions = []
-        for expr in partitions:
-            if isinstance(expr, exp.Identifier):
-                column = exp.to_column(expr.name)
-                column.meta["sql"] = expr.meta["sql"]
-                expr = column
+        partitions = [
+            exp.to_column(expr.name) if isinstance(expr, exp.Identifier) else expr
+            for expr in partitions
+        ]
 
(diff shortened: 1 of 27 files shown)


sqlmesh/utils/pydantic.py
```diff
@@ -10,7 +10,7 @@ class PydanticModel(BaseModel):
     class Config:
         arbitrary_types_allowed = True
         extra = "forbid"
-        json_encoders = {exp.Expression: lambda e: e.sql(dialect=e.meta.get("dialect"))}
+        json_encoders = {exp.Expression: lambda e: e.meta.get("sql") or e.sql()}
```
tobymao, line 13: why would we ever fall back to e.sql()?

GeorgeSittas: Yes, I plan to get rid of this soon -- left over due to testing

GeorgeSittas: @tobymao changed to only do `e.meta["sql"]` but it seems like it's gonna need a bit of work to fix tests & ensure we're adding the `"sql"` key to all expressions (e.g. in model meta fields, etc). If the direction looks good I should be able to complete this tomorrow.

tobymao: what tests is this a problem?

tests/core/test_snapshot.py
```diff
@@ -98,7 +99,7 @@ def test_json(snapshot: Snapshot):
             "name": "name",
             "partitioned_by": [],
             "owner": "owner",
-            "query": "SELECT @EACH(ARRAY(1, 2), x -> x), ds FROM parent.tbl",
+            "query": "SELECT @EACH([1, 2], x -> x), ds FROM parent.tbl",
```
GeorgeSittas, line 102: This test needed to be changed because we're not actually calling `parse_raw` to get back the spark version of the array AST node, but rather just doing a simple json loads which will give us back whatever was stored in the `"sql"` entry of the `meta` map (in this case, it was `@EACH([1, 2], ....)`.

sqlmesh/utils/pydantic.py
GeorgeSittas: Several are failing with a `KeyError` due to the `"sql"` lookup but in many cases it might be because we have `from sqlglot import parse, parse_one` instead of `from sqlmesh.core.dialect import parse, parse_one`, and so the sqlmesh-specific logic might not be kicking off (already fixed some tests in test_state_sync with just this change).

Take a look at Circle CI's report to get a feel if you want.

GeorgeSittas: Brought it back in after all because python and seed models don't have raw sql associated with them and so we need to fallback to `e.sql()` for them.

sqlmesh/core/model/meta.py
```diff
@@ -109,13 +109,17 @@ def _partition_by_validator(
...
             ]
-        partitions = [
-            exp.to_column(expr.name) if isinstance(expr, exp.Identifier) else expr
-            for expr in partitions
-        ]
 
-        for partition in partitions:
-            partition.meta["dialect"] = dialect
+        validated_partitions = []
+        for expr in partitions:
```
tobymao, line 114: i don't think it's as important to do it for this section or the types if that will simplify the code

sqlmesh/utils/pydantic.py
tobymao, line 13: don't you need to do sql(dialect?

sqlmesh/core/model/meta.py
GeorgeSittas: The reason I modified this section is that we can have dialect-specific quotes in identifiers (not sure if it's possible to also have other dialect-specific expressions), and so some tests failed without these changes. For example, in `test_model::test_model_normalization` we have ```partitioned_by foo(`ds`) ```, so the following assertion failed:

```python
assert model.partitioned_by[0].sql(dialect="bigquery") == "foo(`ds`)"
```

sqlmesh/utils/pydantic.py
GeorgeSittas: Assuming the `e.sql()` part is reached only for python / seed models (see https://github.com/TobikoData/sqlmesh/pull/1187#discussion_r1269807721), then this could be consistent with what we did before because those models don't have raw SQL associated with them and so `meta` would not have a `"dialect"` key set.

Now, I'm not sure, but there might be a difference w.r.t. the model meta properties.

sqlmesh/core/model/meta.py
tobymao: it shouldn't fail, i think the issue may be that you're not serializing in the correct dialect?

sqlmesh/utils/pydantic.py
tobymao: yea, this is probably what caused your previous bug with the partitions

izeigerman: Do we want to keep `dialect` as a fallback strategy? Eg. if we modify the original expression we might want to reset the `sql` attribute but keep the `dialect` one

tobymao: sorry i'm unclear what you mean

GeorgeSittas: > Do we want to keep dialect as a fallback strategy? Eg. if we modify the original expression we might want to reset the sql attribute but keep the dialect one

@izeigerman do you mean for the SQL models too, as in keep the `meta["dialect"] = ...` logic in `dialect.py`? If that's what you're suggesting, I understand that we'd instead do:

```python
lambda self, e: e.meta["sql"] or e.sql(dialect=e.meta.get("dialect"))
```

GeorgeSittas: I went ahead and implemented this: seemed like a small overhead to preserve `"dialect"` in the `meta` dict to have a safe fallback which is basically what we did before. What do you think?

sqlmesh/core/model/meta.py
```diff
@@ -109,13 +109,16 @@ def _partition_by_validator(
...
+
         partitions = [
             exp.to_column(expr.name) if isinstance(expr, exp.Identifier) else expr
             for expr in partitions
         ]
 
         for partition in partitions:
             partition.meta["dialect"] = dialect
+
+        for partition in partitions:
```
tobymao, line 121: get rid of this

tobymao (approved):

sqlmesh/core/model/meta.py
GeorgeSittas: Done

sqlmesh/utils/pydantic.py
```diff
@@ -10,7 +10,9 @@ class PydanticModel(BaseModel):
     class Config:
         arbitrary_types_allowed = True
         extra = "forbid"
-        json_encoders = {exp.Expression: lambda e: e.sql(dialect=e.meta.get("dialect"))}
+        json_encoders = {
+            exp.Expression: lambda e: e.meta.get("sql") or e.sql(e.meta.get("dialect"))
```
GeorgeSittas, line 14:
```suggestion
            exp.Expression: lambda e: e.meta.get("sql") or e.sql(dialect=e.meta.get("dialect"))
```

PR #1189, opened 2023-07-19
Fix!: Remove SQL dialect config for dbt projects

treysp:


sqlmesh/dbt/context.py
```diff
@@ -52,8 +50,8 @@ class DbtContext:
     _manifest: t.Optional[ManifestHelper] = None
 
     @property
-    def dialect(self) -> t.Optional[str]:
-        return self.model_defaults.dialect if self.model_defaults else None
+    def dialect(self) -> str:
+        return self.engine_adapter.dialect if self.engine_adapter is not None else ""
```
tobymao, line 54: we should raise here right? why would engine_adapter be none

treysp: Raising makes sense to me, but these lines are a straight reversion to previous so I'm not totally sure. @crericha

crericha: The optional nature of the adapter may not be needed any longer. @izeigerman ?

sqlmesh/cli/main.py
```diff
@@ -96,7 +96,7 @@ def cli(
 
 
 @cli.command("init")
-@click.argument("sql_dialect")
+@click.argument("sql_dialect", required=False)
```
izeigerman, line 99: Why is it not required?

sqlmesh/dbt/context.py
izeigerman: Actually I don't think dbt's context even need an engine adapter at this point. Why don't we just look at the target instead and raise if one is not available, since it should definitely be available at that point.

sqlmesh/dbt/loader.py
```diff
@@ -33,29 +32,17 @@
...
-    )
+    context = DbtContext(project_root=project_root)
     profile = Profile.load(context)
 
     return Config(
         default_gateway=profile.target_name,
         gateways={profile.target_name: GatewayConfig(connection=profile.target.to_sqlmesh(), state_connection=state_connection)},  # type: ignore
         loader=DbtLoader,
-        model_defaults=ModelDefaultsConfig(dialect=default_sql_dialect),
+        model_defaults=ModelDefaultsConfig(dialect=profile.target.type),
```
izeigerman, line 45: We gotta make sure that `model_defaults` is not in `kwargs`.

sqlmesh/cli/main.py
crericha: So that a dbt project can be inited. Later Trey does a check and raises an exception if not specified and not a dbt project.

sqlmesh/dbt/context.py
crericha: Removed engine adapter from context

PR #1191, opened 2023-07-20
Feat: make contextdiff and plans serializable

z3z1ma:
This sets the stage for some very interesting possibilities including decoupling plans and the execution environment/security context, centrally managing plans in a queue, and other stuff.

### DIFF
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -271,9 +271,3 @@ class ContextDiff(PydanticModel):
 
         new, old = self.modified_snapshots[model]
         return old.model.text_diff(new.model)
-
-    def __getstate__(self) -> t.Dict[str, t.Any]:
-        return self.dict()
-
-    def __setstate__(self, state: t.Dict[str, t.Any]) -> None:
-        self.__init__(**state)  # type: ignore
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -271,3 +271,9 @@ class ContextDiff(PydanticModel):
 
         new, old = self.modified_snapshots[model]
         return old.model.text_diff(new.model)
+
+    def __getstate__(self) -> t.Dict[str, t.Any]:
+        return self.dict()
+
+    def __setstate__(self, state: t.Dict[str, t.Any]) -> None:
+        self.__init__(**state)  # type: ignore
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -79,24 +79,24 @@ class Plan(PydanticModel):
     auto_categorization_enabled: bool = True
     include_unmodified: bool = False
 
-    _directly_modified: t.List[Snapshot] = PrivateAttr(None)
-    _indirectly_modified: SnapshotMapping = PrivateAttr(None)
+    _directly_modified: t.List[Snapshot] = None
+    _indirectly_modified: SnapshotMapping = None
 
     _plan_id: str = PrivateAttr(default_factory=random_id)
 
-    _start: t.Optional[TimeLike] = PrivateAttr(None)
-    _end: t.Optional[TimeLike] = PrivateAttr(None)
+    _start: t.Optional[TimeLike] = None
+    _end: t.Optional[TimeLike] = None
     _execution_time: t.Optional[TimeLike] = PrivateAttr(default_factory=now)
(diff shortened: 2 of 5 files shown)


tobymao:
aren't context diffs already serializable? what's the point adding dill

z3z1ma:
Goal is to serialize a `Plan`. 
The plan naturally needs the `ContextDiff`.
The `ContextDiff` is not serializable. (by pickle)
So I made it serializable. `dill` has not entered the picture at that point.

Now onto the plan. There is an `apply` function which can be passed into the plan. The most robust way to serialize it without more code is just using `dill`. Its is in your dependency tree already, so I used it out of convenience. Ofc It should be pinned if it will be used. But there are probably other approaches. This was easiest.

Given the way it is now with the change, both ContextDiffs by themselves are serializable (by pickle), and Plan is serializable.

z3z1ma:
If we drop trying to serialize the callable. We don't need dill and it actually makes way more sense. You would just need to attach a callable which can be the responsibility of whatever is deserializing

sqlmesh/core/context_diff.py
```diff
@@ -271,3 +271,9 @@ def text_diff(self, model: str) -> str:
 
         new, old = self.modified_snapshots[model]
         return old.model.text_diff(new.model)
+
+    def __getstate__(self) -> t.Dict[str, t.Any]:
```
tobymao, line 275: is this necessary?

z3z1ma: Yes. Without it you get this.

```
FAILED tests/core/test_context.py::test_diff - _pickle.PicklingError: Can't pickle <function root at 0x13866c700>: attribute lookup root on __main_...
```

sqlmesh/core/plan/definition.py
```diff
@@ -641,6 +641,18 @@ def _ensure_new_env_with_changes(self) -> None:
                 "No changes were detected. Make a change or run with --include-unmodified to create a new environment without changes."
             )
 
+    def __getstate__(self) -> t.Dict[str, t.Any]:
+        state = self.__dict__.copy()
```
tobymao, line 645: why do you need to get rid of dag?

tobymao:
you shouldn't be using pickle, you should just serialize the context_diff with pydantic's to_dict()

in general we have no guarantees on pickling, we actively avoid it and try to make things serializable using pydantic.

sqlmesh/core/plan/definition.py
```diff
@@ -66,60 +68,69 @@ class Plan:
...
+    forward_only: bool = False
+    environment_ttl: t.Optional[str] = None
+    categorizer_config: CategorizerConfig = Field(default_factory=CategorizerConfig)
+    auto_categorization_enabled: bool = True
+    include_unmodified: bool = False
+
+    _directly_modified: t.List[Snapshot] = None
+    _indirectly_modified: SnapshotMapping = None
+
+    _plan_id: str = PrivateAttr(default_factory=random_id)
```
tobymao, line 85: why do you need privateattr?

tests/core/test_context.py
```diff
@@ -193,6 +193,33 @@ def test_diff(sushi_context: Context, mocker: MockerFixture):
     assert mock_console.show_model_difference_summary.called
 
 
+def test_serialization(sushi_context: Context, mocker: MockerFixture):
```
tobymao, line 196: shouldn't this go in test_plan?

sqlmesh/core/plan/definition.py
z3z1ma: https://docs.pydantic.dev/latest/usage/models/#private-model-attributes

It needs a default factory so rather than use `Field`, `PrivateAttr` seems to be more idiomatic for pydantic here.

tests/core/test_context.py
z3z1ma: Yes, good callout

tobymao (approved):

tobymao:
can you fix the build so we can merge this @z3z1ma

z3z1ma:
@tobymao yes but now that Plan is a Pydantic model, it no longer likes your mocks and throws during validation :) 
So need to manage that

z3z1ma:
@izeigerman have you dealt with this in the past? Should I refactor tests to use ContextDiffs instead of the Mocks so it passes validation or you know a way to disable. Couldn't fish anything up that didnt look janky.

tobymao:
closing for now, i may pick this up

PR #1197, opened 2023-07-20
Feat: sync file system via sse

mykalmax:
- [x] Create File (UI)
- [x] Create File (Local File System)
- [x] Create Directory (UI)
- [x] Create Directory (Local File System)
- [x] Rename File (UI)
- [x] Rename File (Local File System)
- [x] Rename Directory (UI)
- [x] Rename Directory (Local File System)
- [x] Move Single (UI)
- [x] Move Single (Local File System)
- [x] Move Multiple (UI)
- [x] Move Multiple (Local File System)
- [x] Delete Single (UI)
- [x] Delete Single (Local File System)
- [x] Delete Multiple (UI)
- [x] Delete Multiple (Local File System)
- [x] Modify File (UI)
- [x] Modify File (Local File System)

### DIFF
--- a/web/client/src/library/pages/ide/IDE.tsx
+++ b/web/client/src/library/pages/ide/IDE.tsx
@@ -84,6 +84,7 @@ export default function PageIDE(): JSX.Element {
   const findAtrifactByPath = useStoreProject(s => s.findAtrifactByPath)
   const findParentByPath = useStoreProject(s => s.findParentByPath)
 
+  const tabs = useStoreEditor(s => s.tabs)
   const storedTabs = useStoreEditor(s => s.storedTabs)
   const storedTabId = useStoreEditor(s => s.storedTabId)
   const selectTab = useStoreEditor(s => s.selectTab)
@@ -158,6 +159,12 @@ export default function PageIDE(): JSX.Element {
           }
 
           if (artifact instanceof ModelDirectory) {
+            artifact.allFiles.forEach(file => {
+              if (tabs.has(file)) {
+                removeTab(file)
+              }
+            })
+
             artifact.parent?.removeDirectory(artifact)
           }
         }
--- a/web/client/src/library/pages/ide/IDE.tsx
+++ b/web/client/src/library/pages/ide/IDE.tsx
@@ -134,7 +134,7 @@ export default function PageIDE(): JSX.Element {
     const unsubscribeFile = subscribe<{
       changes: any
       models: any
-    }>('file', ({ changes, models }) => {
+    }>('file', (changes: any) => {
       changes.forEach(({ change, path, file, directory, type }: any) => {
         if (change === 'modified') {
           const currentFile = findAtrifactByPath(file.path) as
@@ -181,10 +181,6 @@ export default function PageIDE(): JSX.Element {
         }
       })
 
-      if (isNotNil(models)) {
-        updateModels(models)
-      }
-
       refreshFiles()
     })
 
@@ -198,9 +194,9 @@ export default function PageIDE(): JSX.Element {
       const project = new ModelDirectory(data)
       const files = project.allFiles
 
-      setProject(project)
-      setFiles(files)
       restoreEditorTabsFromSaved(files)
+      setFiles(files)
+      setProject(project)
     })
 
     return () => {
--- a/web/client/src/library/pages/ide/IDE.tsx
+++ b/web/client/src/library/pages/ide/IDE.tsx
@@ -42,6 +42,10 @@ import ModalConfirmation, {
   type Confirmation,
 } from '@components/modal/ModalConfirmation'
 import { ModelDirectory } from '@models/directory'
+import {
+  EnumFileExplorerChange,
+  type FileExplorerChange,
+} from '@components/fileExplorer/context'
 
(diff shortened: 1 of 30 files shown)


web/server/models.py
```diff
@@ -321,3 +321,16 @@ class TestResult(BaseModel):
...
     skipped: t.List[TestSkipped]
+
+
+class ArtifactType(str, enum.Enum):
+    file = "file"
+    directory = "directory"
+
+
+class ArtifactChange(BaseModel):
+    change: str
```
vchan, line 332: How about using the Change enum from watchfiles?

web/server/watcher.py
```diff
@@ -1,38 +1,109 @@
...
-                        ).to_dict()
-                    ),
-                )
+    ignore_entity_patterns.append("^\\.DS_Store$")
+    ignore_paths = [str((context.path / c.CACHE).resolve())]
+    watch_filter = DefaultFilter(
+        ignore_paths=ignore_paths, ignore_entity_patterns=ignore_entity_patterns
+    )
+
+    async for entries in awatch(context.path, watch_filter=watch_filter, force_polling=True):
```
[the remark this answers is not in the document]
mykalmax: Yes, with default I was not getting some events occasionally. I was playing with timeouts and delays but only force_polling worked consistently. Maybe  I was missing something . You can take a look if you feel like there are better way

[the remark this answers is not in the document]
vchan: No, it's fine. I was just curious. I think watchfiles automatically enable it for certain situations or falls back to it when needed.

web/client/src/models/directory.ts
```diff
@@ -157,18 +162,67 @@ export class ModelDirectory extends ModelArtifact<InitialDirectory> {
...
+    directory: ModelDirectory,
+    path: string,
+  ): ModelDirectory {
+    const directories = directory.allDirectories
+    const chain = path.split('/').reduce((acc: string[], path) => {
+      if (acc.length === 0) return [path]
+
+      const last = acc[acc.length - 1] ?? ''
+
+      if (isStringEmptyOrNil(path)) return acc
```
vchan, line 212: Nit: swap these two statements?

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -133,13 +133,17 @@ export default function PageIDE(): JSX.Element {
...
-      directories: any
+      changes: Array<{
+        change: number
+        path: string
+        file: ModelFile
+      }>
+      directories: Record<string, Directory>
     }>('file', ({ changes, directories }) => {
-      changes.sort((a: any) => (a.change === 'deleted' ? -1 : 1))
+      changes.sort((a: any) => (a.change === 3 ? -1 : 1))
```
vchan, line 143: Should we use an enum here that mirrors watchfiles' `Change`? Something like
```typescript
enum Change {
  Added = 1,
  Modified,
  Deleted,
}
```

web/server/watcher.py
```diff
@@ -1,40 +1,98 @@
...
+        context.load()
+        queue.put_nowait(
+            ServerSentEvent(
+                event="models", data=json.dumps(jsonable_encoder(get_all_models(context)))
             )
+        )
+    except Exception:
+        error = ApiException(
+            message="Error refreshing the models while watching file changes",
+            origin="API -> watcher -> _refresh_models",
```
vchan, line 96:
```suggestion
            origin="API -> watcher -> reload_context",
```

web/server/api/endpoints/files.py
```diff
@@ -72,72 +148,32 @@ def walk_path(
...
+    file_path = settings.project_path / path
+
     try:
-        with open(settings.project_path / path) as f:
+        with open(file_path) as f:
             content = f.read()
-    except FileNotFoundError:
-        raise HTTPException(status_code=HTTP_404_NOT_FOUND)
+    except FileNotFoundError as e:
+        raise e
```
vchan, line 173: Do you want to handle or log the exception? If not, no need for the try/catch block.

web/server/api/endpoints/files.py
```diff
@@ -72,72 +148,32 @@ def walk_path(
...
-        path_or_new_path = validate_path(new_path, context)
-        replace_file(settings.project_path / path, settings.project_path / path_or_new_path)
-    else:
-        (settings.project_path / path_or_new_path).write_text(content, encoding="utf-8")
-
-    content = (settings.project_path / path_or_new_path).read_text()
-    return models.File(
-        name=os.path.basename(path_or_new_path),
-        path=path_or_new_path,
+        name=os.path.basename(path),
```
vchan, line 175: I think the following should work since `path` is a Path object and not a string.
```suggestion
        name=path.name,
```

web/server/api/endpoints/files.py
```diff
@@ -55,7 +131,7 @@ def walk_path(
                             files=_files,
                         )
                     )
-                else:
+                if entry.is_file(follow_symlinks=False):
```
vchan, line 134:
```suggestion
                elif entry.is_file(follow_symlinks=False):
```

vchan (approved):

PR #1202, opened 2023-07-21
Fix: Make sure the evaluation progress bar looks good when evaluating a large number of models

izeigerman:
https://github.com/TobikoData/sqlmesh/assets/7747479/696c7c09-6d85-494a-a127-282162b8f105

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -8,12 +8,22 @@ import uuid
 
 from hyperscript import h
 from rich.console import Console as RichConsole
-from rich.progress import BarColumn, Progress, TaskID, TextColumn, TimeElapsedColumn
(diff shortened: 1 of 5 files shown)


izeigerman:
@vchan can you please confirm that the Web UI is fine with the fact that the `current_task_status` dict will be updated dynamically as snapshot evaluation starts as opposed to being preloaded with all snapshots from the start.

vchan:
Unfortunately, this breaks the Web UI since it does expect the full list of snapshots from the start. We'll have to update the UI to handle this more dynamic tasks event. cc @mykalmax

izeigerman:
@vchan no worries, I can update this code to meet the existing UI expectations.

PR #1208, opened 2023-07-24
Fix: use VARCHAR instead of TEXT for MySQL pks, indexed fields

### LINKED ISSUE #1204: MySQL cannot index text fields
MySQL cannot create an index or primary key on columns that are of type text, they need to be VARCHAR with a fixed length.

Currently the migrations don't work on MySQL because we create indices on name which is text. Perhaps we need to do some kind of if statement inside migrations to make limit the name to varchar(...).

georgesittas:
Fixes #1204

### DIFF
--- a/sqlmesh/migrations/v0008_create_intervals_table.py
+++ b/sqlmesh/migrations/v0008_create_intervals_table.py
@@ -10,16 +10,16 @@ def migrate(state_sync):  # type: ignore
     if state_sync.schema:
         intervals_table = f"{state_sync.schema}.{intervals_table}"
 
-    pk_text_type = index_text_type(engine_adapter.dialect)
+    text_type = index_text_type(engine_adapter.dialect)
 
     engine_adapter.create_state_table(
         intervals_table,
         {
-            "id": exp.DataType.build(pk_text_type),
+            "id": exp.DataType.build(text_type),
             "created_ts": exp.DataType.build("bigint"),
-            "name": exp.DataType.build(pk_text_type),
-            "identifier": exp.DataType.build(pk_text_type),
-            "version": exp.DataType.build(pk_text_type),
+            "name": exp.DataType.build(text_type),
+            "identifier": exp.DataType.build(text_type),
+            "version": exp.DataType.build(text_type),
             "start_ts": exp.DataType.build("bigint"),
             "end_ts": exp.DataType.build("bigint"),
             "is_dev": exp.DataType.build("boolean"),
--- a/sqlmesh/migrations/v0001_init.py
+++ b/sqlmesh/migrations/v0001_init.py
@@ -21,14 +21,14 @@ def migrate(state_sync):  # type: ignore
         environments_table = f"{schema}.{environments_table}"
         versions_table = f"{schema}.{versions_table}"
 
-    pk_text_type = index_text_type(engine_adapter.dialect)
+    text_type = index_text_type(engine_adapter.dialect)
 
     engine_adapter.create_state_table(
         snapshots_table,
         {
-            "name": exp.DataType.build(pk_text_type),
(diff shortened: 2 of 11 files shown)


sqlmesh/utils/migrations.py
```diff
@@ -0,0 +1,13 @@
+from sqlglot.dialects.dialect import DialectType
```
izeigerman, line 1: [Nit] Can we not call the module in plural form?

sqlmesh/utils/migrations.py
```diff
@@ -0,0 +1,13 @@
...
+
+# https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#data-types-storage-reqs-strings
+MAX_TEXT_LENGTH = 65535
```
izeigerman, line 4: This seems a little extreme and will impact the page layout in mysql. Can we put something more reasonable in here? I'm sure something like 1024 or even 512 will be more than enough.

GeorgeSittas: Fair enough!

GeorgeSittas: Changed to 1024, left a comment to make it clear that it might not match MySQL's limit.

tobymao (approved):

sqlmesh/migrations/v0008_create_intervals_table.py
```diff
@@ -1,21 +1,25 @@
...
         intervals_table,
         {
             "id": exp.DataType.build("text"),
             "created_ts": exp.DataType.build("bigint"),
-            "name": exp.DataType.build("text"),
-            "identifier": exp.DataType.build("text"),
-            "version": exp.DataType.build("text"),
+            "name": exp.DataType.build(pk_text_type),
+            "identifier": exp.DataType.build(pk_text_type),
+            "version": exp.DataType.build(pk_text_type),
```
vchan, line 22: Please ignore this if I'm mistaken, but isn't the primary key for this table `id` and so `id`'s type needs to be updated but not these columns?

tobymao: everything that has an index needs to be varchar fixed

sqlmesh/utils/migration.py
```diff
@@ -0,0 +1,14 @@
+from sqlglot.dialects.dialect import DialectType
+
+# This is an ad-hoc upper bound which doesn't necessarily match MySQL's
+# TEXT storage specification, but it should be good for most use cases.
+MAX_TEXT_LENGTH = 1024
+
+
+def primary_key_text_type(dialect: DialectType) -> str:
```
tobymao, line 8: change the name of this, it's not primary key, it's index

sqlmesh/migrations/v0008_create_intervals_table.py
vchan: Thank you for clarifying. So does `id` also need to have its type updated to varchar?

tobymao: the problem is that mysql doesn't allow unbounded chars in indices

vchan: Sorry, I didn't mean just varchar, but the varchar(1024) in the pk_text_type variable. This create table will fail in MySQL if the `id` column is type text and it's used in the primary key index, right?

tobymao: yes @GeorgeSittas you need to update "id"

GeorgeSittas: You're right @vchan, thanks for catching this!

sqlmesh/utils/migration.py
GeorgeSittas: Done

PR #1209, opened 2023-07-24
Feat: Respect the model's cron expression in the built-in scheduler

izeigerman:


sqlmesh/core/context.py
```diff
@@ -395,13 +396,16 @@ def run(
...
+            ignore_cron: Whether to ignore the model's cron schedule and run all available missing intervals.
         """
         environment = environment or c.PROD
         self.notification_target_manager.notify(
             NotificationEvent.RUN_START, environment=environment
         )
         try:
-            self.scheduler(environment=environment).run(environment, start, end, latest)
+            self.scheduler(environment=environment).run(
+                environment, start, end, latest, ignore_cron=False
```
tobymao, line 407: did you mean to hardcode ignore_cron here?

izeigerman: just fixed this, thanks

sqlmesh/core/node.py
```diff
@@ -33,6 +33,59 @@ class IntervalUnit(str, Enum):
...
+        if self == IntervalUnit.DAY:
+            return "0 0 * * *"
+        if self == IntervalUnit.MONTH:
+            return "0 0 1 * *"
+        if self == IntervalUnit.YEAR:
+            return "0 0 1 1 *"
+        return ""
+
+    def croniter(self, value: TimeLike) -> CroniterCache:
+        return CroniterCache(self._cron_expr, value)
```
tobymao, line 51: this is getting created every time you call it, maybe it's ok since it's not expensive to create and the cache still hits?

tobymao (commented):
z

sqlmesh/core/snapshot/definition.py
```diff
@@ -669,14 +685,15 @@ def missing_intervals(
...
+        missing = []
         for i in range(len(dates)):
             if dates[i] >= end_ts:
                 break
             current_ts = dates[i]
             next_ts = (
                 dates[i + 1]
                 if i + 1 < len(dates)
-                else to_timestamp(self.model.cron_next(current_ts))
+                else to_timestamp(self.model.interval_unit().cron_next(current_ts))
```
tobymao, line 696: can we store the interval unit above, it's used many times, also this is a tight loop

sqlmesh/core/snapshot/definition.py
```diff
@@ -646,7 +651,18 @@ def missing_intervals(
         )
         latest_ts = to_timestamp(latest)
 
-        croniter = self.model.croniter(start_ts)
+        upper_bound_ts = (
+            to_timestamp(self.model.cron_floor(execution_time or now()))
```
tobymao, line 655: so the upper bound is computed with the regular cron expression, and on line 662 we check it against the normalized interval unit cron?

izeigerman: That is correct. Do you see a problem here?

tobymao: no

sqlmesh/core/node.py
izeigerman: yeah, it doesn't appear that we call it in the loop. The only place where it's called in the loop it's used as a fallback and can only happen on the last iteration: https://github.com/TobikoData/sqlmesh/blob/e478491aaeb397ae3fc8881c593d12f3d243fe90/sqlmesh/core/snapshot/definition.py#L696

tobymao (approved):

PR #1212, opened 2023-07-25
Feat: Add physical schema override

eakmanrq:
Based on this PR: https://github.com/TobikoData/sqlmesh/pull/1032

Thanks @z3z1ma for the idea and most of the implementation. 🙌

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -541,12 +541,12 @@ class Context(BaseContext):
         snapshots = {}
 
         for model in models.values():
+            config = self.config_for_model(model)
             if model.name not in self._models and model.name in remote_snapshots:
                 snapshot = remote_snapshots[model.name]
                 ttl = snapshot.ttl
                 project = snapshot.project
             else:
-                config = self.config_for_model(model)
                 ttl = config.snapshot_ttl
                 project = config.project
 
@@ -557,6 +557,7 @@ class Context(BaseContext):
                 cache=fingerprint_cache,
                 ttl=ttl,
                 project=project,
+                physical_schema_override=config.physical_schema_override,
             )
             snapshots[model.name] = snapshot
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -541,12 +541,12 @@ class Context(BaseContext):
         snapshots = {}
 
         for model in models.values():
-            config = self.config_for_model(model)
             if model.name not in self._models and model.name in remote_snapshots:
                 snapshot = remote_snapshots[model.name]
                 ttl = snapshot.ttl
                 project = snapshot.project
             else:
+                config = self.config_for_model(model)
                 ttl = config.snapshot_ttl
                 project = config.project
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -557,7 +557,6 @@ class Context(BaseContext):
                 cache=fingerprint_cache,
                 ttl=ttl,
                 project=project,
-                physical_schema_override=config.physical_schema_override,
             )
             snapshots[model.name] = snapshot
 
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -24,7 +24,7 @@ from sqlmesh.core.model import (
     OptimizedQueryCache,
     SeedModel,
     create_external_model,
-    load_sql_file_model,
+    load_sql_based_model,
 )
 from sqlmesh.core.model import model as model_registry
 from sqlmesh.utils import UniqueKeyDict
@@ -258,7 +258,7 @@ class SqlMeshLoader(Loader):
                             raise ConfigError(
                                 f"Failed to parse a model definition at '{path}': {ex}."
                             )
-                    return load_sql_file_model(
+                    return load_sql_based_model(
                         expressions,
                         defaults=config.model_defaults.dict(),
                         macros=macros,
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -24,7 +24,7 @@ from sqlmesh.core.model import (
     OptimizedQueryCache,
     SeedModel,
     create_external_model,
-    load_model,
+    load_sql_file_model,
 )
 from sqlmesh.core.model import model as model_registry
 from sqlmesh.utils import UniqueKeyDict
@@ -258,7 +258,7 @@ class SqlMeshLoader(Loader):
                             raise ConfigError(
                                 f"Failed to parse a model definition at '{path}': {ex}."
                             )
-                    return load_model(
+                    return load_sql_file_model(
                         expressions,
                         defaults=config.model_defaults.dict(),
                         macros=macros,
@@ -267,6 +267,7 @@ class SqlMeshLoader(Loader):
                         module_path=context_path,
                         dialect=config.model_defaults.dialect,
                         time_column_format=config.time_column_format,
+                        physical_schema_override=config.physical_schema_override,
                     )
 
                 model = cache.get_or_load_model(path, _load)
@@ -300,6 +301,7 @@ class SqlMeshLoader(Loader):
                         module_path=context_path,
                         defaults=config.model_defaults.dict(),
                         time_column_format=config.time_column_format,
+                        physical_schema_override=config.physical_schema_override,
                     )
                     models[model.name] = model
 
@@ -379,4 +381,9 @@ class SqlMeshLoader(Loader):
                 self._loader._config_mtimes.get(self._context_path),
                 self._loader._config_mtimes.get(self._loader._context.sqlmesh_path),
             ]
-            return str(int(max(m for m in mtimes if m is not None)))
+            return "__".join(
+                [
+                    str(int(max(m for m in mtimes if m is not None))),
+                    self._loader._context.config.fingerprint,
+                ]
+            )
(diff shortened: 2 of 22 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -645,15 +649,15 @@ def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
         """
         raise NotImplementedError
 
-    @property
-    def data_hash(self) -> str:
+    def data_hash(self, additional_fields: t.Optional[t.List[str]] = None) -> str:
```
tobymao, line 652: why are you doing additional fields and not just make it first class?

eakmanrq: Model generates the fingerprint and `physical_schema` is a Snapshot concept and not a model concept. So rather than introducing an unrelated concept to Model I instead allowed Snapshot to pass in additional fields into the data hash so it could share it's context in cases like this.

tobymao: the fingerprint used to get generated in snapshot and it recently got moved to model...

tobymao: i don't like this because now you need to call data_hash with special arguments to get the right hash, i think we need to move physical_schema into the model, or make this purely in the snapshot fingerprint level, meaning it's an arg to the snapshotfingerprint

eakmanrq: @izeigerman thoughts?

z3z1ma: My original contention in the PR was that the physical schema looked like it needed to be part of the Model. In the same way that dialect can be derived from the model_defaults section. My 2 cents. 


Models should have a physical schema and virtual schema config. You can set it in the MODEL properties. At a project level, you can use model_defaults.physical_schema as either a string or map. If it is a map, then it is a map of HashMap<SourceSchema, TargetSchema> where source schema is basically the schema part of the model name and target schema is whatever you want. In this way you can leave the default behavior by doing nothing, force all snapshots into a single schema by supplying a string, or use a mapping to route models to specific schemas, possibly with a `__default__` key in the mapping.

eakmanrq: Talked to @izeigerman. Going to add it to Model but make it a root level config property since we don't want users overriding this on a per-model basis.

sqlmesh/core/loader.py
```diff
@@ -379,4 +381,9 @@ def _model_cache_entry_id(self, model_path: Path) -> str:
                 self._loader._config_mtimes.get(self._context_path),
                 self._loader._config_mtimes.get(self._loader._context.sqlmesh_path),
             ]
-            return str(int(max(m for m in mtimes if m is not None)))
+            return "__".join(
+                [
+                    str(int(max(m for m in mtimes if m is not None))),
+                    self._loader._context.config.fingerprint,
```
eakmanrq, line 387: We didn't include the actual fingerprint of the config in this which means if someone modified the config in values they passed into Context then we could reuse cache even if they changed a config that would modify the behavior of models. So this makes it based on provided config and last modified date.

eakmanrq:
I need to update tests but looking for feedback before investing the time

sqlmesh/core/context.py
```diff
@@ -541,12 +541,12 @@ def snapshots(self) -> t.Dict[str, Snapshot]:
         snapshots = {}
 
         for model in models.values():
+            config = self.config_for_model(model)
```
izeigerman, line 544: Why was this moved here?

sqlmesh/core/loader.py
izeigerman: Is this something we need in the dbt loader too? Maybe not

sqlmesh/core/context.py
eakmanrq: for an earlier version. Looks like it can be moved back now.

sqlmesh/core/model/definition.py
```diff
@@ -1324,7 +1337,7 @@ def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
 Model = t.Union[SqlModel, SeedModel, PythonModel, ExternalModel]
 
 
-def load_model(
+def load_sql_file_model(
```
izeigerman, line 1340: `load_sql_based_model` maybe? There isn't really a file anywhere in this API

sqlmesh/core/loader.py
eakmanrq: I would guess not since most would be loading from files which I'm assuming has a modified timestamp detection like what we are doing here.

sqlmesh/core/model/definition.py
```diff
@@ -1658,6 +1682,7 @@ def _create_model(
                 "jinja_macros": jinja_macros,
                 "dialect": dialect,
                 "depends_on": depends_on,
+                "physical_schema_override": physical_schema_override.get(exp.to_table(name).db),
```
izeigerman, line 1685: I'm a bit confused - do we or do we **not** have to pass dialect into something like `exp.to_table`, @tobymao ?

sqlmesh/core/snapshot/definition.py
```diff
@@ -327,10 +327,8 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
 
     Args:
         name: The snapshot name which is the same as the model name and should be unique per model.
-
         fingerprint: A unique hash of the model definition so that models can be reused across environments.
-        physical_schema: The physical schema that the snapshot is stored in.
-        model: Model object that the snapshot encapsulates.
+        node: Model object that the snapshot encapsulates.
```
izeigerman, line 331: I guess it's `Node object` now

sqlmesh/core/snapshot/definition.py
```diff
@@ -456,21 +454,18 @@ def from_model(
...
-            models: Dictionary of all models in the graph to make the fingerprint dependent on parent changes.
-                If no dictionary is passed in the fingerprint will not be dependent on a model's parents.
+            nodes: Dictionary of all nodes in the graph to make the fingerprint dependent on parent changes.
+                If no dictionary is passed in the fingerprint will not be dependent on a node's parents.
             ttl: A TTL to determine how long orphaned (snapshots that are not promoted anywhere) should live.
             version: The version that a snapshot is associated with. Usually set during the planning phase.
             audits: Available audits by name.
             cache: Cache of model name to fingerprints.
-
         Returns:
```
izeigerman, line 463: Why do you dislike empty lines so much lol?

sqlmesh/core/snapshot/definition.py
eakmanrq: Yeah I think there a lot of these mismatches in docstrings now. I'm just cleaning up the ones I notice.

sqlmesh/core/snapshot/definition.py
```diff
@@ -771,10 +766,7 @@ def version_get_or_generate(self) -> str:
     def physical_schema(self) -> str:
         if self.physical_schema_ is not None:
             return self.physical_schema_
-        schema = exp.to_table(self.name).db
-        if not schema:
-            schema = c.DEFAULT_SCHEMA
```
izeigerman, line 776: I don't see this logic anywhere now. We do need to support a default schema since the model name might just contain a table name without a schema

izeigerman: I take this back. Found this logic in the model definition.

sqlmesh/core/snapshot/definition.py
eakmanrq: Lol I originally put something there and then removed which consumed the newline 🤷 I added it back in.

sqlmesh/core/snapshot/definition.py
eakmanrq: https://github.com/TobikoData/sqlmesh/blob/c091cd2301a0c707c86d30d78e8dd0d4ba63d088/sqlmesh/core/model/definition.py#L555

sqlmesh/migrations/v0019_add_schema_to_data_hash.py
```diff
@@ -0,0 +1,5 @@
+"""Add schema to data_hash (https://github.com/TobikoData/sqlmesh/pull/1212)"""
+
+
+def migrate(state_sync):  # type: ignore
```
izeigerman, line 4: I think the next release will be breaking anyway (0.23.0). Do we still need this script then?

izeigerman: Ah, I guess we do unless we also update sqlglot version

izeigerman (approved):
Beautiful

sqlmesh/core/model/definition.py
tobymao: does the name go through the normalizer? if not then this is incorrect. you need to make sure the the model name validator runs and use that and then you can properly get the db

sqlmesh/core/model/meta.py
```diff
@@ -193,6 +198,14 @@ def _normalize_name(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
         values["name"] = d.normalize_model_name(values["name"], dialect=values.get("dialect"))
         return values
 
+    @classmethod
+    def _set_physical_schema_override(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+        physical_schema_override_map = values.get("physical_schema_override_map", {})
```
tobymao, line 203: values.get("...) OR {}

eakmanrq: Why is that preferred?

eakmanrq: Oh in case it comes back as none

PR #1217, opened 2023-07-26
Feat: Allow specifying of arbitrary table properties as part of the model definition

### LINKED ISSUE #1139: Feat: bigquery require_partition_filter, partition_expiration_date and others

izeigerman:
Closes #1139

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -249,11 +249,13 @@ class SparkEngineAdapter(EngineAdapter):
         partitioned_by: t.Optional[t.List[exp.Expression]] = None,
         partition_interval_unit: t.Optional[IntervalUnit] = None,
         clustered_by: t.Optional[t.List[str]] = None,
+        table_properties: t.Optional[t.Dict[str, t.Any]] = None,
     ) -> t.Optional[exp.Properties]:
         properties: t.List[exp.Expression] = []
 
         if storage_format:
             properties.append(exp.FileFormatProperty(this=exp.Var(this=storage_format)))
+
         if partitioned_by:
             for expr in partitioned_by:
                 if not isinstance(expr, exp.Column):
@@ -265,6 +267,10 @@ class SparkEngineAdapter(EngineAdapter):
                     this=exp.Schema(expressions=partitioned_by),
                 )
             )
+
+        for key, value in (table_properties or {}).items():
+            properties.append(exp.Property(this=key, value=exp.convert(value)))
+
         if properties:
             return exp.Properties(expressions=properties)
         return None
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -249,7 +249,7 @@ class SparkEngineAdapter(EngineAdapter):
         partitioned_by: t.Optional[t.List[exp.Expression]] = None,
         partition_interval_unit: t.Optional[IntervalUnit] = None,
         clustered_by: t.Optional[t.List[str]] = None,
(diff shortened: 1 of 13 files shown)


sqlmesh/core/model/meta.py
```diff
@@ -158,6 +162,46 @@ def _columns_validator(
...
+    ) -> t.Optional[t.Dict[str, TableProperty]]:
+        if v is None or not isinstance(v, exp.Expression):
+            return v
+
+        dialect = values.get("dialect")
+
+        if isinstance(v, (exp.Tuple, exp.Array)):
+            eq_expressions: t.List[exp.Expression] = v.expressions
+        elif isinstance(v, exp.Paren):
+            eq_expressions = [v.this]
```
tobymao, line 177: v.unnest()

sqlmesh/core/model/meta.py
```diff
@@ -158,6 +162,46 @@ def _columns_validator(
...
+            key = eq_expr.this.name
+            value_expr = eq_expr.expression
+            if isinstance(value_expr, exp.Boolean) or value_expr.is_string:
+                value = value_expr.this
+            elif value_expr.is_int:
+                value = int(value_expr.this)
+            elif value_expr.is_number:
+                value = float(value_expr.this)
+            else:
+                value = value_expr.copy()
```
tobymao, line 198: you're expecting an expression here?

do you want to use the python evaluator here to convert sql into python?

izeigerman: No, not really. I think I do want to preserve an expression. If someone does something like `TIMESTAMP('<timestamp>')` I want to pass it as is, don't I?

tobymao: then why are you converting anything at all?

izeigerman: To be consistent with scenario when the dictionary is constructed and passed directly from Python or dbt models.

tobymao: i don't think it makes a difference because you call convert anyways

tobymao: or just call convert in the validator, so when python models pass it in, it is always an expression type

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -249,11 +249,13 @@ def _create_table_properties(
         partitioned_by: t.Optional[t.List[exp.Expression]] = None,
         partition_interval_unit: t.Optional[IntervalUnit] = None,
         clustered_by: t.Optional[t.List[str]] = None,
+        table_properties: t.Optional[t.Dict[str, t.Any]] = None,
```
tobymao, line 252: change to exp.Expression

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -265,6 +267,10 @@ def _create_table_properties(
                     this=exp.Schema(expressions=partitioned_by),
                 )
             )
+
+        for key, value in (table_properties or {}).items():
+            properties.append(exp.Property(this=key, value=exp.convert(value)))
```
tobymao, line 272: remove the convert here and do it in the validator?

sqlmesh/core/engine_adapter/spark.py
izeigerman: done

sqlmesh/core/model/meta.py
izeigerman: updated

sqlmesh/core/engine_adapter/spark.py
izeigerman: udpated

tobymao (approved):

PR #1220, opened 2023-07-27
Feat!: Make Environment Suffix Target Configurable

eakmanrq:
This PR allows users to add environment suffix to either the schema name or the table name. It makes sure that when cleaning up environments/views that we drop the objects using the appropriate strategy.

### DIFF
--- a/sqlmesh/core/config/common.py
+++ b/sqlmesh/core/config/common.py
@@ -1,10 +1,29 @@
 from __future__ import annotations
 
 import typing as t
+from enum import Enum
 
 from pydantic import validator
 
+from sqlmesh.utils import classproperty
 from sqlmesh.utils.errors import ConfigError
+
+
+class EnvironmentSuffixTarget(str, Enum):
+    SCHEMA = "schema"
+    TABLE = "table"
+
+    @property
+    def is_schema(self) -> bool:
+        return self == EnvironmentSuffixTarget.SCHEMA
+
+    @property
+    def is_table(self) -> bool:
+        return self == EnvironmentSuffixTarget.TABLE
+
+    @classproperty
+    def default(cls) -> EnvironmentSuffixTarget:
+        return EnvironmentSuffixTarget.SCHEMA
 
 
 def _concurrent_tasks_validator(v: t.Any) -> int:
--- a/sqlmesh/core/environment.py
+++ b/sqlmesh/core/environment.py
@@ -3,55 +3,28 @@ from __future__ import annotations
 import json
 import typing as t
 
-from pydantic import validator
+from pydantic import Field, validator
 
+from sqlmesh.core import constants as c
+from sqlmesh.core.config import EnvironmentSuffixTarget
 from sqlmesh.core.snapshot import SnapshotId, SnapshotTableInfo
 from sqlmesh.utils import word_characters_only
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.pydantic import PydanticModel
 
 
-class Environment(PydanticModel):
-    """Represents an isolated environment.
-
-    Environments are isolated workspaces that hold pointers to physical tables.
+class EnvironmentNamingInfo(PydanticModel):
+    """
+    Information required for creating an object within an environment
 
     Args:
         name: The name of the environment.
-        snapshots: The snapshots that are part of this environment.
-        start_at: The start time of the environment.
-        end_at: The end time of the environment.
-        plan_id: The ID of the plan that last updated this environment.
-        previous_plan_id: The ID of the previous plan that updated this enviornment.
-        expiration_ts: The timestamp when this environment will expire.
-        finalized_ts: The timestamp when this environment was finalized.
-        promotion_snapshot_ids: The IDs of the snapshots that are promoted in this environment
-            (i.e. for which the views are created). If not specified, all snapshots are promoted.
+        suffix_target: Indicates whether to append the environment name to the schema or table name.
+
     """
 
-    name: str
-    snapshots: t.List[SnapshotTableInfo]
-    start_at: TimeLike
-    end_at: t.Optional[TimeLike]
-    plan_id: str
-    previous_plan_id: t.Optional[str]
-    expiration_ts: t.Optional[int]
-    finalized_ts: t.Optional[int]
-    promoted_snapshot_ids: t.Optional[t.List[SnapshotId]] = None
-
-    @validator("snapshots", pre=True)
-    @classmethod
-    def _convert_snapshots(cls, v: str | t.List[SnapshotTableInfo]) -> t.List[SnapshotTableInfo]:
-        if isinstance(v, str):
-            return [SnapshotTableInfo.parse_obj(obj) for obj in json.loads(v)]
-        return v
-
-    @validator("promoted_snapshot_ids", pre=True)
-    @classmethod
-    def _convert_snapshot_ids(cls, v: str | t.List[SnapshotId]) -> t.List[SnapshotId]:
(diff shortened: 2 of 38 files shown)


sqlmesh/core/config/common.py
```diff
@@ -1,12 +1,31 @@
...
 import typing as t
+from enum import Enum
 
 from pydantic import validator
 
+from sqlmesh.utils import classproperty
 from sqlmesh.utils.errors import ConfigError
 
 
+class EnvironmentSuffixTarget(str, Enum):
```
tobymao, line 12: this class seems unnecessary

sqlmesh/core/environment.py
```diff
@@ -3,55 +3,28 @@
...
 from sqlmesh.utils import word_characters_only
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.pydantic import PydanticModel
 
 
-class Environment(PydanticModel):
-    """Represents an isolated environment.
-
-    Environments are isolated workspaces that hold pointers to physical tables.
+class EnvironmentNamingInfo(PydanticModel):
```
tobymao, line 16: can we just make this a config option and not pass this around everywhere?

tobymao: ok, i don't really like the name, it's a bit verbose and now it's passed everywhere

how about EnvironmentName

eakmanrq: It is a config option but we need information about how a given environment is configured in order to accurately promote/demote snapshots and produce correct qualified view names.

eakmanrq: The naming is designed to match a pattern established by `SnapshotTableInfo`. It is a subset of information contained by the primary class (`Environment` in this class and `Snapshot` in example) and the name designates it's purpose (for determining naming). `EnvironmentName` is shorter but looses the connection the established naming pattern.

sqlmesh/core/config/common.py
eakmanrq: The alternative is doing a bool. So someone would pass in `use_schema_for_environment_suffix: True`. ON this one they provide `environment_suffix_target: schema` (or `table`). Not really a big difference and this has the nice property that it can be expanded in the future (maybe `catalog` later?).

sqlmesh/core/state_sync/base.py
```diff
@@ -248,7 +248,9 @@ def remove_interval(
     @abc.abstractmethod
     def promote(
         self, environment: Environment, no_gaps: bool = False
-    ) -> t.Tuple[t.List[SnapshotTableInfo], t.List[SnapshotTableInfo]]:
+    ) -> t.Tuple[
+        t.List[SnapshotTableInfo], t.Tuple[t.List[SnapshotTableInfo], EnvironmentNamingInfo]
```
tobymao, line 252: make this a triple instead of nested tuples

t.Tuple[a, b, c]

sqlmesh/core/state_sync/cache.py
```diff
@@ -177,7 +177,9 @@ def remove_interval(
 
     def promote(
         self, environment: Environment, no_gaps: bool = False
-    ) -> t.Tuple[t.List[SnapshotTableInfo], t.List[SnapshotTableInfo]]:
+    ) -> t.Tuple[
+        t.List[SnapshotTableInfo], t.Tuple[t.List[SnapshotTableInfo], EnvironmentNamingInfo]
```
tobymao, line 181: same

sqlmesh/core/state_sync/common.py
```diff
@@ -106,12 +115,21 @@ def promote(
...
             table_infos -= set(existing_environment.promoted_snapshots)
 
         self._update_environment(environment)
-        return list(table_infos), [existing_table_infos[name] for name in missing_models]
+        removed = (
+            list(existing_table_infos.values())
+            if environment_suffix_target_changed
+            else [existing_table_infos[name] for name in missing_models]
+        )
+        return list(table_infos), (removed, removed_environment_naming_info)
```
tobymao, line 132: i know you didn't change this but let's do sorted(table_infos) instead of list() so it's deterministic

sqlmesh/core/state_sync/base.py
eakmanrq: I think the nested tuple makes the relationship clearer. The `EnvironmentNamingInfo` that is returned is specific to the `removed` snapshots. By pairing them together this relationship is more clear.

sqlmesh/migrations/v0019_add_env_suffix_target.py
```diff
@@ -0,0 +1,61 @@
...
+            exp.ColumnDef(
+                this=exp.to_column("suffix_target"),
+                kind=exp.DataType.build("text"),
+            )
+        ],
+    )
+    engine_adapter.execute(alter_table_exp)
+
+    environments = []
+    columns = [
```
tobymao, line 26: why are you enumerating the columns? just do a select *

sqlmesh/migrations/v0019_add_env_suffix_target.py
```diff
@@ -0,0 +1,61 @@
...
+        "promoted_snapshot_ids",
+    ]
+    for row in engine_adapter.fetchall(
+        exp.select(*columns).from_(environments_table), quote_identifiers=True
+    ):
+        environments.append({**dict(zip(columns, row)), "suffix_target": "schema"})
+    if environments:
+        engine_adapter.delete_from(environments_table, "TRUE")
+
+        text_type = index_text_type(engine_adapter.dialect)
```
tobymao, line 44: why are you doing this? you should only do this for indices

sqlmesh/migrations/v0019_add_env_suffix_target.py
eakmanrq: I prefer always being explicit.

sqlmesh/migrations/v0019_add_env_suffix_target.py
eakmanrq: This is a copy/paste mistake. Thanks for catching.

tobymao: this whole section is unnecessary, just run update table set suffix_target = 'schema'

sqlmesh/core/state_sync/base.py
tobymao: it makes it harder to use, let's make it either a triple or make a simple data class to store the return type

sqlmesh/migrations/v0019_add_env_suffix_target.py
eakmanrq: Yeah good point will do.

sqlmesh/core/config/common.py
tobymao: i'd prefer a bool for now

sqlmesh/core/state_sync/base.py
eakmanrq: Data class works for me

sqlmesh/core/config/common.py
eakmanrq: Within the code is nearly functions the same so I don't see much of a reason there. For users if you set `use_schema_for_environment_suffix: False` then what does it do instead? It is not obvious. Setting `environment_suffix_target: table` is now very clear. So I just don't see much upside to the bool so I'm going to leave it unless someone wants to strongly advocate for bool.

tobymao: ok

tobymao (approved):
this should be labeled as a breaking change because it performs a migration

PR #1221, opened 2023-07-27
Feat: initial pass of SQLMetrics

tobymao:


sqlmesh/core/loader.py
```diff
@@ -38,6 +39,7 @@
     from sqlmesh.core.context import Context
 
 
+# maybe move these two methods to context
```
izeigerman, line 42: is this a fixme a todo or shouldn't be here?

tobymao: a todo

sqlmesh/core/metric/definition.py
```diff
@@ -0,0 +1,193 @@
...
+    metrics: UniqueKeyDict[str, Metric] = UniqueKeyDict("metrics")
+
+    for name, meta in metas.items():
+        if name not in metrics:
+            metrics[name] = meta.to_metric(metas, metrics)
+
+    return metrics
+
+
+def find_tables(expression: exp.Expression, dialect: str) -> t.Set[str]:
```
izeigerman, line 52: [Nit] Should this be prefixed with `_`? We have a function with the exact same name already and it might not be a bad idea to indicate that this one is specific to just this module.

tobymao: will do

sqlmesh/core/metric/definition.py
```diff
@@ -0,0 +1,193 @@
...
+        return cls._string_validator(v).lower()
+
+    @validator("dialect", "owner", "description", pre=True)
+    def _string_validator(cls, v: t.Any) -> t.Optional[str]:
+        if isinstance(v, exp.Expression):
+            return v.name
+        return str(v) if v is not None else None
+
+    @validator("expression", pre=True)
+    def _validate_expression(
```
izeigerman, line 102: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/common.py#L46

tobymao: i purposefully didn't use this one because it seemed more complicated but i guess i could

sqlmesh/core/metric/definition.py
```diff
@@ -0,0 +1,193 @@
...
+        """Converts a metric meta into a fully expanded and standalone metric."""
+        payload = self.dict(exclude={"expression"})
+        metric_refs = {}
+
+        agg_or_ref = False
+
+        for node, *_ in self.expression.walk():
+            if isinstance(node, exp.Alias):
+                _raise_metric_config_error(
+                    f"Alias found for metric '{self.name}' which is not allowed", self._path
```
izeigerman, line 125: When and why this might occur?

tobymao: user can write bad sql in the definition

sqlmesh/core/metric/definition.py
```diff
@@ -0,0 +1,193 @@
...
+            return d.parse_one(v, dialect=values.get("dialect"))
+        if isinstance(v, exp.Expression):
+            return v
+        return v
+
+    def to_metric(
+        self, metas: t.Dict[str, MetricMeta], metrics: UniqueKeyDict[str, Metric]
+    ) -> Metric:
+        """Converts a metric meta into a fully expanded and standalone metric."""
+        payload = self.dict(exclude={"expression"})
```
izeigerman, line 117: Looks like this is not used

sqlmesh/core/loader.py
izeigerman: so let's probably name it as such

sqlmesh/core/metric/definition.py
izeigerman: I don't think we protect against all sorts of ways a user can write bad sql in there. I'm just curious why this case was picked specifically.

tobymao: because the formulas have a / b as c and user may think this is how they should define their stuff and add an alias

sqlmesh/core/metric/definition.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if metric_refs:
+            expanded = self.expression.copy()
+            for column in expanded.find_all(exp.Column):
+                metric = metric_refs.get(column)
+
+                if metric:
+                    column.replace(metric.expanded.copy())
+        else:
+            expanded = exp.alias_(self.expression, self.name)
```
izeigerman, line 151: Why don't we alias this metric's expression in case when `metric_refs` is not empty? Doesn't this break multiple levels of nestedness? Eg. Metric A -> Metric B -> Metric C

sqlmesh/core/metric/definition.py
izeigerman: ah, I see

sqlmesh/core/metric/definition.py
tobymao: look at the test cases here, but we don't need the aliases on the derived metrics

tobymao: you'd probably get something like

sum(x) as a / sum(b as ) b as c

tobymao: they are for atomic metrics

PR #1225, opened 2023-07-27
Fix: Timeout request after 10s by default

mykalmax:


web/client/src/api/index.ts
```diff
@@ -46,27 +47,28 @@ import {
   useIDE,
   type ErrorIDE,
   EnumErrorKey,
+  type ErrorKey,
 } from '~/library/pages/ide/context'
 
 export function useApiMeta(): UseQueryResult<Meta> {
-  const { addError, removeError } = useIDE()
+  const queryKey = [`/api/meta`]
```
vchan, line 54:
```suggestion
  const queryKey = ['/api/meta']
```

web/client/src/api/index.ts
```diff
@@ -302,6 +304,75 @@ export function useMutationApiSaveFile(
...
+  const queryClient = useQueryClient()
+  const { addError, removeError } = useIDE()
+
+  let timeoutId: ReturnType<typeof setTimeout>
+
+  function timeout(delay: number = DELAY): void {
+    timeoutId = setTimeout(() => {
+      addError(errorKey, {
+        key: errorKey,
+        message: `Request timed out`,
```
vchan, line 335:
```suggestion
        message: 'Request timed out',
```

web/client/src/api/index.ts
```diff
@@ -302,6 +304,75 @@ export function useMutationApiSaveFile(
...
+        description: `Request ${queryKey.join(
+          '->',
+        )} timed out after ${delay}ms`,
+        status: 408,
```
vchan, line 339: Curious how this is used.

vchan (approved):

web/client/src/api/index.ts
mykalmax: Just pushing it to error stack and displaying in Errors, because it can indicate that there are some errors on the server

web/client/src/library/components/plan/Plan.tsx
```diff
@@ -298,42 +271,32 @@ function Plan({
...
+        {
+          type: EnumPlanActions.Dates,
+          start: data?.start,
+          end: data?.end,
+        },
+      ])
 
-        seIsPlanRan(true)
-        setPlanState(EnumPlanState.Init)
+      seIsPlanRan(true)
```
vchan, line 291: Sorry, just noticed this typo.
```suggestion
      setIsPlanRan(true)
```

web/client/src/api/index.ts
vchan: But this is the client timing out, right? It's not the actual server HTTP status code.

vchan (approved):

mykalmax:
> But this is the client timing out, right? It's not the actual server HTTP status code.

Yes, at this point we sent request but server has not responded for 10s and now we should "close" connection. There is no 408 status  from the server. 

Ideally that should hot happen , and server should always send an error or request should not taken logger than 10s. But it may happen if server is running (not 500) but not responsive 

Do you think having that status code is confusing ?

vchan:
I think it could be a little misleading if you're trying to debug the error and the UI says the error is a 408 from the API, but it's actually not.

PR #1228, opened 2023-07-28
Fix: Enable Hive dynamic partition in Spark session

izeigerman:


sqlmesh/engines/spark/db_api/spark_session.py
```diff
@@ -75,6 +75,8 @@ def cursor(self) -> SparkSessionCursor:
             # Note: Spark 3.4+ Only API
             self.spark.catalog.setCurrentCatalog(self.catalog)
         self.spark.conf.set("spark.sql.sources.partitionOverwriteMode", "dynamic")
+        self.spark.conf.set("spark.hadoop.hive.exec.dynamic.partition", True)
```
eakmanrq, line 78: Do you know that passing a bool works? I more commonly see "true" passed in when googling but not sure if that is really needed.

tobymao (approved):

sqlmesh/engines/spark/db_api/spark_session.py
izeigerman: Seems like yes:
```
    @since(2.0)
    def set(self, key: str, value: Union[str, int, bool]) -> None:
```
but I guess better be safe than sorry

PR #1230, opened 2023-07-28
Fix: Use a pendulum datetime object for the start_date attribute in Airflow

izeigerman:


tobymao (approved):

pingzh (approved):
thanks for fixing. tested 

```
In [3]: dag = db.dags['sqlmesh_snapshot_tmp.sqlmesh_full_model_2538461911_dag']

In [4]: dag.timezone
Out[4]: Timezone('+00:00')

In [5]: type(dag.timezone)
Out[5]: pendulum.tz.timezone.FixedTimezone
```

PR #1241, opened 2023-07-31
Fix: Include catalog in table name if defined in model

eakmanrq:


sqlmesh/core/snapshot/definition.py
```diff
@@ -875,7 +875,14 @@ def _ensure_categorized(self) -> None:
 
 def table_name(physical_schema: str, name: str, version: str, is_temp: bool = False) -> str:
```
tobymao, line 876: shouldn't this take a catalog ?

eakmanrq: Well the solution is to take the catalog from the name if it is defined.

eakmanrq: Are you saying as a way to override it?

tobymao: ok nvm i got it now i think

tobymao (approved):

PR #1242, opened 2023-07-31
Fix: Support inserting dataframes into Environment on Databricks

eakmanrq:
When running on latest Databricks I got an error saying it could not infer the schema for a Pandas Dataframe when inserting a new environment record. So this allows us to define the schema when we get a `columns_to_type` and the columns are all non-complex. Otherwise we still pass in None and let it infer.

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -15,6 +15,8 @@ from sqlmesh.utils import nullsafe_join
 from sqlmesh.utils.errors import SQLMeshError
 
 if t.TYPE_CHECKING:
+    from pyspark.sql import types as spark_types
+
     from sqlmesh.core._typing import TableName
     from sqlmesh.core.engine_adapter._typing import (
         DF,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -38,14 +44,53 @@ def spark(self) -> PySparkSession:
...
         return True
 
-    def _ensure_pyspark_df(self, generic_df: DF) -> PySparkDataFrame:
+    @classmethod
+    def convert_columns_to_types_to_pyspark_schema(
+        cls, columns_to_types: t.Dict[str, exp.DataType]
+    ) -> t.Optional[spark_types.StructType]:
+        from pyspark.sql import types as spark_types
+
+        mapping = {
```
tobymao, line 53: make this a constant

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -38,14 +44,53 @@ def spark(self) -> PySparkSession:
...
+            exp.DataType.Type.TEXT: spark_types.StringType,
+            exp.DataType.Type.INT: spark_types.IntegerType,
+            exp.DataType.Type.BIGINT: spark_types.LongType,
+            exp.DataType.Type.FLOAT: spark_types.FloatType,
+            exp.DataType.Type.DOUBLE: spark_types.DoubleType,
+            exp.DataType.Type.DECIMAL: spark_types.DecimalType,
+            exp.DataType.Type.DATE: spark_types.DateType,
+            exp.DataType.Type.TIMESTAMP: spark_types.TimestampType,
+        }
+        try:
```
tobymao, line 64: why are complex types not supported? just because it's harder to write the code? i don't think it's that much more complicated to support maps and arrays etc

eakmanrq: Yes just saving time. I don't have much before I'm OOO and I want to make sure SQLMesh works for Databricks users by getting this essential fixes into main first.

tobymao: ok, put it on your todo to fix this properly

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Ended up not doing this since it requires pyspark types and we don't want to require importing that so I left it where it is.

PR #1249, opened 2023-07-31
Fix: Include catalog name in schema creation

eakmanrq:


sqlmesh/core/test/definition.py
```diff
@@ -63,15 +63,17 @@ def setUp(self) -> None:
...
             df = pd.DataFrame.from_records(rows)  # noqa
             columns_to_types: dict[str, exp.DataType] = {}
             for i, v in rows[0].items():
                 # convert ruamel into python
                 v = v.real if hasattr(v, "real") else v
                 columns_to_types[i] = parse_one(type(v).__name__, into=exp.DataType)
-            self.engine_adapter.create_schema(table)
-            self.engine_adapter.create_view(_test_fixture_name(table), df, columns_to_types)
+            table = exp.to_table(table_name)
+            if table.db:
```
eakmanrq, line 74: This is actually a bugfix. Before if you passed in a name that was just a table name then we would actually create that name as a schema but then correctly create the view where you would expect.

tobymao: should we fix this in crceate_schema?

eakmanrq: Well if `create_schema` is just given a single name then it assumes this a schema name which makes sense. So I think the bug is the caller passing in a table into `create_schema` instead of a schema.

izeigerman: I agree with Ryan. The logic of automatically retrieving a schema if a table name was passed always felt kind of wrong to me. Explicit better than implicit

sqlmesh/core/engine_adapter/base.py
```diff
@@ -456,11 +456,16 @@ def create_view(
             )
         )
 
-    def create_schema(self, schema_name: str, ignore_if_exists: bool = True) -> None:
+    def create_schema(
+        self, schema_name: str, catalog_name: t.Optional[str] = None, ignore_if_exists: bool = True
+    ) -> None:
         """Create a schema from a name or qualified table name."""
+        schema_name = schema_name.split(".")[0]
```
izeigerman, line 463: Why do we continue to do the `split`? If we do, should we also anticipate 3-part names here?

eakmanrq: Nice catch!

izeigerman (approved):

PR #1251, opened 2023-08-01
Feat: Raise error if Run DAG Fails

eakmanrq:
Currently if you do run dag from magics and CLI it will not raise an exception if there is an error while applying. That means the scheduler will think that it was successful when it failed. In addition notifications would tell you it succeeded when it didn't. This change makes it so that the interfaces are aware of the outcome and can raise accordingly.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -394,7 +394,7 @@ class Context(BaseContext):
         execution_time: t.Optional[TimeLike] = None,
         skip_janitor: bool = False,
         ignore_cron: bool = False,
-    ) -> None:
+    ) -> bool:
         """Run the entire dag through the scheduler.
 
         Args:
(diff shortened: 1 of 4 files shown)


sqlmesh/core/context.py
```diff
@@ -422,11 +425,21 @@ def run(
...
-        self.notification_target_manager.notify(NotificationEvent.RUN_END, environment=environment)
+        if success:
+            self.notification_target_manager.notify(
+                NotificationEvent.RUN_END, environment=environment
+            )
+        else:
+            self.notification_target_manager.notify(
+                NotificationEvent.RUN_FAILURE, environment=environment
+            )
+            return success
```
eakmanrq, line 436: I currently have it not run janitor if the run had an error. Let me know if anyone disagrees with that.

z3z1ma:
This had messed me up so times early on that I built a workaround since I was already writing a wrapper and forgot about it...

It was totally masked when `sqlmesh run` would fail. +1 on this change and note to myself to file an issue before I forget next time...

(old screenshot of the issue, ignore the traceback its irrelevant)
image

tobymao (approved):

PR #1256, opened 2023-08-02
Chore: Warn by default if the error occurs during the schema creation

izeigerman:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -456,15 +456,22 @@ def create_view(
...
+                exp.Create(
+                    this=exp.to_identifier(schema_name.split(".")[0]),
+                    kind="SCHEMA",
+                    exists=ignore_if_exists,
+                )
             )
-        )
+        except Exception as e:
+            if not warn_on_error:
+                raise
```
GeorgeSittas, line 473:
```suggestion
                raise e
```

izeigerman: why?

izeigerman: See https://stackoverflow.com/questions/18001721/raise-with-no-argument

> When you raise without arguments, the interpreter looks for the last exception raised and handled. It then acts the same as if you used raise with the most recent exception type, value and traceback.

GeorgeSittas: TIL! Thanks for sharing -- ignore me then.

PR #1257, opened 2023-08-03
Fix: Pass sleep generator parameters directly into the Retry instance in the BigQuery adapter

izeigerman:
This seems to be a more appropriate use of the `Retry` constructor as per implementation: https://github.com/googleapis/python-api-core/blob/main/google/api_core/retry.py#L308-L324

The sleep generator is later created by the `Retry` instance itself here: https://github.com/googleapis/python-api-core/blob/main/google/api_core/retry.py#L346-L348


tobymao (approved):

eakmanrq (approved):
Thanks!

PR #1258, opened 2023-08-03
deschman/mssql

deschman:
Add full SQL Server support:

- add MsSqlEngineAdapter
- add MsSqlConnectionConfig
- add tests where different from base
- add mssql install option


deschman:
Having a hard time understanding why I would be seeing these errors from the 3.7 style & unit tests:
```
sqlmesh/core/engine_adapter/mssql.py:59: error: "Tuple[Any, ...]" has no attribute "catalog_name"  [attr-defined]
sqlmesh/core/engine_adapter/mssql.py:60: error: "Tuple[Any, ...]" has no attribute "schema_name"  [attr-defined]
sqlmesh/core/engine_adapter/mssql.py:61: error: "Tuple[Any, ...]" has no attribute "name"  [attr-defined]
sqlmesh/core/config/connection.py:717: error: Cannot find implementation or library stub for module named "pymssql"  [import]
sqlmesh/core/config/connection.py:717: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 5 errors in 2 files (checked 310 source files)
```

GeorgeSittas:
> Having a hard time understanding why I would be seeing these errors from the 3.7 style & unit tests:

I think for the connection errors you need to include the following couple of lines in the `setup.cfg` file:

```
[mypy-pymssql.*]
ignore_missing_imports = True
```

Can you try to `make style` locally and see if you can reproduce this?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -0,0 +1,62 @@
...
+        dataframe: pd.DataFrame = self.fetchdf(query)
+        return [
+            DataObject(
+                catalog=row.catalog_name,
+                schema=row.schema_name,
+                name=row.name,
+                type=DataObjectType.from_str(row.type),  # type: ignore
+            )
+            for row in dataframe.itertuples()
+        ]
```
GeorgeSittas, line 62: You might need to `# type: ignore` (see also: `base_postgres.py`) this part because mypy doesn't understand that `row` has these attributes, hence the errors you're seeing. Having said that, I'm not sure why it only complains for Python3.7.

setup.py
```diff
@@ -104,6 +104,9 @@
             "langchain",
             "openai",
         ],
+        "mssql": [
+            "pymssql<2.2.8",
```
tobymao, line 108: why is this pinned?

tobymao:
have you tested this? was this all that was needed to support mssql?

setup.py
deschman: 2.2.8 doesn't install on windows, where I personally develop. Safe to assume we want to ensure platform compatibility?

deschman:
Would you recommend any specific tests outside of the unit testing, @tobymao or others?

setup.py
tobymao: yes, although it seems to me like pymssql should work on any platform. i don't think this should be a constrant of sqlmesh as they may fix the issue. additionally, it could just be your machine, so i don't think we should pin it here, it can be a local override for folks who need it

tobymao:
> Would you recommend any specific tests outside of the unit testing, @tobymao or others?

what i mean is, have you actually run an example project using this and verified it worked on a real sql server?

tobymao:
thanks so much for the contribution @deschman, can you sign the cla?

sqlmesh/core/config/connection.py
```diff
@@ -669,6 +669,56 @@ def _connection_factory(self) -> t.Callable:
         return connect
 
 
+class MsSqlConnectionConfig(_ConnectionConfig):
```
izeigerman, line 672: [Nitpick] We capitalize acronyms throughout the codebase. So it should be `MSSQLConnectionConfig`.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -0,0 +1,62 @@
...
+
+import typing as t
+
+import pandas as pd
+
+from sqlmesh.core.engine_adapter.base import EngineAdapterWithIndexSupport
+from sqlmesh.core.engine_adapter.shared import DataObject, DataObjectType
+
+
+class MsSqlEngineAdapter(EngineAdapterWithIndexSupport):
```
izeigerman, line 14: [Nit] Ditto wrt name

izeigerman: You can test this with `sqlmesh fetchdf "<query>"`.

izeigerman: Similarly it doesn't look like MSSQL supports `CREATE OR REPLACE TABLE`, so you might also need `LogicalReplaceQueryMixin`.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+        connection_factory: a callable which produces a new Database API-compliant
+            connection on every call.
+        dialect: The dialect with which this adapter is associated.
+        multithreaded: Indicates whether this adapter will be used by more than one thread.
+    """
+
+    DIALECT: str = "tsql"
+    SUPPORTS_MATERIALIZED_VIEWS: bool = False
```
izeigerman, line 25: Isn't this value set to `False` in the base class? Why override it?

tests/core/engine_adapter/test_mssql.py
```diff
@@ -0,0 +1,546 @@
...
+            call('DROP TABLE IF EXISTS "test_table"'),
+            call('CREATE TABLE IF NOT EXISTS "test_table" ("a" int, "b" int)'),
+            call(
+                'INSERT INTO "test_table" ("a", "b") SELECT CAST("a" AS INTEGER) AS "a", CAST("b" AS INTEGER) AS "b" FROM (VALUES (1, 4), (2, 5), (3, 6)) AS "t"("a", "b")'
+            ),
+        ]
+    )
+
+
+def test_create_table_primary_key(make_mocked_engine_adapter: t.Callable):
```
izeigerman, line 535: Can we also add a test for `create_index`. As far as I know MSSQL doesn't support `CREATE INDEX IF NOT EXISTS` statement I wonder whether sqlglot handles this correctly during transpilation.

tests/core/engine_adapter/test_mssql.py
```diff
@@ -0,0 +1,546 @@
...
+        unique_key=["id", "ts"],
+    )
+    adapter.cursor.execute.assert_called_once_with(
+        'MERGE INTO "target" AS "__MERGE_TARGET__" USING (SELECT CAST("id" AS INTEGER) AS "id", CAST("ts" AS DATETIME2) AS "ts", CAST("val" AS INTEGER) AS "val" FROM (VALUES (1, 4), (2, 5), (3, 6)) AS "t"("id", "ts", "val")) AS "__MERGE_SOURCE__" ON "__MERGE_TARGET__"."id" = "__MERGE_SOURCE__"."id" AND "__MERGE_TARGET__"."ts" = "__MERGE_SOURCE__"."ts" '
+        'WHEN MATCHED THEN UPDATE SET "__MERGE_TARGET__"."id" = "__MERGE_SOURCE__"."id", "__MERGE_TARGET__"."ts" = "__MERGE_SOURCE__"."ts", "__MERGE_TARGET__"."val" = "__MERGE_SOURCE__"."val" '
+        'WHEN NOT MATCHED THEN INSERT ("id", "ts", "val") VALUES ("__MERGE_SOURCE__"."id", "__MERGE_SOURCE__"."ts", "__MERGE_SOURCE__"."val")'
+    )
+
+
+def test_replace_query_pandas(make_mocked_engine_adapter: t.Callable):
```
izeigerman, line 518: I think we also need a test for `test_replace_query` with a query rather than a pandas dataframe. I expect it to yield a statement unsupported by MSSQL.

deschman:
Thanks for the feedback! I'm going to revisit these specific areas and do a bit more testing as I did run into at least one transpiling problem. Would it be preferred if I closed this PR and opened a new one after making my changes or if I left this one open?

tests/core/engine_adapter/test_mssql.py
deschman: fwiw, it looks like this test succeeds but I do think sqlglot might not be parsing this correctly

tests/core/engine_adapter/test_mssql.py
deschman: This appears to be the same situation as above. I had noticed that each engine adapter only had tests that would expect different results than the base engine. So I copied all the tests over from the base and only kept tests in that failed and I had to change something for.

izeigerman: Yeah, the shortcoming of existing tests is that they don't test against an actual database and only test against sql string expectations. This protects code from regression but requires the developer to exercise their judgement when it comes to what is and what is not supported by the target database. 

The reason why I highlighted `replace_query` and `create_index` specifically is that I suspect that their versions inherited from the base class won't produce queries compatible with MSSQL. If it turns out not to be the case, and sqlglot handles conversion correctly, then please feel free to disregard my request for adding additional tests.

tobymao:
you can just use the pr

tobymao:
@deschman are you planning to pick this up again?

deschman:
Yes - I had some priorities shift around but I'm back on it and testing today

tobymao:
> Yes - I had some priorities shift around but I'm back on it and testing today

great let us know if you need any help or have any questions

tests/core/engine_adapter/test_mssql.py
deschman: You're correct on your hunch that sqlglot is producing queries that aren't compatible with MSSQL. I'm planning to take a look to see if I can fix that. Am I correct in understanding that this would be a tsql dialect change in the sqlglot project?

GeorgeSittas: > Am I correct in understanding that this would be a tsql dialect change in the sqlglot project?

Yep, very likely. If you need any help taking on that task feel free to reach out on slack.

deschman: Thank you! This one looked over my head so I created an issue [here](https://github.com/tobymao/sqlglot/issues/2065). I'll be interested to see how change change works in case we run into edge cases in the future!

GeorgeSittas: FYI @deschman I addressed the schema creation issue in https://github.com/tobymao/sqlglot/pull/2083, but haven't fixed the "Replacing data" issues upstream: see https://github.com/tobymao/sqlglot/issues/2065#issuecomment-1682221286.

The fixes to these issues should probably go in SQL Server's engine adapter. Based on the expected results you showed for the `INSERT INTO` statement, I think you just need to pop the `REPLACE WHERE` off of its AST and then create and prepend a `DELETE FROM` statement based on it so that it gets executed first.

Does this make sense?

deschman:
Hey @izeigerman - I'm still working through some of the resulting changes required now that sqlglot is parsing the way that we need it to. I'm hoping to have something ready for review today or tomorrow!

tests/core/engine_adapter/test_mssql.py
deschman: This is going to continue to generate a statement incompatible with MSSQL until a newer version of sqlglot can be used in the project. Aside from that, I think the code is in a pretty good spot for review.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -0,0 +1,185 @@
...
+
+        self.execute(sql)
+
+        result = self.cursor.fetchone()
+
+        return result[0] == 1 if result is not None else False
+
+    @property
+    def connection(self) -> pymssql.Connection:
+        pass
```
izeigerman, line 72: I believe this shouldn't be here.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+        database_name = table.args.get("db")
+        if database_name:
+            sql = sql.where(f"table_schema = '{database_name}'")
+
+        self.execute(sql)
+
+        result = self.cursor.fetchone()
+
+        return result[0] == 1 if result is not None else False
```
izeigerman, line 68: `if result` seems more appropriate in this case since it'd else handle an empty collection.

PR #1259, opened 2023-08-03
Feat: Allow schema override in configuration for dbt projects

izeigerman:


sqlmesh/dbt/loader.py
```diff
@@ -119,7 +122,11 @@ def _load_project(self) -> Project:
             return self._project
 
         self._project = Project.load(
-            DbtContext(project_root=self._context.path, target_name=self._context.gateway)
+            DbtContext(
+                project_root=self._context.path,
+                target_name=self._context.gateway,
+                schema_override=self._schema_override,
+            )
```
crericha, line 129: Is schema_override being used anywhere except for project passing it into ManifestHelper? If not, why not pass it directly into `Project.load()`?

izeigerman: Following the pattern of just passing the `context` everywhere. Can easily change this to what you're suggesting

crericha:
It felt to me like a load only value, so I'd prefer not to make the context
even more complicated. Fine with whatever you decide.

izeigerman:
@crericha @tobymao I've realized that my initial approach is wrong. dbt has complex rules around resolving schemas for model and just simply overriding it for all models is wrong. What I actually want is to override the target schema only and let dbt figure out the rest.

Also to make sure that the manifest cache resets I now require users to override schema in the `dbt_project.yaml` instead of `config.py`. 

Please revisit this PR.

crericha (approved):

PR #1261, opened 2023-08-03
Fix: Rendering of a model definition with table properties specified

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -1873,4 +1873,9 @@ def _single_expr_or_tuple(values: t.Sequence[exp.Expression]) -> exp.Expression
     "tags": _single_value_or_tuple,
     "grain": _single_value_or_tuple,
     "hash_raw_query": exp.convert,
+    "table_properties_": lambda value: exp.Tuple(
+        expressions=[
+            exp.EQ(this=exp.Literal.string(k), expression=v) for k, v in (value or {}).items()
```
tobymao, line 1878: exp.Literal.string(k).eq(v)

izeigerman: nice

izeigerman: fixed

tobymao (approved):

PR #1264, opened 2023-08-03
Feat: Include Jinja macros when rendering a model definition

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -756,6 +760,32 @@ def metadata_hash(self, audits: t.Dict[str, Audit]) -> str:
...
+                        f"{k} = '{v}'" if isinstance(v, str) else f"{k} = {v}"
+                        for k, v in sorted(
+                            self.jinja_macros.global_objs.items(), key=lambda x: x[0]
+                        )
+                    ]
+                )
+            )
+
+        for macro_name, macro_info in sorted(
+            self.jinja_macros.root_macros.items(), key=lambda x: x[0]
```
tobymao, line 779: you don't need key

sorted({"x": 1, "a": 2}.items())

izeigerman: I think I do:
```
In [4]: sorted([MacroInfo(definition="body_b", depends_on=[]), MacroInfo(definition="body_a", depends_on=[])])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 sorted([MacroInfo(definition="body_b", depends_on=[]), MacroInfo(definition="body_a", depends_on=[])])

TypeError: '<' not supported between instances of 'MacroInfo' and 'MacroInfo'
```

izeigerman: ah, I guess, this can never happen due to uniqueness of each key. Never mind me

PR #1272, opened 2023-08-07
Feat: table diff sample by row source

treysp:
This PR modifies the table diff CLI output by:
- Presenting row stats by row source (both tables, source only, target only)
- Calculating column comparison stats for joined rows only
- Removing "_matches" from column comparison stats labels
- Separating row samples by row source (both tables, source only, target only)
- Reordering sample DFs so join columns are in leftmost position
- Displaying only columns with >0 mismatches in the joined sample comparison
- Interleaving source/target cols for joined sample comparison ([s__col1, t__col1, s__col2, t__col2] instead of [s__col1, s__col2, t__col1, t__col2])

---- Previous CLI output ----

Screenshot 2023-08-07 at 2 35 20 PM

---- New CLI output ----

Screenshot 2023-08-07 at 2 35 01 PM

### DIFF
--- a/sqlmesh/core/table_diff.py
+++ b/sqlmesh/core/table_diff.py
@@ -261,7 +261,7 @@ class TableDiff:
                     self.adapter.fetchdf(column_stats_query)
                     .T.rename(
                         columns={0: "pct_match"},
-                        index=lambda x: x.replace("_matches", ""),
+                        index=lambda x: str(x).replace("_matches", "") if x else "",
                     )
                     .drop(index=index_cols)
                 )
--- a/sqlmesh/core/table_diff.py
+++ b/sqlmesh/core/table_diff.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import typing as t
 
 import pandas as pd
@@ -53,6 +54,7 @@ class RowDiff(PydanticModel, frozen=True):
     source: str
     target: str
     stats: t.Dict[str, float]
+    sample: pd.DataFrame
     joined_sample: pd.DataFrame
     s_sample: pd.DataFrame
     t_sample: pd.DataFrame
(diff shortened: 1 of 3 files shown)


sqlmesh/core/table_diff.py
```diff
@@ -223,15 +281,49 @@ def row_diff(self) -> RowDiff:
...
                 )
+                sample = self.adapter.fetchdf(sample_query)
+
+                joined_sample_cols = [f"s__{c}" for c in index_cols]
+                comparison_cols = [
+                    (f"s__{c}", f"t__{c}")
+                    for c in column_stats[column_stats["pct_match"] < 100].index
+                ]
+                for cols in comparison_cols:
+                    joined_sample_cols.extend([*cols])
```
tobymao, line 292: why [*cols] ?

sqlmesh/core/table_diff.py
```diff
@@ -223,15 +281,49 @@ def row_diff(self) -> RowDiff:
...
+
+                s_sample = sample[(sample["s_exists"] == 1) & (sample["rows_joined"] == 0)][
+                    [
+                        *[f"s__{c}" for c in index_cols],
+                        *[f"s__{c}" for c in column_stats.index if not c in index_cols],
+                    ]
+                ]
+
+                t_sample = sample[(sample["t_exists"] == 1) & (sample["rows_joined"] == 0)][
+                    [
```
tobymao, line 310: is the nested array correct?

sqlmesh/core/table_diff.py
treysp: I was getting `TypeError: list.extend() takes exactly one argument (2 given)` with just `*cols`

GeorgeSittas: An idea is to define `comparison_cols` as a list of lists, so then you do `extend(cols)` directly here, I think.

tobymao: why not extend(cols) then?

sqlmesh/core/table_diff.py
treysp: good catch

treysp: assuming the source of non-index cols was what you were commenting on

tobymao (approved):

PR #1278, opened 2023-08-08
Feat: add sqlmesh references and automatic joins

tobymao:


sqlmesh/core/metric/renderer.py
```diff
@@ -28,48 +38,52 @@ def agg(
...
         group_by = ensure_collection(group_by)
-        sources: t.Dict[str, t.Set[exp.AggFunc]] = {name: set() for name in self.sources}
         formulas = []
-        where = where or {}
-
-        if not isinstance(where, dict):
-            where = exp.maybe_parse(where, dialect=dialect)
-            where = {name: where for name in sources}
+        # ensure that relative order of sources stays the same if passed in
+        sources: t.Dict[str, t.Tuple[t.Set[exp.AggFunc], t.Set[str]]] = {
```
izeigerman, line 49: Is there a reasonable type alias for the value type parameter?

sqlmesh/core/metric/definition.py
```diff
@@ -181,10 +185,40 @@ def _raise_metric_config_error(msg: str, path: Path) -> None:
...
-    """Finds all the table references in a metric definition."""
-    return {
-        d.normalize_model_name(exp.table_(*reversed(column.parts[:-1])), dialect=dialect)  # type: ignore
-        for column in expression.find_all(exp.Column)
-        if column.table
-    }
+def _get_ordered_tables(expression: exp.Expression, dialect: str) -> t.List[str]:
+    """Finds all the table references in a metric definition.
+
+    Additionally ensure than the first table returned is the 'measure' or numeric value being aggregated.
```
izeigerman, line 191: Can there only be a single measure table? Should we just express this in a return value as tuple (preferably a named one / pydantic model)

sqlmesh/core/metric/renderer.py
```diff
@@ -28,48 +38,52 @@ def agg(
...
+            name: (set(), set()) for name in self.sources
+        }
 
         for metric in self.metrics:
             formulas.append(metric.formula)
             for agg, tables in metric.aggs.items():
-                # TODO: sort these tables depending on relationships
-                for table_name in tables:
-                    sources[table_name].add(agg)
+                aggs, joins = sources.setdefault(tables[0], (set(), set()))
```
izeigerman, line 56: Relevant to my previous comment: from reading the code it's not at all obvious that a table with index `0` has a special meaning of being a measure table.

sqlmesh/core/model/meta.py
```diff
@@ -37,7 +39,9 @@ class ModelMeta(Node):
     column_descriptions_: t.Optional[t.Dict[str, str]]
     audits: t.List[AuditReference] = []
     tags: t.List[str] = []
-    grain: t.List[str] = []
+    grain: t.Optional[Reference] = None
+    grains: t.List[Reference] = []
```
izeigerman, line 43: I don't think we should be doing this. I think we should only keep `grains` and then `pop` the `grain` key in the `root_validator` and set the `grains` one instead. I'm not sure I like having 2 versions in the model definition itself.

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+from sqlmesh.utils.errors import ConfigError, SQLMeshError
+from sqlmesh.utils.pydantic import PydanticModel
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core.model import Model
+
+
+class Reference(PydanticModel, frozen=True):
+    expression: exp.Expression
+    dialect: str
```
izeigerman, line 18: How is this dialect used?

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+            else:
+                raise ConfigError(
+                    f"Reference '{self.expression}' must have an inferrable name or explicit alias."
+                )
+
+            self._name = "__".join(keys)
+        return self._name
+
+
+class Graph:
```
izeigerman, line 70: I suggest coming up with a more specific name

sqlmesh/core/reference.py
tobymao: used for the parsing of the expression in the serialized form i think

sqlmesh/core/model/meta.py
```diff
@@ -238,6 +242,29 @@ def _depends_on_validator(cls, v: t.Any, values: t.Dict[str, t.Any]) -> t.Option
...
         return v
 
+    @validator("grain", pre=True)
+    def _grain_validator(cls, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[Reference]:
+        if isinstance(v, (exp.Expression, str)):
+            return Reference(expression=v, unique=True, dialect=values.get("dialect"))
+        return Reference(**v)
+
+    @validator("grains", "references", pre=True)
+    def _refs_validator(
```
izeigerman, line 252: Why do we need this whole overhead of storing `Reference` instances as part of the model itself? Seems to be an overkill to just set the `unique` flag, especially provided this information is useless in the model definition itself.

Why can't we construct `Reference` instances on a fly when we add a model to the `Graph`? This will keep the model definition clean.

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+            self.add_model(model)
+
+    def add_model(self, model: Model) -> None:
+        for ref in model.all_references:
+            self._model_refs[model.name][ref.name] = ref
+            self._ref_models[ref.name][model.name] = model.name
+
+    def find_path(
+        self, source: str, target: str, max_depth: int = 3
+    ) -> t.List[t.Tuple[str, Reference]]:
```
izeigerman, line 85: If we construct `Reference` as part of the `add_model` method then we can put the model name directly into it and not have this obscure tuple.

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+                    new_path = path + [(model_name, ref)]
+
+                    if model_name == target:
+                        return new_path
+
+                    if len(new_path) < max_depth:
+                        queue.append(new_path)
+
+        raise SQLMeshError(
+            f"Cannot find path between {source} and {target}. Make sure that references/grains are configured and that a many to many join is not occurring."
```
izeigerman, line 112: [Nit] Quotes around `source` and `target`.

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+            many = False
+
+            for model_name, ref in path:
+                visited.add(model_name)
+                many = many or not ref.unique
+
+            for ref_name in self._model_refs[path[-1][0]]:
+                for model_name in self._ref_models[ref_name].values():
+                    ref = self._model_refs[model_name][ref_name]
+                    if model_name in visited or many and not ref.unique:
```
izeigerman, line 100: Can we add a comment here. After meditating on this code line for some time I still can't tell what's going on here.

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+    @property
+    def name(self) -> str:
+        if not self._name:
+            keys = []
+
+            if isinstance(self.expression, (exp.Tuple, exp.Array)):
+                for e in self.expression.expressions:
+                    if not e.output_name:
+                        raise ConfigError(
+                            f"Reference '{e}' must have an inferrable name or explicit alias."
```
izeigerman, line 56: Do we even allow arbitrary expressions in references? Should we validate this in the expression itself and make sure that only column identifiers are passed?

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,113 @@
...
+                        )
+                    keys.append(e.output_name)
+            elif self.expression.output_name:
+                keys.append(self.expression.output_name)
+            else:
+                raise ConfigError(
+                    f"Reference '{self.expression}' must have an inferrable name or explicit alias."
```
izeigerman, line 63: Ditto

sqlmesh/core/reference.py
izeigerman: Really? Where?

web/client/src/library/components/editor/extensions/SQLMeshDialect.ts
```diff
@@ -27,7 +27,7 @@ export const SQLMeshDialect: ExtensionSQLMeshDialect = function SQLMeshDialect(
   const SQLTypes = options.types
   const SQLMeshModelDictionary = getSQLMeshModelKeywords(dialects)
   const SQLMeshKeywords =
-    'columns grain tags audit model name kind owner cron start storage_format time_column partitioned_by pre post batch_size audits dialect'
+    'columns grains references tags audit model name kind owner cron start storage_format time_column partitioned_by pre post batch_size audits dialect'
```
eakmanrq, line 30: grain is still a keyword right?

sqlmesh/core/reference.py
tobymao: yea, we allow compound ones like (a, b) as compound_ref

sqlmesh/core/reference.py
tobymao: in the validator, parse_expression

sqlmesh/core/model/meta.py
tobymao: we use references in other places like table diff, i can change it to be a customer getter

sqlmesh/core/model/meta.py
```diff
@@ -238,6 +240,40 @@ def _depends_on_validator(cls, v: t.Any, values: t.Dict[str, t.Any]) -> t.Option
 
         return v
 
+    @validator("grains", "references", pre=True)
+    def _refs_validator(cls, vs: t.Any, values: t.Dict[str, t.Any]) -> t.List[Reference]:
```
izeigerman, line 244: The type signature of a return value is wrong

sqlmesh/core/model/meta.py
```diff
@@ -320,6 +356,13 @@ def table_properties(self) -> t.Dict[str, exp.Expression]:
         """A dictionary of table properties."""
         return self.table_properties_ or {}
 
+    @property
+    def all_references(self) -> t.List[Reference]:
+        """All references including grains."""
+        return [Reference(model_name=self.name, expression=e, unique=True) for e in self.grains] + [
```
izeigerman, line 362: Thanks 👍

sqlmesh/core/reference.py
```diff
@@ -0,0 +1,97 @@
...
+
+            for ref in path:
+                visited.add(ref.model_name)
+                many = many or not ref.unique
+
+            ref_name = path[-1].name
+
+            for model_name in self._ref_models[ref_name].values():
+                for ref in self._model_refs[model_name].values():
+                    # paths cannot have loops or contain many to many refs
```
izeigerman, line 83: thanks 👍

izeigerman (approved):

PR #1281, opened 2023-08-08
Fix: Improve interval_unit validation and support the field in dbt projects

crericha:


sqlmesh/core/node.py
```diff
@@ -139,10 +154,28 @@ def _cron_validator(cls, v: t.Any) -> t.Optional[str]:
...
+    @validator("interval_unit_", pre=True)
+    def _interval_unit_validator(
+        cls, v: t.Any, values: t.Dict[str, t.Any]
+    ) -> t.Optional[IntervalUnit]:
+        if v is None:
+            return v
+
+        if not isinstance(v, IntervalUnit):
+            v = IntervalUnit(t.cast(str, str_or_exp_to_str(v)).lower())
+        cron = values.get("cron")
```
izeigerman, line 170: I think `cron` won't be there. I think `values` only contains values that haven't been validated yet and since `interval_unit_` goes after `cron`, it won't be there. I suggest to perform this validation in the `_root_validator`

izeigerman: Even if this works right now I don't think it would work with Pydantic v2.

crericha: I set a breakpoint in tests and verified cron was there, which surprised me too, so I kept it to the validator. I'm fine putting in a root validator instead.

sqlmesh/dbt/model.py
```diff
@@ -269,9 +272,11 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
                 d.parse_one(c, dialect=dialect).name for c in self.cluster_by
             ]
 
-        for field in ["cron", "forward_only", "disable_restatement"]:
+        for field in ["cron", "interval_unit", "forward_only", "disable_restatement"]:
             field_val = getattr(self, field, None) or self.meta.get(field, None)
             if field_val:
+                if field == "interval_unit":
+                    field_val = IntervalUnit(field_val.lower())
```
izeigerman, line 279: This will fail with an obscure error message if the value is invalid. Can we instead just pass it on and let sqlmesh do the validation?

sqlmesh/core/node.py
```diff
@@ -86,6 +86,20 @@ def cron_floor(self, value: TimeLike) -> TimeLike:
         """
         return self.croniter(self.cron_next(value)).get_prev()
 
+    @property
+    def seconds(self) -> int:
```
izeigerman, line 90: Can this be a dict?

crericha: Yeah, that's better. Good suggestion.

sqlmesh/core/node.py
```diff
@@ -232,23 +265,27 @@ def _inferred_interval_unit(self, sample_size: int = 10) -> IntervalUnit:
...
+    croniter = CroniterCache(cron)
+    samples = [croniter.get_next() for _ in range(sample_size)]
+    min_interval = min(b - a for a, b in zip(samples, samples[1:]))
+    if min_interval >= 31536000:
+        return IntervalUnit.YEAR
+    elif min_interval >= 2419200:
+        return IntervalUnit.MONTH
+    elif min_interval >= 86400:
+        return IntervalUnit.DAY
+    elif min_interval >= 3600:
```
izeigerman, line 288: Can we reuse seconds that we compute above? Or better yet - can we create a factory class method in the `IntervalUnit` itself?

sqlmesh/core/node.py
```diff
@@ -29,6 +29,16 @@ class IntervalUnit(str, Enum):
     HOUR = "hour"
     MINUTE = "minute"
 
+    @classmethod
+    def from_cron(klass, cron: str, sample_size: int = 10) -> IntervalUnit:
+        croniter = CroniterCache(cron)
+        samples = [croniter.get_next() for _ in range(sample_size)]
+        min_interval = min(b - a for a, b in zip(samples, samples[1:]))
+        for unit, seconds in sorted(INTERVAL_SECONDS.items(), key=lambda x: x[1], reverse=True):
```
izeigerman, line 37: beautiful

sqlmesh/core/node.py
```diff
@@ -139,10 +163,27 @@ def _cron_validator(cls, v: t.Any) -> t.Optional[str]:
...
 
+    @validator("interval_unit_", pre=True)
+    def _interval_unit_validator(cls, v: t.Any) -> t.Optional[IntervalUnit]:
+        v = str_or_exp_to_str(v)
+        if v:
+            v = v.lower()
+        return v
+
+    @root_validator
+    def _node_root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
```
izeigerman, line 178: thank you!

izeigerman (approved):
Good to go once green

PR #1284, opened 2023-08-08
Fix: Correctly recreate views for snapshots categorized as indirect non-breaking / forward-only

izeigerman:


tobymao (approved):

sqlmesh/core/scheduler.py
```diff
@@ -209,7 +209,7 @@ def run(
         else:
             environment_naming_info = environment
 
-        is_dev = environment != c.PROD
+        is_dev = environment_naming_info.name != c.PROD
```
eakmanrq, line 212: Hmm looks like this was likely caused by the introduction of environment info. Thanks for catching this.

eakmanrq (approved):

PR #1286, opened 2023-08-09
Feat: Support Table Properties on Views

eakmanrq:
Prior to this change `TBLPROPERTIES` could only be set on tables but they are also supported on views in Spark: https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-ddl-create-view.html

This makes it so that table properties can be set for either tables or views. Note that these will not be applied to "environment views" (views that point to the versioned tables/views).

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -321,12 +321,28 @@ class SparkEngineAdapter(EngineAdapter):
                 )
             )
 
-        for key, value in (table_properties or {}).items():
-            properties.append(exp.Property(this=key, value=value))
+        properties.extend(self.__table_properties_to_expressions(table_properties))
 
         if properties:
             return exp.Properties(expressions=properties)
         return None
+
+    def _create_view_properties(
+        self,
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+    ) -> t.Optional[exp.Properties]:
+        """Creates a SQLGlot table properties expression for view"""
+        if not table_properties:
+            return None
+        return exp.Properties(expressions=self.__table_properties_to_expressions(table_properties))
+
+    @classmethod
+    def __table_properties_to_expressions(
+        cls, table_properties: t.Optional[t.Dict[str, exp.Expression]] = None
+    ) -> t.List[exp.Property]:
+        if not table_properties:
+            return []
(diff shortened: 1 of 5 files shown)


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -321,12 +321,28 @@ def _create_table_properties(
...
+        self,
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+    ) -> t.Optional[exp.Properties]:
+        """Creates a SQLGlot table properties expression for view"""
+        if not table_properties:
+            return None
+        return exp.Properties(expressions=self.__table_properties_to_expressions(table_properties))
+
+    @classmethod
+    def __table_properties_to_expressions(
```
tobymao, line 340: i don't understand your naming schema, with the double underscore, can we just stick with one?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -321,12 +321,28 @@ def _create_table_properties(
...
+        cls, table_properties: t.Optional[t.Dict[str, exp.Expression]] = None
+    ) -> t.List[exp.Property]:
+        if not table_properties:
+            return []
+        return [exp.Property(this=key, value=value) for key, value in table_properties.items()]
```
GeorgeSittas, line 345: I think this will mutate `value` by changing its `parent` field, although not sure if it's a problem in this case.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: The double underscore represents a private method to an engine adapter that is not meant to be accessed by potential children and does not override anything of a parent.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Is there a more recommended way to do this?

GeorgeSittas: If we need to avoid mutations here, then you can instead do `value=value.copy()`.

To clarify, the mutation happens inside `Expression`'s constructor, due to the `_set_parent` method:

https://github.com/tobymao/sqlglot/blob/baab165f04714d443bb5faced5a1a139f8eaf81c/sqlglot/expressions.py#L98-L100

PR #1290, opened 2023-08-09
Chore: Support Pydantic 2

izeigerman:


.circleci/continue_config.yml
```diff
@@ -63,6 +63,29 @@ jobs:
           name: Run doc tests
           command: make doc-test
 
+  style_and_unit_tests_pydantic_v2:
+    docker:
+      - image: cimg/python:3.10
```
crericha, line 68: Do you feel like 3.10 is sufficient? Is there an CircleCI impact if we test on the latest and oldest supported python versions too?

izeigerman: Yeah, it'd take longer and cost more credits. I'm not sure I see the value though. Python versions and Pydantic versions are orthogonal to me.

crericha: 90% of the time when I have a CircleCI issue, its either Python 3.7 or 3.11 related. If our pydantic related code has a Python version issue that only manifests in Pydantic V2, we won't know. It'll be a gap in our confidence that the code works imho.

I was under the impression CircleCI was free since we are open source. How does it actually work?

tobymao: 3.10 is fine

sqlmesh/utils/pydantic.py
```diff
@@ -1,32 +1,125 @@
...
-        }
-        underscore_attrs_are_private = True
-        smart_union = True
+PYDANTIC_MAJOR_VERSION = int(pydantic.__version__.split(".")[0])
+
+
+if PYDANTIC_MAJOR_VERSION >= 2:
+
+    def field_validator(*args: t.Any, **kwargs: t.Any) -> t.Callable[[t.Any], t.Any]:
+        kwargs.pop("always", None)
```
crericha, line 23: Consider adding a comment why the kwarg is ignored (default behavior now or something else?)

sqlmesh/dbt/common.py
```diff
@@ -156,15 +176,15 @@ class Dependencies(PydanticModel):
...
+    macros: t.List[MacroReference] = []
+    sources: t.List[str] = []
+    refs: t.List[str] = []
 
     def union(self, other: Dependencies) -> Dependencies:
         return Dependencies(
-            macros=self.macros | other.macros,
-            sources=self.sources | other.sources,
-            refs=self.refs | other.refs,
+            macros=list(set(self.macros) | set(other.macros)),
```
tobymao, line 185: should we use the unique helper so we don't lose order

izeigerman: It was all sets before, so I don't think this matters.

sqlmesh/utils/pydantic.py
```diff
@@ -1,32 +1,125 @@
...
-            exp.Expression: lambda e: e.meta.get("sql") or e.sql(dialect=e.meta.get("dialect"))
-        }
-        underscore_attrs_are_private = True
-        smart_union = True
+PYDANTIC_MAJOR_VERSION = int(pydantic.__version__.split(".")[0])
+
+
+if PYDANTIC_MAJOR_VERSION >= 2:
+
+    def field_validator(*args: t.Any, **kwargs: t.Any) -> t.Callable[[t.Any], t.Any]:
```
tobymao, line 22: this is where the magic happens i guess

izeigerman: Just one of the places.

sqlmesh/dbt/basemodel.py
```diff
@@ -55,6 +57,8 @@ class Hook(DbtConfig):
     sql: SqlStr
     transaction: bool = True  # TODO not yet supported
 
+    _sql_validator = sql_str_validator
```
crericha, line 60: Is this used? I didn't see it called in the diff.

izeigerman: This is a pydantic validator. What do you mean by "called"? Pydantic calls this

sqlmesh/dbt/common.py
crericha: Is this a serialization workaround?

sqlmesh/dbt/basemodel.py
izeigerman: For context: Pydantic V2 no longer automatically coerces regular strings into `SqlStr`, so a validator is required.

crericha (approved):

PR #1292, opened 2023-08-09
Feat: Add model to promote

eakmanrq:


sqlmesh/schedulers/airflow/plan.py
```diff
@@ -82,7 +82,9 @@ def create_plan_dag_spec(
         environment_naming_info=request.environment.naming_info,
         new_snapshots=request.new_snapshots,
         backfill_intervals_per_snapshot=backfill_intervals_per_snapshot,
-        promoted_snapshots=request.environment.snapshots,
+        promoted_snapshots=[
+            x for x in all_snapshots.values() if x.table_info in request.environment.snapshots
```
izeigerman, line 86: Better iterate through `request.environment.snapshots` and then access the name in `all_snapshots` with O(1). Right now this expression is O(n^2(

izeigerman: Btw, storing all these snapshots in the spec seems rather excessive. Can't we get them in the `dag_generator` instead while creating a promotion task?

eakmanrq: Thanks great points. Check out latest version and let me know what you think.

izeigerman: Looks much better, thanks! Is is this still a draft? I think this is is good to go otherwise.

eakmanrq: I was hoping to verify it works at the client before merging.

PR #1294, opened 2023-08-10
Fix!: Exclude redundant keys from the config attribute in dbt models

izeigerman:
Amount of data in the `config` attribute of the `sushi.customer_revenue_by_day` model before the fix: 10KB
After the fix: 1.5KB

### DIFF
--- a/sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
+++ b/sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
@@ -1,0 +1,79 @@
+"""Remove redundant attributes from dbt models."""
+import json
+
+import pandas as pd
+from sqlglot import exp
+
+from sqlmesh.utils.migration import index_text_type
+
+
+def migrate(state_sync):  # type: ignore
+    engine_adapter = state_sync.engine_adapter
+    schema = state_sync.schema
+    snapshots_table = "_snapshots"
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    new_snapshots = []
+
+    for name, identifier, version, snapshot, kind_name in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name").from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        jinja_macros_global_objs = parsed_snapshot["node"]["jinja_macros"]["global_objs"]
+        if "config" in jinja_macros_global_objs and isinstance(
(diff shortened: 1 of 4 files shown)


sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
```diff
@@ -0,0 +1,79 @@
...
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    new_snapshots = []
+
+    for name, identifier, version, snapshot, kind_name in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name").from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        jinja_macros_global_objs = parsed_snapshot["node"]["jinja_macros"]["global_objs"]
```
tobymao, line 24: does this dict always exist?

sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
```diff
@@ -0,0 +1,79 @@
...
+        if "config" in jinja_macros_global_objs and isinstance(
+            jinja_macros_global_objs["config"], dict
+        ):
+            for key in CONFIG_ATTRIBUTE_KEYS_TO_REMOVE:
+                jinja_macros_global_objs["config"].pop(key, None)
```
tobymao, line 29: is this migration necessary? won't it get cleaned up eventually on it's own

sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
izeigerman: It does

crericha (approved):
Nice improvement!

sqlmesh/migrations/v0020_remove_redundant_attributes_from_dbt_models.py
izeigerman: yeah, but what's the point of keeping this dead baggage in the database? Plus I don't think we want to trigger a plan diff

tobymao: would it trigger one? doesn't migrate rows handle it?

izeigerman: Migrate rows reads from the same state, as opposed to a local state, so I don't see how it would help here.

tobymao (approved):

PR #1295, opened 2023-08-10
Fix: resolve self references in rendering closes #1215

### LINKED ISSUE #1215: Make `@this_model` available in models
Currently `@this_model` is only available in audit definitions, but it would be useful in models as well

tobymao:


tobymao:
not necessary

PR #1296, opened 2023-08-10
Fix: respect plan include-unmodified option when no model changes

### LINKED ISSUE #1277: sqlmesh plan `--include-unmodified` flag ignored if no changes
The sqlmesh plan `--include-unmodified` flag creates views for all models in the new environment.

Currently, it is ignored if plan doesn't detect any changes.

treysp:
Closes #1277

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -469,6 +469,15 @@ class TerminalConsole(Console):
             self._prompt_backfill(plan, auto_apply)
         elif plan.context_diff.has_changes and not auto_apply:
             self._prompt_promote(plan)
+        elif (
+            plan.is_dev
+            and not plan.context_diff.is_new_environment
+            and plan.include_unmodified
+            and plan.context_diff.unpromoted_models
+            and not auto_apply
+        ):
+            self.log_status_update("\n[bold]Virtually updating unmodified models\n")
+            self._prompt_promote(plan)
 
     def _prompt_categorize(self, plan: Plan, auto_apply: bool) -> None:
         """Get the user's change category for the directly modified models."""
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -45,6 +45,8 @@ class ContextDiff(PydanticModel):
     """New models."""
     removed: t.Set[str]
     """Deleted models."""
+    existing: t.Set[str]
+    """Existing"""
     modified_snapshots: t.Dict[str, t.Tuple[Snapshot, Snapshot]]
     """Modified snapshots."""
     snapshots: t.Dict[str, Snapshot]
@@ -173,6 +175,7 @@ class ContextDiff(PydanticModel):
             create_from=create_from,
             added=added,
             removed=removed,
+            existing=existing_models,
             modified_snapshots=modified_snapshots,
             snapshots=merged_snapshots,
             new_snapshots=new_snapshots,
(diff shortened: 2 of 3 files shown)


sqlmesh/core/console.py
```diff
@@ -469,6 +469,15 @@ def _show_options_after_categorization(self, plan: Plan, auto_apply: bool) -> No
             self._prompt_backfill(plan, auto_apply)
         elif plan.context_diff.has_changes and not auto_apply:
             self._prompt_promote(plan)
+        elif (
+            plan.is_dev
```
tobymao, line 473: i think we can dry this up, it can be a property on the plan that includes this logic except for auto_aply

sqlmesh/core/context_diff.py
```diff
@@ -204,6 +207,13 @@ def promotable_models(self) -> t.Set[str]:
             *self.modified_snapshots,
         } - self.removed
 
+    @property
+    def unpromoted_models(self) -> t.Set[str]:
+        """The set of model names that have have not yet been promoted in the target environment."""
+        return {
+            *self.existing,
```
tobymao, line 214: why are you doing this? { *self.existing } this is redundant, just do

return self.existing - self.previsouly_promoted_model_names

sqlmesh/core/plan/definition.py
```diff
@@ -440,6 +440,16 @@ def _missing_intervals(self) -> t.Dict[t.Tuple[str, str], Intervals]:
...
         return self.__missing_intervals
 
+    @property
+    def has_unmodified_unpromoted(self) -> bool:
+        """Is the plan for an existing dev environment, has the include unmodified flag, and contains unmodified models that have not been promoted."""
+        return (
+            self.is_dev
+            and not self.context_diff.is_new_environment
+            and self.include_unmodified
+            and len(self.context_diff.unpromoted_models) > 0
```
tobymao, line 450: and self.context_diff.unpromoted_models

treysp: mypy was giving a type error when it was passed like that: 

```sqlmesh/core/plan/definition.py:447: error: Incompatible return value type (got "Union[Literal[False], Set[str]]", expected "bool")  [return-value]```

should i change the return type, type ignore, or something else?

Comment (tobymao) [reply sqlmesh/core/plan/definition.py:450]: bool()

File: sqlmesh/core/context_diff.py [hunk 4]
```diff
@@ -45,6 +45,8 @@ class ContextDiff(PydanticModel):
     """New models."""
     removed: t.Set[str]
     """Deleted models."""
+    existing: t.Set[str]
+    """Existing"""
```
izeigerman, line 49: [Nit] Existing what exactly? :)

sqlmesh/core/context_diff.py
```diff
@@ -204,6 +207,11 @@ def promotable_models(self) -> t.Set[str]:
             *self.modified_snapshots,
         } - self.removed
 
+    @property
+    def unpromoted_models(self) -> t.Set[str]:
+        """The set of model names that have have not yet been promoted in the target environment."""
+        return self.existing - self.previously_promoted_model_names
```
izeigerman, line 213: I don't think we need `existing`. The combination of `added` and `modified_snapshots` should give us all we need as far as I can tell.

izeigerman: Actually, no, isn't it just `snapshots`?

treysp: Good call

izeigerman (approved):

PR #1303, opened 2023-08-11
Feat!: Make all incremental models in a dbt project forward-only by default

izeigerman:
After conversation with @crericha, we concluded that making all incremental models in a dbt project forward-only by default is a safer approach since it's better aligned with dbt users' expectations.  Provided that incremental models in a typical dbt project appear to be either append-only, missing upstream history, or are non-idempotent in some other way, this update should help prevent unexpected or incorrect behavior.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -72,7 +72,7 @@ class ModelConfig(BaseModelConfig):
     interval_unit: t.Optional[str] = None
     batch_size: t.Optional[int] = None
     lookback: t.Optional[int] = None
-    forward_only: t.Optional[bool] = None
+    forward_only: bool = True
     disable_restatement: t.Optional[bool] = None
 
     # DBT configuration fields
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/model.py
```diff
@@ -203,7 +199,8 @@ def model_kind(self, target: TargetConfig) -> ModelKind:
                 IncrementalUnmanagedKind
             )
             return IncrementalUnmanagedKind(
-                insert_overwrite=strategy in INCREMENTAL_BY_TIME_STRATEGIES
+                insert_overwrite=strategy in INCREMENTAL_BY_TIME_STRATEGIES,
+                forward_only=self.forward_only,
```
crericha, line 203: `forward_only` is passed into the model. Why does it also need to be passed into`IncrementalUnmanagedKind`, but not the other incremental kinds?

crericha, line 203: Should you use `foward_only=incremental_kwargs["forward_only"]` in case it is defined in the `meta`?

crericha:
Update the docs and gtg from my perspective.

crericha (approved):

tobymao (approved):

PR #1304, opened 2023-08-12
Fix: Correctly check schema names when cleaning up

eakmanrq:
Prior to this change we would check if the name started with a physical schema name but that assumes there is no catalog on the name. This makes it so we parse the name and then check if the database matches the physical schema.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -449,13 +449,14 @@ class SnapshotEvaluator:
     def _cleanup_snapshot(self, snapshot: SnapshotInfoLike) -> None:
         snapshot = snapshot.table_info
         table_names = [snapshot.table_name()]
-        if snapshot.version != snapshot.fingerprint:
+        if snapshot.version != snapshot.fingerprint.to_version():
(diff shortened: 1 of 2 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -449,13 +449,14 @@ def _demote_snapshot(
     def _cleanup_snapshot(self, snapshot: SnapshotInfoLike) -> None:
         snapshot = snapshot.table_info
         table_names = [snapshot.table_name()]
-        if snapshot.version != snapshot.fingerprint:
+        if snapshot.version != snapshot.fingerprint.to_version():
```
eakmanrq, line 452: @izeigerman I think this has been bugged for a while. The bug would result in this always being true and therefore always adding the table name. Wondering if I am missing something and if we really expect it to behave the way I just fixed it to behave.

tobymao: it's because fingerprint didn't used to be an object probably

eakmanrq: Yeah this code does look old so that makes sense

PR #1305, opened 2023-08-13
Fix: Display column description  + display model references in docs

mykalmax:


web/client/src/library/components/documentation/Documentation.tsx
```diff
@@ -235,39 +243,59 @@ function DetailsItem({
...
-              : 'False'
-            : isDate
-            ? toDateFormat(maybeDate)
-            : value}
-        </p>
-      </div>
+      {isArrayNotEmpty(value) ? (
+        <>
+          <strong className="mr-2 text-xs capitalize">{name}</strong>
+          {(value as TValue[]).map((item: any) => (
```
vchan, line 259: Should item by a TValue?

web/client/src/library/components/documentation/Documentation.tsx
```diff
@@ -277,3 +305,17 @@ Documentation.NotFound = NotFound
...
+  const isDate = isString(value) && !isNaN(maybeDate.getTime())
+  const isBoolean = typeof value === 'boolean'
+
+  if (isBoolean && isTrue(value)) return 'True'
+  if (isBoolean && isFalse(value)) return 'False'
+  if (isDate) return toDateFormat(maybeDate)
+  if (isNumber(value)) return value.toString()
+  if (isString(value)) return value
+
+  return ''
```
vchan, line 320: Why not return a string representation of whatever value is instead of an empty string?

web/client/src/library/components/documentation/Documentation.tsx
```diff
@@ -226,7 +234,7 @@ function Section({
   )
 }
 
-function DetailsItem({
+function DetailsItem<TValue = any>({
```
vchan, line 237: Is TValue a mapping of strings to primitives?

web/client/src/utils/index.ts
```diff
@@ -10,18 +10,26 @@ export function isFalseOrNil(value: unknown): boolean {
...
+export function isString(value: unknown): value is string {
   return typeof value === 'string'
 }
 
+export function isNumber(value: unknown): value is number {
+  return typeof value === 'number'
+}
+
+export function isPrimitive(value: unknown): value is Primitive {
+  return isString(value) || isNumber(value) || typeof value === 'boolean'
```
vchan, line 22: Do you need to do anything to handle symbols?

web/client/src/library/components/documentation/Documentation.tsx
vchan: Also, what if the value is a symbol? Will it reach this line?

web/client/src/utils/index.ts
mykalmax: we don't have a use case for Symbol so i'll just remove it from type

web/client/src/library/components/documentation/Documentation.tsx
mykalmax: if payload for `model` details ever change in the future and be an `object` string will look like `[object Object]`

vchan: In that case, should value not be a primitive? Regardless, I think empty string can be misleading because users might think the value is empty when it actually isn't. Maybe something like "<unknown>" is better?

mykalmax: Good point , so in this case I think having `[object Object]` better since it is clearly shows something wrong with representing details and can be caught and fix faster , because `""` can also be interpreted as intentional empty value

vchan: @mykalmax I was suggesting `"unknown"` above but I think github markup displayed it as "".

mykalmax: so with `unknown` it is still possible to see it as intentional value like `References: Unknown` would make sense, but `References: [object Object]` pretty clear is a bug

PR #1307, opened 2023-08-14
Docs: update configuration reference

treysp:


docs/concepts/models/python_models.md
```diff
@@ -56,14 +56,14 @@ df = context.fetchdf(f"SELECT * FROM {table}")
 
 The `table` method will automatically add the referenced model to the Python model's dependencies.
 
-The only other way to set dependencies of models in Python models is to define them explicitly in the `@model` decorator using the keyword `dependencies`. The dependencies defined in the model decorator take precedence over any dynamic references inside the function.
+The only other way to set dependencies of models in Python models is to define them explicitly in the `@model` decorator using the keyword `depends_on`. The dependencies defined in the model decorator take precedence over any dynamic references inside the function.
```
izeigerman, line 59: > The dependencies defined in the model decorator take precedence over any dynamic references inside the function.

this is not accurate at the moment. At some point

docs/reference/configuration.md
```diff
@@ -149,112 +262,73 @@ gateways:
...
+            type: cloud_composer
+            airflow_url: <airflow_url>
 ```
-This scheduler type shares the same configuration options as the `airflow` type, except for `username` and `password`.
-Cloud Composer relies on `gcloud` authentication, so the `username` and `password` options are not required.
 
-See [Airflow Integration Guide](../integrations/airflow.md) for information on how to integrate Airflow with SQLMesh.
+## Gateway/connection defaults
+
+The default gateway and connection keys specify what gateway/connection should be used if a command does not explicitly specify one or some gateways in the `gateways` dictionaries do not specify every connection.
```
izeigerman, line 280: > if a command does not explicitly

it's not exactly clear what's "command" here

treysp: I'm finding this language tricky - see what you think about the updated version.

PR #1309, opened 2023-08-14
Fix: Format model on save

mykalmax:


web/server/api/endpoints/files.py
```diff
@@ -65,12 +64,21 @@ async def write_file(
     else:
         (settings.project_path / path_or_new_path).write_text(content, encoding="utf-8")
 
+    found_models = [
+        model
+        for model in context.models.values()
+        if model._path == settings.project_path / path_or_new_path
+    ]
+    if found_models:
+        context.format_model(model=found_models[0])
```
vchan, line 73: What about formatting `content` instead of the file? Also, do we want to format when renaming a file? Users might not expect that.

web/server/api/endpoints/files.py
```diff
@@ -65,12 +64,21 @@ async def write_file(
...
+
+    path_or_new_path_mapping = await get_path_mapping(settings=get_settings())
```
vchan, line 75: Settings wouldn't have changed, right?
```suggestion
    path_or_new_path_mapping = await get_path_mapping(settings=settings)
```

web/server/api/endpoints/files.py
```diff
@@ -55,22 +56,25 @@ async def write_file(
...
-    path_mapping: t.Dict[Path, models.FileType] = Depends(get_path_mapping),
 ) -> models.File:
     """Create, update, or rename a file."""
     path_or_new_path = path
     if new_path:
         path_or_new_path = validate_path(new_path, context)
         replace_file(settings.project_path / path, settings.project_path / path_or_new_path)
     else:
+        dialect = context.config_for_path(Path(path_or_new_path)).dialect
+        expressions = parse(content, default_dialect=dialect)
```
vchan, line 67: What happens if the file isn't a sql file?

web/server/api/endpoints/files.py
```diff
@@ -55,22 +57,47 @@ async def write_file(
...
+        full_path = settings.project_path / path
+        is_sql = pathlib.Path(path_or_new_path).suffix == ".sql"
+        model_sql = next(
+            (
+                model
+                for model in context._models.values()
+                if model.is_sql and model._path == full_path
+            ),
+            None,
+        )
```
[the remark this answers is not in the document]
mykalmax: we're using `get_path_mapping` is not just for `models` but for files in general

web/server/api/endpoints/models.py
```diff
@@ -24,14 +23,6 @@ def get_models(
...
 ) -> t.List[models.Model]:
     """Get a mapping of model names to model metadata"""
-    try:
-        context.load()
-    except Exception:
-        raise ApiException(
-            message="Unable to get models",
-            origin="API -> models -> get_models",
-        )
-
```
vchan, line 34: Why is this removed? Is it because the file watcher reloads the context on file changes now?

web/server/api/endpoints/files.py
```diff
@@ -55,22 +63,38 @@ async def write_file(
...
+        dialect = model.dialect if model and model.is_sql else default_dialect
+        if is_sql:
+            try:
+                expressions = parse(content, default_dialect=default_dialect)
+                content = format_model_expressions(expressions, dialect)
+            except Exception:
+                ApiException(
+                    message="Unable to format SQL file",
+                    origin="API -> files -> write_file",
+                )
```
vchan, line 87: This can be removed or replaced with a SSE message warning the file couldn't be formatted. What do you think?

web/server/api/endpoints/files.py
```diff
@@ -55,22 +63,38 @@ async def write_file(
...
     else:
-        (settings.project_path / path_or_new_path).write_text(content, encoding="utf-8")
-
+        full_path = settings.project_path / path
+        is_sql = pathlib.Path(path_or_new_path).suffix == ".sql"
+        path_to_model_mapping = await get_path_to_model_mapping(settings=settings)
+        model = path_to_model_mapping.get(Path(path_or_new_path))
+        default_dialect = context.config_for_path(Path(path_or_new_path)).dialect
+        dialect = model.dialect if model and model.is_sql else default_dialect
+        if is_sql:
```
vchan, line 79: Lines 75-78 can all go into this if block.

vchan (approved):

web/server/api/endpoints/models.py
mykalmax: yes, file watcher reloads the context on  `model`  or `seed` file changes and collecting models to send over SSE

PR #1310, opened 2023-08-14
Fix: Table properties serde

eakmanrq:
Existing behavior results in extra quotes being added to table property values each time `exp.convert` is run on the value. This makes it so after each serialization/deserialization we get the same result.


sqlmesh/core/model/meta.py
```diff
@@ -225,8 +225,9 @@ def _properties_validator(
                 table_properties[eq_expr.this.name] = value_expr
         elif isinstance(v, dict):
             for key, value in v.items():
-                value_expr = exp.convert(value, copy=True)
-                value_expr.meta["dialect"] = dialect
+                value_expr = d.parse_one(str(value), dialect=dialect)
```
tobymao, line 228: not sure this is correct, what are the inputs you're expecting here?

sqlmesh/dbt/model.py
```diff
@@ -294,9 +294,13 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
             table_properties = {}
             if self.partition_expiration_days is not None:
-                table_properties["partition_expiration_days"] = self.partition_expiration_days
+                table_properties["partition_expiration_days"] = exp.convert(
+                    self.partition_expiration_days
+                )
             if self.require_partition_filter is not None:
-                table_properties["require_partition_filter"] = self.require_partition_filter
+                table_properties["require_partition_filter"] = exp.convert(
+                    self.require_partition_filter
```
tobymao, line 302: what is require_partition_filter? what is expected here?

sqlmesh/core/model/decorator.py
```diff
@@ -38,6 +38,10 @@ def __init__(self, name: str, **kwargs: t.Any) -> None:
             )
             for call in calls
         ]
+        if "table_properties" in self.kwargs:
```
tobymao, line 41: are you sure this is needed? isn't this what was done before?

izeigerman: What was done before?

sqlmesh/dbt/model.py
izeigerman: This is a boolean value

izeigerman (approved):

sqlmesh/core/model/meta.py
eakmanrq: Updated

sqlmesh/core/model/decorator.py
tobymao: convert was called in the pydantic validator

sqlmesh/dbt/model.py
tobymao: you shouldn't need to call convert here because it was called in the validator

sqlmesh/core/model/meta.py
```diff
@@ -49,9 +49,10 @@ class ModelMeta(Node, extra="allow"):
     references: t.List[exp.Expression] = []
     hash_raw_query: bool = False
     physical_schema_override: t.Optional[str] = None
-    table_properties_: t.Optional[t.Dict[str, exp.Expression]] = Field(
+    table_properties_: t.Optional[t.Union[exp.Tuple, exp.Array]] = Field(
```
tobymao, line 52: let's just normalize everything into a Tuple

sqlmesh/core/model/meta.py
```diff
@@ -198,40 +199,41 @@ def _columns_validator(
...
+            else:
+                properties = v
         elif isinstance(v, dict):
-            for key, value in v.items():
-                value_expr = exp.convert(value, copy=True)
-                value_expr.meta["dialect"] = dialect
-                table_properties[key] = value_expr
+            properties = exp.Tuple(
+                expressions=[
+                    exp.EQ(
```
tobymao, line 226: exp.Literal.string(key).eq(exp.convert(..)

sqlmesh/core/model/meta.py
tobymao: like black, we're being opinionated

tobymao (approved):

sqlmesh/core/model/meta.py
```diff
@@ -196,42 +195,38 @@ def _columns_validator(
...
+                else v
+            )
         elif isinstance(v, dict):
-            for key, value in v.items():
-                value_expr = exp.convert(value, copy=True)
-                value_expr.meta["dialect"] = dialect
-                table_properties[key] = value_expr
+            properties = exp.Tuple(
+                expressions=[
+                    exp.Literal.string(key).eq(exp.convert(value, copy=True))
```
tobymao, line 222: the convert is not necessary since eq does that automatically

PR #1312, opened 2023-08-15
Fix: styling updates for table and table-diff

mykalmax:


web/client/src/library/components/editor/EditorInspector.tsx
```diff
@@ -485,18 +485,22 @@ function FormDiffModel({
     on,
     where,
   })
+
+  const getTableDiff = useCallback(
+    function getTableDiff(): void {
```
vchan, line 490: Is there a difference if you use an arrow function here instead? Asking to learn what's best practice.

web/client/src/library/components/tableDiff/TableDiff.tsx
```diff
@@ -115,138 +135,182 @@ export default function TableDiff({ diff }: { diff: any }): JSX.Element {
...
-          </small>
+      <div className="flex justify-between items-center px-2 mt-2">
+        <Footer count={rows.all.length} />
+        <div className="flex text-xs">
+          <div className="flex ml-2 items-center">
+            <span className="inline-block w-3 h-3 bg-brand-500 mr-2 rounded-full"></span>
+            <small className="text-neutral-600 dark:text-neutral-400">
+              Grain
+            </small>
+          </div>
```
vchan, line 295: Is it worth it to DRY this up?

vchan (approved):

web/client/src/library/components/editor/EditorInspector.tsx
mykalmax: No, we can use `arrow function` here instead. I don't know if there really a best practice regarding `arrow function`. I prefer using them as callbacks (like in this snippet i need to change it to arrow to. be consistent :)). The main difference for  me is that regular function will be hoisted where arrow function needs to be defined before usage.

PR #1317, opened 2023-08-15
Feat: allow python models to return SQL expressions

### LINKED ISSUE #999: Allow Python models to return SQLGlot expressions
Presently Python models can only return DataFrames (either Pandas or PySpark ones). We need to extend the return value to also support SQLGlot expressions.

This will allow users to define models in Python while simultaneously gaining all the benefits provided by SQL models, such as column-level lineage, automatic change categorization, etc.

GeorgeSittas:
The motivation behind this PR is to allow users of SQLMesh to define python models that return SQL instead of DataFrames, either in the form of SQLGlot expressions or just plain strings:

```python
from sqlglot import exp
from sqlmesh.core.model import model

@model(...)
def entrypoint(...) -> str | exp.Expression:
    ...
```

Defining a model like this is beneficial in cases where its logic is too complex to express in SQL, e.g. because it requires a lot of templating due to runtime dependencies. These models will be loaded as `SqlModel`s whose query is essentially a single macro function invocation under the hood, so one can think of them as being virtually defined as:

```sql
MODEL (
  ...
);

@entrypoint();  -- This will be expanded into the query of interest at render time
```

Implementing them in this way allows users to gain all the benefits provided by SQL models, such as column-level lineage, automatic change categorization, etc.

Fixes #999


tobymao (approved):

sqlmesh/utils/metaprogramming.py
```diff
@@ -276,6 +276,14 @@ def walk(obj: t.Any) -> None:
...
 
     if name not in env:
+        registry_name = getattr(obj, "registry_name", None)
+        if registry_name:
+            from sqlmesh.core.macros import macro
+
+            if registry_name == macro.registry_name:
+                # We only need to add the undecorated code of @macro() functions in env
+                obj = obj.__wrapped__
+
```
GeorgeSittas, line 286: I believe this block is going to be reached when a macro references another macro in its local scope, like how `incremental_by_ds` references `between` in the updated `macros.py`, for example. Based on this, I assumed that the `import` here shouldn't cost _too much_, but ideally I'd like to factor it out so that we don't do it over and over again (it wasn't possible to move outside due to circular `import` issues).

sqlmesh/utils/metaprogramming.py
```diff
@@ -276,6 +276,14 @@ def walk(obj: t.Any) -> None:
                 build_env(v, env=env, name=k, path=path)
 
     if name not in env:
+        registry_name = getattr(obj, "registry_name", None)
```
tobymao, line 279: can you instead just check that it's a sqlmesh macro?

GeorgeSittas: You mean by creating a different attribute like `is_sqlmesh_macro` or something in the decorator's `__call__`?

tobymao: i mean actually checking the name of the macro and seeing that it's macro and that it's path is in sqlmesh?

PR #1322, opened 2023-08-16
Fix: Remove intervals instead missing intervals

eakmanrq:
In main we only remove the intervals that are explicitly passed in for a snapshot. In order to have this work with depends_on_past where the intervals can expand beyond that range we have missing_intervals check if intervals beyond that range should be considered missing.

This change flips that so we now remove intervals beyond the range defined if needed and then missing intervals is now more simple again and just returns what is missing for the given range.

This PR also addresses an issue (thanks for catching @crericha) where if a depends_on_past model removes intervals beyond the range then we need to make sure all downstream snapshots also remove intervals beyond that range.

### DIFF
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -301,12 +301,6 @@ class Plan:
         return self._restatements
 
     @property
-    def restatement_snapshots(self) -> t.List[t.Tuple[Snapshot, Interval]]:
-        return [
-            (self.context_diff.snapshots[s], interval) for s, interval in self.restatements.items()
-        ]
-
-    @property
     def loaded_snapshot_intervals(self) -> t.List[LoadedSnapshotIntervals]:
         loaded_snapshots = []
         for snapshot in self.directly_modified:
@@ -460,7 +454,7 @@ class Plan:
     def _add_restatements(self, restate_models: t.Iterable[str]) -> None:
         def is_restateable_snapshot(snapshot: Snapshot) -> bool:
             if not self.is_dev and snapshot.model.disable_restatement:
-                logger.warning("Restatement is disabled for model '%s'.", snapshot.name)
+                logger.debug("Restatement is disabled for model '%s'.", snapshot.name)
                 return False
             return not snapshot.is_symbolic and not snapshot.is_seed
 
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -482,11 +482,7 @@ class Plan:
             interval = snapshot.get_removal_interval(
                 self.start, self.end, self._execution_time, strict=False
             )
-            upstream_snapshots = [
-                s
-                for s in self._dag.upstream(snapshot_name)
-                if is_restateable_snapshot(snapshot_mapping[s])
-            ]
+            upstream_snapshots = [s for s in self._dag.upstream(snapshot_name)]
             possible_intervals = [
                 restatements.get(s, dummy_interval) for s in upstream_snapshots
             ] + [interval]
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import sys
 import typing as t
 from collections import defaultdict
 from enum import Enum
@@ -19,7 +20,7 @@ from sqlmesh.core.snapshot import (
     merge_intervals,
     missing_intervals,
 )
-from sqlmesh.core.snapshot.definition import format_intervals
+from sqlmesh.core.snapshot.definition import Interval, format_intervals
 from sqlmesh.utils import random_id
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.date import (
@@ -110,7 +111,7 @@ class Plan:
             self._dag.add(name, snapshot.model.depends_on)
 
         self.__missing_intervals: t.Optional[t.Dict[t.Tuple[str, str], Intervals]] = None
-        self._restatements: t.Set[str] = set()
+        self._restatements: t.Dict[str, Interval] = {}
 
         if restate_models and context_diff.new_snapshots:
             raise PlanError(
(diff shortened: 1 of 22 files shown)


sqlmesh/core/plan/definition.py
```diff
@@ -451,30 +452,48 @@ def has_unmodified_unpromoted(self) -> bool:
...
+            if snapshot_name not in restatements:
+                continue
+            snapshot = snapshot_mapping[snapshot_name]
+            interval = snapshot.get_removal_interval(
+                self.start, self.end, self._execution_time, strict=False
+            )
+            upstream_snapshots = [
+                s
+                for s in self._dag.upstream(snapshot_name)
+                if is_restateable_snapshot(snapshot_mapping[s])
```
izeigerman, line 488: Does it matter here? Didn't we already make sure that all snapshots are restateable?

sqlmesh/core/snapshot/definition.py
```diff
@@ -110,11 +110,17 @@ def snapshot_id(self) -> SnapshotId:
...
+    def __lt__(self, other: SnapshotId) -> bool:
+        return self.name < other.name
+
 
 class SnapshotNameVersion(PydanticModel, frozen=True):
     name: str
     version: str
 
+    def __str__(self) -> str:
+        return "__".join([self.name, self.version])
```
izeigerman, line 122: Why is this needed?

sqlmesh/core/plan/definition.py
eakmanrq: No at this point we have made sure that all restated snapshots are restateable and their downstream dependencies. Here we take a snapshot and then go upstream so we need to do this check again because this could include new snapshots that we haven't visited yet.

sqlmesh/core/snapshot/definition.py
eakmanrq: Example of usage: https://github.com/TobikoData/sqlmesh/blob/7c689c680cf29541a1a546414a52a78db8de850c/sqlmesh/core/state_sync/engine_adapter.py#L505

sqlmesh/core/plan/definition.py
eakmanrq: I guess technically if it isn't restateable then it won't be in the restateable dict and therefore we can pull them in here and we will just get a miss on the following line when we check the dict.

eakmanrq: So I do think it can be removed.

sqlmesh/core/snapshot/definition.py
tobymao: isn't it already hashable?

izeigerman: Yeah, I'd expect `frozen=True` to ensure that

izeigerman: Even then, pydantic provides a default `__str__` implementation already:
```
In [3]: str(SnapshotNameVersion(name="test_name", version="test_version"))
Out[3]: "name='test_name' version='test_version'"
```

eakmanrq: Ah great points. Will fix.

sqlmesh/core/plan/definition.py
```diff
@@ -451,30 +452,44 @@ def has_unmodified_unpromoted(self) -> bool:
...
+                    f"Cannot restate from '{snapshot_name}'. Either such model doesn't exist, no other materialized model references it, or restatement was disabled for this model."
                 )
-            self._restatements.update(downstream)
+            restatements[snapshot_name] = dummy_interval
+            for downstream_snapshot_name in self._dag.downstream(snapshot_name):
+                if is_restateable_snapshot(snapshot_mapping[downstream_snapshot_name]):
+                    restatements[downstream_snapshot_name] = dummy_interval
+        # Get restatement intervals for all restated snapshots and make sure that if a snapshot expands it's
+        # restatement range that it's downstream dependencies all expand their restatement ranges as well.
+        for snapshot_name in self._dag:
```
tobymao, line 478: can this be just done in one for loop? why do you loop twice? because you already traverse the downstream

sqlmesh/core/snapshot/definition.py
```diff
@@ -548,39 +559,43 @@ def add_interval(self, start: TimeLike, end: TimeLike, is_dev: bool = False) ->
...
             execution_time: The date/time time reference to use for execution time. Defaults to now.
         """
-        interval = self.inclusive_exclusive(start, end, execution_time, for_removal=True)
         self.intervals = remove_interval(self.intervals, *interval)
         self.dev_intervals = remove_interval(self.dev_intervals, *interval)
 
-    def inclusive_exclusive(
+    def get_removal_interval(
         self,
         start: TimeLike,
```
tobymao, line 575: why is get_removal_interval start, end and remove_interval no longer start and end?

sqlmesh/utils/dag.py
```diff
@@ -99,7 +99,9 @@ def sorted(self) -> t.List[T]:
                 for deps in unprocessed_nodes.values():
                     deps -= next_nodes
 
-                self._sorted.extend(next_nodes)
+                # Sort to make the order deterministic
+                # TODO: Make protocol that makes the type var both hashable and sortable once we are on Python 3.8+
+                self._sorted.extend(sorted(next_nodes))  # type: ignore
```
tobymao, line 104: why was this necessary?

eakmanrq: Well I had a bug in my logic at one point and it would only show up sometimes and I found that this was a source for that inconsistency. So it is necessary in the sense that it makes the behavior of the system more consistent.

sqlmesh/core/plan/definition.py
eakmanrq: I believe the challenge with doing that is that the first loop is on `restate_models` which can be arbitrarily ordered. The algorithm relies on having nodes push the largest interval range down to their children and visiting the graph in an arbitrary order makes that more confusing. I was looking at alternative approaches but it became harder to reason about the implementation.

I did realize though that I don't need to check the full upstream but rather just the immediate parents of a given node. Also I skip traversing graph at all if there are no restatements. https://github.com/TobikoData/sqlmesh/pull/1322/commits/8776e6427a44510594df54586ba0569f1a46f1ce

sqlmesh/core/snapshot/definition.py
eakmanrq: I don't really understand. `get_removal_interval` is getting you an interval object (`t.Tuple[int, int]`) while remove_interval is removing an interval object (`t.Tuple[int, int]`). In order to first make an interval you have a start/end and that needs to be translated into an inclusive/exclusive epoch.

izeigerman (approved):

PR #1324, opened 2023-08-17
Fix: Add ignored snapshots for invalid start date

eakmanrq:
Currently you can set a start date for a model that depends_on_past that is non-contiguous. This doesn't make sense because the model depends on past and therefore these intervals would be invalid.

This change adds a concept of ignored snapshots which are snapshots that will not be included in the plan since their inclusion would result in invalid results.

### DIFF
--- a/examples/sushi/models/items.py
+++ b/examples/sushi/models/items.py
@@ -50,7 +50,7 @@ ITEMS = [
 @model(
     "sushi.items",
     kind=IncrementalByTimeRangeKind(time_column="ds", batch_size=30),
-    start="3 months ago",
+    start="1 week ago",
     cron="@daily",
     columns={
         "id": "int",
--- a/examples/sushi/models/waiter_as_customer_by_day.sql
+++ b/examples/sushi/models/waiter_as_customer_by_day.sql
@@ -6,6 +6,7 @@ MODEL (
   ),
   owner jen,
   cron '@daily',
+  start '1 week ago',
   audits (
     not_null(columns = [waiter_id]),
     forall(criteria = [LENGTH(waiter_name) > 0])
--- a/examples/sushi/models/waiter_as_customer_by_day.sql
+++ b/examples/sushi/models/waiter_as_customer_by_day.sql
@@ -6,7 +6,6 @@ MODEL (
   ),
   owner jen,
   cron '@daily',
-  start '1 week ago',
   audits (
     not_null(columns = [waiter_id]),
     forall(criteria = [LENGTH(waiter_name) > 0])
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -24,7 +24,7 @@ from rich.table import Table
 from rich.tree import Tree
 
 from sqlmesh.core.environment import EnvironmentNamingInfo
-from sqlmesh.core.snapshot import Snapshot, SnapshotChangeCategory
+from sqlmesh.core.snapshot import Snapshot, SnapshotChangeCategory, start_date
 from sqlmesh.core.test import ModelTest
 from sqlmesh.utils import rich as srich
 from sqlmesh.utils.date import to_date, yesterday_ds
@@ -111,7 +111,10 @@ class Console(abc.ABC):
 
     @abc.abstractmethod
     def show_model_difference_summary(
-        self, context_diff: ContextDiff, detailed: bool = False
+        self,
+        context_diff: ContextDiff,
+        detailed: bool = False,
+        ignored_snapshot_names: t.Optional[t.Set[str]] = None,
     ) -> None:
         """Displays a summary of differences for the given models."""
 
@@ -386,14 +389,19 @@ class TerminalConsole(Console):
                 self.log_success("The migration has been completed successfully")
 
     def show_model_difference_summary(
-        self, context_diff: ContextDiff, detailed: bool = False
+        self,
+        context_diff: ContextDiff,
+        detailed: bool = False,
+        ignored_snapshot_names: t.Optional[t.Set[str]] = None,
     ) -> None:
         """Shows a summary of the differences.
 
         Args:
             context_diff: The context diff to use to print the summary
             detailed: Show the actual SQL differences if True.
+            ignored_snapshot_names: A set of snapshot names that are ignored
         """
+        ignored_snapshot_names = ignored_snapshot_names or set()
         if context_diff.is_new_environment:
             self._print(
                 Tree(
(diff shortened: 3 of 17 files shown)


examples/sushi/models/items.py
```diff
@@ -50,7 +50,7 @@
 @model(
     "sushi.items",
     kind=IncrementalByTimeRangeKind(time_column="ds", batch_size=30),
-    start="3 months ago",
+    start="1 week ago",
```
tobymao, line 53: why 1 week? can we only do 1 week for the historical models?

sqlmesh/core/plan/definition.py
```diff
@@ -255,12 +225,21 @@ def missing_intervals(self) -> t.List[SnapshotIntervals]:
     @property
     def snapshots(self) -> t.List[Snapshot]:
         """Gets all the snapshots in the plan/environment."""
-        return list(self.context_diff.snapshots.values())
+        return self._snapshots
+
+    @property
+    def snapshot_mapping(self) -> t.Dict[str, Snapshot]:
```
izeigerman, line 231: [Nit] Can we make this "private" (underscored)? I think `snapshot_mapping` is not a commonly adopter terminology around the code base so I think it makes more sense to keep it as an implementation detail of the plan definition.

izeigerman: Btw, why even make it a property method? Can't we just do:
```
self._snapshot_mapping = self.context_diff.snapshots
```
in the constructor? What's the point of re-creating the dict all the time?

sqlmesh/core/plan/definition.py
```diff
@@ -532,7 +535,9 @@ def _categorize_snapshots(self) -> None:
         returns a list of added and directly modified snapshots as well as the mapping of
         indirectly modified snapshots.
         """
-        for model_name, snapshot in self.context_diff.snapshots.items():
+        for model_name, snapshot in self.snapshot_mapping.items():
+            if snapshot.intervals:
```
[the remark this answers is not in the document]
izeigerman: Btw, the subsequent checks already take care of this:
```
if model_name in self.context_diff.modified_snapshots:
    ...
elif model_name in self.context_diff.added and self.is_new_snapshot(snapshot):
    ...
```
Why have it? I might only suggest to move `upstream_model_names = self._dag.upstream(model_name)` under the 1st branch.

sqlmesh/core/plan/definition.py
eakmanrq: The snapshots in the plan do not equal the snapshots in context diff. Context diff is all snapshots and plan is a subset of that once it takes out snapshots that are `ignored`.

examples/sushi/models/items.py
eakmanrq: 1 week most closely matches what we were doing in the tests. We often set the start to 1 week on prod plans which doesn't really make sense because prod should never be a subset of data.

sqlmesh/core/snapshot/definition.py
```diff
@@ -794,6 +795,43 @@ def version_get_or_generate(self) -> str:
...
+        """Checks if the given start and end are valid for this snapshot.
+        Args:
+            start: The start date/time of the interval (inclusive)
+            snapshot_start: The start date/time of the snapshot (inclusive)
+        """
+        # The snapshot may not have a start defined. If so we use the provided snapshot start.
+        if self.depends_on_past and start:
+            if not snapshot_start:
+                raise SQLMeshError("Snapshot must have a start defined if it depends on past")
+            start_ts = to_timestamp(self.model.cron_floor(start))
```
izeigerman, line 813: I think this should be `self.model.interval_unit.cron_floor(start)`. `self.model.cron_floor(start)` is based on model's cron which is not always representative of the actual interval unit.

sqlmesh/core/plan/definition.py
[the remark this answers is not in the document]
eakmanrq: > Why is this check needed here? Wouldn't it make more sense to check whether the category is set? Otherwise the check looks rather arbitrary in this context.

Yeah I agree this check can be better explained. I'm good with checking the category but I actually just matched what we do here: https://github.com/TobikoData/sqlmesh/blob/b4dbb4587edaf7eb78103479c449d915542f0a51/sqlmesh/core/snapshot/definition.py#L737

sqlmesh/core/snapshot/definition.py
```diff
@@ -794,6 +795,43 @@ def version_get_or_generate(self) -> str:
...
+            # a valid start.
+            missing_intervals = self.missing_intervals(
+                snapshot_start, now(), execution_time=execution_time
+            )
+            earliest_interval = missing_intervals[0][0] if missing_intervals else None
+            if earliest_interval:
+                return earliest_interval >= start_ts
+        return True
+
+    def get_latest(self, snapshot_start: t.Optional[TimeLike] = None) -> t.Optional[TimeLike]:
```
izeigerman, line 827: So `snapshot_start` is more like `default`? Should we call it as such?

sqlmesh/core/plan/definition.py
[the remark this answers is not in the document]
eakmanrq: > Why have it? 

If I don't have this and we set plan start then I hit that assert line linked above and error out. So I just did this just upfront as you can see here.

sqlmesh/core/snapshot/definition.py
eakmanrq: Nice catch!

sqlmesh/core/snapshot/definition.py
```diff
@@ -794,6 +795,43 @@ def version_get_or_generate(self) -> str:
...
+        """The latest interval loaded for the snapshot. Snapshot start is used if intervals are not defined"""
+        return (
+            to_end_date((to_timestamp(max(x[1] for x in self.intervals)), self.model.interval_unit))
```
izeigerman, line 830: Why are we not picking `self.intervals[-1][1]`? Aren't we expecting intervals to always be merged and sorted?

sqlmesh/utils/date.py
```diff
@@ -277,3 +278,14 @@ def time_like_to_str(time_like: TimeLike) -> str:
     if is_date(time_like):
         return to_ds(time_like)
     return to_datetime(time_like).isoformat()
+
+
+def to_end_date(
+    end_and_units: t.Union[t.Tuple[int, IntervalUnit], t.List[t.Tuple[int, IntervalUnit]]]
```
izeigerman, line 284: Shouldn't the 2nd Union item be `t.Tuple[t.List[int], IntervalUnit]`? Are we expecting timestamps with different units here? Why even support this version? It seems like the only place we use it is in `get_latest`. Do we even need this helper function? Seems very specific to just that one scenario.

sqlmesh/core/snapshot/definition.py
eakmanrq: yep I agree will do

sqlmesh/utils/date.py
eakmanrq: I agree. This is fairly old code that I forgot to check again after recent refactor.

sqlmesh/core/console.py
```diff
@@ -565,8 +590,15 @@ def _prompt_backfill(self, plan: Plan, auto_apply: bool) -> None:
...
                 if end:
                     plan.end = end
-
-        if not auto_apply and self._confirm(f"Apply - {backfill_or_preview.capitalize()} Tables"):
+        if plan.ignored_snapshot_names:
+            self._print(
+                self._get_ignored_tree(plan.ignored_snapshot_names, plan.context_diff.snapshots)
+            )
+        if (
+            plan.has_changes
```
izeigerman, line 598: Why did you have to add this? Don't you also need `plan.has_unmodified_unpromoted`?

sqlmesh/core/plan/definition.py
```diff
@@ -308,16 +287,34 @@ def loaded_snapshot_intervals(self) -> t.List[LoadedSnapshotIntervals]:
...
+        return (
+            self.context_diff.is_new_environment
+            or self.context_diff.is_unfinalized_environment
+            or bool(modified_names)
+        )
+
     def is_new_snapshot(self, snapshot: Snapshot) -> bool:
         """Returns True if the given snapshot is a new snapshot in this plan."""
-        return snapshot.snapshot_id in self.context_diff.new_snapshots
+        return snapshot.snapshot_id in [s.snapshot_id for s in self.new_snapshots]
```
izeigerman, line 317: I think we call `is_new_snapshot` quite a bit, including inside other loops iterating over snapshots. This may potentially be `O(N^2)`. Perhaps worth caching and using set

izeigerman (approved):
Noticed a few more things, LGTM otherwise.

sqlmesh/core/console.py
eakmanrq: You are right that this is not needed. Thanks for catching.

examples/sushi/models/waiter_as_customer_by_day.sql
```diff
@@ -6,6 +6,7 @@ MODEL (
   ),
   owner jen,
   cron '@daily',
+  start '1 week ago',
```
tobymao, line 9: let's not add 1 week everywhere, i purposefully didn't add all of them to make it so we test inference of start

eakmanrq: Well adding plan start is not proper since prod environments cannot represent a subset of data

sqlmesh/core/plan/definition.py
```diff
@@ -308,16 +287,34 @@ def loaded_snapshot_intervals(self) -> t.List[LoadedSnapshotIntervals]:
...
                 # We don't want to display indirect non-breaking since to users these are effectively no-op changes
                 if downstream_snapshot.is_indirect_non_breaking:
                     continue
                 loaded_snapshots.append(LoadedSnapshotIntervals.from_snapshot(downstream_snapshot))
         return loaded_snapshots
 
+    @property
+    def has_changes(self) -> bool:
+        modified_names = (
+            set().union(
```
tobymao, line 300: {*self.context_diff.added, *removed...}

sqlmesh/core/plan/definition.py
```diff
@@ -679,6 +681,59 @@ def _ensure_new_env_with_changes(self) -> None:
...
+        self._dag, self._snapshots, self.ignored_snapshot_names = self._build_snapshots_and_dag()
+
+        if self._restate_models and self.new_snapshots:
+            raise PlanError(
+                "Model changes and restatements can't be a part of the same plan. "
+                "Revert or apply changes before proceeding with restatements."
+            )
+
+        self._add_restatements()
+        self._add_restatements()
```
tobymao, line 695: is this supposed to be called twice?

eakmanrq: No typo thanks for catching

examples/sushi/models/waiter_as_customer_by_day.sql
tobymao: how is that relevant? 

the start date of a model should inherit from its parents

eakmanrq: Oh I misunderstood what you said. So make sure all root models have it only.

tobymao (approved):

PR #1326, opened 2023-08-17
Fix: Fetch column types if the type is unknown for a partition column in the BQ engine adapter

izeigerman:


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -391,7 +391,9 @@ def insert_overwrite_by_partition(
             )
 
         with self.session(), self.temp_table(query_or_df, name=table_name) as temp_table_name:
-            if columns_to_types is None:
+            if columns_to_types is None or columns_to_types[
+                partition_column.name
+            ] == exp.DataType.build("unknown"):
```
GeorgeSittas, line 396: I think we can also do `columns_to_types[...].is_type("unknown")` here

eakmanrq (approved):

GeorgeSittas (approved):

PR #1329, opened 2023-08-21
Fix: only show sqlmesh run progress bar if snapshots to process

treysp:


sqlmesh/core/scheduler.py
```diff
@@ -220,49 +220,53 @@ def run(
             restatements=restatements,
             ignore_cron=ignore_cron,
         )
-        dag = self._dag(batches)
+        if batches:
```
eakmanrq, line 223: I would suggest doing something like this instead:
```
if not batches:
   self.console.log_success("No models scheduled to run at this time.")
   return True
```
Then the rest remains unindented.

PR #1331, opened 2023-08-21
Chore: Add dev publish command

eakmanrq:
Dev releases prior to this would not include the latest web code. This now ensures that each build has latest web code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,3 +82,11 @@ ui-up:
 
 ui-down:
 	docker-compose down
+
+ui-build:
+	npm --prefix web/client ci && npm --prefix web/client run build
+
+clean-build:
+	rm -rf build/ && rm -rf dist/ && rm -rf *.egg-info
+
+dev-publish: clean-build ui-build publish
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ ui-down:
 	docker-compose down
 
 ui-build:
-	npm --prefix web/client ci && npm --prefix web/client run build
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -82,3 +82,11 @@ ui-up:
 
 ui-down:
 	docker-compose down
+
+ui-build:
+	npm --prefix web/client ci && npm --prefix web/client run build
```
mykalmax, line 87: to run `npm` as part of `Makefile` we need to have nodejs installed globally

mykalmax: It should work during release but if just run `make ui-build` locally without `node` , `npm` won't be there.

eakmanrq: Yeah I think that is fine the person would get an error. This is something that only people with access to tobiko data creds can do.

eakmanrq: I understand now and didn't realize that even when doing UI work people were building from within docker. Updated: https://github.com/TobikoData/sqlmesh/pull/1331/commits/fec490a23d770982ed57a6de8ef7015b57dc7fef

docker-compose.build.yml
```diff
@@ -0,0 +1,6 @@
+version: "3.11"
+services:
+  app:
+    command: bash -c "npm --prefix /app ci && npm --prefix /app run build"
+    networks: []
+    depends_on: []
```
mykalmax, line 6: we still need to depend on `generate-openapi-spec`

mykalmax (approved):

PR #1333, opened 2023-08-22
Fix: Remove nodes and edges from shared state to avoid re-renders

mykalmax:


web/client/src/library/components/graph/context.tsx
```diff
@@ -101,8 +94,8 @@ export const LineageFlowContext = createContext<LineageFlow>({
   setConnections: () => {},
   setSelectedNodes: () => {},
   setMainNode: () => {},
-  setNodes: () => {},
-  setEdges: () => {},
+  // setNodes: () => {},
+  // setEdges: () => {},
```
vchan, line 98:
```suggestion
```

vchan (approved):

PR #1339, opened 2023-08-23
Fix: Return context if loaded context failed

mykalmax:


izeigerman:
@vchan will appreciate your review here

vchan:
Was there something that prompted this change? Won't this cause problems for endpoints that are expecting a loaded context but instead got an unloaded one? An unloaded context won't have DAG/models and can't plan, produce lineage, etc.

mykalmax:
If we have broken model file and user do `sqlmesh ui` it wont start the server and throw `Error: Failed to parse a model definition`

PR #1345, opened 2023-08-24
Refactor!: move project from snapshot to model/node

tobymao:


sqlmesh/core/model/decorator.py
```diff
@@ -58,6 +58,7 @@ def model(
         dialect: t.Optional[str] = None,
         time_column_format: str = c.DEFAULT_TIME_COLUMN_FORMAT,
         physical_schema_override: t.Optional[t.Dict[str, str]] = None,
+        project: str = "",
```
izeigerman, line 61: Should this be Optional rather than empty string?

sqlmesh/core/node.py
```diff
@@ -154,6 +155,7 @@ class Node(PydanticModel):
     """
 
     name: str
+    project: str = ""
```
izeigerman, line 158: Ditto: should this be optional?

sqlmesh/core/model/decorator.py
tobymao: i wanted to always be filled do you have a strong pref?

izeigerman (approved):
One question, LGTM otherwise

sqlmesh/core/model/decorator.py
izeigerman: Well, empty string is not exactly meaningful, so None is at least explicit. Up to you

PR #1348, opened 2023-08-25
Feature: Support dbt singular tests

crericha:
This PR introduces `StandaloneAudit`, which is a new Snapshot node type that is currently used to support dbt singular tests. A follow-on PR will add `StandaloneAudit` natively to SQLMesh.

The PR also brings in support for dbt's built-in relationship tests.

image

Remaining work to go from draft to PR:
- Add unit tests
- Update comments that assume snapshot == model


sqlmesh/core/audit/definition.py
```diff
@@ -247,22 +261,147 @@ def macro_definitions(self) -> t.List[d.MacroDef]:
...
+    def sorted_python_env(self) -> t.List[t.Tuple[str, Executable]]:
+        """Returns the python env sorted by executable kind and then var name."""
+        return sorted(self.python_env.items(), key=lambda x: (x[1].kind, x[0]))
+
+    @property
+    def dialect(self) -> str:
+        return self.audit.dialect
+
+    @property
+    def data_hash(self) -> str:
```
tobymao, line 333: audits don't have data hashes

sqlmesh/core/console.py
```diff
@@ -416,26 +416,48 @@ def show_model_difference_summary(
...
+            name for name, snapshot in context_diff.added.items() if snapshot.is_model
+        } - ignored_snapshot_names
         if added_model_names:
             added_tree = Tree(f"[bold][added]Added Models:")
             for model_name in added_model_names:
                 added_tree.add(f"[added]{model_name}")
             tree.add(added_tree)
 
-        removed_model_names = context_diff.removed - ignored_snapshot_names
+        added_audit_names = {
```
tobymao, line 428: why did you add this to the ui? we don't need to show audits in the console diff

sqlmesh/core/console.py
```diff
@@ -1094,20 +1118,42 @@ def show_model_difference_summary(
...
+            name for name, snapshot in context_diff.added.items() if snapshot.is_model
+        } - ignored_snapshot_names
         if added_model_names:
             self._print(f"**Added Models:**\n")
             for model_name in added_model_names:
                 self._print(f"- {model_name}\n")
             self._print("\n")
 
-        removed_model_names = context_diff.removed - ignored_snapshot_names
+        added_audit_names = {
```
tobymao, line 1130: remove all of this imo

sqlmesh/core/context_diff.py
```diff
@@ -41,9 +41,9 @@ class ContextDiff(PydanticModel):
     """Whether the currently stored environment record is in unfinalized state."""
     create_from: str
     """The name of the environment the target environment will be created from if new."""
-    added: t.Set[str]
```
tobymao, line 44: does any of this have to change?

sqlmesh/core/console.py
crericha: Discussed offline

sqlmesh/core/context_diff.py
crericha: Added the snapshots to the methods, so that consumers can easily interrogate information about the snapshot. In my case, the type of node.

I originally added an added, removed, modified field for each node type, but that wouldn't scale well if more types were added.

sqlmesh/core/audit/definition.py
```diff
@@ -163,7 +172,7 @@ def load_multiple(
 
     def render_query(
         self,
-        snapshot_or_model: t.Union[Snapshot, _Model],
+        snapshot_or_node: t.Union[Snapshot, _Model, StandaloneAudit],
```
izeigerman, line 175: Why not `t.Union[Snapshot, Node]` then?

izeigerman: This feels wrong to me. Why do we even need to pass `StandaloneAudit` here if we don't even use it? Should this argument just be Optional?

sqlmesh/core/audit/definition.py
```diff
@@ -247,22 +261,154 @@ def macro_definitions(self) -> t.List[d.MacroDef]:
...
+
+    _path: Path = Path()
+    _depends_on: t.Optional[t.Set[str]] = None
+
+    @property
+    def depends_on(self) -> t.Set[str]:
+        if self._depends_on is None:
+            self._depends_on = self.depends_on_ or set()
+
+            query = exp.select("*").from_("table")  # self.render_query(optimize=False)
```
izeigerman, line 312: What is this for? Why are we doing this?

sqlmesh/core/audit/definition.py
```diff
@@ -282,3 +428,7 @@ def _maybe_parse_arg_pair(e: exp.Expression) -> t.Tuple[str, exp.Expression]:
     if isinstance(e, exp.EQ):
         return e.left.name, e.right
     raise_config_error(f"Invalid defaults expression: {e}", error_type=AuditConfigError)
+
+
+def is_audit(node: Node) -> bool:
```
izeigerman, line 433: I think this is used in exactly one place. Do we really need a function for this?

sqlmesh/core/context_diff.py
```diff
@@ -41,9 +41,9 @@ class ContextDiff(PydanticModel):
...
+    added: t.Dict[str, Snapshot]
```
izeigerman, line 44: This feels redundant. Why users can just use the key with the `snapshots` dict?

sqlmesh/core/context_diff.py
```diff
@@ -41,9 +41,9 @@ class ContextDiff(PydanticModel):
...
     """New models."""
-    removed: t.Set[str]
+    removed: t.Dict[str, Snapshot]
```
izeigerman, line 46: Ditto

sqlmesh/core/model/definition.py
```diff
@@ -1916,3 +1917,7 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
     "hash_raw_query": exp.convert,
     "table_properties_": lambda value: value,
 }
+
+
+def is_model(node: Node) -> bool:
```
izeigerman, line 1922: Shouldn't this and `is_audit` be a property of `Node`?

sqlmesh/core/model/kind.py
```diff
@@ -88,6 +88,7 @@ class ModelKindName(str, ModelKindMixin, Enum):
     EMBEDDED = "EMBEDDED"
     SEED = "SEED"
     EXTERNAL = "EXTERNAL"
+    NONE = "NONE"
```
izeigerman, line 91: Where is this used?

sqlmesh/core/plan/definition.py
```diff
@@ -441,7 +442,7 @@ def has_unmodified_unpromoted(self) -> bool:
 
     def _add_restatements(self) -> None:
         def is_restateable_snapshot(snapshot: Snapshot) -> bool:
-            if not self.is_dev and snapshot.model.disable_restatement:
+            if not self.is_dev and snapshot.is_model and snapshot.model.disable_restatement:
```
izeigerman, line 445: I feel like `disable_restatement` should be a part of `Snapshot` and look like this:
```
return self.is_model and self.model.disable_restatement
```

sqlmesh/core/plan/definition.py
```diff
@@ -391,7 +392,7 @@ def _set_effective_from(self, effective_from: t.Optional[TimeLike]) -> None:
         self._effective_from = effective_from
 
         for snapshot in self.new_snapshots:
-            if not snapshot.model.disable_restatement:
+            if snapshot.is_model and not snapshot.model.disable_restatement:
```
izeigerman, line 395: Ditto about making `disable_restatement` a property of Snapshot

sqlmesh/core/scheduler.py
```diff
@@ -140,7 +140,10 @@ def evaluate(
             snapshot.name: snapshot,
         }
 
-        if isinstance(snapshot.model, SeedModel) and not snapshot.model.is_hydrated:
+        if (
+            isinstance(snapshot.node, SeedModel)
+            and not t.cast(SeedModel, snapshot.node).is_hydrated
```
izeigerman, line 145: Why can't this remain `snapshot.model.is_hydrated`?

sqlmesh/core/snapshot/definition.py
```diff
@@ -42,7 +42,8 @@
 
 Interval = t.Tuple[int, int]
 Intervals = t.List[Interval]
-SnapshotNode = Annotated[Model, Field(discriminator="source_type")]
+NodeClasses = t.Union[Model, StandaloneAudit]
```
izeigerman, line 45: Let's not use `Classes` as part of the type name. To me it feels like it's a list of type **instances** not a type. I suggest perhaps following the `Node` (for union) vs `_Node` (for base class) like we do everywhere else.

izeigerman: Seems like it's only used in the next line. Can we just inline?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -142,6 +141,11 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
             **common_render_kwargs,
         )
 
+        if not snapshot.is_model:
```
izeigerman, line 144: Why wait so long, why not return at the top?

sqlmesh/utils/errors.py
```diff
@@ -56,15 +59,20 @@ class AuditConfigError(ConfigError):
...
-        self, audit_name: str, model_name: str, count: int, query: exp.Subqueryable
+        self,
+        audit_name: str,
+        count: int,
+        query: exp.Subqueryable,
+        model: t.Optional[Model] = None,
     ) -> None:
         self.audit_name = audit_name
-        self.model_name = model_name
+        self.model_name = model.name if model else None
```
izeigerman, line 69: `model_name` can just be a property method.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -490,7 +484,7 @@ def _audit(
         if audit.skip:
             return AuditResult(
                 audit=audit,
-                model=snapshot.model,
+                model=snapshot.model if snapshot.is_model else None,
```
izeigerman, line 487: This seems like a fairly common pattern. Can this be a `snapshot.model_or_none` instead?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -767,21 +761,25 @@ def demote(
...
 
     def create(
         self,
-        model: Model,
+        snapshot: Snapshot,
         name: str,
         **render_kwargs: t.Any,
     ) -> None:
+        model = snapshot.model
+        self.adapter.execute(model.render_pre_statements(**render_kwargs))
```
izeigerman, line 781: Why did we move this for `create` but not for `evaluate`?

izeigerman: What if `create` is overridden in one of the strategies, do they also need to invoke pre / post statements too?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -958,10 +964,13 @@ def append(
...
```
izeigerman, line 972: That's what I'm talking about. What's the point of moving pre / post statements into strategies?

sqlmesh/core/state_sync/cache.py
```diff
@@ -68,8 +68,8 @@ def get_snapshots(
             elif snapshot:
                 if (
                     hydrate_seeds
-                    and isinstance(snapshot.model, SeedModel)
-                    and not snapshot.model.is_hydrated
+                    and isinstance(snapshot.node, SeedModel)
+                    and not t.cast(SeedModel, snapshot.model).is_hydrated
```
izeigerman, line 72: Why do we cast in some places but not others?

sqlmesh/core/state_sync/cache.py
```diff
@@ -80,7 +80,7 @@ def get_snapshots(
 
         for snapshot_id, snapshot in existing.items():
             cached = self._from_cache(snapshot_id, now)
-            if cached and (not isinstance(cached.model, SeedModel) or cached.model.is_hydrated):
+            if cached and (not isinstance(cached.node, SeedModel) or cached.node.is_hydrated):
```
izeigerman, line 83: We don't cast here

izeigerman:
Shall we add a standalone audit to the example dbt project? Or is there already one?

sqlmesh/core/audit/definition.py
crericha: The reason it explicitly unions the types is the use of `python_env`, which is not part of the `Node` api and will be needed for SQLMesh standalone tests.

I could change the type to `Node` and check for `StandaloneAudit` or `_Model` in `_create_query_renderer`. That's probably cleaner.

sqlmesh/core/audit/definition.py
crericha: Temporary code. Needs to be fixed.

sqlmesh/core/audit/definition.py
crericha: I did a grep and found it used 6 times.

sqlmesh/core/context_diff.py
crericha: If I'm reading the code correctly, removed snapshots are not part of ContextDiff. If these were added to ContextDiff, then yes, the dicts could be sets.

sqlmesh/core/model/definition.py
crericha: Is the suggestion to add to `Node` and have it return false, then override in `_Model` and `StandaloneAudit`?

sqlmesh/core/model/kind.py
crericha: Default value for `SnapshotTableInfo` `kind_name`, so that `model_kind_name` doesn't have to be optional (which had a spider effect throughout a lot of code).

sqlmesh/core/scheduler.py
crericha: `is_hydrated` is only supported for `SeedModel` and not part of the general `Model` interface, so it needs to be narrowed down. Maybe `snapshot.node.is_hydrated` will work though. I'll try that.

sqlmesh/core/snapshot/definition.py
```diff
@@ -292,7 +310,8 @@ class SnapshotTableInfo(PydanticModel, SnapshotInfoMixin, frozen=True):
     parents: t.Tuple[SnapshotId, ...]
     previous_versions: t.Tuple[SnapshotDataVersion, ...] = ()
     change_category: t.Optional[SnapshotChangeCategory] = None
-    kind_name: ModelKindName
+    kind_name: ModelKindName = ModelKindName.NONE
```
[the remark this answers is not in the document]
crericha: I tried to make it Optional at first, but it spiraled out of control as I updated clients of `SnapshotTableInfo` to support an Optional value for `model_kind_name`. IMHO, this approach made the code more readable.

sqlmesh/core/snapshot/evaluator.py
crericha: I can't recall why. Will move to the top.

sqlmesh/core/snapshot/evaluator.py
crericha: I like that!

sqlmesh/core/snapshot/evaluator.py
crericha: I'll move it back for `create` and early return if not a `model` like done in `evaluate`.

crericha:
> Shall we add a standalone audit to the example dbt project? Or is there already one?

There is one already.

sqlmesh/core/context_diff.py
izeigerman: I see. Do we actually need a full-blown snapshot or the Table Info from the environment will suffice?

sqlmesh/core/model/definition.py
izeigerman: yeah, that would work too

sqlmesh/core/model/kind.py
izeigerman: It did? I thought the impact was minimal since we did `==` everywhere. I could only find like 2 places where it matters. Did you find more?

sqlmesh/core/context_diff.py
crericha: The TableInfo would suffice. We have the removed Snapshot in a local var when constructing context diff. I can add a removed_snapshots dict if that's preferable. What do you think?

sqlmesh/core/model/kind.py
crericha: The method originates from the `ModelKindMixin` and making the return type optional there is what caused the issue.

sqlmesh/core/audit/definition.py
```diff
@@ -247,22 +261,173 @@ def macro_definitions(self) -> t.List[d.MacroDef]:
...
+    python_env_: t.Optional[t.Dict[str, Executable]] = Field(default=None, alias="python_env")
+    source_type: Literal["audit"] = "audit"
+
+    _path: Path = Path()
+    _depends_on: t.Optional[t.Set[str]] = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_fields(
+        cls, values: t.Dict[str, t.Union[t.Tuple[str, ...], t.Optional[str], t.Dict[str, t.Any]]]
```
izeigerman, line 317: I think this type signature is useless to the point that you're better off just using `t.Dict[str, t.Any]`. Nobody can decipher this anyway. Same for return value.

sqlmesh/core/audit/definition.py
```diff
@@ -247,22 +261,173 @@ def macro_definitions(self) -> t.List[d.MacroDef]:
...
+    @property
+    def data_hash(self) -> str:
+        """
+        Computes the data hash for the node.
+
+        Returns:
+            The data hash for the node.
+        """
+        # StandaloneAudits do not have a data hash
+        return ""
```
izeigerman, line 357: This should be either `"0"` or `hash_data("")`.

sqlmesh/core/console.py
```diff
@@ -483,6 +458,75 @@ def _get_ignored_tree(
             )
         return ignored
 
+    def _show_summary_tree_for(
+        self,
+        context_diff: ContextDiff,
+        header: str,
+        filter: t.Callable,
```
izeigerman, line 465: `filter` is an existing Python keyword (or rather builtin function). I suggest coming up with a different name.

izeigerman: If it's a predicate than I think expanding the type signature makes sense: `t.Callable[[Snapshot], bool]`.

sqlmesh/core/console.py
```diff
@@ -483,6 +458,75 @@ def _get_ignored_tree(
...
+        removed_names = {
+            name for name, snapshot in context_diff.removed_snapshots.items() if filter(snapshot)
+        } - filtered_ignored_names
+        modified_names = {
+            name
+            for name, snapshots in context_diff.modified_snapshots.items()
+            if filter(snapshots[0])
+        } - filtered_ignored_names
+
+        if (
```
izeigerman, line 486: [Nit] 
```
name_sets = (added_names, removed_names, modified_names, filtered_ignored_names)
if all(not names for names in name_sets):
    return
```

sqlmesh/core/context_diff.py
```diff
@@ -172,7 +177,9 @@ def create(
             is_unfinalized_environment=bool(env and not env.finalized_ts),
             create_from=create_from,
             added=added,
-            removed=removed,
+            removed_snapshots={
+                name: info for name, info in existing_info.items() if name in removed
```
izeigerman, line 181: [Nit] Considering that we expect `removed` to be empty most of the time I think you're better off iterating over `removed` instead and then getting the `info` instance from the `existing_info` dict.

sqlmesh/core/context_diff.py
```diff
@@ -125,8 +127,11 @@ def create(
             modified = modified_info.get(name)
             existing = stored.get(snapshot.snapshot_id)
 
-            if existing:
-                # Keep the original model instance to preserve the query cache.
+            if modified and snapshot.node_type != modified.node_type:
```
izeigerman, line 130: Shouldn't we remove it from the `modified` mapping then?

sqlmesh/core/audit/definition.py
crericha: The method returns a string. I used `""` to indicate there is no data hash vs a static data dash that `"0" `or `hash_data("") `would produce. Are you thinking there is no value in indicating there isn't a data hash?

sqlmesh/core/context_diff.py
```diff
@@ -202,77 +214,81 @@ def promotable_models(self) -> t.Set[str]:
...
-        if model not in self.modified_snapshots:
+        if node not in self.snapshots:
+            raise SQLMeshError(f"`{node}` does not exist.")
+        if node not in self.modified_snapshots:
             return ""
 
-        new, old = self.modified_snapshots[model]
-        return old.model.text_diff(new.model)
+        new, old = self.modified_snapshots[node]
+        if new.is_model:
```
izeigerman, line 290: Shouldn't `text_diff` be a behavior of a snapshot or a node to avoid this branching?

sqlmesh/core/audit/__init__.py
```diff
@@ -3,7 +3,14 @@
...
 from sqlmesh.core.audit import builtin
-from sqlmesh.core.audit.definition import Audit, AuditResult
+from sqlmesh.core.audit.definition import (
+    Audit,
+    AuditResult,
+    StandaloneAudit,
+    create_standalone_audit,
+)
+
+AuditType = t.Union[Audit, StandaloneAudit]
```
izeigerman, line 13: I don't think a type contain `Type` in its name.

sqlmesh/core/context_diff.py
crericha: modified isn't used again, but I can remove it to prevent bugs in future modifications.

sqlmesh/core/model/kind.py
```diff
@@ -88,17 +88,18 @@ class ModelKindName(str, ModelKindMixin, Enum):
...
```
izeigerman, line 91: Why is the value still here?

sqlmesh/core/snapshot/definition.py
```diff
@@ -291,7 +304,8 @@ class SnapshotTableInfo(PydanticModel, SnapshotInfoMixin, frozen=True):
     parents: t.Tuple[SnapshotId, ...]
     previous_versions: t.Tuple[SnapshotDataVersion, ...] = ()
     change_category: t.Optional[SnapshotChangeCategory] = None
-    kind_name: ModelKindName
+    kind_name: t.Optional[ModelKindName] = None
+    node_type_: NodeType = Field(default=NodeType.MODEL, alias="node_type")
```
izeigerman, line 308: Why do we need an alias if the property method just returns the value as is?

izeigerman: Why is this a property of the Snapshot and not the `_Node`?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -978,10 +987,13 @@ def append(
...
```
izeigerman, line 995: Why is this still here?

sqlmesh/core/model/kind.py
crericha: Not needed. Will remove.

sqlmesh/core/audit/__init__.py
crericha: Habit from that being the coding standard from where I worked.

sqlmesh/core/snapshot/definition.py
crericha: Its a property of SnapshotTableInfo, which doesn't have the node to figure out the type.

sqlmesh/core/snapshot/evaluator.py
crericha: Will fix.

izeigerman:
Please make sure that Airflow tests pass locally before merging

sqlmesh/core/audit/definition.py
```diff
@@ -12,66 +13,215 @@
...
-    dialect: str = ""
-    """The dialect of the audit query."""
-    skip: bool = False
-    """Setting this to `true` will cause this audit to be skipped. Defaults to `false`."""
-    blocking: bool = True
-    """Setting this to `true` will cause the pipeline execution to stop if this audit fails. Defaults to `true`."""
-    defaults: t.Dict[str, exp.Expression] = {}
-    """Default values for the audit query."""
+    dialect: str
+    skip: bool
```
izeigerman, line 56: Why should this be a part of the mixin?

sqlmesh/core/audit/definition.py
```diff
@@ -12,66 +13,215 @@
...
+    @property
+    def macro_definitions(self) -> t.List[d.MacroDef]:
+        """All macro definitions from the list of expressions."""
+        return [s for s in self.expressions if isinstance(s, d.MacroDef)]
+
+    def _create_query_renderer(self, node: _Node) -> QueryRenderer:
+        kwargs: t.Dict[str, t.Any] = {}
+
+        node_dialect = ""
+        if node.is_model:
```
izeigerman, line 160: Now when you use inheritence, can you just override and avoid branching here?

sqlmesh/core/audit/definition.py
```diff
@@ -12,66 +13,215 @@
...
+        Returns:
+            The rendered expression.
+        """
+        from sqlmesh.core.snapshot import Snapshot
+
+        node = snapshot_or_node if isinstance(snapshot_or_node, _Node) else snapshot_or_node.node
+        query_renderer = self._create_query_renderer(node)
+        extra_kwargs = {}
+
+        if node.is_model:
```
izeigerman, line 98: Can we just override the method now and avoid doing these checks?

sqlmesh/core/audit/definition.py
```diff
@@ -287,45 +179,184 @@ def _create_query_renderer(self, node: _Node) -> QueryRenderer:
         )
 
 
-class StandaloneAudit(_Node):
+@field_validator("name", "dialect", mode="before", check_fields=False)
+def audit_string_validator(v: t.Any) -> t.Optional[str]:
```
izeigerman, line 183: Shouldn't this be a class method and take class as the first argument?

sqlmesh/core/audit/definition.py
```diff
@@ -287,45 +179,184 @@ def _create_query_renderer(self, node: _Node) -> QueryRenderer:
...
+    if isinstance(v, exp.Expression):
+        return v.name.lower()
+    return str(v).lower() if v is not None else None
+
+
+@field_validator("defaults", mode="before", check_fields=False)
```
izeigerman, line 189: ditto

sqlmesh/core/node.py
```diff
@@ -299,6 +301,17 @@ def cron_floor(self, value: TimeLike) -> TimeLike:
         """
         return self.croniter(self.cron_next(value)).get_prev()
 
+    def text_diff(self, other: _Node) -> str:
```
izeigerman, line 304: This should be just `Node`.

sqlmesh/core/snapshot/definition.py
```diff
@@ -510,7 +510,7 @@ def from_node(
         node_audits = (
             tuple(t.cast(_Model, node).referenced_audits(audits or {}))
             if node.is_model
-            else tuple([t.cast(StandaloneAudit, node).audit])
+            else tuple()
```
izeigerman, line 513: Shouldn't this be `tuple(node)`?

sqlmesh/dbt/test.py
```diff
@@ -121,27 +121,32 @@ def to_sqlmesh(self, context: DbtContext) -> AuditType:
...
+                name=self.name,
+                dialect=context.dialect,
+                skip=skip,
+                query=query,
+                jinja_macros=jinja_macros,
+                depends_on={model.sql_name for model in test_context.refs.values()}.union(
+                    {source.sql_name for source in test_context.sources.values()}
+                ),
+                tags=self.tags,
+                # TODO node values
```
izeigerman, line 136: what's node values?

sqlmesh/core/audit/definition.py
```diff
@@ -374,54 +401,34 @@ def metadata_hash(self, audits: t.Dict[str, Audit]) -> str:
...
-            else self.audit.render_query(self) or self.audit.query
-        )
+        query = self.query if self.hash_raw_query else self.render_query(self) or self.query
         data.append(query.sql(comments=False))
 
         return hash_data(data)
 
-    def text_diff(self, other: StandaloneAudit) -> str:
-        """Produce a text diff against another standalone audit.
+    def text_diff(self, other: _Node) -> str:
```
izeigerman, line 409: Ditto: this should just be `Node`.

sqlmesh/core/model/definition.py
```diff
@@ -411,15 +412,20 @@ def referenced_audits(self, audits: t.Dict[str, Audit]) -> t.List[Audit]:
                 )
         return referenced_audits
 
-    def text_diff(self, other: Model) -> str:
-        """Produce a text diff against another model.
+    def text_diff(self, other: _Node) -> str:
```
izeigerman, line 415: Ditto

sqlmesh/core/model/definition.py
```diff
@@ -1179,7 +1185,7 @@ def render(
             df[string_columns] = df[string_columns].astype(str)
             yield df
 
-    def text_diff(self, other: Model) -> str:
+    def text_diff(self, other: _Node) -> str:
```
izeigerman, line 1188: Ditto

sqlmesh/core/snapshot/definition.py
crericha: This is for attached audits. I can maybe change the name.

sqlmesh/dbt/test.py
crericha: Node class fields. I plan to add those in the follow-on PR.

sqlmesh/core/snapshot/definition.py
crericha: Refactored to a kwarg, so now only used if the node is a model.

sqlmesh/core/plan/definition.py
```diff
@@ -553,7 +553,7 @@ def _categorize_snapshots(self) -> None:
             snapshot = self._snapshot_mapping.get(name)
             if not snapshot or snapshot.change_category:
                 continue
-            upstream_model_names = self._dag.upstream(name)
+            self._dag.upstream(name)
```
izeigerman, line 556: This is a dead line

izeigerman (approved):
One last comment, LGTM otherwise.

PR #1350, opened 2023-08-26
Feat: Add SCD Type 2 Support

eakmanrq:
This PR adds support for SCD Type 2 using `updated_at` timestamps. 

Follow Up PRs:
* Cleanup engine adapter Pandas overrides so it is easy to get the correct pandas implementation per engine
* Add "check column" support (checking values of columns instead of `updated_at`)
* Add dbt adapter support

### DIFF
--- a/docs/concepts/models/model_kinds.md
+++ b/docs/concepts/models/model_kinds.md
@@ -267,14 +267,16 @@ SCD Type 2 is a model kind that supports [slowly changing dimensions](https://en
 
 SQLMesh achieves this by adding a `valid_from` and `valid_to` column to your model. The `valid_from` column is the date that the record became valid (inclusive) and the `valid_to` column is the date that the record became invalid (exclusive). The `valid_to` column is set to `NULL` for the latest record.
 
-Therefore you can use these models to not only tell you what the latest value is for a given record but also what the values were anytime in the past. Note that maintaining this history does come at a cost of increased storage and compute and this may not be a good fit for sources that change frequently since the history could get very large. 
+Therefore you can use these models to not only tell you what the latest value is for a given record but also what the values were anytime in the past. Note that maintaining this history does come at a cost of increased storage and compute and this may not be a good fit for sources that change frequently since the history could get very large.
+
+Currently SCD Type 2 only supports sourcing from tables that have an "Updated At" timestamp defined in the table that tells you when a given was last updated. Soon we will also be supporting checking column values in cases where an update column is not available.
 
 This example specifies a `SCD_TYPE_2` model kind:
 ```sql linenums="1" hl_lines="3"
 MODEL (
   name db.menu_items,
   kind SCD_TYPE_2 (
-    unique_key (id),
+    unique_key id,
   )
 );
 
@@ -305,7 +307,7 @@ SQLMesh will automatically add the `valid_from` and `valid_to` columns to your t
 MODEL (
   name db.menu_items,
   kind SCD_TYPE_2 (
-    unique_key (id),
+    unique_key id,
     valid_from_name my_valid_from,
     valid_to_name my_valid_to
   )
@@ -329,7 +331,7 @@ The `updated_at` column name can also be changed by adding the following to your
 MODEL (
   name db.menu_items,
   kind SCD_TYPE_2 (
-    unique_key (id),
+    unique_key id,
     updated_at_name my_updated_at
   )
 );
--- a/docs/concepts/models/model_kinds.md
+++ b/docs/concepts/models/model_kinds.md
@@ -261,5 +261,273 @@ FROM db.employees;
 ## SEED
 The `SEED` model kind is used to specify [seed models](./seed_models.md) for using static CSV datasets in your SQLMesh project.
 
+## SCD Type 2
+
+SCD Type 2 is a model kind that supports [slowly changing dimensions](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row) (SCDs) in your SQLMesh project. SCDs are a common pattern in data warehousing that allow you to track changes to records over time. 
+
+SQLMesh achieves this by adding a `valid_from` and `valid_to` column to your model. The `valid_from` column is the date that the record became valid (inclusive) and the `valid_to` column is the date that the record became invalid (exclusive). The `valid_to` column is set to `NULL` for the latest record.
+
+Therefore you can use these models to not only tell you what the latest value is for a given record but also what the values were anytime in the past. Note that maintaining this history does come at a cost of increased storage and compute and this may not be a good fit for sources that change frequently since the history could get very large. 
+
+This example specifies a `SCD_TYPE_2` model kind:
+```sql linenums="1" hl_lines="3"
+MODEL (
+  name db.menu_items,
+  kind SCD_TYPE_2 (
+    unique_key (id),
+  )
+);
+
+SELECT
+  id::INT,
+  name::STRING,
+  price::DOUBLE,
+  updated_at::TIMESTAMP
+FROM 
+  stg.current_menu_items;
+```
+
+SQLMesh will materialize this table with the following structure:
+```sql linenums="1" hl_lines="3"
+TABLE db.menu_items (
+  id INT,
+  name STRING,
+  price DOUBLE,
+  updated_at TIMESTAMP,
+  valid_from TIMESTAMP,
+  valid_to TIMESTAMP
+);
+```
+
+### Column Names
+SQLMesh will automatically add the `valid_from` and `valid_to` columns to your table. If you would like to specify the names of these columns you can do so by adding the following to your model definition:
+```sql linenums="1" hl_lines="3"
+MODEL (
+  name db.menu_items,
+  kind SCD_TYPE_2 (
+    unique_key (id),
+    valid_from_name my_valid_from,
+    valid_to_name my_valid_to
+  )
+);
+```
+
+SQLMesh will materialize this table with the following structure:
+```sql linenums="1" hl_lines="3"
+TABLE db.menu_items (
+  id INT,
+  name STRING,
+  price DOUBLE,
+  updated_at TIMESTAMP,
+  my_valid_from TIMESTAMP,
+  my_valid_to TIMESTAMP
+);
+```
+
+The `updated_at` column name can also be changed by adding the following to your model definition:
+```sql linenums="1" hl_lines="3"
+MODEL (
+  name db.menu_items,
+  kind SCD_TYPE_2 (
+    unique_key (id),
+    updated_at_name my_updated_at
+  )
+);
+
+SELECT
+    id,
+    name,
+    price,
+    my_updated_at
+FROM
+    stg.current_menu_items;
+```
+
+SQLMesh will materialize this table with the following structure:
+```sql linenums="1" hl_lines="3"
+TABLE db.menu_items (
+  id INT,
+  name STRING,
+  price DOUBLE,
+  my_updated_at TIMESTAMP,
+  valid_from TIMESTAMP,
+  valid_to TIMESTAMP
+);
+```
+
+### Deletes
+
+Deletes are supported and when an item is detected as deleted then the `valid_to` column will be set to the time when SQLMesh started running (called `execution_time`). If a deleted column is added back in the source then it will be inserted back into the table with `valid_from` set to the largest of either the `updated_at` timestamp of the new record or the `valid_from` timestamp of the deleted record in the SCD Type 2 table.
+
+### Example of SCD Type 2 in Action
+
+Lets say that you started with the following data in your source table:
+
+| ID | Name             | Price |     Updated At      |
+|----|------------------|:-----:|:-------------------:|
+| 1  | Chicken Sandwich | 10.99 | 2020-01-01 00:00:00 |
+| 2  | Cheeseburger     | 8.99  | 2020-01-01 00:00:00 |
+| 3  | French Fries     | 4.99  | 2020-01-01 00:00:00 |
+
+The target table, which is currently empty, will be materialized with the following data:
+
+| ID | Name             | Price |     Updated At      |     Valid From      | Valid To |
+|----|------------------|:-----:|:-------------------:|:-------------------:|:--------:|
+| 1  | Chicken Sandwich | 10.99 | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 |   NULL   |
+| 2  | Cheeseburger     | 8.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 |   NULL   |
+| 3  | French Fries     | 4.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 |   NULL   |
+
+Now lets say that you update the source table with the following data:
+
+| ID | Name             | Price |     Updated At      |
+|----|------------------|:-----:|:-------------------:|
+| 1  | Chicken Sandwich | 12.99 | 2020-01-02 00:00:00 |
+| 3  | French Fries     | 4.99  | 2020-01-01 00:00:00 |
+| 4  | Milkshake        | 3.99  | 2020-01-02 00:00:00 |
+
+Summary of Changes: 
+* The price of the Chicken Sandwich was increased from $10.99 to $12.99.
+* Cheeseburger was removed from the menu.
+* Milkshakes were added to the menu.
+
+Assuming your pipeline ran at `2020-01-02 02:00:00`, target table will be updated with the following data:
+
+| ID | Name             | Price |     Updated At      |     Valid From      |      Valid To       |
+|----|------------------|:-----:|:-------------------:|:-------------------:|:-------------------:|
+| 1  | Chicken Sandwich | 10.99 | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 | 2020-01-02 00:00:00 |
+| 1  | Chicken Sandwich | 12.99 | 2020-01-02 00:00:00 | 2020-01-02 00:00:00 |        NULL         |
+| 2  | Cheeseburger     | 8.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 | 2020-01-02 02:00:00 |
+| 3  | French Fries     | 4.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 |        NULL         |
+| 4  | Milkshake        | 3.99  | 2020-01-02 00:00:00 | 2020-01-02 00:00:00 |        NULL         |
+
+For our final pass, lets say that you update the source table with the following data:
+
+| ID | Name                | Price |     Updated At      |
+|----|---------------------|:-----:|:-------------------:|
+| 1  | Chicken Sandwich    | 14.99 | 2020-01-03 00:00:00 |
+| 2  | Cheeseburger        | 8.99  | 2020-01-03 00:00:00 |
+| 3  | French Fries        | 4.99  | 2020-01-01 00:00:00 |
+| 4  | Chocolate Milkshake | 3.99  | 2020-01-02 00:00:00 |
+
+Summary of changes:
+* The price of the Chicken Sandwich was increased from $12.99 to $14.99 (must be good!)
+* Cheeseburger was added back to the menu with original name and price.
+* Milkshake name was updated to be "Chocolate Milkshake".
+
+Target table will be updated with the following data:
+
+| ID | Name                | Price |     Updated At      |     Valid From      |      Valid To       |
+|----|---------------------|:-----:|:-------------------:|:-------------------:|:-------------------:|
+| 1  | Chicken Sandwich    | 10.99 | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 | 2020-01-02 00:00:00 |
+| 1  | Chicken Sandwich    | 12.99 | 2020-01-02 00:00:00 | 2020-01-02 00:00:00 | 2020-01-03 00:00:00 |
+| 1  | Chicken Sandwich    | 14.99 | 2020-01-03 00:00:00 | 2020-01-03 00:00:00 |        NULL         |
+| 2  | Cheeseburger        | 8.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 | 2020-01-02 02:00:00 |
+| 2  | Cheeseburger        | 8.99  | 2020-01-03 00:00:00 | 2020-01-03 00:00:00 |        NULL         |
+| 3  | French Fries        | 4.99  | 2020-01-01 00:00:00 | 1970-01-01 00:00:00 |        NULL         |
+| 4  | Milkshake           | 3.99  | 2020-01-02 00:00:00 | 2020-01-02 00:00:00 | 2020-01-03 00:00:00 |
+| 4  | Chocolate Milkshake | 3.99  | 2020-01-03 00:00:00 | 2020-01-03 00:00:00 |        NULL         |
+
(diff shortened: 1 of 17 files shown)


docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
 ## SEED
 The `SEED` model kind is used to specify [seed models](./seed_models.md) for using static CSV datasets in your SQLMesh project.
 
+## SCD Type 2
+
+SCD Type 2 is a model kind that supports [slowly changing dimensions](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row) (SCDs) in your SQLMesh project. SCDs are a common pattern in data warehousing that allow you to track changes to records over time. 
+
+SQLMesh achieves this by adding a `valid_from` and `valid_to` column to your model. The `valid_from` column is the date that the record became valid (inclusive) and the `valid_to` column is the date that the record became invalid (exclusive). The `valid_to` column is set to `NULL` for the latest record.
```
tobymao, line 268: what's the format of valid_from and valid_to?

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
...
+
+Therefore you can use these models to not only tell you what the latest value is for a given record but also what the values were anytime in the past. Note that maintaining this history does come at a cost of increased storage and compute and this may not be a good fit for sources that change frequently since the history could get very large. 
+
+This example specifies a `SCD_TYPE_2` model kind:
+```sql linenums="1" hl_lines="3"
+MODEL (
+  name db.menu_items,
+  kind SCD_TYPE_2 (
+    unique_key (id),
```
tobymao, line 277: does this need parens?

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
...
+  stg.current_menu_items;
+```
+
+SQLMesh will materialize this table with the following structure:
+```sql linenums="1" hl_lines="3"
+TABLE db.menu_items (
+  id INT,
+  name STRING,
+  price DOUBLE,
+  updated_at TIMESTAMP,
```
tobymao, line 296: is updated_at mandatory?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -748,6 +749,153 @@ def _merge(
...
+                    },
+                )
+            )
+        start_time = "CAST('1970-01-01 00:00:00' AS TIMESTAMP)"
+        query = (
+            exp.Select()  # type: ignore
+            .with_(
+                "source",
+                exp.select(*unmanaged_columns)
+                .distinct(*unique_key)
```
tobymao, line 789: is this distinct(...) supported in most dialects?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -748,6 +749,153 @@ def _merge(
...
+                )
+                .from_("deleted")
+                .group_by(*unique_key),
+            )
+            # Do a full join between latest records and source table in order to combine them together
+            .with_(
+                "joined",
+                exp.select(
+                    *(f"latest.{col} AS t_{col}" for col in columns_to_types),
+                    *(f"source.{col} as s_{col}" for col in unmanaged_columns),
```
tobymao, line 828: AS

sqlmesh/core/engine_adapter/base.py
```diff
@@ -748,6 +749,153 @@ def _merge(
...
+                        WHEN t_{valid_from_name} IS NULL 
+                        THEN {start_time} 
+                        ELSE t_{valid_from_name} 
+                    END AS {valid_from_name}""",
+                    f"""
+                    CASE 
+                        WHEN s_{updated_at_name} > t_{updated_at_name} 
+                        THEN s_{updated_at_name} 
+                        WHEN s_{unique_key[0]} IS NULL 
+                        THEN CAST('{to_ts(execution_time)}' AS TIMESTAMP)
```
tobymao, line 856: why ts and the previous with timestamp? just do 1970-01-01 and ds?

sqlmesh/core/model/kind.py
```diff
@@ -277,6 +286,26 @@ class FullKind(_ModelKind):
     name: Literal[ModelKindName.FULL] = ModelKindName.FULL
 
 
+class SCDType2Kind(_ModelKind):
+    name: Literal[ModelKindName.SCD_TYPE_2] = ModelKindName.SCD_TYPE_2
+    unique_key: t.List[str]
+    valid_from_name: str = "valid_from"
```
tobymao, line 292: maybe valid_from_column etc

sqlmesh/utils/date.py
```diff
@@ -212,7 +212,11 @@ def date_dict(
 
 def to_ds(obj: TimeLike) -> str:
     """Converts a TimeLike object into YYYY-MM-DD formatted string."""
-    return to_datetime(obj).isoformat()[0:10]
+    return to_ts(obj)[0:10]
+
+
+def to_ts(obj: TimeLike) -> str:
```
tobymao, line 218: docstring

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,64 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
...S "id", COALESCE("t_name", "s_name") AS "name", COALESCE("t_price", "s_price") AS "price", COALESCE("t_test_updated_at", "s_test_updated_at") AS "test_updated_at", CASE WHEN "t_test_valid_from" IS NULL AND NOT "latest_deleted"."id" IS NULL THEN CASE WHEN "latest_deleted"."test_valid_to" > "s_test_updated_at" THEN "latest_deleted"."test_valid_to" ELSE "s_test_updated_at" END WHEN "t_test_valid_from" IS NULL THEN CAST(\'1970-01-01 00:00:00\' AS TIMESTAMP) ELSE "t_test_valid_from" END AS "test_valid_from", CASE WHEN "s_test_updated_at" > "t_test_updated_at" THEN "s_test_updated_at" WHEN "s_id" IS NULL THEN CAST(\'2020-01-01T00:00:00+00:00\' AS TIMESTAMP) ELSE "t_test_valid_to" END AS "test_valid_to" FROM "joined" LEFT JOIN "latest_deleted" ON "joined"."s_id" = "latest_deleted"."id"), "updated_rows" AS (SELECT "s_id" AS "id", "s_name" AS "name", "s_price" AS "price", "s_test_updated_at" AS "test_updated_at", "s_test_updated_at" AS "test_valid_from", NULL AS "test_valid_to" FROM "joined" WHERE NOT "t_id" IS NULL AND NOT "s_id" IS NULL AND "s_test_updated_at" > "t_test_updated_at") SELECT * FROM "static" UNION ALL SELECT * FROM "non_updated_rows" UNION ALL SELECT * FROM "updated_rows"'''
```
tobymao, line 729: can you format this with sqlglot so it's readable?

tobymao:
nice work, i'll give it another look monday

docs/concepts/models/model_kinds.md
eakmanrq: They are timestamps so the format is irrelevant. I don't define a precision so I'm curious if we later get feedback about that and wanting to customize but I think it is fine for now.

docs/concepts/models/model_kinds.md
eakmanrq: Yes I assert this when trying to evaluate an SCD Type 2.

sqlmesh/core/engine_adapter/base.py
eakmanrq: I actually stole this from logical merge mixin: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/mixins.py#L53

I assumed since it was there that it was considered safe. I can rewrite both if we aren't confident.

sqlmesh/core/engine_adapter/base.py
eakmanrq: I find it easier to follow since we are wanting a timestamp to provide a timestamp. Yes it is correct that we could provide a date and cast it to timestamp I don't see the advantage to that.

sqlmesh/core/model/kind.py
eakmanrq: Actually I'm not sure since one distinct aspect of these parameters is that these are names that we will use when creating the table and not columns that are represented in the SQL itself. So for example `time_column` maps to a column in your SQL so calling this `valid_from_column` matches that pattern and could leave an impression it should be defined in the SQL. Breaking that pattern with `name` can help prevent a user from associating the two (since they are fundamentally different).

sqlmesh/core/engine_adapter/base.py
crericha: We tried `distinct` on 4 or 5 of the most used dialects when logical merge was in development. It was supported in all of them.

tobymao: sounds good

docs/concepts/models/model_kinds.md
tobymao: maybe put that in the initial description

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
...
+  my_valid_to TIMESTAMP
+);
+```
+
+The `updated_at` column name can also be changed by adding the following to your model definition:
+```sql linenums="1" hl_lines="3"
+MODEL (
+  name db.menu_items,
+  kind SCD_TYPE_2 (
+    unique_key (id),
```
tobymao, line 332: can you remove the parenthesis in the examples?

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
...
+
+Deletes are supported and when an item is detected as deleted then the `valid_to` column will be set to the time when SQLMesh started running (called `execution_time`). If a deleted column is added back in the source then it will be inserted back into the table with `valid_from` set to the largest of either the `updated_at` timestamp of the new record or the `valid_from` timestamp of the deleted record in the SCD Type 2 table.
+
+### Example of SCD Type 2 in Action
+
+Lets say that you started with the following data in your source table:
+
+| ID | Name             | Price |     Updated At      |
+|----|------------------|:-----:|:-------------------:|
+| 1  | Chicken Sandwich | 10.99 | 2020-01-01 00:00:00 |
```
tobymao, line 368: i'm afraid if we use time stamp strings like this we may have problems with time zones, i think different engines have different defaults

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,273 @@ FROM db.employees;
...
+
+```sql linenums="1" hl_lines="3"
+SELECT 
+    *
+FROM
+    menu_items
+WHERE
+    id = 1
+    AND '2020-01-02 01:00:00' >= valid_from
+    AND '2020-01-02 01:00:00' < COALESCE(valid_to, '2199-12-31 23:59:59');
```
tobymao, line 469: AND ('...' < valid_to or valid_to is null), i don't like made up random numbers

sqlmesh/core/model/kind.py
```diff
@@ -94,6 +99,17 @@ def model_kind_name(self) -> ModelKindName:
         return self
 
 
+def _unique_key_validator(v: t.Any) -> t.List[str]:
+    if isinstance(v, exp.Identifier):
+        return [v.this]
```
tobymao, line 104: this isn't guaranteed to run a string for tuple, 

maybe change all this to .name

sqlmesh/core/model/kind.py
tobymao: ok

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,314 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
+        adapter.cursor.execute.call_args[0][0]
+        == parse_one(
+            """
+CREATE OR REPLACE TABLE "target" AS
+WITH "source" AS (
+  SELECT DISTINCT ON ("id")
+    "id",
+    "name",
+    "price",
+    "test_updated_at"
```
tobymao, line 739: teste_updated_at doesn't exist from source, only updated_at

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,314 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
+  FROM "static"
+  LEFT JOIN "latest"
+    USING ("id")
+  WHERE
+    "latest"."test_valid_to" IS NULL
+), "latest_deleted" AS (
+  SELECT
+    "id",
+    MAX("test_valid_to") AS "test_valid_to"
+  FROM "deleted"
```
tobymao, line 781: this query means that the join on deleted has to be done an extra time, should we add a rank function on deleted to have a latest deleted flag instead?

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,314 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
+    COALESCE("t_test_updated_at", "s_test_updated_at") AS "test_updated_at",
+    CASE
+      WHEN "t_test_valid_from" IS NULL AND NOT "latest_deleted"."id" IS NULL
+      THEN CASE
+        WHEN "latest_deleted"."test_valid_to" > "s_test_updated_at"
+        THEN "latest_deleted"."test_valid_to"
+        ELSE "s_test_updated_at"
+      END
+      WHEN "t_test_valid_from" IS NULL
+      THEN CAST(\'1970-01-01 00:00:00\' AS TIMESTAMP)
```
tobymao, line 811: why is there a back tick on this single quote, doesn't look right

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,314 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
+  WHERE
+    NOT "t_id" IS NULL AND NOT "s_id" IS NULL AND "s_test_updated_at" > "t_test_updated_at"
+)
+SELECT
+  *
+FROM "static"
+UNION ALL
+SELECT
+  *
+FROM "non_updated_rows"
```
tobymao, line 842: got a bit confused because non_updated_rows is getting updated if it's null and no longer null right?

tests/core/engine_adapter/test_base.py
```diff
@@ -704,6 +705,314 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable):
...
+    "test_updated_at"
+  FROM (
+    SELECT
+      CAST("id1" AS INT) AS "id1",
+      CAST("id2" AS INT) AS "id2",
+      CAST("name" AS VARCHAR) AS "name",
+      CAST("price" AS DOUBLE) AS "price",
+      CAST("test_updated_at" AS TIMESTAMP) AS "test_updated_at"
+    FROM (VALUES
+      (1, 4, \'muffins\', 4.0, \'2020-01-01 10:00:00\'),
```
tobymao, line 901: back ticks?

sqlmesh/core/model/kind.py
```diff
@@ -277,6 +286,26 @@ class FullKind(_ModelKind):
...
+    valid_to_name: str = "valid_to"
+    updated_at_name: str = "updated_at"
+
+    forward_only: bool = True
+    disable_restatement: bool = True
+
+    _unique_key_validator = unique_key_validator
+
+    @property
+    def _managed_columns(self) -> t.Dict[str, exp.DataType]:
```
izeigerman, line 302: [Nit] Why private?

sqlmesh/core/model/definition.py
```diff
@@ -530,8 +541,12 @@ def columns_to_types_or_raise(self) -> t.Dict[str, exp.DataType]:
     @property
     def annotated(self) -> bool:
         """Checks if all column projection types of this model are known."""
-        columns_to_types = self.columns_to_types
-        if columns_to_types is None:
+        if self.columns_to_types is None:
+            return False
+        columns_to_types = {
+            k: v for k, v in self.columns_to_types.items() if k not in self.managed_columns
```
izeigerman, line 547: Why does it matter whether it's a managed column or not?

docs/concepts/models/model_kinds.md
eakmanrq: I see what you are saying but this example then goes with the view example of doing a COALESCE to support between. At the end of the day these are just examples and users can determine what makes sense in their given environment.

docs/concepts/models/model_kinds.md
eakmanrq: Can you explain more your concern? This is just an example so obviously doesn't have an issue. Are you talking about where we use 1970 timestamp string when creating the actual scd type 2 query?

tests/core/engine_adapter/test_base.py
eakmanrq: Damn nice catch

tests/core/engine_adapter/test_base.py
eakmanrq: I considered a rank but wasn't sure how universally it was supported across current and future engines while obviously this would be supported. Really this aggregate should be pretty cheap and the result set should be pretty small since it is only deleted rows. So my preference was to stick with what I knew was universally supported in this case.

tests/core/test_integration.py
```diff
@@ -797,6 +808,173 @@ def test_ignored_snapshots(sushi_context: Context):
...
+        df1 = df1.sort_values(by=["id1", "id2"]).reset_index(drop=True)
+        df2 = df2.sort_values(by=["id1", "id2"]).reset_index(drop=True)
+        pd.testing.assert_frame_equal(df1, df2)
+
+    shutil.copy(str(pathlib.Path("examples/sushi/config.py")), str(tmp_path / "config.py"))
+    (tmp_path / "models").mkdir()
+    with open(tmp_path / "models" / "scd_type_2.sql", "w+") as f:
+        f.write(
+            """
+MODEL (
```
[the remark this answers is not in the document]
eakmanrq: I initially converted `sushi.items` to scd type 2 (it currently it similar and does a daily snapshot history) but I found that we had a bunch of tests written assuming it was a incremental model and I didn't want to go in and update all of those tests. Really the key concern with scd type 2 is how it handles the variety of updates it gets (add, updates, deletes) which is covered by this test. It didn't seem high enough return to actual change `sushi.items` for the time it would take to do given what can be covered with just this test.

sqlmesh/core/model/definition.py
eakmanrq: This method is telling us if all column projections are known (either explicitly by user or inferred from the expression). We obviously know the types of the columns we manage. Therefore when doing this check we just want to consider the user defined columns in order to determine if we have figured out the annotations.

sqlmesh/core/model/kind.py
eakmanrq: Yeah good point just made it public

tests/core/engine_adapter/test_base.py
eakmanrq: Right was just looking at it from a different perspective when I wrote it but I think the way you are framing it is more intuitive. Will update.

docs/concepts/models/model_kinds.md
tobymao: i fundamentally disagree with your inclusion of them

docs/concepts/models/model_kinds.md
tobymao: we store the managed columns as timestamps right? for example if you store a timestamp in spark, it may use local time

tests/core/engine_adapter/test_base.py
tobymao: window functions (rank) is supported in every basically every sql engine, all the ones we support i'm sure

the aggregate is not the problem, it's the delete that has to be done multiple times i believe

tests/core/test_integration.py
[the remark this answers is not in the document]
tobymao: make a new model, don't change a current one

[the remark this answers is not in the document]
eakmanrq: I don't want to pollute the data model with a new scd type 2 just for the sake of having it in there, especially when we already have a model that should be a scd type 2 (`sushi.items`). Can you explain the issue with what is here? Is it the fact that we don't have an example in the examples folder? If so I am intending on fixing this once Trey is done creating the final examples and we are able to isolate these current examples to just tests.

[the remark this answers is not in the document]
tobymao: i don't understand the issue, sushi.examples is a test project and thus we can add directly in there all the new types, that's the whole point. i don't like how this is an inline example that's going to be harder to maintain

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,275 @@ FROM db.employees;
 ## SEED
 The `SEED` model kind is used to specify [seed models](./seed_models.md) for using static CSV datasets in your SQLMesh project.
 
+## SCD Type 2
+
+SCD Type 2 is a model kind that supports [slowly changing dimensions](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row) (SCDs) in your SQLMesh project. SCDs are a common pattern in data warehousing that allow you to track changes to records over time. 
+
+SQLMesh achieves this by adding a `valid_from` and `valid_to` column to your model. The `valid_from` column is the date that the record became valid (inclusive) and the `valid_to` column is the date that the record became invalid (exclusive). The `valid_to` column is set to `NULL` for the latest record.
```
tobymao, line 268: colum in the 'timestamp'

docs/concepts/models/model_kinds.md
```diff
@@ -261,5 +261,275 @@ FROM db.employees;
...
+| 4  | Chocolate Milkshake | 3.99  | 2020-01-03 00:00:00 | 2020-01-03 00:00:00 |        NULL         |
+
+Note: `Cheeseburger` was deleted from `2020-01-02 02:00:00` to `2020-01-03 00:00:00` meaning if you queried the table during that time range then you would not see `Cheeseburger` in the menu. This is the most accurate representation of the menu based on the source data provided. If `Cheeseburger` were added back to the menu with it's original updated at timestamp of `2020-01-01 00:00:00` then the `valid_from` timestamp of the new record would have been `2020-01-02 02:00:00` resulting in no period of time where the item was deleted. Since in this case the updated at timestamp did not change it is likely the item was removed in error and this again most accurately represents the menu based on the source data. 
+
+### Querying SCD Type 2 Models
+
+#### Querying the current version of a record
+
+Although SCD Type 2 models support history, it is still very easy to querying just the latest version of a record. Simply query the model as you would any other table. For example, if you wanted to query the latest version of the `menu_items` table you would simply run:
```
tobymao, line 438: it is still very easy to query

tobymao (commented):
z

sqlmesh/core/engine_adapter/base.py
```diff
@@ -774,6 +775,153 @@ def _merge(
...
+            .with_(
+                "source",
+                exp.select(*unmanaged_columns)
+                .distinct(*unique_key)
+                .from_(source_table.subquery("raw_source")),  # type: ignore
+            )
+            # Historical Records that Do Not Change
+            .with_(
+                "static",
+                exp.Select()
```
tobymao, line 821: exp.select(*columns_to_types).from_...

sqlmesh/core/engine_adapter/base.py
```diff
@@ -774,6 +775,153 @@ def _merge(
...
+                        WHEN s_{unique_key[0]} IS NULL 
+                        THEN CAST('{to_ts(execution_time)}' AS TIMESTAMP) 
+                        ELSE t_{valid_to_name} 
+                    END AS {valid_to_name}""",
+                )
+                .from_("joined")
+                .join(
+                    "latest_deleted",
+                    on=" AND ".join(
+                        [f"joined.s_{col} = latest_deleted.{col}" for col in unique_key]
```
tobymao, line 890: the brackets are unnecessary

tobymao (approved):

Issue #1352, opened 2023-08-29
Bigquery parse error when escaping `-`


serkef:
I'm having some issues with quotes on model names and tables.

## Problem:
For a model with a name: `my-project.example.mymodel` the following table is attempted to be created and fails
```
sqlglot.errors.ParseError: Failed to parse 'my-project.sqlmesh__example."my-project"__example__mymodel__1033600874' into <class 'sqlglot.expressions.Table'>
```
got: `'my-project.sqlmesh__example."my-project"__example__mymodel__1033600874'`
expected: `'my-project.sqlmesh__example.mymodel__1033600874'`

## Reproduction steps:
- Create a big query project containing a dash: eg `my-project`. Create also a dataset and a table to be used as external models. eg `src_dataset` and `src_table`.
- Create a model like the following:
```
MODEL
(
  name `my-project.example.mymodel`,
  kind FULL
);

SELECT id
FROM `my-project.src_dataset.src_table`
```
- Run `create_external_models`, you will see the `schema.yaml` correctly populating source model: `"my-project".src_dataset.src_table`
- Run plan. You will see:
```
├── Added Models:
│   ├── "my-project".src_dataset.src_table
│   └── "my-project".example.mymodel
```
and if you continue with the plan, you will get the `ParseError` above.
I have configured BigQuery dialect and a big query connection with default dataset (`my-project`) in my config.

Community discussion: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1693238988723709

serkef:
I've narrowed it down to the snapshot materialization name logic:
https://github.com/TobikoData/sqlmesh/blob/86a11d51d4b67b9c7ba09f624d5c6d9fc4fef081/sqlmesh/core/snapshot/definition.py#L937

This adds the catalog name to the materialized table.

[Here](https://github.com/TobikoData/sqlmesh/blob/88da4d9fccff7aeab790f9409d0d978aaabb02fd/tests/core/test_snapshot.py#L588), I confirmed with a test, that:
```
make_snapshot(SqlModel(name='"my-catalog".db.table', query=parse_one("select 1, ds"))).table_name(is_dev=False, for_read=False)
> 'my-catalog.sqlmesh__db."my-catalog"__db__table__2528031000'
```
which is illegal name, at least for BQ.

I think the best way is to keep the catalog name, but making it safe first. not sure if we have a util function for that 🤔

PR #1353, opened 2023-08-29
Fix: snapshot table name to support illegal characters

serkef:
Fixes: https://github.com/TobikoData/sqlmesh/issues/1352

Maybe a better solution than `str(exp.to_table(name).parts[0])` would be if sqlglot expression would return the quoted strings by default (`exp.to_table(name).catalog`). but I'm not sure if that's really wanted

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -783,7 +783,8 @@ class MaterializableStrategy(PromotableStrategy):
         **render_kwargs: t.Any,
     ) -> None:
         table = exp.to_table(name)
-        self.adapter.create_schema(table.db, catalog_name=table.catalog)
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -783,7 +783,8 @@ def create(
         **render_kwargs: t.Any,
     ) -> None:
         table = exp.to_table(name)
-        self.adapter.create_schema(table.db, catalog_name=table.catalog)
+        catg = str(table.parts[0]) if table.catalog else ""
```
tobymao, line 786: is this change needed? i don't know if this does anything

tobymao:
@serkef thanks for the PR, i'm going to take to the finish line in another PR.

PR #1354, opened 2023-08-29
Fix: Apply virtualization to rows

mykalmax:


web/client/src/library/components/table/Table.tsx
```diff
@@ -74,6 +74,11 @@ export default function Table({
   })
   const virtualRows = rowVirtualizer.getVirtualItems()
   const totalSize = rowVirtualizer.getTotalSize()
+  const paddingTop = virtualRows.length > 0 ? virtualRows?.[0]?.start ?? 0 : 0
```
vchan, line 77: Can virtualRows be null or undefined? Then shouldn't the optional chaining be done in the condition, i.e. `virtualRows?.length ?? 0 > 0`?

vchan (approved):

web/client/src/library/components/table/Table.tsx
mykalmax: `virtualRows?.[0]` or `virtualRows[0].start`  can be `null` for some reason

PR #1355, opened 2023-08-29
Feat: Support forward-only models in regular plans

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -743,6 +727,21 @@ def _build_snapshots_and_dag(
...
+    def _is_forward_only_model(self, model_name: str) -> bool:
+        snapshot = self._snapshot_mapping.get(model_name)
+        if snapshot and snapshot.model.forward_only:
+            return True
+
+        for upstream in self._dag.upstream(model_name):
+            upstream_snapshot = self._snapshot_mapping.get(upstream)
+            if (
+                upstream_snapshot
+                and upstream_snapshot.is_paused
```
tobymao, line 739: why does it matter if it's paused?

tobymao (approved):

sqlmesh/core/plan/definition.py
izeigerman: Discussed in DM, but paused means that this snapshot is not currently promoted on prod and the physical table associated with its version is not being populated by it. This essentially means that breaking / non-breaking changes can't safely be based on it, and only forward-only changes are supported.

PR #1358, opened 2023-08-29
Fix(bigquery): sanitize physical table name closes #1352

### LINKED ISSUE #1352: Bigquery parse error when escaping `-`

tobymao:


tobymao:
@serkef

sqlmesh/core/engine_adapter/base.py
```diff
@@ -472,12 +472,10 @@ def create_schema(
         warn_on_error: bool = True,
     ) -> None:
         """Create a schema from a name or qualified table name."""
-        if catalog_name:
-            schema_name = f"{catalog_name}.{schema_name}"
         try:
             self.execute(
                 exp.Create(
-                    this=exp.to_table(schema_name),
+                    this=exp.table_(schema_name, catalog_name),
```
serkef, line 478: I think ci fails because catalog_name is passed as empty string when not exists which is a bit weird.

maybe the caller should take care of it
https://github.com/serkef/sqlmesh/blob/4440bbc4a4cce10f7f52f65865d64d3f5b23d8d1/sqlmesh/core/snapshot/evaluator.py#L786C69-L786C69

tobymao: i've fixed this in solglot

tobymao: https://github.com/tobymao/sqlglot/commit/d5bae81e53220f382d7e3bc15fb0afe63b1a502e

eakmanrq (approved):
Messaged about just doing regex on catalog but don't feel strongly.

tobymao:
> Messaged about just doing regex on catalog but don't feel strongly.

edge case that only catalog regex wouldn't handle, duckdb -> bigquery translation where db or table name includes a hyphen

sqlmesh/core/engine_adapter/base.py
serkef: would we need to make the same change for the other adapter functions as well? (create/delete/merge for schema/table/view etc?)

PR #1361, opened 2023-08-29
Feat: Send query on shift-enter + option to open file in new tab

mykalmax:


web/client/src/library/components/fileExplorer/File.tsx
```diff
@@ -138,6 +156,26 @@ function File({
...
+              onClick={(e: MouseEvent) => {
+                e.stopPropagation()
+              }}
+              onSelect={(e: Event) => {
+                e.stopPropagation()
+
+                openFileInNewTab()
+              }}
+            >
+              Open in new tab
```
vchan, line 176:
```suggestion
              Open in New Tab
```

vchan (approved):

PR #1362, opened 2023-08-30
Chore: Improve Engine Adapter Abstraction for Handling DataFrames

eakmanrq:
The core change is to create a single method (`_df_to_source_query`) that can be overridden by any engine to take a Dataframe and create a query. They just need to return a generator that creates expressions. Since it is a generator they can also define their own setup/teardown logic around each expression. This covers all known implementations of either converting pandas to SQL, loading pandas to temp table and then reading from the table, or create a tempView out of a PySpark Dataframe. 

One known regression is that for BigQuery and Snowflake we now always load to a temp table from a Pandas Dataframe and then copy those contents into the target table when in some cases we could have loaded the target directly. Since Pandas Dataframes are smaller data I doubt this have much actual impact and can be improved in a follow-up PR.

Follow Up:
* Make `columns_to_types` required in `SourceQuery`
* Add support for testing Spark
* Add MSSQL to integration tests

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 import re
+import sys
 import typing as t
 from difflib import unified_diff
 from enum import Enum, auto
@@ -654,36 +655,46 @@ def extend_sqlglot() -> None:
 
 
 def select_from_values(
-    values: t.Iterable[t.Tuple[t.Any, ...]],
+    values: t.List[t.Tuple[t.Any, ...]],
     columns_to_types: t.Dict[str, exp.DataType],
-    batch_size: int = 0,
+    batch_size: int = sys.maxsize,
     alias: str = "t",
-) -> t.Generator[exp.Select, None, None]:
+) -> t.Iterator[exp.Select]:
     """Generate a VALUES expression that has a select wrapped around it to cast the values to their correct types.
 
     Args:
         values: List of values to use for the VALUES expression.
         columns_to_types: Mapping of column names to types to assign to the values.
-        batch_size: The maximum number of tuples per batch, if <= 0 then no batching will occur.
+        batch_size: The maximum number of tuples per batch.
         alias: The alias to assign to the values expression. If not provided then will default to "t"
 
     Returns:
         This method operates as a generator and yields a VALUES expression.
     """
+    num_rows = len(values)
+    for i in range(0, num_rows, batch_size):
+        yield select_from_values_for_batch_range(
+            values=values,
+            columns_to_types=columns_to_types,
+            batch_start=i,
+            batch_end=min(i + batch_size, num_rows),
+            alias=alias,
+        )
+
+
+def select_from_values_for_batch_range(
+    values: t.List[t.Tuple[t.Any, ...]],
+    columns_to_types: t.Dict[str, exp.DataType],
+    batch_start: int,
+    batch_end: int,
+    alias: str = "t",
+) -> exp.Select:
     casted_columns = [
         exp.alias_(exp.cast(column, to=kind), column, copy=False)
         for column, kind in columns_to_types.items()
     ]
-    batch = []
-    for row in values:
-        batch.append(row)
-        if batch_size > 0 and len(batch) >= batch_size:
-            values_exp = exp.values(batch, alias=alias, columns=columns_to_types)
-            yield exp.select(*casted_columns).from_(values_exp)
-            batch.clear()
-    if batch:
-        values_exp = exp.values(batch, alias=alias, columns=columns_to_types)
-        yield exp.select(*casted_columns).from_(values_exp)
+    values_exp = exp.values(values[batch_start:batch_end], alias=alias, columns=columns_to_types)
+    return exp.select(*casted_columns).from_(values_exp, copy=False)
 
 
 def pandas_to_sql(
@@ -704,7 +715,7 @@ def pandas_to_sql(
         This method operates as a generator and yields a VALUES expression.
     """
     yield from select_from_values(
-        values=df.itertuples(index=False, name=None),
+        values=list(df.itertuples(index=False, name=None)),
         columns_to_types=columns_to_types or columns_to_types_from_df(df),
         batch_size=batch_size,
         alias=alias,
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -677,7 +677,7 @@ def select_from_values(
     batch = []
     for row in values:
         batch.append(row)
-        if batch_size > 0 and len(batch) > batch_size:
+        if batch_size > 0 and len(batch) >= batch_size:
             values_exp = exp.values(batch, alias=alias, columns=columns_to_types)
             yield exp.select(*casted_columns).from_(values_exp)
             batch.clear()
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -13,6 +13,7 @@ import itertools
 import logging
 import typing as t
 import uuid
+from dataclasses import dataclass
 from enum import Enum
 
 import pandas as pd
@@ -25,7 +26,7 @@ from sqlmesh.core.dialect import pandas_to_sql
 from sqlmesh.core.engine_adapter.shared import DataObject, TransactionType
 from sqlmesh.core.model.kind import TimeColumn
 from sqlmesh.core.schema_diff import SchemaDiffer
-from sqlmesh.utils import double_escape, optional_import
+from sqlmesh.utils import double_escape
 from sqlmesh.utils.connection_pool import create_connection_pool
 from sqlmesh.utils.date import TimeLike, make_inclusive, to_ts
 from sqlmesh.utils.errors import SQLMeshError
@@ -67,6 +68,23 @@ class InsertOverwriteStrategy(Enum):
     def is_replace_where(self) -> bool:
         return self == InsertOverwriteStrategy.REPLACE_WHERE
 
+    @property
+    def requires_condition(self) -> bool:
+        return self.is_replace_where or self.is_delete_insert
+
+
+@dataclass
+class SourceQuery:
+    generator: t.Iterator[Query]
+    columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+    batched: bool = False
+    is_from_df: bool = False
+
+    @property
+    def single_query(self) -> t.Iterator[Query]:
+        assert not self.batched
+        yield from self.generator
+
 
 class EngineAdapter:
     """Base class wrapping a Database API compliant connection.
@@ -107,36 +125,6 @@ class EngineAdapter:
         self.sql_gen_kwargs = sql_gen_kwargs or {}
         self._extra_config = kwargs
 
-    @classmethod
-    def is_pandas_df(cls, value: t.Any) -> bool:
-        return isinstance(value, pd.DataFrame)
-
-    @classmethod
-    def is_pyspark_df(cls, value: t.Any) -> bool:
-        return hasattr(value, "sparkSession")
-
-    @classmethod
-    def is_df(self, value: t.Any) -> bool:
-        return self.is_pandas_df(value) or self.is_pyspark_df(value)
-
-    @classmethod
-    def try_get_df(cls, value: t.Any) -> t.Optional[DF]:
-        if cls.is_df(value):
-            return value
-        return None
-
-    @classmethod
-    def try_get_pyspark_df(cls, value: t.Any) -> t.Optional[PySparkDataFrame]:
-        if cls.is_pyspark_df(value):
-            return value
-        return None
-
-    @classmethod
-    def try_get_pandas_df(cls, value: t.Any) -> t.Optional[pd.DataFrame]:
-        if cls.is_pandas_df(value):
-            return value
-        return None
-
     @property
     def cursor(self) -> t.Any:
         return self._connection_pool.get_cursor()
@@ -144,6 +132,43 @@ class EngineAdapter:
     @property
     def spark(self) -> t.Optional[PySparkSession]:
         return None
+
+    @classmethod
+    def is_pandas_df(cls, value: t.Any) -> bool:
+        return isinstance(value, pd.DataFrame)
+
+    def _ensure_source_query(
+        self,
+        query_or_df: QueryOrDF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        *,
+        target_table: t.Optional[TableName] = None,
+        batch_size: t.Optional[int] = None,
+    ) -> SourceQuery:
+        if isinstance(query_or_df, SourceQuery):
+            return query_or_df
+        batch_size = self.DEFAULT_BATCH_SIZE if batch_size is None else batch_size
+        if isinstance(query_or_df, (exp.Subqueryable, exp.DerivedTable)):
+            return SourceQuery(generator=iter([query_or_df]), columns_to_types=columns_to_types)
+        else:
+            return self._df_to_source_query(
+                query_or_df, columns_to_types, batch_size, target_table=target_table
+            )
+
+    def _df_to_source_query(
+        self,
+        df: DF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        batch_size: int,
+        target_table: t.Optional[TableName] = None,
+    ) -> SourceQuery:
+        assert isinstance(df, pd.DataFrame)
+        return SourceQuery(
+            generator=self._pandas_to_sql(df, columns_to_types, batch_size),
+            columns_to_types=columns_to_types or columns_to_types_from_df(df),
+            batched=0 < batch_size < len(df.index),
+            is_from_df=True,
+        )
 
     def recycle(self) -> None:
         """Closes all open connections and releases all allocated resources associated with any thread
@@ -173,14 +198,10 @@ class EngineAdapter:
             kwargs: Optional create table properties.
         """
         table = exp.to_table(table_name)
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            return self._create_table_from_df(table, df, columns_to_types, replace=True, **kwargs)
-        else:
-            query_or_df = t.cast("Query", query_or_df)
-            return self._create_table_from_query(
-                table, query_or_df, replace=True, columns_to_types=columns_to_types, **kwargs
-            )
+        source_query = self._ensure_source_query(
+            query_or_df, columns_to_types, target_table=table_name
+        )
+        return self._create_table_from_source_query(table, source_query, replace=True, **kwargs)
 
     def create_index(
         self,
@@ -247,14 +268,8 @@ class EngineAdapter:
             exists: Indicates whether to include the IF NOT EXISTS check.
             kwargs: Optional create table properties.
         """
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            self._create_table_from_df(table_name, df, columns_to_types, exists, **kwargs)
-        else:
-            query_or_df = t.cast("Query", query_or_df)
-            self._create_table_from_query(
-                table_name, query_or_df, exists=exists, columns_to_types=columns_to_types, **kwargs
-            )
+        source_query = self._ensure_source_query(query_or_df, columns_to_types)
+        return self._create_table_from_source_query(table_name, source_query, exists, **kwargs)
 
     def create_state_table(
         self,
@@ -308,42 +323,29 @@ class EngineAdapter:
         )
         self._create_table(schema, None, exists=exists, columns_to_types=columns_to_types, **kwargs)
 
-    def _create_table_from_query(
+    def _create_table_from_source_query(
         self,
         table_name: TableName,
-        query: Query,
+        source_query: SourceQuery,
         exists: bool = True,
         replace: bool = False,
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
         **kwargs: t.Any,
     ) -> None:
         table = exp.to_table(table_name)
-        self._create_table(
-            table,
-            query,
-            exists=exists,
-            replace=replace,
-            columns_to_types=columns_to_types,
-            **kwargs,
-        )
-
-    def _create_table_from_df(
-        self,
-        table_name: TableName,
-        df: DF,
-        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
-        exists: bool = True,
-        replace: bool = False,
-        **kwargs: t.Any,
-    ) -> None:
-        if not isinstance(df, pd.DataFrame):
-            raise ValueError("df must be a pandas DataFrame")
-        columns_to_types = columns_to_types or columns_to_types_from_df(df)
-        with self.transaction():
-            if replace:
-                self.drop_table(table_name)
-            self._create_table_from_columns(table_name, columns_to_types, **kwargs)
-            self._insert_append_pandas_df(table_name, df, columns_to_types)
+        if source_query.batched:
+            with self.transaction():
+                if replace:
+                    self.drop_table(table_name)
+                self._create_table_from_columns(
+                    table_name,
+                    source_query.columns_to_types or self.columns(table_name),
+                    **kwargs,
+                )
+                self._insert_append_source_query(table_name, source_query)
+        else:
+            for query in source_query.single_query:
+                self._create_table(table, query, exists=exists, replace=replace, **kwargs)
 
     def _create_table(
         self,
@@ -470,18 +472,13 @@ class EngineAdapter:
             materialized: Whether to create a a materialized view. Only used for engines that support this feature.
             create_kwargs: Additional kwargs to pass into the Create expression
         """
-        schema: t.Optional[exp.Table | exp.Schema] = exp.to_table(view_name)
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            if columns_to_types is None:
-                columns_to_types = columns_to_types_from_df(df)
-
+        source_query = self._ensure_source_query(query_or_df, columns_to_types, batch_size=0)
+        schema: t.Union[exp.Table, exp.Schema] = exp.to_table(view_name)
+        if source_query.columns_to_types:
             schema = exp.Schema(
-                this=schema,
-                expressions=[exp.column(column) for column in columns_to_types],
+                this=exp.to_table(view_name),
+                expressions=[exp.column(column) for column in source_query.columns_to_types],
             )
-            query_or_df = next(self._pandas_to_sql(df, columns_to_types=columns_to_types))
-
         properties = create_kwargs.pop("properties", None)
         if not properties:
             properties = exp.Properties(expressions=[])
@@ -498,16 +495,16 @@ class EngineAdapter:
 
         if properties.expressions:
             create_kwargs["properties"] = properties
-
-        self.execute(
-            exp.Create(
-                this=schema,
-                kind="VIEW",
-                replace=replace,
-                expression=query_or_df,
-                **create_kwargs,
+        for query in source_query.single_query:
+            self.execute(
+                exp.Create(
+                    this=schema,
+                    kind="VIEW",
+                    replace=replace,
+                    expression=query,
+                    **create_kwargs,
+                )
             )
-        )
 
     def create_schema(
         self,
@@ -581,16 +578,10 @@ class EngineAdapter:
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
         contains_json: bool = False,
     ) -> None:
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            self._insert_append_pandas_df(
-                table_name, df, columns_to_types, contains_json=contains_json
-            )
-        else:
-            query = t.cast("Query", query_or_df)
-            if contains_json:
-                query = self._escape_json(query)
-            self._insert_append_query(table_name, query, columns_to_types)
+        source_query = self._ensure_source_query(
+            query_or_df, columns_to_types, target_table=table_name
+        )
+        self._insert_append_source_query(table_name, source_query, contains_json)
 
     @t.overload
     @classmethod
@@ -618,45 +609,18 @@ class EngineAdapter:
             )
         return value
 
-    def _insert_append_query(
+    def _insert_append_source_query(
         self,
         table_name: TableName,
-        query: Query,
-        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
-    ) -> None:
-        column_names = list(columns_to_types or [])
-        self.execute(exp.insert(query, table_name, columns=column_names))
-
-    def _insert_append_pandas_df(
-        self,
-        table_name: TableName,
-        df: pd.DataFrame,
-        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        source_query: SourceQuery,
         contains_json: bool = False,
     ) -> None:
-        connection = self._connection_pool.get()
-        table = exp.to_table(table_name)
-
-        sqlalchemy = optional_import("sqlalchemy")
-        # pandas to_sql doesn't support insert overwrite, it only supports deleting the table or appending
-        if sqlalchemy and isinstance(connection, sqlalchemy.engine.Connectable):
-            df.to_sql(
-                table.sql(dialect=self.dialect),
-                connection,
-                if_exists="append",
-                index=False,
-                chunksize=self.DEFAULT_BATCH_SIZE,
-                method="multi",
-            )
-        else:
-            column_names = list(columns_to_types or [])
-            with self.transaction():
-                for i, expression in enumerate(
-                    self._pandas_to_sql(
-                        df, columns_to_types, self.DEFAULT_BATCH_SIZE, contains_json=contains_json
-                    )
-                ):
-                    self.execute(exp.insert(expression, table_name, columns=column_names))
+        with self.transaction(condition=source_query.batched):
+            columns_to_types = source_query.columns_to_types or self.columns(table_name)
+            for query in source_query.generator:
+                if contains_json:
+                    query = self._escape_json(query)
+                self.execute(exp.insert(query, table_name, columns=list(columns_to_types)))
 
     def insert_overwrite_by_partition(
         self,
@@ -665,9 +629,10 @@ class EngineAdapter:
         partitioned_by: t.List[exp.Expression],
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
     ) -> None:
-        self._insert_overwrite_by_condition(
-            table_name, query_or_df, columns_to_types=columns_to_types
+        source_query = self._ensure_source_query(
+            query_or_df, columns_to_types, target_table=table_name
         )
+        self._insert_overwrite_by_condition(table_name, source_query)
 
     def insert_overwrite_by_time_partition(
         self,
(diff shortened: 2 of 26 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +68,32 @@ def is_insert_overwrite(self) -> bool:
...
+    columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+    batched: bool = False
+    is_from_df: bool = False
+
+    @property
+    def single_query(self) -> t.Iterator[Query]:
+        assert not self.batched
+        yield from self.generator
+
+    def ensure_columns_to_types(
```
izeigerman, line 88: I don't think I like this method. I'm not sure why is it a part of `SourceQuery` to begin with. I'm also not sure why does mutate it. There are quite a few levels of indirection here, while in actuality callers would've been better off just resolving it in place, ie.:
```
columns_to_types = source_query.columns_to_types or self.columns(table_name)
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +68,32 @@ def is_insert_overwrite(self) -> bool:
...
+
+@dataclass
+class SourceQuery:
+    generator: t.Iterator[Query]
+    columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+    batched: bool = False
+    is_from_df: bool = False
+
+    @property
+    def single_query(self) -> t.Iterator[Query]:
```
izeigerman, line 84: If it's a single query, why yield?

izeigerman: I'm guessing you wanted to make it explicit that this property cannot be re-entered just like `generator`, but it's not obvious at all from either the method name or the signature. I'd rather just cache the query instead and always return it.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +68,32 @@ def is_insert_overwrite(self) -> bool:
...
         return self == InsertOverwriteStrategy.REPLACE_WHERE
 
+    @property
+    def requires_condition(self) -> bool:
+        return self.is_replace_where or self.is_delete_insert
+
+
+@dataclass
+class SourceQuery:
+    generator: t.Iterator[Query]
```
izeigerman, line 78: Are we expecting the batches to not fit the memory? If not, then why not materialize all of the queries? Why bother with generators?  Especially provided that we loose the re-entrance property. Basically once generator / single-query is used once it becomes useless and the `SourceQuery` is left in an invalid state.

izeigerman: Generally, using iterators / generators outside iteration / context manager is a dangerous idea to say the least. Let alone as part of a shared stateful object.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +68,32 @@ def is_insert_overwrite(self) -> bool:
     def is_replace_where(self) -> bool:
         return self == InsertOverwriteStrategy.REPLACE_WHERE
 
+    @property
+    def requires_condition(self) -> bool:
+        return self.is_replace_where or self.is_delete_insert
+
+
+@dataclass
```
izeigerman, line 76: Why not pydantic model?

izeigerman:
I don't like `SourceQuery` 😿

sqlmesh/core/engine_adapter/base.py
tobymao: generators won't be able to fit in memory because they could be large sql queries, agreed that it should be in a generator obj or iterator as opposed to an instance variable like this

sqlmesh/core/engine_adapter/base.py
tobymao: why does it matter if this method can be re-entered or not?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -39,6 +40,7 @@ class MSSQLEngineAdapter(
     """
 
     DIALECT: str = "tsql"
+    FALSE_PREDICATE = exp.condition("1=2")
```
tobymao, line 43: just use exp.false()?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -71,79 +73,36 @@ class for other postgres.
...
-            self._merge(
-                target_table=table_name,
-                source_table=query,
-                on=exp.condition("1=2"),
-                match_expressions=[when_not_matched_by_source, when_not_matched_by_target],
-            )
+            try:
+                yield exp.select(*columns_to_types).from_(temp_table)
+            finally:
+                self.drop_table(temp_table)
```
tobymao, line 96: why do all the callers have to drop their own tables?

sqlmesh/core/engine_adapter/base.py
tobymao: SourceQuery should be a base class, and other adapters should inherit it. right now it's strange because you're instantiating it and passing in a "generator" into each one.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +68,32 @@ def is_insert_overwrite(self) -> bool:
...
+    generator: t.Iterator[Query]
+    columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+    batched: bool = False
+    is_from_df: bool = False
+
+    @property
+    def single_query(self) -> t.Iterator[Query]:
+        assert not self.batched
+        yield from self.generator
+
```
tobymao, line 87: @eakmanrq is there an example in this PR where the generator yields more than one query?

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yeah I originally had that but found that I was putting it in a lot of places so I wanted to capture that logic and also cache the result.

sqlmesh/core/engine_adapter/base.py
eakmanrq: https://github.com/TobikoData/sqlmesh/blob/659a972b207abadc9796f44bdb45f64c1598839d/sqlmesh/core/engine_adapter/base.py#L175

sqlmesh/core/engine_adapter/base.py
eakmanrq: I originally had that but then realized it didn't matter since wasn't going to be serialized. I can switch it back though.

sqlmesh/core/engine_adapter/base.py
eakmanrq: > agreed that it should be in a generator obj or iterator as opposed to an instance variable like this

It is a generator object? I basically had a `t.Tuple[t.Iterator, <a bunch of other stuff>]` before so instead of doing that I created a dataclass (or Pydantic model doesn't matter) for it. 

> SourceQuery should be a base class, and other adapters should inherit it.

That could be done but I'm not currently seeing why this would be much better. Each class today implements this the same way just with different generators based on how they handle yielding the expressions after converting a dataframe.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Some engine adapter methods have explicit logic for if they have X number of queries or a single query. For example creating a view on a pandas dataframe wants to assert that it can't get more than a single query back. If we just returned a query though we would loose the lazy evaluation property of the generator which allows the provider of the generator to say how things should behave pre/post the creation of the query. 

For why does it matter if we re-enter, some engines create temporary tables for dataframes and therefore if we re-enter then we would waste resources creating a redundant set of temporary tables.

sqlmesh/core/engine_adapter/mssql.py
eakmanrq: I messages @deschman and he said that False didn't work with SQL Server when he tried but doing this did so I kept this around.

sqlmesh/core/engine_adapter/mssql.py
eakmanrq: I'm not sure what we mean by caller here. The caller of the generator doesn't drop any tables. They get drop automatically after they fully consume the iterator. So effectively "behind the scenes" the table gets dropped but it is not something the caller needs to worry about.

sqlmesh/core/engine_adapter/base.py
izeigerman: The caching was there previously anyway. We'd fetch columns and then pass it as an argument further  downstream. You essentially traded:
```
columns_to_types = source_query.columns_to_types or self.columns(table_name)
```
for
```
source_query. ensure_columns_to_types(table_name, self.columns)
```
So not sure whether we achieved anything here

sqlmesh/core/engine_adapter/base.py
izeigerman: > why does it matter if this method can be re-entered or not?

This means that now at any given place where the `SourceQuery` is used I have to guess in which state it's currently in and whether its generator is valid or not. Provided that the `SourceQuery` is not properly scoped and can be used in a multitude of implementations downstream, I think its stateful nature adds unnecessary complexity.

sqlmesh/core/engine_adapter/base.py
izeigerman: > generators won't be able to fit in memory because they could be large sql queries, agreed that it should be in a generator obj or iterator as opposed to an instance variable like this

I don't get this. We can fir the dataframe in memory but not SQL queries derived from it?

sqlmesh/core/engine_adapter/base.py
izeigerman: IMHO another alternative is to let `SourceQuery` encapsulate a process that **produces** a generator and expose a method called `run` or `generate` that creates the generator and yields results rather than caching the generator itself:
```
class SourceQuery:
    generator_factory: t.Callable[[], t.Iterator[Query]]
    
    def generate(self) -> t.Iterator[Query]:
        yield from generator_factory()
```
This way the object is no longer stateful and the `generate` method can be safely re-entered. So regardless of where and when I call it it'd always produce the same result. We can of course cache individual items if needed and this will introduce state, but at least it will be well-scoped, encapsulated and safe.

izeigerman: Generally though, if I understood the idea correctly, I don’t understand why we should couple something stateless like query generation with side effects like creation of temporary tables. I’d rather make side effects explicit than present them as generic iterator.

sqlmesh/core/engine_adapter/base.py
eakmanrq: I went ahead and reverted back to how it was (which matches your top example).

sqlmesh/core/engine_adapter/base.py
eakmanrq: >  I don’t understand why we should couple something stateless like query generation with side effects like creation of temporary tables

The query is based on the table generated. So temporary table `temp_123` is created and then the query is `select cola, colb from temp_123`. So they are technically coupled. And then destroying the table is coupled to the query to because we only want to drop the temporary table once it is no longer needed and that is once the query has been executed. This is what the current generator provides out-of-box with a contract on callers that yielding queries can only be done once (or run the risk of wasting resources).

I could try to do something where I create assets based on a query and then it could be:
Generate Query -> Process Query with Pre-hook -> Yield Query to caller -> Process query with Post-hook. This would mean you could yield queries from the generator freely and skip hooks if you just want to inspect the queries themselves. Example code (building on what you have above):
```
class SourceQuery:
    generator_factory: t.Callable[[], t.Iterator[Query]]
    pre_hook: t.Optional[t.Callable[[Query], None]] = None
    post_hook: t.Optional[t.Callable[[Query], None]] = None
    columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None

    def generate(self) -> t.Iterator[Query]:
        yield from self.generator_factory()

    @property
    def run(self) -> t.Iterator[Query]:
        for query in self.generate():
            if self.pre_hook:
                self.pre_hook(query)
            yield query
            if self.post_hook:
                self.post_hook(query)
```

Note that this is capturing exactly what is happening currently. It though makes this pre/post hook behavior more explicit and provides an interface for iterating through all objects without doing that extra processing. Note that this is only helpful when creating the query itself can be done cheaply and it assumes all the pre/post hooks need is a query to do their work.

tobymao: you should just implement the iterator in the class, make it an actual iterator / generator

sqlmesh/core/engine_adapter/base.py
tobymao: 1 millon row data frame fits in memory ez, but you convert that into union all statements, you're in trouble, but maybe that's not what ryan was thinking?

tobymao:
let's discuss this PR tomorrow in a meeting, it will probably be faster

sqlmesh/core/engine_adapter/base.py
eakmanrq: With SQL of that size the issue wouldn't be memory of the machine but rather limits on the engine in terms of how big of a SQL statement it will parse. Currently the batching we do with Pandas when generating SQL queries is because of engine limits and not because of local memory limits.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Talked and this is the follow up this: https://github.com/TobikoData/sqlmesh/pull/1362/commits/e7892aa9d51179c07d7c53133ac8bf318809b39a

sqlmesh/core/engine_adapter/base.py
```diff
@@ -67,6 +70,34 @@ def is_insert_overwrite(self) -> bool:
...
+    def __exit__(
+        self,
+        exc_type: t.Optional[t.Type[BaseException]],
+        exc_val: t.Optional[BaseException],
+        exc_tb: t.Optional[types.TracebackType],
+    ) -> bool:
+        if self.cleanup_func:
+            self.cleanup_func()
+        return True
```
izeigerman, line 99: > If an exception is supplied, and the method wishes to suppress the exception (i.e., prevent it from being propagated), it should return a true value. Otherwise, the exception will be processed normally upon exit from this method.

I don't think we should be suppressing an exception here.

izeigerman: https://docs.python.org/3/reference/datamodel.html#object.__exit__

sqlmesh/core/engine_adapter/base.py
```diff
@@ -303,34 +347,22 @@ def _create_table_from_columns(
...
-            if replace:
-                self.drop_table(table_name)
-            self._create_table_from_columns(table_name, columns_to_types, **kwargs)
-            self._insert_append_pandas_df(table_name, df, columns_to_types)
+        table = exp.to_table(table_name)
+        multiple_queries = len(source_queries) > 1
+        with self.transaction(condition=multiple_queries):
+            with source_queries[0] as query:
+                self._create_table(table, query, exists=exists, replace=replace, **kwargs)
+            if multiple_queries:
```
izeigerman, line 364: This is not bad actually

sqlmesh/core/engine_adapter/base.py
```diff
@@ -425,18 +457,16 @@ def create_view(
...
-        schema: t.Optional[exp.Table | exp.Schema] = exp.to_table(view_name)
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            if columns_to_types is None:
-                columns_to_types = columns_to_types_from_df(df)
-
+        source_queries, columns_to_types = self._to_source_queries(
+            query_or_df, columns_to_types, batch_size=0
+        )
+        assert len(source_queries) == 1
```
izeigerman, line 463: Can we have a more meaningful error message here?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -453,16 +483,17 @@ def create_view(
...
-        self.execute(
-            exp.Create(
-                this=schema,
-                kind="VIEW",
-                replace=replace,
-                expression=query_or_df,
-                **create_kwargs,
-            )
-        )
+        for source_query in source_queries:
```
izeigerman, line 486: If we know that there should only be one query, shouldn't we just extract it and avoid an extra layer of nestedness?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -87,6 +88,48 @@ def _job_params(self) -> t.Dict[str, t.Any]:
...
+            self._get_temp_table(target_table or "pandas"), columns_to_types
+        )
+        temp_table = exp.table_(
+            temp_bq_table.table_id,
+            db=temp_bq_table.dataset_id,
+            catalog=temp_bq_table.project,
+        )
+        return [
+            SourceQuery(
+                query_factory=partial(
```
izeigerman, line 122: Why do we need a partial here if `columns_to_types`, `temp_bq_table` and `temp_table` are already local here. Can't `query_factory` reference them directly?

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yes good point. Will fix.

sqlmesh/core/engine_adapter/bigquery.py
eakmanrq: How would I pass along the kwargs to include when calling the factory?

izeigerman: What I'm saying is that if you define `temp_table`, `temp_bq_table` and `columns_to_types` before you define `query_factory` then you can define `query_factory` without any arguments at all since all of them will be in scope. You can just pass it as is: `query_factory=query_factory`.

sqlmesh/core/dialect.py
```diff
@@ -653,36 +654,46 @@ def extend_sqlglot() -> None:
 
 
 def select_from_values(
-    values: t.Iterable[t.Tuple[t.Any, ...]],
+    values: t.List[t.Tuple[t.Any, ...]],
     columns_to_types: t.Dict[str, exp.DataType],
-    batch_size: int = 0,
+    batch_size: int = sys.maxsize,
     alias: str = "t",
 ) -> t.Generator[exp.Select, None, None]:
```
tobymao, line 661: change this t.Iterator[exp.Select]

sqlmesh/core/dialect.py
```diff
@@ -653,36 +654,46 @@ def extend_sqlglot() -> None:
...
-        batch.append(row)
-        if batch_size > 0 and len(batch) > batch_size:
-            values_exp = exp.values(batch, alias=alias, columns=columns_to_types)
-            yield exp.select(*casted_columns).from_(values_exp)
-            batch.clear()
-    if batch:
-        values_exp = exp.values(batch, alias=alias, columns=columns_to_types)
-        yield exp.select(*casted_columns).from_(values_exp)
+    values_exp = exp.values(values[batch_start:batch_end], alias=alias, columns=columns_to_types)
+    return exp.select(*casted_columns).from_(values_exp)
```
tobymao, line 696: from_(values_exp, copy=False)

sqlmesh/core/dialect.py
```diff
@@ -653,36 +654,46 @@ def extend_sqlglot() -> None:
 
 
 def select_from_values(
-    values: t.Iterable[t.Tuple[t.Any, ...]],
+    values: t.List[t.Tuple[t.Any, ...]],
```
tobymao, line 657: why did you have to change this? why not keep it an iterable?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -303,34 +347,22 @@ def _create_table_from_columns(
...
-        if not isinstance(df, pd.DataFrame):
-            raise ValueError("df must be a pandas DataFrame")
-        columns_to_types = columns_to_types or columns_to_types_from_df(df)
-        with self.transaction():
-            if replace:
-                self.drop_table(table_name)
-            self._create_table_from_columns(table_name, columns_to_types, **kwargs)
-            self._insert_append_pandas_df(table_name, df, columns_to_types)
+        table = exp.to_table(table_name)
+        multiple_queries = len(source_queries) > 1
```
tobymao, line 360: why not just

with self.transaction():
  for i, query in enumerate(source_queries):
        if i == 0:
          self._create_table()
      else: 
          self._insert_append(query)

sqlmesh/core/engine_adapter/base.py
```diff
@@ -575,45 +602,20 @@ def _escape_json(cls, value: Query | str) -> Query | str:
...
-        else:
-            column_names = list(columns_to_types or [])
-            with self.transaction():
-                for i, expression in enumerate(
-                    self._pandas_to_sql(
-                        df, columns_to_types, self.DEFAULT_BATCH_SIZE, contains_json=contains_json
-                    )
-                ):
-                    self.execute(exp.insert(expression, table_name, columns=column_names))
+        with self.transaction(condition=len(source_queries) > 0):
```
tobymao, line 612: just always run the transaction, it's fine, remove this condition stuff

sqlmesh/core/engine_adapter/bigquery.py
tobymao: add kwargs to SourceQuery, and have it call the function with those kwargs on use

tobymao: or, what iarslav was saying is saying it's a lambda, just use closures, and reference them directly

sqlmesh/core/engine_adapter/base.py
```diff
@@ -144,6 +145,51 @@ def cursor(self) -> t.Any:
...
+    def _df_to_source_queries(
+        self,
+        df: DF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        batch_size: int,
+        target_table: t.Optional[TableName] = None,
+    ) -> t.Tuple[t.List[SourceQuery], t.Optional[t.Dict[str, exp.DataType]]]:
+        assert isinstance(df, pd.DataFrame)
+        num_rows = len(df.index)
+        columns_to_types = columns_to_types or columns_to_types_from_df(df)
```
tobymao, line 177: what's going on here? why do you have to return columns_to_types? why is that coupled in this function

sqlmesh/core/engine_adapter/duckdb.py
```diff
@@ -3,29 +3,13 @@
...
 
-    def _insert_append_pandas_df(
-        self,
-        table_name: TableName,
-        df: pd.DataFrame,
-        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
-        contains_json: bool = False,
-    ) -> None:
-        column_names = list(columns_to_types or [])
-        self.execute(exp.insert("SELECT * FROM df", table_name, columns=column_names))
```
tobymao, line 27: how is this implemented for duckdb?

sqlmesh/core/engine_adapter/bigquery.py
eakmanrq: Sorry I still don't get it. Putting it in a lambda or passing kwargs to SourceQuery are just two other ways to do functionally what is done here but less clear since they are doing a partial but in other ways.

sqlmesh/core/dialect.py
eakmanrq: I need to be able to get the length of it

sqlmesh/core/engine_adapter/base.py
eakmanrq: This is functionally the same and currently we don't have an insert/append that takes in a `Query` object.

sqlmesh/core/engine_adapter/base.py
eakmanrq: I would prefer if we only issues transactions if they are needed.

sqlmesh/core/engine_adapter/base.py
eakmanrq: In a follow up PR `_to_source_queries` will be returning a non-optional columns_to_types. So this is building towards that.

To answer why is it coupled, when we get dataframes we can infer the columns_to_types (if not provided) based on the schema of that Dataframe. The only exception right now is PySpark. Since this is where we have the logic to convert a dataframe to SQL, and it often requires knowing columns to type, we should have it also return the columns_to_types that it inferred.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -21,6 +23,55 @@ class SnowflakeEngineAdapter(EngineAdapter):
...
+                self._connection_pool.get(),
+                df,
+                temp_table.name,
+                schema=temp_table.db,
+                chunk_size=self.DEFAULT_BATCH_SIZE,
+            )
+            return exp.select(*columns_to_types).from_(temp_table)
+
+        assert isinstance(df, pd.DataFrame)
+        columns_to_types = columns_to_types or columns_to_types_from_df(df)
```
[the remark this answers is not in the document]
eakmanrq: Yes just out of the scope of this PR. Follow up PRs will look to start making columns_to_type required as a response from source query and all internal engine adapter methods.

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -72,14 +74,47 @@ def replace_query(
...
+                    with self.temp_table(
+                        exp.select(*columns_to_types).from_(target_table),
+                        target_table,
+                        columns_to_types,
+                    ) as temp_table:
+                        for table in query.find_all(exp.Table):
+                            if table == target_table:
+                                table.set("this", temp_table.this)
+                                table.set("db", temp_table.args.get("db"))
+                                table.set("catalog", temp_table.args.get("catalog"))
```
eakmanrq, line 113: @tobymao or @GeorgeSittas Although this works, I'm sure there is a better way to write this. Let me know what suggestions you may have.

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -72,14 +74,47 @@ def replace_query(
...
+            columns_to_types = columns_to_types or self.columns(table_name)
+            if len(source_queries) != 1:
+                raise SQLMeshError(
+                    f"Replace query for {table_name} must have exactly one source query."
+                )
+            with source_queries[0] as query:
+                target_table = exp.to_table(table_name)
+                # Check if self-referencing
+                if any(
+                    [
```
tobymao, line 98: array is unnecassyr

sqlmesh/core/engine_adapter/mixins.py
tobymao: you can replace table with temp_table

tobymao: table.replace(temp_table.copy()) or something

sqlmesh/core/engine_adapter/base.py
```diff
@@ -107,43 +139,102 @@ def __init__(
...
+    def _to_utc_timestamp(cls, col: t.Union[str, exp.Literal, exp.Column, exp.Null]) -> exp.Cast:
+        def ensure_utc_exp(
+            ts: t.Union[str, exp.Literal, exp.Column, exp.Null]
+        ) -> t.Union[exp.Literal, exp.Column, exp.Null]:
+            if not isinstance(ts, (str, exp.Literal)):
+                return ts
+            if isinstance(ts, exp.Literal):
+                if not ts.is_string:
+                    raise SQLMeshError("Timestamp literal must be a string")
+                ts = ts.this
```
tobymao, line 164: ts = ts.name, then i don't think you need a cast

sqlmesh/core/engine_adapter/base.py
```diff
@@ -107,43 +139,102 @@ def __init__(
...
+        query_or_df: QueryOrDF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        target_table: TableName,
+        *,
+        batch_size: t.Optional[int] = None,
+    ) -> t.List[SourceQuery]:
+        batch_size = self.DEFAULT_BATCH_SIZE if batch_size is None else batch_size
+        if isinstance(query_or_df, (exp.Subqueryable, exp.DerivedTable)):
+            return [SourceQuery(query_factory=lambda: query_or_df)]  # type: ignore
+        else:
```
tobymao, line 184: you don't need the else here

sqlmesh/core/engine_adapter/base.py
```diff
@@ -107,43 +139,102 @@ def __init__(
...
-        return None
+    def _df_to_source_queries(
+        self,
+        df: DF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        batch_size: int,
+        target_table: TableName,
+    ) -> t.List[SourceQuery]:
+        assert isinstance(df, pd.DataFrame)
+        assert columns_to_types
```
tobymao, line 197: why are you using asserts here? if it's for typing why is it legal for these to be optional then?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -470,18 +547,27 @@ def create_view(
...
-        schema: t.Optional[exp.Table | exp.Schema] = exp.to_table(view_name)
-        df = self.try_get_pandas_df(query_or_df)
-        if df is not None:
-            if columns_to_types is None:
-                columns_to_types = columns_to_types_from_df(df)
-
+        if self.is_pandas_df(query_or_df):
+            values = list(t.cast(pd.DataFrame, query_or_df).itertuples(index=False, name=None))
+            columns_to_types = columns_to_types or self._columns_to_types(query_or_df)
+            assert columns_to_types
```
tobymao, line 553: raise?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -693,68 +760,68 @@ def insert_overwrite_by_time_partition(
...
+        with self.transaction(
+            condition=len(source_queries) > 0 or self.INSERT_OVERWRITE_STRATEGY.is_delete_insert
+        ):
+            columns_to_types = columns_to_types or self.columns(table_name)
+            for i, source_query in enumerate(source_queries):
+                with source_query as query:
+                    query = self._add_where_to_query(query, where)
+                    if i > 0 or self.INSERT_OVERWRITE_STRATEGY.is_delete_insert:
+                        if i == 0:
+                            assert where is not None
```
tobymao, line 807: raise

sqlmesh/core/engine_adapter/base.py
```diff
@@ -813,128 +884,131 @@ def scd_type_2(
...
+                        CASE 
+                            WHEN t_{valid_from_name} IS NULL 
+                                 AND latest_deleted.{unique_key[0]} IS NOT NULL 
+                            THEN CASE 
+                                    WHEN latest_deleted.{valid_to_name} > s_{updated_at_name} 
+                                    THEN latest_deleted.{valid_to_name} 
+                                    ELSE s_{updated_at_name} 
+                                 END 
+                            WHEN t_{valid_from_name} IS NULL 
+                            THEN {self._to_utc_timestamp('1970-01-01 00:00:00+00:00').sql()}
```
tobymao, line 962: sql calls are unnecessary

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -60,7 +62,7 @@ def table_exists(self, table_name: TableName) -> bool:
         )
         database_name = table.args.get("db")
```
tobymao, line 63: database_name = table.db

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -60,7 +62,7 @@ def table_exists(self, table_name: TableName) -> bool:
...
         if database_name:
-            sql = sql.where(f"table_schema = '{database_name}'")
+            sql = sql.where(f"table_schema = '{database_name.this}'")
```
tobymao, line 65: then this can just be database_name

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -88,6 +87,40 @@ def _job_params(self) -> t.Dict[str, t.Any]:
...
         return params
 
+    def _df_to_source_queries(
+        self,
+        df: DF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        batch_size: int,
+        target_table: TableName,
+    ) -> t.List[SourceQuery]:
+        assert columns_to_types
```
tobymao, line 97: raise

sqlmesh/core/engine_adapter/duckdb.py
```diff
@@ -3,28 +3,41 @@
...
+        batch_size: int,
+        target_table: TableName,
+    ) -> t.List[SourceQuery]:
+        assert columns_to_types
+        temp_table = self._get_temp_table(target_table)
+        casted_columns = [
+            exp.alias_(exp.cast(column, to=kind), column, copy=False)
+            for column, kind in columns_to_types.items()
+        ]
+        temp_table_sql = exp.select(*casted_columns).from_("df").sql(dialect=self.dialect)
```
tobymao, line 33: why do you need a temp table?

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -120,3 +151,45 @@ def execute(
         )
         if not self._connection_pool.is_transaction_active:
             self._connection_pool.commit()
+
+
+class InsertOverwriteWithMergeMixin(EngineAdapter):
+    FALSE_PREDICATE: exp.Condition = exp.FALSE
```
tobymao, line 157: exp.false()

sqlmesh/core/engine_adapter/mssql.py
tobymao: is this still needed? i think latest sqlglot has this conversion automatically

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -77,6 +82,44 @@ def convert_columns_to_types_to_pyspark_schema(
...
+
+        def query_factory() -> Query:
+            temp_table = self._get_temp_table(target_table or "spark", table_only=True)
+            df.createOrReplaceGlobalTempView(temp_table.sql(dialect=self.dialect))  # type: ignore
+            temp_table.set("db", "global_temp")
+            return exp.select("*").from_(temp_table)
+
+        if self._use_spark_session:
+            return [SourceQuery(query_factory=query_factory)]
+        else:
```
tobymao, line 120: else is not necessary

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -252,8 +179,15 @@ def replace_query(
...
+        source_queries, columns_to_types = self._get_source_queries_and_columns_to_types(
+            query_or_df, columns_to_types, target_table=table_name
+        )
+        columns_to_types = columns_to_types or self.columns(table_name)
+        if not columns_to_types:
+            raise SQLMeshError("Cannot replace table without columns to types")
+        self.create_table(table_name, columns_to_types)
         return self._insert_overwrite_by_condition(
-            table_name, query_or_df, columns_to_types=columns_to_types, where=exp.condition("1=1")
+            table_name, source_queries, columns_to_types, where=exp.condition("1=1")
```
tobymao, line 190: where=True

sqlmesh/core/engine_adapter/base.py
eakmanrq: Only in the case of Spark PySparkDataframes can `columns_to_types` be None and that will be addressed in future PR.

sqlmesh/core/engine_adapter/duckdb.py
eakmanrq: If you later execute a query that has `select * from df` then it requires that `df` be defined within that scope. So a workaround for this could be to put `df` in globals but similar logic to other engines it should cheap to just make a temp table since dataframes can only be so big.

sqlmesh/core/engine_adapter/base.py
tobymao: you should use overloading then i think, remember, asserts get removed at runtime

PR #1364, opened 2023-08-30
Feat: Use table clones instead of temp tables in dev environments for engines that support this

izeigerman:


eakmanrq:
Can you include Databricks in this? We assume currently in the Databricks engine is using Delta and therefore supports clones. Make sure to do a shallow clone.

izeigerman:
Good point, @eakmanrq , added

tobymao (approved):
that was easier than i thought?

izeigerman:
@tobymao, yeah, that's why Ryan and I didn't anticipate any complexity here.

eakmanrq:
@izeigerman you are missing the `SHALLOW` part for Databricks

izeigerman:
@eakmanrq fixed databricks support. Please revisit

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -390,12 +390,25 @@ def _create_snapshot(
...
+        evaluation_strategy = _evaluation_strategy(snapshot, self.adapter)
+
         with self.adapter.transaction(TransactionType.DDL), self.adapter.session():
             self.adapter.execute(snapshot.model.render_pre_statements(**render_kwargs))
 
-            _evaluation_strategy(snapshot, self.adapter).create(
-                snapshot.model, table_name, **render_kwargs
-            )
+            if is_dev and snapshot.is_materialized and self.adapter.SUPPORTS_CLONING:
+                tmp_table_name = f"{target_table_name}__schema_migration_source"
```
eakmanrq, line 399: Why does it have a suffix of `schema_migration_source`?

izeigerman: Because we use this table as a source of schema when altering the clone. Would you suggest a different suffix?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -390,12 +390,25 @@ def _create_snapshot(
...
+                source_table_name = snapshot.table_name()
+
+                logger.info(f"Cloning table '{source_table_name}' into '{target_table_name}'")
+
+                evaluation_strategy.create(snapshot.model, tmp_table_name, **render_kwargs)
+                try:
+                    self.adapter.clone_table(target_table_name, snapshot.table_name(), replace=True)
+                    self.adapter.alter_table(target_table_name, tmp_table_name)
```
eakmanrq, line 407: This reads to me like this:
- Create clone from prod version using the final dev version table name
- Alter the dev version table name (so the clone we just made) to have a temp table name
- Drop temp table name

I think I'm not following this right.

eakmanrq: Sorry never mind got it now. Was thinking rename for some reason instead of alter meaning change the schema itself.

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Got it now. Thanks!

eakmanrq (approved):

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -142,3 +144,21 @@ def fetchdf(
...
+    def clone_table(
+        self,
+        target_table_name: TableName,
+        source_table_name: TableName,
+        replace: bool = False,
+        clone_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        clone_kwargs = clone_kwargs or {}
+        clone_kwargs["shallow"] = True
```
crericha, line 157: Do we want to enforce `shallow` or only set to `True` if not in kwargs?

PR #1370, opened 2023-08-31
Docs: add extras to installation docs page

treysp:


eakmanrq (approved):
Nice addition! Small (non-functional) nit of formatting the table (Pycharm does it automatically so I just pushed the change).

tobymao (approved):

PR #1372, opened 2023-09-01
Fix!: Support snapshots of added models with forward-only parents

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -732,16 +731,21 @@ def _build_snapshots_and_dag(
         )
 
     def _is_forward_only_model(self, model_name: str) -> bool:
+        is_added_model = model_name in self.context_diff.added
```
tobymao, line 734: why do you need is_added check? aren't all forward only forward only now?

tobymao (commented):
z

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -407,6 +407,19 @@ def _create_snapshot(
         evaluation_strategy = _evaluation_strategy(snapshot, self.adapter)
 
         with self.adapter.transaction(TransactionType.DDL), self.adapter.session():
+            if is_dev and not snapshot.previous_versions:
```
tobymao, line 410: why do you need this section here, shouldn't all snapshots create a table?

tobymao: is it because this snapshot is forward only and we don't create tables on forward only snapshots?

PR #1386, opened 2023-09-05
Fix!: Make interval unit a part of the data hash

izeigerman:
As it turns out the interval unit may impact a partitioning scheme of the created table and hence should be a part of the data hash and not the metadata hash.


tobymao:
need a migratio either through sqlglot or migration file

tobymao (approved):

izeigerman:
I think the sqlglot version has been bumped here https://github.com/TobikoData/sqlmesh/commit/9fdf28cc083280089b60b06931f38f32ad21fa6f

PR #1389, opened 2023-09-06
Fix: Automatic derivation of a partitioning transformation for a time column in BigQuery

izeigerman:


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -506,14 +507,20 @@ def _create_table_properties(
 
             this = partitioned_by[0]
 
-            if isinstance(this, exp.Column):
-                if partition_interval_unit == IntervalUnit.MINUTE:
-                    raise SQLMeshError("BigQuery does not support partitioning by minute")
+            if isinstance(this, exp.Column) and partition_interval_unit != IntervalUnit.MINUTE:
```
crericha, line 510: For the case when partition_by is a column and we get an interval unit of a minute, is the thought to use the BQ error message instead of our own?

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -291,36 +291,70 @@ def test_create_table_date_partition(
...
     execute_mock = mocker.patch(
         "sqlmesh.core.engine_adapter.bigquery.BigQueryEngineAdapter.execute"
     )
     adapter.create_table(
         "test_table",
-        {"a": "int", "b": "int"},
+        {
+            "a": exp.DataType.build("int"),
+            "b": exp.DataType.build("int"),
+            "ds": partition_column_sql_type,
```
crericha, line 349: Can we add test cases where "ds" is not in columns_to_types?

sqlmesh/core/engine_adapter/bigquery.py
izeigerman: Yes, but the error message will only happen if the column is not date and the custom expression was not provided.

tests/core/engine_adapter/test_bigquery.py
izeigerman: Added

sqlmesh/core/engine_adapter/bigquery.py
crericha: If the data type is timestamp or datetime, only providing the column will fail. Does it make sense to provide a better error message to the user of what to do? If they are relying on our "magic", it would be helpful to point them at documentation that informs them what to do.

izeigerman: I'd rather not hide BQ responses for now. If we can derive the partition automatically - great, if not - we should just let user interpret the error and provide a correct `partition_by` based on the error message they got from BQ.

crericha (approved):

PR #1390, opened 2023-09-07
Feat: extend apiconsole

mykalmax:
The goal here is to emit events to UI on every state change for all stages during plan evaluation. We need to track when stage started when next change happed and when it ends. As well as explicitly set when model evaluation started and when it's finished. With all these we can show better statuses and updates on UI during each stage.

https://github.com/TobikoData/sqlmesh/assets/5644753/af25c376-2429-45f3-8243-dfea3fd2f334


web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+
+class ReportBackfillProgress(BaseModel):
+    completed: int
+    total: int
+    view_name: str
+    start: TimeLike
+    end: t.Optional[TimeLike] = None
+
+
+class ReportTestsRusult(BaseModel):
```
vchan, line 385:
```suggestion
class ReportTestsResult(BaseModel):
```

web/server/api/endpoints/commands.py
```diff
@@ -13,6 +13,7 @@
 from sqlmesh.core.test import ModelTest
 from sqlmesh.utils.errors import PlanError
 from web.server import models
+from web.server.console import ApiConsole
```
vchan, line 16: Use api_console.

web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+    status: t.Optional[ReportStatus] = None
+    start_at: t.Optional[int] = None
+    stop_at: t.Optional[int] = None
+    duration: t.Optional[int] = None
+    done: bool = False
+
+    def __init__(self, *arg: t.Any, **kargs: t.Any) -> None:
+        super().__init__(*arg, **kargs)
+        self.status = ReportStatus.init
+        self.start_at = now_timestamp()
```
vchan, line 413: For Pydantic models, I often see people set default values for status and start_at instead of overriding the constructor.

web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+    errors: int
+    successful: int
+    dialect: str
+    details: t.List[ReportTestDetails]
+    traceback: str
+
+
+class ReportMeta(BaseModel):
+    status: t.Optional[ReportStatus] = None
+    start_at: t.Optional[int] = None
```
vchan, line 405:
```suggestion
    status: ReportStatus = ReportStatus.init
    start_at: int = Field(default_factory=now_timestamp)
```

web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+        self.status = ReportStatus.init
+        self.start_at = now_timestamp()
+
+
+class Stage(BaseModel):
+    meta: ReportMeta = ReportMeta()
+
+    def __init__(self, *arg: t.Any, **kargs: t.Any) -> None:
+        super().__init__(*arg, **kargs)
+        self.meta = ReportMeta()
```
vchan, line 421: Do you still need this when `meta` has a default value? This will also overwrite what the user passed in for `meta`.

web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+    done: bool = False
+
+    def __init__(self, *arg: t.Any, **kargs: t.Any) -> None:
+        super().__init__(*arg, **kargs)
+        self.status = ReportStatus.init
+        self.start_at = now_timestamp()
+
+
+class Stage(BaseModel):
+    meta: ReportMeta = ReportMeta()
```
vchan, line 417: Not sure if this will work for classes.
```suggestion
    meta: ReportMeta = Field(default_factory=ReportMeta)
```

web/server/models.py
```diff
@@ -362,3 +372,138 @@ class ArtifactChange(BaseModel):
...
+    dialect: str
+    details: t.List[ReportTestDetails]
+    traceback: str
+
+
+class ReportMeta(BaseModel):
+    status: t.Optional[ReportStatus] = None
+    start_at: t.Optional[int] = None
+    stop_at: t.Optional[int] = None
+    duration: t.Optional[int] = None
```
vchan, line 407: This seems suitable for this field, but I'm not sure if Pydantic v1 supports it: https://docs.pydantic.dev/latest/usage/computed_fields/

web/server/api/endpoints/commands.py
```diff
@@ -59,9 +64,15 @@ async def apply(
     request.app.state.task = plan_task = asyncio.create_task(run_in_executor(plan_func))
     try:
         plan = await plan_task
-    except PlanError as e:
+        report_stage_validate.stop(success=True)
+        report_plan.stop(success=True)
+        api_console.log_event(event=models.ConsoleEvent.report_plan, data=report_plan.dict())
```
vchan, line 69: Move this to a finally block and also remove it from the except block below.

web/server/api/endpoints/plan.py
```diff
@@ -7,6 +7,7 @@
 from sqlmesh.core.context import Context
 from sqlmesh.utils.date import make_inclusive, to_ds
 from web.server import models
+from web.server.console import ApiConsole
```
vchan, line 10: Use api_console.
```suggestion
from web.server.console import api_console
```

web/server/models.py
```diff
@@ -362,3 +372,136 @@ class ArtifactChange(BaseModel):
...
+
+class ReportStagePlanApplyPromote(ReportStage):
+    total_tasks: int
+    num_tasks: int
+    target_environment: str
+
+
+class Report(Stage):
+    environment: str
+    options: t.Optional[t.Dict[str, t.Any]] = None
```
vchan, line 491: Is it possible to enumerate all the possible options in something like a ReportOptions Pydantic model?

web/server/models.py
```diff
@@ -362,3 +372,136 @@ class ArtifactChange(BaseModel):
...
+    def duration(self) -> int | None:
+        return self.stop_at - self.start_at if self.start_at and self.stop_at else None
+
+    def dict(self, *args: t.Any, **kwargs: t.Any) -> t.Dict[str, t.Any]:
+        data = super().dict(*args, **kwargs)
+        data["duration"] = self.duration
+        return data
+
+
+class Stage(BaseModel):
```
vchan, line 419: It's difficult for me to keep track of these models, especially when they're so similarly named. For example, `Stage`, `Report` that is a subclass of Stage, and `ReportStage` that is also a subclass of Stage. Maybe more descriptive names or add some docstrings.

web/server/models.py
```diff
@@ -362,3 +372,136 @@ class ArtifactChange(BaseModel):
...
+    successful: int
+    dialect: str
+    details: t.List[ReportTestDetails]
+    traceback: str
+
+
+class ReportMeta(BaseModel):
+    status: ReportStatus = ReportStatus.init
+    start_at: int = Field(default_factory=now_timestamp)
+    stop_at: t.Optional[int] = None
```
vchan, line 406: Do you want to use end_at? We use "start" and "end" elsewhere in the code base.

web/server/models.py
```diff
@@ -362,3 +374,145 @@ class ArtifactChange(BaseModel):
     path: str
     type: t.Optional[ArtifactType] = None
     file: t.Optional[File] = None
+
+
+class ReportBackfillProgress(BaseModel):
```
vchan, line 379: Can this and `BackfillTask` be combined? They have the same fields.

web/server/models.py
```diff
@@ -39,6 +39,39 @@ class ApplyType(str, enum.Enum):
...
+    backfills = "backfills"
+
+
+class PlanApplyStage(str, enum.Enum):
+    """An enumeration of plan apply stages."""
+
+    creation = "creation"
+    restate = "restate"
+    backfill = "backfill"
+    promote = "promote"
```
vchan, line 72: Nit: I think plans are applied, but plans are not run/ran. Run is for running the DAG through the scheduler.

web/server/models.py
```diff
@@ -362,3 +374,145 @@ class ArtifactChange(BaseModel):
...
+    def duration(self) -> int | None:
+        return self.end - self.start if self.start and self.end else None
+
+    def dict(self, *args: t.Any, **kwargs: t.Any) -> t.Dict[str, t.Any]:
+        data = super().dict(*args, **kwargs)
+        data["duration"] = self.duration
+        return data
+
+
+class TrackableMixin(BaseModel):
```
vchan, line 438: If this is a mixin and not meant to be instantiated, it doesn't have to subclass BaseModel. It'd be used like
```python
PlanStageTracker(TrackableMixin, BaseModel):
    pass
```

web/server/console.py
```diff
@@ -5,132 +5,215 @@
...
-                event="errors",
-                data=ApiException(
-                    message="Tasks failed to a run",
-                    origin="API -> console -> log_exception",
-                ).to_dict(),
-            )
+        self.log_event(
+            event="errors",
+            data=ApiException(
+                message="Tasks failed to a run",
```
vchan, line 214: Should this be?
```suggestion
                message="Tasks failed to run",
```

vchan (approved):

web/server/api/endpoints/plan.py
```diff
@@ -33,6 +34,12 @@ async def run_plan(
             origin="API -> plan -> run_plan",
         )
 
+    tracker = models.PlanApplyStageTracker(
+        environment=environment, options={"skip_tests": plan_options.skip_tests}
```
vchan, line 38: Is this right? Should options be plan_options instead? Something like `plan_options=models.PlanOptions(skip_tests=plan_options.skip_tests` or even `plan_options=plan_options`?

web/server/api/endpoints/plan.py
```diff
@@ -24,7 +25,7 @@ async def run_plan(
     environment: t.Optional[str] = Body(None),
     plan_dates: t.Optional[models.PlanDates] = None,
     plan_options: models.PlanOptions = models.PlanOptions(),
```
vchan, line 27: It's outside the scope of your changes, but I just noticed it. Can you please add a line like `plan_options = plan_options or models.PlanOptions()` below?
```suggestion
    plan_options: models.PlanOptions = None,
```

web/server/models.py
```diff
@@ -362,3 +369,120 @@ class ArtifactChange(BaseModel):
...
+
+        self.meta.end = now_timestamp()
+        self.meta.done = bool(self.meta.start and self.meta.end)
+
+    def update(self, data: t.Dict[str, t.Any]) -> None:
+        for k, v in data.items():
+            setattr(self, k, v)
+
+
+class PlanApplyStageValidation(Trackable):
```
vchan, line 440: What's the difference between PlanApplyStageValidation.start and PlanApplyStageValidation.meta.start?

web/server/api/endpoints/plan.py
mykalmax: yeah, i reverted my previous change and it removed `plan_options=plan_options`

PR #1392, opened 2023-09-07
Fix: ensure external macros are resolved correctly

GeorgeSittas:
Let's say we have a third-party package `foo`, which contains the following macro definition:

```python
# foo/test_macro.py
from sqlmesh.core.macros import macro

@macro()
def test_macro(evaluator):
    return "test_macro"
```

There's currently a bug where if one tries to expose this macro to their models by importing it in `macros/__init__.py`, it's not correctly resolved and will ultimately result in the following error @ runtime:

```python
SQLMeshError: Macro 'test_macro' does not exist
```

This PR aims to solve this issue by handling `Executable`s that are `IMPORT`s in `MacroEvaluator`'s constructor. Before, we'd just ignore them and nothing would be added to the evaluator's `macros` attribute, whereas now we map the macro's name to the corresponding function object after verifying that the import indeed corresponds to a SQLMesh macro.

TODO:
- [x] Add tests
- [ ] Check airflow works as expected

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -105,10 +105,14 @@ class MacroEvaluator:
         self.python_env = python_env or {}
         self._jinja_env: t.Optional[Environment] = jinja_env
         self.macros = {normalize_macro_name(k): v.func for k, v in macro.get_registry().items()}
+
         prepare_env(self.python_env, self.env)
         for k, v in self.python_env.items():
             if v.is_definition:
                 self.macros[normalize_macro_name(k)] = self.env[v.name or k]
+            elif v.is_import and getattr(self.env.get(k), "__wrapped__", None):
+                # External macros that are serialized as imports will have __wrapped__ set
+                self.macros[normalize_macro_name(k)] = self.env[k]
 
     def send(
         self, name: str, *args: t.Any
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -110,7 +110,7 @@ class MacroEvaluator:
         for k, v in self.python_env.items():
             if v.is_definition:
                 self.macros[normalize_macro_name(k)] = self.env[v.name or k]
-            elif v.is_import and getattr(self.env.get(k), "sqlmesh_macro", None):
+            elif v.is_import and getattr(self.env.get(k), "__sqlmesh_macro__", None):
                 self.macros[normalize_macro_name(k)] = self.env[k]
 
     def send(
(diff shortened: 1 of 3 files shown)


sqlmesh/core/macros.py
```diff
@@ -105,10 +105,14 @@ def __init__(
         self.python_env = python_env or {}
         self._jinja_env: t.Optional[Environment] = jinja_env
         self.macros = {normalize_macro_name(k): v.func for k, v in macro.get_registry().items()}
+
         prepare_env(self.python_env, self.env)
         for k, v in self.python_env.items():
             if v.is_definition:
                 self.macros[normalize_macro_name(k)] = self.env[v.name or k]
+            elif v.is_import and getattr(self.env.get(k), "__wrapped__", None):
```
tobymao, line 113: this is true for any function that has a decorator on it, this doesn't seem to be the right way to detect that it's a macro

GeorgeSittas: Refactored the way we detect macros @ runtime

sqlmesh/utils/metaprogramming.py
```diff
@@ -278,7 +278,8 @@ def walk(obj: t.Any) -> None:
     if name not in env:
         # We only need to add the undecorated code of @macro() functions in env, which
         # is accessible through the `__wrapped__` attribute added by functools.wraps
-        env[name] = getattr(obj, "__wrapped__", obj)
+        env[name] = obj.__wrapped__ if getattr(obj, "is_macro", None) else obj
```
GeorgeSittas, line 281: I think this section is more robust this way, because whilst `walk`ing a callable we may encounter a user function that's decorated with `wraps` and isn't a `macro`, so with previous approach we'd "drop" the decorator by only adding the original function into the `env` (due to setting `env[name]` to its `__wrapped__` attribute if present).

sqlmesh/core/macros.py
```diff
@@ -268,6 +271,15 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
 
     registry_name = "macros"
 
+    def __call__(
+        self, func: t.Callable[..., DECORATOR_RETURN_TYPE]
+    ) -> t.Callable[..., DECORATOR_RETURN_TYPE]:
+        wrapper = super().__call__(func)
+
+        # This is useful to identify macros at runtime
+        setattr(wrapper, "is_macro", True)
```
tobymao, line 280: __sqlmesh_macro__

tobymao (approved):

PR #1396, opened 2023-09-08
Feat: Add standalone audits to sqlmesh format

crericha:
Also allow configuration of sqlmesh specific fields (start, cron, stamp, etc.) for dbt singular tests

### DIFF
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import pathlib
 import sys
 import typing as t
+from itertools import zip_longest
 from pathlib import Path
 
 from pydantic import Field
@@ -11,14 +12,15 @@ from sqlglot.optimizer.qualify_columns import quote_identifiers
 
 from sqlmesh.core import constants as c
 from sqlmesh.core import dialect as d
+from sqlmesh.core.macros import MacroRegistry, macro
 from sqlmesh.core.model.common import bool_validator, expression_validator
-from sqlmesh.core.model.definition import _Model
+from sqlmesh.core.model.definition import _Model, _python_env, _single_value_or_tuple
 from sqlmesh.core.node import _Node
 from sqlmesh.core.renderer import QueryRenderer
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.errors import AuditConfigError, SQLMeshError, raise_config_error
 from sqlmesh.utils.hashing import hash_data
-from sqlmesh.utils.jinja import JinjaMacroRegistry
+from sqlmesh.utils.jinja import JinjaMacroRegistry, extract_macro_references
 from sqlmesh.utils.metaprogramming import Executable
 from sqlmesh.utils.pydantic import (
     PydanticModel,
@@ -29,6 +31,7 @@ from sqlmesh.utils.pydantic import (
 
 if t.TYPE_CHECKING:
     from sqlmesh.core.snapshot import Node, Snapshot
+    from sqlmesh.utils.jinja import MacroReference
 
 if sys.version_info >= (3, 9):
     from typing import Literal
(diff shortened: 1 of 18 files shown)


sqlmesh/core/audit/definition.py
```diff
@@ -432,24 +358,109 @@ def text_diff(self, other: Node) -> str:
                 f"Cannot diff audit '{self.name} against a non-audit node '{other.name}'"
             )
 
-        return d.text_diff(self.query, other.query, self.dialect)
+        meta_a, *statements_a = self.render_definition()
+        meta_b, *statements_b = other.render_definition()
+
+        breakpoint()
```
izeigerman, line 364: If the build didn't fail here, then our test coverage is lacking

sqlmesh/core/audit/definition.py
```diff
@@ -432,24 +358,109 @@ def text_diff(self, other: Node) -> str:
...
             self.query,
             self.dialect,
             self.macro_definitions,
             path=self._path or Path(),
             jinja_macro_registry=self.jinja_macros,
-            python_env=audit.python_env,
+            python_env=self.python_env,
         )
 
+    def _jinja_macros_to_exprs(self) -> t.List[exp.Expression]:
```
izeigerman, line 442: This looks like mostly a copy paste from `model/definition.py`. Seems like a good candidate from the utility function in `jinja.py` which takes `JinjaMacroRegistry` and returns a list of expressions.

sqlmesh/core/audit/definition.py
```diff
@@ -29,39 +31,49 @@
...
+    from sqlmesh.utils.jinja import MacroReference
 
 if sys.version_info >= (3, 9):
     from typing import Literal
 else:
     from typing_extensions import Literal
 
 
-class AuditMixin:
+class AuditMeta:
```
izeigerman, line 42: We should probably call this `Mixin` as well. For example `ModelMeta` assumes that we can parse the model definition into it before we even get to complete `Model`. This is not the case here.

sqlmesh/dbt/test.py
```diff
@@ -133,6 +147,8 @@ def to_sqlmesh(self, context: DbtContext) -> Audit:
                     {source.sql_name for source in test_context.sources.values()}
                 ),
                 tags=self.tags,
+                hash_raw_query=True,
+                **self.sqlmesh_config_kwargs,
```
izeigerman, line 151: Why do we pass these here?

sqlmesh/dbt/test.py
```diff
@@ -59,7 +65,11 @@ class TestConfig(GeneralConfig):
     name: str
     sql: SqlStr
     test_kwargs: t.Dict[str, t.Any] = {}
+    model_name: t.Optional[str] = None
```
izeigerman, line 68: Is this a part of the dbt definition?

sqlmesh/dbt/test.py
crericha: Is the question why here and not for audit? These are fields only the standalone audit has.

- `hash_raw_query` is used for the metadata hash.
- `sqlmesh_config_kwargs` are the configurable fields for sqlmesh `node_` concrete classes.

sqlmesh/core/audit/definition.py
crericha: Added test

sqlmesh/core/audit/definition.py
crericha: I couldn't think of a reason not to add directly to JinjaMacroRegistry, so that's what I did.

sqlmesh/dbt/test.py
izeigerman: Sorry, the diff UI confused me. I read this as if we were passing this into `add_globals(`

sqlmesh/utils/jinja.py
```diff
@@ -330,6 +344,28 @@ def merge(self, other: JinjaMacroRegistry) -> JinjaMacroRegistry:
             top_level_packages=[*self.top_level_packages, *other.top_level_packages],
         )
 
+    def macros_to_expressions(self) -> t.List[Expression]:
```
izeigerman, line 347: [Nit] It's not just macros so I think `to_expressions` would suffice.

izeigerman (approved):

PR #1399, opened 2023-09-08
Fix: Github CI/CD bot and merge conflicts

eakmanrq:
The CI/CD bot uses merge commits in order to both create it's PR environments and to do the deployments. These merge commits are a representation of both of what is currently in the PR and what is in main. This doesn't work though if there is currently a merge conflict between the branches. In order for sometime to hit this bug they would need to issue a `/deploy` command or provide a required approval while having a branch that has a merge conflict.

This change makes it so that we will detect a merge conflict and raise when attempting to do a deployment. 

I also put in a fix where users would see output from prior steps in their plan previews. This was because this output was captured and not cleared before generating the preview.

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -11,6 +11,7 @@ import unittest
 from enum import Enum
 from typing import List
 
+import requests
 from hyperscript import Element, h
 from sqlglot.helper import seq_get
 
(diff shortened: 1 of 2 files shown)


sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -360,6 +410,35 @@ def _get_or_create_comment(self, header: str = "**SQLMesh Bot Info**") -> IssueC
...
+    def _get_merge_state_status(self) -> MergeStateStatus:
+        """
+        This feature is currently in preview and therefore not available in the python module.
+        So we query GraphQL directly instead.
+        """
+        headers = {
+            "Authorization": f"Bearer {self._token}",
+            "Accept": "application/vnd.github.merge-info-preview+json",
+        }
+        query = f"""{{
```
tobymao, line 422: graphql?

tobymao (approved):

sqlmesh/integrations/github/cicd/controller.py
eakmanrq: Yeah this is GraphQL

PR #1401, opened 2023-09-08
Docs: add model configuration reference page

treysp:


izeigerman:
Why is this in the `configuration` reference? These fields are part of the model definition which is IMHO separate from configuration.

izeigerman (approved):

PR #1403, opened 2023-09-08
Fix: starting docs and more tests for metrics

tobymao:


sqlmesh/core/reference.py
```diff
@@ -53,18 +53,66 @@ def name(self) -> str:
...
+
+        Args:
+            model: the model to add.
+        """
+        for column in model.columns_to_types or {}:
+            self._dim_models.setdefault(column, set())
+            self._dim_models[column].add(model.name)
+
         for ref in model.all_references:
+            self._model_refs.setdefault(model.name, {})
```
izeigerman, line 74: Just curious why prefer this syntax over
```
self._model_refs[model.name] = {}
```
provided we don't use the return valuie.

izeigerman: ah, it will only be set if the key is not there. Never mind

sqlmesh/core/reference.py
```diff
@@ -53,18 +53,66 @@ def name(self) -> str:
...
+        """
+        models = []
+
+        for model in self._dim_models[column]:
+            try:
+                if model != source:
+                    self.find_path(source, model)
+                models.append(model)
+            except SQLMeshError:
+                pass
```
izeigerman, line 98: why ignore the error?

docs/concepts/metrics/definition.md
```diff
@@ -0,0 +1,80 @@
...
+  name clicks_per_search,
+  expression total_clicks / total_searches
+);
+```
+
+## Properties
+### name
+- The name of the metric. This name is case insensitive and must be unique in a project.
+### expression
+- A SQL expression consisting of an aggregation or a formula consisting of other metrics.
```
izeigerman, line 74: I believe it can be a combo of the 2 too, right?

sqlmesh/core/reference.py
```diff
@@ -53,18 +53,66 @@ def name(self) -> str:
 
 class ReferenceGraph:
     def __init__(self, models: t.Iterable[Model]):
-        self._model_refs: t.DefaultDict[str, t.Dict[str, Reference]] = defaultdict(dict)
-        self._ref_models: t.DefaultDict[str, t.Dict[str, str]] = defaultdict(dict)
+        self._model_refs: t.Dict[str, t.Dict[str, Reference]] = {}
+        self._ref_models: t.Dict[str, t.Dict[str, str]] = {}
```
izeigerman, line 57: What's the point of setting both a key and a value of the nested dict to the same value (model name)? Is this because of dict's ordering guarantees?

izeigerman: In other words, shouldn't this be similar to `_dim_models`?

izeigerman (approved):

sqlmesh/core/reference.py
tobymao: hm yea looks like oversight

sqlmesh/core/reference.py
tobymao: not every model can join to every other

PR #1408, opened 2023-09-11
Fix: Add check for key in migration 0023

z3z1ma:
```
  File "/.../sqlmesh/core/state_sync/engine_adapter.py", line 655, in migrate
    migration.migrate(self)
  File "/.../sqlmesh/migrations/v0023_fix_added_models_with_forward_only_parents.py", line 39, in migrate
    parsed_snapshot = snapshot_mapping[snapshot_id]
KeyError: ('analytics.fct_sales__account_module_unit', '2775306648')
```

Its possible for a migration to fail here. Added a check.

### DIFF
--- a/sqlmesh/migrations/v0023_fix_added_models_with_forward_only_parents.py
+++ b/sqlmesh/migrations/v0023_fix_added_models_with_forward_only_parents.py
@@ -36,6 +36,8 @@ def migrate(state_sync: t.Any) -> None:
     snapshots_to_delete = set()
 
     for snapshot_id in dag:
+        if snapshot_id not in snapshot_mapping:
+            continue
         parsed_snapshot = snapshot_mapping[snapshot_id]
         is_breaking = parsed_snapshot.get("change_category") == 1
         has_previous_versions = bool(parsed_snapshot.get("previous_versions", []))


izeigerman:
@z3z1ma that shouldn't really happen though. There should be no scenario when a snapshot is in the dag, but not in the mapping. Unless there was some race condition between migration and janitor. Could it be the case?

z3z1ma:
I agree following the flow its hard for me to identify anything obvious. 
I see 2 options. Either cache related or race condition. I can verify both that this error occurred and that across multiple migration attempts on two different machines, it failed the same way. Both machines had the `.cache` though. 

This changed allowed to to complete migration but likely there is some nuance upstream of this @izeigerman 

Perhaps we should look back to a `sqlmesh state reconcile` or something which refreshes cache and ensures there are no unrenderable snapshots or anything weird that need to be purged. 🤷 Just a thought as user base expands and state correctness is critical yet irreconcilable without jumping into db if impacted.

tobymao:
should we just merge this since it happened to alex?

izeigerman:
@tobymao I think it can't hurt for this particular migration. I'm just concerned that we're ignoring a larger issue here.

tobymao:
i agree, but solving that is orthogonal

PR #1411, opened 2023-09-12
Feat: Spark Schema to SQLGlot Schema Support

eakmanrq:
Currently if a user provides a PySpark DataFrame then we don't infer the `columns_to_types` from that DataFrame and instead rely on them to provide it (or try to process without it). This change makes it so that we will determine the SQLGlot schema from the provided PySpark DataFrame schema. This means that we now know the `columns_to_types` for both Pandas and PySpark DataFrames and therefore we can assume in code where we are explicitly handling DataFrames that we can get the `columns_to_types`. 

Also added the ability to convert a SQLGlot schema to PySpark schema. This is used when creating a DataFrame from Pandas and we want to enforce the provided schema is applied to the resulting PySpark DataFrame.

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import typing as t
+from functools import partial
 
 import pandas as pd
 from sqlglot import exp
@@ -16,7 +17,7 @@ from sqlmesh.core.engine_adapter.shared import (
     DataObjectType,
     TransactionType,
 )
-from sqlmesh.utils import nullsafe_join
+from sqlmesh.utils import classproperty, nullsafe_join
 from sqlmesh.utils.errors import SQLMeshError
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,137 @@ def _use_spark_session(self) -> bool:
...
+            exp.DataType.Type.TINYINT: spark_types.ByteType,
+            exp.DataType.Type.SMALLINT: spark_types.ShortType,
             exp.DataType.Type.INT: spark_types.IntegerType,
             exp.DataType.Type.BIGINT: spark_types.LongType,
             exp.DataType.Type.FLOAT: spark_types.FloatType,
             exp.DataType.Type.DOUBLE: spark_types.DoubleType,
             exp.DataType.Type.DECIMAL: spark_types.DecimalType,
+            exp.DataType.Type.TEXT: spark_types.StringType,
+            # SQLGlot currently converts these to Strings
+            # exp.DataType.Type.VARCHAR: spark_types.VarcharType,
```
tobymao, line 86: don't you still need these mappings?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,137 @@ def _use_spark_session(self) -> bool:
...
+        }
+
+        SPARK_PRIMITIVE_MAPPING = {v: k for k, v in PRIMITIVE_MAPPING.items()}
+        SPARK_COMPLEX_TYPE_MAPPING = {v: k for k, v in COMPLEX_TYPE_MAPPING.items()}
+
+        def sqlglot_complex_to_spark_complex(complex_type: exp.DataType) -> spark_types.DataType:
+            is_struct = complex_type.is_type(exp.DataType.Type.STRUCT)
+            expressions = []
+            for column_def in complex_type.expressions:
+                col_name = column_def.this.name if is_struct else None
```
tobymao, line 110: what if it's a struct of a struct?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,137 @@ def _use_spark_session(self) -> bool:
...
+                    primitive_func
+                    if primitive_func
+                    else partial(sqlglot_complex_to_spark_complex, data_type)
+                )
+                if is_struct:
+                    expressions.append(spark_types.StructField(col_name, type_func()))
+                else:
+                    expressions.append(type_func())
+            klass = COMPLEX_TYPE_MAPPING[complex_type.this]
+            return klass(*([expressions] if is_struct else expressions))
```
tobymao, line 123: what's going on here? why do you have a nested array? can you add kwargs? klass(expressions=) ?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,137 @@ def _use_spark_session(self) -> bool:
...
+                if col_name:
+                    return exp.ColumnDef(this=exp.to_identifier(col_name), kind=kind)
+                return kind
+
+            expressions = []
+            fields = get_fields(complex_type)
+            for field in fields:
+                col_name = field.name if hasattr(field, "name") else None
+                spark_data_type = field.dataType if hasattr(field, "dataType") else field
+                primitive = SPARK_PRIMITIVE_MAPPING.get(type(spark_data_type))
```
tobymao, line 156: i feel like this code could be cleaner, just need two methods, 

sqlglot_to_spark
spark_to_sqlglot

def sqlglot_to_spark:
   if isinstance(struct):
       ..
   if primitive type:
      ..

eakmanrq: Yeah that is another way to do it. I'm indifferent either way so I can do that.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Good call added them back.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Added a test case to cover this

sqlmesh/core/engine_adapter/spark.py
eakmanrq: This is not a SQLGlot class it is spark. Array/Map want args to be passed in while struct wants a single list to be passed in. I rewrote it a bit if that helps.

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,141 @@ def _use_spark_session(self) -> bool:
...
-                [
-                    spark_types.StructField(col_name, mapping[col_type.this]())
-                    for col_name, col_type in columns_to_types.items()
+        COMPLEX_TYPE_MAPPING: t.Dict[t.Any, t.Any] = {
+            exp.DataType.Type.ARRAY: spark_types.ArrayType,
+            exp.DataType.Type.MAP: spark_types.MapType,
+            exp.DataType.Type.STRUCT: spark_types.StructType,
+        }
+
+        SPARK_PRIMITIVE_MAPPING = {v: k for k, v in PRIMITIVE_MAPPING.items()}
```
tobymao, line 103: is the behavior for for duplicate keys to take the last one?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,141 @@ def _use_spark_session(self) -> bool:
...
+        SPARK_COMPLEX_TYPE_MAPPING = {v: k for k, v in COMPLEX_TYPE_MAPPING.items()}
+
+        def sqlglot_complex_to_spark_complex(complex_type: exp.DataType) -> spark_types.DataType:
```
tobymao, line 106: rename this to sqlglot_to_spark_type or something, remove the complex

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,141 @@ def _use_spark_session(self) -> bool:
         return True
 
     @classmethod
-    def convert_columns_to_types_to_pyspark_schema(
+    @t.overload
+    def convert_between_spark_and_sqlglot(
```
tobymao, line 55: why is this mixing both? it's a much bigger method and makes the types more complicated? just make it one method each.

def spark_to_sqlglot_types

def sqlglot_to_spark_types

tobymao: my guess is you were trying to limit imports of pyspark mapping to one method? just put the import into a method and call it in both

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Yes

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -50,37 +51,141 @@ def _use_spark_session(self) -> bool:
...
+                    else partial(sqlglot_complex_to_spark_complex, data_type)
+                )
+                if is_struct:
+                    expressions.append(spark_types.StructField(col_name, type_func()))
+                else:
+                    expressions.append(type_func())
+            klass = COMPLEX_TYPE_MAPPING[complex_type.this]
+            if is_struct:
+                return klass(expressions)
+            return klass(*list(expressions))
```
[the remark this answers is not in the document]
eakmanrq: Yeah you are right the `list()` is not needed. Removed.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: done

tobymao (approved):

PR #1414, opened 2023-09-12
Fix!: Create new snapshots instead of overwriting the existing ones during migration

izeigerman:


sqlmesh/migrations/v0025_fix_intervals_and_missing_change_category.py
```diff
@@ -0,0 +1,117 @@
+"""Normalize intervals and fix missing change category."""
```
izeigerman, line 1: This migration should've happened a long time but I instead chose (erroneously) to handle this as part of `_migrate_rows`. Correcting this mistake as part of this PR since we no longer get rid of old rows.

tobymao (approved):

PR #1415, opened 2023-09-12
Fix: Use model's columns_to_types if available in tests

vchan:
Had to also fix some of our tests because data were wrong types.


tobymao:
how does it actually leverage columns_to_types?

vchan:
When loading the input fixture data, it will use the corresponding model's columns_to_types when creating the fixture views. Before, it was inferring the SQL types from the Python types ruamel thinks each column might be.

tobymao (approved):

PR #1418, opened 2023-09-12
Chore: Add Automated Engine Adapter Integration Tests

eakmanrq:
This change makes it so we will run these engine adapter integrations on main similar to Airflow integration tests.


tobymao (approved):

eakmanrq:
Will follow up with Databricks, Bigquery, and MSSQL support

PR #1419, opened 2023-09-12
Fix: Prevent duplicate file name in same folder

mykalmax:


web/client/src/library/components/fileExplorer/context.tsx
```diff
@@ -147,7 +147,10 @@ export default function FileExplorerProvider({
   function renameArtifact(artifact: ModelArtifact, newName?: string): void {
     newName = newName?.trim()
 
-    if (isLoading || isStringEmptyOrNil(newName)) return
+    const parentAtrifacts = artifact.parent?.artifacts ?? []
```
vchan, line 150:
```suggestion
    const parentArtifacts = artifact.parent?.artifacts ?? []
```

vchan (approved):

PR #1420, opened 2023-09-12
Fix: Set file extension for created file based on top level parent

mykalmax:
Set default extension for models to `sql` by and for tests to `yaml`

### DIFF
--- a/web/client/src/library/components/fileExplorer/context.tsx
+++ b/web/client/src/library/components/fileExplorer/context.tsx
@@ -36,7 +36,7 @@ interface FileExplorer {
   setArtifactRename: (artifact?: ModelArtifact) => void
   selectArtifactsInRange: (to: ModelArtifact) => void
   createDirectory: (parent: ModelDirectory) => void
-  createFile: (parent: ModelDirectory) => void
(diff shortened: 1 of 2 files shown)


web/client/src/library/components/fileExplorer/context.tsx
```diff
@@ -125,9 +125,19 @@ export default function FileExplorerProvider({
       })
   }
 
-  function createFile(parent: ModelDirectory, extension = '.py'): void {
+  function createFile(parent: ModelDirectory, extension?: string): void {
     if (isLoading) return
 
+    if (isNil(extension)) {
+      if (parent.isModels) {
+        extension = '.sql'
```
vchan, line 133: I guess most users would want SQL models, but we do have Python models.

vchan (approved):

web/client/src/library/components/fileExplorer/context.tsx
mykalmax: It just default, user can rename file to get python model. If mostly users be creating sql and occasionally python it make sense to set default to sql

PR #1423, opened 2023-09-13
Docs: document python-based SQL models

GeorgeSittas:


docs/concepts/models/sql_models.md
```diff
@@ -1,37 +1,40 @@
 # SQL models
 
-SQL models are the main types of models used by SQLMesh. SQL models consist of three sections:
+SQL models are the main type of models used by SQLMesh. These models can be defined using either SQL or Python.
```
tobymao, line 3: it should be types because of the are i think

tobymao: using either SQL or Python that generates SQL

docs/concepts/models/sql_models.md
```diff
@@ -1,37 +1,40 @@
...
 
-* `MODEL` DDL
-* Optional sql statements
-* The model's `SELECT` statement
+## SQL-based definition
 
-SQL models are designed to look and feel like you're simply using SQL, but they can be customized for advanced use cases.
+The SQL-based definition of SQL models is the most common one, and consists of the following sections:
```
tobymao, line 7: it should say SQL models can be defined in pure SQL, without YAML etc, why did you get rid of the simply sql section?

docs/concepts/models/sql_models.md
GeorgeSittas: > it should be types because of the are i think

In my view, the previous form read like there are several _types_, where there's just one: that which represents these SQL models. I'm a bit rusty on my grammar though, so happy to revert if this is incorrect.

GeorgeSittas: > using either SQL or Python that generates SQL

Good point, I like this more

docs/concepts/models/sql_models.md
GeorgeSittas: It's just the git diff here; the "simply SQL" part is just below this section, I haven't removed it.

docs/concepts/models/sql_models.md
GeorgeSittas: > it should be types because of the are i think

Could also become «SQL models constitute the main type of models used by SQLMesh»

docs/concepts/models/sql_models.md
```diff
@@ -40,20 +43,70 @@ SELECT
...
+```python linenums="1"
+from sqlglot import exp
+
+from sqlmesh.core.model import FullKind, model
+from sqlmesh.core.macro import MacroEvaluator
+
+@model(
+    "db.customers",
+    is_sql=True,
+    kind=FullKind(),
```
tobymao, line 91: can you just put the string or no?

tobymao (approved):

docs/concepts/models/sql_models.md
GeorgeSittas: Seems like you can! Updated, thanks.

docs/concepts/models/sql_models.md
treysp: type is probably better here - good catch

docs/concepts/models/sql_models.md
```diff
@@ -1,37 +1,40 @@
...
-To create an example SQL model, add a file named `my_model.sql` into the `models/` directory (or a subdirectory of `models/`) within your SQLMesh project.
+* The `MODEL` DDL
+* Optional pre-statements
+* A single query
+* Optional post-statements
 
-Although the name of the file doesn't matter, it is recommended to name it something that includes the word "model". Each file can only have one model defined within it.
+These models are designed to look and feel like you're simply using SQL, but they can be customized for advanced use cases.
+
+To create a SQL-based model, add a new file with the `.sql` suffix into the `models/` directory (or a subdirectory of `models/`) within your SQLMesh project. Although the name of the file doesn't matter, it is recommended to use a name that preserves the model's name. Each file can only have one model defined within it.
```
treysp, line 16: Instead of "it is recommended...", maybe "it is customary to use the model's name (without the schema) as the file name. For example, the file containing the model `sqlmesh_example.seed_model` file would be named `seed_model.sql`."

docs/concepts/models/sql_models.md
```diff
@@ -1,37 +1,40 @@
...
 */
 CACHE TABLE countries AS SELECT * FROM raw.countries;
 
 /*
-  This is the main SQL statement that comprises your model.
-  Although it is not required, it is best practice to specify explicit types
-  for all columns in the final SELECT statement.
+  This is the single query that defines the model's logic.
+  Although it is not required, it is considered best practice to explicitly
+  specify the type for each one of the model's columns through casting.
```
treysp, line 37: "...through casting. SQLMesh supports the casting format `<column name>::<data type>` in models of any SQL dialect."

docs/concepts/models/sql_models.md
```diff
@@ -40,20 +43,70 @@ SELECT
...
-The `MODEL` DDL is used to specify metadata about the model such as name, [kind](./model_kinds.md), owner, cron, and others. The Model statement should be the first statement in your model SQL file.
+### `MODEL` DDL
+The `MODEL` DDL is used to specify metadata about the model such as its name, [kind](./model_kinds.md), owner, cron, and others. This should be the first statement in your SQL-based model's file.
+
+Refer to `MODEL` [properties](./overview.md#properties) for the full list of allowed properties.
+
+### Optional pre/post-statements
+Optional pre/post-statements can help prepare the model's query and execute "clean-up" actions after it has successfully executed, respectively. For example, you might create temporary views, set permissions, or evict previously cached tables from memory.
+
+However, be careful not to run any statement that could conflict with the execution of another statement when ran concurrently, such as creating a physical table.
```
treysp, line 63: "when ran concurrently --> "if the models run concurrently"

docs/concepts/models/sql_models.md
```diff
@@ -40,20 +43,70 @@ SELECT
...
+
+## Python-based definition
 
-Refer to `model` [properties](./overview.md#properties) for the full list of allowed properties.
+The Python-based definition of SQL models consists of a single python function, decorated with SQLMesh's `@model` [decorator](https://wiki.python.org/moin/PythonDecorators). The decorator is required to have the `is_sql` keyword argument set to `True` to distinguish it from [Python models](./python_models.md) that return DataFrame instances.
 
-### Optional statements
-Optional SQL statements can help you prepare the model's `SELECT` statement. For example, you might create temporary tables or set permissions.
+This function's return value serves as the model's query, and it must be either a SQL string, or a [SQLGlot expression](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py). The `@model` decorator is used to define the model's [metadata](#MODEL-DDL) and, optionally, its pre/post-statements that are also in the form of SQL strings or SQLGlot expressions.
```
treysp, line 72: Remove comma from "either a SQL string, or a".

"...optionally, pre/post-statements in the form of SQL strings..."

docs/concepts/models/sql_models.md
```diff
@@ -40,20 +43,70 @@ SELECT
...
 
-However, be careful not to run any command that could conflict with the execution of the model's query, such as creating a physical table.
+Defining a model like this can be beneficial in cases where its query is too complex to express cleanly in SQL, for example due to having many dynamic components that would require heavy use of [macros](../macros/overview/). Since Python-based models generate SQL, they also support column-level [lineage](../glossary/#lineage), [automatic change categorization](../glossary/#automatic-data-rebasing), etc.
```
treysp, line 74: "Defining a model like this" --> "Defining a SQL model with Python"

"they also support" --> "they support the same features as regular SQL models, such as"

docs/concepts/models/sql_models.md
```diff
@@ -40,20 +43,70 @@ SELECT
...
 
-### Model `SELECT` query
-The model must contain a standalone `SELECT` query. The result of this query will be used to populate the model table or view.
+To create a Python-based model, add a new file with the `.py` suffix into the `models/` directory (or a subdirectory of `models/`) within your SQLMesh project. The file naming conventions of Python-based models are similar to those of SQL-based models. Inside this file, define a function named `entrypoint` with a single `evaluator` argument, as shown in the example below.
+
+### Example
+
+The following example demonstrates how the above `db.customers` model can be defined as a Python-based model, using SQLGlot's `Expression` builder methods:
```
treysp, line 80: remove comma

treysp (approved):
Great work!

GeorgeSittas:
Thanks for the review @treysp 👍, will incorporate the feedback shortly.

PR #1428, opened 2023-09-14
Docs: expand metrics docs

treysp:


docs/concepts/metrics/definition.md
```diff
@@ -1,17 +1,11 @@
...
+Metrics are defined in SQL files in the `metrics/` directory of your SQLMesh project. A single file can contain multiple metric definitions.
 
-```sql linenums="1"
-METRIC (
-  name total_orders_from_active_customers,
-  owner jen,
-  expression COUNT(IF(sushi.customers.status = 'ACTIVE', sushi.orders.id, NULL))
-);
-```
+A metric is defined with the function `METRIC()` and must include the `name` and `expression` keys. The `name` is used to access the metric in a model query, and the `expression` contains the SQL code used to calculate the metric.
```
tobymao, line 5: what do you mean "model query", you mean metric query?

tobymao (approved):

docs/concepts/metrics/definition.md
treysp: I meant that in a model query you access a metric with `METRIC(<metric name>)`. I can see how that's confusing - will delete that clause.

tobymao: you can't access them in model queries yet

PR #1430, opened 2023-09-15
Chore: Wrap JSON strings with PARSE_JSON() in values

vchan:
Example:
```python
from sqlmesh.core.dialect import select_from_values_for_batch_range
values = [
    (1, '2022-01-01', '{"foo":"bar"}'), 
    (2, '2022-01-01', '{"foo":"qaz"}')
]
columns_to_types = {
    'id': exp.DataType.build("int"), 
    'ds': exp.DataType.build("text"), 
    'json_col': exp.DataType.build("json")
}
print(select_from_values_for_batch_range(values, columns_to_types, 0, len(values)).sql(pretty=True))
# SELECT
#   CAST(id AS INT) AS id,
#   CAST(ds AS TEXT) AS ds,
#   CAST(json_col AS JSON) AS json_col
# FROM (VALUES
#   (1, '2022-01-01', PARSE_JSON('{"foo":"bar"}')),
#   (2, '2022-01-01', PARSE_JSON('{"foo":"qaz"}'))) AS t(id, ds, json_col)
```

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -697,7 +697,11 @@ def select_from_values_for_batch_range(
         exp.alias_(exp.cast(column, to=kind), column, copy=False)
         for column, kind in columns_to_types.items()
     ]
-    values_exp = exp.values(values[batch_start:batch_end], alias=alias, columns=columns_to_types)
+    values_exp = exp.values(
+        [tuple(transform_values(v, columns_to_types)) for v in values[batch_start:batch_end]],
+        alias=alias,
+        columns=columns_to_types,
+    )
     return exp.select(*casted_columns).from_(values_exp, copy=False)
 
 
@@ -759,3 +763,13 @@ def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Se
         for table in scope.tables
         if not isinstance(table.this, exp.Func) and exp.table_name(table) not in scope.cte_sources
     }
+
+
+def transform_values(
+    values: t.Tuple[t.Any, ...], columns_to_types: t.Dict[str, exp.DataType]
+) -> t.Iterator[t.Any]:
+    for value, col_type in zip(values, columns_to_types.values()):
+        if col_type == exp.DataType.build("json"):
+            yield exp.func("PARSE_JSON", f"'{value}'")
+        else:
+            yield value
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -759,3 +763,17 @@ def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Se
...
+) -> t.Iterator[t.Any]:
+    """Perform transformations on values given columns_to_types.
+
+    Currently, the only transformation is wrapping JSON columns with PARSE_JSON().
+    """
+    for value, col_type in zip(values, columns_to_types.values()):
+        if col_type == exp.DataType.build("json"):
+            yield exp.func("PARSE_JSON", f"'{value}'")
+        else:
+            yield value
```
GeorgeSittas, line 779: Should we traverse `columns_to_types` once so that we do this transformation using indexes instead of iterating over the whole tuple each time? I'm a bit skeptical about the performance of this.

vchan: Won't it be the same number of iterations since we have to iterate over every element in the `values` tuple anyway?

GeorgeSittas: Hmm maybe, unless we mutate `values`?

vchan: I don't think we can since `values` is a tuple.

GeorgeSittas: Ah, that's right, I was under the impression we used a list. Please ignore me then.

sqlmesh/core/dialect.py
```diff
@@ -759,3 +763,17 @@ def find_tables(expression: exp.Expression, dialect: DialectType = None) -> t.Se
...
+
+def transform_values(
+    values: t.Tuple[t.Any, ...], columns_to_types: t.Dict[str, exp.DataType]
+) -> t.Iterator[t.Any]:
+    """Perform transformations on values given columns_to_types.
+
+    Currently, the only transformation is wrapping JSON columns with PARSE_JSON().
+    """
+    for value, col_type in zip(values, columns_to_types.values()):
+        if col_type == exp.DataType.build("json"):
```
GeorgeSittas, line 776: How about factoring out the `exp.DataType.build("json")` call to avoid rebuilding it over and over again?

GeorgeSittas (approved):
LGTM

sqlmesh/core/dialect.py
vchan: That's a good idea. I'll do that.

PR #1435, opened 2023-09-15
Fix(mssql): enable core functionality

treysp:
- Replace `using` with `on` for all joins, qualify column names in queries
- Append semi-colon to all MSSQL queries
- `TRUNCATE TABLE` instead of `TRUNCATE` for MSSQL
- Fix MSSQL column/table metadata querying
- Fix MSSQL drop_table lack of qualification
- Fix pandas -> MSSQL datetime64 conversion
- Add alternative to `(1, 2) IN ((1, 2))` tuple syntax for engines that don't support it
- Use `varchar` instead of `text` columns for table index columns and columns used in string literal equalities
- Add MSSQL to engine adapter integration tests


sqlmesh/core/engine_adapter/base.py
```diff
@@ -944,7 +946,11 @@ def scd_type_2(
                     "deleted",
                     exp.select(*[f"static.{col}" for col in columns_to_types])
                     .from_("static")
-                    .join("latest", using=unique_key, join_type="left")
+                    .join(
+                        "latest",
+                        on=" AND ".join(f"static.{col} = latest.{col}" for col in unique_key),
```
eakmanrq, line 951: @tobymao @GeorgeSittas Should SQLGlot be updated to correctly transpile using for MSSQL?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -40,25 +42,53 @@ class MSSQLEngineAdapter(
 
     DIALECT: str = "tsql"
     FALSE_PREDICATE = exp.condition("1=2")
```
eakmanrq, line 44: I believe @tobymao said this is no longer needed and you can just put `exp.false()`. I believe SQLGlot now converts that to something like `1=2`. If that is the case then this `FALSE_PREDICATE` can be completely removed since it was just added for mssql.

sqlmesh/core/engine_adapter/base.py
GeorgeSittas: Not sure what part of the transpilation is incorrect; what is the current vs expected behavior in SQLGlot?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -40,25 +42,53 @@ class MSSQLEngineAdapter(
...
+        database_name = table.db
+        if database_name:
+            sql = sql.where(f"table_schema = '{database_name}'")
+
+        self.execute(sql)
+
+        columns_raw = self.cursor.fetchall()
+
+        columns = [
+            (col[0], f"{col[1]}({col[2]})") if col[1] == "varchar" and col[2] else (col[0], col[1])
```
eakmanrq, line 72: Aren't there other data types that have precision bits like `CHAR` and `DECIMAL`?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -40,25 +42,53 @@ class MSSQLEngineAdapter(
...
+        include_pseudo_columns: bool = True,
+    ) -> t.Dict[str, exp.DataType]:
+        """MsSql doesn't support describe so we query information_schema."""
 
-        Reference: https://github.com/aws/amazon-redshift-python-driver/blob/master/redshift_connector/cursor.py#L528-L553
-        """
         table = exp.to_table(table_name)
 
-        catalog_name = table.args.get("catalog") or "master"
+        catalog_name = table.catalog or "master"
```
eakmanrq, line 57: This logic is repeated to assign the default catalog name of `master`. One option would be to create a class variable at the top that is `DEFAULT_CATALOG_NAME` and set that to `master` and then replace this with `table.catalog or self.DEFAULT_CATALOG_NAME`. Another would be to create a method like `__get_catalog(table: Table) -> str: return table.cataog or "master"`. And then do `self.__get_catalog(table)` here.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -83,7 +113,9 @@ def drop_schema(
             # with a `drop_table` call.
             objects = self._get_data_objects(schema_name)
             for obj in objects:
-                self.drop_table(obj.name, exists=ignore_if_not_exists)
+                self.drop_table(
+                    f"{obj.catalog}.{obj.schema_name}.{obj.name}", exists=ignore_if_not_exists
```
eakmanrq, line 117: Nit: You can do `".".join([obj.catalog, obj.schema_name, obj.name])`

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -249,16 +249,19 @@ def delete_expired_environments(self) -> t.List[Environment]:
...
+            self.snapshots_table, where=self._snapshot_id_filter(snapshot_ids, self.snapshots_table)
         )
 
     def snapshots_exist(self, snapshot_ids: t.Iterable[SnapshotIdLike]) -> t.Set[SnapshotId]:
         return {
             SnapshotId(name=name, identifier=identifier)
             for name, identifier in self.engine_adapter.fetchall(
-                exp.select("name", "identifier")
+                exp.select(
+                    exp.to_column(f"{self.snapshots_table}.name"),
```
eakmanrq, line 260: Did you add this because MSSQL requires the full qualified name?

sqlmesh/core/engine_adapter/base.py
eakmanrq: @treysp Can you explain why you replaced all the `using` with `on` throughout this PR?

sqlmesh/migrations/v0011_add_model_kind_name.py
```diff
@@ -55,7 +55,7 @@ def migrate(state_sync):  # type: ignore
                 "identifier": exp.DataType.build(text_type),
                 "version": exp.DataType.build(text_type),
                 "snapshot": exp.DataType.build("text"),
```
eakmanrq, line 57: This is missing the `text_type`.

izeigerman: No, this can't be a text type. Text type is only used for text columns that can be part of a key / index. It has a limited number of characters and shouldn't be applied to `snapshot`.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -551,25 +567,46 @@ def _get_snapshot_intervals(
...
+                exp.to_column(f"{self.intervals_table}.identifier"),
+                "created_ts",
+                "is_removed",
+            )
         )
 
         if uncompacted_only:
             query.join(
-                exp.select("name", "identifier")
+                exp.select(exp.column("name"), exp.column("identifier"))
```
izeigerman, line 590: Why has this been changed?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -832,27 +869,61 @@ def map_data_versions(
...
-    ) -> t.Union[exp.In, exp.Boolean]:
+        self, snapshot_ids: t.Iterable[SnapshotIdLike], fq_table_name: str
+    ) -> t.Union[exp.In, exp.Boolean, exp.Condition]:
         if not snapshot_ids:
             return exp.false()
-
-        return t.cast(exp.Tuple, exp.convert((exp.column("name"), exp.column("identifier")))).isin(
-            *[(snapshot_id.name, snapshot_id.identifier) for snapshot_id in snapshot_ids]
-        )
+        elif self.engine_adapter.SUPPORTS_TUPLE_IN:
```
izeigerman, line 876: This branching seems too much, especially provided we use inheritance. Can we just override this method and `_snapshot_name_version_filter` in the `MSSQLEngineApdater` instead and get rid of this if / else?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -966,14 +972,22 @@ def scd_type_2(
                         *(f"source.{col} AS s_{col}" for col in unmanaged_columns),
                     )
                     .from_("latest")
-                    .join("source", using=unique_key, join_type="left")
+                    .join(
+                        "source",
+                        on=" AND ".join(f"latest.{col} = source.{col}" for col in unique_key),
```
izeigerman, line 977: Shouldn't this be:
```
exp.and(*[exp.to_column(f"latest.{col}").eq(exp.to_column(f"source.{col}")) for col in unique_key) 
```
?
Here and all other places in this query

sqlmesh/core/engine_adapter/base.py
izeigerman: MSSQL doesn't support `USING`. `USING` can't be transpiled into `ON` since the semantics of the two is different.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1294,7 +1308,12 @@ def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.An
         if quote:
             quote_identifiers(expression)
 
-        return expression.sql(**sql_gen_kwargs)  # type: ignore
+        sql = expression.sql(**sql_gen_kwargs)  # type: ignore
+
+        if self.TRAILING_SEMICOLON:
```
izeigerman, line 1313: Again, can we just override this in `MSSQLEngineAdapter`:
```
def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.Any) -> str:
    sql = super()._to_sql(expression, quote=quote, **kwargs):
    return f"{sql};"

File: sqlmesh/core/engine_adapter/mssql.py [hunk 12]
```diff
@@ -146,3 +178,7 @@ def _get_data_objects(
             )
             for row in dataframe.itertuples()
         ]
+
+    def _truncate_table(self, table_name: TableName) -> str:
+        table = quote_identifiers(exp.to_table(table_name))
```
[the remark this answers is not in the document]
izeigerman: Looks like this is jus something that was carried over

sqlmesh/migrations/v0011_add_model_kind_name.py
treysp: I believe the max fixed length for a key/index is higher in MSSQL than in MySQL - should I update it to use that?

And we're setting it for `kind_name` solely due to type (not length). Should I max out the length for kind_name, or reuse the key/index length?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -122,6 +122,8 @@ class EngineAdapter:
     SUPPORTS_MATERIALIZED_VIEWS = False
     SUPPORTS_CLONING = False
     SCHEMA_DIFFER = SchemaDiffer()
+    SUPPORTS_TUPLE_IN = True
+    TRAILING_SEMICOLON = False
```
tobymao, line 126: why do you need this flag? could you just always add semi colons?

sqlmesh/core/engine_adapter/base.py
tobymao, line 951: exp._and(* exp.col(col, table="static").eq(...) for col in unique_key)

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -359,13 +362,26 @@ def _get_snapshots(
...
+            )
         )
         if hydrate_seeds:
-            query = query.select("content").join(
-                self.seeds_table, using=["name", "identifier"], join_type="left outer"
+            query = query.select(exp.to_column(f"{self.seeds_table}.content")).join(
+                self.seeds_table,
+                on=exp.and_(
+                    *[
+                        exp.to_column(f"{self.snapshots_table}.{col}").eq(
```
tobymao, line 378: would prefer exp.column() so we can avoid parsing

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -359,13 +362,26 @@ def _get_snapshots(
...
+                            exp.to_column(f"{self.seeds_table}.{col}")
+                        )
+                        for col in ["name", "identifier"]
+                    ]
+                ),
+                join_type="left outer",
```
tobymao, line 384: "left"

sqlmesh/core/state_sync/engine_adapter.py
tobymao: yea, this is unnecessary

sqlmesh/core/state_sync/engine_adapter.py
treysp: Annoyingly, the joins using `ON` with qualified names means we have to qualify them everywhere

sqlmesh/core/state_sync/engine_adapter.py
treysp: `self.snapshots_table` is of the form `"schema.table"` (no quotes between the two), which doesn't align with the sqlglot-generated `"schema"."table"` and errors. `to_column` parses/generates them with the correct quoting in the context of the query.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -551,25 +567,46 @@ def _get_snapshot_intervals(
     ) -> t.Tuple[t.Set[str], t.List[SnapshotIntervals]]:
         query = (
             exp.select(
-                "id", "name", "identifier", "version", "start_ts", "end_ts", "is_dev", "is_removed"
+                "id",
+                exp.to_column(f"{self.intervals_table}.name"),
```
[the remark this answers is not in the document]
treysp: same issue with schema/table quoting above (`"schema.table"` vs. `"schema"."table"`)

sqlmesh/core/engine_adapter/base.py
eakmanrq: >MSSQL doesn't support USING. USING can't be transpiled into ON since the semantics of the two is different.

Is that right? Isn't USING just ON but with both columns qualified to match the two tables being joined?

eakmanrq:
Can you add mssql to this? https://github.com/TobikoData/sqlmesh/blob/04f73660ad80e5503eacc92181a0a6fc0fbbd2f2/tests/core/engine_adapter/test_integration.py#L190

So it would be: `pytest.param("mssql", marks=[pytest.mark.integration, pytest.mark.engine_integration]),`

I will configure CircleCI. The docker container is already configured here: https://github.com/TobikoData/sqlmesh/blob/df8f0584a818a579ac9bf3afdc708358dd8dbd0a/tests/core/engine_adapter/docker-compose.yaml#L15

Also make sure that all the tests pass when you run it locally. Let me know if you have any questions on how to do that.

eakmanrq (approved):

PR #1441, opened 2023-09-18
Fix!: Correctly drop materialized view

izeigerman:
Make sure that whether a model represents a materialized view is reflected in the data hash.

### DIFF
--- a/sqlmesh/core/model/kind.py
+++ b/sqlmesh/core/model/kind.py
@@ -120,6 +120,10 @@ class _ModelKind(PydanticModel, ModelKindMixin):
     def to_expression(self, **kwargs: t.Any) -> d.ModelKind:
         return d.ModelKind(this=self.name.value.upper(), **kwargs)
 
+    @property
+    def data_hash_fields(self) -> t.List[str]:
+        return [self.name.value]
+
 
 class TimeColumn(PydanticModel):
     column: str
@@ -244,6 +248,10 @@ class ViewKind(_ModelKind):
         if isinstance(v, exp.Expression):
             return bool(v.this)
         return bool(v)
+
+    @property
+    def data_hash_fields(self) -> t.List[str]:
+        return [*super().data_hash_fields, str(self.materialized)]
 
 
 class SeedKind(_ModelKind):
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -1098,7 +1098,15 @@ class ViewStrategy(PromotableStrategy):
         )
 
     def delete(self, name: str) -> None:
-        self.adapter.drop_view(name)
+        try:
+            self.adapter.drop_view(name)
+        except Exception:
+            logger.debug(
+                "Failed to drop view '%s'. Trying to drop the materialized view instead",
+                name,
+                exc_info=True,
+            )
+            self.adapter.drop_view(name, materialized=True)
         logger.info("Dropped view '%s'", name)
 
     def _is_materialized_view(self, model: Model) -> bool:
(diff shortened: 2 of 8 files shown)


sqlmesh/migrations/v0026_fix_view_kind_data_hash.py
```diff
@@ -0,0 +1,5 @@
+"""Reflect the `materialized` flag of the VIEW model kind in the data hash."""
+
+
+def migrate(state_sync):  # type: ignore
+    pass
```
GeorgeSittas, line 5: Just FMI: what does an empty migration achieve?

GeorgeSittas (approved):

sqlmesh/core/model/kind.py
```diff
@@ -120,6 +120,10 @@ def model_kind_name(self) -> t.Optional[ModelKindName]:
     def to_expression(self, **kwargs: t.Any) -> d.ModelKind:
         return d.ModelKind(this=self.name.value.upper(), **kwargs)
 
+    @property
+    def data_hash_fields(self) -> t.List[str]:
```
crericha, line 124: Consider renaming to something like `data_hash_values` since it returns the values and not the field names.

sqlmesh/migrations/v0026_fix_view_kind_data_hash.py
izeigerman: It triggers the fingerprint recomputation. I'm going to drop it because I'm going to have a real migration script in a different PR.

sqlmesh/core/model/kind.py
izeigerman: I actually followed the convention from `model/definition.py`. Should I still rename this?

sqlmesh/migrations/v0026_fix_view_kind_data_hash.py
GeorgeSittas: Gotcha, thanks for explaining

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1098,7 +1098,15 @@ def migrate(
...
     def delete(self, name: str) -> None:
-        self.adapter.drop_view(name)
+        try:
+            self.adapter.drop_view(name)
+        except Exception:
+            logger.debug(
+                "Failed to drop view '%s'. Trying to drop the materialized view instead",
+                name,
+                exc_info=True,
+            )
```
crericha, line 1108: We don't have a way of knowing if the view was materialized, so we need to use a brute force approach?

izeigerman: Yes, unfortunately. This information is not available when we're cleaning things up.

sqlmesh/core/model/kind.py
crericha: In my opinion, the name of the method is misleading. If you agree, change it; otherwise, fine as is.

crericha (approved):

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1098,7 +1098,15 @@ def migrate(
...
+            self.adapter.drop_view(name, materialized=True)
```
tobymao, line 1109: try catch this too?? or no

tobymao (approved):

sqlmesh/core/model/kind.py
izeigerman: renamed

sqlmesh/core/snapshot/evaluator.py
izeigerman: If we're cleaning up a view and neither call succeeded than something is definitely wrong and we should let the error bubble up.

PR #1443, opened 2023-09-18
Feat!: Support CSV parser settings for seed models

### LINKED ISSUE #1434: Support for read_csv configuration
We have a contract in our data pipeline which relies on a particular CSV format. The downstream consumers of this contract are the area where we are integrating sqlmesh. We want to create a seed model which can consume one of these CSVs, but it doesn't conform to pandas defaults.

We'd love to be able to supply a read_csv configuration object.

Here's an example csv we're wanting to create seed models from

```
start_time,end_time,proto_num,description
'2023-01-27 02:51:08+00','2023-01-27 02:52:08+00',17,'Great stuff bud\'s'
```

izeigerman:
Closes #1434

### DIFF
--- a/sqlmesh/core/model/kind.py
+++ b/sqlmesh/core/model/kind.py
@@ -9,7 +9,8 @@ from sqlglot import exp
 from sqlglot.time import format_time
 
 from sqlmesh.core import dialect as d
-from sqlmesh.core.model.common import bool_validator, properties_validator
+from sqlmesh.core.model.common import bool_validator, parse_properties
+from sqlmesh.core.model.seed import CsvSettings
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import (
     PydanticModel,
@@ -121,7 +122,7 @@ class _ModelKind(PydanticModel, ModelKindMixin):
         return d.ModelKind(this=self.name.value.upper(), **kwargs)
 
     @property
-    def data_hash_values(self) -> t.List[str]:
+    def data_hash_values(self) -> t.List[t.Optional[str]]:
         return [self.name.value]
 
 
@@ -250,7 +251,7 @@ class ViewKind(_ModelKind):
         return bool(v)
 
     @property
-    def data_hash_fields(self) -> t.List[str]:
+    def data_hash_values(self) -> t.List[t.Optional[str]]:
         return [*super().data_hash_values, str(self.materialized)]
 
 
@@ -258,9 +259,7 @@ class SeedKind(_ModelKind):
     name: Literal[ModelKindName.SEED] = ModelKindName.SEED
     path: str
     batch_size: int = 1000
-    csv_settings: t.Optional[exp.Tuple] = None
-
-    _csv_settings_validator = properties_validator
+    csv_settings: t.Optional[CsvSettings] = None
 
     @field_validator("batch_size", mode="before")
     @classmethod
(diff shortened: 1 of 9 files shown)


sqlmesh/core/model/kind.py
```diff
@@ -280,6 +279,18 @@ def _parse_path(cls, v: t.Any) -> str:
             return v.this
         return str(v)
 
+    @field_validator("csv_settings", mode="before")
+    @classmethod
+    def _parse_csv_settings(cls, v: t.Any) -> t.Optional[CsvSettings]:
+        if v is None or isinstance(v, CsvSettings):
+            return v
+        if isinstance(v, exp.Expression):
+            tuple_exp = parse_properties(cls, v, {})
```
GeorgeSittas, line 288: Should we pass `dialect` in the `values` arg here?

sqlmesh/core/model/kind.py
```diff
@@ -280,6 +279,18 @@ def _parse_path(cls, v: t.Any) -> str:
...
+            if not tuple_exp:
+                return None
+            return CsvSettings(**{e.left.name: e.right for e in tuple_exp.expressions})
+        return CsvSettings(**v)
```
GeorgeSittas, line 292: Is this dict splat-unpacking always safe?

tobymao (approved):

sqlmesh/core/model/kind.py
izeigerman: Where do we take it from? Additionally it's irrelevant here since we are not parsing a string here neither do we need to store tie dialect in the expresson's meta, since we throw the tuple away.

sqlmesh/core/model/kind.py
izeigerman: Sorry, not sure what's your concern exactly. When do you think it might not be? The idea is that we pass the dict as is and let `CsvSetting`'s validators to take care of it.

GeorgeSittas: If I understand this correctly, if line 292 is hit we assume `v` is a `dict` here, right? If so, my concern was thatthere could be a path where `v` is not a `dict`, so this unpacking would then fail. Should've clarified this in my comment, I don't have a counterexample, just looked like a line I'd double-check - feel free to ignore me if this is not the case.

sqlmesh/core/model/kind.py
GeorgeSittas: Both points make sense 👍

sqlmesh/core/model/kind.py
GeorgeSittas: In hindsight, it makes no sense to expect something other than a dict here, so lgtm.

tests/core/test_model.py
```diff
@@ -1989,3 +2012,68 @@ def test_model_jinja_macro_rendering():
...
+            ),
+        );
+        SELECT 1;
+        """
+    )
+    materiralized_view_model_hash = load_sql_based_model(
+        materialized_view_model_expressions
+    ).data_hash
+
+    assert view_model_hash != materiralized_view_model_hash
```
GeorgeSittas, line 2044: Nit:
```suggestion
    materialized_view_model_hash = load_sql_based_model(
        materialized_view_model_expressions
    ).data_hash

    assert view_model_hash != materialized_view_model_hash
```

GeorgeSittas (approved):

PR #1447, opened 2023-09-19
Fix: allow fixtures to have fewer columns than model defs

tobymao:


izeigerman (approved):
Does this need a unit test?

tobymao:
it's a part of the demo

PR #1451, opened 2023-09-20
Feat: Allow evaluation of partial intervals

izeigerman:


tests/core/test_snapshot.py
```diff
@@ -300,9 +315,19 @@ def test_lookback(snapshot: Snapshot, make_snapshot):
     assert snapshot.missing_intervals("2023-01-28", "2023-01-28", "2023-01-30 05:00:00") == [
         (to_timestamp("2023-01-28"), to_timestamp("2023-01-29")),
     ]
+    assert snapshot.missing_intervals(
+        "2023-01-28", "2023-01-28", "2023-01-30 05:00:00", allow_partial=True
+    ) == [
+        (to_timestamp("2023-01-28"), to_timestamp("2023-01-29")),
```
eakmanrq, line 321: Shouldn't this be `(to_timestamp("2023-01-28"), to_timestamp("2023-01-30"))`?

sqlmesh/core/snapshot/definition.py
```diff
@@ -691,15 +696,20 @@ def missing_intervals(
...
 
         interval_unit = self.node.interval_unit
 
-        upper_bound_ts = to_timestamp(
-            self.node.cron_floor(execution_time) if not ignore_cron else execution_time
-        )
-        end_ts = min(end_ts, to_timestamp(interval_unit.cron_floor(upper_bound_ts)))
+        if allow_partial:
+            upper_bound_ts = to_timestamp(execution_time)
+            end_ts = min(end_ts, upper_bound_ts)
```
tobymao, line 707: why are we taking the upper bound min? why not always do end_ts?

tests/core/test_snapshot.py
izeigerman: Why though? The end was set to `2023-01-28` + partial and we get `2023-01-29`

sqlmesh/core/snapshot/definition.py
tobymao: why does that matter? their source data could have future dates

tests/core/test_snapshot.py
eakmanrq: Yeah sorry get it now.

sqlmesh/core/snapshot/definition.py
izeigerman: yeah, but if the future end results in complete interval we will happily record it

sqlmesh/dbt/model.py
```diff
@@ -263,7 +264,7 @@ def _big_query_partition_by_expr(self) -> exp.Expression:
 
     @property
     def sqlmesh_config_fields(self) -> t.Set[str]:
-        return super().sqlmesh_config_fields | {"cron", "interval_unit"}
+        return super().sqlmesh_config_fields | {"cron", "interval_unit", "allow_partials"}
```
izeigerman, line 267: @crericha can you please confirm that this is sufficient change to enable propagation of this attribute from dbt models?

eakmanrq (approved):

PR #1453, opened 2023-09-20
Chore: Add materialized view engine integration test

eakmanrq:


sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -100,5 +100,21 @@ def replace_query(
...
+    def _get_materialized_views(self, catalog_name: str, schema_name: str) -> exp.Expression:
+        return parse_one(
+            f"""            
+            SELECT
+                {catalog_name} AS catalog_name,
+                viewname AS name,
+                schemaname AS schema_name,
+                'MATERIALIZED_VIEW' AS type
+            FROM
+                pg_views
```
izeigerman, line 112: Provided that in `redshift` views are in the same table as materialized views, aren't they getting included here already https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/base_postgres.py#L117-L123?

eakmanrq: Yes but just represented as a "view". We have this additional union in order to bring in views that are also materialized views.

eakmanrq: See this test demonstrating the current behavior: https://github.com/TobikoData/sqlmesh/blob/b0e7d2bd5c2d06ebb3975cb504582cbbe3e2d720/tests/core/engine_adapter/test_integration.py#L301-L306

izeigerman: wouldn't there be duplicates provided we're using `UNION ALL`?

eakmanrq: Talked and I now understand the issue and it is fixed. We also agreed to drop materialized view support for Redshift due to binding issues.

izeigerman (approved):

PR #1460, opened 2023-09-22
Chore: Add Custom Pydantic Types

eakmanrq:


sqlmesh/utils/pydantic.py
```diff
@@ -185,3 +192,73 @@ def wrapper(cls: t.Type, v: t.Any, values: t.Any, *args: t.Any, **kwargs: t.Any)
...
+        raise ValueError(f"Invalid num {v}. Value must be an integer value")
+    if v <= 0:
+        raise ValueError(f"Invalid num {v}. Value must be a positive integer")
+    return v
+
+
+SqlGlotListOfStrings = t.List[str]
+SqlGlotString = str
+SqlGlotBool = bool
+SqlglotPositiveInt = int
```
eakmanrq, line 243: @izeigerman I had to add this to work with mypy along with the type ignores. I'm not sure if there is another way around it. Basically I assign a type alias and that is how mypy views these but then Pydantic properly uses them as custom types when it does the validation. Technically keeps both happy but not very clean.

sqlmesh/core/model/kind.py
```diff
@@ -9,11 +9,15 @@
 from sqlglot.time import format_time
 
 from sqlmesh.core import dialect as d
-from sqlmesh.core.model.common import bool_validator, parse_properties
+from sqlmesh.core.model.common import parse_properties
 from sqlmesh.core.model.seed import CsvSettings
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import (
     PydanticModel,
+    SqlGlotBool,
```
tobymao, line 17: SQLGlot

sqlmesh/utils/pydantic.py
izeigerman: Any chance you only need this under `t.TYPE_CHECKING`? This way it can look something like:
```
if t.TYPE_CHECKING:
    ...
elif if PYDANTIC_MAJOR_VERSION >= 2:
    ...
else:
    ...
```

sqlmesh/utils/pydantic.py
```diff
@@ -185,3 +192,73 @@ def wrapper(cls: t.Type, v: t.Any, values: t.Any, *args: t.Any, **kwargs: t.Any)
         return func(cls, v, values_dict, *args, **kwargs)
 
     return wrapper
+
+
+T = t.TypeVar("T")
+
+
+class PydanticTypeProxy(t.Generic[T]):
```
izeigerman, line 200: [Nit] Shouldn't this definition be under `else` block since it's only relevant to pydantic v1?

izeigerman (approved):

PR #1466, opened 2023-09-22
Fix: Ignore partial intervals when adding intervals to a snapshot

izeigerman:


sqlmesh/core/snapshot/definition.py
```diff
@@ -617,6 +626,8 @@ def inclusive_exclusive(
         """
         interval_unit = self.node.interval_unit
         start_ts = to_timestamp(interval_unit.cron_floor(start))
+        if start_ts < to_timestamp(start):
+            start_ts = to_timestamp(interval_unit.cron_next(start_ts))
```
izeigerman, line 630: @eakmanrq this is where we make sure that start doesn't exceed the given range.

tests/core/test_integration.py
```diff
@@ -718,7 +723,7 @@ def test_incremental_time_self_reference(
     )
     # Validate that we made 7 calls to the customer_revenue_lifetime snapshot and 1 call to the customer_revenue_by_day snapshot
```
eakmanrq, line 724: Update this comment from `7` to `6`. Also, if you don't find the comment helpful, you could remove it.

tests/core/test_integration.py
```diff
@@ -704,8 +709,8 @@ def test_incremental_time_self_reference(
             SnapshotIntervals(
                 snapshot_name="sushi.customer_revenue_by_day",
                 intervals=[
-                    (to_timestamp(to_date("1 week ago")), to_timestamp(to_date("6 days ago"))),
                     (to_timestamp(to_date("6 days ago")), to_timestamp(to_date("5 days ago"))),
+                    (to_timestamp(to_date("5 days ago")), to_timestamp(to_date("4 days ago"))),
```
eakmanrq, line 713: Why did the intervals get shifted down a date for this model?

izeigerman: because `start` for all models is set to `1 week ago` (not date) which now rounds up instead of down.

eakmanrq: So wouldn't it then have 1 missing interval instead of 2 if the start was brought forward and end remained the same? I think I am just missing something.

izeigerman: > So wouldn't it then have 1 missing interval instead of 2 if the start was brought forward and end remained the same?

Why do you say it's 2? As far as I can see exactly 1 interval has been removed.

More context for completeness sake:
Consider the current time - `2023-09-22 13:12:00`. Thus, `1 week ago` is resolved to `2023-09-15 13:12:00`. Previously this timestamp would be further resolved to `2023-09-15` start timestamp when computing missing intervals for daily models. Now it's resolved to `2023-09-16` instead.

eakmanrq: Right so `2023-09-16` is the new start but the end is still 5 days ago. So before we had missing intervals for 7 days ago and then 6 days ago. Now that start is moved forward we should only have missing intervals for 6 days ago. right? 

So before (starting 7 days ago):
- Missing 7 days ago
- Missing 6 days ago

Now (starting 6 days ago):
- Missing 6 days ago

The PR reads though like we are saying

Now (starting 6 days ago):
- Missing 6 days ago
- Missing 5 days ago

izeigerman: This happened because I set the plan's end to be a date instead of timestamp which added an extra day to the snapshot's end due to the inclusive / exclusive logic. Fixed this.

PR #1468, opened 2023-09-22
Feat: skip croniter when possible for performance

tobymao:


sqlmesh/utils/cron.py
```diff
@@ -21,12 +23,37 @@ def cron_prev(cron: str, time: TimeLike) -> float:
...
+        self.curr: TimeLike = now() if time is None else time
+        self._interval_seconds: t.Optional[int] = None
+
+    @property
+    def interval_seconds(self) -> int:
+        if self._interval_seconds is None:
+            seconds = set()
+            curr = to_datetime(self.curr)
+
+            for _ in range(10):
```
izeigerman, line 35: [Nit] Maybe a named constant like `ESTIMATE_SAMPLES_NUM` or something like that.

sqlmesh/utils/cron.py
```diff
@@ -21,12 +23,37 @@ def cron_prev(cron: str, time: TimeLike) -> float:
...
+                prev = curr
+                curr = to_datetime(cron_next(self.cron, curr))
+                seconds.add(curr - prev)
+
+            if len(seconds) == 1:
+                self._interval_seconds = first(seconds).seconds
+            else:
+                self._interval_seconds = 0
```
izeigerman, line 43: Why 0?

izeigerman: Because it's not `None` but still behaves as false? I'd prefer this to be more explicit

tobymao: docstring

izeigerman (approved):

PR #1469, opened 2023-09-22
Fix : Restore selected env from local storage + add pinned envs to local list

mykalmax:


web/client/src/context/context.ts
```diff
@@ -214,22 +214,57 @@ export const useStoreContext = create<ContextStore>((set, get) => ({
...
+      ModelEnvironment.save({ environments })
       ModelEnvironment.sort(environments)
 
-      const prodEnv = environments.find(({ name }) => name === 'prod')
-      const defaultEnv =
-        environments.find(({ name }) => name === defaultEnvironment) ??
-        s.environment
+      const profileEnv = ModelEnvironment.getEnvironment()
+      let prodEnv = environments.find(({ name }) => name === 'prod')
+      let storredEnv = environments.find(
```
vchan, line 243:
```suggestion
      let storedEnv = environments.find(
```

vchan (approved):

PR #1471, opened 2023-09-23
Feat!: add more interval units and make the smallest 5 minutes

tobymao:
additional performance boosts

still need to add some tests and validate that this doesn't break existing minute level models

### DIFF
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -21,6 +21,9 @@ if t.TYPE_CHECKING:
     from sqlmesh.core.snapshot import Node
 
 
+DEFAULT_SAMPLE_SIZE = 5
+
+
 class IntervalUnit(str, Enum):
     """IntervalUnit is the inferred granularity of an incremental node.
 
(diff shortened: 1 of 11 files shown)


sqlmesh/core/plan/definition.py
```diff
@@ -783,7 +783,9 @@ def is_finished(self) -> bool:
         return self == PlanStatus.FINISHED
 
 
-class SnapshotIntervals(PydanticModel, frozen=True):
+# millions of these can be created, pydantic has significant overhead
```
izeigerman, line 786: Where do millions come from? Isn't it just one per snapshot?

sqlmesh/core/plan/definition.py
```diff
@@ -795,11 +797,12 @@ def format_intervals(self, unit: t.Optional[IntervalUnit] = None) -> str:
         return format_intervals(self.merged_intervals, unit)
 
 
+@dataclass
```
izeigerman, line 800: Ditto

sqlmesh/core/node.py
```diff
@@ -33,17 +36,19 @@ class IntervalUnit(str, Enum):
...
+    def from_cron(klass, cron: str, sample_size: int = DEFAULT_SAMPLE_SIZE) -> IntervalUnit:
         croniter = CroniterCache(cron)
         samples = [croniter.get_next() for _ in range(sample_size)]
         min_interval = min(b - a for a, b in zip(samples, samples[1:]))
         for unit, seconds in sorted(INTERVAL_SECONDS.items(), key=lambda x: x[1], reverse=True):
-            if seconds <= min_interval:
+            if seconds <= min_interval.total_seconds():
                 return unit
-        raise ConfigError(f"Invalid cron '{cron}': must have a cadence of 1 minute or more.")
+        raise ConfigError(f"Invalid cron '{cron}': must have a cadence of 5 minutes or more.")
```
izeigerman, line 51: Do we prohibit using `minute` grain anywhere?

sqlmesh/core/plan/definition.py
tobymao: i gues intervals can be a lot and then pydantic validates it. changing to data class reduces plan by 5x in my tests

sqlmesh/core/node.py
tobymao: yea, it's not an enum anymore

izeigerman (approved):

PR #1476, opened 2023-09-25
Fix: Replace lock icon , relax zoom effect on lineage

mykalmax:


web/client/src/library/components/graph/Graph.tsx
```diff
@@ -164,8 +164,11 @@ const ModelColumnDisplay = memo(function ModelColumnDisplay({
       )}
       <div className="w-full">
         <div className="w-full flex justify-between items-center">
-          <span className={clsx('flex items-center', disabled && 'opacity-50')}>
-            {disabled && <LockClosedIcon className="w-3 h-3 mr-2" />}
+          <span
+            title="No column level lineage for python model"
```
vchan, line 168:
```suggestion
            title="No column level lineage for Python models"
```

vchan (approved):

PR #1477, opened 2023-09-25
Fix: Move columns and rendered query to separate tabs in Editor inspector

mykalmax:
Screenshot 2023-09-25 at 2 48 45 PM
Screenshot 2023-09-25 at 2 48 49 PM

### DIFF
--- a/web/client/src/library/components/editor/EditorInspector.tsx
+++ b/web/client/src/library/components/editor/EditorInspector.tsx
@@ -127,7 +127,7 @@ function InspectorModel({
             className="max-h-[15rem]"
             nodeId={model.name}
             columns={model.columns}
-            disabled={model?.type === 'python'}
+            disabled={model.type === 'python'}
             withHandles={false}
             withSource={false}
             withDescription={true}
--- a/web/client/src/library/components/editor/EditorInspector.tsx
+++ b/web/client/src/library/components/editor/EditorInspector.tsx
@@ -136,9 +136,7 @@ function InspectorModel({
         </Tab.Panel>
         <Tab.Panel
           unmount={false}
-          className={clsx(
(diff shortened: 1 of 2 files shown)


web/client/src/library/components/editor/EditorInspector.tsx
```diff
@@ -110,17 +120,37 @@ function InspectorModel({
...
+            withSource={false}
+            withDescription={true}
+            limit={10}
           />
         </Tab.Panel>
+        <Tab.Panel
+          unmount={false}
+          className={clsx(
+            'text-xs w-full h-full ring-white ring-opacity-60 ring-offset-2 ring-offset-blue-400 focus:outline-none focus:ring-2 p-2',
+          )}
```
vchan, line 141:
```suggestion
          className="text-xs w-full h-full ring-white ring-opacity-60 ring-offset-2 ring-offset-blue-400 focus:outline-none focus:ring-2 p-2"
```

web/client/src/library/components/editor/EditorInspector.tsx
```diff
@@ -110,17 +120,37 @@ function InspectorModel({
...
-            model={model}
-            withCode={false}
-            withModel={false}
-            withQuery={tab.file.isSQLMeshModelSQL}
-            withDescription={false}
+          <ModelColumns
+            className="max-h-[15rem]"
+            nodeId={model.name}
+            columns={model.columns}
+            disabled={model?.type === 'python'}
```
vchan, line 130: Do nodeId and columns need optional chaining on model too?

vchan (approved):

web/client/src/library/components/editor/EditorInspector.tsx
mykalmax: `model` should be present so `model?.type` here does not chaining

PR #1479, opened 2023-09-26
Fix: #1435 introduced invalid sql for querying the state sync

tobymao:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -889,24 +878,24 @@ def _snapshot_id_filter(
                 exp.Tuple,
                 exp.convert(
                     (
-                        exp.to_column(f"{fq_table_name}.name"),
```
izeigerman, line 892: So as far as I remember columns must be qualified in the filter because of of a join using `ON`. At least for MSSQL. @treysp please correct me if I'm wrong.

tobymao: they are qualified using aliases

PR #1482, opened 2023-09-26
Feat: Support custom session properties in the Spark Submit Airflow operator

izeigerman:


sqlmesh/schedulers/airflow/operators/spark_submit.py
```diff
@@ -77,24 +80,59 @@ def execute(self, context: Context) -> None:
                 payload_fd.write(command_payload)
 
             if self._hook is None:
+                if (
+                    isinstance(self._target, SnapshotEvaluationTarget)
+                    and self._target.snapshot.is_model
+                ):
+                    session_properties = self._target.snapshot.model.session_properties
+                    executor_cores: t.Optional[int] = session_properties.pop(  # type: ignore
```
tobymao, line 88: should we make this less restrictive and just pass in as is?

izeigerman: We pass the remainder of configuration as is after we extract the top-level fields.

PR #1484, opened 2023-09-26
Fix: Drop state tables during rollback when first migration fails

vchan:
When a project's first migration fails, state tables could be left over depending on when the failure occurs. This causes problems for future migrates. This PR drops all state tables if the first migration fails.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -725,16 +725,17 @@ class EngineAdapterStateSync(CommonStateSyncMixin, StateSync):
         """Rollback to the previous migration."""
         logger.info("Starting migration rollback.")
         tables = (self.snapshots_table, self.environments_table, self.versions_table)
-        if not any(self.engine_adapter.table_exists(f"{table}_backup") for table in tables):
-            raise SQLMeshError("There are no prior migrations to roll back to.")
-        for table in tables:
-            self._restore_table(table, _backup_table_name(table))
-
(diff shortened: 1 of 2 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -725,16 +725,17 @@ def rollback(self) -> None:
...
+        versions = self.get_versions(validate=False)
+        if versions.schema_version == 0:
+            # Clean up state tables
+            for table in tables + (self.seeds_table, self.intervals_table):
+                self.engine_adapter.drop_table(table)
+        else:
+            if not any(self.engine_adapter.table_exists(f"{table}_backup") for table in tables):
+                raise SQLMeshError("There are no prior migrations to roll back to.")
+            for table in tables + (self.seeds_table, self.intervals_table):
+                if self.engine_adapter.table_exists(_backup_table_name(table)):
```
izeigerman, line 737: The reason why `intervals` and `seeds` tables were handled separately is because they were added in later versions so we didn't expect backups to exist for them. Otherwise, checking table existence twice seems redundant.

vchan: Got it. I'll revert this then.

vchan: @izeigerman I changed line 734 from `any()` to `all()`. If some but not all the backups are there, then it's in a weird state and we shouldn't drop any tables.

tobymao (approved):

PR #1488, opened 2023-09-27
Feat: expose columns to types of parent models in MacroEvaluator

### LINKED ISSUE #1343: Access models in python models and macros
Currently the way to access a table is by doing `context.table(<table_name>)`. This assumes though that there is a database object that can be referenced like a table or a view. This doesn't work then for `embedded`. We should create a new method to return an expression which could either be a table reference or a subquery which would be the embedded logic. If we do this then we need to update our lineage parsing to support that method and consider what we want to do with `table` (continue to support it or make it private and not part of lineage).

GeorgeSittas:
Fixes #1343

This PR enables a SQLMesh user to access the columns-to-types mapping of upstream dependencies in macros and [python models that generate SQL](https://sqlmesh.readthedocs.io/en/latest/concepts/models/sql_models/#python-based-definition), i.e. through `MacroEvaluator`.

One use case for this feature is explained by @z3z1ma in https://github.com/TobikoData/sqlmesh/issues/1343#issuecomment-1701534252. It essentially makes writing maintainable and DRY code easier, by allowing one to leverage model metadata such as its columns' names to normalize them, apply necessary casts, compute common columns, apply masking to PII, etc.

### DIFF
--- a/examples/sushi/models/waiters.py
+++ b/examples/sushi/models/waiters.py
@@ -31,10 +31,10 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
     FROM sushi.orders AS o
     WHERE @incremental_by_ds(ds)
     """
-    blacklist = {"id", "customer_id", "start_ts", "end_ts"}
+    excluded = {"id", "customer_id", "start_ts", "end_ts"}
     projections = []
     for column, dtype in evaluator.columns_to_types("sushi.orders").items():
-        if column not in blacklist:
+        if column not in excluded:
             projections.append(f"{column}::{dtype}")
 
     return (
--- a/examples/sushi/models/waiters.py
+++ b/examples/sushi/models/waiters.py
@@ -12,6 +12,7 @@ from sqlmesh.core.model import EmbeddedKind
     kind=EmbeddedKind(),
     owner="jen",
     cron="@daily",
+    depends_on=["sushi.orders"],
 )
 def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
     """
--- a/examples/sushi/models/waiters.py
+++ b/examples/sushi/models/waiters.py
@@ -30,8 +30,14 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
     FROM sushi.orders AS o
     WHERE @incremental_by_ds(ds)
     """
+    blacklist = {"id", "customer_id", "start_ts", "end_ts"}
+    projections = []
+    for column, dtype in evaluator.columns_to_types("sushi.orders").items():
+        if column not in blacklist:
+            projections.append(f"{column}::{dtype}")
+
     return (
-        exp.select("waiter_id::int as waiter_id", "ds::text as ds")
+        exp.select(*projections)
         .from_("sushi.orders AS o")
         .where(incremental_by_ds(evaluator, exp.to_column("ds")))
         .distinct()
--- a/examples/sushi/models/waiters.py
+++ b/examples/sushi/models/waiters.py
@@ -12,7 +12,6 @@ from sqlmesh.core.model import EmbeddedKind
     kind=EmbeddedKind(),
     owner="jen",
     cron="@daily",
-    depends_on=["sushi.orders"],
 )
 def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
     """
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -13,13 +13,12 @@ from dataclasses import dataclass
 from pathlib import Path
 
 from sqlglot.errors import SchemaError, SqlglotError
-from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
 from sqlglot.schema import MappingSchema
 
 from sqlmesh.core import constants as c
 from sqlmesh.core.audit import Audit, load_multiple_audits
 from sqlmesh.core.dialect import parse, set_default_schema_and_catalog
-from sqlmesh.core.macros import SQLMESH_MOCKED_STAR, MacroRegistry, macro
+from sqlmesh.core.macros import MacroRegistry, macro
 from sqlmesh.core.metric import Metric, MetricMeta, expand_metrics, load_metric_ddl
 from sqlmesh.core.model import (
     Model,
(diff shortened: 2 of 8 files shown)


sqlmesh/core/loader.py
```diff
@@ -61,6 +67,23 @@ def update_model_schemas(
...
 
+def with_mocked_dependencies(loaded_model: Model) -> Model:
+    if isinstance(loaded_model, SqlModel):
+        # We mock dependencies so they can be referenced in macros before they're loaded
+        loaded_model._query_renderer._models = UniqueKeyDict(
+            "mocked_models",
+            {
+                name: _Model(  # type: ignore
+                    name=name,
+                    columns_to_types_={SQLMESH_MOCKED_COLUMN: exp.DataType.build("unknown")},
```
GeorgeSittas, line 78: Adding a single column-type pair here was necessary in order to avoid queries with no projections, since such queries are [deemed invalid](https://github.com/TobikoData/sqlmesh/blob/8b0e3cfd13cc0c6c181a2bc8e4c624ccc5a77940/sqlmesh/core/model/definition.py#L1034) and so we'd crash if we allowed them.

tests/core/test_model.py
```diff
@@ -1320,6 +1321,93 @@ def model2_entrypoint(evaluator: MacroEvaluator) -> str:
     )
 
 
+def test_reference_model_in_python_context(tmp_path, assert_exp_eq) -> None:
```
tobymao, line 1324: it's more robust and easier to maintain if this is just in sushi, can you add an example there?

tobymao: yea but as i said , testing in sushi would also be more robust, can you put a simpl eversion in sushi?

GeorgeSittas: So keep this and also update sushi with a simple example of this functionality?

sqlmesh/core/loader.py
```diff
@@ -39,6 +42,9 @@
     from sqlmesh.core.context import Context
 
 
+SQLMESH_MOCKED_COLUMN = "__SQLMESH_MOCKED_COLUMN__"
```
tobymao, line 45: maybe something like

__SQLMESH_MOCK_STAR__

GeorgeSittas: Ah so it'll generate a `*` instead of a random column name? That could be a good idea

GeorgeSittas: Ah actually I'm not sure if we can do that, because in `SqlModel`'s definition of `columns_to_types` we have:

```python
        if "*" in self._columns_to_types:
            return None
```

tobymao: no, just so that it says star

GeorgeSittas: Why would it say star?

sqlmesh/core/loader.py
```diff
@@ -61,6 +67,22 @@ def update_model_schemas(
             schema.add_table(name, columns_to_types, dialect=model.dialect)
 
 
+def with_mocked_dependencies(loaded_model: Model) -> Model:
+    if isinstance(loaded_model, SqlModel):
+        # We mock dependencies so they can be referenced in macros before they're loaded
+        loaded_model._query_renderer._models = UniqueKeyDict(
```
izeigerman, line 73: This looks like a major hack imho

sqlmesh/core/loader.py
```diff
@@ -280,6 +312,7 @@ def _load() -> Model:
                         physical_schema_override=config.physical_schema_override,
                         project=config.project,
                     )
+                    return with_mocked_dependencies(loaded_model)
```
izeigerman, line 315: So we're going to cache a model rendered with mocked dependencies? Doesn't this defeat the purpose of the caching?

sqlmesh/core/renderer.py
```diff
@@ -69,6 +71,7 @@ def __init__(
         self._only_execution_time = only_execution_time
 
         self._cache: t.Dict[t.Tuple[datetime, datetime, datetime], t.List[exp.Expression]] = {}
+        self._models: t.Optional[UniqueKeyDict[str, Model]] = None
```
izeigerman, line 74: I'm somewhat confused: who sets this when we render a model from a snapshot? I can only see this being set once during the load.

sqlmesh/core/loader.py
tobymao: as long as we always calculate the hash with mock it should be ok

examples/sushi/models/waiters.py
```diff
@@ -30,8 +31,14 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
     FROM sushi.orders AS o
     WHERE @incremental_by_ds(ds)
     """
+    blacklist = {"id", "customer_id", "start_ts", "end_ts"}
```
izeigerman, line 34: let's call this `excluded`

sqlmesh/core/model/definition.py
```diff
@@ -1026,6 +1027,14 @@ def update_schema(
         super().update_schema(
             schema, default_schema=default_schema, default_catalog=default_catalog
         )
+
+        mocked_star = normalize_identifiers(SQLMESH_MOCKED_STAR, dialect=self.dialect)
```
tobymao, line 1031: why do you need to normalize it? just have the mocked star be quoted?

tobymao (approved):

examples/sushi/models/waiters.py
```diff
@@ -30,8 +31,14 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
...
     """
+    excluded = {"id", "customer_id", "start_ts", "end_ts"}
+    projections = []
+    for column, dtype in evaluator.columns_to_types("sushi.orders").items():
+        if column not in excluded:
+            projections.append(f"{column}::{dtype}")
+
     return (
-        exp.select("waiter_id::int as waiter_id", "ds::text as ds")
+        exp.select(*projections)
```
tobymao, line 41: can you add a test for render here?

sqlmesh/core/model/definition.py
GeorgeSittas: It doesn't work for all dialects - Redshift for example doesn't care about quotes.

examples/sushi/models/waiters.py
GeorgeSittas: yeah

GeorgeSittas: Added

PR #1491, opened 2023-09-28
Fix: SCD Type 2 Replace Query Mixin

eakmanrq:
Previously was doing any on an iterator and also applying a transformation. What this meant if that if we got a match and there were further matches then those additional matches didn't get the transform applied to them. This could come up if using SCD Type 2 on certain engines like Postgres.

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -90,19 +90,19 @@ class LogicalReplaceQueryMixin(EngineAdapter):
             with source_queries[0] as query:
                 target_table = exp.to_table(table_name)
                 # Check if self-referencing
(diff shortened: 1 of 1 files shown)


sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -90,19 +90,19 @@ def replace_query(
...
                 target_table = exp.to_table(table_name)
                 # Check if self-referencing
-                if any(
+                matching_tables = [
                     table
                     for table in query.find_all(exp.Table)
                     if quote_identifiers(table) == quote_identifiers(target_table)
-                ):
+                ]
+                if any(matching_tables):
```
tobymao, line 98: if matchig_tables

PR #1495, opened 2023-09-28
Feat: allow default schemas and categories

tobymao:


sqlmesh/core/selector.py
```diff
@@ -12,25 +11,21 @@
...
+    from sqlmesh.core.context import Context
+
 
 class Selector:
     def __init__(
         self,
         state_reader: StateReader,
         models: UniqueKeyDict[str, Model],
-        context_path: Path = Path("."),
+        context: Context,
```
tobymao, line 23: @izeigerman i know this will upset you, but get the right config per model is not trivial which is needed for update_model_schemas

izeigerman: I'm absolutely against this :)

izeigerman: We should then refactor how we get a config per model.

sqlmesh/core/loader.py
```diff
@@ -41,10 +41,10 @@
 
 # TODO: consider moving this to context
 def update_model_schemas(
-    dag: DAG[str], models: UniqueKeyDict[str, Model], context_path: Path
+    dag: DAG[str], models: UniqueKeyDict[str, Model], context: Context
```
izeigerman, line 44: I'm against this as well

izeigerman: Why can't we just pass a map from model names to model defaults?

izeigerman: Or better yet a map from model name to a pair of default schema / catalog

sqlmesh/core/selector.py
izeigerman: Made a better suggestion in the other thread

sqlmesh/core/loader.py
tobymao: it's a hassle to build this

izeigerman (approved):

PR #1500, opened 2023-09-29
Fix: treat Snowflake's staged file paths as MacroVars

GeorgeSittas:
This fixes an issue inadvertently introduced in https://github.com/tobymao/sqlglot/commit/8af4054d4e96b62309b1543d51548728bdba520f (later updated [here](https://github.com/tobymao/sqlglot/commit/17e39d04916e3e406307ed8922cb2597b4a6998a)). Snowflake allows `@foo` to appear in `FROM` clauses, which clashes with our macro variable syntax and hence the parsing was off, causing errors in e.g. audits because they reference `@this_model` in the `FROM` clause.

Verified that with this branch I can successfully run `sqlmesh plan` for the example project using the Snowflake dialect.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -10,6 +10,7 @@ from enum import Enum, auto
 import pandas as pd
 from sqlglot import Dialect, Generator, Parser, Tokenizer, TokenType, exp
 from sqlglot.dialects.dialect import DialectType
+from sqlglot.dialects.snowflake import Snowflake
 from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
 from sqlglot.optimizer.scope import traverse_scope
 from sqlglot.tokens import Token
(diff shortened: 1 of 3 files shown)


sqlmesh/core/dialect.py
```diff
@@ -89,6 +90,10 @@ def output_name(self) -> str:
         return self.this.name
 
 
+class StagedFilePath(exp.Table):
+    """Represents paths to "staged files" in Snowflake."""
+
```
GeorgeSittas, line 95: I decided not to pollute SQLGlot with this definition on purpose: the issue caused by `@` is SQLMesh-specific, so no need to subtype `Table` and introduce unnecessary complexity upstream.

tests/core/test_macros.py
```diff
@@ -76,6 +78,17 @@ def test_macro_var(macro_evaluator):
         macro_evaluator.locals = {"x": k}
         assert macro_evaluator.transform(expression).sql() == v
 
+    # Check that macro vars are rendered as-is if they don't map to something in locals
+    e = parse_one("select * from @staged_table", dialect="snowflake")
```
izeigerman, line 82: Can you please also add a test for snowflake where the macro syntax is used in eg. projection

izeigerman (approved):

tests/core/test_macros.py
GeorgeSittas: Sure, good point

PR #1503, opened 2023-09-30
Fix: Ensure Unit Tests Data Types Match

eakmanrq:
The expected values for a unit tests are expressed in YAML where it is difficult to express expected data types. This fix casts the types provided in the YAML to the types returned by the engine adapter which would match the actual types returned from the model.


izeigerman (approved):

izeigerman:
A test might be helpful

PR #1505, opened 2023-10-02
Fix: don't treat the PARAMETER token as a MacroVar start if it's not @

GeorgeSittas:


sqlmesh/core/dialect.py
```diff
@@ -147,6 +147,9 @@ def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expressio
 
 
 def _parse_macro(self: Parser, keyword_macro: str = "") -> t.Optional[exp.Expression]:
+    if self._prev.text != "@":
+        return self._parse_parameter()
```
GeorgeSittas, line 151: @izeigerman @tobymao should we define a top-level constant to represent the macro prefix? Then we could reuse it in some places where we've hardcoded `@`.

tobymao: ok

GeorgeSittas: Done

izeigerman (approved):

PR #1510, opened 2023-10-02
Feat: support flask v1 via using route

pingzh:
this will allow users to use flask v1


tobymao (approved):

izeigerman:
@pingzh can you please run `make style` to fix the formatting? Good to go otherwise

PR #1514, opened 2023-10-03
Fix!: In Airflow only generate cadence DAGs for snapshots promoted to prod

izeigerman:


tobymao:
we weren't doing this already?

tobymao (approved):

izeigerman:
> we weren't doing this already?

@tobymao  I was surprised too!

PR #1520, opened 2023-10-04
Fix(MSSQL): use autocommit 

treysp:
Concurrent CREATE TABLE statements caused a lock because pymssql always opens transactions. Autocommitting prevents this lock.

https://github.com/pymssql/pymssql/issues/460


eakmanrq:
So the tradeoff here is now we don't control transactions at all anymore since everything would be auto committed. Ideally, since MSSQL supports transactions, our transaction implementation could have continued to work and we could leave this False. 

If we continue with this then I think we should set `supports_transactions` to return False for MSSQL since we effectively disable them with this change.

PR #1521, opened 2023-10-04
Fix(MSSQL): drop views in cascading DROP SCHEMA

treysp:


sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -127,13 +127,18 @@ def drop_schema(
...
-            # with a `drop_table` call.
             objects = self._get_data_objects(schema_name)
             for obj in objects:
-                self.drop_table(
-                    ".".join([obj.catalog, obj.schema_name, obj.name]), exists=ignore_if_not_exists  # type: ignore
-                )
+                if obj.type == DataObjectType.VIEW:
+                    # In MSSQL you can't provide a catalog to DROP VIEW
+                    # https://stackoverflow.com/questions/32828034/sql-server-2008-r2-create-alter-view-does-not-allow-specifying-the-database-n
+                    self.drop_view(
```
eakmanrq, line 135: If we just drop the catalog then we could be making an ambiguous call here if two views exists with same schema name and table name but different catalogs. In the SO link they have this:
```
USE <catalog>
GO

DROP VIEW ...
```
Should we do something like that here?

eakmanrq:
Can you update the integration test to identify this issue? You might be able to reuse an existing test where we create a schema that contains views and just call drop schema on that.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -54,7 +53,9 @@ def columns(
 
         table = exp.to_table(table_name)
 
-        catalog_name = table.catalog or self.DEFAULT_CATALOG_NAME
+        catalog_name = table.catalog
+        if not catalog_name:
+            catalog_name = self.fetchone("select DB_NAME()")[0]
```
tobymao, line 58: make this a method

catalog_name = self.catalog(table)

eakmanrq (approved):

PR #1523, opened 2023-10-04
Fix: Improve Relative Date Categorical Expressions

eakmanrq:
Prior to this change if I said I wanted a plan to run for 1 day ago it would effectively be a no-op since 1 day ago has time and since we don't want to load a partial we move the date up and therefore it is 0 days ago. Really though when I said "1 day ago" I didn't mean "I day ago at this same time" but rather "1 day ignoring time". Therefore we set this now to midnight and there is no longer a partial.


sqlmesh/utils/date.py
```diff
@@ -279,3 +283,16 @@ def time_like_to_str(time_like: TimeLike) -> str:
...
         return to_ds(time_like)
     return to_ts(time_like)
+
+
+def is_catagorical_relative_expression(expression: str) -> bool:
+    # The Freshness Date Data Parser doesn't support plural units so we add the `s?` to the expression
+    freshness_date_parser.PATTERN = re.compile(
+        r"(\d+[.,]?\d*)\s*(%s)s?\b" % freshness_date_parser._UNITS, re.I | re.S | re.U  # type: ignore
+    )
+    if str(expression).lower() in {"today", "yesterday", "tomorrow"}:
```
izeigerman, line 293: here's another one
```
In [1]: import dateparser

In [2]: dateparser.parse("the day before yesterday")
Out[2]: datetime.datetime(2023, 10, 2, 13, 10, 10, 485354)
```

izeigerman: We should `.strip()` the string too.

sqlmesh/utils/date.py
```diff
@@ -146,8 +158,8 @@ def to_datetime(value: TimeLike, relative_base: t.Optional[datetime] = None) ->
     if dt is None:
         raise ValueError(f"Could not convert `{value}` to datetime.")
 
-    if dt.tzinfo:
```
tobymao, line 149: revert this back

sqlmesh/utils/date.py
```diff
@@ -279,3 +291,12 @@ def time_like_to_str(time_like: TimeLike) -> str:
     if is_date(time_like):
         return to_ds(time_like)
     return to_ts(time_like)
+
+
+def is_catagorical_relative_expression(expression: str) -> bool:
+    if str(expression).lower() in {"today", "yesterday", "tomorrow"}:
```
tobymao, line 297: you don't need to call str, it is a str

tobymao, line 297: make the set a constant

sqlmesh/utils/date.py
```diff
@@ -279,3 +291,12 @@ def time_like_to_str(time_like: TimeLike) -> str:
...
+        return True
+    grain_kwargs = freshness_date_parser.FreshnessDateDataParser().get_kwargs(expression)
+    if not grain_kwargs:
+        return False
+    return not any(k in {"hours", "minutes", "seconds"} for k in grain_kwargs.keys())
```
tobymao, line 302: get rid of keys, make the hours a constant

sqlmesh/utils/date.py
```diff
@@ -279,3 +291,12 @@ def time_like_to_str(time_like: TimeLike) -> str:
     if is_date(time_like):
         return to_ds(time_like)
     return to_ts(time_like)
+
+
+def is_catagorical_relative_expression(expression: str) -> bool:
+    if str(expression).lower() in {"today", "yesterday", "tomorrow"}:
+        return True
+    grain_kwargs = freshness_date_parser.FreshnessDateDataParser().get_kwargs(expression)
```
tobymao, line 299: is there state on freshnessdateparser? make that a constant too

tobymao (approved):

PR #1524, opened 2023-10-04
Fix MWAA client's set_variable

max-winderbaum:
We need to escape quotes and escape sequences within the value of this function so Airflow can recognize the string

Additionally, when we try to quote (and escape) with single quotes, Airflow returns a 500 error. We guess this is because Airflow is using escaped single quotes as part of its internal functioning

### DIFF
--- a/sqlmesh/schedulers/airflow/mwaa_client.py
+++ b/sqlmesh/schedulers/airflow/mwaa_client.py
@@ -31,7 +31,7 @@ class MWAAClient(BaseAirflowClient):
         Returns:
             A tuple of stdout and stderr from the MWAA CLI.
         """
-        value=value.replace('\\', '\\\\').replace('"', '\\"')
+        value = value.replace('\\', '\\\\').replace('"', '\\"')
         return self._post(f'variables set {key} "{value}"')
 
     def get_first_dag_run_id(self, dag_id: str) -> t.Optional[str]:
--- a/sqlmesh/schedulers/airflow/mwaa_client.py
+++ b/sqlmesh/schedulers/airflow/mwaa_client.py
@@ -31,7 +31,7 @@ class MWAAClient(BaseAirflowClient):
         Returns:
             A tuple of stdout and stderr from the MWAA CLI.
         """
-        value = value.replace('\\', '\\\\').replace('"', '\\"')
+        value = value.replace("\\", "\\\\").replace('"', '\\"')
         return self._post(f'variables set {key} "{value}"')
 
     def get_first_dag_run_id(self, dag_id: str) -> t.Optional[str]:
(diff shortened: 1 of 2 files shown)


izeigerman (approved):
@max-winderbaum Thanks a lot! Can you please run `make style` and sign the CLA agreement. Good to go otherwise!

sqlmesh/schedulers/airflow/mwaa_client.py
```diff
@@ -31,7 +31,8 @@ def set_variable(self, key: str, value: str) -> t.Tuple[str, str]:
         Returns:
             A tuple of stdout and stderr from the MWAA CLI.
         """
-        return self._post(f"variables set {key} '{value}'")
+        value=value.replace('\\', '\\\\').replace('"', '\\"')
```
tobymao, line 34: what encoding is this? why do you have to do this manually? should you just call json on it?

tobymao: so value = json.dumps(value)

izeigerman: That's actually a great point! So it can be simplified to:
```
return self._post(f"variables set {key} {json.dumps(value)}")
```
Good catch 👍

izeigerman: Reference example confirming it works as expected:
```
In [8]: value = "{\"test\": \"value\n\"}"

In [9]: print(f"variables set test_key {json.dumps(value)}")
variables set test_key "{\"test\": \"value\n\"}"
```

izeigerman: @tobymao on the other hand I'm not sure this is enough. We still need to escape `\`, which `json.dumps` doesn't take care of.

tobymao: we'll handle it from, thanks for the PR!

max-winderbaum: @tobymao Does `we'll handle it from here` mean to close this PR and let you open a new one?

tobymao: no, @izeigerman will merge this and explore

izeigerman, line 34:
```suggestion
        value = value.replace('\\', '\\\\').replace('"', '\\"')
```

sqlmesh/schedulers/airflow/mwaa_client.py
```diff
@@ -31,7 +31,8 @@ def set_variable(self, key: str, value: str) -> t.Tuple[str, str]:
         Returns:
             A tuple of stdout and stderr from the MWAA CLI.
         """
-        return self._post(f"variables set {key} '{value}'")
+        value = value.replace('\\', '\\\\').replace('"', '\\"')
```
izeigerman, line 34:
```suggestion
        value = value.replace("\\", "\\\\").replace('"', '\\"')
```

PR #1526, opened 2023-10-04
Fix: Set updated_at when updating a snapshot record in state

izeigerman:


tobymao (approved):

sqlmesh/core/state_sync/common.py
```diff
@@ -145,11 +145,18 @@ def promote(self, environment: Environment, no_gaps: bool = False) -> PromotionR
...
 
         self._update_environment(environment)
+
         removed = (
             list(existing_table_infos.values())
             if environment_suffix_target_changed
             else [existing_table_infos[name] for name in missing_models]
         )
+        if removed and not environment_suffix_target_changed:
+            removed_snapshots = self._get_snapshots(removed).values()
```
tobymao, line 155: just in line the for loop

tobymao (approved):

PR #1529, opened 2023-10-05
Fix: Property Override Pydantic 2

eakmanrq:
Repro (Pydantic 2 only) showing the issue:
```
class Parent:
    @property
    def _test_prop(self) -> str:
        return "original_val"

    @property
    def test_prop(self) -> str:
        return self._test_prop


class Child(Parent):
    _test_prop = "new_val"


assert Child().test_prop == "new_val"


class ParentPydantic(PydanticModel):
    @property
    def _test_prop(self) -> str:
        return "original_val"

    @property
    def test_prop(self) -> str:
        return self._test_prop


class ChildPydantic(ParentPydantic):
    _test_prop = "new_val"


assert ChildPydantic().test_prop == "original_val"
```

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -676,7 +676,10 @@ class MySQLConnectionConfig(ConnectionConfig):
 
     type_: Literal["mysql"] = Field(alias="type", default="mysql")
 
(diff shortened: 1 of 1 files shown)


tobymao (approved):
better practice anyways, prev version smells

PR #1531, opened 2023-10-05
Fix!: use dialect when generating types for mapping schema

GeorgeSittas:


sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
```diff
@@ -0,0 +1,68 @@
...
+        if not schema:
+            return schema
+
+        for k, v in schema.items():
+            if isinstance(v, dict):
+                _convert_schema_types(v)
+            else:
+                schema[k] = parse_one(v).sql(dialect=engine_adapter.dialect)
+
+        return schema
```
GeorgeSittas, line 27: This is the interesting part @tobymao

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
```diff
@@ -0,0 +1,68 @@
...
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    def _convert_schema_types(schema):
+        if not schema:
+            return schema
+
+        for k, v in schema.items():
+            if isinstance(v, dict):
+                _convert_schema_types(v)
```
tobymao, line 23: why is this nested?

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
```diff
@@ -0,0 +1,68 @@
...
+            else:
+                schema[k] = parse_one(v).sql(dialect=engine_adapter.dialect)
```
tobymao, line 25: shouldn't this be the model dialect?

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
GeorgeSittas: as in "why it's defined as a closure" or the recursion part?

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
GeorgeSittas: right, my bad

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
tobymao: recursion

GeorgeSittas: Since the schemas are nested and we need to update their leaves, this seemed like a simple way to do it, considering that the nesting levels shouldn't be deeper than 2-3. Do you think this is problematic?

tobymao (approved):

sqlmesh/migrations/v0028_generate_schema_types_using_dialect.py
tanghyd: FYI the [maximum nesting level in BigQuery is 15](https://cloud.google.com/bigquery/docs/nested-repeated) - I can't speak for other engines however.

GeorgeSittas: I think that nesting takes into account records / structs, right? We don't nest these IIRC, they're simply represented as strings like `STRUCT<foo STRUCT<... >>`, so it should be fine.

PR #1532, opened 2023-10-05
Fix(mssql): execute within transactions to prevent hanging

treysp:
The pymssql library [always begins transactions](https://github.com/pymssql/pymssql/issues/460) - "pymssql opens transaction on connection start with 'BEGIN TRAN' and after conn.commit() it commits already opened transaction and then open new one again."

In some cases we did not explicitly close them, leading to hung sessions.

The fix wraps calls to `cursor.execute()` in a transaction **in the base engine_adapter**, ensuring they are properly closed. 

Some engines do not support transactions or handle them idiosyncratically, such that this approach does not work. Engine adapters are now given a flag that determines whether their executions occur within transactions. 

At this time, BigQuery, DuckDB, and Spark have the flag set to False.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -948,7 +948,7 @@ class EngineAdapterStateSync(CommonStateSyncMixin, StateSync):
             )
 
     @contextlib.contextmanager
-    def _transaction(self) -> t.Generator[None, None, None]:
+    def _transaction(self) -> t.Iterator[None]:
         with self.engine_adapter.transaction():
             yield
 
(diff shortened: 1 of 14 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -948,8 +948,8 @@ def _snapshot_name_version_filter(
             )
 
     @contextlib.contextmanager
-    def _transaction(self, transaction_type: TransactionType) -> t.Generator[None, None, None]:
-        with self.engine_adapter.transaction(transaction_type=transaction_type):
+    def _transaction(self) -> t.Generator[None, None, None]:
```
tobymao, line 951: -> t.Iterator[None]

tobymao (approved):

PR #1533, opened 2023-10-05
Feat: Use internal state instead of Airflow Variables to store Plan DAG specs

izeigerman:
MWAA relies on shell / bash to set variables in Airflow and bash has limit on a number of characters that can be passed in. 

With this update we no longer rely on Airflow to store our Plan Application DAG specs and instead use our own internal state.  The additional `_plan_dags` table is added to our state to support this.

### DIFF
--- a/sqlmesh/schedulers/airflow/plan.py
+++ b/sqlmesh/schedulers/airflow/plan.py
@@ -2,14 +2,84 @@ from __future__ import annotations
 
 import typing as t
 
+import pandas as pd
+from sqlglot import exp
+
 from sqlmesh.core import scheduler
+from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.environment import Environment
 from sqlmesh.core.plan import can_evaluate_before_promote
 from sqlmesh.core.snapshot import SnapshotTableInfo
-from sqlmesh.core.state_sync import StateSync
+from sqlmesh.core.state_sync import EngineAdapterStateSync, StateSync
+from sqlmesh.core.state_sync.base import DelegatingStateSync
 from sqlmesh.schedulers.airflow import common
+from sqlmesh.utils import nullsafe_join
 from sqlmesh.utils.date import now, now_timestamp
 from sqlmesh.utils.errors import SQLMeshError
+
+
+class PlanDagState:
+    def __init__(self, engine_adapter: EngineAdapter, schema: t.Optional[str]):
+        self.engine_adapter = engine_adapter
+
+        self._plan_dags_table = nullsafe_join(".", schema, "_plan_dags")
+        self._plan_dag_columns_to_types = {
+            "request_id": exp.DataType.build("text"),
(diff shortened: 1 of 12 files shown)


sqlmesh/schedulers/airflow/plan.py
```diff
@@ -2,16 +2,86 @@
...
+        """Returns all DAG specs in the state."""
+        query = exp.select("dag_spec").from_(self._plan_dags_table)
+        return [
+            common.PlanDagSpec.parse_raw(row[0])
+            for row in self.engine_adapter.fetchall(
+                query, ignore_unsupported_errors=True, quote_identifiers=True
+            )
+        ]
+
+    def delete_dag_specs(self, dag_ids: t.Iterable[str]) -> None:
```
tobymao, line 74: make this a Sequence or collection, then you won't have to convert it into a list

sqlmesh/schedulers/airflow/plan.py
```diff
@@ -2,16 +2,86 @@
...
+            columns_to_types=self._plan_dag_columns_to_types,
+            contains_json=True,
+        )
+
+    def get_dag_specs(self) -> t.List[common.PlanDagSpec]:
+        """Returns all DAG specs in the state."""
+        query = exp.select("dag_spec").from_(self._plan_dags_table)
+        return [
+            common.PlanDagSpec.parse_raw(row[0])
+            for row in self.engine_adapter.fetchall(
```
tobymao, line 69: i don't think we should hit the engine adapter raw like, this, add an api to engine adapter for dag_spec querying?

sqlmesh/schedulers/airflow/plan.py
```diff
@@ -2,16 +2,86 @@
...
+from sqlmesh.core.state_sync import EngineAdapterStateSync, StateSync
+from sqlmesh.core.state_sync.base import DelegatingStateSync
 from sqlmesh.schedulers.airflow import common
+from sqlmesh.utils import nullsafe_join
 from sqlmesh.utils.date import now, now_timestamp
 from sqlmesh.utils.errors import SQLMeshError
 
 
+class PlanDagState:
+    def __init__(self, engine_adapter: EngineAdapter, schema: t.Optional[str]):
```
tobymao, line 22: add a catalog here too cc @eakmanrq

sqlmesh/schedulers/airflow/plan.py
tobymao: ah i see, this is an engine adapter proxy?

izeigerman: Yes, it's like StateSync but only for Plan DAGs

sqlmesh/schedulers/airflow/plan.py
izeigerman: we currently don't support it for `StateSync` either.

PR #1536, opened 2023-10-06
Feat: Support sensors for external tables in Airflow

izeigerman:
Docs will be updated in a follow-up PR.

### DIFF
--- a/sqlmesh/schedulers/airflow/operators/hwm_sensor.py
+++ b/sqlmesh/schedulers/airflow/operators/hwm_sensor.py
@@ -3,12 +3,12 @@ import typing as t
 from datetime import datetime
 
 from airflow.models import DagRun
-from airflow.sensors.base import BaseSensorOperator
+from airflow.sensors.base import BaseSensorOperator, PokeReturnValue
 from airflow.utils.context import Context
 
 from sqlmesh.core.snapshot import Snapshot, SnapshotTableInfo
 from sqlmesh.schedulers.airflow import util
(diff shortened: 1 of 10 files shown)


sqlmesh/schedulers/airflow/operators/hwm_sensor.py
```diff
@@ -63,3 +63,33 @@ def _compute_target_high_water_mark(
...
+        )
+        self.snapshot = snapshot
+        self.external_table_sensor_factory = external_table_sensor_factory
+
+    def poke(self, context: Context) -> t.Union[bool, PokeReturnValue]:
+        dag_run = context["dag_run"]
+        signals = self.snapshot.model.render_signals(
+            start=to_datetime(dag_run.data_interval_start),
+            end=to_datetime(dag_run.data_interval_end),
+            execution_time=now(),
```
tobymao, line 92: now or now_timestamp?

tobymao (approved):

sqlmesh/schedulers/airflow/operators/hwm_sensor.py
izeigerman: All other arguments are datetime instances, so I figured I should be consistent.

tobymao: that's fine, just understand that it's minute floored

izeigerman: oh, nice gotcha 😢

PR #1538, opened 2023-10-06
Fix: only include model snapshots in the version creation progress

GeorgeSittas:


sqlmesh/core/plan/evaluator.py
```diff
@@ -136,7 +136,7 @@ def _push(self, plan: Plan) -> None:
         snapshot_id_to_snapshot = {s.snapshot_id: s for s in plan.snapshots}
 
         if plan.new_snapshots:
-            self.console.start_creation_progress(len(plan.new_snapshots))
+            self.console.start_creation_progress(len([s for s in plan.new_snapshots if s.is_model]))
```
eakmanrq, line 139: What if the plan only had new snapshots that were not models? Would we print 0 here?

eakmanrq (commented):
It would be nice if we could get a test written to capture this expected behavior. I haven't worked on this part of the code in a while so if the cost of that seems to high then feel free to skip. I could just see this being easily reverted in the future.

sqlmesh/core/plan/evaluator.py
GeorgeSittas: Good call, I'll need to check that but it might be the case yes

GeorgeSittas: > What if the plan only had new snapshots that were not models?

@eakmanrq how can I reach this state to test the behavior? E.g. run plan and then add an audit and re-run plan?

GeorgeSittas: > What if the plan only had new snapshots that were not models? Would we print 0 here?

@eakmanrq I assume the intended behavior then would be to not display a progress bar at all, correct? That is, we don't version non-model snapshots. If that's what you meant then I believe it should be addressed with the latest change I made.

tobymao (approved):

PR #1539, opened 2023-10-06
Docs: move quickstart project info into interface-specific docs

treysp:
This PR streamlines the quickstart flow by:
1. Moving the example project exposition into collapsible sections of each API-specific doc
2. Altering the quickstart left menu page order to reflect the linear flow among pages
3. Expanding each page's "next steps" to reinforce flow

The PR also removes the main menu "API" link to the Python pdocs and provides additional information on the Python reference page that links to the API docs.


tobymao:
make sure that the base url that we link everywhere still works

treysp:
I intentionally kept that base URL page `/quick_start` in place so it would continue to work, and as far as I can tell it will.

Here's the base URL: https://sqlmesh.readthedocs.io/en/stable/quick_start/
Here's the local URL for updated version: http://127.0.0.1:8000/quick_start/

tobymao (approved):

PR #1540, opened 2023-10-07
Fix!: to_datetime now only accepts millis and not epoch seconds

tobymao:
this solves an issue where dates between 1970 and 1972 are ambiguous.

### DIFF
--- a/sqlmesh/utils/cron.py
+++ b/sqlmesh/utils/cron.py
@@ -27,16 +27,16 @@ class CroniterCache:
         to substitute for the cron call.
         """
         if self._interval_seconds is None:
-            seconds = set()
+            deltas = set()
             curr = self.curr
 
             for _ in range(self.ESTIMATE_SAMPLES_NUM):
                 prev = curr
-                curr = to_datetime(croniter(self.cron, curr).get_next())
-                seconds.add(curr - prev)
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/cron.py
```diff
@@ -32,7 +32,7 @@ def interval_seconds(self) -> int:
 
             for _ in range(self.ESTIMATE_SAMPLES_NUM):
                 prev = curr
-                curr = to_datetime(croniter(self.cron, curr).get_next())
+                curr = to_datetime(croniter(self.cron, curr).get_next() * 1000)
                 seconds.add(curr - prev)
```
izeigerman, line 36: I don't get it: why is this called seconds?

tests/utils/test_date.py
```diff
@@ -19,9 +19,6 @@ def test_to_datetime() -> None:
     assert to_datetime("2020-01-01T05:00:00+05:00") == target
     assert to_datetime("2020-01-01") == target
     assert to_datetime(datetime(2020, 1, 1)) == target
-    assert to_datetime("1577836800") == target
-    assert to_datetime(1577836800) == target
-    assert to_datetime(1577836800.0) == target
```
izeigerman, line 24: Can we add a test for 1970-01-01 - 1972-... range instead?

izeigerman (approved):

PR #1549, opened 2023-10-10
Fix: fix tests due to sqlglot upgrade

tobymao:


sqlmesh/core/renderer.py
```diff
@@ -417,7 +417,13 @@ def _optimize_query(self, query: exp.Subqueryable) -> exp.Subqueryable:
...
                             "",
                         ):
-                            select.replace(exp.alias_(select, select.output_name))
+                            alias = exp.alias_(select, select.output_name)
+                            comments = alias.this.comments
+                            if comments:
+                                alias.add_comments(comments)
+                                comments.clear()
+
+                            select.replace(alias)
```
GeorgeSittas, line 426: Not sure why these changes are needed. What was the issue? I put a breakpoint in the failing test and I couldn't spot a difference in the ASTs so it's weird the comment was generated before the comma.

GeorgeSittas: how did the test work before though?

tobymao: because it compares the ast only

GeorgeSittas: Good catch

PR #1552, opened 2023-10-10
Fix: Automatically restamp model when reverting from forward-only back to breaking / non-breaking

izeigerman:


sqlmesh/core/context.py
```diff
@@ -1255,27 +1259,45 @@ def _snapshots(
...
+                    cache=fingerprint_cache,
+                    ttl=ttl,
+                )
+                snapshots[node.name] = snapshot
+            return snapshots
 
+        snapshots = _nodes_to_snapshots(nodes)
         stored_snapshots = self.state_reader.get_snapshots(snapshots.values())
 
+        non_revertible_snapshot_names = {
```
tobymao, line 1287: irrevertible

sqlmesh/core/context.py
```diff
@@ -1255,27 +1259,45 @@ def _snapshots(
...
+            s.name for s in stored_snapshots.values() if s.name in local_nodes and s.non_revertible
+        }
+        if non_revertible_snapshot_names:
+            for name in non_revertible_snapshot_names:
+                snapshot_id = snapshots[name].snapshot_id
+                logger.info(
+                    "Found a non-revertible snapshot %s. Restamping the model...", snapshot_id
+                )
+                node = local_nodes[name]
+                nodes[name] = node.copy(update={"stamp": f"revert to {snapshot_id.identifier}"})
```
tobymao, line 1297: what does identifier str() turn into?

izeigerman: `identifier` is a string

sqlmesh/core/snapshot/definition.py
```diff
@@ -392,6 +392,8 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
             this snapshot is evaluated on a recurring basis. None indicates that this snapshot is paused.
         effective_from: The timestamp which indicates when this snapshot should be considered effective.
             Applicable for forward-only snapshots only.
+        migrated: Whether or not this snapshot has been created as a result of migration.
+        non_revertible: Whether or not this snapshot can be used to revert its model to a previous version.
```
tobymao, line 396: change all to irrevertible because it's like irreversible

sqlmesh/core/context.py
izeigerman: Per Chat GPT:
> Both "non-revertible" and "irrevertible" are acceptable terms, and they can be used interchangeably to convey the idea that something cannot be reverted or undone. You can choose the one that fits the context and your personal preference.

sqlmesh/core/context.py
izeigerman: it's a hash string

tobymao: perfect, wanted to make sure it's not like Identifier(id: ...)

eakmanrq (approved):

tobymao (approved):

PR #1554, opened 2023-10-10
Chore: refactor the detection of SQLMESH_MOCKED_STAR

GeorgeSittas:
Apparently this part:

```python
if mocked_star.name in (self.columns_to_types or {}):
```

was causing `self.columns_to_types` to render the query in `optimize=True` mode and so we would get a warning pre-evaluation time for the sushi project because the mocked column was unresolved. I think this PR makes it a little bit cleaner, it mimics what we do to detect actual star projections.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -17,7 +17,6 @@ from pydantic import Field
 from sqlglot import diff, exp
 from sqlglot.diff import Insert, Keep
 from sqlglot.helper import ensure_list
-from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
 from sqlglot.schema import MappingSchema, nested_set
 from sqlglot.time import format_time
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1028,8 +1027,10 @@ def update_schema(
             schema, default_schema=default_schema, default_catalog=default_catalog
         )
 
-        mocked_star = normalize_identifiers(SQLMESH_MOCKED_STAR, dialect=self.dialect)
-        if mocked_star.name in (self.columns_to_types or {}):
+        query = self._query_renderer.render(optimize=False)
+        if isinstance(query, exp.Expression) and any(
+            select.output_name.upper() == SQLMESH_MOCKED_STAR for select in query.selects
```
tobymao, line 1032: SQLMESH_MOCKED_STAR in query.named_selects

GeorgeSittas: Way better, thanks

GeorgeSittas: Ah haha this is actually problematic because `named_selects` may have lowercase or uppercase names depending on the dialect and won't match the `SQLMESH_MOCKED_STAR`. :)

tobymao: what if we just made the mock something that's not a text, like __

GeorgeSittas: Hm not sure, it feels like a hack and looks a bit weird. It may also make it harder to search for in the codebase later if needed.

We _could_ though use the `named_selects` you suggested as:
```diff
-        if isinstance(query, exp.Expression) and any(
-            select.output_name.upper() == SQLMESH_MOCKED_STAR for select in query.selects
+        if isinstance(query, exp.Subqueryable) and any(
+            name.upper() == SQLMESH_MOCKED_STAR for name in query.named_selects
```

tobymao: sounds good

tobymao (approved):

PR #1555, opened 2023-10-10
Fix(MSSQL): fix results fetching

treysp:
Background:
- pymssql automatically opens transactions, causing hung sessions on some CREATE statements
- #1532 fixed this by wrapping all `execute` calls in a transaction to ensure transactions would be closed

Current issue:
- The transaction wrapping in #1532 closes the pymssql transaction after executing the query, such that query results are not available to be fetched by the `fetchone` or `fetchall` methods

This PR:
- Creates a new MSSQL method `execute_and_fetch` that both executes a query and fetches its results. The method copies the base engine adapter `execute` code because we must fetch the results from inside the transaction context manager.
- Overrides the base engine adapter `fetchone` and `fetchall` methods.

### DIFF
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -190,13 +190,13 @@ class MSSQLEngineAdapter(
             FROM {catalog_name}.INFORMATION_SCHEMA.TABLES
             WHERE TABLE_SCHEMA LIKE '%{schema_name}%'
         """
-        results = self.fetchall(query)
+        results = self.fetchall(query, quote_identifiers=True)
         return [
             DataObject(
-                catalog=row[0],  # type: ignore
-                schema=row[2],  # type: ignore
-                name=row[1],  # type: ignore
-                type=DataObjectType.from_str(row[3]),  # type: ignore
+                catalog=row[0],
+                schema=row[2],
+                name=row[1],
+                type=DataObjectType.from_str(row[3]),
             )
             for row in results
         ]
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -5,13 +5,14 @@ from __future__ import annotations
 
 import logging
 import typing as t
-from typing import Literal, overload
+from typing import overload
 
 import pandas as pd
 from pandas.api.types import is_datetime64_dtype  # type: ignore
 from sqlglot import exp
 from sqlglot.errors import ErrorLevel
 from sqlglot.optimizer.qualify_columns import quote_identifiers
+from typing_extensions import Literal
 
 from sqlmesh.core.engine_adapter.base import EngineAdapterWithIndexSupport, SourceQuery
 from sqlmesh.core.engine_adapter.mixins import (
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -3,11 +3,13 @@
 
 from __future__ import annotations
 
+import logging
 import typing as t
 
 import pandas as pd
 from pandas.api.types import is_datetime64_dtype  # type: ignore
 from sqlglot import exp
+from sqlglot.errors import ErrorLevel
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -178,7 +180,7 @@ def _get_data_objects(
         Returns all the data objects that exist in the given schema and catalog.
         """
         if not catalog_name:
-            catalog_name = self.fetchone("select DB_NAME()")[0]
+            catalog_name = self.fetchone("select DB_NAME()", quote_identifiers=True)[0]
```
tobymao, line 183: why are you doing quote_identifiers=True here?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -188,15 +190,15 @@ def _get_data_objects(
             FROM {catalog_name}.INFORMATION_SCHEMA.TABLES
             WHERE TABLE_SCHEMA LIKE '%{schema_name}%'
         """
-        dataframe: pd.DataFrame = self.fetchdf(query)
+        results = self.fetchall(query, quote_identifiers=True)
```
tobymao, line 193: same here, if you care just quote your query?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -206,3 +208,80 @@ def _truncate_table(self, table_name: TableName) -> str:
     def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.Any) -> str:
         sql = super()._to_sql(expression, quote=quote, **kwargs)
         return f"{sql};"
+
+    @overload
+    def execute_and_fetch(
```
tobymao, line 214: why do these exists in the public space? i don't think this is necessary

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -206,3 +209,80 @@ def _truncate_table(self, table_name: TableName) -> str:
...
+        self,
+        expression: t.Union[str, exp.Expression],
+        fetch_one: Literal[True],
+        ignore_unsupported_errors: bool = False,
+        quote_identifiers: bool = True,
+        **kwargs: t.Any,
+    ) -> t.Tuple:
+        ...
```
GeorgeSittas, line 233: I think we may be able to reduce the visual noise here by doing something similar to this: https://github.com/tobymao/sqlglot/blob/main/sqlglot/__init__.py#L89

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -206,3 +208,80 @@ def _truncate_table(self, table_name: TableName) -> str:
...
+
+    def execute_and_fetch(
```
tobymao, line 235: move this into fetchall, and then have fetch_one just return 1

tobymao: from fetchall

treysp: So fetchone should call fetchall and return the first entry?

tobymao: yes, and check the behavior on how it should raise

tobymao: actually i change my mind

tobymao: can't you just do 

with self.transaction():
   self.execute()
   return self.cursor.fetchone()

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -188,15 +191,15 @@ def _get_data_objects(
             FROM {catalog_name}.INFORMATION_SCHEMA.TABLES
             WHERE TABLE_SCHEMA LIKE '%{schema_name}%'
         """
-        dataframe: pd.DataFrame = self.fetchdf(query)
```
tobymao, line 191: why did you get rid of this? why don't you just fix fetchdf with a transaction wrap?

PR #1558, opened 2023-10-11
Fix: Support Setting Location in BigQuery Airflow Operator

eakmanrq:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1697020053051689

This results in the Airflow Operator client now having parity with the local client that is created for built-in. 

Note: This is untested in order save time having to actually bring up Airflow and test so I would appreciate extra 👀 on it.

### DIFF
--- a/sqlmesh/schedulers/airflow/hooks/bigquery.py
+++ b/sqlmesh/schedulers/airflow/hooks/bigquery.py
@@ -31,6 +31,7 @@ class SQLMeshBigQueryHook(GoogleBaseHook, DbApiHook):
         gcp_conn_id: str = default_conn_name,
         delegate_to: str | None = None,
         impersonation_chain: str | t.Sequence[str] | None = None,
+        location: t.Optional[str] = None,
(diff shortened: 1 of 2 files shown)


sqlmesh/schedulers/airflow/hooks/bigquery.py
```diff
@@ -50,5 +53,10 @@ def get_conn(self) -> Connection:
             creds, project_id = self._get_credentials_and_project_id()  # type: ignore
         except AttributeError:
             creds, project_id = self.get_credentials_and_project_id()  # type: ignore
-        client = Client(project=project_id, credentials=creds)
+        client = Client(
+            project=project_id,
+            credentials=creds,
+            location=self.location,
+            client_info=client_info.ClientInfo(user_agent="sqlmesh"),
```
eakmanrq, line 60: Matches what we do when creating connection locally

tobymao (approved):

PR #1559, opened 2023-10-11
Fix: Add Proper Catalog Support to Engine Adapter

eakmanrq:
Controversial change to highlight: Removed the field `catalog_name` and instead let callers provide a `TableName` which means it can either be a string or `exp.Table` object. 

Added engine adapter methods:
* `get_current_catalog`
* `set_current_catalog`

New Property: `CatalogSupport`
This tells the `set_catalog` decorator what level of catalog support the engine adapter has in order to set the catalog properly. 3 levels:
* Unsupported
* Requires Set Catalog
* Full Support

`Requires Set Catalog` means that operations cannot be provided with the catalog but instead they need `set_current_catalog` against the catalog and then they can provide the operation but now with the catalog dropped. This always refers to the target of the operation (ex: if drop schema, the schema being drooped) and we assume reads can happen across catalogs. For engines that don't support reads across catalogs we just consider them Unsupported (ex: Postgres). 

Follow up work:
* Have new default catalog integrate with connector default catalog if it supports it
* Add Redshift catalog support
  * The challenge with this is that Redshift doesn't allow changing the catalog from within a connection. A new connection needs to be created with the new catalog defined. This is just for creating objects across catalogs and not querying across catalogs

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -839,3 +839,35 @@ def transform_values(
             yield exp.func("PARSE_JSON", f"'{value}'")
         else:
             yield value
+
+
+def to_schema(sql_path: str | exp.Table) -> exp.Table:
+    if isinstance(sql_path, exp.Table) and sql_path.this is None:
+        return sql_path
+    table = exp.to_table(sql_path.copy() if isinstance(sql_path, exp.Table) else sql_path)
+    table.set("catalog", table.args.get("db"))
+    table.set("db", table.args.get("this"))
+    table.set("this", None)
+    return table
+
+
+def schema_(
+    db: exp.Identifier | str,
+    catalog: t.Optional[exp.Identifier | str] = None,
+    quoted: t.Optional[bool] = None,
+) -> exp.Table:
+    """Build a Schema.
+
+    Args:
+        db: Database name.
+        catalog: Catalog name.
+        quoted: Whether to force quotes on the schema's identifiers.
+
+    Returns:
+        The new Schema instance.
+    """
+    return exp.Table(
+        this=None,
+        db=exp.to_identifier(db, quoted=quoted) if db else None,
+        catalog=exp.to_identifier(catalog, quoted=quoted) if catalog else None,
+    )
--- a/sqlmesh/core/engine_adapter/base_postgres.py
+++ b/sqlmesh/core/engine_adapter/base_postgres.py
@@ -4,12 +4,12 @@ import typing as t
 
 from sqlglot import exp
 
-from sqlmesh.core.engine_adapter.mixins import EngineAdapter
+from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.engine_adapter.shared import DataObject, DataObjectType
 from sqlmesh.utils.errors import SQLMeshError
 
 if t.TYPE_CHECKING:
-    from sqlmesh.core._typing import TableName
+    from sqlmesh.core._typing import SchemaName, TableName
     from sqlmesh.core.engine_adapter.base import QueryOrDF
 
 
@@ -92,13 +92,11 @@ class BasePostgresEngineAdapter(EngineAdapter):
                 **create_kwargs,
             )
 
-    def _get_data_objects(
-        self, schema_name: str, catalog_name: t.Optional[str] = None
-    ) -> t.List[DataObject]:
+    def _get_data_objects(self, schema_name: SchemaName) -> t.List[DataObject]:
         """
         Returns all the data objects that exist in the given schema and optionally catalog.
         """
-        catalog_name = f"'{catalog_name}'" if catalog_name else "NULL"
+        catalog_name = f"'{self.get_current_catalog()}'"
         query = f"""
             SELECT
                 {catalog_name} AS catalog_name,
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -6,7 +6,7 @@ import typing as t
 import pandas as pd
 from sqlglot import Dialect, exp
 
-from sqlmesh.core.engine_adapter.base import InsertOverwriteStrategy
+from sqlmesh.core.engine_adapter.base import CatalogSupport, InsertOverwriteStrategy
 from sqlmesh.core.engine_adapter.spark import SparkEngineAdapter
 from sqlmesh.core.schema_diff import SchemaDiffer
 from sqlmesh.utils import classproperty
@@ -30,6 +30,7 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
         support_nested_operations=True,
         array_element_selector="element",
     )
+    CATALOG_SUPPORT = CatalogSupport.FULL_SUPPORT
 
     def __init__(
         self,
@@ -109,14 +110,14 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
             if catalog:
                 from py4j.protocol import Py4JError
 
-                # Note: Spark 3.4+ Only API
                 try:
-                    self.spark.catalog.setCurrentCatalog(catalog)
+                    # Note: Spark 3.4+ Only API
+                    super().set_current_catalog(catalog)
                 # If `setCurrentCatalog` should work for both non-unity and Unity single user
                 # clusters. If it fails then we try `USE CATALOG` which is Unity only but works
                 # across all clusters
                 except Py4JError:
-                    self.spark.sql(f"USE CATALOG {catalog}")
+                    self.set_current_catalog(catalog)
             self._spark.conf.set("spark.sql.sources.partitionOverwriteMode", "dynamic")
         return self._spark
 
@@ -146,6 +147,15 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
         if not isinstance(df, pd.DataFrame):
             return df.toPandas()
         return df
+
+    def get_current_catalog(self) -> t.Optional[str]:
+        result = self.fetchone("SELECT current_catalog()")
+        if result:
+            return result[0]
+        return None
+
+    def set_current_catalog(self, catalog_name: str) -> None:
+        self.execute(exp.Use(this=exp.to_identifier(catalog_name), kind="CATALOG"))
 
     def clone_table(
         self,
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -10,19 +10,21 @@ from pandas.api.types import is_datetime64_dtype  # type: ignore
 from sqlglot import exp
 from sqlglot.optimizer.qualify_columns import quote_identifiers
 
-from sqlmesh.core.engine_adapter.base import EngineAdapterWithIndexSupport, SourceQuery
+from sqlmesh.core.engine_adapter.base import (
+    CatalogSupport,
+    EngineAdapterWithIndexSupport,
+    SourceQuery,
+)
 from sqlmesh.core.engine_adapter.mixins import (
     InsertOverwriteWithMergeMixin,
     LogicalReplaceQueryMixin,
     PandasNativeFetchDFSupportMixin,
 )
-from sqlmesh.core.engine_adapter.shared import DataObject, DataObjectType
+from sqlmesh.core.engine_adapter.shared import DataObject, DataObjectType, set_catalog
 
 if t.TYPE_CHECKING:
-    import pymssql
-
-    from sqlmesh.core._typing import TableName
-    from sqlmesh.core.engine_adapter._typing import DF, Query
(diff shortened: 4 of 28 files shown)


sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -83,16 +84,14 @@ def _fetch_native_df(
...
+    @set_catalog(schema_field=(1, "schema_name"), override=CatalogSupport.REQUIRES_SET_CATALOG)
+    def _get_data_objects(self, schema_name: t.Union[str, exp.Table]) -> t.List[DataObject]:
         """
         Returns all the data objects that exist in the given schema and optionally catalog.
         """
         from snowflake.connector.errors import ProgrammingError
 
-        target = nullsafe_join(".", catalog_name, schema_name)
-        sql = f"SHOW TERSE OBJECTS IN {target}"
+        sql = f'SHOW TERSE OBJECTS IN "{exp.to_table(schema_name, dialect=self.dialect).sql(dialect=self.dialect)}"'
```
eakmanrq, line 94: I think neither the old approach or this approach properly deals with identifiers in Snowflake. Any advice on the best way to do this for Snowflake?

sqlmesh/core/dialect.py
```diff
@@ -839,3 +839,13 @@ def transform_values(
             yield exp.func("PARSE_JSON", f"'{value}'")
         else:
             yield value
+
+
+def to_schema(sql_path: str | exp.Table) -> exp.Table:
+    if isinstance(sql_path, exp.Table):
```
tobymao, line 845: you should check if table.this is filled, if it is, you need to swap stuff around, this ensures passing in a table will convert it into a schema

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -92,13 +92,11 @@ def create_view(
...
 
-    def _get_data_objects(
-        self, schema_name: str, catalog_name: t.Optional[str] = None
-    ) -> t.List[DataObject]:
+    def _get_data_objects(self, schema_name: SchemaName) -> t.List[DataObject]:
         """
         Returns all the data objects that exist in the given schema and optionally catalog.
         """
-        catalog_name = f"'{catalog_name}'" if catalog_name else "NULL"
+        catalog_name = f"'{self.get_current_catalog()}'"
```
tobymao, line 99: why don't you have to handle null?

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -147,6 +149,15 @@ def fetchdf(
             return df.toPandas()
         return df
 
+    def get_current_catalog(self) -> t.Optional[str]:
+        result = self.fetchone("SELECT current_catalog()")
+        if result:
+            return seq_get(result, 0)
```
tobymao, line 155: why are you doing a seq_get if result is truthy? can't you just do result[0]

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -110,12 +114,21 @@ def table_exists(self, table_name: TableName) -> bool:
 
         return result[0] == 1 if result else False
 
-    @property
-    def connection(self) -> pymssql.Connection:
-        return self.cursor.connection
+    def get_current_catalog(self) -> t.Optional[str]:
+        result = self.fetchone("SELECT DB_NAME()")
+        if result:
+            return seq_get(result, 0)
```
tobymao, line 120: same

sqlmesh/core/engine_adapter/postgres.py
```diff
@@ -23,6 +24,12 @@ class PostgresEngineAdapter(
     DIALECT = "postgres"
     SUPPORTS_INDEXES = True
 
+    def get_current_catalog(self) -> t.Optional[str]:
+        result = self.fetchone("SELECT current_catalog")
+        if result:
+            return seq_get(result, 0)
```
tobymao, line 30: lot of redundancy here, could this be dried up?

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+    def decorator(func: t.Callable) -> t.Callable:
+        @functools.wraps(func)
+        def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
+            def _get_args_pos_and_kwarg_name(
+                func: t.Callable,
+            ) -> t.Optional[t.Tuple[str, int, str]]:
+                spec = inspect.getfullargspec(func)
+                for i, name in enumerate(spec.args):
+                    obj_type = spec.annotations.get(name)
+                    if obj_type == "SchemaName":
```
tobymao, line 73: SchemaName.__name__ ?

tobymao (approved):
z

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+
+
+def set_catalog(
+    *,
+    override: t.Optional[CatalogSupport] = None,
+) -> t.Callable:
+    def decorator(func: t.Callable) -> t.Callable:
+        @functools.wraps(func)
+        def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
+            def _get_args_pos_and_kwarg_name(
```
tobymao, line 67: why is this nested? does it need  to be? makes it hard to read this

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+            # If there is full catalog support then we have nothing to do
+            if catalog_support.is_full_support:
+                return func(*list_args, **kwargs)
+
+            # Get the field value and the container which it came from so we can update it later
+            location = _get_args_pos_and_kwarg_name(func)
+            if location is None:
+                return func(*list_args, **kwargs)
+            name, pos, obj_type = location
+            obj: t.Union[str, exp.Table]
```
tobymao, line 92: nit, i don't like these forward type declarations, prefer inline

sqlmesh/core/engine_adapter/shared.py
eakmanrq: `SchemaName` is part of typing and therefore generally expencted to not be imported and used during runtime. Therefore I would rather keep that than avoid using this string.

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+            container: t.Union[t.Dict, t.List]
+            key: t.Union[int, str]
+            obj, container, key = (
+                (kwargs.get(name), kwargs, name)  # type: ignore
```
tobymao, line 96: better to do a cast than a type ignore if you know what type you want it to be

sqlmesh/core/engine_adapter/shared.py
eakmanrq: Moved out

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+                if kwargs.get(name)
+                else (list_args[pos], list_args, pos)
+            )
+            # mypy shenanigans: https://github.com/python/mypy/issues/10740
+            to_table_func: t.Callable[[t.Union[str, exp.Table]], exp.Table] = exp.to_table
+            to_schema_func: t.Callable[[t.Union[str, exp.Table]], exp.Table] = to_schema
+            to_expression_func = to_table_func if obj_type == "TableName" else to_schema_func
```
tobymao, line 103: obj_type

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+            obj: t.Union[str, exp.Table]
+            container: t.Union[t.Dict, t.List]
+            key: t.Union[int, str]
+            obj, container, key = (
+                (kwargs.get(name), kwargs, name)  # type: ignore
+                if kwargs.get(name)
+                else (list_args[pos], list_args, pos)
+            )
+            # mypy shenanigans: https://github.com/python/mypy/issues/10740
+            to_table_func: t.Callable[[t.Union[str, exp.Table]], exp.Table] = exp.to_table
```
tobymao, line 101: this is garbage, just do a cast

sqlmesh/core/engine_adapter/base_postgres.py
eakmanrq: Postgres should always have a database that it is operating in and therefore a catalog is always returned.

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -47,3 +55,74 @@ class DataObject(PydanticModel):
...
+            catalog_name = expression.args.get("catalog")
+            if not catalog_name:
+                return func(*list_args, **kwargs)
+            # If we have a catalog and this engine doesn't support catalogs then we need to error
+            if catalog_support.is_unsupported:
+                raise UnsupportedCatalogOperationError(
+                    f"{engine_adapter.dialect} does not support catalogs and a catalog was provided: {catalog_name}"
+                )
+            # Remove the catalog name from the argument so the engine adapter doesn't try to use it
+            expression.set("catalog", None)
```
tobymao, line 114: why is this necessary? to_schema should handle it

eakmanrq: `to_schema`'s purpose is to create a table object without a table name but with `db` and `catalog` populated (if defined).

The goal of this line is to take whatever the expression object is and ensure it does not have catalog set. We do this because we are setting the catalog using the engine adapter and therefore we don't want it include in future SQL statements.

sqlmesh/core/engine_adapter/shared.py
eakmanrq: what is the issue here?

sqlmesh/core/state_sync/common.py
```diff
@@ -33,13 +34,20 @@ def cleanup_expired_views(adapter: EngineAdapter, environments: t.List[Environme
...
+            snapshot.qualified_view_name.catalog,
+            snapshot.qualified_view_name.schema_for_environment(environment.naming_info),
+        )
         for environment in expired_schema_environments
         for snapshot in environment.snapshots
         if snapshot.is_model
     }:
-        adapter.drop_schema(expired_schema, ignore_if_not_exists=True, cascade=True)
+        adapter.drop_schema(
+            nullsafe_join(".", expired_catalog, expired_schema),
```
tobymao, line 47: schema_(catalog, schema)

sqlmesh/dbt/adapter.py
```diff
@@ -212,7 +214,7 @@ def list_relations_without_caching(self, schema_relation: BaseRelation) -> t.Lis
 
         assert schema_relation.schema is not None
         data_objects = self.engine_adapter._get_data_objects(
-            schema_name=schema_relation.schema, catalog_name=schema_relation.database
+            to_schema(nullsafe_join(".", schema_relation.database, schema_relation.schema))
```
tobymao, line 217: use schema here?

tobymao (approved):

sqlmesh/core/state_sync/common.py
eakmanrq: I think the qualified view stuff currently has a bug. We parse a snapshot name into a table (that part is fine) but then get the string representation of the identifiers instead of the identifiers themsevles: https://github.com/TobikoData/sqlmesh/blob/f8e45bbc20f5dda759728c260bc194dff8d536ce/sqlmesh/core/snapshot/definition.py#L215-L220

Since we convert to string, I think doing a string join here is correct. If we had identifiers or potentially strings that are properly normalized then we would do `schemo_`. At least that is how I think it is supposed to work based on `to_table` and `table_`.

tobymao: hm not sure i understand, let's chat about this

tests/core/engine_adapter/test_integration.py
```diff
@@ -132,9 +135,13 @@ def output_data(self, data: pd.DataFrame) -> pd.DataFrame:
...
-        return normalize_identifiers(
-            exp.to_table(".".join([schema, table_name]), dialect=self.engine_adapter.dialect),
-            dialect=self.engine_adapter.dialect,
+        return exp.to_table(
+            normalize_model_name(".".join([schema, table_name]), dialect=self.dialect)
+        )
+
+    def schema(self, schema_name: str, catalog_name: t.Optional[str] = None) -> str:
+        return normalize_model_name(
+            nullsafe_join(".", catalog_name, schema_name), dialect=self.dialect
```
tobymao, line 144: change this

PR #1560, opened 2023-10-11
Fix: improve error message when a lib needs to be upgraded

GeorgeSittas:


tobymao (approved):

GeorgeSittas:
Just lowercased the `lib` to avoid displaying "SQLGlot" in the pip command. For example:

```
➜  foo sqlmesh plan
Error: SQLGlot (local) is using version '18.10.2.dev14' which is behind '18.11.6' (remote). Please upgrade SQLGlot ('pip install SQLGlot --upgrade' command).
➜  foo sqlmesh plan
Error: SQLGlot (local) is using version '18.10.2.dev14' which is behind '18.11.6' (remote). Please upgrade SQLGlot ('pip install sqlglot --upgrade' command).
```

eakmanrq:
@GeorgeSittas What do you think about providing the full command so they can just copy/paste?

GeorgeSittas:
> @GeorgeSittas What do you think about providing the full command so they can just copy/paste?

Which command do you mean? I think `pip install sqlglot --upgrade` suffices here for example right?

eakmanrq:
Well it doens't include the version to you. So if they just copy/paste that and SQLGlot is ahead then it will install the wrong version.

GeorgeSittas:
> Well it doens't include the version in it. So if they just copy/paste that and SQLGlot is ahead then it will install the wrong version.

Ah my bad, good catch. I'll fix this, thanks!

GeorgeSittas:
Updated message:

```
➜  foo sqlmesh plan
Error: SQLGlot (local) is using version '18.10.1' which is behind '18.12.0' (remote). Please upgrade SQLGlot ('pip install "sqlglot==18.12.0"' command).
```

eakmanrq:
@GeorgeSittas Don't you need `--upgrade` included?

GeorgeSittas:
I thought the `==` forces the installation of that specific version? I think `--upgrade` is required if we want to force-update even though the currently used version is compatible with what we have in `setup.py`. Am I missing something?

eakmanrq:
@GeorgeSittas Ah you might be right. I just always do it. As you said it shouldn't hurt though.

PR #1561, opened 2023-10-11
Fix: Initialize dev intervals with prod intervals for new forward-only snapshots

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -414,28 +414,30 @@ def _missing_intervals(self) -> t.Dict[t.Tuple[str, str], Intervals]:
             # to choose non-breaking / forward only. this will change the version of the snapshot on the fly
             # thus changing the missing intervals. additionally we replace any snapshots with the old copies
             # because they have intervals and the ephemeral ones don't
-            snapshots = {
-                (snapshot.name, snapshot.version_get_or_generate()): snapshot
-                for snapshot in self.snapshots
+            old_snapshots = {
+                (old.name, old.version_get_or_generate()): old
+                for _, old in self.context_diff.modified_snapshots.values()
             }
```
tobymao, line 420: why did this change from snapshots to modified_snapshots

sqlmesh/core/plan/evaluator.py
```diff
@@ -376,3 +378,13 @@ def can_evaluate_before_promote(
     return not snapshot.is_paused_forward_only and not any(
         snapshots[p_id].is_paused_forward_only for p_id in snapshot.parents
     )
+
+
+def update_intervals_for_new_snapshots(
```
tobymao, line 383: why is this done in the plan evaluator?

sqlmesh/core/plan/definition.py
izeigerman: because the original logic didn't make much sense. We're only missing intervals for newly created snapshots and that's what we should hydrate instead of replacing new snapshots with old ones.

sqlmesh/core/plan/evaluator.py
izeigerman: Because it's a part of the plan evaluation. Where should it be instead?

sqlmesh/core/plan/definition.py
tobymao: is the comment above still valid? i remember this section was quite tricky and i added this in order to fix a very tricky bug with changing categorizations

tobymao (approved):

PR #1562, opened 2023-10-12
Feat: expose a macro var to check if we're evaluating snapshots

GeorgeSittas:
This is the first step towards adding support for the following pattern (it's still a WIP - will finish in a followup PR):

```sql
MODEL (...);

@IF(@runtime_stage = 'evaluating', <statement>);  -- only execute <statement> @ snapshot evaluation time

...
```

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import typing as t
+from enum import Enum
 from functools import reduce
 from string import Template
 
@@ -27,6 +28,12 @@ from sqlmesh.utils.jinja import JinjaMacroRegistry, has_jinja
 from sqlmesh.utils.metaprogramming import Executable, prepare_env, print_exception
 
 SQLMESH_MOCKED_STAR = "__SQLMESH_MOCKED_STAR__"
+
+
+class RuntimeStage(Enum):
+    LOADING = "loading"
+    CREATING = "creating"
+    EVALUATING = "evaluating"
 
 
 class MacroStrTemplate(Template):
@@ -103,11 +110,11 @@ class MacroEvaluator:
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         jinja_env: t.Optional[Environment] = None,
         schema: t.Optional[t.Dict[str, t.Any]] = None,
-        evaluating: bool = False,
+        runtime_stage: RuntimeStage = RuntimeStage.LOADING,
     ):
         self.dialect = dialect
         self.generator = MacroDialect().generator()
-        self.locals: t.Dict[str, t.Any] = {"evaluating": evaluating}
+        self.locals: t.Dict[str, t.Any] = {"runtime_stage": runtime_stage}
         self.env = {**ENV, "self": self}
         self.python_env = python_env or {}
         self._jinja_env: t.Optional[Environment] = jinja_env
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -103,11 +103,11 @@ class MacroEvaluator:
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         jinja_env: t.Optional[Environment] = None,
         schema: t.Optional[t.Dict[str, t.Any]] = None,
-        execute: t.Optional[bool] = None,
+        evaluating: t.Optional[bool] = None,
     ):
         self.dialect = dialect
         self.generator = MacroDialect().generator()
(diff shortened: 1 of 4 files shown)


sqlmesh/core/macros.py
```diff
@@ -103,10 +103,11 @@ def __init__(
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         jinja_env: t.Optional[Environment] = None,
         schema: t.Optional[t.Dict[str, t.Any]] = None,
+        execute: t.Optional[bool] = None,
```
tobymao, line 106: maybe it should be evaluating?

sqlmesh/core/renderer.py
```diff
@@ -203,10 +208,12 @@ def _cache_key(
         start: t.Optional[TimeLike] = None,
         end: t.Optional[TimeLike] = None,
         execution_time: t.Optional[TimeLike] = None,
-    ) -> t.Tuple[datetime, datetime, datetime]:
+        execute: t.Optional[bool] = None,
+    ) -> t.Tuple[datetime, datetime, datetime, bool]:
```
tobymao, line 212: make this a named type

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -410,6 +410,7 @@ def _create_snapshot(
             engine_adapter=self.adapter,
             snapshots=parent_snapshots_by_name,
             is_dev=is_dev,
+            execute=True,
```
tobymao, line 413: wait why is this one True? shouldn't it be false? this is the section where we create the table

GeorgeSittas: Yeah this was an oversight, thanks for catching

sqlmesh/core/macros.py
```diff
@@ -103,10 +103,11 @@ def __init__(
         python_env: t.Optional[t.Dict[str, Executable]] = None,
         jinja_env: t.Optional[Environment] = None,
         schema: t.Optional[t.Dict[str, t.Any]] = None,
+        evaluating: t.Optional[bool] = None,
```
tobymao, line 106: evaluating: bool = False

sqlmesh/core/macros.py
```diff
@@ -103,10 +103,11 @@ def __init__(
...
     ):
         self.dialect = dialect
         self.generator = MacroDialect().generator()
-        self.locals: t.Dict[str, t.Any] = {}
+        self.locals: t.Dict[str, t.Any] = {"evaluating": bool(evaluating)}
```
tobymao, line 110: don't need the bool here

sqlmesh/core/renderer.py
```diff
@@ -203,10 +208,12 @@ def _cache_key(
         start: t.Optional[TimeLike] = None,
         end: t.Optional[TimeLike] = None,
         execution_time: t.Optional[TimeLike] = None,
-    ) -> t.Tuple[datetime, datetime, datetime]:
+        evaluating: t.Optional[bool] = None,
```
tobymao, line 211: evaluating bool false

tobymao (approved):

sqlmesh/core/renderer.py
GeorgeSittas: kept this one because we pass `kwargs.get("evaluating")` which can be `None`

sqlmesh/core/renderer.py
```diff
@@ -307,7 +314,7 @@ def render(
         Returns:
             The rendered expression.
         """
-        cache_key = self._cache_key(start, end, execution_time)
+        cache_key = self._cache_key(start, end, execution_time, kwargs.get("evaluating"))
```
izeigerman, line 317: Why not make it a top-level argument?

sqlmesh/core/renderer.py
```diff
@@ -87,8 +89,8 @@ def _render(
             The rendered expressions.
         """
 
-        cache_key = self._cache_key(start, end, execution_time)
-        start_dt, end_dt, execution_dt = cache_key
+        cache_key = self._cache_key(start, end, execution_time, kwargs.get("evaluating"))
```
izeigerman, line 92: ditto

sqlmesh/core/renderer.py
GeorgeSittas: Addressed in https://github.com/TobikoData/sqlmesh/pull/1562/commits/74f69aa9097c6c7904b2f4f6f5ebb620acea8600

izeigerman (approved):

PR #1565, opened 2023-10-12
Fix: In Airflow skip state migration when loading the SQLMesh plugin on the Webserver instance

izeigerman:


max-winderbaum:
So smart 🧠

PR #1566, opened 2023-10-12
Feat: Support BigQuery No-Op Coerce Types

eakmanrq:
If types are compatible (i.e: alterable) then that also means for BigQuery that the inverse relationship can be a no-operation for us. We can't alter but we can insert new values without issue. 

This can later be expanded to more engines but just focusing on BigQuery for now.


tobymao (approved):
will there be a warning?

eakmanrq:
> will there be a warning?

Added

PR #1568, opened 2023-10-12
Feat!: Sync Connection Catalog With Model Default Catalog

eakmanrq:
This PR makes sure that default catalog is synced with the connection catalog. If they say different things then an error is raised. Also if a user sets a default catalog on an engine that doesn't support catalogs then an error is raised.


eakmanrq:
Reworking this change in new PR

PR #1569, opened 2023-10-13
Feat: Support When Matched Expression for Merge

eakmanrq:
Users can now override the when matched logic for `INCREMENTAL_BY_UNQIUE_KEY` model kind. Note that macros are not supported in this PR.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -294,15 +294,17 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
     if not key:
         return None
 
-    if self._match(TokenType.L_PAREN):
-        value: t.Optional[exp.Expression] = self.expression(
+    name = key.name.lower()
+    if name == "when_matched":
+        value: t.Optional[exp.Expression] = self._parse_when_matched()[0]
+    elif self._match(TokenType.L_PAREN):
+        value = self.expression(
             exp.Tuple, expressions=self._parse_csv(lambda: _parse_prop_value(self))
         )
         self._match_r_paren()
     else:
         value = self._parse_bracket(self._parse_field(any_token=True))
 
-    name = key.name.lower()
     if name == "path" and value:
         # Make sure if we get a windows path that it is converted to posix
         value = exp.Literal.string(value.this.replace("\\", "/"))
--- a/sqlmesh/core/model/kind.py
+++ b/sqlmesh/core/model/kind.py
@@ -18,6 +18,7 @@ from sqlmesh.utils.pydantic import (
     SQLGlotPositiveInt,
     SQLGlotString,
     field_validator,
+    field_validator_v1_args,
     model_validator,
     model_validator_v1_args,
 )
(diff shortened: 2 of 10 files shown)


sqlmesh/core/dialect.py
```diff
@@ -294,15 +294,17 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
     if not key:
         return None
 
-    if self._match(TokenType.L_PAREN):
-        value: t.Optional[exp.Expression] = self.expression(
+    name = key.name.lower()
+    if name == "when_matched":
+        value: t.Optional[exp.Expression] = self._parse_when_matched()[0]
```
tobymao, line 299: can't there be multiple when matched expressions? 

when matched, when not matched etc

sqlmesh/core/model/kind.py
```diff
@@ -208,8 +208,36 @@ def to_expression(self, dialect: str = "", **kwargs: t.Any) -> d.ModelKind:
...
+    @classmethod
+    def _when_matched_validator(cls, v: t.Optional[exp.When]) -> t.Optional[exp.When]:
+        def replace_table_references(expression: exp.Expression) -> exp.Expression:
+            from sqlmesh.core.engine_adapter.base import (
+                MERGE_SOURCE_ALIAS,
+                MERGE_TARGET_ALIAS,
+            )
+
+            if isinstance(expression, exp.Column):
+                if expression.table == "target":
```
tobymao, line 225: lower

sqlmesh/core/model/kind.py
```diff
@@ -208,8 +208,36 @@ def to_expression(self, dialect: str = "", **kwargs: t.Any) -> d.ModelKind:
...
     unique_key: t.List[exp.Expression]
+    when_matched: t.Optional[exp.When] = None
+
     _unique_key_validator = _unique_key_validator
 
+    @field_validator("when_matched", mode="before")
+    @classmethod
+    def _when_matched_validator(cls, v: t.Optional[exp.When]) -> t.Optional[exp.When]:
+        def replace_table_references(expression: exp.Expression) -> exp.Expression:
+            from sqlmesh.core.engine_adapter.base import (
```
tobymao, line 219: do we need to set the dialect meta here? look at other places where we validate expressions

sqlmesh/core/dialect.py
eakmanrq: This PR is just implementing when matched. I believe that is all that is needed for now and we can add more later if that is required.

tobymao: you don't think it's worth it to just implement both? complexity doesn't seem too high

sqlmesh/core/model/kind.py
eakmanrq: Do you know why the parser doesn't automatically do this when returning expressions?

eakmanrq: Also not sure what the purpose of this is and the impact if it is not included.

eakmanrq: I did add it.

tobymao: parser doesn't add it to every node automatically

tobymao (approved):

PR #1570, opened 2023-10-13
Fix: BigQuery Operator/Hook kwarg passing

eakmanrq:


sqlmesh/schedulers/airflow/operators/bigquery.py
```diff
@@ -22,16 +22,26 @@ def __init__(
         *,
         target: BaseTarget,
         bigquery_conn_id: str = SQLMeshBigQueryHook.default_conn_name,
+        delegate_to: str | None = None,
+        impersonation_chain: str | t.Sequence[str] | None = None,
+        location: t.Optional[str] = None,
```
izeigerman, line 27: [Nit] We use Optional here but ` | None` for other args.

izeigerman (approved):

PR #1577, opened 2023-10-16
Feat: add support for @IF(cond, statement) macro

GeorgeSittas:
This PR along with https://github.com/TobikoData/sqlmesh/commit/c637908ab859617b18dafc54335fa38dd6a08dd6 allows SQLMesh users to conditionally execute pre/post-statements based on the current runtime stage, using the good ol' `@IF` macro: `@IF(condition, statement)`.

~The reason I didn't just implement this using the good ol' `@IF(...)` macro, is that we don't account for statements / commands in normal function calls and hence we'd need to override its parser to account for them, thus making the `_parse_if` method more complex.~ Refactored to use `@IF` instead.

### DIFF
--- a/tests/core/test_snapshot_evaluator.py
+++ b/tests/core/test_snapshot_evaluator.py
@@ -189,13 +189,10 @@ def test_runtime_stages(capsys, mocker, adapter_mock, make_snapshot):
     evaluator.evaluate(snapshot, "2020-01-01", "2020-01-02", "2020-01-02", snapshots={})
     assert f"RuntimeStage value: {RuntimeStage.EVALUATING.value}" in capsys.readouterr().out
 
-    adapter_mock.execute.assert_has_calls(
-        [
-            call([]),
-            call([]),
-            call([parse_one("ALTER TABLE test_schema.foo MODIFY COLUMN c SET MASKING POLICY p")]),
-            call([]),
-        ]
+    non_empty_calls = [c for c in adapter_mock.execute.mock_calls if c != call([])]
+    assert len(non_empty_calls) == 1
+    assert non_empty_calls[0] == call(
+        [parse_one("ALTER TABLE test_schema.foo MODIFY COLUMN c SET MASKING POLICY p")]
     )
 
 
--- a/tests/core/test_snapshot_evaluator.py
+++ b/tests/core/test_snapshot_evaluator.py
@@ -158,7 +158,7 @@ def test_runtime_stages(capsys, mocker, adapter_mock, make_snapshot):
     @macro()
     def increment_stage_counter(evaluator) -> None:
         # Hack which allows us to intercept the different runtime stage values
-        print(f"RuntimeStage value: {evaluator.locals['runtime_stage'].value}")
+        print(f"RuntimeStage value: {evaluator.locals['runtime_stage']}")
 
     model = load_sql_based_model(
         parse(  # type: ignore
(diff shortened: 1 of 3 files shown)


tobymao:
i'd rather just making @if more complicated, how hard is that?

GeorgeSittas:
> i'd rather just making @if more complicated, how hard is that?

I think we'd need to either solve this by overriding `_parse_if` to use try-except and use the logic of `execute_if` as implemented here, or mess with sqlglot's implementation to change the parsing logic for the `true` and `false` args. The parser in sqlglot is also not trivial to do a simple patch here without duplicating the code too much:

```python
    def _parse_if(self) -> t.Optional[exp.Expression]:
        if self._match(TokenType.L_PAREN):
            args = self._parse_csv(self._parse_conjunction)
            this = self.validate_expression(exp.If.from_arg_list(args), args)
            self._match_r_paren()
        else:
            index = self._index - 1
            condition = self._parse_conjunction()

            if not condition:
                self._retreat(index)
                return None

            self._match(TokenType.THEN)
            true = self._parse_conjunction()
            false = self._parse_conjunction() if self._match(TokenType.ELSE) else None
            self._match(TokenType.END)
            this = self.expression(exp.If, this=condition, true=true, false=false)

        return self._parse_window(this)
```

tests/core/test_snapshot_evaluator.py
```diff
@@ -158,7 +158,7 @@ def test_runtime_stages(capsys, mocker, adapter_mock, make_snapshot):
     @macro()
     def increment_stage_counter(evaluator) -> None:
         # Hack which allows us to intercept the different runtime stage values
-        print(f"RuntimeStage value: {evaluator.locals['runtime_stage'].value}")
+        print(f"RuntimeStage value: {evaluator.locals['runtime_stage']}")
```
tobymao, line 161: why does this call print?

tests/core/test_snapshot_evaluator.py
```diff
@@ -188,6 +189,15 @@ def increment_stage_counter(evaluator) -> None:
     evaluator.evaluate(snapshot, "2020-01-01", "2020-01-02", "2020-01-02", snapshots={})
     assert f"RuntimeStage value: {RuntimeStage.EVALUATING.value}" in capsys.readouterr().out
 
+    adapter_mock.execute.assert_has_calls(
+        [
+            call([]),
```
tobymao, line 194: why are you checking these 3 empty calls?

tests/core/test_snapshot_evaluator.py
GeorgeSittas: Because intercepting the macro call at runtime stages besides evaluation wasn't trivial for this test. The pattern we're using in the snapshot evaluator tests is to pass a `payload` dict in `evaluate` so it can populate it but that misses the load / create stages. See https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_snapshot_evaluator.py#L80.

tests/core/test_snapshot_evaluator.py
GeorgeSittas: Because the pre-statements are rendered as `None` and so they're filtered out, producing an empty list to be passed to `EngineAdapter.execute`. Only the `@IF` macro is rendered as non-`None` once, when we are at evaluation time.

tobymao: should probaly only check that the if is called, don't need to check all of them

tobymao (approved):

PR #1579, opened 2023-10-17
Feat: Add Trino Hive Connector Support

eakmanrq:


docs/guides/connections.md
```diff
@@ -87,3 +87,4 @@ default_gateway: local_db
 * [Redshift](../integrations/engines/redshift.md)
 * [Snowflake](../integrations/engines/snowflake.md)
 * [Spark](../integrations/engines/spark.md)
+* Trino(../integrations/engines/trino.md)
```
tobymao, line 90: formatting [Trino]

sqlmesh/core/audit/builtin.py
```diff
@@ -59,7 +59,6 @@
     query="""
 SELECT 1
 FROM @this_model
-LIMIT @threshold + 1
```
tobymao, line 62: why was this needed to be removed? trino error?

sqlmesh/core/engine_adapter/trino.py
```diff
@@ -0,0 +1,106 @@
...
+
+class TrinoEngineAdapter(
+    PandasNativeFetchDFSupportMixin,
+    LogicalReplaceQueryMixin,
+    HiveMetastoreTablePropertiesMixin,
+    GetCurrentCatalogFromFunctionMixin,
+):
+    DIALECT = "trino"
+    ESCAPE_JSON = False
+    INSERT_OVERWRITE_STRATEGY = InsertOverwriteStrategy.INTO_IS_OVERWRITE
```
tobymao, line 36: what does this property mean

tobymao:
lgtm

sqlmesh/core/audit/builtin.py
eakmanrq: Ah yes I meant to make a comment on this. 

I don't understand how this query works. For example if I put this into Bigquery:
```
SELECT 1 FROM `sushi.items__fixture` HAVING COUNT(*) <= 100
```
I get this error:
```
The HAVING clause only allows aggregation if GROUP BY or SELECT list aggregation is present 
```
Which makes sense to me.

sqlmesh/core/engine_adapter/trino.py
eakmanrq: In Trino we basically set a configuration that tells it to treat an `INTO` as if it was an `OVERWRITE`. Therefore we set this strategy and it tells the base adapter to just pretend like we are doing an overwrite while keeping `INTO`.

https://github.com/TobikoData/sqlmesh/blob/bd71fffd1051111ff5f2442fb29324b61124b88d/sqlmesh/core/engine_adapter/trino.py#L58-L64

sqlmesh/core/audit/builtin.py
eakmanrq: The reason why I touched this at all is because Trino does not support the `+` in the limit. The more the read the query the more I was confused how this is supposed to work.

tobymao: ok change it to this

```
SELECT COUNT(*) AS cnt
FROM @this_model
LIMIT @threshold + 1
HAVING cnt <= @threshold
```

i will make sqlglot fix the limit @threshold

eakmanrq:
```
SELECT COUNT(*) AS cnt
FROM @this_model
LIMIT @threshold + 1
HAVING cnt <= @threshold
```

The limit in this won't do anything. Limit is last in terms of order of operations and therefore since `count(*) will only return 1 row you are limiting the results of something that has a single row. 

If you want the limit on read then you need to do something like this:
```
SELECT COUNT(*) as cnt
FROM (
   SELECT 1
   FROM @this_model
   LIMIT @threshold + 1
)
HAVING cnt <= @threshold
```

Btw I don't think all engines support using an alias in having like that but not sure. It also might be solved by SQLGlot.

tobymao: i can do that, but i'm also not aware of an engine that doesn't support that

tobymao: let's do that subquery as you propose, i'm aware that bigquery has problems NOT using the alias (opposite of what you said)

eakmanrq: Ok will you update SQLGlot for Postgres to handle the limit issue?

eakmanrq: He updated here:  https://github.com/tobymao/sqlglot/commit/d9958f9438eecf8137361ab76f02acda63e59790

examples/sushi/models/items.py
```diff
@@ -63,18 +63,18 @@
...
-        "float_prop": 1.0,
-        "bool_prop": True,
-    },
-    session_properties={
-        "string_prop": "some_value",
-        "int_prop": 1,
-        "float_prop": 1.0,
-        "bool_prop": True,
-    },
+    # table_properties={
```
tobymao, line 66: temporary?

eakmanrq: mistake thanks

PR #1580, opened 2023-10-17
Chore: update macro docs to include runtime_stage macro

GeorgeSittas:


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -318,6 +328,21 @@ SELECT
...
 ```
 
+Another example is conditionally executing a pre/post-statement depending on the current [runtime stage](./macro_variables.md#predefined-variables). For instance, the following `@IF` pre-statement will only be executed at model creation time:
+
+```sql linenums="1"
+MODEL (
+  ...
+);
+
+@IF(
```
tobymao, line 338: have you tested this? i think you want this as a post statement and not a pre statement

GeorgeSittas: not yet, i can try to test this specific example though soon

GeorgeSittas: to be clear, i did test using both `create schema ...` and `create table ...` as pre-statements, seemed good

GeorgeSittas: yeah what you're saying makes sense, I'll try to refine the example

GeorgeSittas: updated the example to use an evaluation-time post-statement instead

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -260,10 +260,22 @@ It includes three elements:
...
 
 The value to return if the condition is `FALSE` is optional - if it is not provided and the condition is `FALSE`, then the macro has no effect on the resulting query.
 
+It's also possible to use this macro in order to conditionally execute pre/post-statements:
+
+```sql linenums="1"
+@IF([logical condition], [statement to execute if TRUE])
+```
+
+Note that the statement to be executed must not end with a semi-colon.
```
treysp, line 277: Looking at the example below, is it correct that the `@IF()` statement SHOULD end with a semicolon but the statement within should not?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -260,10 +260,22 @@ It includes three elements:
...
-These elements are specified as `@IF([logical condition], [value if TRUE], [value if FALSE])`.
+These elements are specified as:
+
+```sql linenums="1"
+@IF([logical condition], [value if TRUE], [value if FALSE])
+```
 
 The value to return if the condition is `FALSE` is optional - if it is not provided and the condition is `FALSE`, then the macro has no effect on the resulting query.
 
+It's also possible to use this macro in order to conditionally execute pre/post-statements:
```
treysp, line 271: Is it possible to have both unconditional and conditional pre/post statements? Like if I have one pre-statement that should always run and a second pre-statement that should only run sometimes?

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: That's correct, I can update this note if it doesn't seem clear

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Yeah that's correct. All pre/post-statements are always rendered, but the ones wrapped in `@IF` are simply rendered to `None` IIRC and so they won't included in the execution? I'll need to double check that.

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Updated

treysp (approved):

tobymao (approved):

PR #1588, opened 2023-10-18
Fix: Register source and ref dependencies from macros

crericha:


sqlmesh/utils/jinja.py
```diff
@@ -130,19 +132,19 @@ def find_call_names(node: nodes.Node, vars_in_scope: t.Set[str]) -> t.Iterator[t
             for arg in child_node.args:
                 vars_in_scope.add(arg.name)
         elif isinstance(child_node, nodes.Call):
-            name = call_name(child_node)
+            name = node_name(child_node)
             if name[0][0] != "'" and name[0] not in vars_in_scope:
-                yield name
+                yield (name, child_node)
```
tobymao, line 137: why did you change if it you don't use this on line 147?

crericha: I use it in manifest.py

sqlmesh/utils/jinja.py
```diff
@@ -99,23 +99,25 @@ def _advance(self, times: int = 1) -> None:
         )
 
 
-def call_name(node: nodes.Expr) -> t.Tuple[str, ...]:
+def node_name(node: nodes.Expr) -> t.Tuple[str, ...]:
     if isinstance(node, nodes.Name):
         return (node.name,)
     if isinstance(node, nodes.Const):
-        return (f"'{node.value}'",)
+        return (node.value,)
```
crericha, line 106: @izeigerman why were constants being quoted? All the tests pass if I don't quote, so I'm wondering what it was solving? source args are constants, so the quotes were messing up getting the source package and name. I can add them back in and work around them if the quotes are indeed needed.

sqlmesh/dbt/manifest.py
```diff
@@ -260,6 +270,18 @@ def _load_profile(self) -> Profile:
             target_override=self.target.name,
         )
 
+    def _macro_dependencies(
```
izeigerman, line 273: Why do we need this? Doesn't the top-level node capture all nested macro dependencies already?

sqlmesh/dbt/package.py
```diff
@@ -24,6 +24,7 @@ class MacroConfig(PydanticModel):
     """Class to contain macro configuration"""
 
     info: MacroInfo
+    dependencies: Dependencies
```
izeigerman, line 27: Why do we need this?

sqlmesh/utils/jinja.py
```diff
@@ -103,7 +103,7 @@ def call_name(node: nodes.Expr) -> t.Tuple[str, ...]:
     if isinstance(node, nodes.Name):
         return (node.name,)
     if isinstance(node, nodes.Const):
-        return (f"'{node.value}'",)
+        return (f"'node.value'",)
```
izeigerman, line 106: This is wrong. Please see the previous version. Would you mind adding a test case for this too please?

sqlmesh/utils/jinja.py
```diff
@@ -132,17 +134,17 @@ def find_call_names(node: nodes.Node, vars_in_scope: t.Set[str]) -> t.Iterator[t
...
 
-def extract_call_names(jinja_str: str) -> t.List[t.Tuple[str, ...]]:
+def extract_call_names(jinja_str: str) -> t.List[t.Tuple[t.Tuple[str, ...], nodes.Call]]:
     return list(find_call_names(ENVIRONMENT.parse(jinja_str), set()))
 
 
 def extract_macro_references(jinja_str: str) -> t.Set[MacroReference]:
     result = set()
-    for call_name in extract_call_names(jinja_str):
+    for call_name, _node in extract_call_names(jinja_str):
```
izeigerman, line 147: `_node` is not used, should be a placeholder (`_`)

sqlmesh/dbt/manifest.py
izeigerman: Ah, I guess because we need to capture additional sources and refs now

tests/dbt/test_manifest.py
```diff
@@ -38,6 +42,8 @@ def test_manifest_helper():
     assert waiter_revenue_by_day_config.dependencies == Dependencies(
         macros={
             MacroReference(name="log_value"),
+            MacroReference(name="should_full_refresh"),
```
izeigerman, line 45: Why has this set been updated?

crericha: The `is_incremental` macro calls `should_full_refresh`. dbt did not report the dependency, so it wasn't picked-up originally.

tests/dbt/test_manifest.py
```diff
@@ -20,7 +20,11 @@ def test_manifest_helper():
     assert helper.models()["top_waiters"].materialized == "view"
 
     assert helper.models()["waiters"].dependencies == Dependencies(
-        macros={MacroReference(name="incremental_by_time")},
+        macros={
+            MacroReference(name="incremental_by_time"),
```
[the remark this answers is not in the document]
crericha: Same

PR #1589, opened 2023-10-19
Fix(mysql): use 8-character random ID for temp table names

treysp:
MySQL table names must be <=64 characters long. 

Some SQLMesh temp table names exceed that, so this PR changes the length of the random ID embedded in the table name from a 32-character UUID4 to 8 alphanumeric characters.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ install-dev:
 	pip3 install -e ".[dev,web,slack]"
 
 install-engine-integration:
-	pip3 install -e ".[dev,web,slack,mysql,postgres,databricks,redshift,bigquery,snowflake]"
+	pip3 install -e ".[dev,web,slack,mysql,postgres,databricks,redshift,bigquery,snowflake,trino]"
 
 install-pre-commit:
 	pre-commit install
--- a/sqlmesh/utils/__init__.py
+++ b/sqlmesh/utils/__init__.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import copy
 import importlib
 import os
+import random
 import re
+import string
 import sys
 import time
 import traceback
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ install-dev:
 	pip3 install -e ".[dev,web,slack]"
 
 install-engine-integration:
-	pip3 install -e ".[dev,web,slack,mysql,postgres,databricks,redshift,bigquery,snowflake]"
+	pip3 install -e ".[dev,web,slack,mysql,postgres,databricks,redshift,bigquery,snowflake,trino]"
```
eakmanrq, line 7: nice catch

tests/core/engine_adapter/__init__.py
```diff
@@ -17,3 +17,9 @@ def to_sql_calls(adapter: EngineAdapter, identify: bool = True) -> t.List[str]:
         )
         output.append(sql)
     return output
+
+
+def temp_table_name(table_name: str, random_id: str):
```
eakmanrq, line 22: It looks like this is just used by the tests. If so, it would be better to have the function just exist in the tests then. You can make it be a callable fixture that can be passed in. You will see examples of this where we follow the pattern of `make_*`.

sqlmesh/utils/__init__.py
```diff
@@ -47,7 +49,10 @@ def unique(iterable: t.Iterable[T], by: t.Callable[[T], t.Any] = lambda i: i) ->
     return list({by(i): None for i in iterable})
 
 
-def random_id() -> str:
+def random_id(short: bool = False) -> str:
+    if short:
+        return "".join(random.choices(string.ascii_letters + string.digits, k=8))
```
tobymao, line 54: make this a constant, ascii_letters + string.digits

tests/conftest.py
```diff
@@ -260,3 +260,13 @@ def delete_cache(project_paths: str | t.List[str]) -> None:
             rmtree(path + "/.cache")
         except FileNotFoundError:
             pass
+
+
+@pytest.fixture
+def make_temp_table_name(mocker: MockerFixture) -> t.Callable:
+    def _make_function(table_name: str, random_id: str) -> exp.Table:
+        temp_table = t.cast(exp.Table, exp.to_table(table_name))
```
tobymao, line 268: the cast here isn't necessary

PR #1601, opened 2023-10-24
fix!: Remove Most Fields from dbt Target

eakmanrq:


sqlmesh/dbt/target.py
```diff
@@ -42,6 +42,14 @@
     t.Type[IncrementalUnmanagedKind],
 ]
 
+# We only serialize a subset of fields in order to avoid persisting sensitive information
+SERIALIZABLE_FIELDS = {
+    "type",
+    "name",
+    "database",
+    "schema_",
```
crericha, line 50: Is this field serialized as `schema_` or `schema`. It needs to be the latter.

crericha (approved):

sqlmesh/dbt/target.py
crericha: Unit test confirms its the latter.

eakmanrq: lol was wondering the same

eakmanrq:
@izeigerman do you mind also taking a look when you get a chance (I know you are traveling)? Just want to get another set of eyes since this is breaking/requires a migration.

PR #1603, opened 2023-10-24
Fix: mysql does not support FULL OUTER JOIN

treysp:
Update sushi model customer_revenue_lifetime.sql to use `LEFT JOIN`/`UNION`/`RIGHT JOIN` instead of `FULL OUTER JOIN` so it can run on mysql


tobymao:
i think this can just be a sqlglot transpilation cc @GeorgeSittas and @eakmanrq 

what do you think?

full join into left right union

tobymao:
let me see how easy this

eakmanrq:
Yeah handling this in transpilation would be ideal.

GeorgeSittas:
+1, seems like a good improvement for sqlglot

PR #1604, opened 2023-10-24
chore: include schema version in cache key

eakmanrq:
For end users this shouldn't matter since the SQLMesh version will change with new migration scripts this does help with local testing where we add a migration script but we still get cached results because no version change has happened.

### DIFF
--- a/sqlmesh/utils/cache.py
+++ b/sqlmesh/utils/cache.py
@@ -107,5 +107,15 @@ class FileCache(t.Generic[T]):
             major, minor = __version_tuple__[0], __version_tuple__[1]
         except ImportError:
             major, minor = 0, 0
-        entry_file_name = f"{name}__{major}__{minor}__{SQLGLOT_MAJOR_VERSION}__{SQLGLOT_MINOR_VERSION}__{SCHEMA_VERSION}__{entry_id}"
+        entry_file_name = "__".join(
+            [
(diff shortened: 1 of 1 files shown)


sqlmesh/utils/cache.py
```diff
@@ -99,11 +99,23 @@ def put(self, name: str, entry_id: str, value: T) -> None:
...
             major, minor = 0, 0
-        entry_file_name = f"{name}__{major}__{minor}__{SQLGLOT_MAJOR_VERSION}__{SQLGLOT_MINOR_VERSION}__{entry_id}"
+        entry_file_name = "__".join(
+            [
+                name,
+                str(major),
+                str(minor),
+                SQLGLOT_MAJOR_VERSION,
+                SQLGLOT_MINOR_VERSION,
+                str(SCHEMA_VERSION),
```
tobymao, line 117: why don't we have the sqlmesh cache bust on every sqlmesh change, which is more frequent than schema version

eakmanrq: What do you mean by every sqlmesh change?

tobymao: use sqlmesh version instead of schema version

eakmanrq: we resolved in chat

PR #1605, opened 2023-10-24
Fix: Get dbt classes without all config fields present

crericha:
Now that we don't store all target config fields in the snapshot, we need to be able to get the dbt classes (relation, column, and quote_policy) without instantiating the TargetConfig class.

### DIFF
--- a/sqlmesh/dbt/target.py
+++ b/sqlmesh/dbt/target.py
@@ -111,20 +111,20 @@ class TargetConfig(abc.ABC, DbtConfig):
         fields["target_name"] = self.name
         return AttributeDict(fields)
 
-    @property
-    def quote_policy(self) -> Policy:
+    @classmethod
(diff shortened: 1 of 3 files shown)


sqlmesh/dbt/target.py
```diff
@@ -111,20 +111,20 @@ def attribute_dict(self) -> AttributeDict:
         fields["target_name"] = self.name
         return AttributeDict(fields)
 
-    @property
-    def quote_policy(self) -> Policy:
+    @classmethod
```
eakmanrq, line 114: These all can be classproperties: https://github.com/TobikoData/sqlmesh/blob/b62dfab5b6f741486c9b3dbd69b6a3a1dfe6145b/sqlmesh/utils/__init__.py#L237

eakmanrq (approved):

PR #1608, opened 2023-10-25
 Fix: improve detection of mocked star column 

GeorgeSittas:
The previous check in `update_schema` where we rendered and then checked for a mocked star projection was insufficient. For example, if a user transforms the column names returned by `MacroEvaluator.columns_to_types`, then the equality check with `SQLMESH_MOCKED_STAR` will never hold.

This PR makes the cache clearing check more explicit, by setting special metadata & propagating it to the model when the `columns_to_types` method is invoked, instead of relying on the rendered query's form which can be anything.

cc @z3z1ma

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -111,6 +111,7 @@ class MacroEvaluator:
         jinja_env: t.Optional[Environment] = None,
         schema: t.Optional[t.Dict[str, t.Any]] = None,
         runtime_stage: RuntimeStage = RuntimeStage.LOADING,
+        query_meta: t.Dict[str, t.Any] = {},
     ):
         self.dialect = dialect
         self.generator = MacroDialect().generator()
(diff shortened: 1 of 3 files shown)


sqlmesh/core/macros.py
```diff
@@ -120,6 +121,7 @@ def __init__(
         self._jinja_env: t.Optional[Environment] = jinja_env
         self.macros = {normalize_macro_name(k): v.func for k, v in macro.get_registry().items()}
         self._schema = MappingSchema(schema, dialect=dialect, normalize=False) if schema else {}
+        self._query_meta = query_meta
```
tobymao, line 124: i don't get how this is propogated to query.meta

GeorgeSittas: same dict refence

tobymao: but on line 182 you do self.query.meta, how is that related to this _query_meta

GeorgeSittas: when you create a renderer, you pass in `self.query` which is stored in the renderer's `self.expression` attribute and so `self._expression.meta` is the exact same dict object as `self.query.meta`, where the former one is passed to the macro evaluator and stored in `self._query_meta`

tobymao:
alternative https://github.com/TobikoData/sqlmesh/pull/1615

PR #1609, opened 2023-10-25
Feat: batch snapshot querying to prevent query limit issues

tobymao:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -360,56 +361,55 @@ def _get_snapshots(
...
             A dictionary of snapshot ids to snapshots for ones that could be found.
         """
-        query = (
-            exp.select(exp.column("snapshot", table="snapshots"))
-            .from_(exp.to_table(self.snapshots_table).as_("snapshots"))
-            .where(
-                None
-                if snapshot_ids is None
-                else self._snapshot_id_filter(snapshot_ids, "snapshots")
+        for where in (
```
izeigerman, line 364: Doesn't this only return snapshots from the last batch since we do:
```
snapshots: t.Dict[SnapshotId, Snapshot] = {}
```
on each iteration?

tests/core/test_state_sync.py
```diff
@@ -1341,5 +1346,21 @@ def test_get_snapshots(mocker):
...
-    cache.get_snapshots(None)
-    mock.get_snapshots.assert_called()
+
+def test_snapshot_batching(state_sync, mocker):
+    mock = mocker.Mock()
+
+    state_sync.SNAPSHOT_BATCH_SIZE = 2
+    state_sync.engine_adapter = mock
+
+    state_sync.delete_snapshots(
```
izeigerman, line 1356: Can we also have a test for `get_snapshots` ?

tests/core/test_state_sync.py
```diff
@@ -1341,5 +1360,55 @@ def test_get_snapshots(mocker):
...
+        [
+            [
+                make_snapshot(
+                    SqlModel(name="a", query=parse_one("select 3")),
+                ).json(),
+            ],
+        ],
+    ]
+
+    snapshots = state_sync._get_snapshots(
```
izeigerman, line 1404: [Nit] Maybe a test for intervals too?

izeigerman (approved):

tests/core/test_state_sync.py
tobymao: i added modified the an interval test to make sure that code path is hit

PR #1610, opened 2023-10-25
Feat!: Differentiate between deployable and non-deployable snapshots during plan evaluation

izeigerman:
Previously we used the `is_dev` flag everywhere to determine the following:

1. Whether to use dev intervals or regular intervals when computing missing intervals.
2. Whether to reference a prod table or a dev / temp table when evaluating a snapshot (physical table mapping).
3. Whether to write into a prod table or a dev / temp table.
4. Order of the backfill during plan evalution.

Using `is_dev` had the following shortcomings:

1. We had to repeat the same logic in many places. Eg. something `if is_dev and snapshot.is_paused and (snapshot.is_forward_only or snapshot.is_non_breaking)...` had to be repeated all over the code base.
2. The decision about whether to use a prod table or a dev / temp table could only be derived from the current state of an individual snapshot. There was no way to impact the decision based on the current state of the DAG as a whole. We found use-cases in which the state of parent snapshots may impact this decision.

In essence, during evaluation, each snapshot can be in 1 of 3 states:
1. Deployable: this means that the output that snapshot produces can be safely reused when promoted to prod.
2. Non-deployable: this means that the output can't be safely reused in prod and should be discarded.
3. Representative: this means that the snapshot might not be deployable due to its nature (eg. `FORWARD_ONLY` or `INDIRECT_NON_BREAKING`), but since it's currently promoted in prod, we should refer to its prod table when computing intervals or determining the physical table mapping. Therefore, **all** _deployable_ snapshots are also _representative_, but the opposite is not always true.

To the best of my knowledge, these three states encompass all the use cases we're dealing with here.

To identify which state each snapshot is in, I created a new entity called `DeployabilityIndex`. It is built once at the beginning of plan evaluation and then used unchanged throughout the process. It's constructed using the entire DAG which means that if a parent is non-deployable this property also propagates to its children.

Thus, `DeployabilityIndex` is now a single source of truth for whether or not a snapshot produces deployable results, while the clients no longer need to worry about the current snapshot state (eg. whether it's paused or not) or which target environment it is.

Additionally, `DeployabilityIndex` presents an opportunity to inform a user at plan creation time about whether or not what they're about to build is going to be deployable. This, however, is out of scope of this PR.


sqlmesh/core/snapshot/definition.py
```diff
@@ -1035,12 +1008,171 @@ def _ensure_categorized(self) -> None:
 SnapshotNameVersionLike = t.Union[SnapshotNameVersion, SnapshotTableInfo, Snapshot]
 
 
-def table_name(physical_schema: str, name: str, version: str, is_temp: bool = False) -> str:
+class DeployabilityIndex(PydanticModel, frozen=True):
```
tobymao, line 1011: small nit but I'd prefer like DeployabilityMap or something else, index reminds me of array indexing

sqlmesh/core/snapshot/definition.py
```diff
@@ -1035,12 +1008,171 @@ def _ensure_categorized(self) -> None:
...
+        }
+        non_deployable_ids = set(snapshots) - deployable_ids
+
+        # Pick the smaller set to reduce the size of the serialized object.
+        if len(deployable_ids) <= len(non_deployable_ids):
+            return cls(
+                deployable_ids=deployable_ids,
+                deployed_shared_version_ids=deployed_shared_version_ids,
+            )
+        else:
```
tobymao, line 1159: else isn't needed

sqlmesh/core/snapshot/definition.py
izeigerman: Isn't it more like database index? We have IDs, lookup, etc.

izeigerman: Sorry, I'm confused. There is neither map nor dict in there :D

tobymao: ah right

sqlmesh/core/snapshot/definition.py
```diff
@@ -1040,12 +1018,179 @@ class SnapshotTableCleanupTask(PydanticModel):
...
-def table_name(physical_schema: str, name: str, version: str, is_temp: bool = False) -> str:
+class DeployabilityIndex(PydanticModel, frozen=True):
+    """Contains information about deployability of every snapshot.
+
+    Deployability is defined as whether or not the output that a snapshot produces during the
+    current evaluation can be reused in (deployed to) the production environment.
+    """
+
+    deployable_ids: t.Optional[t.FrozenSet[str]] = None
+    non_deployable_ids: t.Optional[t.FrozenSet[str]] = None
```
eakmanrq, line 1029: Could this be simplified to be a single non-optional `FrozenSet[str]` and then a flag that says if it is deployable or non-deployable?

sqlmesh/core/plan/definition.py
```diff
@@ -426,6 +427,12 @@ def _missing_intervals(self) -> t.Dict[t.Tuple[str, str], Intervals]:
                 if new.is_forward_only:
                     new.dev_intervals = new.intervals.copy()
 
+            deployability_index = (
+                DeployabilityIndex.create({s.snapshot_id: s for s in self.snapshots})
```
tobymao, line 431: would probably clean things up if create took in both a collection of snapshots and also the dict

sqlmesh/core/snapshot/definition.py
```diff
@@ -1040,12 +1011,166 @@ class SnapshotTableCleanupTask(PydanticModel):
...
+        """Returns true if the output produced by the given snapshot in a development environment can be reused
+        in (deployed to) production
+
+        Args:
+            snapshot: The snapshot to check.
+
+        Returns:
+            True if the snapshot is deployable, False otherwise.
+        """
+        snapshot_id = self._snapshot_id_key(snapshot.snapshot_id)
```
tobymao, line 1050: return (self.is_opposite_index and snapshot_id not in self.indexed_ids) or (not self.is_opposite and snapshot_id in self.indexed_ids)

sqlmesh/core/snapshot/definition.py
```diff
@@ -1040,12 +1011,166 @@ class SnapshotTableCleanupTask(PydanticModel):
...
+        them as such when constructing a plan, building a physical table mapping or computing missing intervals.
+
+        Args:
+            snapshot: The snapshot to check.
+
+        Returns:
+            True if the snapshot is representative, False otherwise.
+        """
+        snapshot_id = self._snapshot_id_key(snapshot.snapshot_id)
+        deployed = (
```
tobymao, line 1072: shouldn't this variable be representative

sqlmesh/core/snapshot/definition.py
```diff
@@ -1040,12 +1011,166 @@ class SnapshotTableCleanupTask(PydanticModel):
...
+    @classmethod
+    def all_deployable(cls) -> DeployabilityIndex:
+        return cls(indexed_ids=frozenset(), is_opposite_index=True)
+
+    @classmethod
+    def none_deployable(cls) -> DeployabilityIndex:
+        return cls(indexed_ids=frozenset())
+
+    @classmethod
+    def create(cls, snapshots: t.Dict[SnapshotId, Snapshot]) -> DeployabilityIndex:
```
tobymao, line 1102: snapshots: t.Dict[SnapshotId, Snapshot] | t.Iterable[Snapshot]

sqlmesh/core/snapshot/definition.py
```diff
@@ -1040,12 +1011,166 @@ class SnapshotTableCleanupTask(PydanticModel):
...
+        dag = snapshots_to_dag(snapshots.values())
+        reversed_dag = dag.reversed.graph
+
+        deployability_mapping: t.Dict[SnapshotId, bool] = {}
+        representative_shared_version_ids: t.Set[SnapshotId] = set()
+
+        def _visit(node: SnapshotId, deployable: bool = True) -> None:
+            if node in deployability_mapping and (
```
tobymao, line 1110: if deployability_mapping.get(node) in (False, deployable)

tobymao (approved):

PR #1611, opened 2023-10-25
Chore: raise if we can't make progress when topo-sorting a DAG

GeorgeSittas:
This is just a sanity check that helps detecting infinite loops.

### DIFF
--- a/sqlmesh/utils/dag.py
+++ b/sqlmesh/utils/dag.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 
 import typing as t
 
+from sqlmesh.utils.errors import SQLMeshError
+
 T = t.TypeVar("T", bound=t.Hashable)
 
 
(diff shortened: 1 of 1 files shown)


sqlmesh/utils/dag.py
```diff
@@ -93,6 +95,9 @@ def sorted(self) -> t.List[T]:
             while unprocessed_nodes:
                 next_nodes = {node for node, deps in unprocessed_nodes.items() if not deps}
 
+                if not next_nodes:
+                    raise SQLMeshError("Failed to topologically sort a DAG.")
```
tobymao, line 99: maybe a better error message like

"Detected a cycle in the DAG. Please check for cyclical references"

GeorgeSittas: I'm not sure if this is related to cycles, though. The issue at hand is related to Snowflake and my guess is that `deps` doesn't get updated in `deps -= next_nodes` just below due to inconsistencies in the name normalization.

I'll investigate a bit more and update this thread.

GeorgeSittas: Updated the message to hint about circular references after all.

PR #1613, opened 2023-10-25
Fix(mssql, snowflake): convert datetime64 with timezone

treysp:
The MSSQL and Snowflake connector libraries do not handle pandas datetime64 columns correctly, so we convert them to strings and let the engine convert to the proper table data type when inserting the data.

The current implementation uses `is_datetime64_dtype()` to detect datetime64 columns, but it returns `False` for datetime64 columns with a time zone. 

This PR replaces it with the more general `is_datetime64_any_dtype()` to detect columns and includes the time zone UTC offset in the string when the column has a timezone specified.

### DIFF
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import typing as t
 
 import pandas as pd
-from pandas.api.types import is_datetime64_dtype  # type: ignore
+from pandas.api.types import is_datetime64_any_dtype  # type: ignore
 from sqlglot import exp
 from sqlglot.optimizer.qualify_columns import quote_identifiers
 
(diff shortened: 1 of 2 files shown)


tobymao (approved):

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -185,9 +185,12 @@ def query_factory() -> Query:
             # pymssql doesn't convert Pandas Timestamp (datetime64) types
             # - this code is based on snowflake adapter implementation
             for column, kind in (columns_to_types or {}).items():
-                if kind.is_type("date") and is_datetime64_dtype(df.dtypes[column]):  # type: ignore
+                if kind.is_type("date") and is_datetime64_any_dtype(df.dtypes[column]):  # type: ignore
```
eakmanrq, line 188: if I am seeing this correct all of these conditions check `is_datetime64_any_dtype(df.dtypes[column])`. Although it would increase nesting, I think it would be easier to read if we did an if check on that and then did the following checks inside:
```
if is_datetime64_any_dtype(df.dtypes[column]):
   if kind.is_type("date"):
        ....
   elif getattr(df.dtypes[column], "tz", None) is not None:
        ...
   else:
      ....
```

eakmanrq (approved):
Do you have a test that captures this expected functionality? Not worth holding up merging the PR but it would be nice to validate the expected behavior.

PR #1615, opened 2023-10-25
Fix: skip the cache if columns_to_types is called in a macro

tobymao:


tobymao:
cc @z3z1ma

GeorgeSittas:
Clever approach, nice!

PR #1617, opened 2023-10-26
chore: add local spark to integration tests

eakmanrq:
Adds Spark to our integration tests and since it is deployed locally it means it will be tested on all merges to main. 

I ended up adding the local config as a checked in YAML so now users just need to have connection config to external services defined locally. 

Note: Iceberg was just included because I needed a way to add a catalog and I was pulling my hair out trying to figure out how to do it when the catalog isn't a Iceberg/Delta catalog. I just eventually gave up and made it an Iceberg catalog. The tests involving creating objects or loading data are all done agains the default catalog and therefore it is reading/writing Parquet files to "S3" (Minio).

### DIFF
--- a/sqlmesh/core/engine_adapter/shared.py
+++ b/sqlmesh/core/engine_adapter/shared.py
@@ -101,7 +101,7 @@ def set_catalog(
                 exp.to_table if obj_type == "TableName" else to_schema,
             )
             expression = to_expression_func(obj.copy() if isinstance(obj, exp.Table) else obj)
-            catalog_name = expression.args.get("catalog")
+            catalog_name = expression.catalog
             if not catalog_name:
                 return func(*list_args, **kwargs)
             # If we have a catalog and this engine doesn't support catalogs then we need to error
--- a/tests/core/engine_adapter/test_integration.py
+++ b/tests/core/engine_adapter/test_integration.py
@@ -1,6 +1,7 @@
 # type: ignore
 from __future__ import annotations
 
+import os
 import pathlib
 import sys
 import typing as t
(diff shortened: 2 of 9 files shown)


sqlmesh/core/engine_adapter/shared.py
```diff
@@ -101,7 +101,7 @@ def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
                 exp.to_table if obj_type == "TableName" else to_schema,
             )
             expression = to_expression_func(obj.copy() if isinstance(obj, exp.Table) else obj)
-            catalog_name = expression.args.get("catalog")
+            catalog_name = expression.catalog
```
eakmanrq, line 104: Bug I found while testing. This would be an identifier before but now will be a string (if it is defined).

tests/core/engine_adapter/test_integration.py
```diff
@@ -700,6 +715,10 @@ def test_merge(ctx: TestContext):
 def test_scd_type_2(ctx: TestContext):
     if ctx.dialect == "tsql":
         pytest.skip(f"MSSQL scd type 2 functionality waiting on sqlglot cte in FROM fix")
+    if ctx.dialect == "spark":
+        pytest.skip(
+            "Spark SCD Type 2 does not currently work due to this error: `[UNSUPPORTED_OVERWRITE.TABLE] Can't overwrite the target that is also being read from`"
```
eakmanrq, line 720: https://github.com/TobikoData/sqlmesh/issues/1618

tobymao (approved):

PR #1622, opened 2023-10-26
Fix: Only capture extra refs and sources that are statically defined and enabled

crericha:


sqlmesh/dbt/manifest.py
```diff
@@ -353,29 +400,7 @@ def _convert_jinja_test_to_macro(test_jinja: str) -> str:
...
-        elif call_name[0] == "ref":
-            ref = ".".join(_jinja_call_arg_name(arg) for arg in node.args)
-            if ref:
-                dependencies.refs.append(ref)
-
-    return dependencies
-
-
 def _jinja_call_arg_name(node: nodes.Node) -> str:
-    if isinstance(node, nodes.Name):
```
tobymao, line 377: why was this removed?

sqlmesh/dbt/manifest.py
```diff
@@ -63,6 +63,8 @@ def __init__(
         self._macros_per_package: t.Dict[str, MacroConfigs] = defaultdict(dict)
 
         self._tests_by_owner: t.Dict[str, t.List[TestConfig]] = defaultdict(list)
+        self._disabled_refs: t.Set[str] = set()
```
tobymao, line 66: this should be optional and set to None

sqlmesh/dbt/manifest.py
```diff
@@ -270,6 +267,33 @@ def _load_profile(self) -> Profile:
             target_override=self.target.name,
         )
 
+    def _disabled_ref(self, ref: str) -> bool:
+        if not self._disabled_refs:
```
tobymao, line 271: if self._disabled_refs is None:
    for ... 

if you have no disabled refs, you're going to be constantly reloading the manifest with this current code

sqlmesh/dbt/manifest.py
```diff
@@ -270,6 +267,33 @@ def _load_profile(self) -> Profile:
...
+            self._load_disabled()
+
+        return ref in self._disabled_refs
+
+    def _disabled_source(self, source: str) -> bool:
+        if not self._disabled_sources:
```
tobymao, line 277: make the same change here, just do it inline here

sqlmesh/dbt/manifest.py
crericha: This is the crux of the error that was reported. Since we are statically parsing, we can only resolve constants - `source("package", "source")`, not vars - `source(package_var, source_table_var)`.

tobymao: so what happens if a var is passed?

crericha: We don't include it in our supplemental dependency detection. It was never included before. This is all new since yesterday. The majority of source and ref calls are defined with constants, so that's the best bang for our buck. In the future, if we find dbt is missing a lot more dependencies from the manifest, we can look into adding spend more time here.

sqlmesh/dbt/manifest.py
```diff
@@ -270,6 +267,35 @@ def _load_profile(self) -> Profile:
             target_override=self.target.name,
         )
 
+    def _disabled_ref(self, ref: str) -> bool:
```
izeigerman, line 270: [Nit] `is_ref_disabled`. This name looks like it returns a ref

sqlmesh/dbt/manifest.py
```diff
@@ -270,6 +267,35 @@ def _load_profile(self) -> Profile:
...
+        if self._disabled_refs is None:
+            self._load_disabled()
+
+        return ref in self._disabled_refs  # type: ignore
+
+    def _disabled_source(self, source: str) -> bool:
```
izeigerman, line 276: ditto

tobymao (approved):

sqlmesh/dbt/manifest.py
crericha: I secretly want a question mark in the method name

izeigerman (approved):

PR #1624, opened 2023-10-27
Fix: start dates of models are now relative to cron

tobymao:
if a project does not have a start date set, models with monthly cron would never run because the default start date was yesterday.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -843,7 +843,7 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
     def is_valid_start(
         self,
         start: t.Optional[TimeLike],
-        snapshot_start: t.Optional[TimeLike] = None,
+        snapshot_start: TimeLike,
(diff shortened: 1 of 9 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -1258,40 +1255,44 @@ def missing_intervals(
...
-    snapshots = list(snapshots)
-    earliest = to_datetime(default_value or yesterday().date())
     if snapshots:
-        return min(start_date(snapshot, snapshots, cache) or earliest for snapshot in snapshots)
-    return earliest
+        return min(
+            start_date(snapshot, snapshots, cache=cache, relative_to=relative_to)
+            for snapshot in snapshots
+        )
+    return yesterday()
```
izeigerman, line 1277: [Nit] Should we just raise if snapshots are empty?

tobymao: thought about it and think it may be dangerous

PR #1628, opened 2023-10-28
fix: Have Python depends_on Actually Override

eakmanrq:


sqlmesh/core/model/definition.py
```diff
@@ -1726,7 +1726,7 @@ def create_python_model(
     depends_on = (
         _parse_depends_on(entrypoint, python_env) - {name}
         if depends_on is None and python_env is not None
-        else None
+        else depends_on
```
tobymao, line 1729: what happened here? bug?

tobymao (approved):

sqlmesh/core/model/definition.py
eakmanrq: Yeah I believe this is a bug. If this function was called with a `depends_on` it would have been ignored because of this `else None`.

PR #1630, opened 2023-10-30
fix: duckdb init catalogs at cursor level

eakmanrq:
Prior approach didn't take into account that a connection could be closed (like after running tests) and once re-opened it wouldn't have the catalogs applied. Originally in another PR I modified the connection factory to init the catalogs but that doesn't work because the cursor object created by DuckDB is unique from the connection object from which it is created (despite them saying it is a duplicate). Therefore we need to make sure that whenever a cursor is created that it gets the catalog initialized.

This is more thoroughly tested in the default catalog PR that actually uses these catalogs and ensures the connection is configured correctly.


tobymao:
i'll let @izeigerman look at this because it deals with connection pools but lgtm

PR #1635, opened 2023-10-30
Chore: improve error message when cycle is detected in model graph

GeorgeSittas:
This is with the changes introduced in this PR:

```
➜  cat models/a.sql
model (name a, kind full);

select id from b
➜ cat models/b.sql
model (name b, kind full);

select id from a
➜  cat models/normal.sql
model (name normal, kind full);

select 1 as a%
➜  sqlmesh info
Error: Detected a cycle in the DAG. Please make sure there are no circular references between models.
Last models added to the DAG: normal
Possible model candidates to check for circular references: b, a
```

If I get rid of the `normal.sql` model, then the error message becomes:
```
➜  sqlmesh info
Error: Detected a cycle in the DAG. Please make sure there are no circular references between models.
Possible model candidates to check for circular references: b, a
```

### DIFF
--- a/sqlmesh/utils/dag.py
+++ b/sqlmesh/utils/dag.py
@@ -100,9 +100,10 @@ class DAG(t.Generic[T]):
                 next_nodes = {node for node, deps in unprocessed_nodes.items() if not deps}
 
                 if not next_nodes:
+                    # Sort cycle candidates to make the order deterministic
                     cycle_candidates_msg = (
                         "\nPossible candidates to check for circular references: "
-                        + ", ".join(str(node) for node in cycle_candidates)
+                        + ", ".join(str(node) for node in sorted(cycle_candidates))
                     )
 
                     if last_processed_nodes:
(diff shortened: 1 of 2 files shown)


sqlmesh/utils/dag.py
```diff
@@ -92,13 +92,28 @@ def sorted(self) -> t.List[T]:
...
                 next_nodes = {node for node, deps in unprocessed_nodes.items() if not deps}
 
                 if not next_nodes:
+                    cycle_candidates_msg = (
+                        "\nPossible model candidates to check for circular references: "
+                        + ", ".join(str(node) for node in unprocessed_nodes)
+                    )
+
+                    if last_processed_nodes:
+                        last_processed_msg = "\nLast models added to the DAG: " + ", ".join(
```
izeigerman, line 107: Just a heads up, that the `DAG` implementation is pretty generic and may not just contain models. Should we reflect this fact in the error message and change `model` to eg. `node`?

GeorgeSittas: That makes sense, though it may make the error message a bit more cryptic because it might not be obvious that "nodes" are essentially models. I don't have a strong opinion on this though, happy to go either way.

GeorgeSittas: Btw I chatted with Chris and I'm looking into alternative metrics that are more helpful when troubleshooting circular DAGs. For example, display the last processed nodes & those nodes that had no dependencies removed in the last iteration.

GeorgeSittas: Replaced to use `node` instead of `model`

izeigerman (approved):
Can we maybe add a test for this?

GeorgeSittas:
Yes, definitely. Planning to do it in a bit.

GeorgeSittas:
Added a few unit tests

tobymao (approved):

PR #1638, opened 2023-10-31
Fix!: show proper pip install command when sqlmesh schema version is behind

GeorgeSittas:
Current behavior:

```
Error: SQLMesh (local) is using version '30' which is behind '31' (remote). Please upgrade SQLMesh ('pip install --upgrade "sqlmesh==31"' command).
```

The suggested command is incorrect because we're interpolating `versions.schema_version` instead of the actual SQLMesh package version. To overcome this issue I added a new column `sqlmesh_version` to the `_versions` state table which is currently only used to improve the UX when an error message is raised.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -37,6 +37,7 @@ class Versions(PydanticModel):
 
     schema_version: int
     sqlglot_version: str
+    sqlmesh_version: str
 
     @property
     def minor_sqlglot_version(self) -> t.Tuple[int, int]:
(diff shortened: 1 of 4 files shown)


tobymao (approved):

sqlmesh/core/state_sync/base.py
```diff
@@ -158,17 +159,30 @@ def get_versions(self, validate: bool = True) -> Versions:
...
             if ahead:
                 raise SQLMeshError(
-                    f"{lib} (local) is using version '{local}' which is ahead of '{remote}' (remote). Please run a migration ('sqlmesh migrate' command)."
+                    f"{lib} (local) is using version '{local}' which is ahead of '{remote}' (remote). "
+                    "Please run a migration ('sqlmesh migrate' command)."
                 )
             raise SQLMeshError(
-                f"{lib} (local) is using version '{local}' which is behind '{remote}' (remote). Please upgrade {lib} ('pip install --upgrade \"{lib.lower()}=={remote}\"' command)."
+                f"{lib} (local) is using version '{local}' which is behind '{remote}' (remote). "
+                f"Please upgrade {lib} ('pip install --upgrade \"{lib.lower()}=={remote_package_version or remote}\"' command)."
```
izeigerman, line 176: Why even fallback on `remote`? Isn't it meaningless?

GeorgeSittas: It's used when SQLGlot local version is behind

GeorgeSittas: We could be explicit though and use the `remote_package_version` instead of falling back to `remote`.

izeigerman: ah, I see

izeigerman: it's really confusing that we use `remote_package_version` for sqlmesh and `remote` for sqlglot

GeorgeSittas: tried to clean this up here: https://github.com/TobikoData/sqlmesh/pull/1638/commits/f6a789795901423736e03e72ebfa9bd607ac8b34

izeigerman (approved):

PR #1640, opened 2023-10-31
feat: Add Single Catalog Support to Postgres Engines

eakmanrq:
Postgres engines allow only creating objects within the catalog assigned in the connection itself. Therefore they do support catalogs but only creating objects within the connection catalog. This change will make sure that the catalog being targeted matches the connection catalog and if so it will remove the catalog and perform the operation.

Redshift allows querying across catalogs (not creating objects) but this distinction does not matter for the `set_catalog` decorator since it is focused on the objects actually being created and not the expressions being used.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import os
+import pathlib
 import sys
 import typing as t
 from enum import Enum
@@ -78,9 +79,20 @@ class ConnectionConfig(abc.ABC, BaseConfig):
             ),
             multithreaded=self.concurrent_tasks > 1,
             cursor_kwargs=self._cursor_kwargs,
+            default_catalog=self.get_catalog(),
             cursor_init=self._cursor_init,
             **self._extra_engine_config,
         )
+
+    def get_catalog(self) -> t.Optional[str]:
+        """The catalog for this connection"""
+        if hasattr(self, "catalog"):
+            return self.catalog
+        if hasattr(self, "database"):
+            return self.database
+        if hasattr(self, "db"):
+            return self.db
+        return None
 
 
 class DuckDBConnectionConfig(ConnectionConfig):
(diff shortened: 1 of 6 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -87,13 +87,18 @@ def requires_condition(self) -> bool:
 
 class CatalogSupport(Enum):
     UNSUPPORTED = 1
-    REQUIRES_SET_CATALOG = 2
-    FULL_SUPPORT = 3
+    SINGLE_CATALOG_ONLY = 2
+    REQUIRES_SET_CATALOG = 3
+    FULL_SUPPORT = 4
```
eakmanrq, line 92: These enums are not persisted anywhere so they can be switched around like this.

sqlmesh/core/config/connection.py
```diff
@@ -78,10 +79,21 @@ def create_engine_adapter(self) -> EngineAdapter:
...
+            default_catalog=self.get_catalog(),
             cursor_init=self._cursor_init,
             **self._extra_engine_config,
         )
 
+    def get_catalog(self) -> t.Optional[str]:
+        """The catalog for this connection"""
+        if hasattr(self, "catalog"):
+            return self.catalog
+        elif hasattr(self, "database"):
```
izeigerman, line 91: [Nit] `elif` not needed

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -112,6 +112,13 @@ def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
             # Remove the catalog name from the argument so the engine adapter doesn't try to use it
             expression.set("catalog", None)
             container[key] = expression  # type: ignore
+            if catalog_support.is_single_catalog_only:
+                if catalog_name != engine_adapter.default_catalog:
+                    raise UnsupportedCatalogOperationError(
+                        f"{engine_adapter.dialect} requires that all catalog operations be against a single catalog: {engine_adapter.default_catalog}"
+                    )
+                else:
```
izeigerman, line 120: `else` is not necessary

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -112,6 +112,13 @@ def wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
             # Remove the catalog name from the argument so the engine adapter doesn't try to use it
             expression.set("catalog", None)
             container[key] = expression  # type: ignore
+            if catalog_support.is_single_catalog_only:
+                if catalog_name != engine_adapter.default_catalog:
+                    raise UnsupportedCatalogOperationError(
```
izeigerman, line 117: Do we need a test for this?

izeigerman (approved):

PR #1642, opened 2023-10-31
Fix(spark): support replacing self-referencing queries

### LINKED ISSUE #1618: Add SCD Type 2 Support to Spark
We currently get this error when using Spark: `[UNSUPPORTED_OVERWRITE.TABLE] Can't overwrite the target that is also being read from`

Workaround could be creating a temp table and then loading from that temp table.

treysp:
Closes #1618


tobymao (approved):
cc @eakmanrq

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -338,6 +342,21 @@ def replace_query(
         columns_to_types = columns_to_types or self.columns(table_name)
         if not columns_to_types:
             raise SQLMeshError("Cannot replace table without columns to types")
+
+        # Self-referential queries: cannot insert overwrite a SELECT from itself, so
+        # use LogicalReplaceQuery (which creates a temp table and SELECTs from it)
+        if len(source_queries) > 1:
+            raise SQLMeshError("Cannot replace table with a batched data frame")
```
eakmanrq, line 349: [Nit]: `dataframe` instead of `data frame`

tests/core/engine_adapter/test_mixins.py
```diff
@@ -54,32 +54,35 @@ def test_logical_replace_query_does_not_exist(
...
+        )
+
+        adapter.replace_query(table_name, parse_one(f"SELECT col + 1 AS col FROM {table_name}"))
+
+        assert to_sql_calls(adapter) == [
+            f"CREATE SCHEMA IF NOT EXISTS {id_char}db{id_char}",
+            f"CREATE TABLE IF NOT EXISTS {id_char}db{id_char}.{id_char}__temp_table_{temp_table_id}{id_char} AS SELECT {id_char}col{id_char} FROM {id_char}db{id_char}.{id_char}table{id_char}",
+            f"TRUNCATE {id_char}db{id_char}.{id_char}table{id_char}",
+            f"INSERT INTO {id_char}db{id_char}.{id_char}table{id_char} ({id_char}col{id_char}) SELECT {id_char}col{id_char} + 1 AS {id_char}col{id_char} FROM {id_char}db{id_char}.{id_char}__temp_table_{temp_table_id}{id_char}",
+            f"DROP TABLE IF EXISTS {id_char}db{id_char}.{id_char}__temp_table_{temp_table_id}{id_char}",
```
eakmanrq, line 84: What do you think about leaving this test as is and then adding a new `replace_query` test to Spark that has a select reference query and then verify that it then outputs SQL in the above pattern? That way when I look at the Spark tests I would see tests for regular replace_where query, replace_where dataframe, and then replce_where self-reference: https://github.com/TobikoData/sqlmesh/blob/e2f498bfc99fe909663492e686e19e00eef9468d/tests/core/engine_adapter/test_spark.py#L108-L129

eakmanrq (approved):

PR #1644, opened 2023-11-01
Chore: update docs to include MacroEvaluator.columns_to_types

GeorgeSittas:
cc @z3z1ma

### DIFF
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -962,7 +962,7 @@ Having access to the schema of an upstream model can be useful for various reaso
 - Selecting only a subset of columns that satisfy some criteria (e.g. columns whose names start with a specific prefix)
 - Applying transformations to columns, such as masking PII or computing various statistics based on the column types
 
-It also encourages writing code according to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, as one can implement these transformations in a single function instead of manually duplicating them for each model of interest.
+Thus, leveraging `columns_to_types` can also enable one to write code according to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, as they can implement these transformations in a single function instead of duplicating them in each model of interest.
 
 ## Mixing macro systems
 
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -922,7 +922,7 @@ from sqlglot import exp
 from sqlmesh.core.macros import macro
 
 @macro()
-def foo(evaluator, model_name, prefix):
+def prefix_and_cast_columns(evaluator, model_name, prefix):
     prefix = prefix.name
     renamed_projections = []
 
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -940,7 +940,7 @@ def prefix_columns(evaluator, model_name, prefix):
     return renamed_projections
 ```
 
-One example of this macro being used in the query of a SQL model is shown below:
+This can then be used in a SQL model as shown below:
 
 ```sql linenums="1"
 MODEL (
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -959,8 +959,8 @@ Note that `columns_to_types` expects an _unquoted model name_, such as `schema.p
 Having access to the schema of an upstream model can be useful for various reasons:
 
 - Renaming columns so that downstream consumers are not tightly coupled to external or source tables
-- Selecting only a subset of columns that satisfy some criteria (e.g., columns whose names start with a specific prefix)
-- Applying transformations to columns, such as masking PII or computing different statistics based on the column types
+- Selecting only a subset of columns that satisfy some criteria (e.g. columns whose names start with a specific prefix)
+- Applying transformations to columns, such as masking PII or computing various statistics based on the column types
 
 It also encourages writing code according to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle, as one can implement these transformations in a single function instead of manually duplicating them for each model of interest.
 
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -926,7 +926,7 @@ def prefix_columns(evaluator, model_name, prefix):
     prefix = prefix.name
     renamed_projections = []
 
-    # Convert the SQLGlot expression to a lookup key.
+    # Convert the SQLGlot expression into a lookup key.
     #
     # Assumes that `model_name` does not contain quotes - if it did, we would have to
     # generate SQL for each part of `model_name` separately and then concatenate them,
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -913,7 +913,7 @@ Column expressions are sub-classes of the [Condition class](https://sqlglot.com/
 
 #### Accessing model schemas
 
-Model schemas can be accessed within a Python macro function through its evaluation context's `column_to_types` method, when they can be statically determined.
+Model schemas can be accessed within a Python macro function through its evaluation context's `column_to_types` method, when they can be statically determined. For instance, a schema of an external model can be accessed only after the `sqlmesh create_external_models` command has been executed.
 
 For example, consider the following macro function which aims to rename the columns of a target model by adding a prefix to them:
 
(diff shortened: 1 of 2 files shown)


z3z1ma:
Nice 2 questions. 

1) Where does this _not_ work. You mention "where columns can be statically determined". Is there any way we can succinctly enumerate a case or two. I have only used the feature on external models thus far. Does it work on other sqlmesh models?

2) Is this there preferred way to get the correct key for the `columns_to_types` dict? I could see that being a point of contention if it is not clear as you could end up trying to access model_name.this, model_name.key, model_name.alias or who knows. Its obvious enough once you use the lib, but emphasizing the access pattern would be good. 

`evaluator.columns_to_types(model_name.sql())`

cc @GeorgeSittas

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+    for name, dtype in evaluator.columns_to_types(model_name.sql()).items():
+        new_name = prefix + name
+        renamed_projections.append(exp.cast(exp.column(name), dtype).as_(new_name))
+
+    return renamed_projections
+```
+
+Having access to the schema of an upstream model can be useful for various reasons:
+
+- Normalizing columns so that downstream consumers are not tightly coupled e.g. to external or source tables
```
treysp, line 938: What does "normalizing" mean in this context?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+- Projecting only a subset of columns that satisfy some criteria
```
treysp, line 939: "Projecting" -> "Selecting" 

Add parenthetical example like `(e.g., columns whose names start with a specific prefix)`

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+    renamed_projections = []
+
+    for name, dtype in evaluator.columns_to_types(model_name.sql()).items():
+        new_name = prefix + name
+        renamed_projections.append(exp.cast(exp.column(name), dtype).as_(new_name))
+
+    return renamed_projections
+```
+
+Having access to the schema of an upstream model can be useful for various reasons:
```
treysp, line 936: Is it fair to say that the most common use case will be when the columns in upstream tables are either unknown or dynamic?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+
+- Normalizing columns so that downstream consumers are not tightly coupled e.g. to external or source tables
+- Projecting only a subset of columns that satisfy some criteria
+- Adding necessary casts
+- Computing common columns
```
treysp, line 941: What does this mean?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+```python linenums="1"
+from sqlglot import exp
+from sqlmesh.core.macros import macro
+
+@macro()
+def foo(evaluator, model_name, prefix):
+    prefix = prefix.name
+    renamed_projections = []
+
+    for name, dtype in evaluator.columns_to_types(model_name.sql()).items():
```
treysp, line 929: Pull out `model_name.sql()` into its own line and add comment describing what it does

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+
+For example, consider the following macro function which aims to rename the columns of a target model by adding a prefix to them:
+
+```python linenums="1"
+from sqlglot import exp
+from sqlmesh.core.macros import macro
+
+@macro()
+def foo(evaluator, model_name, prefix):
+    prefix = prefix.name
```
treysp, line 926: Does this assume `prefix` is a sqlglot expression?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+    renamed_projections = []
+
+    for name, dtype in evaluator.columns_to_types(model_name.sql()).items():
+        new_name = prefix + name
+        renamed_projections.append(exp.cast(exp.column(name), dtype).as_(new_name))
+
+    return renamed_projections
+```
```
treysp, line 934: Add brief example of using this macro in a model

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -911,6 +911,37 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+from sqlmesh.core.macros import macro
+
+@macro()
+def foo(evaluator, model_name, prefix):
+    prefix = prefix.name
+    renamed_projections = []
+
+    for name, dtype in evaluator.columns_to_types(model_name.sql()).items():
+        new_name = prefix + name
+        renamed_projections.append(exp.cast(exp.column(name), dtype).as_(new_name))
```
treysp, line 931: Do we need the cast here?

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Anything really. One way to normalize them is to add prefixes as shown in the exampe, another is to transform `__` to `_` as done in an example Alex had shared. The point is to transform them in some way that establishes a naming protocol for downstream consumers.

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: I don't think so: unknown columns can't be returned by `columns_to_types` because that'd mean we can infer them statically and they wouldn't really be "unknown". Unless I misread your comment?

Correct me if I'm wrong, but my understanding is that the most common use case is creating a dependency inversion layer between e.g. external models and their downstream consumers by applying certain transformations (renames, casts, etc - basically what I mentioned in the bullets)

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: You can combine columns from upstream models, e.g. by adding them, thus computing new columns. We should probably improve this section a bit and add more examples, I understand that it's a bit hand-wavy at the moment.

@z3z1ma any practical use-case comes to mind from your experience?

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Good point, thanks!

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Yeah, macro arguments are always sqlglot expressions. Same goes for the returned value - it's parsed into a sqlglot expression (if it's not one already).

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: It's not "needed" per se, but it's used to demonstrate that you can apply transformations to columns such as enforcing certain types. Perhaps one could do this to make sure there are explicit casts in the model queries.

GeorgeSittas:
Thank you for reviewing, @z3z1ma @treysp! Agree with most of the points you've made and will update this PR soon.

docs/concepts/macros/sqlmesh_macros.md
z3z1ma: if you do rule based naming on an upstream column, for example `n_<column>` always corresponds to a numeric -- you can autocompute statistics like a pandas describe. And subsequent `plans` will simply include them by virtue of updating the external model for example. very dry.

Some columns can be determined to be PII based on name and a blanket hash may be sufficient for any column containing `email` or `address`, etc.

z3z1ma: Actually even better, you can use the `type` 😆 
So generating summary tables that are always in sync with upstream source may be cake.

docs/concepts/macros/sqlmesh_macros.md
treysp: Maybe use "Renaming" to avoid ambiguity with normalizing/denormalizing tables?

GeorgeSittas: > Maybe use "Renaming" to avoid ambiguity with normalizing/denormalizing tables?

Sounds good

GeorgeSittas:
> Nice 2 questions.
> 
> 1. Where does this _not_ work. You mention "where columns can be statically determined". Is there any way we can succinctly enumerate a case or two. I have only used the feature on external models thus far. Does it work on other sqlmesh models?
> 
> 2. Is this there preferred way to get the correct key for the `columns_to_types` dict? I could see that being a point of contention if it is not clear as you could end up trying to access model_name.this, model_name.key, model_name.alias or who knows. Its obvious enough once you use the lib, but emphasizing the access pattern would be good.
> 
> 
> `evaluator.columns_to_types(model_name.sql())`
> 
> cc @GeorgeSittas

@z3z1ma regarding (1), IIRC it doesn't work when SQLMesh doesn't know or can't infer the schema of a model. For example, if you don't run `create_external_models`, then it's supposed to fail if `columns_to_types` is called with an external model's name for which we don't know the schema.

(2) is probably trickier because it depends on the use-case. If for example a macro gets passed a `"foo.bar"` then you don't want to generate that arg using `.sql()` because it'll include the quotes which need to be omitted for the lookup.

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Addressed

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Perfect, incorporated the statistics example in the latest commit

docs/concepts/macros/sqlmesh_macros.md
GeorgeSittas: Addressed

treysp (approved):

PR #1646, opened 2023-11-01
Chore! (spark, trino): raise if spark/trino used as state connection

treysp:


examples/sushi/config.py
```diff
@@ -61,7 +62,9 @@
 # Due to a 3.7 mypy bug we ignore. Can remove once 3.7 support is dropped.
 airflow_config_docker = Config(  # type: ignore
     default_scheduler=AirflowSchedulerConfig(airflow_url="http://airflow-webserver:8080/"),
-    gateways=GatewayConfig(connection=SparkConnectionConfig()),
+    gateways=GatewayConfig(
+        connection=SparkConnectionConfig(), state_connection=DuckDBConnectionConfig()
+    ),
```
eakmanrq, line 67: @izeigerman This would be a problem right?

izeigerman: It likely will be. Why do we do this?

treysp: It can no longer use Spark for the state connection - should it be a stateful Duckdb connection, or another DB entirely?

eakmanrq: In this part you are changing the Airflow connection information and Airflow is designed to run across multiple workers (it is distributed). Therefore conceptually it is not possible to share an in-memory db connection.

I think this isn't needed though since we we will actually use Airflow's state by default and therefore we aren't currently using Spark for these tests. Therefore this comment is misleading and it should be that we use DWG unless Airflow is configured then we use Airflow's state by default. 
https://github.com/TobikoData/sqlmesh/blob/15972e1538909dc186271590d6497fe3d5bfe39a/sqlmesh/core/config/gateway.py#L17-L18
@izeigerman please confirm.

treysp: Also also, do the airflow integration tests pass with the in-memory DuckDB config due to a coincidence of how our tests are set up, or is SQLMesh falling back to airflow state because the DuckDB state isn't working?

sqlmesh/core/config/scheduler.py
```diff
@@ -60,6 +63,11 @@ def create_state_sync(self, context: Context) -> StateSync:
         engine_adapter = (
             state_connection.create_engine_adapter() if state_connection else context.engine_adapter
         )
+        if isinstance(engine_adapter, (SparkEngineAdapter, TrinoEngineAdapter)):
```
eakmanrq, line 66: What do you think about doing the FLAG on engine adapter and checking for that? So add `SUPPORTS_ROW_LEVEL_OP` and `True` by default but `False` for these and then check for that?

eakmanrq (approved):

PR #1648, opened 2023-11-01
Chore: Make create_external_schema warn instead of raise when source table doesn't exist

crericha:


GeorgeSittas (approved):

sqlmesh/core/schema_loader.py
```diff
@@ -53,17 +53,26 @@ def create_schema_file(
         external_tables -= existing_models
 
     with ThreadPoolExecutor(max_workers=max_workers) as pool:
+
+        def _get_columns(table: str) -> t.Optional[t.Dict[str, t.Any]]:
+            try:
+                return adapter.columns(table, include_pseudo_columns=True)
+            except:
+                logger.warning(f"Unable to get schema for '{table}'")
```
tobymao, line 61: you should log the exception and not just swallow it

PR #1649, opened 2023-11-01
fix: include dialect when making dbt seed models

eakmanrq:


sqlmesh/core/model/definition.py
```diff
@@ -1632,6 +1632,7 @@ def create_seed_model(
     name: str,
     seed_kind: SeedKind,
     *,
+    dialect: t.Optional[str] = None,
```
crericha, line 1635: Should the call to `create_seed_model` in core also pass in dialect?

eakmanrq: It does it is part of the common_kwargs: https://github.com/TobikoData/sqlmesh/blob/b8820bc1ae8368a7bc18643554c2bd0c85fdcb4c/sqlmesh/core/model/definition.py#L1547

crericha (approved):

GeorgeSittas (approved):

PR #1650, opened 2023-11-01
Fix: Include "this" jinja object in dbt singular tests

crericha:


tobymao (approved):

sqlmesh/dbt/manifest.py
```diff
@@ -172,13 +172,15 @@ def _load_tests(self) -> None:
             )
 
             test_model = _test_model(node)
+            node_config = _node_base_config(node)
+            node_config["database"] = node.database or self.target.database
```
izeigerman, line 176: why do we do this?

crericha: For singular tests, dbt returns the test database in the "this" obj. This is to match it.

izeigerman: So the problem with this is that we no longer know whether the `database` was actually overridden for any given model.

crericha: This is only for a `test` node.

izeigerman (approved):

PR #1652, opened 2023-11-01
Chore: set schema_version default value directly in the pydantic model

GeorgeSittas:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -485,9 +485,7 @@ def _get_versions(self, lock_for_update: bool = False) -> Versions:
             return no_version
 
         return Versions(
-            schema_version=row[0],
-            sqlglot_version=row[1],
-            sqlmesh_version=seq_get(row, 2) or "0.0.0",
+            schema_version=row[0], sqlglot_version=row[1], sqlmesh_version=seq_get(row, 2)
```
tobymao, line 488: same with line 474?

GeorgeSittas: should we then apply defaults to sqlglot_version and schema_version too?

tobymao: sure

GeorgeSittas: updated

tobymao (approved):

PR #1654, opened 2023-11-02
Fix: add validators to handle None values in Versions class

GeorgeSittas:
...sigh

### DIFF
--- a/tests/core/test_state_sync.py
+++ b/tests/core/test_state_sync.py
@@ -1062,6 +1062,14 @@ def test_get_version(state_sync: EngineAdapterStateSync) -> None:
         state_sync.get_versions()
     state_sync.get_versions(validate=False)
 
+    for empty_versions in (
+        Versions(),
+        Versions(schema_version=None, sqlglot_version=None, sqlmesh_version=None),
+    ):
+        assert empty_versions.schema_version == 0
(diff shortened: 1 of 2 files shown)


tobymao (approved):
can you make a test that actually migrated from n-1 to n?

GeorgeSittas:
I can try, maybe I need to mock out the schema version so that it doesn't pick v0032 or something

GeorgeSittas:
Added a unit test for this.

tests/core/test_state_sync.py
```diff
@@ -1545,3 +1554,89 @@ def test_snapshot_batching(state_sync, mocker, make_snapshot):
     assert len(snapshots) == 3
     calls = mock.fetchall.call_args_list
     assert len(calls) == 2
+
+
+def test_migrate_only_to_add_sqlmesh_version_column(duck_conn, monkeypatch) -> None:
+    from sqlmesh import __version__ as SQLMESH_VERSION
+
+    n_migrations_before_sqlmesh_version = 31
```
tobymao, line 1562: make this n -1 so we always test the latest migration right

GeorgeSittas: It's not that simple because this test has logic specific to the 31->32 migration, i.e. patching _update_versions and assuming that prior to that migration we don't have `sqlmesh_version`

PR #1655, opened 2023-11-02
Chore: Refactor to remove legacy QuotingConfig class

crericha:
Also remove residual QuotingConfig from before the switch to using the manifest.

### DIFF
--- a/sqlmesh/dbt/basemodel.py
+++ b/sqlmesh/dbt/basemodel.py
@@ -21,7 +21,6 @@ from sqlmesh.dbt.common import (
     DbtConfig,
     Dependencies,
     GeneralConfig,
-    QuotingConfig,
     SqlStr,
     sql_str_validator,
 )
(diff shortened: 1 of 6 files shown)


izeigerman:
Need a little more context here as well. Looks a lot like the `database` change. Is this consistent with how dbt behaves?

crericha:
> Need a little more context here as well. Looks a lot like the `database` change. Is this consistent with how dbt behaves?

There are two things going on here:
1. dbt returns a dict for `quoting`. We were storing that dict in our `QuotingConfig` class, which was created before we used the manifest. Since we don't need to validate the input (we trust the manifest), it's overkill and I removed it to simplify the code.
2. When `quoting` is not overriden in the node, the `quoting` field is set to an empty dict in the manifest. By default, if we received an empty dict, we'd set quoting to true for each component of the fqn, which is correct for the majority of dialects; however, is incorrect for Snowflake, where by default nothing should be quoted (this is what dbt does too). This change fixes it by using the target dwh's quoting policy when one is not provided in the manifest.

PR #1656, opened 2023-11-02
fix: spark temp tables

eakmanrq:
It appears when writing local files with spark it does not like a table that starts with double underscore. So I updated temp table for spark to remove it.

Also re-organized things a bit to remove the pattern where we got a query from source query and then made a source query again.

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -12,7 +12,7 @@ from sqlmesh.utils.errors import SQLMeshError
 
 if t.TYPE_CHECKING:
     from sqlmesh.core._typing import TableName
-    from sqlmesh.core.engine_adapter._typing import DF
+    from sqlmesh.core.engine_adapter._typing import DF, Query
     from sqlmesh.core.engine_adapter.base import QueryOrDF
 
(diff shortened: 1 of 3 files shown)


izeigerman (approved):

treysp (approved):

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -69,6 +69,38 @@ def merge(
...
+        with engine_adapter.temp_table(
+            exp.select(*columns_to_types).from_(target_table),
+            target_table,
+            columns_to_types,
+        ) as temp_table:
+
+            def replace_table(
+                node: exp.Expression, curr_table: exp.Table, new_table: exp.Table
+            ) -> exp.Expression:
+                if isinstance(node, exp.Table) and quote_identifiers(node) == quote_identifiers(
```
eakmanrq, line 92: @treysp FYI I think this was another bug outside of the spark issue. `node == curr_table` could say that two tables don't match if one doesn't have quotes and the other does. So we quote them both now when comparing.

treysp: Good catch - thanks

PR #1659, opened 2023-11-03
Feat!: add more capabilities to macros

tobymao:


sqlmesh/core/macros.py
```diff
@@ -285,6 +288,27 @@ def columns_to_types(self, model_name: str) -> t.Dict[str, exp.DataType]:
...
+
+    @property
+    def engine_adapter(self) -> EngineAdapter:
+        engine_adapter = self.locals.get("engine_adapter")
+        if not engine_adapter:
+            raise SQLMeshError(
+                "The engine adapter is not available while models are loading."
+                " Rendering queries happens frequently and should be fast and determinstic so it is not recommended"
+                " to make calls to the warehouse at load time."
+                " You can gate these calls by checking in Python: evaluator.runtime_stage != 'loading' or SQL: @runtime_stage <> 'loading'."
```
GeorgeSittas, line 308: It's not like this is a recommendation, since we're prohibiting the use of this property at load time, so perhaps the reasoning behind why this can't happen should go in the docs instead? I think I'd personally keep only the first and last sentences here.

sqlmesh/core/macros.py
```diff
@@ -285,6 +288,27 @@ def columns_to_types(self, model_name: str) -> t.Dict[str, exp.DataType]:
...
+            )
+        return self.locals["engine_adapter"]
```
GeorgeSittas, line 310:
```suggestion
        return engine_adapter
```

sqlmesh/core/renderer.py
```diff
@@ -90,12 +90,22 @@ def _render(
             The rendered expressions.
         """
 
-        cache_key = self._cache_key(start, end, execution_time, runtime_stage)
-        start_dt, end_dt, execution_dt, runtime_stage = cache_key
-        if cache_key not in self._cache:
+        cache_key = self._cache_key(start, end, execution_time)
+        start_dt, end_dt, execution_dt = cache_key
+        if runtime_stage != RuntimeStage.LOADING or cache_key not in self._cache:
```
GeorgeSittas, line 95: Was this a bug?

GeorgeSittas (commented):
Should we add some tests for these additions?

sqlmesh/core/renderer.py
izeigerman: Do we ignore cache when not loading?

sqlmesh/core/renderer.py
```diff
@@ -90,12 +90,22 @@ def _render(
...
             expressions = [self._expression]
 
             render_kwargs = {
+                "resolve_tables": lambda e, **kwargs: self._resolve_tables(
```
izeigerman, line 99: what's this key? Should there be a better name?

sqlmesh/core/renderer.py
tobymao: it's kind of unnecessary, the cache is really to prevent recomputation for loads, and so i removed the complexity of runtime stage in cache, 

it just means that when evaluating / creating tables, we won't hit the query cache, but that's fine, and may be intentional if your sql has side effects

sqlmesh/core/renderer.py
tobymao: it's an expression, i can update it

sqlmesh/core/macros.py
```diff
@@ -285,6 +288,27 @@ def columns_to_types(self, model_name: str) -> t.Dict[str, exp.DataType]:
...
+        return self.locals["resolve_tables"](query, **kwargs)
+
+    @property
+    def runtime_stage(self) -> RuntimeStage:
+        """Returns the current runtime stage of the macro evaluation."""
+        return self.locals["runtime_stage"]
+
+    @property
+    def engine_adapter(self) -> EngineAdapter:
+        engine_adapter = self.locals.get("engine_adapter")
```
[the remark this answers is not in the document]
tobymao: it is passed in

sqlmesh/core/renderer.py
tobymao: going to revert this

GeorgeSittas: Did it end up causing issues?

tobymao: it's annoying because i need to also not set the cache when the runtime stage is not loading

PR #1661, opened 2023-11-03
Fix!: normalize identifiers before calling engine adapter methods in dbt

GeorgeSittas:


sqlmesh/dbt/adapter.py
```diff
@@ -255,15 +259,19 @@ def get_missing_columns(
...
         if relation.schema is not None:
-            self.engine_adapter.create_schema(relation.schema)
+            schema = self._normalize(relation.schema, self.quote_policy.schema)
+            self.engine_adapter.create_schema(schema)
 
     def drop_schema(self, relation: BaseRelation) -> None:
         if relation.schema is not None:
-            self.engine_adapter.drop_schema(relation.schema)
+            schema = self._normalize(relation.schema, self.quote_policy.schema)
+            self.engine_adapter.drop_schema(schema)
```
eakmanrq, line 268: Although the default catalog stuff is not in this PR, you could normalize the catalog here still right? And then do `schema_` on the schema name and catalog name.

sqlmesh/dbt/adapter.py
```diff
@@ -255,15 +259,19 @@ def get_missing_columns(
...
 
     def drop_relation(self, relation: BaseRelation) -> None:
         if relation.schema is not None and relation.identifier is not None:
-            self.engine_adapter.drop_table(f"{relation.schema}.{relation.identifier}")
+            schema = self._normalize(relation.schema, self.quote_policy.schema)
+            identifier = self._normalize(relation.identifier, self.quote_policy.identifier)
```
eakmanrq, line 273: same with catalog here.

sqlmesh/dbt/adapter.py
```diff
@@ -309,13 +317,38 @@ def resolve_identifier(self, relation: BaseRelation) -> t.Optional[str]:
...
 
     def _map_table_name(
-        self, database: t.Optional[str], schema: t.Optional[str], identifier: t.Optional[str]
+        self,
+        database: t.Optional[str | exp.Identifier],
+        schema: t.Optional[str | exp.Identifier],
+        identifier: t.Optional[str | exp.Identifier],
     ) -> exp.Table:
-        name = ".".join(p for p in (database, schema, identifier) if p is not None)
+        name = ".".join(
```
eakmanrq, line 325: you could do `nullsafe_join(".", database, schema, identifier)` here.

tests/dbt/test_adapter.py
```diff
@@ -58,6 +63,76 @@ def test_adapter_relation(sushi_test_project: Project, runtime_renderer: t.Calla
...
+        "{%- set relation = api.Relation.create(schema='bla') -%}"
+        "{{ adapter.create_schema(relation) }}"
+    )
+    adapter_mock.create_schema.assert_has_calls([call(bla_id)])
+
+    renderer(
+        "{%- set relation = api.Relation.create(schema='bla') -%}"
+        "{{ adapter.drop_schema(relation) }}"
+    )
+    adapter_mock.drop_schema.assert_has_calls([call(bla_id)])
```
crericha, line 97: Why twice with the same params?

sqlmesh/dbt/adapter.py
```diff
@@ -255,15 +259,19 @@ def get_missing_columns(
...
-            self.engine_adapter.drop_schema(relation.schema)
+            schema = self._normalize(relation.schema, self.quote_policy.schema)
+            self.engine_adapter.drop_schema(schema)
 
     def drop_relation(self, relation: BaseRelation) -> None:
         if relation.schema is not None and relation.identifier is not None:
-            self.engine_adapter.drop_table(f"{relation.schema}.{relation.identifier}")
+            schema = self._normalize(relation.schema, self.quote_policy.schema)
+            identifier = self._normalize(relation.identifier, self.quote_policy.identifier)
+            self.engine_adapter.drop_table(exp.Table(this=identifier, db=schema))
```
tobymao, line 274: exp.table_(identefier, db=schema)

sqlmesh/dbt/adapter.py
tobymao: don't use null safe join

do 

table_name = exp.table_(identefier, db=schema, catalog=database)

sqlmesh/dbt/adapter.py
```diff
@@ -309,13 +317,38 @@ def resolve_identifier(self, relation: BaseRelation) -> t.Optional[str]:
...
+
+    @t.overload
+    def _normalize(self, name: None, quoted: bool = True) -> None:
+        ...
+
+    @t.overload
+    def _normalize(self, name: str, quoted: bool = True) -> exp.Identifier:
+        ...
+
+    def _normalize(self, name: t.Optional[str], quoted: bool = True) -> t.Optional[exp.Identifier]:
```
tobymao, line 346: why doesn't normalize take in like a table or something? that way you don't have to call it multiple times, or it can take multiple args

tests/dbt/test_adapter.py
GeorgeSittas: One should be for `create_schema`, the other for `drop_schema`

sqlmesh/dbt/adapter.py
tobymao: nullsafe_join is going to be deprecated soon, it's only used in a test

tests/dbt/test_adapter.py
```diff
@@ -58,6 +63,76 @@ def test_adapter_relation(sushi_test_project: Project, runtime_renderer: t.Calla
...
+    select_star_query: exp.Select = exp.maybe_parse("SELECT * FROM t", dialect="snowflake")
+
+    # The following call to run_query won't return dataframes and so we're expected to
+    # raise in adapter.execute right before returning from the method
+    with pytest.raises(AssertionError):
+        renderer("{{ run_query('SELECT * FROM t') }}")
+    adapter_mock.fetchdf.assert_has_calls([call(select_star_query, quote_identifiers=False)])
+
+    renderer("{% call statement('something') %} {{ 'SELECT * FROM t' }} {% endcall %}")
+    adapter_mock.execute.assert_has_calls([call(select_star_query, quote_identifiers=False)])
```
crericha, line 114: Let's move run_query and statement to their own tests each

sqlmesh/dbt/adapter.py
GeorgeSittas: I guess same question applies here

sqlmesh/dbt/adapter.py
GeorgeSittas: Ah right thanks, I forgot about `table_`

sqlmesh/dbt/adapter.py
GeorgeSittas: thanks

tests/dbt/test_adapter.py
GeorgeSittas: Will do

sqlmesh/dbt/adapter.py
eakmanrq: No you can do this without the catalog PR. Each relation can have a database defined (or catalog in our terms) and therefore you should include that in the call here if one is defined.

tests/dbt/test_adapter.py
GeorgeSittas: what about just moving all these normalization tests in a new one instead?

sqlmesh/dbt/adapter.py
GeorgeSittas: I think there might actually be an issue here because if we create a table and then generate SQL from it there can be quotes included. Is there a utility to just generate the name part without quotes?

sqlmesh/dbt/adapter.py
GeorgeSittas: done

tests/dbt/test_adapter.py
GeorgeSittas: moved them to `test_normalization` so the intent is clear

sqlmesh/dbt/adapter.py
GeorgeSittas: @eakmanrq changed

GeorgeSittas:
I think I've incorporated the PR feedback - please take another look and let me know.

tobymao (approved):

crericha (approved):

PR #1662, opened 2023-11-03
chore: remove fallback values

eakmanrq:


tobymao:
what's the motivation for this change?

eakmanrq:
Just cleanup. Ran into it in default catalog PR and wanted to split it out since it is a standalone change.

izeigerman:
I added the fallbacks. Not sure why do we need to fail instead

eakmanrq:
@izeigerman This should never happen right? I figure if we did end up in a situation where we would use these fallbacks that we would prefer to error instead.

izeigerman:
@eakmanrq sorry for the delayed reply. I noticed that in some cases in the past (maybe not right now) we called these property methods (`_dag`, `_snapshot_mapping`) recursively while computing the `__dag` and `__snapshot_mapping` values. I see why shouldn't we fallback to defaults in this case

PR #1663, opened 2023-11-03
Feat: Optionally disable formatting of model sql in render

z3z1ma:
**Problem:**

This has bugged me for awhile, but current way `sqlmesh render` works has a lot of issues if your SQL is wider than console length. It automatically wraps and insert new lines which breaks usage in scripts and even copy paste. Furthermore it pads the end of line all the way out to the terminal width. Which is also really annoying. Lastly, and I don't know if this is consistent -- I get truncated text sometimes too. 

**Solution:**

Let people just dump the output as is to stdout, so we can use it in our vimscripts and so on. Or even so its cleaner to yank. 

Add `--no-pretty-print` to render command which just uses simple print stmt

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -143,9 +143,7 @@ def init(
     type=str,
     help="The SQL dialect to render the query as.",
 )
-@click.option(
-    "--no-pretty-print", is_flag=True, help="Disable pretty printing of the query."
-)
+@click.option("--no-pretty-print", is_flag=True, help="Disable pretty printing of the query.")
 @click.pass_context
 @error_handler
 def render(
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -143,7 +143,7 @@ def init(
     type=str,
     help="The SQL dialect to render the query as.",
 )
-@click.option("--no-pretty-print", is_flag=True, help="Disable pretty printing of the query.")
+@click.option("--no-format", is_flag=True, help="Disable fancy formatting of the query.")
 @click.pass_context
 @error_handler
 def render(
@@ -154,7 +154,7 @@ def render(
     execution_time: t.Optional[TimeLike] = None,
     expand: t.Optional[t.Union[bool, t.Iterable[str]]] = None,
     dialect: t.Optional[str] = None,
-    no_pretty_print: bool = False,
+    no_format: bool = False,
 ) -> None:
     """Renders a model's query, optionally expanding referenced models."""
     rendered = ctx.obj.render(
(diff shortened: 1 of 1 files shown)


z3z1ma:
image
 
 
Padding seen above, worst though is the line break inserting and truncation -- but you get the point

sqlmesh/cli/main.py
```diff
@@ -153,6 +154,7 @@ def render(
     execution_time: t.Optional[TimeLike] = None,
     expand: t.Optional[t.Union[bool, t.Iterable[str]]] = None,
     dialect: t.Optional[str] = None,
+    no_pretty_print: bool = False,
```
tobymao, line 157: i was confused because i thought you meant sqlglot pretty, but you're keeping this, it's really in the console, show sql.

maybe it should be

--no-format

z3z1ma: right, true

sqlmesh/cli/main.py
```diff
@@ -164,7 +166,10 @@ def render(
     )
 
     sql = rendered.sql(pretty=True, dialect=ctx.obj.config.dialect if dialect is None else dialect)
-    ctx.obj.console.show_sql(sql)
+    if no_format:
+        print(sql)
+    else:
+        ctx.obj.console.show_sql(sql)
```
GeorgeSittas, line 172: I think this needs to instead be `pretty=not no_format` or something?

tobymao: what do you mean?

tobymao (approved):

sqlmesh/cli/main.py
GeorgeSittas: i may have misunderstood the problem - is it that we're pretty-printing, or that we're using the fancy rich printer instead of python's `print`?

tobymao: fancy rich printer, that confused me too

GeorgeSittas: oh i just re-read the description, it seems it's the latter  🤦 -- ignore me

z3z1ma: The intent is actually just printing rendered query to stdout **without** going through `rich.print` + rich console + `rich.syntax` which the combination were adding unnecessary noise to stdout, which made the command difficult to use. I think `--pretty/no-pretty` could be a separate option -- fyi @GeorgeSittas

GeorgeSittas: Gotcha thanks

GeorgeSittas: Re: I think --pretty/no-pretty could be a separate option

Sure, that could be helpful I guess if you wanna avoid SQLGlot's formatter for any reason? cc @tobymao

tobymao: can be done later

PR #1664, opened 2023-11-03
Feat: add more core macros

z3z1ma:


sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+@macro()
+def star(
+    evaluator: MacroEvaluator,
+    relation: exp.Table,
+    alias: t.Optional[exp.Identifier] = None,
+    except_: t.Optional[exp.Array] = None,
+    prefix: t.Optional[exp.Literal] = None,
+    suffix: t.Optional[exp.Literal] = None,
+    quote_identifiers: exp.Boolean = exp.Boolean(this=True),
+) -> t.List[exp.Expression]:
```
tobymao, line 660: let's add a doc string

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
     return evaluator.eval_expression(condition)
 
 
+@macro()
+def star(
+    evaluator: MacroEvaluator,
+    relation: exp.Table,
+    alias: t.Optional[exp.Identifier] = None,
+    except_: t.Optional[exp.Array] = None,
```
tobymao, line 656: this should be | Tuple right

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    prefix: t.Optional[exp.Literal] = None,
+    suffix: t.Optional[exp.Literal] = None,
+    quote_identifiers: exp.Boolean = exp.Boolean(this=True),
+) -> t.List[exp.Expression]:
+    if alias and not isinstance(alias, exp.Identifier):
```
tobymao, line 661: you don't have control over this being an Identifier, it's going to be a column type

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
 
+@macro()
+def star(
+    evaluator: MacroEvaluator,
+    relation: exp.Table,
+    alias: t.Optional[exp.Identifier] = None,
+    except_: t.Optional[exp.Array] = None,
+    prefix: t.Optional[exp.Literal] = None,
+    suffix: t.Optional[exp.Literal] = None,
+    quote_identifiers: exp.Boolean = exp.Boolean(this=True),
```
tobymao, line 659: exp.true()

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    if not isinstance(quote_identifiers, exp.Boolean):
+        raise SQLMeshError(
+            f"Invalid quote_identifiers '{quote_identifiers}'. Expected a boolean."
+        )
+    projections = []
+    exclude = []
+    quoted = quote_identifiers.this
+    if except_ and isinstance(except_.expressions, list):
+        exclude.extend(
+            e.name for e in except_.expressions if isinstance(e, exp.Identifier)
```
tobymao, line 678: are you sure you're getting identifiers in here and not columns?

tobymao: this should be a set

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,48 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
     return evaluator.eval_expression(condition)
 
 
+@macro()
+def star(
+    evaluator: MacroEvaluator,
+    relation: exp.Table,
+    alias: t.Optional[exp.Identifier] = None,
+    except_: t.Optional[exp.Array] = None,
+    prefix: t.Optional[exp.Literal] = None,
```
tobymao, line 657: you could make this Literal.string("") then down where you use it, you can just do prefix.name

sqlmesh/core/macros.py
z3z1ma: I see, good to know! Its a little ambiguous to me why this is the case. But if it is logical and concrete, I will lean into this knowledge.

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,87 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    if not isinstance(quote_identifiers, exp.Boolean):
+        raise SQLMeshError(
+            f"Invalid quote_identifiers '{quote_identifiers}'. Expected a boolean."
+        )
+    projections = []
+    exclude = []
+    kwargs = {"quoted": quote_identifiers.this}
+    if alias:
+        kwargs["table"] = alias.name
+    if except_ and isinstance(except_.expressions, list):
```
tobymao, line 687: there's no need for this check here, expressions is always a list

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,87 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+                f"{p}{column}{s}", quoted=kwargs["quoted"]
+            )
+        )
+    return projections
+
+
+@macro()
+def generate_surrogate_key(
+    _: MacroEvaluator, *fields: exp.Column | exp.Identifier
+) -> exp.Func:
```
tobymao, line 709: @eakmanrq do you have opinions on this

sqlmesh/core/macros.py
z3z1ma: image
 type checker suggests any in unio type but will drop check

tobymao: missing type, i'll fix it

tobymao:
let's fix the build, add some tests, and i'll merge this in, thanks @z3z1ma

z3z1ma:
bigger plans, give me just a sec

tobymao:
> bigger plans, give me just a sec

lol, what kind of plans? do you want to save that for another PR?

z3z1ma:
no just be nice to get more than 2 macros --there are more low hanging fruit and im on a roll

tobymao:
> no just be nice to get more than 2 macros --there are more low hanging fruit and im on a roll

ok go for it, please add some tests, and i'll do a final review of all the macros

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    null_cond = exp.and_(*[exp.func("IS_NULL", field) for field in fields])
+    case = exp.Case().when(null_cond, exp.null())
+    terms = []
+    for field in fields:
+        terms.append(exp.func("COALESCE", field, exp.Literal.number(0)))
+    return case.else_(reduce(lambda a, b: a + b, terms))  # type: ignore
+
+
+@macro()
+def safe_subtract(_: MacroEvaluator, *fields: exp.Column) -> exp.Func:
```
tobymao, line 746: maybe safe_sub and safe_div ?

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    """Divides numbers, returning NULL if all terms are NULL.
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "SELECT @SAFE_DIVIDE(a, b) FROM foo"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        "SELECT a / NULLIF(b, 0) FROM foo"
+    """
+    return numerator / exp.func("NULLIF", denominator, exp.Literal.number(0))
```
tobymao, line 775: you can just do 0 instead of the Literal.number here

tobymao: is nullif ansi standard? can you switch it to a case statement?

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    case = exp.Case().when(null_cond, exp.null())
+    terms = []
+    for field in fields:
+        terms.append(exp.func("COALESCE", field, exp.Literal.number(0)))
+    return case.else_(reduce(lambda a, b: a - b, terms))  # type: ignore
+
+
+@macro()
+def safe_divide(_: MacroEvaluator, numerator: exp.Column, denominator: exp.Column) -> exp.Div:
+    """Divides numbers, returning NULL if all terms are NULL.
```
tobymao, line 766: returns null if the denominator is 0

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    """Subtract numbers, returning NULL if all terms are NULL.
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "SELECT @SAFE_SUBTRACT(a, b) FROM foo"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        "SELECT CASE WHEN a IS NULL AND b IS NULL THEN NULL ELSE COALESCE(a, 0) - COALESCE(b, 0) END FROM foo"
+    """
+    null_cond = exp.and_(*[exp.func("IS_NULL", field) for field in fields])
```
tobymao, line 756: is_null is not a standard function do

field.is_(exp.null()) instead

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    case = exp.Case().when(null_cond, exp.null())
+    terms = []
+    for field in fields:
+        terms.append(exp.func("COALESCE", field, exp.Literal.number(0)))
```
tobymao, line 760: 0

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    return case.else_(reduce(lambda a, b: a + b, terms))  # type: ignore
+
+
+@macro()
+def safe_subtract(_: MacroEvaluator, *fields: exp.Column) -> exp.Func:
+    """Subtract numbers, returning NULL if all terms are NULL.
```
tobymao, line 747: substitutes nulls for 0s and only returns null if all fields are null

sqlmesh/core/macros.py
tobymao: fields type is wrong, because the input can be any expression

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+                "COALESCE",
+                exp.cast(field, exp.DataType.build("string")),
+                default_null_value,
+            )
+        )
+    return exp.func("MD5", exp.func("CONCAT", *string_fields))
+
+
+@macro()
+def safe_add(_: MacroEvaluator, *fields: exp.Column) -> exp.Func:
```
tobymao, line 727: field type wrong, return value can be more strict

sqlmesh/core/macros.py
```diff
@@ -648,6 +648,133 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    """Adds numbers together, returning NULL if all terms are NULL.
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "SELECT @SAFE_ADD(a, b) FROM foo"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        "SELECT CASE WHEN a IS NULL AND b IS NULL THEN NULL ELSE COALESCE(a, 0) + COALESCE(b, 0) END FROM foo"
+    """
+    null_cond = exp.and_(*[exp.func("IS_NULL", field) for field in fields])
```
tobymao, line 737: use is_(exp.null())

sqlmesh/core/macros.py
```diff
@@ -648,6 +649,228 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "SELECT @SAFE_DIV(a, b) FROM foo"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        'SELECT a / CASE WHEN b = 0 THEN NULL ELSE b END FROM foo'
+    """
+    return numerator / exp.Case().when(denominator.eq(0), exp.null()).else_(denominator)
+
+
+@macro()
+def union(evaluator: MacroEvaluator, *tables: exp.Table) -> exp.Union:
```
tobymao, line 780: i'm not sure about this one...

sqlmesh/core/macros.py
```diff
@@ -648,6 +649,228 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+    projection = [
+        exp.cast(exp.column(name), typ).as_(name)
+        for name, typ in sorted(superset, key=lambda c: precedence[c[0]])
+    ]
+    selects: t.List[exp.Unionable] = [exp.select(*projection).from_(t) for t in tables]
+    return t.cast(exp.Union, reduce(lambda a, b: a.union(b), selects))
+
+
+@macro()
+def width_bucket(
```
tobymao, line 808: width bucket is already a sql function in many dialects, this shouldn't be here, 

if this isn't in bigquery, it should be in sqlglot, can you remove this one

sqlmesh/core/macros.py
z3z1ma: oh, this one was actually one i was pretty happy with functionally as a net-positive value add. its pretty powerful. can you explain the issue

sqlmesh/core/macros.py
z3z1ma: ya sure, would you mind giving a pointer as to how I would PR there? good to keep working between the codebases
maybe on slack l8er

sqlmesh/core/macros.py
z3z1ma: i thought the code, implementation, use cases, etc on this one was probably my favorite of what i have implemented. its resilient to columns appearing in different orders, differing columns entirely, differing column types, etc. all total PITAs in sql alone. and this is well like enough in dbt. the alternative is a lot of repeated code, sources of truth in 2 places. the plan apply flow guarantees deterministic behavior. only if a column is added of the same type the the tables being unioned and becomes part of the superset would it be added to the projection (via standard plan + apply)

it doesnt stop you from just using unions regularly when you need em.

z3z1ma: but curious on thoughts

z3z1ma: Think about the scenario you have N tables which are essentially the same but come from multiple sources upstream. You want to union them all in the data warehouse to create a base table to select from. By using this macro, you gain a number of benefits. 


- If an upstream model gets a new column but the others do not, it doesn't break where a select * would
- If columns are out of order, it doesnt break where a select * would
- You don't need to repeat information already defined in your external models, aka enumerate all cols to circumvent points 1 & 2 above
- The columns are still managed explicitly via code change _somewhere_, there is nothing dynamic happening here
- When all the upstream tables get a new column of the same type, it is brought downstream


So maybe just add docs on when or why you would use it.

sqlmesh/core/macros.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import typing as t
+from collections import OrderedDict
```
tobymao, line 4: this is not necessary, python 3.6 dicts are ordered

sqlmesh/core/macros.py
tobymao: you'd create a new expression type widthbucket and handle generation of it in bigquery

tobymao: let's delete this for now

sqlmesh/core/macros.py
```diff
@@ -648,6 +649,265 @@ def eval_(evaluator: MacroEvaluator, condition: exp.Condition) -> t.Any:
...
+                * exp.func("COS", exp.func("RADIANS", lat2))
+                * exp.func("POWER", exp.func("SIN", exp.func("RADIANS", (lon2 - lon1) / 2)), 2),
+            ),
+        )
+        * conversion_rate
+    )
+
+
+@macro()
+def pivot(
```
tobymao, line 875: is this an existing dbt function?

z3z1ma: yeah

z3z1ma:
Adding tests @tobymao , hive is rendering the `exp.Cast` -> `TRY_CAST` despite the fact `safe` is not being set to true. Not sure if bug. So it fails test on the AST assertion (though string equality assertion passes). Since that was sqlglot level behavior, I tried switch the dialect to postgres. I then noticed postgres has special handling in its `exp.Coalesce` I think that is doing an extra cast to text. I can update my test case to expect this but didnt want to make a test depend on that behavior if you thought it was prone to change. Any suggestions? Keeping `hive` doesn't seem like its workable if we use `cast` and `assert_exp_eq` (try yourself if you want). 

Once you have an idea on the above, I can add the last couple and commit the tests so we can merge -- they are pretty simple:

```python
        (
            """select @star(test1, x, [c], 't1_') FROM test1 AS x""",
            """SELECT CAST("x"."a" AS TEXT) AS "t1_a", CAST("x"."b" AS INT) AS "t1_b" FROM test1 AS x""",
            {},
        ),
        (
            """select @generate_surrogate_key(a, b), c FROM test1""",
            """SELECT MD5(CONCAT(COALESCE(CAST(a AS TEXT), '_sqlmesh_surrogate_key_null_'), '|', COALESCE(CAST(b AS TEXT), '_sqlmesh_surrogate_key_null_'))), c FROM test1""",
            {},
        )
```

z3z1ma:
We only use `exp.cast` in our macros, so we know `safe=False` which is the default on the node. But look at what happens here:
```
>>> e = MacroEvaluator("hive")
>>> e.transform(parse_one("select @star(test1, x, [c], 't1_') FROM test1 AS x")).sql()
'SELECT TRY_CAST("x"."a" AS TEXT) AS "t1_a", TRY_CAST("x"."b" AS INT) AS "t1_b" FROM test1 AS x'
```

So maybe something specific to how sql is generated for hive in specific instances

tobymao:
try_cast is what hive parses cast as. not super sure, but i plan to iterate on your PR anyways, so i wouldn't worry about it. just finish up what you have and i'll refactor it on my own

z3z1ma:
Tests fail cuz of the cast so probably not worth pushing any -- just have the above 2 anyway. So I think we are good! Appreciate your help here. 
The only other macro I had in mind was unpivot but its a bit hairy so may as well wait on it til I can see how you refactor.

sqlmesh/core/macros.py
eakmanrq: Seems fine. It is an opinionated way to generate surrogate keys but if we add arguments like the hash function to use then it somewhat defeats the purpose since now it may not be implemented consistently across an org. Also this has an opinionated constant for null but yet again it needs to be consist so it makes sense to now allow overriding. I think for larger orgs they will implement their own with their own opinions but this can be used as a reference. 

If you are asking because of past conversations about creating new columns by concating that was specific to doing that automatically within joins and therefore forcing shuffles. Doesn't apply here.

PR #1665, opened 2023-11-06
Feat: Support variable overrides per dbt config instance

izeigerman:


sqlmesh/core/config/root.py
```diff
@@ -50,6 +50,7 @@ class Config(BaseConfig):
         username: Name of a single user who should receive approvals/notification, instead of all users in the `users` list.
         pinned_environments: A list of development environment names that should not be deleted by the janitor task.
         loader: Loader class used for loading project files.
+        loader_kwargs: Key-value arguments to pass to the loader instance.
```
tobymao, line 53: is this dbt only?

tobymao: should we respect it in sqlmesh?

izeigerman: yeah, why not? It works for both

izeigerman: we just don't have anything for sqlmesh right now

crericha (approved):

PR #1668, opened 2023-11-07
chore: Add Logging to Github CI/CD Bot

eakmanrq:
It can be tricky to understand what is going on with the bot so now if a user sets `SQLMESH_DEBUG` to `1` then they will get the normal debug logging and then also bot specific log messages when using the CI/CD bot.

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -337,7 +337,7 @@ class GithubController:
             if user.is_required_approver and user.github_username
         ]
         logger.debug(
-            f"Required approvers: {', '.join([user.github_username for user in required_approvers if user.github_username])}"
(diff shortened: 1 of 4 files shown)


sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -349,7 +361,9 @@ def pr_environment_name(self) -> str:
     @property
     def do_required_approval_check(self) -> bool:
         """We want to skip required approval check if no users have this role"""
-        return bool(self._required_approvers)
+        do_required_approval_check = bool(self._required_approvers)
+        logger.debug(f"Do required approval check: {str(do_required_approval_check)}")
```
GeorgeSittas, line 365:
```suggestion
        logger.debug(f"Do required approval check: {do_required_approval_check}")
```

GeorgeSittas (approved):

PR #1672, opened 2023-11-07
chore: cleanup new macros

tobymao:


sqlmesh/core/macros.py
```diff
@@ -705,28 +723,21 @@ def star(
...
+
+    exclude = {e.name for e in except_.expressions}
+    quoted = quote_identifiers.this
+
+    return [
+        exp.cast(exp.column(column, table=alias.name, quoted=quoted), type_).as_(
+            f"{prefix.this}{column}{suffix.this}", quoted=quoted
         )
-    return projections
+        for column, type_ in evaluator.columns_to_types(relation).items()
```
GeorgeSittas, line 734: I think we need to convert `relation` to a string here. Not sure if `.sql()` suffices because it might generate quotes and the lookup inside the evaluator will probably fail (?).

tobymao: we shouldn't need to,  columns_to_types expects a table

sqlmesh/core/macros.py
```diff
@@ -798,44 +806,48 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
         >>> from sqlmesh.core.macros import MacroEvaluator
         >>> sql = "SELECT @SAFE_DIV(a, b) FROM foo"
         >>> MacroEvaluator().transform(parse_one(sql)).sql()
-        'SELECT a / CASE WHEN b = 0 THEN NULL ELSE b END FROM foo'
+        'SELECT a / NULLIF(b, 0) FROM foo'
     """
-    return numerator / exp.Case().when(denominator.eq(0), exp.null()).else_(denominator)
+    return numerator / exp.func("NULLIF", denominator, 0)
```
z3z1ma, line 811: I thought this wasn't ANSI so we use case when?

z3z1ma (commented):
This is great. I see how you were able to simplify and condense operations

z3z1ma:
I figured you'd appreciate the usage of the operators you added on the `exp` class. 😄  So cool

sqlmesh/core/macros.py
tobymao: after further research, nullif seems ubiquitous

sqlmesh/core/macros.py
```diff
@@ -798,44 +806,48 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
+    columns = {
+        column
+        for column, _ in reduce(
+            lambda a, b: a & b,  # type: ignore
+            (evaluator.columns_to_types(table).items() for table in tables),
+        )
+    }
+
+    projections = [
+        exp.cast(column, kind).as_(column)
```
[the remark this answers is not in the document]
tobymao: thanks

sqlmesh/core/macros.py
GeorgeSittas: Ah I think the type hint is probably wrong then?

```
    # `model_name` should be `str | Table`
    def columns_to_types(self, model_name: str) -> t.Dict[str, exp.DataType]:
```

GeorgeSittas: @tobymao `relation` isn't actually a `Table` though, right? I believe it will be parsed into a `Column` by the macro evaluator and so this will fail inside of `exp.to_table` (which is invoked in `MacroEvaluator.columns_to_Types`)

tobymao: i handle that

GeorgeSittas: Ah sorry I just saw the update 👍

PR #1673, opened 2023-11-07
Fix!: Use LONGTEXT data type when storing large blobs of text in MySQL

izeigerman:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -103,17 +104,20 @@ def __init__(
...
+        index_type = index_text_type(self.engine_adapter.dialect)
+        blob_type = blob_text_type(self.engine_adapter.dialect)
+
         self._snapshot_columns_to_types = {
-            "name": exp.DataType.build("text"),
-            "identifier": exp.DataType.build("text"),
-            "version": exp.DataType.build("text"),
-            "snapshot": exp.DataType.build("text"),
-            "kind_name": exp.DataType.build("text"),
+            "name": exp.DataType.build(index_type),
```
izeigerman, line 111: @GeorgeSittas any reason we didn't update this before?

tobymao: update what?

tobymao:
i'm fine with what you've done here, but you do understand that this isn't a proper migration because old systems won't change right. the proper way would have been to make a new migration to change the types

sqlmesh/core/state_sync/engine_adapter.py
GeorgeSittas: Hmm I'm not fully sure - maybe an oversight?

izeigerman:
@tobymao yes, you're right. Going to add a migration script for that

tobymao:
then you should revert all the changes to the previous migrations

georgesittas (commented):
The index type changes in the engine_adapter, plan modules look good, might have been an oversight on my end (at least for the engine adapter).

PR #1675, opened 2023-11-08
Fix!: allow partial schemas to display accurate missing schema warnings

georgesittas:
cc @z3z1ma

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -552,20 +552,6 @@ class _Model(ModelMeta, frozen=True):
                     tuple(str(part) for part in table.parts),
                     {k: v.sql(dialect=self.dialect) for k, v in mapping_schema.items()},
                 )
-            else:
-                logger.warning(
-                    "Missing schema for model '%s' referenced in model '%s'. "
-                    "Run `sqlmesh create_external_models` and / or make sure that the model '%s' "
-                    "can be rendered at parse time",
-                    dep,
-                    self.name,
-                    dep,
-                )
-
-                # Reset the entire mapping if at least one upstream dependency is missing from the mapping
-                # to prevent partial mappings from being used.
-                self.mapping_schema.clear()
-                return
 
     @property
     def depends_on(self) -> t.Set[str]:
@@ -575,12 +561,6 @@ class _Model(ModelMeta, frozen=True):
             A list of all the upstream table names.
         """
         return self.depends_on_ or set()
-
-    @property
-    def contains_star_projection(self) -> t.Optional[bool]:
-        """Returns True if the model contains a star projection, False if it does not, and None if this
-        cannot be determined at parse time."""
-        return False
 
     @property
     def columns_to_types(self) -> t.Optional[t.Dict[str, exp.DataType]]:
@@ -1026,13 +1006,6 @@ class SqlModel(_SqlBasedModel):
 
             self._depends_on -= {self.name}
         return self._depends_on
-
-    @property
-    def contains_star_projection(self) -> t.Optional[bool]:
-        query = self._query_renderer.render(optimize=False)
-        if query is None:
-            return None
-        return any(isinstance(expression, exp.Star) for expression in query.expressions)
 
     @property
     def columns_to_types(self) -> t.Optional[t.Dict[str, exp.DataType]]:
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -553,16 +553,17 @@ class _Model(ModelMeta, frozen=True):
                     {k: v.sql(dialect=self.dialect) for k, v in mapping_schema.items()},
                 )
             else:
+                logger.warning(
+                    "Missing schema for model '%s' referenced in model '%s'. "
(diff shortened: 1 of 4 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -553,16 +553,17 @@ def update_schema(
                     {k: v.sql(dialect=self.dialect) for k, v in mapping_schema.items()},
                 )
             else:
+                logger.warning(
```
izeigerman, line 556: Wouldn't this result in warning for models that are on the edge (at the root for DAG) even if they carefully list all projections and their types? Not sure this is a good experience.

tobymao: i think maybe we should go with the approach where we don't clear the mapping schema

GeorgeSittas: yea we were just chatting about this, i'll look into refactoring that section to have a more explicit signal for avoiding the optimizer

GeorgeSittas:
FYI @tobymao the tests were changed due to the `NULLIF` addition in SQLGlot: I was getting a couple of failures locally with latest main because `assert_exp_equal` was parsing the string using the default dialect I guess as opposed to the rewritten query being parsed using DuckDB

sqlmesh/core/model/definition.py
```diff
@@ -552,19 +552,6 @@ def update_schema(
...
-                if self.contains_star_projection:
-                    logger.warning(
-                        "Missing schema for model '%s' referenced in model '%s'. Run `sqlmesh create_external_models` "
-                        "and / or make sure that the model '%s' can be rendered at parse time",
-                        dep,
-                        self.name,
-                        dep,
-                    )
-                self.mapping_schema.clear()
-                return
```
georgesittas, line 567: The reasoning / assumption I made for getting rid of this section was that we already optimize @ load time anyway, so we will warn about missing schemas there and hence not needed to warn here as well earlier.

Regarding @izeigerman's [comment](https://github.com/TobikoData/sqlmesh/pull/1675#discussion_r1385780345), I don't think we want to omit warnings even for edge models, because we can't disambiguate between a valid, but external, model reference and an invalid one (e.g. that contains a typo). Happy to continue this discussion though as I may be missing context, though to my understanding this PR also doesn't significantly change the previous behaviour as we would already warn for these models @ optimize time.

georgesittas: Btw `self.mapping_schema` is involved in the computation of the metadata hash. Does this mean if we get rid of this section we'd also need to make a new migration script?

tobymao: yea, so if you upgrade sqlglot minor, it should be fine

tobymao (approved):

sqlmesh/core/model/definition.py
georgesittas: For reference, the current behavior in main w.r.t. the edge models argument (with both selecting everything & a single column w/ explicit type and alias):

```
➜  cat models/test.sql
model (name test, kind full);

select * from external
➜  sqlmesh render test
Missing schema for model 'external' referenced in model 'test'. Run `sqlmesh create_external_models` and / or make sure that the model 'external' can be rendered at parse time
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time
SELECT
  *
FROM "external" AS "external"
➜  cat models/test.sql
model (name test, kind full);

select a::int as a from external
➜  sqlmesh render test
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time
SELECT
  CAST("a" AS INT) AS "a"
FROM "external" AS "external"
```

Whereas, this PR changes to:
```
➜  cat models/test.sql
model (name test, kind full);

select * from external
➜  sqlmesh render test
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time.
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time.
SELECT
  *
FROM "external" AS "external"
➜  cat models/test.sql
model (name test, kind full);

select a::int as a from external
➜  sqlmesh render test
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time.
Query cannot be optimized due to missing schema for model 'external'. Run `sqlmesh create_external_models` and / or make sure that the model 'test' can be rendered at parse time.
SELECT
  CAST("a" AS INT) AS "a"
FROM "external" AS "external"
```

georgesittas:
@tobymao added a few tests, only one is currently failing. Can add others later as well.

georgesittas:
@tobymao thanks for the changes - looks good to go from my end I believe. Waiting for other reviews but can otherwise merge when ready.

PR #1678, opened 2023-11-08
fix: github cicd support uncategorized snapshots

eakmanrq:
I believe the way this happens is if a PR environment is created and then we go to report what was loaded compared to prod and prod has changed. If that is the case then when we diff we could get new snapshots that have not been categorized. This change allows those to flow through and just get reported as uncategorized snapshots.

### DIFF
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -272,27 +272,32 @@ class Plan:
         return self._restatements
 
     @property
-    def loaded_snapshot_intervals(self) -> t.List[LoadedSnapshotIntervals]:
+    def loaded_snapshot_intervals(
+        self,
+    ) -> t.Tuple[t.List[LoadedSnapshotIntervals], t.List[Snapshot]]:
         loaded_snapshots = []
+        unloaded_snapshots = []
         for snapshot in self.directly_modified:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/plan/definition.py
```diff
@@ -272,19 +272,32 @@ def restatements(self) -> t.Dict[str, Interval]:
...
         loaded_snapshots = []
+        unloaded_snapshots = []
         for snapshot in self.directly_modified:
-            if not snapshot.change_category:
-                continue
-            loaded_snapshots.append(LoadedSnapshotIntervals.from_snapshot(snapshot))
+            if snapshot.change_category:
+                loaded_snapshots.append(LoadedSnapshotIntervals.from_snapshot(snapshot))
+            else:
+                logger.debug(f"Got an unloaded snapshot. Snapshot: {snapshot.name}")
```
izeigerman, line 284: What's `unloaded snapshot`?

eakmanrq: A snapshot with no intervals and in this case because it doesn't have a change category. It is to contrast a `LoadedSnapshot` that is created within this method.

izeigerman (approved):

PR #1680, opened 2023-11-08
Fix: In dbt adapter normalize the model name before checking the table mapping

izeigerman:


sqlmesh/dbt/adapter.py
```diff
@@ -311,7 +311,10 @@ def resolve_identifier(self, relation: BaseRelation) -> t.Optional[str]:
     def _map_table_name(
         self, database: t.Optional[str], schema: t.Optional[str], identifier: t.Optional[str]
     ) -> exp.Table:
-        name = ".".join(p for p in (database, schema, identifier) if p is not None)
+        name = normalize_model_name(
+            exp.table_(identifier or "", db=schema, catalog=database),
```
tobymao, line 315: is this gonna work? "" as a default will just crash won't it? or is this just to trick mypy?

tobymao (approved):

sqlmesh/dbt/adapter.py
```diff
@@ -311,7 +311,10 @@ def resolve_identifier(self, relation: BaseRelation) -> t.Optional[str]:
     def _map_table_name(
         self, database: t.Optional[str], schema: t.Optional[str], identifier: t.Optional[str]
     ) -> exp.Table:
-        name = ".".join(p for p in (database, schema, identifier) if p is not None)
+        name = normalize_model_name(
```
georgesittas, line 314: is this call to `normalize_model_name` necessary? I think we'll get normalized identifiers when calling this?

cc @tobymao

tobymao: well it fixes the issue we're facing. i don't know if it's right

PR #1683, opened 2023-11-09
Chore: test migrations from N-1 to N

georgesittas:
@tobymao is this what you had in mind?

### DIFF
--- a/tests/core/test_state_sync.py
+++ b/tests/core/test_state_sync.py
@@ -35,6 +35,7 @@ from sqlmesh.core.state_sync import (
     cleanup_expired_views,
 )
 from sqlmesh.core.state_sync.base import (
+    MIGRATIONS,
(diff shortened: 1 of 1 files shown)


georgesittas:
Closing this for now as there may be problems in the future if the codebase state changes for version `N` compared to that in version `N-1` (chatted with Toby about this).

This happened for the v0031 -> v0032 migration because v0032 assumed the existence of `sqlmesh_version` in state sync and this was reflected in some parts of the codebase, whereas v0031 had no concept of that column and so the v0000 -> v0031 migration would reach these updated parts that were inconsistent with the state it represented and hence it'd fail.

One idea to explore if we want to add this kind of test in the future is to inject pre/post hooks in the test to monkeypatch the relevant parts of the codebase needed for specific values of `N`.

PR #1684, opened 2023-11-09
Feat: Add filter for up/downstream models  in lineage

mykalmax:
Screenshot 2023-11-09 at 10 44 36 AM

### DIFF
--- a/web/client/src/library/components/graph/Graph.tsx
+++ b/web/client/src/library/components/graph/Graph.tsx
@@ -947,7 +947,7 @@ function ModelColumnLineage({
           position="top-right"
           className="bg-theme !m-0 w-full"
         >
-          <GrpahControls nodes={nodes} />
+          <GraphControls nodes={nodes} />
         </Panel>
         <Controls className="bg-light p-1 rounded-md !border-none !shadow-lg" />
         <Background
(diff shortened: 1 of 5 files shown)


web/client/src/library/components/graph/Graph.tsx
```diff
@@ -970,80 +945,9 @@ function ModelColumnLineage({
...
-                  withColumns && 'Columns',
-                  hasBackground && 'Background',
-                  withConnected && 'Connected',
-                  withImpacted && 'Impact',
-                  withSecondary && 'Secondary',
-                ].filter(Boolean) as string[]
-              }
-            />
-          </div>
+          <GrpahControls nodes={nodes} />
```
vchan, line 950: GraphControls?

web/client/src/library/components/graph/Graph.tsx
```diff
@@ -1082,3 +986,167 @@ function ColumnLoading({
     </>
   )
 }
+
+function GrpahControls({ nodes = [] }: { nodes: Node[] }): JSX.Element {
```
vchan, line 990: Same comment as above about GraphControls.

vchan (approved):

PR #1685, opened 2023-11-09
Fix(mssql): bulk copy fails for timestamps with tz

treysp:


sqlmesh/core/audit/builtin.py
```diff
@@ -57,9 +57,9 @@
 number_of_rows_audit = ModelAudit(
     name="number_of_rows",
     query="""
-SELECT COUNT(*)
+SELECT COUNT(*) as _count
 FROM (
-   SELECT 1
+   SELECT 1 as _one
```
eakmanrq, line 62: I'm probably just out of the loop but is this not something we think SQLGlot should handle? I could see us forgetting this with future changes.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -191,14 +193,25 @@ def query_factory() -> Query:
...
+                        # bulk_copy() doesn't work with TZ timestamp, so load into string column and cast to
+                        # timestamp in SELECT statement
+                        columns_to_types_tztext = columns_to_types.copy()
+                        columns_to_types_tztext[column] = exp.DataType.build("TEXT")
                     else:  # type: ignore
                         df[column] = pd.to_datetime(df[column]).dt.strftime("%Y-%m-%d %H:%M:%S.%f")  # type: ignore
 
-            self.create_table(temp_table, columns_to_types)
+            self.create_table(
+                temp_table, columns_to_types_tztext if columns_to_types_tztext else columns_to_types
```
eakmanrq, line 205: [Nit] More common pattern in this codebase: `columns_to_types_tztext or columns_to_types`

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -191,14 +193,23 @@ def query_factory() -> Query:
                     elif getattr(df.dtypes[column], "tz", None) is not None:  # type: ignore
                         # MSSQL requires a colon in the offset (+00:00) so we use isoformat() instead of strftime()
                         df[column] = pd.to_datetime(df[column]).map(lambda x: x.isoformat(" "))  # type: ignore
+
+                        # bulk_copy() doesn't work with TZ timestamp, so load into string column and cast to
+                        # timestamp in SELECT statement
+                        columns_to_types_tztext = columns_to_types.copy()
```
tobymao, line 199: let's just make the copy upfront on line 189, that way we aren't copying n times for every tz

tobymao: so just modify this in place?

treysp: I see the problem, but I'm not following what you mean by modify in place. 

What if we move the dt64 detection outside the loop so we know up front whether there's any need to copy?

``` python
dt64_cols = {k:v for (k,v) in columns_to_types.items() if is_datetime64_any_dtype(df.dtypes[k])}
columns_to_types_tztext = columns_to_types.copy() if dt64_cols else None
for column, kind in dt64_cols.items():
    if kind.is_type("date"):  # type: ignore
    ...
```

tobymao: what i meant was

```python

for column, kind in (columns_to_types or {}).copy().items():

     columns_to_types[column] = Text

File: sqlmesh/core/engine_adapter/mssql.py [hunk 4]
```diff
@@ -184,21 +184,25 @@ def _df_to_source_queries(
...
+                        columns_to_types[column] = exp.DataType.build("TEXT")
                     else:  # type: ignore
                         df[column] = pd.to_datetime(df[column]).dt.strftime("%Y-%m-%d %H:%M:%S.%f")  # type: ignore
 
             self.create_table(temp_table, columns_to_types)
             rows: t.List[t.Tuple[t.Any, ...]] = list(df.itertuples(index=False, name=None))  # type: ignore
             conn = self._connection_pool.get()
             conn.bulk_copy(temp_table.sql(dialect=self.dialect), rows)
-            return exp.select(*columns_to_types).from_(temp_table)
+            return exp.select(*self._casted_columns(columns_to_types)).from_(temp_table)  # type: ignore
```
tobymao, line 205: @treysp where do we cast timestamps back? it looks like it's still a text type right

treysp: Ah, right - now that it's modified in place there isn't a copy of the original version containing the timestamp type.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -182,23 +182,29 @@ def _df_to_source_queries(
         temp_table = self._get_temp_table(target_table or "pandas")
 
         def query_factory() -> Query:
+            columns_to_types_create = columns_to_types.copy()
+
             # pymssql doesn't convert Pandas Timestamp (datetime64) types
             # - this code is based on snowflake adapter implementation
-            for column, kind in (columns_to_types or {}).items():
+            for column, kind in (columns_to_types_create or {}).items():
```
tobymao, line 189: you don't need the {} i assume columns_to_types is never none

tobymao (approved):

PR #1692, opened 2023-11-10
Feat: Support model selection for backfilling when running a plan against a dev environment

izeigerman:


docs/reference/cli.md
```diff
@@ -203,12 +203,13 @@ Options:
...
   --no-auto-categorization  Disable automatic change categorization.
-  --effective-from          The effective date from which to apply forward-only
-                            changes on production.
-  --include-unmodified      Whether to include unmodified models in the target
-                            development environment.
-  --model-selections        Model selection strings to filter the
-                            models that should be included in this plan.
+  --include-unmodified      Include unmodified models in the target
+                            environment.
+  --select-model TEXT       Select specific model changes that should be
```
tobymao, line 208: can this support multiple? should it be models?

sqlmesh/cli/main.py
```diff
@@ -287,14 +287,27 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     multiple=True,
     help="Select specific model changes that should be included in the plan.",
 )
+@click.option(
+    "--backfill-model",
```
tobymao, line 291: plurals?

sqlmesh/core/plan/definition.py
```diff
@@ -700,6 +714,13 @@ def _refresh_dag_and_ignored_snapshots(self) -> None:
                 "Revert or apply changes before proceeding with restatements."
             )
 
+        if self._input_backfill_models is not None:
+            if not self.is_dev:
```
tobymao, line 718: do we need to be so strict?

sqlmesh/cli/main.py
izeigerman: No, see `--select-model` and `--restate-model`. Multiple entires in CLI look like following:
```
sqlmesh plan --backfill-model model_a --backfill-model "<expression>"
```

sqlmesh/core/plan/definition.py
izeigerman: I strongly believe so. Why would anyone want to backfill prod partially?

sqlmesh/core/plan/definition.py
```diff
@@ -700,6 +714,13 @@ def _refresh_dag_and_ignored_snapshots(self) -> None:
...
+                raise PlanError(
+                    "Selecting models to backfill is only supported for development environments."
+                )
+            self._models_to_backfill = set(self.__dag.subdag(*self._input_backfill_models).sorted)
```
tobymao, line 722: so this will select all models upstream of models selected to ensure the chain is proper right

izeigerman: this is right

sqlmesh/core/plan/definition.py
tobymao: you never know

tobymao (approved):

docs/reference/cli.md
```diff
@@ -203,12 +203,13 @@ Options:
                             will fail.
   --auto-apply              Automatically apply the new plan after creation.
   --no-auto-categorization  Disable automatic change categorization.
-  --effective-from          The effective date from which to apply forward-only
```
tobymao, line 206: why was this removed?

tobymao:
this is a breaking change because you're changing user facing cli

docs/reference/cli.md
```diff
@@ -203,12 +203,13 @@ Options:
...
-                            changes on production.
-  --include-unmodified      Whether to include unmodified models in the target
-                            development environment.
-  --model-selections        Model selection strings to filter the
-                            models that should be included in this plan.
+  --include-unmodified      Include unmodified models in the target
+                            environment.
+  --select-model TEXT       Select specific model changes that should be
+                            included in the plan.
+  --backfill-model TEXT     Backfill only the models whose names match the
```
tobymao, line 210: does this support special syntax?

eakmanrq: For example did we decide to not support a syntax for saying directly modified only?

sqlmesh/core/plan/definition.py
eakmanrq: I agree with being strict here unless we hear a good use case. The product is opinionated about characteristics that make up prod and we so far this seems like a good bet.

docs/reference/cli.md
izeigerman: Depends on what you mean by a special syntax. If you mean `+`, `*`, etc. then yes. If you mean directly modified, then no. I realized supporting directly modified doesn't have much value.

docs/reference/cli.md
izeigerman: it was mentioned twice

eakmanrq (approved):

PR #1695, opened 2023-11-10
Fix: Wrapping of long lines when presenting a SQL diff to a user

izeigerman:


sqlmesh/core/console.py
```diff
@@ -581,7 +583,7 @@ def _prompt_categorize(self, plan: Plan, auto_apply: bool) -> None:
         self._show_categorized_snapshots(plan)
 
         for snapshot in plan.uncategorized:
-            self._print(Syntax(plan.context_diff.text_diff(snapshot.name), "sql"))
+            self.show_sql(plan.context_diff.text_diff(snapshot.name))
```
tobymao, line 586: shouldn't this depend on if the snapshot is sql or n ot?

tobymao (approved):

PR #1699, opened 2023-11-13
Fix: allow parsing functions as property values in model meta

georgesittas:
- Updates documentation to get rid of redundant parentheses in coalesced unique key
- Makes the parser more lenient so it can recognize parenthesized coalesced unique keys

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -281,14 +281,6 @@ def _parse_order(
     return macro
 
 
-def _parse_prop_value(self: Parser) -> t.Optional[exp.Expression]:
-    this = self._parse_string() or self._parse_function() or self._parse_id_var()
-    if self._match(TokenType.EQ):
-        this = exp.EQ(this=this, expression=self._parse_string() or self._parse_id_var())
-
-    return this
-
-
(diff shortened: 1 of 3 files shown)


sqlmesh/core/dialect.py
```diff
@@ -282,7 +282,7 @@ def _parse_order(
 
 
 def _parse_prop_value(self: Parser) -> t.Optional[exp.Expression]:
-    this = self._parse_string() or self._parse_id_var()
+    this = self._parse_string() or self._parse_function() or self._parse_id_var()
```
tobymao, line 285: should this be like parse_field ?

georgesittas: Ah probably, forgot about that

georgesittas: Ended up getting rid of the whole method, seems ok at first glance but haven't double-checked about edge cases. It seems like I added that section a while ago in https://github.com/TobikoData/sqlmesh/pull/1442/files#diff-32e7379e6d699794516fd766eee46a0a093be9bb23c3fdb4f8ea20117c2a44c3R272, not sure why I didn't use the equality parser at the time 🤷

tobymao (approved):

PR #1701, opened 2023-11-13
New UI layout (modules)

mykalmax:
New layout breaks all major interface units into separate modules. 
Modules can be navigated using modules navigation bar and modules located next to environment status group (with changes, backfills and current env state)
Screenshot 2024-01-18 at 1 54 51 PM
 
Old Layout
Screenshot 2024-01-18 at 1 49 22 PM

New Layout
Screenshot 2024-01-18 at 1 51 56 PM

New layout allows conditionally load different modules making it possible to have a `read-only` version of UI with only `Docs` module present


web/client/src/models/file.ts
```diff
@@ -44,6 +44,10 @@ export class ModelFile extends ModelArtifact<InitialFile> {
       initial?.content ?? this.initial.content ?? ''
   }
 
+  get shortName(): string {
```
vchan, line 47: Is this the filename without the extension? I've heard it called `basename` or more commonly `stem`.

web/server/models.py
```diff
@@ -113,6 +113,7 @@ class Directory(BaseModel):
 class Meta(BaseModel):
     version: str
     has_running_task: bool = False
+    modules: t.List[str] = []
```
vchan, line 116: What are all the possible modules? Maybe create an enum for them.

web/client/src/library/pages/docs/Docs.tsx
```diff
@@ -1,79 +1,106 @@
...
+                <div className="center">
+                  <h3 className="text-lg mb-4">
+                    Still Waiting For Models To Load...
+                  </h3>
+                  <p>It should take few more seconds.</p>
+                  <p>
+                    If it takes too long, probably, there is a problem with API
+                    response.
+                  </p>
+                  <p>Check if the server is running and refresh.</p>
```
vchan, line 81: Should we say this stuff at all?
```suggestion
                  <p>It should take a few more seconds.</p>
                  <p>
                    If it takes too long, there is probably a problem with the API
                    server.
                  </p>
                  <p>Check if the server is running and refresh.</p>
```

web/client/src/library/pages/plan/Content.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <PlanProvider>
+      {isNil(environment) ? (
+        <NotFound
+          link={EnumRoutes.Plan}
+          description={
+            isNil(environmentName)
+              ? undefined
+              : `Model ${environmentName} Does Not Exist`
```
vchan, line 35: Environment does not exist?

web/client/src/library/pages/root/NotFound.tsx
```diff
@@ -19,9 +19,9 @@ export default function NotFound({
           <h1 className="text-[4rem] md:text-[6rem] lg:text-[9rem] text-secondary-10 dark:text-primary-10 mb-4">
             Not Found
           </h1>
-          {descritpion != null && (
+          {description != null && (
```
vchan, line 22: isNotNil?

web/server/api/endpoints/meta.py
```diff
@@ -22,4 +22,5 @@ def get_api_meta(
     return models.Meta(
         version=_sqlmesh_version(),
         has_running_task=hasattr(request.app.state, "task") and not request.app.state.task.done(),
+        modules=[models.Modules.EDITOR, models.Modules.PLANS],
```
vchan, line 25: What does this do? Is there a way to disable the docs module?

web/server/models.py
```diff
@@ -43,6 +43,16 @@ class EventName(str, enum.Enum):
     PLAN_CANCEL = "plan-cancel"
 
 
+class Modules(str, enum.Enum):
+    EDITOR = "editor"
+    DOCS = "docs"
+    PLANS = "plans"
+    PLAN_ACTIVE = "plan-active"
```
vchan, line 50: What is the plan active module?

vchan (approved):

web/server/models.py
mykalmax: so intention was to display running  plan or plan apply as separate `module` so one can have `plans` module disabled but still be able to see running tasks and updates.  But it actually just placeholder , i remove it for now and add this option as separate PR

web/server/api/endpoints/meta.py
mykalmax: and option to read modules from global variable

vchan: Sounds good. Let's have it be consistent with the other modules so it can be enabled/disabled dynamically.

PR #1704, opened 2023-11-14
Fix: No gaps check for non-representative snapshots

izeigerman:


sqlmesh/core/plan/evaluator.py
```diff
@@ -176,16 +176,17 @@ def on_complete(snapshot: SnapshotInfoLike) -> None:
...
     def _promote(
-        self, plan: Plan, deployability_index: t.Optional[DeployabilityIndex] = None
+        self, plan: Plan, no_gaps_snapshot_names: t.Optional[t.Set[str]] = None
     ) -> PromotionResult:
         """Promote a plan.
 
         Args:
             plan: The plan to promote.
-            deployability_index: Indicates which snapshots are deployable in the context of this promotion.
+            no_gaps_snapshot_names: The names of snapshots to check for gaps. If not provided,
```
eakmanrq, line 185: `If not provided, and plan has no gaps enabled, all snapshots are checked`

sqlmesh/core/plan/evaluator.py
```diff
@@ -101,7 +101,7 @@ def evaluate(self, plan: Plan) -> None:
             self._push(plan, deployability_index)
             self._restate(plan)
             self._backfill(plan, before_promote_snapshots, deployability_index)
-            promotion_result = self._promote(plan, deployability_index)
+            promotion_result = self._promote(plan, before_promote_snapshots)
```
eakmanrq, line 104: Was the issue that we were getting the latest snapshots and those might have gaps but we weren't promoting latest but instead just promoting what we just backfilled?

izeigerman: No, I don't think so. Those were all latest snapshots. It just some snapshots can only be evaluated after promotion, but the no gap check was performed for all of them.

eakmanrq (approved):

PR #1712, opened 2023-11-14
Fix: Make sure there is only one copy of a new snapshot with the same ID when migration snapshot rows

izeigerman:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -902,10 +903,18 @@ def _migrate_rows(self) -> None:
...
-            snapshot_mapping[snapshot.snapshot_id] = new_snapshot
-            logger.debug(f"{snapshot.snapshot_id} mapped to {new_snapshot.snapshot_id}.")
+            new_snapshot_id = new_snapshot.snapshot_id
 
-        if not snapshot_mapping:
+            if (
+                new_snapshot_id not in new_snapshots
+                or new_snapshot.created_ts > new_snapshots[new_snapshot_id].created_ts
+            ):
+                new_snapshots[new_snapshot.snapshot_id] = new_snapshot
```
tobymao, line 912: should this be new_snapshot_id instead of new_snapshot.snapshot_id

tobymao (approved):

PR #1715, opened 2023-11-15
Fix!: remove auto expansion in render of unpushed snapshots

tobymao:
prior to this commit, render a model would auto expand any snapshot that did not have a version (uncommitted). this way the render query would always generate a runnable sql query. however, with large projects, expansion is extremely expensive and makes the default render unusable.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -202,19 +202,54 @@ class BaseExpressionRenderer:
         if not snapshots and not table_mapping and not expand:
             return expression
 
+        from sqlmesh.core.snapshot import to_table_mapping
+
         expression = expression.copy()
         with _normalize_and_quote(expression, self._dialect) as expression:
-            return _resolve_tables(
-                expression,
-                snapshots=snapshots,
-                table_mapping=table_mapping,
-                expand=expand,
-                deployability_index=deployability_index,
-                start=start,
-                end=end,
-                execution_time=execution_time,
-                **kwargs,
-            )
+            snapshots = snapshots or {}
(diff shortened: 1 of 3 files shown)


sqlmesh/core/renderer.py
```diff
@@ -202,19 +202,54 @@ def _resolve_tables(
...
-                snapshots=snapshots,
-                table_mapping=table_mapping,
-                expand=expand,
-                deployability_index=deployability_index,
-                start=start,
-                end=end,
-                execution_time=execution_time,
-                **kwargs,
-            )
+            snapshots = snapshots or {}
```
izeigerman, line 209: why is this no longer a function?

sqlmesh/core/renderer.py
```diff
@@ -202,19 +202,54 @@ def _resolve_tables(
...
+            table_mapping = table_mapping or {}
+            mapping = {**to_table_mapping(snapshots.values(), deployability_index), **table_mapping}
+            expand = {d.normalize_model_name(name, dialect=self._dialect) for name in expand} | {
```
izeigerman, line 212: Is this the only line that actually changed? Does it make sense to reduce the PR to just this relevant part?

sqlmesh/core/renderer.py
tobymao: it's not needed, it's not called anywhere outside of this instance method, and i needed  dialect

sqlmesh/core/renderer.py
tobymao: yes, this is the only line that should have changed

izeigerman (approved):

PR #1718, opened 2023-11-15
Chore: test python model table properties inline

treysp:
Some engines only allow specific table properties

### DIFF
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -1985,12 +1985,12 @@ def test_model_table_properties() -> None:
     def my_model(context, **kwargs):
         pass
 
-    m = model.get_registry()["my_model"].model(
+    python_model = model.get_registry()["my_model"].model(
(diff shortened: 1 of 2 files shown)


izeigerman (approved):

tests/core/test_model.py
```diff
@@ -1969,16 +1969,36 @@ def test_custom_interval_unit():
...
+            "format": "PARQUET",
+            "bucket_count": 0,
+            "orc_bloom_filter_fpp": 0.05,
+            "auto_purge": False,
+        },
+    )
+    def my_model(context, **kwargs):
+        pass
+
+    m = model.get_registry()["my_model"].model(
```
georgesittas, line 1988: [Nit] should we rename this to `python_model = ...` for consistency with the `sql_model` changes below?

georgesittas (approved):

PR #1723, opened 2023-11-15
Chore: fix quoting in sushi python model fetchdf calls

treysp:


examples/sushi/models/raw_marketing.py
```diff
@@ -22,7 +23,7 @@
     columns={
         "customer_id": "int",
         "status": "text",
-        "updated_at": "timestamp",
+        "updated_at": "timestamptz",
```
eakmanrq, line 26: Since it is UTC, does this make sure the `+0000` is preserved?

treysp: Yes - with just `timestamp` the `+0000` is being ignored by some engines (depending on their default timestamp data type behavior)

eakmanrq (approved):

PR #1732, opened 2023-11-16
Fix: Include old snapshot into the previous versions of the migrated snapshot

izeigerman:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -919,23 +916,16 @@ def _migrate_rows(self) -> None:
...
-            )
+            if versions and versions[-1].snapshot_id(name) in snapshot_id_mapping:
+                new_snapshot = new_snapshots[snapshot_id_mapping[versions[0].snapshot_id(name)]]
+                return new_snapshot.all_versions
+            return versions
 
         for from_snapshot_id, to_snapshot_id in snapshot_id_mapping.items():
             from_snapshot = all_snapshots[from_snapshot_id]
             to_snapshot = new_snapshots[to_snapshot_id]
-            to_snapshot.previous_versions = map_data_versions(
```
izeigerman, line 936: We don't need to do this since we no longer delete old snapshots. Instead we should include the old snapshot into the version history of the new one which we already do here:
https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/state_sync/engine_adapter.py#L892

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -919,23 +916,16 @@ def _migrate_rows(self) -> None:
...
-            version_ids = ((version.snapshot_id(name), version) for version in versions)
-
-            return tuple(
-                new_snapshots[snapshot_id_mapping[version_id]].data_version
-                if version_id in snapshot_id_mapping
-                else version
-                for version_id, version in version_ids
-            )
+            if versions and versions[-1].snapshot_id(name) in snapshot_id_mapping:
+                new_snapshot = new_snapshots[snapshot_id_mapping[versions[0].snapshot_id(name)]]
```
izeigerman, line 922: We don't need to map every version. We only need to check whether the latest one was mapped and fetch the version history from it.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -899,9 +899,6 @@ def _migrate_rows(self) -> None:
             if new_snapshot.fingerprint == snapshot.fingerprint:
                 logger.debug(f"{new_snapshot.snapshot_id} is unchanged.")
                 continue
-            if new_snapshot.snapshot_id in all_snapshots:
```
izeigerman, line 902: Even if the snapshot already exists and there was a hash collission we should still preserver it in the mapping in order to update environments correctly.

tobymao (approved):

PR #1738, opened 2023-11-18
fix: don't check for dialects when they are already known

tobymao:
before this change, dialect.parse always does a regex to find the model dialect override. this is expensive and not necessary when the dialect is known ahead of time like in serialization.


georgesittas (approved):
Makes sense 👍

PR #1745, opened 2023-11-21
Fix(mssql): pymssql doesn't convert datetime.date objects

treysp:


georgesittas:
Should we add a test for this?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -187,10 +187,12 @@ def query_factory() -> Query:
...
+                # pymssql errors if the column contains a datetime.date object
+                if kind.is_type("date"):  # type: ignore
+                    df[column] = pd.to_datetime(df[column]).dt.strftime("%Y-%m-%d")  # type: ignore
+
                 if is_datetime64_any_dtype(df.dtypes[column]):
-                    if kind.is_type("date"):  # type: ignore
-                        df[column] = pd.to_datetime(df[column]).dt.strftime("%Y-%m-%d")  # type: ignore
-                    elif getattr(df.dtypes[column], "tz", None) is not None:  # type: ignore
+                    if getattr(df.dtypes[column], "tz", None) is not None:  # type: ignore
                         # MSSQL requires a colon in the offset (+00:00) so we use isoformat() instead of strftime()
```
georgesittas, line 196: Just FMI: if we previously hit the `is_type("date")`, then it seems like we wouldn't hit the the timezone `elif` branch below it. Is this still true now that the former check has been moved before the `datetime64_any_dtype` branch? i.e. will we ever hit both the `is_type("date")` branch _and_ the timezone one now? Does that matter?

treysp: Good point - once the first conversion is done the column is guaranteed not to be datetime64 so we can skip that evaluation.

If the column type is date then any time-related info was just being ignored during the cast to date on the mssql side, so not hitting that branch wouldn't be an issue anyway.

treysp:
> Should we add a test for this?

Good idea, but I'm not sure how to implement. That code calls `bulk_copy`, which isn't executed as regular sql (so we can't validate based on the sql itself).

I think we'd need to look at the data rows themselves to ensure they're converted from date/time object to character, but I'm not sure how to reach in and examine them from the test. Is there a way to capture what `df.itertuples` returns?

https://github.com/TobikoData/sqlmesh/blob/9a5a34dbf5dd23ab3e4c2b4f91e2adb5c85d27c0/sqlmesh/core/engine_adapter/mssql.py#L204

georgesittas:
> Is there a way to capture what `df.itertuples` returns?

I think you can `mock.patch` it?

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -171,6 +171,24 @@ def drop_view(
...
+                df[column] = pd.to_datetime(df[column]).dt.strftime("%Y-%m-%d")  # type: ignore
+            elif is_datetime64_any_dtype(df.dtypes[column]):  # type: ignore
+                if getattr(df.dtypes[column], "tz", None) is not None:  # type: ignore
+                    # MSSQL requires a colon in the offset (+00:00) so we use isoformat() instead of strftime()
+                    df[column] = pd.to_datetime(df[column]).map(lambda x: x.isoformat(" "))  # type: ignore
+
+                    # bulk_copy() doesn't work with TZ timestamp, so load into string column and cast to
+                    # timestamp in SELECT statement
+                    columns_to_types[column] = exp.DataType.build("TEXT")
+                else:  # type: ignore
```
georgesittas, line 189: Nit: is the type ignore needed in L189?

georgesittas (commented):
Looks good

georgesittas (approved):

Issue #1637, opened 2023-10-30
Various unit test improvements


georgesittas:
- [x] Add option for allowing empty / unspecified columns in unit tests (related Slack thread [here](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1698620706683929))

  The use-case for this is that one can have wide tables but they may only want to write tests that focus on a group of related columns and ignore the rest.

- [x] Make unit test output easier to read (previous ticket: #948)

  When a unit test fails, the whole row shows as different from expected. For wide tables, this makes it difficult to parse where the difference was. The output should highlight which columns were different.

- [x] Automatically generate unit tests

  Writing unit tests by hand is cumbersome and error-prone, so one idea is to generate them automatically by specifying queries that will be used to populate the input models.

- [x] Allow running tests against a subset of resulting columns

  Users could have a really wide table and are wanting to test the output against a specific subset of the columns. This change would allow defining just that subset in the expected result and would not throw an error. 

- [x] Support nested data types

  Currently we don't properly support nested data types ([Arrays/Lists/Maps/Structs](https://duckdb.org/docs/sql/data_types/overview.html#nested--composite-types)). Slack context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1703270196743739

- [x] Improve error message when there are missing rows in the DataFrame diff (related Slack thread [here](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1711028889119499?thread_ts=1710984577.606679&cid=C044BRE5W4S)).

  If the actual and expected DataFrames have a different row count, we currently only report that there's a shape mismatch between the two, without providing details. One idea to improve the UX here is to report exactly which rows are missing from the actual data, or limit them using a threshold in case there are many of them.

- [x] Generate CTE fixtures too when using the `create_test` command

  The first iteration on the automatic unit test generation didn't include logic for producing CTE outputs.

- [x] Add optional setting which allows users to freeze time in order to test `CURRENT_TIMESTAMP` values and the like

  We can leverage [freezegun](https://github.com/spulec/freezegun) for this. The idea is that if we have e.g. a `freeze_time: <date or timestamp>` setting in a test, we'll wrap the execution of a query in the relevant context with the current time set to that value.

- [x] Ensure unit test fixtures are unique to avoid issues when the test connection is used by multiple users concurrently.

- [x] Add support for "dumping" duckdb state on test failure to speed up debugging

[...]

PR #1756, opened 2023-11-27
Feat: automate unit test generation

georgesittas:
This PR introduces a way to automate the generation of unit tests for a given model, by specifying how its dependencies should be populated with data. This is achieved through user-supplied queries which are executed against the default warehouse, but the interface can certainly be improved over time.

Currently, the CLI command looks like this:

```
$ sqlmesh create_test model_to_test -q dep1 'SELECT ...' -q dep2 'SELECT ...' ...
```

To be addressed in a future PR:
- Update other interfaces to add support for this command (Notebooks, UI)
- Add an "append" mode which updates an existing test, see https://github.com/TobikoData/sqlmesh/pull/1756#discussion_r1406937195

Notes:
- Updated some type hints for consistency, e.g. `dict[...]` to `t.Dict[...]`
- Updated the example project's `inputs` indentation for consistency
- Fixed a bug in `test/definition.py`:

```diff
    def tearDown(self) -> None:
-        """Drop all input tables"""
+        """Drop all fixture tables."""
        for table in self.body.get("inputs", {}):
-            self.engine_adapter.drop_view(table)
+            self.engine_adapter.drop_view(_test_fixture_name(table))
```

In the context of this PR, this was important because we would previously try to drop the actual model views instead of the fixtures.

Implements the third feature in https://github.com/TobikoData/sqlmesh/issues/1637.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -125,12 +125,6 @@ class ModelTest(unittest.TestCase):
         """Compute a version of this test's path relative to the `other` path"""
         return self.path.relative_to(other) if self.path else None
 
-    def _add_missing_columns(self, df: pd.DataFrame, columns: t.Iterable) -> None:
-        """Add missing columns to a given dataframe with None values."""
-        for index, column in enumerate(columns):
-            if column not in df:
-                df.insert(index, column, None)  # type: ignore
-
     @staticmethod
     def create_test(
         body: t.Dict[str, t.Any],
@@ -170,6 +164,12 @@ class ModelTest(unittest.TestCase):
 
     def __str__(self) -> str:
         return f"{self.test_name} ({self.path})"
+
+    def _add_missing_columns(self, df: pd.DataFrame, columns: t.Iterable) -> None:
+        """Add missing columns to a given dataframe with None values."""
+        for index, column in enumerate(columns):
+            if column not in df:
+                df.insert(index, column, None)  # type: ignore
 
     def _normalize_test(self, dialect: str | None) -> None:
         """Normalizes all identifiers in this test according to the given dialect."""
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -361,7 +361,9 @@ def generate_test(
     if isinstance(model, SqlModel):
         mapping = {name: _test_fixture_name(name) for name in models.keys() | inputs.keys()}
         model_query = model.render_query_or_raise(
-            **variables, engine_adapter=engine_adapter, table_mapping=mapping
+            **t.cast(t.Dict[str, t.Any], variables),
+            engine_adapter=engine_adapter,
+            table_mapping=mapping,
         )
         output = t.cast(SqlModelTest, test)._execute(model_query)
     else:
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -303,7 +303,9 @@ def generate_test(
     input_queries: t.Dict[str, str],
     models: t.Dict[str, Model],
     engine_adapter: EngineAdapter,
+    project_path: Path,
     overwrite: bool = False,
+    variables: t.Optional[t.Dict[str, str]] = None,
     path: t.Optional[str] = None,
     name: t.Optional[str] = None,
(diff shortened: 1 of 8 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -160,19 +168,67 @@ def create_test(
 
         raise TestError(f"Model '{model_name}' is an unsupported model type for testing at {path}")
 
+    @staticmethod
```
izeigerman, line 171: [Nit] I see little reason for it to be a class method since it doesn't actually return an instance of the class. I suggest to just make it a function

sqlmesh/core/test/definition.py
```diff
@@ -160,19 +168,67 @@ def create_test(
...
+        """
+        inputs: t.Dict[str, t.Any] = {}
+        for dep, query in input_queries.items():
+            if dep not in models:
+                raise ConfigError(f"Cannot find model for '{dep}'.")
+
+            inputs[dep] = engine_adapter.fetchdf(query).to_dict(orient="records")
+
+        outputs: t.Dict[str, t.Any] = {"query": {}}
+        test_name = f"test_{model.view_name}"
```
izeigerman, line 195: I think this should be overridable.

sqlmesh/core/test/definition.py
```diff
@@ -160,19 +168,67 @@ def create_test(
...
+            )
+            output = t.cast(SqlModelTest, test).execute(model_query)
+        else:
+            output = t.cast(PythonModelTest, test).evaluate()
+
+        outputs["query"] = output.to_dict(orient="records")
+
+        test.tearDown()
+
+        with open(Path(c.TESTS) / f"{test_name}.yaml", "w", encoding="utf-8") as file:
```
izeigerman, line 222: Shouldn't we check if the file already exists?

sqlmesh/core/test/definition.py
```diff
@@ -281,16 +337,17 @@ def __init__(
             models=models,
         )
 
+    def evaluate(self) -> pd.DataFrame:
```
izeigerman, line 340: We need to disambiguate the name. In a context of a test this reads as it evaluates this test

georgesittas: Good point

georgesittas:
@izeigerman re: overridability of path etc - agreed, can definitely work into adding that, this mostly serves as an mvp but adding a `--path` option was in my mind as well.

sqlmesh/core/test/definition.py
georgesittas: Makes sense

sqlmesh/core/test/definition.py
georgesittas: @izeigerman FWIW there's also an `execute` method on the `SqlModel` class. I'm thinking of adding a leading underscore to both of them and renaming this one to `_execute_model` or `_evaluate_model`. What do you think?

sqlmesh/core/test/definition.py
georgesittas: Updated

sqlmesh/core/test/definition.py
georgesittas: > Shouldn't we check if the file already exists?
> I think the file name should be customizable as well.

Addressed

sqlmesh/core/test/definition.py
```diff
@@ -281,25 +283,109 @@ def __init__(
...
+    path = path or f"{test_name}.yaml"
+
+    extension = path.split(".")[-1].lower()
+    if extension not in ("yaml", "yml"):
+        path = f"{path}.yaml"
+
+    fixture_path = project_path / c.TESTS / path
+    if not overwrite and fixture_path.exists():
+        raise ConfigError(
+            f"Fixture '{fixture_path}' already exists, make sure to set --overwrite if it can be safely overwritten."
```
izeigerman, line 339: Just curious: can't we just append in this case?

georgesittas: I guess we can merge the dicts in the test (i.e. update `inputs` and `query`), though that should probably require an action from the user right? Like `--append` or `--update`.

PR #1759, opened 2023-11-27
Feat!: Preliminary work to enable WAP support

izeigerman:


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -377,9 +334,14 @@ def audit(
                     execution_time=execution_time,
                     raise_exception=raise_exception,
                     deployability_index=deployability_index,
+                    wap_id=wap_id,
                     **kwargs,
                 )
             )
+
+        if wap_id is not None:
+            self._wap_publish_snapshot(snapshot, wap_id, deployability_index)
```
izeigerman, line 343: @tobymao @eakmanrq how do you feel about publishing implicitly as part of the successful audit call? Or should we make it explicit and introduce the `publish` method to `SnapshotEvaluator` and call it externally?

sqlmesh/core/context.py
```diff
@@ -679,11 +679,11 @@ def evaluate(
         """
         snapshot = self.get_snapshot(model_or_snapshot, raise_if_missing=True)
 
-        df = self.snapshot_evaluator.evaluate(
+        df = self.snapshot_evaluator.evaluate_and_fetch(
```
tobymao, line 682: this is a breaking change to the python api so add !

sqlmesh/core/snapshot/evaluator.py
tobymao: i think it's fine, because if you pass wap_id, you're opting into it

tobymao (approved):

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Yeah seems fine.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1290,6 +1290,18 @@ def fetch_pyspark_df(
         """Fetches a PySpark DataFrame from the cursor"""
         raise NotImplementedError(f"Engine does not support PySpark DataFrames: {type(self)}")
 
+    def wap_supported(self, table_name: TableName) -> bool:
```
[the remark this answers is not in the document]
izeigerman: Ah, this is a good point. Let me think about it

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -404,6 +366,143 @@ def close(self) -> None:
...
+
+        evaluation_strategy = _evaluation_strategy(snapshot, self.adapter)
+
+        wap_id: t.Optional[str] = None
+        if table_name and self.adapter.wap_supported(table_name):
+            wap_id = random_id()
+            logger.info("Using WAP ID '%s' for snapshot %s", wap_id, snapshot.snapshot_id)
+
+        def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
+            nonlocal wap_id
```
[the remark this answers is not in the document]
izeigerman: In my first iteration I was mutating the value.

sqlmesh/core/engine_adapter/base.py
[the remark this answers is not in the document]
eakmanrq: Btw I am making an assumption here which is that if a model is configured to store a table in Iceberg format that the engine supports WAP on Iceberg. This is generally true but I don't think it is always true. For example if I use Iceberg on Snowflake can I do WAP? What about Iceberg on BigQuery? These are more edge cases though.

PR #1760, opened 2023-11-28
Chore(docs): unify and use more expressive time_column names

treysp:


tobymao (approved):

docs/concepts/models/model_kinds.md
```diff
@@ -17,12 +17,12 @@ This example implements an `INCREMENTAL_BY_TIME_RANGE` model by specifying the `
 MODEL (
   name db.events,
   kind INCREMENTAL_BY_TIME_RANGE (
-    time_column ds
+    time_column event_date
```
mewwts, line 20: To me it is still a bit unclear that `time_column` is a magic keyword and `event_date` is a column in the model

mewwts: but it's better now. I thought `ds` was a magic variable earier

treysp: Updated it with some additional explanation - does that clarify?

PR #1761, opened 2023-11-28
Feat: expose snapshots to the MacroEvaluator

georgesittas:


sqlmesh/core/macros.py
```diff
@@ -114,6 +115,7 @@ def __init__(
         schema: t.Optional[t.Dict[str, t.Any]] = None,
         runtime_stage: RuntimeStage = RuntimeStage.LOADING,
         resolve_tables: t.Optional[t.Callable[[exp.Expression], exp.Expression]] = None,
+        snapshots: t.Optional[t.Dict[str, Snapshot]] = None,
```
tobymao, line 118: this will conflict with @eakmanrq work

tobymao:
let's add an example to sushi + document it in the macros section, and sync with ryan

georgesittas:
Cool, working on testing this now.

sqlmesh/core/macros.py
georgesittas: > Conclusion: I will update PR to have users go through a method to get the snapshot instead of calling the dict directly. When Ryan rebases this change into his PR he will update the snapshots dict to be fqn -> snapshot and then normalize within the method what the user passes in in order to get the normalized fqn to get the snapshot from the dict.

sqlmesh/core/macros.py
```diff
@@ -297,6 +300,15 @@ def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, ex
 
         return columns_to_types
 
+    def get_snapshot(self, model_name: TableName | exp.Column) -> t.Optional[Snapshot]:
```
tobymao, line 303: don't you need to normalize the model name here

georgesittas: Good point; fwiw I think this section will be rewritten soon by @eakmanrq when he rebases to get these changes.

For now I've addressed this in https://github.com/TobikoData/sqlmesh/pull/1761/commits/e67912b6d38559ecfea70eae642936e7c6b956d5

sqlmesh/core/macros.py
```diff
@@ -303,11 +304,10 @@ def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, ex
     def get_snapshot(self, model_name: TableName | exp.Column) -> t.Optional[Snapshot]:
         """Returns the snapshot that corresponds to the given model name."""
         if isinstance(model_name, str):
-            key = model_name
-        else:
-            key = ".".join(part.name for part in model_name.parts)
+            model_name = exp.to_table(model_name, dialect=self.dialect)
 
-        return self._snapshots.get(key)
+        normalize_model_name = normalize_identifiers(model_name, dialect=self.dialect)
```
tobymao, line 309: shouldn't this be normalize_model_name?

georgesittas: ah thanks I keep forgetting about these utilities. Done.

georgesittas: There's actually a small issue with `normalize_model_name` in case we get a column actually, because tables are built out of columns using `exp.table_(*reversed(table.parts[:-1]))`, but in the case of a macro the `Column` is actually just a `Table` represented differently and we don't want to skip `this` (i.e. the last element in `parts`). Perhaps we should convert the `Column` into a `Table` first..

georgesittas: Took a stab at addressing it like this in https://github.com/TobikoData/sqlmesh/pull/1761/commits/4c1086889257ad93a0a60d85c8268db7eea0bdf3

sqlmesh/core/macros.py
```diff
@@ -286,17 +290,20 @@ def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, ex
...
         columns_to_types = self._schema.find(exp.to_table(model_name))
         if columns_to_types is None:
             raise SQLMeshError(f"Schema for model '{model_name}' can't be statically determined.")
 
         return columns_to_types
 
+    def get_snapshot(self, model_name: TableName | exp.Column) -> t.Optional[Snapshot]:
+        """Returns the snapshot that corresponds to the given model name."""
+        if isinstance(model_name, exp.Column):
+            model_name = self._column_to_table(model_name)
```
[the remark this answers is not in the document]
georgesittas: See https://github.com/TobikoData/sqlmesh/pull/1761#discussion_r1409813524

sqlmesh/core/macros.py
tobymao: then normalize_model_name must be bugged?

georgesittas: It depends on the interface it's supposed to implement. If it wants to extract out & normalize the `catalog.db.table` part given a column like `catalog.db.table.col`, then it's not bugged. If it was meant to interpret a `Column` as a table, then yeah it's bugged. I bet it's the former though because the latter is unintuitive and not documented.

georgesittas: Ok let me check

georgesittas: It doesn't seem that way, some tests are breaking if I change the logic in `normalize_model_name`:

```
    def test_get_measure_and_dim_tables():
>       assert _get_measure_and_dim_tables(d.parse_one("SUM(a.x)"), "") == ("a", ())
E       AssertionError: assert ('a.x', ()) == ('a', ())
E         At index 0 diff: 'a.x' != 'a'
E         Use -v to get more diff
```

The code diff I get this with is

```diff
-        table = exp.table_(*reversed(table.parts[:-1]))  # type: ignore
-    else:
-        table = exp.to_table(table, dialect=dialect)
+        table = exp.table_(table.this, db=table.args.get("table"), catalog=table.args.get("db"))

-    return exp.table_name(normalize_identifiers(table, dialect=dialect))
+    table = exp.to_table(table, dialect=dialect)
+    return exp.table_name(
+        normalize_identifiers(exp.to_table(table, dialect=dialect), dialect=dialect)
+    )

tobymao (approved):

PR #1766, opened 2023-11-29
Feat: Add support for duckdb extensions and connector config

crericha:


sqlmesh/core/config/connection.py
```diff
@@ -101,11 +101,15 @@ class DuckDBConnectionConfig(ConnectionConfig):
...
         database: The optional database name. If not specified, the in-memory database will be used.
         catalogs: Key is the name of the catalog and value is the path.
+        extensions: A list of autoloadable extensions to load.
+        connector_config: A dictionary of connection configuration to pass into the duckdb connector.
         concurrent_tasks: The maximum number of tasks that can use this connection concurrently.
     """
 
     database: t.Optional[str] = None
     catalogs: t.Optional[t.Dict[str, str]] = None
+    extensions: t.Optional[t.List[str]] = None
```
eakmanrq, line 111: [Nit] Since this is Pydantic, I believe you could do this:
```
extensions: t.List[str] = []
```

Which then below you can just do:
```
for extension in self.extensions:
    ...
```

eakmanrq (approved):
nice!

PR #1767, opened 2023-11-29
Fix: handle empty list in select_from_values_for_batch_range

georgesittas:
The current implementation of select_from_values_for_batch_range constructs an empty VALUES () clause in case it's passed an empty list for its 'values' argument, which is usually a syntax error.

This commit makes it so we instead return a dummy query that's guaranteed to not return any rows (using a WHERE FALSE filter).

Additionally, it fixes a few bugs in the unit test module related to test cases that contain empty input / output rows.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -759,7 +759,7 @@ def select_from_values_for_batch_range(
     if not values:
         # Ensures we don't generate an empty VALUES clause & forces a zero-row output
         where = exp.false()
-        expressions = [tuple(exp.cast(exp.null(), to=kind) for _, kind in columns_to_types.items())]
+        expressions = [tuple(exp.cast(exp.null(), to=kind) for kind in columns_to_types.values())]
     else:
         where = None
         expressions = [
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -759,7 +759,7 @@ def select_from_values_for_batch_range(
     if not values:
         # Ensures we don't generate an empty VALUES clause & forces a zero-row output
         where = exp.false()
-        expressions =  [(exp.null(),) * len(columns_to_types)]
+        expressions = [(exp.null(),) * len(columns_to_types)]
     else:
         where = None
         expressions = [
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -755,23 +755,21 @@ def select_from_values_for_batch_range(
         exp.alias_(exp.cast(column, to=kind), column, copy=False)
         for column, kind in columns_to_types.items()
     ]
-    select = exp.select(*casted_columns)
 
-    # Make sure we don't generate an empty VALUES clause
-    if values:
-        values_exp = exp.values(
-            [tuple(transform_values(v, columns_to_types)) for v in values[batch_start:batch_end]],
-            alias=alias,
-            columns=columns_to_types,
-        )
-        return select.from_(values_exp, copy=False)
+    if not values:
+        where = exp.false()  # Force a zero-row output
+        values = [(exp.null(),) * len(columns_to_types)]  # Don't generate an empty VALUES clause
+    else:
+        where = None
+        values = values[batch_start:batch_end]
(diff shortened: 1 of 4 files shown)


sqlmesh/core/dialect.py
```diff
@@ -755,12 +755,23 @@ def select_from_values_for_batch_range(
...
+    if values:
+        values_exp = exp.values(
+            [tuple(transform_values(v, columns_to_types)) for v in values[batch_start:batch_end]],
+            alias=alias,
+            columns=columns_to_types,
+        )
+        return select.from_(values_exp, copy=False)
+
+    # If there are no values, return a dummy query that always produces no rows
+    for projection in select.selects:
```
tobymao, line 770: select.set("expressions", [exp.Literal.number(1)])

sqlmesh/core/dialect.py
```diff
@@ -755,12 +755,23 @@ def select_from_values_for_batch_range(
...
+        # There is no source to select from anymore so we swap out the cast columns with literals
+        projection.this.replace(exp.Literal.number(1))
+
+    return select.where(exp.false(), copy=False)
```
tobymao, line 774: let's just use copy false when things can get big, it litters the code base if we just put it everywhere

sqlmesh/core/dialect.py
tobymao: ah you need the same number? in the case, it should be nulls

tobymao: select.set("expressions", [exp.null().as_(alias) for alias in select.named_selects])

georgesittas: Got it makes sense, yeah I need the same number to match the aliases in things like create view etc

sqlmesh/core/dialect.py
georgesittas: sure

z3z1ma: Bigquery doesn't support a where clause without a from clause FYI

georgesittas: Interesting, thanks @z3z1ma. Perhaps make a dummy single value `VALUES` clause? :P

sqlmesh/core/dialect.py
```diff
@@ -755,12 +755,21 @@ def select_from_values_for_batch_range(
...
+
     values_exp = exp.values(
-        [tuple(transform_values(v, columns_to_types)) for v in values[batch_start:batch_end]],
+        [tuple(transform_values(v, columns_to_types)) for v in values],
         alias=alias,
         columns=columns_to_types,
     )
-    return exp.select(*casted_columns).from_(values_exp, copy=False)
+
+    return exp.select(*casted_columns).from_(values_exp).where(where)
```
tobymao, line 772: since this is taking a where statement here with values, this could be large, so we want to copy=False

georgesittas: yeah fair point, I added the `copy=False` kwarg back and did another refactor

tobymao (approved):

sqlmesh/core/dialect.py
```diff
@@ -755,12 +755,19 @@ def select_from_values_for_batch_range(
...
-        [tuple(transform_values(v, columns_to_types)) for v in values[batch_start:batch_end]],
-        alias=alias,
-        columns=columns_to_types,
-    )
-    return exp.select(*casted_columns).from_(values_exp, copy=False)
+
+    if not values:
+        # Ensures we don't generate an empty VALUES clause & forces a zero-row output
+        where = exp.false()
+        expressions = [tuple(exp.cast(exp.null(), to=kind) for _, kind in columns_to_types.items())]
```
georgesittas, line 762: FYI I added casts for the `NULL`s inside of the `VALUES` clause too because of a BQ issue I came across:

```
SELECT CAST(json_col AS JSON) AS json_col
FROM UNNEST([STRUCT(NULL AS id, NULL AS ds, NULL AS json_col)])
WHERE FALSE
```

Apparently this fails with "Invalid cast from INT64 to JSON at [1:13]"

PR #1769, opened 2023-11-29
Feat: Add Iceberg WAP support to the Spark engine adapter

izeigerman:


sqlmesh/core/model/kind.py
```diff
@@ -53,6 +53,7 @@ def is_incremental(self) -> bool:
             self.is_incremental_by_time_range
             or self.is_incremental_by_unique_key
             or self.is_incremental_unmanaged
+            or self.is_scd_type_2
```
eakmanrq, line 56: Can you explain why this change was made? SCD Type 2 doesn't inherit from the `_Incremental` model kind for example. What aspect of it makes it incremental in order to be added here?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -450,7 +469,11 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
 
         with self.adapter.transaction(), self.adapter.session():
             wap_id: t.Optional[str] = None
-            if table_name and snapshot.is_materialized and self.adapter.wap_supported(table_name):
+            if (
+                table_name
+                and snapshot.is_materialized
+                and (model.wap_supported or self.adapter.wap_supported(table_name))
```
eakmanrq, line 475: One tricky thing is what about engines like Trino or Databricks. Technically I could create Iceberg tables but Databricks wouldn't work since it relies on having a Spark Session which isn't always true for Databricks and obviously not implemented at all on Trino.

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -420,6 +419,91 @@ def create_view(
...
+        self.execute(
+            f"CALL {fqn.catalog}.system.cherrypick_snapshot('{fqn.db}.{fqn.name}', {iceberg_snapshot_id})"
+        )
+        self.execute(f"ALTER TABLE {fqn.sql(dialect=self.dialect)} DROP BRANCH {branch_name}")
+
     def _truncate_table(self, table_name: TableName) -> str:
         table = quote_identifiers(exp.to_table(table_name))
         return f"TRUNCATE TABLE {table.sql(dialect=self.dialect)}"
+
+    def _ensure_fqn(self, table_name: TableName) -> exp.Table:
```
eakmanrq, line 497: Even with my PR merged the fqn for a model may not include catalog/database if the user only specified a table. In that case we don't qualify with a catalog since a database doesn't exist. So we would have to keep this still right? Is it really required that they must have a database defined?

eakmanrq (commented):
Were you able to ensure that if an audit fails that the user gets an output that includes the full table path so they can check the data that was being audited?

sqlmesh/core/model/kind.py
izeigerman: Because this kind is conceptually incremental. Today we only use this method (`is_incremental`) when we check for data gaps. I think the check should apply to scd type 2 as well.

sqlmesh/core/snapshot/evaluator.py
izeigerman: It would still work if the format is set to `iceberg` explicitly, which I think would be required anyway in case of Databricks.

sqlmesh/core/engine_adapter/spark.py
izeigerman: Yes. Another reason to keep this is Airflow

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -321,6 +321,21 @@ def audit(
...
+            original_table_name = snapshot.table_name(
+                is_deployable=deployability_index.is_deployable(snapshot)
+            )
+            wap_table_name = self.adapter.wap_table_name(original_table_name, wap_id)
+            logger.info(
+                "Auditing WAP table '%s', snapshot %s", wap_table_name, snapshot.snapshot_id
+            )
+
+            table_mapping = kwargs.get("table_mapping") or {}
+            table_mapping[snapshot.name] = wap_table_name
```
izeigerman, line 335: @eakmanrq please note that this will have to be reflected in your PR if this gets merged eariler.

sqlmesh/core/model/kind.py
eakmanrq: Want to change this to something like `can_be_partiall_loaded` or `allows_loading_subset`?

sqlmesh/core/engine_adapter/spark.py
eakmanrq: We talked and even when Airflow support default catalog we still need this since in the case where a database is not defined the wap branch syntax won't work since it uses a `.` to separate and the only way to know that there is a branch being supplied is if the name is fully qualified and therefore there are 4 `.`.

sqlmesh/core/model/kind.py
izeigerman: what's wrong with `is_incremental`?

sqlmesh/core/snapshot/evaluator.py
izeigerman: Updated databricks

eakmanrq (approved):

PR #1773, opened 2023-11-29
Fix!: Include sql statements in metadata hash.

crericha:


sqlmesh/core/model/definition.py
```diff
@@ -942,6 +948,10 @@ def _data_hash_values(self) -> t.List[str]:
             *[e.sql(comments=False) for e in (*pre_statements, *post_statements, *macro_defs)],
         ]
 
+    @property
+    def _additional_metadata(self) -> t.List[str]:
+        return [s.sql(comments=False) for s in (*self.pre_statements, *self.post_statements)]
```
tobymao, line 953: this should be cleaned up, data_hash_values already does this when hash_raw_query is true, can you dry it up?

also, in the metadata one, you can remove the comments=False, because you want all metadata changes

tobymao (commented):
z

tobymao (approved):

PR #1780, opened 2023-12-01
Fix: Take into account dbt vars, sources, and refs when computing snapshot fingerprint

izeigerman:


sqlmesh/dbt/common.py
```diff
@@ -178,12 +178,14 @@ class Dependencies(PydanticModel):
...
     sources: t.List[str] = []
     refs: t.List[str] = []
+    variables: t.List[str] = []
 
     def union(self, other: Dependencies) -> Dependencies:
         return Dependencies(
             macros=list(set(self.macros) | set(other.macros)),
             sources=list(set(self.sources) | set(other.sources)),
             refs=list(set(self.refs) | set(other.refs)),
+            variables=list(set(self.variables) | set(other.variables)),
```
tobymao, line 188: why are these stored as randomly sorted lists?

sqlmesh/dbt/context.py
```diff
@@ -238,9 +238,13 @@ def context_for_dependencies(self, dependencies: Dependencies) -> DbtContext:
             else:
                 raise ConfigError(f"Source '{source}' was not found.")
 
+        variable_deps = set(dependencies.variables)
```
tobymao, line 241: and here you convert back from list to set, maybe these should all just be sets?

sqlmesh/dbt/manifest.py
```diff
@@ -341,6 +367,10 @@ def _extra_dependencies(self, target: str) -> Dependencies:
                     ref = ".".join(args)
                     if not self._is_disabled_ref(ref):
                         dependencies.refs.append(ref)
+            elif call_name[0] == "var":
+                args = [_jinja_call_arg_name(arg) for arg in node.args]
```
tobymao, line 371: arg = seq_get(...)
if arg:
dependencies.append(arg)

sqlmesh/utils/jinja.py
```diff
@@ -358,6 +359,26 @@ def to_expressions(self) -> t.List[Expression]:
...
 
+    @property
+    def data_hash_values(self) -> t.List[str]:
+        data = []
+
+        for macro_name, macro in sorted(self.root_macros.items()):
+            data.append(macro_name)
+            data.append(macro.definition)
+
+        for _, package in sorted(self.packages.items(), key=lambda x: x[0]):
```
tobymao, line 370: you don't need to pass in key right, can't you just do 

sorted({"x": 1}.items())

lambda x: x[0] is effectively doing the same thing?

izeigerman: I think you're right

sqlmesh/dbt/common.py
```diff
@@ -176,21 +176,18 @@ class Dependencies(PydanticModel):
     """
 
     macros: t.List[MacroReference] = []
```
crericha, line 178: What's the purpose of macros remaining as a list, while the others are now sets?

tobymao: you can use a frozenset if you want hashability

izeigerman: How `frozenset` helps with hashability? The problem is that `MacroReference` is serialized into dict which is not hasbable. Set is not a problem

sqlmesh/dbt/context.py
```diff
@@ -238,9 +238,13 @@ def context_for_dependencies(self, dependencies: Dependencies) -> DbtContext:
             else:
                 raise ConfigError(f"Source '{source}' was not found.")
 
+        variable_deps = dependencies.variables
```
crericha, line 241: Are you using `variable_deps` so that the next piece fits on one line?

izeigerman: mostly, yes

sqlmesh/dbt/manifest.py
```diff
@@ -308,17 +310,41 @@ def _load_disabled(self) -> None:
...
         dependencies = Dependencies()
         for macro in macros:
-            macro_config = self._macros_per_package[macro.package or default_package].get(
+            macro_package = macro.package or default_package
+
+            if (macro_package, macro.name) in visited:
+                continue
+            visited.add((macro_package, macro.name))
+
+            macro_dependencies = self._macro_flatten_dependencies.get(macro_package, {}).get(
```
crericha, line 330: Should we use `none` as the default for the case when the macro truly has no dependencies?

crericha (approved):

sqlmesh/dbt/context.py
crericha: My personal preference is not to introduce extra variables. black does a good enough job imo. Up to you though.

sqlmesh/utils/jinja.py
```diff
@@ -358,6 +359,26 @@ def to_expressions(self) -> t.List[Expression]:
...
+            data.append(macro_name)
+            data.append(macro.definition)
+
+        for _, package in sorted(self.packages.items()):
+            for macro_name, macro in sorted(package.items()):
+                data.append(macro_name)
+                data.append(macro.definition)
+
+        trimmed_global_objs = {
+            k: self.global_objs[k] for k in ("refs", "sources", "vars") if k in self.global_objs
```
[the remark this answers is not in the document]
izeigerman: I consider that but it feels redundant to serialize this and store this in state when we know these values at compile time. Additionally, if we change our minds in future, we'll have to migrate these values, which is even more annoying.

sqlmesh/dbt/manifest.py
izeigerman: If the dict for that *package* wasn't set yet, we fallback to `{}` to safely chain with a subsequent `get()`.

crericha: Ahh, my eyes completely glanced over that since it wasn't a changed line. Makes sense now.

Issue #1768, opened 2023-11-29
CICD bot diff preview not showing change to macro variable


plaflamme:
I have a model with a local variable defined:

```
MODEL(...)
@DEF(
    cols, 
    [a,b,c]
);
...
```

When adding an item to the list, e.g.:

```
MODEL(...)
@DEF(
  cols,
  [a,b,c,d]
)
...
```

The rendered diff preview looks like this in the Github comment:

```diff
--- 

+++ 

@@ -1 +1 @@

-@DEF(cols, )
+@DEF(cols, )
```

eakmanrq:
@georgesittas There shouldn't be anything specific to CI/CD bot that would create this behavior. So if your PR fixed the diff in normal plan/apply output then it should also be fixed in the bot. Is that the case?

georgesittas:
I couldn't repro this in the CLI after checking out to the parent of https://github.com/TobikoData/sqlmesh/commit/8f25d5be55e026ee94b710edc9381fdb1ca5306f:

```
➜  test sqlmesh init duckdb
➜  test sqlmesh render sqlmesh_example.full_model
SELECT
  "incremental_model"."item_id" AS "item_id",
  COUNT(DISTINCT "incremental_model"."id") AS "num_orders",
  1 IN (1, 2, 3, 4) AS "_col_2"
FROM "sqlmesh_example"."incremental_model" AS "incremental_model"
GROUP BY
  "incremental_model"."item_id"
ORDER BY
  "item_id"
➜  test sqlmesh plan
...
➜  test sqlmesh render sqlmesh_example.full_model
SELECT
  "incremental_model"."item_id" AS "item_id",
  COUNT(DISTINCT "incremental_model"."id") AS "num_orders",
  1 IN (1, 2, 3, 4, 5) AS "_col_2"
FROM "sqlmesh__sqlmesh_example"."sqlmesh_example__incremental_model__3958293195" AS "incremental_model"
GROUP BY
  "incremental_model"."item_id"
ORDER BY
  "item_id"
➜  test sqlmesh plan
======================================================================
Successfully Ran 0 tests against duckdb
----------------------------------------------------------------------
Summary of differences against `prod`:
Models:
└── Directly Modified:
    └── sqlmesh_example.full_model
---

+++

@@ -5,7 +5,7 @@

   audits (ASSERT_POSITIVE_ORDER_IDS()),
   grains (item_id)
 )
-@DEF(vals, [1, 2, 3, 4])
[...]

PR #1781, opened 2023-12-01
Fix: preserve Array macro vars instead of converting them to Tuple

georgesittas:
This fixes the following bug: let's assume we add a `@DEF(cols, ['1', '2', '3', '4']);` in the example full model and include it in the projection list. Then, without this PR we have:

```
➜  sqlmesh render sqlmesh_example.full_model
SELECT
  "incremental_model"."item_id" AS "item_id",
  COUNT(DISTINCT "incremental_model"."id") AS "num_orders",
  ('1', '2', '3', '4') AS "_col_2"  -- Invalid DuckDB, we need an array here
FROM "sqlmesh__sqlmesh_example"."sqlmesh_example__incremental_model__3958293195" AS "incremental_model"
GROUP BY
  "incremental_model"."item_id"
ORDER BY
  "item_id"
```

Including the PR we get:

```
➜  sqlmesh render sqlmesh_example.full_model
SELECT
  "incremental_model"."item_id" AS "item_id",
  COUNT(DISTINCT "incremental_model"."id") AS "num_orders",
  ['1', '2', '3', '4'] AS "_col_2"
FROM "sqlmesh__sqlmesh_example"."sqlmesh_example__incremental_model__3958293195" AS "incremental_model"
GROUP BY
  "incremental_model"."item_id"
ORDER BY
  "item_id"
```

This may or may not be related to #1768, cc @eakmanrq if you take a look into it.

### DIFF
--- a/sqlmesh/core/model/decorator.py
+++ b/sqlmesh/core/model/decorator.py
@@ -29,15 +29,19 @@ class model(registry_decorator):
         self.is_sql = is_sql
         self.kwargs = kwargs
 
-        # Make sure that argument values are expressions in order to
-        # pass validation in ModelMeta.
+        # Make sure that argument values are expressions in order to pass validation in ModelMeta.
         calls = self.kwargs.pop("audits", [])
         self.kwargs["audits"] = [
             (call, {})
             if isinstance(call, str)
             else (
                 call[0],
-                {arg_key: exp.convert(arg_value) for arg_key, arg_value in call[1].items()},
+                {
+                    arg_key: exp.convert(
+                        tuple(arg_value) if isinstance(arg_value, list) else arg_value
+                    )
+                    for arg_key, arg_value in call[1].items()
+                },
             )
             for call in calls
         ]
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/decorator.py
```diff
@@ -29,15 +29,19 @@ def __init__(self, name: str, is_sql: bool = False, **kwargs: t.Any) -> None:
...
         calls = self.kwargs.pop("audits", [])
         self.kwargs["audits"] = [
             (call, {})
             if isinstance(call, str)
             else (
                 call[0],
-                {arg_key: exp.convert(arg_value) for arg_key, arg_value in call[1].items()},
+                {
+                    arg_key: exp.convert(
+                        tuple(arg_value) if isinstance(arg_value, list) else arg_value
```
izeigerman, line 41: Very nice! I hope this was the only reason 🤞 Good place to fix this otherwise.

izeigerman (approved):

sqlmesh/core/model/decorator.py
georgesittas: I hope so too! 🤞 😅

tobymao:
i remember why this was done now. because in audits, people would
specify values =[1,2] but you can do where x in [] and need to convert it to a tuple

PR #1782, opened 2023-12-01
fix(redshift): ctas now uses false limit 0 instead of from false prop…

tobymao:
…ogation

in redshift, there's a bug where false propogation to from statements with md5(...) was resulting in varchar(1) types instead of varchar(32).

it seems like where false limit 0 is safe in most engines.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -417,11 +417,13 @@ class _Model(ModelMeta, frozen=True):
         Return:
             The mocked out ctas query.
         """
-        query = self.render_query_or_raise(**render_kwarg)
         # the query is expanded so it's been copied, it's safe to mutate.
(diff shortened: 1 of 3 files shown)


georgesittas:
@tobymao what about [this](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/dialect.py#L767) section? Is it also affected by this bug?

tobymao:
> @tobymao what about [this](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/dialect.py#L767) section? Is it also affected by this bug?

probably

tobymao:
> > @tobymao what about [this](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/dialect.py#L767) section? Is it also affected by this bug?
> 
> probably

no, it's not affected  because you have a cast

PR #1783, opened 2023-12-01
Feat: add flag to skip generating files in example project

georgesittas:
I've found that many times I just want an empty SQLMesh project with the directories set up for me. Currently I'm just running `sqlmesh init <dialect>` and then get rid of the files myself, but that's cumbersome.

This adds a new template `empty` which only contains the SQLMesh project directories and no files under them.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -177,8 +177,7 @@ Usage: sqlmesh init [OPTIONS] [SQL_DIALECT]
 
 Options:
   -t, --template TEXT  Project template. Supported values: airflow, dbt,
-                       default.
-  -e, --empty          Create empty project directories.
+                       default, empty.
   --help               Show this message and exit.
 ```
 
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -105,14 +105,12 @@ def cli(
     "-t",
     "--template",
     type=str,
-    help="Project template. Supported values: airflow, dbt, default.",
+    help="Project template. Supported values: airflow, dbt, default, empty.",
 )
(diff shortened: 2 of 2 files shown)


tobymao (approved):

docs/reference/cli.md
```diff
@@ -171,14 +171,14 @@ Options:
...
 
-  Create a new SQLMesh repository. Argument SQL_DIALECT is required unless the dbt
-  template option is specified.
+  Create a new SQLMesh repository.
 
 Options:
-  -t, --template TEXT  Project template. Support values: airflow, dbt,
+  -t, --template TEXT  Project template. Supported values: airflow, dbt,
                        default.
+  -e, --empty          Create empty project directories.
```
izeigerman, line 181: Rather than adding a new flag, can we just add another template with name `empty`?

georgesittas: I like this more, didn't occur to me. Thanks!

georgesittas: Ah on second thought @izeigerman, isn't generating an empty project orthogonal to the template? For example we might want _both_ and empty project _and_ and airflow template.

izeigerman: We can yet another template for this if needed. Conversely, the `dbt` template doesn't generate any model files, and is technically `empty` as it is.

georgesittas: The `dbt` template currently only creates the config file, so no SQLMesh project structure.

FWIW I'm ok either way, we can just roll out `--template empty` (current version of the PR) if it looks good.

izeigerman (approved):

PR #1785, opened 2023-12-01
chore: improve readability of logs

tobymao:


sqlmesh/core/model/definition.py
```diff
@@ -1158,9 +1157,6 @@ def _data_hash_values(self) -> t.List[str]:
 
         return data
 
-    def __repr__(self) -> str:
```
izeigerman, line 1161: Why all model-related reprs have been removed?

sqlmesh/core/model/definition.py
```diff
@@ -1330,9 +1326,6 @@ def _data_hash_values(self) -> t.List[str]:
             data.append(column_hash)
         return data
 
-    def __repr__(self) -> str:
-        return f"Model<name: {self.name}, seed: {self.kind.path}>"
```
izeigerman, line 1334: I think this is still useful

sqlmesh/core/node.py
```diff
@@ -181,10 +182,15 @@ class _Node(PydanticModel):
...
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
+    _path: Path = Path()
 
     _croniter: t.Optional[CroniterCache] = None
     __inferred_interval_unit: t.Optional[IntervalUnit] = None
 
+    def __str__(self) -> str:
+        path = f": {self._path.name}" if self._path else ""
+        return f"{self.__class__.__name__}<{self.name}{path}>"
```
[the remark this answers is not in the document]
izeigerman: Please note that `path` is not available after reading from state

sqlmesh/core/model/definition.py
tobymao: pydantic

sqlmesh/core/snapshot/definition.py
```diff
@@ -299,6 +311,20 @@ class SnapshotTableInfo(PydanticModel, SnapshotInfoMixin, frozen=True):
...
+        "version",
+        "physical_schema",
+        "change_category",
+        "kind_name",
+    )
+
+    def __str__(self) -> str:
+        args = ", ".join(f"{k}: {getattr(self, k)}" for k in self.REPR_FIELDS)
+        temp_version = f", {self.temp_version}" if self.temp_version else ""
+        return f"SnapshotInfo<{args}{temp_version}>"
```
izeigerman, line 326: We should use the original class name. `SnapshotInfo` refers to an entity which doesn't exist

sqlmesh/core/model/definition.py
tobymao: still have it

sqlmesh/core/snapshot/definition.py
```diff
@@ -119,11 +125,17 @@ def snapshot_id(self) -> SnapshotId:
     def __lt__(self, other: SnapshotId) -> bool:
         return self.name < other.name
 
+    def __str__(self) -> str:
+        return f"{self.name}: {self.identifier}"
```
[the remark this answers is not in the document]
izeigerman: IMHO I'd rather not invent new formats and just rely on pydantic.

sqlmesh/core/model/definition.py
izeigerman: Can you perhaps show an example of how this looks like for seed models now?

sqlmesh/core/node.py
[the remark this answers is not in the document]
tobymao: how is it not using path?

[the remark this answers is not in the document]
izeigerman:
```
f"{self.__class__.__name__}<{self.name}{path}>"
```
Path is missing curly braces and hence is not interpolated, right?

izeigerman (approved):

PR #1789, opened 2023-12-01
fix: make redshift late binding again

eakmanrq:
I also updated the new project integration test (thanks again @treysp!) to also do a run dag to expose issues like this one.

Edit: Removed the adding of dag since it caused weird fatal errors when running locally. Will look to add it back later.

### DIFF
--- a/tests/core/engine_adapter/test_integration.py
+++ b/tests/core/engine_adapter/test_integration.py
@@ -884,11 +884,8 @@ def test_scd_type_2(ctx: TestContext):
 
 
 def test_sushi(ctx: TestContext):
-    if ctx.test_type == "df":
+    if ctx.test_type != "query":
         pytest.skip("Sushi end-to-end tests only need to run for query and pyspark tests")
-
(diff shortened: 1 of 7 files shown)


tests/core/engine_adapter/test_integration.py
```diff
@@ -884,72 +885,76 @@ def test_scd_type_2(ctx: TestContext):
 
 
 def test_sushi(ctx: TestContext):
-    if ctx.test_type == "df":
+    if ctx.test_type != "query":
```
eakmanrq, line 888: @treysp I realized that we don't need to run pyspark. The reason is that the way it works is that it takes the given dataframe input and converts it to pyspark (and converts it to SQL for the query tests). Since we aren't providing explicit inputs here like the other tests then we don't convert anything to pyspark so the pyspark test was just the query test run again. This also means we can remove the init since we just aren't including pyspark.

tobymao (approved):

PR #1797, opened 2023-12-02
feat: redshift plan parsing to address ctas varchar(1) bug

tobymao:
in redshift, CTAS types are not determinstic based on the logical plan of a query. adding limits or where false to an md5(...) will change from the expected varchar(32) into a varchar(1). this pr tries to detect these and then force casts projections to the right types.

### DIFF
--- a/sqlmesh/core/engine_adapter/redshift.py
+++ b/sqlmesh/core/engine_adapter/redshift.py
@@ -5,6 +5,7 @@ import uuid
 
 import pandas as pd
 from sqlglot import exp
+from sqlglot.errors import ErrorLevel
 
 from sqlmesh.core.engine_adapter.base_postgres import BasePostgresEngineAdapter
 from sqlmesh.core.engine_adapter.mixins import (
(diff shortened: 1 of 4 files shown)


eakmanrq:
It would be nice if we could have a test that actually generates a plan on the engine and confirms the format has not changed from what we have assumed the format is for this parser to work. Do you know if postgres explain plans are the same format? If so we could add an engine test just for Postgres that will verify the explain plan matches what we expect.

tobymao:
ready for review

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -71,6 +72,48 @@ def _create_table_from_source_queries(
...
+            for i, target in enumerate(plan["targetlist"] if plan else []):  # type: ignore
+                if target["name"] == "TARGETENTRY":
+                    resdom = target["resdom"]
+                    # https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat
+                    if resdom["restype"] == "1043" and resdom["restypmod"] == "- 1":
+                        select = statement.selects[i]
+
+                        if not isinstance(select, exp.Cast):
+                            select.replace(
+                                exp.cast(select.copy(), "VARCHAR(MAX)", dialect="redshift")
```
izeigerman, line 112: [Nit] `dialect=self.dialect`

izeigerman (approved):

PR #1798, opened 2023-12-03
Fix: set DataFrame column names explicitly in unit tests

### LINKED ISSUE #1786: Omitting columns in the test inputs doesn't work in particular cases
# Issue description

Let's say we have an external model `model1` defined as follows:
```yaml
- name: dummy.model1
  columns:
    id: INT
    name: STRING
    price: DECIMAL(38, 9)
    quantity: DECIMAL(38, 9)
```
And a model `model2` using `model1` defined as follows:
```sql
MODEL (
  name dummy.model2,
  kind FULL,
);

SELECT
  *
FROM
  dummy.model1
```
And finally a unit test for `model2` defined as follows:
```yaml
test_omitted_column:
  model: dummy.model2
  inputs:
    dummy.model1:
      rows:
        - id: 1234
          quantity: 0.1
        - id: 9876
          name: hello
          quantity: 0.2
  outputs:
    query:
      rows:
        - id: 1234
          quantity: 0.1
        - id: 9876
          name: hello
          quantity: 0.2
```
In row #1, the columns `name` and `quantity` are omitted.
In row #2, the column `quantity` is omitted.

When running this test, I get the following error:
```bash
  duckdb.duckdb.ConversionException: Conversion Error: Could not convert string "hello" to DECIMAL(38,9)
```
If I swap the order of rows #1 and #2 (in the inputs and expected outputs) everything works fine.
⇨ it seems to works if the larger subset of columns appears first in the `inputs`.

# Analysis

When rendering the fixture view used to mock the `model1` data, the generated query is the following:
```sql
SELECT
  CAST(id AS INT) AS id,
  CAST(name AS VARCHAR) AS name,
  CAST(price AS DECIMAL(38, 9)) AS price,
  CAST(quantity AS DECIMAL(38, 9)) AS quantity
FROM (
  VALUES(
    1234, -- id
    0.1,  -- quantity instead of name
    NULL,
    NULL
  ), (
    9876, -- id
    0.2,  -- quantity instead of name
    NULL,
   'hello' -- name instead of quantity
  )
) AS t(id, name, price, quantity)
```

We can see that there is a mismatch in the order of the non-omitted columns specified in the inputs and the values used in the fixture query.

georgesittas:
Fixes #1786

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -82,7 +82,7 @@ class ModelTest(unittest.TestCase):
                     schema_(table.args["db"], table.args.get("catalog"))
                 )
 
-            df = pd.DataFrame.from_records(rows, columns=columns_to_types)
+            df = pd.DataFrame.from_records(rows, columns=columns_to_types)  # type: ignore
             self.engine_adapter.create_view(_test_fixture_name(table_name), df, columns_to_types)
 
     def tearDown(self) -> None:
@@ -243,7 +243,7 @@ class SqlModelTest(ModelTest):
         # Test model query
         query_rows = self.body["outputs"].get("query")
         if query_rows is not None:
-            expected_df = pd.DataFrame.from_records(query_rows, columns=self.model.columns_to_types)
+            expected_df = pd.DataFrame.from_records(query_rows, columns=self.model.columns_to_types)  # type: ignore
             actual_df = self._execute(query)
             self.assert_equal(expected_df, actual_df)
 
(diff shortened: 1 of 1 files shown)


tobymao (approved):

sqlmesh/core/test/definition.py
```diff
@@ -25,7 +25,6 @@ class TestError(SQLMeshError):
 
 class ModelTest(unittest.TestCase):
     __test__ = False
-    view_names: t.List[str] = []
```
georgesittas, line 28: Note: this is unused.

PR #1808, opened 2023-12-05
Fix: Skip null macro dependencies when extracting macro references from the dbt manifest

izeigerman:


crericha:
Should we do the same for ref and source dependencie?

izeigerman:
@crericha so far the problem was only observed with the `depends_on` manifest block. I'd rather not rush into adding workarounds in places where we don't need them.

tobymao (approved):

crericha (approved):

PR #1812, opened 2023-12-07
feature!: remove graphviz and use html for dag gen

tobymao:


tobymao:
Screenshot_2023-12-06_19 45 04

PR #1822, opened 2023-12-07
Feat: Detect circular references from dbt tests

crericha:
In the dbt workflow, tests are run separately from transformation, so it is okay to have circular test references. In SQLMesh, tests (aka audits) are run during transformation to prevent bad data from propagating and thus the test dependencies are included in dag construction. This PR provides a clearer error message on how to fix these circular test references, so that the project can use the SQLMesh workflow.

### DIFF
--- a/sqlmesh/dbt/basemodel.py
+++ b/sqlmesh/dbt/basemodel.py
@@ -218,6 +218,18 @@ class BaseModelConfig(GeneralConfig):
         return dependencies
 
     def check_for_circular_test_refs(self, context: DbtContext) -> None:
+        """
+        Checks for direct circular references between two models and raises an exception if found.
+        This addresses the most common circular reference seen when importing a dbt project -
+        relationship tests in both directions. In the future, we may want to increase coverage by
+        checking for indirect circular references.
+
+        Args:
+            context: The dbt context this model resides within.
+
+        Returns:
+            None
+        """
         for test in self.tests:
             for ref in test.dependencies.refs:
                 model = context.refs[ref]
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/basemodel.py
```diff
@@ -217,12 +217,33 @@ def tests_ref_source_dependencies(self) -> Dependencies:
...
+            for ref in test.dependencies.refs:
+                model = context.refs[ref]
+                if ref == self.name or ref in self.dependencies.refs:
+                    continue
+                elif self.name in model.dependencies.refs:
+                    raise ConfigError(
+                        f"Test '{test.name}' for model '{self.name}' depends on downstream model '{model.name}'."
+                        " Move the test to the downstream model to avoid circular references."
+                    )
+                elif self.name in model.tests_ref_source_dependencies.refs:
```
izeigerman, line 231: Does this mean that it only checks circular references 1 level deep? What about a graph like this:
```
A <- B <- C
```
and then model `A` has a relationship test with model `C`?

sqlmesh/dbt/basemodel.py
```diff
@@ -217,12 +217,33 @@ def tests_ref_source_dependencies(self) -> Dependencies:
         dependencies.macros = []
         return dependencies
 
+    def check_for_circular_test_refs(self, context: DbtContext) -> None:
+        for test in self.tests:
+            for ref in test.dependencies.refs:
+                model = context.refs[ref]
+                if ref == self.name or ref in self.dependencies.refs:
+                    continue
+                elif self.name in model.dependencies.refs:
```
izeigerman, line 226: Does this mean that it only checks circular references 1 level deep? What about a graph like this:
```
A <- B <- C
```
and then model A has a relationship test with model C?

izeigerman (approved):
Let's just add a comment, that we only check one level of references, so that the next person who reads this code knows that that was an intentional choice.

PR #1825, opened 2023-12-08
Fix!: workaround for weird date dtype bug exposed by duckdb v0.9.2

### LINKED ISSUE #1799: Test on model producing NULL date values fail with empty diff
# Issue description

Let's say we have an external model `model1` defined as follows:
```yaml
- name: dummy.model1
  columns:
    id: INT
    date: DATE
```
And a model `model2` using `model1` defined as follows:
```sql
MODEL (
  name dummy.model2,
  kind FULL,
);

SELECT
  *,
  DATE_TRUNC('month', date)::DATE AS month,
  NULL::DATE AS null_date,
FROM
  dummy.model1
```
And finally a unit test for `model2` defined as follows:
```yaml
test_null_dates:
  model: dummy.model2
  inputs:
    dummy.model1:
      rows:
        - id: 1234
          date: 2023-01-12
        - id: 9876
          date: 2023-02-10
  outputs:
    query:
      rows:
        - id: 1234
          date: 2023-01-12
          month: 2023-01-01
          null_date:
        - id: 9876
          date: 2023-02-10
          month: 2023-02-01
          null_date:
```

When running this test, I get the following error:
```bash
FAIL: test_null_dates (tests/test_model2.yaml)
AssertionError: Data differs (exp: expected, act: actual)

Empty DataFrame
Columns: []
Index: []

Ran 1 test in 0.036s

FAILED (failures=1)
```

# Analysis

In the `ModelTest.assert_equal()` method the issue seems to be related to [this line](https://github.com/TobikoData/sqlmesh/blob/573951086edf1c51366e8783e49dfb60fe24e914/sqlmesh/core/test/definition.py#L104).
This line updates the `actual` dataframe types, replace the `DATE` columns schemas.

Before this line:
```
actual.dtypes=('id', dtype('int32')) ('date', dtype('<M8[us]')) ('month', dtype('<M8[us]')) ('null_date', dtype('<M8[us]'))
```

After this lines:
```
actual.dtypes=('id', dtype('int32')) ('date', dtype('O')) ('month', dtype('O')) ('null_date', dtype('O'))
```

georgesittas:
Fixes #1799

Not quite sure if this change will have unintended side effects, but normalizing "non-existent" values to `np.nan` instead of `None` (previous approach) seemed just as reasonable to me.

@vchan the previous logic was introduced in https://github.com/TobikoData/sqlmesh/pull/1110, just tagging to make sure I'm not missing additional context here.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -100,8 +100,8 @@ class ModelTest(unittest.TestCase):
             actual_types, errors="ignore"
         )
 
-        expected = expected.replace({np.nan: None, "nan": None})
-        actual = actual.replace({np.nan: None, "nan": None})
+        expected = expected.replace({None: np.nan, "nan": np.nan})
+        actual = actual.replace({None: np.nan, "nan": np.nan})
 
         try:
             pd.testing.assert_frame_equal(
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -100,8 +100,8 @@ class ModelTest(unittest.TestCase):
             actual_types, errors="ignore"
         )
 
-        expected = expected.replace({None: np.nan, "nan": np.nan})
-        actual = actual.replace({None: np.nan, "nan": np.nan})
+        expected = expected.replace({None: np.nan})
+        actual = actual.replace({None: np.nan})
 
         try:
             pd.testing.assert_frame_equal(
--- a/tests/core/test_test.py
+++ b/tests/core/test_test.py
@@ -222,34 +222,9 @@ test_foo:
     assert result and result.wasSuccessful()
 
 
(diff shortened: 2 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -100,8 +100,8 @@ def assert_equal(self, expected: pd.DataFrame, actual: pd.DataFrame) -> None:
             actual_types, errors="ignore"
         )
 
-        expected = expected.replace({np.nan: None, "nan": None})
-        actual = actual.replace({np.nan: None, "nan": None})
+        expected = expected.replace({None: np.nan})
+        actual = actual.replace({None: np.nan})
```
tobymao, line 104: yea i think in general with pandas you want to use nan and not None

PR #1827, opened 2023-12-08
Chore: Add plan action class to control plan flow

mykalmax:
Previous update to plan flow has introduced some regression. It inaccurately display stages and messages (Button Labels). Also not showing changes to Model Metadata and when there are some backfills but no changes.
I added class to control current `plan-action` from single source.

### DIFF
--- a/web/client/src/library/components/plan/PlanActionsDescription.tsx
+++ b/web/client/src/library/components/plan/PlanActionsDescription.tsx
@@ -59,7 +59,7 @@ export default function PlanActionsDescription(): JSX.Element {
         )}
         {isFalse(isStringEmptyOrNil(restate_models)) && (
           <span className="inline-block mr-1">
-            and restate folowing models <b>{restate_models}</b>
+            and restate the followingmodels <b>{restate_models}</b>
           </span>
         )}
         {include_unmodified && (
--- a/web/client/src/library/components/plan/PlanActionsDescription.tsx
+++ b/web/client/src/library/components/plan/PlanActionsDescription.tsx
@@ -1,0 +1,71 @@
+import { useStoreContext } from '~/context/context'
+import { usePlan } from './context'
+import { isFalse, isStringEmptyOrNil } from '@utils/index'
+
+export default function PlanActionsDescription(): JSX.Element {
+  const {
+    start,
+    end,
+    skip_backfill,
+    no_gaps,
+    no_auto_categorization,
+    forward_only,
+    restate_models,
+    include_unmodified,
+  } = usePlan()
+
+  const environment = useStoreContext(s => s.environment)
+
+  return (
+    <div className="w-full flex px-4 py-2">
+      <p className="ml-2 text-xs">
+        <span>Plan for</span>
+        <b className="text-primary-500 font-bold mx-1">{environment.name}</b>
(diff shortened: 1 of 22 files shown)


web/client/src/library/components/plan/PlanActionsDescription.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </span>
+        )}
+        {no_auto_categorization && (
+          <span className="inline-block mr-1">
+            also set <b>Change Category</b> manually
+          </span>
+        )}
+        {isFalse(isStringEmptyOrNil(restate_models)) && (
+          <span className="inline-block mr-1">
+            and restate folowing models <b>{restate_models}</b>
```
vchan, line 62:
```suggestion
            and restate the following models <b>{restate_models}</b>
```

web/client/src/library/components/plan/PlanApplyStageTracker.tsx
```diff
@@ -90,41 +91,46 @@ export default function PlanApplyStageTracker(): JSX.Element {
           </small>
         </div>
       )}
-      {isFalseOrNil(hasChanges) || isNil(changes) ? (
+      {planAction.isRunning || (isNotNil(changes) && isTrue(hasChanges)) ? (
```
vchan, line 94: I know this isn't exactly new, but what's the difference been changes and hasChanges?

web/client/src/library/components/plan/PlanOptions.tsx
```diff
@@ -142,8 +138,16 @@ export default function PlanOptions({
                         info="Ensure that new snapshots have no data gaps when
               comparing to existing snapshots for matching
               models in the target environment"
-                        enabled={Boolean(no_gaps)}
-                        disabled={isInitialPlanRun}
+                        enabled={
```
vchan, line 141: Can InputToggle be enabled and disabled at the same time?

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -328,17 +327,30 @@ export default function PageIDE(): JSX.Element {
...
 
     setPlanApply(planApply)
   }
 
+  function updateEnviroments(data: Optional<Environments>): void {
+    if (
+      isNil(data) ||
+      isNil(data.environments) ||
+      isObjectEmpty(data) ||
+      isObjectEmpty(data.environments)
```
vchan, line 340: Can this be condensed to something like isObjectNotEmpty(data?.environments)?

vchan (approved):

web/client/src/library/components/plan/PlanOptions.tsx
mykalmax: `enabled` is referring to Toggle State but `disabled` is referring to input state, so `enabled` Toggle can be `disabled`. It is poor naming, i will rename it to something like `checked` in separate PR

web/client/src/library/components/plan/PlanApplyStageTracker.tsx
mykalmax: it is possible these two don't much, need to refactor it so `hasChanges` is just derived from `changes `

PR #1829, opened 2023-12-09
refactor: improve query performance

tobymao:


sqlmesh/core/dialect.py
```diff
@@ -824,30 +826,26 @@ def normalize_model_name(
...
-        if not isinstance(table.this, exp.Func) and exp.table_name(table) not in scope.cte_sources
-    }
+    if TABLES_META not in expression.meta:
+        expression.meta[TABLES_META ] = {
+            normalize_model_name(table, dialect=dialect)
+            for scope in traverse_scope(expression)
+            for table in scope.tables
+            if not isinstance(table.this, exp.Func) and exp.table_name(table) not in scope.cte_sources
+        }
+    return expression.meta[TABLES_META ]
```
izeigerman, line 848: Doesn't `black` catch this?

tobymao: you assume i ran black lol

sqlmesh/core/macros.py
```diff
@@ -177,32 +182,23 @@ def _transform_node(node: exp.Expression) -> exp.Expression:
...
-        transformed = evaluate_macros(expression)
+        transformed = evaluate_macros(expression.copy())
 
         if changed:
             # the transformations could have corrupted the ast, turning this into sql and reparsing ensures
             # that the ast is correct
             if isinstance(transformed, list):
-                return [self.parse_one(node.sql(dialect=self.dialect)) for node in transformed]
+                return [
+                    self.parse_one(node.sql(dialect=self.dialect, copy=False))
```
izeigerman, line 197: what does `copy=False` mean for the `sql` call?

sqlmesh/core/model/definition.py
```diff
@@ -1053,6 +1059,7 @@ def update_schema(
         )
         self._columns_to_types = None
         self._query_renderer._optimized_cache = {}
+        self._query_renderer.update_schema(self.mapping_schema)
```
izeigerman, line 1062: Was this a bug?

sqlmesh/core/macros.py
tobymao: when we generate sql using sqlglot.sql(), it by default does a copy for safety, sometimes write sql in a dialect has to do transformations. in this case, i know it's safe to copy=False here even though the sql generation may cause mutation.

sqlmesh/core/model/definition.py
izeigerman: Should 
```
self._query_renderer._optimized_cache = {}
```
happen inside the renderer's `update_schema` then?

tobymao: no, the code used to generate mapping schema every time you call render / macro evaluator etc. so it was sharing the mapping_schema dict and renormalizing the mapping schema. now i only do it once or when the schema is updated. it's cleaner this way anyways.

sqlmesh/core/renderer.py
```diff
@@ -166,7 +167,19 @@ def _render(
 
                 if expression:
                     with _normalize_and_quote(expression, self._dialect) as expression:
-                        pass
+                        if hasattr(expression, "selects"):
+                            for select in expression.selects:
+                                if not isinstance(select, exp.Alias) and select.output_name not in (
```
izeigerman, line 172: Shouldn't we do this check **after** we qualify the query in the `_optimize_query`?

sqlmesh/core/model/definition.py
izeigerman: ah, right. I didn't like the fact that we had to rely on the mutable nature of that dict. Good 👍

sqlmesh/core/renderer.py
tobymao: it used to be there, but there's no need for it. optimize_query, if successful already does exactly that.

PR #1832, opened 2023-12-12
refactor!: use simplified gen to make data hash more performant

tobymao:


izeigerman (approved):
Make sure to either bump sqlglot or add a migration file

PR #1835, opened 2023-12-12
fix!: add rust tokenizer support with new parser

tobymao:


sqlmesh/core/dialect.py
```diff
@@ -98,32 +98,6 @@ class StagedFilePath(exp.Table):
     """Represents paths to "staged files" in Snowflake."""
 
 
-def _scan_var(self: Tokenizer) -> None:
```
georgesittas, line 101: so we basically got rid of this and now it's handled at parse time?

tobymao: yea

sqlmesh/core/dialect.py
```diff
@@ -150,6 +124,52 @@ def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expressio
...
+    self: Parser,
+    any_token: bool = True,
+    tokens: t.Optional[t.Collection[TokenType]] = None,
+) -> t.Optional[exp.Expression]:
+
+    if self._match(TokenType.L_BRACE):
+        identifier = self.__parse_id_var(any_token=any_token, tokens=tokens)  # type: ignore
+        if not self._match(TokenType.R_BRACE):
+            self.raise_error("Expecting }")
+        identifier.args["this"] = f"@{{{identifier.args['this']}}}"
```
georgesittas, line 137:
```suggestion
        identifier.args["this"] = f"@{{{identifier.name}}}"
```

sqlmesh/core/dialect.py
```diff
@@ -150,6 +124,52 @@ def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expressio
...
+
+    while (
+        identifier
+        and self._is_connected()
+        and (
+            self._match_texts(("{", SQLMESH_MACRO_PREFIX), advance=False)
+            or self._curr.token_type not in self.RESERVED_TOKENS
+        )
+    ):
+        this = identifier.args["this"]
```
georgesittas, line 149:
```suggestion
        this = identifier.name
```

sqlmesh/core/dialect.py
```diff
@@ -150,6 +124,52 @@ def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expressio
...
+        brace = self._match(TokenType.L_BRACE)
+
+        if brace:
+            this += "{"
+
+        next_id = self._parse_id_var(any_token=False)
+
+        if next_id:
+            this += next_id.this
```
georgesittas, line 158: here too I guess?
```suggestion
            this += next_id.name
```

sqlmesh/core/dialect.py
```diff
@@ -150,6 +124,52 @@ def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expressio
...
+        else:
+            return identifier
+
+        if brace:
+            if self._match(TokenType.R_BRACE):
+                this += "}"
+            else:
+                self.raise_error("Expecting }")
+
```
georgesittas, line 167: Note: maybe out of scope for now, but I believe we have this pattern of "assert we match something" in sqlglot's parser as well -- perhaps we could factor it out using an extra flag in `_match`?

```python
self._match(TokenType.R_BRACE, ensure_matched=True)
```

georgesittas (approved):

PR #1836, opened 2023-12-13
Fix!: Include macro dependencies missed by dbt manifest parser

crericha:


sqlmesh/dbt/manifest.py
```diff
@@ -347,15 +350,15 @@ def _flatten_dependencies_from_macros(
...
+    def _extra_dependencies(self, target: str, package: str) -> Dependencies:
         # We sometimes observe that the manifest doesn't capture all macros, refs, and sources within a macro.
         # This behavior has been observed with macros like dbt.current_timestamp(), dbt_utils.slugify(), and source().
         # Here we apply our custom extractor to make a best effort to supplement references captured in the manifest.
         dependencies = Dependencies()
         for call_name, node in extract_call_names(target):
-            if len(call_name) == 2 and call_name[0] in ("dbt", "dbt_utils"):
-                dependencies.macros.append(MacroReference(package=call_name[0], name=call_name[1]))
-            elif call_name[0] == "source":
+            if call_name[0] in ("config"):
```
izeigerman, line 359:
```suggestion
            if call_name[0] == "config":
```

sqlmesh/dbt/manifest.py
```diff
@@ -371,6 +374,24 @@ def _extra_dependencies(self, target: str) -> Dependencies:
...
+                    f"macro.{package}.{macro_name}" not in self._manifest.macros
+                    and f"macro.dbt.{macro_name}" in self._manifest.macros
+                ):
+                    package_name: t.Optional[str] = "dbt"
+                else:
+                    # dbt doesn't include the package name for project macros
+                    package_name = package if package != self._project_name else None
+                dependencies.macros.append(MacroReference(package=package_name, name=macro_name))
+            else:
+                if call_name[0] not in ("adapter"):
```
izeigerman, line 391:
```suggestion
                if call_name[0] == "adapter":
```

sqlmesh/dbt/manifest.py
```diff
@@ -371,6 +374,24 @@ def _extra_dependencies(self, target: str) -> Dependencies:
                 args = [_jinja_call_arg_name(arg) for arg in node.args]
                 if args and args[0]:
                     dependencies.variables.add(args[0])
+            elif len(call_name) == 1:
+                macro_name = call_name[0]
+                if macro_name in {**BUILTIN_GLOBALS, **BUILTIN_FILTERS}:
```
izeigerman, line 379: This dict will be created for each macro.

crericha: When it's a very small dict (~20), does it impact performance? Originally I checked in both separately, but thought this was more readable.

tobymao: make a new constant and use that here

tobymao: depending on how many times this is called, it could have a significant impact on performance

izeigerman (approved):

PR #1843, opened 2023-12-17
Fix: Do not load environments to get their naming info

vchan:
Very slow, especially if there are lots of models.


tobymao (approved):

web/server/api/endpoints/plan.py
```diff
@@ -140,7 +140,7 @@ def _get_plan_backfills(context: Context, plan: Plan) -> t.Dict[str, t.Any]:
                 model_name=interval.snapshot_name,
                 view_name=plan.context_diff.snapshots[
                     interval.snapshot_name
-                ].qualified_view_name.for_environment(plan.environment.naming_info)
+                ].qualified_view_name.for_environment(plan.environment_naming_info)
```
tobymao, line 143: why does this still take 30 seconds?

PR #1844, opened 2023-12-18
Chore: Display ellipsis for long model names

mykalmax:


web/client/src/library/components/documentation/Documentation.tsx
```diff
@@ -258,7 +261,12 @@ function DetailsItem<TValue = Record<string, Primitive>>({
     >
       {isArrayNotEmpty<TValue>(value) ? (
         <>
-          <strong className="mr-2 text-xs capitalize">{name}</strong>
+          <strong
+            title={title}
```
vchan, line 265: What are these titles used for?

mykalmax: when hovering over element, browser displays default tooltip with full name in this case

vchan (approved):

PR #1846, opened 2023-12-18
Chore: Log and send SSE updates after each stage completion for plan_overview

mykalmax:


web/server/settings.py
```diff
@@ -57,7 +57,6 @@ def _get_path_mappings(context: Context) -> dict[Path, FileType]:
 
 @lru_cache()
 def _get_loaded_context(path: str | Path, config: str, gateway: str) -> Context:
-    print(config)
```
vchan, line 60: Oops.

web/client/src/models/plan-action.ts
```diff
@@ -111,10 +111,14 @@ export class ModelPlanAction<
...
-  get displayStatus(): string {
+  displayStatus(planOverview: ModelPlanOverviewTracker): string {
     if (this.isRunningTask) return 'Running Task...'
     if (this.isApplying) return 'Applying Plan...'
-    if (this.isRunning) return 'Getting Changes...'
+    if (this.isRunning && isNil(planOverview.hasChanges))
+      return 'Getting Changes...'
+    if (this.isRunning && isNil(planOverview.hasBackfills))
+      return 'Getting Backfills...'
+    if (this.isRunning) return 'Checking Plan...'
```
vchan, line 121: What does Checking Plan mean?

vchan (approved):

web/client/src/models/plan-action.ts
mykalmax: so it is possible to have a delay between state change .

PR #1848, opened 2023-12-18
fix!: improve perf of model loading by reducing the scope of type ann…

tobymao:
…otations to just the select

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -15,7 +15,6 @@ from pydantic import Field
 from sqlglot import diff, exp
 from sqlglot.diff import Insert, Keep
 from sqlglot.helper import ensure_list
-from sqlglot.optimizer.annotate_types import annotate_types
 from sqlglot.optimizer.simplify import gen
 from sqlglot.schema import MappingSchema, nested_set
 from sqlglot.time import format_time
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1009,7 +1009,10 @@ def columns_to_types(self) -> t.Optional[t.Dict[str, exp.DataType]]:
 
             self._columns_to_types = {
                 select.output_name: select.type or exp.DataType.build("unknown")
-                for select in annotate_types(query, schema=self._query_renderer.schema).selects
+                for select in (
+                    annotate_types(select, schema=self._query_renderer.schema)
+                    for select in query.selects
+                )
```
georgesittas, line 1015: I'm not sure if you can do this, don't you need the tables in scope to fetch the corresponding column types?

georgesittas: Refer to this section https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/annotate_types.py#L339

tobymao: we're only doing types for selects

georgesittas: Yeah but I mean I'm not sure you can isolate the `annotate_types` call to just each select. For example if you don't have a proper query I think the scope creation will not work as expected and you won't annotate types?

PR #1855, opened 2023-12-19
Chore: Support verbose CLI output for the 'plan' command

izeigerman:


eakmanrq (commented):
I thought you were making the argument this should be default? If I passed in a verbose flag I would generally expect a bit more than just adding creation/promotion. I could see us later wanting to add more and then consider these current items as normal and not verbose anymore.

PR #1856, opened 2023-12-19
Docs: add custom measures to observer guide

treysp:


docs/guides/observer.md
```diff
@@ -206,4 +208,99 @@ Next, the Loaded Intervals section displays the time intervals that have been lo
...
+
+### Definition
+
+Each custom measure is associated with a model and is defined by a SQL query in the model file.
+
+A measure query is specified in the `@measure` macro after the primary model query. The measure query will be executed during a SQLMesh `plan` or `run` after the primary model query is executed.
+
+A measure query may return multiple columns, each of which will be stored as individual measures. Multiple measure queries may be specified in a model.
+
+This example shows a model definition that includes a measure query calculating the total number of rows in the table and the average value of the model's `numeric_col` column:
```
crericha, line 225: that includes a measure query, `which defines two measures: row_count (the total number of rows in the table) and num_col_avg (the average value of the model's numeric_col column).`

docs/guides/observer.md
```diff
@@ -206,4 +208,99 @@ Next, the Loaded Intervals section displays the time intervals that have been lo
...
+    SELECT
+        COUNT(*) AS row_count, -- Table's row count
+        AVG(numeric_col) AS num_col_avg -- Average value of `numeric_col`
+    FROM custom_measure.example -- Select FROM the name of the model
+);
+```
+
+Every time the `custom_measure.example` model is executed, Observer will execute the measure query and store the value it returns.
+
+Observer will automatically associate the value with the execution time of the `plan`/`run` that calculated it and will record each measure under its column name. This example will create two custom measures: `row_count` and `num_col_avg`.
```
crericha, line 248: Suggest "By default the measure's timestamp will be the execution time of the `plan`/`run` that captured the measure."

docs/guides/observer.md
```diff
@@ -206,4 +208,99 @@ Next, the Loaded Intervals section displays the time intervals that have been lo
...
+
+Observer charts will display the value of the two measures on the y-axis and the execution time of the associated `plan`/`run` on the x-axis, allowing you to monitor whether the value has meaningfully changed since the previous execution.
```
crericha, line 250: "of the two measures" is confusing to me. Can we say "of the selected measure"

crericha (approved):
I made a few wording suggestions that sound clearer to me personally. I'll defer to your judgement on whether to take them or not.

docs/guides/observer.md
```diff
@@ -206,4 +208,99 @@ Next, the Loaded Intervals section displays the time intervals that have been lo
...
+
+## Custom measures
+
+SQLMesh Observer allows you to calculate and track custom measures in addition to the ones it [automatically calculates](#data).
+
+### Definition
+
+Each custom measure is associated with a model and is defined by a SQL query in the model file.
+
+A measure query is specified in the `@measure` macro after the primary model query. The measure query will be executed during a SQLMesh `plan` or `run` after the primary model query is executed.
```
crericha, line 221: The `@measure` macro is used to define custom measures within a model. The body of the @measure macro is the a query, that contains the measures to capture as columns. The column name will be used as the measure name.

crericha:
The same measure name can be used across models. Should we make that clear?

PR #1863, opened 2023-12-20
Feat: implement a cache for dbt's adapter.list_relations method

georgesittas:
Still a WIP, looking to get some early feedback.

- [ ] Add a config flag that controls whether or not we'll use this cache at runtime so that there's a path to switch this off in case it turns out the implementation is unsafe and / or needs to be refactored (we'll use the cache by default).
- [ ] Add some more tests for cache clearing, updating etc

### DIFF
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -339,7 +339,7 @@ class RuntimeAdapter(BaseAdapter):
 
     def _table_to_relation(self, table: exp.Table) -> BaseRelation:
         return self.relation_type.create(
-            database=table.catalog or None,
+            database=table.catalog or self.engine_adapter.default_catalog,
             schema=table.db,
             identifier=table.name,
             quote_policy=self.quote_policy,
@@ -406,7 +406,9 @@ class CachingEngineAdapter:
         table = exp.to_table(relation_name)
         schema = to_schema(table)
         relation = self.relation_type.create(
-            database=table.catalog or None, schema=table.db, identifier=table.name
+            database=table.catalog or self.engine_adapter.default_catalog,
+            schema=table.db,
+            identifier=table.name,
         )
 
         relations = self.cache.get(schema)
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -8,7 +8,7 @@ import pandas as pd
 from dbt.contracts.relation import Policy
 from sqlglot import exp, parse_one
 
-from sqlmesh.core.dialect import normalize_model_name
+from sqlmesh.core.dialect import normalize_model_name, to_schema
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.renderer import _normalize_and_quote
 from sqlmesh.core.snapshot import DeployabilityIndex, Snapshot, to_table_mapping
@@ -20,6 +20,8 @@ if t.TYPE_CHECKING:
     from dbt.adapters.base import BaseRelation
     from dbt.adapters.base.column import Column
     from dbt.adapters.base.impl import AdapterResponse
+
+    from sqlmesh.core._typing import TableName
 
 
 logger = logging.getLogger(__name__)
@@ -47,10 +49,7 @@ class BaseAdapter(abc.ABC):
 
     @abc.abstractmethod
     def list_relations(self, database: t.Optional[str], schema: str) -> t.List[BaseRelation]:
-        """Gets all relations in a given schema and optionally database.
-
-        TODO: Add caching functionality to avoid repeat visits to DB
-        """
+        """Gets all relations in a given schema and optionally database."""
 
     @abc.abstractmethod
     def list_relations_without_caching(self, schema_relation: BaseRelation) -> t.List[BaseRelation]:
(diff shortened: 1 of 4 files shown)


sqlmesh/dbt/adapter.py
```diff
@@ -355,3 +362,54 @@ def _normalize(self, input_table: exp.Table) -> exp.Table:
             normalized_table.set("db", normalized_table.this)
             normalized_table.set("this", None)
         return normalized_table
+
+
+class CachingAdapter:
```
izeigerman, line 367: I'm a little confused: why do we have cache here but also in the RuntimeAdapter? Do we even use this cache right now or is this still WIP?

izeigerman: Oh, I suppose this should be `CachingEngineAdapter`, not dbt adapter.

sqlmesh/dbt/adapter.py
```diff
@@ -355,3 +362,54 @@ def _normalize(self, input_table: exp.Table) -> exp.Table:
...
+
+    def create_view(self, view_name: TableName, *args: t.Any, **kwargs: t.Any) -> None:
+        self.engine_adapter.create_view(view_name, *args, **kwargs)
+        self._cache_relation(view_name)
+
+    def _cache_relation(self, relation_name: TableName) -> None:
+        table = exp.to_table(relation_name)
+        schema = to_schema(table)
+        relation = self.relation_type.create(
+            database=table.catalog or None, schema=table.db, identifier=table.name
```
izeigerman, line 407: Should it be `default_catalog` instead of `None`?

sqlmesh/dbt/adapter.py
```diff
@@ -219,7 +219,14 @@ def list_relations(self, database: t.Optional[str], schema: str) -> t.List[BaseR
         reference_relation = self.relation_type.create(
             database=database, schema=schema, quote_policy=self.quote_policy
         )
-        return self.list_relations_without_caching(reference_relation)
+        normalized_schema = self._normalize(self._schema(reference_relation))
+
+        relations = self.cache.get(normalized_schema)
```
izeigerman, line 224: I might suggest making the `CachingAdapter` an owner of the `cache` and expose an API to get it

sqlmesh/dbt/adapter.py
```diff
@@ -219,7 +219,14 @@ def list_relations(self, database: t.Optional[str], schema: str) -> t.List[BaseR
...
+        if relations is None:
+            relations = self.list_relations_without_caching(reference_relation)
+            self.cache[normalized_schema] = relations
```
izeigerman, line 227: Should we update the cache directly here, or override `_get_data_objects` in the `CachingAdapter` and cache it there (but not return from the cache so that `list_relations_without_caching` still works as expected) ?

sqlmesh/dbt/adapter.py
georgesittas: Ah, I forgot to update this part it seems 🤦 

My intention was to do `self.engine_adapter = CachingEngineAdapter(...)` in the `RuntimeAdapter`'s init method.

sqlmesh/dbt/adapter.py
georgesittas: Not sure - I was trying to follow what is done in https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/dbt/adapter.py#L334 but we can double-check this

sqlmesh/dbt/adapter.py
georgesittas: That's right! This was my intention all along, I forgot to update the constructor

sqlmesh/dbt/adapter.py
georgesittas: Updated in https://github.com/TobikoData/sqlmesh/pull/1863/commits/68fdc54372c59e7b27601112df0ed6c153a6d9b2, please take another look and let me know if something looks off.

sqlmesh/dbt/adapter.py
georgesittas: I will probably update the logic to move `self.cache` in the caching adapter.

sqlmesh/dbt/adapter.py
georgesittas: Perhaps we can do it there as well yeah, let me try.

georgesittas: On the other hand that will add caching in `list_relations_without_caching`, which contradicts the method name?

izeigerman: That's why I suggested to only storing in the cache but not serving from it in `_get_data_objects`:

> (but not return from the cache so that list_relations_without_caching still works as expected)

georgesittas: Ah I see, makes sense, let me look what that'd look like 👍

georgesittas: Okay, I took a quick look and I'm not really sure about the value of this refactor - it introduces some complexity:

1. need to override `_get_data_objects`,
2. should probably replicate or factor out [logic](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/dbt/adapter.py#L229) that exists in `list_relations_without_caching`, so that we can store the appropriate relation objects in the cache.

Also, `_get_data_objects` is used infrequently at the moment, so it looks like we wouldn't get much value back from doing this. The current approach feels a little less involved. Thoughts?

georgesittas:
Closing this to reduce the PR noise as it's become stale and will open another one soon when I make some progress.

PR #1865, opened 2023-12-20
Chore: Clean up plan action

mykalmax:


web/client/src/library/components/plan/PlanActions.tsx
```diff
@@ -91,7 +91,6 @@ export default function PlanActions({
                   EnumPlanAction.ApplyBackfill,
                   EnumPlanAction.ApplyVirtual,
                   EnumPlanAction.ApplyChangesAndBackfill,
-                  EnumPlanAction.ApplyChanges,
```
mykalmax, line 94: removing `ApplyChanges`  sinse it is a virtual update

vchan (approved):

PR #1867, opened 2023-12-20
Chore: Return 204 on empty reuqest

mykalmax:
when returning `None` from endpoint with `200` response has `null` in it. So returning 204  is more predictable

### DIFF
--- a/web/server/api/endpoints/commands.py
+++ b/web/server/api/endpoints/commands.py
@@ -5,7 +5,8 @@ import io
 import typing as t
 
 import pandas as pd
-from fastapi import APIRouter, Body, Depends, Request
+from fastapi import APIRouter, Body, Depends, Request, Response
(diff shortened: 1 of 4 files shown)


vchan (approved):

web/server/api/endpoints/commands.py
```diff
@@ -28,6 +29,7 @@
 @router.post("/apply", response_model=t.Optional[models.PlanApplyStageTracker])
```
vchan, line 29: I haven't tested it, but I think you might be able to do the following:
```suggestion
@router.post("/apply", response_model=t.Optional[models.PlanApplyStageTracker], status_code=HTTP_204_NO_CONTENT)
```

mykalmax: giving me error `Status code 204 must not have a response body`

PR #1878, opened 2023-12-21
Feat: Register table and column comments with engine

treysp:
This PR allows SQLMesh to automatically register SQL model comments with the underlying SQL engine. It supports both table and column comments.

### Configuration
Comment registration is enabled via the `register_comments` config key in the connection config (default `true` for all engines but Snowflake). Even when enabled, comments are never registered for temp tables or views in non-prod environments.

### Specification
Table comments are specified before a model's DDL or as the `description` key in the DDL (`description` used if both are present). Column comments are specified in-line in the model's query (comments on the same line as a column definition).

### Engine support
Engines vary widely in their support and methods for registering comments. Some engines don't support them at all (e.g., DuckDB), some support physical tables only (e.g., Redshift), and some support physical tables and views. This PR includes a table of engine support in the model overview concepts document.

### Registration methods
There are two methods of registering comments: embedding both table and column comments in an object's `CREATE` call or executing registration commands after the object has been created. 

Most engines support both methods, but some support only post-creation commands (e.g., postgres). Engine support can vary based on whether the object is a physical table or view.

Typically, post-creation commands require a separate call for each comment, so we avoid them everywhere possible. However, `CREATE` calls almost always require column names and types, so we must use post-creation commands if columns_to_types is unavailable.

One special case is views in the production environment. They are universally defined as `SELECT *` to allow the columns in the underlying physical table to change, so we can never specify column names and types. Some cloud engines support multiple comments in a single post-creation command using their Python API. This PR implements that method for BigQuery, and a future PR will implement it for Snowflake (if possible).


sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +489,82 @@ def _create_table_from_columns(
...
         )
-        schema = exp.Schema(
-            this=table,
-            expressions=[
-                exp.ColumnDef(this=exp.to_identifier(column), kind=kind)
-                for column, kind in columns_to_types.items()
-            ]
-            + primary_key_expression,
+
+        column_descriptions = kwargs.pop("column_descriptions", None)
```
izeigerman, line 493: We should probably make `column_descriptions` and `table_description` actual named arguments of the `_create_table_from_columns` method.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +489,82 @@ def _create_table_from_columns(
...
+        )
+
+        table_description = kwargs.pop("table_description", None)
+        self._create_table(
+            schema,
+            None,
+            exists=exists,
+            columns_to_types=columns_to_types,
+            table_description=table_description
+            if self.COMMENT_CREATION == CommentCreation.IN_SCHEMA_DEF
```
izeigerman, line 508: I'd probably do this check closer to the source to make sure we don't mess this logic up in other places where we call `_create_table`. So let's move this into `_create_table`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +489,82 @@ def _create_table_from_columns(
...
-                exp.ColumnDef(this=exp.to_identifier(column), kind=kind)
-                for column, kind in columns_to_types.items()
-            ]
-            + primary_key_expression,
+
+        column_descriptions = kwargs.pop("column_descriptions", None)
+        schema = self._build_schema_exp(
+            table,
+            columns_to_types,
+            column_descriptions if self.COMMENT_CREATION == CommentCreation.IN_SCHEMA_DEF else None,
```
izeigerman, line 497: Ditto, probably worth keeping it closer to the source.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +489,82 @@ def _create_table_from_columns(
...
+                    exp.ColumnDef(
+                        this=exp.to_identifier(column),
+                        kind=kind,
+                        constraints=self._build_col_comment_exp(column, column_descriptions),
+                    )
+                    for column, kind in columns_to_types.items()
+                ]
+                + expressions,
+            )
+        else:
```
izeigerman, line 547: [Nit] Else is not needed here since we return in both branches.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -501,23 +576,47 @@ def _create_table_from_source_queries(
         **kwargs: t.Any,
     ) -> None:
         table = exp.to_table(table_name)
+
+        # Build a schema expression with column comments if the engine supports it
+        table_description = kwargs.pop("table_description", None)
```
izeigerman, line 581: Ditto probably worth having named arguments for these

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +489,82 @@ def _create_table_from_columns(
...
+        Build a schema expression for a table, columns, column comments, and additional schema properties.
+        """
+        if column_descriptions:
+            return exp.Schema(
+                this=table,
+                expressions=[
+                    exp.ColumnDef(
+                        this=exp.to_identifier(column),
+                        kind=kind,
+                        constraints=self._build_col_comment_exp(column, column_descriptions),
```
izeigerman, line 541: Is this the only thing that differs between branches? Can we just have a single branch instead and just do:
```
constraints=self._build_col_comment_exp(column, column_descriptions) if column_descriptions else None
```
?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -501,23 +576,47 @@ def _create_table_from_source_queries(
...
                     if i == 0:
                         self._create_table(
-                            table,
+                            schema if schema else table,
                             query,
                             columns_to_types=columns_to_types,
                             exists=exists,
                             replace=replace,
+                            table_description=table_description
+                            if self.COMMENT_CREATION == CommentCreation.IN_SCHEMA_DEF
```
izeigerman, line 600: This is another reason to put this closer to the source - to avoid duplication like this.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -700,8 +799,12 @@ def create_view(
             if not self.SUPPORTS_MATERIALIZED_VIEW_SCHEMA and isinstance(schema, exp.Schema):
                 schema = schema.this
 
-        create_view_properties = self._create_view_properties(
-            create_kwargs.pop("table_properties", None)
+        table_description = create_kwargs.pop("table_description", None)
```
izeigerman, line 802: [Ditto] named argument

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1469,6 +1582,11 @@ def temp_table(
         source_queries, columns_to_types = self._get_source_queries_and_columns_to_types(
             query_or_df, columns_to_types=columns_to_types, target_table=name
         )
+
+        # do not add comments to temp tables
+        kwargs.pop("table_description", None)
```
izeigerman, line 1587: Btw, you won't have to do this if both of these were named arguments everywhere

sqlmesh/core/engine_adapter/base.py
```diff
@@ -112,6 +117,12 @@ def is_supported(self) -> bool:
         return self.is_requires_set_catalog or self.is_full_support
 
 
+class CommentCreation(Enum):
+    UNSUPPORTED = 1
+    IN_SCHEMA_DEF = 2
+    COMMENT_COMMAND_ONLY = 3
```
eakmanrq, line 123: Can you move this to `shared.py`? We have `DataObjectType` enum in there which is shared across these adapters. Also can you follow it's implementation of implementing `str, Enum` and then creating the `is_*` properties. I like doing this since it can help avoid doing an import and creating a circular dependency when I just need to check an Enum value is something I would generally expect.

eakmanrq:
I know this might be asking a lot, so happy to talk about it, but I think it would be great if there were tests in the [test_integration.py](github.com/tobikodata/sqlmesh/blob/a6d8d90bec8804f3bcc8dbd2ea99f343b49e19ac/tests/core/engine_adapter/test_integration.py#l24-l23) that would test creating an object with comments and then describe that table to verify they are actually there. If I'm reading this PR correctly, it would test creating comments on tables and views (if supported) for each engine.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +529,70 @@ def _create_table_from_columns(
...
+                table_name,
+                table_description,
+                column_descriptions,
+            )
+
+    def _build_schema_exp(
+        self,
+        table: exp.Table,
+        columns_to_types: t.Dict[str, exp.DataType],
+        column_descriptions: t.Optional[t.Dict[str, str]],
```
georgesittas, line 563: I think we also need to initialize this with `None` since it's optional

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +529,70 @@ def _create_table_from_columns(
...
                 for column, kind in columns_to_types.items()
             ]
-            + primary_key_expression,
+            + expressions,
         )
-        self._create_table(schema, None, exists=exists, columns_to_types=columns_to_types, **kwargs)
+
+    def _build_col_comment_exp(
+        self, col_name: str, column_descriptions: t.Dict[str, str]
+    ) -> t.List[exp.ColumnConstraint | None]:
```
georgesittas, line 587: Doesn't look like we can return a list that contains `None` here ever

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -469,8 +470,55 @@ def _create_table_properties(
         if clustered_by:
             properties.append(exp.Cluster(expressions=[exp.column(col) for col in clustered_by]))
 
+        if table_description:
+            properties.append(
+                exp.Property(
+                    this=exp.Identifier(this="description", quoted=False),
```
georgesittas, line 476: I think this needs to be `exp.var` instead of an identifier?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -469,8 +470,55 @@ def _create_table_properties(
...
+        self, col_name: str, column_descriptions: t.Dict[str, str]
+    ) -> t.List[exp.ColumnConstraint | None]:
+        comment = column_descriptions.get(col_name, None)
+        if comment:
+            return [
+                exp.ColumnConstraint(
+                    kind=exp.Properties(
+                        expressions=[
+                            exp.Property(
+                                this=exp.Identifier(this="description", quoted=False),
```
georgesittas, line 498: Ditto about the description. Should we just factor this part out into a helper method?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -469,8 +470,55 @@ def _create_table_properties(
...
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+        table_description: t.Optional[str] = None,
+    ) -> t.Optional[exp.Properties]:
+        """Creates a SQLGlot table properties expression for view"""
+        properties: t.List[exp.Expression] = []
+
+        if table_description:
+            properties.append(
+                exp.Property(
+                    this=exp.Identifier(this="description", quoted=False),
```
georgesittas, line 518: Ditto

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -469,8 +470,55 @@ def _create_table_properties(
...
+                    value=exp.Literal.string(table_description),
+                )
+            )
+
         for key, value in (table_properties or {}).items():
-            properties.append(exp.Property(this=key, value=value))
+            properties.append(exp.Property(this=key, value=value.copy()))
+
```
georgesittas, line 483: We may be able to reuse https://github.com/TobikoData/sqlmesh/blob/c0cf31f821ae4768ed18f7d5b79d204ebbc8492b/sqlmesh/core/engine_adapter/mixins.py#L216 here

tests/core/test_snapshot_evaluator.py
```diff
@@ -156,16 +156,19 @@ def x(evaluator, y=None) -> None:
         partition_interval_unit=IntervalUnit.DAY,
         clustered_by=[],
         table_properties={},
+        table_description=None,
```
georgesittas, line 159: Why do we need to specify `None` here? Is this something that can be set by default?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1567,6 +1739,35 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+                    kind=table_kind,
+                    expression=exp.Literal.string(table_comment),
+                )
+            )
+
+        if column_comments:
+            for col, comment in column_comments.items():
+                self.execute(
+                    exp.Comment(
+                        this=exp.to_column(f"{table_name}.{col}"),
```
georgesittas, line 1765: This looks a bit iffy, if `table_name` is a `Table` expression then this silently invokes `.sql` on it which can produce unexpected SQL if there are quotes

treysp: Yes - thanks for raising this. What is the correct/safe way to build a comment from table and column names in this context?

treysp:
> I know this might be asking a lot, so happy to talk about it, but I think it would be great if there were tests in the [test_integration.py](github.com/tobikodata/sqlmesh/blob/a6d8d90bec8804f3bcc8dbd2ea99f343b49e19ac/tests/core/engine_adapter/test_integration.py#l24-l23) that would test creating an object with comments and then describe that table to verify they are actually there. If I'm reading this PR correctly, it would test creating comments on tables and views (if supported) for each engine.
> 
> Also might want to verify on the `test_sushi` test that the comments defined on models are actually created on the target engine.

@eakmanrq I have most of the necessary components written from my own testing, so shouldn't be too big a lift.

sqlmesh/core/engine_adapter/base.py
treysp: Since we loop through all the columns on 580, this gets called for columns even if they don't have a comment in `column_descriptions` (so will return `[]` here).

Or am I misunderstanding?

sqlmesh/core/engine_adapter/bigquery.py
treysp: I was mimicking that sqlglot returns for a parsed statement - should I have identified the correct expression via another mechanism?

```
❯ python
Python 3.11.6 (main, Nov  2 2023, 04:39:43) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> parse_one("CREATE TABLE IF NOT EXISTS `test_table` (`a` int OPTIONS (description='a description'), `b` int) OPTIONS (description='test description')", dialect="bigquery")
(CREATE this: 
  (SCHEMA this: 
    (TABLE this: 
      (IDENTIFIER this: test_table, quoted: True)), expressions: 
    (COLUMNDEF this: 
      (IDENTIFIER this: a, quoted: True), kind: 
      (DATATYPE this: Type.INT, nested: False, prefix: False), constraints: 
      (COLUMNCONSTRAINT kind: 
        (PROPERTIES expressions: 
          (PROPERTY this: 
            (IDENTIFIER this: description, quoted: False), value: 
            (LITERAL this: a description, is_string: True))))), 
    (COLUMNDEF this: 
      (IDENTIFIER this: b, quoted: True), kind: 
      (DATATYPE this: Type.INT, nested: False, prefix: False))), kind: TABLE, exists: True, properties: 
  (PROPERTIES expressions: 
    (PROPERTY this: 
      (IDENTIFIER this: description, quoted: False), value: 
      (LITERAL this: test description, is_string: True))))
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -481,15 +529,70 @@ def _create_table_from_columns(
...
+
+        # Register comments with commands if the engine doesn't support comments in the schema or CREATE
+        if (
+            table_description or column_descriptions
+        ) and self.COMMENT_CREATION == CommentCreation.COMMENT_COMMAND_ONLY:
+            self._create_comments(
+                table_name,
+                table_description,
+                column_descriptions,
+            )
```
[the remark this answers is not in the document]
treysp: Since `_create_comments` is already a separate method, maybe we can just push the if logic down into the method and always call it here?

sqlmesh/core/engine_adapter/bigquery.py
treysp: @eakmanrq If we pull this out of the mixin so it can be reused, where should it go?

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: I see - no your reasoning makes sense, maybe the way this is parsed is not fully correct as well. You can ignore my comment for the time being and I'll get back to you on this once I dig a bit deeper.

sqlmesh/core/engine_adapter/bigquery.py
eakmanrq: You could move it to base.

sqlmesh/core/engine_adapter/base.py
tobymao: this method can never return a None, look at line 591

treysp: Ah, I was misunderstanding - `t.List[exp.ColumnConstraint]` encompasses `[]` and doesn't need the `| None`

sqlmesh/core/engine_adapter/base.py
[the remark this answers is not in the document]
georgesittas: Yep could do that too 👍

sqlmesh/core/engine_adapter/base.py
georgesittas: You could create a table expression first like on L1755 and then construct the column expression by manually setting its `table`, `db` and `catalog` args to the corresponding args of the table.

@tobymao do we already have a helper for this in sqlmesh? I don't recall having something in sqlglot.

tobymao: what exactly are you trying to do?

treysp: Build an expression that renders like this with the right identifier quoting: `COMMENT ON COLUMN my_table.my_column IS 'column comment'`

tobymao: do this at the top

table = exp.to_table(...)

then you can do

exp.column(col, *reverse(table.parts))

sqlmesh/core/engine_adapter/bigquery.py
tobymao, line 476: all of these should be exp.identifier(..., quoted=True)

tests/core/test_snapshot_evaluator.py
izeigerman: I think this is because it gets passed explicitly in the chain of calls

sqlmesh/core/engine_adapter/mysql.py
```diff
@@ -73,3 +80,52 @@ def _get_data_objects(self, schema_name: SchemaName) -> t.List[DataObject]:
...
+        """
+        table = exp.to_table(table_name)
+        table_sql = table.sql(dialect=self.dialect, identify=True)
+
+        if table_comment:
+            try:
+                self.execute(
+                    f"ALTER TABLE {table_sql} COMMENT = '{table_comment}'",
+                )
+            except:
```
izeigerman, line 102: I think generally it's not a good idea to catch all possible exceptions. Can we at least do `except Exception`?

sqlmesh/core/engine_adapter/mysql.py
```diff
@@ -73,3 +80,52 @@ def _get_data_objects(self, schema_name: SchemaName) -> t.List[DataObject]:
...
+                for col in column_comments:
+                    col_def = col_def_exps.get(col)
+                    col_def.args["constraints"].extend(  # type: ignore
+                        self._build_col_comment_exp(col_def.alias_or_name, column_comments)  # type: ignore
+                    )
+                    self.execute(
+                        f"ALTER TABLE {table_sql} MODIFY {col_def.sql(dialect=self.dialect, identify=True)}",  # type: ignore
+                    )
+            except:
+                logger.warning(
```
izeigerman, line 129: ditto

sqlmesh/core/engine_adapter/base.py
```diff
@@ -425,61 +517,98 @@ def _create_table_from_source_queries(
...
                         )
 
+        # Register comments with commands if the engine doesn't support comments in the schema
+        if (
+            (table_description or column_descriptions)
+            and self.COMMENT_CREATION.is_comment_command_only
+        ) or (
+            column_descriptions
+            and self.COMMENT_CREATION.is_in_schema_def
+            and (not self.SUPPORTS_CTAS_SCHEMA_COMMENTS or not columns_to_types)
```
izeigerman, line 561: I'm really confused by this one. If `SUPPORTS_CTAS_SCHEMA_COMMENTS` is True but `columns_to_types` are missing why do we still need a separate statement to create comments?

Overall this condition is really hairy and I suggest finding a way to rewrite it.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -425,61 +517,98 @@ def _create_table_from_source_queries(
...
 
-    def _create_table_exp(
+    def _build_create_table_exp(
         self,
         table_name_or_schema: t.Union[exp.Schema, TableName],
         expression: t.Optional[exp.Expression],
         exists: bool = True,
         replace: bool = False,
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        table_description: t.Optional[str] = None,
```
izeigerman, line 601: Seems like in case of this function, having a named `table_description` doesn't really contribute into the ease of use, so I might suggest keeping it in `kwargs`.

sqlmesh/core/engine_adapter/mysql.py
treysp: This try/except is motivated by permissions errors, which should all come from the database connector library (and not be raised as SQLMesh exceptions by sqlmesh itself). 

What do you think about raising for any sqlmesh exception and warning for everything else?

```
            try:
                ...
            except SQLMeshError as e:
                raise e
            except Exception as e:
                logger.warning(...)
```

sqlmesh/core/engine_adapter/base.py
treysp: To register the comments in the CTAS call for engines with `SUPPORTS_CTAS_SCHEMA_COMMENTS is True`, we have to build a full schema expression like `(a INTEGER COMMENT 'comment')`. 

`columns_to_types` is required to have the types for that schema expression, so if we don't have it available we need to use separate statements instead.

The `schema` object will be `None` unless we have already registered the comments in the CTAS call, so how about this?

```
        if (
            (table_description or column_descriptions)
            and self.COMMENT_CREATION.is_comment_command_only
        ) or (
            column_descriptions
            and not self.COMMENT_CREATION.is_unsupported
            and schema is None
        ):
```

treysp: The simplest approach here would be to just always use the separate commands, but that results in one command per column comment.

Empirically, if columns_to_types will rarely be available then the benefits of comments in the schema won't ever accrue and this complexity is for nought. Do you have a sense for the proportion of CTAS calls that will have columns_to_types available?

tests/core/engine_adapter/test_integration.py
```diff
@@ -421,6 +587,28 @@ def test_temp_table(ctx: TestContext):
...
+        table_description="test table description",
+        column_descriptions={"id": "test id column description"},
+    )
+    results = ctx.get_metadata_results()
+    assert len(results.tables) == 1
+    assert len(results.views) == 0
+    assert len(results.materialized_views) == 0
+    assert results.tables[0] == table.name
+
+    if not ctx.engine_adapter.COMMENT_CREATION.is_unsupported:
```
eakmanrq, line 605: Want to make a `is_supported` property? Just a bit easier to read.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -475,13 +482,86 @@ def _create_table_properties(
...
+        if table_description:
+            properties.append(
+                self._build_description_property_exp(table_description),
+            )
 
         if properties:
             return exp.Properties(expressions=properties)
         return None
 
+    def _create_comments(
```
eakmanrq, line 529: If I am seeing this correctly, this duplicates the same logic as base but just changes the exec SQL syntax to match what BQ expected. Want to add a `_build_comment_exp` and then have BQ override it what what you are doing here? MySQL/Spark would also be updated to follow this pattern.

eakmanrq (approved):
Thanks for the great work on this! One nit comment, one small refactor comment.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -424,14 +424,21 @@ def _fetch_native_df(
         self.execute(query, quote_identifiers=quote_identifiers)
         return self._query_job.to_dataframe()
 
-    def _create_table_properties(
+    def _build_description_property_exp(self, description: str) -> exp.Property:
+        return exp.Property(
+            this=exp.Identifier(this="description", quoted=True),
```
tobymao, line 429:
```suggestion
            exp.to_identifier("description", quoted=True),
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -101,6 +104,7 @@ def __init__(
         cursor_init: t.Optional[t.Callable[[t.Any], None]] = None,
         default_catalog: t.Optional[str] = None,
         execute_log_level: int = logging.DEBUG,
+        register_comments: t.Optional[bool] = None,
```
izeigerman, line 107: Given it's a bool, I'd suggest to just pick False or True as a default.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -134,6 +140,10 @@ def cursor(self) -> t.Any:
     def spark(self) -> t.Optional[PySparkSession]:
         return None
 
+    @property
+    def comments_enabled(self) -> t.Optional[bool]:
```
izeigerman, line 144: Shouldn't be Optional I think

sqlmesh/core/engine_adapter/base.py
izeigerman: If not here, then bellow: 
```
self.register_comments = register_comments if register_comments is not None else <default value>
```

sqlmesh/core/config/connection.py
```diff
@@ -35,6 +35,7 @@
 
 class ConnectionConfig(abc.ABC, BaseConfig):
     concurrent_tasks: int
+    register_comments: bool = False
```
izeigerman, line 38: I don't think it's enough to just add it here. `ConnectionConfig` is an abstract base class and not a pydantic model. I believe you need to include it explicitly in all child classes. See how `concurrent_tasks` is used.

At the same time my recollection might be inaccurate I encourage you to test my claim.

sqlmesh/core/config/connection.py
```diff
@@ -84,6 +85,7 @@ def create_engine_adapter(self) -> EngineAdapter:
             cursor_kwargs=self._cursor_kwargs,
             default_catalog=self.get_catalog(),
             cursor_init=self._cursor_init,
+            register_comments=register_comments or self.register_comments,
```
izeigerman, line 88: If `None` means `False` anyway why not just use `False` as a default argument value.

sqlmesh/core/config/connection.py
izeigerman: Also I thought we wanted it to be `True` for everything but Snowflake.

sqlmesh/core/context.py
```diff
@@ -288,10 +290,13 @@ def __init__(
 
         self._connection_config = self.config.get_connection(self.gateway)
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
+        self.register_comments = register_comments or self._connection_config.register_comments
```
izeigerman, line 293: Why do we need this? We don't use it anywhere.

sqlmesh/core/config/connection.py
```diff
@@ -71,7 +72,7 @@ def _cursor_init(self) -> t.Optional[t.Callable[[t.Any], None]]:
         """A function that is called to initialize the cursor"""
         return None
 
-    def create_engine_adapter(self) -> EngineAdapter:
+    def create_engine_adapter(self, register_comments: bool = False) -> EngineAdapter:
```
eakmanrq, line 75: Want to call this `register_comments_override`? At first I was confused why this was needed since you define if you want to register comments at the connection config level but I think you have this so test connections can override. Let me know if I am missing something.

eakmanrq (approved):
Great job getting this to the finish line! Just one minor code comment and a potential change to docs.

PR #1883, opened 2023-12-21
Fix: Handle 0 batch_size in select_from_values

vchan:
This fixes a bug when calling render from the CLI or the UI.
```
  File "sqlmesh/core/dialect.py", line 780, in select_from_values
    for i in range(0, num_rows, batch_size):
ValueError: range() arg 3 must not be zero
```

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -756,7 +756,7 @@ def extend_sqlglot() -> None:
 def select_from_values(
     values: t.List[t.Tuple[t.Any, ...]],
     columns_to_types: t.Dict[str, exp.DataType],
-    batch_size: int = sys.maxsize,
+    batch_size: int = 0,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -810,7 +810,7 @@ def select_from_values_for_batch_range(
 def pandas_to_sql(
     df: pd.DataFrame,
     columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
-    batch_size: int = 0,
+    batch_size: int = sys.maxsize,
```
tobymao, line 813: i don’t think this is the right fix. batch size 0 should work and mean infinite

tobymao: you can see the comment indicates 0 means no batch

tobymao (approved):

PR #1891, opened 2023-12-22
Fix:  Update overview stage

mykalmax:
simplifies Plan Interface a bit 

https://github.com/TobikoData/sqlmesh/assets/5644753/8f3ddfa4-90a8-4355-9ca6-1f350b1a0491

### DIFF
--- a/web/client/src/library/components/plan/PlanActionsDescription.tsx
+++ b/web/client/src/library/components/plan/PlanActionsDescription.tsx
@@ -74,7 +74,7 @@ export default function PlanActionsDescription(): JSX.Element {
           )}
           {isTrue(forward_only) && (
             <span className="inline-block mr-1">
-              consider as a <b>Breaking Change</b>
+              all changes will be <b>Forward Only</b>
             </span>
           )}
           {isTrue(no_auto_categorization) && (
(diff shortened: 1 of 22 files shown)


tobymao:
we shouln't display fqn, we should always display model names or display names. it should match the cli. so you should never see memory.x.y

mykalmax:
> we shouln't display fqn, we should always display model names or display names. it should match the cli. so you should never see memory.x.y

`fqn` fixes were in another PR and video clip is made before those changes merged.

web/client/src/library/components/plan/PlanActionsDescription.tsx
```diff
@@ -1,77 +1,97 @@
...
+            </span>
+          )}
+          {isTrue(no_auto_categorization) && (
+            <span className="inline-block mr-1">
+              also set <b>Change Category</b> manually
+            </span>
+          )}
+          {isFalse(isStringEmptyOrNil(restate_models)) && (
+            <span className="inline-block mr-1">
+              and restate the followingmodels <b>{restate_models}</b>
```
vchan, line 87:
```suggestion
              and restate the following models <b>{restate_models}</b>
```

web/client/src/library/components/plan/PlanActionsDescription.tsx
```diff
@@ -1,77 +1,97 @@
...
+          {isTrue(skip_backfill) && (
+            <span className="inline-block mr-1">
+              without <b>Backfills</b>
+            </span>
+          )}
+          {isTrue(forward_only) && (
+            <span className="inline-block mr-1">
+              consider as a <b>Breaking Change</b>
+            </span>
+          )}
```
vchan, line 79: Is this right?

vchan (approved):

web/client/src/library/components/plan/PlanActionsDescription.tsx
mykalmax: changed text to Forward Only

PR #1893, opened 2023-12-24
Fix: Send fqn when requesting column lineage

mykalmax:


vchan (approved):

web/server/api/endpoints/lineage.py
```diff
@@ -54,17 +51,17 @@ async def column_lineage(
     context: Context = Depends(get_loaded_context),
 ) -> t.Dict[str, t.Dict[str, LineageColumn]]:
     """Get a column's lineage"""
-    model_fqn = normalize_model_name(model_name, context.default_catalog, context.config.dialect)
+    model_name = normalize_model_name(model_name, context.default_catalog, context.config.dialect)
```
tobymao, line 54: this should just be

model = context.get_model(model_name)

then you can get the fqn with model.fqn

web/server/api/endpoints/lineage.py
```diff
@@ -78,20 +75,17 @@ async def column_lineage(
     for i, node in enumerate(node.walk()):
         if i > 0:
             node_name = _get_table(node) or node.name
-            node_name = (
-                context.snapshots[node_name].name if node_name in context.snapshots else node_name
-            )
+            node_name = normalize_model_name(node_name, None)
```
tobymao, line 78: this shouldn't be necessary

web/server/api/endpoints/lineage.py
```diff
@@ -78,20 +75,17 @@ async def column_lineage(
...
             column_name = exp.to_column(node.name).name
         if column_name in graph.get(node_name, []):
             continue
 
-        node_fqn = normalize_model_name(node_name, context.default_catalog, context.config.dialect)
-        dialect = context.models[node_fqn].dialect if node_fqn in context.models else ""
+        dialect = context.models[node_name].dialect if node_name in context.models else ""
```
tobymao, line 83: context.get_model(...)

web/server/api/endpoints/lineage.py
mykalmax: without `normalize_model_name` node (table) name coming without quotes 

Screenshot 2023-12-28 at 9 44 08 AM

web/server/api/endpoints/lineage.py
mykalmax: how `context.get_model(...)` is different from `context.models[node_name]` since we still need to check if node is actually a model

web/server/api/endpoints/lineage.py
tobymao: the issue is you're calling _get_table which is incorrect. instead it should be 

exp.table_name(node, identify=True)

tobymao: so add identify=True into _get_table helper

web/server/api/endpoints/lineage.py
tobymao: get_model normalizes stuff for you, so you don't need to call normalize_model_name manually

mykalmax: yes, but what if node is not a table nor a model (cte, or  a random external model) ?
Screenshot 2023-12-28 at 10 11 30 AM

tobymao: change get_table to get the expression and find the identifier / alias and call sql(identify=True)

web/server/api/endpoints/lineage.py
```diff
@@ -73,25 +69,22 @@ async def column_lineage(
...
-                context.snapshots[node_name].name if node_name in context.snapshots else node_name
-            )
+            node_name = _get_table(node)
             column_name = exp.to_column(node.name).name
         if column_name in graph.get(node_name, []):
             continue
 
-        node_fqn = normalize_model_name(node_name, context.default_catalog, context.config.dialect)
-        dialect = context.models[node_fqn].dialect if node_fqn in context.models else ""
+        # At his point node_name should be fqn/normalized/quoted
```
tobymao, line 81: at this point

tobymao (approved):

PR #1894, opened 2023-12-26
Fix: Don't send sqlmesh type with files, and handle that independently on ui

mykalmax:
Decouple files from models to speed up loading of file system (files don't need loaded_context).
Move checking for whether a file is a model and of which type to Model class

### DIFF
--- a/web/client/src/library/components/editor/EditorFooter.tsx
+++ b/web/client/src/library/components/editor/EditorFooter.tsx
@@ -5,7 +5,7 @@ import { useEffect } from 'react'
 import { isNil, isNotNil, isStringEmptyOrNil } from '~/utils'
 import Input from '@components/input/Input'
 import { EnumSize } from '~/types/enum'
-import { ModelSQLMeshModel } from '@models/sqlmesh-model'
+import { useStoreContext } from '@context/context'
 
 const EnumFileType = {
   Model: 'model',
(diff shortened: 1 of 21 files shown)


web/client/src/library/components/editor/EditorFooter.tsx
```diff
@@ -73,24 +90,39 @@ export default function EditorFooter({ tab }: { tab: EditorTab }): JSX.Element {
...
-          text="SQLMesh Type"
-        >
-          <EditorIndicator.Text
-            text={tab.file.isSQLMeshModel ? 'Model' : 'Unsupported'}
-          />
-        </EditorIndicator>
-      )}
+      {ModelSQLMeshModel.isSQLMeshModel(tab.file) &&
+        isNotNil(tab.dialect) &&
+        tab.dialect !== '' && (
```
vchan, line 95: Nit: What about using !isStringEmptyOrNil?

web/client/src/library/components/editor/EditorFooter.tsx
```diff
@@ -5,6 +5,23 @@ import { useEffect } from 'react'
...
+  Test: 'test',
+  Audit: 'audit',
+  Macro: 'macro',
+  Hook: 'hook',
+  Log: 'log',
+  Config: 'config',
+  Seed: 'seed',
+  Metric: 'metric',
+  Schema: 'schema',
+  None: 'none',
```
vchan, line 21: Is "unknown" more accurate? None sounds fine too.

web/client/src/models/sqlmesh-model.ts
```diff
@@ -74,4 +92,8 @@ export class ModelSQLMeshModel<
       }
     }
   }
+
+  static isSQLMeshModel(file?: ModelFile): boolean {
+    return Boolean(file?.path?.startsWith('models'))
```
vchan, line 97: What if there are files in the models directory that aren't models?

vchan (approved):

PR #1899, opened 2023-12-28
feat!: Add config for view catalog

eakmanrq:
Adds support for changing the catalog used for an environment schema based on a regex pattern.

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -229,7 +229,7 @@ If you had a model name of `my_schema.table`, the physical table would be create
 
 This key only applies to the _physical tables_ that SQLMesh creates - the views are still created in `my_schema` (prod) or `my_schema__<env>`.
 
-#### View schemas
+#### Environment view schemas
 
 By default, SQLMesh appends the environment name to the schema name when creating new environments. This can be changed to append a suffix at the end of table/view name instead.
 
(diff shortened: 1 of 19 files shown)


sqlmesh/core/config/root.py
```diff
@@ -120,6 +122,16 @@ def _gateways_ensure_dict(cls, value: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
         except Exception:
             return value
 
+    @field_validator("environment_catalog_mapping", mode="before")
+    @classmethod
+    def _validate_regex_keys(cls, value: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
```
tobymao, line 127: can you return a regex compiled string so you don't have to compile it again?

docs/guides/configuration.md
```diff
@@ -260,6 +260,67 @@ If you created a `dev` environment for a project containing a model named `my_sc
 
 The default behavior of appending the suffix to schemas is recommended because it leaves production with a single clean interface for accessing the views. However, if you are deploying SQLMesh in an environment with tight restrictions on schema creation then this can be a useful way of reducing the number of schemas SQLMesh uses.
 
+#### View catalogs
```
izeigerman, line 263: Should this concept be called `Environment catalogs` instead? I feel like "view" is a broader term that extends beyond what's captured by this feature.

docs/guides/configuration.md
```diff
@@ -260,6 +260,67 @@ If you created a `dev` environment for a project containing a model named `my_sc
...
+
+By default, SQLMesh creates environment views in the same [catalog](../concepts/glossary.md#catalog) as the table it is pointing to. 
+The table's catalog is determine by either the catalog defined in it's name or the default catalog defined in the connection.
+A common use case though would be to have the "prod" catalog contain just the tables and "prod" environment views and then put all the "dev" views in a separate catalog. 
+Another use case could be if you have a CI/CD pipeline creating environments, like the [Github Action CI/CD Bot](../integrations/github.md), and you want to put these environments in a dedicated catalog since there can be many of them.
+
+To configure this, you provide a mapping of a [regex pattern](https://en.wikipedia.org/wiki/Regular_expression) to a given catalog name.
```
izeigerman, line 270:
```suggestion
To configure this, you provide a mapping from a [regex pattern](https://en.wikipedia.org/wiki/Regular_expression) to a given catalog name.
```

izeigerman: I'd also clarify that the regex applies to environment names specifically

examples/sushi/config.py
```diff
@@ -125,3 +125,19 @@
     default_test_connection=DuckDBConnectionConfig(catalogs=CATALOGS),
     model_defaults=ModelDefaultsConfig(dialect="duckdb"),
 )
+
+environment_catalog_mapping_config = Config(
+    default_connection=DuckDBConnectionConfig(
+        catalogs={
+            "physical": ":memory:",
```
izeigerman, line 132: What do we need this one for?

sqlmesh/core/context.py
```diff
@@ -306,6 +306,14 @@ def __init__(
         self.users = list({user.username: user for user in self.users}.values())
         self._register_notification_targets()
 
+        if (
+            self.config.environment_catalog_mapping
+            and not self.engine_adapter.CATALOG_SUPPORT.is_multi_catalog_supported
```
izeigerman, line 311: Should we also check whether the cross-catalog querying is supported?

sqlmesh/core/state_sync/common.py
```diff
@@ -128,20 +128,29 @@ def promote(
...
+            environment_suffix_target_changed = (
+                environment.suffix_target != existing_environment.suffix_target
+            )
+            if environment_suffix_target_changed:
+                removed_environment_naming_info.suffix_target = existing_environment.suffix_target
+            environment_catalog_changed = (
+                environment.catalog_name_override != existing_environment.catalog_name_override
+            )
+            if environment_catalog_changed:
+                removed_environment_naming_info.catalog_name_override = (
```
izeigerman, line 150: Do we need parenthesis here?

examples/sushi/config.py
eakmanrq: This is for the tests I run. The top catalog becomes the default catalog which I call "physical". I am testing to confirm that all the tables are created in "physical" and the views are created in their expected prod or dev catalogs.

sqlmesh/core/context.py
eakmanrq: That is what this is checking

docs/guides/configuration.md
eakmanrq: I see your point but I am following the pattern from the previous titles. It goes like this:

* Physical table schemas
...
* View schemas
...
* View catalogs
....

So that matches the pattern.

eakmanrq: What do you think about `Environment view schemas` and `Environment view catalogs`?

izeigerman: Yep, your suggestion sounds good, thanks

sqlmesh/core/context.py
izeigerman: Ah, I thought this only checked whether multiple catalogs are supported at all. Got confused

izeigerman (approved):

PR #1905, opened 2023-12-28
fix: add more caching to intervals and cron

tobymao:


sqlmesh/core/plan/definition.py
```diff
@@ -332,7 +332,10 @@ def has_changes(self) -> bool:
 
     def is_new_snapshot(self, snapshot: Snapshot) -> bool:
         """Returns True if the given snapshot is a new snapshot in this plan."""
-        return snapshot.snapshot_id in {s.snapshot_id for s in self.new_snapshots}
```
tobymao, line 335: @eakmanrq you missed this section in your refactor, it's an N^2 operation

eakmanrq: Thanks for catching. Want to just update `new_snapshots` so it is a dict of snapshot ID to snapshot?

tobymao: ok i can do that

izeigerman (approved):

sqlmesh/core/plan/definition.py
tobymao: decided not to do that because then i have to change other places to values

PR #1908, opened 2023-12-29
feat: add model tag restatement selector

eakmanrq:


sqlmesh/core/context.py
```diff
@@ -867,6 +874,14 @@ def plan(
         if restate_models is not None:
             restate_models = model_selector.expand_model_selections(restate_models)
 
+        if restate_tags is not None:
+            restate_models = restate_models or set()
+            restate_models.update(model_selector.expand_model_tags(restate_tags))
+            if not restate_models:
+                raise PlanError(
+                    f"No models match the provided tags. Tags: {', '.join(restate_tags)}"
+                )
```
eakmanrq, line 883: Do we think we should raise here?

sqlmesh/core/selector.py
```diff
@@ -108,35 +178,29 @@ def expand_model_selections(self, model_selections: t.Iterable[str]) -> t.Set[st
...
+                selection,
+                include_upstream,
+                include_downstream,
+            ) = self._get_value_and_dependency_inclusion(selection)
 
             if "*" in selection:
                 for model in self._models.values():
-                    if fnmatch.fnmatch(model.name, selection):
-                        _add_model(model.fqn, include_upstream, include_downstream)
+                    if fnmatch.fnmatchcase(model.name, selection):
```
eakmanrq, line 193: @izeigerman I think you wanted to use `fnmatchcase` here instead of `fnmatch`. `fnmatch` determines if the matching is case-sensitive based on the OS which feels irrelevant within this context. Therefore I think always matching on case makes more sense here.

mkdocs.yml
```diff
@@ -105,6 +105,7 @@ theme:
         icon: material/weather-night
         name: Switch to light mode
   features:
+    - content.tabs.link
```
eakmanrq, line 108: This makes it so that if a user picks an option on a page (like "Model" or "Tag" in this example) then it updates all of the codeblocks to match that selection.

izeigerman:
Instead of adding new arguments like `--restate-tag`, `--select-tag`, etc., can we instead extend the model selection syntax to incorporate tags? Eg. `--select-model tag:my_tag`?

sqlmesh/core/context.py
izeigerman: Should this just be a noop instead?

eakmanrq:
> Rather than adding new arguments like `--restate-tag`, `--select-tag`, etc., can we instead extend the model selection syntax to incorporate tags? Eg. `--select-model tag:my_tag`?

Great suggestion will do.

sqlmesh/core/context.py
```diff
@@ -877,12 +882,14 @@ def plan(
                 environment or c.PROD,
                 snapshots=self._snapshots(models_override),
                 create_from=create_from,
+                force_no_diff=(restate_models is not None and not expanded_restate_models)
```
izeigerman, line 885: Why do we need this?

sqlmesh/core/selector.py
```diff
@@ -97,7 +113,67 @@ def select_models(
 
         return models
 
-    def expand_model_selections(self, model_selections: t.Iterable[str]) -> t.Set[str]:
+    @classmethod
```
izeigerman, line 116: maybe `@staticmethod` instead

sqlmesh/core/selector.py
```diff
@@ -106,37 +182,38 @@ def expand_model_selections(self, model_selections: t.Iterable[str]) -> t.Set[st
...
-                    include_upstream,
-                    include_downstream,
-                )
+                model_fqn = normalize_model_name(selection, self._default_catalog, self._dialect)
+                if model_fqn in models:
+                    matched_models.add(model_fqn)
 
-        return result
+            if not matched_models:
+                logger.warning(f"Model {selection} doesn't match any models.")
```
izeigerman, line 214:
```suggestion
                logger.warning(f"Expression '{selection}' doesn't match any models.")
```

sqlmesh/core/selector.py
```diff
@@ -97,7 +113,67 @@ def select_models(
...
+
+        if "*" in selection:
+            for model_tag in self._models_by_tag:
+                if fnmatch.fnmatchcase(model_tag, selection):
+                    matched_tags.add(model_tag)
+        elif selection in self._models_by_tag:
+            matched_tags.add(selection)
+
+        if not matched_tags:
+            logger.warning(f"Tag {tag_selection} doesn't match any models.")
```
izeigerman, line 166:
```suggestion
            logger.warning(f"Expression '{tag_selection}' doesn't match any models.")
```

izeigerman:
@eakmanrq  Just to confirm: this update doesn't cover using a superset of models from both the local state and the target environment?

eakmanrq:
@izeigerman Is this what you are referring to? https://github.com/TobikoData/sqlmesh/blob/e30e60a5a7e157648e6ae2b81f5ac5f413429b7f/sqlmesh/core/selector.py#L89-L91

izeigerman:
@eakmanrq yes, thank you, overlooked that

izeigerman (approved):

sqlmesh/core/context.py
eakmanrq: If you don't do this then you will have a diff since you are including snapshots and other configuration in plan.

sqlmesh/core/selector.py
```diff
@@ -97,7 +113,67 @@ def select_models(
...
+
+        if "*" in selection:
+            for model_tag in self._models_by_tag:
+                if fnmatch.fnmatchcase(model_tag, selection):
+                    matched_tags.add(model_tag)
+        elif selection in self._models_by_tag:
+            matched_tags.add(selection)
+
+        if not matched_tags:
+            logger.warning(f"Expression '{tag_selection}' doesn't match any models.")
```
izeigerman, line 166:
```suggestion
            logger.warning(f"Expression 'tag:{tag_selection}' doesn't match any models.")
```
Sorry, I've just realized that you strip the `tag:` prefix upstream.

PR #1909, opened 2023-12-29
Fix: Set the correct start timestamp for a plan with restatements in the presence of missing intervals

izeigerman:


tests/core/test_plan.py
```diff
@@ -189,6 +191,55 @@ def test_restate_models(sushi_context_pre_scheduling: Context):
         sushi_context_pre_scheduling.plan(restate_models=['"unknown_model"'], no_prompts=True)
 
 
+@freeze_time()
```
eakmanrq, line 194: Would it be better to freeze at a specific time? That way you don't have to worry about day switching issues. Also you could hardcode values in the test.

tobymao (approved):

tests/core/test_plan.py
izeigerman: I'm freezing here to prevent day switching issue. The specific day doesn't really matter and the current timestamp at the time of freezing works just fine.

eakmanrq: Ok nevermind not an issue.

PR #1910, opened 2023-12-29
Fix: Run reload_context in file watcher in executor

mykalmax:


web/server/watcher.py
```diff
@@ -80,6 +79,9 @@ async def watch_project() -> None:
             )
         )
 
+        if should_load_context:
+            asyncio.create_task(run_in_executor(reload_context, context))
```
vchan, line 83: I think you can await it. No need to wrap it in a task.

vchan (approved):

PR #1911, opened 2023-12-29
feat: add full snowflake private key support

eakmanrq:


sqlmesh/dbt/target.py
```diff
@@ -216,6 +216,11 @@ class SnowflakeConfig(TargetConfig):
     # SSO authentication
     authenticator: t.Optional[str] = None
 
+    # Key Pair Auth
+    private_key: t.Optional[str] = None
+    private_key_path: t.Optional[str] = None
+    private_key_passphrase: t.Optional[str] = None
+
     # TODO add other forms of authentication
```
izeigerman, line 224: Is this TODO relevant after this update?

izeigerman (approved):

sqlmesh/dbt/target.py
eakmanrq: There is still work to, at a minimum, add other configuration properties. I'm also not sure if we have parity with Snowflake auth method behavior. We got a report for example that we keep re-asking to auth in browser when using SSO on each connection when dbt does not do that. So still work needed here.

PR #1912, opened 2023-12-29
Feat: Explicitly label preview backfills in the plan summary

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -177,7 +179,7 @@ def modified_snapshots(self) -> t.Dict[SnapshotId, Snapshot]:
     @property
     def start(self) -> TimeLike:
         """Returns the start of the plan or the earliest date of all snapshots."""
-        if not self.override_start and (self._restate_models or not self._missing_intervals):
+        if not self._start and (self._restate_models or not self._missing_intervals):
```
izeigerman, line 182: @eakmanrq we should not rely on `override_start` because sometimes `_start` is set to `default_start` in which case the `override_start` flag won't be set. At the same time we should still use the existing value of `_start`.

sqlmesh/core/plan/definition.py
```diff
@@ -744,8 +748,8 @@ def _refresh_dag_and_ignored_snapshots(self) -> None:
                 ).sorted
             }
 
-        self._add_restatements()
         self.__missing_intervals = None
+        self._add_restatements()
```
izeigerman, line 752: @eakmanrq another potential issue. We used to add restatements before resetting previous intervals, which means that the `self.start` used to compute restatement intervals could've been derived from previous missing intervals.

tobymao (approved):

eakmanrq:
Maybe not in this PR, but are there tests we can write that expose the issues you listed above?

izeigerman:
> Maybe not in this PR, but are there tests we can write that expose the issues you listed above?

The only issue here that I'm not sure how to materialize is the wrong order of reseting missing intervals vs adding restatements. Once I fixed this issue the existing tests exposed the other `_start` / `override_start` related issue.

Unfortunately, I currently struggle to come up with a scenario in which this would matter. It should probably involve a combination of ignored snapshots, forward-only and partially backfilled DAG?

izeigerman:
I'm hoping to refactor this whole logic soon and break up the loop that currently exists between missing intervals, plan start and restatements. This would eliminate the possibility of any issues of this sort.

PR #1913, opened 2023-12-30
Fix: rollback should not fail if versions are not in alignment since …

z3z1ma:
…its essentially expected. Rollbacks should make an effort to proceed. Otherwise its difficult for users to reconcile depending on how they manage their prod environment. In my particular case, a dev install I did (the one I was asked to test) ended up having a newer schema version than the prod release which incorporated the changes. It means prod would not work and it was asking me to "update" from `0.62.1` to  `0.61.1dev8` or something along those lines which doesnt even make sense from a semver perspective. 

Since rollbacks just swap tables, this seems fine. 

Another unrelated error, **but** also affecting rollbacks (on postgres unfortunately) fixed here:
https://github.com/tobymao/sqlglot/pull/2736

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -157,11 +157,12 @@ class StateReader(abc.ABC):
     def close(self) -> None:
         """Closes all open connections and releases all allocated resources."""
 
-    def get_versions(self, validate: bool = True) -> Versions:
+    def get_versions(self, validate: bool = True, force: bool = False) -> Versions:
         """Get the current versions of the SQLMesh schema and libraries.
 
         Args:
             validate: Whether or not to raise error if the running version is ahead of state.
+            force: Whether or not to force the versions to be fetched and not raise an error.
 
         Returns:
             The versions object.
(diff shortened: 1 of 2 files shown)


sqlmesh/core/state_sync/base.py
```diff
@@ -190,23 +191,23 @@ def raise_error(
                 f"{lib} (local) is using version '{local}' which is behind '{remote}' (remote).{upgrade_suggestion}"
             )
 
-        if SCHEMA_VERSION < versions.schema_version:
+        if SCHEMA_VERSION < versions.schema_version and not force:
```
izeigerman, line 194: I think we can simplify this slightly and instead of adding `not force` to every branch we can just add the following at line 171:
```
if not force:
    return versions
```
to circumvent all validations. 

Alternatively, @tobymao @z3z1ma  , shouldn't `validate=False` take care of this?

tobymao: yea i'm a bit confused about why this new flag was needed.

izeigerman: @z3z1ma let's remove the `force` flag and immediately return `versions if`validate=False`

tobymao:
closed in favor of https://github.com/TobikoData/sqlmesh/pull/1917

PR #1915, opened 2024-01-02
Feat: Simple inline macro functions

z3z1ma:
An attempt at allowing inline defs to be callables like macros. Works for the simplest case.

```sql
MODEL(...);

-- A macrodef should be able to take a simple lambda and do substitution as a callable
-- Allows for highly specific macros to be colocated with the model
-- Makes models with complex biz logic easier to break down, inline
@DEF(convert_to_int, x -> case when left(x, 1) = 'A' then 1 when left(x, 1) = 'B' then 2 when left(x, 1) = 'C' then 3 end);

SELECT
  id,
  cust_rank_1, 
  cust_rank_2,
  cust_rank_3
  @convert_to_int(cust_rank_1) as cust_rank_1_int,
  @convert_to_int(cust_rank_2) as cust_rank_2_int,
  @convert_to_int(cust_rank_3) as cust_rank_3_int
FROM
  some.model
```

Can add a test and run this direction if you think its the right way. Otherwise defer to you all 😃

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -225,8 +225,10 @@ class MacroEvaluator:
     def evaluate(self, node: MacroFunc) -> exp.Expression | t.List[exp.Expression] | None:
         if isinstance(node, MacroDef):
             if isinstance(node.expression, exp.Lambda):
-                _, fn = _norm_var_arg_lambda(self, node.expression, node.expression.this.name)
-                self.macros[normalize_macro_name(node.name)] = lambda _, *args: fn(*args)
+                _, fn = _norm_var_arg_lambda(self, node.expression)
+                self.macros[normalize_macro_name(node.name)] = lambda _, *args: fn(
+                    exp.Tuple(expressions=args) if len(args) > 1 else args
+                )
             else:
                 self.locals[node.name] = self.transform(node.expression)
             return node
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -224,7 +224,13 @@ class MacroEvaluator:
 
     def evaluate(self, node: MacroFunc) -> exp.Expression | t.List[exp.Expression] | None:
         if isinstance(node, MacroDef):
-            self.locals[node.name] = self.transform(node.expression)
+            if isinstance(node.expression, exp.Lambda):
+                _, func = _norm_var_arg_lambda(
+                    self, node.expression, node.expression.this.name
+                )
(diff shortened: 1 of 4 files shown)


z3z1ma:
Also handles multiple args (leveraging existing code), contrived examples below

```sql
@DEF(pythag, (x,y) -> pow(x, 2) + pow(y, 2));

SELECT
  @pythag(sideA, sideB) AS sideC
FROM
  some.triangle
```


```sql
@DEF(nrr, (starting_mrr, expansion_mrr, churned_mrr) -> (starting_mrr + expansion_mrr - churned_mrr) / starting_mrr);

SELECT
  @nrr(fy21_mrr, fy21_expansions, fy21_churns) AS fy21_net_retention_rate,
  @nrr(fy22_mrr, fy22_expansions, fy22_churns) AS fy22_net_retention_rate,
  @nrr(fy23_mrr, fy23_expansions, fy23_churns) AS fy23_net_retention_rate,
FROM
  some.revenue
```

sqlmesh/core/macros.py
```diff
@@ -224,7 +224,13 @@ def template(self, text: t.Any, local_variables: t.Dict[str, t.Any]) -> str:
 
     def evaluate(self, node: MacroFunc) -> exp.Expression | t.List[exp.Expression] | None:
         if isinstance(node, MacroDef):
-            self.locals[node.name] = self.transform(node.expression)
+            if isinstance(node.expression, exp.Lambda):
+                _, fn = _norm_var_arg_lambda(self, node.expression)
+                self.macros[normalize_macro_name(node.name)] = lambda _, *args: fn(
+                    exp.Tuple(expressions=args) if len(args) > 1 else args
```
tobymao, line 230: args here should be a a list, list(args)

tobymao:
this looks good alex, let's add a few tests

eakmanrq:
Also please add documentation

sqlmesh/core/model/definition.py
```diff
@@ -1823,6 +1823,10 @@ def _python_env(
             for macro_func_or_var in expression.find_all(d.MacroFunc, d.MacroVar):
                 if macro_func_or_var.__class__ is d.MacroFunc:
                     name = macro_func_or_var.this.name.lower()
+                    if name not in macros:
+                        # This macro is assumed to be in the same file as the model definition.
+                        # If it's not, it will be caught later when the macro is rendered.
+                        continue
```
tobymao, line 1829: let's copy the below,

if name in macros:
   used_macros[name] = macros[name]

PR #1916, opened 2024-01-02
fix: only SQLGlot format sql files

eakmanrq:
Prior to this if you had a Python model that returned a SQLGlot expression it would try to format it assuming it contained just SQL when it contained Python.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -738,7 +738,7 @@ class Context(BaseContext):
     def format(self, transpile: t.Optional[str] = None, newline: bool = False) -> None:
         """Format all models in a given directory."""
         for model in self._models.values():
-            if not model.is_sql:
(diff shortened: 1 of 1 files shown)


sqlmesh/core/context.py
```diff
@@ -738,7 +738,7 @@ def evaluate(
     def format(self, transpile: t.Optional[str] = None, newline: bool = False) -> None:
         """Format all models in a given directory."""
         for model in self._models.values():
-            if not model.is_sql:
+            if not model.file_extension.is_sql:
```
tobymao, line 741: just do model._path.suffix == ".sql"

eakmanrq: updated

tobymao (approved):

georgesittas (approved):

PR #1919, opened 2024-01-03
feat: Update Notebook Magics to match CLI functionality

eakmanrq:
Key changes:
* Make sure all CLI commands have a magic command with the same arguments
* Add tests that verify behavior in Jupyter notebooks
* Add an example Jupyter notebook that can be run with `make jupyter-example` in order to see what the output looks like across most commands when a dev is making changes

Note: A follow up PR will be focused on testing and validating Databricks output. More limited in options for that but will see what we can do. It might just be manually running stuff and verify the output works.


Makefile
```diff
@@ -104,3 +104,6 @@ clean-build:
 	rm -rf build/ && rm -rf dist/ && rm -rf *.egg-info
 
 dev-publish: ui-build clean-build publish
+
+jupyter-example:
+	jupyter notebook tests/integrations/jupyter/example_outputs.ipynb
```
tobymao, line 109: can we default to lab instead, jupyter classic is deprecated i think

setup.py
```diff
@@ -75,6 +75,7 @@
             "freezegun",
             "google-auth",
             "isort==5.10.1",
+            "jupyter",
```
tobymao, line 78: jupyterlab

tobymao:
make sure to update all the docs

sqlmesh/core/console.py
```diff
@@ -1674,12 +1674,19 @@ def get_console(**kwargs: t.Any) -> TerminalConsole | DatabricksMagicConsole | N
...
+    runtime_env = RuntimeEnv.get()
 
     runtime_env_mapping = {
         RuntimeEnv.DATABRICKS: DatabricksMagicConsole,
         RuntimeEnv.JUPYTER: NotebookMagicConsole,
         RuntimeEnv.TERMINAL: TerminalConsole,
         RuntimeEnv.GOOGLE_COLAB: NotebookMagicConsole,
     }
-    return runtime_env_mapping[runtime_env](**kwargs)
+    rich_console_kwargs: t.Dict[str, t.Any] = dict(theme=srich.theme)
```
tobymao, line 1687: {"theme": srich.theme}

sqlmesh/magics.py
```diff
@@ -24,48 +28,52 @@
...
+logger = logging.getLogger(__name__)
+
 CONTEXT_VARIABLE_NAMES = [
     "context",
     "ctx",
     "sqlmesh",
 ]
 
 
+def pass_sqlmesh_context(func: t.Callable) -> t.Callable:
```
tobymao, line 40: why was this added?

eakmanrq: Allows magics that need a context to access a context while ensuring it is configured properly. It kind of already existed but with a property but now this ensure you only get the context once and we don't have to implement any caching.

setup.py
```diff
@@ -75,6 +75,8 @@
             "freezegun",
             "google-auth",
             "isort==5.10.1",
+            "jupyerlab",
+            "jupyterlab_server==2.19.0",
```
tobymao, line 79: why do you need server?

tobymao (approved):

setup.py
eakmanrq: This is something I am testing. Trying to fix dependency issues.

PR #1925, opened 2024-01-04
Chore: Refactor Plan definition into immutable Plan and mutable PlanBuilder

izeigerman:


sqlmesh/core/plan/builder.py
```diff
@@ -0,0 +1,617 @@
...
+        if not self._start and is_dev and forward_only:
+            self._start = default_start or yesterday_ds()
+
+        self._environment_naming_info = EnvironmentNamingInfo.from_environment_catalog_mapping(
+            environment_catalog_mapping or {},
+            name=self._context_diff.environment,
+            suffix_target=environment_suffix_target,
+        )
+
+        self._override_start = start is not None
```
tobymao, line 109: just make these public and delete the properties

sqlmesh/core/plan/builder.py
```diff
@@ -0,0 +1,617 @@
...
+        self._override_end = end is not None
+        self._plan_id: str = random_id()
+        self._model_fqn_to_snapshot = {s.name: s for s in self._context_diff.snapshots.values()}
```
tobymao, line 112: isn't this exactly the same as context_diff.snapshots?

tobymao: ah i guess snapshots is snapshot_id

sqlmesh/core/plan/builder.py
izeigerman: This is intentional to signal that these values are not supposed to be modified

tobymao (approved):

sqlmesh/core/context.py
```diff
@@ -826,6 +826,88 @@ def plan(
...
+            run=run,
+        )
+
+        self.console.plan(
+            plan_builder, auto_apply, self.default_catalog, no_diff=no_diff, no_prompts=no_prompts
+        )
+
+        return plan_builder.build()
+
+    def plan_builder(
```
eakmanrq, line 855: Sure you want to make this public? Not sure if we want to commit to exposing this.

eakmanrq (approved):
Excited for this!

sqlmesh/core/plan/builder.py
tobymao: meh, no one's gonna modify it

izeigerman: famous last words lol. Ok will change this

z3z1ma: Interesting 👀

sqlmesh/core/context.py
izeigerman: Discussed in person. We concluded that this method makes a lot more sense when using SQLMesh API as opposed to the `plan` one which is clearly geared towards using in CLI or Notebook.

PR #1927, opened 2024-01-04
Fix: Only push new records when migrating snapshot rows

izeigerman:


tobymao (approved):

tobymao:
how did you run into this?

eakmanrq (approved):

eakmanrq:
> how did you run into this?

Someone ran a double migration (migrated and then did another migration to a newer version without running plan).

PR #1929, opened 2024-01-04
feat!: mapping schema should not have quotes

tobymao:


sqlmesh/core/model/definition.py
```diff
@@ -533,7 +533,7 @@ def update_schema(
             if mapping_schema:
                 nested_set(
                     self.mapping_schema,
-                    tuple(str(part) for part in table.parts),
+                    tuple(part.sql(copy=False)[1: -1] for part in table.parts),
```
georgesittas, line 536: Is it guaranteed we'll always have quotes here?

georgesittas (approved):
Left a comment to consider. If there's no guarantee for the quotes we could wrap the slice in a lambda that checks for quotes to make the fix more robust (albeit a bit slower)

sqlmesh/core/model/definition.py
tobymao: yes, because all tables are fqn quoted

izeigerman:
I think this might require a migration

tobymao:
@izeigerman that's why i upgraded sqlglot

PR #1930, opened 2024-01-04
feat: add model caching for snapshots

tobymao:


sqlmesh/core/selector.py
```diff
@@ -91,6 +91,12 @@ def select_models(
         )
 
         dag: DAG[str] = DAG()
+        subdag = set()
+
+        for fqn in all_selected_models:
+            subdag.add(fqn)
+            subdag.update(self._dag.downstream(fqn))
```
izeigerman, line 98: This may get rather expensive given the selector expression itself may include downstream through the trailing +.

PR #1931, opened 2024-01-05
Feat: Use max interval end of selected models when picking a default plan end

izeigerman:
This update ensures that when using model selector the default plan end will be inferred using the last backfilled interval of the selected model on prod.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -139,13 +139,13 @@ class StateReader(abc.ABC):
 
     @abc.abstractmethod
     def max_interval_end_for_environment(
-        self, environment: str, model_fqns: t.Optional[t.Set[str]] = None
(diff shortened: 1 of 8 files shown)


sqlmesh/core/state_sync/base.py
```diff
@@ -138,11 +138,15 @@ def get_environments(self) -> t.List[Environment]:
         """
 
     @abc.abstractmethod
-    def max_interval_end_for_environment(self, environment: str) -> t.Optional[int]:
+    def max_interval_end_for_environment(
+        self, environment: str, model_fqns: t.Optional[t.Set[str]] = None
```
tobymao, line 142: i don't think we need to use "fqn" everywhere, there's really only one representation now, could just call it models, the type sig is a str, but whatever up to you

tobymao (approved):

PR #1933, opened 2024-01-05
chore: reorganize test marks

eakmanrq:
No logic changes but changes to CI/CD workflow.

Expanded test marks to make it easier to test specific subsets of tests. See new marks here: https://github.com/TobikoData/sqlmesh/blob/eakmanrq/refactor_test_markings/pytest.ini

If working on core features and want quick iteration cycle, `make core-unit-test` now provides a quicker feedback loop. Going to also look into parallelizing our tests but that will require some changes for tests that are not properly isolated. 

For make commands assumed that if a user wants integration tests then they also want unit tests (since they are so fast in comparison). As a result CI/CD now just runs unit/integration tests together. 

This change also now makes it much easier to isolate specific engines for testing compared to before (which was just commenting out code).

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -171,7 +171,7 @@ jobs:
           command: sudo apt-get update && sudo apt-get install libpq-dev
       - run:
           name: Install dependencies
-          command: make install-engine-it
+          command: make install-engine-test
       - run:
           name: Bring up Dockerized Engines
           command: make engine-up
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -37,7 +37,7 @@ commands:
             - run: circleci-agent step halt
 
 jobs:
-  style_and_unit_tests:
+  style_and_integration_tests:
     parameters:
       python_version:
         type: string
@@ -57,13 +57,13 @@ jobs:
           name: Run linters and code style checks
           command: make py-style
       - run:
-          name: Run unit tests
-          command: make unit-test
+          name: Run integration tests
+          command: make it-test
       - run:
           name: Run doc tests
           command: make doc-test
 
-  style_and_unit_tests_pydantic_v1:
+  style_and_integration_tests_pydantic_v1:
     docker:
       - image: cimg/python:3.10
     resource_class: small
@@ -83,8 +83,8 @@ jobs:
           name: Run linters and code style checks
           command: make py-style
       - run:
-          name: Run unit tests
-          command: make unit-test
+          name: Run integration tests
+          command: make it-test
 
   ui_style:
     docker:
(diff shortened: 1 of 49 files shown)


.circleci/continue_config.yml
```diff
@@ -57,13 +57,13 @@ jobs:
           name: Run linters and code style checks
           command: make py-style
       - run:
-          name: Run unit tests
-          command: make unit-test
+          name: Run integration tests
```
tobymao, line 60: why did you get rid of unit tests? is it moved somewhere?

Makefile
```diff
@@ -107,3 +73,95 @@ dev-publish: ui-build clean-build publish
...
+	pytest -m "docker and engine"
+
+engine-remote-test:
+	pytest -m "remote and engine"
+
+engine-test:
+	pytest -m "engine"
+
+dbt-test:
+	pytest -m "dbt"
```
crericha, line 125: Would you mind adding a

.circleci/continue_config.yml
eakmanrq: Unit tests still exists. See make file. CI/CD now runs integration which includes includes unit tests.

tobymao:
the only thing that i'd comment on is that we should be consistent with it vs integration, can we just put "it" everywhere instead of integration?

tobymao (approved):

tests/core/engine_adapter/test_base_postgres.py
```diff
@@ -2,10 +2,13 @@
 import typing as t
 from unittest.mock import call
 
+import pytest
 from sqlglot import exp, parse_one
 
 from sqlmesh.core.engine_adapter.base_postgres import BasePostgresEngineAdapter
 
+pytestmark = [pytest.mark.postgres, pytest.mark.redshift]
```
izeigerman, line 10: Shouldn't this include `pytest.mark.engine` since `engine` is:

> engine: test all engine adapters

? Eg. if I run with `-m engine` I'd expect all engine tests to run but it will actually only run the common logic. Maybe rename to `engine_common`?

tests/core/test_plan.py
```diff
@@ -215,6 +216,7 @@ def test_restate_models(sushi_context_pre_scheduling: Context):
     assert not plan.requires_backfill
 
 
+@pytest.mark.integration
```
izeigerman, line 219: Should this one https://github.com/TobikoData/sqlmesh/blob/9ca41c82abcc93bab41935793338080883f721e7/tests/core/test_plan.py#L1239 be integration too?

tests/core/engine_adapter/test_base_postgres.py
eakmanrq: Yep thanks

PR #1936, opened 2024-01-08
Fix: Revert to the old plan end behavior

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -72,8 +72,8 @@ def start(self) -> TimeLike:
         return t.cast(TimeLike, self._start)
 
     @property
-    def end_or_now(self) -> TimeLike:
-        return self.end or now()
+    def end(self) -> TimeLike:
+        return self.provided_end or now()
```
crericha, line 76: Should we cache now(), so it doesn't change if called multiple times during apply? Worst case being 23:59:59:999 and 00:00:00:000

crericha (approved):

tobymao (approved):

PR #1939, opened 2024-01-09
chore: parallelize tests by module

eakmanrq:
Low effort way to improve test performance. Cut times locally in half. Changed Circle from Small to Medium and nearly cut the test time (just test) in half. 

Biggest issue now is that we load balance the tests across the workers by module. This is because some module assume we aren't running in parallel and therefore share resources across tests. This results in a skew across the workers. Fix shouldn't be too bad but taking the easy win for now.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -28,6 +28,7 @@ markers =
     snowflake: test for Snowflake
     spark: test for Spark
     trino: test for Trino
+addopts = -n auto --dist=loadscope
 
 # Set this to True to enable logging during tests
 log_cli = False
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import datetime
 import typing as t
+import uuid
 from pathlib import Path
-from shutil import rmtree
+from shutil import copytree, rmtree
 from tempfile import TemporaryDirectory
 from unittest import mock
 from unittest.mock import PropertyMock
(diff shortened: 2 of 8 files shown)


pytest.ini
```diff
@@ -28,6 +28,7 @@ markers =
     snowflake: test for Snowflake
     spark: test for Spark
     trino: test for Trino
+addopts = -n auto --dist=loadscope
```
eakmanrq, line 31: @izeigerman `-n <X>` is the same as `--tx <X>*popen//python=python` as long as the Python interpreter you are targeting is the same as what you are running Pytest with. In other words, `--tx` is for assigning work to another python interpreter and you can shortcut this with `-n` is you are running on the same interpreter.

izeigerman: Beautiful, thanks!

tests/conftest.py
```diff
@@ -200,56 +201,58 @@ def sushi_context_fixed_date(mocker: MockerFixture) -> Context:
...
-    paths: str | t.List[str],
-    mocker: MockerFixture,
-    config="test_config",
-) -> t.Tuple[Context, Plan]:
-    delete_cache(paths)
-    sushi_context = Context(paths=paths, config=config)
-    confirm = mocker.patch("sqlmesh.core.console.Confirm")
-    confirm.ask.return_value = False
+@pytest.fixture
+def init_and_plan_context(copy_to_temp_path, mocker) -> t.Callable:
```
izeigerman, line 242: Is `copy_to_temp_path` used here?

eakmanrq: Thanks that is a mistake when refactoring before opening PR

tests/conftest.py
```diff
@@ -313,6 +316,27 @@ def _make_function(klass: t.Type[T], dialect: t.Optional[str] = None) -> T:
...
+@pytest.fixture
+def copy_to_temp_path(tmp_path: Path) -> t.Callable:
+    def ignore(src, names):
+        if Path(src).name in {".cache", "__pycache__", "logs", "data"}:
+            return names
+        return []
+
+    def _make_function(
+        paths: t.Union[t.Union[str, Path], t.Collection[t.Union[str, Path]]]
+    ) -> t.List[Path]:
```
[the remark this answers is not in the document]
eakmanrq: It now does after fixing that mistake: https://github.com/TobikoData/sqlmesh/pull/1939/commits/63e660fcfdca960f24e6a4d5637090102934e2cf

PR #1941, opened 2024-01-09
Feat: parse multiline model documentation in a nice way

z3z1ma:
This is predicated on how sqlglot / sqlmesh parses triple quoted strings, but the idea is that these are equivalent:

```sql
MODEL(
    name analytics_activation.license_utilization_report_v2,
    kind FULL,
    ...,
    description '''
    This is the primary output for Domo to report on license usage. It is a filtered
    view of the fct_license_utilization table with some additional columns to make
    reporting easier.
    ''',
    grain account_module_unit_key
);
```

```sql
MODEL(
    name analytics_activation.license_utilization_report_v2,
    kind FULL,
    ...,
    description 'This is the primary output for Domo to report on license usage. It is a filtered view of the fct_license_utilization table with some additional columns to make reporting easier.',
    grain account_module_unit_key
);
```

Descriptions could get very long but are valuable to have inline with the model. We should support multiline strings.

The argument could be made for column comments too. 

```sql
SELECT
  complex_computation, /*
    What is happening here is ...
    
    E = mc2
    
    And the output is as such */
  more,
  columns
FROM
  something
```

LMK thoughts.

### DIFF
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -235,7 +235,7 @@ class _Node(PydanticModel):
     def _description_validator(cls, v: t.Any) -> t.Optional[str]:
         maybe_str = str_or_exp_to_str(v)
         if isinstance(maybe_str, str):
-            return " ".join(inspect.cleandoc(maybe_str).splitlines())
+            return inspect.cleandoc(maybe_str)
         return maybe_str
 
     @field_validator("owner", "stamp", mode="before")
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import typing as t
 from datetime import datetime
 from enum import Enum
@@ -229,7 +230,15 @@ class _Node(PydanticModel):
                 raise ConfigError(f"Invalid cron expression '{cron}'")
         return cron
 
-    @field_validator("owner", "description", "stamp", mode="before")
+    @field_validator("description", mode="before")
+    @classmethod
+    def _description_validator(cls, v: t.Any) -> t.Optional[str]:
+        maybe_str = str_or_exp_to_str(v)
+        if isinstance(maybe_str, str):
+            return " ".join(inspect.cleandoc(maybe_str).splitlines())
+        return maybe_str
+
+    @field_validator("owner", "stamp", mode="before")
     @classmethod
     def _string_expr_validator(cls, v: t.Any) -> t.Optional[str]:
         return str_or_exp_to_str(v)


tobymao:
@z3z1ma i think this can be purely a UI formatting feature. so we can just put this in the format_models code and the IDE -- as opposed to changing it in the model base class, what do you think?

z3z1ma:
The problem, to me, isn't in how its formatted, its in how its parsed. 

This:

```sql
    '''
    This is the primary output for Domo to report on license usage. It is a filtered
    view of the fct_license_utilization table with some additional columns to make
    reporting easier.
    '''
```

Would be parsed as this:

`'\n    This is the primary output for Domo to report on license usage. It is a filtered\n    view of the fct_license_utilization table with some additional columns to make\n    reporting easier.\n    '`


Without my change. And with my change, this:

`'This is the primary output for Domo to report on license usage. It is a filtered view of the fct_license_utilization table with some additional columns to make reporting easier.'`


There are a lot of environments users may be viewing the model, even in Github browsing the code, the ability to express it as a multiline string while pydantic cleans it up is a nice property. What do you think? Maybe both your point and mine are simultaneously valid answers/approaches.

tobymao:
ah i see, i think i misunderstood. doesn't the current behavior make more sense because we preserve the original user formatting, whereas if we do what you're saying, we'd lose the original formatting? we can always go to a condensed representation, but we can't go back to the original unless we keep it as.

z3z1ma:
Yeah thats fair. I think just applying `inspect.cleandoc` from the stdlib is sufficient to preserve user formatting while allowing for a more aesthetic definition. Since this is ugly.

```sql
MODEL(
  ...,
  description '''This is some
model without cleandoc
being applied''',
  grain ...
);
```


and using `inspect.cleandoc` lets you do this as exactly equivalent (hence why it exists)

```sql
MODEL(
  ...,
  description '''
  This is some
  model without cleandoc
  being applied
  ''',
  grain ...
);
```


I will drop the condensing and let you reassess @tobymao

tobymao:
@z3z1ma how does this behave when you do format model? does it mess up the formatting? i guess my point is that we can do this all in the UI, and just leave the raw data as raw, we don't need to preprocess.

z3z1ma:
Ah yeah formatting will convert the string back into a single line IIRC since thats how sqlglot's generator does it. 
I don't use the formatting feature for the time being so that hasn't affected me. 


> just leave the raw data as raw, we don't need to preprocess.

Okay np, we can close it then. If users are looking at a model outside the UI though, its likely they will want to wrap it themselves. So maybe a step like this can be deferred to when you are leveraging the raw data like in https://github.com/TobikoData/sqlmesh/pull/1878

PR #1943, opened 2024-01-09
Fix: Only drop a materialized view if it's supported by the target engine

izeigerman:


docs/concepts/models/model_kinds.md
```diff
@@ -288,7 +288,6 @@ MODEL (
 
 * BigQuery
 * Databricks
-* Postgres
```
eakmanrq, line 291: Thanks for catching this.

tobymao (approved):

docs/concepts/models/model_kinds.md
izeigerman: Wasn't me unfortunately :) This was reported by a user in Slack

PR #1947, opened 2024-01-09
Fix: Send sse error if file watcher throws

mykalmax:
When saving files and file watcher catch error , it should send SSE event and continue collecting changes. And eventually it should send SSE with changes back to update UI.

### DIFF
--- a/web/server/api/endpoints/files.py
+++ b/web/server/api/endpoints/files.py
@@ -46,21 +46,22 @@ def get_file(
     """Get a file, including its contents."""
     try:
         file_path = Path(path)
-        file = _get_file_with_content(file_path, settings)
+        file = _get_file_with_content(settings.project_path / file_path, str(file_path))
     except FileNotFoundError:
         raise HTTPException(status_code=HTTP_404_NOT_FOUND)
 
     return file
 
 
-@router.post("/{path:path}", response_model=models.File)
+@router.post("/{path:path}", response_model=t.Optional[models.File])
 async def write_file(
+    response: Response,
     content: str = Body("", embed=True),
     new_path: t.Optional[str] = Body(None, embed=True),
     path: str = Depends(validate_path),
     settings: Settings = Depends(get_settings),
     context: Context = Depends(get_context),
-) -> models.File:
+) -> t.Optional[models.File]:
     """Create, update, or rename a file."""
     path_or_new_path = path
     if new_path:
@@ -86,12 +87,11 @@ async def write_file(
                     ServerSentEvent(event="errors", data=json.dumps(error))
                 )
 
-        full_path.write_text(content, encoding="utf-8")
+        full_path.write_text(content)
 
-    content = (settings.project_path / path_or_new_path).read_text()
-    return models.File(
-        name=os.path.basename(path_or_new_path), path=path_or_new_path, content=content
-    )
+    response.status_code = HTTP_204_NO_CONTENT
+
+    return None
 
 
 @router.delete("/{path:path}")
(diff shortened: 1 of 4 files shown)


web/server/settings.py
```diff
@@ -81,10 +81,9 @@ async def get_loaded_context(settings: Settings = Depends(get_settings)) -> Cont
         )
 
 
-async def get_context(settings: Settings = Depends(get_settings)) -> Context:
+def get_context(settings: Settings = Depends(get_settings)) -> Context:
     try:
-        async with get_context_lock:
```
vchan, line 86: I think this was added because multiple requests were asking for a context at the same time and simultaneous context creations wasn't supported. Is this not the case anymore? Also, have you checked if this works on smaller projects, e.g. the sushi example project?

web/server/api/endpoints/files.py
```diff
@@ -162,18 +187,16 @@ def walk_path(
...
-    settings: Settings,
-) -> models.File:
+def _get_file_with_content(file_path: Path, path: str) -> models.File:
     """Get a file, including its contents."""
-    file_path = settings.project_path / path
-
-    with open(file_path) as f:
-        content = f.read()
+    try:
+        content = str(file_path.read_text())
```
vchan, line 193: Do you still need to convert to a string?

web/server/api/endpoints/files.py
```diff
@@ -162,18 +187,16 @@ def walk_path(
...
+    except Exception:
+        print(f"Unable to read file: {file_path}")
+        content = None
```
vchan, line 196: Shouldn't we let the error bubble up?

web/server/api/endpoints/files.py
```diff
@@ -162,18 +187,16 @@ def walk_path(
     )
 
 
-def _get_file_with_content(
-    path: Path,
-    settings: Settings,
-) -> models.File:
+def _get_file_with_content(file_path: Path, path: str) -> models.File:
```
vchan, line 190: The parameters are very similar. Is it because you don't want to pass in the settings object? Then how about `project_path: Path` and `relative_path: str`?

web/server/api/endpoints/files.py
```diff
@@ -46,21 +46,22 @@ def get_file(
...
+@router.post("/{path:path}", response_model=t.Optional[models.File])
 async def write_file(
+    response: Response,
     content: str = Body("", embed=True),
     new_path: t.Optional[str] = Body(None, embed=True),
     path: str = Depends(validate_path),
     settings: Settings = Depends(get_settings),
     context: Context = Depends(get_context),
-) -> models.File:
+) -> t.Optional[models.File]:
```
vchan, line 64: This can no longer return a File object. Can the return type be just None?

vchan (approved):

web/server/watcher.py
```diff
@@ -40,31 +40,46 @@ async def watch_project() -> None:
...
+                error = ApiException(
+                    message="Error updating file",
+                    origin=f"API -> watcher -> watch_project",
+                    trigger=path_str,
+                ).to_dict()
+                api_console.queue.put_nowait(
+                    ServerSentEvent(event="errors", data=json.dumps(error))
                 )
 
+                continue
```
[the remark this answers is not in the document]
mykalmax: my try catch is inside `for loop` , i want to continue iteration in case of error

web/server/api/endpoints/files.py
mykalmax: `File` may not be picked up by openapi to convert to typescript types if it is not exported

PR #1949, opened 2024-01-10
Fix: pass dialect to name field value converter

z3z1ma:
External models are written like this:

image

But they fail when the method render definition is called due to the meta converter lookup not having a dialect argument. 

image

---

This PR fixes that with what is likely the correct behavior

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -178,6 +178,13 @@ class _Model(ModelMeta, frozen=True):
                             value=field_value.to_expression(dialect=self.dialect),
                         )
                     )
+                elif field_name == "name":
+                    expressions.append(
+                        exp.Property(
+                            this=field_info.alias or field_name,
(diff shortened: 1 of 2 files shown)


izeigerman:
@z3z1ma This looks really great, thank you 👍  Any chance you can add a small test for this? Good to go otherwise

sqlmesh/core/model/definition.py
```diff
@@ -178,6 +178,13 @@ def render_definition(self, include_python: bool = True) -> t.List[exp.Expressio
                             value=field_value.to_expression(dialect=self.dialect),
                         )
                     )
+                elif field_name == "name":
+                    expressions.append(
+                        exp.Property(
+                            this=field_info.alias or field_name,
```
tobymao, line 184: i think this can just be field_name

eakmanrq (approved):
Thanks for this @z3z1ma!

PR #1952, opened 2024-01-10
chore: don't parallelize tests by default

eakmanrq:


tobymao (approved):

izeigerman:
Was going to suggest exactly this, but you're way ahead of me. Thanks!

PR #1953, opened 2024-01-10
fix!: ensure deps are not functions closes #1945

### LINKED ISSUE #1945: Incomplete test error since v0.63.0
# Issue description

Let's say we have a model defined as follows:
```sql
SELECT
  r.range::DATE AS month
FROM
  RANGE(DATE '2023-01-01', DATE '2023-04-01', INTERVAL 1 MONTH) AS r
```
And a unit test for this model defined as follows:
```yaml
test_xyz:
  model: xyz
  outputs:
    query:
      rows:
        - month: 2023-01-01
        - month: 2023-02-01
        - month: 2023-03-01
```

When running this test with v0.62.1, everything works fine.

When running it with v0.63.0, I get the following error:
```bash
Error: Incomplete test, missing input for table RANGE(CAST('2023-01-01' AS DATE), CAST('2023-04-01' AS DATE), INTERVAL '1' MONTH) at [...]
```

tobymao:


izeigerman (approved):

tests/core/test_test.py
```diff
@@ -496,3 +496,34 @@ def test_test_generation(tmp_path: Path) -> None:
     test = load_yaml(context.path / c.TESTS / "foo/bar.yaml")
     assert len(test) == 1
     assert "new_name" in test
+
+
+def test_foo() -> None:
```
georgesittas, line 501: This isn't a very descriptive name

tests/core/test_test.py
```diff
@@ -496,3 +496,34 @@ def test_test_generation(tmp_path: Path) -> None:
...
+            """
+        ),
+        models={},  # type: ignore
+        engine_adapter=Context(
+            config=Config(model_defaults=ModelDefaultsConfig(dialect="duckdb"))
+        )._test_engine_adapter,
+        dialect="duckdb",
+        path=None,
+    ).run()
+    raise
```
georgesittas, line 527: leftover?

georgesittas (approved):

tests/core/test_test.py
```diff
@@ -496,3 +496,33 @@ def test_test_generation(tmp_path: Path) -> None:
     test = load_yaml(context.path / c.TESTS / "foo/bar.yaml")
     assert len(test) == 1
     assert "new_name" in test
+
+
+def test_source_func() -> None:
+    result = SqlModelTest(
```
georgesittas, line 502: fyi could also use the `_create_test` helper if you wanted to refactor this I think

georgesittas: i guess check out the other test methods, iirc there's a pattern to how these are tested with the helpers

PR #1957, opened 2024-01-11
fix: ignore external models during cleanup

eakmanrq:


sqlmesh/core/state_sync/common.py
```diff
@@ -46,7 +46,7 @@ def cleanup_expired_views(adapter: EngineAdapter, environments: t.List[Environme
         )
         for environment in expired_schema_environments
         for snapshot in environment.snapshots
-        if snapshot.is_model
+        if snapshot.is_model and not snapshot.is_external
```
tobymao, line 49: i think this should be not is_symbolic, symbolic being embedded + external

tests/core/test_context.py
```diff
@@ -416,32 +416,26 @@ def test_janitor(sushi_context, mocker: MockerFixture) -> None:
...
                 cascade=True,
                 ignore_if_not_exists=True,
             ),
-        ],
-        any_order=True,
+        ]
     )
     # Assert that the views are dropped for each snapshot just once and make sure that the name used is the
     # view name with the environment as a suffix
-    # TODO: It appears we try to drop views for external models which shouldn't hurt but we should fix this
```
tobymao, line 434: interesting comment. glad we have this fixed now

tobymao (approved):

PR #1960, opened 2024-01-11
feat: add limited dbt snapshot support

eakmanrq:
Adds limited support for dbt snapshots with those limitations/behavioral difference documented.

Previous scd_type_2 test was moved to be a dbt test that compares running dbt runtime, dbt adapter, and sqlmesh to make sure they all return the same result (except for documented behavioral difference). 

Follow up PRs will add full snapshot support.

### DIFF
--- a/sqlmesh/core/model/kind.py
+++ b/sqlmesh/core/model/kind.py
@@ -327,6 +327,7 @@ class SCDType2Kind(_ModelKind):
     valid_from_name: SQLGlotString = "valid_from"
     valid_to_name: SQLGlotString = "valid_to"
     updated_at_name: SQLGlotString = "updated_at"
+    updated_at_as_valid_from: SQLGlotBool = False
 
     forward_only: SQLGlotBool = True
     disable_restatement: SQLGlotBool = True
--- a/sqlmesh/dbt/manifest.py
+++ b/sqlmesh/dbt/manifest.py
@@ -193,8 +193,21 @@ class ManifestHelper:
 
     def _load_models_and_seeds(self) -> None:
         for node in self._manifest.nodes.values():
-            if node.resource_type not in ("model", "seed"):
+            if node.resource_type not in ("model", "seed", "snapshot"):
                 continue
+
+            if node.resource_type == "snapshot":
+                config = _config(node)
+                if config.get("strategy", "timestamp") != "timestamp":
+                    logging.warning(
+                        f"{node.relation_name}: SQLMesh only supports timestamp snapshot strategy. Skipping model."
+                    )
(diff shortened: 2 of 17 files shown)


sqlmesh/core/model/kind.py
```diff
@@ -327,6 +327,7 @@ class SCDType2Kind(_ModelKind):
     valid_from_name: SQLGlotString = "valid_from"
     valid_to_name: SQLGlotString = "valid_to"
     updated_at_name: SQLGlotString = "updated_at"
+    use_updated_at_as_valid_from: SQLGlotBool = False
```
tobymao, line 330: what about just 

`updated_at_as_valid_from`

sqlmesh/dbt/manifest.py
```diff
@@ -193,17 +193,30 @@ def _load_tests(self) -> None:
 
     def _load_models_and_seeds(self) -> None:
         for node in self._manifest.nodes.values():
-            if node.resource_type not in ("model", "seed"):
+            if node.resource_type not in ("model", "seed", "snapshot"):
                 continue
 
+            if node.resource_type == "snapshot":
+                config = _config(node)
+                if config.get("strategy", "timestamp") != "timestamp":
```
tobymao, line 201: in dbt, is None == timestamp?

tests/dbt/test_integration.py
```diff
@@ -0,0 +1,347 @@
...
+        return self.is_sqlmesh or self.is_dbt_adapter
+
+
+@pytest.fixture(params=list(TestType))
+def test_type(request):
+    return request.param
+
+
+@pytest.fixture
+def create_scd_type_2_dbt_project(tmp_path):
```
tobymao, line 55: Could we just add an scd type 2 model in the the dbt sushi project? I'd rather have these big model types in those examples because they're run all over the place.

In general, I'm afraid of the maintainability of these custom integration tests, which is why I lean towards adding these directly to the test projects.

sqlmesh/dbt/manifest.py
eakmanrq: According to current documentation strategy is required. Not sure if this changed though over time. It is also recommended. So, although I think it should always be there, if some old version didn't require it then it would default to timestamp.

tests/dbt/test_integration.py
eakmanrq: The test dbt sushi project does have an scd type 2 model. It actually has 3 of them (to verify what does or does not work). 

Lets discuss the maintainability concern.

tobymao (approved):

PR #1968, opened 2024-01-12
Feat: Send sse on file formatting

mykalmax:


web/server/models.py
```diff
@@ -28,6 +28,21 @@
 SUPPORTED_EXTENSIONS = {".py", ".sql", ".yaml", ".yml", ".csv"}
 
 
+class EventName(str, enum.Enum):
+    """An enumeration of possible SSE names."""
+
+    PING = "ping"
+    ERRORS = "errors"
+    WARNINGS = "warning"
```
vchan, line 36: Should this be warnings?

web/server/models.py
```diff
@@ -554,3 +560,8 @@ class PlanApplyStageTracker(PlanStageTracker):
 
 class PlanCancelStageTracker(PlanStageTracker):
     cancel: t.Optional[PlanStageCancel] = None
+
+
+class FormatFileStatus(BaseModel):
+    path: str
+    status: Status = Status.SUCCESS
```
vchan, line 567: Nit: Does a status of INIT mean anything for formatting a file?

web/client/src/library/components/editor/EditorFooter.tsx
```diff
@@ -59,6 +59,14 @@ export default function EditorFooter({ tab }: { tab: EditorTab }): JSX.Element {
           <EditorIndicator.Light ok={tab.isSaved} />
         </EditorIndicator>
       )}
+      {tab.file.isSQL && (
+        <EditorIndicator
+          className="mr-2"
+          text="Format"
```
vchan, line 65: Should this be Formatted?

vchan (approved):

web/server/models.py
mykalmax: It is possible to have `indeterminate` state, where it is not yet clean if file has been formatted not (sinse we are not storing previous state for files)

vchan: Should the default be INIT then?

vchan (approved):

PR #1974, opened 2024-01-14
fix!: allow columns to types to have spaces

tobymao:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -1494,6 +1494,10 @@ def _truncate_table(self, table_name: TableName) -> None:
         table = exp.to_table(table_name)
         self.execute(f"TRUNCATE TABLE {table.sql(dialect=self.dialect, identify=True)}")
 
+    @classmethod
+    def _select_columns(cls, columns: t.Iterable[str]) -> exp.Select:
+        return exp.select(*(f'"{c}"' for c in columns))
```
izeigerman, line 1499: Are these identifier quotes? Shouldn't they be dialect specific?

tobymao: no because the context that these are used are sqlglot dialect

eakmanrq:
Since the fix is just applied at the select expressions, I'm wondering if there are other places that we use `columns_to_types` that doesn't properly support spaces. Can you make sure that pandas handles it fine without quotes? Also these are some other places to think about:

https://github.com/TobikoData/sqlmesh/blob/6dc113a4e49b3ab6ea0dc22cefdcccdc179a64bf/sqlmesh/core/engine_adapter/duckdb.py#L43
https://github.com/TobikoData/sqlmesh/blob/6dc113a4e49b3ab6ea0dc22cefdcccdc179a64bf/sqlmesh/core/engine_adapter/bigquery.py#L266
https://github.com/TobikoData/sqlmesh/blob/6dc113a4e49b3ab6ea0dc22cefdcccdc179a64bf/sqlmesh/core/engine_adapter/mixins.py#L66
https://github.com/TobikoData/sqlmesh/blob/6dc113a4e49b3ab6ea0dc22cefdcccdc179a64bf/sqlmesh/core/engine_adapter/spark.py#L163

PR #1975, opened 2024-01-14
fix: share connection config state sync instead of adapters

eakmanrq:
Previously we shared actual engine adapters across the data warehouse connection and state sync connection (if one isn't provided). It seems more proper to have each one have a unique connection but share a common config. That way the two connections can make changes (like changing catalog settings) independent of another.


izeigerman:
@eakmanrq did you confirm that this works when using `DuckdDB` that points to the same DB file?

eakmanrq:
> @eakmanrq did you confirm that this works when using `DuckdDB` that points to the same DB file?

Good callout and, a bit to my surprise, this appears to work fine. I created multiple environments with changes and queried the data and the state metadata using `fetchdf` and everything was fine.

izeigerman (approved):

PR #1976, opened 2024-01-16
Feat: Allow using $root in seed paths to specify path relative to context path

z3z1ma:
Depending on how you organize your sqlmesh project, you may want to declare a seed in some nested directed colocated with a relevant model file. The `../../../../` gets kludgy. Our desire is often to specify a path from the *context root*. This should be straightforward to do. Here I have implemented it via using `~` to be indicative of the module path / context root. If we don't like the overloading we could try something else but to be fair Path.expanduser is never called anyway so it never would have worked and seeds being located in a real user home dir is an anti-pattern to say the least. 

So lets provide something nicer.

EDIT: pivoted to use a special `$root` (case-insensitive) substitution to avoid confusion

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1598,9 +1598,9 @@ def create_seed_model(
             from the macro registry.
     """
     seed_path = Path(seed_kind.path)
-    base, *subdir = seed_path.parts
-    if base == "~":
-        seed_path = module_path.joinpath(*subdir)
+    base, *subdirs = seed_path.parts
+    if base.lower() == "$root":
+        seed_path = module_path.joinpath(*subdirs)
     elif not seed_path.is_absolute():
         seed_path = path / seed_path if path.is_dir() else path.parent / seed_path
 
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1598,8 +1598,11 @@ def create_seed_model(
             from the macro registry.
     """
     seed_path = Path(seed_kind.path)
-    if not seed_path.is_absolute():
-        seed_path = path / seed_path if path.is_dir() else path.parents[0] / seed_path
+    base, *subdir = seed_path.parts
+    if base == "~":
```
izeigerman, line 1602: Any chance we can use [expanduser()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.expanduser) instead?

izeigerman: Ah, sorry, this is not correct. I didn't realize we overload `~` to mean the project path. Not sure I like such implicit overloading. I feel like `~` is so universally used that substituting its meaning like can cause quite a bit of confusion.

z3z1ma: Sure, I commented the same thought. What symbol should we prefer. We need a way to indicate the context root. 

How about something longer in the flavor of bash substitution `$root/seeds/whatever.csv`

izeigerman (approved):
I like this one better, thanks!

PR #1977, opened 2024-01-17
feat: add snowflake oauth access token support

mkilpat:
https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-with-oauth


eakmanrq (commented):
Thanks @mkilpat this looks great! Really appreciate you adding support to both native and dbt adapter + tests.

Do you mind updating the documentation to explicitly list the OATH authentication method with an example? You can see below the table you updated we list out different auth methods with examples of how to use them. Just want to make sure OATH is explicitly listed there.

PR #1988, opened 2024-01-19
Feat: Add @default_catalog to the macro environment

z3z1ma:
EDIT: The original crux of the issue was to fix the below

```sql
MODEL (
  name my.model,
  kind FULL,
  dialect bigquery,
  grain id
);

SELECT * FROM upstream.table;

ALTER TABLE my.model ADD PRIMARY KEY (id) NOT ENFORCED; -- this is not mapped to snapshot
```

We have fixed this in sqlglot. Part 2 is making available default catalog. Part 3 would be properly parsing `GRANTs` into dedicated nodes? Otherwise the mapping from name -> snapshot will never work?

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -106,7 +106,6 @@ class BaseExpressionRenderer:
                     cache_key[0] if not self._only_execution_time else None,
                     cache_key[1] if not self._only_execution_time else None,
                 ),
-                "default_catalog": self._default_catalog,
                 **kwargs,
             }
 
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -106,6 +106,7 @@ class BaseExpressionRenderer:
                     cache_key[0] if not self._only_execution_time else None,
                     cache_key[1] if not self._only_execution_time else None,
                 ),
+                "default_catalog": self._default_catalog,
                 **kwargs,
             }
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/renderer.py
```diff
@@ -527,6 +527,11 @@ def _optimize_query(self, query: exp.Subqueryable, all_deps: t.Set[str]) -> exp.
 
 @contextmanager
 def _normalize_and_quote(query: E, dialect: str, default_catalog: t.Optional[str]) -> t.Iterator[E]:
+    if isinstance(query, (exp.Command, exp.AlterTable)) and default_catalog:
```
tobymao, line 530: i don't think this is the right fix. shouldn't we just fix qualify_tables in sqlglot to work with alter tables?

z3z1ma: Also `exp.Command` parses everything after as a literal, I just realized
Is there a `@default_catalog` macro var? If not, I think that might be the move here. A 2 pronged approach. Fix `qualify_tables` in sqlglot, and use `@default_catalog` for users trying to run `GRANT`s while not hardcoding a catalog
@tobymao

tobymao: makes sense, do you want to just make a PR to ensure default_catalog is exposed as a macro var? we can handle qualify_tables cc @georgesittas

z3z1ma: Sure. Also Im on the default tables thing now, just give me 2 mins. 😃 add test

z3z1ma:
https://github.com/tobymao/sqlglot/pull/2862

cc @tobymao , @georgesittas

sqlmesh/core/renderer.py
```diff
@@ -106,6 +106,7 @@ def _render(
                     cache_key[0] if not self._only_execution_time else None,
                     cache_key[1] if not self._only_execution_time else None,
                 ),
+                "default_catalog": self._default_catalog,
```
tobymao, line 109: can you refactor this a bit. add default_catalog to kwargs, and add kwargs to the cache key on line 98?

z3z1ma: yeah that was moreso just putting some pen to the paper so I have a starting point, I am not sure if there is a single source of truth "canonical" place you add variables to the render context. I also think, based on the CI error, that it might already be available to the context 🤔 

I have a feeling this might be more of a testing/documentation exercise on the sqlmesh side.

z3z1ma: So I validated that default catalog did not work and was thus not in the context. But its looks like I found the right place to add it. This should help users who need to run `GRANTs` or any other command which parses the body as a literal since I THINK at least macros will be expanded.

PR #1989, opened 2024-01-19
Fix!: Prefer the builtin is_incremental macro over the dbt one

izeigerman:


sqlmesh/dbt/manifest.py
```diff
@@ -41,7 +41,7 @@
 MacroConfigs = t.Dict[str, MacroConfig]
 
 
-BUILTIN_CALLS = {*BUILTIN_GLOBALS, *BUILTIN_FILTERS}
+BUILTIN_CALLS = {*BUILTIN_GLOBALS, *BUILTIN_FILTERS, "is_incremental"}
```
crericha, line 44: There are other bulitin calls besides `is_incremental` that are added in `create_builtin_globals` -- `flags` as one example. Is there any impact in omitting them here?

sqlmesh/dbt/builtin.py
```diff
@@ -281,6 +281,8 @@ def _try_literal_eval(value: str) -> t.Any:
     "as_text": lambda v: "" if v is None else str(v),
 }
 
+OVERRIDDEN_MACROS = {MacroReference(package="dbt", name="is_incremental")}
```
crericha, line 284: We are okay with customers overriding `is_incremental`?

sqlmesh/dbt/manifest.py
izeigerman: That's an artifact that I forgot to remove. This should not be here. Thanks for catching.

crericha (approved):

PR #1990, opened 2024-01-19
Feat: Add get model endpoint.

vchan:
Also make query rendering optional.

### DIFF
--- a/tests/web/test_main.py
+++ b/tests/web/test_main.py
@@ -430,6 +430,19 @@ def test_fetchdf(web_sushi_context: Context) -> None:
     assert not df.empty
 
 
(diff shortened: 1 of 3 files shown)


tests/web/test_main.py
```diff
@@ -430,6 +430,19 @@ def test_fetchdf(web_sushi_context: Context) -> None:
...
 
 
+def test_get_model(web_sushi_context: Context) -> None:
+    response = client.get("/api/models/sushi.customers")
+
+    assert response.status_code == 200
+
+    test_model = response.json()
+
+    assert test_model.get("name")
```
tobymao, line 440: why not test some equivalence

PR #1991, opened 2024-01-19
Feat: Only promote selected models in dev when selector is used

izeigerman:


sqlmesh/core/plan/definition.py
```diff
@@ -168,13 +168,23 @@ def environment(self) -> Environment:
             else None
         )
 
+        snapshots_by_name = self.context_diff.snapshots_by_name
         snapshots = [s.table_info for s in self.snapshots.values()]
         promoted_snapshot_ids = None
         if self.is_dev and not self.include_unmodified:
+            promotable_snapshot_ids = self.context_diff.promotable_snapshot_ids.copy()
+            if self.models_to_backfill is not None:
```
tobymao, line 176:
```suggestion
            if self.models_to_backfill:
```

tobymao (approved):

sqlmesh/core/plan/definition.py
izeigerman: Empty set would mean no models to backfill and hence nothing to promote

tobymao: ah woops, you reference previous_promoted_snapshot_ids as well,

PR #1993, opened 2024-01-20
Chore: Add groupby filter to utils package

crericha:


sqlmesh/utils/filters.py
```diff
@@ -0,0 +1,15 @@
+import typing as t
+from collections import defaultdict
+
+ITEM = t.TypeVar("ITEM")
+GROUP = t.TypeVar("GROUP")
+
+
+def groupby(
```
tobymao, line 8: i'd just put it in __init__ until we have more in this category

PR #1994, opened 2024-01-20
Render handles for Model Column only when that column has been select

mykalmax:
Render Lineage node handles for Model Column only when that column has been selected to build column level lineage. Rendering handles (blue circle on screenshot)  for all nodes with columns upfront slows down rendering lineage and creates unnecessary html elements
Screenshot 2024-01-19 at 5 17 35 PM

Also, show columns for nodes when `withColumns: false`:
 - when node selected
 - when node has selected columns
 - when hovered over node
 - when main node (selected model)

### DIFF
--- a/web/client/src/library/components/graph/ModelLineage.tsx
+++ b/web/client/src/library/components/graph/ModelLineage.tsx
@@ -209,7 +209,6 @@ function ModelColumnLineage(): JSX.Element {
       withConnected,
       withImpacted,
       withSecondary,
-      withColumns,
     )
     const newEdges = getUpdatedEdges(
       allEdges,
@@ -253,8 +252,6 @@ function ModelColumnLineage(): JSX.Element {
             setIsBuildingLayout(false)
           }, 100)
         }
-
-        setActiveNodes(newActiveNodes)
       })
 
     return () => {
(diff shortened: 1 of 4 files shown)


web/client/src/library/components/graph/ModelLineage.tsx
```diff
@@ -265,17 +262,6 @@ function ModelColumnLineage(): JSX.Element {
     }
   }, [lineageIndex, withColumns])
 
-  useEffect(() => {
```
mykalmax, line 268: redundant sinse will be updated in next useEffect anyways

web/client/src/library/components/graph/help.ts
```diff
@@ -664,10 +665,9 @@ function getUpdatedNodes(
   connectedNodes: Set<string>,
   selectedNodes: Set<string>,
   connections: Map<string, Connections>,
-  withConnected: boolean = true,
-  withImpacted: boolean = true,
-  withSecondary: boolean = true,
-  withColumns: boolean = true,
```
mykalmax, line 670: `withColumns` triggers rebuilding and repositioning of the lineage sinse node with columns has different heigh, so no need for additionally update

vchan (approved):

PR #1995, opened 2024-01-20
Feat: add a debugger friendly console mode

z3z1ma:
Debugging in the terminal using something like pdb is a nightmare. Lets make it nicer. 

Setting `SQLMESH_DEBUG=1 sqlmesh ...` will make our lives much easier!

### DIFF
--- a/sqlmesh/__init__.py
+++ b/sqlmesh/__init__.py
@@ -57,7 +57,7 @@ class RuntimeEnv(str, Enum):
         except NameError:
             pass
 
-        if os.getenv("SQLMESH_PDB"):
+        if debug_mode_enabled():
             return RuntimeEnv.DEBUGGER
         return RuntimeEnv.TERMINAL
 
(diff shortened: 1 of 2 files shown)


z3z1ma:
Simplified output

image

---

And with `pdb`, can actually see whats going on

image

georgesittas:
This is great @z3z1ma! 🙌

sqlmesh/__init__.py
```diff
@@ -56,6 +57,8 @@ def get(cls) -> RuntimeEnv:
         except NameError:
             pass
 
+        if os.getenv("SQLMESH_PDB"):
```
izeigerman, line 60: Instead of introducing a new environment variable should we consider reusing an existing [debug mode](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/__init__.py#L219-L220)?

z3z1ma: I thought about that, my only thought is that debug mode is common for your end users to gather logs. Do we want the simplified console output in tandem; is that fine or surprising behavior? The real beneficiary is for interactive tools like pdb, hence `SQLMESH_PDB`. Buttt I dont feel strongly. I can just use the existing debug.

z3z1ma: I suppose the output is actually more informative than the progress bars so it is probably fitting for SQLMESH_DEBUG in general

izeigerman (approved):

PR #1996, opened 2024-01-20
Fix: expand seed $root marker in constructor

z3z1ma:
We must "expand" any markers (`$root` is the only one currently) eagerly in the SeedModel constructor. (vs currently just doing it for the sake of the `Seed` container.

Main reason is so the [_track_file](https://github.com/z3z1ma/sqlmesh/blob/284990baa4fe72310f5beeef35f9201cc6e2f107/sqlmesh/core/loader.py#L311-L313) call here does not throw (not that anyone is using this feature just yet), but also looking forward so `SeedModel.seed_path` is correct without any additional work and  the SeedKind path makes sense.

I may slip in a documentation update in this PR too mentioning the marker in the SeedKind section

### DIFF
--- a/docs/concepts/models/seed_models.md
+++ b/docs/concepts/models/seed_models.md
@@ -28,7 +28,7 @@ MODEL (
   )
 );
 ```
-The `path` attribute contains the path to the seed's CSV file **relative** to the path of the model's `.sql` file.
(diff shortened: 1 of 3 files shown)


izeigerman (approved):
Thanks!

PR #1997, opened 2024-01-21
feat: Add SCD Type 2 Column Support

eakmanrq:
Prior to this PR you could only create an SCD Type 2 table if you had an "Updated At" timestamp in the source table. This PR makes it so that you can create an SCD Type 2 from any source by checking if specific columns have changed. Since you no longer have "Updated At" to tell you when that change was made it uses `execution_time` instead. As a result you can think of it as a less precise approach to SCD Type 2.

This PR includes support both for native SQLMesh and dbt adapter. 

One challenge when testing the dbt runtime is that dbt doesn't allow freezing "now()" (their execution time). Before I had a simple way of patching their "now()" with the frozen now but it had a bug and I fixed it in this PR. This bug is actually what created the perceived behavior difference between dbt and SQLMesh so now they actually appear to behave the same.

I decided to leave the `SCD_TYPE_2` model kind in place and alias it to be `SCD_TYPE_2_BY_TIME` and this is what this PR is not a breaking change. To me this seems fair since by time is the recommended and default approach so having the unqualified version f the name point to what we recommend seems fine. If others disagree then I can remove and make this a breaking change.


plaflamme:
@eakmanrq we have a slightly similar use-case, but where the "updated at" timestamp could still be inferred from the source.

The source table is a daily snapshot of the raw data that changes over time. It has a "snapshot_date" timestamp, this column doesn't tell you if the record changed (so it doesn't fulfill the `updated_at` semantics), but it could be used as the resulting `valid_from / valid_to` column values. For example:

```
name,price,snapshot_date
foo,20,2024-01-01
foo,20,2024-01-02
foo,21,2024-01-03
foo,21,2024-01-04
food,21,2024-01-05
```

We can build an SCD type 2 from this and use `snapshot_date` instead of `execution_time` (if I understand correctly), the expected result would be

```
name,price,valid_from,valid_to
foo,20,2024-01-01,2024-01-02
foo,21,2024-01-03,2024-01-04
food,21,2024-01-05,NULL
```

Would it make sense to add support for this case out of the box? Perhaps this is already supported by the new kind?

eakmanrq:
> @eakmanrq we have a slightly similar use-case, but where the "updated at" timestamp could still be inferred from the source.
> 
> The source table is a daily snapshot of the raw data that changes over time. It has a "snapshot_date" timestamp, this column doesn't tell you if the record changed (so it doesn't fulfill the `updated_at` semantics), but it could be used as the resulting `valid_from / valid_to` column values. For example:
> 
> ```
> name,price,snapshot_date
> foo,20,2024-01-01
> foo,20,2024-01-02
> foo,21,2024-01-03
> foo,21,2024-01-04
> food,21,2024-01-05
> ```
> 
> We can build an SCD type 2 from this and use `snapshot_date` instead of `execution_time` (if I understand correctly), the expected result would be
> 
> ```
> name,price,valid_from,valid_to
> foo,20,2024-01-01,2024-01-02
> foo,21,2024-01-03,2024-01-04
> food,21,2024-01-05,NULL
> ```
> 
> Would it make sense to add support for this case out of the box? Perhaps this is already supported by the new kind?

Thanks @plaflamme for sharing this use case. Basically you are wanting to create a SCD Type 2 table out of a snapshot table and that use case makes sense. There is already a bool which is `updated_at_as_valid_from` but that is exclusive to `SCD_TYPE_2_BY_TIME`. Conceptually this makes sense since if you have an `updated_at` then you would just use `BY_TIME` but you have a good example here where you want it `BY_COLUMN` but still use an `updated_at` value. 

I like the idea of adding support for this but leaning towards adding support in another PR. The reason is that this one is already complex enough. It is a cool use case though and I would like to see support added.

plaflamme:
@eakmanrq great! glad to hear this might be a useful addition.

Another PR sounds like a good idea. I'll open a separate issue so it this can be tracked separately, feel free to close it if it's not useful.

sqlmesh/core/model/kind.py
```diff
@@ -341,6 +348,24 @@ def managed_columns(self) -> t.Dict[str, exp.DataType]:
...
+class SCDType2ByTimeKind(_SCDType2Kind):
+    name: Literal[ModelKindName.SCD_TYPE_2_BY_TIME] = ModelKindName.SCD_TYPE_2_BY_TIME
+    updated_at_name: SQLGlotString = "updated_at"
+    updated_at_as_valid_from: SQLGlotBool = False
+
+
+# Legacy alias to original SCD Type 2 Kind that existed prior to adding by column support
+# Only used in discriminator for Pydantic
+class SCDType2Kind(SCDType2ByTimeKind):
+    name: Literal[ModelKindName.SCD_TYPE_2] = ModelKindName.SCD_TYPE_2_BY_TIME  # type: ignore
```
izeigerman, line 360: Why don't the literal type and the value match?

tests/conftest.py
```diff
@@ -368,3 +369,20 @@ def _make_function(table_name: str, random_id: str) -> exp.Table:
         return temp_table
 
     return _make_function
+
+
+def create_df(data: t.Sequence[t.Tuple], schema: t.Dict[str, str]) -> pd.DataFrame:
```
izeigerman, line 374: This is probably minor but I'd suggest reserving `conftest.py` only for fixtures, plugins, hooks and other pytest-specific mechanisms. I'd suggest creating a separate test utils module for this.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1145,18 +1204,102 @@ def scd_type_2(
...
-            updated_at_name
-            if updated_at_as_valid_from
-            else self._to_utc_timestamp("1970-01-01 00:00:00+00:00")
-        )
+        # If a star is provided, we include all unmanaged columns in the check.
+        # This unnecessarily includes unique key columns but since they are used in the join, and therefore we know
+        # they are equal or not, the extra check is not a problem and we gain simplified logic here.
+        # If we want to change this, then we just need to check the expressions in unique_key and pull out the
+        # column names and then remove them from the unmanaged_columns
+        if check_columns and check_columns[0] == exp.Star():
```
tobymao, line 1234: check_columns[0].is_star ?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1145,18 +1204,102 @@ def scd_type_2(
...
+                    "Cannot use `updated_at_as_valid_from` without `updated_at_name` for SCD Type 2"
+                )
+            update_valid_from_start = updated_at_name
+        elif execution_time_as_valid_from:
+            update_valid_from_start = execution_ts
+        else:
+            update_valid_from_start = self._to_utc_timestamp("1970-01-01 00:00:00+00:00")
+        insert_valid_from_start = execution_ts if check_columns else updated_at_name
+        if check_columns:
+            updated_row_filter = " OR ".join(
```
tobymao, line 1250: could be safer to use sqlglot builder instead of raw string manipulation, in case col has bad identifiers in them

```python
col.neq(tcol).or_(tcol.is(exp.null()).and_(col.is(exp.null()).not_())
for col in check_columns

File: sqlmesh/core/model/kind.py (hunk 1 above)
Comment (eakmanrq) [reply sqlmesh/core/model/kind.py]: Thanks for your feedback and helping me to dig deeper into this. I removed what we talked about (including this line) but then tested actually loading SCD_TYPE_2 from state. This commit fixes the issues identified: https://github.com/TobikoData/sqlmesh/pull/1997/commits/9a76e6fdf610492c6c004e13356548850ca1bbb3

Btw the reason why I left the default values at all is because it makes it easy in testing to do `FullKind()` and let the default put in full. We could always create a fixture that properly constructs models with the name included but it felt like I was going down the rabbit hole a bit and stopped here.

File: sqlmesh/core/engine_adapter/base.py (hunk 4 above)
Comment (eakmanrq) [reply sqlmesh/core/engine_adapter/base.py]: Addressed here: https://github.com/TobikoData/sqlmesh/pull/1997/commits/b61e09420ae8676fca71433b5a6fe60db1845adb

File: sqlmesh/core/engine_adapter/base.py [hunk 5]
```diff
@@ -1187,23 +1246,138 @@ def scd_type_2(
...
+            )
+        insert_valid_from_start = execution_ts if check_columns else updated_at_name
+        if check_columns:
+            row_check_conditions = []
+            for col in check_columns:
+                t_col = col.copy()
+                if isinstance(col, exp.Column):
+                    t_col.this.set("this", f"t_{col.name}")
+                else:
+                    t_col.set("this", f"t_{col.name}")
```
eakmanrq, line 1309: @tobymao What is the best way to implement this?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1187,23 +1246,132 @@ def scd_type_2(
...
         ]
-        select_source_columns.append(exp.cast(updated_at_name, time_data_type).as_(updated_at_name))
-        valid_from_start = (
-            updated_at_name
-            if updated_at_as_valid_from
-            else self._to_utc_timestamp("1970-01-01 00:00:00+00:00", time_data_type)
-        )
+        if updated_at_name:
+            select_source_columns.append(
+                exp.cast(updated_at_name, time_data_type).as_(updated_at_name)
```
tobymao, line 1277: time_data_type is a actual exp.DataType?

sqlmesh/core/engine_adapter/base.py
tobymao: this wasn't resolved?

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yes it is

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1187,23 +1246,131 @@ def scd_type_2(
...
+            update_valid_from_start = execution_ts
+        else:
+            update_valid_from_start = self._to_utc_timestamp(
+                "1970-01-01 00:00:00+00:00", time_data_type
+            )
+        insert_valid_from_start = execution_ts if check_columns else updated_at_name
+        if check_columns:
+            row_check_conditions = []
+            for col in check_columns:
+                t_col = exp.to_column(col.copy())
```
tobymao, line 1305: why are you calling to_column here? aren't they already columns?

tobymao (commented):
z

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1187,23 +1246,131 @@ def scd_type_2(
...
+                exp.cast(updated_at_name, time_data_type).as_(updated_at_name)
+            )
+
+        # If a star is provided, we include all unmanaged columns in the check.
+        # This unnecessarily includes unique key columns but since they are used in the join, and therefore we know
+        # they are equal or not, the extra check is not a problem and we gain simplified logic here.
+        # If we want to change this, then we just need to check the expressions in unique_key and pull out the
+        # column names and then remove them from the unmanaged_columns
+        if check_columns and check_columns[0] == exp.Star():
+            check_columns = [exp.to_column(col) for col in unmanaged_columns]
```
tobymao, line 1286: this to_column conversion isn't necessary

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yep just making sure since the type is `exp.Expression` and not `exp.Column`. It follows the same pattern as `unique_key`.

tobymao: yea, i wouldn't do that. instead, use column.assert_is(exp.Column), or even better switch the type in the function

sqlmesh/core/engine_adapter/base.py
eakmanrq: `unmanaged_columns` is a list of strings and `check_columns` is a list of expression (or columns after update) so it does need this.

tobymao: change the type of check_columns to exp.Column

eakmanrq: That is what I said here: "is a list of expression (or columns after update)"

So we are agreeing this is needed right?

tobymao: the call exp.column(col) instead

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1356,11 +1504,11 @@ def scd_type_2(
                     "inserted_rows",
                     exp.select(
                         *unmanaged_columns,
-                        f"{updated_at_name} as {valid_from_name}",
+                        f"{insert_valid_from_start} as {valid_from_name}",
```
tobymao, line 1507: exp.column(insert_valid_start).as_(valid_from_name)

sqlmesh/utils/pydantic.py
```diff
@@ -262,18 +265,83 @@ def positive_int_validator(v: t.Any) -> int:
     return v
 
 
+def _get_expressions(
+    v: t.Any,
+    values: t.Union[t.Dict, ValidationInfo],
+    coerce_func: t.Optional[t.Callable[[t.Any], t.Any]] = None,
+) -> t.List[exp.Expression]:
```
tobymao, line 272: is this newly added? why did these get added?

sqlmesh/utils/pydantic.py
```diff
@@ -262,18 +265,83 @@ def positive_int_validator(v: t.Any) -> int:
...
+        expressions = [
+            parse_one(entry, dialect=dialect) if isinstance(entry, str) else entry
+            for entry in ensure_list(v)
+        ]
+
+    results = []
+
+    for expr in expressions:
+        expr = normalize_identifiers(
+            exp.column(expr) if isinstance(expr, exp.Identifier) else coerce_func(expr),
```
tobymao, line 291: why did we add this all of a sudden?

sqlmesh/utils/pydantic.py
eakmanrq: Working on splitting out validators for creating Expressions and Columns

sqlmesh/utils/pydantic.py
eakmanrq: Working on splitting out validators for creating Expressions and Columns

sqlmesh/core/engine_adapter/base.py
tobymao: if it's a string (unmanaged_columns), then you should use exp.column, but if it's a column, you don't need it and should just fix the type checking

sqlmesh/utils/pydantic.py
tobymao: did this need to be done in this pr? it's hard for me to re-review things with big new changes like this

sqlmesh/utils/pydantic.py
tobymao: i don't believe this is correct, we may not always want columns in any generic list of expressions

sqlmesh/utils/pydantic.py
eakmanrq: I thought this is what you wanted?

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yeah so it is a string of columns. So it sounds like `exp.to_column` should change to `exp.column`

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1319,7 +1319,7 @@ def _scd_type_2(
             update_valid_from_start = self._to_utc_timestamp(
                 "1970-01-01 00:00:00+00:00", time_data_type
             )
-        insert_valid_from_start = execution_ts if check_columns else updated_at_name
+        insert_valid_from_start = execution_ts if check_columns else exp.to_column(updated_at_name)  # type: ignore
```
tobymao, line 1322: should this use column and not to_column? what are the expected inputs of updated_at_name?

eakmanrq: It is just a string representing the column name so yeah `exp.column` seems like a better fit

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+
+        # If a star is provided, we include all unmanaged columns in the check.
+        # This unnecessarily includes unique key columns but since they are used in the join, and therefore we know
+        # they are equal or not, the extra check is not a problem and we gain simplified logic here.
+        # If we want to change this, then we just need to check the expressions in unique_key and pull out the
+        # column names and then remove them from the unmanaged_columns
+        if check_columns and check_columns == exp.Star():
+            check_columns = [exp.column(col) for col in unmanaged_columns]
+        execution_ts = self._to_utc_timestamp(to_ts(execution_time), time_data_type)
+        update_valid_from_start: t.Union[str, exp.Expression]
```
tobymao, line 1309: can you delete this declaration and move the type into line 1315 update_valid_from_start: t.Union[...] = updated_at_name

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+                    "Cannot use `updated_at_as_valid_from` without `updated_at_name` for SCD Type 2"
+                )
+            update_valid_from_start = updated_at_name
+        elif execution_time_as_valid_from:
+            update_valid_from_start = execution_ts
+        else:
+            update_valid_from_start = self._to_utc_timestamp(
+                "1970-01-01 00:00:00+00:00", time_data_type
+            )
+        insert_valid_from_start = execution_ts if check_columns else exp.to_column(updated_at_name)  # type: ignore
```
tobymao, line 1322: should this be exp.column(updated_at_name) ?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+        if check_columns:
+            row_check_conditions = []
+            for col in check_columns:
+                t_col = col.copy()
+                t_col.this.set("this", f"t_{col.name}")
```
tobymao, line 1327: this is a bit tricky, could we do

t_col.set("this", exp.to_identifier(f"t_{col.name}") 

it's a bit more explicit imo, setting the idintefier this like you've done makes me feel antsy

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+                        col.neq(t_col),
+                        exp.and_(t_col.is_(exp.Null()), col.is_(exp.Null()).not_()),
+                        exp.and_(t_col.is_(exp.Null()).not_(), col.is_(exp.Null())),
+                    ]
+                )
+            row_value_check = exp.or_(*row_check_conditions)
+            unique_key_conditions = []
+            for col in unique_key:
+                t_col = col.copy()
+                t_col.this.set("this", f"t_{col.name}")
```
tobymao, line 1339: same here please

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+            valid_to_case_stmt = (
+                exp.Case()
+                .when(
+                    exp.or_(
+                        exp.column("_exists", "joined").is_(exp.Null()),
+                        updated_row_filter,
+                    ),
+                    execution_ts,
+                )
+                .else_(exp.to_column(f"t_{valid_to_name}"))
```
tobymao, line 1357: exp.column()

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+                .as_(valid_to_name)
+            )
+            valid_from_case_stmt = exp.func(
+                "COALESCE",
+                exp.column(f"t_{valid_from_name}"),
+                update_valid_from_start,
+            ).as_(valid_from_name)
+        else:
+            assert updated_at_name is not None
```
tobymao, line 1366: is this just for mypy? could be simplified to assert updated_at_name

tobymao: or why not make 1365 elif updated_at_name?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1208,23 +1267,131 @@ def scd_type_2(
...
+            updated_row_filter = exp.to_column(updated_at_name) > exp.to_column(
+                f"t_{updated_at_name}"
+            )
+            valid_to_case_stmt = (
+                exp.Case()
+                .when(updated_row_filter, exp.to_column(updated_at_name))
```
tobymao, line 1372: exp.column() everywhere

sqlmesh/core/model/kind.py
```diff
@@ -404,10 +431,14 @@ def _model_kind_validator(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) ->
         if time_data_type:
             props["time_data_type"] = exp.DataType.build(time_data_type, dialect=dialect)
         name = v.this if isinstance(v, d.ModelKind) else props.get("name")
+        # We want to ensure whatever name is provided to construct the class is the same name that will be
```
tobymao, line 434: this comment explains what, but i'm not sure it explains why? what's the reasoning behind this

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1230,11 +1236,13 @@ def create(
...
     ) -> None:
         model = snapshot.model
-        assert isinstance(model.kind, SCDType2Kind)
+        assert isinstance(model.kind, (SCDType2ByTimeKind, SCDType2ByColumnKind))
         if model.annotated:
             logger.info("Creating table '%s'", name)
             columns_to_types = model.columns_to_types_or_raise
-            columns_to_types[model.kind.updated_at_name] = model.kind.time_data_type
+            if model.kind.is_scd_type_2_by_time:
+                assert isinstance(model.kind, SCDType2ByTimeKind)
```
tobymao, line 1244: why do you need these asserts? for mypy updated_at_name?

tobymao: given how many times you have to do assert isinstance to please mypy i'm inclined to believe there's probably a better way to accomplish this. i think perhaps the problem is the reliance on these helper methods is_scd, 

if instead you just did

if isinstance(model.kind, ScdType2ByTimeKind):
   
this would be cleaner, thoughts @izeigerman ?

sqlmesh/dbt/model.py
```diff
@@ -113,6 +115,13 @@ class ModelConfig(BaseModelConfig):
     def _validate_list(cls, v: t.Union[str, t.List[str]]) -> t.List[str]:
         return ensure_list(v)
 
+    @field_validator("check_cols", mode="before")
+    @classmethod
+    def _validate_check_cols(cls, v: t.Union[str, t.List[str]]) -> t.Union[str, t.List[str]]:
+        if isinstance(v, str) and v == "all":
```
tobymao, line 121: all is a special value here? what if it's upper case? what if you have a column called all?

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yes it is for mypy and it is impossible to hit this and not have an `uppdated_at_name`. The public methods that are used to call this private method ensure that is true.

sqlmesh/dbt/model.py
tobymao: ah, this is dbt, i guess the last question doesn't hold then.

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Yes for mypy

sqlmesh/dbt/model.py
```diff
@@ -226,18 +235,28 @@ def model_kind(self, context: DbtContext) -> ModelKind:
...
                     if target.type == "bigquery"
                     else exp.DataType.build("TIMESTAMP")
                 ),
-            )
+            }
+            if self.snapshot_strategy.is_check:
+                return SCDType2ByColumnKind(
+                    columns=self.check_cols, execution_time_as_valid_from=True, **shared_kwargs
+                )
+            else:
```
tobymao, line 256: no need for else here

sqlmesh/dbt/model.py
eakmanrq: This is dbt so we didn't create this contract. We use a `*` to help avoid that issue. 

Although from the docs it seems like it is expected to be lowercase I can just do a case-insensitive check here to be sure.

sqlmesh/core/model/kind.py
eakmanrq: I was thinking the example explained why: 
```
Ex: Pass in `SCD_TYPE_2` then we want to ensure we get `SCD_TYPE_2` as the kind name instead of `SCD_TYPE_2_BY_TIME`.
```
In other words, for legacy aliases we want to ensure we maintain the legacy name instead of using the new name which is what would happen if we relied on the default value.

tobymao: yea why though? why can't SCD_TYPE_2 map to the new name? why do we need to keep the old mapping?

eakmanrq: Oh to avoid having it be detected as a change to the model when you run plan/apply. Will add that note.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Well technically possible if someone called the private method directly but impossible in practice since the evaluator only calls public methods.

sqlmesh/core/snapshot/evaluator.py
eakmanrq: Updated: https://github.com/TobikoData/sqlmesh/pull/1997/commits/97aa0d5ab6d30dbd9ffdae5c89a8ccae7e64f0dd

tobymao (approved):

PR #1999, opened 2024-01-22
Chore: Add animation to plan stages transition

mykalmax:
Animates transition between different stages when `run plan/apply plan`. Animation is very simple and need some work but makes plan UX less choppy


web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -328,10 +328,6 @@ export default function PageIDE(): JSX.Element {
   function updatePlanOverviewTracker(data: PlanOverviewTracker): void {
     planOverview.update(data)
 
-    if (planOverview.meta?.status === Status.init) {
```
mykalmax, line 331: we are manually resetting when user click `Plan` button

PR #2001, opened 2024-01-22
Feat: Add CLI flag for logging to stdout

treysp:


eakmanrq (approved):
I think this should be a "feat"? It adds new functionality for users. I think of "chore" and an internal detail that doesn't affect the user experience at all.

PR #2002, opened 2024-01-22
Feat: Fetch model's sql from API on demand

mykalmax:
`Models` API endpoint returns only model's essential details and omits `sql`. In order to get `rendered sql` we need to send another API call to `Model` API endpoint. Also when `model's` file's changed, include `rendered sql` in payload sent via SSE.

### DIFF
--- a/web/server/api/endpoints/models.py
+++ b/web/server/api/endpoints/models.py
@@ -46,8 +46,14 @@ def get_model(
     return serialize_model(model, context.path, render_query=True)
 
 
-def serialize_all_models(context: Context) -> t.List[models.Model]:
-    return [serialize_model(model, context.path) for model in context.models.values()]
+def serialize_all_models(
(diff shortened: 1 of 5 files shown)


web/server/api/endpoints/models.py
```diff
@@ -46,8 +46,14 @@ def get_model(
     return serialize_model(model, context.path, render_query=True)
 
 
-def serialize_all_models(context: Context) -> t.List[models.Model]:
-    return [serialize_model(model, context.path) for model in context.models.values()]
+def serialize_all_models(
+    context: Context, render_queries: t.Optional[t.List[str]] = None
+) -> t.List[models.Model]:
+    render_queries = render_queries or []
```
vchan, line 52: How about using a set for faster lookup?

vchan (approved):

PR #2011, opened 2024-01-24
Fix: Adjustments to new layout based on feedback

mykalmax:


web/server/models.py
```diff
@@ -44,12 +44,15 @@ class EventName(str, enum.Enum):
...
-    ERRORS = "errors"
+    EDITOR = "editor"  # include file editor and file explorer
+    DOCS = "docs"  # include docs
+    PLANS = "plans"  # include ability to run/apply plans
+    PLAN_PROGRESS = "plan-progress"  # include ability to see plan progress
+    TESTS = "tests"  # include ability to run tests
+    AUDITS = "audits"  # include ability to run audits
+    ERRORS = "errors"  # include ability to see errors
+    ENVIRONMENTS = "environments"  # include ability to see environments
+    ENVIRONMENT_PROD = "environment-prod"  # include ability to promote to prod
```
vchan, line 55: This seems less like a module and more like a permissions.

web/client/src/library/pages/root/Page.tsx
```diff
@@ -268,3 +344,397 @@ function ActionStatus({
...
+              </Menu.Items>
+            </div>
+          </Transition>
+        </>
+      )}
+    </Menu>
+  )
+}
+
+function AddEnvironemnt({
```
vchan, line 608:
```suggestion
function AddEnvironment({
```

vchan (approved):

web/server/models.py
mykalmax: I was experimenting with this idea, but you are right, it is different concept so will remove.
We can add `like a permissions` later in separate PR if we need to

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -267,13 +274,23 @@ export default function PageIDE(): JSX.Element {
...
+    if (
+      models.size > 0 &&
+      isFalse(planAction.isProcessing) &&
+      isFalse(planAction.isRunningTask)
+    ) {
+      planApply.reset()
+
       void planRun()
     }
   }, [models])
```
vchan, line 286: Is it better to have the dependency be models.size instead of models? Or it doesn't matter?

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -267,13 +274,23 @@ export default function PageIDE(): JSX.Element {
...
 
   useEffect(() => {
```
vchan, line 288: This looks very similar to the effect hook above it. Is that intentional?

web/client/src/library/pages/ide/IDE.tsx
```diff
@@ -305,9 +322,11 @@ export default function PageIDE(): JSX.Element {
       planCancel,
     })
 
-    if (isNotNil(value)) {
-      setPlanAction(new ModelPlanAction({ value }))
-    }
+    // This inconsistency can happen when we recived flag from the backend
```
vchan, line 325:
```suggestion
    // This inconsistency can happen when we receive flag from the backend
```

web/server/api/endpoints/meta.py
```diff
@@ -7,6 +11,41 @@
 router = APIRouter()
 
 
+class Mode(str, enum.Enum):
+    DEFAULT = "default"  # Allow all modules
```
vchan, line 15: Nit: How about ALL instead of DEFAULT?

vchan (approved):

web/client/src/library/pages/ide/IDE.tsx
mykalmax: having `models.size` may not catch when some `value` that `models` map holds has changed

PR #2012, opened 2024-01-24
fix: have dbt bq use timestamp instead of datetime

eakmanrq:


tests/core/engine_adapter/test_spark.py
```diff
@@ -445,7 +445,147 @@ def test_scd_type_2(
         "CREATE TABLE IF NOT EXISTS `db`.`temp_target_abcdefgh` AS SELECT `id`, `name`, `price`, `test_updated_at`, `test_valid_from`, `test_valid_to` FROM `db`.`target`",
         "TRUNCATE TABLE `db`.`target`",
         parse_one(
-            """INSERT INTO `db`.`target` (
+            """WITH `source` AS (
```
eakmanrq, line 448: I think this is just a SQLGlot change. When I first generated this SQLGlot put the with statement in CTE but that wasn't compatible with Spark 2.X so SQLGlot changed to this format. Either way I think it would get transpiled so that is why it worked before.

sqlmesh/dbt/model.py
```diff
@@ -232,6 +232,9 @@ def model_kind(self, context: DbtContext) -> ModelKind:
                 valid_to_name="dbt_valid_to",
                 updated_at_name=self.updated_at,
                 updated_at_as_valid_from=True,
+                time_data_type=exp.DataType.build("TIMESTAMPTZ")
+                if target.type == "bigquery"
+                else exp.DataType.build("TIMESTAMP"),
```
eakmanrq, line 237: Just doing this for BQ for now until we can verify what the expected behavior is for other adapters.

izeigerman (approved):

PR #2021, opened 2024-01-24
fix: model lineage section on small screens

caiters:
These use container queries so the screenshots below should reflect when this section is in either position (either the right column or on the bottom)

Large screen (should be mostly unchanged)
image

Medium (also mostly unchanged, just showing how it resizes)
image

Search collapses (closed):
image

Search collapses (open):
image

Details collapses (closed):
image

Info collapses (open):
image


web/client/src/library/components/graph/ModelLineage.tsx
```diff
@@ -346,7 +348,7 @@ function ModelColumnLineage(): JSX.Element {
       >
         <Panel
           position="top-right"
-          className="bg-theme !m-0 w-full"
+          className="bg-theme !m-0 w-full !z-10"
```
caiters, line 351: without this, ran into an issue where the lineage information popover was underneath the react flow controls when I was working on my tiny laptop screen 

before
image

after
image

web/client/src/library/components/search/SearchList.tsx
```diff
@@ -137,7 +137,7 @@ export default function SearchList<
         }
       }}
     >
-      <Popover className="relative flex">
+      <Popover className="relative flex w-full">
```
caiters, line 140: i did not see any negative effects elsewhere as a result of this, but i feel like this is the change that has the most potential to do something we don't want.

web/client/src/library/components/graph/ModelLineage.tsx
```diff
@@ -440,72 +425,27 @@ function GraphControls({ nodes = [] }: { nodes: Node[] }): JSX.Element {
...
-            <b>Data Sources</b>: {countDataSources}
-          </span>
-        )}
-        {countCTEs > 0 && (
-          <span className="mr-2 whitespace-nowrap">
-            <b>CTEs:</b> {countCTEs}
-          </span>
-        )}
+        <Popover
+          className={clsx('flex @lg:hidden bg-none border-none px-2 py-1')}
```
vchan, line 431: Do we need clsx here?

vchan (approved):

web/client/src/library/components/graph/ModelLineage.tsx
caiters: ahah, no, must have added it and then removed the conditional at some point. I'll fix it

web/client/src/library/components/graph/ModelLineageSearch.tsx
```diff
@@ -0,0 +1,85 @@
...
+      className={clsx(
+        'w-full',
+        showSearchInput
+          ? 'block absolute left-0 right-0 z-10 pr-10 bg-dark'
+          : 'items-end justify-end flex',
+      )}
+    >
+      <Button
+        className={clsx(
+          'flex @[40rem]:hidden border-none !py-1 !rounded-full',
```
mykalmax, line 47: there is `shape`  prop on `Button` and you can pass `EnumButtonShape.Circle`

mykalmax (approved):

PR #2022, opened 2024-01-25
fix: further improve performance of dag by introducing a dummy node

tobymao:
nodes are now connected to a terminal node to avoid many to many edges

before:

Screenshot_2024-01-24_19 15 44

after:

Screenshot_2024-01-24_19 14 38

104 edges -> 60 edges

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -265,6 +265,8 @@ class Scheduler:
                 raise CircuitBreakerError()
 
             snapshot_name, ((start, end), batch_idx) = node
+            if batch_idx == -1:
+                return
             snapshot = snapshots_by_name[snapshot_name]
 
             self.console.start_snapshot_evaluation_progress(snapshot)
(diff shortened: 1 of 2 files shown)


izeigerman:
Do we need a dummy node for snapshots that only have a single interval to backfill?

sqlmesh/core/scheduler.py
```diff
@@ -329,29 +331,23 @@ def _dag(self, batches: SnapshotToBatches) -> DAG[SchedulingUnit]:
         }
 
         dag = DAG[SchedulingUnit]()
+        dummy_node = ((to_datetime(0), to_datetime(0)), -1)
```
izeigerman, line 334: should we call this `terminal_node` or `batch_end_node` to make its purpose clear?

tobymao:
reduced to 58 edges

Screenshot_2024-01-25_16 50 42

izeigerman (approved):

PR #2025, opened 2024-01-26
fix: changing catalog env override existing catalog matches

eakmanrq:
Scenario: A user has views being created in a catalog we will call `main` based on the default catalog from the connection (could also come from the model itself). They then add a `environment_catalog_mapping` which maps that models to the same catalog that they were already being mapped to which would mean having the views still be created in `main`. Before this change, we would consider this a catalog change and therefore say that we need to both add and remove those views from the environment. Since add runs before remove this would result in the views being dropped.

We now check the actual view paths and only demote if we see a change. So if the catalog change ends up being no change for the object itself then we won't remove it.

### DIFF
--- a/sqlmesh/core/state_sync/common.py
+++ b/sqlmesh/core/state_sync/common.py
@@ -189,11 +189,28 @@ class CommonStateSyncMixin(StateSync):
 
         self._update_environment(environment)
 
-        removed = (
-            list(existing_table_infos.values())
-            if environment_suffix_target_changed or environment_catalog_changed
-            else [existing_table_infos[name] for name in missing_models]
-        )
+        removed = [existing_table_infos[name] for name in missing_models]
+        if environment_catalog_changed:
+            removed = list(
+                {
+                    table_info
+                    for table_info in existing_table_infos.values()
+                    # Can remove the `table_info.fully_qualified_table.catalog is not None` once we make sure that
+                    # `default` schema is included in the fully qualified table and therefore all fully qualified
+                    # tables have a catalog.
+                    if (
+                        table_info.fully_qualified_table.catalog is not None
+                        and table_info.fully_qualified_table.catalog
+                        != environment.catalog_name_override
+                    )
+                    or (
+                        existing_environment
+                        and existing_environment.catalog_name_override is not None
+                    )
+                }.union(removed)
+            )
+        elif environment_suffix_target_changed:
+            removed = list(existing_table_infos.values())
 
         return PromotionResult(
             added=sorted(table_infos),
--- a/sqlmesh/core/state_sync/common.py
+++ b/sqlmesh/core/state_sync/common.py
@@ -15,6 +15,7 @@ from sqlmesh.core.snapshot import (
     SnapshotIdLike,
     SnapshotInfoLike,
     SnapshotNameVersionLike,
+    SnapshotTableInfo,
     start_date,
 )
 from sqlmesh.core.state_sync.base import PromotionResult, StateSync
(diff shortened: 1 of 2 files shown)


izeigerman (approved):

sqlmesh/core/state_sync/common.py
```diff
@@ -195,11 +195,28 @@ def promote(
...
+                    # Can remove the `table_info.fully_qualified_table.catalog is not None` once we make sure that
+                    # `default` schema is included in the fully qualified table and therefore all fully qualified
+                    # tables have a catalog.
+                    if (
+                        table_info.fully_qualified_table.catalog is not None
+                        and table_info.fully_qualified_table.catalog
+                        != environment.catalog_name_override
+                    )
+                    or (
+                        existing_environment
```
izeigerman, line 213: Why not just use existing `environment_catalog_changed`?

eakmanrq: I completely rewrote this logic: https://github.com/TobikoData/sqlmesh/pull/2025/commits/7bc8e359310df6594b0a2e5c5990003642d525dd

tests/core/test_state_sync.py
```diff
@@ -465,19 +465,21 @@ def test_promote_snapshots(state_sync: EngineAdapterStateSync, make_snapshot: t.
 
     promotion_result = promote_snapshots(state_sync, [snapshot_a, snapshot_b_old], "prod")
 
-    assert set(promotion_result.added) == set([snapshot_a.table_info, snapshot_b_old.table_info])
+    assert sorted([x.json() for x in promotion_result.added]) == sorted(
```
tobymao, line 468: why did you need to change this test?

tobymao:
very nice changes, just one comment on the test changes

tests/core/test_state_sync.py
eakmanrq: It seems related to Pydantic version. With Pydantic 1 and I _think_ some Pydantic 2 versions the old version passes but when upgraded to latest it says they are different. So I serialize/sort the results and they correctly determine if they are equal.

tobymao: can you take a look at the actual pydantic error and let’s fix equality? this may patch the test but will still fail if we use eq check in regular code. id rather fix the equality in general as opposed to working around it in tests

eakmanrq: The issue is that Pydantic includes private attributes in the equality check:
https://github.com/pydantic/pydantic/blob/b785d5b1db4eccb922dd2bf17faea5fc842e6a57/pydantic/main.py#L880

We use these for caching values. In this case it is the fully qualified table: https://github.com/TobikoData/sqlmesh/blob/d5fd1b23a8cbb6c0eca1313af65762ebf09c006c/sqlmesh/core/snapshot/definition.py#L371-L375

This is actually a project wide issue that we don't currently address. A bad fix, just to demonstrate the issue, would be this:
```
    def __eq__(self, other):
        return (self.copy().__pydantic_private__ or {}).clear().__eq__((other.copy().__pydantic_private__ or {}).clear()) 
```

The problem obviously being that we would create copies on each equality check. It would be good to sync with @izeigerman on some options since we knows Pydantic the best. I'm fine with fixing this issue being out of scope of this PR.

tobymao: i don't really agree with that given you've already altered the scope of this PR by changing the tests to json(), i think we should fix it, in snapshot for example, we already over equality to simply check the snapshot id, i would not make an equality check that does copy()

tobymao: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/definition.py#L601

eakmanrq: I think we need a general solution that doesn't require making sure to override `__eq__` for each equality check.

eakmanrq: Opening separate PR to try to find a solution everyone is happy with

eakmanrq: Created here: https://github.com/TobikoData/sqlmesh/pull/2035

izeigerman (approved):

PR #2028, opened 2024-01-26
Feat: Create snapshot tables on demand

izeigerman:
This update makes sure that physical tables for snapshots are created lazily when they are actually needed. This is particularly helpful when using the model selector, in which case we create tables for selected models only.

Additionally, this update has a nice side effect of adding self-healing properties to the plan / apply flow by automatically re-creating snapshot tables if they happen to be missing.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -1406,7 +1406,7 @@ class EngineAdapter:
                 )
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
-    def list_data_objects(
+    def get_data_objects(
         self, schema_name: SchemaName, object_names: t.Optional[t.Set[str]] = None
     ) -> t.List[DataObject]:
         """Lists all data objects in the target schema.
@@ -1419,6 +1419,8 @@ class EngineAdapter:
             A list of data objects in the target schema.
         """
         if object_names is not None:
+            if not object_names:
+                return []
             object_names_list = list(object_names)
             batches = [
                 object_names_list[i : i + self.DATA_OBJECT_FILTER_BATCH_SIZE]
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -79,6 +79,7 @@ class EngineAdapter:
 
     DIALECT = ""
     DEFAULT_BATCH_SIZE = 10000
+    DATA_OBJECT_FILTER_BATCH_SIZE = 1000
     ESCAPE_JSON = False
     SUPPORTS_TRANSACTIONS = True
     SUPPORTS_INDEXES = False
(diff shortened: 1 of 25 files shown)


sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -103,15 +103,44 @@ def _fetch_native_df(
...
+                    exp.Literal.string("MATERIALIZED_VIEW"),
+                )
+                .else_(exp.column("TABLE_TYPE"))
+                .as_("type"),
+            )
+            .from_(exp.table_("TABLES", db="INFORMATION_SCHEMA", catalog=catalog_name))
+            .where(exp.column("TABLE_SCHEMA").eq(schema.db))
+        )
+        if object_names:
+            # FIXME: Find a way to properly normalize object names.
```
tobymao, line 139: arent names already normalized? where do they come from?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -230,12 +232,32 @@ def create(
...
-            [s.table_name() for s in target_snapshots if s.is_model and not s.is_symbolic]
-        )
+        snapshots_with_table_names: t.List[t.Tuple[Snapshot, str]] = []
+        tables_by_schema = defaultdict(set)
+        for snapshot in target_snapshots:
+            if not snapshot.is_model or snapshot.is_symbolic:
+                continue
+            table = exp.to_table(snapshot.table_name(False), dialect=snapshot.model.dialect)
+            snapshots_with_table_names.append((snapshot, table.name.lower()))
+            tables_by_schema[d.schema_(table.db, catalog=table.catalog)].add(table.name)
```
tobymao, line 242: why is snapshots_with_table_names lower table name but tables_by_schema not?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -230,12 +232,32 @@ def create(
...
+
+        existing_objects: t.Set[str] = set()
+        for schema, object_names in tables_by_schema.items():
+            logger.info("Listing data objects in schema %s", schema.sql())
+            objs = self.adapter.list_data_objects(schema, object_names)
+            existing_objects.update(obj.name.lower() for obj in objs)
```
tobymao, line 248: why are we changing case sensitivy here? why not keep it as is?

sqlmesh/core/engine_adapter/snowflake.py
izeigerman: Actually you might be right

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1405,6 +1406,29 @@ def rename_table(
                 )
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
+    def list_data_objects(
```
eakmanrq, line 1409: should we make this consistent with the private name? Seems strange to have `list` and then `get`.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -103,15 +103,42 @@ def _fetch_native_df(
...
+            .from_(exp.table_("TABLES", db="INFORMATION_SCHEMA", catalog=catalog_name))
+            .where(exp.column("TABLE_SCHEMA").eq(schema.db))
+        )
+        if object_names:
+            query = query.where(exp.column("TABLE_NAME").isin(*object_names))
         try:
-            df = self.fetchdf(sql, quote_identifiers=True)
+            df = self.fetchdf(query, quote_identifiers=True)
         except ProgrammingError as e:
             if "Object does not exist" in str(e):
```
eakmanrq, line 143: This probably isn't relevant anymore now that we aren't using "SHOW OBJECTS...".

eakmanrq:
Since you added a public method usually these get a test added in `test_integration`. Specifically it would be good to test the object names functionality and possibly with the batching too. So it would be something like create table and view and make sure they show up and then filter and make sure the filter limits the number of results. If you want something quick you could just update the `_get_data_objects` call to instead use the new public method which would at least have us using that public path.

sqlmesh/core/engine_adapter/snowflake.py
izeigerman: The object might still not exist if the catalog doesn't exist, right?

eakmanrq: Maybe? Not sure that is the error you would get if you reference a non-existent catalog. Should verify if that is the intent.

izeigerman:
@eakmanrq added additional tests

eakmanrq (approved):

PR #2030, opened 2024-01-26
Fix: When deleting expired snapshots in janitor only fetch relevant snapshot records

izeigerman:
Previously,  the janitor process would fetch all snapshot records to figure out what needs to be cleaned up. This led to a large memory footprint for projects with many models and / or changes.

This update ensures that only records that actually expired are fetched from the state DB.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -57,12 +57,15 @@ from sqlmesh.core.snapshot.definition import (
 )
 from sqlmesh.core.state_sync.base import MIGRATIONS, SCHEMA_VERSION, StateSync, Versions
 from sqlmesh.core.state_sync.common import CommonStateSyncMixin, transactional
-from sqlmesh.utils import major_minor, random_id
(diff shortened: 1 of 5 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -256,6 +259,66 @@ def invalidate_environment(self, name: str) -> None:
...
+            .where(exp.column("expiration_ts") <= current_ts)
+        )
+        expired_candidates = {
+            SnapshotId(name=name, identifier=identifier): SnapshotNameVersion(
+                name=name, version=version
+            )
+            for name, identifier, version in self.engine_adapter.fetchall(expired_query)
+        }
+
+        snapshots = self._get_snapshots_with_same_version(set(expired_candidates.values()))
```
tobymao, line 278: is there anyway to batch this as well for the worst case?

izeigerman: What do you mean by batch here?

tobymao (approved):

PR #2031, opened 2024-01-26
fix: Use Insert/Overwrite for Replace to Fix Dataframe Batching

eakmanrq:
Some engines don't support "CREATE OR REPLACE AS SELECT..." which means we had to come up with another way to replace the contents in a given table. Prior to this PR, some engines had custom logic to do this replacement. You could though think of a `CREATE OR REPLACE AS SELECT...` as a INSERT/OVERWRITE where the table already exists. Therefore this change makes `replace_query` use INSERT/OVERWRITE when it can't replace. Since INSERT/OVERWRITE already properly supports batching this will resolve the issues users were reporting while simplifying the code.


tests/core/engine_adapter/test_integration.py
```diff
@@ -853,6 +853,59 @@ def test_replace_query(ctx: TestContext):
         ctx.compare_with_current(table, replace_data)
 
 
+def test_replace_query_batched(ctx: TestContext):
```
eakmanrq, line 856: This was added to specifically test the issue users reported (and used to fail for some engines).

sqlmesh/core/engine_adapter/base.py
```diff
@@ -304,15 +305,59 @@ def replace_query(
...
-            table,
-            source_queries,
-            columns_to_types,
-            replace=True,
-            table_description=table_description,
-            column_descriptions=column_descriptions,
-            **kwargs,
+        columns_to_types = columns_to_types or self.columns(table)
+        query = source_queries[0].query_factory()
+        target_table = exp.to_table(table_name)
```
tobymao, line 310: isn't table already exp.to_table(table_name) what's going on here

eakmanrq: Yep just looks like a copy/paste mistake.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -304,15 +305,59 @@ def replace_query(
...
+        self_referencing = any(
+            quote_identifiers(table) == quote_identifiers(target_table)
```
tobymao, line 312: you're shadowing the already table var,

it should be

quote_identifiers(t) == quote_identifiers(table) for t in ...

eakmanrq: just stuck with targe_tablet: https://github.com/TobikoData/sqlmesh/pull/2031/commits/b2cecadd3c7c9a7d80a53df66d320ca6613ce150

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -176,13 +91,16 @@ def _insert_overwrite_by_condition(
...
+        insert_overwrite_strategy_override: t.Optional[InsertOverwriteStrategy] = None,
     ) -> None:
         """
         Some engines do not support `INSERT OVERWRITE` but instead support
         doing an "INSERT OVERWRITE" using a Merge expression but with the
         predicate being `False`.
         """
         columns_to_types = columns_to_types or self.columns(table_name)
+        if not where:
+            where = exp.true()
```
tobymao, line 103: why is this needed? where=None should just mean no where statement right

eakmanrq: I think some engines require a condition on an overwrite? I will test.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -300,19 +301,62 @@ def replace_query(
...
-            **kwargs,
+        columns_to_types = columns_to_types or self.columns(target_table)
+        query = source_queries[0].query_factory()
+        self_referencing = any(
+            quote_identifiers(table) == quote_identifiers(target_table)
+            for table in query.find_all(exp.Table)
         )
+        # If a query references itself then it must have a table created regardless of approach used.
+        if self_referencing:
+            self._create_table_from_columns(
```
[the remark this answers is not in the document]
eakmanrq: Yeah but that is fine right? Since it self-references then it would do a `CREATE OR REPLACE AS SELECT...` and reference itself so it would actually run. Without this it would error since it would reference a table that doesn't exist.

sqlmesh/core/engine_adapter/mixins.py
tobymao: ah, i see, but it wasn't there before which is why i was wondering why it was added

sqlmesh/core/engine_adapter/base.py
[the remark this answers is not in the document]
tobymao: what happened before this was added?

tobymao (approved):

sqlmesh/core/engine_adapter/base.py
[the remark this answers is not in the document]
eakmanrq: Looks like it would error if self-referencing: https://github.com/TobikoData/sqlmesh/blob/c0bbe57041b090bd8c802e8ca04174040d8ba2f8/sqlmesh/core/engine_adapter/base.py#L618-L626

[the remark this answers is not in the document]
eakmanrq: Note that we wouldn't see this within SQLMesh since we always have a table but this now properly ensures that is the case regardless of context.

sqlmesh/core/engine_adapter/mixins.py
eakmanrq: Tests ran fine without it so I will go ahead and remove.

PR #2032, opened 2024-01-26
Fix: Fetch snapshot records on demand when generating Airflow dags

izeigerman:
Currently we fetch all snapshots on the scheduler side which may lead to high memory consumption there. Additionally we had passed all these snapshots into underlying DAG tasks potentially causing high memory consumption at the execution time which could lead to tasks being SIGKILLed.

### DIFF
--- a/sqlmesh/schedulers/airflow/integration.py
+++ b/sqlmesh/schedulers/airflow/integration.py
@@ -12,6 +12,7 @@ from airflow.utils.session import provide_session
 from sqlalchemy.orm import Session
 
 from sqlmesh.core import constants as c
+from sqlmesh.core.state_sync import StateReader
 from sqlmesh.engines import commands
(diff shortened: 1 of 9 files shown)


tobymao (approved):

sqlmesh/schedulers/airflow/integration.py
```diff
@@ -123,29 +124,27 @@ def dags(self) -> t.List[DAG]:
         """
         self.set_default_catalog(self._default_catalog)
         with util.scoped_state_sync() as state_sync:
-            stored_snapshots = state_sync.get_snapshots(None)
```
tobymao, line 126: so this is the main change right? you got rid of this

izeigerman: exactly

PR #2033, opened 2024-01-27
fix: Spark Create Table DDL WAP

eakmanrq:
Fix after this PR: https://github.com/TobikoData/sqlmesh/pull/2031

The PR introduced a flow where we would create a table if not exists if the query is a self-referencing query. This doesn't work for Spark WAP since spark doesn't let you create a table with a WAP ID at the end. This makes sense because to have the branch you needed to create the table first and then altered it. Therefore this PR updates Spark's create table to issue the create table on the table itself without the branch. 

While working on that fix I noticed a different issue: If `_create_table` was called with `IF NOT EXISTS` included and was Iceberg and WAP capable then we would do the "Dummy" Insert into call: https://github.com/TobikoData/sqlmesh/blob/f583fcec955e9433f755737ecbab279888643fc0/sqlmesh/core/engine_adapter/spark.py#L466

The problem would be if the table has data in it which is possible since we could have done "CREATE IF NOT EXIST" and it did exist and have data. In that case we would duplicate all of the data. Therefore I also add an exists check to ensure that we only create the dummy snapshot if the table was actually created.

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -443,7 +443,12 @@ class SparkEngineAdapter(GetCurrentCatalogFromFunctionMixin, HiveMetastoreTableP
             if wap_id.startswith(f"{self.BRANCH_PREFIX}{self.WAP_PREFIX}"):
                 table_name.set("this", table_name.this.this)
 
-        already_exists = False if not exists else self.table_exists(table_name)
+        wap_supported = (
+            kwargs.get("storage_format") or ""
+        ).lower() == "iceberg" or self.wap_supported(table_name)
+        do_dummy_insert = (
+            False if not wap_supported or not exists else not self.table_exists(table_name)
+        )
         super()._create_table(
             table_name_or_schema,
             expression,
@@ -459,11 +464,7 @@ class SparkEngineAdapter(GetCurrentCatalogFromFunctionMixin, HiveMetastoreTableP
             if isinstance(table_name_or_schema, exp.Schema)
             else exp.to_table(table_name_or_schema)
         )
-        if (
-            not already_exists
-            and (kwargs.get("storage_format") or "").lower() == "iceberg"
-            or self.wap_supported(table_name)
-        ):
+        if do_dummy_insert:
             # Performing a dummy insert to create a dummy snapshot for Iceberg tables
             # to workaround https://github.com/apache/iceberg/issues/8849.
             dummy_insert = exp.insert(exp.select("*").from_(table_name), table_name)
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -61,6 +61,9 @@ class SparkEngineAdapter(GetCurrentCatalogFromFunctionMixin, HiveMetastoreTableP
     # currently check for storage formats we say we don't support REPLACE TABLE
     SUPPORTS_REPLACE_TABLE = False
 
+    WAP_PREFIX = "wap_"
+    BRANCH_PREFIX = "branch_"
+
     @property
     def spark(self) -> PySparkSession:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -428,6 +431,19 @@ def _create_table(
...
+            else exp.to_table(table_name_or_schema)
+        )
+        # Spark doesn't support creating a wap table DDL. Therefore we check if this is a wap table and if it is,
+        # this is not a replace, and the table already exists then we can safely just return. Otherwise we let it error.
+        if not expression and isinstance(table_name.this, exp.Dot):
+            wap_id = table_name.this.parts[-1].name
+            if wap_id.startswith(f"{self.BRANCH_PREFIX}{self.WAP_PREFIX}"):
+                table_name.set("this", table_name.this.this)
+
+        already_exists = False if not exists else self.table_exists(table_name)
```
izeigerman, line 446: Does it mean we now check table existence on each table creation call? Should we change `exists` to False` by default then?

izeigerman: Also why call the parent method if the table exists?

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -443,8 +459,10 @@ def _create_table(
             if isinstance(table_name_or_schema, exp.Schema)
             else exp.to_table(table_name_or_schema)
         )
-        if (kwargs.get("storage_format") or "").lower() == "iceberg" or self.wap_supported(
-            table_name
+        if (
+            not already_exists
+            and (kwargs.get("storage_format") or "").lower() == "iceberg"
+            or self.wap_supported(table_name)
```
izeigerman, line 465: This condition doesn't look right to me. Wouldn't it always be true if `wap_supported` is true regardless of the `already_exists` value?

sqlmesh/core/engine_adapter/spark.py
izeigerman: Can we at least set it to False when creating tables in the snapshot evaluator since we already check table existence there.

izeigerman: Also I feel like the check is only relevant if it's iceberg.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Thanks should have caught this

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Now does check only if WAP supported and `IF NOT EXISTS`: https://github.com/TobikoData/sqlmesh/pull/2033/commits/ae6bb0e329b0f48285ccb1277db3e3601c755cdb

Can you provide more context for this feedback? I don't follow.

>Also why call the parent method if the table exists?

>Can we at least set it to False when creating tables in the snapshot evaluator since we already check table existence there.

izeigerman (approved):

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -428,6 +431,24 @@ def _create_table(
...
+        if not expression and isinstance(table_name.this, exp.Dot):
+            wap_id = table_name.this.parts[-1].name
+            if wap_id.startswith(f"{self.BRANCH_PREFIX}{self.WAP_PREFIX}"):
+                table_name.set("this", table_name.this.this)
+
+        wap_supported = (
+            kwargs.get("storage_format") or ""
+        ).lower() == "iceberg" or self.wap_supported(table_name)
+        do_dummy_insert = (
+            False if not wap_supported or not exists else not self.table_exists(table_name)
```
izeigerman, line 450: That's a lot of negations btw :)

PR #2035, opened 2024-01-28
fix!: ignore private attributes doing equality pydantic

eakmanrq:
We have an issue that Pydantic includes private attributes in the equality check:
https://github.com/pydantic/pydantic/blob/b785d5b1db4eccb922dd2bf17faea5fc842e6a57/pydantic/main.py#L880

We use these for caching values. When we do equality checks our intent is to check if the public attributes of the class are the same and not if the cached attributes are equal. 

This PR has a way of solving this problem but involves mutating the actual objects themselves so it would be best if we had a solution that didn't involve that. Are there Pydantic config properties that can help us here? @izeigerman do you have any ideas? I want to find a solution that works for all Pydantic classes instead requiring each one to provide an override if they introduce private attributes.

### DIFF
--- a/sqlmesh/utils/pydantic.py
+++ b/sqlmesh/utils/pydantic.py
@@ -23,7 +23,9 @@ if t.TYPE_CHECKING:
 
 T = t.TypeVar("T")
 DEFAULT_ARGS = {"exclude_none": True, "by_alias": True}
-PYDANTIC_MAJOR_VERSION = int(pydantic.__version__.split(".")[0])
+PYDANTIC_MAJOR_VERSION, PYDANTIC_MINOR_VERSION = [int(p) for p in pydantic.__version__.split(".")][
+    :2
+]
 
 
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/pydantic.py
```diff
@@ -24,6 +24,7 @@
 T = t.TypeVar("T")
 DEFAULT_ARGS = {"exclude_none": True, "by_alias": True}
 PYDANTIC_MAJOR_VERSION = int(pydantic.__version__.split(".")[0])
+PYDANTIC_MINOR_VERSION = int(pydantic.__version__.split(".")[1])
```
izeigerman, line 27: [Nit] 
```
PYDANTIC_MAJOR_VERSION, PYDANTIC_MINOR_VERSION = [int(p) for p in pydantic.__version__.split(".")][:2]
```
or something to eliminate split duplication

izeigerman (approved):

Issue #2037, opened 2024-01-28
DuckDB catalogs config option doesn't commit changes


eakmanrq:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1706126317634699

Commit with repro: https://github.com/plaflamme/sqlmesh/commit/0b50c8bb877892053fcc1a4fe6585b316b6667b1

Summary: basically, running sqlmesh plan and immediately running sqlmesh fetchdf … fails when using the catalogs configuration (at least on some models).

eakmanrq:
@plaflamme This should now be fixed with `0.69.0`: https://github.com/TobikoData/sqlmesh/releases/tag/v0.69.0

PR #2038, opened 2024-01-29
fix: context close on Create/Update/Delete operations

eakmanrq:
Part of the fix for this issue: https://github.com/TobikoData/sqlmesh/issues/2037

It makes sense that we close connections when using CLI prior to exiting. Open question about how we would mimic this in environments with long-running python instances (Notebook/UI) but fixing in CLI for now.

### DIFF
--- a/sqlmesh/cli/__init__.py
+++ b/sqlmesh/cli/__init__.py
@@ -5,6 +5,7 @@ from functools import wraps
 import click
 from sqlglot.errors import SqlglotError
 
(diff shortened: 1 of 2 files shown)


izeigerman (approved):
Show we do this in the `except` block?

tobymao:
or finally?

izeigerman:
yes, thanks, @tobymao , that's what I meant

PR #2039, opened 2024-01-29
fix: share duckdb adapter is overlapping data files

eakmanrq:
Part of fix for: https://github.com/TobikoData/sqlmesh/issues/2037

Result of this change: https://github.com/TobikoData/sqlmesh/pull/1975

Prior to this change we would allow multiple adapters to share a common data file for DuckDB. This actually works fine until multiple adapters are created again pointed to the already created file and then the second adapter (typically state sync) will not read any data. This is why testing before of having multiple adapter run a plan and check the result did not show any issues.

The solution is to check when creating a new adapter for DuckDB if the new adapter is going to point to file that overlaps a file already used by another adapter. If so, we just reuse the past adapter. This has a downside that any additional configuration associated with the new adapter is ignored. One could consider merging the previous and new catalogs together if this overlap is detected but then it doesn't address other config. For example, what if the new one has other connection config set? How do we merge those? I'm wondering if this naive approach will work in most cases to where we can handle the more complex situations later. For example we will eventually want to support Motherduck and that could be a good time to rework things with a more robust use-case in mind.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import base64
+import logging
 import os
 import pathlib
 import sys
@@ -31,6 +32,9 @@ if sys.version_info >= (3, 9):
     from typing import Literal
 else:
     from typing_extensions import Literal
+
+
+logger = logging.getLogger(__name__)
 
 
 class ConnectionConfig(abc.ABC, BaseConfig):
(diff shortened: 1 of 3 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -182,6 +188,29 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
         return init
 
+    def create_engine_adapter(self, register_comments_override: bool = False) -> EngineAdapter:
+        """Checks if another engine adapter has already been created that shares a catalog that points to the same data
+        file. If so, it uses that same adapter instead of creating a new one. As a result, any additional configuration
+        associated with the new adapter will be ignored."""
+        data_files = set((self.catalogs or {}).values())
+        if self.database:
+            data_files.add(self.database)
+        data_files = {x for x in data_files if x != ":memory:"}
```
tobymao, line 198: data_files -= ":memory"

tobymao (approved):

PR #2040, opened 2024-01-29
Fix: Improve plan flow transitions

mykalmax:


web/server/api/endpoints/plan.py
```diff
@@ -2,6 +2,7 @@
 
 import asyncio
 import typing as t
+from time import sleep
```
vchan, line 5: Why do we need all the sleeps?

mykalmax: sleeps  just to slow down state change a bit so UI can transition smoother. I can delay it on UI but it will involve touching many places and adding bunch of `setTimeout`s and `useState`

tobymao: that should definitely not be done in the api

web/client/src/library/pages/root/Page.tsx
```diff
@@ -383,62 +377,79 @@ function ActionStatus({
...
-          changes={planOverview.backfills}
+          changes={planOverviewTracker.backfills}
           className="ml-4 group-hover:ml-2 z-[6]"
         />
       ) : (
         <div className="ml-2 group-hover:ml-2 z-[6]"></div>
       )}
     </span>
+  ) : (
+    <span className="block ml-1 px-3 py-1 first-child:ml-0 rounded-full whitespace-nowrap font-bold text-xs text-center bg-neutral-5 dark:bg-neutral-20 cursor-default text-neutral-500  dark:text-neutral-300">
```
vchan, line 450:
```suggestion
    <span className="block ml-1 px-3 py-1 first-child:ml-0 rounded-full whitespace-nowrap font-bold text-xs text-center bg-neutral-5 dark:bg-neutral-20 cursor-default text-neutral-500 dark:text-neutral-300">
```

web/client/src/library/components/plan/PlanActions.tsx
```diff
@@ -90,83 +92,112 @@ export default function PlanActions({
...
               {ModelPlanAction.getActionDisplayName(
                 planAction,
-                [],
-                'Start Over',
+                planAction.isProcessing && isFalse(planCancel.isSuccessed)
+                  ? [EnumPlanAction.Cancelling]
+                  : [],
+                planAction.isProcessing
+                  ? planCancel.isSuccessed
+                    ? 'Finishing Cancelation...'
```
vchan, line 195:
```suggestion
                    ? 'Finishing Cancellation...'
```

web/client/src/library/components/editor/EditorTabs.tsx
```diff
@@ -60,7 +60,7 @@ export default function EditorTabs(): JSX.Element {
       addTab(newTab)
     }
 
-    selectTab(newTab)
+    setTimeout(() => selectTab(newTab))
```
vchan, line 63: I understand at a very basic level what this does, but why do this here (setTimeout without a delay)?

vchan (approved):

web/client/src/library/components/editor/EditorTabs.tsx
mykalmax: Forgot to remove. There was a bug where when selecting file in docs and go back to editor, Edito Tab is not selected for selected file. I fixed it [here](https://github.com/TobikoData/sqlmesh/pull/2076)

vchan (approved):

PR #2041, opened 2024-01-29
Feat: Macro typing coercion

z3z1ma:
A spike at allowing users to explicitly define types for macro arguments. This makes the interface a little bit nicer since you know exactly what you are gonna get. The implementation is a proof of concept at where we may add this functionality.

Thoughts?


z3z1ma:
Test failures unrelated to my change AFAICT

tests/core/test_macros.py
```diff
@@ -27,6 +27,15 @@ def upper_case(evaluator: MacroEvaluator, expression: exp.Condition) -> str:
     def noop(evaluator: MacroEvaluator):
         return None
 
+    @macro("add_ints", coerce_args=True)
+    def add_lits(evaluator: MacroEvaluator, x: int, y: int) -> int:
+        return x + y
```
z3z1ma, line 32: This is a contrived test to show it working, prime use cases are just making the interface more explicit such as in the following test along with bailing on types which are not coercible.

sqlmesh/core/macros.py
```diff
@@ -369,10 +370,31 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
...
-        self, func: t.Callable[..., DECORATOR_RETURN_TYPE]
+        self,
+        func: t.Callable[..., DECORATOR_RETURN_TYPE],
     ) -> t.Callable[..., DECORATOR_RETURN_TYPE]:
-        wrapper = super().__call__(func)
+        @wraps(func)
+        def _typed_func(*args: t.Any, **kwargs: t.Any) -> DECORATOR_RETURN_TYPE:
+            spec = inspect.getfullargspec(func)
+            kwargs = inspect.getcallargs(func, *args, **kwargs)
+            for param, value in kwargs.items():
```
tobymao, line 385: what happens if there are no types in the macro definition?

tobymao:
can you update the documentation to indicate this new behavior of type annotations?

sqlmesh/core/macros.py
z3z1ma: Then `spec.annotations` is an empty dict
image

We would see it manifest as a KeyError currently but we can catch that and reraise with something more informative, IE all types must be annotated when `coerce_args=True` or something.

tobymao: no, we shouldn't require types. it should be an optional thing, so make sure we have a test for that and add code to handle untyped definitions

sqlmesh/core/macros.py
```diff
@@ -369,10 +370,31 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
...
+                coercible_type = spec.annotations[param]
+                if isinstance(value, coercible_type):
+                    continue
+                elif issubclass(coercible_type, exp.Expression):
+                    kwargs[param] = parse_one(value.sql(), into=coercible_type)
+                elif coercible_type in (int, float, str) and isinstance(value, exp.Literal):
```
tobymao, line 391: i think we can have an option which is if the type is a string, but it's not a literal, you can just call value.sql(dialect=self.dialect)

sqlmesh/core/macros.py
tobymao: what do you think about not having coercable_types at all, and just doing it if you add type annotations?

sqlmesh/core/macros.py
z3z1ma: Sure that makes sense. Very convenient.

sqlmesh/core/macros.py
z3z1ma: If we make the types optional, meaning no annotation just passes it through as-is to the function body -- then yes we can technically just drop coercible_args since you can opt out by not typing it in the sig and doing it in the body manually. 

I can do that

z3z1ma: @tobymao done

sqlmesh/core/macros.py
z3z1ma: done

sqlmesh/core/macros.py
```diff
@@ -349,6 +350,25 @@ def engine_adapter(self) -> EngineAdapter:
...
         return self.locals["engine_adapter"]
 
+    def _coerce(self, expr: exp.Expression, typ: t.Type) -> t.Any:
+        """Coerces the given expression to the specified type on a best-effort basis."""
+        try:
+            if typ is None:
+                return expr
+            if isinstance(expr, typ):
+                return expr
+            elif issubclass(typ, exp.Expression):
```
tobymao, line 360: you don't need elif when all statments are returns, in the codebase we try to do the pattern

if...
return
if...
return

sqlmesh/core/macros.py
```diff
@@ -349,6 +350,25 @@ def engine_adapter(self) -> EngineAdapter:
...
+                return parse_one(expr.sql(self.dialect), into=typ, read=self.dialect)
```
tobymao, line 361: here, since you're doing an imediate parse_one from the sql, you don't need to specify the dialect.

you can just do parse_one(expr.sql(), into=typ)

sqlmesh/core/macros.py
```diff
@@ -370,9 +390,30 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
...
+        @wraps(func)
+        def _typed_func(*args: t.Any, **kwargs: t.Any) -> DECORATOR_RETURN_TYPE:
+            """Coerce arguments where possible to the user-defined type annotations."""
+            evaluator: MacroEvaluator = args[0]
+            argmap = inspect.getcallargs(func, *args, **kwargs)
+            for param, value in argmap.items():
+                coercible_type = spec.annotations.get(param)
+                if not coercible_type:
+                    continue
+                if isinstance(coercible_type, str):
```
tobymao, line 407: what does it mean for the coercible type to be a string? why do you continue here? is this like 

def x(y: "my custom class") ?

sqlmesh/core/macros.py
```diff
@@ -370,9 +390,30 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
...
+                    continue
+                argmap[param] = evaluator._coerce(value, coercible_type)
+                if (ix := spec.args.index(param)) > -1:
+                    args = args[:ix] + (argmap[param],) + args[ix + 1 :]
```
tobymao, line 411: is this code necessary? can you just pass argmap into func?

sqlmesh/core/macros.py
z3z1ma: Ah its actually extremely interesting @tobymao. This specifically happens when you have `from __future__ import annotations` at the top of the file. The solution is to use `typing.get_type_hints`. 

See PEP 563 https://peps.python.org/pep-0563/

I was chasing this earlier today. Really nice to get to the bottom of it.

image

z3z1ma: I will update the code shortly, current code was just avoiding the problem since I hadnt fully gotten it yet.

sqlmesh/core/macros.py
z3z1ma: Maybe, just need to account for variadic arguments. So `**argmap` wont work unless I pop the variadic key, which i can do actually. So lets simplify yes.

sqlmesh/core/macros.py
```diff
@@ -370,9 +390,28 @@ def add_one(evaluator: MacroEvaluator, column: exp.Literal) -> exp.Add:
...
-        self, func: t.Callable[..., DECORATOR_RETURN_TYPE]
+        self,
+        func: t.Callable[..., DECORATOR_RETURN_TYPE],
     ) -> t.Callable[..., DECORATOR_RETURN_TYPE]:
-        wrapper = super().__call__(func)
+        spec = inspect.getfullargspec(func)
+        annotations = t.get_type_hints(func)
+
+        @wraps(func)
+        def _typed_func(*args_: t.Any, **kwargs_: t.Any) -> DECORATOR_RETURN_TYPE:
```
tobymao, line 400: def _typed_func(evaluator: MacroEvaluator, *args, **kwargs)

z3z1ma: true, doh

tobymao:
one last small suggestion.

can we get a lot more tests, basically covering the gamut of coercisions and docs? thanks @z3z1ma !

z3z1ma:
cool thanks for help @tobymao you rock. i will get to the tests and docs tomorrow

z3z1ma:
@tobymao I added tests that hit every code path I think, I might add another 1 or 2. But otherwise going to focus on the docs.

sqlmesh/core/macros.py
```diff
@@ -349,6 +354,65 @@ def engine_adapter(self) -> EngineAdapter:
...
 
+    def _coerce(self, expr: exp.Expression, typ: t.Type) -> t.Any:
+        """Coerces the given expression to the specified type on a best-effort basis."""
+        try:
+            if typ is None:
+                return expr
+            base = t.get_origin(typ) or typ
+            if isinstance(expr, base):
+                return expr
+            if t.get_origin(typ) is t.Union:
```
tobymao, line 365: this is kind of concerning how complicated this is going to get with type coercision.

sqlmesh/core/macros.py
```diff
@@ -349,6 +354,65 @@ def engine_adapter(self) -> EngineAdapter:
...
+                    try:
+                        return self._coerce(expr, t_)
+                    except Exception:
+                        pass
+                raise SQLMeshError(f"Failed to coerce expression '{expr}' to type '{typ}'.")
+            if issubclass(base, exp.Expression):
+                d = Dialect.get_or_raise(self.dialect)
+                into = base if base in d.parser().EXPRESSION_PARSERS else None
+                coerced = parse_one(
+                    expr.this if isinstance(expr, exp.Literal) else expr.sql(), into=into
```
tobymao, line 376: this doesn't look right to me anymore. i don't think you should call parse_one(sql()) if there's no into and not a literal

sqlmesh/core/macros.py
```diff
@@ -349,6 +354,65 @@ def engine_adapter(self) -> EngineAdapter:
...
+            base = t.get_origin(typ) or typ
+            if isinstance(expr, base):
+                return expr
+            if t.get_origin(typ) is t.Union:
+                for t_ in t.get_args(typ):
+                    try:
+                        return self._coerce(expr, t_)
+                    except Exception:
+                        pass
+                raise SQLMeshError(f"Failed to coerce expression '{expr}' to type '{typ}'.")
```
tobymao, line 371: this raise is repeated several times, could we make this more dry?

sqlmesh/core/macros.py
```diff
@@ -349,6 +354,65 @@ def engine_adapter(self) -> EngineAdapter:
...
+            if base in (int, float, str) and isinstance(expr, exp.Literal):
+                return base(expr.this)
+            if base is bool and isinstance(expr, exp.Boolean):
+                return expr.this
+            if base is str and isinstance(expr, exp.Expression):
+                return expr.sql(self.dialect)
+            if base is tuple and isinstance(expr, (exp.Tuple, exp.Array)):
+                generic = t.get_args(typ)
+                if not generic:
+                    return tuple(expr.expressions)
```
tobymao, line 391: not every expression defines expressions

sqlmesh/core/macros.py
z3z1ma: Yeah, I think there is a line in the sand on what cases you want to handle. I think I drew the line in a pretty safe place. I wouldn't add more logic here personally. Its the users responsibility if they have some crazy marshaling to do. This handles all the basic cases I can think of with no further code which is extremely nice.

sqlmesh/core/macros.py
z3z1ma: we already asserted its an `exp.Tuple` or `exp.Array`

sqlmesh/core/macros.py
z3z1ma: Sure

sqlmesh/core/macros.py
z3z1ma: I will add a dedicated test set for just this function so you can see it play out directly, instead of testing it indirectly via macros currently

z3z1ma: From there, we can assert that's all the cases we want to handle and anything else is user territory, which is already fine.

sqlmesh/core/macros.py
z3z1ma: I'll take another look. My aim was to be super relaxed and make a best-effort in all cases since we will always fallback to the originally parsed type. There is an `isinstance` after this to make sure we don't give back anything jank and just give back the original if we didnt get it right. 
But I think what you say on the lit and into None combo being its own case...

tobymao: what case are you expecting with no into and expr.sql()?

z3z1ma: yeah I see what you mean, itd be a noop at best. fixed

tobymao (approved):
amazing

PR #2042, opened 2024-01-29
Fix: Show all value for deleted rows (table-diff)

mykalmax:
Show all values for deleted rows

Screenshot 2024-01-29 at 11 35 56 AM


web/client/src/library/components/editor/Editor.tsx
```diff
@@ -221,6 +222,10 @@ function EditorMain({ tab }: { tab: EditorTab }): JSX.Element {
     setPreviewDiff(undefined)
   }, [tab.id, tab.file.fingerprint])
 
+  useEffect(() => {
+    setPreviewDiff(undefined)
+  }, [environment])
```
mykalmax, line 227: if changing environment we need to reset `Diff Preview`

vchan (approved):

PR #2044, opened 2024-01-29
Fi: Bg for modal component disappear while still open

mykalmax:


web/client/src/library/components/modal/ModalDrawer.tsx
```diff
@@ -1,56 +0,0 @@
-import { Transition, Dialog } from '@headlessui/react'
-import { Fragment } from 'react'
-
-export default function ModalDrawer({
```
mykalmax, line 4: not using this component anymore

vchan (approved):

PR #2048, opened 2024-01-30
Feat: more configuration for format command

z3z1ma:
Not enough configurability. Expose very nice implementation in Generator to the benefit of all users while preserving existing behavior by default.

### DIFF
--- a/sqlmesh/magics.py
+++ b/sqlmesh/magics.py
@@ -618,16 +618,49 @@ class SQLMeshMagics(Magics):
         help="Transpile project models to the specified dialect.",
     )
     @argument(
-        "--new-line",
+        "--newline",
         action="store_true",
         help="The output dialect of the sql string.",
+    )
+    @argument(
+        "--normalize",
+        action="store_true",
+        help="Whether or not to normalize identifiers to lowercase.",
+    )
+    @argument(
+        "--pad",
+        type=int,
+        help="Determines the pad size in a formatted string.",
+        default=2,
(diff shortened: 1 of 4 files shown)


eakmanrq:
@z3z1ma Can you also add these to the magic command?

https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/magics.py#L627

z3z1ma:
You bet! Done @eakmanrq

sqlmesh/magics.py
```diff
@@ -618,16 +618,49 @@ def rewrite(self, context: Context, line: str, sql: str) -> None:
...
+        help="The max number of characters in a segment before creating new lines in pretty mode.",
+        default=80,
+    )
     @line_magic
     @pass_sqlmesh_context
     def format(self, context: Context, line: str) -> None:
         """Format all SQL models."""
         args = parse_argstring(self.format, line)
-        context.format(args.transpile, args.new_line)
+        context.format(**vars(args))
```
tobymao, line 663: what's vars?

z3z1ma: The arg parser returns a Namespace. `vars` accesses the `__dict__` attribute basically. On namespaces, the is well defined behavior to convert it to a dict.

z3z1ma: See stdlib docs mention the same

https://docs.python.org/3/library/argparse.html#the-namespace-object

PR #2055, opened 2024-01-30
Feat: add condition to most audits

z3z1ma:
Most builtin audits can be made further general purpose by adding an optional `@condition` parameter. This keeps existing behavior as-is but allows more flexibility before needing to roll your own 

cc @plaflamme


plaflamme (approved):

sqlmesh/core/audit/builtin.py
```diff
@@ -7,22 +7,26 @@
 # not_null(columns=[column_1, column_2])
 not_null_audit = ModelAudit(
     name="not_null",
+    defaults={"condition": exp.true()},
```
izeigerman, line 10: I can't recall whether plain `True` universally works across all engines. I remember having to do `1 = 1` expression instead. Or maybe it's only relevant when there is just one condition in the `WHERE` clause. @tobymao @eakmanrq @treysp any chance one of you knows about what I'm referring to?

tobymao: why do you need any condition at all? sqlglot will remove empty where statements

tobymao: i think TRUE doesn't work in tsql, only 1 =1, but i think we may have a sqlglot transpiler for that

treysp: sqlglot transpiles to 1=1

```
>>> exp.true().sql(dialect="tsql")
'(1 = 1)'
```

z3z1ma: @tobymao what do you mean, how would a user be able to _optionally_ supply a condition?
If they dont provide one, we need a default in the context so there is no KeyError when trying to render the expression?
If the user does provide a condition, we need to specify where it goes (in the where clause with @condition)

Can you share and example of what you mean?

tobymao: i guess we need a new macro def and_ or def or_ that will take in optionals and construct the proper query, can chat with you over slack if you need more info

sqlglot has helpers

exp.and_ that does this i think

tobymao:
taking this to the finish line here https://github.com/TobikoData/sqlmesh/pull/2255

PR #2059, opened 2024-01-31
Fix: Lineage 

vchan:
This PR
- Normalizes table names so the UI can correctly match them with their model's fqn.
- Handles the cases where a lineage node's expression are unions and CTEs.

### DIFF
--- a/tests/web/test_lineage.py
+++ b/tests/web/test_lineage.py
@@ -192,6 +192,31 @@ def test_get_lineage_join(project_context: Context) -> None:
     assert response_json['"baz"']["price"]["models"] == {'"external_baz"': ["price"]}
 
 
+def test_get_lineage_multiple_columns(project_context: Context) -> None:
+    project_tmp_path = project_context.path
+    models_dir = project_tmp_path / "models"
+    models_dir.mkdir()
+    foo_sql_file = models_dir / "foo.sql"
+    foo_sql_file.write_text(
+        """MODEL (name foo);
+           SELECT id, bar.value * bar.multiplier AS col FROM bar;"""
+    )
+    bar_sql_file = models_dir / "bar.sql"
+    bar_sql_file.write_text(
+        """MODEL (name bar);
+           SELECT id, value, multiplier FROM external_bar;"""
+    )
+    project_context.load()
+
+    response = client.get("/api/lineage/foo/col")
+    assert response.status_code == 200, response.json()
+    response_json = response.json()
+    assert "value" in response_json['"foo"']["col"]["models"]['"bar"']
+    assert "multiplier" in response_json['"foo"']["col"]["models"]['"bar"']
+    assert response_json['"bar"']["value"]["models"] == {'"external_bar"': ["value"]}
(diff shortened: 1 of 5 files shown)


web/server/api/endpoints/lineage.py
```diff
@@ -27,16 +28,25 @@ def _get_table(node: Node, dialect: t.Optional[DialectType] = None) -> t.Optiona
     table: t.Union[exp.Table, str] = node.alias or node.name
     if isinstance(node.expression, exp.Table):
         table = node.expression
+    elif isinstance(node.expression, exp.Alias):
+        table = node.expression.this.table
 
     try:
-        return exp.table_name(table, identify=True, dialect=dialect)
+        return normalize_model_name(table, None, dialect=dialect)
```
tobymao, line 35: why are you passing None?

web/server/api/endpoints/lineage.py
```diff
@@ -27,16 +28,25 @@ def _get_table(node: Node, dialect: t.Optional[DialectType] = None) -> t.Optiona
...
     except sqlglot.errors.ParseError:
         # Cannot extract table from node. One reason this can happen is node is
         # '*' because a model selects * from an external model for which we do
         # not know the schema.
         return None
 
 
+def _get_column_name(node: Node) -> str:
+    if isinstance(node.expression, exp.Alias):
```
tobymao, line 44: why not node.expression.alias_or_name?

web/server/api/endpoints/lineage.py
```diff
@@ -106,25 +134,25 @@ async def column_lineage(
...
-            column_name = exp.to_column(node.name).name
-        if column_name in graph.get(node_name, []):
+        column_name = _get_column(node, dialect)
+        if column_name in graph.get(table, []):
             continue
 
         # At this point node_name should be fqn/normalized/quoted
-        graph[node_name] = {
+        graph[table] = {
             column_name: LineageColumn(
```
mykalmax, line 152: i think , we need to append here as well

web/server/api/endpoints/lineage.py
vchan: Sorry, the PR wasn't done yet. I'm passing in default catalog now.

tests/web/test_lineage.py
```diff
@@ -0,0 +1,438 @@
...
+
+    response = client.get("/api/lineage/foo/col")
+    assert response.status_code == 200, response.json()
+    response_json = response.json()
+    assert response_json['"foo"']["col"]["models"] == {'"my_cte1"': ["col"]}
+    assert response_json['"my_cte1"']["col"]["models"] == {
+        '"my_cte2"': ["col"],
+        '"external_table1"': ["col"],
+    }
+    assert response_json['"my_cte2"']["col"]["models"] == {'"bar"': ["col"]}
```
tobymao, line 434: wouldn't it be easier just to check the whole json response as opposed to manually checking each snippet

vchan: The whole json response can get quite cluttered with the expression and source sql.

PR #2063, opened 2024-01-31
Fix: do not include a schema def in CTAS unless all column types are known

### LINKED ISSUE #2060: DDL being sent to engine adapter with UNKOWN data type
Seems related to #1878 but cant say for certain. Log was linked in Slack.

treysp:
CTAS calls include a schema definition if column comments are present and columns_to_types are available. Currently, the schema definition could include `UNKNOWN` or `NULL` data types, which cause the CTAS call to error out.

This PR ensures that the schema definition is only included if all column types are known.

Closes #2060

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -39,7 +39,7 @@ from sqlmesh.core.engine_adapter.shared import (
 )
 from sqlmesh.core.model.kind import TimeColumn
 from sqlmesh.core.schema_diff import SchemaDiffer
-from sqlmesh.utils import double_escape, random_id
+from sqlmesh.utils import columns_to_types_all_known, double_escape, random_id
 from sqlmesh.utils.connection_pool import create_connection_pool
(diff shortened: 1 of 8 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -604,9 +604,14 @@ def _create_table_from_source_queries(
         # types, and for evaluation methods like `LogicalReplaceQueryMixin.replace_query()`
         # calls and SCD Type 2 model calls.
         schema = None
+        columns_to_types_all_known = columns_to_types and all(
```
tobymao, line 607: can you make the all() check into a helper method?

delete the extra columns_to_types check on line 613

tobymao: can you add a test for this?

treysp: Done. The check on line 613 was for mypy - i added a type ignore where it complained instead.

izeigerman (approved):

tobymao (approved):

PR #2064, opened 2024-01-31
Feat: allow configuring default formatting behavior in the UI via sql…

z3z1ma:
…mesh config. In subsequent iterations perhaps this serves as the "default options" while a modal allows overriding them for the session. But as a general source of truth, I think its better here since you have consistency across the entire team as well as version control. 

```yaml
default_gateway: duckdb
gateways:
  duckdb:
    connection:
      type: duckdb

model_defaults:
  dialect: 'duckdb'

format:
  pad: 4
  indent: 4
  leading_commas: true
  max_text_width: 120

ui:
  format_on_save: true
```


z3z1ma:
I wondered the same, personally I think top level would be best. Then the CLI flags would just serve as overrides.
This would ensure uniformity between both uses of SQLMesh without having to ensure your CLI flags match ui opts.

web/server/api/endpoints/files.py
```diff
@@ -67,18 +67,22 @@ async def write_file(
...
+            default_dialect = config.dialect
             dialect = model.dialect if model and model.is_sql else default_dialect
-
             try:
                 expressions = parse(content, default_dialect=default_dialect)
-                content = format_model_expressions(expressions, dialect)
+                content = format_model_expressions(
+                    expressions, dialect, **config.ui.format_options.dict(exclude={"newline"})
+                )
+                if config.ui.format_options.newline:
```
izeigerman, line 84: why do we need this?

izeigerman:
@z3z1ma @mykalmax  +1 to making format settings top-level

web/server/api/endpoints/files.py
z3z1ma: To match the CLI and because by default the formatter strips trailing newlines completely, or rather doesn't generate them. And having a single trailing newline is a style preference some users (such as myself) may have. Also it makes tools like `cat`, `wc`, and likely others work properly.

z3z1ma: We could name it append_newline everywhere if preferable, to make its intention clearer @izeigerman

z3z1ma:
@izeigerman I cleaned it up a bit, renamed the newline option to make its intent clear, and made sure format_options is a top level config applicable to both the CLI and UI, and lastly that CLI flags take precedence over the config settings as would be expected.

z3z1ma:
Seems like a pydantic inconsistency on v1?

sqlmesh/core/config/ui.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+import typing as t
+
+from pydantic import Field
+
+from sqlmesh.core.config.base import BaseConfig
+
+
+class FormatOptions(BaseConfig):
```
tobymao, line 10: i don't think this belongs in ui.py, since this is also used for the cli isn't it

mykalmax:
do we actually need an option to override global `format` by passing flags to CLI, every `override` will be adding number of changes to the project repo and maybe be `overridden` by someone running format without `override` flags later

should we then do the same for `ui` part for config ?
(bringing back) 
```yaml
ui:
  format_on_save: true
  format_options:
    pad: 4
    indent: 4
    leading_commas: true
    max_text_width: 120
```

z3z1ma:
@mykalmax I think the user-specific config in the home directory supported by sqlmesh will already solve for user specific overrides in the ui technically. i think longer term, there should be a modal in the UI with options preset based on the `format` config. And they can tweak there in theory.

z3z1ma:
I also think CLI precedence is typical enough to expect in a tool, it allows for faster iteration -- or imperative flows which dont use the config.py/yaml. I dont think its necessary to constrain it personally 🤷

sqlmesh/core/config/ui.py
z3z1ma: Done

z3z1ma:
Thinking about it more @mykalmax , 
I could go either way really on dropping the cli flags and using just the yaml honestly
 
Since there is some simplification in the code then. And is easier to grok than stacking CLI flags. And its uniform with UI by default as you said.

sqlmesh/core/config/format.py
```diff
@@ -0,0 +1,29 @@
...
+        append_newline: Whether to append a newline to the end of the file or not.
+        normalize: Whether to normalize the SQL code or not.
+        pad: The number of spaces to use for padding.
+        indent: The number of spaces to use for indentation.
+        normalize_functions: The functions to normalize.
+        leading_comma: Whether to use leading commas or not.
+        max_text_width: The maximum text width.
+    """
+
+    append_newline: bool = Field(default=False, exclude=True)
```
izeigerman, line 23: Why do we need `exclude=False` here? I think this is what trips up Pydantic V1

mykalmax:
@z3z1ma would it make more sense to have `format` top level with `enable: true` as default (and `false` to disable formatting), then drop `ui` from config and if `format` is set for a project `format-on-save` on ui just be reading options from config.
Remove all overrides (flags), so only way to change/set format is to change it in config for entire project (or disable).
In this case if projects needs `formatting`: pass options to config and it will be consistent in cli and ui and if formatting not needed , just disable for entire project.

sqlmesh/core/config/format.py
z3z1ma: Because everything is a kwarg to the `Generator` class except this option which is handled in sqlmesh side @izeigerman

z3z1ma: So it simplifies dumping to dict and passing through to `format_model_expressions` or w/e

z3z1ma: See here:
https://github.com/z3z1ma/sqlmesh/blob/4f8d2fb29afd511ca5db4a847f964555dfd539d8/sqlmesh/core/context.py#L766-L768
https://github.com/z3z1ma/sqlmesh/blob/4f8d2fb29afd511ca5db4a847f964555dfd539d8/web/server/api/endpoints/files.py#L81

where we would otherwise need to remember to pass `exclude={"append_newline"}` to the dict call to be compatible with `Generator` @izeigerman 
The pydantic v1 error may make it inevitable, so its nbd

z3z1ma:
I think thats fine to consolidate and drop flags if everyone agrees @mykalmax ! 

I would maybe nest the ui format-on-save option though.

```yaml
format:
  pad: 4
  indent: 4
  normalize: false
  ui:
    enabled: true # auto format on save
```

I think being able to run `sqlmesh format` in the CLI and the UI **auto-formatting** files on save are 2 different concerns and the one worth disabling is the uncontrollable implicit one (the UI). The CLI is by contrast a very explicit action and a user running it, typically wants to run it. And if they dont want to run it, they dont run it. We don't need to make it a no-op for the sake of consistency with UI autoformat.

I will await decision. (can always pivot later 😄)

sqlmesh/core/config/format.py
z3z1ma: Okay i just added a dedicated prop to solve for this instead of implicit pydantic behavior
tbd if you want to change the name

sqlmesh/core/config/format.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    normalize: bool = False
+    pad: int = 2
+    indent: int = 2
+    normalize_functions: t.Optional[str] = None
+    leading_comma: bool = False
+    max_text_width: int = 80
+    append_newline: bool = False
+    ui: UIFormatterConfig = UIFormatterConfig()
```
izeigerman, line 39: This feels backwards to me. Shouldn't there be a separate top-level UI config in which there is `format: bool` flag? CC: @mykalmax. The PR looks good otherwise

sqlmesh/core/config/format.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @property
+    def generator_options(self) -> t.Dict[str, t.Any]:
+        """Options which can be passed through to the SQLGlot Generator class.
+
+        Returns:
+            The generator options.
+        """
+        return {
```
izeigerman, line 48: Can't this be just `self.dict(exclude={"append_newline"})` just like you did before?

z3z1ma: Ya it's just something you have to remember to do if trying to use the API. Instead of a prop which does it for you in 1 place that's easy to change or extend. 

Whichever way you prefer I can keep.

mykalmax:
@izeigerman @z3z1ma my suggestion was to drop `ui:` from `config` at all and if formatting is `enabled` for a project `UI` just picks up settings from `format:` in `config`. 
But with option to opt-out from `auto-formatting` on save on `UI` it should be separate top level `ui:` to avoid nesting.
Sorry for confusion 
```yaml
...

format:
  pad: 4
  indent: 4
  leading_commas: true
  max_text_width: 120

ui:
  format_on_save: true

...
```

sqlmesh/core/config/format.py
izeigerman: No, `generator_options` is fine, I like it. My question is mostly why not use `self.dict(...)` in its body?

z3z1ma: Oh I see, yes thats fair

z3z1ma:
@izeigerman can you retrigger test. Seems to be hanging? Or I can do another commit I guess

izeigerman (approved):

izeigerman:
Looks great! Thanks a lot for addressing comments 👍

PR #2075, opened 2024-02-02
fix: replace query unknown self-reference

eakmanrq:
Prior to this change if you had a self-referencing query that is full refresh (SCD Type 2 is this) then we would always do a create table if not exists since we need a table to reference if one doesn't exist. If a column was unknown then we would get an error trying to create the table.

Now we check if column types are all known in create table and if they aren't then we raise an error. The exception is if the table already exists and we get IF NOT EXISTS then we just return since we aren't creating the table anyways. This addresses the error case here since even though we have an unknown it is ok since the table is already created.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -495,6 +495,17 @@ class EngineAdapter:
             kwargs: Optional create table properties.
         """
         table = exp.to_table(table_name)
+
+        if not columns_to_types_all_known(columns_to_types):
+            # It is ok if the columns types are not known if the table already exists and IF NOT EXISTS is set
+            if exists and self.table_exists(table_name):
+                return
+            raise SQLMeshError(
+                "Cannot create a table without knowing the column types. "
+                "Try casting the columns to an expected type or defining the columns in the model metadata. "
+                f"Columns to types: {columns_to_types}"
+            )
+
         primary_key_expression = (
             [exp.PrimaryKey(expressions=[exp.to_column(k) for k in primary_key])]
             if primary_key and self.SUPPORTS_INDEXES
--- a/tests/core/engine_adapter/test_spark.py
+++ b/tests/core/engine_adapter/test_spark.py
@@ -239,10 +239,6 @@ def test_replace_query_self_ref_not_exists(
     make_mocked_engine_adapter: t.Callable, mocker: MockerFixture, make_temp_table_name: t.Callable
 ):
     mocker.patch(
-        "sqlmesh.core.engine_adapter.spark.SparkEngineAdapter.table_exists",
-        return_value=False,
-    )
-    mocker.patch(
         "sqlmesh.core.engine_adapter.spark.SparkEngineAdapter.get_current_catalog",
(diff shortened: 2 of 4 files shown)


tests/core/engine_adapter/test_spark.py
```diff
@@ -270,9 +270,10 @@ def test_replace_query_self_ref_not_exists(
 
     assert to_sql_calls(adapter) == [
         "CREATE TABLE IF NOT EXISTS `db`.`table` (`col` INT)",
-        # This second query doesn't do anything but we can't just run this because we need the table
-        # to exist since it is self-referencing
-        "CREATE TABLE IF NOT EXISTS `db`.`table` AS SELECT `col` + 1 AS `col` FROM `db`.`table`",
```
eakmanrq, line 275: This old behavior looks like a bug.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -495,6 +495,16 @@ def _create_table_from_columns(
             kwargs: Optional create table properties.
         """
         table = exp.to_table(table_name)
+
+        if not columns_to_types_all_known(columns_to_types):
+            # It is ok if the columns types are not known if the table already exists and IF NOT EXISTS is set
+            if exists and self.table_exists(table_name):
+                return
+            raise SQLMeshError(
+                f"Cannot create a table from a self-referencing query without knowing the column types. "
```
tobymao, line 504: i think it'd be helpful for the user to receive a suggestion on a solution, perhaps suggest a user casts all projections

tobymao: i'm a bit confused why this error message explicitly calls out "self-referencing". this method is called create_table_from_columns, is only ever used for self-referencing queries?

eakmanrq: Ah thank you for catching

sqlmesh/core/engine_adapter/base.py
```diff
@@ -495,6 +495,17 @@ def _create_table_from_columns(
             kwargs: Optional create table properties.
         """
         table = exp.to_table(table_name)
+
+        if not columns_to_types_all_known(columns_to_types):
+            # It is ok if the columns types are not known if the table already exists and IF NOT EXISTS is set
+            if exists and self.table_exists(table_name):
+                return
+            raise SQLMeshError(
+                "Cannot create a table from a self-referencing query without knowing the column types. "
```
tobymao, line 504: left a comment in the previous version, but i'm a bit confused by this error message. why does this explicitly call out self-referencing. the method named seems very generic _create_table_from_columns. how do you know this branch is specific to self-referencing queries?

tobymao (approved):

tests/core/engine_adapter/test_spark.py
eakmanrq: It is actually an artifact of the test itself. There is a scenario the test took where it said the table does not exist when it has to since the SQL statement above created it. So the test was updated to dynamically determine if the table exists based on if a create table command was run and now the output is correct. 

TL;DR: Test bug not an actual behavior bug.

PR #2082, opened 2024-02-03
feat: Virtualized docs list, auto scroll to active item on refresh

caiters:
Three main things being added here

- virtualizing the list of items in the docs list using tanstack react virtualizer
- when user refreshes the page, if the active item is not visible in the docs list, list will be scrolled until the active item is visible
- added a button so if you scroll away from the currently active item and want to return to it, it will be easy to do so

==

Video showing page refresh results in active item being visible, and the "scroll to" button when it is scrolled out of frame:

https://github.com/TobikoData/sqlmesh/assets/81378/469b7318-dd77-4d34-9865-9d97a26b8488

### DIFF
--- a/web/client/src/library/components/sourceList/SourceList.tsx
+++ b/web/client/src/library/components/sourceList/SourceList.tsx
@@ -265,8 +265,8 @@ export function SourceListItem({
             ? variant === EnumVariant.Primary
               ? 'text-primary-500 bg-primary-10'
               : variant === EnumVariant.Danger
-                ? 'text-danger-500 bg-danger-5'
-                : 'text-neutral-500 bg-neutral-10'
+              ? 'text-danger-500 bg-danger-5'
+              : 'text-neutral-500 bg-neutral-10'
             : 'hover:bg-neutral-10 text-neutral-400 dark:text-neutral-300',
         )
       }
--- a/web/client/src/library/components/sourceList/SourceList.tsx
+++ b/web/client/src/library/components/sourceList/SourceList.tsx
@@ -1,5 +1,5 @@
 import Input from '@components/input/Input'
-import { useVirtualizer } from '@tanstack/react-virtual'
+import { type Virtualizer, useVirtualizer } from '@tanstack/react-virtual'
 import { isArrayEmpty, isNil, isNotNil } from '@utils/index'
 import clsx from 'clsx'
 import { useEffect, useMemo, useRef, useState } from 'react'
@@ -27,7 +27,7 @@ export default function SourceList<
   items = [],
   types,
   by = 'id',
-  activeItemIndex = -1,
+  isActive,
   byName,
   byDescription,
   to,
@@ -39,7 +39,7 @@ export default function SourceList<
   to: string
   items?: TItem[]
   types?: TType
-  activeItemIndex?: number
+  isActive?: (id: string) => boolean
   byName?: string
   disabled?: boolean
   byDescription?: string
@@ -49,23 +49,51 @@ export default function SourceList<
 
   const scrollableAreaRef = useRef<HTMLDivElement>(null)
 
-  const filtered =
-    filter === ''
-      ? items
-      : items.filter(item => {
-          const id = item[by] ?? ''
-          const description = String(
-            isNil(byDescription) ? '' : item?.[byDescription] ?? '',
-          )
-          const name = String(isNil(byName) ? '' : item?.[byName] ?? '')
-          const type = String(types?.[id] ?? '')
+  const getActiveIndex = (itemsList: TItem[]): number =>
+    itemsList.findIndex(item => isNotNil(isActive) && isActive(item[by]))
 
-          return (
-            name.includes(filter) ||
-            description.includes(filter) ||
-            type.includes(filter)
-          )
-        })
+  const [activeItemIndex, filtered] = useMemo(() => {
+    if (filter === '') {
+      return [getActiveIndex(items), items]
+    }
+    let activeIndex = -1
+    const filteredList: TItem[] = []
+    items.forEach((item, index) => {
+      const id = item[by] ?? ''
+      const description = String(
+        isNil(byDescription) ? '' : item?.[byDescription] ?? '',
+      )
+      const name = String(isNil(byName) ? '' : item?.[byName] ?? '')
+      const type = String(types?.[id] ?? '')
+      if (
+        name.includes(filter) ||
+        description.includes(filter) ||
+        type.includes(filter)
+      ) {
+        filteredList.push(item)
+        if (isNotNil(isActive) && isActive(item[by])) {
+          activeIndex = index
+        }
+      }
+    })
+    // const filteredList = items.filter(item => {
+    //   const id = item[by] ?? ''
+    //   const description = String(
+    //     isNil(byDescription) ? '' : item?.[byDescription] ?? '',
+    //   )
+    //   const name = String(isNil(byName) ? '' : item?.[byName] ?? '')
+    //   const type = String(types?.[id] ?? '')
+
+    //   return (
+    //     name.includes(filter) ||
+    //     description.includes(filter) ||
+    //     type.includes(filter)
+    //   )
+    // })
+    return [activeIndex, filteredList]
+  }, [items, filter])
+
+  console.log({ activeItemIndex, filtered })
 
   const rowVirtualizer = useVirtualizer({
     count: filtered.length,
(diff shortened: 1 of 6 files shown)


web/client/src/library/pages/docs/Docs.tsx
```diff
@@ -40,15 +40,21 @@ export default function PageDocs(): JSX.Element {
...
   return (
     <Page
       sidebar={
         <SourceList
-          key={location.pathname}
           by="displayName"
           byName="displayName"
           to={EnumRoutes.IdeDocsModels}
           items={list}
+          activeItemIndex={activeItemIndex}
```
mykalmax, line 57: I guess it'll be better if we pass  `key/id` (like `active=${EnumRoutes.Docs}/${item.id}` in this case and same for other pages) instead of `index` so we can handle getting actual `index` inside component itself.

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
+  return (
+    <div className={clsx('flex flex-col w-full h-full relative', className)}>
+      {shouldShowReturnButton && (
+        <Button
+          className="absolute right-0 top-0 z-10 text-ellipsis !block overflow-hidden no-wrap max-w-[90%]"
+          onClick={() => scrollToItem({ itemIndex: filteredItemIndex })}
+          size="sm"
+          variant="neutral"
+        >
+          Scroll to {items[activeItemIndex]?.name}
```
mykalmax, line 142: maybe just `Scroll to Active` or `... Selected` , names can get pretty long What do you think if we make button `opacity-50` and `opacity-100` on hover?

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
+    return false
+  }, [filteredItemIndex, rowVirtualizer.range])
+
+  // scroll to the active item when the activeItemIndex changes
+  useEffect(() => {
+    if (
+      filteredItemIndex > -1 &&
+      isNotNil(rowVirtualizer.range) &&
+      (rowVirtualizer.range.startIndex > filteredItemIndex ||
+        rowVirtualizer.range.endIndex < filteredItemIndex)
```
mykalmax, line 127: how about refactor this to be a method?

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
+    return filteredIndex
+  }, [filtered, activeItemIndex])
+
+  /**
+   * The return button should appear when the
+   * active item is available in the list (not
+   * filtered out) and it is not in the visible
+   * range of the virtualized list
+   */
+  const shouldShowReturnButton = useMemo(() => {
```
mykalmax, line 104: I think having `useMemo` in this case not going to do much optimization

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -36,13 +39,16 @@ export default function SourceList<
...
   byName?: string
   disabled?: boolean
   byDescription?: string
   className?: string
 }): JSX.Element {
   const [filter, setFilter] = useState('')
 
+  const scrollableAreaRef = useRef<HTMLDivElement>(null)
+
   const filtered =
```
mykalmax, line 52: i think we can have this return tuple: `active index` and `filtered list` (just list if no `filter` text provided)

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
+    isSmoothScroll?: boolean
+  }): void => {
+    rowVirtualizer.scrollToIndex(itemIndex, {
+      align: 'center',
+      behavior: isSmoothScroll ? 'smooth' : 'auto',
+    })
+  }
+
+  // the index of the active item in the filtered list
+  const filteredItemIndex = useMemo(() => {
```
mykalmax, line 90: we can combine this with iteration from previous useMemo

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
+          variant="neutral"
+        >
+          Scroll to {items[activeItemIndex]?.name}
+        </Button>
+      )}
+      <div
+        className="p-2 h-full overflow-auto hover:scrollbar scrollbar--horizontal scrollbar--vertical"
+        ref={scrollableAreaRef}
+      >
+        <div
```
mykalmax, line 149: can this still be `ul` or we need to change `li`s to maybe `div`s or `span`s

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -61,51 +67,142 @@ export default function SourceList<
...
             <li
-              key={id}
-              className={clsx(
-                'text-sm font-normal',
-                disabled && 'cursor-not-allowed',
-              )}
-              tabIndex={id === filter ? -1 : 0}
+              key="not-found"
+              className="p-2"
+              onClick={() => {
```
mykalmax, line 157: maybe just keep it `not clickable`. I would not expect clicking on `No Result` to reset filter. Also i think you can add `type="search"` to Input component and it should display `x` and clear filter on `Esc`

caiters: oh yeah, the onClick was there before but the diff makes it hard to see. I definitely can remove this because I agree that it isn't very intuitive

web/client/src/library/components/sourceList/SourceList.tsx
caiters: yes thank you, I meant to revisit that after I got this all working with divs (because I had to add an extra wrapper inside) and I forgot. I'm switching it back to a list!

web/client/src/library/components/sourceList/SourceList.tsx
caiters: great point!

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -36,76 +39,187 @@ export default function SourceList<
...
-          )
-        })
+  const scrollableAreaRef = useRef<HTMLDivElement>(null)
+
+  const getActiveIndex = (itemsList: TItem[]): number =>
+    itemsList.findIndex(item => isNotNil(isActive) && isActive(item[by]))
+
+  const [activeItemIndex, filtered] = useMemo(() => {
+    if (filter === '') {
+      return [getActiveIndex(items), items]
```
mykalmax, line 57: we can handle this case inside `items.forEach` and remove `getActiveIndex`

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -36,76 +39,187 @@ export default function SourceList<
...
+        description.includes(filter) ||
+        type.includes(filter)
+      ) {
+        filteredList.push(item)
+        if (isNotNil(isActive) && isActive(item[by])) {
+          activeIndex = index
+        }
+      }
+    })
+    // const filteredList = items.filter(item => {
```
mykalmax, line 79: i guess , we can remove this

web/client/src/library/components/sourceList/SourceList.tsx
```diff
@@ -36,76 +39,187 @@ export default function SourceList<
...
+    //   return (
+    //     name.includes(filter) ||
+    //     description.includes(filter) ||
+    //     type.includes(filter)
+    //   )
+    // })
+    return [activeIndex, filteredList]
+  }, [items, filter])
+
+  console.log({ activeItemIndex, filtered })
```
mykalmax, line 96: debug

web/client/src/library/pages/docs/Docs.tsx
```diff
@@ -40,15 +40,19 @@ export default function PageDocs(): JSX.Element {
...
   return (
     <Page
       sidebar={
         <SourceList
-          key={location.pathname}
           by="displayName"
           byName="displayName"
           to={EnumRoutes.IdeDocsModels}
           items={list}
+          isActive={isActive}
```
mykalmax, line 55: Nit: 
```
<SourceList
  ...
  isActive={id => `${EnumRoutes.IdeDocsModels}/${id}` === pathname}
  ...
/>
```

web/client/src/library/components/sourceList/SourceList.tsx
caiters: dang it, thanks!

web/client/src/library/components/sourceList/SourceList.tsx
caiters: dang it, thanks!

web/client/src/library/components/sourceList/SourceList.tsx
caiters: are you thinking that since we have to loop over at least some of the array anyway that we should do the forEach regardless of whether there is a filter? (either way since getActiveIndex is no longer used more than once I can get rid of it)

mykalmax: yes, we have to loop over items to get activeIndex anyway. filter is kinda like side effect here

mykalmax (approved):

PR #2084, opened 2024-02-05
ibis example project

andrew-sha:
This demonstrates how ibis can be embedded within a sqlmesh project. The project is largely the same as the default `init` project, with the addition of `ibis_full_model.py` which is a Python model that uses ibis to replicate the model created by `full_model.sql`


examples/ibis/models/ibis_full_model.py
```diff
@@ -0,0 +1,25 @@
...
+
+@model(
+    "ibis.ibis_full_model",
+    is_sql=True,
+    kind="FULL",
+    audits=["assert_positive_order_ids"]
+)
+def entrypoint(evaluator: MacroEvaluator) -> str:
+    # connect to database
+    con = ibis.duckdb.connect(database="data/local.duckdb")
```
crericha, line 14: @tobymao is it possible to get the database from `MacroEvaluator`? If not, maybe use an environment variable here and in config.py, so we don't have to duplicate the database file location.

tobymao: the macro evaluator has a reference to engine_adapter

tobymao: i'd just stick with a constant at that point.

examples/ibis/models/ibis_full_model.py
```diff
@@ -0,0 +1,19 @@
...
+from sqlmesh.core.model import model
+
+
+@model("ibis.ibis_full_model", is_sql=True, kind="FULL", audits=["assert_positive_order_ids"])
+def entrypoint(evaluator: MacroEvaluator) -> str:
+    # connect to database
+    con = ibis.duckdb.connect(database="data/local.duckdb")
+
+    # retrieve table
+    incremental_model = con.table("incremental_model", schema="ibis")
```
tobymao, line 13: you need to get the table from the evaluator so that dependencies are properly tracked.

table_name = context.table("ibis.incremental_model")

examples/ibis/models/ibis_full_model.py
tobymao: hm wait a second, this is a python sql model and not a python model. is it possible to have one example for both? in this case, why do you need any connection to the database, you just need to generate sql?

examples/ibis/models/ibis_full_model_python.py
```diff
@@ -0,0 +1,41 @@
...
+)
+def execute(
+    context: ExecutionContext,
+    start: datetime,
+    end: datetime,
+    execution_time: datetime,
+    **kwargs: t.Any,
+) -> pd.DataFrame:
+    # connect ibis to database
+    con = ibis.duckdb.connect(database="data/local.duckdb")
```
tobymao, line 28: can you make data/local.duckdb a constant within the project and import it?

examples/ibis/models/ibis_full_model_python.py
```diff
@@ -0,0 +1,41 @@
...
+    # retrieve table
+    incremental_model = con.table("incremental_model", schema="ibis")
+
+    # build query
+    count = incremental_model.id.nunique()
+    aggregate = incremental_model.group_by("item_id").aggregate(num_orders=count)
+    query = aggregate.order_by("item_id")
+
+    # convert ibis table into dataframe
+    df = query.to_pandas()
```
tobymao, line 39: return query.to_pandas()

examples/ibis/models/ibis_full_model_python.py
```diff
@@ -0,0 +1,41 @@
...
+from sqlmesh.core.model import FullKind
+
+
+@model(
+    "ibis.ibis_full_model_python",
+    kind=FullKind(),
+    columns={
+        "item_id": "int",
+        "num_orders": "int",
+    },
```
tobymao, line 17: could you write a description here explaining this is an ibis model that actually runs query and the other one is an ibis model that generates sql?

examples/ibis/models/ibis_full_model_sql.py
```diff
@@ -0,0 +1,26 @@
+import ibis  # type: ignore
+from ibis.expr.operations import Namespace, UnboundTable  # type: ignore
+
+from sqlmesh.core.macros import MacroEvaluator
+from sqlmesh.core.model import model
+
+
+@model("ibis.ibis_full_model_sql", is_sql=True, kind="FULL", audits=["assert_positive_order_ids"])
+
+# This model uses ibis to generate a SQL string.
```
tobymao, line 10: are these passed through automatically, or do you have to do description= in the @model tag?

andrew-sha: oh I didn't even realize you could include a description in the tag, will change

tobymao (approved):

izeigerman:
Have we tested this with Airflow?

tobymao:
> Have we tested this with Airflow?

i don't think it's necessary, it's just an example project.

PR #2085, opened 2024-02-05
feat: add dbt scd type 2 feature flag

eakmanrq:
Adds a feature flag for disabling SCD Type 2 Support (or called "Snapshots" in dbt) using the dbt adapter. This would disable all Snapshot models. The plan is to remove this in a few weeks if we find it is not needed as we continue to rollout full Snapshot support to dbt models.

### DIFF
--- a/sqlmesh/core/config/feature_flag.py
+++ b/sqlmesh/core/config/feature_flag.py
@@ -1,0 +1,9 @@
+from sqlmesh.utils.pydantic import PydanticModel
+
+
+class DbtFeatureFlag(PydanticModel):
+    scd_type_2_support: bool = True
+
+
+class FeatureFlag(PydanticModel):
+    dbt: DbtFeatureFlag = DbtFeatureFlag()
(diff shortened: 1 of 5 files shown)


sqlmesh/core/config/feature_flag.py
```diff
@@ -0,0 +1,9 @@
+from sqlmesh.utils.pydantic import PydanticModel
+
+
+class DbtFeatureFlag(PydanticModel):
+    scd_type_2_support: bool = True
+
+
+class FeatureFlag(PydanticModel):
+    dbt: DbtFeatureFlag = DbtFeatureFlag()
```
crericha, line 9: We want the feature flag to be dbt specific?

tobymao: my guess is that this is a categorization for dbt specific feature flags

eakmanrq: Yeah this feature flag is dbt specific. If native users start using a new SCD Type 2 feature and they hit an error they can just remove it and not affect prod. In dbt, since we used to ignore snapshots and now we process them, their prod becomes affected and therefore this FF is an easy to reach stability again. Therefore it is dbt focused.

izeigerman (approved):

PR #2086, opened 2024-02-05
Fix: do not convert missing seed string column values to string

treysp:
If a seed file contained a missing value in a string column, its `NaN` value was coerced to string `"nan"` with pandas `astype(str)`. This PR ensures only non-missing values are coerced to string.

Also:
- Converts `NaN` values to `None` for MSSQL engine pymssql library
- Adds integration test confirming successful `NaN` -> db `NULL` -> `NaN`/`None` roundtrip

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1171,7 +1171,7 @@ class SeedModel(_SqlBasedModel):
                 df[column] = pd.to_datetime(df[column])
             df[bool_columns] = df[bool_columns].apply(lambda i: str_to_bool(str(i)))
             df[string_columns] = df[string_columns].mask(
-                lambda x: x.notna(), df[string_columns].astype(str)  # type: ignore
+                cond=lambda x: x.notna(), other=df[string_columns].astype(str)  # type: ignore
             )
             yield df
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1170,7 +1170,9 @@ class SeedModel(_SqlBasedModel):
             for column in date_or_time_columns:
                 df[column] = pd.to_datetime(df[column])
             df[bool_columns] = df[bool_columns].apply(lambda i: str_to_bool(str(i)))
-            df[string_columns] = df[string_columns].astype(str)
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1170,7 +1170,9 @@ def render(
             for column in date_or_time_columns:
                 df[column] = pd.to_datetime(df[column])
             df[bool_columns] = df[bool_columns].apply(lambda i: str_to_bool(str(i)))
-            df[string_columns] = df[string_columns].astype(str)
+            df[string_columns] = df[string_columns].mask(
+                lambda x: x.notna(), df[string_columns].astype(str)  # type: ignore
```
tobymao, line 1174: what does this do? specifically the mask part

tests/core/engine_adapter/test_integration.py
```diff
@@ -820,6 +821,34 @@ def test_drop_schema(ctx: TestContext):
...
+        ]
+    )
+    ctx.engine_adapter.create_table(table, ctx.columns_to_types)
+    ctx.engine_adapter.replace_query(
+        table,
+        ctx.input_data(input_data),
+        columns_to_types=ctx.columns_to_types,
+    )
+    results = ctx.get_metadata_results()
+    assert len(results.views) == 0
```
tobymao, line 845: you could just do assert not results.views

sqlmesh/core/model/definition.py
tobymao: is this saying, if x is not na, convert it to a string?

treysp: It's an if-else construct - if not NA, assign astype(str), else keep existing value.

There's probably a more idiomatic way to do that in pandas.

eakmanrq: Yeah I wondered the same and I believe that is it. It might help to use kwargs here to make it easier to understand:
`mask(cond=lambda x: x.notna(), other=df[string_columns].astype(str))`

tobymao: is this call vectorized?

tobymao (approved):

eakmanrq (approved):

PR #2091, opened 2024-02-06
Chore: test CLI plan/run commands

treysp:


tests/cli/test_cli.py
```diff
@@ -0,0 +1,169 @@
...
+
+default_gateway: local
+
+model_defaults:
+    dialect: duckdb
+"""
+        )
+
+
+def reset_example_project(temp_dir) -> None:
```
eakmanrq, line 39: I wonder if we should just split each "reset" into a separate test. It would result in specific commands having multiple tests but I would prefer that over having to reset state between tests and potentially have that shared state, if not properly reset, affect the results.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,169 @@
...
+    reset_example_project(tmp_path)
+
+    # Successful test run message should not appear if `--skip-tests` is passed and plan is applied
+    result = runner.invoke(cli, ["--paths", tmp_path, "plan", "--skip-tests"], input="y\n")
+    assert result.exit_code == 0
+    assert "Successfully Ran 1 tests against duckdb" not in result.output
+    assert_new_env(result)
+    assert_backfill_success(result)
+
+    rmtree(tmp_path)
```
eakmanrq, line 91: This shouldn't be needed. `tmp_path` is unique per test so each test is running in it's own isolated space and therefore you can let pytest deal with the cleanup. It actually can be nice to leave the old path around for debugging purposes. I think pytest clears out past tests once they are over 5 test runs old.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+def assert_virtual_update(result) -> None:
+    assert "Virtual Update executed successfully" in result.output
+
+
+def test_version(runner):
+    result = runner.invoke(cli, ["--version"])
+    assert result.exit_code == 0
+    assert re.fullmatch("\d+\.\d+\.\d+(\..*)?\n?", result.output)
```
eakmanrq, line 130: Don't we know what version should be reported based on the value in the version file? I'm wondering if we should assert we get that as a response instead of the softer assert that we get something that matches a regex pattern.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+    assert_plan_success(result)
+
+
+def test_plan_skip_tests(runner, tmp_path):
+    create_example_project(tmp_path)
+
+    # Successful test run message should not appear if `--skip-tests` is passed and plan is applied
+    result = runner.invoke(cli, ["--paths", tmp_path, "plan", "--skip-tests"], input="y\n")
+    assert result.exit_code == 0
+    assert "Successfully Ran 1 tests against duckdb" not in result.output
```
eakmanrq, line 155: I think we should assert this does show up in the `test_plan` test. Right now we could never print that tests passed and we wouldn't have a test fail here. 

Btw I'm saying to put it in `test_plan` since I was originally thinking "Oh have a test that makes sure that we see tests when `--skip-tests` is not passed in" but that would be exactly what `test_plan` is.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+    assert_new_env(result)
+    assert_backfill_success(result)
+
+
+def test_plan_restate_model(runner, tmp_path):
+    create_example_project(tmp_path)
+
+    # Create prod and backfill
+    result = runner.invoke(cli, ["--paths", tmp_path, "plan", "--auto-apply"])
```
eakmanrq, line 164: I'm wondering if this should be a dedicated function to init prod. Could also do some simple asserts to check that prod was indeed created.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+    # plan with no changes and full_model restated
+    result = runner.invoke(
+        cli,
+        ["--paths", tmp_path, "plan", "--restate-model", "sqlmesh_example.full_model"],
+        input="\n\ny\n",
```
eakmanrq, line 170: Could you add comment explaining what these inputs are mimicking? I think it is skipping a start and end and then confirming apply but it would be nice to have it spelled out.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+def test_plan_verbose(runner, tmp_path):
+    create_example_project(tmp_path)
+
+    result = runner.invoke(cli, ["--paths", tmp_path, "plan", "--verbose"], input="y\n")
+    assert_plan_success(result)
+    assert "sqlmesh_example.seed_model created" in result.output
+    assert "sqlmesh_example.seed_model evaluated in" in result.output
+    assert "sqlmesh_example.seed_model promoted" in result.output
```
eakmanrq, line 214: Want to assert that these do not show up in the test above this? Right now these could always show up and not fail a test.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,505 @@
...
+    create_example_project(tmp_path)
+
+    # Create and backfill `prod` environment
+    runner.invoke(cli, ["--paths", tmp_path, "plan", "--no-prompts", "--auto-apply"])
+
+    # No error and no output if `prod` environment exists and cron has not elapsed
+    result = runner.invoke(cli, ["--paths", tmp_path, "run"])
+    assert result.exit_code == 0
+
+    # INFO messages are printed to console when tests run in some environments.
```
eakmanrq, line 487: What do you mean by some environments here? Btw pytest has a `caplog` fixture that can be used to check what is logged during a test. Not sure if that would be helpful here.

tests/cli/test_cli.py
treysp: `assert_plan_success()` covers that - it includes a test that "Successfully Ran 1 tests against duckdb" is in the output

eakmanrq: Ah sorry missed that. Thanks!

tests/cli/test_cli.py
treysp: Yeah, it's weird. 

The test was passing with `output == ""` when I ran the test from VSCode or with `pytest <file name>` in CLI, but when I ran `make fast-test` the output contained those info messages and test failed.

I assumed that we can't fully control the behavior of info messages across all places these tests will run, so this version of the assert handles output with or without them.

eakmanrq (approved):
Can you mark this as slow? Example: https://github.com/TobikoData/sqlmesh/blob/7bf0c16761370536edb2ae596d94204b6f4b410d/tests/core/test_state_sync.py#L48

PR #2093, opened 2024-02-06
fix!: airflow truncate tag and task id names

eakmanrq:
Prior to this change users could hit character limits on either the snapshot name tag on a dag (100 character limit) or issues sending metrics through statsd for task instances (250 character limit). This limits both of those.

### DIFF
--- a/sqlmesh/schedulers/airflow/dag_generator.py
+++ b/sqlmesh/schedulers/airflow/dag_generator.py
@@ -30,6 +30,7 @@ from sqlmesh.schedulers.airflow.operators.hwm_sensor import (
 from sqlmesh.schedulers.airflow.operators.notification import (
     BaseNotificationOperatorProvider,
 )
+from sqlmesh.schedulers.airflow.util import truncate_task_id_if_needed
 from sqlmesh.utils import sanitize_name
 from sqlmesh.utils.date import TimeLike, to_datetime, yesterday_timestamp
 from sqlmesh.utils.errors import SQLMeshError
@@ -59,6 +60,8 @@ DAG_DEFAULT_ARGS = {
         )
     ),
 }
+
+AIRFLOW_TAG_CHARACTER_LIMIT = 100
 
 
 class SnapshotDagGenerator:
@@ -116,7 +119,7 @@ class SnapshotDagGenerator:
             tags=[
                 common.SQLMESH_AIRFLOW_TAG,
                 common.SNAPSHOT_AIRFLOW_TAG,
-                snapshot.name,
+                snapshot.name[-AIRFLOW_TAG_CHARACTER_LIMIT:],
             ],
             default_args={
                 **DAG_DEFAULT_ARGS,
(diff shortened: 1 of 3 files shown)


sqlmesh/schedulers/airflow/dag_generator.py
```diff
@@ -116,15 +119,15 @@ def _create_cadence_dag_for_snapshot(
             tags=[
                 common.SQLMESH_AIRFLOW_TAG,
                 common.SNAPSHOT_AIRFLOW_TAG,
-                snapshot.name,
+                snapshot.name[-AIRFLOW_TAG_CHARACTER_LIMIT:],
```
tobymao, line 122: is this going to have any unexpected issues, what about collision etc?

eakmanrq: This tag is non-functional so even if there is a collision it will not have any functional effect.

For the task ids, the ones that include the snapshot name also include the version. Since we keep the trailing characters our hope is that the version is included and therefore should have a low chance of collision. Collision though will always be a risk if we are truncating. 

One thing to note in this implementation I don't define what is the minimal required for a given task ID to have a very low chance of collision and then error or avoid truncating if we don't hit that. I basically say that as long as _some_ of the task ID is preserved then it was an ok truncation. So current approach is fairly naive with a basic safety check and we could go deeper if we need.

sqlmesh/schedulers/airflow/util.py
```diff
@@ -169,3 +171,25 @@ def discover_engine_operator(name: str, sql_only: bool = False) -> t.Type[BaseOp
...
+def truncate_task_id_if_needed(
+    dag_id: str,
+    task_id: str,
+    max_length: int = (STATSD_METRIC_NAME_MAX_LENGTH - STATSD_METRIC_PREFIX_LENGTH),
+    dag_task_sep: str = ".",
+) -> str:
+    task_id_max_length = max_length - len(dag_id) - len(dag_task_sep)
+    if task_id_max_length <= 0:
+        raise SQLMeshError(
+            f"The dag id '{dag_id}' is too long for statsd and therefore task_id cannot be truncated"
```
tobymao, line 185: do we need to error here? i think there's a workaround where in statsd you can rename the metrics yourself. if you raise here, what should the user do to fix this?

eakmanrq: Some context here is that this function is just returning the task ID to use within the Airflow DAG. The statsd metric name is then derived from the dag name + task id. Therefore we aren't directly setting the metric name but rather truncating the task name which is then used to create the metric name.

>if you raise here, what should the user do to fix this?

The hope is that this doesn't happen. Their dag name would need to be so long that we can't fit in any task id. It would suggest we need to do more aggresive truncating to make this work at attack this problem from a different angle.

eakmanrq:
Using different approach instead: https://github.com/TobikoData/sqlmesh/pull/2095

PR #2095, opened 2024-02-07
fix: airflow use model name instead of snapshot name

eakmanrq:
In order to help reduce the change of character limits when sending stats in Airflow for tasks, we are switching to model name instead of snapshot name for the task ids. We also truncate the tag model name to the max characters to ensure users don't get an error when defining this non-functional tag.


izeigerman (approved):
Seems like applying the same approach to DAG IDs is not as straightforward. I think this is the right direction though.

PR #2097, opened 2024-02-07
Fix: Set min-width for split panes

mykalmax:


caiters (approved):
what you have here looks good to me - I don't know if it's also a SplitPane but is there a minimum width on the lineage panel as well when it is in column view?

mykalmax:
We still want to have ability to resize to 0 in case where there is no Data to preview

PR #2101, opened 2024-02-07
Chore!: bump sqlglot to v21.0.0

georgesittas:
Was super annoying to detect the need for the L458 change in schema_diff.py (triggered by a recent [change](https://github.com/tobymao/sqlglot/commit/b4e886877ecfbafdd64c515c765c3c54764bd987#diff-7857fedd1d1451b1b9a5b8efaa1cc292c02e7ee4f0d04d7e2f9d5bfb9565802cR3837) in sqlglot), may need to refactor this at some point to avoid the mutations if possible.

### DIFF
--- a/sqlmesh/core/schema_diff.py
+++ b/sqlmesh/core/schema_diff.py
@@ -453,7 +453,9 @@ class SchemaDiffer(PydanticModel):
         root_struct: exp.DataType,
         new_kwarg: exp.ColumnDef,
     ) -> t.List[TableAlterOperation]:
-        current_type = exp.DataType.build(current_type)
+        # We don't copy on purpose here because current_type may need to be mutated inside
+        # _get_operations (struct.expressions.pop and struct.expressions.insert)
+        current_type = exp.DataType.build(current_type, copy=False)
         if self.support_nested_operations:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/schema_diff.py
```diff
@@ -453,7 +453,9 @@ def _alter_operation(
         root_struct: exp.DataType,
         new_kwarg: exp.ColumnDef,
     ) -> t.List[TableAlterOperation]:
-        current_type = exp.DataType.build(current_type)
+        # We don't copy on purpose here because current_type may need to be mutated inside
```
tobymao, line 456: why does data.build mutate?

georgesittas: It doesn't mutate, it copies `current_type` which caused an issue in this module because we are mutating the struct expression in several places (search for `.pop` and `.insert` calls, as mentioned in the comment)

georgesittas: so before my sqlglot commit we weren't copying inside `DataType.build` and this worked because the mutations were being propagated here, now that we copy they don't unless I switch it off here

tobymao (approved):

PR #2107, opened 2024-02-08
test: add test to reproduce missing documentation on bigquery `STRUCT` fields

plaflamme:
I noticed that BigQuery `STRUCT` fields do not get their comment added as their "description", e.g.:

```sql
MODEL(
 name foo
);
SELECT
  STRUCT<
     a INT, -- this comment will not be added
  > s, -- this comment will be added
FROM bar
```

This adds a test to reproduce the issue. I'm not sure how to refer to struct inner fields in the dict of field comments (i.e.: `Dict[str, str]`), but it should be close enough to understand.


tobymao:
discussed on slack, core team is not planning to implement but open to external contributions

PR #2108, opened 2024-02-08
Feat!: Automatically compute previews for changed forward-only models in dev

izeigerman:
This introduces the additional `--enable-preview` flag (and a corresponding config option) to the plan command, which enables automatic computation of data previews for changes applied to forward-only models in development environments. 

Previously, to achieve the same outcome (compute previews), users had to run two plan commands: 1) to apply forward-only changes, and 2) to restate affected models in dev so that the portion of data is recomputed to make the preview available.

### DIFF
--- a/sqlmesh/core/config/root.py
+++ b/sqlmesh/core/config/root.py
@@ -13,7 +13,6 @@ from sqlmesh.cicd.config import CICDBotConfig
 from sqlmesh.core import constants as c
 from sqlmesh.core.config import EnvironmentSuffixTarget
 from sqlmesh.core.config.base import BaseConfig, UpdateStrategy
-from sqlmesh.core.config.categorizer import CategorizerConfig
 from sqlmesh.core.config.connection import (
     ConnectionConfig,
     DuckDBConnectionConfig,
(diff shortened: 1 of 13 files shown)


sqlmesh/core/config/root.py
```diff
@@ -149,10 +146,16 @@ def _validate_regex_keys(
 
     @model_validator(mode="before")
     @model_validator_v1_args
-    def _normalize_fields(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+    def _normalize_and_validate_fields(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
         if "gateways" not in values and "gateway" in values:
             values["gateways"] = values.pop("gateway")
 
+        for plan_deprecated in ("auto_categorize_changes", "include_unmodified"):
```
tobymao, line 153: this is a breaking change for users so we should do feat!

tobymao (approved):

sqlmesh/core/config/root.py
izeigerman: thanks, good point

PR #2110, opened 2024-02-09
fix: init bigquery with correct type

eakmanrq:
If you init the example project it won't work because we try to put a string in for the partition column. This makes it so that bigquery uses a date data type.


tobymao:
can we always just use date?

PR #2112, opened 2024-02-09
feat!: add tstz support

eakmanrq:
TODO: Add more tests

Currently if you are using BigQuery then the `*_ts` macros will create `DATETIME` objects and therefore if you are using `TIMESTAMP` then you will run into errors. `*_tstz` is being introduced to allow users to have `TIMESTAMP` objects to use instead.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -14,7 +14,6 @@ import itertools
 import logging
 import sys
 import typing as t
-from datetime import datetime, timezone
 from functools import partial
 
 import pandas as pd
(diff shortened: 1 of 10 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -152,23 +151,11 @@ def is_pandas_df(cls, value: t.Any) -> bool:
...
-            if isinstance(ts, exp.Literal):
-                if not ts.is_string:
-                    raise SQLMeshError("Timestamp literal must be a string")
-                ts = ts.name
-            return exp.Literal.string(
-                datetime.fromisoformat(ts).astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S")
-            )
-
-        return exp.cast(exp.cast(ensure_utc_exp(col), "TIMESTAMP"), time_data_type)
+        column_to_cast = col if isinstance(col, exp.Null) else exp.Literal.string(to_ts(col))
```
tobymao, line 157: what are these changes?

eakmanrq:
Split out across these two PRs:
https://github.com/TobikoData/sqlmesh/pull/2112
https://github.com/TobikoData/sqlmesh/pull/2119

PR #2121, opened 2024-02-13
feat!: add tstz support

eakmanrq:
Prior to this change `ts` could contain timezone information if that was passed in and would cause issues for some engines. Now we make sure that `ts` does not contain timezone and `tstz` does. 

Took the opportunity to reduce some extra code in engine adapter and write some tests to ensure we could cast the values correctly across all engines. Also ran the engine adapter integration tests across all engines (including remote) to ensure it works.

### DIFF
--- a/tests/core/engine_adapter/test_integration.py
+++ b/tests/core/engine_adapter/test_integration.py
@@ -5,7 +5,7 @@ import os
 import pathlib
 import sys
 import typing as t
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 import numpy as np
 import pandas as pd
@@ -17,7 +17,7 @@ from sqlmesh.core.config import load_config_from_paths
 from sqlmesh.core.dialect import normalize_model_name
 from sqlmesh.core.engine_adapter.shared import DataObject, DataObjectType
 from sqlmesh.utils import random_id
-from sqlmesh.utils.date import now, to_date, to_ds, yesterday
+from sqlmesh.utils.date import now, to_date, to_ds, to_time_column, yesterday
 from sqlmesh.utils.errors import UnsupportedCatalogOperationError
 from sqlmesh.utils.pydantic import PydanticModel
 from tests.conftest import SushiDataValidator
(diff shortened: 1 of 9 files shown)


tobymao (approved):

tests/core/engine_adapter/test_integration.py
```diff
@@ -1898,3 +1898,85 @@ def test_dialects(ctx: TestContext):
...
+                "duckdb": pd.Timestamp("2020-01-01"),
+            },
+        ),
+        (
+            "2020-01-01 00:00:00+00:00",
+            exp.DataType.build("TIMESTAMPTZ"),
+            None,
+            {
+                "default": pd.Timestamp("2020-01-01 00:00:00+00:00"),
+                # TODO: Figure out why this is bytes and if this is really ok
```
treysp, line 1935: Unfortunately, this is an outstanding issue with pymssql https://github.com/pymssql/pymssql/issues/649

tests/core/engine_adapter/test_integration.py
```diff
@@ -1898,3 +1898,85 @@ def test_dialects(ctx: TestContext):
...
+                "spark": pd.Timestamp("2020-01-01 00:00:00"),
+            },
+        ),
+        (
+            "2020-01-01 00:00:00+00:00",
+            exp.DataType.build("TIMESTAMP"),
+            None,
+            {
+                "default": pd.Timestamp("2020-01-01 00:00:00"),
+                # TODO: Figure out why this has a timezone and if this is really ok
```
treysp, line 1947: Databricks' timestamp type is tz-aware: 
"Represents values comprising values of fields year, month, day, hour, minute, and second, with the session local time-zone. The timestamp value represents an absolute point in time." https://docs.databricks.com/en/sql/language-manual/data-types/timestamp-type.html

They are adding a non-aware version `TIMESTAMP_NTZ` that's currently in public preview  - you have to specify a table option to use it: 
"Feature support is enabled automatically when you create a new Delta table with a column of TIMESTAMP_NTZ type. It is not enabled automatically when you add a column of TIMESTAMP_NTZ type to an existing table. To enable support for TIMESTAMP_NTZ columns, support for the feature must be explicitly enabled for the existing table." https://docs.databricks.com/en/sql/language-manual/data-types/timestamp-ntz-type.html

tests/core/engine_adapter/test_integration.py
```diff
@@ -1898,3 +1898,85 @@ def test_dialects(ctx: TestContext):
...
+                "default": 20200101,
+            },
+        ),
+    ],
+)
+def test_to_time_column(
+    ctx: TestContext, time_column, time_column_type, time_column_format, result
+):
+    if ctx.test_type != "query":
+        pytest.skip("Sushi end-to-end tests only need to run for query")
```
treysp, line 1973:
```suggestion
        pytest.skip("Time column tests only need to run for query")
```

treysp (approved):

tests/core/engine_adapter/test_integration.py
eakmanrq: Thanks for linking that! Will update the comment.

tests/core/engine_adapter/test_integration.py
eakmanrq: Ah ok will update the comment

tests/core/engine_adapter/test_integration.py
eakmanrq: thank you!

PR #2125, opened 2024-02-14
feat!: include sqlmesh in version check

eakmanrq:
Prior to this change we considered SQLMesh versions "compatible", or that they could be run together, if they shared a common sqlglot minor version and were on the same migration release. Sometimes though we introduce incompatibility that doesn't require a migration and we represent this with a minor version bump. Therefore this change now considers SQLMesh's minor version too when deciding if releases are compatible.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -42,6 +42,10 @@ class Versions(PydanticModel):
     def minor_sqlglot_version(self) -> t.Tuple[int, int]:
         return major_minor(self.sqlglot_version)
 
+    @property
+    def minor_sqlmesh_version(self) -> t.Tuple[int, int]:
+        return major_minor(self.sqlmesh_version)
+
     @field_validator("sqlglot_version", "sqlmesh_version", mode="before")
     @classmethod
     def _package_version_validator(cls, v: t.Any) -> str:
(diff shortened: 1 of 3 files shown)


tests/core/test_state_sync.py
```diff
@@ -1250,29 +1250,65 @@ def test_get_version(state_sync: EngineAdapterStateSync) -> None:
...
-    # sqlglot version is behind, always raise
+    # sqlglot version is behind
     sqlglot_version = f"{major}.{int(minor) + 1}.{patch}"
     error = (
         rf"SQLGlot \(local\) is using version '{SQLGLOT_VERSION}' which is behind '{sqlglot_version}' \(remote\). "
         rf"""Please upgrade SQLGlot \('pip install --upgrade "sqlglot=={sqlglot_version}"' command\)."""
     )
     state_sync._update_versions(sqlglot_version=sqlglot_version)
+    with pytest.raises(SQLMeshError, match=error):
+        state_sync.get_versions()
```
eakmanrq, line 1299: This looks like a typo in this test. `error` was defined but never used. Updated to check for it similar to what is done below.

sqlmesh/core/state_sync/base.py
```diff
@@ -220,12 +226,23 @@ def raise_error(
                     remote_package_version=versions.sqlglot_version,
                 )
 
+            if major_minor(SQLMESH_VERSION) < major_minor(versions.sqlmesh_version):
```
tobymao, line 229: can this just be 

if ... != ...
raise_eror(ahead == ...)

Issue #1998, opened 2024-01-21
INCREMENTAL_BY_TIME_RANGE models generate invalid SQL on Trino / Iceberg


erindru-BitMEX:
**Note:** I'm aware that only Hive is officially supported in the SQLMesh Trino connector, i'm just raising this because I think Iceberg is on the todo list so this information might be useful when implementing Iceberg support.

Versions:
- Trino 435 with Iceberg connector pointed at AWS Glue Data Catalog / S3
- sqlmesh 0.68.0

Model:
```
MODEL (
    name landing.model1,
    kind INCREMENTAL_BY_TIME_RANGE (
        time_column timestamp,
    ),
    start '2023-06-01',
    cron '0 * * * *',
);

select id, timestamp
from base_schema.base_table
```

Generated SQL:
```
CREATE TABLE IF NOT EXISTS "catalog"."sqlmesh__landing"."model1__3844959144__temp"
WITH (PARTITIONED_BY=ARRAY['timestamp'])
AS SELECT ..... <rest of query is fine>
```

This produces an error:
`io.trino.spi.TrinoException: Catalog 'catalog' table property 'partitioned_by' does not exist`

The issue is that the table property for partitioning is called `partitioning` in the [Iceberg connector](https://trino.io/docs/current/connector/iceberg.html#partitioned-tables), not `partitioned_by`.

Note that `partitioned_by` is completely valid for a Hive table, so this isnt a bug per se. It's annoying that Trino isn't consistent with its property naming across official connectors.

tobymao:
this is a sqlglot issue cc @georgesittas

georgesittas:
so, depending on the connector, the trino property syntax changes?

georgesittas:
Yeah, we're able to control the generation logic at runtime as well using dialect options so this shouldn't be too much to fix once we understand the scope a bit better.

treysp:
@erindru Did #2129 complete this, or is there remaining work to do?

PR #2129, opened 2024-02-15
Feat: Add support for Trino Iceberg tables

erindru:
- Add the Iceberg connector to the integration test Trino instance
- Add a configuration to run the integration tests against the Iceberg connector
- Add the ability to probe the type of a catalog within an EngineAdapter
- Use this ability to change how table properties are generated in the Trino adapter so it can use `partitioned_by` for Hive tables and `partitioning` for Iceberg tables

ref: Issue #1998

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -266,6 +266,10 @@ class EngineAdapter:
         """Intended to be overridden for data virtualization systems like Trino that,
         depending on the target catalog, require slightly different properties to be set when creating / updating tables
         """
+        if self.CATALOG_SUPPORT.is_unsupported:
+            raise UnsupportedCatalogOperationError(
+                f"{self.dialect} does not support catalogs and a catalog was provided: {catalog}"
+            )
         return self.DEFAULT_CATALOG_TYPE
 
     @property
@@ -687,7 +691,9 @@ class EngineAdapter:
         if not isinstance(table_name_or_schema, exp.Schema):
             table_name_or_schema = exp.to_table(table_name_or_schema)
         properties = (
-            self._build_table_properties_exp(**kwargs, columns_to_types=columns_to_types)
+            self._build_table_properties_exp(
+                **kwargs, table=table_name_or_schema.this, columns_to_types=columns_to_types
+            )
             if kwargs
             else None
         )
@@ -1859,6 +1865,7 @@ class EngineAdapter:
 
     def _build_table_properties_exp(
         self,
+        table: exp.Table,
         storage_format: t.Optional[str] = None,
         partitioned_by: t.Optional[t.List[exp.Expression]] = None,
         partition_interval_unit: t.Optional[IntervalUnit] = None,
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -95,6 +95,7 @@ class EngineAdapter:
     SUPPORTS_ROW_LEVEL_OP = True
     HAS_VIEW_BINDING = False
     SUPPORTS_REPLACE_TABLE = True
+    DEFAULT_CATALOG_TYPE = DIALECT
 
     def __init__(
         self,
@@ -260,6 +261,16 @@ class EngineAdapter:
     def set_current_catalog(self, catalog: str) -> None:
         """Sets the catalog name of the current connection."""
         raise NotImplementedError()
+
+    def get_catalog_type(self, catalog: t.Optional[str]) -> str:
+        """Intended to be overridden for data virtualization systems like Trino that,
+        depending on the target catalog, require slightly different properties to be set when creating / updating tables
+        """
+        return self.DEFAULT_CATALOG_TYPE
+
+    @property
+    def current_catalog_type(self) -> str:
+        return self.get_catalog_type(self.get_current_catalog())
 
     def replace_query(
         self,
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -149,8 +149,7 @@ class HiveMetastoreTablePropertiesMixin(EngineAdapter):
                         f"PARTITIONED BY contains non-column value '{expr.sql(dialect='spark')}'."
                     )
 
-            property: exp.Property
-            property = exp.PartitionedByProperty(
+            property: exp.Property = exp.PartitionedByProperty(
                 this=exp.Schema(expressions=partitioned_by),
             )
 
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -128,6 +128,7 @@ class InsertOverwriteWithMergeMixin(EngineAdapter):
 class HiveMetastoreTablePropertiesMixin(EngineAdapter):
     def _build_table_properties_exp(
         self,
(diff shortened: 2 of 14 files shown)


tests/core/engine_adapter/docker-compose.yaml
```diff
@@ -50,47 +50,54 @@ services:
     volumes:
       - ./trino/catalog:/etc/trino/catalog
 
-  trino_datalake_metastore_db:
+  trino_metastore_db:
```
erindru, line 53: I refactored this slightly to make it take up less memory. Rather than having an entire Postgres instance per metastore, I changed it to a single Postgres instance with a database per metastore

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -147,11 +147,19 @@ def _build_table_properties_exp(
...
                     )
-            properties.append(
-                exp.PartitionedByProperty(
-                    this=exp.Schema(expressions=partitioned_by),
+
+            if self.dialect == "trino" and self.current_catalog_type == "iceberg":
+                # On the Trino Iceberg catalog, the table property is called "partitioning" - not "partitioned_by"
+                # ref: https://trino.io/docs/current/connector/iceberg.html#table-properties
+                properties.append(
+                    exp.Property(this="partitioning", value=exp.Schema(expressions=partitioned_by))
```
erindru, line 155: I'm not super familiar with sqlglot, is this a valid approach?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -261,6 +262,16 @@ def set_current_catalog(self, catalog: str) -> None:
...
 
+    def get_catalog_type(self, catalog: t.Optional[str]) -> str:
+        """Intended to be overridden for data virtualization systems like Trino that,
+        depending on the target catalog, require slightly different properties to be set when creating / updating tables
+        """
+        return self.DEFAULT_CATALOG_TYPE
+
+    @property
+    def current_catalog_type(self) -> str:
+        return self.get_catalog_type(self.get_current_catalog())
```
eakmanrq, line 273: It seems like we should raise `UnsupportedCatalogOperationError` here if `self.CATALOG_SUPPORt.is_unsupported`. It is strange for example for MySQL to return a `mysql` current catalog when it doesn't support catalogs at all.

sqlmesh/core/engine_adapter/mixins.py
eakmanrq: @georgesittas Can you confirm the best approach here?

georgesittas: I'm not quite sure - what SQL code is this supposed to generate? Is there a way we can add a test to ensure we generate the correct DDL statement?

One observation is that instead of `this="partitioning"`, we may need to do something like `this=exp.var("partitioning")` to have an expression in `this` vs a string (which is usually the case for `Property` expressions IIRC).

sqlmesh/core/engine_adapter/trino.py
```diff
@@ -58,6 +59,15 @@ def set_current_catalog(self, catalog: str) -> None:
         """Sets the catalog name of the current connection."""
         self.execute(exp.Use(this=schema_(db="information_schema", catalog=catalog)))
 
+    def get_catalog_type(self, catalog: t.Optional[str]) -> str:
+        if catalog:
+            connector_name = self.fetchone(
+                f"select connector_name from system.metadata.catalogs where catalog_name='{catalog}'"
+            )
+            if len(connector_name) > 0:
+                return connector_name[0]
```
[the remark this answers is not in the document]
erindru: Catalogs in Trino need to have unique names or Trino will fail to start, so there should only ever be a single row returned.

Since `fetchone()` returns a single row as a tuple, so the length check is to see if there was a value in the first column `connector_name`, and if there is, it gets returned.

If no records are returned, we just return the default

sqlmesh/core/engine_adapter/base.py
erindru: Fair call, it didn't even occur to me that catalogs might not be supported by some databases

erindru:
Thanks for your review @eakmanrq , i'll make some tweaks based on the feedback.

> One thing that is missing is documentation. Can you update the Trino documentation to include a note about Iceberg support?

Actually, I wasn't intending to commit you guys to supporting Iceberg until I was sure it was working myself :) this PR was an initial implementation but i'm expecting other issues to crop up so I wasn't going to state it was "supported" just yet.

However, I can update the docs if you're ok with that

> Also from reviewing this PR it seems like the code currently assumes a single catalog is used when connecting to Trino. For example it always checks "current_catalog" when wanting to know what the connector is when the object it is creating could actually be in a different catalog and therefore be a different connector. If that is correct, I think that constraint is fine but should also be documented.

You're absolutely right, I just wanted to gauge feedback before spending too much time making it flexible. `_build_table_properties_exp` doesn't have a parameter containing the table/catalog being built and I was trying to avoid changing method signatures across all the adapters in the first instance. I'll revisit this

sqlmesh/core/engine_adapter/trino.py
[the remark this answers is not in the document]
eakmanrq: Ah I see now. You could then use `seq_get(connector_name, 0) or self.DEFAULT_CATALOG_TYPE` or this is fine too.

eakmanrq:
>Actually, I wasn't intending to commit you guys to supporting Iceberg until I was sure it was working myself :) this PR was an initial implementation but i'm expecting other issues to crop up so I wasn't going to state it was "supported" just yet.

We currently define supported as being able to pass all the integration tests. Since that is the case with your change then I think it would be correct to consider it supported.

sqlmesh/core/engine_adapter/trino.py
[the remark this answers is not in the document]
erindru: I like that, its more concise. I added it

erindru:
> We currently define supported as being able to pass all the integration tests. Since that is the case with your change then I think it would be correct to consider it supported.

Cool, works for me! I've added a section to the docs.

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -147,12 +148,25 @@ def _build_table_properties_exp(
...
                     )
-            properties.append(
-                exp.PartitionedByProperty(
-                    this=exp.Schema(expressions=partitioned_by),
-                )
+
+            property: exp.Property
+            property = exp.PartitionedByProperty(
+                this=exp.Schema(expressions=partitioned_by),
             )
```
georgesittas, line 155:
```suggestion
            property: exp.Property = exp.PartitionedByProperty(
                this=exp.Schema(expressions=partitioned_by),
            )
```

sqlmesh/core/engine_adapter/mixins.py
georgesittas: Looks good, thanks 👍

georgesittas:
The PR looks good to me, nice work @erindru! As I'm not very familiar with some of the details here, could you also take another looks @eakmanrq?

eakmanrq (approved):
Looks great! Thanks @erindru!

Feel free to merge if you ready.

erindru:
Thanks! I can't merge it, GitHub shows the following message:

> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

PR #2131, opened 2024-02-15
Fix: transpile default catalog name for default DuckDB test connection

treysp:
SQLMesh currently assumes the default DuckDB test connection can execute queries containing the main connection's default catalog name, which is not always true due to different identifier quotes across engines. 

This PR transpiles the default catalog name from the main connection dialect to the DuckDB dialect so DuckDB can accommodate default catalogs from all engines.

Background
- A SQLMesh project's main connection may specify a default catalog. If the catalog name is quoted, it will use the identifier quotes of the connection's own SQL dialect. 
- By default, SQLMesh executes model unit tests on a local DuckDB connection.
- The catalog name is part of a model's fully qualified name, which must be usable by DuckDB when it runs the test queries.

### DIFF
--- a/sqlmesh/core/config/root.py
+++ b/sqlmesh/core/config/root.py
@@ -6,6 +6,7 @@ import typing as t
 import zlib
 
 from pydantic import Field
+from sqlglot import exp
 from sqlglot.helper import first
 from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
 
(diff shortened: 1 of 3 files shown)


tobymao:
lgtm

sqlmesh/core/config/root.py
```diff
@@ -169,10 +170,21 @@ def _normalize_fields_after(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.An
...
+        default_catalog_dialect: t.Optional[str] = None,
     ) -> ConnectionConfig:
         return self.default_test_connection_ or DuckDBConnectionConfig(
-            catalogs=None if default_catalog is None else {default_catalog: ":memory:"}
+            catalogs=(
+                None
+                if default_catalog is None
+                else {
+                    # transpile catalog name from main connection dialect to DuckDB
+                    exp.parse_identifier(default_catalog, dialect=default_catalog_dialect).sql(
```
eakmanrq, line 183: Nice job thinking through this detail to make sure the default catalog dialect is used to parse it.

eakmanrq (approved):

PR #2133, opened 2024-02-16
Feat: ignore SQL model unit test row order unless query contains ORDER BY

treysp:
SQLMesh unit tests currently require the rows returned by the test query to be in the same order as the output rows specified in the test definition. Different engines may return query rows in different orders, meaning the same unit test will pass on some engines and fail on others.

This PR makes SQL model unit tests row-order agnostic unless the model outer SELECT contains an `ORDER BY` clause (or a CTE's outer SELECT for CTE-specific tests).

Python model tests are order-dependent because there is no universally reliable way to determine whether the model orders the rows.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -101,7 +101,7 @@ class ModelTest(unittest.TestCase):
                 else table
             )
 
-    def assert_equal(self, expected: pd.DataFrame, actual: pd.DataFrame) -> None:
+    def assert_equal(self, expected: pd.DataFrame, actual: pd.DataFrame, sort: bool) -> None:
         """Compare two DataFrames"""
 
         # Two astypes are necessary, pandas converts strings to times as NS,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -241,12 +250,13 @@ def test_ctes(self, ctes: t.Dict[str, exp.Expression]) -> None:
                     )
 
                 cte_query = ctes[cte_name].this
+                query_has_orderby = any(cte_query.find_all(exp.Order))
```
tobymao, line 253: the order by should only be on the final projection query right?

this checks for order by on ANY part of the sql qurey

georgesittas: I believe you can either check `args.get("order")` or try `find_all_in_scope(exp.Order)`

tests/core/test_test.py
```diff
@@ -234,6 +234,74 @@ def test_column_order(sushi_context: Context, full_model_without_ctes: SqlModel)
...
+  vars:
+    start: 2022-01-01
+    end: 2022-01-01
+        """
+    )
+    result = _create_test(body, "test_foo", model, sushi_context).run()
+    assert result and not result.wasSuccessful()
+
+
+def test_row_order_no_orderby(sushi_context: Context, full_model_without_ctes: SqlModel) -> None:
```
georgesittas, line 274: What's different in this test, e.g. compared to https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_test.py#L212?

georgesittas: Not sure where the `select *` is located, it seems like we just use [full_model_without_ctes](https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_test.py#L56) which explicitly lists its columns? Correct me if I'm wrong.

I was thinking if it'd make sense to consolidate the two tests basically. You could rename the older one to something like `test_order` and check both cases there?

georgesittas (approved):

PR #2134, opened 2024-02-16
Fix: motherduck only supports duckdb==0.9.2

treysp:
Motherduck only supports `duckdb==0.9.2` as of 2024-02-16

https://motherduck.com/docs/getting-started/connect-query-from-python/installation-authentication/#installing-duckdb

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,9 @@ setup(
             "langchain",
             "openai",
         ],
+        "motherduck": [
+            "duckdb==0.9.2",
+        ],
         "mssql": [
             "pymssql",
         ],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -121,6 +121,9 @@
             "langchain",
             "openai",
         ],
+        "motherduck": [
+            "duckdb==0.9.2",
```
eakmanrq, line 125: I think we could safely say `duckdb<0.10.0` to make sure we get bug fixes until they support ` `0.10.0`. It would be strange for them to release a `0.9.3` and then not support that bugfix in motherduck.

georgesittas (approved):

PR #2139, opened 2024-02-19
Feat: Support config class overriding in the loader and context

crericha:
This PR enables loading a different Config class that is derived from `Config` into the SQLMesh context, which enables storage of additional config parameters that are not directly used by SQLMesh.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -42,6 +42,8 @@ class ConnectionConfig(abc.ABC, BaseConfig):
     concurrent_tasks: int
     register_comments: bool
 
+    type_: str = Field(alias="type", default="none")
+
     @property
     @abc.abstractmethod
     def _connection_kwargs_keys(self) -> t.Set[str]:
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -42,6 +42,7 @@ class ConnectionConfig(abc.ABC, BaseConfig):
     concurrent_tasks: int
     register_comments: bool
 
+    # Pydantic V2 does not include type field in dict unless also in base class
     type_: str = Field(alias="type", default="none")
 
     @property
--- a/sqlmesh/core/config/loader.py
+++ b/sqlmesh/core/config/loader.py
@@ -13,12 +13,15 @@ from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.metaprogramming import import_python_file
 from sqlmesh.utils.yaml import load as yaml_load
 
+C = t.TypeVar("C", bound=Config)
+
 
 def load_configs(
-    config: t.Optional[str | Config],
+    config: t.Optional[str | t.Union[Config, C]],
+    config_type: t.Type[C],
     paths: t.Union[str | Path, t.Iterable[str | Path]],
     sqlmesh_path: t.Optional[Path] = None,
-) -> t.Dict[Path, Config]:
+) -> t.Dict[Path, C]:
     sqlmesh_path = sqlmesh_path or c.SQLMESH_PATH
     config = config or "config"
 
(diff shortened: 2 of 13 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -42,6 +42,8 @@ class ConnectionConfig(abc.ABC, BaseConfig):
     concurrent_tasks: int
     register_comments: bool
 
+    type_: str = Field(alias="type", default="none")
```
crericha, line 45: Pydantic V2 `dict()` does not include `type` in for child instances unless `type` is also defined in the base class.

sqlmesh/core/config/loader.py
```diff
@@ -13,12 +13,15 @@
 from sqlmesh.utils.metaprogramming import import_python_file
 from sqlmesh.utils.yaml import load as yaml_load
 
+C = t.TypeVar("C", bound=Config)
+
 
 def load_configs(
-    config: t.Optional[str | Config],
+    config: t.Optional[str | t.Union[Config, C]],
+    config_type: t.Type[C],
```
crericha, line 21: I couldn't decide if `config_type` should be a required field or not. I went with required for now, so that if more loader calls are added, the caller will be aware that they should pass in `config_type` for overriding to work on that load path. I could be convinced otherwise.

sqlmesh/core/config/connection.py
izeigerman: Adding this as a code comment might be helpful

sqlmesh/core/config/loader.py
```diff
@@ -13,12 +13,15 @@
 from sqlmesh.utils.metaprogramming import import_python_file
 from sqlmesh.utils.yaml import load as yaml_load
 
+C = t.TypeVar("C", bound=Config)
+
 
 def load_configs(
-    config: t.Optional[str | Config],
+    config: t.Optional[str | t.Union[Config, C]],
```
izeigerman, line 20: Could this be `t.Optional[str, C]`?

sqlmesh/core/config/loader.py
izeigerman: I feel like having the default to be `Config` makes sense and simplifies the API

sqlmesh/core/config/loader.py
```diff
@@ -27,6 +30,8 @@ def load_configs(
     ]
 
     if isinstance(config, Config):
+        if type(config) != config_type:
```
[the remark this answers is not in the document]
crericha: The check is really seeing if we have a string or config obj.

izeigerman (approved):

sqlmesh/core/context.py
```diff
@@ -119,6 +119,7 @@
     ModelOrSnapshot = t.Union[str, Model, Snapshot]
     NodeOrSnapshot = t.Union[str, Model, StandaloneAudit, Snapshot]
 
+C = t.TypeVar("C", bound=Config)
```
izeigerman, line 122: Should we just import one from the `config/loader.py`?

sqlmesh/core/config/loader.py
```diff
@@ -158,3 +175,10 @@ def load_config_from_env() -> t.Dict[str, t.Any]:
             target_dict[segments[-1]] = value
 
     return config_dict
+
+
+def convert_config_type(
+    config_obj: Config,
+    config_type: t.Type[C],
+) -> C:
+    return config_type.parse_obj(config_obj.dict())
```
eakmanrq, line 184: I've been recently working on something similar and I think you could reduce of the cost of this with this:
```
config_type(**{k: v for k, v in model})
```
That avoids having serde overhead of going to and from a dict by just iterating over the root level fields of the model.

eakmanrq (approved):

sqlmesh/core/config/loader.py
eakmanrq: Oh I see good point

PR #2141, opened 2024-02-19
Fix: Track unknownModels to diff from CTEs

mykalmax:


web/client/src/library/components/graph/ModelLineage.tsx
```diff
@@ -77,9 +79,23 @@ export default function ModelLineage({
...
+        lineageModels.forEach(modelName => {
+          modelName = encodeURI(modelName)
+
+          if (
+            isFalse(models.has(modelName)) &&
+            isFalse(unknownModels.has(modelName))
+          ) {
+            unknownModels.add(modelName)
+
+            setUnknownModels(new Set(unknownModels))
```
vchan, line 94: Can we do this outside of the loop?

vchan (approved):

PR #2142, opened 2024-02-19
Fix: Make sure that missing intervals don't exceed the target end date when applying plan

izeigerman:
Missing intervals might exceed the given end date in cases when the `allow_partials` or the `lookback` attribute is configured for a model. Even though, this behavior makes sense when executing the `sqlmesh run` command, it's confusing and undesirable when applying a plan (especially to a dev environment).

This update prevents unexpected backfills by bounding missing intervals to the target end date when applying a plan.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -56,6 +56,8 @@ class PlanBuilder:
         default_start: The default plan start to use if not specified.
         default_end: The default plan end to use if not specified.
         enable_preview: Whether to enable preview for forward-only models in development environments.
+        end_bounded: If set to true, the missing intervals will be bounded by the target end date, disregarding lookback,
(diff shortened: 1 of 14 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -56,6 +56,8 @@ class PlanBuilder:
         default_start: The default plan start to use if not specified.
         default_end: The default plan end to use if not specified.
         enable_preview: Whether to enable preview for forward-only models in development environments.
+        end_bounded: If set to true, the missing intervals will be bounded by the target end date, disregarding lookback,
+            allow_partials, and other attributes that could cause the intervals to exceed the target end date.
```
eakmanrq, line 60: Extra tab here

eakmanrq (approved):

tobymao (approved):

PR #2145, opened 2024-02-20
feat: propogate descriptions in the ui if they are simply pass throug…

tobymao:
…h and don't have multiple sources

### DIFF
--- a/sqlmesh/core/lineage.py
+++ b/sqlmesh/core/lineage.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import typing as t
+from collections import defaultdict
+
+from sqlglot import exp
+from sqlglot.helper import first
+from sqlglot.lineage import lineage
+
+from sqlmesh.core.dialect import normalize_model_name
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core.context import Context
+    from sqlmesh.core.model import Model
+
+
+def _render_query(model: Model) -> exp.Subqueryable:
+    """Render a model's query, adding in managed columns"""
+    query = model.render_query_or_raise()
+    if model.managed_columns:
+        query.select(
+            *[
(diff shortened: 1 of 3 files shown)


sqlmesh/core/lineage.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
```
izeigerman, line 1: Is this only used for web? Should we move this into the `web` package?

sqlmesh/core/lineage.py
```diff
@@ -0,0 +1,64 @@
...
+                continue
+
+            table = d.expression.find(exp.Table)
+            if table:
+                name = normalize_model_name(table, default_catalog=context.default_catalog, dialect=model.dialect)
+                parents[name].add(exp.to_column(d.name).name)
+    return dict(parents)
+
+
+def column_description(context: Context, model_name: str, column: str) -> t.Optional[str]:
```
izeigerman, line 48: We'll need some tests for this

sqlmesh/core/lineage.py
```diff
@@ -0,0 +1,64 @@
...
+    dependencies = column_dependencies(context, model_name, column)
+
+    if len(dependencies) != 1:
+        return None
+
+    for parent, columns in dependencies.items():
+        if len(columns) != 1:
+            return None
+
+        for column in columns:
```
izeigerman, line 63: [Nit] Seems like an unnecessary level of nestedness

sqlmesh/cli/main.py
```diff
@@ -591,6 +591,7 @@ def ui(
         port=port,
         log_level="info",
         timeout_keep_alive=300,
+        reload=True,
```
vchan, line 594: I think we should remove this.

tobymao: yea i will

tobymao:
@vchan merging this after tests pass

PR #2148, opened 2024-02-20
Fix: restrict scope of sqlmesh statement parsers

georgesittas:
Related Slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1708432374934119

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -107,15 +107,23 @@ def _parse_statement(self: Parser) -> t.Optional[exp.Expression]:
 
     parser = PARSERS.get(self._curr.text.upper())
 
-    if parser and self._prev is None:
+    if parser:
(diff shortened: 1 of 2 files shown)


tobymao (approved):

izeigerman:
@georgesittas I think this is still failing on your branch:
```
In [1]: from sqlmesh.core import dialect as d

In [2]: d.parse_one("metric", dialect="bigquery")
----------------------------------------------------
```

georgesittas:
Yeah the reasoning was that we wouldn't ever encounter a standalone `metric` (or other statement-starting keyword) in a model file, but it's still an issue apparently because we rely on `parse_one` internally to e.g. parse meta fields. I'm currently looking to refactor.

tobymao (approved):

izeigerman (approved):

PR #2149, opened 2024-02-20
Fix: convert pandas timestamp to py datetime for unit test generation

treysp:
**Problem/background**
- `generate_test()` creates a SQLMesh unit test YAML file based on a user-specified model name and input query
- When we fetchdf the query results to Pandas, all datetime columns are converted to `pd.Timestamp` type (at least when queries run on DuckDB)
- We use `ruamel.yaml` to serialize the data values returned by the query to YAML, and it does not support `pd.Timestamp` objects

**This PR**
- Converts pd.Timestamp to datetime objects before writing query result data values to YAML
- We want to maintain fidelity to the source column type. If it is `DATE`, we extract the date component from the `datetime` object before writing so a time component of `00:00:00` isn't included in the YAML.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -14,6 +14,7 @@ from sqlmesh.core.dialect import normalize_model_name, schema_
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.model import Model, PythonModel, SqlModel
 from sqlmesh.utils import UniqueKeyDict, yaml
+from sqlmesh.utils.date import pandas_timestamp_to_pydatetime
 from sqlmesh.utils.errors import ConfigError, SQLMeshError
 
 Row = t.Dict[str, t.Any]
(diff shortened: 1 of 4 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -379,8 +380,14 @@ def generate_test(
             f"Fixture '{fixture_path}' already exists, make sure to set --overwrite if it can be safely overwritten."
         )
 
+    # ruamel.yaml does not support pandas Timestamps, so we must convert them to python
+    # datetime or datetime.date objects based on column type
     inputs = {
-        dep: engine_adapter.fetchdf(query).to_dict(orient="records")
+        models[dep]
+        .name: pandas_timestamp_to_pydatetime(
```
georgesittas, line 387: Have you double-checked the use of `name` vs `fqn` in this section? E.g. should this be `models[dep].fqn`?

sqlmesh/utils/date.py
```diff
@@ -337,3 +340,19 @@ def to_time_column(
     if time_column_type.this in exp.DataType.NUMERIC_TYPES:
         return exp.Literal.number(time_column)
     return exp.convert(time_column)
+
+
+def pandas_timestamp_to_pydatetime(
+    df: pd.DataFrame, columns_to_types: t.Dict[str, exp.DataType]
+) -> pd.DataFrame:
+    for column, column_type in columns_to_types.items():
+        # Query projections may not include all columns, so we check if the column exists in the dataframe
```
georgesittas, line 349: what's an example of this?

georgesittas (commented):
Is this currently tested in the integration suite? Should we also add a unit test for this?

georgesittas (approved):
Discussed in Slack about changing the for loop order for perf. reasons, but other than that LGTM 👍

georgesittas (approved):
Looks good again. One final question is whether CTE outputs also need to be converted.

PR #2162, opened 2024-02-21
Feat: update init example project to run on all engines

treysp:
And add integration tests


eakmanrq (approved):
Such a relief to know that we can now ensure this works across all engines. Nice job!

PR #2168, opened 2024-02-22
feat: dbt multi repo

tobymao:


sqlmesh/dbt/loader.py
```diff
@@ -124,59 +126,65 @@ def _load_audits(
...
 
         return audits
 
-    def _load_project(self) -> Project:
-        if self._project:
-            return self._project
-
-        target_name = self._context.gateway or self._context.config.default_gateway
+    def _load_projects(self) -> t.List[Project]:
+        if not self._projects:
```
izeigerman, line 142: I know it has been there before but I feel like we should just use `@cached_property` and get rid of an extra level nestedness

izeigerman (approved):

Issue #1273, opened 2023-08-07
Introduce `sqlmesh clean` command


izeigerman:
A command should cleanup cache files and other build artifacts if available.

izeigerman:
Implemented

PR #2170, opened 2024-02-22
Feat: Introduce `sqlmesh clean` command

jacksenechal:
- [x] Add a `clear_caches` method to the `Context`
- [x] Add a `clean` command to the CLI
- [x] Document changes in [CLI reference](https://github.com/TobikoData/sqlmesh/blob/main/docs/reference/cli.md)
- [x] Add a `clean` command to the Magic commands
- [x] Document changes in [Notebooks reference](https://github.com/TobikoData/sqlmesh/blob/main/docs/reference/notebook.md)
- [ ] ~~Add build artifact cleaning~~


jacksenechal:
Hey @izeigerman, I've been working with @tobymao on this PR to address https://github.com/TobikoData/sqlmesh/issues/1273. The basic `sqlmesh clean` command works at this point, and I wanted to get your eyes on it and ask a couple of questions.

1. With regard to cleaning build artifacts, I wanted to clarify what is needed. I see that the `Makefile` has the `clean-build` command, which would deal more with development artifacts. It seems like the `sqlmesh clean` command shouldn't be concerned with those. I also see `airflow-clean`, which sounds like the kind of artifacts that would be generated in the course of end-user use of the tools. Can you clarify which build artifacts need to be cleaned with the `sqlmesh clean` command, and point me to others that I've missed?
2. The current command has no guards before deleting the cache directory/directories. Should this command ask for confirmation before proceeding?

sqlmesh/magics.py
```diff
@@ -791,6 +791,14 @@ def rollback(self, context: Context, line: str) -> None:
         """Rollback SQLMesh to the previous migration."""
         context.rollback()
 
+    @magic_arguments()
+    @line_magic
+    @pass_sqlmesh_context
+    def clean(self, context: Context, line: str) -> None:
+        """Clears the SQLMesh cache and any build artifacts."""
+        context.clear_caches()
+        context.console.log_success("SQLMesh cache and build artifacts cleared")
```
jacksenechal, line 800: Do we like this feedback here? Some methods have it, some don't.

sqlmesh/cli/main.py
```diff
@@ -720,3 +720,11 @@ def prompt(
...
     if evaluate:
         context.console.log_success(context.fetchdf(query))
+
+
+@cli.command("clean")
+@click.pass_obj
+@error_handler
+def clean(obj: Context) -> None:
+    """Clears the SQLMesh cache and any build artifacts."""
+    obj.clear_caches()
```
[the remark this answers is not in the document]
izeigerman: I think exit code 0 should be sufficient.

sqlmesh/core/context.py
```diff
@@ -1571,6 +1572,13 @@ def _has_environment_changed() -> bool:
 
         return success
 
+    def clear_caches(self) -> None:
+        for path in self.configs.keys():
```
izeigerman, line 1576: [Nit] `.keys()` is not necessary:
```
for path in self.configs:
```

sqlmesh/core/context.py
```diff
@@ -1571,6 +1572,13 @@ def _has_environment_changed() -> bool:
...
+            self._clear_cache(path)
+
+    def _clear_cache(self, path: Path) -> None:
+        rmtree(path / c.CACHE)
```
izeigerman, line 1580: Do we even need this method? Should we just call `rmtree` directly above?

sqlmesh/magics.py
izeigerman: Yeah, here it probably makes more sense, since there's no exit code to rely on like in case of the CLI.

sqlmesh/core/context.py
jacksenechal: Removing it is cleaner :)

tests/cli/test_cli.py
```diff
@@ -517,3 +518,20 @@ def test_run_cron_elapsed(runner, tmp_path):
...
+    # Confirm cache exists
+    cache_path = Path(tmp_path) / ".cache"
+    assert cache_path.exists()
+    assert len(list(cache_path.iterdir())) > 0
+
+    # Invoke the clean command
+    result = runner.invoke(cli, ["--paths", tmp_path, "clean"])
+
+    # Confirm cache was cleared
+    assert result.exit_code == 0
```
jacksenechal, line 536: This is the operational bit here. We could remove the Path based assertions before and after and eliminate the extra import if preferred. I added it here to verify the test setup.

izeigerman:
@jacksenechal 

> Can you clarify which build artifacts need to be cleaned with the sqlmesh clean command, and point me to others that I've missed?

I think cleaning the cache is sufficient for now.

> The current command has no guards before deleting the cache directory/directories. Should this command ask for confirmation before proceeding?

No, I don't think so. Cleaning cache is a pretty low-risk operation. No additional hindrances are necessary.

tests/cli/test_cli.py
izeigerman: I think this is great

izeigerman (approved):
Thanks a lot for your contribution!

PR #2175, opened 2024-02-23
Feat: support dbt-formatted projects on MSSQL/sqlserver engine

### LINKED ISSUE #2154: Enable SQL Server support for the dbt adapter
Even though SQLMesh supports SQL Server natively, this feature has never been integrated into the dbt adapter.

treysp:
Only user/password and Windows authentication methods are supported

Fixes #2154


sqlmesh/dbt/target.py
```diff
@@ -563,11 +566,108 @@ def to_sqlmesh(self) -> ConnectionConfig:
...
+    driver: t.Optional[str] = None
+    encrypt: t.Optional[bool] = None
+    trust_cert: t.Optional[bool] = None
+    retries: t.Optional[int] = None
+
+    # Unused authentication parameters (not supported by pymssql)
+    windows_login: t.Optional[bool] = None  # pymssql doesn't require this flag for Windows Auth
+    tenant_id: t.Optional[str] = None  # Azure Active Directory auth
+    client_id: t.Optional[str] = None  # Azure Active Directory auth
+    client_secret: t.Optional[str] = None  # Azure Active Directory auth
```
eakmanrq, line 609: Is it possible to define these and use the `sql` auth method?

treysp: You could put them in the profile, but they would be ignored for `sql` auth method.

Or are you asking if dbt-sqlserver errors if they're included when auth is `sql`?

sqlmesh/dbt/context.py
```diff
@@ -281,3 +281,13 @@ def create_relation(
...
+
+TARGET_TYPE_TO_DIALECT: t.Dict[str, str] = {
+    "databricks": "databricks",
+    "duckdb": "duckdb",
+    "postgres": "postgres",
+    "redshift": "redshift",
+    "snowflake": "snowflake",
+    "bigquery": "bigquery",
+    "sqlserver": "tsql",
+}
```
eakmanrq, line 293: What do you think about calling this `TARGET_TYPE_TO_DIALECT_OVERRIDE` and then only having `sqlserver` -> `tsql` in it. Calls do a get and then default to the `target.type` if no match. Avoids having to make sure we have entries for target types that directly map to a dialect (which is the common case).

sqlmesh/dbt/target.py
eakmanrq: No that answers it. Making sure it it was not possible for someone to define these for `sql` and then have them be silently ignored by the adapter. Or at least if we were intended to silently ignore that it was an intention decision.

eakmanrq (approved):
🚀

PR #2177, opened 2024-02-23
feat: log BigQuery job details link

plaflamme:
This will log a clickable link to the BigQuery job within BigQuery Studio. This should help debugging queries within BigQuery directly.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -615,6 +615,14 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
             job_config=job_config,
             timeout=self._extra_config.get("job_creation_timeout_seconds"),
         )
+
+        logger.debug(
+            "BigQuery job created: https://console.cloud.google.com/bigquery?project=%s&j=bq:%s:%s",
+            self._query_job.project,
+            self._query_job.location,
+            self._query_job.job_id,
+        )
+
         results = self._db_call(
             self._query_job.result,
             timeout=self._extra_config.get("job_execution_timeout_seconds"),  # type: ignore


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -615,6 +615,11 @@ def _execute(
             job_config=job_config,
             timeout=self._extra_config.get("job_creation_timeout_seconds"),
         )
+
+        if logger.isEnabledFor(logging.DEBUG):
```
tobymao, line 619: logger.isEOnabled is not necessary, you can just log to the logger
also best practice to domething like


logger.debug("Bigquery job created: %s", job_link)

plaflamme: Sounds good. I wanted to be defensive here and not actually generate the job link at all if we're not going to log it. i can redefine the whole log line to inline the `%s` there instead for the same result.

izeigerman: If we want to get rid of `logger.isEnabledFor` I'd this instead:
```
logger.debug(f"BigQuery job created: https://console.cloud.google.com/bigquery?project=%s&j=bq:%s:%s", ...)
```
To avoid creating a variable not used outside the logging setting.

plaflamme: fixed in https://github.com/TobikoData/sqlmesh/pull/2177/commits/cbdb133c47904d83dcf1cf9206cb031ea28d3b87

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -615,6 +615,14 @@ def _execute(
             job_config=job_config,
             timeout=self._extra_config.get("job_creation_timeout_seconds"),
         )
+
+        logger.debug(
```
izeigerman, line 619: beautiful 👍

izeigerman:
@plaflamme could you please rebase against latest main? This should help address the failing build.

plaflamme:
@izeigerman done, I also squashed the commits while I was at it

PR #2178, opened 2024-02-24
Chore: add unixodbc dependency to circleci docker images

treysp:


tobymao (approved):
why are we adding this? are we switch the mssql lib?

treysp:
No, dbt-sqlserver imports pyodbc

PR #2181, opened 2024-02-25
add bigquery execution project config support

Kayrnt:
Support for execution_project (which is going to be the GCP project used for execution as in the resources used ... opposed to the storage). It's mandatory for organizations that split "storage project" vs "execution project" to enable fine grained resources management.

For reference, the similar change I made for dbt 3 years ago: https://github.com/dbt-labs/dbt-bigquery/commit/42cd89d08e38a871743ffd49716708475bb48d1f

I added the fallback to `project` for backward compatibility.

I couldn't find tests for bigquery related config... so I made one but it's not as useful as I'd like 🙄

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -614,6 +614,7 @@ class BigQueryConnectionConfig(ConnectionConfig):
     method: BigQueryConnectionMethod = BigQueryConnectionMethod.OAUTH
 
     project: t.Optional[str] = None
+    execution_project: t.Optional[str] = None
     location: t.Optional[str] = None
     # Keyfile Auth
     keyfile: t.Optional[str] = None
@@ -675,7 +676,7 @@ class BigQueryConnectionConfig(ConnectionConfig):
         else:
             raise ConfigError("Invalid BigQuery Connection Method")
         client = google.cloud.bigquery.Client(
-            project=self.project,
+            project=self.execution_project or self.project,
             credentials=creds,
             location=self.location,
             client_info=client_info.ClientInfo(user_agent="sqlmesh"),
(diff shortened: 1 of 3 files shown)


eakmanrq:
@Kayrnt Were you able to confirm that this change works for you? In that you were able to run a project, like the sushi project, and have it execute in a different project but reference objects in another project? 

Also, to make sure I understand, with this change you can refere to datasets in any project right and it just ensures that the execution project is where the execution takes place (and therefore is ultimately billed)?

Kayrnt:
@eakmanrq, thanks for the quick look!

>Were you able to confirm that this change works for you?
Not yet, it's just a change I already did in dbt and it works. I'll confirm on my end.

> In that you were able to run a project, like the sushi project, and have it execute in a different project but reference objects in another project?

If `get_catalog` is well implemented and properly used to build the reference to the table, then yes, it should work like that.

> Also, to make sure I understand, with this change you can refere to datasets in any project right and it just ensures that the execution project is where the execution takes place (and therefore is ultimately billed)?

Yes, I expect the project from the connection config to be used as the "storage project" (where the data is being stored) and the execution project, where the execution happens (somehow billing but there's a more complex mapping to resources using commitments & reservations to link actual BQ slots to GCP projects).

Kayrnt:
Ok tested it locally with that kind of config
```
gateways:
    local:
        connection:
            type: bigquery
            project: my_storage_rpoject
            execution_project: my_execution_project
            location: US
            method: oauth
            priority: interactive
        state_connection:
            type: duckdb
            database: state.db

default_gateway: local

model_defaults:
    dialect: bigquery
```

It works as intended

eakmanrq (approved):
>It works as intended

Great thanks @Kayrnt!

PR #2187, opened 2024-02-28
fix: propogate false and limit to every select as some queries do not…

tobymao:
… have sufficient predicate pushdown

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -423,11 +423,13 @@ class _Model(ModelMeta, frozen=True):
             The mocked out ctas query.
         """
         # the query is expanded so it's been copied, it's safe to mutate.
-        query = (
-            self.render_query_or_raise(**render_kwarg)
-            .limit(0, copy=False)
-            .where(exp.false(), copy=False, append=False)
-        )
+        query = self.render_query_or_raise(**render_kwarg)
+
+        for select in query.find_all(exp.Select):
+            if select.args.get("from"):
+                select.where(exp.false(), copy=False)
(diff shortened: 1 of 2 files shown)


izeigerman (approved):
LGTM once green. Do we have a test for this?

tobymao:
> LGTM once green. Do we have a test for this?

yea we have a couple of test cases around this

sqlmesh/core/model/definition.py
```diff
@@ -423,11 +423,13 @@ def ctas_query(self, **render_kwarg: t.Any) -> exp.Query:
...
-            .limit(0, copy=False)
-            .where(exp.false(), copy=False, append=False)
-        )
+        query = self.render_query_or_raise(**render_kwarg)
+
+        for select in query.find_all(exp.Select):
+            if select.args.get("from"):
+                select.where(exp.false(), copy=False)
+                if not isinstance(select.parent, exp.Union):
+                    select.limit(0, copy=False)
```
georgesittas, line 432: What happens if we have a model with a `UNION` as its query? Don't we wanna limit it?

tobymao: ethos is not legal 

select * limit 0 union select *

georgesittas: agreed, but you can do <query1> union <query2> limit X

georgesittas: I see. Maybe we were limiting the query because it's less costly that way, so removing it for unions would incur a full scan? I guess I'm not sure if getting rid of the limit here is a good idea vs fully addressing this now.

Just trying to understand the impact of this change.

PR #2189, opened 2024-02-28
Feat: Support the Snowflake Warehouse model config for dbt projects

izeigerman:
Support for https://docs.getdbt.com/reference/resource-configs/snowflake-configs#configuring-virtual-warehouses


crericha (approved):

crericha:
Is this a per model config property? Not something that can be inferred from the profile target?

izeigerman:
@crericha yes, it is. Updated the description with a reference to relevant docs.

PR #2190, opened 2024-02-28
Chore: tests that import dbt adapters run on CI/CD only

treysp:


.circleci/continue_config.yml
```diff
@@ -165,9 +165,6 @@ jobs:
       - run:
           name: Install envsubst
           command: sudo apt-get update && sudo apt-get install gettext-base
-      - run:
-          name: Install ODBC
-          command: sudo apt-get install unixodbc-dev
```
georgesittas, line 170: We don't need these anymore?

tobymao: i assume we still need this for ci/cd tests right?

Makefile
```diff
@@ -84,26 +87,29 @@ engine-down:
...
-	pytest -n auto -m "fast and not web and not github and not dbt and not airflow and not jupyter"
+	pytest -n auto -m "fast and not web and not github and not dbt and not airflow and not jupyter and not cicdonly"
 
 core-slow-test:
-	pytest -n auto -m "(fast or slow) and not web and not github and not dbt and not airflow and not jupyter"
+	pytest -n auto -m "(fast or slow) and not web and not github and not dbt and not airflow and not jupyter and not cicdonly"
 
 airflow-fast-test:
-	pytest -n auto -m "fast and airflow"
+	pytest -n auto -m "fast and airflow and not cicdonly"
```
tobymao, line 105: do you have to add these everywhere?isn't it a subset of fast?

treysp: It's a subset of dbt so can be removed from most of these. I added it everywhere to make it easier to add more in the future, but i guess these should be rare.

.circleci/continue_config.yml
treysp: The airflow tests don't include the ci/cd only tests, so the airflow-specific docker doesn't need the dependency

Makefile
tobymao: but fast already includes cicdonly, so if you just added that to airflow, would that work?

treysp: yes, i think the only place we need it is `fast-test`, `slow-test`, and `dbt-test`

since cicdonly isn't part of airflow, "fast and airflow" shouldn't need the additional cicdonly exclusion

.circleci/continue_config.yml
```diff
@@ -71,13 +71,13 @@ jobs:
           command: sudo apt-get install unixodbc-dev
       - run:
           name: Install dependencies
-          command: make install-dev
+          command: make install-cicd-test
       - run:
           name: Run linters and code style checks
           command: make py-style
       - run:
           name: Run slow tests
```
tobymao, line 79: change the desciption maybe

tobymao (approved):

PR #2197, opened 2024-03-01
Fix:  Manual categorization

mykalmax:
I removed list of parents from indirectly modified to simplify

Screenshot 2024-02-29 at 4 23 07 PM

Added confirmation when user has uncategorized and trying to Apply

Screenshot 2024-02-29 at 3 39 19 PM
Screenshot 2024-02-29 at 3 40 28 PM

### DIFF
--- a/web/client/src/library/components/plan/PlanChangePreview.tsx
+++ b/web/client/src/library/components/plan/PlanChangePreview.tsx
@@ -84,7 +84,7 @@ function PlanChangePreviewDefault({
   changes: ModelSQLMeshChangeDisplay[]
 }): JSX.Element {
   return (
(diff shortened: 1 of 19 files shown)


web/client/src/library/components/plan/PlanChangePreview.tsx
```diff
@@ -78,12 +84,12 @@ function PlanChangePreviewDefault({
   changes: ModelSQLMeshChangeDisplay[]
 }): JSX.Element {
   return (
-    <ul>
+    <ul className="">
```
vchan, line 87: Is this needed?

vchan (approved):

PR #2199, opened 2024-03-01
Fix: Reduce the memory footprint during the migration

izeigerman:
Currently the memory footprint of the sqlmesh process during migration is way too high for large projects.

For example, it takes about 10-11GB of RAM to migrate ~10k snapshot records. 

During the investigation I've noticed that the raw snapshot payload doesn't consume that much memory. It's when we deserialize all snapshots at once the footprint grows drastically. 

This update ensures that we only keep a limited number of snapshots in deserialized form in memory during the migration.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -148,6 +148,11 @@ class SnapshotIntervals(PydanticModel, frozen=True):
         return SnapshotId(name=self.name, identifier=self.identifier)
 
 
+class SnapshotIndirectVersion(PydanticModel, frozen=True):
+    version: str
+    change_category: t.Optional[SnapshotChangeCategory] = None
+
+
 class SnapshotDataVersion(PydanticModel, frozen=True):
     fingerprint: SnapshotFingerprint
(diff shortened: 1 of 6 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -173,6 +178,13 @@ def is_new_version(self) -> bool:
         """Returns whether or not this version is new and requires a backfill."""
         return self.fingerprint.to_version() == self.version
 
+    @property
+    def indirect_version(self) -> SnapshotIndirectVersion:
```
eakmanrq, line 182: How does adding this property/class relate to the other changes? Or is something you noticed while working on this?

izeigerman: By adding this we no longer need to map indirect versions during the migration. The purpose of these is twofold: 
1. It simplifies the migration logic drastically.
2. We no longer need to keep all snapshots in memory in the deserialized form. The mapping logic used to rely on this assumption heavily.

eakmanrq (approved):

PR #2200, opened 2024-03-01
Fix: map dbt model default dialect sqlserver -> tsql

treysp:


tobymao (approved):

sqlmesh/dbt/adapter.py
```diff
@@ -290,12 +290,14 @@ def execute(
             self.engine_adapter.fetchdf if fetch else self.engine_adapter.execute  # type: ignore
         )
 
-        expression = parse_one(sql, read=self.engine_adapter.dialect)
+        # parse SQL in its dialect
+        expression = parse_one(sql, read=self.dialect)
```
izeigerman, line 294: As per convo in slack: I believe that `self.dialect` and `self.engine_adapter.dialect` are actually the same dialect since `self.dialect` is derived from the type of the dbt target.

crericha: Yeah, these are the same value. The constructor sets dialect to engine_adapter.dialect. I guess it isn't bad to use dialect here, but we should be consistent throughout then. I think I saw other places where we do engine_adapter.dialect.

sqlmesh/dbt/context.py
```diff
@@ -11,6 +11,7 @@
 from sqlmesh.dbt.manifest import ManifestHelper
 from sqlmesh.dbt.target import TargetConfig
 from sqlmesh.utils import AttributeDict
+from sqlmesh.utils.dbt import TARGET_TYPE_TO_DIALECT_OVERRIDE
```
crericha, line 14: I'm not a fan of the override being done like this. IMO it should be a property in the target classes and we use that instead of type.

sqlmesh/dbt/adapter.py
```diff
@@ -290,9 +288,9 @@ def execute(
             self.engine_adapter.fetchdf if fetch else self.engine_adapter.execute  # type: ignore
         )
 
-        expression = parse_one(sql, read=self.engine_adapter.dialect)
+        expression = parse_one(sql, read=self.dialect)
         with _normalize_and_quote(
-            expression, self.engine_adapter.dialect, self.engine_adapter.default_catalog
+            expression, self.dialect, self.engine_adapter.default_catalog  # type: ignore
```
crericha, line 293: Why is the type ignore needed?

crericha (approved):

sqlmesh/dbt/adapter.py
treysp: Dialect is optional in the BaseAdapter class but required by `_normalize_and_quote`. 

We know it will exist here because we infer it from the required engine_adapter passed to the RuntimeAdapter constructor.

For context, we call the RuntimeAdapter and ParsetimeAdaper constructors in [`create_builtin_globals()` in builtin.py](https://github.com/TobikoData/sqlmesh/blob/1acb6efdb22c9e137cd14302332abd212e218f06/sqlmesh/dbt/builtin.py#L338).

treysp: If you're good with keeping `self.engine_adapter.dialect` here instead of converting to `self.dialect` there's no type issue

crericha: I slightly prefer the cast, so we are using dialect everywhere, but I'm okay with getting it from engine adapter too.

PR #2208, opened 2024-03-03
fix: snowflake column level lineage issue due to normalization

tobymao:


sqlmesh/dbt/context.py
```diff
@@ -62,7 +62,7 @@ class DbtContext:
     def dialect(self) -> str:
         if not self.target:
             raise SQLMeshError("Target must be configured before calling the dialect property.")
-        return self.target.dialect
+        return self.sqlmesh_config.model_defaults.dialect or self.target.dialect
```
tobymao, line 65: i changed this because i was trying to run a client's snowflake dbt in duckdb but was unable to do so.

sqlmesh/dbt/loader.py
```diff
@@ -41,8 +41,10 @@ def sqlmesh_config(
     project_root = project_root or Path()
     context = DbtContext(project_root=project_root)
     profile = Profile.load(context, target_name=dbt_target_name)
-    model_defaults = kwargs.get("model_defaults", ModelDefaultsConfig())
-    model_defaults.dialect = profile.target.dialect
+    model_defaults = kwargs.pop("model_defaults", ModelDefaultsConfig())
```
tobymao, line 44: same with this change cc @treysp @crericha

sqlmesh/dbt/context.py
tobymao: @treysp @crericha

crericha: Transpiling is currently not supported via the dbt adapter. If we want to support, there is a larger change set then these two changes required. Basically, there are some places in the dbt adapter where we need to know the target DWH dialect and other places that need to know the model's dialect. Right now, they are treated as the same thing.

tobymao: well for basic changes like this, it works fine, just trying to see column level lineage

crericha: There are other basic changes that would hit an issue, so its more so happenstance that you didn't hit one of the other places that need to be changed. I'm on call next week and can add transpiling to the dbt adapter if this is a priority.

tobymao: no it’s not but this was enough to unblock me. is it incorrect?

crericha: Yes, I believe it is incorrect. The context dialect is sometimes used for the target DWH.

crericha: I think we'd want context to have a default_model_dialect method and then find the places in the adapter where that should be used instead of the target dialect.

crericha: As a bonus, we could add dialect as a config property and allow individual models to override the default.

tobymao: i'll remove these changes for now

sqlmesh/dbt/loader.py
tobymao: ok this is the only change now

PR #2211, opened 2024-03-04
Chore: Update dbt-core dependency to 1.7.9 and older

crericha:


tobymao (approved):
why pin it at all?

crericha:
> why pin it at all?

If you prefer, we can remove the pin and reactively pin when there is a breaking change in dbt-core. Let me know.

tobymao:
yea let's do that pin dbt < 2

PR #2212, opened 2024-03-04
Feat: allow setting register_comments for dbt connection

treysp:


sqlmesh/dbt/loader.py
```diff
@@ -48,9 +49,13 @@ def sqlmesh_config(
     if variables is not None:
         loader_kwargs["variables"] = variables
 
+    connection_kwargs = kwargs.pop("connection_kwargs", {})
+    if register_comments is not None:
+        connection_kwargs["register_comments"] = register_comments
```
georgesittas, line 54: Where do `kwargs` get the `connection_kwargs` from? I'm trying to wrap my head around the data flow.

sqlmesh/dbt/loader.py
```diff
@@ -48,9 +49,13 @@ def sqlmesh_config(
     if variables is not None:
         loader_kwargs["variables"] = variables
 
+    connection_kwargs = kwargs.pop("connection_kwargs", {})
```
izeigerman, line 52: Why do we need this? There are no `connection_kwargs` in the `Config`

treysp: We don't. Should I do something like this, or is the dict unpacking approach more generally wrong for conditionally passing register_comments if it is not none?

```
    to_sqlmesh_args = {}
    if register_comments is not None:
        to_sqlmesh_args["register_comments"] = register_comments
```

georgesittas (approved):

eakmanrq (approved):

izeigerman (approved):

PR #2214, opened 2024-03-04
Fix: Serialization of the ConnectionConfig objects

izeigerman:
The serialization of abstract base classes has been intentionally broken in Pydantic V2. See https://docs.pydantic.dev/latest/concepts/serialization/#serializing-with-duck-typing for more details.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -24,6 +24,7 @@ from sqlmesh.core.config.common import (
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import (
+    PYDANTIC_MAJOR_VERSION,
     field_validator,
     model_validator,
     model_validator_v1_args,
@@ -41,9 +42,6 @@ logger = logging.getLogger(__name__)
 class ConnectionConfig(abc.ABC, BaseConfig):
     concurrent_tasks: int
     register_comments: bool
-
-    # Pydantic V2 does not include type field in dict unless also in base class
-    type_: str = Field(alias="type", default="none")
 
     @property
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -1276,3 +1277,14 @@ def _connection_config_validator(
...
+
+if t.TYPE_CHECKING:
+    SerializableConnectionConfig: t.TypeAlias = ConnectionConfig  # type: ignore
+elif PYDANTIC_MAJOR_VERSION >= 2:
+    import pydantic
+
+    # Workaround for https://docs.pydantic.dev/latest/concepts/serialization/#serializing-with-duck-typing
+    SerializableConnectionConfig = pydantic.SerializeAsAny[ConnectionConfig]  # type: ignore
+else:
+    SerializableConnectionConfig = ConnectionConfig  # type: ignore
```
georgesittas, line 1290: If the branch order isn't important, perhaps we could merge the first and the last one?

georgesittas (approved):

sqlmesh/core/config/connection.py
izeigerman: So this part:
```
...: t.TypeAias
```
is crucial for mypy to work correctly, but `TypeAlias` itself has not been introduced until Python 3.10, which means that it'd fail at runtime if evaluated.

izeigerman: Will add a comment

sqlmesh/core/config/connection.py
```diff
@@ -1276,3 +1274,16 @@ def _connection_config_validator(
...
+
+if t.TYPE_CHECKING:
+    # TypeAlias hasn't been introduced until Python 3.10 which means that we can't use it
+    # outside the TYPE_CHECKING guard.
+    SerializableConnectionConfig: t.TypeAlias = ConnectionConfig  # type: ignore
+elif PYDANTIC_MAJOR_VERSION >= 2:
+    import pydantic
+
+    # Workaround for https://docs.pydantic.dev/latest/concepts/serialization/#serializing-with-duck-typing
+    SerializableConnectionConfig = pydantic.SerializeAsAny[ConnectionConfig]  # type: ignore
```
crericha, line 1287: Is this a pattern we'll need more? If so, consider wrapping the logic in a method provided by our pydantic util module.

crericha (approved):

sqlmesh/core/config/connection.py
izeigerman: Don't want to make it generic until we actually need it. This code is trash and I'd rather keep it contained for now.

PR #2216, opened 2024-03-04
Chore: In GenericContext make the config type a class member

izeigerman:


crericha (approved):

crericha:
@izeigerman can we release this too?

PR #2218, opened 2024-03-05
non standard redshift plans with limits

tobymao:
some redshift explain plans come in with unexpected blocks, ignore them and only parse blocks with RESULTS


izeigerman (approved):
❤️

PR #2220, opened 2024-03-05
Fix: Improve error display for failed plan

mykalmax:
Provide me details on plan errors as well as better display error state on UI during `plan` or `apply`
Screenshot 2024-03-06 at 8 12 38 AM
Screenshot 2024-03-06 at 8 13 05 AM
Screenshot 2024-03-06 at 8 13 33 AM

### DIFF
--- a/web/client/src/library/components/plan/Plan.tsx
+++ b/web/client/src/library/components/plan/Plan.tsx
@@ -50,10 +50,10 @@ function Plan(): JSX.Element {
   function reset(): void {
     setPlanAction(new ModelPlanAction({ value: EnumPlanAction.Run }))
 
-    resetPlanTrackers()
     clearPlanApply()
-
     cleanUp()
+
+    setTimeout(() => resetPlanTrackers(), 500)
   }
 
   function cancel(): void {
@@ -78,7 +78,6 @@ function Plan(): JSX.Element {
   function apply(): void {
     setPlanAction(new ModelPlanAction({ value: EnumPlanAction.Applying }))
 
-    resetPlanTrackers()
     clearPlanApply()
 
     dispatch([{ type: EnumPlanActions.ResetTestsReport }])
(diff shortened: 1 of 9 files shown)


web/client/src/library/components/plan/Plan.tsx
```diff
@@ -53,10 +53,10 @@ function Plan(): JSX.Element {
   function reset(): void {
     setPlanAction(new ModelPlanAction({ value: EnumPlanAction.Run }))
 
-    resetPlanTrackers()
     clearPlanApply()
-
     cleanUp()
+
+    setTimeout(() => resetPlanTrackers(), 500)
```
mykalmax, line 59: we delay reset to make state change animation smoother

web/server/utils.py
```diff
@@ -26,15 +26,42 @@ def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
...
+                    message = exception.message
+                if exception.origin:
+                    origin = exception.origin
+                if exception.trigger:
+                    trigger = exception.trigger
+                if exception.status_code:
+                    status_code = exception.status_code
+
+            api_console.log_exception(
+                ApiException(
```
mykalmax, line 58: we need to re-instantiate `ApiException` here to catch actual traceback of an error, because that's where error actually occurred

web/server/api/endpoints/plan.py
```diff
@@ -37,6 +37,10 @@ async def initiate_plan(
                 plan_options,
                 environment,
                 plan_dates,
+                exception=ApiException(
+                    message="Unable to run a plan",
+                    origin="API -> plan -> initiate_plan",
+                ),
```
vchan, line 43: What do you think about this pattern instead of passing an exception to run_in_executor?
```
def get_plan_builder(...):
    try:
        _get_plan_builder(...)
    except Exception as e:
        tracker_stage_changes.stop(success=False)
        tracker.stop(success=False)
        api_console.log_event_plan_overview()
        raise ApiException(
            message="Unable to get plan changes",
            origin="API -> plan -> run_plan",
        ) from e

def _get_plan_builder(...):
    # original get_plan_builder function
```

PR #2222, opened 2024-03-05
Feat: use 1 thread for dbt duckdb targets

treysp:
DuckDB doesn't support concurrent writes, so we automatically use 1 thread even if the dbt profile specifies >1


crericha (approved):
Can we add a test?

tobymao (approved):

PR #2223, opened 2024-03-05
Feat: add dbt trino support

treysp:


treysp:
@eakmanrq I didn't see a dbt counterpart to SQLMesh's `BASIC` Trino authentication method on the [dbt-trino page](https://docs.getdbt.com/docs/core/connect-data-platform/trino-setup#authentication-parameters). Do I need to cover that case?

See `method_to_auth_enum` dict in Trino's `to_sqlmesh()` method in dbt/target.py.

sqlmesh/dbt/target.py
```diff
@@ -672,6 +676,120 @@ def to_sqlmesh(self) -> ConnectionConfig:
...
+        method: The Trino authentication method to use
+        host: The server host to connect to
+        port: The MSSQL server port to connect to
+        database: Name of the database
+        roles: Trino catalog roles
+        session_properties: Trino session properties
+        retries: Number of times to retry if the Trino connector encounters an error
+        timezone: The timezone to use for the Trino session
+        http_headers: HTTP Headers to send alongside requests to Trino
+        http_scheme: The HTTP scheme to use for requests to Trino (default: http, or https if kerberos, ldap or jwt auth)
```
crericha, line 693: Some of the fields are missing from args

sqlmesh/dbt/target.py
```diff
@@ -672,6 +676,120 @@ def to_sqlmesh(self) -> ConnectionConfig:
...
+        return TrinoRelation
+
+    @classproperty
+    def column_class(cls) -> t.Type[Column]:
+        from dbt.adapters.trino.column import TrinoColumn
+
+        return TrinoColumn
+
+    def to_sqlmesh(self) -> ConnectionConfig:
+        method_to_auth_enum = {
```
crericha, line 752: Should this be a class constant?

crericha (approved):

crericha:
> @eakmanrq I didn't see a dbt counterpart to SQLMesh's `BASIC` Trino authentication method on the [dbt-trino page](https://docs.getdbt.com/docs/core/connect-data-platform/trino-setup#authentication-parameters). Do I need to cover that case?
> 
> See PR's `method_to_auth_enum` dict in Trino's `to_sqlmesh()` method in dbt/target.py.

As long as we cover dbt's auth methods, we are good imo.

sqlmesh/dbt/target.py
eakmanrq: Example for reference: https://github.com/TobikoData/sqlmesh/blob/3e4f25ad31ee7bbdcacb51d68d9f2506187dd00c/sqlmesh/core/config/base.py#L100

eakmanrq:
> > @eakmanrq I didn't see a dbt counterpart to SQLMesh's `BASIC` Trino authentication method on the [dbt-trino page](https://docs.getdbt.com/docs/core/connect-data-platform/trino-setup#authentication-parameters). Do I need to cover that case?
> > See PR's `method_to_auth_enum` dict in Trino's `to_sqlmesh()` method in dbt/target.py.
> 
> As long as we cover dbt's auth methods, we are good imo.

Yeah I agree if there is not a way to express the `BASIC` way in dbt then we don't have to worry about covering it in the mapping.

examples/sushi_dbt/config.py
```diff
@@ -1,10 +1,11 @@
 from pathlib import Path
 
-from sqlmesh.core.config import AirflowSchedulerConfig
+from sqlmesh.core.config import AirflowSchedulerConfig, DuckDBConnectionConfig
 from sqlmesh.dbt.loader import sqlmesh_config
 
 config = sqlmesh_config(Path(__file__).parent)
 
+trino_config = sqlmesh_config(Path(__file__).parent, state_connection=DuckDBConnectionConfig())
```
eakmanrq, line 8: did you mean to add this? If so, why is it the only extra config in here?

eakmanrq (approved):

examples/sushi_dbt/config.py
treysp: This is a leftover from testing, but do you think it would be useful to have a config with a separate state connection so it's more convenient to run on trino/spark where we don't support main connection state? 

It's easy enough to add one if needed, though, so maybe just clutter?

tobymao: in general, we shouldn’t have testing code around.

PR #2225, opened 2024-03-06
Fix: Support syntethic projections added to the plan by redshift

izeigerman:
When a query involves window functions, Redshift would extend the plan with additional projections for the exclusive purpose of helping it compute the return value.

In this case the name of the projection is set to `<>`. This update makes sure that we skip these columns.

### DIFF
--- a/sqlmesh/core/engine_adapter/redshift.py
+++ b/sqlmesh/core/engine_adapter/redshift.py
@@ -20,6 +20,7 @@ from sqlmesh.core.engine_adapter.shared import (
     SourceQuery,
     set_catalog,
 )
+from sqlmesh.utils.errors import SQLMeshError
 
 if t.TYPE_CHECKING:
     from sqlmesh.core._typing import SchemaName, TableName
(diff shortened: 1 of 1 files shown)


sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -136,6 +137,17 @@ def _build_create_table_exp(
                 for target in plan["targetlist"]:  # type: ignore
                     if target["name"] == "TARGETENTRY":
                         resdom = target["resdom"]
+                        resname = resdom["resname"]
+                        if resname == "<>":
+                            # A synthetic column added by Redshift to compute a window function.
+                            continue
+                        if resname == "? column ?":
```
georgesittas, line 144: Does the plan add spaces between `?` and `column`? I don't remember seeing them but I could be wrong.

georgesittas (approved):

sqlmesh/core/engine_adapter/redshift.py
izeigerman: Tokenizer does. Just like the `- 1` example below

tobymao (approved):

PR #2231, opened 2024-03-06
expose additional fields in PlanDagSpec

andrew-sha:
this is meant to help `EnterpriseSnapshotDagGenerator` get some additional plan-related information for observer

### DIFF
--- a/sqlmesh/schedulers/airflow/client.py
+++ b/sqlmesh/schedulers/airflow/client.py
@@ -198,6 +198,9 @@ class AirflowClient(BaseAirflowClient):
         models_to_backfill: t.Optional[t.Set[str]] = None,
         end_bounded: bool = False,
         ensure_finalized_snapshots: bool = False,
+        directly_modified_snapshots: t.List[SnapshotId] = [],
+        indirectly_modified_snapshots: t.Dict[SnapshotId, t.Set[SnapshotId]] = {},
+        removed_snapshots: t.List[SnapshotId] = [],
     ) -> None:
         request = common.PlanApplicationRequest(
             new_snapshots=list(new_snapshots),
@@ -215,6 +218,9 @@ class AirflowClient(BaseAirflowClient):
             models_to_backfill=models_to_backfill,
             end_bounded=end_bounded,
             ensure_finalized_snapshots=ensure_finalized_snapshots,
+            directly_modified_snapshots=directly_modified_snapshots,
+            indirectly_modified_snapshots=indirectly_modified_snapshots,
+            removed_snapshots=removed_snapshots,
         )
 
         response = self._session.post(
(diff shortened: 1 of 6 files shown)


sqlmesh/schedulers/airflow/common.py
```diff
@@ -52,6 +52,9 @@ class PlanApplicationRequest(PydanticModel):
     models_to_backfill: t.Optional[t.Set[str]]
     end_bounded: bool
     ensure_finalized_snapshots: bool
+    directly_modified_snapshots: t.List[SnapshotId]
+    indirectly_modified_snapshots: t.Dict[SnapshotId, t.Set[SnapshotId]]
```
izeigerman, line 56: This key is likely not going to work with pydantic. AFAIK, Pydantic doesn't support (de)serialization of complex types as dict keys / set items.

sqlmesh/schedulers/airflow/client.py
```diff
@@ -198,6 +198,9 @@ def apply_plan(
         models_to_backfill: t.Optional[t.Set[str]] = None,
         end_bounded: bool = False,
         ensure_finalized_snapshots: bool = False,
+        directly_modified_snapshots: t.List[SnapshotId] = [],
```
izeigerman, line 201: We should never use `[]` / `{}` as default arguments in Python. See, for example, this https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

andrew-sha: oh nice I didn't know about this, thanks

sqlmesh/schedulers/airflow/client.py
```diff
@@ -198,6 +198,9 @@ def apply_plan(
         models_to_backfill: t.Optional[t.Set[str]] = None,
         end_bounded: bool = False,
         ensure_finalized_snapshots: bool = False,
+        directly_modified_snapshots: t.Optional[t.List[SnapshotId]] = None,
+        indirectly_modified_snapshots: t.Optional[t.List[SnapshotId]] = None,
```
izeigerman, line 202: Will we ever care about which snapshot caused the indirectly modified one? We can consider making it:
```
t.Dict[str, t.List[SnapshotId]]
```
and use the snapshot name as a key.

izeigerman: CC: @crericha

andrew-sha: for the specific thing im trying to do currently it isn't needed but i think this information would be nice to have in general so its probably worth including as a dict key

crericha: Yeah, which snapshot caused the indirect mod is good info. In an ideal world, Airflow would provide a `Plan` object with a full `ContextDiff`.

izeigerman (approved):

PR #2233, opened 2024-03-07
Feat: truncate bigquery comments at max allowed length 1024 chars

treysp:
Bigquery errors if a comment is >1024 characters, so this PR automatically truncates them to that length.

https://cloud.google.com/bigquery/quotas#table_limits


crericha:
Do we think this limitation is unique to BQ? I'd expect most engines have a max description length.

PR #2235, opened 2024-03-08
fix!: recursively check for unknown data types

eakmanrq:
Prior to this change you could have a struct with a unknown data type in it and we would consider this as having all the data types defined. This PR makes the data type check recursive so we will find unknown types in nested data types. 

`SchemaDiffer` was changed because I converted a classmethod it had into a utility function. No logic changes to that class or it's tests.

### DIFF
--- a/sqlmesh/utils/__init__.py
+++ b/sqlmesh/utils/__init__.py
@@ -308,11 +308,35 @@ def groupby(
     return grouped
 
 
+def columns_to_types_to_struct(
+    columns_to_types: t.Union[t.Dict[str, exp.DataType], t.Dict[str, str]]
+) -> exp.DataType:
+    """
+    Converts a dict of column names to types to a struct.
+    """
+    return exp.DataType(
+        this=exp.DataType.Type.STRUCT,
+        expressions=[
+            exp.ColumnDef(this=exp.to_identifier(k), kind=v) for k, v in columns_to_types.items()
+        ],
+        nested=True,
+    )
+
+
 def columns_to_types_all_known(columns_to_types: t.Dict[str, exp.DataType]) -> bool:
     """
     Checks that all column types are known and not NULL.
     """
-    return all(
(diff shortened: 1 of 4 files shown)


sqlmesh/utils/__init__.py
```diff
@@ -308,11 +308,37 @@ def groupby(
...
-    )
+
+    def check_struct(struct: exp.DataType) -> bool:
+        for column_def in struct.expressions:
+            if column_def.kind.is_type(exp.DataType.Type.UNKNOWN, exp.DataType.Type.NULL):
+                return False
+            if column_def.kind.is_type(exp.DataType.Type.STRUCT):
+                if not check_struct(column_def.kind):
+                    return False
+            if column_def.kind.is_type(exp.DataType.Type.ARRAY):
```
tobymao, line 338: Type.MAP

sqlmesh/utils/__init__.py
```diff
@@ -308,11 +308,37 @@ def groupby(
...
+                if column_def.kind.expressions[0].is_type(exp.DataType.Type.STRUCT):
```
tobymao, line 339: expressions[0] may not exist? this may be an index error

eakmanrq: Does it not have to exist if it is an array?

tobymao: i think it's possible to just have an ArrayType, not sure if it's practical, but the code path exists

sqlmesh/utils/__init__.py
eakmanrq: thanks I always forget about map

sqlmesh/utils/__init__.py
eakmanrq: Ok I can just check and return false if not

eakmanrq: Ended up rewriting a bit to orient around expressions instead of structs since we don't care about column names (like in SchemaDiffer).

georgesittas (approved):

PR #2238, opened 2024-03-08
Feat: truncate comments to engine-specific maximum

treysp:
Some engines error if comments are longer than a maximum length. With this PR they will be automatically truncated to the engine's max.

Some engines, like Spark and Trino, have modular compute, metadata, and storage layers. The max comment length may differ across choices of metadata layer, such as Hive (max 4000 chars) or Iceberg (no max). For modular engines, we handle truncation in the Hive mixin rather than the engine adapter itself.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -584,7 +584,7 @@ class EngineAdapter:
             return [
                 exp.ColumnConstraint(
                     kind=exp.CommentColumnConstraint(
-                        this=exp.Literal.string(self._truncate_comment(comment, "column"))
+                        this=exp.Literal.string(self._truncate_column_comment(comment))
                     )
                 )
             ]
@@ -1889,7 +1889,7 @@ class EngineAdapter:
         if table_description:
             properties.append(
                 exp.SchemaCommentProperty(
-                    this=exp.Literal.string(self._truncate_comment(table_description, "table"))
+                    this=exp.Literal.string(self._truncate_table_comment(table_description))
                 )
             )
 
@@ -1908,7 +1908,7 @@ class EngineAdapter:
         if table_description:
             properties.append(
                 exp.SchemaCommentProperty(
-                    this=exp.Literal.string(self._truncate_comment(table_description, "table"))
+                    this=exp.Literal.string(self._truncate_table_comment(table_description))
                 )
             )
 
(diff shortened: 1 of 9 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -1898,12 +1906,28 @@ def _build_view_properties_exp(
...
+                    this=exp.Literal.string(self._truncate_table_comment(table_description))
+                )
+            )
 
         if properties:
             return exp.Properties(expressions=properties)
         return None
 
+    def _truncate_comment(self, comment: str, length: t.Optional[int]) -> str:
+        if length is None:
```
crericha, line 1920: nit: you can make this a one-liner `return comment[:length] if length else comment`.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -451,10 +455,14 @@ def _create_column_comments(
...
-    def _build_description_property_exp(self, description: str) -> exp.Property:
+    def _build_description_property_exp(
+        self,
+        description: str,
+        trunc_method: t.Callable,
+    ) -> exp.Property:
         return exp.Property(
             this=exp.to_identifier("description", quoted=True),
-            value=exp.Literal.string(description),
+            value=exp.Literal.string(trunc_method(description)),
```
crericha, line 465: Just confirming that `trunc_method` is correct here.

crericha:
Approach looks good. Nice job, @treysp. I left a couple minor comments. I didn't deeply review where the truncation was added. I'll leave that to @eakmanrq who knows that code better than I.

sqlmesh/core/engine_adapter/bigquery.py
treysp: Yep - this method handles both table and column comments, so I pass the correct `_truncate_XXX_comment` method to the `trunc_method` arg here

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -184,14 +191,32 @@ def _build_view_properties_exp(
...
 
+    def _truncate_table_comment(self, comment: str) -> str:
+        # iceberg tables do not have a comment length limit
+        if self.current_catalog_type == "iceberg":
+            return comment
+
+        return super()._truncate_table_comment(comment)
+
+    def _truncate_column_comment(self, comment: str) -> str:
+        # iceberg tables do not have a comment length limit
```
eakmanrq, line 214: Copy/paste typo: `tables` -> `columns`

It might be better to add this check to `def _truncate_comment` so it is not repeated and it has no length limit regardless of comment type.

eakmanrq (approved):
Nice! One small change requested.

PR #2239, opened 2024-03-08
Fix redshift explain verbose no where limit

izeigerman:


sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -125,10 +125,18 @@ def _build_create_table_exp(
             # is applied to a ctas statement, VARCHAR types default to 1 in some instances.
             # this checks the explain plain from redshift and tries to detect when these optimizer
             # bugs occur and force a cast
-            sql = statement.sql(
+            explain_statement = statement.copy()
+            for select in explain_statement.find_all(exp.Select):
+                if select.args.get("from"):
+                    select.args["limit"] = None
+                    select.args["where"] = None
```
georgesittas, line 132: Nit(style): we usually do this for popping off args that may not exist in sqlglot:
```suggestion
                    select.set("limit", None)
                    select.set("where", None)
```

georgesittas (approved):

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -359,7 +374,25 @@ def parse_nested() -> t.Optional[t.Dict]:
...
 
-        if nested and nested.get("name") == "RESULT":
+        if nested and nested.get("name") in ("RESULT", "SEQSCAN"):
             return nested
         advance()
     return None
+
+
+# https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat
+REDSHIFT_PLAN_TYPE_MAPPINGS = {
```
tobymao, line 384: is this sufficient?

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -125,10 +125,18 @@ def _build_create_table_exp(
...
             # this checks the explain plain from redshift and tries to detect when these optimizer
             # bugs occur and force a cast
-            sql = statement.sql(
+            explain_statement = statement.copy()
+            for select in explain_statement.find_all(exp.Select):
+                if select.args.get("from"):
+                    select.set("limit", None)
+                    select.set("where", None)
+
+            explain_statement_sql = explain_statement.sql(
```
tobymao, line 134: copy=False here so that there isn't a double copy, you explicitly copy on 128 now.  sql() calls always perform a copy by default because some sql calls do mutations. but in this case you don't care, so we don't need the extra overhead

tobymao (approved):

sqlmesh/core/engine_adapter/redshift.py
izeigerman: This covers the most common Redshift types as per https://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html. We'll add more later if needed

PR #2240, opened 2024-03-08
Fix: configure duckdb with connector_config settings.

crericha:
`duckdb.connect` does not properly parse config dicts that are passed into it, so instead set each config setting directly in duckdb when we init the cursor.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -148,7 +148,7 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
 
             for field, setting in self.connector_config.items():
                 try:
-                    cursor.execute(f'SET {field} = "{setting}"')
+                    cursor.execute(f"SET {field} = '{setting}'")
                 except Exception as e:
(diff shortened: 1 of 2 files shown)


seandavi:
Thx, @crericha, for the quick fix. A checkout does fix the parsing issue from what I can tell. I'm not yet able to read from s3 due to a new error, but I'll file something once I narrow down the problem.

sqlmesh/core/config/connection.py
```diff
@@ -150,6 +146,12 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
                 except Exception as e:
                     raise ConfigError(f"Failed to load extension {extension}: {e}")
 
+            for field, setting in self.connector_config.items():
+                try:
+                    cursor.execute(f"SET {field} = '{setting}'")
```
izeigerman, line 151: Is value always a string literal?

crericha: This mimics dbt's behavior.

izeigerman (approved):

PR #2245, opened 2024-03-11
chore: serialize engine integration tests

eakmanrq:
pytest-xdist will now start evenly distributing the tests across the N workers but make sure that tests with a common `xdist_group` end up on the same worker and therefore run sequentially. The only tests with this grouping applied are these integration tests. 

Only negative with this change is we lose the "worksteal" functionality where idle workers are greedy and steal work from other workers. This is a minor negative though. 

Edit: impact of not having "worksteal" is bigger then I thought. It appears "worksteal" still tried to group modules together but did not provide guarantees of this. Therefore some module functions couldn't be parallelized but we never saw this because worksteal would group them together and therefore run them sequentially. Added grouping for those and we might heave some flaky tests for a bit until we get everything grouped together correctly. https://github.com/pytest-dev/pytest-xdist/issues/890


izeigerman (approved):
LGTM once green

PR #2255, opened 2024-03-12
Audits

tobymao:


tobymao:
> Can you please share some context around this change in the description

i’m fixing @z3z1ma pr

z3z1ma:
I see what you mean did herw tih the `@AND` and `@OR`, that makes a ton of sense. Simplifies the optional condition very cleanly. Also the usage of `@WHERE`, I hadn't cosidered thatd work with a truthy/false value like that.

z3z1ma (approved):
LGTM based on what I was trying to achieve.

PR #2263, opened 2024-03-13
Fix: Improve performance of migration

izeigerman:
This update migrates snapshot records in the DFS order which lets us achieve the following benefits:

* The fingerprint cache is now reused per snapshot parent which leads to a higher number of cache hits.
* The DFS order allows us to efficiently evict parsed snapshots from the cache, since once the child is processed we can be sure we won't be needing its parsed record any longer.

Additionally, this update introduces a new migration config option: `promoted_snapshots_only`.  When set to `True` only snapshots that are part of at least one environment will be promoted, reducing the total number of migrated records substantially. This can be helpful for really big states. The tradeoff is that some snapshots will no longer be available for a quick rollback of changes immediately following the migration.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -417,7 +417,12 @@ class StateSync(StateReader, abc.ABC):
         """
 
     @abc.abstractmethod
-    def migrate(self, default_catalog: t.Optional[str], skip_backup: bool = False) -> None:
+    def migrate(
+        self,
+        default_catalog: t.Optional[str],
+        skip_backup: bool = False,
+        promoted_snapshots_only: bool = True,
+    ) -> None:
         """Migrate the state sync to the latest SQLMesh / SQLGlot version."""
 
     @abc.abstractmethod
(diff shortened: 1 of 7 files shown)


sqlmesh/core/state_sync/base.py
```diff
@@ -417,7 +417,12 @@ def compact_intervals(self) -> None:
         """
 
     @abc.abstractmethod
-    def migrate(self, default_catalog: t.Optional[str], skip_backup: bool = False) -> None:
+    def migrate(
+        self,
+        default_catalog: t.Optional[str],
+        skip_backup: bool = False,
+        promoted_snapshots_only: bool = False,
```
tobymao, line 424: default True

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -899,7 +900,12 @@ def _restore_table(
         )
 
     @transactional()
-    def migrate(self, default_catalog: t.Optional[str], skip_backup: bool = False) -> None:
+    def migrate(
+        self,
+        default_catalog: t.Optional[str],
+        skip_backup: bool = False,
+        promoted_snapshots_only: bool = False,
```
tobymao, line 907: ditto

tobymao (approved):

PR #2264, opened 2024-03-13
Feat: add support for YAML dictionaries in unit tests

georgesittas:
Still a WIP, a refactor is under way for this. Expecting to have an MVP soon.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -145,6 +145,7 @@ class ModelTest(unittest.TestCase):
             actual = actual.sort_values(by=actual.columns.to_list()).reset_index(drop=True)
             expected = expected.apply(lambda col: col.map(_to_hashable))
             expected = expected.sort_values(by=expected.columns.to_list()).reset_index(drop=True)
+
         try:
             pd.testing.assert_frame_equal(
                 expected,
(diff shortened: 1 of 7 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -82,10 +82,17 @@ def setUp(self) -> None:
...
             if not columns_to_types and rows:
                 for i, v in rows[0].items():
-                    # convert ruamel into python
-                    v = v.real if hasattr(v, "real") else v
-                    v_type = annotate_types(exp.convert(v)).type or type(v).__name__
-                    columns_to_types[i] = exp.maybe_parse(v_type, into=exp.DataType)
+                    if isinstance(v, dict):
+                        # Since a YAML dict can be interpreted in many ways (e.g. STRUCT, MAP),
+                        # we parse it to figure out what the correct representation is per dialect
+                        v_ast: exp.Expression = exp.maybe_parse(str(v), dialect=self.dialect)
```
tobymao, line 88: i don't understand how this works. what is v? how can you parse it as sql?

georgesittas: Afaik the possible values in the yaml format are strings, numerics, booleans, null, and lists/dicts that can be nested and contain such values. These should all be parseable as sql.

georgesittas: We've discussed about this and decided to follow a different approach, so I'm temporarily making this a draft.

tobymao (approved):

PR #2265, opened 2024-03-14
chore: set airflow as library for some tests

eakmanrq:
Airflow code: https://github.com/apache/airflow/blob/7bba05d782a64da0b5cce68ec92ac6f5c7b49221/airflow/__init__.py#L55-L61

We don't want to set this generally but for these tests it should be fine. We may need to later serialize other airflow tests, or have them also set this env var, but doing these for now since I know they are problematic. I don't have a repro for the issue currently.

### DIFF
--- a/tests/schedulers/airflow/operators/test_hwm_sensor.py
+++ b/tests/schedulers/airflow/operators/test_hwm_sensor.py
@@ -1,5 +1,3 @@
-import os
-from unittest import mock
 from unittest.mock import call
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/schedulers/airflow/operators/test_hwm_sensor.py
```diff
@@ -13,10 +15,17 @@
 )
 from sqlmesh.utils.date import to_datetime
 
+pytest_plugins = ["tests.schedulers.airflow.operators.fixtures"]
 pytestmark = pytest.mark.airflow
 
 
-def test_no_current_hwm(mocker: MockerFixture, make_snapshot, random_name):
+@pytest.fixture
+def set_airflow_as_library():
```
izeigerman, line 23: Can we somehow have a single copy of this fixture? Maybe put it into `.../operators/conftest.py`?

eakmanrq: This is a typo the single copy is in `tests/schedulers/airflow/operators/fixtures.py` since it is just used in here

izeigerman (approved):

PR #2267, opened 2024-03-14
Fix!: Quote identifiers when deserializing model SQL attributes

izeigerman:
This ensures that model attributes like `partitioned_by`, `unique_key`, etc are not only normalized during the deserialization but that identifiers are quoted as well.


sqlmesh/core/model/meta.py
```diff
@@ -277,7 +278,14 @@ def _kind_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
 
             if hasattr(kind, "unique_key"):
                 kind.unique_key = [
-                    normalize_identifiers(key, dialect=dialect) for key in kind.unique_key
+                    quote_identifiers(normalize_identifiers(key, dialect=dialect), dialect=dialect)
+                    for key in kind.unique_key
+                ]
+
+            if hasattr(kind, "columns") and isinstance(kind.columns, list):
```
izeigerman, line 285: @eakmanrq I'm assuming this should've been added in order to support the `SCDType2ByColumnKind` kind correctly but was overlooked. Please confirm.

tobymao: what about time_column?

tobymao: should we make a function normalize_and_quote?

izeigerman: `time_column` is a string, so the quoting seems meaningless.

sqlmesh/utils/pydantic.py
```diff
@@ -309,10 +310,9 @@ def _get_fields(
     results = []
 
     for expr in expressions:
-        expr = normalize_identifiers(
-            exp.column(expr) if isinstance(expr, exp.Identifier) else expr,
-            dialect=dialect,
-        )
+        expr = exp.column(expr) if isinstance(expr, exp.Identifier) else expr
+        if dialect:
```
tobymao, line 314: what are the instances where there is or isn't a dialect?

sqlmesh/core/model/meta.py
izeigerman: See: https://github.com/TobikoData/sqlmesh/blob/98a4610f575a06fb32bbcad5ca3b11e81d34e66b/sqlmesh/core/model/meta.py#L277

tests/core/test_model.py
```diff
@@ -99,7 +99,7 @@ def test_load(assert_exp_eq):
     assert model.owner == "owner_name"
     assert model.dialect == "spark"
     assert model.storage_format == "iceberg"
-    assert [col.sql() for col in model.partitioned_by] == ["a", "d"]
+    assert [col.sql() for col in model.partitioned_by] == ["a", '"d"']
```
tobymao, line 102: why is only d quoted? is it because a is inferred? this again seems problematic?

sqlmesh/utils/pydantic.py
izeigerman: If the validator is used as part of the Model `kind` object the dialect won't be a part of the `values`. If it's used as par of the `ModelMeta` then it's available and can be used.

tests/core/test_model.py
izeigerman: `a` is a time column which is a plan string in the model definition.

sqlmesh/core/model/meta.py
tobymao: why the heck is it a string

sqlmesh/utils/pydantic.py
tobymao: is that ok? i haven't thought about that enough

sqlmesh/core/model/meta.py
izeigerman: it's been a long time lol

eakmanrq: Yes I agree thanks for catching this.

sqlmesh/utils/pydantic.py
eakmanrq: I'm also wondering if this is ok. Are you considering fixing that out of scope of this PR or not something that needs to be fixed?

izeigerman: So it's only relevant when we deserialize kind from state. In which case it's fine to not have a dialect cause those values will be serialized without a dialect as well.

sqlmesh/core/model/kind.py
```diff
@@ -153,53 +153,66 @@ def data_hash_values(self) -> t.List[t.Optional[str]]:
...
-                kwargs = {
-                    key: v.expressions[i].name
-                    for i, key in enumerate(("column", "format")[: len(v.expressions)])
-                }
-                return TimeColumn(**kwargs)
+                column_expr = v.expressions[0]
+                return TimeColumn(
+                    column=(
+                        exp.column(column_expr)
+                        if isinstance(column_expr, exp.Identifier)
```
tobymao, line 167: you tested this? you get back an identifier here?

izeigerman: I tested this and, yes, there are cases when I get an identifier there.

tests/core/test_state_sync.py
```diff
@@ -1426,16 +1426,16 @@ def test_rollback(state_sync: EngineAdapterStateSync, mocker: MockerFixture) ->
     state_sync.rollback()
     calls = {(a.sql(), b.sql()) for (a, b), _ in restore_table_spy.call_args_list}
     assert (
-        f'"{state_sync.schema}"."_snapshots"',
-        f'"{state_sync.schema}"._snapshots_backup',
+        f"{state_sync.schema}._snapshots",
```
eakmanrq, line 1429: Is this right?

eakmanrq: Why would it be unquoted now?

sqlmesh/core/model/kind.py
```diff
@@ -153,53 +159,76 @@ def data_hash_values(self) -> t.List[t.Optional[str]]:
...
-                return TimeColumn(column=v.name)
+                column_expr = v.expressions[0]
+                column = (
+                    exp.column(column_expr)
+                    if isinstance(column_expr, exp.Identifier)
+                    else column_expr
+                )
+                format = v.expressions[1].name if len(v.expressions) > 1 else None
+            elif isinstance(v, exp.Expression):
+                column = exp.column(v) if isinstance(v, exp.Identifier) else v
```
tobymao, line 179: can you just make this a validator for TimeColumn

tests/core/test_state_sync.py
izeigerman: Because we used to mutate expressions in the engine adapter before. Please note that this is the `restore_table` call, not the final sql

sqlmesh/core/model/kind.py
izeigerman: No, because I'll also have to propagate dialect into `TimeColumn` and not just `kind`.

tobymao (approved):

PR #2269, opened 2024-03-14
Fix: Reload context on config change

mykalmax:
- Optionally rely on `context` when working with files
- Reload `context` if `config.*` has changed
- Track newly added files to trigger `context.load()`

### DIFF
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -156,7 +156,7 @@ class Loader(abc.ABC):
             True if a modification is found; False otherwise
         """
         return any(
-            path.stat().st_mtime > initial_mtime
+            not path.exists() or path.stat().st_mtime > initial_mtime
             for path, initial_mtime in self._path_mtimes.items()
         )
 
--- a/web/server/api/endpoints/files.py
+++ b/web/server/api/endpoints/files.py
@@ -25,16 +25,9 @@ router = APIRouter()
 
 
 @router.get("", response_model=models.Directory)
-def get_files(
-    context: t.Optional[Context] = Depends(get_context),
-    settings: Settings = Depends(get_settings),
-) -> models.Directory:
+async def get_files(settings: Settings = Depends(get_settings)) -> models.Directory:
     """Get all project files."""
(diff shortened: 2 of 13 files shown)


sqlmesh/core/loader.py
```diff
@@ -156,7 +156,7 @@ def reload_needed(self) -> bool:
             True if a modification is found; False otherwise
         """
         return any(
-            path.stat().st_mtime > initial_mtime
+            not path.exists() or path.stat().st_mtime > initial_mtime
```
mykalmax, line 159: when deleting model file i was getting error `No such file or directory`

web/server/api/endpoints/files.py
```diff
@@ -26,22 +26,20 @@
...
 
 @router.get("/{path:path}", response_model=models.File)
 def get_file(
-    path: str = Depends(validate_path), settings: Settings = Depends(get_settings)
+    path: str,
+    settings: Settings = Depends(get_settings),
+    context: t.Optional[Context] = Depends(get_context),
 ) -> models.File:
     """Get a file, including its contents."""
+    path = validate_path(path, context) if context else path
```
vchan, line 42: How about we change validate_path to accept a settings object instead of a context and use settings.project_path to validate?

mykalmax: so we using context here to get `ignore_patterns` from `context.config_for_path` in case `multirepo`

vchan: Good point. I still think we can follow my suggestion and have validate_path attempt to get a context from the settings object that is passed in. We still want validate_path to prevent the app from editing files outside the project_path.

mykalmax: so basically to use both ? if context passed ,get ignore_patterns and if not juts use settings ?

vchan: I think you can pass in settings only and call get_context(settings) to get a Context (or None).

vchan (approved):

vchan (approved):

crericha (approved):

PR #2270, opened 2024-03-14
chore: publish sqlmesh tests

eakmanrq:
Publishes to an internal repo within Tobiko SQLMesh core tests so they can be easily tested from other repos. 

Also includes some adjustments to http state sync tests to make them more portable.

### DIFF
--- a/tests/core/test_state_sync.py
+++ b/tests/core/test_state_sync.py
@@ -28,6 +28,7 @@ from sqlmesh.core.snapshot import (
     Snapshot,
     SnapshotChangeCategory,
     SnapshotId,
+    SnapshotIntervals,
     SnapshotTableCleanupTask,
     missing_intervals,
 )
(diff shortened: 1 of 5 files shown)


tests/core/test_state_sync.py
```diff
@@ -211,12 +212,21 @@ def test_snapshots_exists(state_sync: EngineAdapterStateSync, snapshots: t.List[
     assert state_sync.snapshots_exist(snapshot_ids) == snapshot_ids
 
 
-def get_snapshot_intervals(state_sync, snapshot):
-    intervals = state_sync._get_snapshot_intervals([snapshot])[-1]
-    return intervals[0] if intervals else None
```
eakmanrq, line 216: This is a truly private method that was wrapped by a function for testing. By making this a fixture it can now be overriden.

tests/core/test_state_sync.py
```diff
@@ -1248,13 +1264,17 @@ def test_unpause_snapshots_remove_intervals(
         ),
         version="a",
     )
+    snapshot.categorize_as(SnapshotChangeCategory.BREAKING)
+    snapshot.version = "a"
```
eakmanrq, line 1268: In another context, I would get an error getting table info from these snapshots because they were not categorized. Therefore this categorizes them but then maintains the expected version. I believe this maintain the original intent of this test.

tests/core/test_state_sync.py
```diff
@@ -372,16 +384,20 @@ def test_get_snapshot_intervals(
         ),
         version="c",
     )
-    state_sync.add_interval(snapshot_c, "2020-01-03", "2020-01-03")
     state_sync.push_snapshots([snapshot_c])
+    state_sync.add_interval(snapshot_c, "2020-01-03", "2020-01-03")
```
eakmanrq, line 388: Confirmed privately that moving this down a line is fine for the intent of the test.

izeigerman (approved):

PR #2271, opened 2024-03-14
Feat: Record successful tests and return them in the API

vchan:


sqlmesh/core/test/result.py
```diff
@@ -1,10 +1,17 @@
 from __future__ import annotations
 
 import types
+import typing as t
 import unittest
 
 
 class ModelTextTestResult(unittest.TextTestResult):
+    successes: t.List[unittest.TestCase]
```
georgesittas, line 9: do we need this here?

georgesittas (approved):
Should we add a test for this?

sqlmesh/core/test/result.py
vchan: I added it for mypy. I think successes needs to be annotated somewhere, but it could also be done in the __init__ function. I figured here made it more apparent there's a `successes` attribute for this class.

mykalmax (approved):

PR #2275, opened 2024-03-15
Feat: add optional @limit macro

z3z1ma:
Adds limit macro. Pending an update in sqlglot to add the `skip_limit_token` kwarg.


z3z1ma:
Depends on https://github.com/tobymao/sqlglot/pull/3149

sqlmesh/core/macros.py
```diff
@@ -539,158 +539,30 @@ def filter_(evaluator: MacroEvaluator, *args: t.Any) -> t.List[t.Any]:
...
-    return expression if evaluator.eval_expression(condition) else None
-
-
-@macro()
-def order_by(
-    evaluator: MacroEvaluator,
-    condition: exp.Condition,
-    expression: exp.Order,
-) -> t.Optional[exp.Order]:
-    """Inserts ORDER BY expression when the condition is True
```
tobymao, line 677: i don't like how we lose documentation, any way we can keep these docs? perhaps in _optional expression?

z3z1ma: I added the examples back to the `_optional_expression`. These run on doc tests and show usage for the various permutations. Is this what you meant?

z3z1ma: also if its not obvious in the diff BTW, `_optional_expression` already had docs, so we werent losing docs but dropping strings that are 90% repeated.

tobymao: exactly

tobymao (approved):

PR #2281, opened 2024-03-15
Docs: python model pre/post statements

treysp:


docs/concepts/models/python_models.md
```diff
@@ -78,6 +78,40 @@ Given an execution `ExecutionContext` "context", you can fetch a DataFrame with
 df = context.fetchdf("SELECT * FROM my_table")
 ```
 
+## Optional pre/post-statements
+
+Optional pre/post-statements can help prepare the SQL engine for a model and execute "clean-up" actions after it has successfully executed, respectively. For example, you might create temporary views, set permissions, or evict previously cached tables from memory.
```
izeigerman, line 83: [Nit] The doc refers to a "clean-up" in post-statements but the actual example refers to index creation.

tobymao (approved):

PR #2287, opened 2024-03-16
Feat: Add the ability to specify a model end date

erindru:
This feature allows setting an explicit end date on an incremental model, rather than using the implicit end date of "right now".

The use-case is addressing historical / "archive" datasets that only have data for a certain date range (and typically these would be defined as forward-only as well).

In these cases, its there is no use in generating new intervals after a certain date.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -1482,12 +1482,17 @@ def missing_intervals(
             snapshot.intervals = snapshot.intervals.copy()
             snapshot.remove_interval(interval, execution_time)
 
+        missing_interval_end_date = snapshot_end_date
+        node_end_date = snapshot.node.end
+        if node_end_date and (to_datetime(node_end_date) < to_datetime(snapshot_end_date)):
+            missing_interval_end_date = node_end_date
+
         intervals = snapshot.missing_intervals(
             max(
                 to_datetime(snapshot_start_date),
                 to_datetime(start_date(snapshot, snapshots, cache, relative_to=snapshot_end_date)),
             ),
-            snapshot_end_date,
+            missing_interval_end_date,
             execution_time=execution_time,
             deployability_index=deployability_index,
             ignore_cron=ignore_cron,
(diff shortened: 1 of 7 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -1482,12 +1482,17 @@ def missing_intervals(
             snapshot.intervals = snapshot.intervals.copy()
             snapshot.remove_interval(interval, execution_time)
 
+        missing_interval_end_date = snapshot_end_date
```
tobymao, line 1485:
```suggestion
        snapshot_end_date = min(to_datetime(snapshot_end_date), to_datetime(snapshot.node.end or snapshot_end_date)
```

tobymao:
this looks good to me, thanks so much for the pr @erindru

sqlmesh/core/snapshot/definition.py
erindru: My initial implementation did this, however I observed that passing datetime objects instead of the strings to `snapshot.missing_intervals` caused "off by 1 interval" issues in the existing test suite.

So I did the check and then passed the string value, rather than passing the datetime value

tobymao: ah tricky one. yes, dates are inclusives and date times are exclusive

izeigerman (approved):
This is really good, @erindru. Well done!

izeigerman:
@erindru I've just realized that one thing to keep in mind, that this doesn't prevent model DAGs on Airflow from continuing the evaluation because they don't rely on missing intervals but on whatever intervals Airflow itself supplies. 

We should make sure to set the `end_date` here as well: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/schedulers/airflow/dag_generator.py#L114

erindru:
> @erindru I've just realized that one thing to keep in mind is that this doesn't prevent model DAGs on Airflow from continuing the evaluation because they don't rely on missing intervals but on whatever intervals Airflow itself supplies.
> 
> We should make sure to set the `end_date` here as well: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/schedulers/airflow/dag_generator.py#L114

I've given this a go. However, i'm a little unclear on how to test this change. How are you guys getting Airflow working locally? `make airflow-init` just errors out for me with a DNS error in the psql container:

`psql: error: could not translate host name "airflow-postgres-1" to address: Name or service not known`

izeigerman:
@erindru that's ok. Let's merge this and if Airflow tests fail on main, I'll take care of it. Thanks again for introducing this!

PR #2292, opened 2024-03-18
Fix: Fix partitioning property for Trino iceberg tables

erindru:
Currently, using iceberg column transforms as the `partitioned_by` model property, eg:

```
MODEL (
     name foo,
        time_column transacttime,
    ),
    partitioned_by day(transacttime),
  ....
)
```

throw a SQLMesh error:

`Error: partitioned_by field ''day(transacttime)'' does not contain a column`

A closer look at [the docs](https://trino.io/docs/current/connector/iceberg.html#partitioned-tables) reveal that, for an Iceberg table, the partition columns need to be defined as `ARRAY[varchar]`.

So this PR addresses that by wrapping the partition columns in `exp.Array` instead of `exp.Schema` and rendering them as varchar literals.

Note that Trino works with both the following syntax:

`partitioning = ARRAY['day("cola")', '"colb"']`
or
`partitioning = ARRAY['day(cola)','colb']`

(that is, if the column names are quoted within the string literals, it still works as expected)

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -157,11 +157,11 @@ class HiveMetastoreTablePropertiesMixin(EngineAdapter):
                 properties.append(
                     exp.Property(
                         this=exp.var("PARTITIONING"),
-                        value=exp.Array(
-                            expressions=[
+                        value=exp.array(
+                            *(
                                 exp.Literal.string(e.sql(dialect=self.dialect))
                                 for e in partitioned_by
-                            ]
+                            )
                         ),
                     )
                 )
@@ -169,7 +169,7 @@ class HiveMetastoreTablePropertiesMixin(EngineAdapter):
                 for expr in partitioned_by:
                     if not isinstance(expr, exp.Column):
                         raise SQLMeshError(
-                            f"PARTITIONED BY contains non-column value '{expr.sql(dialect='spark')}'."
+                            f"PARTITIONED BY contains non-column value '{expr.sql(dialect=self.dialect)}'."
                         )
 
                 properties.append(
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -146,28 +146,37 @@ class HiveMetastoreTablePropertiesMixin(EngineAdapter):
             properties.append(exp.FileFormatProperty(this=exp.Var(this=storage_format)))
 
         if partitioned_by:
-            for expr in partitioned_by:
-                if not isinstance(expr, exp.Column):
-                    raise SQLMeshError(
-                        f"PARTITIONED BY contains non-column value '{expr.sql(dialect='spark')}'."
-                    )
-
-            property: exp.Property = exp.PartitionedByProperty(
-                this=exp.Schema(expressions=partitioned_by),
-            )
-
             if (
                 self.dialect == "trino"
                 and self.get_catalog_type(catalog_name or self.get_current_catalog()) == "iceberg"
             ):
                 # On the Trino Iceberg catalog, the table property is called "partitioning" - not "partitioned_by"
                 # In addition, partition column transform expressions like `day(col)` or `bucket(col, 5)` are allowed
+                # Also, column names and transforms need to be strings and supplied as an ARRAY[varchar]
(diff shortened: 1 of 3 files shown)


tests/core/test_model.py
```diff
@@ -245,17 +245,20 @@ def test_model_qualification():
...
-        ("e", "", ConfigError),
-        ("(a, e)", "", ConfigError),
-        ("(a, a)", "", ConfigError),
+        ("a", ["`a`"], "bigquery", None),
+        ("(a, b)", ["`a`", "`b`"], "bigquery", None),
+        ("TIMESTAMP_TRUNC(`a`, DAY)", ["TIMESTAMP_TRUNC(`a`, DAY)"], "bigquery", None),
+        ("e", "", "bigquery", ConfigError),
+        ("(a, e)", "", "bigquery", ConfigError),
+        ("(a, a)", "", "bigquery", ConfigError),
+        ("(day(a),b)", ['DAY("a")', '"b"'], "trino", None),
```
erindru, line 256: I added another parameter to this test to check that parsing was working as expected. Since this test was hardcoded to `dialect=bigquery`, I added another parameter for which dialect `partition_by_output` should be rendered in

tests/core/engine_adapter/test_trino.py
```diff
@@ -148,26 +151,36 @@ def test_partitioned_by_iceberg_transforms(
...
+
+        SELECT 1::timestamp AS cola, 2::varchar as colb, 'foo' as colc;
+    """
+    )
+    model: SqlModel = t.cast(SqlModel, load_sql_based_model(expressions))
 
     adapter.create_table(
-        "test_table",
-        columns_to_types,
-        partitioned_by=[exp.to_column("day(cola)"), exp.to_column("truncate(colb, 8)")],
```
erindru, line 159: The existing test "cheated" by doctoring values for `partitioned_by` that dont actually reflect what the parser would produce when parsing a SQL Model.

So I reimplemented it based on specifying a model string more like a user would

tests/core/engine_adapter/test_trino.py
```diff
@@ -148,26 +151,36 @@ def test_partitioned_by_iceberg_transforms(
...
 
-    columns_to_types = {
-        "cola": exp.DataType.build("INT"),
-        "colb": exp.DataType.build("TEXT"),
-    }
+    expressions = d.parse(
+        f"""
+        MODEL (
+            name test_table,
+            partitioned_by (day(cola), truncate(colb, 8), colc),
```
erindru, line 158: Is this the correct syntax for specifying multiple partition columns in a SQL model? I couldn't actually find it documented anywhere. It seems to work but im not sure if I should be using `[ ]` array syntax instead

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -146,28 +146,37 @@ def _build_table_properties_exp(
...
                 # On the Trino Iceberg catalog, the table property is called "partitioning" - not "partitioned_by"
                 # In addition, partition column transform expressions like `day(col)` or `bucket(col, 5)` are allowed
+                # Also, column names and transforms need to be strings and supplied as an ARRAY[varchar]
                 # ref: https://trino.io/docs/current/connector/iceberg.html#table-properties
-                property = exp.Property(
-                    this=exp.var("PARTITIONING"), value=exp.Schema(expressions=partitioned_by)
+                properties.append(
+                    exp.Property(
+                        this=exp.var("PARTITIONING"),
+                        value=exp.Array(
```
tobymao, line 160: exp.array(*(exp.Literal.string(e.sql(dialect=self.dialect)) for e in partitioned_by)

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -146,28 +146,37 @@ def _build_table_properties_exp(
...
+                                for e in partitioned_by
+                            ]
+                        ),
+                    )
+                )
+            else:
+                for expr in partitioned_by:
+                    if not isinstance(expr, exp.Column):
+                        raise SQLMeshError(
+                            f"PARTITIONED BY contains non-column value '{expr.sql(dialect='spark')}'."
```
tobymao, line 172: dialect=self.dialect

eakmanrq (approved):
Thanks @erindru

PR #2294, opened 2024-03-18
Fix: normalize and quote properly in table diff module

georgesittas:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1710667738734639

I could repro the issue against Snowflake, and also can verify that with these changes the issues were resolved. I tested this using the example project:

```
➜  test sqlmesh table_diff prod:dev sqlmesh_example.incremental_model

Schema Diff Between 'PROD' and 'DEV' environments for model 'sqlmesh_example.incremental_model':
└── Added Columns:
    ├── FOO (DECIMAL(38, 0))
    └── BAR (DECIMAL(38, 0))


Row Counts:
├──  COMMON: 7 rows
├──  PROD ONLY: 0 rows
└──  DEV ONLY: 0 rows

        pct_match
ITEM_ID     100.0
```

I can try to write a test for this if needed.

### DIFF
--- a/sqlmesh/core/table_diff.py
+++ b/sqlmesh/core/table_diff.py
@@ -5,7 +5,8 @@ import typing as t
 
 import pandas as pd
 from sqlglot import exp
-from sqlglot.dialects.dialect import DialectType
+from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
+from sqlglot.optimizer.qualify_columns import quote_identifiers
 
 from sqlmesh.utils.pydantic import PydanticModel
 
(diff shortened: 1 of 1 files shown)


sqlmesh/core/table_diff.py
```diff
@@ -106,7 +107,6 @@ def __init__(
         target: TableName,
         on: t.List[str] | exp.Condition,
         where: t.Optional[str | exp.Condition] = None,
-        dialect: DialectType = None,
```
georgesittas, line 109: This was unused - not sure if we meant this to be part of the API, given that we have the engine adapter in here. Happy to revert this change if I missed a use case, cc @tobymao please take a look.

PR #2305, opened 2024-03-19
Fix!: Normalize dbt model cluster_by field

### LINKED ISSUE #2293: [dbt integration] Impossible to cluster on `connectionId` field (reserved field name?!)
Using latest build (0.78.dev8), I ran into following error:

If creating a `my_model.sql` with following content:
```
{{
    config(
        materialized='table',
        cluster_by = ["connectionId"],
    )
}}

SELECT 1 as connectionId
```

It fails with:
```
sqlmesh.utils.errors.ConfigError: clustered_by keys ['connectionId'] are missing in the model definition at 'my_model.sql'
```

As it should obviously works it looks like there's something with `connectionId` value that doesn't work for some reason.

crericha:
There was a gap in `clustered_by` normalization where we didn't normalize if passed in a list. This PR adds normalization for list types.

Fixes #2293

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -11,6 +11,7 @@ from sqlmesh.core.config import (
     GatewayConfig,
     ModelDefaultsConfig,
     PlanConfig,
+    PostgresConnectionConfig,
     SparkConnectionConfig,
 )
 from sqlmesh.core.notification_target import (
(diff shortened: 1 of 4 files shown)


izeigerman (approved):
Can we please add a test specific to dbt since this is where the issue occurred?

PR #2307, opened 2024-03-19
Fix: escape backslashes in comments

### LINKED ISSUE #2290: Backslahes in comments should be escaped
I had a model like this:

```sql
/* This is the comment with a \ in it */
MODEL(
  name foo.bar
);
SELECT * FROM baz
```

Results in the following error in BigQuery:

```
google.api_core.exceptions.BadRequest: 400 Syntax error: Illegal escape sequence: \  at [1:2231]; reason: invalidQuery, location: query, message: Syntax error: Illegal escape sequence: \  at [1:2231]
```

treysp:
- Escape by default for all engines but Postgres
- Do not escape when directly modifying a BigQuery table object to create column comments

Closes #2290

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -1921,18 +1921,21 @@ class EngineAdapter:
         return None
 
     def _truncate_comment(
-        self, comment: str, length: t.Optional[int], escape_backslash: t.Optional[bool] = None
+        self, comment: str, length: t.Optional[int], escape_backslash: bool = False
     ) -> str:
-        escape = escape_backslash if escape_backslash is not None else self.ESCAPE_COMMENT_BACKSLASH
-        if escape:
+        if escape_backslash:
             comment = comment.replace("\\", "\\\\")
         return comment[:length] if length else comment
 
     def _truncate_table_comment(self, comment: str) -> str:
-        return self._truncate_comment(comment, self.MAX_TABLE_COMMENT_LENGTH)
+        return self._truncate_comment(
+            comment, self.MAX_TABLE_COMMENT_LENGTH, self.ESCAPE_COMMENT_BACKSLASH
+        )
 
     def _truncate_column_comment(self, comment: str) -> str:
-        return self._truncate_comment(comment, self.MAX_COLUMN_COMMENT_LENGTH)
+        return self._truncate_comment(
+            comment, self.MAX_COLUMN_COMMENT_LENGTH, self.ESCAPE_COMMENT_BACKSLASH
+        )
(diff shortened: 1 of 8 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -1919,7 +1920,12 @@ def _build_view_properties_exp(
             return exp.Properties(expressions=properties)
         return None
 
-    def _truncate_comment(self, comment: str, length: t.Optional[int]) -> str:
+    def _truncate_comment(
+        self, comment: str, length: t.Optional[int], escape_backslash: t.Optional[bool] = None
```
tobymao, line 1924: why is this an optional bool? why isn't it just a bool with true false?

treysp: We will almost always use the value of `self.ESCAPE_COMMENT_BACKSLASH`, so I'm wanting to replicate the theoretical behavior of `def _truncate_comment(self, ..., escape_backslash = self.ESCAPE_COMMENT_BACKSLASH)` if it was legal. 

My initial implementation gave `escape_backslash` a default `True`, but then I needed to pass `escape_backslash=self.ESCAPE_COMMENT_BACKSLASH` every time the method was called.

This seemed like a better approach since that arg could be omitted unless we wanted to explicitly set it to `True` or `False`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -87,6 +87,7 @@ class EngineAdapter:
     COMMENT_CREATION_VIEW = CommentCreationView.IN_SCHEMA_DEF_AND_COMMANDS
     MAX_TABLE_COMMENT_LENGTH: t.Optional[int] = None
     MAX_COLUMN_COMMENT_LENGTH: t.Optional[int] = None
+    ESCAPE_COMMENT_BACKSLASH = True
```
georgesittas, line 90: What's the reasoning behind only switching this off for postgres?

treysp: Postgres does the escaping itself - if we also do it the backslashes are doubled in the registered comment

georgesittas: Ah, is this only specific to table / column comments? I was wondering if this is something that can be controlled by `dialect.tokenizer_class.STRING_ESCAPES` or some other constant available through SQLGlot.

treysp: Yep, just for table/column comments

georgesittas (approved):

tobymao (approved):

PR #2308, opened 2024-03-20
Fix: Create a temporary view to source the model schema instead of parsing the Redshift plan

izeigerman:
This simplifies the implementation and eliminates the need to parser the planner's output to determine column types.


sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -126,10 +126,9 @@ def _build_create_table_exp(
             # this checks the explain plain from redshift and tries to detect when these optimizer
             # bugs occur and force a cast
             explain_statement = statement.copy()
-            for select in explain_statement.find_all(exp.Select):
-                if select.args.get("from"):
-                    select.set("limit", None)
-                    select.set("where", None)
+            for select_or_union in explain_statement.find_all(exp.Select, exp.Union):
+                select_or_union.set("limit", None)
+                select_or_union.set("where", None)
```
tobymao, line 131: is there any concern of a potential cross join here? in sql you can do 

select 
from x, y
where xid =yid

removing the where significantly alters the query but maybe won’t matter

tobymao (approved):

sqlmesh/core/engine_adapter/redshift.py
tobymao: select_or_union.pop("limit")
select_or_union.pop("where")

georgesittas (approved):

PR #2314, opened 2024-03-21
Feat: Allow setting depends_on_past explicitly instead of inferring it

erindru:
ref: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1710202877518519?thread_ts=1710201667.408719&cid=C044BRE5W4S

Sometimes, you need an incremental model to run sequentially.

Setting `batch_size: 1` makes a task per interval, but it still runs batches in parallel based on the `concurrent_tasks` setting of the gateway connection.

There is an existing `depends_on_past` property that triggers sequential execution but it's currently derived and cant be set by the end user.

This PR allows it to be set as part of the model `Kind`


sqlmesh/core/model/definition.py
```diff
@@ -615,8 +615,9 @@ def is_seed(self) -> bool:
 
     @cached_property
     def depends_on_past(self) -> bool:
-        if self.kind.is_incremental_by_unique_key:
-            return True
+        explicit_depends_on_past = getattr(self.kind, "depends_on_past", None)
```
erindru, line 618: note that my original implementation did something like:

```
if self.kind.is_incremental and self.kind.depends_on_past is not None:
  return self.kind.depends_on_past
```

but this caused all the `ScdType2Kind` tests to fail because those `Kind`'s still return `True` from the `is_incremental` check but they dont inherit from `_Incremental`

sqlmesh/core/model/kind.py
```diff
@@ -270,6 +271,7 @@ class IncrementalByUniqueKeyKind(_Incremental):
     name: Literal[ModelKindName.INCREMENTAL_BY_UNIQUE_KEY] = ModelKindName.INCREMENTAL_BY_UNIQUE_KEY
     unique_key: SQLGlotListOfFields
     when_matched: t.Optional[exp.When] = None
+    depends_on_past: SQLGlotBool = True
```
erindru, line 274: This preserves the existing `INCREMENTAL_BY_UNIQUE_KEY` behaviour where depends_on_past is hardcoded to `True`

izeigerman:
As I mentioned previously, the `depends_on_past` has implications beyond just sequential execution of batches. It also impacts whether the plan start is appropriate for this model and if not a model can be excluded from the plan when iterating in dev. 

That's why I'm not sure I like the idea of somebody setting this property at will. It makes the configuration more complex, it's easy to misuse and it feels rather redundant given the mechanisms we have in place already that can achieve the same behavior.

How do you like the idea of including the model name as part of the `depends_on` attribute manually to signal that you want to opt into the `depends_on_past` behavior? I've made this behavior possible here: https://github.com/TobikoData/sqlmesh/pull/2323

Let me know what you think.

erindru:
Sure, that would work too, I didn't think of that.

I guess both methods are an indirect way of achieving the goal of making a model run sequentially.

To make the intent more clear, would a `sequential: True` property (that under the hood sets depends_on -> self) be appropriate?

Otherwise i'm happy to go with your suggestion since you have the wider context :)

izeigerman:
@erindru yes, I like `sequential`! I think it captures the intention much better

PR #2316, opened 2024-03-21
feat: improve lineage perf

tobymao:
this should significantly increase the perf of lineage, will require sqlglot 23.1

### DIFF
--- a/sqlmesh/core/lineage.py
+++ b/sqlmesh/core/lineage.py
@@ -7,6 +7,7 @@ from sqlglot import exp
 from sqlglot.helper import first
 from sqlglot.lineage import Node
 from sqlglot.lineage import lineage as sqlglot_lineage
+from sqlglot.optimizer import Scope, build_scope, qualify
 
 from sqlmesh.core.dialect import normalize_mapping_schema, normalize_model_name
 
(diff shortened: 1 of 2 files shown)


vchan (approved):
Should we trim selects here as well? https://github.com/TobikoData/sqlmesh/blob/main/web/server/api/endpoints/lineage.py#L56

tobymao:
> Should we trim selects here as well? https://github.com/TobikoData/sqlmesh/blob/main/web/server/api/endpoints/lineage.py#L56

no, we shouldn't. but this breaks some tests so i'll need some time to fix it and test

sqlmesh/core/lineage.py
```diff
@@ -15,44 +16,63 @@
...
-            *[
-                exp.alias_(exp.cast(exp.Null(), to=col_type), col)
-                for col, col_type in model.managed_columns.items()
-                if col not in query.named_selects
-            ],
-            append=True,
-            copy=False,
-        )
-    return query
+CACHE: t.Dict[str, t.Tuple[int, exp.Expression, Scope]] = {}
```
georgesittas, line 19: Should this be global? Can it ever be modified concurrently & if so, would there be issues?

vchan (approved):

web/server/api/endpoints/lineage.py
```diff
@@ -115,16 +115,10 @@ async def column_lineage(
...
-        return create_lineage_adjacency_list(model_name, column_name, context)
-    except Exception:
-        raise ApiException(
-            message="Unable to get column lineage",
-            origin="API -> lineage -> column_lineage",
-        )
+    model_name = context.get_model(model_name).fqn
+    if models_only:
+        return create_models_only_lineage_adjacency_list(model_name, column_name, context)
+    return create_lineage_adjacency_list(model_name, column_name, context)
```
georgesittas, line 121: We don't need to raise the `ApiException` here anymore?

sqlmesh/core/lineage.py
```diff
@@ -15,44 +16,63 @@
...
+    if not query or not scope:
+        query = model.render_query_or_raise()
+
+        if model.managed_columns:
+            query = query.select(
+                *(
+                    exp.alias_(exp.cast(exp.Null(), to=col_type), col)
+                    for col, col_type in model.managed_columns.items()
+                    if col not in query.named_selects
+                )
```
georgesittas, line 47: Should we set `copy=False` here?

georgesittas (approved):
Looks good, only some questions FMI.

sqlmesh/core/lineage.py
tobymao: in general our code is not really thread safe, so this is fine i think

sqlmesh/core/lineage.py
tobymao: no, we don't want to do mutate query

PR #2318, opened 2024-03-21
Fix: cast to timestamp to guarantee safe coalesce in SCD_TYPE_2_BY_COLUMN logic

z3z1ma:
SCD_TYPE_2_BY_COLUMN do not work due to a coalesce with a datetime column with a value that may be a timestamp. 

Not sure if this fix is better than making sure we get timestamps in the generated table, but for BQ this is totally broken without this.

EDIT: I manage to get out of the state on the latest release meaning this fix isnt the root cause. There is some bad state thats possible, obviously, and it seems to have to do with columns_to_types in the _scd_type_2 function where the `time_column_type` is pulled out. Its totally possible I get back into the same state tomorrow. I just have no idea how/why.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -1376,8 +1376,8 @@ class EngineAdapter:
             )
             valid_from_case_stmt = exp.func(
                 "COALESCE",
-                exp.column(f"t_{valid_from_name}"),
-                update_valid_from_start,
+                exp.cast(exp.column(f"t_{valid_from_name}"), "TIMESTAMPTZ"),
+                exp.cast(update_valid_from_start, "TIMESTAMPTZ"),
             ).as_(valid_from_name)
         else:
             assert updated_at_name is not None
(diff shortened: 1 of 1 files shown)


plaflamme:
FWIW: we're facing the same problem. The model works initially when doing `sqlmesh plan` and gets successfully constructed, but the next `sqlmesh run` fails with the `COALESCE` error where it's doing this:

```sql
COALESCE(`t_valid_from`, CAST('1970-01-01 00:00:00+00:00' AS TIMESTAMP))
```

and where `t_valid_from` ultimately has `DATETIME` type coming from the model's resulting table.

It does look like there's a missing cast somewhere and something defaults to `DATETIME`.

eakmanrq:
Thanks @z3z1ma and @plaflamme  for reporting this and we have a fix in this PR: https://github.com/TobikoData/sqlmesh/pull/2318

We are looking to create another PR that will change the default data type in BQ to be `TIMESTAMP` instead of `DATETIME`. Let me know if either of you have any concerns with this. Migration might be tricky is that is something I will be testing.

PR #2329, opened 2024-03-22
Fix: treat 'unknown' time_column fields as 'timestamp'

erindru:
Given a model like:

```
        MODEL (
            name test_table,
            kind INCREMENTAL_BY_TIME_RANGE(
                time_column ds,
            )
        );

        SELECT ds from some_other_table
```

If the type of `ds` cant be determined, it defaults to `Type.UNKNOWN`. However, as a user I _know_ that `ds` is a timestamp column - its the whole reason why its specified as the `time_column` of my model.

The problem with defaulting to `Type.UNKNOWN` (at least on the Trino adapter) is that when applying intervals, sqlmesh starts generating invalid queries like:

`DELETE FROM test_table WHERE ds BETWEEN '2024-03-20' AND '2024-03-21'`

Which causes the query engine to throw errors like:

`Cannot check if timestamp(6) is BETWEEN varchar(10) and varchar(10)`

If the column is detected as a `timestamp` instead of `unknown`, the correct query is generated:

`DELETE FROM test_table WHERE ds BETWEEN CAST('2024-03-20 00:00:00' AS TIMESTAMP) AND CAST('2024-03-20 23:59:59.999999' AS TIMESTAMP)`

Can you guys see an issue with defaulting the `time_column` to `Type.TIMESTAMP` if its type couldn't be automatically detected? It would certainly help my use-case out a lot.

Note that I cant just change the query to `select cast(ds as timestamp) as ds` because in Trino's case it can prevent predicate pushdown to the underlying connector and trigger a table scan which I definitely dont want

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1003,6 +1003,18 @@ class SqlModel(_SqlBasedModel):
                 for select in query.selects
             }
 
+            if self.time_column:
+                time_column_name = self.time_column.column.output_name
+                if self._columns_to_types[time_column_name] == exp.DataType.build(
+                    exp.DataType.Type.UNKNOWN
+                ):
+                    # if the user has specified a time column, but we cant detect its type - assume its a timestamp
+                    # if we assume its "unknown" then in practice gets treated like a string and the incremental model delete queries
+                    # will happily try to filter a timestamp column based on a string literal, which is invalid in some engines
+                    self._columns_to_types[time_column_name] = exp.DataType.build(
+                        exp.DataType.Type.TIMESTAMP
+                    )
+
         if "*" in self._columns_to_types:
             return None
 
--- a/tests/core/test_audit.py
+++ b/tests/core/test_audit.py
@@ -20,7 +20,7 @@ from sqlmesh.utils.metaprogramming import Executable
 def model() -> Model:
     return create_sql_model(
         "db.test_model",
-        parse_one("SELECT a, b, ds"),
+        parse_one("SELECT a, b, ds::varchar"),
         kind=IncrementalByTimeRangeKind(time_column="ds"),
     )
 
@@ -29,7 +29,7 @@ def model() -> Model:
 def model_default_catalog() -> Model:
     return create_sql_model(
         "db.test_model",
-        parse_one("SELECT a, b, ds"),
+        parse_one("SELECT a, b, ds::varchar"),
         kind=IncrementalByTimeRangeKind(time_column="ds"),
         default_catalog="test_catalog",
     )
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -1373,14 +1373,16 @@ def test_parse(assert_exp_eq):
 
         SELECT
           id::INT AS id,
-          ds
+          ds,
+          foo
         FROM x
         WHERE ds BETWEEN '{{ start_ds }}' AND @end_ds
     """
     )
     model = load_sql_based_model(expressions, dialect="hive")
     assert model.columns_to_types == {
-        "ds": exp.DataType.build("unknown"),
+        "foo": exp.DataType.build("unknown"),
+        "ds": exp.DataType.build("timestamp"),
         "id": exp.DataType.build("int"),
     }
     assert not model.annotated
@@ -1392,7 +1394,8 @@ def test_parse(assert_exp_eq):
         """
       SELECT
         CAST("id" AS INT) AS "id",
-        "ds" AS "ds"
+        "ds" AS "ds",
+        "foo" AS "foo"
       FROM "x" AS "x"
       WHERE
         "ds" BETWEEN '1970-01-01' AND '1970-01-01'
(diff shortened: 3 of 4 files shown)


tests/core/test_audit.py
```diff
@@ -20,7 +20,7 @@
 def model() -> Model:
     return create_sql_model(
         "db.test_model",
-        parse_one("SELECT a, b, ds"),
+        parse_one("SELECT a, b, ds::varchar"),
```
erindru, line 23: Every test in this class was based around the `time_column` being treated as a string, so I made this explicit instead of updating all the assertions

tests/core/test_model.py
```diff
@@ -1373,14 +1373,16 @@ def test_parse(assert_exp_eq):
...
+          ds,
+          foo
         FROM x
         WHERE ds BETWEEN '{{ start_ds }}' AND @end_ds
     """
     )
     model = load_sql_based_model(expressions, dialect="hive")
     assert model.columns_to_types == {
-        "ds": exp.DataType.build("unknown"),
+        "foo": exp.DataType.build("unknown"),
```
erindru, line 1384: This test was explicitly checking for the time_column to be `unknown`. Since my change treats unknown time_column's as timestamps, I added another non time_column to this test to show that `unknown` still works for columns that arent the `time_column`

tobymao:
have you tried specifying column types in the model def?

erindru:
Yes, the problem with that is that it's viral. You cant just specify the type of one column, you have to specify all of them

sqlmesh/core/model/definition.py
```diff
@@ -1003,6 +1003,18 @@ def columns_to_types(self) -> t.Optional[t.Dict[str, exp.DataType]]:
...
+            if self.time_column:
+                time_column_name = self.time_column.column.output_name
+                if self._columns_to_types[time_column_name] == exp.DataType.build(
+                    exp.DataType.Type.UNKNOWN
+                ):
+                    # if the user has specified a time column, but we cant detect its type - assume its a timestamp
+                    # if we assume its "unknown" then in practice gets treated like a string and the incremental model delete queries
+                    # will happily try to filter a timestamp column based on a string literal, which is invalid in some engines
+                    self._columns_to_types[time_column_name] = exp.DataType.build(
+                        exp.DataType.Type.TIMESTAMP
```
tobymao, line 1015: i think we need to think about the implications of this and test every single engine. the original design defaults to date string because those are common in hive / spark.

additionally this may not be the place to do this. we could probably just add this into the section where we add the time filter

erindru: Agreed, although the existing test suite does pass no problem.

Even if we change where this is done, I guess the core question still remains - is it valid to default the column specifically designated as the incremental model ` time_column`  to `timestamp` if its type is unknown? I personally think it is; how could you have an incremental model based on a `time_column` that isnt a date or timestamp type?

Another approach might be to make `columns (.. )` on the model definition override the auto-detected column types rather than replace them entirely (which would also be a potentially breaking change I think).

Then I could just override the type of the `time_column` to be what I need and leave everything else alone

tobymao: well it’s valid because your time column could be a string, which is common in spark. and comparing a date string to a timestamp works fine in spark. it just doesn’t work in trino which has stricter types

erindru: Is it detected as a string though? This would only change it to a timestamp if it was detected as "unknown". If it was detected as "varchar" or whatever the Spark string type is, it would remain as a string

tobymao:
@eakmanrq @izeigerman thoughts here?

sqlmesh/core/model/definition.py
izeigerman: Based on this conversation, it feels like there's no one-size-fits-all solution, and the behavior largely depends on a specific engine. As such I'd suggest solving this for individual engines as well. We should probably introduce this in specific engine adapter implementations.

erindru: Ok, no problem, I can make this specific to the Trino adapter instead.

There is another potential approach though - and that is allowing the `columns` property on the Model to be merged into `columns_to_types` without replacing them entirely.

Then I could just do:
```
MODEL (
  ...
  columns (
    ds timestamp
  )
);
```

What do you think of this approach?

This would also help some of my other models where part of the resultset is determined by a Trino table function which means that some of the types cannot be determined just by analysing the SQL. Right now I have to painstakingly specify every single column type, even the ones that can be inferred, but it would be nice not to have to :)

erindru:
To summarize the big thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1711487799758899

- Incremental models should be written with an explicit WHERE clause that includes the correct timestamp filter. The implicit WHERE predicate is a safety guard and shouldn't be relied on
- This reduces the impact of `cast()`'ing columns to the right type, because you can `cast()` the output column but filter on the uncasted input column which addresses the problem of predicates on casted columns not pushing down.
- So the recommendation is, when SQLMesh doesnt detect a timestamp type correctly, to cast the column to a timestamp in the main query and add a WHERE clause that filters on the uncasted version
- Type hints might still be useful; this has been deferred to the backlog

PR #2330, opened 2024-03-22
Feat: support Trino Delta Lake connector

treysp:
- Only tested with Hive catalog type
- Converts timestamp types in DDL schemas
    - delta_lake only supports two timestamp data types. We convert other timestamp types to those two for use in DDL statements. Trino/delta automatically converts the data values to the correct type on write, so we only need to handle the column types in DDL. https://trino.io/docs/current/connector/delta-lake.html#delta-lake-to-trino-type-mapping
        - `timestamp(6)` for non-timezone-aware
        - `timestamp(3) with time zone` for timezone-aware

### DIFF
--- a/sqlmesh/core/engine_adapter/trino.py
+++ b/sqlmesh/core/engine_adapter/trino.py
@@ -194,7 +194,7 @@ class TrinoEngineAdapter(
         is_view: bool = False,
     ) -> exp.Schema:
         if self.current_catalog_type == "delta_lake":
-            columns_to_types = self.to_delta_ts(columns_to_types)
+            columns_to_types = self._to_delta_ts(columns_to_types)
 
         return super()._build_schema_exp(
             table, columns_to_types, column_descriptions, expressions, is_view
@@ -218,7 +218,7 @@ class TrinoEngineAdapter(
         column_descriptions: t.Optional[t.Dict[str, str]] = None,
     ) -> None:
         if columns_to_types and self.current_catalog_type == "delta_lake":
-            columns_to_types = self.to_delta_ts(columns_to_types)
+            columns_to_types = self._to_delta_ts(columns_to_types)
 
         return super()._scd_type_2(
             target_table,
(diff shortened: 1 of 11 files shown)


sqlmesh/core/engine_adapter/trino.py
```diff
@@ -183,3 +184,78 @@ def _df_to_source_queries(
...
+        )
+
+    # delta_lake only supports two timestamp data types. This method converts other
+    # timestamp types to those two for use in DDL statements. Trino/delta automatically
+    # converts the data values to the correct type on write, so we only need to handle
+    # the column types in DDL.
+    # - `timestamp(6)` for non-timezone-aware
+    # - `timestamp(3) with time zone` for timezone-aware
+    # https://trino.io/docs/current/connector/delta-lake.html#delta-lake-to-trino-type-mapping
+    def to_delta_ts(self, columns_to_types: t.Dict[str, exp.DataType]) -> t.Dict[str, exp.DataType]:
```
eakmanrq, line 247: This can be private (`_to_delta_ts`). Nothing outside of Trino itself uses it and we don't want to commit to maintaining it in the public API once we do a 1.0.0 release.

sqlmesh/core/engine_adapter/trino.py
```diff
@@ -183,3 +184,78 @@ def _df_to_source_queries(
...
+        ts6 = exp.DataType.build("timestamp(6)")
+        ts3_tz = exp.DataType.build("timestamp(3) with time zone")
+
+        delta_columns_to_types = {
+            k: ts6 if v.is_type(exp.DataType.Type.TIMESTAMP) and not v == ts6 else v
```
eakmanrq, line 252: Could this be `ts6 if v.is_type(exp.DataType.Type.TIMESTAMP) else v`? Same for the ts3 check below this.

tests/core/engine_adapter/docker-compose.yaml
```diff
@@ -60,6 +60,15 @@ services:
     volumes:
       - ./trino/initdb.sql:/docker-entrypoint-initdb.d/initdb.sql
 
+  trino_iceberg_delta_metastore_db:
```
eakmanrq, line 63: Can you leave a comment explaining why it has a separate db? I believe it is because of errors about too many connections so it would be good to document in case someone in the future wants to reduce this down.

tests/core/engine_adapter/test_trino.py
```diff
@@ -89,32 +95,37 @@ def mock_fetchone(sql):
...
-        return_value="datalake_hive",
-    )
-    assert adapter.get_catalog_type("datalake_hive") == "hive"
+    storage_types = ("hive", "delta")
 
-    columns_to_types = {
-        "cola": exp.DataType.build("INT"),
-        "colb": exp.DataType.build("TEXT"),
-    }
+    for storage_type in storage_types:
```
eakmanrq, line 106: What do you think about doing this with `@pytest.mark.parametrize(` having delta and hive be parameters? That would remove this inner loop and allow the test to assert just two commands are run instead of 2 * 2. Same applies to other tests where you have this inner loop over storage types.

PR #2334, opened 2024-03-22
Fix: keyboard bindings to work on win/mac

mykalmax:


web/client/src/library/components/editor/Editor.tsx
```diff
@@ -95,7 +95,7 @@ function EditorMain({ tab }: { tab: EditorTab }): JSX.Element {
     return [
       ...defaultKeymapsEditorTab,
       {
-        key: 'Ctrl-Enter',
+        key: 'Shift-Mod-Enter',
```
vchan, line 98: What's Mod?

vchan (approved):

web/client/src/library/components/editor/Editor.tsx
mykalmax: Codemirror using `Mod` a shorthand for Cmd- on Mac and Ctrl- on other platforms. So Mod-b is Ctrl-b on Linux but Cmd-b on macOS.

PR #2338, opened 2024-03-24
Feat: improve unit test failure reporting when rows are missing

georgesittas:
Fixes sixth item in https://github.com/TobikoData/sqlmesh/issues/1637

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import typing as t
 import unittest
+from collections import Counter
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -485,6 +484,49 @@ def generate_test(
...
+        referenced_columns = {col for row in rows for col in row}
+        _raise_if_unknown_columns(columns, referenced_columns)
+
+        if partial:
+            columns = list(referenced_columns)
+
+    return pd.DataFrame.from_records(rows, columns=columns)
+
+
+def _raise_if_unknown_columns(
```
eakmanrq, line 500: would `_raise_if_extra_columns` make more sense? Based on the name I was expecting a check for `UNKNOWN` data type.

georgesittas: Yeah, I can see the confusion. Maybe an even better name could be `_raise_if_unexpected_columns`?

PR #2340, opened 2024-03-25
Feat: A CLI command to print the name of the physical table for a model

izeigerman:
Adding a new CLI command to help figure out the name of the physical table given a model name. The local model snapshot is used for this purpose.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -341,6 +341,18 @@ Options:
   --help           Show this message and exit.
 ```
 
+## table_name
+```
+Usage: sqlmesh table_name [OPTIONS] MODEL_NAME
+
+  Prints the name of the physical table for the given model.
+
+Options:
+  --dev   Print the name of the snapshot table used for previews in
+          development environments.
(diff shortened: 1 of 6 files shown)


tobymao:
can you add notebook?

izeigerman:
@tobymao what's the use case for it? Can't notebook call the context method directly?

tobymao:
we try to mimic all cli commands as magics, do we want to change that?

izeigerman:
@tobymao added the magic

eakmanrq:
@izeigerman Can you also add tests for both CLI and magics? 

https://github.com/TobikoData/sqlmesh/blob/2fd6721594dc3f7dc2cfafe57384d23943d8817b/tests/cli/test_cli.py
https://github.com/TobikoData/sqlmesh/blob/2fd6721594dc3f7dc2cfafe57384d23943d8817b/tests/integrations/jupyter/test_magics.py

izeigerman:
@eakmanrq added tests, thanks

tobymao (approved):

PR #2347, opened 2024-03-26
fix: ensure start if using end

eakmanrq:
Prior to this PR a user would get validation error if defining an end without start that wasn't clear on the issue. This validates prior to the actual value validation if an end is defined without a start and gives a more clear error.

I think requiring start with end seems reasonable. Open to other thoughts though.

### DIFF
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -258,7 +258,11 @@ class _Node(PydanticModel):
                 raise ConfigError(
                     f"Interval unit of '{interval_unit}' is larger than cron period of '{cron}'"
                 )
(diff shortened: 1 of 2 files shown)


tobymao:
would there be an issue if the project had a start?

eakmanrq:
> would there be an issue if the project had a start?

If you are talking about a model default start, then that is covered with this test: https://github.com/TobikoData/sqlmesh/blob/e42201548d30f092455ce6eac68e9b1b1cdb2282/tests/core/test_model.py#L3328-L3344

tobymao (approved):
nice

PR #2351, opened 2024-03-27
Fix: Normalize array before timestamps during test creation

vchan:


georgesittas (approved):
Looks good.

We should probably refactor these tests soon, they've started becoming very noisy and it seems like some parts could be factored out into helpers.

PR #2356, opened 2024-03-27
docs: fix docs for dag CLI

andrew-sha:
Documentation for the `sqlmesh dag` CLI on the models guide was incorrect and contradicted documentation for the command in other places. Also, a minor change to incremental time guide: in that sentence 'it' could refer to the model or the command so I made it explicit

### DIFF
--- a/docs/guides/models.md
+++ b/docs/guides/models.md
@@ -244,6 +244,6 @@ sudo apt-get install graphviz
 
 To view the DAG, enter the following command:
 
-`sqlmesh dag`
+`sqlmesh dag FILE`
 
-The generated files (.gv and .jpeg formats) will be placed at the root of your project folder.
+A file containing your project's DAG will be placed at the root of your project folder. Save this file using the `.html` extension and open it in your browser to view.
--- a/docs/guides/models.md
+++ b/docs/guides/models.md
@@ -246,4 +246,4 @@ To view the DAG, enter the following command:
 
 `sqlmesh dag FILE`
 
-A file containing your project's DAG will be placed at the root of your project folder. Save this file using the `.html` extension and open it in your browser to view.
+An html file containing your project's DAG will be placed at the root of your project folder. The DAG can then be viewed by opening this file in your browser.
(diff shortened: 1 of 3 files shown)


docs/guides/models.md
```diff
@@ -244,6 +244,6 @@ sudo apt-get install graphviz
 
 To view the DAG, enter the following command:
 
-`sqlmesh dag`
+`sqlmesh dag FILE`
 
-The generated files (.gv and .jpeg formats) will be placed at the root of your project folder.
\ No newline at end of file
+A file containing your project's DAG will be placed at the root of your project folder. Save this file using the `.html` extension and open it in your browser to view.
```
eakmanrq, line 249: The file is automatically saved as an html file right? I wonder if this would be better:
```
An html file containing your project's DAG will be placed at the root of your project folder. The html file can be viewed by opening it with your browser.
```

andrew-sha: Doesn't look like it--I did `sqlmesh dag DAG` and an extension-less file named `DAG` got created that can't be opened using browser until I append the `.html` extension

eakmanrq: Oh dang. I see why you worded it this way. Want to potentially improve that? Not sure if this is a meaningful reason why we don't at least make sure it is easy for the user to open.

andrew-sha: yes, happy to take a look!

sqlmesh/core/context.py
```diff
@@ -1261,6 +1261,8 @@ def render_dag(self, path: str, select_models: t.Optional[t.Collection[str]] = N
             path: filename to save the dag html to
             select_models: A list of model selection strings that should be included in the dag.
         """
+        if not path.endswith(".html"):
+            path += ".html"
```
eakmanrq, line 1265: pathlib is useful for getting the file extension (they call it suffix for some reason):
```
assert pathlib.Path("example.txt") == ".txt"
```
So what you could do is something like this:
```
file_path = pathlib.Path(path)
suffix = pathlib.Path(file_path)
if suffix != ".html":
  logger.warning("Provided a file extension that is not `.html` but the output is an `.html` file. Updating to be an html file.")
  path = str(file_path.with_suffix(".html"))
```

This way you don't end up with something like `file.txt.html`.

eakmanrq (approved):

PR #2358, opened 2024-03-27
fix: correctly skip loading archived snapshots

eakmanrq:
Prior to this PR if you had a plan that required loading beyond a model's end date, for example catching up a group of models when deploying to prod since they fell behind, you would get an error saying that start cannot exceed end. 

This PR makes it so that cases where start > end then we will just return that there are no missing intervals. This should preserve the intended behavior of allowing these models to be active in dags, and potentially have changes, but never exceeding their desired end date.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -771,6 +771,9 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
         Returns:
             A list of all the missing intervals as epoch timestamps.
         """
+        # If the node says that it has an end, and we are wanting to load past it, then we can return no empty intervals
+        if self.node.end and to_datetime(start) > to_datetime(self.node.end):
+            return []
         # If the amount of time being checked is less than the size of a single interval then we
         # know that there can't being missing intervals within that range and return
         validate_date_range(start, end)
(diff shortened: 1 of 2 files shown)


tests/core/test_scheduler.py
```diff
@@ -273,3 +273,12 @@ def test_intervals_with_end_date_on_model(mocker: MockerFixture, make_snapshot):
...
     assert batches[0] == (to_datetime("2023-01-01"), to_datetime("2023-01-02"))
     assert batches[-1] == (to_datetime("2023-01-10"), to_datetime("2023-01-11"))
+
+    # generate for the last day of range
+    batches = scheduler.batches(start="2023-01-31", end="2023-01-31")[snapshot]
+    assert len(batches) == 1
+    assert batches[0] == (to_datetime("2023-01-31"), to_datetime("2023-02-01"))
+
+    # generate for future days to ensure no future batches are loaded
+    snapshot_to_batches = scheduler.batches(start="2023-02-01", end="2023-02-28")
```
eakmanrq, line 283: This test raises an exception is run on main and is fixed in this PR.

vchan (approved):

izeigerman (approved):

sqlmesh/core/snapshot/definition.py
```diff
@@ -771,6 +771,9 @@ def missing_intervals(
         Returns:
             A list of all the missing intervals as epoch timestamps.
         """
+        # If the node says that it has an end, and we are wanting to load past it, then we can return no empty intervals
+        if self.node.end and to_datetime(start) > to_datetime(self.node.end):
```
izeigerman, line 775: Shouldn't this be `>=`?

eakmanrq: This test would fail if it was like that (why I added the test): https://github.com/TobikoData/sqlmesh/blob/bbe357ea48fa4ccb83f1ca6e1a12900eb9514097/tests/core/test_scheduler.py#L278-L280

PR #2361, opened 2024-03-28
Feat: allow UNION macros as model queries

georgesittas:


sqlmesh/core/model/definition.py
```diff
@@ -1477,8 +1477,12 @@ def load_sql_based_model(
         **meta_fields,
     )
 
-    if query_or_seed_insert is not None and isinstance(
-        query_or_seed_insert, (exp.Query, d.JinjaQuery)
+    if query_or_seed_insert is not None and (
+        isinstance(query_or_seed_insert, (exp.Query, d.JinjaQuery))
+        or (
+            isinstance(query_or_seed_insert, d.MacroFunc)
```
tobymao, line 1483: we should allow any macro function because it could return a query right?

georgesittas: I'm not sure, some macro functions don't return queries and we'll silently ignore that and crash at render time if we try to get their named selects or something.

tobymao: don’t we only get named selects after render?

tobymao: i guess fail later on?

georgesittas: Ok sounds good. It seems like we [raise](https://github.com/TobikoData/sqlmesh/blob/b8f9d1e245c9916c3266055ae23d6c1679b420ee/sqlmesh/core/renderer.py#L433) in `render_query` so it should catch it:

georgesittas: Ah, it's actually not that simple. Some macro funcs are not meant as queries (e.g. `@DEF`). So, if we allow any `MacroFunc` here would be ambiguous. I tried changing the conditions to get rid of the `== "union"` part and ~10 tests failed.

tobymao: what tests?

georgesittas:
```
FAILED tests/core/test_model.py::test_no_query - Failed: DID NOT RAISE <class 'sqlmesh.utils.errors.ConfigError'>
FAILED tests/core/test_model.py::test_model_pre_post_statements - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_model.py::test_load - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_model.py::test_seed_pre_post_statements - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at 'examples/sushi/models/test_model.sql'
FAILED tests/core/test_model.py::test_json_serde - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_model.py::test_render_definition - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_snapshot_evaluator.py::test_runtime_stages - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_snapshot_evaluator.py::test_evaluate - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
FAILED tests/core/test_model.py::test_render_query - sqlmesh.utils.errors.ConfigError: Only one SELECT query is allowed per model at '.'
```

> can we allow user defined macros?

Allowing arbitrary macros to be used for model queries introduces ambiguity. For example:

```sql
MODEL (...);

@macro1();
@macro2();
@macro3();
@macro4();
```

How do you know which is the query? Any of these could be pre/post-statements as well.

georgesittas: Imo it doesn't seem like supporting arbitrary macros as SQL model queries is a pattern we really want to encourage. If someone wants to do this, they could just create a python / SQL model. `@UNION` looks like an exception in this case, as it's the only builtin macro that definitely renders into a `Query`.

tobymao (approved):

PR #2362, opened 2024-03-28
Feat: add support for freezing time in unit tests

georgesittas:
Addresses the 8th item in https://github.com/TobikoData/sqlmesh/issues/1637

This seemed like a nice QOL feature to have in unit tests. The relevant Slack thread is: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1711540655441619.


sqlmesh/core/test/definition.py
```diff
@@ -387,6 +398,17 @@ def runTest(self) -> None:
 
             self.assert_equal(expected, actual_df, sort=False, partial=partial)
 
+    def _execute_model(self) -> pd.DataFrame:
+        """Executes the python model and returns a DataFrame."""
+        with (
+            freeze_time(self._freeze_time) if self._freeze_time else nullcontext(),  # type: ignore
```
eakmanrq, line 404: TIL: `nullcontext()`

sqlmesh/core/test/definition.py
```diff
@@ -331,6 +344,11 @@ def runTest(self) -> None:
 
             self.assert_equal(expected, actual, sort=sort, partial=partial)
 
+    def _execute(self, query: exp.Expression) -> pd.DataFrame:
+        """Executes the query with the engine adapter and returns a DataFrame."""
+        with patch.dict(self._engine_adapter_dialect.generator_class.TRANSFORMS, self._transforms):
```
izeigerman, line 349: Does `Dialect.get_or_raise(self.engine_adapter.dialect)` return a global instance or a new one each time? If it's the latter, can't we just set transformers once and not patch it?

georgesittas: Unfortunately we can't do this, because `TRANSFORMS` is a class attribute and since dialects are singletons we'd end up mutating every dialect if we didn't patch it.

eakmanrq (approved):

PR #2364, opened 2024-03-28
Fix!: Handle maps during test generation

vchan:
This PR depends on https://github.com/tobymao/sqlglot/pull/3241.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -592,4 +592,7 @@ def _normalize_dataframe(value: t.Any) -> t.Any:
     """Normalize data in a pandas dataframe so ruamel and sqlglot can deal with it."""
     if isinstance(value, np.ndarray):
         return [_normalize_dataframe(v) for v in value]
+    if isinstance(value, dict) and "key" in value and "value" in value:
+        # {'key1': 10, 'key2': 20, 'key3': 30} looks like {'key': ['key1', 'key2', 'key3'], 'value': [10, 20, 30]}
+        return dict(zip(value["key"], _normalize_dataframe(value["value"])))
     return value
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -592,7 +592,10 @@ def _normalize_dataframe(value: t.Any) -> t.Any:
     """Normalize data in a pandas dataframe so ruamel and sqlglot can deal with it."""
     if isinstance(value, np.ndarray):
         return [_normalize_dataframe(v) for v in value]
-    if isinstance(value, dict) and "key" in value and "value" in value:
-        # {'key1': 10, 'key2': 20, 'key3': 30} looks like {'key': ['key1', 'key2', 'key3'], 'value': [10, 20, 30]}
-        return dict(zip(value["key"], _normalize_dataframe(value["value"])))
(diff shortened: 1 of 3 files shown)


tests/core/test_test.py
```diff
@@ -893,6 +893,94 @@ def test_test_generation_with_array(tmp_path: Path) -> None:
         "query": [{"array_col": [["value1"], ["value2", "value3"]]}]
     }
 
+    # Array of maps
+    input_queries = {
+        "sqlmesh_example.bar": "SELECT [MAP {'key1': 'value1'}, MAP {'key2': 'value2'}] AS array_col"
+    }
+
+    context.create_test(
```
georgesittas, line 901: Can we reduce these three `create_test` calls into a single one by adding a couple more projections?

georgesittas: Ditto about the maps tests

georgesittas: Maybe one way to reduce the clutter is to have a single model we wanna create a test for. It would then contain as many projections as there are cases we want to test and we'd do it in one go to avoid the boilerplate. Wdyt?

sqlmesh/core/test/definition.py
```diff
@@ -549,4 +549,7 @@ def _normalize_dataframe(value: t.Any) -> t.Any:
     """Normalize data in a pandas dataframe so ruamel and sqlglot can deal with it."""
     if isinstance(value, np.ndarray):
         return [_normalize_dataframe(v) for v in value]
+    if isinstance(value, dict) and "key" in value and "value" in value:
+        # {'key1': 10, 'key2': 20, 'key3': 30} looks like {'key': ['key1', 'key2', 'key3'], 'value': [10, 20, 30]}
+        return dict(zip(value["key"], _normalize_dataframe(value["value"])))
```
georgesittas, line 554: We may need to parameterize this dict normalization logic per dialect. This will depend on whether other engines follow this format for maps.

tests/core/test_test.py
vchan: Isn't that what we're doing with the input_queries? We're reusing `sqlmesh_example.foo` and `sqlmesh_example.bar` and changing input_queries to the projections we want to test. There are a lot of unnecessary asserts that were copied and pasted I think we can remove.

sqlmesh/core/test/definition.py
```diff
@@ -549,4 +549,10 @@ def _normalize_dataframe(value: t.Any) -> t.Any:
     """Normalize data in a pandas dataframe so ruamel and sqlglot can deal with it."""
     if isinstance(value, np.ndarray):
         return [_normalize_dataframe(v) for v in value]
+    if isinstance(value, dict):
+        if "key" in value and "value" in value:
+            # Maps have the following structure: {'key': ['key1', 'key2', 'key3'], 'value': [10, 20, 30]}
+            # so we convert to {'key1': 10, 'key2': 20, 'key3': 30}
```
georgesittas, line 555:
```suggestion
            # Maps returned by DuckDB have the following structure: {'key': ['key1', 'key2', 'key3'], 'value': [10, 20, 30]}
            # so we convert to {'key1': 10, 'key2': 20, 'key3': 30}
            # TODO: handle more dialects here
```

georgesittas (approved):

PR #2370, opened 2024-03-30
fix: don't warn duckdb recommended state

eakmanrq:
In order to ensure new user experience doesn't have warning, we are not warning on duckdb. We are going to assume that if someone is using Duckdb then they know it is not recommended as a state sync.

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -79,7 +79,12 @@ class _EngineAdapterStateSyncSchedulerConfig(_SchedulerConfig):
                 f"The {engine_adapter.DIALECT.upper()} engine cannot be used to store SQLMesh state - please specify a different `state_connection` engine."
                 + " See https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#gateways for more information."
             )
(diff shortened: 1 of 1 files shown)


sqlmesh/core/config/scheduler.py
```diff
@@ -79,7 +79,12 @@ def create_state_sync(self, context: GenericContext) -> StateSync:
                 f"The {engine_adapter.DIALECT.upper()} engine cannot be used to store SQLMesh state - please specify a different `state_connection` engine."
                 + " See https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#gateways for more information."
             )
-        if not state_connection.is_recommended_for_state_sync:
+        # We don't want to log warning on duckdb in order to avoid warning when testing out SQLMesh
+        # We assume the user knows that they shouldn't be using duckdb in production
+        if (
+            not state_connection.is_recommended_for_state_sync
+            and state_connection.type_ != "duckdb"
```
tobymao, line 86: I think we should add duckdb to is_recommended_for_state_sync because it's totally fine for a single user to use duckdb in production.

PR #2376, opened 2024-04-01
Fix!: Data hash values for the model kind

izeigerman:
Turns out we weren't including many relevant model kind attributes into the snapshot fingerprint.


sqlmesh/core/model/kind.py
```diff
@@ -255,6 +260,23 @@ def _kind_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
             raise ValueError("batch_size cannot be less than lookback")
         return values
 
+    @property
+    def data_hash_values(self) -> t.List[t.Optional[str]]:
+        return [
+            *super().data_hash_values,
+            self.dialect,
+            str(self.lookback) if self.lookback is not None else None,
```
eakmanrq, line 268: Couldn't this be a metadata change? It affects run behavior there do we really need to rebuild the data itself for the past?

izeigerman: Wouldn't it mean that the previous data ends up being wrong? Cause we had to rerun those intervals but didn't?

eakmanrq: They could restate if that is the case right? Otherwise we are considering all of the past as invalid which is also not correct. It is just (potentially) recent past.

eakmanrq: Ok nevermind this seems fine

PR #2391, opened 2024-04-03
Feat!: Support variables in native SQLMesh projects

izeigerman:
This PR introduces support for user-defined variables to SQLMesh native projects. The variables can now be accessed in the following contexts:

* SQL model definitions
* Python model definitions
* Standalone audits
* Native macros
* Jinja macros

Additionally, SQLMesh intelligently detects variable names referenced in all aforementioned contexts and only includes values for referenced variables into snapshots. This means that if a variable changes, only models that actually use it will be impacted.

However, this capability comes with a few constraints, as the usage detection is performed statically by introspecting relevant SQL/Python/Jinja code.

* When using the `@VAR` macro in SQL, the first argument must always be a string literal. The same limitation applies to the `{{ var(...) }}` macro in Jinja code.
* When variables are accessed from within Python models / native macros, either the `context.var` or `evaluator.var` methods must be used, with the `context` / `evaluator` reference names preserved. Using different reference names breaks the variable detection logic, thus preventing the relevant values from being captured into snapshots.

The `@gateway` macro has been refactored to use variables to benefit from the described static analysis.

@treysp please review the docs. Please also note that a dedicated guide will likely be needed for this in future.

### DIFF
--- a/docs/concepts/macros/macro_variables.md
+++ b/docs/concepts/macros/macro_variables.md
@@ -100,3 +100,22 @@ Other macro variables:
     * 'creating' - The model tables are being created.
     * 'evaluating' - The models' logic is being evaluated.
 * @gateway - A string value that represents the name of the selected [gateway](../../guides/connections.md).
+
+## User-defined Variables
+
+Users can also pass custom variables in [configuration](../../reference/configuration.md#variables). To access variable values the `@var` macro should be used in model definitions. For example:
+
+```sql linenums="1"
+SELECT *
+FROM table
+WHERE some_value = @var('<var_name>')
+```
+
+The default value can be passed as a second argument to the `@var` macro, which will be used as a fallback when the variable with the given name is missing:
+```sql linenums="1"
+SELECT *
+FROM table
+WHERE some_value = @var('<var_name>', <default_value>)
+```
+
+The similar API is available for [Python models](../models/python_models.md) via the `context.var` method, as well as in [macros](./sqlmesh_macros.md) via `evaluator.var`.
--- a/docs/concepts/macros/macro_variables.md
+++ b/docs/concepts/macros/macro_variables.md
@@ -100,22 +100,3 @@ Other macro variables:
     * 'creating' - The model tables are being created.
     * 'evaluating' - The models' logic is being evaluated.
 * @gateway - A string value that represents the name of the selected [gateway](../../guides/connections.md).
-
-## User-defined Variables
-
-Users can also pass custom variables in [configuration](../../reference/configuration.md#variables). To access variable values the `@var` macro should be used in model definitions. For example:
-
-```sql linenums="1"
-SELECT *
-FROM table
-WHERE some_value = @var('<var_name>')
-```
-
-The default value can be passed as a second argument to the `@var` macro, which will be used as a fallback when the variable with the given name is missing:
-```sql linenums="1"
-SELECT *
-FROM table
-WHERE some_value = @var('<var_name>', <default_value>)
-```
-
-The similar API is available for [Python models](../models/python_models.md) via the `context.var` method, as well as in [macros](./sqlmesh_macros.md) via `evaluator.var`.
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -1061,6 +1061,22 @@ def some_macro(evaluator):
     ...
 ```
 
+### Accessing variables
+
+[User-defined variables](../../reference/configuration.md#variables) can be accessed within the macro's body using the `evaluator.var` method. For example:
+
+```python linenums="1"
+from sqlmesh.core.macros import macro
+
+@macro()
+def some_macro(evaluator):
+    var_value = evaluator.var("<var_name>")
+    another_var_value = evaluator.var("<another_var_name>", "default_value")
+    ...
+```
+
+
+
 ## Mixing macro systems
 
 SQLMesh supports both SQLMesh and [Jinja](./jinja_macros.md) macro systems. We strongly recommend using only one system in a model - if both are present, they may fail or behave in unintuitive ways.
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -29,7 +29,7 @@ It uses the following five step approach to accomplish this:
 2. Examine the placeholder values to classify them as one of the following types:
 
     - Creation of user-defined macro variables with the `@DEF` operator (see more about [user-defined macro variables](#user-defined-variables))
-    - Macro variables, both [SQLMesh pre-defined](./macro_variables.md) and [user-defined](#User-defined-variables)
+    - Macro variables: [SQLMesh pre-defined](./macro_variables.md), [user-defined local](#local-variables), and [user-defined global](#global-variables)
     - Macro functions, both [SQLMesh's](#sqlmesh-macro-operators) and [user-defined](#user-defined-macro-functions)
 
 3. Substitute macro variable values where they are detected. In most cases, this is direct string substitution as with a templating system.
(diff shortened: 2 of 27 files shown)


docs/concepts/macros/macro_variables.md
```diff
@@ -100,3 +100,22 @@ Other macro variables:
...
 * @gateway - A string value that represents the name of the selected [gateway](../../guides/connections.md).
+
+## User-defined Variables
+
+Users can also pass custom variables in [configuration](../../reference/configuration.md#variables). To access variable values the `@var` macro should be used in model definitions. For example:
+
+```sql linenums="1"
+SELECT *
+FROM table
+WHERE some_value = @var('<var_name>')
```
tobymao, line 111: why use `@var` and not just the variable name?

docs/concepts/models/python_models.md
```diff
@@ -143,6 +143,28 @@ def execute(
...
+```python linenums="1"
+@model(
+    "my_model.name",
+)
+def execute(
+    context: ExecutionContext,
+    start: datetime,
+    end: datetime,
+    execution_time: datetime,
+    **kwargs: t.Any,
```
tobymao, line 160: the purpose of kwargs was to handle vars

sqlmesh/core/audit/definition.py
```diff
@@ -352,7 +354,7 @@ def metadata_hash(self, audits: t.Dict[str, ModelAudit]) -> str:
         ]
 
         query = self.render_query(self) or self.query
-        data.append(query.sql(comments=False))
+        data.append(query.sql(dialect=self.dialect, comments=False))
```
tobymao, line 357: this is a breaking change if you do this so need to switch to feat!

maybe should switch this to gen?

sqlmesh/core/macros.py
```diff
@@ -866,6 +876,17 @@ def or_(evaluator: MacroEvaluator, *expressions: t.Optional[exp.Expression]) ->
     return exp.or_(*conditions, dialect=evaluator.dialect)
 
 
+@macro("VAR")
+def var(
+    evaluator: MacroEvaluator, var_name: exp.Expression, default: exp.Expression = exp.null()
```
tobymao, line 881: default should be optional null, python doesn't allow defaults types here because they are mutable and singletons

default = None
default = exp.null() if default is None

sqlmesh/core/model/definition.py
```diff
@@ -1441,14 +1449,15 @@ def load_sql_based_model(
...
-    jinja_macro_references: t.Set[MacroReference] = {
-        r
-        for references in [
-            *[extract_macro_references(e.sql()) for e in pre_statements],
-            *[extract_macro_references(e.sql()) for e in post_statements],
-        ]
-        for r in references
-    }
+    jinja_macro_references, used_variables = extract_macro_references_and_variables(
+        *(e.sql(dialect=dialect) for e in pre_statements),
```
tobymao, line 1453: gen

docs/concepts/macros/macro_variables.md
```diff
@@ -100,3 +100,22 @@ Other macro variables:
     * 'creating' - The model tables are being created.
     * 'evaluating' - The models' logic is being evaluated.
 * @gateway - A string value that represents the name of the selected [gateway](../../guides/connections.md).
+
+## User-defined Variables
```
treysp, line 104: Since we already have "user-defined macro variables," I think it would help people differentiate if we had another descriptor for these. 

How about "user-defined project variables" since they are a top-level key in the project config?

georgesittas: Maybe "parameters" is a better term here? Wdyt?

treysp: If we're going to use the `@var()` macro name I think it would be good to use "variables" so the connection is obvious. I'm guessing the term "vars" was chosen because that's what dbt uses? 

If we were just going for clarity, I think `@project_param()` or `@project_var()` (with yaml key of same name) would both be  good.

docs/concepts/macros/macro_variables.md
izeigerman: More importantly though, having a function allows you to specify a default value in the second argument.

docs/concepts/models/python_models.md
izeigerman: It doesn't work though cause it'd impossible to statically detect whether variable is used or not.

sqlmesh/core/macros.py
izeigerman: I can see us doing exactly this here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/macros.py#L822-L823. What's the difference?

docs/concepts/macros/macro_variables.md
izeigerman: Not a fan of `@project_var`. Too verbose.

georgesittas: Makes sense

github-christophe-oudar:
Hello, it looks like a great feature 👍 

I think it could be helpful to clarify few things regarding that feature:
- @DEF is meant to be "model local variable" while @VAR would be "global context variable" right? Then maybe it could be helpful to update and clarify the concepts in https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#user-defined-variables
- How @VAR are being defined? Through config.yml / config.py? I don't see related documentation in the PR
- Then you would like to set a precedence system with the CLI (to be implemented in a future PR)?

One major pain point in dbt (at least for me) with variables is that any variables change trigger a recompilation of the models as the variables can change the DAG (for instance, `if (var1 == true) FROM my_schema.table1 else FROM my_schema.my_table2`). Then it means that we can't "precompile" the project for production and recompiling it on dbt can take minutes on big projects (because it fetches the schema/tables from the engine). I'm not sure if SQLMesh is affected by the same problem but on runtime variables (such as start/end/runtime dates), you wouldn't change that DAG (right?) and then you could expect the model to be "identical" (and skip any kind of computing).

docs/concepts/models/python_models.md
izeigerman: I think we can do this if check which variables user specified explicitly in the list of arguments. However, I can see several issues with that:

1. A user might expect variables to be available through `kwargs` by key, but that won't be the case.
2. A user has to make sure to provide a default value for the argument, otherwise it'd fail at runtime when a variable is missing.
3. The interface looks quite different from what we do in macros.

So there's more implicit magic involved imho. Wdyt?

izeigerman:
> How @VAR are being defined? Through config.yml / config.py? I don't see related documentation in the PR

There's a documentation update to the `reference/configuration.md` file.

>  I'm not sure if SQLMesh is affected by the same problem but on runtime variables (such as start/end/runtime dates), you wouldn't change that DAG (right?) and then you could expect the model to be "identical" (and skip any kind of computing).

These variables are not runtime variables. They are set and versioned for each model individually at plan time. Changing a variable will produce new versions for models that use the affected variable.

>  with variables is that any variables change trigger a recompilation of the models as the variables can change the DAG

Yes, that would be the case here as well to an extent. We'd still benefit from caching, but for example, rendered (compiled) queries will need to be rebuilt.

github-christophe-oudar:
> There's a documentation update to the reference/configuration.md file.

Alright, I watched the whole configuration page and it's clearer thanks 👍
I'm likely biased but I feel like the template as in https://docs.getdbt.com/reference/dbt_project.yml is easier to understand what's available at which level.

> These variables are not runtime variables. They are set and versioned for each model individually at plan time. Changing a variable will produce new versions for models that use the affected variable.

Thanks that's what I thought it was too 👍 
There are other use cases for actual run time variables but it's likely a "niche" (adhoc templated reports mostly) compared to global variables usage so I think it shouldn't be a priority for you anyway (especially considering you have already start/end/runtime date as runtime ones well integrated).

izeigerman:
Thanks, @github-christophe-oudar, appreciate your feedback!

izeigerman:
@github-christophe-oudar I think @treysp did a really good job addressing your comments in the other PR: https://github.com/TobikoData/sqlmesh/pull/2396

docs/concepts/macros/macro_variables.md
izeigerman: After additional conversation we decided to support both styles

github-christophe-oudar:
Right 👍 
That looks good 🚀

plaflamme:
@izeigerman This is great! We've got our own poor-man's version of gateway-scoped variables, but it requires using forward-only plans whenever we make change to the set of variables. So this feature will be welcomed addition!

One thing I'm unclear on is how someone would use this in practice. Is the idea that the same project would have 2 separate configs with different values for these variables?

In our case, we have a set of variables that changes per **gateway**. Each gateway connects to different versions of the same "warehouse": dev, staging and production. The data is different, but everything is expected to be the same. The main reason why we need variables is because the models have to select from external tables which need to be fully qualified and thus are different between each environment, i.e.: `select * from gcp-<dev|stg|prod>-project.dataset.table`.

So ideally for our use-case, `variables` would be defined on the gateways. In fact, when you think about it, the existing `@gateway` variable is such a gateway-scoped variable already: its value is different per gateway, not per project.

Have you considered allowing "Gateway Variables"? So something like `Project Variables < Gateway Variables < Local Variables`.

izeigerman:
@plaflamme 

> So ideally for our use-case, variables would be defined on the gateways. In fact, when you think about it, the existing @gateway variable is such a gateway-scoped variable already: its value is different per gateway, not per project.

That's an excellent point! Yes, we're considering gateway variables. So far the only challenge I'm seeing is what should we do in the multi-repo case when different repos might not have gateways configured / named the same way. How do select a correct set of variables for each individual subrepo? Other than that it should be doable.

plaflamme:
> what should we do in the multi-repo case when different repos might not have gateways configured / named the same way

That sounds like it would already be a problem, regardless of variables? i.e.: what happens in this situation

```
sqlmesh --gateway=prod -p project/a -p project/b plan
```

and where `project/a` and `project/b` don't agree on a `prod` gateway?

izeigerman:
@plaflamme good point! In your example we'll just pick the gateway from the first project specified in the arguments.

sqlmesh/core/macros.py
tobymao: that's a bug

z3z1ma:
This is sweet 🔥

izeigerman:
@plaflamme I've settled on emitting a warning if a gateway with the given name doesn't exist in a project that is part of the multi-repo setup. Otherwise, the gateway variables will be picked up accordingly.

plaflamme:
@izeigerman this is great, exactly what we needed, thank you!

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -41,7 +41,74 @@ It uses the following five step approach to accomplish this:
...
+            my_var: 1
+        ...
+```
+
+A model definition could access the `int_var` value in a `WHERE` clause like this:
+
+```sql linenums="1"
+SELECT *
+FROM table
+WHERE int_variable = @INT_VAR
```
tobymao, line 86: do we note anywhere in the docs that vars are always case insensitive? if not i think we should

sqlmesh/core/macros.py
```diff
@@ -171,13 +172,14 @@ def evaluate_macros(
 
             if isinstance(node, MacroVar):
                 changed = True
-                if node.name not in self.locals:
+                variables = self.locals.get(c.SQLMESH_VARS, {})
+                if node.name not in self.locals and node.name.lower() not in variables:
```
tobymao, line 176: why is lower() only called on the second one and not the first one?

sqlmesh/core/macros.py
```diff
@@ -375,6 +377,15 @@ def engine_adapter(self) -> EngineAdapter:
...
         return self.locals["engine_adapter"]
 
+    @property
+    def gateway(self) -> t.Optional[str]:
+        """Returns the gateway name."""
+        return self.var(c.GATEWAY)
+
+    def var(self, var_name: str, default: t.Optional[t.Any] = None) -> t.Optional[t.Any]:
+        """Returns the value of the specified variable, or the default value if it doesn't exist."""
+        return (self.locals.get(c.SQLMESH_VARS) or {}).get(var_name, default)
```
tobymao, line 387: maybe this method should always normalize input? var_name.lower() so it's consistent

tests/core/test_model.py
```diff
@@ -3407,3 +3409,276 @@ def test_end_no_start():
...
+        return exp.convert(evaluator.var("test_var_d") + 10)
+
+    expressions = parse(
+        """
+        MODEL(
+            name test_model,
+            kind FULL,
+        );
+
+        SELECT @VAR('test_var_a') AS a, @VAR('test_var_b', 'default_value') AS b, @VAR('test_var_c') AS c, @TEST_MACRO_VAR() AS d;
```
tobymao, line 3426: let's add an uppercase test case, @var('TEST_VAR_a')

tobymao (approved):

sqlmesh/core/macros.py
izeigerman: Seems like we rely on case sensitivity in locals. Some tests failed when I tried that. Investigating this is out of scope of this update.

tobymao: ok

PR #2396, opened 2024-04-04
Docs: expand global variables docs

treysp:


izeigerman (approved):
So much better than what I originally wrote. Really appreciate it!

PR #2400, opened 2024-04-05
fix: allow macros to recursively resolve closes #2399

### LINKED ISSUE #2399: Conditions in the Built-in Audits dont support macro vars
Slack context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1712179233739539

Built-in audits (such as `unique_values`) have a `condition`  property that can be used to pass extra conditions like so:

```
@MODEL (
   ...
  audits (
    unique_values(columns=(id), condition=foo='bar')
  )
)
```

However, if you want to use `@start_dt` and `@end_dt` in the `condition` to limit the amount of data scanned on the underlying engine, they dont expand out correctly.

For example, this `condition` fails:
```
condition=ts between @start_dt and @end_dt

-- ValueError: Unsupported expression type MacroVar
```

You can wrap it in `@SQL` to get further:
```
condition=@SQL('ts between @start_dt and @end_dt')

-- sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 38.
  ts between 2023-03-01 00:00:00+00:00 and 2023-03-10 23:59:59.999999+00:00
Error: Failed to resolve macro for expression. Error trying to eval macro.

Generated code: self.parse_one(self.template('ts between @start_dt and @end_dt', locals()))

Original sql: @SQL('ts between @start_dt and @end_dt') at '.'
```

It seems that it doesn't take the current dialect into account and just does a string substitution. That same expression in a normal model would correctly expand to `CAST('@start_dt' AS TIMESTAMP) AND CAST('@end_dt' AS TIMESTAMP)`

You can work around the dialect issue by writing SQL to convert the string values to timestamps:

```
condition=@SQL('ts between cast(''@start_dt'' as timestamp) and cast(''@end_dt'' as timestamp)')
```

However, that's a bit hacky. It would be good if the rules for the model SQL dialect were applied

tobymao:
macro variables were not resolving additional macro references. also, we ensure conversion of sql to the right dialect when using string templates like `@sql`

cc @erindru

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import typing as t
 from enum import Enum
 from functools import reduce
+from itertools import chain
 from string import Template
 
 import sqlglot
@@ -177,7 +178,15 @@ class MacroEvaluator:
                     return node
 
                 value = self.locals[node.name]
-                return exp.convert(tuple(value) if isinstance(value, list) else value)
+                if isinstance(value, list):
+                    return exp.convert(
+                        tuple(
+                            self.transform(v) if isinstance(v, exp.Expression) else v for v in value
+                        )
+                    )
+                return exp.convert(
+                    self.transform(value) if isinstance(value, exp.Expression) else value
+                )
             if isinstance(node, exp.Identifier):
                 text = self.template(node.this, self.locals)
                 if node.this != text:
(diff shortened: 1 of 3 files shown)


tobymao:
@erindru can you test that this works for your use case?

sqlmesh/core/macros.py
```diff
@@ -222,7 +231,20 @@ def template(self, text: t.Any, local_variables: t.Dict[str, t.Any]) -> str:
         Returns:
            The rendered string.
         """
-        return MacroStrTemplate(str(text)).safe_substitute(self.locals, **local_variables)
+        mapping = {}
+
+        for k, v in chain(self.locals.items(), local_variables.items()):
+            if not isinstance(v, str):
+                try:
+                    v = exp.convert(v)
```
izeigerman, line 239: I'm not sure I follow what's going on in this block. An explanatory comment would be appreciated. Like why do we convert but ignore the failure? Is there a better way to know whether we should convert or not?

tobymao: i’d have to enumerate every type. additionally there’s a recursive error due to magic mock

georgesittas: +1 about using a comment here

georgesittas (approved):

PR #2404, opened 2024-04-07
Feat: add flag for preserving unit test fixtures to improve debuggability

georgesittas:
Addresses the tenth item in https://github.com/TobikoData/sqlmesh/issues/1637. I didn't address the CTE fixture item because creating views for CTEs may not be possible, e.g. if we don't have or can't correctly infer their `columns_to_types` mappings.

> Add support for "dumping" duckdb state on test failure to speed up debugging

As I mentioned in a reply in the above issue, I believe this PR should suffice because one can simply override the test connection database to "dump" the state for DuckDB, so the fixtures are not transient (due to creating them in-memory).

For example:

```yaml
gateways:
    local:
        connection:
            type: duckdb
            database: db.db
        test_connection:
            type: duckdb
            catalogs:
              db: test.db

default_gateway: local

model_defaults:
    dialect: duckdb
```

I'm using this config for the example project, and I can jump into `test.db` directly after running:

```
sqlmesh test --preserve-fixtures
```

Observe:

```
➜ duckdb test.db
v0.10.0 20b1486d11
Enter ".help" for usage hints.
D .schema
CREATE VIEW sqlmesh_example.incremental_model__fixture__mzn485dz (id, item_id, event_date) AS SELECT CAST(id AS INTEGER) AS id, CAST(item_id AS INTEGER) AS item_id, CAST(event_date AS DATE) AS event_date FROM (SELECT * FROM (VALUES (1, 1, '2020-01-01'), (2, 1, '2020-01-02'), (3, 2, '2020-01-03')) AS valueslist) AS t(id, item_id, event_date);

D select * from sqlmesh_example.incremental_model__fixture__mzn485dz;
┌───────┬─────────┬────────────┐
│  id   │ item_id │ event_date │
│ int32 │  int32  │    date    │
├───────┼─────────┼────────────┤
│     1 │       1 │ 2020-01-01 │
│     2 │       1 │ 2020-01-02 │
│     3 │       2 │ 2020-01-03 │
└───────┴─────────┴────────────┘
```

### DIFF
--- a/sqlmesh/magics.py
+++ b/sqlmesh/magics.py
@@ -783,9 +783,9 @@ class SQLMeshMagics(Magics):
     )
     @argument("--verbose", "-v", action="store_true", help="Verbose output.")
     @argument(
-        "--persist-fixtures",
+        "--preserve-fixtures",
         action="store_true",
-        help="Persist the fixture tables in the testing database, useful for debugging.",
+        help="Preserve the fixture tables in the testing database, useful for debugging.",
     )
     @line_magic
     @pass_sqlmesh_context
@@ -796,7 +796,7 @@ class SQLMeshMagics(Magics):
             match_patterns=args.pattern,
             tests=args.tests,
             verbose=args.verbose,
-            persist_fixtures=args.persist_fixtures,
+            preserve_fixtures=args.preserve_fixtures,
         )
 
     @magic_arguments()
--- a/sqlmesh/magics.py
+++ b/sqlmesh/magics.py
@@ -782,12 +782,22 @@ class SQLMeshMagics(Magics):
         help="Only run tests that match the pattern of substring.",
     )
     @argument("--verbose", "-v", action="store_true", help="Verbose output.")
+    @argument(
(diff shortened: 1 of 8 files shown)


sqlmesh/magics.py
```diff
@@ -782,12 +782,22 @@ def create_test(self, context: Context, line: str) -> None:
         help="Only run tests that match the pattern of substring.",
     )
     @argument("--verbose", "-v", action="store_true", help="Verbose output.")
+    @argument(
+        "--persist-fixtures",
```
izeigerman, line 786: "Persist" seems a little misleading. The fixtures are persisted regardless of whether the flag is provided. It's whether they are cleaned up is the question. Should this be `preserve` or something indicated the lack of cleanup.

For context, I initially thought that this command dumps the fixtures to some other place.

georgesittas: Ah, I can see the confusion, thanks for letting me know. I think `--preserve-fixtures` looks good.

izeigerman (approved):

PR #2406, opened 2024-04-08
Chore: update mssql test from bytes to timestamp return value

treysp:
pymssql fixed a bug and now correctly converts `DATETIMEOFFSET` type to Python datetime

### DIFF
--- a/tests/core/engine_adapter/test_integration.py
+++ b/tests/core/engine_adapter/test_integration.py
@@ -2035,8 +2035,6 @@ def test_dialects(ctx: TestContext):
             None,
             {
                 "default": pd.Timestamp("2020-01-01 00:00:00+00:00"),
-                # https://github.com/pymssql/pymssql/issues/649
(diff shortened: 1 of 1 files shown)


tests/core/engine_adapter/test_integration.py
```diff
@@ -2035,8 +2035,6 @@ def test_dialects(ctx: TestContext):
             None,
             {
                 "default": pd.Timestamp("2020-01-01 00:00:00+00:00"),
-                # https://github.com/pymssql/pymssql/issues/649
```
tobymao, line 2038: doesn’t this depend on the version? if you’re gonna remove this you may have to pin the version, if not you’ll need backwards compatibility right?

treysp: To my knowledge, no sqlmesh behavior will change - just the test. Our plan was to have people handle the bytes -> python conversion themselves if it ever came up.

tobymao: ah ok great

tobymao (approved):

eakmanrq (approved):

PR #2409, opened 2024-04-08
fix: widen spark support

eakmanrq:
I've updated Spark's catalog operations to not rely solely on 3.4+. In general after testing it seems like Spark may only support a single catalog for managing SQLMesh models. 

For `set_catalog` override I tested and I couldn't find any reason why `create_schema` and `drop_schema` were to requiring set catalog to work and it seems like on both Spark and Databricks the full catalog.db reference is compatible. So I removed this entirely. 

Databricks does require the set catalog when getting objects outside of the default catalog. 
Screenshot 2024-04-08 at 12 46 11 PM

For Spark if I try to do this within my local environment I get an error about not being able to do this on v2 tables. So I think potentially our current implementation requires the use of `spark_catalog`?

### DIFF
--- a/sqlmesh/core/engine_adapter/spark.py
+++ b/sqlmesh/core/engine_adapter/spark.py
@@ -330,23 +330,19 @@ class SparkEngineAdapter(GetCurrentCatalogFromFunctionMixin, HiveMetastoreTableP
         ]
 
     @property
-    def _spark_semver(self) -> t.Tuple[int, int, int]:
-        return tuple(int(x) for x in self.spark.version.split(".")[:3])  # type: ignore
+    def _spark_major_minor(self) -> t.Tuple[int, int]:
+        return tuple(int(x) for x in self.spark.version.split(".")[:2])  # type: ignore
 
     def get_current_catalog(self) -> t.Optional[str]:
-        # Spark 3.4+ API
         if self._use_spark_session:
-            major, minor, _ = self._spark_semver
-            if major > 3 or (major == 3 and minor >= 4):
+            if self._spark_major_minor >= (3, 4):
                 return self.spark.catalog.currentCatalog()
             else:
                 return self._default_catalog or "spark_catalog"
         return super().get_current_catalog()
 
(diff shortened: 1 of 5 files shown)


tobymao (approved):

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -334,39 +329,41 @@ def _get_data_objects(
             for row in results  # type: ignore
         ]
 
+    @property
+    def _spark_semver(self) -> t.Tuple[int, int, int]:
```
izeigerman, line 333: [Nit] Doesn't look like we care about the patch version anywhere.

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -334,39 +329,41 @@ def _get_data_objects(
...
+            else:
+                # We just assume it is the default catalog
+                return "spark_catalog"
         return super().get_current_catalog()
 
     def set_current_catalog(self, catalog_name: str) -> None:
         # Spark 3.4+ API
-        self.spark.catalog.setCurrentCatalog(catalog_name)
+        major, minor, _ = self._spark_semver
+        if major > 3 or (major == 3 and minor >= 4):
```
izeigerman, line 354: Seems like we can simplify with `(major, minor) >= (3, 4)`. Or if we get rid of the patch version we can compare the tuple directly and not unpack it.

izeigerman (approved):

sqlmesh/core/engine_adapter/spark.py
eakmanrq: nice forgot about that pattern which is used for Python version checking. Updated: https://github.com/TobikoData/sqlmesh/pull/2409/commits/2f702b731393a32a90f69012197458ff38fb6158

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Removed

PR #2415, opened 2024-04-09
Fix: Avoid forcing quoted identifiers when creating views in trino and spark

izeigerman:
To ensure interoperability between Spark and Trino when both technologies are used to access the same tables and views, we must avoid including dialect-specific quotes whenever possible when creating views.


georgesittas (approved):

sqlmesh/core/renderer.py
```diff
@@ -293,6 +295,14 @@ def _cache_key(
             runtime_stage,
         )
 
+    @contextmanager
+    def _normalize_and_quote(self, query: E) -> t.Iterator[E]:
```
tobymao, line 299: why don't we use the method in dialect now? add the quote_identifiers flag to that one and use it?

izeigerman: That's what I did initially, but then I noticed that we only care about not quoting in this one case and I concluded that the difference in logic is sufficient enough to just re-implement it here. Do you suggest to share the logic anyway?

tobymao: yea, i think so

izeigerman: fixed

tobymao (approved):

PR #2417, opened 2024-04-09
Fix: get physical table name in ibis example

andrew-sha:
Previously, applying a plan in this project from an empty state would fail since the table `ibis.incremental_model` would not yet exist. This gets the physical table name from context and passes that into the ibis connection instead

### DIFF
--- a/examples/ibis/models/ibis_full_model_python.py
+++ b/examples/ibis/models/ibis_full_model_python.py
@@ -17,7 +17,7 @@ from sqlmesh.core.model import FullKind
         "num_orders": "int",
     },
     audits=["assert_positive_order_ids"],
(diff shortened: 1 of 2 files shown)


examples/ibis/models/ibis_full_model_python.py
```diff
@@ -26,11 +26,14 @@ def execute(
     execution_time: datetime,
     **kwargs: t.Any,
 ) -> pd.DataFrame:
+    # get physical table name
+    table_name = context.table("ibis.incremental_model").split(".")
```
tobymao, line 30: table = exp.to_table(context.table("ibis.incremental_model").split("."))

tobymao (approved):

PR #2423, opened 2024-04-10
Feat: Support restatements for the incremental unmanaged model kind

izeigerman:
This update adds ability to restate model kinds for which restatements weren't supported in the past: incremental by unique key, unmanaged incremental (dbt incremental), and SCD Type 2.

When restating models of the aforementioned types, the entire model history will be reprocessed from scratch, making this operation equivalent to the full-refresh command in dbt. Start/end dates specified by a user in the plan will be ignored for such models.

For dbt projects we now rely on the [full_refresh ](https://docs.getdbt.com/reference/resource-configs/full_refresh) model configuration to determine whether the restatement should be disabled or not.

### DIFF
--- a/docs/concepts/plans.md
+++ b/docs/concepts/plans.md
@@ -133,7 +133,7 @@ For this reason, the `plan` command supports the `--restate-model`, which allows
 
 Application of a plan will trigger a cascading backfill for all specified models (other than external tables), as well as all models downstream from them. The plan's date range determines the data intervals that will be affected.
 
-Please note that models of kinds [INCREMENTAL_BY_UNIQUE_KEY](models/model_kinds.md#INCREMENTAL_BY_UNIQUE_KEY), [SCD_TYPE_2_BY_TIME](models/model_kinds.md#scd-type-2), and [SCD_TYPE_2_BY_COLUMN](models/model_kinds.md#scd-type-2) cannot be partially restated. Therefore, such models will be fully refreshed regardless of the start/end dates provided by a user in the plan.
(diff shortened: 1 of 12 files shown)


docs/concepts/plans.md
```diff
@@ -133,6 +133,8 @@ For this reason, the `plan` command supports the `--restate-model`, which allows
 
 Application of a plan will trigger a cascading backfill for all specified models (other than external tables), as well as all models downstream from them. The plan's date range determines the data intervals that will be affected.
 
+Please note that models of kinds [INCREMENTAL_BY_UNIQUE_KEY](models/model_kinds.md#INCREMENTAL_BY_UNIQUE_KEY), [SCD_TYPE_2_BY_TIME](models/model_kinds.md#scd-type-2), and [SCD_TYPE_2_BY_COLUMN](models/model_kinds.md#scd-type-2) cannot be partially restated. Therefore, such models will be fully refreshed regardless of the start/end dates provided by a user in the plan.
+
```
eakmanrq, line 137: Should each model also get a note stating this information?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1485,3 +1504,11 @@ def delete(self, name: str) -> None:
 
     def _is_materialized_view(self, model: Model) -> bool:
         return isinstance(model.kind, ViewKind) and model.kind.materialized
+
+
+def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> Intervals:
+    return (
+        snapshot.intervals
+        if deployability_index.is_deployable(snapshot)
+        else snapshot.dev_intervals
```
tobymao, line 1513: are there any edge cases where since we've never run in dev, intervals is empty so we do the wrong thing?

tobymao (approved):

sqlmesh/core/snapshot/evaluator.py
izeigerman: What would the wrong thing be? If it never ran in dev then we'd just recreate the dev table.

docs/concepts/plans.md
izeigerman: added

PR #2429, opened 2024-04-12
Feat: allow partial schema inference in unit test

georgesittas:
Prior to this change, if there was at least one unknown type in the schema of an input model we would try to infer all of its column types again using a combination of `convert` and `annotate_types`, thus ignoring the types that we'd previously correctly inferred.

However, relying solely on the `convert` and `annotate_types` approach is problematic, because the former is dialect/type-agnostic and thus it would always convert dictionaries into `MAP`s, which led to type mismatch errors (eg. when we would expect `STRUCT`s instead).

This PR ensures we leverage the inferred types, and only fall back to the above strategy for the columns that have unknown types. It also enforces that we'll only create an input fixture if its inferred schema contains known types.

This will open a path for users to fix invalid conversion issues by simply casting their nested objects into the appropriate types.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -18,7 +18,7 @@ from sqlmesh.core import constants as c
 from sqlmesh.core.dialect import normalize_model_name, schema_
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.model import Model, PythonModel, SqlModel
-from sqlmesh.utils import UniqueKeyDict, random_id, yaml
+from sqlmesh.utils import UniqueKeyDict, random_id, type_is_known, yaml
 from sqlmesh.utils.date import pandas_timestamp_to_pydatetime
 from sqlmesh.utils.errors import ConfigError, TestError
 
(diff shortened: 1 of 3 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -101,18 +101,30 @@ def shortDescription(self) -> t.Optional[str]:
...
         for table_name, values in self.body.get("inputs", {}).items():
+            all_types_are_known = False
             columns_to_types: t.Dict[str, exp.DataType] = {}
             if table_name in self.models:
                 model = self.models[table_name]
-                columns_to_types = model.columns_to_types if model.annotated else {}
+                columns_to_types = {
+                    c: t for c, t in model.columns_to_types.items() if type_is_known(t)
+                }
+                all_types_are_known = len(columns_to_types) == len(model.columns_to_types)
```
izeigerman, line 111: I thought `model.columns_to_types` could be None. How come we don't need to account for this here?

sqlmesh/core/test/definition.py
```diff
@@ -101,18 +101,30 @@ def shortDescription(self) -> t.Optional[str]:
     def setUp(self) -> None:
         """Load all input tables"""
         for table_name, values in self.body.get("inputs", {}).items():
+            all_types_are_known = False
             columns_to_types: t.Dict[str, exp.DataType] = {}
             if table_name in self.models:
                 model = self.models[table_name]
-                columns_to_types = model.columns_to_types if model.annotated else {}
+                columns_to_types = {
```
izeigerman, line 108: Maybe clarify that this is `known` columns_to_types.

sqlmesh/core/test/definition.py
georgesittas: Good catch, this was an oversight. Will amend.

sqlmesh/core/test/definition.py
georgesittas: Done

sqlmesh/core/test/definition.py
```diff
@@ -94,29 +94,46 @@ def shortDescription(self) -> t.Optional[str]:
...
-            columns_to_types = {
-                **(model.columns_to_types if model and model.annotated else {}),
-                **values.get("columns", {}),
-            }
+            if model:
+                inferred_columns_to_types = model.columns_to_types or {}
+                known_columns_to_types = {
+                    c: t for c, t in inferred_columns_to_types.items() if type_is_known(t)
+                }
+                all_types_are_known = len(known_columns_to_types) == len(inferred_columns_to_types)
```
izeigerman, line 106: This doesn't look right. If we couldn't infer the columns, the `model.columns_to_types` returns None and the length will be 0. Wouldn't we incorrectly conclude that all types are known then?

sqlmesh/core/test/definition.py
```diff
@@ -94,29 +94,46 @@ def shortDescription(self) -> t.Optional[str]:
...
+            known_columns_to_types.update(values.get("columns", {}))
 
             rows = values["rows"]
-            if not columns_to_types and rows:
-                for i, v in rows[0].items():
-                    v_type = annotate_types(exp.convert(v)).type or type(v).__name__
-                    columns_to_types[i] = exp.maybe_parse(
-                        v_type, into=exp.DataType, dialect=self.dialect
-                    )
+            if (not all_types_are_known or not known_columns_to_types) and rows:
```
izeigerman, line 112: I guess we check this here but now the condition looks contradictory. Like all types are known but there are no known types.

sqlmesh/core/test/definition.py
izeigerman:
```
all_types_are_known = inferred_columns_to_types  and len(known_columns_to_types) == len(inferred_columns_to_types)
```

izeigerman (approved):
Confusing condition, LGTM otherwise.

sqlmesh/core/test/definition.py
georgesittas: Right, I was skeptical whether to do it like this or to `and` them like you suggested above. I'll apply your suggestion.

PR #2433, opened 2024-04-12
Fix!: Add the missing migration script for the gateway variable

izeigerman:
This migration script should've been a part of #2391

### DIFF
--- a/sqlmesh/migrations/v0045_move_gateway_variable.py
+++ b/sqlmesh/migrations/v0045_move_gateway_variable.py
@@ -1,0 +1,67 @@
+"""Move the gateway variable."""
+
+import ast
+import json
+
+import pandas as pd
+from sqlglot import exp
+
+from sqlmesh.utils.migration import index_text_type
+
+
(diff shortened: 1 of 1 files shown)


sqlmesh/migrations/v0045_move_gateway_variable.py
```diff
@@ -0,0 +1,67 @@
...
+    new_snapshots = []
+
+    for name, identifier, version, snapshot, kind_name, expiration_ts in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name", "expiration_ts").from_(
+            snapshots_table
+        ),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        python_env = parsed_snapshot["node"].get("python_env", {})
```
georgesittas, line 29: Nit: it seems like we can get rid of the default `{}` value here?

georgesittas (approved):

tobymao (approved):

PR #2442, opened 2024-04-15
docs: column name consistency

dbrtly:
This model does not match the model created by `sqlmesh init bigquery`. Looks like accidentally missed in an update.


georgesittas (approved):
Thanks!

Issue #2441, opened 2024-04-15
'Self-referencing models require inferrable column types' error when using macros to list columns


erindru:
Given the basic example project from `sqlmesh create duckdb`, modifying the default `incremental_model.sql` model to reference itself like so:

```
MODEL (
    name sqlmesh_example.incremental_model,
    kind INCREMENTAL_BY_TIME_RANGE (
        time_column event_date
    ),
    start '2020-01-01',
    cron '@daily',
    grain (id, event_date),
    depends_on [sqlmesh_example.incremental_model]
);

SELECT
    id,
    item_id,
    event_date,
FROM
    sqlmesh_example.seed_model
WHERE
    event_date BETWEEN @start_date AND @end_date
```

Produces no error, `sqlmesh plan` can successfully complete. This shows that SQLMesh can infer the column types no problem.

However, changing the query to produce the field list via a macro:

```
@DEF(fields, [
    id,
    item_id,
    event_date,
]);

SELECT
    @each(fields, f -> f)
FROM
    sqlmesh_example.seed_model
WHERE
    event_date BETWEEN @start_date AND @end_date
```

Causes `sqlmesh plan` to throw the following error:

```
Error: Self-referencing models require inferrable column types.
```

It appears to lose the column types if the column names are inserted via a macro

z3z1ma:
I am pretty sure you can also repro with an SCD by unique key model without all columns specified.

tobymao:
https://github.com/TobikoData/sqlmesh/pull/2451

PR #2445, opened 2024-04-15
fix: temporarily remove validation on annotated depends on to fix issues

tobymao:


tobymao:
closes #2441

izeigerman (approved):
Maybe add a FIXME for why we do this

tobymao:
> Maybe add a FIXME for why we do this

i'll fix it soonish so no need

PR #2450, opened 2024-04-15
Feat!: Add the ability to control concurrency between model batches during the evaluation

### LINKED ISSUE #2384: Add support for specifying maximum concurrency per model
BigQuery only supports 2 concurrent DML queries against a table (see [here](https://cloud.google.com/bigquery/docs/data-manipulation-language#update_delete_merge_dml_concurrency)).

Within sqlmesh, you can specify `concurrent_tasks` ([here](https://sqlmesh.readthedocs.io/en/latest/guides/configuration/#concurrent-tasks)) so increase concurrency, but setting this to more than 2 means that some queries will be queued.

This isn't a problem until you consider query timeouts as well since the time a query spends being queued counts towards its timeout.

If you have a subset of your models that have particularly expensive queries, it is more or less impossible to set `concurrent_tasks` to anything higher than 2 for BigQuery since you cannot predict how many queries will be queued and thus how long they will run for (on average). Obviously, you can set the timeout to be some very high value, but that's not desirable in practice. This can severely limit concurrency in larger projects.

This feature request is to add a setting to set a maximum number of concurrent queries *per model*. Something like this:

```yaml
concurrent_tasks: 6 # maximum of 6 concurrent queries globally
concurrent_tasks_per_model: 2 # maximum of 2 concurrent queries on any individual model
```

Alternatively, don't expose this configuration property and special-case engines that have a similar limitation to BigQuery (i.e.: just always restrict concurrency-per-model to 2 in the BQ engine)

izeigerman:
Closes #2384


sqlmesh/core/model/definition.py
```diff
@@ -101,9 +100,6 @@ class _Model(ModelMeta, frozen=True):
             The start date can be a static datetime or a relative datetime like "1 year ago"
         end: The date that the model will be backfilled up until. Follows the same syntax as 'start',
             should be omitted if there is no end date.
-        batch_size: The maximum number of incremental intervals that can be run per backfill job. If this is None,
```
tobymao, line 104: why was the doc removed? duplicate?

izeigerman: `batch_size` hasn't been a top-level model attribute for a long time now.

sqlmesh/core/config/model.py
```diff
@@ -31,7 +28,6 @@ class ModelDefaultsConfig(BaseConfig):
     cron: t.Optional[str] = None
     owner: t.Optional[str] = None
     start: t.Optional[TimeLike] = None
-    batch_size: t.Optional[int] = None
```
izeigerman, line 34: Removing this value, since it breaks the project if set anyway:
```
Error: 1 validation error for SqlModel
batch_size
  Extra inputs are not permitted [type=extra_forbidden, input_value=1, input_type=int]
    For further information visit https://errors.pydantic.dev/2.5/v/extra_forbidden at '/Users/izeigerman/github/sqlmesh/examples/sushi/models/marketing.sql'
```
It assumes that the `batch_size` is the top-level model attribute.

sqlmesh/core/scheduler.py
```diff
@@ -361,15 +361,28 @@ def _dag(self, batches: SnapshotToBatches) -> DAG[SchedulingUnit]:
...
             for i, interval in enumerate(intervals):
                 node = (snapshot.name, (interval, i))
                 dag.add(node, upstream_dependencies)
 
                 if len(intervals) > 1:
                     dag.add((snapshot.name, terminal_node), [node])
 
-                if snapshot.depends_on_past and i > 0:
-                    dag.add(node, [(snapshot.name, (intervals[i - 1], i - 1))])
+                if batch_concurrency and i >= batch_concurrency:
```
tobymao, line 375: so basically if batch concurrency is 2, you will queue up two concurrent batches. but the 3rd will wait on the 1st one. and the 4th will wait for the 2nd.

i guess that makes sense

tobymao (approved):

PR #2456, opened 2024-04-16
Fix: Improve the depends_on_past definition

izeigerman:
Previously we only considered the following models to be dependent on their past:
1. `INCREMENTAL_BY_UNIQUE_KEY` models.
2. Models that reference self in their queries.

There are, however, more cases when the model depends on its past, like eg. SCD2 and `IncrementalUnmanaged` models. This update extends the definition of the `depends_on_past` property to encompass these.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -685,11 +685,7 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
         *,
         strict: bool = True,
     ) -> Interval:
-        end = (
-            execution_time or now()
-            if self.depends_on_past or self.full_history_restatement_only
-            else end
-        )
+        end = execution_time or now() if self.depends_on_past else end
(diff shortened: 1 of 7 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -1051,8 +1047,18 @@ def audit(self) -> StandaloneAudit:
 
     @property
     def depends_on_past(self) -> bool:
-        """Whether or not this models depends on past intervals to be accurate before loading following intervals."""
-        return self.is_model and self.model.depends_on_past
+        """Whether or not this models depends on past intervals to be populated before loading following intervals.
```
tobymao, line 1050: can you add the specific example about incremental by time with a self reference not fitting in full_history_restatement, that is why we need to do the OR here

izeigerman: Added an example.

tobymao (approved):

PR #2458, opened 2024-04-17
fix!: normalize scd type 2 columns

eakmanrq:
Prior to this change we would treat the column properties in SCD Type 2 model kind as strings. This would cause issues in dialects like Snowflake where we need to properly normalize (uppercase) in order for the column to be used properly. 

This change makes it so that we now consider these fields as columns and normalize/quote them. As shown by the tests, this means casing is preserved properly. 

* Breaking: Once someone starts using this version they will start storing column values in state in a different format than someone on a different version.
* Empty migration script: The validator is designed to handle columns, identifiers, or strings and so strings that are currently stored in state will be processed correctly in the migration that happens from the empty migration script.


tobymao (approved):

sqlmesh/core/model/kind.py
```diff
@@ -422,14 +424,19 @@ class FullKind(_ModelKind):
...
+    valid_from_name: SQLGlotColumn = Field(exp.column("valid_from"), validate_default=True)
+    valid_to_name: SQLGlotColumn = Field(exp.column("valid_to"), validate_default=True)
     invalidate_hard_deletes: SQLGlotBool = False
     time_data_type: exp.DataType = Field(exp.DataType.build("TIMESTAMP"), validate_default=True)
 
     forward_only: SQLGlotBool = True
     disable_restatement: SQLGlotBool = True
 
+    # Remove once Pydantic 1 is deprecated
+    _always_validate_column = field_validator(
```
izeigerman, line 436: Why didn't we need this for `SQLGlotString`?

sqlmesh/core/model/kind.py
```diff
@@ -446,9 +453,10 @@ def _time_data_type_validator(
 
     @property
     def managed_columns(self) -> t.Dict[str, exp.DataType]:
+        self.valid_from_name.meta.get("dialect")
```
izeigerman, line 456: Dead code?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1331,52 +1338,62 @@ def _scd_type_2(
...
 
         existing_rows_query = exp.select(*table_columns).from_(target_table)
         if truncate:
             existing_rows_query = existing_rows_query.limit(0)
 
         with source_queries[0] as source_query:
+            prefixed_columns_to_types = []
+            for column in columns_to_types:
+                prefixed_col = exp.column(column).copy()
+                prefixed_col.this.set("this", f"t_{prefixed_col.name}")
```
izeigerman, line 1390: Why not identifier? Here and other places

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1282,25 +1285,29 @@ def _scd_type_2(
...
+        prefixed_valid_to_col = valid_to_col.copy()
+        prefixed_valid_to_col.this.set("this", f"t_{prefixed_valid_to_col.name}")
+        prefixed_valid_from_col = valid_from_col.copy()
+        prefixed_valid_from_col.this.set("this", f"t_{valid_from_col.name}")
         if check_columns:
             row_check_conditions = []
             for col in check_columns:
                 t_col = col.copy()
-                t_col.set("this", exp.to_identifier(f"t_{col.name}"))
+                t_col.this.set("this", f"t_{col.name}")
```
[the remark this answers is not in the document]
tobymao: identifier is still there, he’s just changing the string directly instead of making a new one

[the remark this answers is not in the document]
eakmanrq: The identifier still exists. What I am doing is replacing the string found inside the identifier. This way if the identifier has quoted=True or other properties they are preserved.

sqlmesh/core/model/kind.py
eakmanrq: We need this for when we get a default value. The validator for `SQLGlotString` just ensures that we get a string result which we know will happen since the default was a string. In this case we need to make sure that the default is normalized to the dialect so we always need to validate.

sqlmesh/core/model/kind.py
eakmanrq: ah thanks for catching

sqlmesh/core/engine_adapter/base.py
izeigerman: Thanks for clarifying

izeigerman (approved):

sqlmesh/migrations/v0046_change_scd_string_to_column.py
```diff
@@ -0,0 +1,5 @@
+"""Changes the SCD Type 2 columns from strings to columns."""
```
izeigerman, line 1: Please note that the `46` version already exists in main

eakmanrq: Thanks fixed 👍

PR #2459, opened 2024-04-17
Add support for loading multiple yaml files containing external models

erindru:


sqlmesh/core/constants.py
```diff
@@ -43,6 +43,7 @@
 MACROS = "macros"
 METRICS = "metrics"
 MODELS = "models"
+EXTERNAL_MODELS = "external_models"
```
erindru, line 46: It was unclear if this should be `schemas` or `external_models`. Easy to change once its been decided

tobymao: @izeigerman i guess i prefer schemas/ but i'll let you make the final call here

sqlmesh/core/loader.py
```diff
@@ -182,9 +182,17 @@ def _load_metrics(self) -> UniqueKeyDict[str, MetricMeta]:
...
+            schema_path = Path(context_path / c.SCHEMA_YAML)
+            external_models_path = context_path / c.EXTERNAL_MODELS
 
-            if path.exists():
+            paths_to_load = []
+            if schema_path.exists():
+                paths_to_load.append(schema_path)
+
+            if external_models_path.exists() and external_models_path.is_dir():
+                paths_to_load.extend(p for p in external_models_path.glob("*.yaml"))
```
tobymao, line 193: do you need to do an iterator here? doesn't glob just return a list? extend(x.glob())

erindru: Oh, true. Have simplified it

sqlmesh/core/constants.py
izeigerman: I strongly dislike the existing `schema.yaml` name, because "schema" as a concept is so generic and overused that it says little about the purpose of the file. Furthermore, it doesn't inform user at all about the fact that only schemas of *external* tables are expected to be there. What's even worse, "schema" means something completely different in dbt.

So far the project folder naming has been very purpose-driven, and I think we should keep it that way. If the purpose of the folder is to store definitions of external models, then we should call it as such.

docs/concepts/models/external_models.md
```diff
@@ -18,7 +18,7 @@ The querying model's [`kind`](./model_kinds.md), [`cron`](./overview.md#cron), a
 
 ## Generating an external models schema file
 
-External models are defined in the `schema.yaml` file in the SQLMesh project's root folder.
+External models can be defined in the `schema.yaml` file in the SQLMesh project's root folder.
```
izeigerman, line 21: Can we also update this line: https://github.com/TobikoData/sqlmesh/blob/054f665e8e62eb7296bcab46d1db52167276177e/docs/concepts/models/external_models.md?plain=1#L41

to include something like:
```
Alternatively, additional external models can also be defined in the [external_models/](#using-the--external_models--directory) folder.
```
(Haven't tested the link)

izeigerman (approved):
Looks great! Thanks a lot for updating the docs too! One small ask to add one more note in the doc, LGTM otherwise.

sqlmesh/core/constants.py
tobymao: what if we also make create_external_models now create external_models.yaml by default, still accepting schema.yaml for now, but then we align and remove schema.yaml later?

tobymao (approved):
we can get this in for now

sqlmesh/core/constants.py
izeigerman: I like the idea

docs/concepts/models/external_models.md
erindru: Sure, have updated it

PR #2461, opened 2024-04-17
 Fix!: set default catalog for unit test fixtures, create new schema for them

georgesittas:
We currently create unit test fixtures within the specified catalog of the corresponding models, which can lead to permission issues for some external models.

This PR:
- Sets each input's catalog to the default one, for which we should have write access
- Sets each input's schema to a new testing schema, to ensure consistency and environment isolation

To avoid name clashes, we now generate fixture names using the format: `[original_catalog__][original_shema__]name`.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -364,8 +364,12 @@ class ModelTest(unittest.TestCase):
         table = self._fixture_table_cache.get(name)
         if not table:
             table = exp.to_table(name, dialect=self.dialect)
+
+            # We change both the schema and the catalog, so we need to ensure there are no name clashes
+            table.this.set("this", "__".join(part.name for part in table.parts))
             table.set("db", exp.to_identifier(self._test_schema))
             table.set("catalog", exp.to_identifier(self.default_catalog))
+
             self._fixture_table_cache[name] = table
 
         return table
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -141,8 +141,7 @@ class ModelTest(unittest.TestCase):
     def tearDown(self) -> None:
         """Drop all fixture tables."""
         if not self.preserve_fixtures:
-            for name in self.body.get("inputs", {}):
-                self.engine_adapter.drop_view(self._fixture_table_cache[name])
+            self.engine_adapter.drop_schema(self._test_schema, cascade=True)
 
     def assert_equal(
         self,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -129,10 +129,9 @@ def setUp(self) -> None:
                         known_columns_to_types[col] = v_type
 
             test_fixture_table = self._test_fixture_table(name)
-            if test_fixture_table.db:
-                self.engine_adapter.create_schema(
-                    schema_(test_fixture_table.args["db"], test_fixture_table.args.get("catalog"))
-                )
+            self.engine_adapter.create_schema(
```
izeigerman, line 132: Do we clean it up?

tests/core/test_test.py
```diff
@@ -123,21 +124,14 @@ def test_ctes(sushi_context: Context, full_model_with_two_ctes: SqlModel) -> Non
...
+        ).run()
     )
 
-    random_id = "jzngz56a"
-    test._test_id = random_id
-    _check_successful_or_raise(test.run())
-
-    assert len(test._fixture_table_cache) == len(sushi_context.models) + 1
-    for table in test._fixture_table_cache.values():
-        assert table.name.endswith(f"__fixture__{random_id}")
```
izeigerman, line 139: Do we no longer test names of created fixtures?

sqlmesh/core/test/definition.py
georgesittas: Good catch, thanks! I intended to do this but forgot.

georgesittas: I assumed `drop_schema` with `cascade=True` can take care of this - see: https://github.com/TobikoData/sqlmesh/pull/2461/commits/7c29450f61a4274a023156acddd970217b54bbf6.

sqlmesh/core/test/definition.py
```diff
@@ -337,7 +335,8 @@ def _test_fixture_table(self, name: str) -> exp.Table:
         table = self._fixture_table_cache.get(name)
         if not table:
             table = exp.to_table(name, dialect=self.dialect)
-            table.this.set("this", f"{table.this.this}__fixture__{self._test_id}")
+            table.set("db", exp.to_identifier(self._test_schema))
```
izeigerman, line 338: Don't we want to somehow encode the original schema name in the test one? Or otherwise include the original schema name in the table name? What happens if the 2 upstream tables share the same table name but are in different schemas?

georgesittas: Yes, that's a valid point. We could prepend the schema of each fixture, if any, to its table name. So for example `sushi.foo` would be accessed as `sqlmesh_test_some_id.sushi_foo`. Wdyt about this approach?

PR #2463, opened 2024-04-17
Chore: Remove the obsolete snapshot rebasing logic

izeigerman:
The existing logic has been introduced at the project's inception and was based on assumptions that didn't hold up over time. 
As a result, the current implementation is fundamentally wrong and can cause unexpected issues and behaviors. 

Some of the existing issues include:
1. Categorizing a child snapshot before categorizing its parent by inheriting the rebased version. If a parent is BREAKING, inheriting the existing version is completely wrong.
2. Not taking into account a scenario when both the local and the remote versions are `INDIRECT_BREAKING`.

Given the complexity, error-proneness, and obsoleteness of this implementation, as well as a fact that it covers a pretty niche set of use cases, we decided to drop it for now and reintroduce it later if needed.


tobymao (approved):

docs/concepts/glossary.md
```diff
@@ -76,4 +73,4 @@ SQLMesh's unique approach to environment that allows it to provide both environm
 Term used to describe a plan that can be applied without having to load any additional data or build any additional tables. See [Virtual Update](plans.md#virtual-update) for more information.
 
 ## Virtual Preview
-Term used to describe the ability to create an environment without having to build any additional tables. By comparing the version of models in the repo against what currently exists, SQLMesh can create an environment that exactly represents what is in the repo without by just updating views.
\ No newline at end of file
+Term used to describe the ability to create an environment without having to build any additional tables. By comparing the version of models in the repo against what currently exists, SQLMesh can create an environment that exactly represents what is in the repo without by just updating views.
```
vchan, line 76: I don't think you changed anything here, but there may be an existing typo.
```suggestion
Term used to describe the ability to create an environment without having to build any additional tables. By comparing the version of models in the repo against what currently exists, SQLMesh can create an environment that exactly represents what is in the repo by just updating views.
```

izeigerman: Thanks!

PR #2468, opened 2024-04-18
fix: Github action bot handle test exceptions

eakmanrq:
Prior to this PR if a test had an exception when executing, not a failure which means the tests didn't pass, the bot would incorrectly report this as "skipped" instead of a failure. Now the bot will properly categorize this as a failure and output the exception in the status summary.

Also updated another test to verify the error output upon failure.

### DIFF
--- a/tests/integrations/github/cicd/test_github_commands.py
+++ b/tests/integrations/github/cicd/test_github_commands.py
@@ -16,7 +16,7 @@ from sqlmesh.integrations.github.cicd.controller import (
     GithubCheckConclusion,
     GithubCheckStatus,
 )
-from sqlmesh.utils.errors import PlanError
+from sqlmesh.utils.errors import PlanError, TestError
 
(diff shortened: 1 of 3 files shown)


tests/integrations/github/cicd/test_github_commands.py
```diff
@@ -465,6 +466,144 @@ def test_run_all_test_failed(
...
+    assert not controller._context.invalidate_environment.called
+
+    assert len(created_comments) == 0
+
+    with open(github_output_file, "r") as f:
+        output = f.read()
+        assert (
+            output
+            == "run_unit_tests=failure\nhas_required_approval=success\npr_environment_name=hello_world_2\npr_environment_synced=skipped\nprod_plan_preview=skipped\nprod_environment_synced=skipped\n"
+        )
```
eakmanrq, line 535: I don't know why Github represents this as an add but all of this logic already exists and is unchanged. Only lines 469 - 479 were added.

tobymao (approved):

PR #2474, opened 2024-04-19
feat: there are some commands that cannot be parsed, so we fallback t…

tobymao:
…o a table macro to allow for substition

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -92,7 +92,6 @@ class BaseExpressionRenderer:
         Returns:
             The rendered expressions.
         """
-
         cache_key = self._cache_key(start, end, execution_time, runtime_stage)
 
         if cache_key not in self._cache:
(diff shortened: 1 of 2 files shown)


sqlmesh/core/renderer.py
```diff
@@ -114,6 +113,16 @@ def _render(
                 table_mapping=table_mapping,
                 deployability_index=deployability_index,
                 default_catalog=self._default_catalog,
+                table=lambda e, comments=True: self._resolve_tables(
```
izeigerman, line 116: A few comments:
1. Calling `resolve_tables` seems like an overkill. Provided we only need this line:
```
mapping = {**to_table_mapping(snapshots.values(), deployability_index), **table_mapping}
```
and normalization.

2. I don't like the name `table`. Can we be more specific? Eg. `physical_table` or `resolve_table`?

izeigerman:
Please update documentation.

sqlmesh/core/renderer.py
izeigerman: 3. Seems like our built-in macros are all over the place. Can this be a macro in `macros.py`?

izeigerman: Btw, why do we need this 2nd argument in the lambda?

PR #2479, opened 2024-04-20
feat: add this_model var to model statements

tobymao:
additionally only cache queries with no arguments since it's getting complicated caching everything

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import logging
 import typing as t
 from contextlib import contextmanager
-from datetime import datetime
 from pathlib import Path
 
 from sqlglot import exp, parse
@@ -15,7 +14,7 @@ from sqlglot.optimizer.simplify import simplify
 from sqlmesh.core import constants as c
 from sqlmesh.core import dialect as d
 from sqlmesh.core.macros import MacroEvaluator, RuntimeStage
-from sqlmesh.utils.date import TimeLike, date_dict, make_inclusive, to_datetime
+from sqlmesh.utils.date import TimeLike, date_dict, make_inclusive_end, to_datetime
 from sqlmesh.utils.errors import (
     ConfigError,
     MacroEvalError,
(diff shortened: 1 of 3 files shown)


sqlmesh/core/renderer.py
```diff
@@ -61,11 +59,22 @@ def __init__(
         self._default_catalog = default_catalog
         self._quote_identifiers = quote_identifiers
         self.update_schema({} if schema is None else schema)
-        self._cache: t.Dict[CacheKey, t.List[t.Optional[exp.Expression]]] = {}
+        self._cache: t.List[t.Optional[exp.Expression]] = []
+        self._model_fqn = model_fqn
```
izeigerman, line 63: I don't see this being passed into the `ExpressionRenderer` anywhere? Shouldn't we do this while rendering pre-/post- statements?

sqlmesh/core/renderer.py
```diff
@@ -61,11 +59,22 @@ def __init__(
...
 
     def update_schema(self, schema: t.Dict[str, t.Any]) -> None:
         self.schema = d.normalize_mapping_schema(schema, dialect=self._dialect)
 
+    def _should_cache(self, runtime_stage: RuntimeStage, *args: t.Any) -> bool:
+        return runtime_stage == RuntimeStage.LOADING and not any(args)
+
+    def _to_table_mapping(
```
izeigerman, line 71: [Nit] Can we move private methods to the bottom of the class def

sqlmesh/core/renderer.py
```diff
@@ -61,11 +59,22 @@ def __init__(
...
+        snapshots: t.Iterable[Snapshot], deployability_index: t.Optional[DeployabilityIndex]
+    ) -> t.Dict[str, str]:
+        from sqlmesh.core.snapshot import to_table_mapping
+
+        return to_table_mapping(snapshots, deployability_index)
```
izeigerman, line 76: Why not just call this directly?

tobymao: because i’d have to import teicr

sqlmesh/core/renderer.py
tobymao: done

izeigerman (approved):

PR #2481, opened 2024-04-20
Chore!: Use version instead of identifier in the seeds table

izeigerman:
The contents of the seed are only different if the snapshot's version is different too. Therefore, it doesn't make sense to store seed content by the (name, identifier) tuple since it leads to duplication and more complexity and inefficiencies downstream. For example, we don't need to migrate seed records if we were to store them by (name, version), since only identifiers change during the migration.

### DIFF
--- a/sqlmesh/migrations/v0049_replace_identifier_with_version_in_seeds_table.py
+++ b/sqlmesh/migrations/v0049_replace_identifier_with_version_in_seeds_table.py
@@ -1,0 +1,57 @@
+"""Use version instead of identifier in the seeds table."""
+
+from sqlglot import exp
+
+from sqlmesh.utils.migration import index_text_type
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
+    engine_adapter = state_sync.engine_adapter
+
+    snapshots_table = "_snapshots"
+    seeds_table = "_seeds"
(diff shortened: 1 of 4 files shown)


sqlmesh/migrations/v0049_replace_identifier_with_version_in_seeds_table.py
```diff
@@ -0,0 +1,57 @@
...
+    new_seeds_table = f"{seeds_table}_v49"
+
+    if state_sync.schema:
+        snapshots_table = f"{state_sync.schema}.{snapshots_table}"
+        seeds_table = f"{state_sync.schema}.{seeds_table}"
+        new_seeds_table = f"{state_sync.schema}.{new_seeds_table}"
+
+    index_type = index_text_type(engine_adapter.dialect)
+
+    engine_adapter.drop_table(new_seeds_table)
```
tobymao, line 22: just to be safe?

tobymao (approved):

sqlmesh/migrations/v0049_replace_identifier_with_version_in_seeds_table.py
izeigerman: Yeah. Eg. a cleanup after an unsuccessful migration.

PR #2497, opened 2024-04-23
fix: push snapshots before updating intervals

eakmanrq:
Prior to this we would add a reference to the intervals tables to a snapshot that doesn't exist when doing forward-only changes. This flips the order so we first add the snapshot and then start adding intervals.


izeigerman:
I think the same problem exists in case of Airflow https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/schedulers/airflow/plan.py#L104 but there the fix is a lot more involved potentially. What's the problem with interval being created before the snapshot?

izeigerman (approved):

PR #2502, opened 2024-04-24
Fix: allow this_model to be used as a sqlmesh native macro var

georgesittas:
The updated test currently fails in main:

```
E         Skipping 54 identical leading characters in diff, use -v to show
E         - "x" FROM "db"."table")
E         + "x" FROM """db"".""table""")
E         ?          ++    + +       ++
```

I also saw a different error when I was testing locally with a duckdb project, which makes sense because we stored a dict in the macro evaluator's `locals` and then `exp.convert`ed it into a `MAP`:

```
duckdb.duckdb.CatalogException: Catalog Error: Table Function with name map does not exist!
Did you mean "range"?
LINE 1: ...table" ("col") SELECT * FROM MAP(['"db"."dev"."table"'], [...
                                                  ^
```

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -101,11 +101,12 @@ class BaseExpressionRenderer:
 
         if self._model_fqn and "this_model" not in kwargs:
             if snapshots:
-                kwargs["this_model"] = self._to_table_mapping(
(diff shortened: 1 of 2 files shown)


georgesittas:
Closing as we don't really need to expose this_model as a sqlmesh native macro.

PR #2504, opened 2024-04-24
feat: support `condition` argument in `forall` audit

plaflamme:
This isn't strictly necessary since you can add the condition in the `criteria`. But there are several advantages to supporting `condition`:

1. it makes `forall` consistent with other audits
2. it is simpler to reason about, consider `foreall(criteria=[f != 42 OR a > b])` vs `forall(criteria=[a > b], condition=(f=42))`
3. more concise if you want to apply multiple criteria to the same subset of rows

### DIFF
--- a/sqlmesh/core/audit/builtin.py
+++ b/sqlmesh/core/audit/builtin.py
@@ -83,16 +83,21 @@ HAVING COUNT(*) <= @threshold
 # ))
 forall_audit = ModelAudit(
     name="forall",
+    defaults={"condition": exp.null()},
     query="""
 SELECT *
 FROM @this_model
-WHERE @REDUCE(
-  @EACH(
-    @criteria,
-    c -> NOT (c)
-  ),
-  (l, r) -> l OR r
-)
+WHERE
+  @AND(
+    @REDUCE(
+      @EACH(
+        @criteria,
+        c -> NOT (c)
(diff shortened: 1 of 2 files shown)


izeigerman:
Hey @plaflamme, thanks a lot for the contribution! I'm not sure I follow: is there anything that the `condition` argument can expression that the `criteria` one can't? 

Otherwise, I'm somewhat hesitant to introduce additional arguments that significantly overlap in functionality with the existing ones. I'm concerned that it will create confusion around which one to use and also create multiple styles of achieving same things.

izeigerman:
@plaflamme this is helpful context and examples. Any chance you can add it in the corresponding docs section: https://github.com/TobikoData/sqlmesh/blob/main/docs/concepts/audits.md?plain=1#L101? LGTM otherwise.

plaflamme:
@izeigerman 

> Any chance you can add it in the corresponding docs section

Are you referring to documenting `condition` or having a specific example of how `condition` is useful for this audit? I'm asking because none of the audits that support the `condition` parameter actually document that fact currently. Perhaps adding this documentation should be a separate task?

izeigerman:
@plaflamme 

That's fair. I didn't realize we didn't document `condition` anywhere. Sounds good

PR #2509, opened 2024-04-24
Fix: Support of the table renaming in Snowflake and Databricks

izeigerman:


georgesittas (approved):
Perhaps this is something that we can fix upstream in sqlglot?

tobymao:
this is already fixed in sqlglot isn't it?

georgesittas:
I think we only drop the db, ie. the other way around: https://github.com/tobymao/sqlglot/blob/fb9a7ad8f2af98a248e4576677b7b615b9d4c3e7/sqlglot/generator.py#L2970-L2974

tobymao:
ah, let's fix it there then

PR #2511, opened 2024-04-24
Toby/typed macros

tobymao:


examples/sushi/macros/utils.py
```diff
@@ -1,9 +1,20 @@
+from sqlglot import exp
+
 from sqlmesh import macro
 
 
 @macro()
-def add_one(evaluator, column):
-    return f"{column} + 1"
+def add_one(evaluator, column: int):
```
izeigerman, line 7: Just a heads up that I don't think that this is how this meant to be used. I think it's suppose to return the SQL `Add` expression, not perform the operation in place.

sqlmesh/core/macros.py
```diff
@@ -388,6 +423,75 @@ def var(self, var_name: str, default: t.Optional[t.Any] = None) -> t.Optional[t.
...
+                if isinstance(coerced, base):
+                    return coerced
+                raise SQLMeshError(base_err_msg)
+
+            if base in (int, float, str) and isinstance(expr, exp.Literal):
+                return base(expr.this)
+            if base is bool and isinstance(expr, exp.Boolean):
+                return expr.this
+            if base is str and isinstance(expr, exp.Expression):
+                return expr.sql(self.dialect)
```
izeigerman, line 466: This seems sketchy. What's the use case for this? IMO I think we should only coerce string literals and identifiers into strings.

tobymao: this seems reasonable to me. for example, you may pass a column or a table into a function, and just care about the string representation

izeigerman: Sure, I think this is reasonable for columns, tables and literals. But this logic will apply even if I pass an arbitrary SQL expression, right?

tobymao: without this coercision to string, usage like this is flawed because x will become sqlglot dialect sql

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1451,6 +1451,116 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
...
+```python linenums="1"
+@macro()
+def stamped(evaluator, query: exp.Select) -> exp.Subquery:
+    return query.select(exp.Literal.string(str(datetime.now())).as_("stamp")).subquery()
+
+# Coercing to a complex node like `exp.Select` works as expected given a string literal input
+# SELECT * FROM @stamped('SELECT a, b, c')
+```
+
+When coercion fails, there will always be a warning logged but we will not crash. We believe the macro should be flexible by default, meaning the default behavior is preserved if we cannot coerce. Give that, the user can express whatever level of additional checks they want. For example, if you would like to raise an error when the coercion fails, you can use an `assert` statement. For example:
```
georgesittas, line 1517:
```suggestion
When coercion fails, there will always be a warning logged but we will not crash. We believe the macro system should be flexible, meaning the default behavior is preserved if we cannot coerce. Given that, the user can express whatever level of additional checks they want. For example, if you would like to raise an error when the coercion fails, you can use an `assert` statement. For example:
```

PR #2513, opened 2024-04-24
fix: action bot properly output all changes

eakmanrq:
Prior to this change the bot displayed changed by using the output generated when categorizing changes which would give you the directly modified snapshot and then the indirectly modified models in a subtree. The problem is that this doesn't cover all possible changes and therefore removed snapshots (and likely other things) were not being represented.

This change uses the normal output that is created instead which captures everything. This was actually already written but not used since we switched to the above approach so it just needed updated. I tested all the new output that is generated against this to make sure it formats properly: https://jbt.github.io/markdown-editor/

The main regression here is that users no longer see the directly modified and then below it what was indirectly modified and instead they are listed in separate lists. Will consider improving this based on user feedback.

### DIFF
--- a/tests/integrations/github/cicd/test_integration.py
+++ b/tests/integrations/github/cicd/test_integration.py
@@ -154,7 +154,12 @@ def test_merge_pr_has_non_breaking_change(
     assert GithubCheckStatus(prod_plan_preview_checks_runs[1]["status"]).is_in_progress
     assert GithubCheckStatus(prod_plan_preview_checks_runs[2]["status"]).is_completed
     assert GithubCheckConclusion(prod_plan_preview_checks_runs[2]["conclusion"]).is_success
-    expected_prod_plan_summary = """```diff
+    expected_prod_plan_summary = """**Summary of differences against `prod`:**
+
+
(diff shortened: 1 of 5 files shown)


tests/integrations/github/cicd/test_integration.py
```diff
@@ -1581,14 +1560,7 @@ def test_overlapping_changes_models(
  LEFT JOIN current_marketing AS m
    ON o.customer_id = m.customer_id
 ```
-
-```
-
-Directly Modified: sushi.customers (Non-breaking)
```
izeigerman, line 1587: Why did the output change so drastically?

izeigerman (approved):

tests/integrations/github/cicd/test_integration.py
eakmanrq: I tried to cover this with the PR description. I used to use the output that we had for categorization in order to get this nice direct change -> indirect relationship but switching to this other console method loses that. The problem was that the other method was only designed to help with categorization so it was missing things like removed models.

PR #2514, opened 2024-04-25
Chore: Refactor migration script application

vchan:


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -908,6 +908,24 @@ def _restore_table(
             new_table_name=table_name,
         )
 
+    def apply_migrations(
```
izeigerman, line 911: Let's make it private

tobymao (approved):

PR #2515, opened 2024-04-25
Chore: update to docs to recommend using utf-8

georgesittas:


docs/concepts/models/overview.md
```diff
@@ -1,6 +1,6 @@
 # Overview
 
-Models are comprised of metadata and queries that create tables and views, which can be used by other models or even outside of SQLMesh. They are defined in the `models/` directory of your SQLMesh project and live in `.sql` files.
+Models comprise metadata and queries that create tables and views, which can be used by other models or even outside of SQLMesh. They are defined in the `models/` directory of your SQLMesh project and live in `.sql` files.
```
tobymao, line 3: this should be reverted

tobymao (approved):

docs/concepts/models/overview.md
georgesittas: "comprised of" is an incorrect form

Issue #2014, opened 2024-01-24
[Bigquery] add labels property to VIEWs


treysp:
SQLMesh allows a model to pass arbitrary table properties to its `CREATE` statement with the [table_properties](https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/#table_properties) model configuration parameter. 

Most of the properties this is used for relate to physical storage or partitioning and are only relevant for physical tables. Therefore, the properties are not included in view creation statements.

Bigquery supports the [labels](https://cloud.google.com/bigquery/docs/adding-labels#adding_table_and_view_labels) metadata property, which can be applied to both tables and views. They are similar to tags, but they are key-value pairs rather than strings.

This issue requests that a model's labels property be propagated into its views as well as its physical tables.

seblum:
We are looking into something like this as well. We already had the case that some tables have indeed been labeled in BigQuery after adding `table_properties [ lables=[()] ]`. So, I am not sure whether this has been implemented in some cases already.. However, every change to the table, update, or also within virtual tables removed those labels. 

Would be great to have this implemented in general!

izeigerman:
#2517 introduced propagation of `table_properties` into VIEWS in the **physical** layer. We still don't set them in the virtual layer and it's not quite clear whether we even should.

plaflamme:
@izeigerman agreed, it's unclear to me that these should be forwarded since `table_properties` is definitely about the physical layer.

That being said, having the ability to specify labels on the virtual environment views in BigQuery would be nice for integration with other GCP tooling. Having an equivalent `view_properties` for the "virtual layer" would solve this.

PR #2517, opened 2024-04-25
Fix: Propagate table properties as options when creating views in BQ

izeigerman:
BigQuery supports the `OPTIONS` clause when creating views. Previously we weren't propagating `table_properties` set at the model level into the view creation statement.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -568,6 +568,8 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
                 ),
             )
 
+        properties.extend(self._table_properties_to_expressions(table_properties))
+
         if properties:
             return exp.Properties(expressions=properties)
         return None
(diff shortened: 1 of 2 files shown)


vchan (approved):

tobymao (approved):

plaflamme:
This was reverted, but I came here to mention that not all properties should be forwarded since some apply might only apply to the physical table.

Thankfully, the documentation is pretty clear:

* [list of options that apply to tables](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#table_set_options_list)
* [list of options that apply to views](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#view_set_options_list)

tobymao:
it was not reverted, i reverted a bad commit, cc @izeigerman

plaflamme:
Also note that this would close https://github.com/TobikoData/sqlmesh/issues/2014

tobymao:
@plaflamme this is only applied to the physical layer, so it's up to the user to specify the correct properties. if they specify table level properties to a view materialization, that's user error.

plaflamme:
@tobymao I see, my mistake, carry on!

PR #2520, opened 2024-04-26
feat: vararg and kwarg macro defs closes #2484

### LINKED ISSUE #2484: Add support for named arguments in SQLMesh macros
There is no clean way to skip optional arguments in macros today. For example, if we want to specify an `except` value for the `@STAR` [macro](https://sqlmesh.readthedocs.io/en/latest/concepts/macros/sqlmesh_macros/#star), we also need to specify an alias, otherwise the SQL arguments won't map to the correct python arguments.

One way to address this is to add support for named arguments. We could use `PropertyEQ` (ie. `:=`) to achieve this, so the `@STAR` issue would be resolved by simply doing:

```sql
SELECT @STAR(foo, except := [a, b, c])
```

tobymao:


tobymao:
cc @z3z1ma

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1210,6 +1210,27 @@ SELECT
 FROM table
 ```
 
+#### Vararg and kwarg arguments
```
georgesittas, line 1213: I'd go for `#### Variable-length and named arguments` or something, and then establish the short-hand terminology using parentheses.

georgesittas (approved):
Nice!

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1210,6 +1210,27 @@ SELECT
...
+Varargs (variable arguments) and kwargs (keyword arguments) are possible to use in Python macros definitions. Kwargs can be used in SQL with the property assignment operator `:=`.
```
georgesittas, line 1214: Maybe we should expand a little on what "varargs" and "kwargs" mean and perhaps show a small python function as an example to establish the connection between the two.

PR #2523, opened 2024-04-26
feat: adds plan id to apply notifications

isaiahgm:
Makes changes to how apply notifications are called in context.py to include plan_id. Also modifies the Airflow plan_request_id to be the same as plan_id so that apply notifications can include links to the Airflow job. All of this will help our users find logs when they want to see what SQL queries SQL Mesh runs or track down an error.
Currently when a SQL Mesh apply fails with Airflow acting as scheduler, the stack trace sent to the notification integration doesn't include any information about why the apply failed. Getting the actual stack trace out of airflow and into the notification would be very complicated, so a good work around is just to include a link to the DAG so users can quickly explore the logs for their run.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -499,5 +499,4 @@ def update_intervals_for_new_snapshots(
         if snapshot.is_forward_only:
             snapshot.dev_intervals = snapshot.intervals.copy()
             for start, end in snapshot.dev_intervals:
-                from sqlmesh.utils import random_id
-state_sync.add_interval(snapshot, start, end, is_dev=True)
+                state_sync.add_interval(snapshot, start, end, is_dev=True)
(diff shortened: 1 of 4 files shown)


sqlmesh/core/plan/evaluator.py
```diff
@@ -499,4 +499,5 @@ def update_intervals_for_new_snapshots(
         if snapshot.is_forward_only:
             snapshot.dev_intervals = snapshot.intervals.copy()
             for start, end in snapshot.dev_intervals:
-                state_sync.add_interval(snapshot, start, end, is_dev=True)
+                from sqlmesh.utils import random_id
+state_sync.add_interval(snapshot, start, end, is_dev=True)
```
tobymao, line 503: this is formatting error

isaiahgm: Huh... must of missed that. Thanks! Probably a typo when I was catching my repo up with the changes since I forked it. I'll push a fix shortly.

isaiahgm:
Looks like the test cases for notify() and notify_user() need to be updated or the logic needs to be changed to still be backwards compatible.

izeigerman (approved):

PR #2524, opened 2024-04-26
Fix: force-lowercase dialect names, normalize macro columns_to_types

georgesittas:
This fixes a couple of bugs I came across while trying to get the sushi project to run in snowflake.

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -402,14 +402,12 @@ class MacroEvaluator:
             self.columns_to_types_called = True
             return {"__schema_unavailable_at_load__": exp.DataType.build("unknown")}
 
-        if isinstance(model_name, exp.Column):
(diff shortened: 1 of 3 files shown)


sqlmesh/core/macros.py
```diff
@@ -402,14 +402,12 @@ def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, ex
             self.columns_to_types_called = True
             return {"__schema_unavailable_at_load__": exp.DataType.build("unknown")}
 
-        if isinstance(model_name, exp.Column):
```
tobymao, line 405: why wasn't this needed anymore?

georgesittas: it's handled in `normalize_model_name`

tobymao (approved):

PR #2525, opened 2024-04-26
Docs: clarify macro function keyword args

treysp:


georgesittas (approved):
LGTM 👍

There's an additional functionality that's now supported: variable-length argument lists. For example, you can have a macro definition which looks like:

```python
@macro()
def foo(evaluator, *args):
    return sum(args)
```

So you can use it in a model as follows:

```sql
model (name a);

select @foo(1, 2, 3, 4) as c
```

And this will render to `SELECT 10 AS "c"`. Perhaps we should document this as well.

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1210,26 +1210,25 @@ SELECT
...
 from sqlmesh import macro
 
 @macro()
-def kwarg_macro(evaluator, a, b=None, c=None):
-  if b:
-    return b
-  if c:
-    return c
-  return a
+def add_args(evaluator, argument_1=1, argument_2=2, argument_3=3):
```
tobymao, line 1223: this definition is not right, your defaults are int, but they won't be integers unless you add a type annotation `:int`. as is, arguments will be sqlglot expressions

treysp: Thanks, I'll update and add more information about typing. Is our general position that macro arg types are required?

Question: why is this rendering correctly without the types?

Macro definition:
```
@macro()
def add_args(evaluator, argument_1=1, argument_2=2, argument_3=3):
    return argument_1 + argument_2 + argument_3
```
Call in model:
```
SELECT
    item_id,
    COUNT(DISTINCT id) AS num_orders,
    @add_args(argument_2 := 1) as add_args,
FROM
    sqlmesh_example.incremental_model
GROUP BY item_id
```
Rendered query:
```
❯ sqlmesh render sqlmesh_example.full_model
SELECT                                                                                                                                                                                                                                                        
  "incremental_model"."item_id" AS "item_id",                                                                                                                                                                                                                 
  COUNT(DISTINCT "incremental_model"."id") AS "num_orders",                                                                                                                                                                                                   
  5 AS "add_args"                                                                                                                                                                                                                                             
FROM "db"."sqlmesh__sqlmesh_example"."sqlmesh_example__incremental_model__1732922178" AS "incremental_model" /* db.sqlmesh_example.incremental_model */                                                                                                       
GROUP BY                                                                                                                                                                                                                                                      
  "incremental_model"."item_id" 
```

treysp: Answer: i got "luck" - in this specific example, the behavior of the sqlglot expressions happens to result in the same output as if types had been present

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1210,27 +1214,93 @@ SELECT
...
+from sqlmesh import macro
+
+@macro()
+def arg_in_123(evaluator, my_arg: int): # Type annotation `my_arg: int`
+    return my_arg in [1,2,3]
+```
+
+Now the macro call will return `SELECT TRUE` because the value is coerced to a Python integer before the function code executes and `1` is in `[1,2,3]`.
+
+Type annotations should **always** be specified for arguments with a default value because, in contrast to user-supplied values, the default is not parsed by SQLGlot before the function code executes. Type annotation ensures that both default and user-supplied argument values are of the same type in the function body.
```
tobymao, line 1259: type annotations don't need to be specified with default arguments, they just need to match.

```
def foo(evaluator, column=None):
```

is fine or

```
def foo(evaluator, column = Literal.string("hello"))
```
is also fine

georgesittas (approved):
Looks great!

PR #2536, opened 2024-04-30
fix: mssql integration tests

tobymao:
the tests needed to be fixed because the were written with raw strings and parsed in each dialect. tsql doesn't have a real timestamp type.

the macros were resulting in a syntax error, so i cleaned them up and removed the ones with warnings

i also changed sql server to 2019 for linux compatibility

### DIFF
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -1255,7 +1255,7 @@ def test_render_query(assert_exp_eq, sushi_context):
         WITH "test_macros" AS (
           SELECT
             2 AS "lit_two",
-            "waiter_revenue_by_day"."revenue" > 0 AS "sql_exp",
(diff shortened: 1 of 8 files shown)


tests/core/test_model.py
```diff
@@ -3870,7 +3870,7 @@ def test_this_model() -> None:
             start="2020-01-01",
             snapshots={snapshot.name: snapshot},
         ).sql(dialect="bigquery")
-        == "SELECT '`sqlmesh__project-1`.`project_1__table__1178373199`' AS `x`"
+        == "SELECT '`sqlmesh__project-1`.`project_1__table__2235533348`' AS `x`"
```
georgesittas, line 3873: can we just do `f"{snapshot.version}"` here instead?

georgesittas (approved):
LGTM

izeigerman (approved):

PR #2537, opened 2024-04-30
fix: add model prop comments closes #2533

### LINKED ISSUE #2533: Comments on `MODEL` properties are removed by `sqlmesh format`
For example:

```sql
MODEL(
  name foo.bar, -- my model is awesome
  kind VIEW,
);
SELECT * FROM foo_bar
```

Will produce

```sql
MODEL(
  name foo.bar,
  kind VIEW,
);
SELECT
  *
FROM foo_bar
```

tobymao:


sqlmesh/core/dialect.py
```diff
@@ -499,9 +499,14 @@ def parse(self: Parser) -> t.Optional[exp.Expression]:
             if isinstance(value, exp.Expression):
                 value.meta["sql"] = self._find_sql(start, self._prev)
 
-            expressions.append(self.expression(exp.Property, this=key, value=value))
-
-            if not self._match(TokenType.COMMA):
+            if self._match(TokenType.COMMA):
```
georgesittas, line 502: I think we can just do `self._match(TokenType.COMMA, expression=prop)` here, assuming we did `prop = self.expression(...)` above this

georgesittas (approved):

PR #2543, opened 2024-04-30
chore: add interval inclusive/exclusive method

eakmanrq:
Prior to this change the contract to add an interval was that you must provide a whole snapshot in order to add an interval when really intervals just need snapshot name, identifier, and version. This change provides an endpoint that just requires those minimal fields. Once split out, the functionality of `add_interval` is really just called `inclusive_exclusive` on a snapshot which would likely be universal across all state syncs. Therefore it was converted to not be abstract anymore and call the new `add_inclusive_exclusive_interval` method.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -18,7 +18,7 @@ from sqlmesh.core.snapshot import (
     SnapshotTableCleanupTask,
     SnapshotTableInfo,
 )
-from sqlmesh.core.snapshot.definition import Interval
+from sqlmesh.core.snapshot.definition import Interval, SnapshotIntervals
 from sqlmesh.utils import major_minor
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.errors import SQLMeshError
@@ -304,25 +304,6 @@ class StateSync(StateReader, abc.ABC):
         """
 
     @abc.abstractmethod
-    def add_inclusive_exclusive_interval(
-        self,
-        snapshot_id: SnapshotId,
-        snapshot_version: str,
-        start_ts: int,
-        end_ts: int,
-        is_dev: bool = False,
-    ) -> None:
-        """Add an inclusive-exclusive interval to a snapshot and sync it to the store.
-
-        Args:
-            snapshot_id: The snapshot id to add an interval to.
-            snapshot_version: The snapshot version to add an interval to.
-            start_ts: The inclusive start of the interval to add.
-            end_ts: The exclusive end of the interval to add.
-            is_dev: Indicates whether the given interval is being added while in development mode
-        """
-
-    @abc.abstractmethod
     def remove_interval(
         self,
         snapshot_intervals: t.Sequence[t.Tuple[SnapshotInfoLike, Interval]],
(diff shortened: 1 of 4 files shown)


sqlmesh/core/state_sync/base.py
```diff
@@ -429,6 +428,28 @@ def migrate(
     def rollback(self) -> None:
         """Rollback to previous backed up state."""
 
+    def add_interval(
+        self,
+        snapshot: Snapshot,
```
tobymao, line 433: why don't we just make add_interval accept a snapshot id? the reason for this in the past was probably because intervals were attached to snapshots but this is no longer true

why do we need to split out this method?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -639,12 +649,21 @@ def add_interval(
             "dev interval" if is_dev else "interval",
             start_ts,
             end_ts,
-            snapshot.snapshot_id,
+            snapshot_id,
+        )
+        snapshot_intervals = SnapshotIntervals(
```
izeigerman, line 654: Should we just make the new method accept `SnapshotIntervals` with populated `intervals` / `dev_intervals`? This will make the API cleaner IMHO

izeigerman: eg:
```
def add_snapshot_intervals(self, intervals: SnapshotIntervals)
```

sqlmesh/core/state_sync/base.py
eakmanrq: We discussed this option over a call.

sqlmesh/core/state_sync/engine_adapter.py
tobymao: ya i like that

eakmanrq: https://github.com/TobikoData/sqlmesh/pull/2543/commits/016a36b60d9beaa14107e767fcf31d715e22ef3f

sqlmesh/core/state_sync/base.py
```diff
@@ -304,23 +304,11 @@ def invalidate_environment(self, name: str) -> None:
...
-        self,
-        snapshot: Snapshot,
-        start: TimeLike,
-        end: TimeLike,
-        is_dev: bool = False,
-    ) -> None:
-        """Add an interval to a snapshot and sync it to the store.
-
-        Snapshots must be pushed before adding intervals to them.
+    def _add_snapshot_intervals(self, snapshot_intervals: SnapshotIntervals) -> None:
```
izeigerman, line 307: [Nit] Let's move "private" method to the bottom of the definition

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1312,6 +1333,27 @@ def _intervals_to_df(
...
+        [
+            _interval_to_df(
+                snapshot_intervals,
+                start_ts,
+                end_ts,
+                is_dev=is_dev,
+                is_removed=is_removed,
+            )
+            for is_dev in (False, True)
+            for start_ts, end_ts in getattr(
```
izeigerman, line 1350: What happens if it's empty?

izeigerman (approved):

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: If both are empty then it would return an empty pandas dataframe. The contract here is to always return a dataframe so that seems like expected behavior. The place where this is currently used checks for empty intervals and just does nothing in that case.

PR #2544, opened 2024-05-01
chore: make connection factory property with kwargs

eakmanrq:
Standardizing this as a separate property instead of creating it within the create engine adapter call.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -8,6 +8,7 @@ import pathlib
 import sys
 import typing as t
 from enum import Enum
+from functools import partial
 
 from pydantic import Field
 from sqlglot import exp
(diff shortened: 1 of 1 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -84,15 +85,21 @@ def is_recommended_for_state_sync(self) -> bool:
         """Whether this connection is recommended for being used as a state sync for production state syncs"""
         return self.type_ in RECOMMENDED_STATE_SYNC_ENGINES
 
+    @property
+    def _connection_factory_with_kwargs(self) -> t.Callable:
```
izeigerman, line 89: Should we more specific with type here? Eg. specify that it expects no arguments

izeigerman (approved):

PR #2545, opened 2024-05-01
Chore: Additional tests for lineage with derived tables

vchan:


web/server/api/endpoints/lineage.py
```diff
@@ -27,6 +27,9 @@ def get_source_name(node: Node, default_catalog: t.Optional[str], dialect: str)
     if node.reference_node_name:
         # CTE name
         return node.reference_node_name
+    column = exp.to_column(node.name)
+    if getattr(column, "table", None):
```
tobymao, line 31: why are you calling getattr? can you just do column.table?

web/server/api/endpoints/lineage.py
```diff
@@ -27,6 +27,9 @@ def get_source_name(node: Node, default_catalog: t.Optional[str], dialect: str)
     if node.reference_node_name:
         # CTE name
         return node.reference_node_name
+    column = exp.to_column(node.name)
```
tobymao, line 30: why do we call to_column? can't we just get the alias name of the subquery?

tobymao (approved):

PR #2547, opened 2024-05-01
Feat: @model decorator `kind` arg should be passed a dict

### LINKED ISSUE #2540: Use Python dict instead of a Pydantic model type to set a kind for Python models
By allowing users to specify model kind as a Pydantic model we give up a certain amount of control over how certain kind fields are being set (eg. `dialect`). As a result we sometimes have to resort to ghastly workarounds to set these values after the fact.

Instead of letting user create the kind instance directly, we should only let them set the kind's settings as a Python dict, so that the actual instance construction can happen downstream where SQLMesh controls the construction process.

Therefore, a typical Python model definition that looks like this:
```
@model(
    kind=IncrementalByTimeRangeKind(time_column="ds", batch_size=10),
    ...,
)
```
should look like this:
```
@model(
    kind=dict(name=ModelKindName.INCREMENTAL_BY_TIME_RANGE, time_column="ds", batch_size=10),
    ...,
)
```

Code examples in docs should be updated accordingly.

treysp:
Closes #2540

### DIFF
--- a/docs/concepts/models/python_models.md
+++ b/docs/concepts/models/python_models.md
@@ -41,29 +41,32 @@ If the function output is too large, it can also be returned in chunks using Pyt
 
 The arguments provided in the `@model` specification have the same names as those provided in a SQL model's `MODEL` DDL.
 
-Most of the arguments are simply Python-formatted equivalents of the SQL version, but Python model `kind`s are specified with model kind objects. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties). A model's `kind` object must be imported at the beginning of the model definition file before use in the model specification.
+Python model `kind`s are specified with a Python dictionary containing the kind's name and arguments. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties).
 
-Supported model kind objects include:
+The model `kind` dictionary must contain a `name` key whose value is a member of the [`ModelKindName` enum class](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ModelKindName). The `ModelKindName` class must be imported at the beginning of the model definition file before use in the `@model` specification.
 
-- [ViewKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ViewKind)
-- [FullKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#FullKind)
-- [SeedKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#SeedKind)
-- [IncrementalByTimeRangeKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#IncrementalByTimeRangeKind)
-- [IncrementalByUniqueKeyKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#IncrementalByUniqueKeyKind)
-- [SCDType2KindByTimeKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#SCDType2ByTimeKind)
-- [SCDType2KindByColumnKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#SCDType2ByColumnKind)
-- [EmbeddedKind()](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#EmbeddedKind)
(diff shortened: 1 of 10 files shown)


docs/concepts/models/python_models.md
```diff
@@ -41,29 +41,32 @@ If the function output is too large, it can also be returned in chunks using Pyt
 
 The arguments provided in the `@model` specification have the same names as those provided in a SQL model's `MODEL` DDL.
 
-Most of the arguments are simply Python-formatted equivalents of the SQL version, but Python model `kind`s are specified with model kind objects. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties). A model's `kind` object must be imported at the beginning of the model definition file before use in the model specification.
+Python model `kind`s are specified with a Python dictionary containing the kind's name and arguments. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties).
 
-Supported model kind objects include:
+The model `kind` dictionary must contain a `name` key whose value is a member of the [`ModelKindName` enum class](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ModelKindName). The `ModelKindName` class must be imported at the beginning of the model definition file before use in the `@model` specification.
```
georgesittas, line 46:
```suggestion
The model `kind` dictionary must contain a `name` key whose value is a member of the [`ModelKindName` enum class](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ModelKindName). The `ModelKindName` class must be imported at the beginning of the model definition file before used in the `@model` specification.
```

tests/core/test_model.py
```diff
@@ -1616,6 +1618,76 @@ def model2_entrypoint(evaluator: MacroEvaluator) -> str:
     )
 
 
+def test_python_model_decorator_kind() -> None:
+    logger = logging.getLogger("sqlmesh.core.model.decorator")
+
+    # no kind specified -> default View kind
+    @model("default_kind", columns={'"COL"': "int"})
+    def a_model(context):
+        pass
```
georgesittas, line 1627: FYI I believe this behavior is about to change so that python models have a `FULL` kind by default, cc @crericha.

tobymao (approved):
great thanks!

sqlmesh/core/model/decorator.py
```diff
@@ -27,6 +31,18 @@ def __init__(self, name: str, is_sql: bool = False, **kwargs: t.Any) -> None:
         if not is_sql and "columns" not in kwargs:
             raise ConfigError("Python model must define column schema.")
 
+        kind = kwargs.get("kind", None)
+        if kind is not None:
+            if isinstance(kind, _ModelKind):
+                logger.warning(
+                    f"""Python model "{name}"'s `kind` argument was passed a SQLMesh `{type(kind).__name__}` object. This is no longer supported - provide a dictionary instead."""
+                )
```
georgesittas, line 39: The `_ModelKind` object will still work until we actually deprecate this, right? Should we instead say that we plan to deprecate it instead of "is no longer supported"?

georgesittas (approved):

sqlmesh/core/model/decorator.py
treysp: Good call. @tobymao at some point are we going to error if we get a model instance or will we always allow with warning?

If we'll always allow it, maybe better wording is "may exhibit unexpected behavior"?

tobymao: yea the latter, we encourage people to not use the python api and it's not really supported

docs/concepts/models/python_models.md
```diff
@@ -41,29 +41,32 @@ If the function output is too large, it can also be returned in chunks using Pyt
 
 The arguments provided in the `@model` specification have the same names as those provided in a SQL model's `MODEL` DDL.
 
-Most of the arguments are simply Python-formatted equivalents of the SQL version, but Python model `kind`s are specified with model kind objects. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties). A model's `kind` object must be imported at the beginning of the model definition file before use in the model specification.
+Python model `kind`s are specified with a Python dictionary containing the kind's name and arguments. All model kind arguments are listed in the [models configuration reference page](../../reference/model_configuration.md#model-kind-properties).
 
-Supported model kind objects include:
+The model `kind` dictionary must contain a `name` key whose value is a member of the [`ModelKindName` enum class](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ModelKindName). The `ModelKindName` class must be imported at the beginning of the model definition file before used in the `@model` specification.
```
izeigerman, line 46:
```suggestion
The model `kind` dictionary must contain a `name` key whose value is a member of the [`ModelKindName` enum class](https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/model/kind.html#ModelKindName). The `ModelKindName` class must be imported at the beginning of the model definition file before being used in the `@model` specification.
```
Is something missing here?

izeigerman (approved):

PR #2548, opened 2024-05-01
Improve error message when the parsing of 'cluster_by' / 'partition_by' fields fails in the dbt loader

izeigerman:
Make the messages more specific about which field caused the failure.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -4,6 +4,7 @@ import logging
 import typing as t
 
 from sqlglot import exp
+from sqlglot.errors import SqlglotError
 from sqlglot.helper import ensure_list
 
 from sqlmesh.core import dialect as d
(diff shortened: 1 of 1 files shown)


sqlmesh/dbt/model.py
```diff
@@ -346,9 +352,13 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
         if self.cluster_by:
-            optional_kwargs["clustered_by"] = [
-                d.parse_one(c, dialect=dialect).name for c in self.cluster_by
-            ]
+            clustered_by = []
+            for c in self.cluster_by:
+                try:
+                    clustered_by.append(d.parse_one(c, dialect=dialect).name)
+                except SqlglotError as e:
+                    raise ConfigError(f"Failed to parse cluster_by field '{c}': {e}") from e
```
georgesittas, line 360: Small QOL - should we aggregate these errors so they're all reported at once after the loop instead of 1 by 1?

georgesittas (approved):
LGTM - just a suggestion

sqlmesh/dbt/model.py
izeigerman: Why? It seems like our general strategy has been failing fast

tobymao (approved):

PR #2551, opened 2024-05-01
Fix: Model kind change should not be categorized as forward-only

izeigerman:
Before this fix when transitioning from a non-materialized kind like `VIEW` to a materialized incremental kind for which the `forward_only` flag is set, SQLMesh would categorized change as `FORWARD_ONLY` and try to reuse the existing physical table which is actually a view which can't be reused.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -581,9 +581,8 @@ class PlanBuilder:
         snapshot = self._context_diff.snapshots[s_id]
         if snapshot.name in self._context_diff.modified_snapshots:
             _, old = self._context_diff.modified_snapshots[snapshot.name]
-            # If the model kind has been changed from non-materialized to materialized,
-            # then we should not consider this to be a forward-only change.
-            if not old.is_materialized:
+            # If the model kind has changed, then we should not consider this to be a forward-only change.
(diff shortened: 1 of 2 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -577,8 +577,14 @@ def _apply_effective_from(self) -> None:
...
                 snapshot.effective_from = self._effective_from
 
-    def _is_forward_only_model(self, s_id: SnapshotId) -> bool:
+    def _is_forward_only_change(self, s_id: SnapshotId) -> bool:
         snapshot = self._context_diff.snapshots[s_id]
+        if snapshot.name in self._context_diff.modified_snapshots:
+            _, old = self._context_diff.modified_snapshots[snapshot.name]
+            # If the model kind has been changed from non-materialized to materialized,
+            # then we should not consider this to be a forward-only change.
+            if not old.is_materialized:
```
tobymao, line 586: can we change any kind in a forward only way? incremental -> unique by key -> scd type 2 ???

izeigerman: That is a good point. Let me think about this more and make a change accordingly.

izeigerman: Updated

tobymao (approved):

PR #2553, opened 2024-05-02
Feat!: add ability to override the engine adapter of a unit test

georgesittas:
Addresses the 11th item in https://github.com/TobikoData/sqlmesh/issues/1637

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -324,11 +324,8 @@ class GenericContext(BaseContext, t.Generic[C]):
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
         self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
-        test_connection_config = self.config.get_test_connection(
+        self._test_connection_config = self.config.get_test_connection(
             self.gateway, self.default_catalog, default_catalog_dialect=self.engine_adapter.DIALECT
-        )
-        self._test_engine_adapter = test_connection_config.create_engine_adapter(
-            register_comments_override=False
         )
 
         self._snapshot_evaluator: t.Optional[SnapshotEvaluator] = None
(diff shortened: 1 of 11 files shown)


sqlmesh/core/context.py
```diff
@@ -1372,42 +1372,46 @@ def test(
...
-                    stream=stream,
-                    default_catalog=self.default_catalog,
                 )
-        finally:
-            self._test_engine_adapter.close()
+
+            result = run_tests(
+                model_test_metadata=test_meta,
+                models=self._models,
+                engine_adapter=self._test_engine_adapter,
```
izeigerman, line 1405: Why even pass the engine_adapter here? Shouldn't we figure the default one out downstream?

georgesittas: Yeah I was thinking about this as well. The reason I did this was to avoid re-instantiating the test engine adapter. We instantiate it once in the `Context` constructor and use it in a few other places besides `Context.test`, so I couldn't just move it in `run_tests` too.

If that's ok I'd be happy to move it here.

izeigerman: To be fair the only other place it's used is in the `info` command. Provided that tests may refer to multiple adapters, the usefulness of testing the test connection as part of the `info` command becomes somewhat less clear. Also users can just run tests to test the connection.

Should we move the test connection management out of the context altogether?

georgesittas: Yep, this seems reasonable to me.

One gotcha is that `create_test` also [uses the testing adapter](https://github.com/TobikoData/sqlmesh/blob/f0ae0a189d63a91f179cc30e9d9bf9211a49e3b2/sqlmesh/core/context.py#L1351), but that shouldn't be a problem I think. We could move the test connection management in `run_tests` and then separately create on-demand the testing adapter for `create_test` in its body, making sure to close the connection after it's done. Wdyt?

izeigerman: Yes, I like this

georgesittas: Gotcha thanks, will refactor

georgesittas: Updated, added docs as well.

izeigerman (approved):

PR #2555, opened 2024-05-02
Fix: format CAST as :: less aggressively

### LINKED ISSUE #2554: SQLMesh format converts SAFE_CAST to CAST
This is not semantically the same operation. The issue will need to be resolved in sqlglot I imagine but posting here because of its relevance in SQLMesh given the format CLI command.

image

georgesittas:
Fixes #2554

I went with the `any` approach instead of `.. and not args.get("safe")` to ensure we there are no future regressions if we were to add new args to `Cast`.

### DIFF
--- a/examples/wursthall/models/src/customer_details.py
+++ b/examples/wursthall/models/src/customer_details.py
@@ -50,7 +50,7 @@ def execute(
         for _ in range(random.choice(range(10, 20))):
             customer_details.append(
                 CustomerDetails(
-                    id=faker.uuid4(),
+                    id=t.cast(str, faker.uuid4()),
                     name=faker.name(),
                     phone=faker.phone_number(),
                     email=faker.ascii_email(),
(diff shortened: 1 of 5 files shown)


tobymao (approved):

examples/wursthall/models/src/customer_details.py
```diff
@@ -50,7 +50,7 @@ def execute(
         for _ in range(random.choice(range(10, 20))):
             customer_details.append(
                 CustomerDetails(
-                    id=faker.uuid4(),
+                    id=str(faker.uuid4()),
```
georgesittas, line 53: FYI this change was triggered due to a mypy [error](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/12841/workflows/1a26c970-de75-46a1-8580-0991362ac09c/jobs/94167), probably caused by the latest Faker [release](https://github.com/joke2k/faker/releases/tag/v25.0.1).

I verified that the uuid is a string value, but I went ahead and used `str` here to enforce this instead of `t.cast`.

```python
>>> import faker
>>> type(faker.Faker().uuid4())
<class 'str'>
```

PR #2556, opened 2024-05-02
Fix: Include catalog if available when calling USE SCHEMA in snowflake

crericha:
USE SCHEMA fails when a seed is in a different db than the main project.

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -90,7 +90,10 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
                 #
                 # The above issue has already been fixed upstream, but we keep the following
                 # line anyway in order to support a wider range of Snowflake versions.
-                self.cursor.execute(f'USE SCHEMA "{temp_table.db}"')
+                schema = f'"{temp_table.db}"'
+                if temp_table.catalog:
+                    schema = f'"{temp_table.catalog}".{schema}'
+                self.cursor.execute(f"USE SCHEMA {schema}")
 
                 # See: https://stackoverflow.com/a/75627721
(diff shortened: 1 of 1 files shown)


izeigerman (approved):

izeigerman:
Test please

crericha:
Any suggestions on how to test? I manually verified.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -90,7 +90,10 @@ def query_factory() -> Query:
                 #
                 # The above issue has already been fixed upstream, but we keep the following
                 # line anyway in order to support a wider range of Snowflake versions.
-                self.cursor.execute(f'USE SCHEMA "{temp_table.db}"')
+                schema = f'"{temp_table.db}"'
+                if temp_table.catalog:
+                    schema = f'"{temp_table.catalog}".{schema}'
```
eakmanrq, line 95: The double quotes around `schema` were accidentally removed here

eakmanrq:
> Any suggestions on how to test? I manually verified.

You can do a test like this where you include the dataframe in the call where the target includes a catalog and then verify the `USE SCHEMA` expression that should have been called includes the catalog. 

https://github.com/TobikoData/sqlmesh/blob/430336fdfa5ee900d346422161f846ab51293687/tests/core/engine_adapter/test_base.py#L2410-L2423

sqlmesh/core/engine_adapter/snowflake.py
crericha: The double quotes are already in the `schema` var. Here is the output from the snowflake log - `USE SCHEMA "SQLMESH_TEST2"."sqlmesh__EXAMPLE2"`

eakmanrq: sorry yep missed that

PR #2561, opened 2024-05-03
Feat: Add virtual properties and rename table_properties to physical_properties

### LINKED ISSUE #2014: [Bigquery] add labels property to VIEWs

Kayrnt:
closes #2014 

- A new property `virtual_properties` is introduced: it is a dictionary that propagate to the engine the provided properties to the virtual layer (the "target view")
- `physical_properties` replaces `table_properties` (we keep the value as an alias for backward compatibility)
- A deprecation warning will be issued to use `physical_properties`instead of `table_properties`

To be figured out:
- What should we do in the case of dbt adapter? Should we apply the label to:
   -  Just the physical layer
   - Just the virtual layer
   - Both
   ? So far, it's just on the physical layer.


sqlmesh/core/model/common.py
```diff
@@ -137,7 +137,7 @@ def depends_on(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[
 
 
 properties_validator = field_validator(
-    "table_properties_",
+    "physical_properties_",
```
tobymao, line 140: did you need to add view_properties here?

sqlmesh/core/model/meta.py
```diff
@@ -64,6 +67,8 @@ class ModelMeta(_Node):
     references: t.List[exp.Expression] = []
     physical_schema_override: t.Optional[str] = None
     table_properties_: t.Optional[exp.Tuple] = Field(default=None, alias="table_properties")
+    physical_properties_: t.Optional[exp.Tuple] = Field(default=None, alias="physical_properties")
```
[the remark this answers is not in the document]
tobymao: https://github.com/TobikoData/sqlmesh/tree/main/sqlmesh/migrations

tobymao:
is the error due to a new version of pydantic?

sqlmesh/core/model/common.py
Kayrnt: I guess I should put all properties isn't it?
`table_properties` and `virtual_properties` as well

tobymao: i'd like to remove table_properties from the model, and only have table_properties be accepted as an input from legacy projects if that makes sense, so internally, we'd only ever have to deal with the two.

sqlmesh/core/model/meta.py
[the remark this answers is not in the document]
Kayrnt: Isn't these fields meant for Pydantic field parsing?
My goal is to be able to parse:
- table_properties ("deprecated")
- physical_properties (the new table_properties)
- virtual_properties (the new properties for virtual layer)

Should that append with a custom `Field` to factor `table_properties`and `physical_properties`.

> can we remove this and do a migration to move all physical_properties in the database to virtual_properties?
Do you mean all `table_properties` to `physical_properties` to migrate?

Kayrnt:
> is the error due to a new version of pydantic?

No I think it's PR code related as it works if I just use `physical_properties`. I suspect it's in the common code you pointed out 👀

sqlmesh/core/model/meta.py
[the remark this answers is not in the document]
tobymao: so i imagine pydantic will only have two fields, physical/virtual_properties. we migrate in the state all usage of table_properties -> physical properties.

at parse time, if we detect table_properties, we log a message and remap that to physical_properties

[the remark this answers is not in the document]
izeigerman: I don't think we need to migrate anything. Can't we just add a `table_properties` alias to `physical_properties_`?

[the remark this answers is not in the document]
izeigerman: If not, we can also handle this in the root model validator. And just pop `table_properties`, emit a warning, and then set it as `physical_properties`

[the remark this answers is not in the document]
izeigerman: In any case, only `physical_properties` should stay, and no updates should be necessary to the `EngineAdapter` layer.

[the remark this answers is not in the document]
Kayrnt: I also don't think a migration is needed or maybe I'm missing?
@izeigerman, in [this commit](https://github.com/TobikoData/sqlmesh/pull/2561/commits/32bdcbd10bcf5198c5a8925f651cb40e7c51b52f) I've updated the Pydantic parsing to use both with the related warning. Isn't it what you suggest?
Regarding `EngineAdapter`, I think we could indeed skip the `physical` / `virtual` naming and skip to the actual materialization: `table_properties` / `view_properties` WDYT?

tests/core/test_model.py
```diff
@@ -2582,6 +2582,150 @@ def my_model(context, **kwargs):
...
+            physical_properties [
+                labels = [('test-label', 'label-value')]
+            ]
+        );
+        SELECT a FROM tbl;
+        """
+        )
+    )
+    assert sql_model.physical_properties == {
+        "labels": exp.Array(
```
tobymao, line 2600: let's use the builder api for the tests,

exp.array("('test-label', 'label-value')")

sqlmesh/core/model/meta.py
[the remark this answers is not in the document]
Kayrnt: In [this commit](https://github.com/TobikoData/sqlmesh/pull/2561/commits/1632b9866fd1935e0b57381c359d318c3611423b) I tried to set proper boundaries with physical/virtual until the evaluator that translate those into table/view (depending on the actual chosen materialization).

tests/core/test_model.py
Kayrnt: Right, it's more compact 👍

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -745,10 +745,29 @@ def test_view_table_properties(make_mocked_engine_adapter: t.Callable, mocker: M
         },
     )
 
-    adapter.create_view("test_table", parse_one("SELECT 1"), table_properties={})
+    adapter.create_view(
+        "test_table",
+        parse_one("SELECT 1"),
+        table_description="some description",
+        view_properties={
+            "labels": exp.Array(
```
tobymao, line 753: can you use sqlglot builder here ?

tests/core/test_model.py
```diff
@@ -2582,6 +2582,139 @@ def my_model(context, **kwargs):
...
+                labels = [('test-virtual-label', 'label-virtual-value')],
+            )
+        );
+        SELECT a FROM tbl;
+        """
+        )
+    )
+    assert sql_model.physical_properties == {
+        "partition_expiration_days": exp.convert(7),
+        "labels": exp.Array(
```
tobymao, line 2622: ditto

tests/core/test_model.py
```diff
@@ -2582,6 +2582,139 @@ def my_model(context, **kwargs):
...
+                        exp.Literal.string("test-physical-label"),
+                        exp.Literal.string("label-physical-value"),
+                    ]
+                )
+            ]
+        ),
+    }
+
+    assert sql_model.virtual_properties == {
+        "labels": exp.Array(
```
tobymao, line 2635: ditto

tests/fixtures/migrations/snapshots.json
```diff
@@ -1 +1 @@
...], \"audits\": [], \"expressions\": [], \"jinja_macros\": {\"packages\": {}, \"root_macros\": {}, \"global_objs\": {}}, \"seed\": {\"content\": \"id,name\\n0,Toby\\n1,Tyson\\n2,Ryan\\n3,George\\n4,Chris\\n5,Max\\n6,Vincent\\n7,Iaroslav\\n8,Emma\\n9,Maia\\n10,Jim\\n\"}, \"source_type\": \"seed\"}, \"parents\": [], \"audits\": [], \"intervals\": [[1672531200000, 1673136000000]], \"dev_intervals\": [], \"created_ts\": 1680814464932, \"updated_ts\": 1680814464932, \"ttl\": \"in 1 week\", \"previous_versions\": [{\"fingerprint\": {\"data_hash\": \"1876476880\", \"metadata_hash\": \"570478986\", \"parent_data_hash\": \"0\", \"parent_metadata_hash\": \"0\"}, \"version\": \"2505706914\"}], \"indirect_versions\": {\"sushi.waiter_as_customer_by_day\": [{\"fingerprint\": {\"data_hash\": \"486172035\", \"metadata_hash\": \"1992853678\", \"parent_data_hash\": \"2154574190\", \"parent_metadata_hash\": \"1349779748\"}, \"version\": \"1267397572\"}, {\"fingerprint\": {\"data_hash\": \"486172035\", \"metadata_hash\": \"1992853678\", \"parent_data_hash\": \"2824767713\", \"parent_metadata_hash\": \"1349779748\"}, \"version\": \"3668757715\"}]}, \"version\": \"1204702829\", \"change_category\": 1}"}}
```
tobymao, line 1: what changed here?

tobymao:
seems almost ready, @izeigerman can you take a final look

docs/concepts/models/overview.md
```diff
@@ -219,20 +219,33 @@ MODEL (
 ### description
 - Optional description of the model. Automatically registered as a table description/comment with the underlying SQL engine (if supported by the engine).
 
-### table_properties
-- A key-value of arbitrary table properties specific to the target engine. For example:
+### physical_properties (previously table_properties)
+- A key-value of arbitrary table properties specific to the target engine applied on SQLMesh physical layer. For example:
```
georgesittas, line 223:
```suggestion
- A key-value mapping of arbitrary table properties specific to the target engine applied on SQLMesh's physical layer. For example:
```

docs/concepts/models/overview.md
```diff
@@ -219,20 +219,33 @@ MODEL (
...
+  physical_properties (
     partition_expiration_days = 7,
     require_partition_filter = true
   )
 );
 
 ```
 
+### virtual_properties
+- A key-value of arbitrary table properties specific to the target engine applied on SQLMesh virtual layer. For example:
```
georgesittas, line 237:
```suggestion
- A key-value mapping of arbitrary table properties specific to the target engine applied on SQLMesh's virtual layer. For example:
```

docs/reference/model_configuration.md
```diff
@@ -28,7 +28,8 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
...
-| `table_properties` | Arbitrary table properties specific to the target engine. Specified as key-value pairs (`key = value`)                                                                                                                                                                                                                           |       dict        |    N     |
+| `physical_properties` | Arbitrary table properties specific to the target engine applied on SQLMesh physical layer. Specified as key-value pairs (`key = value`)                                                                                                                                    
+| `virtual_properties` | Arbitrary table properties specific to the target engine applied on SQLMesh virtual layer. Specified as key-value pairs (`key = value`)                                                                                         |       dict        |    N     |
```
georgesittas, line 32:
```suggestion
| `physical_properties` | Arbitrary table properties specific to the target engine applied on SQLMesh's physical layer. Specified as key-value pairs (`key = value`)                                                                                                                                    
| `virtual_properties` | Arbitrary table properties specific to the target engine applied on SQLMesh's virtual layer. Specified as key-value pairs (`key = value`)                                                                                         |       dict        |    N     |
```

sqlmesh/core/model/meta.py
```diff
@@ -261,6 +265,19 @@ def _pre_root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
                     f"Cannot use argument 'grain' ({grain}) with 'grains' ({grains}), use only grains"
                 )
             values["grains"] = ensure_list(grain)
+
+        table_properties = values.pop("table_properties", None)
+        if table_properties:
+            model_name = values.get("name")
```
georgesittas, line 271: Nit: I think this is expected to always exist right? Given that `name` is required.
```suggestion
            model_name = values["name"]
```

sqlmesh/core/model/meta.py
```diff
@@ -328,13 +345,23 @@ def batch_concurrency(self) -> t.Optional[int]:
...
-            for expression in self.table_properties_.expressions:
-                table_properties[expression.this.name] = expression.expression
-            return table_properties
+    def physical_properties(self) -> t.Dict[str, exp.Expression]:
+        """A dictionary of properties that will be applied to the physical layer. It replaces table_properties which is deprecated."""
+        if self.physical_properties_:
+            physical_properties = {}
+            for expression in self.physical_properties_.expressions:
+                physical_properties[expression.this.name] = expression.expression
+            return physical_properties
```
georgesittas, line 354:
```suggestion
            return {e.this.name: e.expression for e in self.physical_properties_.expressions}
```

sqlmesh/core/model/meta.py
```diff
@@ -328,13 +345,23 @@ def batch_concurrency(self) -> t.Optional[int]:
...
+        return {}
+
+    @cached_property
+    def virtual_properties(self) -> t.Dict[str, exp.Expression]:
+        """A dictionary of properties that will be applied to the virtual layer."""
+        if self.virtual_properties_:
+            virtual_properties = {}
+            for expression in self.virtual_properties_.expressions:
+                virtual_properties[expression.this.name] = expression.expression
+            return virtual_properties
```
georgesittas, line 364:
```suggestion
            return {e.this.name: e.expression for e in self.virtual_properties_.expressions}
```

georgesittas (commented):
LGTM as well, a few small suggestions.

tests/fixtures/migrations/snapshots.json
georgesittas: `table_properties` was renamed to `physical_properties` and since this is a single line of JSON the whole thing is included in the diff.

tests/core/engine_adapter/test_bigquery.py
Kayrnt: Sure, then I'll update [the existing code](https://github.com/TobikoData/sqlmesh/pull/2561/files/8fc49c3d99c6df4e564fe007d9898f8a5656a6c3#diff-31d4a537b0e60bd12859a67b8e1432a28bcf6f0fb48dc970111a65cdc3b2fb54R735-R744) too to be consistent

sqlmesh/core/model/meta.py
github-christophe-oudar: Yes, it feels like it should always exists as it's explicitly added in the provided validation fields. I'll add that change so that it fails eagerly if that behavior changes (thus preventing a silent switch to a `Python model "None"'` log).

sqlmesh/core/model/meta.py
Kayrnt: Thanks, those suggestions appears more "Pythonic" 👍

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1847,13 +1849,14 @@ def temp_table(
             finally:
                 self.drop_table(table)
 
-    def _table_properties_to_expressions(
-        self, table_properties: t.Optional[t.Dict[str, exp.Expression]] = None
+    def _model_properties_to_expressions(
```
izeigerman, line 1852: Model properties means something else throughout the code so let's not use it here. Additionally, the engine adapter is too low level of a construct to be aware of something as high-level as "model".

I'm ok keeping with keeping this as `table_properties` or just `properties` since that's what we create here. If you want to be more specific I'm also fine with something like `table_or_view_properties`.

sqlmesh/core/model/meta.py
```diff
@@ -261,6 +265,19 @@ def _pre_root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
                     f"Cannot use argument 'grain' ({grain}) with 'grains' ({grains}), use only grains"
                 )
             values["grains"] = ensure_list(grain)
+
+        table_properties = values.pop("table_properties", None)
+        if table_properties:
```
izeigerman, line 270: This looks great 👍

sqlmesh/core/model/meta.py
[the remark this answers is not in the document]
izeigerman: Hey @Kayrnt, after additional consideration, I've realized that you do need the migration after all. Since we're getting rid of the `table_properties` field altogether we do need to reflect this in the existing state.

The fact that you had to change the `fixtures/migrations/snapshots.json` file confirms this suspicion. If we don't migrate, the next time users run the `plan` command after migration, it will show the diff because the field name has changed.

You can test this by first initializing a project with duckdb using a previous SQLMesh version and then upgrading to your current branch and running the plan command.

Sorry about the confusion initially, I really hoped we could avoid it.

tests/fixtures/migrations/snapshots.json
izeigerman: I think this file shouldn't have changed. The fact that it did indicates that a migration script is necessary.

izeigerman:
Unfortunately, the migration script is required to rename `table_properties` to `physical_properties` in the existing SQLMesh state. One minor naming comment. Looks great otherwise, thanks!

sqlmesh/core/engine_adapter/base.py
Kayrnt: Ok I see what you mean!

The goal of that change was to be consistent with the call parameters since we have both:
- https://github.com/Kayrnt/sqlmesh/blob/8bae4b47b7bd5a038f05bd3d51a126c4c747fcf8/sqlmesh/core/engine_adapter/bigquery.py#L532
- https://github.com/Kayrnt/sqlmesh/blob/8bae4b47b7bd5a038f05bd3d51a126c4c747fcf8/sqlmesh/core/engine_adapter/bigquery.py#L571

I would also suggest for `table_or_view_properties` (and `table_or_view_properties_to_expressions`) then as I'd rather have future maintainers understand that the function expects one or the other (and it's not a bug).

I'll make an update

tests/fixtures/migrations/snapshots.json
Kayrnt: I think we could keep `table_properties` as the Pydantic model is backward compatible (unless it's used another way?).
I didn't look into migrations so far so I'm not sure how it works and what's needed.

Kayrnt:
@izeigerman Looking a bit at the migration system, what you suggest is to create a new migration (something like `v0050_table_properties_to_physical_properties`) where:
- I parse `snapshot` from `snapshots` table
- I substitute the `table_properties` with an equivalent `physical_properties`
- I save again updated `snapshot` from `snapshots` table

I could do it in SQL but I wonder if I can safely assume that all state connections can support JSON modifications or else try a `replace` on `"table_properties"` but then it could conflict with existing values. Any preference? I think the most simpler/safer would be to parse the JSON in Python but hopefully there's no installation with millions of rows to migrate else it might take ages to migrate.

izeigerman:
@Kayrnt yes, the steps look accurate to me. 

> I think the most simpler/safer would be to parse the JSON in Python but hopefully there's no installation with millions of rows to migrate else it might take ages to migrate.

Yes, that's the safest approach which works uniformly across all supported engines. You shouldn't worry about millions of rows. It's going to be 10s of thousands at most.

izeigerman (approved):
This looks great, thanks a lot for addressing changes!

PR #2564, opened 2024-05-04
Feat: provide helpful error for invalid model defaults keys

treysp:


sqlmesh/core/config/loader.py
```diff
@@ -113,9 +114,19 @@ def load_config_from_paths(
...
+        supported_model_defaults = ModelDefaultsConfig.all_fields()
+        for default in model_defaults_dict:
+            if not default in supported_model_defaults:
+                raise ConfigError(
+                    f"'{default}' is not a valid model default configuration key. Please remove it from the `model_defaults` specification in your config file."
+                )
 
-    non_python_config = config_type.parse_obj(merge_dicts(*non_python_configs))
+    non_python_config_dict = merge_dicts(*non_python_configs)
+    _check_model_defaults_keys(non_python_config_dict.get("model_defaults", {}))
```
georgesittas, line 126: Should we inline this instead of using a closure?

georgesittas (approved):

sqlmesh/core/config/loader.py
treysp: Good call - I initially thought I'd need to use it again for the python check, but people pass a class instance directly so pydantic already handles it.

izeigerman (approved):

Issue #2336, opened 2024-03-23
Full DuckDB Attach Support


eakmanrq:
DuckDB allows defining not only a name to attach to but also a "type" for that name. An example is attaching a SQLite database: https://duckdb.org/docs/extensions/sqlite.html

We should consider matching dbt's format for this which would also make having this work through the adapter easier: https://github.com/duckdb/dbt-duckdb?tab=readme-ov-file#attaching-additional-databases

I think we could keep our `catalogs` property for this but make it be `t.Dict[str, t.Union[str, t.Dict[str, str]]`. This would provide a simple mapping for most cases and then a more complex option for doing the advanced attached options.

wseaton:
Hey @eakmanrq, I'd love to take a crack at this! (I'm @wseaton in the Tobiko slack)

eakmanrq:
@wseaton Hey Will! Sounds good! I think the trickiest part is trying to make sure we can maintain backwards compatibility since we will be changing something that is currently expected to just be a string to also support a dict too. Need to make sure Pydantic handles that properly. If not we could consider adding a new property.

PR #2565, opened 2024-05-05
 Feat: Full support for Duckdb attach options

wseaton:
This PR adds additional config surface to allow for DuckDB's attach options to be leveraged, similar to duckdb-dbt.

Ref: #2336

cc @eakmanrq

### DIFF
--- a/tests/core/test_config.py
+++ b/tests/core/test_config.py
@@ -204,14 +204,12 @@ config = Config(default_connection=DuckDBConnectionConfig())
     )
 
     with pytest.raises(
-        ConfigError,
-        match=r"^Default model SQL dialect is a required configuration parameter.*",
+        ConfigError, match=r"^Default model SQL dialect is a required configuration parameter.*"
     ):
         load_config_from_paths(Config, project_paths=[tmp_path / "config.yaml"])
 
     with pytest.raises(
-        ConfigError,
-        match=r"^Default model SQL dialect is a required configuration parameter.*",
+        ConfigError, match=r"^Default model SQL dialect is a required configuration parameter.*"
     ):
         load_config_from_paths(Config, project_paths=[tmp_path / "config.py"])
 
@@ -342,10 +340,7 @@ model_defaults:
     [
         (
             "'^dev$': dev_catalog\n    '^other$': other_catalog",
-            {
-                re.compile("^dev$"): "dev_catalog",
-                re.compile("^other$"): "other_catalog",
-            },
+            {re.compile("^dev$"): "dev_catalog", re.compile("^other$"): "other_catalog"},
             "duckdb",
             "",
         ),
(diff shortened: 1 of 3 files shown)


wseaton:
There is some formatting churn here caused by my editor setup that shouldn't be happening anymore. Will fixup, but wanted to note that the duckdb integration tests are passing w/ the following patch:

```patch
diff --git a/tests/core/engine_adapter/config.yaml b/tests/core/engine_adapter/config.yaml
index 66be470e..ca07c66a 100644
--- a/tests/core/engine_adapter/config.yaml
+++ b/tests/core/engine_adapter/config.yaml
@@ -4,7 +4,10 @@ gateways:
       type: duckdb
       catalogs:
         memory: ':memory:'
-        testing: 'testing.duckdb'
+        testing:
+          type: duckdb
+          path: 'testing.duckdb'
+          read_only: false
   inttest_trino:
     connection:
       type: trino
diff --git a/tests/core/engine_adapter/test_integration.py b/tests/core/engine_adapter/test_integration.py
index 6bbc746a..64e60d10 100644
--- a/tests/core/engine_adapter/test_integration.py
+++ b/tests/core/engine_adapter/test_integration.py
@@ -526,8 +526,9 @@ def engine_adapter(mark_gateway: t.Tuple[str, str], config) -> EngineAdapter:
         engine_adapter.DEFAULT_BATCH_SIZE = 1
     # Clear our any local db files that may have been left over from previous runs
     if mark == "duckdb":
-        for raw_path in (connection_config.catalogs or {}).values():
-            pathlib.Path(raw_path).unlink(missing_ok=True)
+        for options in (connection_config.catalogs or {}).values():
+            path = options if isinstance(options, str) else options.path
+            pathlib.Path(path).unlink(missing_ok=True)
     return engine_adapter
```

tests/core/test_config.py
```diff
@@ -477,12 +482,43 @@ def test_variables():
...
+              postgres:
+                type: 'postgres'
+                path: 'dbname=postgres user=postgres host=127.0.0.1'
+                read_only: true
+model_defaults:
+    dialect: ''
+        """
+        )
+
+    assert load_config_from_paths(
```
eakmanrq, line 521: Can you assert some characteristics about the config in this test? So these catalogs are defined and have the values expected?

tests/core/test_config.py
```diff
@@ -477,12 +482,43 @@ def test_variables():
...
     with pytest.raises(
-        ConfigError, match="Unsupported variable value type: <class 'sqlglot.expressions.Column'>"
+        ConfigError,
+        match="Unsupported variable value type: <class 'sqlglot.expressions.Column'>",
     ):
         Config(variables={"invalid_var": exp.column("sqlglot_expr")})
+
+
+def test_load_duckdb_attach_config(tmp_path_factory):
+    config_path = tmp_path_factory.mktemp("yaml_config") / "config_duckdb_attach.yaml"
```
eakmanrq, line 499: Since this is function scoped, you can do `tmp_path` which will give you a unique folder path for this function. Then this line would change to `tmp_path / "config_duckdb_attach.yaml"`

tests/core/test_config.py
wseaton: Absolutely, added a few in 28a7aaf

eakmanrq (approved):

tobymao (approved):

PR #2571, opened 2024-05-06
fix!: change databricks to replace where

eakmanrq:
Behavior Change: Prior to this change if using Databricks and doing an `INCREMENTAL_BY_TIME_RANGE` and lets say you were loading `2024-01-01` through `2024-01-10`. If your new data was missing `2024-01-02` then that partition would be untouched in the target table if it already exists. Now `2024-01-02` will be deleted. We don't document anywhere that users should expect this behavior and I don't anticipate this to be an issue.

As Databricks rolls out new products they seem to consistently lack support for dynamic insert overwrite support. Therefore we are going to switch Databricks to now do REPLACE WHERE by default.

In testing REPLACE WHERE behaves like DELETE FROM ... WHERE...; INSERT INTO...;. 

Example:

Table:

| id | ds                 |
|----|------------|
| 1  | 2024-01-01 |

Run this:
```sql
INSERT INTO table REPLACE WHERE TRUE SELECT id, ds FROM VALUES (2, '2024-01-02')
```

Results in

| id | ds                 |
|----|------------|
| 2  | 2022-01-02 |

Also tested the behavior change above and it functioned like a DELETE FROM ... WHERE...

Since REPLACE WHERE doesn't support providing columns we are relying on this function which will detect if the columns in `columns_to_types` doesn't match what is in the query: https://github.com/TobikoData/sqlmesh/blob/d7caf8ee138143c8ea1c39970048fe541898dbc8/sqlmesh/core/engine_adapter/base.py#L1955-L1977

Ran manual tests and also verified that the integration tests for Databricks passed.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -1105,7 +1105,6 @@ class EngineAdapter:
                         insert_exp = exp.insert(
                             query,
                             table,
-                            # Change once Databricks supports REPLACE WHERE with columns
                             columns=(
                                 list(columns_to_types)
                                 if not insert_overwrite_strategy.is_replace_where
@@ -1114,7 +1113,7 @@ class EngineAdapter:
                             overwrite=insert_overwrite_strategy.is_insert_overwrite,
                         )
                         if insert_overwrite_strategy.is_replace_where:
-                            insert_exp.set("where", where)
+                            insert_exp.set("where", where or exp.true())
                         self.execute(insert_exp)
 
     def update_table(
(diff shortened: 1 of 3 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -1114,7 +1113,7 @@ def _insert_overwrite_by_condition(
                             overwrite=insert_overwrite_strategy.is_insert_overwrite,
                         )
                         if insert_overwrite_strategy.is_replace_where:
-                            insert_exp.set("where", where)
+                            insert_exp.set("where", where or exp.true())
```
izeigerman, line 1116: In which case `where` might be missing?

izeigerman: Never mind, I see a test case

izeigerman (approved):

PR #2573, opened 2024-05-07
fix: execution_time was not getting passed in from plan

tobymao:


izeigerman:
How about this spot:https://github.com/TobikoData/sqlmesh/blob/8e237207b546f30d06366495a39ffa0b5ecced36/sqlmesh/schedulers/airflow/plan.py#L133

izeigerman:
Also I think we need to pass it here as well: https://github.com/TobikoData/sqlmesh/blob/8e237207b546f30d06366495a39ffa0b5ecced36/sqlmesh/schedulers/airflow/plan.py#L116

sqlmesh/core/state_sync/base.py
```diff
@@ -307,7 +307,6 @@ def invalidate_environment(self, name: str) -> None:
     def remove_interval(
         self,
         snapshot_intervals: t.Sequence[t.Tuple[SnapshotInfoLike, Interval]],
-        execution_time: t.Optional[TimeLike] = None,
```
izeigerman, line 310: @eakmanrq this changes the State Sync interface

izeigerman (approved):

PR #2574, opened 2024-05-07
Feat: allow macro functions to be used as model meta properties

georgesittas:


izeigerman (approved):
@tobymao can you take a look as well please

tobymao (approved):

PR #2576, opened 2024-05-07
Feat!: forward-only changes schema modification check

treysp:
Some forward-only changes require schema modifications that drop existing data (e.g., changing a column type from `str` to `int`). If a user inadvertently makes such a change, they will lose existing data. This feature identifies destructive changes and optionally warns/errors before they occur to avoid accidental data loss. 

The feature introduces a new model kind property `on_destructive_change` with valid values `allow`, `warn`, and `error` (default is `error`). The `on_destructive_change` property only applies to incremental models and can be specified as a model default.

Users may want to temporarily allow a destructive change to a forward-only model. The feature introduces to the `plan` command a new `--allow-destructive-model` selector to temporarily allow destructive changes.

**Of note**

- Defaults
    - `on_destructive_change` is a property of a model kind
        - Only incremental models are checked for destructive changes, and their default is `ERROR`
        - Non-incremental models default is `ALLOW`
    - Users may specify a default in the model_defaults configuration key - it will only apply to incremental models, as non-incremental models are always allowed

- Plan time check
    - Uses `SchemaDiffer` and engine-specific rules to determine whether a change is destructive.
    - We pass the current engine's SchemaDiffer into the plan builder to perform the comparison.
    - Evaluating a change requires that `columns_to_types` is known.
    - Evaluates both direct and indirect changes to a model. A complete evaluation requires knowing `columns_to_types` for the model and all directly modified ancestors.

- Run time check
    - Uses the set of `Alter`s passed from `SchemaDiffer` to `alter_table()` to determine whether a change is destructive.

- dbt
    - We map dbt's incremental model `on_schema_change` setting to sqlmesh `on_destructive_change`
    - `on_schema_change` values are mapped to `on_destructive_change` as follows:
        - ignore -> WARN
        - append_new_columns -> WARN
        - fail -> ERROR
        - sync_all_columns -> ALLOW


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -72,8 +73,13 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
...
         },
-        support_coercing_compatible_types=True,
-    )
+        "support_coercing_compatible_types": True,
+    }
+    SCHEMA_DIFFER = SchemaDiffer(**SCHEMA_DIFFER_ARGS)
+
+    @classproperty
+    def schema_differ_args(cls) -> t.Dict[str, t.Any]:
+        return cls.SCHEMA_DIFFER_ARGS
```
eakmanrq, line 82: Why not just do `cls.SCHEMA_DIFFER_ARGS` instead of creating this property which then does that?

sqlmesh/core/model/kind.py
```diff
@@ -152,6 +152,26 @@ def __repr__(self) -> str:
         return str(self)
 
 
+class OnSchemaChange(str, Enum):
```
eakmanrq, line 155: Although I like the simplicity of `OnSchemaChange`, it is a bit misleading. It really is `OnDestructiveSchemaChange`. Also I don't currently intuitively link `OnSchemaChange` with `allow_destructive_models`. So I think this naming still needs some work. 

What about `OnDestructiveChange`? Or I'm thinking we want to find something better than "Destructive"?

sqlmesh/cli/main.py
```diff
@@ -303,6 +303,12 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     help="Create a plan for forward-only changes.",
     default=None,
 )
+@click.option(
+    "--allow-destructive-model",
```
izeigerman, line 307: Do we need a selector here? Shouldn't this be a global plan flag `--allow-descrtuctive`?

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -35,18 +35,23 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
...
+        "array_element_selector": "element",
+    }
+    SCHEMA_DIFFER = SchemaDiffer(**SCHEMA_DIFFER_ARGS)
 
     def __init__(self, *args: t.Any, **kwargs: t.Any):
         super().__init__(*args, **kwargs)
         self._spark: t.Optional[PySparkSession] = None
 
+    @classproperty
+    def schema_differ_args(cls) -> t.Dict[str, t.Any]:
```
izeigerman, line 52: Seems like this always just returns the `SCHEMA_DIFFER_ARGS`. Why do we need it?

sqlmesh/core/model/definition.py
```diff
@@ -1813,17 +1821,25 @@ def _create_model(
     dialect = dialect or ""
     physical_schema_override = physical_schema_override or {}
 
+    # Assumption: any specified default that is not a ModelMeta field is kind-specific.
```
izeigerman, line 1824: Seems like a dangerous assumption. Should we be more specific in our defaults and eg. have defaults per model kind?

sqlmesh/core/model/kind.py
```diff
@@ -579,14 +605,34 @@ def _model_kind_validator(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) ->
...
+            for k, v in (values.get("kind_specific_defaults") or {}).items()
+            if k in kind_type_fields and not k in props
+        }
+        props.update(user_defaults_to_set)
+
+        # OnSchemaChange: convert string or exp.Identifier to enum
+        on_schema_change = props.get("on_schema_change", None)
+        if on_schema_change and not isinstance(on_schema_change, OnSchemaChange):
+            props["on_schema_change"] = OnSchemaChange(
+                on_schema_change.this.upper()
```
izeigerman, line 631: Looks like something that should be handled in its own field validator.

sqlmesh/core/plan/builder.py
```diff
@@ -429,15 +442,35 @@ def _categorize_snapshots(
                         # previous version for non-seed models.
                         # New snapshots of seed models are considered non-breaking ones.
                         if not snapshot.is_seed:
+                            if (
```
izeigerman, line 445: This check seems to be repeated over and over again. Consider moving it inside `_validate_schema_change`

sqlmesh/core/engine_adapter/__init__.py
```diff
@@ -18,6 +18,7 @@
 from sqlmesh.core.engine_adapter.trino import TrinoEngineAdapter
 
 DIALECT_TO_ENGINE_ADAPTER = {
+    "hive": SparkEngineAdapter,
```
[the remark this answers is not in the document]
tobymao: yea this looks problematic to me, we should just say which engine adapter you want and not map from dialect, could do that in a follow up

sqlmesh/core/engine_adapter/base.py
```diff
@@ -137,6 +138,10 @@ def with_log_level(self, level: int) -> EngineAdapter:
 
         return adapter
 
+    @classproperty
+    def schema_differ_args(cls) -> t.Dict[str, t.Any]:
```
tobymao, line 142: why is there a property?

sqlmesh/cli/main.py
eakmanrq: I think selector seems correct since it requires the user to be intentional about what they are allowing. Even if it does require some extra work on their end I think of that as a feature rather than something to make easier.

sqlmesh/core/engine_adapter/base.py
treysp: We don't need it - removed

sqlmesh/core/engine_adapter/databricks.py
treysp: We don't - removed

sqlmesh/core/model/kind.py
treysp: Agreed - so I'll change both the enum name here and the kind option (`on_schema_change` --> `on_destructive_change`)?

sqlmesh/core/model/definition.py
treysp: I originally had a private field in the model defaults that listed which fields were kind-specific, peeled those off here, then matched them against all of a kind's fields in the kind validator. 

Downside of that approach is that if we expose a kind-specific default for any kind it would be set for all kinds that have that field. 

Doing it per kind seems like a better choice - it would set us up to be more granular in the future, such that we could expose defaults for only a subset of kinds.

sqlmesh/core/engine_adapter/__init__.py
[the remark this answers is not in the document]
treysp: We have the engine adapter available [in context where we instantiate `PlanBuilder`](https://github.com/TobikoData/sqlmesh/blob/aeb8878962a57fa6dbbceca1edc310246275b3a5/sqlmesh/core/context.py#L1054).

Should I just pass its SchemaDiffer in? Then we don't have to do any indirection at all.

sqlmesh/core/model/kind.py
```diff
@@ -288,6 +309,7 @@ def metadata_hash_values(self) -> t.List[t.Optional[str]]:
             *super().metadata_hash_values,
             str(self.batch_size) if self.batch_size is not None else None,
             str(self.forward_only),
+            str(self.on_schema_change),
```
tobymao, line 312: this is a breaking change, you're gonna need to upgrade sqlglot or have a migration

georgesittas: > you're gonna need to upgrade sqlglot

FYI sqlglot was bumped to v23.14.0 recently and this hasn't been released yet, see https://github.com/TobikoData/sqlmesh/compare/v0.94.0...main.

sqlmesh/core/plan/builder.py
```diff
@@ -94,6 +103,7 @@ def __init__(
         self._skip_backfill = skip_backfill
         self._is_dev = is_dev
         self._forward_only = forward_only
+        self._allow_destructive_models = set(allow_destructive_models or [])
```
tobymao, line 106: this code is not correct given that allow_destructive_models is an iterable, iterables are never falsey. so you need to check if it's None instead

sqlmesh/core/plan/builder.py
```diff
@@ -429,15 +442,35 @@ def _categorize_snapshots(
                         # previous version for non-seed models.
                         # New snapshots of seed models are considered non-breaking ones.
                         if not snapshot.is_seed:
+                            if (
+                                snapshot.is_model
```
tobymao, line 446: can this condition not be repeated as well?

sqlmesh/core/plan/builder.py
```diff
@@ -650,6 +709,85 @@ def _ensure_new_env_with_changes(self) -> None:
...
+        force_warn: bool = False,
+    ) -> None:
+        """Determine if a schema change requires a DROP operation"""
+        s_id_snapshot = t.cast(Snapshot, self._context_diff.snapshots.get(s_id))
+
+        if not s_id_snapshot.model.on_schema_change.is_ignore or force_warn:
+            schema_differ_args = getattr(
+                DIALECT_TO_ENGINE_ADAPTER[s_id_snapshot.model.dialect], "schema_differ_args"
+            )
+            schema_differ = SchemaDiffer(**schema_differ_args)
```
tobymao, line 732: why do we need to instantiate this here? isn't this already an attribute of the engine adapter?

sqlmesh/core/plan/builder.py
treysp: Thank you - should I go ahead and change it for `self._restate_models` below as well?

tobymao: yes

sqlmesh/core/plan/builder.py
```diff
@@ -650,6 +709,85 @@ def _ensure_new_env_with_changes(self) -> None:
...
+
+            subdag = [s_id, *upstream_snapshot_ids]
+
+            # if we already know that the change is destructive, we don't need to evaluate any more snapshots
+            warning_msg = f"PLAN TIME CHECK: Plan results in a destructive change to forward-only model '{s_id_snapshot.name}'s schema."
+            error_msg = f"{warning_msg} To allow this, change the model's `on_schema_change` setting to `warn` or `ignore` or include it in the plan's `--allow-destructive-model` option."
+
+            if any(self._snapshot_change_is_destructive.get(id.name, None) for id in subdag):
```
tobymao, line 740: you're using any a lot, is there another way we can do this or another data structure?

sqlmesh/core/model/kind.py
```diff
@@ -288,11 +328,12 @@ def metadata_hash_values(self) -> t.List[t.Optional[str]]:
             *super().metadata_hash_values,
             str(self.batch_size) if self.batch_size is not None else None,
             str(self.forward_only),
+            str(self.on_destructive_change),
```
izeigerman, line 331: Shouldn't this be a part of the `metadata_hash_values` in `_Incremental`? This way we don't need to duplicate this

sqlmesh/core/model/meta.py
```diff
@@ -53,6 +54,9 @@ class ModelMeta(_Node):
 
     dialect: str = ""
     name: str
+    on_destructive_change_: t.Optional[OnDestructiveChange] = Field(
```
izeigerman, line 57: Why do we have this both here and in the model kind?

treysp: There's two ways for a user to specify this value: in a model definition and in the model defaults.

If they specify in the definition, it's captured when we parse that definition and available to the kind constructor in the _model_kind_validator.

If they specify in the model defaults, this is the path for that value to get to the kind constructor in the _model_kind_validator.

sqlmesh/core/plan/builder.py
```diff
@@ -553,6 +583,22 @@ def _set_choice(
 
             is_forward_only_child = self._is_forward_only_change(child_s_id)
 
+            if is_forward_only_child:
+                # no DAG if manual categorization
+                if not dag:
+                    logger.info(
+                        f"Unable to evaluate model '{child_s_id.name}'s upstream parents for destructive schema changes at plan time."
```
tobymao, line 590: what does this error mean, and how is it actionable or informative?

sqlmesh/core/plan/builder.py
```diff
@@ -553,6 +583,22 @@ def _set_choice(
...
+                    )
+
+                self._validate_destructive_change(
+                    child_s_id,
+                    (
```
tobymao, line 595: is this parenthesis needed?

sqlmesh/core/plan/builder.py
```diff
@@ -553,6 +583,22 @@ def _set_choice(
...
+                        self._upstream_directly_modified(child_s_id, directly_modified, dag)
```
tobymao, line 596: _upstream_directly_modified is called 3x with the same argument, always with validate_destructive_change.

can we move this call to validate_destructive_change since it seems like it's required?

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+    def _upstream_directly_modified(
+        self, s_id: SnapshotId, directly_modified: t.Set[SnapshotId], dag: DAG[SnapshotId]
+    ) -> t.List[SnapshotId]:
+        return [
+            upstream_id for upstream_id in dag.upstream(s_id) if upstream_id in directly_modified
+        ]
+
+    def _validate_destructive_change(
+        self,
+        s_id: SnapshotId,
```
tobymao, line 731: maybe let's change this to take in directly_modified and dag, and have it compute upstream_snapshot_ids

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+        upstream_snapshot_ids: t.List[SnapshotId] = [],
+        force_check: bool = False,
+    ) -> None:
+        """Determine if a schema change requires a destructive operation"""
+        s_id_snapshot = t.cast(Snapshot, self._context_diff.snapshots.get(s_id))
```
tobymao, line 736: why is this called s_id_snapshot, and not just snapshot?

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+
+        if (
+            s_id_snapshot.is_model
+            and s_id_snapshot.name not in self._allow_destructive_models
+            and not s_id_snapshot.model.on_destructive_change.is_ignore
+        ):
+            warning_msg = f"PLAN TIME CHECK: Plan results in a destructive change to forward-only model '{s_id_snapshot.name}'s schema."
```
tobymao, line 743: these are not used more than once, can you put them inline on 748 and 749?

PlanError(f...)
warning(...)

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+            error_msg = f"{warning_msg} To allow this, change the model's `on_destructive_change` setting to `warn` or `ignore` or include it in the plan's `--allow-destructive-model` option."
+
+            def _raise_or_warn() -> None:
+                if s_id_snapshot.model.on_destructive_change.is_error:
+                    raise PlanError(error_msg)
+                logger.warning(warning_msg)
+
+            subdag = set([s_id, *upstream_snapshot_ids])
```
tobymao, line 751: subdag = {s_id, *upstream}

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+
+            # get already classified snapshots
+            for id in subdag:
+                if id.name in self._snapshot_change_is_destructive:
+                    subdag_to_process.remove(id)
+                    value = self._snapshot_change_is_destructive[id.name]
+                    if value:
+                        _raise_or_warn()
+                        return
+                    if value is None:
```
tobymao, line 763: i don't understand what value is? according to the name, it should be whether or not snapshot_change_is_destructive

but if value is None, why does that imply no_cols_to_types?

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+            warning_msg = f"PLAN TIME CHECK: Plan results in a destructive change to forward-only model '{s_id_snapshot.name}'s schema."
+            error_msg = f"{warning_msg} To allow this, change the model's `on_destructive_change` setting to `warn` or `ignore` or include it in the plan's `--allow-destructive-model` option."
+
+            def _raise_or_warn() -> None:
+                if s_id_snapshot.model.on_destructive_change.is_error:
+                    raise PlanError(error_msg)
+                logger.warning(warning_msg)
+
+            subdag = set([s_id, *upstream_snapshot_ids])
+            subdag_to_process = subdag.copy()
```
tobymao, line 752: i wonder if this method could be more succint something like

subdag = set()

for id in (s_id, *upstream):
  if ...:
    subdag.add(id)

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                        _raise_or_warn()
+                        return
+                    if value is None:
+                        subdag_no_cols_to_types.add(id)
+
+            # process unclassified snapshots
+            for id in subdag_to_process:
+                snapshot = t.cast(
+                    Snapshot,
+                    s_id_snapshot if id == s_id else self._context_diff.snapshots.get(id),
```
tobymao, line 770: this logic seems unnecessary, can just do

snapshot = self._context_diff.snapnshots.get(id)

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                )
+                new, old = self._context_diff.modified_snapshots[snapshot.name]
+
+                # we must know all columns_to_types to determine whether a change is destructive
+                old_columns_to_types = old.model.columns_to_types
+                new_columns_to_types = new.model.columns_to_types
+                if (
+                    not old_columns_to_types
```
tobymao, line 778: maybe let's change columns_to_types to accept optionals, then we don't have to make this check so verbose

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    or not columns_to_types_all_known(old_columns_to_types)
+                    or not new_columns_to_types
+                    or not columns_to_types_all_known(new_columns_to_types)
+                ):
+                    self._snapshot_change_is_destructive[id.name] = None
+                    subdag_no_cols_to_types.add(id)
+                    continue
+
+                schema_differ = self._engine_schema_differ
+                schema_diff = schema_differ.compare_columns(
```
tobymao, line 788: schema_differ is never used again, let's do self._engine_schema_differ.compare_columns(...)

sqlmesh/core/plan/builder.py
tobymao: either that or do columns_to_types or {}

then you don't need to cast on line 790 

 t.cast(t.Dict[str, exp.DataType], old_columns_to_types) this can be removed

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    new.model.name,
+                    t.cast(t.Dict[str, exp.DataType], old_columns_to_types),
+                    t.cast(t.Dict[str, exp.DataType], new_columns_to_types),
+                )
+
+                has_drop = any(
+                    [
```
tobymao, line 795: i don't think the [ is necessary, you can just pass a generator into any

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    ]
+                )
+                self._snapshot_change_is_destructive[id.name] = has_drop
+                if has_drop:
+                    _raise_or_warn()
+                    return
+
+            if subdag_no_cols_to_types:
+                logger.info(
+                    f"Unable to determine at plan time if changes cause a destructive schema change to model '{s_id.name}'."
```
tobymao, line 808: why is it unable to determine it? (because we don't have the types, how can they avoid this warning in the future)

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1522,3 +1551,30 @@ def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> I
...
+def _check_destructive_schema_change(
+    snapshot: Snapshot,
+    alter_expressions: t.List[exp.AlterTable],
+    allow_destructive_snapshots: t.Set[str],
+) -> None:
+    if (
+        snapshot.is_model
+        and snapshot.is_forward_only
+        and snapshot.name not in allow_destructive_snapshots
+        and not snapshot.model.on_destructive_change.is_ignore
```
tobymao, line 1565: i feel like i've seen these conditions chained together multiple times, maybe this means we should have a method on the snapshot that computes this similar logic?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1522,3 +1551,30 @@ def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> I
...
+        and any(
+            [
```
tobymao, line 1567: same

sqlmesh/migrations/v0050_add_on_destructive_change.py
```diff
@@ -0,0 +1,5 @@
+"""Adds the on_destructive_change setting to model kinds."""
```
tobymao, line 1: is this needed? a sqlglot bump could also achieve this

sqlmesh/core/plan/builder.py
treysp: Each check can return one of three values: drop present, drop not present, can't be determined. I'm representing those as `True`, `False`, and `None`, respectively.

Is that the wrong approach, or should I rename `snapshot_change_is_destructive` to reflect the potential `None` value?

sqlmesh/core/plan/builder.py
izeigerman: This is not a warning though

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    or not columns_to_types_all_known(new_columns_to_types)
+                ):
+                    self._snapshot_change_is_destructive[id.name] = None
+                    subdag_no_cols_to_types.add(id)
+                    continue
+
+                schema_differ = self._engine_schema_differ
+                schema_diff = schema_differ.compare_columns(
+                    new.model.name,
+                    t.cast(t.Dict[str, exp.DataType], old_columns_to_types),
```
izeigerman, line 790: Can we somehow get rid of this case here? Eg. `old_columns_to_types or {}`?

sqlmesh/core/plan/builder.py
tobymao: ok that's fine, but maybe instead of value, let's call it destructive

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    raise PlanError(error_msg)
+                logger.warning(warning_msg)
+
+            subdag = set([s_id, *upstream_snapshot_ids])
+            subdag_to_process = subdag.copy()
+            subdag_no_cols_to_types = set()
+
+            # get already classified snapshots
+            for id in subdag:
+                if id.name in self._snapshot_change_is_destructive:
```
tobymao, line 757: what's case where a snapshot is not in this dictionary?

sqlmesh/core/plan/builder.py
```diff
@@ -673,6 +719,95 @@ def _ensure_new_env_with_changes(self) -> None:
...
+                    subdag_to_process.remove(id)
+                    value = self._snapshot_change_is_destructive[id.name]
+                    if value:
+                        _raise_or_warn()
```
izeigerman, line 761: I'm a little confused. Don't we call `_raise_or_warn` when we set the key to the `_snapshot_change_is_destructive` on line 802? How do we end in this block? Do we warn more than once?

sqlmesh/core/plan/builder.py
tobymao: i think the other confusing thing is that there is a 4th state, a snapshot may not be in the dictionary

sqlmesh/core/plan/builder.py
tobymao: yes exactly, i just suggested this

sqlmesh/core/plan/builder.py
tobymao: then what is it?

sqlmesh/core/plan/builder.py
izeigerman: Never mind, I think I get it. `_raise_or_warn` always checks the target snapshot and not the one for which the `_snapshot_change_is_destructive` was set.

sqlmesh/core/model/meta.py
izeigerman: I feel like we need to find a different way to propagate the default. The current way looks more like a workaround.

sqlmesh/core/plan/builder.py
treysp: If this is the first time we've classified any plan snapshot, the dictionary will be empty. As we classify snapshots their values get added. 

Snapshots may have many of the same models in their upstream. If one of those models has already been classified, we grab the value from this dict instead of reclassifying it.

sqlmesh/core/plan/builder.py
treysp: Right - the approach is "as soon as we determine that any of the target snapshot or an upstream has a destructive change, _raise_or_warn() for the target snapshot instead of looking at more snapshots"

sqlmesh/core/plan/builder.py
treysp: I embedded the warning message in the error message to prevent writing the same text in both of them. With that approach, at least the warning will have to be assigned (then used inline in the error message).

Should I repeat the text instead of embedding?

sqlmesh/core/model/meta.py
treysp: I think it's fair to say it's a workaround for pydantic field validators only having access to fields prior to them in the pydantic model def.

sqlmesh/core/snapshot/evaluator.py
treysp: Moved it to a snapshot staticmethod

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1522,3 +1551,20 @@ def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> I
...
+    allow_destructive_snapshots: t.Set[str],
+) -> None:
+    if snapshot.do_destructive_check(
+        allow_destructive_snapshots, True
+    ) and snapshot.has_drop_alteration(alter_expressions):
+        warning_msg = f"RUN TIME CHECK: Plan results in a destructive change to forward-only table '{snapshot.name}'s schema."
+        if snapshot.model.on_destructive_change.is_warn:
+            logger.warning(warning_msg)
+            return
+        raise RuntimeError(
```
izeigerman, line 1568: We should probably raise `SQLMeshError` instead

sqlmesh/core/snapshot/definition.py
```diff
@@ -553,6 +553,14 @@ def hydrate_with_intervals_by_identifier(
 
         return result
 
+    @staticmethod
+    def has_drop_alteration(alter_expressions: t.List[exp.AlterTable]) -> bool:
```
izeigerman, line 557: Why is this a part of the snapshot definition? I don't think this belongs here

sqlmesh/core/snapshot/definition.py
```diff
@@ -927,6 +935,16 @@ def to_end_date(end: int, unit: IntervalUnit) -> TimeLike:
...
         )
 
+    def do_destructive_check(
+        self, allow_destructive_snapshots: t.Set[str], check_forward_change: bool = False
+    ) -> bool:
+        return (
+            self.is_model
+            and not self.model.on_destructive_change.is_ignore
+            and self.name not in allow_destructive_snapshots
+            and (self.is_forward_only if check_forward_change else True)
```
izeigerman, line 945: Why do we care about `check_forward_change`?

sqlmesh/core/snapshot/definition.py
treysp: I wasn't sure where to put it

sqlmesh/core/snapshot/definition.py
treysp: A snapshot needs to be checked if either:
1. It's a forward-only model with on_destructive_change != ignore
2. It's a model in a forward-only plan with on_destructive_change != ignore

We determine 1 in both the plantime and runtime checks.

We can determine 2 In the plantime check because we know if we're in a forward-only plan. 

In the runtime check, we do not know if we're In a forward-only plan but can infer 2 if change==FORWARD_ONLY. The `check_forward_change` arg lets us use this method for both plantime and runtime checks.

sqlmesh/dbt/model.py
```diff
@@ -181,16 +182,26 @@ def model_kind(self, context: DbtContext) -> ModelKind:
...
         target = context.target
         materialization = self.model_materialization
+
+        # args common to all sqlmesh incremental kinds, regardless of materialization
+        incremental_kind_kwargs: t.Dict[str, t.Any] = {}
+        if self.on_schema_change:
+            incremental_kind_kwargs["on_destructive_change"] = (
+                OnDestructiveChange.ALLOW
+                if self.on_schema_change == "sync_all_columns"
+                else OnDestructiveChange.ERROR
```
izeigerman, line 192: This doesn't look right to me. What informed this choice?

sqlmesh/core/snapshot/definition.py
izeigerman: There's nothing specific to snapshot in this method. Shouldn't this be a part of the `SchemaDiffer`?

sqlmesh/core/snapshot/definition.py
```diff
@@ -927,6 +935,26 @@ def to_end_date(end: int, unit: IntervalUnit) -> TimeLike:
...
 
+    def needs_destructive_check(
+        self,
+        allow_destructive_snapshots: t.Set[str],
+        is_forward_only_plan: bool = False,
+        runtime_check: bool = False,
+    ) -> bool:
+        return (
+            self.is_model
+            # is a forward-only change happening?
```
izeigerman, line 946: I think this method is too overloaded and hard to digest. We should only keep the logic that is shared between plantime and runtime here and do plan / run specific stuff upstream.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1549,3 +1578,20 @@ def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> I
...
+
+def _check_destructive_schema_change(
+    snapshot: Snapshot,
+    alter_expressions: t.List[exp.AlterTable],
+    allow_destructive_snapshots: t.Set[str],
+) -> None:
+    if snapshot.needs_destructive_check(
+        allow_destructive_snapshots, runtime_check=True
+    ) and snapshot.has_drop_alteration(alter_expressions):
+        warning_msg = f"RUN TIME CHECK: Plan results in a destructive change to forward-only table '{snapshot.name}'s schema."
```
izeigerman, line 1591: `RUN TIME CHECK` - where does this message prefix come from? Is this some new convention?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -652,7 +661,13 @@ def _create_snapshot(
                 )
                 try:
                     self.adapter.clone_table(target_table_name, snapshot.table_name(), replace=True)
-                    self.adapter.alter_table(target_table_name, tmp_table_name)
+                    alter_expressions = self.adapter.get_alter_expressions(
```
izeigerman, line 664: Why do we check this here? This only happens in dev, why should we fail in dev?

sqlmesh/core/plan/builder.py
```diff
@@ -425,6 +436,64 @@ def _adjust_new_snapshot_intervals(self) -> None:
             if new.is_forward_only:
                 new.dev_intervals = new.intervals.copy()
 
+    def _check_destructive_changes(
```
izeigerman, line 439: This looks so much better now!

sqlmesh/core/snapshot/evaluator.py
treysp: Oh yeah, i forgot that cloning only happens in dev - is the `snapshot.previous_versions` clause how this conditions on dev env?

```
            if (
                snapshot.is_forward_only
                and snapshot.is_materialized
                and snapshot.previous_versions
                and self.adapter.SUPPORTS_CLONING
            ):
```

sqlmesh/dbt/model.py
treysp: Responded in slack thread https://tobikodata.slack.com/archives/C04DTU8RF4J/p1716262698103299

sqlmesh/core/plan/builder.py
```diff
@@ -425,6 +436,64 @@ def _adjust_new_snapshot_intervals(self) -> None:
...
+        snapshot_is_destructive = {}
+        for s_id in dag:
+            # Only process snapshots that are modified by the plan
+            if s_id.name in self._context_diff.modified_snapshots:
+                snapshot = self._context_diff.snapshots[s_id]
+
+                if snapshot and snapshot.is_model:
+                    # should we raise/warn if this snapshot has/inherits a destructive change?
+                    should_raise_or_warn = (
+                        snapshot.model.forward_only or self._forward_only
```
izeigerman, line 460: Wouldn't this fail if the snapshot is not a model? We should probably check `needs_destructive_check` first

sqlmesh/core/model/kind.py
```diff
@@ -152,8 +152,44 @@ def __repr__(self) -> str:
...
+
+
+on_destructive_change_validator = field_validator("on_destructive_change", mode="before")(
+    _on_destructive_change_validator
+)
+
+
 class _ModelKind(PydanticModel, ModelKindMixin):
     name: ModelKindName
+    on_destructive_change: OnDestructiveChange = OnDestructiveChange.ALLOW
```
izeigerman, line 190: I think I've removed it in my PR. Why do we need this here if the property only applies to incremental models?

sqlmesh/core/plan/builder.py
treysp: Won't the `if snapshot and snapshot.is_model:` ensure only models hit this?

izeigerman: Ah, shoot, missed it. Thanks for checking

izeigerman (approved):

PR #2579, opened 2024-05-07
Feat: allow user-supplied queries to generate unit test data

### LINKED ISSUE #2542: Invalid SQL generated for the ROW type in Trino unit test fixtures
Here is another Trino edge case which im not sure if it's SQLMesh or SQLGlot.

Given a model like so:
```
MODEL (
    name test.test_model,
    kind FULL
);

select
    max(meta.day) as day,
    meta.type as type
from test.metadata
group by 1, 2
```

And a test like so:
```
test_correct_grouping:
  model: test.test_model
  inputs:
    test.metadata:
      columns:
        meta: ROW(day date, type varchar)
      rows:
        - meta:
            day: 2024-03-30
            type: foo
        - meta:
            day: 2024-03-31
            type: bar
        - meta:
            day: 2024-04-01
            type: baz
  outputs:
    query:
      rows:
        - day: 2024-04-01
          type: baz
```

When running (against a `default_test_connection` pointed at Trino), it fails:

```
$ sqlmesh --debug test tests/test_model.yaml 
E
ERROR: test_sqlmesh (tests/test_model.yaml)
trino.exceptions.TrinoUserError: TrinoUserError(type=USER_ERROR, name=MISMATCHED_COLUMN_ALIASES, message="line 1:147: Column alias list has 1 entries but 't' has 2 columns available", query_id=20240430_192625_00091_yysrm)
```

The reason for this failure is the SQL that SQLMesh generates to create the test fixture. From the logs, it tries to execute:
```
CREATE OR REPLACE VIEW datalake.sqlmesh_test_g087nhkb."datalake__test__metadata" AS
SELECT CAST(meta AS ROW(day DATE, type VARCHAR)) AS meta
FROM (
  VALUES
    (CAST(ROW(CAST('2024-03-30' AS DATE), 'foo') AS ROW(day DATE, type VARCHAR))),
    (CAST(ROW(CAST('2024-03-31' AS DATE), 'bar') AS ROW(day DATE, type VARCHAR))),
    (CAST(ROW(CAST('2024-04-01' AS DATE), 'baz') AS ROW(day DATE, type VARCHAR)))
) AS t(meta)
```

The problem is that when `CAST(ROW(CAST('2024-03-30' AS DATE), 'foo') AS ROW(day DATE, type VARCHAR))` is put inside `VALUES()`, Trino unpacks the `ROW` into two columns. This can be shown like so:

Running `SELECT CAST(ROW(CAST('2024-03-30' AS DATE), 'foo') AS ROW(day DATE, type VARCHAR))` correctly produces a `ROW`:

| _col0                                         |
|------------------------------------------|
|{day=2024-04-30, type=foo} |

However, wrapping it in `VALUES` by running `SELECT * FROM ( VALUES ( CAST(ROW(CAST('2024-03-30' AS DATE), 'foo') AS ROW(day DATE, type VARCHAR)) ) )` "helpfully" *unpacks* the `ROW` into multiple columns:

| _col0             | _col1 |
|------------------|-------   |
| 2024-04-30 | foo     |

This is what causes the error: `Column alias list has 1 entries but 't' has 2 columns available`

The correct syntax re-assembles the `ROW` type from the top-level columns something like:
```
SELECT CAST((t.col1, t.col2) AS ROW(day DATE, type VARCHAR)) AS meta
FROM (
  VALUES
    (CAST(ROW(CAST('2024-03-30' AS DATE), 'foo') AS ROW(day DATE, type VARCHAR))),
    (CAST(ROW(CAST('2024-03-31' AS DATE), 'bar') AS ROW(day DATE, type VARCHAR))),
    (CAST(ROW(CAST('2024-04-01' AS DATE), 'baz') AS ROW(day DATE, type VARCHAR)))
) AS t(col1, col2)
```

Alternatively - instead of SQLMesh trying to generate a query from the yaml to produce a test fixture, maybe we could add a feature that allows the user to supply their own `SELECT` query to produce the data?

georgesittas:
Fixes #2542, TL;DR this PR adds a new format for specifying unit test fixture/expected data: SQL. :-)

- [x] Discuss identifier normalization decisions. Initial approach was to execute user-supplied queries without normalizing or quoting them, but this led to issues in the `create_view` method because the fully-qualified view name in the `CREATE VIEW <name> AS ...` statement won't be quoted if we set `quote_identifiers=False`, which may cause problems.

@erindru I ran your example against Trino and it seemed to work fine. I feel like the flexibility this feature provides will allow users to bypass conversion issues in unit tests pretty easily. Thanks for the idea!

EDIT: will make a followup PR to update the unit test documentation.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -25,7 +25,10 @@ from sqlmesh.utils import UniqueKeyDict, random_id, type_is_known, yaml
 from sqlmesh.utils.date import pandas_timestamp_to_pydatetime
 from sqlmesh.utils.errors import ConfigError, TestError
 
-Row = t.Dict[str, t.Any]
+if t.TYPE_CHECKING:
+    from sqlglot.dialects.dialect import DialectType
+
+    Row = t.Dict[str, t.Any]
 
 
 class ModelTest(unittest.TestCase):
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -121,10 +121,10 @@ class ModelTest(unittest.TestCase):
                     len(known_columns_to_types) == len(inferred_columns_to_types)
                 )
 
-            # Types specified in the test take precedence over the corresponding inferred ones
+            # Types specified in the test will override the corresponding inferred ones
             known_columns_to_types.update(values.get("columns", {}))
 
-            rows = values["rows"]
+            rows = values.get("rows")
             if not all_types_are_known and rows:
                 for col, value in rows[0].items():
                     if col not in known_columns_to_types:
@@ -144,9 +144,13 @@ class ModelTest(unittest.TestCase):
 
                         known_columns_to_types[col] = v_type
 
-            df = _create_df(rows, columns=known_columns_to_types)
+            if rows is None:
+                query_or_df = values["query"]
+            else:
+                query_or_df = self._create_df(values, columns=known_columns_to_types)
+
             self.engine_adapter.create_view(
-                self._test_fixture_table(name), df, known_columns_to_types
+                self._test_fixture_table(name), query_or_df, known_columns_to_types
             )
 
     def tearDown(self) -> None:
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -310,17 +314,38 @@ def _validate_and_normalize_test(self) -> None:
...
+            if query is not None:
+                if rows is not None:
+                    _raise_error(
+                        f"Invalid test, cannot set both 'query' and 'rows' for '{name}'", self.path
+                    )
+
+                # We parse the user-supplied query using the testing adapter dialect, but we
+                # normalize its identifiers according to the model's dialect, so that, e.g.,
+                # the projection names match those in its `columns_to_types` field
+                values["query"] = normalize_identifiers(
```
tobymao, line 337: this is confusing, but i think it makes sense, we need to normalize by the model, that's always true or else references will be wrong.

the parsing should be supplied by the actual test engine so there are no nuanences with transpilation

tobymao (approved):

sqlmesh/core/test/definition.py
georgesittas: both points are exactly what I had in mind 👍

erindru:
Allowing users to supply arbitrary SQL to produce a resultset for tests is a nice escape hatch imo.

Of course in an ideal world SQLGlot would be able to generate the correct query 100% of the time, but the variations in SQL syntax and obscure data types between engines are essentially infinite.

I think allowing users to un-block themselves by supplying the query manually when they encounter an edge case will go a long way to reducing friction

PR #2580, opened 2024-05-08
chore: upgrade dev spark to fix integration tests

tobymao:


georgesittas:
What was the issue?

tobymao:
@georgesittas pandas 2 and spark 3.4 have issues, previously, i think we were pinned to pandas 1

PR #2584, opened 2024-05-08
Chore: replace autoflake+black+isort with ruff

georgesittas:
Ruff rules for reference: https://docs.astral.sh/ruff/rules/#error-e


sqlmesh/core/engine_adapter/trino.py
```diff
@@ -176,11 +176,16 @@ def _df_to_source_queries(
         batch_size: int,
         target_table: TableName,
     ) -> t.List[SourceQuery]:
+        assert isinstance(df, pd.DataFrame)
```
georgesittas, line 179: I added this because `df` is of type `DF`, which includes Spark dataframes, and for these `dtypes` [returns](https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.sql.DataFrame.dtypes.html#pyspark.sql.DataFrame.dtypes) a `list` which is not indexable by `str` (`column`). Should be fine, because `super()._df_to_source_queries` also makes this assertion.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -543,14 +544,20 @@ def apply(query_or_df: QueryOrDF, index: int = 0) -> None:
...
-                            limit,
-                            execute(exp.select(existing_limit.expression)).rows[0][0],
-                        )
-                return query_or_df.head(limit) if hasattr(query_or_df, "head") else self.adapter._fetch_native_df(query_or_df.limit(limit))  # type: ignore
+                if isinstance(query_or_df, PyDataFrame):
+                    return query_or_df.limit(limit)
+                if isinstance(query_or_df, pd.DataFrame):
+                    return query_or_df.head(limit)
+
+                assert isinstance(query_or_df, exp.Query)
```
georgesittas, line 552: I think there was a bug here previously, which was silenced due to `type: ignore`. When `query_or_df` is a Spark dataframe, `head` will [return](https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.sql.DataFrame.head.html#pyspark.sql.DataFrame.head) a `Row` or a list of `Row`s, i.e. _not_ a `DF`, and so invariants may break. E.g. in https://github.com/TobikoData/sqlmesh/blob/9e19b98338ca97e6aefd1525256d325371bfee9f/sqlmesh/core/snapshot/evaluator.py#L173 we'd return a different object from what the method's type hint says.

georgesittas: The to ensure we have a `Query` happened because the type union also contains `DerivedTable` for some reason (maybe to take CTEs and / or UDTFs into account? not sure..). Only `Query` expressions support the `limit` builder, so if for whatever reason `query_or_df` was not a `Query`, the `.limit(...)` would crash.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1279,7 +1279,7 @@ def _snapshot_name_version_filter(
         batches = self._snapshot_batches(name_versions)
 
         if not name_versions:
-            return exp.false()
+            yield exp.false()
```
georgesittas, line 1282: Mypy treated this as an "unexpected return value", because the return type of the method is an iterator (due to the `yield`s). Are there any weird semantics in place that I may have missed here? cc @tobymao.

georgesittas (commented):
Pointed out a few things I found interesting, to help with the review.

setup.py
```diff
@@ -59,19 +59,17 @@
...
             "cryptography~=42.0.4",
             "dbt-core",
             "dbt-duckdb>=1.7.1",
             "Faker",
             "google-auth",
             "google-cloud-bigquery",
             "google-cloud-bigquery-storage",
-            "isort==5.10.1",
-            "mypy~=1.8.0",
+            "mypy~=1.10.0",
```
georgesittas, line 72: Please also note the mypy bump. Let me know if we'd pinned to 1.8.0 for a reason.

tobymao (approved):

sqlmesh/core/audit/__init__.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: F401
```
tobymao, line 1: why don't we just always ignore f401, what is this?

sqlmesh/core/state_sync/engine_adapter.py
tobymao: ah yea should be good i think

tobymao (approved):

sqlmesh/core/audit/__init__.py
georgesittas: We don't want to because it checks [unused imports](https://docs.astral.sh/ruff/rules/unused-import/#legend). The `__init__` files though export names, so they're not really "unused". We also [do this in sqlglot](https://github.com/tobymao/sqlglot/blob/main/sqlglot/__init__.py).

There may be an issue with _how_ we're exporting names. IIRC there was a mypy issue in the past that was resolved after doing `import foo as foo` vs `import foo`, so maybe adding these aliases will allow us to get rid of the `noqa`s.

tobymao: yea let's do import foo as foo right

georgesittas: yep planning to do another pass to verify this & fix if it's the culprit

sqlmesh/__init__.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: E402
```
georgesittas, line 1: I used this because we need `extend_sqlglot()` to happen right after it's imported to kick off the dialect patching. Otherwise we'd get a [module-import-not-at-top-of-file](https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/#module-import-not-at-top-of-file-e402) error for the subsequent imports.

PR #2589, opened 2024-05-09
tobiko-app image size reduced from 3GB to 1.3GB

singulart:
Hey, Team! Came across your repository and thought I could help in optimizing the Docker image size.

### DIFF
--- a/Dockerfile.api
+++ b/Dockerfile.api
@@ -1,8 +1,15 @@
-FROM python:3.11
-
-WORKDIR /sqlmesh
-
+FROM python:3.11-slim-bullseye AS venv-image
+RUN apt-get update && apt-get -y install libpq-dev gcc
 COPY setup.py setup.py
 COPY sqlmesh/_version.py sqlmesh/_version.py
+ENV VIRTUAL_ENV=/.venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN python3 -m venv $VIRTUAL_ENV
+RUN pip3 install -e .[dev,web]
 
-RUN pip install -e .[dev,web]
+FROM python:3.11-slim-bullseye as app-image
+WORKDIR /app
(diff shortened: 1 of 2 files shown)


Dockerfile.api
```diff
@@ -1,8 +1,15 @@
...
+RUN apt-get update && apt-get -y install libpq-dev gcc
 COPY setup.py setup.py
 COPY sqlmesh/_version.py sqlmesh/_version.py
+ENV VIRTUAL_ENV=/.venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN python3 -m venv $VIRTUAL_ENV
+RUN pip3 install -e .[dev,web]
 
-RUN pip install -e .[dev,web]
+FROM python:3.11-slim-bullseye as app-image
```
tobymao, line 10: why are you referencing python 3.11 both times? can you explain a bit about what this pr is doing?

singulart: The Dockerfile was updated the following way:
1) The "slimmer" base image `python:3.11-slim-bullseye` was used.
2) Multi-stage build approach was used: https://docs.docker.com/build/building/multi-stage

The #2 is why the image is being referenced twice. Hope that helps! Let me know if you need more clarity on this ;)

tobymao:
@singulart thanks for the PR. since this docker image is used for development and never shipped to the user, I don't think we need to add complexity to reduce the size of the image.

so for now i'm going to close this pr.

singulart:
> @singulart thanks for the PR. since this docker image is used for development and never shipped to the user, I don't think we need to add complexity to reduce the size of the image.
> 
> so for now i'm going to close this pr.

Sure @tobymao! Should the Docker image size become an issue going forward, you know how to solve it. Good luck!

tobymao:
thanks so much

Issue #2335, opened 2024-03-23
Support session properties in model defaults 


izeigerman:
When the value is set in a specific model definition, the overriding should take place on a per key basis.

izeigerman:
Fixed in https://github.com/TobikoData/sqlmesh/pull/2596

PR #2596, opened 2024-05-11
#2335 Add support for session properties in model defaults

ciprian-cgr:
This should allow you to set session_properties on the project configuration which will then be propagated on the models. If custom properties are also set on the specific model, those will not be overwritten by the default ones. It might need some more work when adding the default props to handle quoted/unquoted identifiers better; right now, it sets them by default as quoted=False.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1897,7 +1897,7 @@ def _resolve_custom_session_properties(
                 session_properties[k] = exp.Literal.string(k).eq(v)
 
     if session_properties:
-        return exp.Tuple(expressions=session_properties.values())
+        return exp.Tuple(expressions=list(session_properties.values()))
 
     return None
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1877,21 +1877,19 @@ def _split_sql_model_statements(
     return query, expressions[:pos], expressions[pos + 1 :]
 
 
-def _resolve_custom_session_params(defaults: t.Dict[str, t.Any], **kwargs: t.Any): 
-    if  kwargs.get("session_properties") and defaults and defaults.get("session_properties"):
+def _resolve_custom_session_params(
+    defaults: t.Dict[str, t.Any] | None, **kwargs: t.Dict[str, t.Any]
+) -> None:
+    if kwargs.get("session_properties") and defaults and defaults.get("session_properties"):
         session_properties = kwargs["session_properties"]
         session_props = {expr.this.name for expr in session_properties}
         for k, v in defaults["session_properties"].items():
             if k not in session_props:
                 new_eq = exp.EQ(
-                    this=exp.Literal(this=k),
-                    expression=exp.Column(
-                    this=exp.Identifier(this=v)
-                    )
+                    this=exp.Literal(this=k), expression=exp.Column(this=exp.Identifier(this=v))
                 )
                 session_properties.expressions.append(new_eq)
         kwargs["session_properties"] = session_properties
-
 
 
 def _validate_model_fields(klass: t.Type[_Model], provided_fields: t.Set[str], path: Path) -> None:
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1812,7 +1812,10 @@ def _create_model(
     **kwargs: t.Any,
 ) -> Model:
     _validate_model_fields(klass, {"name", *kwargs} - {"grain", "table_properties"}, path)
-    _resolve_custom_session_params(defaults, **kwargs)
+
+    kwargs["session_properties"] = _resolve_custom_session_properties(
+        defaults, kwargs.get("session_properties")
+    )
 
     dialect = dialect or ""
     physical_schema_override = physical_schema_override or {}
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1876,6 +1877,21 @@ def _split_sql_model_statements(
     return query, expressions[:pos], expressions[pos + 1 :]
 
 
+def _resolve_custom_session_params(
```
izeigerman, line 1880: If the purpose of the function is to only resolve session properties it feels somewhat counterintuitive to require 2 dicts in arguments and mutation one of them. How about the following signature:
```
def _resolve_custom_session_params(
    default: t.Optional[t.Dict[str, t.Any]], provided: t.Optional[exp.Expression]
) -> t.Optional[exp.Expression]:
    if default is not None and provided is not None:
        ...

```

sqlmesh/core/model/definition.py
```diff
@@ -1876,6 +1877,21 @@ def _split_sql_model_statements(
...
+    defaults: t.Dict[str, t.Any] | None, **kwargs: t.Dict[str, t.Any]
+) -> None:
+    if kwargs.get("session_properties") and defaults and defaults.get("session_properties"):
```
izeigerman, line 1883: So what happens if there session properties in defaults but not in model definitions? Don't we still need to convert the dict into a sqlglot expression?

sqlmesh/core/model/definition.py
```diff
@@ -1876,6 +1877,21 @@ def _split_sql_model_statements(
...
+        session_properties = kwargs["session_properties"]
+        session_props = {expr.this.name for expr in session_properties}
+        for k, v in defaults["session_properties"].items():
+            if k not in session_props:
+                new_eq = exp.EQ(
+                    this=exp.Literal(this=k), expression=exp.Column(this=exp.Identifier(this=v))
```
izeigerman, line 1889: The value can also be a sqlglot expression, that's why I'd suggest doing:
```
this=exp.Literal(this=k), expression=exp.convert(v)
```
Also CC: @georgesittas @tobymao to correct me if I'm wrong.

izeigerman:
Make sure to run `make style` to run linters and code formatter

tests/core/test_model.py
```diff
@@ -2698,6 +2698,47 @@ def test_model_table_properties_conflicts() -> None:
         sql_model.physical_properties
 
 
+def test_session_properties_on_model_and_project(sushi_context):
```
izeigerman, line 2701: Let's also add a test when there are only defaults but nothing is set in the model definition.

sqlmesh/core/model/definition.py
```diff
@@ -1876,6 +1880,28 @@ def _split_sql_model_statements(
...
+    provided: t.Optional[exp.Expression] | t.Optional[t.Dict[str, t.Any]],
+) -> t.Optional[exp.Expression]:
+    if provided is not None and isinstance(provided, dict):
+        session_properties = {k: exp.Literal.string(k).eq(v) for k, v in provided.items()}
+    else:
+        session_properties = (
+            {expr.this.name: expr for expr in provided} if provided is not None else {}
+        )
+
+    if defaults is not None and defaults.get("session_properties") is not None:
```
izeigerman, line 1894: Nit:
```
if defaults and "session_properties" in defaults:
```

sqlmesh/core/model/definition.py
```diff
@@ -1876,6 +1880,28 @@ def _split_sql_model_statements(
     return query, expressions[:pos], expressions[pos + 1 :]
 
 
+def _resolve_custom_session_properties(
+    defaults: t.Optional[t.Dict[str, t.Any]],
+    provided: t.Optional[exp.Expression] | t.Optional[t.Dict[str, t.Any]],
+) -> t.Optional[exp.Expression]:
+    if provided is not None and isinstance(provided, dict):
```
izeigerman, line 1887: I'm pretty sure you just need `isinstance(provided, dict)` since `isinstance(None, dict)` returns false. Let me know if I'm wrong.

tests/core/test_model.py
```diff
@@ -2698,6 +2698,76 @@ def test_model_table_properties_conflicts() -> None:
...
+        "spark.executor.memory": "1G",
+        "some_bool": True,
+        "some_float": 0.1,
+        "quoted_identifier": exp.column("quoted identifier", quoted=True),
+        "unquoted_identifier": exp.column("unquoted_identifier", quoted=False),
+        "project_level_property": "project_property",
+    }
+
+
+def test_project_level_session_properties(sushi_context):
```
izeigerman, line 2742: Thanks for adding this test! Since you caught an edge case which involved a Python model would it make sense to add a test which tests specifically the `dict` route?

sqlmesh/core/model/definition.py
ciprian-cgr: Indeed, I thought it might check for an empty dict as well, though that won't cause any issue as it resolved to an empty dict anyway.

sqlmesh/core/model/definition.py
ciprian-cgr: yes, that was my preference too, I just thought you might have some convention I'm not aware of based on this comment 
image

izeigerman (approved):

PR #2601, opened 2024-05-13
Chore: Support dbt 1.8

izeigerman:


sqlmesh/dbt/manifest.py
```diff
@@ -256,6 +256,7 @@ def _load_manifest(self) -> Manifest:
             profiles_dir=str(self.profiles_path),
             target=self.target.name,
             macro_debugging=False,
+            REQUIRE_RESOURCE_NAMES_WITHOUT_SPACES=True,
```
georgesittas, line 259: I assume this is backwards compatible?

crericha (approved):

georgesittas (approved):

sqlmesh/dbt/manifest.py
izeigerman: That's the expectation, yes. If the build passes that would be the confirmation of that.

PR #2602, opened 2024-05-13
Feat: Enable anonymized analytics for the SQLMesh core

izeigerman:
We prioritize our development work based on the needs of SQLMesh users. Some users share their needs via our Slack or Github communities, but many do not. We have added some simple anonymized usage information (telemetry) to SQLMesh to ensure the needs of all users are heard.

All information is anonymized with hash functions, so we could not link data to a specific company, user, or project even if we wanted to (which we don't!). No information is related to credentials or authentication.

We collect anonymized information about SQLMesh project complexity and usage - for example, number of models, count of model kinds, project load time, whether an error occurred during a plan/run (no stacktraces or error message), and names (but not values) of the arguments passed to CLI commands.

The collection of anonymized usage information can be disabled with these methods:

- Set the root `disable_anonymized_analytics: true` key in the SQLMesh project configuration file
- Execute SQLMesh commands with an environment variable `SQLMESH__DISABLE_ANONYMIZED_ANALYTICS` set to `1`, `true`, `t`, `yes`, or `y`

### DIFF
--- a/sqlmesh/core/analytics/__init__.py
+++ b/sqlmesh/core/analytics/__init__.py
@@ -13,8 +13,7 @@ from sqlmesh.utils import str_to_bool
 def init_collector() -> AnalyticsCollector:
     dispatcher = (
         NoopEventDispatcher()
-        if "DO_NOT_TRACK" in os.environ
-        or str_to_bool(os.getenv("SQLMESH__DISABLE_ANONYMIZED_ANALYTICS", "false"))
+        if str_to_bool(os.getenv("SQLMESH__DISABLE_ANONYMIZED_ANALYTICS", "false"))
         else AsyncEventDispatcher()
     )
     return AnalyticsCollector(dispatcher=dispatcher)
--- a/sqlmesh/core/analytics/__init__.py
+++ b/sqlmesh/core/analytics/__init__.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+import atexit
+import os
+import typing as t
+from functools import wraps
+
+from sqlmesh.core.analytics.collector import AnalyticsCollector
+from sqlmesh.core.analytics.dispatcher import AsyncEventDispatcher, NoopEventDispatcher
+from sqlmesh.utils import str_to_bool
+
+
+def init_collector() -> AnalyticsCollector:
+    dispatcher = (
+        NoopEventDispatcher()
+        if "DO_NOT_TRACK" in os.environ
+        or str_to_bool(os.getenv("SQLMESH__DISABLE_ANONYMIZED_ANALYTICS", "false"))
+        else AsyncEventDispatcher()
+    )
+    return AnalyticsCollector(dispatcher=dispatcher)
+
+
+collector = init_collector()
+
+
+atexit.register(collector.shutdown, flush=True)
+
+
+def disable_analytics() -> None:
+    global collector
+    collector = AnalyticsCollector(dispatcher=NoopEventDispatcher())
+
+
+def cli_analytics(func: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
+    @wraps(func)
+    def wrapper(*args: t.List[t.Any], **kwargs: t.Any) -> t.Any:
+        import click
+        from click.core import ParameterSource
+
+        arg_keys = set()
+        command_chain = []
+
+        cli_context: click.Context = click.get_current_context()
+        cli_context_cursor: t.Optional[click.Context] = cli_context
+        while cli_context_cursor:
+            arg_keys |= {
+                k
+                for k, source in cli_context_cursor._parameter_source.items()
+                if source == ParameterSource.COMMANDLINE
+            }
+            command_chain.append(cli_context_cursor.info_name)
+            cli_context_cursor = cli_context_cursor.parent
+
+        command_name, *parent_command_names = command_chain
+
+        common_context: t.Dict[str, t.Any] = {
+            "command_name": command_name,
+            "command_args": arg_keys,
+            "parent_command_names": parent_command_names,
+        }
+
+        if "github" in parent_command_names:
+            cicd_bot_config = None
+            github_controller = cli_context.obj.get("github")
+            if github_controller:
+                cicd_bot_config = github_controller._context.config.cicd_bot
+            collector.on_cicd_command(**common_context, cicd_bot_config=cicd_bot_config)
+        else:
+            collector.on_cli_command(**common_context)
+
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def python_api_analytics(func: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
+    @wraps(func)
+    def wrapper(*args: t.List[t.Any], **kwargs: t.Any) -> t.Any:
+        import inspect
+
+        from sqlmesh import magics
+
+        should_log = True
+        for frame in inspect.stack():
+            if "click/" in frame.filename or frame.filename == magics.__file__:
+                # Magics and CLI are reported separately.
+                should_log = False
+                break
+
+        if should_log:
+            collector.on_python_api_command(command_name=func.__name__, command_args=kwargs)
+
+        return func(*args, **kwargs)
+
+    return wrapper
--- a/sqlmesh/core/analytics/__init__.py
+++ b/sqlmesh/core/analytics/__init__.py
@@ -27,6 +27,7 @@ atexit.register(collector.shutdown, flush=True)
 
 def disable_analytics() -> None:
     global collector
+    collector.shutdown(flush=False)
     collector = AnalyticsCollector(dispatcher=NoopEventDispatcher())
 
 
--- a/sqlmesh/core/analytics/collector.py
+++ b/sqlmesh/core/analytics/collector.py
@@ -10,7 +10,7 @@ from sqlmesh.core import constants as c
 from sqlmesh.core.analytics.dispatcher import AsyncEventDispatcher, EventDispatcher
 from sqlmesh.utils import random_id
 from sqlmesh.utils.date import now_timestamp
-from sqlmesh.utils.hashing import sha256
+from sqlmesh.utils.hashing import md5
 from sqlmesh.utils.pydantic import PydanticModel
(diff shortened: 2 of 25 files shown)


sqlmesh/core/config/scheduler.py
```diff
@@ -97,7 +97,7 @@ def state_sync_fingerprint(self, context: GenericContext) -> str:
         state_connection = (
             context.config.get_state_connection(context.gateway) or context._connection_config
         )
-        return sha256([state_connection.json(sort_keys=True)])
+        return md5([state_connection.json(sort_keys=True)])
```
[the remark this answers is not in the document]
izeigerman: We're not actually trying to obfuscate anything here. This is more of a fingerprint.

sqlmesh/core/analytics/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    return AnalyticsCollector(dispatcher=dispatcher)
+
+
+collector = init_collector()
+
+
+atexit.register(collector.shutdown, flush=True)
+
+
+def disable_analytics() -> None:
```
tobymao, line 28: why did you choose to do this pattern? should collector instead be a constant like COLLECTOR, which can then be disabled?

izeigerman: This allowed me to reuse existing constructs (eg dispatcher) without explicitly adding `disable` API to the collector. I don't feel strongly about either approach. Do you see any advantages to doing this the way you suggested?

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+        command_name: str,
+        command_args: t.Collection[str],
+        parent_command_names: t.Collection[str],
+    ) -> None:
+        """Called when a CLI command is executed.
+
+        Args:
+            command_name: The name of the command.
+            command_args: The arguments of the command.
+            parent_command_names: The names of the parent commands.
```
tobymao, line 76: what is a parent command, why does cli have it and not magic?

sqlmesh/core/analytics/__init__.py
tobymao: no

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+        project_type = project_type.upper()
+        if project_type == "DBT":
+            from dbt import version
+
+            dbt_version = version.__version__
+
+        self._add_event(
+            "PROJECT_LOADED",
+            {
+                "project_type": project_type.upper(),
```
tobymao, line 136: upper() is unnecessary, you call it on line 127

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+                "models_count": models_count,
+                "audits_count": audits_count,
+                "standalone_audits_count": standalone_audits_count,
+                "macros_count": macros_count,
+                "jinja_macros_count": jinja_macros_count,
+                "load_time_ms": int(load_time_sec * 1000),
+                "state_sync_fingerprint": state_sync_fingerprint,
+                "project_name_hash": _anonymize(project_name),
+                "dbt_version": dbt_version,
```
tobymao, line 145: i think dbt_version should probably only be in the dict if it exists. 

you could create the dict above the dbt check, and then set the dict inside the dbt if

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+            audits_count: The number of audits in the project.
+            standalone_audits_count: The number of standalone audits in the project.
+            macros_count: The number of macros in the project.
+            jinja_macros_count: The number of Jinja macros in the project.
+            load_time_sec: The time it took to load the project in (fractional) seconds.
+            state_sync_fingerprint: The fingerprint of the state sync configuration.
+            project_name: The name of the project.
+        """
+        dbt_version = None
+        project_type = project_type.upper()
```
tobymao, line 127: just because i'm ocd, let's use lower(), since dbt and native should just be lower case, also we use lower for engine types

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+            state_sync_type: The type of the engine used to store the SQLMesh state.
+            scheduler_type: The type of the scheduler being used. Eg. "builtin" or "airflow".
+        """
+        self._add_event(
+            "PLAN_APPLY_START",
+            {
+                "plan_id": plan.plan_id,
+                "engine_type": engine_type.lower() if engine_type is not None else None,
+                "state_sync_type": state_sync_type.lower() if state_sync_type is not None else None,
+                "scheduler_type": scheduler_type.upper(),
```
tobymao, line 171: let's make everything lower

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+        if not new_snapshots:
+            return
+        snapshots = []
+        for snapshot in new_snapshots:
+            snapshots.append(
+                {
+                    "name_hash": _anonymize(snapshot.name),
+                    "identifier": snapshot.identifier,
+                    "version": snapshot.version,
+                    "node_type": snapshot.node_type.upper(),
```
tobymao, line 218: lower

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+                "event": json.dumps(event),
+            }
+        )
+
+    @cached_property
+    def _user(self) -> User:
+        return User.load_or_create(self._sqlmesh_path)
+
+
+class User(PydanticModel):
```
tobymao, line 330: there's no reason for this to be a pydantic model

sqlmesh/core/analytics/dispatcher.py
```diff
@@ -0,0 +1,194 @@
...
+        self._session.headers.update(
+            {
+                "Content-Type": "application/json",
+                "Content-Encoding": "gzip",
+                "User-Agent": f"SQLMesh/{sqlmesh_version}",
+            }
+        )
+
+    def emit(self, events: t.List[t.Dict[str, t.Any]]) -> None:
+        url = urljoin(self.base_url, "sqlmesh/")
```
tobymao, line 48: maybe move this to init so it's not called every time we emit

sqlmesh/core/analytics/dispatcher.py
```diff
@@ -0,0 +1,194 @@
...
+        try:
+            self.emitter.emit(events_to_emit)
+        except Exception as e:
+            logger.info("Failed to emit events: %s", e)
+            if isinstance(e, ApiClientError):
+                if e.code == 429 and self._emit_interval_sec < self._max_emit_interval_sec:
+                    self._emit_interval_sec = min(
+                        self._emit_interval_sec * 2, self._max_emit_interval_sec
+                    )
+                    logger.info(
```
tobymao, line 145: should this be debug?

sqlmesh/core/analytics/dispatcher.py
```diff
@@ -0,0 +1,194 @@
...
+        if flush and self._events:
+            # Reduce the connect timeout to avoid blocking the shutdown.
+            self.emitter.connect_timeout = 2
+            self.flush()
+        self.emitter.close()
+
+    def _add_events(self, events: t.List[t.Dict[str, t.Any]], prepend: bool = False) -> None:
+        with self._events_lock:
+            if not prepend:
+                self._events += events
```
tobymao, line 172: you don't want to mutate? self._events.extend(...) ?

sqlmesh/core/analytics/collector.py
izeigerman: In `click` there's a hierarchy of commands. Eg:
```
sqlmesh plan
```
There are 2 commands here: `sqlmesh` and `plan`. In this case the command name is `plan` and its parent is `sqlmesh.`

With the CICD cli there are 2 levels of commands. Eg:
```
sqlmesh_cicd github run-all
```
`run-all` is the command name, while `sqlmesh_cicd` and `github` are the parents. 

There's no such hierarchy when using magics.

sqlmesh/core/analytics/dispatcher.py
izeigerman: Yes, it probably should. Good point

sqlmesh/core/analytics/collector.py
izeigerman: Why? Isn't convention for enum values to be upper case?

sqlmesh/core/analytics/collector.py
izeigerman: What do you mean? What do you suggest to do instead?

sqlmesh/core/analytics/collector.py
tobymao: maybe? but you make engine_type and state_sync_type lower, so now it's inconsistent, additionaly, things like dbt, native, sqlmesh etc, are all normally represented as lower case

sqlmesh/core/analytics/collector.py
tobymao: it can just be a data class

sqlmesh/core/analytics/collector.py
```diff
@@ -0,0 +1,351 @@
...
+            standalone_audits_count: The number of standalone audits in the project.
+            macros_count: The number of macros in the project.
+            jinja_macros_count: The number of Jinja macros in the project.
+            load_time_sec: The time it took to load the project in (fractional) seconds.
+            state_sync_fingerprint: The fingerprint of the state sync configuration.
+            project_name: The name of the project.
+        """
+        dbt_version = None
+        project_type = project_type.upper()
+        if project_type == "DBT":
```
tobymao, line 128: since you use "dbt" here and context.py:502 probably should make it a constant, and i think it should just be "dbt", not "DBT"

sqlmesh/core/analytics/collector.py
izeigerman: I also serialize it into a dict before dumping into YAML. Not sure I want to introduce new patterns here.

tobymao (approved):

PR #2603, opened 2024-05-13
Chore: Cosmetic changes to how we resolve session properties for a model

izeigerman:


georgesittas (approved):

sqlmesh/core/model/definition.py
```diff
@@ -1880,21 +1880,20 @@ def _split_sql_model_statements(
...
-    defaults: t.Optional[t.Dict[str, t.Any]],
-    provided: t.Optional[exp.Expression] | t.Optional[t.Dict[str, t.Any]],
+def _resolve_session_properties(
+    default: t.Optional[t.Dict[str, t.Any]],
+    provided: t.Optional[exp.Expression | t.Dict[str, t.Any]],
 ) -> t.Optional[exp.Expression]:
     if isinstance(provided, dict):
         session_properties = {k: exp.Literal.string(k).eq(v) for k, v in provided.items()}
+    elif provided:
+        session_properties = {expr.this.name: expr for expr in provided}
```
tobymao, line 1890: what is expr? what are you expecting it to be such that expr.this.name works

tobymao (approved):

sqlmesh/core/model/definition.py
izeigerman: It's a tuple of properties.

izeigerman: so we expect this to be `EQ` with the left operand being a string literal or an identifer

tobymao: sg

PR #2604, opened 2024-05-13
 Fix: make macro function argument resolution more robust

### LINKED ISSUE #2599: "Expecting )" error when using named args `@STAR(tbl, except := [col])`
https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#star

> For example, we might omit the alias argument with `@STAR(foo, except := [c])`

But I get an error:
```
Error: Failed to parse a model definition at '/Users/anentropic/dev/etl/models/orders.sql': Expecting ). Line 51, Col: 28.
  _TRG_USERACCESS_D
        )
    order by W_INSERT_DATE, ORDER_WID
)
select
    @STAR(ranked_orders, except := [row_num])
from ranked_orders
where
    row_num = 1
    and snowflake_last_ingested_at between @.
```

omitting the `except := [row_num]` allows the plan to work

I'm on sqlmesh 0.95.1 and if I look at the files in my virtualenv I appear to have the changes from this PR https://github.com/TobikoData/sqlmesh/issues/2484

georgesittas:
Fixes #2599

Prior to this PR, we would bind macro arguments to the corresponding formal parameters **only** if annotations were present, and the existing logic for that case was brittle and complicated. This PR addresses these points:

- Parameter binding will now always happen, ensuring that macros will get called with the correct arguments
- The binding logic has been offloaded to [`inspect.Signature.bind`](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind), which handles all sorts of edge cases for us that would otherwise be tedious and tricky to implement correctly (see added test cases).

### DIFF
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -600,12 +600,12 @@ If the column data types are known, the resulting query `CAST`s columns to their
 
 - `relation`: The relation/table whose columns are being selected
 - `alias` (optional): The alias of the relation (if it has one)
-- `except` (optional): A list of columns to exclude
+- `except_` (optional): A list of columns to exclude
 - `prefix` (optional): A string to use as a prefix for all selected column names
 - `suffix` (optional): A string to use as a suffix for all selected column names
 - `quote_identifiers` (optional): Whether to quote the resulting identifiers, defaults to true
 
-Like all SQLMesh macro functions, omitting an argument when calling `@STAR` requires passing all subsequent arguments with their name and the special `:=` keyword operator. For example, we might omit the `alias` argument with `@STAR(foo, except := [c])`. Learn more about macro function arguments [below](#positional-and-keyword-arguments).
+Like all SQLMesh macro functions, omitting an argument when calling `@STAR` requires passing all subsequent arguments with their name and the special `:=` keyword operator. For example, we might omit the `alias` argument with `@STAR(foo, except_ := [c])`. Learn more about macro function arguments [below](#positional-and-keyword-arguments).
 
 As a `@STAR` example, consider the following query:
 
@@ -635,15 +635,15 @@ FROM foo AS bar
 Note these aspects of the rendered query:
 - Each column is `CAST` to its data type in the table `foo` (e.g., `a` to `TEXT`)
 - Each column selection uses the alias `bar` (e.g., `"bar"."a"`)
-- Column `c` is not present because it was passed to `@STAR`'s `except` argument
+- Column `c` is not present because it was passed to `@STAR`'s `except_` argument
 - Each column alias is prefixed with `baz_` and suffixed with `_qux` (e.g., `"baz_a_qux"`)
 
 Now consider a more complex example that provides different prefixes to `a` and `b` than to `d` and includes an explicit column `my_column`:
 
 ```sql linenums="1"
 SELECT
-  @STAR(foo, bar, except=[c, d], 'ab_pre_'),
-  @STAR(foo, bar, except=[a, b, c], 'd_pre_'),
+  @STAR(foo, bar, except_=[c, d], 'ab_pre_'),
+  @STAR(foo, bar, except_=[a, b, c], 'd_pre_'),
   my_column
 FROM foo AS bar
 ```
@@ -661,7 +661,7 @@ FROM foo AS bar
 
 Note these aspects of the rendered query:
 - Columns `a` and `b` have the prefix `"ab_pre_"` , while column `d` has the prefix `"d_pre_"`
-- Column `c` is not present because it was passed to the `except` argument in both `@STAR` calls
+- Column `c` is not present because it was passed to the `except_` argument in both `@STAR` calls
 - `my_column` is present in the query
 
 ### @GENERATE_SURROGATE_KEY
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -183,8 +183,12 @@ class MacroEvaluator:
         if not callable(func):
             raise SQLMeshError(f"Macro '{name}' does not exist.")
 
-        # Bind the macro's actual parameters to its formal parameters
-        bound = inspect.signature(func).bind(self, *args, **kwargs)
+        try:
+            # Bind the macro's actual parameters to its formal parameters
+            bound = inspect.signature(func).bind(self, *args, **kwargs)
+        except Exception as e:
+            print_exception(e, self.python_env)
+            raise MacroEvalError("Error trying to eval macro.") from e
 
         try:
             annotations = t.get_type_hints(func)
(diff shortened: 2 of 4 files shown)


z3z1ma:
I like the use of BoundArguments here. Makes sense. However we need to use parameters to determine variadic positional args and variadic kwargs. The current check of tuple/dict is not correct. 

This will correct the implementation 
https://docs.python.org/3/library/inspect.html#inspect.Parameter

You can get Parameters from the Signature object.

georgesittas:
Hm I guess you were talking about this note in the docs?

> Should be used in conjunction with [Signature.parameters](https://docs.python.org/3/library/inspect.html#inspect.Signature.parameters) for any argument processing purposes.

I may have missed this one earlier, perhaps there could be ordering issues the way I'm currently accessing the arguments. Not sure though, will double-check. In any case, an example would still be appreciate :)

z3z1ma:
I got you. Give me just a min

georgesittas:
One bug I found with this approach is the following:

main:
```python
>>> from sqlmesh.core.macros import *
>>> from sqlglot import *
>>>
>>> @macro()
... def foo(evaluator: MacroEvaluator, a1: int = 1, a2: int = exp.Literal.number(2)):
...     return sum([a1, a2])
...
>>> MacroEvaluator().evaluate(parse_one("@foo()")).sql()
'3'
```

this branch:

```python
>>> from sqlmesh.core.macros import *
>>> from sqlglot import *
>>>
>>> @macro()
... def foo(evaluator: MacroEvaluator, a1: int = 1, a2: int = exp.Literal.number(2)):
...     return sum([a1, a2])
...
>>> MacroEvaluator().evaluate(parse_one("@foo()")).sql()
'1 + 2'
```

It seems like default arguments are not properly handled in the coercion loop.

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -600,12 +600,12 @@ If the column data types are known, the resulting query `CAST`s columns to their
...
 - `relation`: The relation/table whose columns are being selected
 - `alias` (optional): The alias of the relation (if it has one)
-- `except` (optional): A list of columns to exclude
+- `except_` (optional): A list of columns to exclude
 - `prefix` (optional): A string to use as a prefix for all selected column names
 - `suffix` (optional): A string to use as a suffix for all selected column names
 - `quote_identifiers` (optional): Whether to quote the resulting identifiers, defaults to true
 
-Like all SQLMesh macro functions, omitting an argument when calling `@STAR` requires passing all subsequent arguments with their name and the special `:=` keyword operator. For example, we might omit the `alias` argument with `@STAR(foo, except := [c])`. Learn more about macro function arguments [below](#positional-and-keyword-arguments).
+Like all SQLMesh macro functions, omitting an argument when calling `@STAR` requires passing all subsequent arguments with their name and the special `:=` keyword operator. For example, we might omit the `alias` argument with `@STAR(foo, except_ := [c])`. Learn more about macro function arguments [below](#positional-and-keyword-arguments).
```
tobymao, line 608: i thought you were going to call it exclude instead of except_

tobymao:
@georgesittas i'm a bit confused, isn't main's behavior correct? if the default value is a literal number, but the type says it's an int, it should be coerced.

docs/concepts/macros/sqlmesh_macros.md
georgesittas: Yep, I decided to not include it in this PR though and since there are some other things I was planning to look at afterwards I thought I'd fix the docs temporarily, until that refactor is made.

In any case though, `except_` won't be deprecated at first so this will still be useful I guess.

georgesittas:
> @georgesittas i'm a bit confused, isn't main's behavior correct? if the default value is a literal number, but the type says it's an int, it should be coerced.

There's no issue with main, this is an issue with this branch. I was replying to Alex's concern above.

tobymao (approved):

z3z1ma:
Small changes below. cc @georgesittas 
I havent ran this, but this is what I had in mind.

```python
def send(
    self, name: str, *args: t.Any, **kwargs: t.Any
) -> t.Union[None, exp.Expression, t.List[exp.Expression]]:
    func = self.macros.get(normalize_macro_name(name))

    if not callable(func):
        raise SQLMeshError(f"Macro '{name}' does not exist.")

    try:
        # Bind the macro's actual parameters to its formal parameters
        sig = inspect.signature(func)
        bound = sig.bind(self, *args, **kwargs)
    except Exception as e:
        print_exception(e, self.python_env)
        raise MacroEvalError("Error trying to eval macro.") from e

    try:
        annotations = t.get_type_hints(func)
    except NameError:  # forward references aren't handled
        annotations = {}

    # If the macro is annotated, we try coerce the actual parameters to the corresponding types
    if annotations:
        for arg, value in bound.arguments.items():
            typ = annotations.get(arg)
            if not typ:
                continue

            # Changes to bound.arguments will reflect in bound.args and bound.kwargs
            # https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.arguments
            param = sig.parameters[arg]
            if param.default is value:
                continue
            elif param.kind is inspect.Parameter.VAR_POSITIONAL:
                bound.arguments[arg] = tuple(self._coerce(v, typ) for v in value)
            elif param.kind is inspect.Parameter.VAR_KEYWORD:
                bound.arguments[arg] = {k: self._coerce(v, typ) for k, v in value.items()}
            else:
                bound.arguments[arg] = self._coerce(value, typ)

    try:
        return func(*bound.args, **bound.kwargs)
    except Exception as e:
        print_exception(e, self.python_env)
        raise MacroEvalError("Error trying to eval macro.") from e

```

georgesittas:
Ah I see, makes sense. Thanks for catching!

sqlmesh/core/macros.py
```diff
@@ -183,35 +183,39 @@ def send(
         if not callable(func):
             raise SQLMeshError(f"Macro '{name}' does not exist.")
 
+        try:
+            # Bind the macro's actual parameters to its formal parameters
+            sig = inspect.signature(func)
+            bound = sig.bind(self, *args, **kwargs)
+            bound.apply_defaults()
```
georgesittas, line 190: This ensures default values will also be coerced, even when omitted.

georgesittas:
@z3z1ma refactored the previous approach, feel free to take another look. I thought a bit about the `isinstance(value, tuple)` and `isinstance(value, dict)` branches and I'm still not sure if there's a code path where they'd produce incorrect behavior, but I do agree that the suggested approach is more robust / clear. Let me know if you had a particular example in mind that would crash using that approach and I can add another test.

z3z1ma:
You would have hit that codepath anytime someone asked passes in something that meets that isinstance check and has typing. This includes any arg instead of just variadics. And the top-level annotation was being passed inside the comprehension which could've also been wrong since generic tuples apply the inner types to the tuples values.

PR #2605, opened 2024-05-14
Fix: Ignore non-restateable models and print out a warning instead of failing the plan

izeigerman:
Previously SQLMesh would fail the plan if the restatement list contained a model which couldn't be restated.

When this behavior is combined with selectors, it results in a subpar experience, as the selector expression may encompass a large set of models, only a small portion of which cannot be restated.

This update changes the behavior so that an error is printed out for models that are part of the restatement list but cannot be restated instead failing the plan. It also completely ignores symbolic and seed models, as restating those is a no-op

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import re
 import sys
 import typing as t
+import logging
 from collections import defaultdict
 from datetime import datetime
 from functools import cached_property
@@ -27,6 +28,9 @@ from sqlmesh.utils import random_id
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.date import TimeLike, now, to_datetime, yesterday_ds
 from sqlmesh.utils.errors import NoChangesPlanError, PlanError, SQLMeshError
+
+
+logger = logging.getLogger(__name__)
 
 
 class PlanBuilder:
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlmesh/core/plan/builder.py
```diff
@@ -298,18 +306,25 @@ def is_restateable_snapshot(snapshot: Snapshot) -> bool:
...
-                not forward_only_preview_needed and not is_restateable_snapshot(snapshot)
-            ):
-                raise PlanError(
-                    f"Cannot restate from '{model_fqn}'. Either such model doesn't exist, no other materialized model references it, or restatement was disabled for this model."
-                )
+            if not snapshot:
+                raise PlanError(f"Cannot restate model '{model_fqn}'. Model does not exist.")
+            if not forward_only_preview_needed:
+                if not self._is_dev and snapshot.disable_restatement:
+                    self._console.log_error(
```
eakmanrq, line 321: This is a warn and not an error right?

sqlmesh/core/plan/builder.py
```diff
@@ -298,18 +306,25 @@ def is_restateable_snapshot(snapshot: Snapshot) -> bool:
...
+                        f"Cannot restate model '{model_fqn}'. Restatement is disabled for this model."
+                    )
+                    continue
+                elif snapshot.is_symbolic or snapshot.is_seed:
```
eakmanrq, line 325: Should we log something here? Maybe info?

eakmanrq (approved):

PR #2608, opened 2024-05-14
fix: add support for untyped snowflake arrays

eakmanrq:
Prior to this change the schema differ assumed all arrays had a type which isn't true for Snowflake. This adds proper support for "untyped" arrays.

Also did some cleanup where old `self` referenced weren't renamed when methods were changed to classmethods.

### DIFF
--- a/sqlmesh/core/schema_diff.py
+++ b/sqlmesh/core/schema_diff.py
@@ -43,8 +43,8 @@ class TableAlterColumn(PydanticModel):
     quoted: bool = False
 
     @classmethod
-    def primitive(self, name: str, quoted: bool = False) -> TableAlterColumn:
-        return self(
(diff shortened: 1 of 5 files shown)


sqlmesh/core/schema_diff.py
```diff
@@ -456,6 +458,9 @@ def _alter_operation(
                     root_struct,
                 )
             if new_type.this == current_type.this == exp.DataType.Type.ARRAY:
+                # Some engines (i.e. Snowflake) don't support defining types on arrays
+                if not new_type.expressions or not current_type.expressions:
```
tobymao, line 462: what if something is untyped and then it becomes typed?

eakmanrq: There is no known way to do this today. The only untyped example must remain untyped.

tobymao (approved):

PR #2610, opened 2024-05-14
Fix: Parsing of session properties when only one key-value pair is provided

izeigerman:
Expression
```
session_properties (
    'warehouse' = 'test_warehouse'
)
```
is being parsed into the `exp.Paren` expression instead of `exp.Tuple` which was not accounted for while merging the model's values with the default ones.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1887,6 +1887,8 @@ def _resolve_session_properties(
     if isinstance(provided, dict):
         session_properties = {k: exp.Literal.string(k).eq(v) for k, v in provided.items()}
     elif provided:
+        if isinstance(provided, exp.Paren):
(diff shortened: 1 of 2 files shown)


eakmanrq (approved):

sqlmesh/core/model/definition.py
```diff
@@ -1887,6 +1887,8 @@ def _resolve_session_properties(
     if isinstance(provided, dict):
         session_properties = {k: exp.Literal.string(k).eq(v) for k, v in provided.items()}
     elif provided:
+        if isinstance(provided, exp.Paren):
+            provided = exp.Tuple(expressions=[provided.this])
```
tobymao, line 1891: provided.this.unnest()

then you can handle many parenthesis lol

tobymao (approved):

PR #2615, opened 2024-05-15
chore: split up get snapshots operations

eakmanrq:
Took two operations that `_get_snapshots` currently does, which are creating an expression to get the snapshots and parse the response into a snapshot, into a method and a function. This was done to make these operations accessible elsewhere without having to call the whole method.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -459,6 +459,45 @@ class EngineAdapterStateSync(CommonStateSyncMixin, StateSync):
             return query.lock(copy=False)
         return query
 
+    def _get_snapshots_expressions(
+        self,
+        snapshot_ids: t.Optional[t.Iterable[SnapshotIdLike]] = None,
+        lock_for_update: bool = False,
+        hydrate_seeds: bool = False,
+        batch_size: t.Optional[int] = None,
+    ) -> t.Iterator[exp.Expression]:
+        for where in (
+            [None]
+            if snapshot_ids is None
+            else self._snapshot_id_filter(snapshot_ids, alias="snapshots", batch_size=batch_size)
+        ):
+            query = (
+                exp.select(
+                    "snapshots.snapshot",
+                    "snapshots.name",
+                    "snapshots.identifier",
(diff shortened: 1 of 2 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1452,6 +1463,27 @@ def _snapshot_to_json(snapshot: Snapshot) -> str:
...
+    payload = json.loads(serialized_snapshot)
+
+    def loader() -> Node:
+        return parse_obj_as(Node, payload["node"])  # type: ignore
+
+    payload["node"] = model_cache.get_or_load(f"{name}_{identifier}", loader=loader)  # type: ignore
+    snapshot = Snapshot(**payload)
+
+    if seed_content and isinstance(snapshot.node, SeedModel):
+        snapshot.node = t.cast(SeedModel, snapshot.node).to_hydrated(seed_content)
```
tobymao, line 1482: why did you have to cast, there's an isinstance check on 1481

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -504,17 +492,43 @@ def _get_snapshots(
                     ),
                     join_type="left",
                 )
-            elif lock_for_update:
+            else:
+                query = query.select(exp.Null().as_("content"))
```
tobymao, line 496: what's going on here?

sqlmesh/core/state_sync/engine_adapter.py
tobymao: seed_content could be None, is that ok? the previous code checked that seed_content was not falsey

tobymao (approved):

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: Ensures that we always return the same number of rows regardless if we are hydrating or not

sqlmesh/core/state_sync/engine_adapter.py
eakmanrq: Removed cast. The `if seed_content` does this check.

PR #2616, opened 2024-05-15
Fix: Take batch_size into account when filling snapshots for INCREMENTAL_BY_UNIQUE_KEY models

### LINKED ISSUE #2609: INCREMENTAL_BY_UNIQUE_KEY models not taking `batch_size` into account
On Trino/Iceberg (and potentially other engines; unconfirmed), the following model doesnt backfill correctly:

```
MODEL (
    name  datalake_iceberg.stg_premium_day4,
         unique_key id,
         batch_size 1
    ),
    start '2020-01-01',
    cron '@daily'
);

select * from datalake_iceberg.seed_model --note: this is the seed_model from `sqlmesh init`
where event_date between @start_dt and @end_dt
```

The goal here is to backfill a "daily" model, in batches of 1 day.

However, it does not execute correctly.  It triggers a DELETE+INSERT (where the DELETE has `WHERE TRUE` and clears the whole table) for every interval/batch, even though it should only trigger a DELETE+INSERT for the _first_ batch (to support restatements) and then a MERGE for subsequent batches.

Killing `sqlmesh plan` before it finishes with ctrl+c and then running `sqlmesh plan` again, which picks up where it left off, triggers the correct MERGE behaviour.

It appears that the `batch_size` is not being taken into account correctly when deciding whether to clear the table or merge into the table

erindru:
This PR makes the `batch_index` available when filling snapshots, so `EvaluationStrategy`'s can take it into account when making decisions.

Prior to this, for `kind`'s like INCREMENTAL_BY_UNIQUE_KEY, if:
 - You had a model that defines intervals (say `@daily`) and runs them in multiple batches (say, `batch_size=1`, which generates a `Snapshot` per interval)
 - When running the `Snapshot`'s, SQLMesh would treat them all as "clear table -> insert data" rather than just the first one

After this change, for `Snapshot`'s with no existing intervals at plan time, SQLMesh can check the index in the batch and just execute the "clear table" logic for the _first_ snapshot and then the "merge into existing table" logic for subsequent snapshots 

Fixes #2609


tests/core/engine_adapter/test_integration.py
```diff
@@ -96,6 +101,20 @@ def dialect(self) -> str:
     def current_catalog_type(self) -> str:
         return self.engine_adapter.current_catalog_type
 
+    @property
+    def supports_merge(self) -> bool:
```
erindru, line 105: I found that this needed to be checked in multiple places so I made it a property of the TestContext

In addition, it's become more complex. It's no longer "engine either supports merge or it doesnt", it depends on both the engine _and_ the catalog

tests/core/engine_adapter/test_integration.py
```diff
@@ -376,6 +395,116 @@ def non_temp_tables(self) -> t.List[str]:
         return [x for x in self.tables if not x.startswith("__temp") and not x.startswith("temp")]
 
 
+class ProjectCreator(PydanticModel):
```
erindru, line 398: I had trouble creating a test case to reproduce the issue because it touched so many concepts.

What I really wanted was to be able to set up a small isolated SQLMesh project to expose the issue from the users perspective (`sqlmesh plan` throws error when I use a model defined like so), so I wrote this small fixture to create a minimal project I could run against the current engine adapter

I realise there is some overlap with `sqlmesh.cli.example_project.init_example_project` but I didnt need the example models and they just slowed things down

tests/core/engine_adapter/test_integration.py
```diff
@@ -376,6 +395,116 @@ def non_temp_tables(self) -> t.List[str]:
...
+    SEEDS_DIR: str = "seeds"
+    MODELS_DIR: str = "models"
+
+    output_path: pathlib.Path
+    engine_adapter: EngineAdapter
+    gateway: str
+    schema_name: str
+    _context: t.Optional[Context] = PrivateAttr(default=None)
+
+    def add_seed(self, model_name: str, columns: t.Dict[str, str], rows: t.List[t.Dict[str, str]]):
```
tobymao, line 414: check out core/dialect.py::pandas_to_sql

or just use create_view in the engine_adapter, that accepts a pandas dataframe

erindru: So the goal was to produce a project from a user perspective, based on the files in the filesystem that a user would create.

A user doesn't create a seed as a pandas DataFrame (unless theyre using Python models I guess), they create a .csv file in the `seeds/` directory and expose it in a SQL model using `kind SEED`.

However, i've implemented the `pandas_to_sql` version because in this case the seed data is just a vehicle to expose an issue in the INCREMENTAL_BY_UNIQUE_KEY model

erindru:
@izeigerman I had a go at implementing your suggestions on the Airflow scheduler code but I ran out of time to add a test case that proves I implemented them correctly.

I'm back on Monday and can resume this up then; otherwise feel free to finish it off if this bug is blocking someone

sqlmesh/engines/commands.py
```diff
@@ -37,6 +37,7 @@ class EvaluateCommandPayload(PydanticModel):
     end: TimeLike
     execution_time: TimeLike
     deployability_index: DeployabilityIndex
+    batch_index: int = 0
```
izeigerman, line 40: [Nit] I don't think we need a default value here, to make sure that all command's attributes are set explicitly upstream.

examples/airflow/docker_compose_decorator.py
```diff
@@ -20,6 +20,9 @@
     ]
 )
 
+# Dont load Airflow example DAGs because they cause visual pollution
+docker_compose["x-airflow-common"]["environment"]["AIRFLOW__CORE__LOAD_EXAMPLES"] = "false"
```
izeigerman, line 24: Nice 👍  Let's move this into `Dockerfile.template` : https://github.com/TobikoData/sqlmesh/blob/main/examples/airflow/Dockerfile.template#L49

examples/airflow/spark_conf/hive-site.xml
```diff
@@ -1,7 +1,7 @@
 <configuration>
    <property>
      <name>javax.jdo.option.ConnectionURL</name>
-     <value>jdbc:postgresql://airflow-postgres-1:5432/metastore_db</value>
+     <value>jdbc:postgresql://postgres:5432/metastore_db</value>
```
izeigerman, line 4: Just curious: why did you have to change this?

tests/core/engine_adapter/test_integration.py
izeigerman: Interesting. What prompted you to change this? Did you run into any issues? Or did you notice that merge is not being used?

examples/airflow/spark_conf/hive-site.xml
erindru: I was having trouble with new `docker compose` vs legacy `docker-compose` (which I think is docker compose V2 vs docker compose V1).

- The name `airflow-postgres-1` is just the container name, autogenerated by docker-compose. Since the service is defined as `postgres` in the `docker-compose.yml`, the [network name](https://docs.docker.com/compose/networking/) that should be referenced is `postgres`
- Something seems to have changed in the networking that `docker compose` sets up in newer versions. The name `airflow-postgres-1` was not reachable anymore, but `postgres` was. So I couldnt get the tests to run without this change

examples/airflow/docker_compose_decorator.py
erindru: Do you mean add an ENV entry to `Dockerfile.template` to set it?

The problem is, it gets overridden again the second the containers are instantiated, because its explicitly defined in the upstream `docker-compose.yml` file

tests/core/engine_adapter/test_integration.py
erindru: I noticed that some other tests, like `test_merge()` werent actually testing all the combinations because they were being skipped where `dialect=spark` or `dialect=trino`.

Spark and Trino _do_ support MERGE, just not on all catalogs, so I improved the check.

But, I wanted to use the same logic in the ` test_batch_size_on_incremental_by_unique_key_model` test (because unless i'm missing something, `INCREMENTAL_BY_UNIQUE_KEY` only works on engines that support `MERGE`). So rather than duplicate the checks, I made them a property of the `TestContext` fixture so they could be used in both places.

Arguably this kind of thing could also live on the EngineAdapter

tests/schedulers/airflow/test_dag_generator.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        *,
+        target: BaseTarget,
+        **kwargs: t.Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.target = target
+
+
+def test_generate_plan_application_dag__batch_index_populated(mocker: MockerFixture, make_snapshot):
```
erindru, line 32: I couldn't find any existing tests for `SnapshotDagGenerator` so I added this file

tests/core/engine_adapter/test_integration.py
erindru: I ended up removing this entirely in favour of adding the core methods to `TestContext` instead

sqlmesh/engines/commands.py
erindru: Fair call, i've removed the default

izeigerman (approved):
This looks great, thank you!

PR #2619, opened 2024-05-16
Refactor: clean up some sections in the unit test module

georgesittas:
cc @z3z1ma

### DIFF
--- a/tests/core/test_test.py
+++ b/tests/core/test_test.py
@@ -1034,7 +1034,7 @@ test_foo:
     )
     _check_successful_or_raise(test.run())
 
-    assert len(test._fixture_table_cache) == len(sushi_context.models) + 1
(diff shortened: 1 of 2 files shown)


tests/core/test_test.py
```diff
@@ -1034,7 +1034,7 @@ def test_create_external_model_fixture(sushi_context: Context, mocker: MockerFix
     )
     _check_successful_or_raise(test.run())
 
-    assert len(test._fixture_table_cache) == len(sushi_context.models) + 1
+    assert len(test._fixture_table_cache) == 1
```
georgesittas, line 1037: This changed because we used to have `self.models.keys() | self.body.get("inputs", {})`, but I believe `self.models` is redundant here so I went ahead and removed it.

tobymao (approved):

PR #2629, opened 2024-05-17
Chore: pass dialect in builders used within sushi python models

georgesittas:
Prior to this change, the `from_` builder would produce a `ParseError` when run against specific engines. The bug was spotted for BigQuery, because we'd pass a backtick-quoted name in the builder, which can't be parsed by the sqlglot base dialect.

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -38,7 +38,7 @@ bigquery_config = Config(
         )
     },
     default_gateway="gcp",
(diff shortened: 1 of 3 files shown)


tobymao (approved):

examples/sushi/config.py
```diff
@@ -38,7 +38,7 @@
         )
     },
     default_gateway="gcp",
```
crericha, line 40: Can you change the gateway name to `bq`?

PR #2632, opened 2024-05-17
Feat: Introduce support for 'pre ping' to detect stale or lost connections

izeigerman:
For long running plans, a state connection established at the beginning of a plan application might go stale by the end of it.

This update introduces a "pre-ping" mechanism, ensuring that the connection is not stale before initiating a new transaction. The ping implementation itself is dialect-specific but essentially comes down to executing a `SELECT 1` query in all cases except MySQL, which has a first-class API for this purpose.

The new behavior is configurable and is turned off by default for all engines except for those likely to be used to store the SQLMesh state: MySQL, MSSQL, and Postgres. This is also only supported for engines with transaction support.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -109,26 +109,28 @@ class EngineAdapter:
         default_catalog: t.Optional[str] = None,
         execute_log_level: int = logging.DEBUG,
         register_comments: bool = True,
+        pre_ping: bool = False,
         **kwargs: t.Any,
     ):
         self.dialect = dialect.lower() or self.DIALECT
         self._connection_pool = create_connection_pool(
             connection_factory, multithreaded, cursor_kwargs=cursor_kwargs, cursor_init=cursor_init
         )
-        self.sql_gen_kwargs = sql_gen_kwargs or {}
+        self._sql_gen_kwargs = sql_gen_kwargs or {}
         self._default_catalog = default_catalog
         self._execute_log_level = execute_log_level
         self._extra_config = kwargs
-        self.register_comments = register_comments
+        self._register_comments = register_comments
+        self._pre_ping = pre_ping
 
     def with_log_level(self, level: int) -> EngineAdapter:
         adapter = self.__class__(
             lambda: None,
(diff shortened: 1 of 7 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -2038,6 +2048,12 @@ def _rename_table(
     ) -> None:
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
+    def _ping(self) -> None:
+        try:
+            self._execute(exp.select(exp.Literal.number(1)).sql(dialect=self.dialect))
+        finally:
+            self._connection_pool.close_cursor()
```
eakmanrq, line 2055: Why close in all cases? If we just verified the connection is good, why not keep it?

eakmanrq: Is it because we are just validating the connection?

izeigerman: This closing the cursor (not connection) to discard the result of the `SELECT 1` query which can technically be fetched.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2038,6 +2049,12 @@ def _rename_table(
     ) -> None:
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
+    def _ping(self) -> None:
+        try:
+            self._execute(exp.select(exp.Literal.number(1)).sql(dialect=self.dialect))
```
tobymao, line 2054:
```suggestion
            self.execute(exp.select("1"))
```

tobymao (approved):

sungchun12:
Really glad to see this PR :)

sqlmesh/core/engine_adapter/base.py
izeigerman: This was intentional. `execute` always makes sure there's a transaction, while starting a transaction might trigger a ping which would cause an infinite loop.

izeigerman: Also this will select a string literal, not a number 1.

PR #2634, opened 2024-05-18
Fix `%%model` and `%model` ipython magic  error for seed model with csv column name with periods

butchland:
* Add missing tests for `%model` and `%%model` line magic
   * add tests for full and view kind of models
   * add tests for seed model (happy path)
 * Add test for seed model with period in column name
 *  Fix error for seed models with period in column name

### DIFF
--- a/sqlmesh/core/model/seed.py
+++ b/sqlmesh/core/model/seed.py
@@ -41,9 +41,6 @@ class CsvSettings(PydanticModel):
             return v
         return v.this
 
-def clean_csv_column(col: str) -> str:
-    # return col
-    return col.replace(".","")
 class CsvSeedReader:
     def __init__(self, content: str, dialect: str, settings: CsvSettings):
         self.content = content
@@ -83,7 +80,7 @@ class CsvSeedReader:
             )
             self._df = self._df.rename(
                 columns={
-                    col: clean_csv_column(normalize_identifiers(col, dialect=self.dialect).name)
+                    col: normalize_identifiers(col, dialect=self.dialect).name
                     for col in self._df.columns
                 },
             )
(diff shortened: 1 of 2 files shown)


butchland:
weird - the tests are failing on stuff I didn't touch. I thought removing the `@pytest.mark.slow` annotation would fix it but it seems not. 
¯\_(ツ)_/¯   I don't know  what else to do get my PR accepted.

georgesittas:
Hey @butchland, rebasing off of main should fix the CI issues. Can you share some context around this fix? What exactly was the issue / error you saw?

butchland:
hi @georgesittas sorry for the late reply - here is a[ colab notebook](https://colab.research.google.com/drive/1UDHJ3ornoi_-Vqp_nKgFQMbSWI9VZFMk?usp=sharing) showing the error that occurs when the seed data contains a csv file with a column name that has  a period in it..

georgesittas:
I see, thanks for sharing the notebook @butchland. I understand the issue now.

I don't think this approach to solving it is correct, though. Seeds with dots in the column names are perfectly legal, and we shouldn't just drop that information to make the parser happy. Also, this doesn't address similar issues with other "invalid" identifier characters.

The issue here is that we're trying to parse these names as if they are valid SQL, even though that's not really the intention. The proper way to handle them is to parse them into identifiers, and if they contain any "special" characters, ensure that they are quoted. That way, we'll convert them to valid SQL.

georgesittas:
I'll go ahead and close this PR for now, but I've opened a [ticket](https://github.com/TobikoData/sqlmesh/issues/2677) to make sure this is tracked. We'll take a look soon, thanks for bringing this into our attention and appreciate the PR!

PR #2638, opened 2024-05-20
Chore: enforce consistent indentation in docs

georgesittas:
Currently, the indentation in both SQL and YAML snippets is inconsistent: we use 4 spaces in some examples and 2 spaces in others. This PR changes this so that we default to 2 spaces for consistency, which also seems to be the width we use in the example projects (sushi, etc). It also includes a few minor fixups, like typo corrections.


tobymao (approved):

treysp (approved):
Thank you!

PR #2644, opened 2024-05-21
Feat: Improve the CLI output of the table diff command

izeigerman:
The following improvements have been introduced:
- Split `COMMON`, which represented the number of joined rows ,into `FULL MATCH` and `PARTIAL MATCH` to show the number of rows that are identical vs the number of rows with only  a partial match accordingly.
- Add percentage values for `FULL MATCHES` and `PARTIAL MATCHES`.
- Only display the number of rows if it's not 0.
- Add ability to set the number of decimal places to use when comparing floating point columns. This is set to 3 by default.

Screenshot 2024-05-21 at 10 46 42 AM

### DIFF
--- a/sqlmesh/core/table_diff.py
+++ b/sqlmesh/core/table_diff.py
@@ -87,14 +87,47 @@ class RowDiff(PydanticModel, frozen=True):
         return int(self.stats["join_count"])
 
     @property
+    def full_match_count(self) -> int:
+        """The number of rows for which shared columns have same values."""
+        return int(self.stats["full_match_count"])
+
+    @property
+    def full_match_pct(self) -> float:
+        """The percentage of rows for which shared columns have same values."""
+        return self._pct(2 * self.full_match_count)
+
+    @property
+    def partial_match_count(self) -> int:
+        """The number of rows for which some shared columns have same values."""
+        return self.join_count - self.full_match_count
+
+    @property
+    def partial_match_pct(self) -> float:
+        """The percentage of rows for which some shared columns have same values."""
+        return self._pct(2 * self.partial_match_count)
+
+    @property
     def s_only_count(self) -> int:
         """Count of rows only present in source."""
         return int(self.stats["s_only_count"])
 
     @property
+    def s_only_pct(self) -> float:
+        """The percentage of rows that are only present in source."""
+        return self._pct(self.s_only_count)
+
+    @property
     def t_only_count(self) -> int:
         """Count of rows only present in target."""
         return int(self.stats["t_only_count"])
+
+    @property
+    def t_only_pct(self) -> float:
+        """The percentage of rows that are only present in target."""
+        return self._pct(self.t_only_count)
+
+    def _pct(self, numerator: int) -> float:
+        return round((numerator / (self.source_count + self.target_count)) * 100, 2)
 
 
 class TableDiff:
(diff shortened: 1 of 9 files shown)


sungchun12:
This is out of the scope for this PR, but is there a reason we don't show data diffs side by side like a git diff? We do it in the UI but not for the CLI. I imagine a lot of people want something similar in the CLI.

Example library that does this: https://github.com/paulfitz/daff

image

izeigerman:
@sungchun12 , thanks, this is definitely on my mind! I even submitted an issue at the same time as this PR: https://github.com/TobikoData/sqlmesh/issues/2645. Any chance you can share this info there?

treysp:
I naturally compare the row counts percentages and counts, and full match 6 is not 92.31% of 7. (I recognize why it's calculated the way it is, but I don't think we can expect users to.)

I think that calculation is going to be confusing unless we display percentages that sum to 100%.

izeigerman:
@treysp I'm not sure we can, in good faith, show 100% when the number of rows doesn't match, right?

treysp:
I meant show a percentage next to each count such that all the percentages sum to 100% - that way the user can tell how the denominator was calculated

izeigerman:
@treysp ah, that's a good point, thanks. Updated my PR:

sungchun12:
> @sungchun12 , thanks, this is definitely on my mind! I even submitted an issue at the same time as this PR: #2645. Any chance you can share this info there?

will do!

sqlmesh/core/table_diff.py
```diff
@@ -187,9 +222,20 @@ def row_diff(self) -> RowDiff:
                     t_index.append(col)
             index_cols = list(dict.fromkeys(index_cols))
 
+            matched_columns = {
+                c: t for c, t in self.source_schema.items() if t == self.target_schema.get(c)
+            }
+
+            def _column_expr(name: str, table: str) -> exp.Expression:
+                if matched_columns[name].this in exp.DataType.FLOAT_TYPES:
+                    return exp.Round(
```
tobymao, line 231: exp.func("ROUND", exp.column(name, table), self.decimals)

sqlmesh/core/table_diff.py
```diff
@@ -244,14 +289,31 @@ def name(e: exp.Expression) -> str:
                 .where(self.where)
             )
 
+            query = exp.select(
+                "*",
+                exp.Case()
+                .when(
+                    exp.and_(
+                        *[
+                            exp.column(f"{c}_matches").eq(exp.Literal.number(1))
```
tobymao, line 298: eq(1)

sqlmesh/core/table_diff.py
```diff
@@ -244,14 +289,31 @@ def name(e: exp.Expression) -> str:
...
+                            for c in matched_columns
+                        ]
+                    ),
+                    exp.Literal.number(1),
```
tobymao, line 302: 1 and 0 for else_

sqlmesh/magics.py
```diff
@@ -598,6 +598,12 @@ def create_external_models(self, context: Context, line: str) -> None:
         action="store_true",
         help="Show a sample of the rows that differ. With many columns, the output can be very wide.",
     )
+    @argument(
+        "--decimals",
+        type=int,
+        default=20,
```
tobymao, line 604: 3 not 20 right

tobymao (approved):

sqlmesh/magics.py
izeigerman: ah, thanks for catching this. Not sure how this happened.

PR #2650, opened 2024-05-21
Chore: include local sqlmesh configs in integration test context

treysp:
Remote engine connection configs may be stored in the local `.sqlmesh` directory. We get a key error for the gateway If that is the case and those configs aren't included in the context.


treysp:
@erindru fyi

erindru:
Ah, I did wonder how the tests got run against things like Redshift :)

izeigerman (approved):

Issue #2656, opened 2024-05-22
Running 'sqlmesh test' no longer displays dots for successful test cases


giovannipcarvalho:
I noticed a few versions ago that `sqlmesh test` stopped logging `...` for successes. 

Bisecting through the revisions I pinpointed it to the introduction of the method `addSuccess` for the class `ModelTextTestResult` in d8ff3fd4c8946ab756609fc38c6d43552aea103e.

I think the original behavior is a very helpful visual feedback of test progression and this seems like a minor oversight.

repro:

```console
λ uv venv
Using Python 3.11.8 interpreter at: /home/giovanni/.local/bin/python
Creating virtualenv at: .venv
Activate with: source .venv/bin/activate

λ . .venv/bin/activate
λ uv pip install sqlmesh==v0.78.2
Resolved 52 packages in 795ms
Downloaded 1 package in 630ms
Installed 52 packages in 54ms
 [...]
λ sqlmesh init duckdb
λ sqlmesh test
.    <-------- dots are printed
----------------------------------------------------------------------
Ran 1 test in 0.020s

OK

λ uv pip install sqlmesh==v0.78.3
Resolved 52 packages in 284ms
Downloaded 1 package in 719ms
Installed 1 package in 19ms
 - sqlmesh==0.78.2
 + sqlmesh==0.78.3

λ rm -f db.db && sqlmesh test
    <-------- dots are missing
----------------------------------------------------------------------
Ran 1 test in 0.020s

OK
```

for completeness, it still happens on the latest version

```console
λ uv pip install sqlmesh --upgrade
Resolved 53 packages in 82ms
Installed 4 packages in 10ms
 + freezegun==1.5.1
 - sqlglot==22.5.0
 + sqlglot==23.17.0
 - sqlglotrs==0.1.2
 + sqlglotrs==0.2.5
 - sqlmesh==0.78.3
 + sqlmesh==0.97.0

λ rm -f db.db && sqlmesh test

----------------------------------------------------------------------
Ran 1 test in 0.022s

OK
```

georgesittas:
Fixed by #2657 - thanks!

PR #2657, opened 2024-05-22
fix: Fix test runner not displaying dots for successes (#2656)

giovannipcarvalho:
Since the introduction of the custom method `addSuccess` for `ModelTextTestResult` in d8ff3fd4c8946ab756609fc38c6d43552aea103e, the test runner stopped displaying dots for successful test cases.

This is due to the logging being part of the base implementation itself:

    # unittest.TextTestResult::addSuccess
    self.stream.write('.')
    self.stream.flush()

This patch adds a call the superclass implementation of `addSucess` to achieve the original behavior of reporting successes.

---

I actually wanted to add a test case for this change to ensure the behavior is maintained, but I couldn't find a reference point for tests that check the CLI output of `sqlmesh test`.  I can do so with some guidance. Thanks.

### DIFF
--- a/sqlmesh/core/test/result.py
+++ b/sqlmesh/core/test/result.py
@@ -36,4 +36,5 @@ class ModelTextTestResult(unittest.TextTestResult):
         Args:
             test: The test case
         """
+        super().addSuccess(test)
         self.successes.append(test)


tobymao (approved):
thank you!

tobymao:
not sure how to test this either, i don’t think it’s worth the hassle

PR #2666, opened 2024-05-23
feat!: snowpark

tobymao:


sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -72,28 +75,28 @@ def session(self, properties: SessionProperties) -> t.Iterator[None]:
...
         temp_table = self._get_temp_table(target_table or "pandas")
 
         def query_factory() -> Query:
-            from snowflake.connector.pandas_tools import write_pandas
-
-            if not self.table_exists(temp_table):
-                # Workaround for https://github.com/snowflakedb/snowflake-connector-python/issues/1034
-                #
-                # The above issue has already been fixed upstream, but we keep the following
-                # line anyway in order to support a wider range of Snowflake versions.
```
eakmanrq, line 92: If you are going to drop support for older versions, we should update the min version on Snowflake then right to reflect that?

tobymao: this bug only affects snowflake users on old versions who happen to also use pandas. i could add a min but didn't think it was that necessary

tobymao: after reading the dbt issues related to this, i think i'll keep the workaround for now

eakmanrq (approved):

sqlmesh/core/engine_adapter/base.py
```diff
@@ -146,14 +147,14 @@ def cursor(self) -> t.Any:
...
+    @property
+    def snowpark(self) -> t.Optional[SnowparkSession]:
+        return None
+
     @property
     def comments_enabled(self) -> bool:
         return self._register_comments and self.COMMENT_CREATION_TABLE.is_supported
 
-    @classmethod
-    def is_pandas_df(cls, value: t.Any) -> bool:
```
izeigerman, line 154: Why are we getting rid of this?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -53,6 +53,7 @@
         PySparkSession,
         Query,
         QueryOrDF,
+        SnowparkSession,
```
izeigerman, line 56: I don't understand how this import can possible work with Python 3.12

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -72,22 +75,30 @@ def session(self, properties: SessionProperties) -> t.Iterator[None]:
         yield
         self.execute(f"USE WAREHOUSE {current_warehouse_sql}")
 
+    @property
```
izeigerman, line 78: Should we cache this?

sqlmesh/core/engine_adapter/base.py
tobymao: because it was confusing mypy

sqlmesh/core/engine_adapter/base.py
tobymao: because it's an if else statement in the typing

tobymao: ah i see, you're right, i don't know how it works

PR #2667, opened 2024-05-23
feat!: bigframes

z3z1ma:


z3z1ma:
Dependency hell, unfortunately. 
Would  be nice if upstreams had looser upper range pins

z3z1ma:
I managed to wade my way through dep hell and get it to work FWIW. The last thing I was having issues with was ensuring the `to_gbq` table was persisted. The query would run but it seemed like a session scoped table. I could see someone forking the lib since bigframes/ibis are both pretty powerful here. I might even do it. So with that in mind, the only option is  the `optional_import` pattern. 

Alternatively I can maintain a `.patch` file and patch sqlmesh during standardized environment setup. But dep hell may be annoying enough to not be worth it too, even for the intrepid pythonista.

tobymao:
i'm going to close this for now to avoid dependency issues, but happy to chat more about this if you need it @z3z1ma

thomas-trividic-lv:
it would be nice to have this feature in the future 👍

tobymao:
@thomas-trividic-lv feel free to file an issue on the big frames github repo

PR #2669, opened 2024-05-23
Fix: make schemadiffer dialect aware

treysp:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -89,7 +89,7 @@ class EngineAdapter:
     SUPPORTS_MATERIALIZED_VIEWS = False
     SUPPORTS_MATERIALIZED_VIEW_SCHEMA = False
     SUPPORTS_CLONING = False
-    SCHEMA_DIFFER = SchemaDiffer()
+    SCHEMA_DIFFER = SchemaDiffer(dialect=DIALECT)
```
tobymao, line 92: is this gonna work? this only evaluates once right because it's a constant

treysp: Meaning `dialect=DIALECT` vs. explicitly passing the dialect `dialect=""`?

Won't it get evaluated once when the class is imported then be "frozen" as adapter instances are created? Like even if you overwrite DIALECT after creating an instance, the schemadiffer will still have dialect="duckdb" because it was evaluated at import time?

tobymao: if you don't redefine SCHEMA_DIFFER in a subclass, i believe that the subclass SCHEMA_DIFFER dialect will be ""

eakmanrq: I had a similar reaction. Maybe it would be better to create a `schema_differ` property that does `return SchemaDiffer(dialect=self.DIALECT)`. Then it could be overridden by other classes.

treysp: | Maybe it would be better to create a schema_differ property that does return SchemaDiffer(dialect=self.DIALECT).

So for Bigquery, I'd move all of this into a property?

```
    SCHEMA_DIFFER = SchemaDiffer(
        compatible_types={
            exp.DataType.build("INT64", dialect=DIALECT): {
                exp.DataType.build("NUMERIC", dialect=DIALECT),
                exp.DataType.build("FLOAT64", dialect=DIALECT),
                exp.DataType.build("BIGNUMERIC", dialect=DIALECT),
            },
            exp.DataType.build("NUMERIC", dialect=DIALECT): {
                exp.DataType.build("FLOAT64", dialect=DIALECT),
                exp.DataType.build("BIGNUMERIC", dialect=DIALECT),
            },
            exp.DataType.build("BIGNUMERIC", dialect=DIALECT): {
                exp.DataType.build("FLOAT64", dialect=DIALECT),
            },
            exp.DataType.build("DATE", dialect=DIALECT): {
                exp.DataType.build("DATETIME", dialect=DIALECT),
            },
        },
        support_coercing_compatible_types=True,
        dialect=self.DIALECT,
    )
```

PR #2674, opened 2024-05-23
Fix: For MSSQL apply the same workaround for the varchar size as we use for Redshift

### LINKED ISSUE #2671: MSSQL tables with incorrect VARCHAR size
Looks like MSSQL is prone to the same issue as Redshift where tables get created with invalid size for some VARCHAR columns due to the fact that SQLMesh creates table using CTAS with `LIMIT 0`. Further investigation is required.

izeigerman:
Closes #2671

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -277,8 +277,6 @@ class VarcharSizeWorkaroundMixin(EngineAdapter):
             assert not isinstance(table_name_or_schema, exp.Schema)
             # redshift and mssql have a bug where CTAS statements have non determistic types. if a limit
             # is applied to a ctas statement, VARCHAR types default to 1 in some instances.
-            # this checks the explain plain from redshift and tries to detect when these optimizer
-            # bugs occur and force a cast
             select_statement = statement.expression.copy()
             for select_or_union in select_statement.find_all(exp.Select, exp.Union):
                 select_or_union.set("limit", None)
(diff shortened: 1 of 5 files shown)


sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -246,3 +246,65 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+
+        if (
+            statement.expression
+            and statement.expression.args.get("limit") is not None
+            and statement.expression.args["limit"].expression.this == "0"
+        ):
+            assert not isinstance(table_name_or_schema, exp.Schema)
+            # redshift and mssql have a bug where CTAS statements have non determistic types. if a limit
+            # is applied to a ctas statement, VARCHAR types default to 1 in some instances.
+            # this checks the explain plain from redshift and tries to detect when these optimizer
```
georgesittas, line 280:
```suggestion
            # this checks the explain plan from redshift and tries to detect when these optimizer
```

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -246,3 +246,65 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+            columns_to_types=columns_to_types,
+            table_description=table_description,
+            **kwargs,
+        )
+
+        if (
+            statement.expression
+            and statement.expression.args.get("limit") is not None
+            and statement.expression.args["limit"].expression.this == "0"
+        ):
```
georgesittas, line 276:
```suggestion
        limit = statement.expression.args.get("limit")
        if statement.expression and limit and limit.expression.this == "0":
```

georgesittas (approved):

tobymao (approved):

sqlmesh/core/engine_adapter/mixins.py
izeigerman: `statement.expression` can be None too

georgesittas: yeah makes sense, missed that

PR #2687, opened 2024-05-27
Feat: Introduce INCREMENTAL_BY_PARTITION model kind

izeigerman:
This PR introduces support for the `INCREMENTAL_BY_PARTITION` kind which incrementally updates only the partition key values that are returned by the model query on any given run while keeping the remaining partitions intact.

The new kind is `forward_only` and the restatements are disabled for it by default.

For all engines but Apache Spark the evaluation will yield the following steps:
1. Create a temporary table where the output of the query is stored.
2. Get all partition values from the temporary table and use them as a filter by which to delete values in the target table.
3. Insert into the target table from the temporary table.

For Apache Spark we just rely on the `INSERT OVERWRITE` semantics with support for dynamic partition keys.


sqlmesh/core/engine_adapter/base.py
```diff
@@ -2006,6 +2015,41 @@ def _truncate_table(self, table_name: TableName) -> None:
...
         self.execute(f"TRUNCATE TABLE {table.sql(dialect=self.dialect, identify=True)}")
 
+    def _delete_then_insert_by_key(
+        self,
+        target_table: TableName,
+        source_table: QueryOrDF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        key: t.Sequence[exp.Expression],
+        is_unique_key: bool,
+    ) -> None:
```
eakmanrq, line 2025: I think it should be possible to have this support Databrick's replace where right? That would allow it to still not rely on transactions.

tobymao (approved):
just needs docs

sqlmesh/core/engine_adapter/base.py
izeigerman: Added support for RELACE WHERE. Please let me know if I did this correctly.

PR #2688, opened 2024-05-27
Chore: Don't store the seed's content in the state

### LINKED ISSUE #2668: Table diffing for large seed tables takes a long time. (mssql)
**Description**
I have a seed table with 1.5m rows which are all the countries/zip codes in the world.
The initial load takes 10 minutes with the standard configuration.
When I increase the batch_size to 100.000 it goes down to 3.5 minutes.

So first issue would be to see if the default batch size makes sense?

Then when changing a row I noted that it takes around 10 minutes after "sqlmesh plan" to print the first output, then it prints all 1.5m rows which takes about 7 minutes. Then the apply takes the same amount of time than the first apply.

**Suggestion**
Can we skip the diffing on large tables? Also then maybe remove the contents in the _seeds table content column, so it does not have to be inserted twice (I assume double the upload time and more storage).

**Reproduce**
To reproduce here download the allCountries https://download.geonames.org/export/zip/
I then added the header rows
`countryCode postalCode	placeName	stateCode	stateName	countryProvinceCode	countryProvinceName	communityCode	communityName	latitude	longitude	accuracy`

and used the model
```
MODEL (
  name seed.countries_utf8,
  kind SEED (
    path '$root/seeds/allCountriesUTF8.csv',
    batch_size 100000,
    csv_settings (
      delimiter='	'
    )
  ),
  columns (
    countryCode NVARCHAR(2),
    postalCode NVARCHAR(20),
    placeName NVARCHAR(180),
    stateCode NVARCHAR(20),
    stateName NVARCHAR(100),
    countryProvinceCode NVARCHAR(20),
    countryProvinceName NVARCHAR(100),
    communityCode NVARCHAR(20),
    communityName NVARCHAR(100),
    latitude NUMERIC(9, 6),
    longitude NUMERIC(9, 6),
    accuracy INTEGER
  )
)
```

izeigerman:
Closes #2668

We only really need to store the seed's content in order for the diffing to work at plan time. It appears that this convenience feature cannot justify the storage and compute overhead it incurs, especially for larger seed files.

This update ensures that we no longer rely on the seed's content stored in the state.

### DIFF
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -119,28 +119,10 @@ class ContextDiff(PydanticModel):
             if snapshot.snapshot_id not in added
             and snapshot.fingerprint != remote_snapshot_name_to_info[snapshot.name].fingerprint
         }
-        modified_local_seed_snapshot_ids = {
-            s.snapshot_id
-            for s in snapshots.values()
-            if s.is_seed and s.name in modified_snapshot_name_to_snapshot_info
-        }
-        modified_remote_snapshot_ids = {
-            s.snapshot_id for s in modified_snapshot_name_to_snapshot_info.values()
-        }
 
-        stored = {
-            **state_reader.get_snapshots(
-                [
-                    snapshot.snapshot_id
-                    for snapshot in snapshots.values()
-                    if snapshot.snapshot_id not in modified_local_seed_snapshot_ids
-                ]
-            ),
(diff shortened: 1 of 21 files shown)


tobymao:
migration to delete the seeds table?

izeigerman:
@tobymao created a file but forgot to add it. Thanks

PR #2690, opened 2024-05-28
Fix: unescape tabs and backslashes in CSV settings

### LINKED ISSUE #2681: Pass seed file tab delimiter specification correctly
Currently, the tab character `\t` isn't passed from the model specification parser to the `read_csv` function with usable escaping.

This specification:
```
kind SEED (
    path 'file.txt',
    csv_settings (
        delimiter='\t',
    ),
)
```

Produces this error:
```
Error: Falling back to the 'python' engine because the 'c' engine does not support regex separators (separators > 1 char and different from '\s+' are interpreted as regex), but this causes 'low_memory' to be ignored as it is not supported by the 'python' engine.
```

georgesittas:
Fixes #2681

### DIFF
--- a/sqlmesh/core/model/seed.py
+++ b/sqlmesh/core/model/seed.py
@@ -39,7 +39,14 @@ class CsvSettings(PydanticModel):
     def _str_validator(cls, v: t.Any) -> t.Optional[str]:
         if v is None or not isinstance(v, exp.Expression):
             return v
-        return v.this
+
+        v = v.this
+        if v == "\\t":
+            return "\t"
+        if v == "\\\\":
+            return "\\"
+
+        return v
 
 
 class CsvSeedReader:
--- a/sqlmesh/core/model/seed.py
+++ b/sqlmesh/core/model/seed.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import pandas as pd
 from sqlglot import exp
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/seed.py
```diff
@@ -39,7 +39,14 @@ def _bool_validator(cls, v: t.Any) -> t.Optional[bool]:
     def _str_validator(cls, v: t.Any) -> t.Optional[str]:
         if v is None or not isinstance(v, exp.Expression):
             return v
-        return v.this
+
+        v = v.this
+        if v == "\\t":
```
izeigerman, line 44: Do we need to take into account other escaped characters like `\r`, `\n`, etc?

izeigerman (approved):

sqlmesh/core/model/seed.py
georgesittas: Yeah will amend by doing a quick sqlglot refactor.

georgesittas: Addressed via https://github.com/tobymao/sqlglot/commit/89a8984b8db3817d934b4395e190f3848b1ee77a and https://github.com/TobikoData/sqlmesh/pull/2690/commits/f6429761f8ed66f3855bf6030e5edba26a2c23c0.

PR #2697, opened 2024-05-29
Feat: Support name inference from schema directories

Themiscodes:
This is an opt-in feature that controls whether model names in your project are automatically inferred from their directory structure. The `model_naming` config is used to enable it:

```yaml
model_naming:
  infer_names: true
```

- **Default Behavior**: If `model_naming` is not defined or if `infer_names` is set to `false` model names must be explicitly provided.
- **Inferred Names**: When `infer_names` is set to `true`, model names are inferred based on their file paths. 

### Example

For example with inference enabled and assuming the file structure is:

  ```
  models/
    ├── catalog/
    │   └── schema/
    │       └── model.sql
    └── other_model.sql
  ```

The inferred names would be:

  - `models/catalog/schema/model.sql` would be named `catalog.schema.model`.
  - `models/other_model.sql` would be named `other_model`.

### Overriding Inferred Names

If you want to provide a specific name for a model, you can do so directly in the model definition, and this name will take precedence over the inferred one. For example:

```yaml
# File: models/catalog/schema/model.sql
name: custom_model_name
```

In this case, despite the inferred name being `catalog.schema.model`, the model will be named `custom_model_name`.


sqlmesh/core/model/definition.py
```diff
@@ -1492,6 +1492,15 @@ def load_sql_based_model(
         meta_fields["dialect"] = meta_fields["dialect"].name
 
     name = meta_fields.pop("name", "")
+
+    # If the sql file is under a schema directory infer the name from path
+    if (
+        (not name or not name.db)
```
georgesittas, line 1498: We only care if `name` is not present. The `name.db` is irrelevant in this condition.

sqlmesh/core/model/definition.py
```diff
@@ -1492,6 +1492,15 @@ def load_sql_based_model(
...
+        and "models" in path.parts
+        and len(path.parts) - path.parts.index("models") == 3
+    ):
+        name = f"{path.parts[-2]}.{path.stem}"
```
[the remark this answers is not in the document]
georgesittas: Finally, what about catalgos? For example, we could have: `models/catalog/schema/name.sql`. I believe we wanted to name this as `catalog.schema.name`.

sqlmesh/core/model/definition.py
Themiscodes: Condition was altered. Thought this was needed to be handled as well but now it defaults to setting to `default.foo`.

sqlmesh/core/model/definition.py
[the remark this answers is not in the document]
Themiscodes: Thank you you're right addressed the above issues by removing the redundant checks. Also to account for catalogs so now this is handled by setting the name using the catalog information as well. Tests were adapted accordingly to check for this behaviour, let me know what you think

sqlmesh/core/model/definition.py
```diff
@@ -1492,6 +1492,14 @@ def load_sql_based_model(
         meta_fields["dialect"] = meta_fields["dialect"].name
 
     name = meta_fields.pop("name", "")
+
+    # If the sql file is under a schema directory infer the name from path
+    if not name:
+        name = path.stem
+        subdirectories = path.parts[path.parts.index("models") + 1 : -1]
+        if subdirectories:
+            name = ".".join(subdirectories) + "." + name
```
izeigerman, line 1501: What if the nestedness is more than 3 levels deep? Should we fail then? Or only take the last 3 parts?

How about Python models?

tests/core/test_model.py
```diff
@@ -4445,3 +4445,42 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
...
+        ],
+        [
+            """dialect snowflake,""",
+            """models/test_model.sql,""",
+            "test_model",
+        ],
+        [
+            """dialect snowflake,""",
+            """models/catalog1/catalog2/test_schema/test_model.sql,""",
+            "catalog1.catalog2.test_schema.test_model",
```
izeigerman, line 4466: Is this a correct name?

tobymao: no, this is not a legal name

sqlmesh/core/model/definition.py
Themiscodes: For python models I overriden the `__call__` method in the model `decorator` to derive the name from the path there, I'm open to suggestions though if we should handle it in a better way.

For the nestedness I adapted the code to take the last 3 parts.

tests/core/test_model.py
Themiscodes: Logic was adapted to go only 3 levels deep as discussed, so this test was adapted accordingly.

sqlmesh/core/model/decorator.py
```diff
@@ -18,16 +24,26 @@
 logger = logging.getLogger(__name__)
 
 
+DECORATOR_RETURN_TYPE = t.TypeVar("DECORATOR_RETURN_TYPE")
```
georgesittas, line 27: We shouldn't redefine this here, it's already available in `sqlmesh.core.model`.

sqlmesh/core/model/decorator.py
```diff
@@ -18,16 +24,26 @@
...
 
-    def __init__(self, name: str, is_sql: bool = False, **kwargs: t.Any) -> None:
-        if not name:
+    def __call__(
+        self, func: t.Callable[..., DECORATOR_RETURN_TYPE]
+    ) -> t.Callable[..., DECORATOR_RETURN_TYPE]:
+        self.name = self.name or get_model_name(Path(inspect.getfile(func)))
+
+        if not self.name:
             raise ConfigError("Python model must have a name.")
```
georgesittas, line 42: 1. Can this ever _not_ be set after these changes?
2. The `ConfigError` doesn't really make sense anymore, since we supposedly don't require the model to have a name anymore.

georgesittas: If we do it like this, we should move the `kind is not None` branch within `__call__`, otherwise `name` would be interpolated as `None` in the warning.

sqlmesh/core/model/definition.py
```diff
@@ -2122,3 +2124,12 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
     "allow_partials": exp.convert,
     "signals": lambda values: exp.Tuple(expressions=values),
 }
+
+
+def get_model_name(path: Path) -> str:
+    name = path.stem
+    subdirectories = path.parts[path.parts.index("models") + 1 : -1]
```
georgesittas, line 2131: Maybe let's gracefully handle `"models"` not existing in `path.parts`. If for whatever reason that happens, it'll be better UX if we say "If you wanna skip the name in the model, then you need to define it in a file under the project's `models/` directory".

sqlmesh/core/model/decorator.py
```diff
@@ -43,7 +59,7 @@ def __init__(self, name: str, is_sql: bool = False, **kwargs: t.Any) -> None:
                         f"""Python model "{name}"'s `kind` dictionary must contain a `name` key with a valid ModelKindName enum value."""
                     )
 
-        self.name = name
+        self.name = name or ""
```
georgesittas, line 62: Should we comment that this is temporarily set to `""` to allow the `__call__` method to infer the name based on the model's path?

sqlmesh/core/model/definition.py
```diff
@@ -1491,7 +1491,9 @@ def load_sql_based_model(
     if isinstance(meta_fields.get("dialect"), exp.Expression):
         meta_fields["dialect"] = meta_fields["dialect"].name
 
-    name = meta_fields.pop("name", "")
+    # If the sql file is under a schema directory infer the name from path
```
georgesittas, line 1494:
```suggestion
    # The name of the model will be inferred from its path relative to `models/`, if it's not explicitly specified
```

tests/core/test_model.py
```diff
@@ -4445,3 +4445,42 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
             """
         )
         load_sql_based_model(expressions)
+
+
+@pytest.mark.parametrize(
+    ["model_def", "path", "assertion"],
```
georgesittas, line 4451:
```suggestion
    ["model_def", "path", "expected_name"],
```

tests/core/test_model.py
```diff
@@ -4445,3 +4445,42 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
...
+        [
+            """dialect duckdb,""",
+            """models/inventory/db/test_schema/test_model.sql,""",
+            "db.test_schema.test_model",
+        ],
+        ["""name test_model,""", """models/schema/test_model.sql,""", "test_model"],
+    ],
+)
+def test_model_table_name_inference(
+    sushi_context: Context, model_def: str, path: str, assertion: str
```
georgesittas, line 4472:
```suggestion
    sushi_context: Context, model_def: str, path: str, expected_name: str
```

tests/core/test_model.py
```diff
@@ -4445,3 +4445,42 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
...
+        MODEL (
+            {model_def}
+        );
+        SELECT a FROM tbl;
+        """,
+            default_dialect="duckdb",
+        ),
+        path=Path(f"$root/{path}"),
+    )
+    assert model.name == assertion
```
georgesittas, line 4486:
```suggestion
    assert model.name == expected_name
```

sqlmesh/core/model/definition.py
Themiscodes: I think this is ok since it is checked before. In `/cli/main.py::cli` if the project doesn't have any models:
`if load and not context.models:`
it raises an exception. Though I can add an additional check here as well

sqlmesh/core/model/decorator.py
Themiscodes: Yes correct I'm moving this part of code to `__call__`. The name should also be derived always from the filename so the check is redundant.

sqlmesh/core/model/definition.py
georgesittas: Btw, I think we can simplify this function by reducing it to:

```python
path_parts = path.parts[...] + [path.stem]
return ".".join(path_parts[-3:]
```

plaflamme:
Will this be an opt-in feature? Our directory structure doesn't map to schema / table names directly, so we don't ever want this feature to kick in for us; for example, we'd much rather have an error like "name is missing" at load time than any weird behaviour at runtime.

sqlmesh/core/config/naming.py
```diff
@@ -0,0 +1,14 @@
...
+
+class NameInferenceConfig(BaseConfig):
+    """Configuration for name inference of models from directory structure.
+
+    Args:
+        name_inference: A flag indicating whether name inference is enabled.
+                        Default is False.
+    """
+
+    name_inference: bool = False
```
tobymao, line 14: nit: can we replace this and all refs to infer_names

Themiscodes: Sure done!

tobymao (approved):
need documentation

sqlmesh/core/config/naming.py
```diff
@@ -0,0 +1,14 @@
...
+
+from sqlmesh.core.config.base import BaseConfig
+
+
+class NameInferenceConfig(BaseConfig):
+    """Configuration for name inference of models from directory structure.
+
+    Args:
+        infer_names: A flag indicating whether name inference is enabled.
+                        Default is False.
```
georgesittas, line 11: This seems redundant:
```suggestion
```

sqlmesh/core/config/naming.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+from sqlmesh.core.config.base import BaseConfig
+
+
+class NameInferenceConfig(BaseConfig):
+    """Configuration for name inference of models from directory structure.
```
georgesittas, line 7: Did we conclude this deserves its own config class? @izeigerman thoughts?

sqlmesh/core/model/decorator.py
```diff
@@ -24,26 +30,17 @@ class model(registry_decorator):
     registry_name = "python_models"
     _dialect: DialectType = None
 
-    def __init__(self, name: str, is_sql: bool = False, **kwargs: t.Any) -> None:
-        if not name:
-            raise ConfigError("Python model must have a name.")
+    def __call__(
+        self, func: t.Callable[..., DECORATOR_RETURN_TYPE]
+    ) -> t.Callable[..., DECORATOR_RETURN_TYPE]:
+        return super().__call__(func)
```
georgesittas, line 36: It looks like after the recent refactor this became redundant. Let's remove.

sqlmesh/core/model/decorator.py
```diff
@@ -24,26 +30,17 @@ class model(registry_decorator):
...
-                    f"""Python model "{name}"'s `kind` argument was passed a SQLMesh `{type(kind).__name__}` object. This may result in unexpected behavior - provide a dictionary instead."""
-                )
-            elif isinstance(kind, dict):
-                if "name" not in kind or not isinstance(kind.get("name"), ModelKindName):
-                    raise ConfigError(
-                        f"""Python model "{name}"'s `kind` dictionary must contain a `name` key with a valid ModelKindName enum value."""
-                    )
-
-        self.name = name
+        # Set the name to an empty string, allowing __call__ method to infer the name based on the model's path
```
georgesittas, line 42: This comment is outdated as well, I'd just remove it altogether.

sqlmesh/core/model/decorator.py
```diff
@@ -90,10 +87,31 @@ def model(
...
+        infer_names: t.Optional[bool] = False,
     ) -> Model:
         """Get the model registered by this function."""
         env: t.Dict[str, t.Any] = {}
         entrypoint = self.func.__name__
+        self.name = (
+            self.name
+            if not infer_names
+            else self.name or get_model_name(Path(inspect.getfile(self.func)))
+        )
```
georgesittas, line 99:
```suggestion
        if not self.name and infer_names:
            self.name = get_model_name(Path(inspect.getfile(self.func)))
```

sqlmesh/core/model/definition.py
```diff
@@ -1491,7 +1492,13 @@ def load_sql_based_model(
     if isinstance(meta_fields.get("dialect"), exp.Expression):
         meta_fields["dialect"] = meta_fields["dialect"].name
 
-    name = meta_fields.pop("name", "")
+    # The name of the model will be inferred from its path relative to `models/`, if it's not explicitly specified
+    name = (
+        meta_fields.pop("name", "")
+        if not infer_names
+        else meta_fields.pop("name", "") or get_model_name(path)
+    )
```
georgesittas, line 1500:
```suggestion
    name = meta_fields.pop("name", "")
    if not name and infer_names:
        name = get_model_name(path)
```

tests/core/test_model.py
```diff
@@ -1687,26 +1687,38 @@ def b_model(context):
...
+    @model("kind_empty_dict", kind=dict(), columns={'"COL"': "int"})
+    def my_model(context):
+        pass
+
     # error if kind dict with no `name` key
     with pytest.raises(ConfigError, match="`kind` dictionary must contain a `name` key"):
+        python_model = model.get_registry()["kind_empty_dict"].model(
+            module_path=Path("."),
+            path=Path("."),
+        )
```
georgesittas, line 1699: Correct me if I'm wrong, but it looks like we haven't tested the name inference for python models. If that's true, let's add one.

tests/core/test_model.py
```diff
@@ -4445,3 +4457,43 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
             """
         )
         load_sql_based_model(expressions)
+
+
+@pytest.mark.parametrize(
+    ["model_def", "path", "expected_name"],
+    [
+        [
+            """dialect duckdb,""",
```
georgesittas, line 4466: Can't we get rid of `model_def` altogether? This doesn't look necessary for the test. The `dialect` property can be left unspecified.

docs/guides/configuration.md
```diff
@@ -944,6 +944,42 @@ This example demonstrates how to specify an incremental by time range model kind
 
 Learn more about specifying Python models at the [Python models concepts page](../concepts/models/python_models.md#model-specification).
 
+
+#### Model Naming
+
+The `model_naming` configuration controls if model names are inferred based on the schema directory structure. If `model_naming` is not defined or `infer_names` is set to false, the model names must be provided explicitly.
```
georgesittas, line 950:
```suggestion
The `model_naming` configuration controls if model names are inferred based on the project's directory structure. If `model_naming` is not defined or `infer_names` is set to false, the model names must be provided explicitly.
```

docs/guides/configuration.md
```diff
@@ -944,6 +944,42 @@ This example demonstrates how to specify an incremental by time range model kind
...
+
+With `infer_names` set to true, model names are inferred based on their path. For example, a model located at `/models/catalog/schema/model.sql` would be named `catalog.schema.model`. However, if a name is provided in the model definition, it will take precedence over the inferred name.
```
georgesittas, line 952:
```suggestion
With `infer_names` set to true, model names are inferred based on their path. For example, a model located at `models/catalog/schema/model.sql` would be named `catalog.schema.model`. However, if a name is provided in the model definition, it will take precedence over the inferred name.
```

docs/guides/configuration.md
```diff
@@ -944,6 +944,42 @@ This example demonstrates how to specify an incremental by time range model kind
...
+
+Supported values:
+
+* `true`: Automatically infer model names based on the directory structure.
+* `false`: Do not infer model names automatically.
```
georgesittas, line 957: Imo this looks redundant, given that this setting is obviously a boolean flag. Let's see what others think too.

docs/guides/configuration.md
```diff
@@ -944,6 +944,42 @@ This example demonstrates how to specify an incremental by time range model kind
...
+=== "YAML"
+
+    ```yaml linenums="1"
+    model_naming:
+      infer_names: false
+    ```
+
+=== "Python"
+
+    The Python `model_naming` argument takes a configuration object with the `infer_names` attribute set to `True` or `False`.
```
georgesittas, line 970: Should this be outside the `"Python"` code block?

docs/reference/model_configuration.md
```diff
@@ -10,7 +10,7 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
...
 | --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :---------------: | :------: |
-| `name`                | The model name. Must include at least a qualifying schema (`<schema>.<model>`) and may include a catalog (`<catalog>.<schema>.<model>`).                                                                                                                                                         |        str        |    Y     |
+| `name`                | The model name. Must include at least a qualifying schema (`<schema>.<model>`) and may include a catalog (`<catalog>.<schema>.<model>`).                                                                                                                                                         |        str        |    N     |
```
georgesittas, line 13: Update this as well to shortly mention the name inference option (perhaps link the relevant section?)

docs/reference/model_configuration.md
```diff
@@ -48,6 +48,16 @@ The SQLMesh project-level `model_defaults` key supports the following options, d
...
 - on_destructive_change (described [below](#incremental-models))
 
+
+### Model Naming
+
+Configuration option for name inference. Learn more in the [model naming guide](../guides/configuration.md#model-naming).
+
+| Option          | Description                                                                             |  Type   | Required |
+| --------------- | --------------------------------------------------------------------------------------- | :-----: | :------: |
+| `infer_names`   | Whether to automatically infer model names based on the directory structure (Default: true) | boolean |    N     |
```
georgesittas, line 58: Is this table going to be rendered correctly?

georgesittas (approved):
Nice work @Themiscodes 👍

cc @treysp can you take a quick look at the docs to verify if all relevant sections have been properly updated?

georgesittas:
> Will this be an opt-in feature? Our directory structure doesn't map to schema / table names directly, so we don't ever want this feature to kick in for us; for example, we'd much rather have an error like "name is missing" at load time than any weird behaviour at runtime.

FYI @plaflamme, PR has been updated since and this is now an opt-in feature.

docs/reference/model_configuration.md
Themiscodes: Yes though the default and type were wrong I corrected them

tests/core/test_model.py
georgesittas: Nvm, ignore this

tests/core/test_model.py
```diff
@@ -4450,3 +4468,77 @@ def test_incremental_by_partition(sushi_context, assert_exp_eq):
...
+        [
+            """models/inventory/db/test_schema/test_model.py""",
+            "db.test_schema.test_model",
+        ],
+    ],
+)
+def test_python_model_name_inference(tmp_path: Path, path: str, expected_name: str) -> None:
+    init_example_project(tmp_path, dialect="duckdb")
+    config = Config(
+        default_connection=DuckDBConnectionConfig(),
```
georgesittas, line 4529: This is the [default value](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/config/root.py#L81) for `default_connection` so no need to set it here:
```suggestion
```

georgesittas (approved):
LGTM, only pending [question](https://github.com/TobikoData/sqlmesh/pull/2697/files#r1624251339) from me is related to the config type addition.

tobymao (approved):

PR #2703, opened 2024-05-29
feat: databricks allow disabling spark session

eakmanrq:
Prior to this change if a user was running in a databricks notebook we assumed they would want to use the local spark session. This isn't potentially the case though if the user wants to connect to new Serverless SQL from a notebook. So this change allows the user to disable spark session access so they can connect to Serverless SQL.


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -97,11 +100,15 @@ def spark(self) -> PySparkSession:
...
-                token=self._extra_config["databricks_connect_access_token"],
-                cluster_id=self._extra_config["databricks_connect_cluster_id"],
-            ).getOrCreate()
+            self._spark = (
+                DatabricksSession.builder.remote(
+                    host=self._extra_config["databricks_connect_server_hostname"],
+                    token=self._extra_config["databricks_connect_access_token"],
+                    cluster_id=self._extra_config["databricks_connect_cluster_id"],
+                )
+                .userAgent("sqlmesh")
```
eakmanrq, line 109: Missed this `userAgent` when I added support before.

docs/integrations/engines/databricks.md
```diff
@@ -38,6 +38,7 @@ Note: If using Databricks Connect please note the [requirements](https://docs.da
...
 | `force_databricks_connect`           | When running locally, force the use of Databricks Connect for all model operations (so don't use SQL Connector for SQL models)                                                                                                                                                                  |  bool  |    N     |
 | `disable_databricks_connect`         | When running locally, disable the use of Databricks Connect for all model operations (so use SQL Connector for all models)                                                                                                                                                                      |  bool  |    N     |
+| `disable_spark_session`              | Do not use SparkSession if it is available (like when running in a notebook).                                                                                                                                                                                                                   |  bool  |    N     |
```
treysp, line 41: Does this feature allow you to not use SparkSession even if it is available, or it allows you to not get warnings when it's not available?

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -47,12 +46,20 @@ def __init__(self, *args: t.Any, **kwargs: t.Any):
...
 
-    @classproperty
-    def can_access_spark_session(cls) -> bool:
+    @classmethod
+    def can_access_spark_session(cls, disable_spark_session: bool) -> bool:
         from sqlmesh import RuntimeEnv
 
-        if RuntimeEnv.get().is_databricks:
-            return True
+        if disable_spark_session:
```
izeigerman, line 53: This seems counterintuitive to me. IMHO if `disable_spark_session` is False we shouldn't even call this method, no? What's the point of calling it to immediately return from it?

eakmanrq: To make sure the caller considers the possibility of the spark session being disabled. It is easy to just say if I'm in a notebook then use SparkSession and so I wanted to make sure when someone checks for this we bundle in the check for disabling spark session.

docs/integrations/engines/databricks.md
eakmanrq: The former. So when running in a Databricks notebook you will always have a SparkSession available and this lets you say even though it is available you don't want to use it.

sqlmesh/core/engine_adapter/databricks.py
eakmanrq: I can't rely on state of the class since we do this check from outside and inside the class (back in connection config).

izeigerman (approved):

PR #2712, opened 2024-05-31
Feat: column comments in python models

### LINKED ISSUE #2710: How to add comments to columns using a Python model
Hi,
When creating a sql model you can add comments to every column and it will get registered in the table and view schemas.
Is it possible to do the same when creating a Python model?

I'm using pyspark

treysp:
Closes #2710


georgesittas (approved):

sqlmesh/core/model/meta.py
```diff
@@ -62,7 +62,9 @@ class ModelMeta(_Node):
     default_catalog: t.Optional[str] = None
     depends_on_: t.Optional[t.Set[str]] = Field(default=None, alias="depends_on")
     columns_to_types_: t.Optional[t.Dict[str, exp.DataType]] = Field(default=None, alias="columns")
-    column_descriptions_: t.Optional[t.Dict[str, str]] = None
+    column_descriptions_: t.Optional[t.Dict[str, str]] = Field(
+        default=None, alias="column_descriptions"
```
izeigerman, line 66: How did this work before? Couldn't use specify `column_descriptions` in a SQL model?

izeigerman (approved):

tobymao (approved):

sqlmesh/core/model/meta.py
treysp: We automatically infer them from inline comments in SQL models - I guess you technically could have passed a `column_descriptions_` key.

I described it as a "python model only" key in the docs, but this will allow it to be passed in SQL models too. Hmm...

sqlmesh/core/model/meta.py
```diff
@@ -212,6 +214,41 @@ def _columns_validator(
...
+                for v in vs
+            }
+        )
+
+        columns_to_types = values.get("columns_to_types_")
+        if columns_to_types:
+            for column_name in col_descriptions:
+                if column_name not in columns_to_types:
+                    raise ConfigError(
+                        f"In model '{values.get('name')}', a description is provided for column '{column_name}' but it is not a column in the model."
```
georgesittas, line 247: I think the name should always be available here?
```suggestion
                        f"In model '{values['name']}', a description is provided for column '{column_name}' but it is not a column in the model."
```

docs/concepts/models/overview.md
```diff
@@ -75,11 +75,31 @@ Some SQL engines support registering comments as metadata associated with a tabl
...
+#### Model comment
 
 SQLMesh will register a comment specified before the `MODEL` DDL block as the table comment in the underlying SQL engine. If the [`MODEL` DDL `description` field](#description) is also specified, SQLMesh will register it with the engine instead.
 
-SQLMesh will automatically detect comments in a query's column selections and register each column's final comment in the underlying SQL engine.
+#### Explicit column comments
+
+You may explicitly specify column comments in the `MODEL` DDL `column_descriptions` field.
+
+Specify them as an dictionary of key/value pairs separated by an equals sign `=`, where the column name is the key and the column comment is the value. For example:
```
georgesittas, line 86:
```suggestion
Specify them as a dictionary of key/value pairs separated by an equals sign `=`, where the column name is the key and the column comment is the value. For example:
```

docs/concepts/models/overview.md
```diff
@@ -75,11 +75,31 @@ Some SQL engines support registering comments as metadata associated with a tabl
...
+    id = 'This is the ID column comment'
+  )
+);
+```
+
+If the `column_descriptions` key is present, SQLMesh will not detect and register inline column comments from the model query.
+
+#### Inline column comments
+
+If the `column_descriptions` key is not present in the `MODEL` definition, SQLMesh will automatically detect comments in a query's column selections and register each column's final comment in the underlying SQL engine.
```
georgesittas, line 102: Nit: not sure if `final` is the right term to use here. Do we use that consistently? Perhaps we can get rid of it altogether if we explain that only comments _following_ the projection (e.g., after the comma) are registered.

sqlmesh/core/model/meta.py
```diff
@@ -212,6 +214,34 @@ def _columns_validator(
...
+            return None
+
+        if isinstance(vs, exp.Paren):
+            vs = vs.flatten()
+
+        if isinstance(vs, (exp.Tuple, exp.Array)):
+            vs = vs.expressions
+
+        col_descriptions = (
+            vs if isinstance(vs, dict) else {v.this.name: v.expression.alias_or_name for v in vs}
```
georgesittas, line 232: I believe we can _slightly_ simplify this; there should be no alias in the right-hand side, so `alias_or_name` doesn't seem suitable:
```suggestion
            vs if isinstance(vs, dict) else {v.name: v.expression.name for v in vs}
```

georgesittas (approved):
LGTM - final minor comments to consider.

docs/concepts/models/overview.md
treysp: I may not understand how SQLMesh behaves - what gets registered if there are two comments after the comma?

georgesittas: Ah I forgot I changed this recently. 🤦

> what gets registered if there are two comments after the comma

The final one, so what you have here is actually the correct thing!

PR #2715, opened 2024-05-31
feat: add the ability to have audits on external models

tobymao:


sqlmesh/core/model/meta.py
```diff
@@ -108,16 +108,33 @@ def extract(v: exp.Expression) -> t.Tuple[str, t.Dict[str, exp.Expression]]:
...
+                if isinstance(entry, tuple):
+                    audits.append((
+                        entry[0].lower(),
+                        {
+                            key: d.parse_one(value) if isinstance(value, str) else value
+                            for key, value in entry[1].items()
+                        },
+                    ))
+                elif isinstance(entry, dict):
+                    if len(entry) != 1:
```
izeigerman, line 123: Not sure I understand what the format is going to look like. Should we add a test?

sqlmesh/core/snapshot/definition.py
```diff
@@ -742,6 +745,11 @@ def merge_intervals(self, other: t.Union[Snapshot, SnapshotIntervals]) -> None:
             for start, end in other.dev_intervals:
                 self.add_interval(start, end, is_dev=True)
 
+
+    @property
+    def has_intervals(self) -> bool:
```
izeigerman, line 750: Should we pick some purpose-driven name? Maybe something like `is_evaluatable`?

tests/core/test_snapshot.py
```diff
@@ -1746,3 +1746,10 @@ def test_missing_intervals_node_start_end(make_snapshot):
     assert missing_intervals([snapshot], start="2024-03-01", end=to_datetime("2024-03-10")) == {}
     assert missing_intervals([snapshot], start="2024-03-13", end="2024-03-14") == {}
     assert missing_intervals([snapshot], start="2024-03-14", end="2024-03-30") == {}
+
+
+def test_external_model_audits(sushi_context):
```
izeigerman, line 1751: I think we also need a test in the `test_snapshot_evaluator.py` to ensure that audits are actually called for external model snapshots.

izeigerman (approved):

PR #2717, opened 2024-06-01
External models default yaml

dbrtly:
As discussed, renaming schema.yaml to external_models.yaml for consistency.

This was all the references I could find in this repo.
I can update the example repos too.

Open invitation to nitpick as much as you want. Happy to take on any/all feedback.


web/client/src/library/components/editor/EditorFooter.tsx
```diff
@@ -130,6 +130,8 @@ function getFileType(path?: string): FileType {
   if (path.startsWith('metrics')) return EnumFileType.Metric
   if (['config.yaml', 'config.yml', 'config.py'].includes(path))
     return EnumFileType.Config
+  if (['external_models.yaml', 'external_models.yml'].includes(path))
```
tobymao, line 133: why did you add yml? we don’t support that right

sqlmesh/core/context.py
```diff
@@ -1598,17 +1598,17 @@ def rollback(self) -> None:
...
+        The external models file contains all columns and types of external models, allowing for more
+        robust lineage, validation, and optimizations.
         """
         if not self._models:
             self.load(update_schemas=False)
 
         for path, config in self.configs.items():
-            create_schema_file(
-                path=path / c.SCHEMA_YAML,
+            create_external_models_file(
```
izeigerman, line 1610: To preserve compatibility with existing projects should we check whether a file with the old name already exists and use the old name if it does?

sqlmesh/core/loader.py
```diff
@@ -183,9 +183,13 @@ def _load_external_models(self) -> UniqueKeyDict[str, Model]:
         models: UniqueKeyDict[str, Model] = UniqueKeyDict("models")
         for context_path, config in self._context.configs.items():
             schema_path = Path(context_path / c.SCHEMA_YAML)
+            external_models_default_path = context_path / c.EXTERNAL_MODELS_DEFAULT_YAML
             external_models_path = context_path / c.EXTERNAL_MODELS
 
             paths_to_load = []
+            if external_models_default_path.exists():
```
izeigerman, line 190: Should we just load one or other rather than both? In other words:
```
if external_models_default_path.exists():
    ...
else if schema_path.exists():
    ...
```

sqlmesh/core/context.py
```diff
@@ -1598,17 +1598,22 @@ def rollback(self) -> None:
...
         """
         if not self._models:
             self.load(update_schemas=False)
 
         for path, config in self.configs.items():
-            create_schema_file(
-                path=path / c.SCHEMA_YAML,
+            external_models_yaml = (
+                c.EXTERNAL_MODELS_DEFAULT_YAML
+                if not (Path(self.path / c.EXTERNAL_MODELS_ALTERNATE_YAML)).exists()
```
izeigerman, line 1612: This should be just `path` instead of `self.path`.

sqlmesh/core/loader.py
```diff
@@ -182,12 +182,15 @@ def _load_metrics(self) -> UniqueKeyDict[str, MetricMeta]:
     def _load_external_models(self) -> UniqueKeyDict[str, Model]:
         models: UniqueKeyDict[str, Model] = UniqueKeyDict("models")
         for context_path, config in self._context.configs.items():
-            schema_path = Path(context_path / c.SCHEMA_YAML)
+            external_models_default_yaml = Path(context_path / c.EXTERNAL_MODELS_DEFAULT_YAML)
+            external_models_alternate_yaml = Path(context_path / c.EXTERNAL_MODELS_ALTERNATE_YAML)
```
izeigerman, line 186: Ditto about alternate and default.

sqlmesh/core/context.py
izeigerman: Also we don't need another `Path` instance, it can just be:
```
if not (path / c.EXTERNAL_MODELS_ALTERNATE_YAML).exists()
```

izeigerman: Additionally, if we want to avoid creating the full path twice we can something like:
```
deprecated_yaml = path / c.EXTERNAL_MODELS_DEPRECATED_YAML

external_models_yaml = (
    path / c.EXTERNAL_MODELS_YAML
    if not deprecated_yaml.exists()
    else deprecated_yaml
)

create_external_models_file(
    path=external_models_yaml,
    ...
)
```

sqlmesh/core/context.py
```diff
@@ -1598,17 +1598,24 @@ def rollback(self) -> None:
...
-                path=path / c.SCHEMA_YAML,
+            deprecated_yaml = path / c.EXTERNAL_MODELS_DEPRECATED_YAML
+
+            external_models_yaml = (
+                path / c.EXTERNAL_MODELS_YAML
+                if not deprecated_yaml.exists()
+                else deprecated_yaml
+            )
+            create_external_models_file(
+                path=path / external_models_yaml,
```
izeigerman, line 1618:
```suggestion
                path=external_models_yaml,
```
`external_models_yaml` already contains the full path

sqlmesh/core/constants.py
```diff
@@ -40,15 +40,16 @@
...
 METRICS = "metrics"
 MODELS = "models"
-EXTERNAL_MODELS = "external_models"
 SEEDS = "seeds"
 TESTS = "tests"
-CACHE = ".cache"
-SCHEMA_YAML = "schema.yaml"
 
+EXTERNAL_MODELS_YAML = "external_models.yaml"
+DEPRECATED_YAML = "schema.yaml"
```
izeigerman, line 52: I liked the previous name better :D Now it's quite unclear which YAML has been deprecated. Sorry about the back and forth.

dbrtly:
btw I'm anticipating that the reviewer resolves the conversations.

izeigerman (approved):
This looks great. Thank you so much for addressing comments 👍

PR #2722, opened 2024-06-03
Refactor!: normalize environment names according to the engine's dialect

georgesittas:
We previously treated environment names as case-insensitive, lowercasing them regardless of the target engine.

This was problematic for engines like Snowflake, because it essentially forced users to use quotes in queries that referenced tables within these environments. So, for example instead of FOO_DEV, we'd create FOO_dev.

This PR aims to address this issue by normalizing the environment names (& the `default` schema) according to the rules of the targeted engine's dialect. So, for the above example users will now be able to query FOO_DEV.table without using quotes for the schema part.

This PR should not affect existing environments, which is achieved by setting their `normalize_name` column to `FALSE`. Hence that column will only be `TRUE` for new environments.


sqlmesh/core/snapshot/definition.py
```diff
@@ -1291,10 +1314,12 @@ def table_name(
     return exp.table_name(table)
 
 
+# TODO: should we find a way to propagate the engine dialect into `for_environment` here?
```
georgesittas, line 1317:
```suggestion
```

sqlmesh/core/plan/builder.py
```diff
@@ -137,6 +137,7 @@ def __init__(
             environment_catalog_mapping or {},
             name=self._context_diff.environment,
             suffix_target=environment_suffix_target,
+            normalize_name=self._context_diff.is_new_environment,
```
tobymao, line 140: is this for backwards compatibility? how does this work?

sqlmesh/migrations/v0052_add_normalize_name_in_environment_naming_info.py
```diff
@@ -0,0 +1,27 @@
...
+from sqlglot import exp
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
+    engine_adapter = state_sync.engine_adapter
+    environments_table = "_environments"
+    if state_sync.schema:
+        environments_table = f"{state_sync.schema}.{environments_table}"
+
+    alter_table_exp = exp.AlterTable(
```
tobymao, line 12: in the future, we should make a builder for alter table, low pri

sqlmesh/core/snapshot/definition.py
```diff
@@ -213,22 +219,36 @@ def catalog_for_environment(
...
         if (
             environment_naming_info.name.lower() != c.PROD
             and environment_naming_info.suffix_target.is_schema
         ):
-            schema = f"{schema}__{environment_naming_info.name}"
+            env_name = environment_naming_info.name
+            if environment_naming_info.normalize_name:
+                env_name = normalize_identifiers(env_name, dialect=dialect).name
+
+            schema = f"{schema}__{env_name}"
```
tobymao, line 234: how does this work? do we not normalize schema? doesn't that mean it could be

sqlmesh_DEV ?

tobymao: or default_DEV

sqlmesh/core/plan/builder.py
georgesittas: Correct, new environments should be normalized but old ones shouldn't.

sqlmesh/migrations/v0052_add_normalize_name_in_environment_naming_info.py
georgesittas: yep 👍

sqlmesh/core/snapshot/definition.py
georgesittas: Ah thanks, I think this is an oversight. I'll do another pass at this soon and make sure we normalize everything.

georgesittas: I think we should only normalize the `default` schema, right? If a user's set their own schema, then that should have been normalized prior to flowing into this method. I verified this in my tests because `sqlmesh plan foo` ended up creating `FOO__DEV`.

georgesittas:
I plan to wrap this up soon. Remaining items on my end are:

- [ ] Normalize default schema name ([relevant discussion](https://github.com/TobikoData/sqlmesh/pull/2722#discussion_r1626847339))
- [ ] Add tests

sqlmesh/core/plan/builder.py
tobymao: why aren't old environments normalized? is this permanent?

georgesittas: I think this may be incorrect after all, thanks for double checking it. I'll revisit.

sqlmesh/core/snapshot/definition.py
georgesittas: Addressed

sqlmesh/core/plan/builder.py
georgesittas: Addressed a bit differently, see https://github.com/TobikoData/sqlmesh/pull/2722/commits/9ebcc31f5f2b75addda9c12af838a29cea0e2bd1.

TL;DR added new field (`normalize_environment_name: bool`) in `ContextDiff` which is now set as:

```
is_new_environment or bool(env and env.normalize_name)
```

So, if an environment is new, we use the new logic. Otherwise, we check if the fetched environment's `normalize_name` column is `TRUE`.

georgesittas:
This is ready for another review. I added a [simple test](https://github.com/TobikoData/sqlmesh/pull/2722/commits/9ebcc31f5f2b75addda9c12af838a29cea0e2bd1#diff-5af59db99ac14709aa0e02acc49ee767af92605439b2f23c567b4c7d0618b2b1R288-R310) that demonstrates how the default schema & environment name get normalized, but I haven't tested that old environment names won't be normalized (though I did test against snowflake and it seems to work fine). Any suggestions on how we should do this, if needed?

sqlmesh/core/console.py
```diff
@@ -251,6 +251,8 @@ def __init__(
 
         self.verbose = verbose
 
+        self.dialect = kwargs.pop("dialect", None)
```
izeigerman, line 254: Should we just make this an argument?

sqlmesh/core/console.py
```diff
@@ -1094,6 +1100,8 @@ def __init__(
         self.missing_dates_output = widgets.Output()
         self.dynamic_options_after_categorization_output = widgets.VBox()
 
+        self.dialect = kwargs.pop("dialect", None)
```
izeigerman, line 1103: ditto

sqlmesh/core/context_diff.py
```diff
@@ -183,10 +186,18 @@ def create_no_diff(cls, environment: str) -> ContextDiff:
         Returns:
             The ContextDiff object.
         """
+        env = state_reader.get_environment(environment.lower())
+
+        if env is None:
+            normalize_environment_name = True
+        else:
+            normalize_environment_name = env.normalize_name
```
izeigerman, line 194:
```
normalize_environment_name = env is None or not env.normalize_name
```

sqlmesh/core/context_diff.py
```diff
@@ -183,10 +186,18 @@ def create_no_diff(cls, environment: str) -> ContextDiff:
         Returns:
             The ContextDiff object.
         """
+        env = state_reader.get_environment(environment.lower())
```
izeigerman, line 189: [Nit] I'd probably just pass the environment here instead of the entire state reader.

izeigerman: Also makes it easier to test

sqlmesh/core/snapshot/definition.py
```diff
@@ -219,22 +225,39 @@ def catalog_for_environment(
...
         if (
             environment_naming_info.name.lower() != c.PROD
             and environment_naming_info.suffix_target.is_schema
         ):
-            schema = f"{schema}__{environment_naming_info.name}"
+            env_name = environment_naming_info.name
+            if environment_naming_info.normalize_name:
+                env_name = normalize_identifiers(env_name, dialect=dialect).name
+                if not self.schema_name:
+                    # We also normalize the schema if it's the default one
```
izeigerman, line 240: Why weren't we doing this before? A bug?

georgesittas: Yeah, I think so. If we'd set the schema earlier in the control flow, we wouldn't have this issue. But until now, we were basically interpolating a lowercase `default` schema if none was specified, i.e. `"default"."FOO"` would be created in SF.

sqlmesh/core/context_diff.py
georgesittas: That's a good point, thank you

sqlmesh/core/console.py
georgesittas: Yep I can do that

tobymao (approved):

izeigerman (approved):

PR #2724, opened 2024-06-03
Fix!: add migration script for column_descriptions_ rename

treysp:


sqlmesh/migrations/v0051_rename_column_descriptions.py
```diff
@@ -0,0 +1,61 @@
...
+    for name, identifier, version, snapshot, kind_name, expiration_ts in engine_adapter.fetchall(
+        exp.select("name", "identifier", "version", "snapshot", "kind_name", "expiration_ts").from_(
+            snapshots_table
+        ),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+
+        column_descriptions = parsed_snapshot["node"].pop("column_descriptions_", None)
+        if column_descriptions:
```
izeigerman, line 30: Should this be `is None`? I'm concerned that if it's an empty dict for whatever reason, we won't migrate.

izeigerman: Actually, checking the existence of the key will be the safest option.

treysp: Ah, good call

izeigerman (approved):

PR #2727, opened 2024-06-04
Resolve warnings magics

dbrtly:
Actually turns out to 4 warnings total resolved comparing slow-test before and after.

### DIFF
--- a/sqlmesh/schedulers/airflow/dag_generator.py
+++ b/sqlmesh/schedulers/airflow/dag_generator.py
@@ -118,7 +118,7 @@ class SnapshotDagGenerator:
 
         with DAG(
             dag_id=dag_id,
-            schedule_interval=snapshot.node.cron,
+            schedule=snapshot.node.cron,
             start_date=pendulum.instance(to_datetime(snapshot.unpaused_ts)),
             end_date=end_date,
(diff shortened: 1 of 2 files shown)


sqlmesh/schedulers/airflow/dag_generator.py
```diff
@@ -118,7 +118,7 @@ def _create_cadence_dag_for_snapshot(
 
         with DAG(
             dag_id=dag_id,
-            schedule_interval=snapshot.node.cron,
+            schedule=snapshot.node.cron,
```
izeigerman, line 121: It looks like Airflow 2.3.0 and prior still relies on `schedule_interval` (https://github.com/apache/airflow/blob/2.3.0/airflow/models/dag.py#L312). So we can't just switch to `schedule`. We should either suppress the warning, or have conditional logic which picks the attribute name based on the Airflow version.

dbrtly: Understood. Is the range of supported airflow explicit?

izeigerman: Ideally we should support the entire range of the Airflow 2 major release

PR #2736, opened 2024-06-05
Chore: increase timeout limit for engine adapter tests

georgesittas:
Lots of timeouts recently when running engine adapter tests. This should fix it.


tobymao (approved):
30 minute maybec

PR #2737, opened 2024-06-05
Fix: convert pyspark/snowpark dataframes into pandas when testing

georgesittas:
Some `pd.DataFrame` methods such as `apply` are not supported by pyspark / snowpark dataframes, which caused issues in unit tests because we assumed we'd get back pandas dataframes from the python models, and that's not always the case.

This PR fixes the issue by converting pyspark / snowpark dataframes into pandas ones for unit tests, so that there's a single representation to work with => keep the implementation simpler.

### DIFF
--- a/tests/integrations/jupyter/test_magics.py
+++ b/tests/integrations/jupyter/test_magics.py
@@ -201,9 +201,9 @@ def test_evaluate(notebook, loaded_sushi_context):
 def test_format(notebook, sushi_context):
     with capture_output():
         test_model_path = sushi_context.path / "models" / "test_model.sql"
-        test_model_path.write_text("MODEL(name db.test); SELECT 1 AS foo FROM table")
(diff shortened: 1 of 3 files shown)


tests/integrations/jupyter/test_magics.py
```diff
@@ -201,9 +201,9 @@ def test_evaluate(notebook, loaded_sushi_context):
 def test_format(notebook, sushi_context):
     with capture_output():
         test_model_path = sushi_context.path / "models" / "test_model.sql"
-        test_model_path.write_text("MODEL(name db.test); SELECT 1 AS foo FROM table")
+        test_model_path.write_text("MODEL(name db.test); SELECT 1 AS foo FROM t")
```
georgesittas, line 204: Unrelated change, will become relevant when we bump sqlglot due to https://github.com/tobymao/sqlglot/commit/3e38912cd0de2e3939221b6ad8ae194e68cfe288.

tobymao (approved):

Issue #2689, opened 2024-05-27
Support per-gateway external models


erindru:
Right now, if you have a model that depends on a dynamically named external model, it's hard to `sqlmesh plan` across gateways.

In this scenario, each gateway represents a different _application_ environment. The model structure is the same for each gateway but the naming and actual data is gateway-specific.

For example, consider a model defined like so:
```
MODEL (
   name @{gateway}_landing.my_model
);

select * from @{gateway}_app.source_table;
```

When you run `sqlmesh --gateway dev create_external_models`, this will create an entry in `schema.yml` for `dev_app.source_table`, eg:

```
- name: '"datalake"."dev_app"."source_table"'
  columns: ...
```

However, running for another gateway, eg `sqlmesh --gateway prod create_external_models` just clobbers the existing entry:
```
- name: '"datalake"."prod_app"."source_table"'
  columns: ...
```

Which means you cant run `sqlmesh --gateway dev plan` anymore without re-creating the external models. And the second you do that, cant run `sqlmesh --gateway prod plan` because the external models are now all the dev models.

You end up with warnings like:
```
WARNING - SELECT * cannot be expanded due to missing schema(s) for model(s). Run `sqlmesh create_external_models`
```
and errors like:
```
Error: Removed '"datalake"."prod_app"."source_table"' are referenced in '"datalake"."prod_landing"."my_model"'. Please remove broken references before proceeding.
```

I can see a few options here:
- Put the dynamism in `schema.yml` and resolve it on load. Eg persist the model as `- name: '"datalake"."@{gateway}_app"."source_table"'` and then resolve `@{gateway}` when it's loaded so the downstream code doesn't need to change
- Add a `gateway: ` key on writing the external models and then filter based on it when reading them back. Omitting the `gateway: ` key would mean that the external model gets loaded for all gateways. Something like:
- ```
  - name: '"datalake"."prod_app"."source_table"'
    gateway: prod
    columns: ...
  ```
- Take the opportunity to, instead of using a global `schema.yml` in the root dir, when someone runs `sqlmesh create_external_models`, write the result to `external_models/<gateway>.yml`. The external model loader could still load all of them, but perhaps a post-processing step could be done to prune the list of external models down to just the ones that are referenced in the current `sqlmesh plan` (which would be gateway-specific)

Thoughts?

georgesittas:
@erindru I assume we can close this now right?

erindru:
[...]

PR #2746, opened 2024-06-06
Feat: Gateway-specific external models

erindru:
Add an optional `gateway: ` key to the external models definitions which allows gateway-specific external models and addresses the problems highlighted in issue #2689 

After this change, the behaviour is as follows:
`sqlmesh create_external_models` - no change, current behaviour. Models get written without the new `gateway: ` key.
`sqlmesh --gateway dev create_external_models` - Models get written with `gateway: dev`
`sqlmesh --gateway prod create_external_models` - Models get written with `gateway: prod`. The `gateway: dev` models remain in the file.

When running plans:
`sqlmesh plan` - will pick up all external models with no `gateway: ` set, plus any external models with `gateway: <default gateway>`
`sqlmesh --gateway dev plan` - will pick up all external models with no `gateway: ` set, plus any external models with `gateway: dev`

### DIFF
--- a/docs/concepts/models/external_models.md
+++ b/docs/concepts/models/external_models.md
@@ -42,6 +42,45 @@ All of a SQLMesh project's external models are defined in a single `external_mod
 
 Alternatively, additional external models can also be defined in the [external_models/](#using-the-external_models-directory) folder.
 
+### Using CLI
+
+Instead of creating the `external_models.yaml` file manually, SQLMesh can generate it for you with the [create_external_models](../../reference/cli.md#create_external_models) CLI command.
+
+The command identifies all external tables referenced in your SQLMesh project, fetches their column information from the SQL engine's metadata, and then stores the information in the `external_models.yaml` file.
+
+If SQLMesh does not have access to an external table's metadata, the table will be omitted from the file and SQLMesh will issue a warning.
+
+`create_external_models` solely queries SQL engine metadata and does not query external tables themselves.
+
+### Gateway-specific external models
+
+In some use-cases such as [isolated systems with multiple gateways](../../guides/isolated_systems.md#multiple-gateways), there are external models that only exist on a certain gateway.
+
+Consider the following model that queries an external table with a dynamic database based on the current gateway:
+
+```
+MODEL (
+  name my_db.my_table,
+  kind FULL
+);
+
+SELECT
+  *
+FROM
+  @{gateway}_db.external_table;
+```
+
+This table will be named differently depending on which `--gateway` SQLMesh is run with. For example:
+
+- `sqlmesh --gateway dev plan` - SQLMesh will try to query `dev_db.external_table`
+- `sqlmesh --gateway prod plan` - SQLMesh will try to query `prod_db.external_table`
+
+To ensure SQLMesh can look up the correct schema when the relevant gateway is set, run `create_external_models` with the `--gateway` argument. For example:
+
+- `sqlmesh --gateway dev create_external_models`
+
+This will set `gateway: dev` on the external model and ensure that it is only loaded when the current gateway is set to `dev`.
+
 ### Writing YAML by hand
 
 This example demonstrates the structure of a `external_models.yaml` file:
@@ -57,21 +96,17 @@ This example demonstrates the structure of a `external_models.yaml` file:
   columns:
     column_c: bool
     column_d: float
+- name: external_db.gateway_specific_external_table
+  description: Another external table that only exists when the gateway is set to "test"
+  gateway: test
+  columns:
+    column_e: int
+    column_f: varchar
 ```
 
-It contains each `EXTERNAL` model's name, an optional description, and each of the external table's columns' name and data type.
+It contains each `EXTERNAL` model's name, an optional description, an optional gateway and each of the external table's columns' name and data type.
 
 The file can be constructed by hand using a standard text editor or IDE.
-
-### Using CLI
-
-Instead of creating the `external_models.yaml` file manually, SQLMesh can generate it for you with the [create_external_models](../../reference/cli.md#create_external_models) CLI command.
-
-The command identifies all external tables referenced in your SQLMesh project, fetches their column information from the SQL engine's metadata, and then stores the information in the `external_models.yaml` file.
-
-If SQLMesh does not have access to an external table's metadata, the table will be omitted from the file and SQLMesh will issue a warning.
-
-`create_external_models` solely queries SQL engine metadata and does not query external tables themselves.
 
 ### Using the `external_models` directory
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -340,9 +340,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
         self._test_connection_config = self.config.get_test_connection(
-            self.gateway,
-            self.default_catalog,
-            default_catalog_dialect=self.engine_adapter.DIALECT,
+            self.gateway, self.default_catalog, default_catalog_dialect=self.engine_adapter.DIALECT
         )
 
         self._snapshot_evaluator: t.Optional[SnapshotEvaluator] = None
@@ -749,9 +747,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         if expand and not isinstance(expand, bool):
             expand = {
                 normalize_model_name(
-                    x,
-                    default_catalog=self.default_catalog,
-                    dialect=self.default_dialect,
+                    x, default_catalog=self.default_catalog, dialect=self.default_dialect
                 )
                 for x in expand
             }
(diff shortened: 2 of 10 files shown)


docs/concepts/models/external_models.md
```diff
@@ -42,6 +42,45 @@ All of a SQLMesh project's external models are defined in a single `external_mod
 
 Alternatively, additional external models can also be defined in the [external_models/](#using-the-external_models-directory) folder.
 
+### Using CLI
```
erindru, line 45: Note: I moved this section up because there were references to using the CLI before the section on the CLI was introduced

sqlmesh/core/context.py
```diff
@@ -1625,6 +1630,7 @@ def create_external_models(self) -> None:
                 adapter=self._engine_adapter,
                 state_reader=self.state_reader,
                 dialect=config.model_defaults.dialect,
+                gateway=self.gateway,
```
erindru, line 1633: This is the only change I made to this file. The rest are due to `pre-commit` reformatting the file

sqlmesh/core/context.py
```diff
@@ -948,7 +952,7 @@ def plan(
             auto_apply if auto_apply is not None else self.config.plan.auto_apply,
             self.default_catalog,
             no_diff=no_diff if no_diff is not None else self.config.plan.no_diff,
-            no_prompts=no_prompts if no_prompts is not None else self.config.plan.no_prompts,
+            no_prompts=(no_prompts if no_prompts is not None else self.config.plan.no_prompts),
```
izeigerman, line 955: Has this been changed by the linter or manually?

sqlmesh/core/context.py
izeigerman: It's weird that the current build passes on main without these other changes.

izeigerman: Could it be that pre-commit using a different version / config of `ruff`?

sqlmesh/core/schema_loader.py
```diff
@@ -76,5 +81,12 @@ def _get_columns(table: str) -> t.Optional[t.Dict[str, t.Any]]:
             if columns
         ]
 
+        # dont clobber existing entries from other gateways
+        entries_to_keep = (
+            [e for e in yaml.load(path) if e.get("gateway", None) != gateway]
```
izeigerman, line 86: Wouldn't this result in duplicated entries for folks with existing external models?

sqlmesh/core/context.py
erindru: The linter. It appears to operate on just the files that you've added to the git staging area for commit. Since I modified  `context.py`, it ran for the whole file and adjusted all the formatting.

I was surprised at the number of formatting changes it made but I figured this file hadn't been touched in a while

sqlmesh/core/schema_loader.py
erindru: I'll add another test to verify the behaviour.

The intention was that the `gateway: `-specific entry would override the entry with no gateway specified when the file is loaded. So both can exist in the file without conflict, and one supersedes the other on load.

Does that seem like reasonable behaviour?

sqlmesh/core/context.py
erindru: Possibly. I'm not sure what version it's meant to be using, but here is the version it is using:

```
$ ruff --version
ruff 0.4.7
```

izeigerman: I have the same one

sqlmesh/core/schema_loader.py
izeigerman: Yeah, I'm more worried, that when users with existing files run the command to refresh their schemas, that'd see a huge delta cause we'd just insert all the entires all over again while also keeping the old ones.

sqlmesh/core/context.py
erindru: I think I understand what happened. My IDE originally had the `black` formatter enabled and I must have saved this file before I turned it off.

`ruff` apparently didnt mind the changes `black` made so it didn't change them back. 

I've rebased this file from main and the formatting changes have gone

sqlmesh/core/model/definition.py
```diff
@@ -1805,7 +1806,7 @@ def create_external_model(
     path: Path = Path(),
     defaults: t.Optional[t.Dict[str, t.Any]] = None,
     **kwargs: t.Any,
-) -> Model:
+) -> ExternalModel:
```
erindru, line 1809: I wanted to collect the output of `create_external_model` into a typed list. The type needed to be `ExternalModel` and not `Model` so I could access the `gateway` property without the type checker complaining.

So I changed the return type of `create_external_model()` to be `ExternalModel` but it did mean I needed to cast the result of `_create_model`

sqlmesh/core/schema_loader.py
erindru: So it depends on whether users are explicitly setting `--gateway` or not. I observed that unless `--gateway` is explicitly set, it defaults to `None`.

There is no top-level fallback that sets it to the default gateway and then passes that to the loading / saving code. It seems that any code dealing with `gateway` must decide how to handle `None` vs a specific value.

So, a user wont see any duplicate entries if they usually don't set `--gateway` at all, because it will default to `None` and all the entries will be written without a gateway (which is the current behaviour).

If the user _does_ set `--gateway`, then yes the entries will be duplicated. However, this wont pose a problem on load because any gateway-specific entries will override entries with no gateway defined.

It will pose a problem on a `git diff` however. Since the user is using SQLMesh to manage their external models file anyway, I think it's reasonable to ask them to delete and re-create it. This should be minimal effort and will result in one set of external model definitions going forward.

I'm open to other suggestions though! I also thought about writing the gateway-specific models to their own file in the `external_models/` directory so that we wouldn't have to mutate an existing file. However, there is a potential for the file name to clash with a user-defined file

izeigerman: I think as long as the original behavior is preserved if the user doesn't explicitly provide a gateway, I'm fine with that.

sqlmesh/core/context.py
izeigerman: Thanks!

sqlmesh/core/loader.py
```diff
@@ -208,7 +210,17 @@ def _load_external_models(self) -> UniqueKeyDict[str, Model]:
                             project=config.project,
                             default_catalog=self._context.default_catalog,
                         )
+                        external_models.append(model)
+
+                    # external models with no explicit gateway defined form the base set
+                    for model in (e for e in external_models if e.gateway is None):
```
izeigerman, line 216: I wonder whether this can be simplified to one pass through:
```
for model in external_models:
    if model.fqn not in models or model.gateway == gateway:
        models[model.fqn] = model
```
Does this achieve the same thing?

sqlmesh/core/schema_loader.py
```diff
@@ -62,10 +64,13 @@ def _get_columns(table: str) -> t.Optional[t.Dict[str, t.Any]]:
                 logger.warning(f"Unable to get schema for '{table}': '{e}'.")
                 return None
 
+        gateway_part = {"gateway": gateway} if gateway else {}
```
izeigerman, line 67: Provided that `gateway` is optional in the arguments, should we just always set it and avoid dict manipulations?

erindru: Do you mean like write `gateway: None` to the YAML?

The goal of this was to omit the key entirely if the gateway wasn't specified

sqlmesh/core/loader.py
erindru: Almost. That achieves:

```ValueError: Duplicate key found in UniqueKeyDict<models>. Call dict.update(...) if this is intentional.```

However, if I change it to `models.update({...})` then it still breaks a couple of tests because it adds external models from other gateways (not just `None` and the current gateway).

If it change it to:
```
                    for model in external_models:
                        if model.gateway is None or model.gateway == gateway:
                            models.update({model.fqn: model})
```

Then there is the potential to get different results depending on which order the models are loaded. For example, a model with `gateway=None` could override a model with the same name and `gateway=<current gateway>` if it's loaded second.

The two-pass method avoids this by populating the base set (models with `gateway=None`) and then overriding it just with models from the current gateway.

I get that it's less efficient to iterate the list twice. If I tried to sort the list (to put all the `gateway=None` entries at the top) and iterate it once, I guess that's still iterating it twice because the sort will need to iterate it

Comment (izeigerman) [reply sqlmesh/core/loader.py:216]: Oh, if you're getting:
> ValueError: Duplicate key found in UniqueKeyDict<models>. Call dict.update(...) if this is intentional.

Doesn't this mean that your previous implementation doesn't work either, since no override actually takes place?

Comment (izeigerman) [reply sqlmesh/core/loader.py:216]:
```
for model in external_models:
    if model.fqn not in models or (model.gateway and model.gateway == gateway):
        models.update({model.fqn: model})
```
Shouldn't this work then?

Comment (izeigerman) [reply sqlmesh/core/loader.py:216]: Anyway, this is a minor thing.

**Review by izeigerman — APPROVED** [review]

File: sqlmesh/core/loader.py (hunk 6 above)
Comment (erindru) [reply sqlmesh/core/loader.py:216]: > Doesn't this mean that your previous implementation doesn't work either, since no override actually takes place?

The overrides take place on the second pass, where `update` is called instead of using the `foo[bar] = "baz"` syntax. This ensures that duplicate models without a gateway defined will still cause an error (current behaviour)

> Shouldn't this work then?

It doesnt, because the first check `if model.fqn not in models` can return `True` for models specific to an unrelated gateway. 

Let's say I had a model `@{gateway}_foo` and had two versions of it defined in the external models file:

```
- name: dev_foo
  gateway: dev
- name: prod_foo
  gateway: prod
```

Using your code, loading this file using `--gateway dev` will put `prod_foo` in the list of external models, because `model.fqn not in models` returns `True`

Does that make sense? You can try it as well, some tests that check for this specifically will fail

PR #2749, opened 2024-06-06
WIP: Add Materialize Engine Adapter

bobbyiliev:
As a follow up to the https://github.com/tobymao/sqlglot/pull/3577 PR, starting the initial work to add the SQLMesh Materialize engine adapter.

This is still WIP, just starting with the initial work.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -203,6 +203,79 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
         return init
 
 
+class MaterializeConnectionConfig(ConnectionConfig):
+    """
+    Materialize Connection Configuration.
+
+    Args:
+        host: The hostname of the Materialize instance.
+        user: The username to use for authentication.
+        password: The password to use for authentication.
+        database: The name of the database to connect to.
+        port: The port number of the Materialize instance. Default is 6875.
+        sslmode: The SSL mode to use for the connection. Default is 'disable'.
+        connect_timeout: The number of seconds before the connection to the server will timeout.
+        application_name: The name of the application connecting to Materialize.
+        cluster: The name of the cluster to connect to.
+    """
+
+    host: str
+    user: str
+    password: str
+    database: str
+    port: int = 6875
+    sslmode: str = "disable"
+    connect_timeout: int = 10
+    application_name: t.Optional[str] = None
+    cluster: t.Optional[str] = None
+
+    concurrent_tasks: int = 4
+    register_comments: bool = True
+    pre_ping: bool = True
+
+    type_: Literal["materialize"] = Field(alias="type", default="materialize")
+
+    @property
+    def _connection_kwargs_keys(self) -> t.Set[str]:
+        return {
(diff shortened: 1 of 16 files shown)


eakmanrq:
Excited about the progress! When you are ready, make sure to add materialize here: https://github.com/TobikoData/sqlmesh/blob/0607f65800d231d70fdc2c50c0550c77b39eb873/tests/core/engine_adapter/test_integration.py#L538

These tests ensure that the engine implements the engine adapter interface correctly which tends to pick up most of the issue. It also includes a sushi test now which involves actually running a SQLMesh project. Typically passing these tests is where most of your time will be spent when adding a new engine.

sqlmesh/core/config/connection.py
```diff
@@ -203,6 +203,79 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
+
+    @property
+    def _connection_factory(self) -> t.Callable:
+        from psycopg2 import connect
+
+        return connect
+
+    @property
+    def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
+        kwargs = {
```
izeigerman, line 264: I don't think you need to enumerate all of these since you've already specified these keys in _connection_kwargs_keys. So they will be propagated as is from the config object. I think you should only keep options here since it requires additional processing.

tests/core/engine_adapter/test_integration.py
```diff
@@ -536,6 +536,15 @@ def config() -> Config:
...
         ),
+        pytest.param(
+            "materialize",
+            marks=[
+                pytest.mark.docker,
+                pytest.mark.engine,
+                pytest.mark.materialize,
+                pytest.mark.xdist_group("engine_integration_materialize"),
+            ],
+        ),
```
bobbyiliev, line 547: These tests are picking up 2 main issues, which I will be looking into!

tobymao:
hey @bobbyiliev how's progress on this? do you need any assistance?

bobbyiliev:
> hey @bobbyiliev how's progress on this? do you need any assistance?

Hey @tobymao, sorry for the delay here. I will be picking this up again starting tomorrow.

The two main things that are currently left are the fact that Materialize does not support `ALTER TABLE` and `CREATE TABLE ..  AS SELECT` which I will be looking into tomorrow.

bobbyiliev:
Hey @tobymao, quick update here, I had a chat with the team regarding the limitation where Materialize does not support `ALTER TABLE .. ADD COLUMN ..` and we will actually be working on this soon and hope to have this implemented by the beginning of the next quarter. I was thinking that in this case, it might be better to wait for that to get implemented, as SQMesh migrations seem to depend on it quite a bit, and as far as I can tell all other drivers that you currently have, all support it. Let me know what you think!

tobymao:
@bobbyiliev that's fine. thanks for digging into it. looking forward to adding materialize when it's ready :)

tobymao:
i'm going to close this now but feel free to reopen when you have @bobbyiliev

Issue #2505, opened 2024-04-24
Feature request: support user-defined macros written as SQL


plaflamme:
Currently, user-defined macros need to be defined as python. An alternative SQL-based syntax would be more consistent with `MODEL` and `AUDIT`. That is, having a `MACRO` function to define user-defines macros in `.sql` files:

```sql
MACRO(
  name forall,
  defaults (
    condition NULL
  )
);
SELECT *
FROM @this_model
WHERE
  @AND(
    @REDUCE(@criteria, (l,r) -> l AND r),
    @condition,
  )
```

For supporting named arguments, perhaps a more general `arguments` could work, i.e.: you'd have to define each argument and its default value (and maybe its type for coercion)?:

```sql
MACRO(
  name accepted_range,
  arguments (
    min_v NULL,
    max_v NULL,
    inclusive TRUE,
    condition NULL,
  )
);
...
```

tobymao:
i believe you can do this already

 https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#local-variables

tobymao:
closing for now but happy to discuss alternatives

treysp:
We do currently support reusable SQL macros written with jinja

https://sqlmesh.readthedocs.io/en/stable/concepts/macros/jinja_macros/#user-defined-macro-functions

plaflamme:
Which, presumably, is an argument for having an equivalent in SQL directly. As a SQLMesh user that doesn't have to deal with `dbt` or `jinja`, it would be a more consistent experience to be able to write these as `SQL`. Writing them in python is okay, but less accessible to non-developer types of users (e.g.: "data analysts" or similar SQL-mostly roles).

StuffbyYuki:
@plaflamme Is this still an issue?

plaflamme:
@StuffbyYuki this is a feature request, looks like there were 2 attempts at implementing it

PR #2760, opened 2024-06-10
WIP: Add support for reusable SQL macros

Themiscodes:
This is WIP for supporting reusable SQL based macros: [#2505](https://github.com/TobikoData/sqlmesh/issues/2505)

This update would be an alternative to Jinja with user defined SQL macros in the `.sql` files under the `macros` directory. The provisional syntax is as follows:

```
MACRO_BEGIN pythag(a, b);
    SQRT(POW(@a, 2) + POW(@b, 2));
MACRO_END;
```

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -661,7 +661,7 @@ def _is_command_statement(command: str, tokens: t.List[Token], pos: int) -> bool
         return False
 
 
-MACRO = "MACRO"
+MACRO_BEGIN = "MACRO_BEGIN"
 MACRO_END = "MACRO_END"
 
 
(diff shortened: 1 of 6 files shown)


sqlmesh/core/dialect.py
```diff
@@ -658,11 +661,23 @@ def _is_command_statement(command: str, tokens: t.List[Token], pos: int) -> bool
         return False
 
 
+MACRO_BEGIN = "MACRO_BEGIN"
```
izeigerman, line 664: I think this should just be `MACRO (...)`

Themiscodes: Yes and it was to avoid collisionswith SQL dialects that use MACRO statements

tobymao: i think it should be MACRO() since a macro can only have one statement.

what dialects have a macro keyword?

tobymao: macro's shouldn't require custom parsing, it can probably be something like

```@def(MY_MACRO, arg -> ...)```

Themiscodes: Duckdb has macro statements that use the macro keyword. Ok, I will try with a syntax like: `@def(MY_MACRO, arg -> ...)` so that it is like the sqlmesh macro definitions in the models sql files

PR #2768, opened 2024-06-13
Feat: make SchemaDiffer precision aware

treysp:
Some data types can include user-specified precision parameters, such as decimal precision `DECIMAL(10,1)` or character length `VARCHAR(100)`. 

Currently, the SchemaDiffer outputs a DROP/ADD for all  precision changes, even though precision increases can be done with an ALTER. This PR makes the SchemaDiffer aware of precision such that it outputs ALTER statements when precision is increased.

Precision is implemented differently across engines and types:
- Some types have one parameter `VARCHAR(1)`, and others have two `DECIMAL(10,1)`.
- Sometimes a parameter is optional and has a default value (e.g., `VARCHAR` = `VARCHAR(256)`, `DECIMAL(10)` = `DECIMAL(10,0)`). The SchemaDiffer will automatically provide these default values where possible.
- Some engines/types allow explicitly making a type its maximum precision (e.g., `VARCHAR(max)`).  SchemaDiffer substitutes the corresponding numeric precision.
- Sometimes a parameter is optional and the default is "unlimited" precision (e.g., `VARCHAR` allows strings of any length up to system limits). SchemaDiffer will allow ALTERs from any explicit precision to an unlimited type.
- 
Default values for all engines and types are documented in [this Notion doc](https://www.notion.so/SchemaDiffer-type-docs-4464135715254cc597d638e4550c20dd?pvs=4).

## Complexities

Type aliases:
- Some engines provide multiple aliases for the same underlying type. For example, DuckDB has a single "unlimited" length character type and maps all of `VARCHAR`, `VARCHAR(1)`, `CHAR`, and others to that underlying type.
- In some engines, the aliases behave differently depending on whether a parameter is provided. For example, in Postgres `BPCHAR(1)` == `CHAR(1)` but `BPCHAR` != `CHAR`. 
- Therefore, direct comparison of parameter values only occurs WITHIN a SQLGlot `exp.DataType.Type`. If two type aliases are parsed to the same underlying SQLGlot type their parameters will be compared, if not SchemaDiffer will return DROP/ADD.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -74,8 +74,8 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
         },
         support_coercing_compatible_types=True,
         parameterized_type_defaults={
-            exp.DataType.build("DECIMAL", dialect=DIALECT).this: {0: (38, 9), 1: (0,)},
-            exp.DataType.build("BIGDECIMAL", dialect=DIALECT).this: {0: (76.76, 38), 1: (0,)},
+            exp.DataType.build("DECIMAL", dialect=DIALECT).this: [(38, 9), (0,)],
+            exp.DataType.build("BIGDECIMAL", dialect=DIALECT).this: [(76.76, 38), (0,)],
         },
         types_with_unlimited_length={
             # parameterized `STRING(n)` can ALTER to unparameterized `STRING`
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -73,6 +73,20 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
             },
         },
         support_coercing_compatible_types=True,
+        parameterized_type_defaults={
+            exp.DataType.build("DECIMAL", dialect=DIALECT).this: {0: (38, 9), 1: (0,)},
+            exp.DataType.build("BIGDECIMAL", dialect=DIALECT).this: {0: (76.76, 38), 1: (0,)},
+        },
+        types_with_unlimited_length={
+            # parameterized `STRING(n)` can ALTER to unparameterized `STRING`
+            exp.DataType.build("STRING", dialect=DIALECT).this: {
+                exp.DataType.build("STRING", dialect=DIALECT).this,
+            },
+            # parameterized `BYTES(n)` can ALTER to unparameterized `BYTES`
+            exp.DataType.build("BYTES", dialect=DIALECT).this: {
+                exp.DataType.build("BYTES", dialect=DIALECT).this,
+            },
+        },
     )
 
     @property
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -26,6 +26,7 @@ from sqlmesh.core.engine_adapter.shared import (
     SourceQuery,
     set_catalog,
 )
+from sqlmesh.core.schema_diff import SchemaDiffer
 
 if t.TYPE_CHECKING:
     from sqlmesh.core._typing import SchemaName, TableName
@@ -48,6 +49,25 @@ class MSSQLEngineAdapter(
     COMMENT_CREATION_TABLE = CommentCreationTable.UNSUPPORTED
     COMMENT_CREATION_VIEW = CommentCreationView.UNSUPPORTED
     SUPPORTS_REPLACE_TABLE = False
+    SCHEMA_DIFFER = SchemaDiffer(
+        parameterized_type_defaults={
+            exp.DataType.build("DECIMAL", dialect=DIALECT).this: {0: (18, 0), 1: (0,)},
+            exp.DataType.build("BINARY", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("VARBINARY", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("CHAR", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("VARCHAR", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("NCHAR", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("NVARCHAR", dialect=DIALECT).this: {0: (1,)},
+            exp.DataType.build("TIME", dialect=DIALECT).this: {0: (7,)},
+            exp.DataType.build("DATETIME2", dialect=DIALECT).this: {0: (7,)},
+            exp.DataType.build("DATETIMEOFFSET", dialect=DIALECT).this: {0: (7,)},
+        },
+        types_with_max_parameter={
+            exp.DataType.build("VARBINARY", dialect=DIALECT).this,
+            exp.DataType.build("VARCHAR", dialect=DIALECT).this,
+            exp.DataType.build("NVARCHAR", dialect=DIALECT).this,
+        },
+    )
 
     def columns(
         self,
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -62,10 +62,10 @@ class MSSQLEngineAdapter(
             exp.DataType.build("DATETIME2", dialect=DIALECT).this: {0: (7,)},
             exp.DataType.build("DATETIMEOFFSET", dialect=DIALECT).this: {0: (7,)},
         },
-        types_with_max_parameter={
-            exp.DataType.build("VARBINARY", dialect=DIALECT).this,
-            exp.DataType.build("VARCHAR", dialect=DIALECT).this,
-            exp.DataType.build("NVARCHAR", dialect=DIALECT).this,
+        max_parameter_length={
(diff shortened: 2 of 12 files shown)


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -73,6 +73,20 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
             },
         },
         support_coercing_compatible_types=True,
+        parameterized_type_defaults={
+            exp.DataType.build("DECIMAL", dialect=DIALECT).this: [(38, 9), (0,)],
```
tobymao, line 77: you don't need to pass in dialect if you're just looking for a data type. just pass in the enum directly so you don't have to call this as well

tobymao: it kind of feels like these constants maybe should go into sqlglot, but perhaps we keep it for now

sqlmesh/core/schema_diff.py
```diff
@@ -294,15 +293,41 @@ class SchemaDiffer(PydanticModel):
...
+        support_nested_operations: Whether the engine for which the diff is being computed supports modifications to
+            nested data types like STRUCTs and ARRAYs.
+        compatible_types: Types that are compatible and automatically coerced in actions like UNION ALL. Dict key is data
+            type, and value is the set of types that are compatible with it.
+        support_coercing_compatible_types: Whether or not the engine for which the diff is being computed supports direct
+            coercion of compatible types.
+        parameterized_type_defaults: Default values for parameterized data types. Dict key is a sqlglot exp.DataType.Type,
+            but in the engine adapter specification we build it from the dialect string instead of specifying it directly.
+            Example: `exp.DataType.build("STRING", dialect=DIALECT).this` instead of the underlying `exp.DataType.Type.TEXT`
+            to which it parses. We do that because parameter default replacement will silently break if we specify type
```
tobymao, line 310: i'm not sure that's a good reason for it, more indication that this should go into sqlglot

sqlmesh/core/schema_diff.py
```diff
@@ -319,10 +344,16 @@ def coerceable_types(self) -> t.Dict[exp.DataType, t.Set[exp.DataType]]:
...
-        if current_type == new_type:
+        # types are identical or both types are parameterized and new has higher precision
+        # - default parameter values are automatically provided if not present
+        if current_type == new_type or self._is_precision_increase(current_type, new_type):
             return True
+        # types are un-parameterized and compatible
         if current_type in self.compatible_types:
             return new_type in self.compatible_types[current_type]
+        # new type is un-parameterized and has unlimited length, current type is compatible
+        if len(new_type.expressions) == 0 and new_type.this in self.types_with_unlimited_length:
```
tobymao, line 355: not new_type.expressions

sqlmesh/core/schema_diff.py
```diff
@@ -337,6 +368,47 @@ def _is_coerceable_type(self, current_type: exp.DataType, new_type: exp.DataType
...
+            current_params = self._get_type_parameters(current_type)
+            new_params = self._get_type_parameters(new_type)
+
+            if len(current_params) != len(new_params):
+                return False
+
+            return all(new >= current for current, new in zip(current_params, new_params))
+        return False
+
+    def _is_nested_type(self, type: exp.DataType) -> bool:
```
tobymao, line 382: DataType.NESTED_TYPES has this, with more stuff

sqlmesh/core/schema_diff.py
```diff
@@ -337,6 +368,47 @@ def _is_coerceable_type(self, current_type: exp.DataType, new_type: exp.DataType
             return is_coerceable
         return False
 
+    def _is_precision_increase(self, current_type: exp.DataType, new_type: exp.DataType) -> bool:
+        if current_type.this == new_type.this and not self._is_nested_type(current_type):
```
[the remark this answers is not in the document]
treysp: It looks like `is_type` includes parameters in the comparison if they're present in both, but not if only present in one. Is that intended behavior?

```
prec120 = exp.DataType.build("VARCHAR(120)")
prec121 = exp.DataType.build("VARCHAR(121)")
precNone = exp.DataType.build("VARCHAR")

prec120.is_type(prec121)
False
prec120.is_type(precNone)
True
```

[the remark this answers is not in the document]
georgesittas: That's the intended behavior, but perhaps we could add a flag to force a recursive (structural) equality check. I'd also be fine with just keeping `==`.

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -48,6 +49,25 @@ class MSSQLEngineAdapter(
...
     SUPPORTS_REPLACE_TABLE = False
+    SCHEMA_DIFFER = SchemaDiffer(
+        parameterized_type_defaults={
+            exp.DataType.build("DECIMAL", dialect=DIALECT).this: [(18, 0), (0,)],
+            exp.DataType.build("BINARY", dialect=DIALECT).this: [(1,)],
+            exp.DataType.build("VARBINARY", dialect=DIALECT).this: [(1,)],
+            exp.DataType.build("CHAR", dialect=DIALECT).this: [(1,)],
+            exp.DataType.build("VARCHAR", dialect=DIALECT).this: [(1,)],
+            exp.DataType.build("NCHAR", dialect=DIALECT).this: [(1,)],
+            exp.DataType.build("NVARCHAR", dialect=DIALECT).this: [(1,)],
```
georgesittas, line 60: Are we using 1 intentionally here? It seems like the default for casts is 30:

> When n isn't specified in a data definition or variable declaration statement, the default length is 1. When n isn't specified with the CAST function, the default length is 30.

sqlmesh/core/schema_diff.py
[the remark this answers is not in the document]
treysp: To summarize:
`type == new_type` -> true if equality of both underlying type and parameters
`type.this == new_type.this` -> true if equality of underlying type regardless of parameter presence in either
`type.is_type(new_type)` -> true if (same as first equality above) OR (type has parameters, new_type does not, same underlying type)

sqlmesh/core/engine_adapter/mssql.py
treysp: The schema differ ingests two schemas and determines the sequence of ALTERs/DROPs/ADDs needed to convert the types/structure of the former to the latter. We want to ALTER if a precision is increased.

If you tell me a schema contains `VARCHAR` and I need to determine whether we can ALTER to `VARCHAR(29)`, we would only want to say yes if we're certain that's possible without truncating data. 

So I think we want to use 1 as the conservative default since we can't know for sure?

georgesittas: Makes sense

sqlmesh/core/schema_diff.py
```diff
@@ -337,6 +368,46 @@ def _is_coerceable_type(self, current_type: exp.DataType, new_type: exp.DataType
...
+            _str_to_number(param.this.this, type.this in self.max_parameter_length)
+            for param in type.expressions
+        ]
+
+        # maybe get default parameter values
+        param_defaults: t.Tuple[t.Union[int, float], ...] = ()
+        if type.this in self.parameterized_type_defaults:
+            param_defaults_list = self.parameterized_type_defaults[type.this]
+            if len(params) < len(param_defaults_list):
+                param_defaults = param_defaults_list[len(params)]
```
tobymao, line 407: should this be param_defaults_list[len(params):], a splice?

how does this work?

tobymao (approved):

sqlmesh/core/schema_diff.py
treysp: The index returns a tuple with all the parameters missing from a type, given the number of parameters it already has: 

```
# Bigquery's decimal parameters 
# Note that second number in first tuple 9 is different from the second tuple 0 even though they both correspond to the decimal's scale parameter
bq_decimal_params = [(38, 9), (0,)]

# "DECIMAL" has 0 params so indexes to (38,9) -> DECIMAL(38,9)
# "DECIMAL(1)" has 1 param so indexes to (0) -> DECIMAL(1,0)
# "DECIMAL(1,0)" has 2 params, so len(2) !< len(bq_decimal_params) is a no-op
```

The splice approach is cleaner, but it would only work if the second number was constant with respect to the number of parameters the provided type already has.

This approach feels opaque and inelegant to me - any ideas for a cleaner implementation?

sqlmesh/core/schema_diff.py
```diff
@@ -337,6 +368,46 @@ def _is_coerceable_type(self, current_type: exp.DataType, new_type: exp.DataType
...
+    def _get_type_parameters(self, type: exp.DataType) -> t.List[t.Union[int, float]]:
+        def _str_to_number(string: str, allows_max_param: bool) -> t.Union[int, float]:
+            try:
+                return int(string)
+            except ValueError:
+                try:
+                    return float(string)
+                except ValueError:
+                    if allows_max_param and string.upper() == "MAX":
+                        return self.max_parameter_length[type.this]
```
izeigerman, line 393: Should we ever expect the value to be missing from this dict?

izeigerman: Never mind, I see the check on line 398

izeigerman (approved):

PR #2770, opened 2024-06-13
Fix: redshift leverages postgres adapter dispatch implementations for common functionality

crericha:
The redshift dbt adapter depends on the postgres adapter. For adapter dispatch, this means that postgres dispatch macros are valid when a corresponding redshift macro does not exist.

Note: The databricks adapter also depends on the spark adapter, so this fix applies there too.

### DIFF
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -97,32 +97,27 @@ class BaseAdapter(abc.ABC):
 
     def dispatch(self, name: str, package: t.Optional[str] = None) -> t.Callable:
         """Returns a dialect-specific version of a macro with the given name."""
+        package = package or "dbt"
         target_type = self.jinja_globals["target"]["type"]
         macro_suffix = f"__{name}"
 
-        def _macro_relevance(val: t.Tuple[str, str]) -> int:
+        def _relevance(macro: MacroReference) -> int:
             """Lower scores more relevant."""
-            package_name, macro_name = val
-            score = 0
-            if package_name != package:
-                score += 2
-            if macro_name.startswith("default"):
-                score += 3
-            elif not macro_name.startswith(target_type):
-                score += 1
-            return score
+            if macro.name.startswith("default"):
+                return 2
+            elif macro.name.startswith(target_type):
+                return 0
+            return 1
 
-        relevant_macros = [
-            (package_name, macro_name)
+        macros = [
+            MacroReference(package=package_name, name=macro_name)
             for package_name, macros in self.jinja_macros.packages.items()
+            if (package_name == package) or (package == "dbt" and package_name.startswith("dbt_"))
             for macro_name in macros.keys()
             if macro_name.endswith(macro_suffix)
         ]
-
-        for package_name, macro_name in sorted(relevant_macros, key=_macro_relevance):
-            macro_callable = self.jinja_macros.build_macro(
-                MacroReference(package=package_name, name=macro_name), **self.jinja_globals
-            )
+        for macro in sorted(macros, key=_relevance):
+            macro_callable = self.jinja_macros.build_macro(macro, **self.jinja_globals)
             if macro_callable is not None:
                 return macro_callable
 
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/adapter.py
```diff
@@ -98,14 +98,31 @@ def quote(self, identifier: str) -> str:
...
+                score += 2
+            if macro_name.startswith("default"):
+                score += 3
+            elif not macro_name.startswith(target_type):
+                score += 1
+            return score
+
+        relevant_macros = [
+            (package_name, macro_name)
+            for package_name, macros in self.jinja_macros.packages.items()
```
izeigerman, line 117: Shouldn't we also check `root_macros` if the `package` is None?

sqlmesh/dbt/adapter.py
```diff
@@ -98,14 +98,31 @@ def quote(self, identifier: str) -> str:
...
-            MacroReference(package=package, name=f"{target_type}__{name}"),
-            MacroReference(package=package, name=f"default__{name}"),
+        macro_suffix = f"__{name}"
+
+        def _macro_relevance(val: t.Tuple[str, str]) -> int:
+            """Lower scores more relevant."""
+            package_name, macro_name = val
+            score = 0
+            if package_name != package:
+                score += 2
```
izeigerman, line 108: I think this mechanics is overly complex. Can we just stick to 3 states: 0-2 with 2 being `default__`, 1 for the same package name and `__{name}`, and 0 for `{package}__` + `__{name}`

izeigerman: Sorry, I meant in reversed order.

tests/dbt/test_adapter.py
```diff
@@ -148,9 +148,10 @@ def test_adapter_dispatch(sushi_test_project: Project, runtime_renderer: t.Calla
     context = sushi_test_project.context
     renderer = runtime_renderer(context)
     assert renderer("{{ adapter.dispatch('current_engine', 'customers')() }}") == "duckdb"
+    assert renderer("{{ adapter.dispatch('current_engine')() }}") == "duckdb"
 
-    with pytest.raises(ConfigError, match=r"Macro 'current_engine'.*was not found."):
```
izeigerman, line 152: This doesn't look right to me. Why did this test change? The dispatch call should've failed since  the `current_engine` macro is in the `customers` package.

sqlmesh/dbt/adapter.py
crericha: The above suggestion is missing different package + expected name. For example, `dbt_redshift.redshift__{name}` would be the correct macro and should be preferred over `dbt.postgres__{name}`.

crericha: I ranked the best match as 0, so I didn't have to add reverse to the sort. Do you have a strong opinion here?

tests/dbt/test_adapter.py
crericha: `current_engine` is a dispatch, so it resolves if in another package. We can narrow what the dispatch finds if we don't want this. For example, maybe if the requested package is `dbt.*`, we look in all `dbt.*` packages. Otherwise, we only look in the requested package?

sqlmesh/dbt/adapter.py
izeigerman: > I ranked the best match as 0, so I didn't have to add reverse to the sort. Do you have a strong opinion here?

No, I don't.

> For example, dbt_redshift.redshift__{name} would be the correct macro and should be preferred over dbt.postgres__{name}.

Doesn't this:

> and 0 for `{package}__` + `__{name}`

cover it?

tests/dbt/test_adapter.py
izeigerman: I think we absolutely need to take the package into account. I wouldn't hardcode the `dbt` package though.

sqlmesh/dbt/adapter.py
crericha: According to dbt documentation, use the dbt internal packages if no package specified.

izeigerman (approved):

PR #2776, opened 2024-06-14
Feat: Support 'enabled' flag in model definitions

izeigerman:
Now that we support project-wide variables in model meta, it also makes sense to support the `enabled` flag to allow users to conditionally enable / disable certain models.


tobymao (approved):

serkef:
I love this feature! I don't have to rename or move my wip models any more!

z3z1ma:
Same, I used to remove the `.sql` extension to similar effect lol

Now I am going to use a macro to populate the value dynamically

PR #2777, opened 2024-06-14
Proposal to Add 6 Hour Interval

isaiahgm:
We have many models that we want to update more than once a day, but having a cron schedule like "0 */6 * * *" will currently cause SQL Mesh to create intervals on an hourly increment. This can cause high water mark sensor to not succeed even if the upstream table was updated properly if the cron schedules are in different hours. For example, if upstream is "55 */6 * * *" and downstream is "0 1/6 * * *", downstream will not run because it will look for an interval in a different hour, even though the two runs were 5 minutes apart.

I might be misunderstanding how scheduling should work, so I am open to any solution. However, I wanted to start with good faith attempt to resolve the issue as un-intrusively as possible.


tobymao:
i'm not 100% sure this is the right approach perhaps this is something specific to HWM sensors. @izeigerman do you have thoughts here?

izeigerman:
Hey @isaiahgm , thanks a lot for the submission.

Let me make sure I understand the problem correctly.

Let's imagine that the upstream model A runs at `6:55` according to it's configured cron. Since the interval unit is hourly, it will process data up to and including hour 5, but it won't process hour 6.

Now the downstream model B runs at `7:00`. Since the interval unit is again hourly, it actually wants to process data up to and including hour 6, which is 1 hour ahead of what's been backfilled by model A. 

Since hour 6 was not populated by the most recent run of the model A, the model B is forced to wait for the next run of model A.

Am I understanding this correctly? I'm just curious why is model A scheduled to run before the model B?

isaiahgm:
Hi @izeigerman, yes the captures the essence of the issue well. I think the intent is that Model A and Model B can be scheduled to run at arbitrary times. With many people creating their own models with a complex web of dependencies, it is common to see job scheduled on different cadences based on what is most convenient for them.

As an example, we could have Model A created by Team 1 that they want to update every 6 hours, so they assign the schedule `30 */6 * * *`. Team 2, thinks the data in Model A would be useful to them, so they build Model B which uses Model A plus some additional logic. Team 2 has a strict business requirement to send the data in Model B at a specific time once a day, so they schedule it at `0 8 * * *`. Unfortunately, b/c Model A will default to hour increments, Model B will not run on time (if at all).

This example could be extended to many situations, and it isn't limited to just 6 hour increments. Because cron is so flexible a tool, any number of schedules could be generated, and I'd argue that SQL Mesh should be able to support most cron schedules.

sqlmesh/core/node.py
```diff
@@ -71,7 +72,7 @@ def is_day(self) -> bool:
 
     @property
     def is_hour(self) -> bool:
-        return self == IntervalUnit.HOUR
+        return self in (IntervalUnit.HOUR, IntervalUnit.SIX_HOUR)
```
izeigerman, line 75: I don't think this is right. I think we need a separate `is_six_hour` property method here, but since we don't have a use case for it we can add it later if needed. In any case `is_hour` shouldn't be updated

izeigerman:
@isaiahgm , thanks for the context.

One danger I see with supporting interval units like this one is that if a cron expression represents a more irregular pattern, such as every 7, 8, or 9 hours, all of those will be identified as 6 hours long as opposed to 1 hour long, which is IMHO feels more appropriate.

What do you think?

isaiahgm:
My change is only a half-measure in solving the underlying issue. There will be problems with all potential intervals outside of the handful supported by SQL Mesh. It may warrant exploring an entirely different approach to determining intervals. I admittedly don't fully understand all of this project's code, so I don't know if it is possible, but I think it would be more flexible to build intervals by interpreting the give cron schedule directly rather than trying to bucket the schedules. 

This would obviously be a significant undertaking, and would surely have an affect on many aspects of the project, which is why I proposed this half-measure as a stop gap.

izeigerman:
I'm somewhat concerned that the proposed solution only works in a very narrow set of circumstances.

I see how this solves this very specific problem for Model A cron `55 */6 * * *` and Model B cron `0 1/6 * * *`, but if you change this initial setup slightly, you will run into the same issue all over again.

For example, if the Model B cron was set to something like `0 */10 * * *`, it would result in Model B being blocked for at least 4 hours.

My recommendation would be to solve this at the cron level for now until we have a more robust solution.

Closing this PR for now but happy to continue the discussion here or in Slack.

isaiahgm:
Thank you for the review! If anything I hope this sparks development of this system so that SQL Mesh can handle a variety of cron setups. We have a serviceable workaround right now by just requiring jobs to execute in the same hour, but that may not always work, especially as we have more and more users adopting the tool.

PR #2779, opened 2024-06-17
Feat: allow overriding the dialects' normalization strategies

georgesittas:


docs/guides/configuration.md
```diff
@@ -893,6 +893,27 @@ Example configuration:
 
 The default model kind is `VIEW` unless overridden with the `kind` key. For more information on model kinds, refer to [model concepts page](../concepts/models/model_kinds.md).
 
+##### Identifier resolution
```
georgesittas, line 896: Not sure if this is overkill, but since there has been some discussion in Slack, I thought I'd do at least a first pass and then we could chat about it.

cc @tobymao @treysp

docs/guides/configuration.md
```diff
@@ -893,6 +893,27 @@ Example configuration:
...
+
+When a SQL engine receives a query such as `SELECT id FROM "some_table"`, it eventually needs to understand what database objects the identifiers `id` and `"some_table"` correspond to. This process is usually referred to as identifier (or name) resolution.
+
+Different SQL dialects implement different rules when resolving identifiers in queries. For example, certain identifiers may be treated as case-sensitive (e.g. if they're quoted), and a case-insensitive identifier is usually either lowercased or uppercased, before the engine actually looks up what object it corresponds to.
+
+SQLMesh analyzes model queries so that it can extract useful information from them, such as computing Column-Level Lineage. To facilitate this analysis, it _normalizes_ and _quotes_ all identifiers in those queries, [respecting each dialect's resolution rules](https://sqlglot.com/sqlglot/dialects/dialect.html#Dialect.normalize_identifier).
+
+The "normalization strategy", i.e. whether case-insensitive identifiers are lowercased or uppercased, is configurable per dialect. For example, to treat all identifiers as caes-sensitive in a BigQuery project, one can do:
```
georgesittas, line 904:
```suggestion
The "normalization strategy", i.e. whether case-insensitive identifiers are lowercased or uppercased, is configurable per dialect. For example, to treat all identifiers as case-sensitive in a BigQuery project, one can do:
```

sqlmesh/core/context.py
```diff
@@ -325,6 +325,13 @@ def __init__(
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
 
+        # This allows overriding the default dialect's normalization strategy, so for example
+        # one can do `dialect="duckdb,normalization_strategy=lowercase"` and this will be
+        # applied to the DuckDB globally
```
georgesittas, line 330:
```suggestion
        # applied to the DuckDB dialect globally
```

sqlmesh/core/context.py
```diff
@@ -325,6 +325,13 @@ def __init__(
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
 
+        # This allows overriding the default dialect's normalization strategy, so for example
+        # one can do `dialect="duckdb,normalization_strategy=lowercase"` and this will be
+        # applied to the DuckDB dialect globally
+        if "normalization_strategy" in str(self.config.dialect):
+            dialect = Dialect.get_or_raise(self.config.dialect)
+            type(dialect).NORMALIZATION_STRATEGY = dialect.normalization_strategy
```
izeigerman, line 333: So I can't have different normalization strategies for the same dialect? Furthermore it can't be overridden on per model basis? Should this be a top-level config then instead of being a part of a dialect? Feels a little misleading otherwise.

georgesittas: > So I can't have different normalization strategies for the same dialect?

That's right - but it seems reasonable to me, because you can only have a single default strategy for a given dialect. This simply makes the `default_dialect` override that default globally for said dialect, but it _can_ be overridden on a model basis, so if you do `dialect duckdb,normalization_strategy=...` in a model meta, then I'm pretty sure that should take precedence over the class constant set here.

georgesittas: > So I can't have different normalization strategies for the same dialect?

Ah, I guess you were asking if you can override that strategy per model? If so, my answer is a bit contradictory. You can override it wherever you can specify a dialect basically.

tobymao: sorry i'm confused, can you elaborate?

georgesittas: - You can set `dialect: foo,normalization_strategy=...` at the model defaults config level. This will monkey patch its `NORMALIZATION_STRATEGY` [class constant](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py#L216), so it applies globally.
- You can override this behavior per model by doing e.g. `dialect foo,normalization_strategy=...`

izeigerman: I'm still confused though. Wouldn't setting it in a model still override the strategy globally?

tobymao: how would it work if you set the model config to normarizaltion=x. would that get propogated to the engine adapter?

georgesittas: > Wouldn't setting it in a model still override the strategy globally?

Nope, because it gets stored as an attributed in the dialect and we ignore the `NORMALIZATION_STRATEGY` constant (see [constructor](https://github.com/tobymao/sqlglot/blob/main/sqlglot/dialects/dialect.py#L513-L519)). So there's no global overriding.

> how would it work if you set the model config to normarizaltion=x. would that get propogated to the engine adapter?

If you do this at the model default config, then the dialect will be patched => the engine adapter's dialect will also have its strategy changed. This was the main motivation behind the patching.

If we didn't patch it, then setting `normalization=..` in the project config would override the strategy of the dialect stored in the config, but not the one [stored in the engine adapter](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/base.py#L116), because we don't really pass the config's dialect to the adapters.

tobymao (approved):

Issue #1992, opened 2024-01-19
Better support for long lists of indirectly modified models printed in CLI output


izeigerman:
Sometimes a directly modified model may be associated with 100s indirectly modified downstream models. 

When this happens the output becomes cumbersome and overwhelming to a user. We need to revisit how we print out the directly / indirectly modified trees and potentially omit model names once their total number exceeds a certain threshold. 

For example one way to present a really long list could look like this:

```
Summary of differences against `prod`:
Models:
├── Directly Modified:
│   └── sushi.orders
└── Indirectly Modified:
    ├── sushi.waiter_revenue_by_day
    ├── ... N more ....
    └── sushi.top_waiters
  ```

PR #2788, opened 2024-06-18
Feat: Limit long lists of indirectly modified models in CLI output

Themiscodes:
This is an improvement to trim long indirectly modified trees in the output, fixes: #1992 

```
Summary of differences against `dev2`:
Models:
├── Directly Modified:
│   └── sushi__dev2.seed_model
└── Indirectly Modified:
    ├── sushi__dev2.full_model
    ├── ... 30 more ...
    └── sushi__dev2.incremental_model
```

With this update, the names of indirectly modified models will be omitted once their total number exceeds a certain threshold, currently set to 20. This behavior can be overridden with the `verbose` flag, in which case all model names will be shown regardless of their total number.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -286,6 +286,8 @@ class TerminalConsole(Console):
 
         self.loading_status: t.Dict[uuid.UUID, Status] = {}
 
+        self.modified_threshold: int = 20
+
         self.verbose = verbose
         self.dialect = dialect
 
@@ -628,6 +630,16 @@ class TerminalConsole(Console):
         if auto_apply:
             plan_builder.apply()
 
+    def _limit_model_names(self, tree: Tree) -> Tree:
+        modified_length = len(tree.children)
+        if modified_length > self.modified_threshold:
+            tree.children = [
+                tree.children[0],
+                Tree(f".... {modified_length-2} more ...."),
+                tree.children[-1],
+            ]
+        return tree
+
     def _get_ignored_tree(
         self,
         ignored_snapshot_ids: t.Set[SnapshotId],
@@ -728,7 +740,7 @@ class TerminalConsole(Console):
             if direct.children:
                 tree.add(direct)
             if indirect.children:
-                tree.add(indirect)
+                tree.add(indirect if self.verbose else self._limit_model_names(indirect))
             if metadata.children:
                 tree.add(metadata)
         if selected_ignored_snapshot_ids:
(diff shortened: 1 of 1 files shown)


sqlmesh/core/console.py
```diff
@@ -286,6 +286,8 @@ def __init__(
 
         self.loading_status: t.Dict[uuid.UUID, Status] = {}
 
+        self.modified_threshold: int = 20
```
georgesittas, line 289: No need for the type hint here:

```suggestion
        self.modified_threshold = 20
```

Also:
1) should we make this configurable?
2) if not, we can make it a class attribute, i.e.:

```python
class TerminalConsole(Console):
    INDIRECTLY_MODIFIED_DISPLAY_THRESHOLD = 20
```

sqlmesh/core/console.py
```diff
@@ -728,7 +740,7 @@ def _show_summary_tree_for(
             if direct.children:
                 tree.add(direct)
             if indirect.children:
-                tree.add(indirect)
+                tree.add(indirect if self.verbose else self._limit_model_names(indirect))
```
georgesittas, line 743: should we just push down the verbose check in `_limit_model_name`? So it's only done in one place.

georgesittas:
What about these two?

- `MarkdownConsole.show_model_difference_summary`
- `MarkdownConsole._show_categorized_snapshots`

sqlmesh/core/console.py
Themiscodes: Ok, I'll make it a class attribute for now

sqlmesh/core/console.py
Themiscodes: Refactored the function to contain the check.

Themiscodes:
Thank you @georgesittas for the review, addressed your comments and extended the functionality for MarkdownConsole as well

sqlmesh/core/console.py
```diff
@@ -1550,10 +1569,19 @@ def show_model_difference_summary(
...
                 self._print("\n**Indirectly Modified:**")
-                for snapshot in sorted(indirectly_modified):
+                modified_length = len(indirectly_modified)
+                if (
+                    not self.verbose
+                    and modified_length > self.INDIRECTLY_MODIFIED_DISPLAY_THRESHOLD
+                ):
                     self._print(
-                        f"- `{snapshot.display_name(environment_naming_info, default_catalog, dialect=self.dialect)}`"
+                        f"- `{indirectly_modified[0].display_name(environment_naming_info, default_catalog, dialect=self.dialect)}`\n- `.... {modified_length-2} more ....`\n- `{indirectly_modified[-1].display_name(environment_naming_info, default_catalog, dialect=self.dialect)}`"
```
georgesittas, line 1578: Let's make this a bit easier to read. Another concern is that the order here won't necessarily match that in the `else` branch, because we don't sort here. I'd sort beforehand.

Themiscodes: Thank you, refactored the code to address these

georgesittas (approved):

PR #2791, opened 2024-06-18
Feat: increase default type precision in redshift/mssql view workaround

treysp:
Redshift and MSSQL use the `VarcharSizeWorkaroundMixin` to infer the size of a varchar or other "max" column when it cannot be resolved from model code.

Data may be truncated if it is inferred to the default length (256 in Redshift, 1 in MSSQL) and that was not intended by the user, so this PR increases the length to "max" if a column with the default length is detected.

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -244,7 +244,7 @@ class VarcharSizeWorkaroundMixin(EngineAdapter):
                 parameter = self.SCHEMA_DIFFER.get_type_parameters(col_type)
                 type_default = types_with_max_default_param[col_type.this]
                 if parameter == type_default:
-                    col_type.set("expressions", [exp.DataTypeParam(this=exp.Var(this="max"))])
+                    col_type.set("expressions", [exp.DataTypeParam(this=exp.var("max"))])
 
         return columns_to_types
 
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -238,7 +238,10 @@ class VarcharSizeWorkaroundMixin(EngineAdapter):
             if k in self.SCHEMA_DIFFER.parameterized_type_defaults
         }
 
-        # if type supports "max" and has default length, convert to "max" length
+        # Redshift and MSSQL have a bug where CTAS statements have non-deterministic types. If a LIMIT
+        # is applied to a CTAS statement, VARCHAR (and possibly other) types sometimes revert to their
+        # default length of 256 (Redshift) or 1 (MSSQL). If we detect that a type has its default length
+        # and supports "max" length, we convert it to "max" length to prevent inadvertent data truncation.
         for col_name, col_type in columns_to_types.items():
             if col_type.this in types_with_max_default_param and col_type.expressions:
                 parameter = self.SCHEMA_DIFFER.get_type_parameters(col_type)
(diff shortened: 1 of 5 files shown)


sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -227,6 +227,26 @@ def _truncate_table(self, table_name: TableName) -> None:
 
 
 class VarcharSizeWorkaroundMixin(EngineAdapter):
+    def _default_precision_to_max(
+        self, columns_to_types: t.Dict[str, exp.DataType]
+    ) -> t.Dict[str, exp.DataType]:
+        # get default lengths for types that support "max" length
+        types_with_max_default_param = {
+            k: [self.SCHEMA_DIFFER.parameterized_type_defaults[k][0][0]]
```
izeigerman, line 235: I'm not sure we care about all types. I think we only care about `Varchar` don't we?

izeigerman: Wouldn't this simplify the implementation?

izeigerman: Or do we think the same issue applies to types `VARBINARY`?

tests/core/engine_adapter/test_redshift.py
```diff
@@ -29,6 +29,32 @@ def test_columns(adapter: t.Callable):
     assert resp == {"col": exp.DataType.build("INT")}
 
 
+def test_varchar_size_workaround(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture):
```
izeigerman, line 32: Can we also have a test that validates this flow all the way to the table creation expression?

sqlmesh/core/engine_adapter/mixins.py
treysp: I figured it could happen for any of these types due to transpilation, but VARCHAR is much more widely used than NVARCHAR/CHAR/VARBINARY so in practice it's probably the only one that matters. I'll update.

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -227,6 +227,26 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+            if k in self.SCHEMA_DIFFER.parameterized_type_defaults
+        }
+
+        # if type supports "max" and has default length, convert to "max" length
+        for col_name, col_type in columns_to_types.items():
+            if col_type.this in types_with_max_default_param and col_type.expressions:
+                parameter = self.SCHEMA_DIFFER.get_type_parameters(col_type)
+                type_default = types_with_max_default_param[col_type.this]
+                if parameter == type_default:
+                    col_type.set("expressions", [exp.DataTypeParam(this=exp.Var(this="max"))])
```
georgesittas, line 246:
```suggestion
                    col_type.set("expressions", [exp.DataTypeParam(this=exp.var("max"))])
```

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -228,6 +228,26 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+        self, columns_to_types: t.Dict[str, exp.DataType]
+    ) -> t.Dict[str, exp.DataType]:
+        # get default lengths for types that support "max" length
+        types_with_max_default_param = {
+            k: [self.SCHEMA_DIFFER.parameterized_type_defaults[k][0][0]]
+            for k in self.SCHEMA_DIFFER.max_parameter_length
+            if k in self.SCHEMA_DIFFER.parameterized_type_defaults
+        }
+
+        # if type supports "max" and has default length, convert to "max" length
```
izeigerman, line 241: Let's maybe add some more context in comments around why we do this.

izeigerman (approved):

PR #2792, opened 2024-06-19
Fix: normalize default schema name when targeting prod too

georgesittas:
Minor fixup: if we didn't specify an environment name, the default schema wouldn't be normalized.


tobymao (approved):

tobymao:
let's get this in and deploy

PR #2794, opened 2024-06-19
Fix: truncate backfill models and adjust threshold

Themiscodes:
- Adjust `INDIRECTLY_MODIFIED_DISPLAY_THRESHOLD` to 10.
- Limit the output for backfill, as well as added and removed models.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -701,20 +701,20 @@ class TerminalConsole(Console):
         tree = Tree(f"[bold]{header}:")
         if added_snapshot_ids:
             added_tree = Tree("[bold][added]Added:")
-            for s_id in added_snapshot_ids:
+            for s_id in sorted(added_snapshot_ids):
                 snapshot = context_diff.snapshots[s_id]
                 added_tree.add(
                     f"[added]{snapshot.display_name(environment_naming_info, default_catalog, dialect=self.dialect)}"
                 )
-            tree.add(added_tree)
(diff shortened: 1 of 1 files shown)


sqlmesh/core/console.py
```diff
@@ -1509,10 +1511,17 @@ def show_model_difference_summary(
         added_snapshot_models = {s for s in added_snapshots if s.is_model}
         if added_snapshot_models:
             self._print("\n**Added Models:**")
-            for snapshot in sorted(added_snapshot_models):
-                self._print(
-                    f"- `{snapshot.display_name(environment_naming_info, default_catalog, dialect=self.dialect)}`"
-                )
+            added_models = sorted(added_snapshot_models)
+            list_length = len(added_models)
+            if not self.verbose and list_length > self.INDIRECTLY_MODIFIED_DISPLAY_THRESHOLD:
```
izeigerman, line 1516: Where do we set the level of verbosity? Can the `plan` command be run in a verbose mode?

Themiscodes: Yes `plan` command has the option to be executed in `verbose` mode. and it is then set in [/cli/main.py#L393](https://github.com/TobikoData/sqlmesh/blob/b46206bb54237f837c3b83fcbb4ae7c9e94b8bf0/sqlmesh/cli/main.py#L393)

izeigerman: nice, thanks!

izeigerman (approved):

PR #2795, opened 2024-06-19
Refactor!: error when trying to construct source queries from empty DFs

### LINKED ISSUE #2769: Empty Python model DataFrames can lead to no-op EngineAdapter.ctas calls
What the title says. We observed this for an `IncrementalByUniqueKey` Redshift model. The bug was that during a run, the `ctas` call didn't actually create a temp table that was later referenced and Redshift crashed with a "relation does not exist" error. We should make sure that we handle this case gracefully.

georgesittas:
Fixes #2769

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -182,10 +182,18 @@ class EngineAdapter:
         batch_size = self.DEFAULT_BATCH_SIZE if batch_size is None else batch_size
         if isinstance(query_or_df, (exp.Query, exp.DerivedTable)):
             return [SourceQuery(query_factory=lambda: query_or_df)]  # type: ignore
+
         if not columns_to_types:
             raise SQLMeshError(
-                "It is expected that if a DF is passed in then columns_to_types is set"
+                "It is expected that if a DataFrame is passed in then columns_to_types is set"
             )
+        if query_or_df.empty:
+            raise SQLMeshError(
+                "Cannot construct source query from an empty DataFrame. This error is commonly "
+                "related to Python models that produce no data. For such models, consider yielding "
+                "from an empty generator if the resulting set is empty, i.e. use `yield from ()`."
+            )
+
         return self._df_to_source_queries(
             query_or_df, columns_to_types, batch_size, target_table=target_table
         )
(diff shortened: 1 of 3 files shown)


tobymao (approved):

sqlmesh/core/engine_adapter/base.py
```diff
@@ -182,10 +182,18 @@ def _get_source_queries(
...
             return [SourceQuery(query_factory=lambda: query_or_df)]  # type: ignore
+
         if not columns_to_types:
             raise SQLMeshError(
-                "It is expected that if a DF is passed in then columns_to_types is set"
+                "It is expected that if a DataFrame is passed in then columns_to_types is set"
+            )
+        if query_or_df.empty:
+            raise SQLMeshError(
+                "Cannot construct source query from an empty DataFrame. This error is commonly "
```
tobymao, line 192: would it be easier if we just handled the df empty case?

georgesittas: discussed about it with Iaroslav too, but we concluded that this approach would be simpler as we don't have to think about edge cases.

georgesittas: I _think_ it'd involve tracking down uses of `_get_source_queries_and_columns_to_types` and handling the empty `SourceQuery` list case. E.g. in `EngineAdapter.ctas` we could create a table with the `CREATE TABLE <name> <schema>` syntax instead of a CTAS. It's just a bit of a PITA because there are quite a few places where `_get_source_queries_and_columns_to_types` is used.

tests/core/test_context.py
```diff
@@ -285,6 +287,32 @@ def test_plan_execution_time():
...
+    capsys.readouterr()
+    with pytest.raises(SQLMeshError):
+        sushi_context.plan(no_prompts=True, auto_apply=True)
+
+    assert (
+        "Cannot construct source query from an empty \nDataFrame. This error "
+        "is commonly related to Python models that produce no data.\nFor such "
+        "models, consider yielding from an empty generator if the resulting set "
+        "\nis empty, i.e. use `yield from ()`"
+    ) in capsys.readouterr().out
```
georgesittas, line 313: I tested this using `capsys` because if I do `pytest.raises(SQLMeshError, match=...)` it captures the `Plan application failed.` exception which [doesn't include](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/plan/evaluator.py#L174-L175) the exception of interest in its context.

izeigerman (approved):

PR #2796, opened 2024-06-20
Feat: Support custom materialization strategies

izeigerman:
This update introduces an ability for users to provide implementation for their custom materialization strategies.

The custom implementation is provided in Python and in its simplest form looks something like this:

```python
import typing as t

from sqlmesh import CustomMaterialization, Model

if t.TYPE_CHECKING:
    from sqlmesh import QueryOrDF


class MyCustomMaterialization(CustomMaterialization):
    NAME = "my_custom_materialization"

    def insert(
        self,
        table_name: str,
        query_or_df: QueryOrDF,
        model: Model,
        is_first_insert: bool,
        **kwargs: t.Any,
    ) -> None:
        config_value = model.custom_materialization_properties["config_key"]
        self.adapter.execute(...)
```

In order to use the custom materialization strategy a new model kind `CUSTOM` has been introduced:

```sql
MODEL (
  ...,
  kind CUSTOM (
    materialization 'my_custom_materialization',
    materialization_properties (
      config_key = 'config_value'
    )
  )
);
```

There are two ways to add a new materialization strategy to a project: by adding a Python module with the implementation to the `materializations/ `project folder or by installing a Python package that contains the implementation.

## `materializations/` folder

SQLMesh loads all `*.py` files in this folder and registers custom materializations at project load time. This is the simplest way to add a custom materialization. 

However, this approach doesn't work when using the multi-repo setup or when using SQLMesh with Airflow. In the latter case, the custom implementation must be made available in the Airflow cluster, which can be achieved by packaging the custom materializations into a Python package.

## Python package

Custom materializations can be packaged into a Python package and exposed via [setuptools entrypoints](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata) mechanism. Once the package is installed, SQLMesh will automatically load custom materializations from the entrypoint list.

Entrypoints can be specified in the `pyproject.toml` file of the package:
```
[project.entry-points."sqlmesh.materializations"]
my_materialization = "my_package.my_materialization:MyCustomMaterialization"
```

Or in `setup.py`:
```python
setup(
    ...,
    entry_points={
        "sqlmesh.materializations": [
            "my_materialization = my_package.my_materialization:MyCustomMaterialization",
        ],
    },
)
```

Documentation updates and guides are going to be a part of a follow up PR(s).


Makefile
```diff
@@ -1,16 +1,16 @@
 .PHONY: docs
 
 install-dev:
-	pip3 install -e ".[dev,web,slack]"
+	pip3 install -e ".[dev,web,slack]" && cd ./examples/custom_materializations && pip3 install -e .
```
tobymao, line 4: this is kinda invasive, now i can't do pip install -e .dev anymore.

is there a way to add this into setup.py ?

sqlmesh/core/model/kind.py
```diff
@@ -601,6 +609,72 @@ class ExternalKind(_ModelKind):
...
+            *super().metadata_hash_values,
+            str(self.forward_only),
+            str(self.disable_restatement),
+        ]
+
+    def to_expression(self, **kwargs: t.Any) -> d.ModelKind:
+        return super().to_expression(
+            expressions=[
+                exp.Property(
+                    this=exp.Var(this="materialization"),
```
tobymao, line 660: exp.var("materialization")

sqlmesh/core/snapshot/definition.py
```diff
@@ -423,6 +425,13 @@ class SnapshotTableInfo(PydanticModel, SnapshotInfoMixin, frozen=True):
     # This can be removed from this model once Pydantic 1 support is dropped (must remain in `Snapshot` though)
     base_table_name_override: t.Optional[str] = None
 
+    custom_materialization: t.Optional[str] = None
```
tobymao, line 428: why is this needed on the table info?

izeigerman: Because in snapshot evaluator we don't always have the full snapshot available. Particularly during demotion and cleanup.

Makefile
izeigerman: I couldn't find a non-hacky way to do this through `setup.py` unfortunately.

sqlmesh/core/snapshot/definition.py
tobymao: in demotion, why do you need the custom materialization? do you know what the materialization is for a non custom materialization is for demotion/ cleanup?

izeigerman: > do you know what the materialization is for a non custom materialization is for demotion/ cleanup?

we do based on the kind name

> in demotion, why do you need the custom materialization?

eg. I could decide to use branches in LakeFS or Iceberg, etc. instead of views. Even if not for demotion, we still need this for cleanup.

tobymao: so then don't you only need kind name, why do you also need properties?

should we just have kind_name be the custom kind name instead of custom?

Makefile
tobymao: is it something like this? https://stackoverflow.com/questions/35668295/how-to-include-and-install-local-dependencies-in-setup-py-in-python

sqlmesh/core/snapshot/definition.py
izeigerman: I'd rather not mix kind name and the materialization name. They are different concepts. We need properties to correctly initialize the custom materialization class instance.

Makefile
izeigerman: Yes, exactly. As far as I understand the `egg` package should already be available which doesn't solve the original problem, but I could be wrong and can try this approach out if you feel like it's better.

sqlmesh/core/snapshot/definition.py
tobymao: why do custom materializations need properties for initialization but not regular materializations?

Makefile
tobymao: there's gotta be a way to install from source

sqlmesh/core/snapshot/definition.py
izeigerman: Because in regular materializations the logic is hardcoded. Properties provide additional flexibility for developers to reuse the same implementation with slightly different settings in different models.

Makefile
izeigerman: I think I fixed this

sqlmesh/core/snapshot/definition.py
tobymao: ok

izeigerman: Also builtin materialization use the model definition directly and it just happens that the cleanup and demotion behavior don't depend on a specific model.

izeigerman: Actually, I think that custom materialization properties can be sourced from the model definition itself which is available in all cases except demotion and cleanup. This doesn't seem like a big deal but makes the implementation cleaner. So I'll probably go ahead and change that.

izeigerman: Done that

tobymao (approved):

PR #2801, opened 2024-06-20
Fix: Check all dialects used in a project when retrieving a model by its name

izeigerman:
Before this fix, the `get_model` method won't find a model if the model's dialect is set to eg. `snowflake` while the default dialect is something like `duckdb`.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -624,21 +624,27 @@ class GenericContext(BaseContext, t.Generic[C]):
             The expected model.
         """
         if isinstance(model_or_snapshot, str):
-            normalized_name = normalize_model_name(
-                model_or_snapshot,
-                dialect=self.default_dialect,
-                default_catalog=self.default_catalog,
-            )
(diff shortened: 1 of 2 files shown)


sqlmesh/core/context.py
```diff
@@ -624,21 +624,27 @@ def get_model(
...
         """
         if isinstance(model_or_snapshot, str):
-            normalized_name = normalize_model_name(
-                model_or_snapshot,
-                dialect=self.default_dialect,
-                default_catalog=self.default_catalog,
-            )
-            model = self._models.get(normalized_name)
+            # We should try all dialects referenced in the project for cases when models use mixed dialects.
+            all_dialects = {m.dialect for m in self._models.values() if m.dialect} | {
```
tobymao, line 628: cache this otherwise get_model becomes n^2

izeigerman: I don't think we ever use `get_model` in a loop, do we?

tobymao: no but it could be

PR #2802, opened 2024-06-21
Fix: use isEmpty() for non pandas dataframes

georgesittas:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -187,7 +187,11 @@ def _get_source_queries(
             raise SQLMeshError(
                 "It is expected that if a DataFrame is passed in then columns_to_types is set"
             )
-        if query_or_df.empty:
+
+        df_is_empty = (
+            query_or_df.empty if isinstance(query_or_df, pd.DataFrame) else query_or_df.isEmpty()
```
tobymao, line 192: pyspark and snowpark? should we check hasattr instead?

georgesittas: Yes the intention was to handle pandas vs snowpark/pyspark differently (should fix CI failures), but I'm actually not sure if `isEmpty()` is a valid snowpark method - I'll double check this. We could also use `hasattr` here, I don't feel strongly about it, only  a gut feeling that it may be more hacky/brittle. Happy to go either way, though.

tobymao: check snowpark it needs to work with that too

georgesittas: yep just did, it doesn't support `isEmpty()`, we need to use `count()` for snowpark dataframes

tobymao: you could use count for all?

tobymao: actually that’s bad for spark

tobymao: is there even an issue imfor empty snowpark and pyspark? this should only be an issue for empty pandas dataframes to sql

tobymao: i think you only need to check empty for pandas dataframes

izeigerman (approved):

PR #2806, opened 2024-06-21
Feat: enable dbt transpilation

treysp:
This PR allows dbt project models and tests to be written in a different SQL dialect than the target engine on which the project is being run.

A project has a default dialect - all models/tests without an explicitly configured dialect are assumed to be written in the default. In a dbt project without an explicit default dialect specified, the target engine dialect is the default. 

Specify a default dialect other than the target engine in the `config.py` file model defaults argument
  - Example: `config=sqlmesh_config(Path(__file__).parent, model_defaults=ModelDefaultsConfig(dialect="bigquery"))

Specify dialects for individual models/tests in the standard way:
- Configuration jinja at top of file: `{{ config(dialect='bigquery') }}`
- YAML configuration files in file directories 
- dbt_project.yml

Limitations
- SQLMesh allows kind-specific parameters like `time_column` to be specified in a different dialect than the model query itself. In dbt, kind-specific parameters are parsed with the model's dialect.

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -302,7 +302,7 @@ def create_builtin_globals(
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
     api = Api(target)
-    dialect = jinja_globals.pop("dialect", None) or dict(target or {}).get("dialect")
+    dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)  # type: ignore
 
     builtin_globals["api"] = api
 
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -43,9 +43,9 @@ class Exceptions:
 
 
 class Api:
-    def __init__(self, target: t.Optional[AttributeDict] = None) -> None:
-        if target:
-            config_class = TARGET_TYPE_TO_CONFIG_CLASS[target["type"]]
+    def __init__(self, dialect: str) -> None:
+        if dialect:
+            config_class = TARGET_TYPE_TO_CONFIG_CLASS[dialect]
             self.Relation = config_class.relation_class
             self.Column = config_class.column_class
             self.quote_policy = config_class.quote_policy
@@ -301,8 +301,8 @@ def create_builtin_globals(
     jinja_globals = jinja_globals.copy()
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
-    api = Api(target)
     project_dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)  # type: ignore
+    api = Api(project_dialect)
 
     builtin_globals["api"] = api
 
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -302,7 +302,7 @@ def create_builtin_globals(
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
     api = Api(target)
-    dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)  # type: ignore
+    project_dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)  # type: ignore
 
     builtin_globals["api"] = api
 
@@ -349,13 +349,14 @@ def create_builtin_globals(
             snapshots=jinja_globals.get("snapshots", {}),
             table_mapping=jinja_globals.get("table_mapping", {}),
             deployability_index=jinja_globals.get("deployability_index"),
+            project_dialect=project_dialect,
         )
     else:
         builtin_globals["flags"] = Flags(which="parse")
         adapter = ParsetimeAdapter(
             jinja_macros,
             jinja_globals={**builtin_globals, **jinja_globals},
-            dialect=dialect,
+            target_dialect=(target.dialect if target else None),  # type: ignore
         )
 
     sql_execution = SQLExecution(adapter)
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -302,7 +302,7 @@ def create_builtin_globals(
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
     api = Api(target)
-    dialect = target.dialect if target else None  # type: ignore
+    dialect = jinja_globals.pop("dialect", None) or dict(target or {}).get("dialect")
 
     builtin_globals["api"] = api
 
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -43,7 +43,7 @@ class Exceptions:
 
 
 class Api:
-    def __init__(self, dialect: str) -> None:
+    def __init__(self, dialect: t.Optional[str]) -> None:
         if dialect:
             config_class = TARGET_TYPE_TO_CONFIG_CLASS[dialect]
             self.Relation = config_class.relation_class
@@ -301,7 +301,9 @@ def create_builtin_globals(
     jinja_globals = jinja_globals.copy()
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
-    project_dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)  # type: ignore
+    project_dialect = jinja_globals.pop("dialect", None) or (
+        target.get("dialect") if target else None
+    )
     api = Api(project_dialect)
 
     builtin_globals["api"] = api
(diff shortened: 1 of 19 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -302,7 +302,7 @@ def create_builtin_globals(
 
     target: t.Optional[AttributeDict] = jinja_globals.get("target", None)
     api = Api(target)
-    dialect = target.dialect if target else None  # type: ignore
+    dialect = jinja_globals.pop("dialect", None) or dict(target or {}).get("dialect")
```
crericha, line 305:
```suggestion
    dialect = jinja_globals.pop("dialect", None) or (target.dialect if target else None)
```

sqlmesh/dbt/context.py
```diff
@@ -229,6 +227,9 @@ def jinja_globals(self) -> t.Dict[str, JinjaGlobalAttribute]:
             output["project_name"] = self.project_name
         if self._target is not None:
             output["target"] = self._target.attribute_dict()
+        # pass user-specified default dialect if we have already loaded the config
+        if self.sqlmesh_config.dialect:
+            output["dialect"] = self.sqlmesh_config.dialect
```
crericha, line 232: Since we now need to handle dialect in the jinja, would it make sense to always pass in dialect here? It would simplify `create_builtin_globals` code.

sqlmesh/dbt/context.py
```diff
@@ -59,10 +59,8 @@ class DbtContext:
     _manifest: t.Optional[ManifestHelper] = None
 
     @property
-    def dialect(self) -> str:
-        if not self.target:
-            raise SQLMeshError("Target must be configured before calling the dialect property.")
-        return self.target.dialect
+    def default_dialect(self) -> str:
+        return self.sqlmesh_config.dialect or self.target.dialect
```
izeigerman, line 63: Do we no longer expect `target` to be missing?

sqlmesh/dbt/profile.py
```diff
@@ -26,19 +26,27 @@ def __init__(
         path: Path,
         target_name: str,
         target: TargetConfig,
+        project_model_config: t.Dict[str, t.Any],
```
izeigerman, line 29: This doesn't look right to me. IMHO the `profile` object should not be aware of the project config.

izeigerman: What's the idea behind putting it here?

sqlmesh/dbt/loader.py
```diff
@@ -31,6 +31,36 @@
     from sqlmesh.core.context import GenericContext
 
 
+def get_default_dialect(
+    model_defaults_config: ModelDefaultsConfig, profile: Profile, profile_name: t.Optional[str]
+) -> ModelDefaultsConfig:
```
crericha, line 36: I don't think this method is needed. Either you've specified a default dialect in sqlmesh, the model has a dialect via config (that will be set in basemodel, or you use the target. I believe all we need to do is set the model defaults in context and let the overrides of that happen later in the model logic.

sqlmesh/utils/jinja.py
```diff
@@ -567,6 +567,7 @@ def create_builtin_globals(
     jinja_macros: JinjaMacroRegistry, global_vars: t.Dict[str, t.Any], *args: t.Any, **kwargs: t.Any
 ) -> t.Dict[str, t.Any]:
     global_vars.pop(c.GATEWAY, None)
+    global_vars.pop("dialect", None)  # may be set in dbt context
```
izeigerman, line 570: I don't think this method can ever be invoked in the context of a dbt project.

sqlmesh/dbt/context.py
crericha: Also, do we want to pass in the default dialect? Wouldn't it make more sense to pass in the model dialect during `to_sqlmesh` in the model?

sqlmesh/dbt/loader.py
```diff
@@ -42,8 +72,9 @@ def sqlmesh_config(
     project_root = project_root or Path()
     context = DbtContext(project_root=project_root)
     profile = Profile.load(context, target_name=dbt_target_name)
-    model_defaults = kwargs.pop("model_defaults", ModelDefaultsConfig())
-    model_defaults.dialect = profile.target.dialect
+    model_defaults = get_default_dialect(
```
crericha, line 75: Following up on the above. All we need here is the model's default dialect if specified in sqlmesh.

sqlmesh/dbt/model.py
```diff
@@ -201,7 +201,13 @@ def model_kind(self, context: DbtContext) -> ModelKind:
         if materialization == Materialization.VIEW:
             return ViewKind()
         if materialization == Materialization.INCREMENTAL:
-            incremental_materialization_kwargs: t.Dict[str, t.Any] = {"dialect": context.dialect}
+            # SQLMesh `_IncrementalBy` and SCD models can set dialect at both the model and
+            # kind levels, where model level applies to query and kind level applies to kind-
+            # specific components like time_column. In dbt, the kind level is constrained to
+            # be the same dialect as the model level.
```
crericha, line 207: I'm not sure the comment is helpful. The code speaks for itself imo.

sqlmesh/dbt/model.py
```diff
@@ -201,7 +201,13 @@ def model_kind(self, context: DbtContext) -> ModelKind:
...
+            incremental_materialization_kwargs: t.Dict[str, t.Any] = {
+                "dialect": self.dialect or context.default_dialect
```
crericha, line 209: Since you are using `self.dialect or context.default_dialect` multiple times, I recommend making it a property that you call rather than duplicating the logic.

sqlmesh/dbt/model.py
```diff
@@ -358,14 +364,22 @@ def sqlmesh_config_fields(self) -> t.Set[str]:
...
-                [exp.to_column(val) for val in self.partition_by]
+                [
+                    exp.to_column(
+                        t.cast(
+                            exp.Column, d.parse_one(val, dialect=model_dialect, into=exp.Column)
+                        ),
+                        dialect=context.target.dialect,
+                    )
+                    for val in self.partition_by
+                ]
```
crericha, line 382: I'm not sure this is needed. Aren't we creating directly into the AST and the sqlglot generator will handle transpiling for us? How is `partitioned_by` handled on the sqlmesh side?

sqlmesh/dbt/profile.py
crericha, line 29: Is it possible to put this in the `DbtContext` instead of adding to the profile, which is a dbt-ism?

sqlmesh/dbt/seed.py
```diff
@@ -50,7 +50,7 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
         return create_seed_model(
             self.canonical_name(context),
             SeedKind(path=seed_path),
-            dialect=context.dialect,
+            dialect=context.target.dialect,
```
crericha, line 53: Should you be allowed to override dialect here? I was thinking yes.

sqlmesh/dbt/test.py
```diff
@@ -142,7 +143,7 @@ def to_sqlmesh(self, context: DbtContext) -> Audit:
             jinja_macros.add_globals({"this": self.relation_info})
             audit = StandaloneAudit(
                 name=self.name,
-                dialect=context.dialect,
+                dialect=self.dialect or context.default_dialect,
```
crericha, line 146: property

sqlmesh/utils/jinja.py
crericha, line 570: What's the reason this is needed here? We `pop` in `create_builtins`, no? I believe there are other globals specific to dbt handled there.

sqlmesh/dbt/model.py
treysp: `exp.to_column(val)` is parsing the raw column string from the model definition and returning an expression. If someone includes dialect-specific characters, I think it will choke. 

cluster_by is also parsing column names just below and passes a dialect, so maybe it not being here already is an oversight?
https://github.com/TobikoData/sqlmesh/blob/b7c659e66b6eab5ec30a6f5d341937d8746de683/sqlmesh/dbt/model.py#L377

Either way, this initial implementation is messy, so I'll clean it up and we can remove if it's not necessary.

sqlmesh/dbt/context.py
treysp: We can't pass in self.default_dialect because target hasn't been set the first time this is called and we'll error.

I'm not following your second comment - where in `to_sqlmesh` can we modify the globals? Here?
https://github.com/TobikoData/sqlmesh/blob/b7c659e66b6eab5ec30a6f5d341937d8746de683/sqlmesh/dbt/model.py#L382

izeigerman: @crericha  I think this is only relevant in the context of dbt projects and doesn't apply to native projects, so not sure whether we need the dialect as a top-level argument.

tobymao (approved):

tobymao (approved):

PR #2809, opened 2024-06-24
Feat: Managed models

erindru:
Initial implementation of Managed models. This defines the general semantics and contains an implementation for Snowflake, based on Dynamic Tables.

As a side effect, this PR also fixes [issue 1049](https://github.com/TobikoData/sqlmesh/issues/1049).

Follow-up PR's:
 - Support for other engines
 - Cleanup / compaction support, essentially the ability for something like `sqlmesh clean` to drop any objects that aren't referenced in a specific environment. The assumption up until now is that storage is cheap so keeping old snapshots around for a while isn't a problem. However, costs could add up quickly with managed models because users would be paying for models to be refreshed that aren't actually being referenced

### DIFF
--- a/docs/concepts/models/managed_models.md
+++ b/docs/concepts/models/managed_models.md
@@ -58,7 +58,7 @@ MODEL (
     target_lag = '2 minutes',
     data_retention_time_in_days = 2
   )
-)
+);
 
 SELECT
   event_date::DATE as event_date,
--- a/docs/concepts/models/managed_models.md
+++ b/docs/concepts/models/managed_models.md
@@ -26,7 +26,7 @@ Managed models follow the same lifecycle as other models:
 
 - Creating a Virtual Environment creates a pointer to the current model snapshot
 - Modifying the model causes a new snapshot to be created
-- Breaking upstream changes cause a new snapshot to be created
+- Any upstream changes cause a new snapshot to be created
 - The model can be deployed and rolled back via the usual pointer swap mechanism
 - Once the TTL expires, model snapshots are cleaned up
 
--- a/docs/concepts/models/managed_models.md
+++ b/docs/concepts/models/managed_models.md
@@ -1,0 +1,107 @@
+# Managed models
+Unlike normal tables where the user is responsible for managing the data within the table, some database engines have a concept of a table where the engine itself ensures that the data within the table is up to date. These tables are typically based on a query that reads from other tables within the database. Each time these other tables are updated, the database will ensure that the managed table reflects the changes without the user having to do anything special (such as issue a `REFRESH` command).
+
+Under the hood, each supported database engine achieves this in a slightly different way but most of them have background processes that run and automatically keep the tables up to date, within the parameters you define when you create the table.
+
+For supported engines, we expose this functionality through Managed models. This indicates to SQLMesh that the underlying database engine will ensure that the data remains up to date and all SQLMesh needs to do is maintain the schema.
+
+Due to this, managed models would typically be built off an [External Model](./external_models.md) rather than another SQLMesh model. Since SQLMesh already ensures that models it's tracking are kept up to date, the main benefit of managed models comes when they read from external tables that arent tracked by SQLMesh.
+
+## Difference from materialized views
+The difference between an Managed model and a materialized view is down to semantics and in some engines there is no difference.
+
+SQLMesh has support for [materialized views](./model_kinds#materialized-views) already. However, depending on the engine, these are subject to some limitations, such as:
+
+- A Materialized View query can only be derived from a single base table
+- The Materialized View is not automatically maintained by the engine. To refresh the data, a `REFRESH MATERIALIZED VIEW` or equivalent command must be issued
+
+Managed models are different in that:
+
+- The engine updates the table data automatically when a base table changes
+- When performing updates, the engine has a semantic understanding of the query and can decide if an incremental or full refresh should be applied
+- There is no need to issue manual `REFRESH` commands. The engine maintains the table transparently in a background process
+
+## Lifecycle in SQLMesh
+Managed models follow the same lifecycle as other models:
+
+- Creating a Virtual Environment creates a pointer to the current model snapshot
+- Modifying the model causes a new snapshot to be created
+- Breaking upstream changes cause a new snapshot to be created
+- The model can be deployed and rolled back via the usual pointer swap mechanism
+- Once the TTL expires, model snapshots are cleaned up
+
+However, there is usually extra vendor-imposed costs associated with Managed models. For example, Snowflake has [additional costs](https://docs.snowflake.com/en/user-guide/dynamic-tables-cost) for Dynamic Tables.
+
+Therefore, we try to not create managed tables unnecessarily. For example, in [forward-only plans](../plans.md#forward-only-change) we just create a normal table to preview the changes and only re-create the managed table on deployment to prod.
+
+## Supported Engines
+SQLMesh supports managed models in the following database engines:
+
+| Engine                                               | Implementatation                                                               |
+| ---------------------------------------------------- | ------------------------------------------------------------------------------ |
+| [Snowflake](../../integrations/engines/snowflake.md) | [Dynamic Table](https://docs.snowflake.com/en/user-guide/dynamic-tables-intro) |
+
+To define a managed model, you can use the [`MANAGED`](./model_kinds.md#managed) model Kind.
+
+### Snowflake
+
+Managed Models are in Snowflake are implemented as [Dynamic Tables](https://docs.snowflake.com/en/user-guide/dynamic-tables-intro).
+
+Here is an example of a SQLMesh model that will result in a dynamic table being created:
+
+```sql linenums="1"
+MODEL (
+  name db.events,
+  kind MANAGED,
+  physical_properties (
+    warehouse = datalake,
+    target_lag = '2 minutes',
+    data_retention_time_in_days = 2
+  )
+)
+
+SELECT
+  event_date::DATE as event_date,
+  event_payload::TEXT as payload
+FROM raw_events
+```
+
+results in:
+
+```sql linenums="1"
+CREATE OR REPLACE DYNAMIC TABLE db.events
+  WAREHOUSE = "datalake",
+  TARGET_LAG = '2 minutes'
+  DATA_RETENTION_TIME_IN_DAYS = 2
+AS SELECT
+  event_date::DATE as event_date,
+  event_payload::TEXT as payload
+FROM raw_events
+```
+
+!!! info
+
+    Note that SQLMesh will not create intervals and run this model for each interval, so there is no need to add a WHERE clause with date filters like you would for a normal incremental model. How the data in this model is refreshed is completely up to Snowflake.
+
+#### Table properties
+
+Dynamic Tables have some properties that affect things like how often the data is refreshed by Snowflake, when the initial data is populated, how long data is retained for etc. The list of available properties is located in the [Snowflake documentation](https://docs.snowflake.com/sql-reference/sql/create-dynamic-table).
+
+In SQLMesh, these properties are set on the model definition.
+
+The following Dynamic Table properties are set on the model [`physical_properties`](../models/overview.md#physical_properties-previously-table_properties):
+
+| Snowflake Property              | Required | Notes
+| ------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| target_lag                      | Y        |                                                                                                                                         |
+| warehouse                       | N        | In Snowflake, this is a required property. However, if not specified, then SQLMesh will use the result of `select current_warehouse()`. |
+| refresh_mode                    | N        |                                                                                                                                         |
+| initialize                      | N        |                                                                                                                                         |
+| data_retention_time_in_days     | N        |                                                                                                                                         |
+| max_data_extension_time_in_days | N        |                                                                                                                                         |
+
+The following Dynamic Table properties can be set directly on the model:
+
+| Snowflake Property | Required   | Notes                                                                                                                                                                   |
+| ------------------ | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| cluster by         | N          | `clustered_by` is a [standard model property](../models/overview.md#clustered_by), so set `clustered_by` on the model to add a `CLUSTER BY` clause to the Dynamic Table |
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -90,6 +90,7 @@ class EngineAdapter:
     SUPPORTS_MATERIALIZED_VIEWS = False
     SUPPORTS_MATERIALIZED_VIEW_SCHEMA = False
     SUPPORTS_CLONING = False
+    SUPPORTS_MANAGED_MODELS = False
     SCHEMA_DIFFER = SchemaDiffer()
     SUPPORTS_TUPLE_IN = True
     CATALOG_SUPPORT = CatalogSupport.UNSUPPORTED
@@ -428,6 +429,35 @@ class EngineAdapter:
             **kwargs,
         )
 
+    def create_managed_table(
+        self,
+        table_name: TableName,
+        query: Query,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        partitioned_by: t.Optional[t.List[exp.Expression]] = None,
+        clustered_by: t.Optional[t.List[str]] = None,
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """Create an managed table using a query.
+
+        "Managed" means that once the table is created, the data is kept up to date by the underlying database engine and not SQLMesh.
+
+        Args:
+            table_name: The name of the table to create. Can be fully qualified or just table name.
+            query: The SQL query for the engine to base the managed table on
+            columns_to_types: A mapping between the column name and its data type.
+            partitioned_by: The partition columns or engine specific expressions, only applicable in certain engines. (eg. (ds, hour))
+            clustered_by: The cluster columns, only applicable in certain engines. (eg. (ds, hour))
+            table_properties: Optional mapping of engine-specific properties to be set on the managed table
+            table_description: Optional table description from MODEL DDL.
+            column_descriptions: Optional column descriptions from model query.
+            kwargs: Optional create table properties.
+        """
+        raise NotImplementedError(f"Engine does not support managed tables: {type(self)}")
+
     def ctas(
         self,
         table_name: TableName,
@@ -610,6 +640,7 @@ class EngineAdapter:
         replace: bool = False,
         table_description: t.Optional[str] = None,
         column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        table_kind: t.Optional[str] = None,
         **kwargs: t.Any,
     ) -> None:
         table = exp.to_table(table_name)
@@ -652,6 +683,7 @@ class EngineAdapter:
                             exists=exists,
                             replace=replace,
                             table_description=table_description,
+                            table_kind=table_kind,
                             **kwargs,
                         )
                     else:
@@ -679,6 +711,7 @@ class EngineAdapter:
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
         table_description: t.Optional[str] = None,
         column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        table_kind: t.Optional[str] = None,
         **kwargs: t.Any,
     ) -> None:
         self.execute(
@@ -693,6 +726,7 @@ class EngineAdapter:
                     if self.COMMENT_CREATION_TABLE.supports_schema_def and self.comments_enabled
                     else None
                 ),
+                table_kind=table_kind,
                 **kwargs,
             )
         )
@@ -704,6 +738,8 @@ class EngineAdapter:
         exists: bool = True,
         replace: bool = False,
         columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        table_description: t.Optional[str] = None,
+        table_kind: t.Optional[str] = None,
         **kwargs: t.Any,
     ) -> exp.Create:
         exists = False if replace else exists
@@ -717,14 +753,18 @@ class EngineAdapter:
 
         properties = (
             self._build_table_properties_exp(
-                **kwargs, catalog_name=catalog_name, columns_to_types=columns_to_types
+                **kwargs,
+                catalog_name=catalog_name,
+                columns_to_types=columns_to_types,
+                table_description=table_description,
+                table_kind=table_kind,
             )
-            if kwargs
+            if kwargs or table_description
             else None
         )
         return exp.Create(
             this=table_name_or_schema,
-            kind="TABLE",
+            kind=table_kind or "TABLE",
             replace=replace,
             exists=exists,
             expression=expression,
@@ -791,7 +831,30 @@ class EngineAdapter:
             table_name: The name of the table to drop.
             exists: If exists, defaults to True.
         """
-        drop_expression = exp.Drop(this=exp.to_table(table_name), kind="TABLE", exists=exists)
+        self._drop_tablelike_object(table_name, exists)
+
+    def drop_managed_table(self, table_name: TableName, exists: bool = True) -> None:
+        """Drops a managed table.
+
+        Args:
+            table_name: The name of the table to drop.
+            exists: If exists, defaults to True.
+        """
+        raise NotImplementedError(f"Engine does not support managed tables: {type(self)}")
+
+    def _drop_tablelike_object(
+        self, table_name: TableName, exists: bool = True, kind: str = "TABLE"
+    ) -> None:
+        """Drops a "tablelike object".
+
+        A "tablelike object" could be a TABLE or a DYNAMIC TABLE or a TEMPORARY TABLE etc depending on the context.
+
+        Args:
+            table_name: The name of the table to drop.
+            exists: If exists, defaults to True.
+            kind: What kind of object to drop. Defaults to TABLE
+        """
+        drop_expression = exp.Drop(this=exp.to_table(table_name), kind=kind, exists=exists)
         self.execute(drop_expression)
 
     def get_alter_expressions(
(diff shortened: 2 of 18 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1661,6 +1663,83 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+        table_name: str,
+        model: Model,
+        is_table_deployable: bool,
+        **render_kwargs: t.Any,
+    ) -> None:
+        rendered_query = model.render_query_or_raise(**render_kwargs)
+
+        # only create the managed table if this snapshot can actually be deployed to prod (eg if an upstream model doesnt have a forward-only change)
+        # vendors providing managed tables (such as Snowflake's dynamic tables) tend to charge more for them, they arent as cheap as normal tables
+        if is_table_deployable:
```
erindru, line 1678: I'm still not 100% clear on what to do here.

I notice that the SnapshotEvaluator calls `create()` twice, once for the `_temp` version (`is_table_deployable=False`) and then immediately afterwards for the actual version `is_table_deployable=True`.

The assumption is that the `_temp` version can just be a normal table because it would only ever be populated with preview data. However, I still don't fully understand how this all ties in with the case where the production table gets shallow cloned in order to test a forward-only change in a non-prod environment

tobymao:
not sure i like engine managed. 

what about dynamic, or live, or auto

erindru:
> not sure i like engine managed.
> 
> what about dynamic, or live, or auto

Dynamic = Snowflake terminology
Live = Databricks terminology

I'm not against those names per se, but I did think we were trying to remain vendor neutral.

Auto = could work? "Automatic Models"? 

The other way to do this could be to drop any pretense of a generic implementation have a Kind per Engine. That way, we can name them accordingly and also "strongly type" the `physical_properties` and make them part of the Kind instead of arbitrary key-value pairs, which would allow easier validation via pydantic.

For example:
```
MODEL (
   name foo,
   kind SNOWFLAKE_DYNAMIC_TABLE (
     warehouse "small",
     target_lag '20 minutes'
   ),
  ...
```

The downside is that engine-specific Kind's sort of defeat the purpose

tobymao:
maybe auto then

sqlmesh/core/engine_adapter/base.py
```diff
@@ -784,16 +823,25 @@ def clone_table(
             )
         )
 
-    def drop_table(self, table_name: TableName, exists: bool = True) -> None:
+    def drop_table(self, table_name: TableName, exists: bool = True, kind: str = "TABLE") -> None:
```
izeigerman, line 826: I think this is leaking engine implementation details to the outside. Eg. a caller of the `drop_table ` now need to worry about whether it's a "DYNAMIC TABLE" or something else. 

If you want to reuse this API to implement `drop_engine_managed_table`, then I'd suggest having a "private" method that does take the `kind` argument and is called by both `drop_table` and `drop_engine_managed_table`.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -91,6 +87,12 @@ def session(self, properties: SessionProperties) -> t.Iterator[None]:
         yield
         self.execute(f"USE WAREHOUSE {current_warehouse_sql}")
 
+    @property
+    def current_warehouse(self) -> exp.Identifier:
```
izeigerman, line 91: [Nit] Looks like an implementation detail to me. Should this be `_current_warehouse`?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -99,6 +101,97 @@ def snowpark(self) -> t.Optional[SnowparkSession]:
...
+        table_properties = {k.upper(): v for k, v in (table_properties or {}).items()}
+
+        # the WAREHOUSE property is required for a Dynamic Table
+        if "WAREHOUSE" not in table_properties:
+            table_properties["WAREHOUSE"] = self.current_warehouse
+
+        # so is TARGET_LAG
+        # ref: https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table
+        if "TARGET_LAG" not in table_properties:
+            raise SQLMeshError(
```
izeigerman, line 129: Can we also validate this in the model definition to make the feedback loop faster.

erindru: I guess we could for `TARGET_LAG` but not for `WAREHOUSE` since that requires access to either the DB connection config or the ability to run queries

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1661,6 +1663,83 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+            logger.info("Creating managed table: %s", table_name)
+            self.adapter.create_engine_managed_table(
```
izeigerman, line 1680: Shouldn't we defer the creation of the deployable table until the change is promoted to production?

sqlmesh/core/model/definition.py
```diff
@@ -1331,6 +1336,14 @@ class PythonModel(_Model):
     entrypoint: str
     source_type: Literal["python"] = "python"
 
+    @model_validator(mode="after")
+    def verify_compatibility(self) -> Self:
```
izeigerman, line 1340: I think this won't work with Pydantic V1. You need the `@model_validator_v1_args` decorator and return `t.Dict[str, t.Any]` instead.

izeigerman: For example: https://github.com/TobikoData/sqlmesh/blob/273e913847173d20e9eb291df3eeaf998ec679e7/sqlmesh/core/audit/definition.py#L298-L299

sqlmesh/core/model/definition.py
```diff
@@ -12,7 +12,7 @@
 import pandas as pd
 import numpy as np
 from astor import to_source
-from pydantic import Field
+from pydantic import Field, model_validator
```
izeigerman, line 15: Wrong import for `model_validator`. You need `from sqlmesh.utils.pydantic import ` one

sqlmesh/core/model/definition.py
erindru: Yeah, I realised this after the check failed and Toby explained why we still needed v1 compatibility.

I moved the check into `validate_definition()` so it gets run when the Context loads the model rather than when Pydantic creates the instance

sqlmesh/core/engine_adapter/snowflake.py
erindru: Done, i've added a branch in `validate_definition` although this method in looks like it might need to be refactored at some point because it does a bunch of checks that perhaps should live in the model Kind classes

sqlmesh/core/engine_adapter/snowflake.py
erindru: Ah yeah I just copied the same approach as the `snowpark` property. Have updated

sqlmesh/core/engine_adapter/base.py
erindru: Yeah this did feel a bit awkward.

I've reinstated the previous public API and created a new private method for dropping table-like objects

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1661,6 +1663,93 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
     return strategy_type
 
 
+class EngineManagedStrategy(PromotableStrategy):
```
erindru, line 1666: Right, so, I think I have figured out what to do here.

On `create()`, because the DeployabilityIndex always returns False to allow creating self-referential models, we just create the dev table and carry on. The call to create the prod table is a no-op because we dont yet know if it will be deployable.

On `insert()`, the DeployabilityIndex works as expected. If the current snapshot is deployable, we call `create_managed_table`. Otherwise, we replace the dev table with the preview data. We only do this for the "first insert" because managed models are not incremental.

This approach means that promotion works as expected with no need for a table creation side-effect.

sqlmesh/core/model/definition.py
```diff
@@ -58,6 +58,11 @@
 else:
     from typing_extensions import Literal
 
+if sys.version_info >= (3, 11):
+    pass
+else:
+    pass
```
georgesittas, line 64: let's get rid of this?

georgesittas:
FYI in case it's somehow relevant to this PR: https://github.com/tobymao/sqlglot/commit/60fa5e3f8a6eab3abb12064366a6bde907d9e9de.

sqlmesh/core/snapshot/evaluator.py
izeigerman: This sounds really good 👍

izeigerman: Actually, this won't work if the "MANAGED" model snapshot IS deployable and has downstream dependencies that are part of the same plan. The downstream models will attempt to create their own tables referencing the deployable version of the "MANAGED" model and it will be missing at that point.

izeigerman: What you probably need to do is to check whether the snapshot is deployable **before** we modify the deployability index to make the current snapshot non-deployable. Eg. add something like this:
```
is_snapshot_deployable = deployability_index.is_deployable(snapshot)
```

erindru:
> FYI in case it's somehow relevant to this PR: [tobymao/sqlglot@60fa5e3](https://github.com/tobymao/sqlglot/commit/60fa5e3f8a6eab3abb12064366a6bde907d9e9de).

Ahh, thanks, good to know that exists now.

I think it's more relevant in a future PR. Using it now would involve modifying a bunch of method signatures (for example, `drop_table` would need to be modified to take table properties and the current mechanism of specifying table properties elsewhere assumes key-value pairs instead of individual properties for anything that isnt "top level")

sqlmesh/core/model/definition.py
erindru: Well spotted, the linter removed the unused imports when I refactored and I didnt notice the conditionals were still there

sqlmesh/core/snapshot/evaluator.py
erindru: Of course :disappointed: , I was so focused on upstream models and if they were forward-only or not that I completely forgot about downstream models.

I've implemented your suggestion to pass the snapshot deployability to `create()` since we cant rely on the DeployabilityIndex there.

This changes the logic to:

- on `create()`, if both the table and snapshot are deployable, we create a managed table. Otherwise, we fallback to the superclass logic which just creates a normal table
- on `insert()`, we populate the table only if the snapshot is NOT deployable (which would have resulted in a normal table being created in `create()`). If the snapshot IS deployable, then it's a no-op because the data would have been populated in `create()`.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1071,6 +1081,7 @@ def create(
         table_name: str,
         model: Model,
         is_table_deployable: bool,
+        is_snapshot_deployable: bool,
```
izeigerman, line 1084: Can we not make it a part of the public API to avoid making it more complex for users willing to implement a custom strategy. Can we hide this in `kwargs`?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1666,6 +1684,82 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+        query_or_df: QueryOrDF,
+        model: Model,
+        is_first_insert: bool,
+        **kwargs: t.Any,
+    ) -> None:
+        deployability_index: DeployabilityIndex = kwargs["deployability_index"]
+        snapshot: Snapshot = kwargs["snapshot"]
+        is_snapshot_deployable = deployability_index.is_deployable(snapshot)
+
+        if is_first_insert:
```
izeigerman, line 1731:
```
if is_first_insert and not is_snapshot_deployable:
```

izeigerman: Btw, why do we do this only on the first insert? Shouldn't we always update the dev table in dev?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1666,6 +1684,82 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+            if not is_snapshot_deployable:
+                # Snapshot isnt deployable; update the preview table instead
+                # If the snapshot was deployable, then data would have already been loaded in create() because a managed table would have been created
+                logger.info("Updating preview table: %s", table_name)
```
izeigerman, line 1735: Let's clarify that the preview table is related to a managed model specifically.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1666,6 +1684,82 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+    def create(
+        self,
+        table_name: str,
+        model: Model,
+        is_table_deployable: bool,
+        is_snapshot_deployable: bool,
+        **render_kwargs: t.Any,
+    ) -> None:
+        if is_table_deployable and is_snapshot_deployable:
+            # We could deploy this to prod; create a proper managed table
```
izeigerman, line 1697: I've now realized that this should actually work nicely. Even if the snapshot is not deployable in dev, next time the change is applied to prod, SQLMesh should detect a missing table and try to run this logic again, at which point the snapshot should already be deployable.

izeigerman: No, that's actually not true. I think the snapshot will still be non-deployable the first time it's promoted to prod. I think we're still missing a story around creating a managed table when a non-deployable snapshot gets promoted to prod.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1666,6 +1684,82 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+    def migrate(
+        self,
+        target_table_name: str,
+        source_table_name: str,
+        snapshot: Snapshot,
+        **kwargs: t.Any,
+    ) -> None:
+        # Not entirely true, many engines support modifying some of the metadata fields on a managed table
+        # eg Snowflake allows you to ALTER DYNAMIC TABLE foo SET WAREHOUSE=my_other_wh;
+        raise ConfigError("Cannot mutate a managed table")
```
izeigerman, line 1756: Let's add the target table name to the message

sqlmesh/core/snapshot/evaluator.py
erindru: My lack of knowledge around exactly when and how `insert()` is called multiple times.

Managed models are not incremental. The main model would only be created once if it was deployable, so I figured we should only populate the dev preview table once to match?

If the user changes something, isnt that a new snapshot / new table?

sqlmesh/core/snapshot/evaluator.py
erindru: > I think the snapshot will still be non-deployable the first time it's promoted to prod

I tested this manually and it seemed to work (since `is_snapshot_deployable` doesnt come from the doctored DeployabilityIndex that gets passed to `create()`, it comes from the actual DeployabilityIndex).

However, of course a manual test wont verify the behaviour in code or prevent regressions, so i'll see how I can add an automated test for this scenario

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1669,6 +1677,114 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+
+            table_metadata = self.adapter.get_data_objects(
+                table_name_parsed.db, {table_name_parsed.name}
+            )
+            if len(table_metadata) == 0:
+                raise SQLMeshError(
+                    f"Table {table_name} exists but isnt being returned in the metadata. Cannot determine if it's a managed table or not"
+                )
+
+            if table_metadata[0].type != DataObjectType.MANAGED_TABLE:
```
erindru, line 1735: This is where the "forward-only in a virtual environment that might eventually get deployed" edge case is resolved.

I added an integration test `test_managed_model_upstream_forward_only` to exercise this code

sqlmesh/core/snapshot/evaluator.py
erindru: Done

sqlmesh/core/snapshot/evaluator.py
erindru: Done

sqlmesh/core/snapshot/evaluator.py
erindru: There's now an extra case added to `insert()`. If a snapshot that was not deployable at creation (and thus would have been created as a normal table) suddenly becomes deployable - then we "upgrade" it to a managed table by dropping it and recreating it as a managed table

sqlmesh/core/snapshot/evaluator.py
erindru: Done, it's been banished back to `kwargs`

sqlmesh/core/snapshot/evaluator.py
erindru: I guess there is no harm is doing it for all `insert()`'s, i've adjusted the logic

sqlmesh/core/engine_adapter/base.py
```diff
@@ -90,6 +90,7 @@ class EngineAdapter:
     SUPPORTS_MATERIALIZED_VIEWS = False
     SUPPORTS_MATERIALIZED_VIEW_SCHEMA = False
     SUPPORTS_CLONING = False
+    SUPPORTS_MANAGED_MODELS = False
```
erindru, line 93: I'm wondering if this should be `SUPPORTS_MANAGED_TABLES` instead. It's mainly just used to tell pytest to skip the managed model integration tests for adapters that dont support managed models

docs/concepts/models/managed_models.md
```diff
@@ -0,0 +1,107 @@
...
+```sql linenums="1"
+MODEL (
+  name db.events,
+  kind MANAGED,
+  physical_properties (
+    warehouse = datalake,
+    target_lag = '2 minutes',
+    data_retention_time_in_days = 2
+  )
+)
```
georgesittas, line 61:
```suggestion
);
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -428,6 +429,35 @@ def create_table(
...
+        query: Query,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        partitioned_by: t.Optional[t.List[exp.Expression]] = None,
+        clustered_by: t.Optional[t.List[str]] = None,
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """Create an managed table using a query.
```
georgesittas, line 444:
```suggestion
        """Create a managed table using a query.
```

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -99,6 +102,97 @@ def snowpark(self) -> t.Optional[SnowparkSession]:
...
+                "`target_lag` must be specified in the model physical_properties for a Snowflake Dynamic Table"
+            )
+
+        source_queries, columns_to_types = self._get_source_queries_and_columns_to_types(
+            query, columns_to_types, target_table=target_table
+        )
+
+        self._create_table_from_source_queries(
+            quote_identifiers(target_table),
```
georgesittas, line 139: Why are we quoting here? Note that by default `_to_sql` quotes the identifiers in the expression.

tests/core/engine_adapter/test_integration.py
```diff
@@ -2242,3 +2293,199 @@ def _mutate_config(current_gateway_name: str, config: Config):
 
     finally:
         ctx.cleanup(context)
+
+
+def test_managed_model_upstream_forward_only(ctx: TestContext):
```
georgesittas, line 2298: Note: I liked how detailed this test is, thanks for providing context 👍

sqlmesh/core/snapshot/evaluator.py
izeigerman: I'm not sure I understand how this scenario can happen.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1676,6 +1684,116 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+class EngineManagedStrategy(MaterializableStrategy):
+    def create(
+        self,
+        table_name: str,
+        model: Model,
+        is_table_deployable: bool,
+        render_kwargs: t.Dict[str, t.Any],
+        **kwargs: t.Any,
+    ) -> None:
+        is_snapshot_deployable: bool = render_kwargs["is_snapshot_deployable"]
```
izeigerman, line 1696: This should be `kwargs`, right?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1676,6 +1684,116 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+        query_or_df: QueryOrDF,
+        model: Model,
+        is_first_insert: bool,
+        **kwargs: t.Any,
+    ) -> None:
+        deployability_index: DeployabilityIndex = kwargs["deployability_index"]
+        snapshot: Snapshot = kwargs["snapshot"]
+        is_snapshot_deployable = deployability_index.is_deployable(snapshot)
+
+        if is_first_insert and is_snapshot_deployable and self.adapter.table_exists(table_name):
```
izeigerman, line 1733: When do we handle the case when the **deployable** (prod) table doesn't exist?

sqlmesh/core/snapshot/evaluator.py
erindru: Ah, yes, since the merge that broke out `render_kwargs` and `kwargs` into two separate things. It looks like there is a missing test case, i'll add one

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1676,6 +1684,116 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+                table_name=table_name,
+                query=model.render_query_or_raise(**render_kwargs),
+                columns_to_types=model.columns_to_types,
+                partitioned_by=model.partitioned_by,
+                clustered_by=model.clustered_by,
+                table_properties=model.physical_properties,
+                table_description=model.description,
+                column_descriptions=model.column_descriptions,
+            )
+        else:
```
izeigerman, line 1711:
```
else is_snapshot_deployable:
```

sqlmesh/core/engine_adapter/snowflake.py
erindru: I think I saw it used elsewhere in a similar situation. Monkey see, monkey do :)

But you're right, it appears to be unnecessary. I've removed it

sqlmesh/core/snapshot/evaluator.py
erindru: Actually, this was in the `render_kwargs`, but i'll move it

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -652,6 +652,8 @@ def _create_snapshot(
             **common_render_kwargs,
             # Refers to self as non-deployable to successfully create self-referential tables / views.
             deployability_index=deployability_index.with_non_deployable(snapshot),
+            # However, it can still be useful to know if the snapshot was actually deployable
+            is_snapshot_deployable=deployability_index.is_deployable(snapshot),
```
erindru, line 656: The deployability index is already in the render kwargs which is why I put `is_snapshot_deployable` right next to it

erindru: Moved to `kwargs`

sqlmesh/core/snapshot/evaluator.py
erindru: To summarise: it was due to unecessarily creating the "main" table next to the dev table, when we only needed to create the dev table

sqlmesh/core/snapshot/evaluator.py
erindru: The logic has been adjusted to create the managed table if it doesnt exist, rather than upgrading an existing non-managed table that didnt need to be created in the first place

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1676,6 +1687,99 @@ def get_custom_materialization_type(name: str) -> t.Type[CustomMaterialization]:
...
+        query_or_df: QueryOrDF,
+        model: Model,
+        is_first_insert: bool,
+        **kwargs: t.Any,
+    ) -> None:
+        deployability_index: DeployabilityIndex = kwargs["deployability_index"]
+        snapshot: Snapshot = kwargs["snapshot"]
+        is_snapshot_deployable = deployability_index.is_deployable(snapshot)
+
+        if is_first_insert and is_snapshot_deployable and not self.adapter.table_exists(table_name):
```
izeigerman, line 1739: Do we need the `self.adapter.table_exists(table_name)` check at all? I think we don't expect the table here and if it exists, then something went wrong and we should fail, shouldn't we?

sqlmesh/core/model/kind.py
```diff
@@ -755,6 +765,15 @@ def to_expression(
         )
 
 
+class ManagedKind(_ModelKind):
+    name: Literal[ModelKindName.MANAGED] = ModelKindName.MANAGED
+    disable_restatement: SQLGlotBool = True
```
izeigerman, line 770: Let's make it `t.Literal[True] = True` so that the value can be overridden by a user.

izeigerman: Or make it a property method which returns True

sqlmesh/core/snapshot/evaluator.py
erindru: the table might already exist if the snapshot was deployable when `create()` was called

docs/concepts/models/managed_models.md
```diff
@@ -0,0 +1,107 @@
...
+- The engine updates the table data automatically when a base table changes
+- When performing updates, the engine has a semantic understanding of the query and can decide if an incremental or full refresh should be applied
+- There is no need to issue manual `REFRESH` commands. The engine maintains the table transparently in a background process
+
+## Lifecycle in SQLMesh
+Managed models follow the same lifecycle as other models:
+
+- Creating a Virtual Environment creates a pointer to the current model snapshot
+- Modifying the model causes a new snapshot to be created
+- Breaking upstream changes cause a new snapshot to be created
```
izeigerman, line 29: [Nit] Any upstream changes cause a new snapshot to be created

izeigerman (approved):
A few more smaller comments. LGTM otherwise!

sqlmesh/core/snapshot/evaluator.py
izeigerman: Ah, excellent point. Ignore my suggestion then

Issue #546, opened 2023-03-16
Inline model definition audits


crericha:
Allow sql audits to be defined inline within a model definition instead of requiring a separate file

PR #2812, opened 2024-06-24
Feat: Add support for audits within model definition

Themiscodes:
Τhis update introduces the ability to define audits directly within a model definition instead of a separate file, fixes:  #546 

The syntax for defining audits remains the same ([Audits Docs](https://sqlmesh.readthedocs.io/en/stable/concepts/audits/)), while multiple audits can be specified within the same SQL file:

```
MODEL (
    name sushi.items,
    audits(does_not_exceed_threshold(column := id, threshold := 1000), price_is_not_null)
);
SELECT  ...;

AUDIT (name does_not_exceed_threshold);
SELECT * FROM @this_model
WHERE @column >= @threshold;

AUDIT (name price_is_not_null);
SELECT * FROM @this_model
WHERE price IS NULL;
```


sqlmesh/core/model/definition.py
```diff
@@ -113,6 +113,7 @@ class _Model(ModelMeta, frozen=True):
     python_env_: t.Optional[t.Dict[str, Executable]] = Field(default=None, alias="python_env")
     jinja_macros: JinjaMacroRegistry = JinjaMacroRegistry()
     mapping_schema: t.Dict[str, t.Any] = {}
+    inline_audits: t.Optional[t.Dict[str, ModelAudit]] = None
```
tobymao, line 116: should we make this a dict? and then you don't have to check for existence on line 472?

sqlmesh/core/model/definition.py
```diff
@@ -1906,22 +1925,34 @@ def _split_sql_model_statements(
     Raises:
         ConfigError: If the model definition contains more than one SELECT query or `@INSERT_SEED()` call.
     """
+    from sqlmesh.core.audit.definition import load_audit, ModelAudit
+
     query_positions = []
-    for idx, expression in enumerate(expressions):
-        if (
+    inline_audits: UniqueKeyDict[str, ModelAudit] = UniqueKeyDict("inline_audits")
+    idx = 0
```
tobymao, line 1932: why do you need idx manipulation?

Themiscodes: So that when an `Audit` instance is detected at the current index, to load it the corresponding next SQL statement and so that they are removed from the pre or post statement lists.

sqlmesh/core/model/definition.py
Themiscodes: Sure I can try this

tobymao (approved):

Themiscodes:
Addressed comments and adapted logic and tests for snapshot with inline_audits

sqlmesh/core/model/definition.py
```diff
@@ -762,13 +773,21 @@ def metadata_hash(self, audits: t.Dict[str, ModelAudit]) -> str:
...
         for audit_name, audit_args in sorted(self.audits, key=lambda a: a[0]):
             metadata.append(audit_name)
-
+            audit = None
             if audit_name in BUILT_IN_AUDITS:
                 for arg_name, arg_value in audit_args.items():
                     metadata.append(arg_name)
                     metadata.append(gen(arg_value))
+            elif audit_name in self.inline_audits:
+                audit = load_audit(self.inline_audits[audit_name], dialect=self.dialect)
```
georgesittas, line 782: Can we not call `load_audit` twice for inline audits? E.g. can we parse them only once at validation time? Otherwise, consider caching them or something.

sqlmesh/core/model/definition.py
```diff
@@ -762,13 +773,21 @@ def metadata_hash(self, audits: t.Dict[str, ModelAudit]) -> str:
...
+                if not isinstance(audit, ModelAudit):
+                    audit = None
```
georgesittas, line 784: How can this ever _not_ be a `ModelAudit`? Is this something we should be asserting instead? The alternative is `StandaloneAudit`, but I don't _think_ we should ever have to deal with these here.

sqlmesh/core/model/definition.py
georgesittas: This section is a bit complicated, I'd spend some more time thinking how it can be refactored.

georgesittas:
@Themiscodes I think the direction here is good, just need to clean some parts up a bit. Let's also define an inline audit within a model under examples/sushi, so that it's tested end-to-end.

Themiscodes:
Fixed Pydantic issues with @georgesittas in the validator, moved audit loading in the parsing stage, added docs and tests.

sqlmesh/core/model/meta.py
```diff
@@ -359,6 +378,10 @@ def _kind_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
                 raise ValueError(f"partitioned_by field is required for {kind.name} models")
         return values
 
+    @property
+    def inline_audits(self) -> t.Dict[str, t.Any]:
+        return self.inline_audits_
```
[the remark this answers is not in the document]
Themiscodes: Right missed this, removed these!

PR #2813, opened 2024-06-24
Feat: allow marking audits as [non-]blocking at use site

georgesittas:
Until now, audits could be marked as blocking (default) or non-blocking only at their definition. Some users reasonably found this counter-intuitive, since the audit's logic and whether it blocks execution or not are orthogonal concepts.

This PR adds support for marking an audit as blocking or non-blocking at its use-site, i.e. in a model that references it. The syntax to do that is demonstrated below:

```sql
MODEL (
  name model_name,
  kind FULL,
  audits (
    some_audit(blocking := false)
  )
);

...
```


georgesittas:
FYI @plaflamme

izeigerman:
Can we please update documentation

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -831,6 +831,10 @@ def _audit(
                 skipped=True,
             )
 
+        # Model's "blocking" argument takes precedence over the audit's default setting
+        blocking = audit_args.pop("blocking", None)
+        blocking = blocking == exp.true() if blocking else audit.blocking
```
tobymao, line 836: wonder if there's a better way to detect truthy

georgesittas: Yea I was looking for a better way too, but didn't find something.. I thought of overriding `__eq__` for `Boolean` to check for `True` / `False` but didn't like the idea much. Open to suggestions.

georgesittas:
> Can we please update documentation

Done, thanks for pointing it out 👍

sqlmesh/core/snapshot/evaluator.py
izeigerman: Does this mean we can't pass a macro result in here?

izeigerman: Or I guess it will be evaluated during rendering? Can we have a test for this?

georgesittas: Let me take a look, I didn't think much about this use case but it makes sense.

georgesittas: Done - check out https://github.com/TobikoData/sqlmesh/pull/2813/commits/15d6cab0c01d4414d111aac80f4128966fbde3a3

tobymao (approved):

PR #2815, opened 2024-06-24
add ux help

sungchun12:


docs/guides/custom_materializations.md
```diff
@@ -120,6 +120,7 @@ class CustomFullMaterialization(CustomMaterialization):
     ) -> None:
         config_value = model.custom_materialization_properties["config_key"]
         # Proceed with implementing the insertion logic.
+        # Example for existing materialization for look and feel: https://github.com/TobikoData/sqlmesh/blob/273e913847173d20e9eb291df3eeaf998ec679e7/sqlmesh/core/snapshot/evaluator.py#L1227
```
izeigerman, line 123: Should we keep the commit hash in the URL? I'm afraid that this URL may lose its relevance really quickly. Either we update the API and forgot to update this specific comment to point at a new commit hash, or the line numbers drift if we don't include the hash.

sungchun12: updated the link

sungchun12: It's okay if we just point them to the file. Given the intended audience, they will likely search for method names in the github ui after they open this link

tobymao: isn't this a permalink?

Issue #2297, opened 2024-03-18
Support "external signals" for built-in scheduler


plaflamme:
Currently, sqlmesh supports synchronizing with externally scheduled dependencies in the Airflow scheduler via "[external signals](https://sqlmesh.readthedocs.io/en/latest/integrations/airflow/#external-signals)". That is, one can define an "external signal" on a sqlmesh model which allows Airflow to add a node to the DAG to represent a dependency between that model and some external state.

This issue is about extending this capability to the "built-in scheduler" as well: allowing the declaration of external dependencies which are evaluated by the build-in scheduler to determine if a model is "ready" to be evaluated within a `sqlmesh run`.

A public abstraction similar to `BaseSensorOperator` could be defined by sqlmesh, and implemented by sqlmesh users. A "factory method" (equivalent to `create_external_sensor` in the Airflow scheduler) could be provided to the build-in scheduler at config time (i.e.: in `config.py`)

tobymao:
#2822

PR #2822, opened 2024-06-26
feat: signals for the built-in scheduler

plaflamme:
This is just a sketch for #2297. I'll add tests and documentation later on. I'd like feedback on the approach, namely:

1. how should the user provide `signal_factory`?
    1. should they provide a function when building `Context` (like here) or should there be a different mechanism?
1. should waiting occur where it is in this PR or elsewhere? 
1. what about feedback in the console? should there be specific steps output when waiting on signals?
1. should the interface implemented by users be polling or simply blocking (like in the PR)
    1. it's simpler for users to implement a single polling method (i.e.: check the signal once) and the looping is done once
    1. it removes the ability for users to have custom polling / looping (both a good and bad thing)
    1. a poll-based approach makes it easier to give feedback in the console (if necessary)
    1. could also allow a global `poll_interval` and `timeout` configuration (used for every signal)

An alternate interface would be something like this:

```python
class Signal():
  def poll_ready() -> bool:
    """Return true if the signal is ready, false otherwise"
```

And then the looping would be done in the scheduler, e.g.:

```python
poll_timeout = ... # configurable polling timeout
poll_interval = ... # configurable polling interval

for signal in signals:
  start_polling_ts = now()
  while not signal.poll():
    if now() - start_polling_ts > poll_timeout:
      raise TimeoutError
    time.sleep(poll_interval)
```

Ideally, we'd release the thread to go do some other stuff in the meantime, say because there are 2 threads, 3 tasks available, but only 2 of them have a signal; ideally, the 2 signals wouldn't block the third task. It's an optimization, but maybe it's worth it?


izeigerman:
Hey @plaflamme , thanks a lot for submitting this PR and kicking off the conversation.

Part of me wonders whether the builtin SQLMesh scheduler is fundamentally too naive and simplistic to support something like external signals.

Here's one specific example that bothers me. Imagine we have some Model A which has a set of upstream dependencies U and a set of downstream dependencies D. In Airflow, when the model A blocks on external signals, obviously all its downstream dependencies (D) are blocked as well.  At the same time, the upstream set U is free to refresh at its own cadence, unimpacted by the fact that Model A is blocked.

Now, in the case of the built-in scheduler, the blocked Model A will block the entire `sqlmesh run` process until the signal requirement is met or the timeout occurs. Which means that upstream models U (as well as models that have no connection to Model A whatsoever) won't be updated until the signal unblocks, the process exits, and subsequently runs again.

The argument can be made that this sort of issue can be mitigated by setting a reasonable timeout. But what is reasonable? What if the external source is experiencing an outage and is delayed by hour(s)? What if the model's cron doesn't match the source update cadence and the source is consistently late? Should we set the timeout to some low value and just fail the `sqlmesh run` execution during this time? That could result in a subpar experience. Or should the scheduler be smarter and run continuously instead of in a fire-and-forget fashion?

TL;DR I believe significant improvements might be required to the built-in scheduler itself before we can consider supporting external signals for it.

tobymao:
I have another idea that we can use to create "signals" for the built in scheduler. Build in scheduler "signals" will simply be boolean blockers. If they return True, then they process, else they stop. It's up to your cron to retry, let's say every 5 minutes. 

Imagine U -> (S)ignal -> A.

Run, U process, S blocks. Exit.
Run, U process, S Runs. A runs.

It's a very simple mechanism that allows you to choose whether or not the dag continues. And it's simple to maintain because you just do sqlmesh run at your minimum frequency.

izeigerman:
I kind of like the @tobymao's idea. This feels like a natural extension of how we treat cron expressions. 

SQLMesh doesn't block and wait until the designated time to run; it just skips the execution and picks it up again the next time it runs.

plaflamme:
@tobymao That does simplify things. One thing that comes to mind though is that the dependencies aren't between models, but between model **intervals**. So, I believe you can get into a "permanently blocked" situation if we're not blocking.

Imagine `U` -> `E`xternal process -> `S`ignal -> `A`.

Where `E` is some external process that waits for intervals in `U` to show up and produce its own output.
And `S` is configured to look for intervals in `E`, say by using `@start_ds` and `@end_ds` in its definition.

Now imagine a stream of `sqlmesh run`s:

| Run | Intervals to run | Signals to check | Signals Ready | Intervals Ready |
| --- |---|---|---|---|
| 1 | `U1`, `A1` | `S1` | | `U1` |
| 2 | `U2`, `A1-2`| `S1-2` | `S1` | `U2` |
| 3 | `U3`, `A1-3`| `S1-3` | `S1-2` |`U3` |

So because on each `sqlmesh run`, the intervals for `A` "grow" at the same rate as its upstream dependency, the signal can never say it is ready, because it will always depend on the interval being run now.

With the blocking approach, this is resolved eventually and `A` will run at some point (baring errors and timeouts).

Maybe this is an edge case, but it doesn't seem far fetched to want to coordinate sqlmesh with external systems like this. Especially during a migration or similar.

Presumably, the signal could return a list of intervals that are ready instead of just a boolean value, but that seems pretty challenging to integrate in practice. Or maybe the scheduler can be smart enough to try different batches somehow.

Another thing that comes to mind is that maybe the user can solve this using `lookback`. That is, if `A` has `lookback 1` then perhaps its `S`ignal can also ignore the absence of the latest interval since it knows it will run it again later. Perhaps this can be provided as a workaround for users with this situation.

FWIW: I don't think this situation applies to us, so I'd be happy to move forward regardless.

sqlmesh/core/scheduler.py
```diff
@@ -431,6 +438,37 @@ def compute_interval_params(
         ignore_cron=ignore_cron,
         end_bounded=end_bounded,
     ).items():
+
+        def intersect(left: SnapshotInterval, intervals: list[Interval]) -> SnapshotInterval | None:
+            """Returns the intersection of `left` against `intervals`. Returns None if it does not intersect. Assumes `intervals` are disjoint."""
```
tobymao, line 443: isn't a SnapshotInterval an Interval, why are the types different?

tobymao: list[Interval] is a Batch

tobymao: can we define this lambda outside of the for loop, the nestedness is a bit too much

sqlmesh/core/signal.py
```diff
@@ -0,0 +1,14 @@
...
+import typing as t
+from datetime import datetime
+
+Interval = t.Tuple[datetime, datetime]
+
+
+class Signal(abc.ABC):
+    @abc.abstractmethod
+    def poll_ready_intervals(self) -> bool | list[Interval]:
+        """Poll the signal and return True if it is ready, otherwise return False."""
```
tobymao, line 11: i think we should make it clear that the intervals here need to be aligned with the interval unit, so there's no guessing games on what these values should be

sqlmesh/core/scheduler.py
```diff
@@ -431,6 +438,37 @@ def compute_interval_params(
...
+
+            left_lower, left_upper = left
+            for right in intervals:
+                right_lower, right_upper = to_timestamp(right[0]), to_timestamp(right[1])
```
tobymao, line 447: how exactly is the api supposed to work for this? i think users should probably be expected to return discrete intervals that align to the interval unit.

sqlmesh/core/signal.py
tobymao: nit can we rename this to 

get_intervals 

poll i feel indicates that it will happen many times, but with sqlmesh's way of doing things, it's usually one and done

sqlmesh/core/scheduler.py
```diff
@@ -431,6 +438,37 @@ def compute_interval_params(
...
+            for signal in snapshot.model.render_signals(
+                start=start, end=end, execution_time=execution_time
+            ):
+                ready_intervals = signal_factory(signal).poll_ready_intervals()
+                if isinstance(ready_intervals, bool):
+                    if not ready_intervals:
+                        intervals = []
+                        break
+                elif isinstance(ready_intervals, list):
+                    intervals = [
```
tobymao, line 464: can this just be set to the intervals returned by the function?

plaflamme: This was addressed in the new interface in https://github.com/TobikoData/sqlmesh/pull/2822/commits/e9d9119249d07679bc4c04c913d4563ffdf8aef7

sqlmesh/core/scheduler.py
plaflamme: This was addressed in the new interface in https://github.com/TobikoData/sqlmesh/pull/2822/commits/e9d9119249d07679bc4c04c913d4563ffdf8aef7

sqlmesh/core/signal.py
plaflamme: Addressed in https://github.com/TobikoData/sqlmesh/pull/2822/commits/e9d9119249d07679bc4c04c913d4563ffdf8aef7

sqlmesh/core/scheduler.py
plaflamme: > isn't a SnapshotInterval an Interval, why are the types different?

From what I can tell, `missing_intervals` (defined [here](https://github.com/TobikoData/sqlmesh/blob/910261867a7cf5c8f9fbb3b2f4349f37dd348d2f/sqlmesh/core/snapshot/definition.py#L1572-L1581)) returns `sqlmesh.core.snapshot.Intervals` which is defined [here](https://github.com/TobikoData/sqlmesh/blob/910261867a7cf5c8f9fbb3b2f4349f37dd348d2f/sqlmesh/core/snapshot/definition.py#L50-L51) as `list[t.Tuple[int,int]]`. The built-in scheduler has its own `Interval` defined [here](https://github.com/TobikoData/sqlmesh/blob/910261867a7cf5c8f9fbb3b2f4349f37dd348d2f/sqlmesh/core/scheduler.py#L38-L39).

So `Interval = (datetime, datetime)` and `SnapshotInterval = (int, int)`.

Please see https://github.com/TobikoData/sqlmesh/pull/2822/commits/e9d9119249d07679bc4c04c913d4563ffdf8aef7 where we convert back and forth. Doesn't seem great, but it's unclear to me what an alternative might be.

sqlmesh/core/signal.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+import abc
+import typing as t
+from datetime import datetime
+
+Interval = t.Tuple[datetime, datetime]
```
tobymao, line 7: move this type typing file?

sqlmesh/core/signal.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+import abc
+import typing as t
+from sqlmesh.core._typing import Batch
```
izeigerman, line 5: Is this something users suppose to import? In this case let's probably export this in the `sqlmesh/__init__.py`. Also let's put this under `if t.TYPE_CHECKING`

sqlmesh/core/scheduler.py
```diff
@@ -431,6 +436,27 @@ def compute_interval_params(
         ignore_cron=ignore_cron,
         end_bounded=end_bounded,
     ).items():
+        if signal_factory:
+            batch = [(to_datetime(start), to_datetime(end)) for start, end in intervals]
+            for signal in snapshot.model.render_signals(
+                start=start, end=end, execution_time=execution_time
+            ):
+                ready_intervals = signal_factory(signal).get_intervals(batch=batch)
+                if isinstance(ready_intervals, bool):
```
izeigerman, line 445:
```
if isinstance(ready_intervals, bool) and not ready_intervals:
    ...
```

sqlmesh/core/signal.py
plaflamme: I believe users would need `Signal` and `Batch` (for its method signature). These are all concerns of the `scheduler` (where `Batch` used to be defined). I've created a `signal.py` file, but that seems unnecessary and it's causing some `import` pains.

Any objection to moving `Signal` and `SignalFactory` from `signal.py` into `scheduler.py`? Then we can re-export whatever is necessary from `sqlmesh/__init__.py` to simplify users' life.

plaflamme: Took the liberty to move things around in https://github.com/TobikoData/sqlmesh/pull/2822/commits/f1aa0455f7442d974076f9380e5906814e48a2b2 PTAL

tobymao (approved):
this looks good to me, thanks so much for doing this @plaflamme

plaflamme:
@tobymao all checks are passing please let me know if this needs anything else

achicoine-coveo:
This is very interesting. Does it get used the same as the airflow signals? I would be interested in a very basic example on how to register a check_interval from a python function in a model.

plaflamme:
@achicoine-coveo More work needs to be done to make it usable through `sqlmesh run`. But in the meantime, it would look like this.

Create a python package, say `my_package` and define a class that extends `sqlmesh.core.scheduler.Signal` that you want to execute as a signal, for example checking if a file exists somewhere in S3:

```python
from datetime import date
from sqlmesh.core.scheduler import Batch, Signal

class MyS3Signal(Signal):
  def check_file_exists(self, dt: date) -> bool:
    """Checks if a file exists for the specified date"""
    pass
  
  # concrete implementation of abstraction method from Signal
  def check_intervals(self, batch: Batch) -> bool | Batch:
    """"Filter the batch to only return the intervals for which the file exists"""
    return [(start, end) for (start, end) in batch if check_file_exists(start.sate())]
```

Next, define a function that converts the `signal` metadata from models to a concrete `Signal` implementation:

```python
# see `sqlmesh.core.scheduler.SignalFactory` for the expected function signature
def signal_factory(signal_meta: dict[str, str | int | float | bool]) -> Signal:
   if signal_meta["kind"] == "MyS3Signal":
    return MyS3Signal()
  throw ValueError("unknown signal")
```

Lastly, provide your `signal_factory` to sqlmesh (this is the part that needs to be improved):

```python
def run():
  config: Config = ... # load sqlmesh config
  
  # create a sqlmesh context
  context = Context(configs=config, signal_factory=my_package.signal_factory)

  # equivalent to `sqlmesh run` cli without any parameters
  context.run()
```

From that point, you can use this signal in a model:

```sql
MODEL(
  name foo.bar,
  signals [
    (kind="MyS3Signal", arg_1=TRUE, , arg_2=42)
  ]
);
SELECT ...
```

Hope this helps, bonne chance!

PR #2833, opened 2024-06-27
Feat: Introduce metadata only flag for macro functions

vchan:
This PR allows flagging macro functions in a SQL model's pre and post hooks to be included in the model's metadata hash instead of its data hash.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -962,11 +962,13 @@ class _SqlBasedModel(_Model):
 
         for statement in (*self.pre_statements, *self.post_statements):
             statement_exprs: t.List[exp.Expression] = []
-            if isinstance(statement, d.MacroFunc) and not self._is_metadata_statement(statement):
+            if isinstance(statement, d.MacroDef):
                 statement_exprs = [statement]
             else:
                 rendered = self._statement_renderer(statement).render()
-                if rendered is not None:
+                if self._is_metadata_statement(statement):
+                    continue
+                if rendered:
                     statement_exprs = rendered
                 else:
                     statement_exprs = [statement]
@@ -989,9 +991,11 @@ class _SqlBasedModel(_Model):
             return False
         if isinstance(statement, d.MacroFunc):
             target_macro = macro.get_registry().get(statement.name)
-            if not target_macro:
-                target_macro = self.python_env.get(statement.name)
-            return getattr(target_macro, c.SQLMESH_METADATA, False)
+            if target_macro:
+                return getattr(target_macro.func, c.SQLMESH_METADATA, False)
+            target_macro = self.python_env.get(statement.name)
+            if isinstance(target_macro, Executable):
+                return target_macro.is_metadata is True
         return False
 
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -962,10 +962,8 @@ class _SqlBasedModel(_Model):
 
         for statement in (*self.pre_statements, *self.post_statements):
             statement_exprs: t.List[exp.Expression] = []
-            if isinstance(statement, d.MacroFunc):
-                target_macro = macro.get_registry().get(statement.name)
-                if not getattr(target_macro, c.SQLMESH_METADATA, False):
-                    statement_exprs = [statement]
+            if isinstance(statement, d.MacroFunc) and not self._is_metadata_statement(statement):
+                statement_exprs = [statement]
             else:
                 rendered = self._statement_renderer(statement).render()
                 if rendered is not None:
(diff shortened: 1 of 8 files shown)


tobymao:
why does it need to be a flag? why isn’t it always metadata?

tests/core/test_model.py
```diff
@@ -5096,3 +5096,40 @@ def test_model_kind_to_expression():
...
+        SELECT 1;
+
+        @noop();
+    """
+    )
+    new_model = load_sql_based_model(
+        expressions, path=Path("./examples/sushi/models/test_model.sql")
+    )
+
+    assert model.metadata_hash(audits={}) == new_model.metadata_hash(audits={})
```
crericha, line 5132: Can you also check datahash here and after the other assert?

crericha (approved):

vchan:
> why does it need to be a flag? why isn’t it always metadata?

@tobymao Good question. I had a chat with @izeigerman and we felt it'd be safer to use a flag. A macro func could do something to warrant a backfill even though it doesn't render any expressions to execute. Happy to discuss this more though. I'm not that familiar with the macros system.

tobymao:
> > why does it need to be a flag? why isn’t it always metadata?
> 
> @tobymao Good question. I had a chat with @izeigerman and we felt it'd be safer to use a flag. A macro func could do something to warrant a backfill even though it doesn't render any expressions to execute. Happy to discuss this more though. I'm not that familiar with the macros system.

ok that's fine with me, but we should edit the docs to indicate this exists

sqlmesh/core/model/definition.py
```diff
@@ -934,6 +936,18 @@ def _data_hash_values(self) -> t.List[str]:
 
         return data_hash_values
 
+    @property
+    def _additional_metadata(self) -> t.List[str]:
+        additional_metadata = super()._additional_metadata
+
+        for statement in (*self.pre_statements, *self.post_statements):
+            if isinstance(statement, d.MacroFunc) and not isinstance(statement, d.MacroDef):
+                target_macro = macro.get_registry().get(statement.name)
```
izeigerman, line 945: 944-946 look exactly like above. Can we have a function for this?
```
def _is_metadata_statement(statement: exp.Expression) -> bool:
    if isinstance(statement, d.MacroDef):
        return True
    if isinstance(statement, d.MacroFunc):
        ...
    return False

Comment (izeigerman) [reply sqlmesh/core/model/definition.py:945]: Also `macro.get_registry()` might not always be populated correctly (eg. Airflow). Should we use `self.python_env` instead?

Comment (izeigerman) [reply sqlmesh/core/model/definition.py:945]: you might need to extend `Executable` with the metadata flag however.

**Comment by vchan** [comment]
> > > why does it need to be a flag? why isn’t it always metadata?
> > 
> > 
> > @tobymao Good question. I had a chat with @izeigerman and we felt it'd be safer to use a flag. A macro func could do something to warrant a backfill even though it doesn't render any expressions to execute. Happy to discuss this more though. I'm not that familiar with the macros system.
> 
> ok that's fine with me, but we should edit the docs to indicate this exists
> 
> so the default is a data hash or metadata hash? default should be metadata right?

Okay, I'll update the docs. As for the default, it's currently data hash since it's safer (will trigger a backfill for your plan vs wait until next run). What do you think?

File: sqlmesh/core/model/definition.py [hunk 3]
```diff
@@ -966,14 +966,38 @@ def _data_hash_values(self) -> t.List[str]:
                 statement_exprs = [statement]
             else:
                 rendered = self._statement_renderer(statement).render()
-                if rendered is not None:
+                if self._is_metadata_statement(statement):
+                    continue
+                if rendered:
```
vchan, line 971: For macros that return None, `rendered` is an empty list and will be excluded from the data hash.

tobymao: that’s fine, the args to the func are still included right

vchan: Yes, statement includes the args to the func.

sqlmesh/core/model/definition.py
```diff
@@ -966,14 +966,38 @@ def _data_hash_values(self) -> t.List[str]:
...
+
+        for statement in (*self.pre_statements, *self.post_statements):
+            if self._is_metadata_statement(statement):
+                additional_metadata.append(gen(statement))
+
+        return additional_metadata
+
+    def _is_metadata_statement(self, statement: exp.Expression) -> bool:
+        if isinstance(statement, d.MacroDef):
+            return False
```
tobymao, line 991: should macro def always be metadata, there's no side effects

tests/core/test_macros.py
```diff
@@ -599,3 +600,19 @@ def test_macro_parameter_resolution(macro_evaluator):
...
+    @macro()
+    def noop(evaluator) -> None:
+        return None
+
+    @macro(is_metadata=True)
+    def noop_metadata_only(evaluator) -> None:
+        return None
+
+    assert not hasattr(noop, c.SQLMESH_METADATA)
+    assert macro.get_registry()["noop"].is_metadata is False
```
tobymao, line 615: assert not is_metadata

sqlmesh/core/model/definition.py
izeigerman: yeah, I think we can do that

sqlmesh/utils/metaprogramming.py
```diff
@@ -360,6 +361,7 @@ def serialize_env(env: t.Dict[str, t.Any], path: Path) -> t.Dict[str, Executable
                     # Do `as_posix` to serialize windows path back to POSIX
                     path=t.cast(Path, file_path).relative_to(path.absolute()).as_posix(),
                     alias=k if name != k else None,
+                    is_metadata=getattr(v, c.SQLMESH_METADATA, None),
```
izeigerman, line 364: Can the default be `False` instead of `None`?

izeigerman: I guess `is_metadata` would still have to be `Optional` so that the existing state can be deserialized. So doesn't matter.

izeigerman (approved):

PR #2836, opened 2024-06-28
Revert "Fix: redshift leverages postgres adapter dispatch implementations for common functionality"

izeigerman:
Reverts TobikoData/sqlmesh#2770


izeigerman:
Looks like our assumption was wrong. If the package name is not specified in the `adapter.dispatch` call then the default value is not `dbt` but the name of the package from which `adapter.dispatch` has been called.

Issue #2810, opened 2024-06-24
BUG: data diffing with SQLMesh does NOT work correctly if primary keys match with duplicates


sungchun12:
<img width="909" alt="image" src="https://github.com/TobikoData/sqlmesh/assets/19176976/fdbf5592-9ea2-4f98-aa02-b48934f56257">

The above is an example tablediff using the default sqlmesh projected generated with the CLI and using duckdb. The expected result should error out and say that the primary key (`grain`) is not unique and not null and prompt the user to update the `grain` or fix the primary key issue. 

Steps to reproduce:
- `sqlmesh init duckdb`
- `sqlmesh plan` and apply the changes
- make changes to `full_model` below
- `sqlmesh plan dev `and apply the changes
- `sqlmesh table_diff prod:dev sqlmesh_example.full_model --show-sample`  

```sql
MODEL (
  name sqlmesh_example.full_model,
  kind FULL,
  cron '@daily',
  grain item_id,
  audits (assert_positive_order_ids),
);

SELECT
  item_id,
  count(distinct id) AS num_orders,
FROM
    sqlmesh_example.incremental_model
GROUP BY item_id
union all
SELECT
  item_id,
  count(distinct id) AS num_orders,
FROM
    sqlmesh_example.incremental_model
GROUP BY item_id
```

tobymao:
this is user error. grain is not validated and up to the user to ensure interegrity through an audit.

sungchun12:
That's a non-obvious thing for a user trying this for the first time or the 50th time. We should reduce that friction with a clear message that for accurate results, the grain should be tested. Example below.

```shell
(venv) ➜  sqlmesh-demos git:(migration-demo) ✗ sqlmesh table_diff prod:dev demo.full_model --show-sample
grain should have unique and not_null audits for accurate results

Schema Diff Between 'PROD' and 'DEV' environments for model 'demo.full_model':
└── Schemas match


Row Counts:
└──  FULL MATCH: 6 rows (100.0%)

COMMON ROWS column comparison stats:
            pct_match
num_orders      100.0


COMMON ROWS sample data differences:
  All joined rows match

```

tobymao:
maybe a count distinct could be shoved into the data diff query, if you or perhaps @Themiscodes has time you could make a pr for this.

basically i'm looking for a way to put in some kind of uniqueness check into the singular data diff query, it could be an optional flag that is injected by default but turned off due to overhead

sungchun12:
[...]

PR #2840, opened 2024-06-28
Fix: Check for null or duplicate primary keys when running table_diff

Themiscodes:
This adds a check for the uniqueness of the primary key (grain) or whether it contains null values and accordingly shows a warning to the user, fixes: #2810 

By default, this behavior is turned off due to the overhead of the additional query, but it can be enabled with the `check_grain` flag.

```
sqlmesh table_diff prod:dev sqlmesh_example.full_model --show-sample --check-grain

Grain should have unique and not-null audits for accurate results.

Row Counts:
└──  FULL MATCH: 3 rows (100.0%)

            pct_match
num_orders      100.0

COMMON ROWS sample data differences:
  All joined rows match
```

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -743,6 +743,11 @@ def create_external_models(obj: Context) -> None:
     default=3,
     help="The number of decimal places to keep when comparing floating point columns. Default: 3",
 )
+@click.option(
+    "--check-grain",
+    is_flag=True,
+    help="An optional check if a primary key (grain) is missing or is not unique.",
+)
 @click.pass_obj
 @error_handler
 @cli_analytics
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -744,9 +744,9 @@ def create_external_models(obj: Context) -> None:
     help="The number of decimal places to keep when comparing floating point columns. Default: 3",
 )
 @click.option(
-    "--check-grain",
+    "--skip-grain-check",
     is_flag=True,
-    help="An optional check if a primary key (grain) is missing or is not unique.",
+    help="Disable the check for a primary key (grain) that is missing or is not unique.",
 )
 @click.pass_obj
 @error_handler
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -255,7 +255,7 @@ class Console(abc.ABC):
 
     @abc.abstractmethod
     def show_row_diff(
-        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = True
+        self, row_diff: RowDiff, show_sample: bool = True, skip_grain_check: bool = False
     ) -> None:
         """Show table summary diff."""
 
@@ -1068,7 +1068,7 @@ class TerminalConsole(Console):
         self.console.print(tree)
 
     def show_row_diff(
-        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = True
+        self, row_diff: RowDiff, show_sample: bool = True, skip_grain_check: bool = False
     ) -> None:
         source_name = row_diff.source
         if row_diff.source_alias:
@@ -1078,7 +1078,7 @@ class TerminalConsole(Console):
             target_name = row_diff.target_alias.upper()
 
         if row_diff.stats["null_grain_count"] > 0 or (
-            check_grain
+            not skip_grain_check
             and (
                 row_diff.stats["distinct_count_s"] != row_diff.stats["s_count"]
                 or row_diff.stats["distinct_count_t"] != row_diff.stats["t_count"]
@@ -2048,7 +2048,7 @@ class DebuggerTerminalConsole(TerminalConsole):
         self._write(schema_diff)
 
     def show_row_diff(
-        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = True
+        self, row_diff: RowDiff, show_sample: bool = True, skip_grain_check: bool = False
     ) -> None:
         self._write(row_diff)
 
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -255,7 +255,7 @@ class Console(abc.ABC):
 
     @abc.abstractmethod
     def show_row_diff(
-        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = False
+        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = True
     ) -> None:
         """Show table summary diff."""
 
@@ -982,20 +982,6 @@ class TerminalConsole(Console):
         ):
             plan_builder.apply()
 
-    def _check_grain_uniqueness(self, stats: t.Dict[str, float], check_grain: bool = False) -> bool:
-        if stats["null_grain_count"] > 0:
-            return True
-
-        if check_grain:
-            join_count = stats["join_count"]
-            return all(
-                count != join_count
-                for key, count in stats.items()
-                if key.startswith("distinct_count_")
-            )
-
-        return False
-
     def log_test_results(
         self, result: unittest.result.TestResult, output: str, target_dialect: str
     ) -> None:
@@ -1082,7 +1068,7 @@ class TerminalConsole(Console):
         self.console.print(tree)
 
     def show_row_diff(
-        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = False
+        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = True
     ) -> None:
         source_name = row_diff.source
         if row_diff.source_alias:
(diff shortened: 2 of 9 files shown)


sqlmesh/core/table_diff.py
```diff
@@ -277,6 +277,20 @@ def name(e: exp.Expression) -> str:
                         1,
                         0,
                     ).as_("row_joined"),
+                    exp.func(
+                        "IF",
+                        exp.and_(
```
tobymao, line 282: why is this an and?

shouldn't it be a null grain if any part of the key is null?

sqlmesh/core/table_diff.py
```diff
@@ -322,6 +337,47 @@ def name(e: exp.Expression) -> str:
                 stats_df["t_only_count"] = stats_df["t_count"] - stats_df["join_count"]
                 stats = stats_df.iloc[0].to_dict()
 
+                if check_grain:
```
tobymao, line 340: can this be included in the main query without having to do another query?

sqlmesh/core/console.py
```diff
@@ -1024,14 +1026,24 @@ def show_schema_diff(self, schema_diff: SchemaDiff) -> None:
...
         target_name = row_diff.target
         if row_diff.target_alias:
             target_name = row_diff.target_alias.upper()
 
+        if check_grain and (
+            row_diff.stats["null_grain_count"] > 0
+            or row_diff.stats["join_count"] != row_diff.stats["distinct_grain_count"]
+        ):
+            self.console.print(
+                "[b][red]Grain should have unique and not-null audits for accurate results.[/red][/b]"
```
tobymao, line 1044: the null grains is cheap, there's no reason why that has to be covered by the flag

sqlmesh/core/table_diff.py
tobymao: can't we just do a count distinct?

Themiscodes: Sure, I wanted to keep it in a separate query, to control it by the flag, but I will try to add it in the main query

sqlmesh/core/console.py
Themiscodes: Right, I will have this check then always on by default

sqlmesh/core/table_diff.py
Themiscodes: Right, that's a mistake I'll remove the `and_`

sqlmesh/cli/main.py
```diff
@@ -725,6 +725,11 @@ def create_external_models(obj: Context) -> None:
     default=3,
     help="The number of decimal places to keep when comparing floating point columns. Default: 3",
 )
+@click.option(
+    "--check-grain",
+    is_flag=True,
+    help="An optional check if a primary key (grain) is missing or is not unique.",
+)
```
georgesittas, line 732: Let's mirror this for magics too.

sqlmesh/core/table_diff.py
georgesittas: > Sure, I wanted to keep it in a separate query, to control it by the flag

You can still control it using the flag and append to the existing query instead of making a new one, e.g. either by mutating it or just calling `select` on it and just overwriting the local variable.

georgesittas (commented):
Let's test this & update documentation (CLI, magics reference pages).

sqlmesh/core/table_diff.py
Themiscodes: Yes this is how I currently have it so that is in the main query, I will push this implementation, when I address the rest of your suggestions as well

georgesittas:
> Thank you @georgesittas addressed your comments and suggestions

Thanks 👍, I'll take another look soon.

sqlmesh/core/console.py
```diff
@@ -1024,14 +1040,21 @@ def show_schema_diff(self, schema_diff: SchemaDiff) -> None:
 
         self.console.print(tree)
 
-    def show_row_diff(self, row_diff: RowDiff, show_sample: bool = True) -> None:
+    def show_row_diff(
+        self, row_diff: RowDiff, show_sample: bool = True, check_grain: bool = False
```
tobymao, line 1044: default should be true

sqlmesh/core/table_diff.py
```diff
@@ -277,6 +277,20 @@ def name(e: exp.Expression) -> str:
                         1,
                         0,
                     ).as_("row_joined"),
+                    exp.func(
+                        "IF",
+                        exp.or_(
+                            *(
+                                exp.and_(
```
tobymao, line 284: will s and t ever both be NULL? is this correct? have you tested it

sqlmesh/core/table_diff.py
```diff
@@ -309,13 +323,33 @@ def name(e: exp.Expression) -> str:
...
+                    exp.func("SUM", "null_grain").as_("null_grain_count"),
                     exp.func("SUM", "row_full_match").as_("full_match_count"),
                     *(exp.func("SUM", name(c)).as_(c.alias) for c in comparisons),
-                ).from_(table)
+                ]
+
+                if check_grain:
+                    distincts = [
+                        *(
+                            exp.func("COUNT", exp.func("DISTINCT", f"s__{c}")).as_(
```
tobymao, line 338: i don't think this is proper ast, distinct is not a function

sqlmesh/core/table_diff.py
```diff
@@ -309,13 +323,33 @@ def name(e: exp.Expression) -> str:
...
+                                "distinct_count_s"
+                            )
+                            for c in index_cols
```
tobymao, line 341: why are you checking each distinct column, shouldn't it be the whole key? each column may not be unique on it's own, it's the combination of the index_cols that creates the unique key right?

tests/core/test_table_diff.py
```diff
@@ -134,3 +135,111 @@ def test_data_diff_decimals(sushi_context_fixed_date):
...
+
+@pytest.mark.slow
+def test_distinct_grain_check(sushi_context_fixed_date):
+    engine_adapter = sushi_context_fixed_date.engine_adapter
+
+    engine_adapter.ctas(
+        "table_diff_source",
+        pd.DataFrame(
+            {
+                "key": [1, 2, 3, 3, 2],
```
tobymao, line 148: your tests needs to have multiple keys and Null values

sqlmesh/core/table_diff.py
```diff
@@ -205,7 +205,7 @@ def schema_diff(self) -> SchemaDiff:
             model_name=self.model_name,
         )
 
-    def row_diff(self) -> RowDiff:
+    def row_diff(self, check_grain: bool = False) -> RowDiff:
```
tobymao, line 208: check_grains should be default True

sqlmesh/core/table_diff.py
Themiscodes: Yes I tested it and since this query is performed on the resulting table of the full join, a null primary key in table 's' will have a null values for the primary key as well in table 't' in the same row:

```
item_id │ num_orders │ item_id │ num_orders │
────────│────────────│─────────│────────────
   int32 │     int64  │   int32 │     int64
────────│────────────│─────────│────────────
      1 │         5  │       1 │         5
      2 │         1  │       2 │         1
        │            │         │         3
        │            │         │         2
```

So I couldn't think of another way to construct this case to include it in the main query for the fully joined table. Except if the check for null values was with a separate query on the individual tables themselves

tests/core/test_table_diff.py
Themiscodes: Sure I will enhance these tests and add more

sqlmesh/core/table_diff.py
Themiscodes: Yes that's my mistake, I adapted this now to check for distinct of the whole key instead

Themiscodes:
Addressed comments, refactored distinct query, condensed tests to one with cases for multiple column keys and null values

sqlmesh/core/table_diff.py
tobymao: hey @Themiscodes can you explain this condition to me? you're trying to check that any of the index cols are null right. why are you using an and condition?

should it be if( s_a is null or s_b is null or t_a is null or t_b is null) ?

Themiscodes: The issue with the condition `s_a is null or s_b is null or t_a is null or t_b` is it would evaluate to `True` in cases where there are no actual nulls in the original tables, since the query is applied to the `fully joined` table which fills nulls in the rows that are not matching.

If for example we had the tables:

**Prod:**
```
┌─────────┬────────────┐
│ item_id │ num_orders │
│  int32  │   int64    │
├─────────┼────────────┤
│       1 │          5 │
│       3 │          1 │
│       2 │          1 │
└─────────┴────────────┘
```

**Dev:**
```
┌─────────┬────────────┐
│ item_id │ num_orders │
├─────────┼────────────┤
│       3 │          1 │
│       1 │          5 │
│       2 │          1 │
│       4 │          6 │
└─────────┴────────────┘
```

The resulting full join would be:
```
┌───────────┬──────────────┬───────────┬──────────────┐
│ s__item_id│ s__num_orders│ t__item_id│ t__num_orders│
├───────────┼──────────────┼───────────┼──────────────┤
│        1.0│           5.0 │         1 │            5 │
│        3.0│           1.0 │         3 │            1 │
│        2.0│           1.0 │         2 │            1 │
│        NaN│           NaN │         4 │            6 │
└───────────┴──────────────┴───────────┴──────────────┘
```

So having the `and` in the condition prevents false positives like this. And since if a key is null in one table, it will also be null in the same row of the full join table, actual null keys are properly identified

tobymao (approved):

PR #2841, opened 2024-06-28
Fix: Improve the dbt adapter dispatch resolution when the package name is not specified

izeigerman:
When the package name is not specified the root package must be searched and not just the `dbt*` ones.

### DIFF
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -97,15 +97,27 @@ class BaseAdapter(abc.ABC):
 
     def dispatch(self, name: str, package: t.Optional[str] = None) -> t.Callable:
         """Returns a dialect-specific version of a macro with the given name."""
+        package = package or "dbt"
         target_type = self.jinja_globals["target"]["type"]
-        references_to_try = [
-            MacroReference(package=f"{package}_{target_type}", name=f"{target_type}__{name}"),
-            MacroReference(package=package, name=f"{target_type}__{name}"),
-            MacroReference(package=package, name=f"default__{name}"),
-        ]
+        macro_suffix = f"__{name}"
 
-        for reference in references_to_try:
-            macro_callable = self.jinja_macros.build_macro(reference, **self.jinja_globals)
+        def _relevance(macro: MacroReference) -> int:
+            """Lower scores more relevant."""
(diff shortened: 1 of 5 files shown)


sqlmesh/dbt/adapter.py
```diff
@@ -98,19 +98,49 @@ def quote(self, identifier: str) -> str:
...
+        candidates = {}
+        for macro_package in packages_to_check:
+            macros: t.Dict[str, t.Any] = (
+                jinja_env.get(macro_package, {}) if macro_package else jinja_env  # type: ignore
+            )
+            matching_macro_names = [k for k in macros if k.endswith(macro_suffix)]
+            candidates.update(
+                {
+                    (macro_package, macro_name): macros[macro_name]
+                    for macro_name in matching_macro_names
```
crericha, line 131: Is it possible to do the matching macro names work here instead of its own loop? 

```
(macro_package, macro_name): macros[macro_name]
for macro_name in macros
if macro_name.endswith(macro_suffix)
```

sqlmesh/dbt/adapter.py
```diff
@@ -98,19 +98,49 @@ def quote(self, identifier: str) -> str:
...
+                }
+            )
 
-        for reference in references_to_try:
-            macro_callable = self.jinja_macros.build_macro(reference, **self.jinja_globals)
-            if macro_callable is not None:
-                return macro_callable
+        if candidates:
+            sorted_candidates = sorted(candidates, key=_relevance)
+            return candidates[sorted_candidates[0]]
```
crericha, line 137: Can this be `return sorted(candidates, key=_relevance)[0]`?

sqlmesh/dbt/adapter.py
izeigerman: ah, good point

crericha (approved):

sqlmesh/dbt/adapter.py
izeigerman: What do you mean? We need to return the callable and not just the name

sqlmesh/dbt/adapter.py
izeigerman: Fixed this

PR #2844, opened 2024-07-01
Feat: Allow the janitor to be run on-demand

erindru:
This exposes the janitor task to the CLI so that users may run it on-demand and not as part of `sqlmesh run`.

To trigger the janitor, just run `sqlmesh janitor` and it calls the janitor task in the exact same way that `sqlmesh run` would have.

The main benefit is being able to decouple the janitor process and the missing intervals process so they dont interfere with each other.

I've noticed that they can interfere with each other when there are many large tables to drop, particularly on systems like Trino / Iceberg where a `DROP TABLE` statement blocks until all the files have been removed from S3. This causes the main run to be blocked until the janitor is complete, which is undesirable when you want it to refresh data ASAP.

I also added an option `--ignore-ttl` to ignore the snapshot TTL when identifying snapshots to remove. Iterating heavily on a SQLMesh project leaves lots of unused tables lying around. If you want to clean them up now - you're stuck. You have no way to do it without either waiting for the TTL to expire or manually modifying the state database to adjust the TTL to be in the past.

This option gives you a way to do it. I understand there is some concern around race conditions, so it throws up a warning and requires the user to manually indicate they want to proceed before continuing. The idea isn't to use it all the time, just when it makes sense to perform an early cleanup.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -30,6 +30,7 @@ Commands:
   info                    Print information about a SQLMesh project.
   init                    Create a new SQLMesh repository.
   invalidate              Invalidate the target environment, forcing its...
+  janitor                 Run the janitor task on-demand.
   migrate                 Migrate SQLMesh to the current running version.
   plan                    Apply local changes to the target environment.
   prompt                  Uses LLM to generate a SQL query from a prompt.
@@ -227,6 +228,21 @@ Options:
               janitor process. This option requires a connection to the data
               warehouse.
   --help      Show this message and exit.
+```
+
+## janitor
+
+```
+Usage: sqlmesh janitor [OPTIONS]
+
+  Run the janitor task on-demand.
+
+  The janitor cleans up old environments and expired snapshots.
+
+Options:
+  --ignore-ttl  Cleanup snapshots that are not referenced in any environment,
+                regardless of when they're set to expire
+  --help        Show this message and exit.
 ```
 
 ## migrate
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -30,7 +30,7 @@ Commands:
   info                    Print information about a SQLMesh project.
   init                    Create a new SQLMesh repository.
   invalidate              Invalidate the target environment, forcing its...
-  janitor                 Run the janitor task on-demand.
+  janitor                 Run the janitor process on-demand.
   migrate                 Migrate SQLMesh to the current running version.
   plan                    Apply local changes to the target environment.
   prompt                  Uses LLM to generate a SQL query from a prompt.
@@ -235,7 +235,7 @@ Options:
 ```
 Usage: sqlmesh janitor [OPTIONS]
 
-  Run the janitor task on-demand.
+  Run the janitor process on-demand.
 
   The janitor cleans up old environments and expired snapshots.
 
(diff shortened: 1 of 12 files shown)


docs/reference/cli.md
```diff
@@ -30,6 +30,7 @@ Commands:
   info                    Print information about a SQLMesh project.
   init                    Create a new SQLMesh repository.
   invalidate              Invalidate the target environment, forcing its...
+  janitor                 Run the janitor task on-demand.
```
georgesittas, line 33: Should we also expose a notebook magic for this?

georgesittas (approved):

sqlmesh/core/console.py
```diff
@@ -441,10 +460,26 @@ def stop_creation_progress(self, success: bool = True) -> None:
         self.environment_naming_info = EnvironmentNamingInfo()
         self.default_catalog = None
 
+    def start_cleanup(self, ignore_ttl: bool) -> bool:
+        if ignore_ttl:
+            self._print(
+                "Are you sure you want to delete all orphaned snapshots regardless of their environment?"
```
izeigerman, line 466: I don't understand this message. Doesn't `oprhaned` mean they no longer belong to any environments?

sqlmesh/core/console.py
```diff
@@ -441,10 +460,26 @@ def stop_creation_progress(self, success: bool = True) -> None:
...
+            )
+            if not self._confirm("This may affect other users. Proceed?"):
```
izeigerman, line 468: This is not very API friendly

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -257,14 +257,16 @@ def invalidate_environment(self, name: str) -> None:
...
+        self, ignore_ttl: bool = False
+    ) -> t.List[SnapshotTableCleanupTask]:
         current_ts = now_timestamp(minute_floor=False)
 
-        expired_query = (
-            exp.select("name", "identifier", "version")
-            .from_(self.snapshots_table)
-            .where(exp.column("expiration_ts") <= current_ts)
-        )
+        expired_query = exp.select("name", "identifier", "version").from_(self.snapshots_table)
```
izeigerman, line 265: There will be a significant runtime penalty

erindru: Only if the user specifies `--ignore-ttl`, which means that they do want every snapshot checked. I think if theyre going down this road then they can accept the penalty

docs/reference/cli.md
erindru: Added

sqlmesh/core/console.py
erindru: I've adjusted the wording to make it clearer about what will happen. Is it better?

sqlmesh/core/console.py
erindru: True, and ironically it also makes it hard to test the Notebook magic because it requires user input.

However, I dont think this is something we should encourage the users to be running automatically anyway. It's more of an escape hatch to speed something up outside of the normal.

I could add a `--no-prompts` flag like some of the other operations, would that be better?

izeigerman (approved):

PR #2850, opened 2024-07-02
feat: allow users to configure high watermark sensor behavior

eakmanrq:
This PR makes it so that users have the ability to change the behavior for HighWaterMark and HighWaterMarkExternal sensors.

Also removed the default values for `poke_interval` and `timeout` since they were just copies of Airflow's default we will just defer to Airflow to set them. 

Ran the Airflow local tests and confirmed they passed.


sqlmesh/schedulers/airflow/dag_generator.py
```diff
@@ -74,6 +74,8 @@ def __init__(
             t.Callable[[t.Dict[str, t.Any]], BaseSensorOperator]
         ],
         sensor_mode: str,
+        high_water_mark_sensor_args: t.Optional[t.Dict[str, t.Any]],
+        high_water_mark_external_sensor_args: t.Optional[t.Dict[str, t.Any]],
```
izeigerman, line 78: Calling external sensors "high watermark" was a mistake and likely a result of a copy paste. Can we just all this `external_sensor_args`?

sqlmesh/schedulers/airflow/dag_generator.py
```diff
@@ -83,6 +85,8 @@ def __init__(
         self._external_table_sensor_factory = external_table_sensor_factory
         self._state_reader = state_reader
         self._sensor_mode = sensor_mode
+        self._high_water_mark_sensor_args = high_water_mark_sensor_args or {}
```
izeigerman, line 88: Should we remove the `sensor_mode` argument since it can now just be passed as part of arguments?

eakmanrq: One reason why why `sensor_mode` would be treated different than the rest is because it is changing the default behavior (default is `poke`). That is why I initially kept it at least since it is making explicit/obvious this behavior change that SQLMesh is doing.

sqlmesh/schedulers/airflow/dag_generator.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/2850/commits/66e41db19c895065b66bfe41d4e99c1621faeb19

izeigerman (approved):

PR #2853, opened 2024-07-02
Fix: Use the default dialect when constructing a table mapping key in the dbt adapter

izeigerman:
The keys in the mapping have been serialized using the default dialect so the same dialect should be used for lookup.

### DIFF
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -6,6 +6,7 @@ import typing as t
 
 import pandas as pd
 from sqlglot import exp, parse_one
+from sqlglot.optimizer.qualify_columns import quote_identifiers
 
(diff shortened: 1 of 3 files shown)


sqlmesh/dbt/adapter.py
```diff
@@ -353,7 +353,9 @@ def resolve_identifier(self, relation: BaseRelation) -> t.Optional[str]:
         return identifier if identifier else None
 
     def _map_table_name(self, table: exp.Table) -> exp.Table:
-        name = table.sql(dialect=self.project_dialect)
+        # Use the default dialect since this is the dialect used to normalize and quote keys in the
+        # mapping table.
+        name = table.sql()
```
tobymao, line 358: is table always going to be quoted here? or should we force the quotes?

tobymao (approved):

PR #2856, opened 2024-07-03
Fix: cast NULLs in VALUES clause (df->values) to avoid coercion issues

### LINKED ISSUE #2684: Bigquery unit tests with partial inputs fail on uncoercable types
The `UNNEST([STRUCT(...), ...])` statement thats generated in BQ does not pass in any types. This causes all elements to be inferred as a number type, `INT64` I think. The `CAST` that then happens in the `SELECT` statement fails on uncoercible types. Most obviously with `DATE` / `TIMESTAMP`.  You can reproduce with this:

`select cast(x as timestamp) from unnest([null]) as x` 

Action:

`select cast(x as timestamp) from unnest([cast(null as timestamp)]) as x`

Adding a cast to just the first index of the `[...]` array is sufficient for BQ to type it correctly. With or without the outer `CAST` which would then be a no-op. A real life query generated by the UT framework may look like this:

```sql
SELECT CAST(`id` AS STRING), CAST(`opp1` AS STRING), CAST(`sort_order` AS INT64), ... 
FROM UNNEST([STRUCT('line1' AS `id`, 'opp1' AS `opportunity_id`, NULL AS `sort_order`, NULL AS `pricebook_entry_id`, 'product1' AS `product2_id`, NULL AS `product_code`, NULL AS `name`, 1 AS `quantity`, NULL AS `discount`, NULL AS `subtotal`, NULL AS `total_price`, NULL AS `unit_price`, NULL AS `list_price`, NULL AS `service_date`, NULL AS `description`, NULL AS `created_date`, NULL AS `created_by_id`,  ...
```

georgesittas:
Fixes #2684

Gave it a whirl. Not sure if this is too intrusive to fix the linked issue - happy to discuss and refactor if needed.

cc @z3z1ma

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -897,6 +897,14 @@ def select_from_values_for_batch_range(
         ]
 
     values_exp = exp.values(expressions, alias=alias, columns=columns_to_types)
+    if values:
+        # BigQuery crashes on `SELECT CAST(x AS TIMESTAMP) FROM UNNEST([NULL]) AS x`, but not
+        # on `SELECT CAST(x AS TIMESTAMP) FROM UNNEST([CAST(NULL AS TIMESTAMP)]) AS x`. This
+        # ensures nulls under the `Values` expression are cast to avoid similar issues.
+        for value, kind in zip(values_exp.expressions[0].expressions, columns_to_types.values()):
+            if isinstance(value, exp.Null):
+                value.replace(exp.cast(value, to=kind))
+
     return exp.select(*casted_columns).from_(values_exp, copy=False).where(where, copy=False)
 
 
--- a/tests/core/test_dialect.py
+++ b/tests/core/test_dialect.py
@@ -375,6 +375,20 @@ def test_select_from_values_for_batch_range_json():
     )
 
 
+def test_select_from_values_that_include_null():
+    values = [(1, exp.null())]
+    columns_to_types = {
+        "id": exp.DataType.build("int", dialect="bigquery"),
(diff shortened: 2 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -897,6 +897,14 @@ def select_from_values_for_batch_range(
         ]
 
     values_exp = exp.values(expressions, alias=alias, columns=columns_to_types)
+    if values:
```
tobymao, line 900: if we cast values, directly, do we need to also cast the select?

georgesittas: are you referring to the projections? because we already cast them [earlier in this function](https://github.com/TobikoData/sqlmesh/pull/2856/files/6f9e7ee997868e3f3674ad8a506cb7841b6395a6#diff-32e7379e6d699794516fd766eee46a0a093be9bb23c3fdb4f8ea20117c2a44c3R884-R887)

tobymao: yea but now it's a double cast, cast in the values, and cast in the projection, should we just do one?

georgesittas: Ah gotcha.. I guess we can try? I was being conservative because I'm not sure what the implications are in case we remove those casts, maybe we'll fail to infer the type of a projection down the line somewhere? On the other hand, we seem to have the `columns_to_types` mapping already, so perhaps it's fine.

tobymao: no big deal

tobymao (approved):

PR #2861, opened 2024-07-03
Fix: duplicate columns in table diff query

### LINKED ISSUE #2821: table_diff incorrect query syntax on MSSQL
## Bug
`table_diff` command fails to run on MSSQL possibly due to SQLGlot parsing error.

## Description
Thanks to the detailed debug log, I managed to track down where the query went wrong. 

This code portion from [core/table_diff.py](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/table_diff.py#L255-L279) will produce syntax error on MSSQL.

```python
query = (
    exp.select(
        *s_selects.values(),
        *t_selects.values(),
        exp.func("IF", exp.or_(*(c.not_().is_(exp.Null()) for c in s_index)), 1, 0).as_(
            "s_exists"
        ),
        exp.func("IF", exp.or_(*(c.not_().is_(exp.Null()) for c in t_index)), 1, 0).as_(
            "t_exists"
        ),
        exp.func(
            "IF",
            exp.and_(
                *(
                    exp.and_(
                        exp.column(c, "s").eq(exp.column(c, "t")),
                        exp.column(c, "s").not_().is_(exp.Null()),
                        exp.column(c, "t").not_().is_(exp.Null()),
                    )
                    for c in index_cols
                ),
            ),
            1,
            0,
        ).as_("row_joined"),
...
```

#### DuckDB (working)

The following snippet is from `sqlmesh table_diff prod:dev sqlmesh_example.incremental_model` on the DuckDB instance created by `sqlmesh init duckdb`.

*Although this code runs fine, the `NOT "table"."field" IS NULL` clause gets duplicated in `s_exists` and `t_exists`.*

```sql
CASE
	WHEN NOT "s"."id" IS NULL
	OR NOT "s"."event_date" IS NULL
	OR NOT "s"."id" IS NULL
	OR NOT "s"."event_date" IS NULL THEN 1
	ELSE 0
END AS "s_exists",
CASE
	WHEN NOT "t"."id" IS NULL
	OR NOT "t"."event_date" IS NULL
	OR NOT "t"."id" IS NULL
	OR NOT "t"."event_date" IS NULL THEN 1
	ELSE 0
END AS "t_exists",
CASE
	WHEN ("s"."id" = "t"."id"
	AND NOT "s"."id" IS NULL
	AND NOT "t"."id" IS NULL)
	AND ("s"."event_date" = "t"."event_date"
	AND NOT "s"."event_date" IS NULL
	AND NOT "t"."event_date" IS NULL) THEN 1
	ELSE 0
END AS "row_joined",
```

#### MSSQL (error)

`NOT [s].[id] <> 0 IS NULL` is not a valid syntax causing the fatal error.

```sql
IIF(NOT [s].[id] <> 0 IS NULL
    OR NOT [s].[event_date] <> 0 IS NULL
    OR NOT [s].[id] <> 0 IS NULL
    OR NOT [s].[event_date] <> 0 IS NULL,
	1,
	0) AS [s_exists],
IIF(NOT [t].[id] <> 0 IS NULL
    OR NOT [t].[event_date] <> 0 IS NULL
    OR NOT [t].[id] <> 0 IS NULL
    OR NOT [t].[event_date] <> 0 IS NULL,
	1,
	0) AS [t_exists],
IIF(([s].[id] = [t].[id]
    AND NOT [s].[id] <> 0 IS NULL
    AND NOT [t].[id] <> 0 IS NULL)
    AND ([s].[event_date] = [t].[event_date]
    AND NOT [s].[event_date] <> 0 IS NULL
    AND NOT [t].[event_date] <> 0 IS NULL),
	1,
	0) AS [row_joined],
```

## Environment

```
sqlmesh 0.104.1 and 0.106.1
sqlglot 25.0.3 and 25.3.1
duckdb 1.0.0
mssql 2016
```

I really wish I could help fix the code, but SQLGlot is too daunting for me now. Give me some time and I'll try to contribute more. 😅 

Massive thanks! 🙏

na399:
Prior to this fix, the table diff query has duplicate columns in their source and table index lists, `s_index` and `t_index`, respectively. This issue was also reported in #2821 as a minor issue. While #2825 fixed the major issue in #2821, `table_diff` still fails to run on MSSQL due to the following error:
> A column has been specified more than once in the order by list. Columns in the order by list must be unique.

Generated query prior to the fix: 
```sql
SELECT ... FROM sqlmesh.__temp_diff_xyz WHERE ... ORDER BY s__foo, s__bar, s__foo, s__bar, t__foo, t__bar, t__foo, t__bar
```

With this fix:
```sql
SELECT ... FROM sqlmesh.__temp_diff_xyz WHERE ... ORDER BY s__foo, s__bar, t__foo, t__bar
```

---

These `s_index` and `t_index` are used at only another section in the same [table_diff.py](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/table_diff.py#L259-L264), but they are `OR` expressions so the duplicates don't cause fatal errors there.


tobymao (approved):
can you add a test case please?

thanks so much for the fix

na399:
Thanks, @tobymao.
Did you mean a test case for mssql specifically or the general [test_table_diff](https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_table_diff.py)? 

I went through code and find no other uses of the duplicate column identifiers elsewhere. So this change should not change the overall behavior of the function. I might be wrong though. Please advice.

tobymao:
general test case for test_table_diff that checks that the sql is generated as expected

tests/core/test_table_diff.py
```diff
@@ -217,3 +218,43 @@ def test_grain_check(sushi_context_fixed_date):
...
+    )
+
+    query_sql = 'CREATE TABLE IF NOT EXISTS "sqlmesh_temp"."__temp_diff_6mexbir9" AS SELECT *, CASE WHEN "key_matches" = 1 AND "value_matches" = 1 THEN 1 ELSE 0 END AS "row_full_match" FROM (SELECT "s"."key" AS "s__key", "s"."value" AS "s__value", "t"."key" AS "t__key", "t"."value" AS "t__value", CASE WHEN NOT "s"."key" IS NULL THEN 1 ELSE 0 END AS "s_exists", CASE WHEN NOT "t"."key" IS NULL THEN 1 ELSE 0 END AS "t_exists", CASE WHEN "s"."key" = "t"."key" AND NOT "s"."key" IS NULL AND NOT "t"."key" IS NULL THEN 1 ELSE 0 END AS "row_joined", CASE WHEN "s"."key" IS NULL AND "t"."key" IS NULL THEN 1 ELSE 0 END AS "null_grain", CASE WHEN "s"."key" = "t"."key" THEN 1 WHEN ("s"."key" IS NULL) AND ("t"."key" IS NULL) THEN 1 WHEN ("s"."key" IS NULL) OR ("t"."key" IS NULL) THEN 0 ELSE 0 END AS "key_matches", CASE WHEN ROUND("s"."value", 3) = ROUND("t"."value", 3) THEN 1 WHEN ("s"."value" IS NULL) AND ("t"."value" IS NULL) THEN 1 WHEN ("s"."value" IS NULL) OR ("t"."value" IS NULL) THEN 0 ELSE 0 END AS "value_matches" FROM "table_diff_source" AS "s" FULL JOIN "table_diff_target" AS "t" ON ("s"."key" = "t"."key") OR (("s"."key" IS NULL) AND ("t"."key" IS NULL))) AS "stats"'
```
georgesittas, line 247: @Themiscodes isn't the `6mexbir9` part randomly produced here? Not sure how this passed CI, but something seems off. I'd expect the `random_id` to be patched for this to work properly.

Themiscodes: Right, I adapted this to use patch instead thanks! It was generating the same randomid for this test before, but I haven't identified the reason why it was doing this yet

tests/core/test_table_diff.py
```diff
@@ -217,3 +218,45 @@ def test_grain_check(sushi_context_fixed_date):
...
+    summary_query_sql = 'SELECT SUM("s_exists") AS "s_count", SUM("t_exists") AS "t_count", SUM("row_joined") AS "join_count", SUM("null_grain") AS "null_grain_count", SUM("row_full_match") AS "full_match_count", SUM("key_matches") AS "key_matches", SUM("value_matches") AS "value_matches", COUNT(DISTINCT ("s__key")) AS "distinct_count_s", COUNT(DISTINCT ("t__key")) AS "distinct_count_t" FROM "sqlmesh_temp"."__temp_diff_abcdefgh"'
```
georgesittas, line 248: Do we care about these two? Isn't `sample_query_sql` the only one we're interested in since it has the `ORDER` clause?

georgesittas (approved):

tests/core/test_table_diff.py
Themiscodes: Yes `sample_query_sql` suffices for this particular fix. Checking `query_sql` was mainly to see that s_exists and t_exists cases have no duplicates (though this wouldn't throw an error) and thought it would be a good idea to include a check for the `summary_query` as well to ensure that any future changes to `table_diff` still generate the expected results there, but I can remove these

georgesittas: Ah got it, ignore me then. What you did makes sense, I forgot we're using `s/t_index` in those places too.

PR #2862, opened 2024-07-03
Fix!: Move metadata only macro funcs from model's data hash to metadata hash

vchan:


sqlmesh/core/model/definition.py
```diff
@@ -727,7 +727,13 @@ def data_hash(self) -> str:
     @property
     def _data_hash_values(self) -> t.List[str]:
         data = [
-            str(self.sorted_python_env),
+            str(  # Exclude metadata only macro funcs
+                [
+                    (k, v)
+                    for k, v in self.sorted_python_env
+                    if isinstance(v, Executable) and not v.is_metadata
```
izeigerman, line 734: Shouldn't this be:
```
if not isinstance(v, Executable) or not v.is_metadata
```

izeigerman: Actually, when it's ever non `Executable`?

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1545,6 +1545,17 @@ The methods are available because the `column` argument is parsed as a SQLGlot [
 
 Column expressions are sub-classes of the [Condition class](https://sqlglot.com/sqlglot/expressions.html#Condition), so they have builder methods like [`between`](https://sqlglot.com/sqlglot/expressions.html#Condition.between) and [`like`](https://sqlglot.com/sqlglot/expressions.html#Condition.like).
 
+#### Metadata only macros as model pre/post-statements
+When you first use your macro functions in your models as pre/post-statements, SQLMesh will identify those models as directly modified the next time you create a plan. These models will then need backfills. The same thing applies when you edit or remove these pre/post-statements. If your macro does not have any effect on your models' data and you do not want it to trigger backfills, you can configure your macro to be part of a model's metadata. That way, SQLMesh can still detect changes and create new snapshots for your models when you add, edit, or delete your macro pre/post-statements. To do this, pass in True to the `is_metadata` parameter of the `@macro()` decorator.
+
+```python linenums="1"
+from sqlmesh import macro
+
+@macro(is_metadata=True)
```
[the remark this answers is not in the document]
vchan: That's a good point. No, there wasn't a reason. I can rename it.

sqlmesh/core/model/definition.py
vchan: You're right. I should've taken a closer look at the type annotation of python_env.

erindru (approved):

tobymao (approved):

PR #2863, opened 2024-07-04
Fix!: Support for trailing comments in SQL model definitions

izeigerman:
Before this fix, the renderer would return a `Semicolon` expression, resulting in an empty SQL string and subsequent failure when evaluating in the engine adapter.


tobymao (approved):
needs migration i assume…

izeigerman:
@tobymao I don't think anyone who ran into this was able to run a plan.

izeigerman:
@tobymao actually, I think you're right

PR #2867, opened 2024-07-04
Feat: add support for native SQLMesh SQL macros

georgesittas:
This is a hacky and somewhat incomplete (*) solution, but I tried a few others and they all had issues related to serialization. So I'm mostly opening this PR to begin a discussion and if needed I can refactor, perhaps with the help of @Themiscodes.

Regarding the serialization issues, take the following example as an alternative approach (loader.py):

```python
for definition in macro_definitions:
    _, fn = _norm_var_arg_lambda(MacroEvaluator(dialect=dialect), definition.expression)
    macro(definition.name)(lambda evaluator, *args: fn(evaluator, *args)
```

At serialization time, `normalize_source` would produce a payload for the lambda that looks like `"lambda ...: ... fn ..."` and when we actually try to evaluate this, `fn` would be an undefined reference. Other [approaches](https://github.com/TobikoData/sqlmesh/compare/main...themis/reusable_macros) had similar issues. Maybe we could explore other options, i.e. avoid converting a SQL definition into a python function that simulates a macro?

(*) kwargs are not supported for these macros, because `exp.Lambda` only accepts positional arguments.

Fixes https://github.com/TobikoData/sqlmesh/issues/2505

### DIFF
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -11,10 +11,11 @@ from pathlib import Path
 
 from sqlglot.errors import SchemaError, SqlglotError
 from sqlglot.schema import MappingSchema
+from sqlglot import TokenType, exp, tokenize
 
 from sqlmesh.core import constants as c
 from sqlmesh.core.audit import Audit, load_multiple_audits
-from sqlmesh.core.dialect import parse
+from sqlmesh.core.dialect import parse, MacroDef
 from sqlmesh.core.macros import MacroRegistry, macro
 from sqlmesh.core.metric import Metric, MetricMeta, expand_metrics, load_metric_ddl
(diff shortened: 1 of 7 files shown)


sqlmesh/core/loader.py
```diff
@@ -261,6 +263,7 @@ def _load_scripts(self) -> t.Tuple[MacroRegistry, JinjaMacroRegistry]:
                         else macro_file_mtime
                     )
 
+            macro_definitions = []
```
izeigerman, line 266: Let's be more specific with the name. There are too many different kinds of macros in this method

georgesittas:
Closing this for now, may revisit in the future with a more robust implementation.

PR #2870, opened 2024-07-05
Fix!: Snowflake adapter

erindru:
This fixes the Snowflake adapter so that the integration tests pass again (and `sushi` example compatibility is restored).

The main issue was the catalog from the Snowflake gateway config not being respected. This change modifies every query `Expression` before it gets converted to SQL to replace the default catalog references (which are almost always wrong in a multi-dialect project because they contain the catalog normalized against the source dialect, not Snowflake) with whatever the user supplied in the Gateway config.

There are also some other issues around how the `write_pandas()` function from the Snowflake python lib handles things as well as recent changes to normalizing environment names that broke the `SushiDataValidator` in the `test_sushi` integration test.

Note that the integration tests are still not being run as part of the build; that will be a future PR

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -292,10 +292,6 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
 
         schema = to_schema(schema_name)
         catalog_name = schema.catalog or self.get_current_catalog()
-        if catalog_name:
-            catalog_name = normalize_identifiers(catalog_name, dialect=self.dialect).sql(
-                dialect=self.dialect
-            )
 
         query = (
             exp.select(
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -200,7 +200,9 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
         batch_size: int,
         target_table: TableName,
     ) -> t.List[SourceQuery]:
-        temp_table = self._get_temp_table(target_table or "pandas")
+        temp_table = self._get_temp_table(
+            target_table or "pandas", quoted=False
+        )  # write_pandas() re-quotes everything without checking if its already quoted
 
         def query_factory() -> Query:
             if snowpark and isinstance(df, snowpark.dataframe.DataFrame):
@@ -211,10 +213,10 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
                 # Workaround for https://github.com/snowflakedb/snowflake-connector-python/issues/1034
                 # The above issue has already been fixed upstream, but we keep the following
                 # line anyway in order to support a wider range of Snowflake versions.
-                schema = f'"{temp_table.db}"'
+                schema = temp_table.db
                 if temp_table.catalog:
-                    schema = f'"{temp_table.catalog}".{schema}'
-                self.cursor.execute(f"USE SCHEMA {schema}")
+                    schema = f"{temp_table.catalog}.{schema}"
+                self.set_current_schema(schema)
 
                 # See: https://stackoverflow.com/a/75627721
                 for column, kind in columns_to_types.items():
@@ -240,7 +242,11 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
                     df,
                     temp_table.name,
                     schema=temp_table.db or None,
-                    database=temp_table.catalog or None,
+                    database=normalize_identifiers(temp_table.catalog, dialect=self.dialect).sql(
+                        dialect=self.dialect
+                    )
+                    if temp_table.catalog
+                    else None,
                     chunk_size=self.DEFAULT_BATCH_SIZE,
                     overwrite=True,
                     table_type="temp",  # if you dont have this, it will convert the table we created above into a normal table and it wont get dropped when the session ends
@@ -252,7 +258,13 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
 
             return exp.select(*self._casted_columns(columns_to_types)).from_(temp_table)
 
-        return [SourceQuery(query_factory=query_factory)]
+        # the cleanup_func technically isnt needed because the temp table gets dropped when the session ends
+        # but boy does it make our multi-adapter integration tests easier to write
+        return [
+            SourceQuery(
+                query_factory=query_factory, cleanup_func=lambda: self.drop_table(temp_table)
+            )
+        ]
 
     def _fetch_native_df(
         self, query: t.Union[exp.Expression, str], quote_identifiers: bool = False
@@ -280,6 +292,11 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
 
         schema = to_schema(schema_name)
         catalog_name = schema.catalog or self.get_current_catalog()
+        if catalog_name:
+            catalog_name = normalize_identifiers(catalog_name, dialect=self.dialect).sql(
+                dialect=self.dialect
+            )
+
         query = (
             exp.select(
                 exp.column("TABLE_CATALOG").as_("catalog"),
@@ -308,6 +325,8 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
             )
             .from_(exp.table_("TABLES", db="INFORMATION_SCHEMA", catalog=catalog_name))
             .where(exp.column("TABLE_SCHEMA").eq(schema.db))
+            # Snowflake seems to have delayed internal metadata updates and will sometimes return duplicates
+            .distinct()
         )
         if object_names:
             query = query.where(exp.column("TABLE_NAME").isin(*object_names))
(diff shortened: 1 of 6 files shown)


sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -102,6 +102,10 @@ def snowpark(self) -> t.Optional[SnowparkSession]:
             ).getOrCreate()
         return None
 
+    def get_current_catalog(self) -> t.Optional[str]:
+        current_catalog = super().get_current_catalog()
+        return current_catalog.lower() if current_catalog else current_catalog
```
tobymao, line 107: this doesn't seem correct, why did you need this?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -326,7 +349,40 @@ def _get_data_objects(
...
+
+    def set_current_schema(self, schema: str) -> None:
+        self.execute(exp.Use(kind="SCHEMA", this=to_schema(schema)))
+
+    def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.Any) -> str:
+        # note: important to use self._default_catalog instead of the self.default_catalog property
+        # otherwise we get RecursionError: maximum recursion depth exceeded
+        # because it calls get_current_catalog(), which executes a query, which needs the default catalog, which calls get_current_catalog()... etc
+        if self._default_catalog:
+            default_catalog_compare = exp.to_identifier(self._default_catalog).output_name.replace(
```
erindru, line 362: Is there a better way to unquote things that are potentially quoted?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -200,7 +204,9 @@ def _df_to_source_queries(
         batch_size: int,
         target_table: TableName,
     ) -> t.List[SourceQuery]:
-        temp_table = self._get_temp_table(target_table or "pandas")
+        temp_table = self._get_temp_table(
+            target_table or "pandas", quoted=False
```
tobymao, line 208: why do we need to unquote this? can we just make sure evyrtihng is quoted?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -308,6 +329,8 @@ def _get_data_objects(
             )
             .from_(exp.table_("TABLES", db="INFORMATION_SCHEMA", catalog=catalog_name))
             .where(exp.column("TABLE_SCHEMA").eq(schema.db))
+            # Snowflake seems to have delayed internal metadata updates and will sometimes return duplicates
+            .group_by(exp.Group(all=True))
```
tobymao, line 333: can you call distinct() instead

sqlmesh/core/engine_adapter/snowflake.py
tobymao: exp.to_identifier(...).name, there will be no quotes

sqlmesh/core/engine_adapter/snowflake.py
erindru: Nope, because these get passed to `write_pandas()`. This function comes from the Snowflake library. It has a bunch of problems:

- It has a flag called `quote_identifiers` which defaults to `True`. So if you give it already quoted identifiers, it happily double quotes them.
- If you set `quote_identifiers=False`, then it doesnt quote the column names. Therefore, the table gets created with columns like `ID` but the rest of the SQLMesh code wants to query it as `"id"`, causing an error.

The path of least resistance was to simply not quote the identifiers first. This means that:
 - `write_pandas()` quotes them and creates correctly named tables with correctly named columns
 - `EngineAdapter.execute()` also quotes them when `create_table()` is called, because it see's they're unquoted when it goes to run the query

sqlmesh/core/engine_adapter/snowflake.py
erindru: It was to make a test pass, i've fixed the test instead and removed this

sqlmesh/core/engine_adapter/snowflake.py
erindru: Done

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -240,7 +242,11 @@ def query_factory() -> Query:
...
                     schema=temp_table.db or None,
-                    database=temp_table.catalog or None,
+                    database=normalize_identifiers(temp_table.catalog, dialect=self.dialect).sql(
+                        dialect=self.dialect
+                    )
+                    if temp_table.catalog
+                    else None,
                     chunk_size=self.DEFAULT_BATCH_SIZE,
                     overwrite=True,
                     table_type="temp",  # if you dont have this, it will convert the table we created above into a normal table and it wont get dropped when the session ends
```
izeigerman, line 252: Is this comment still relevant? Seems misleading

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -326,7 +345,39 @@ def _get_data_objects(
...
+
+            def catalog_rewriter(node: exp.Expression) -> exp.Expression:
+                if isinstance(node, exp.Table):
+                    if node.catalog:
+                        # only replace the catalog on the model with the target catalog if the two are functionally equivalent
+                        if unquote_and_lower(node.catalog) == default_catalog_unquoted:
+                            node.set("catalog", default_catalog_normalized)
+                return node
+
+            # Rewrite whatever default catalog is present on the query to be compatible with what the user supplied in the
```
izeigerman, line 375: I thought that was by design? And that's exactly why we required users to specify the catalog in quotes in the connection config when using a cross-dialect setup. @eakmanrq please correct me if I'm wrong.

erindru: So what happens if you specify `database: '"ERIN"'` in the Snowflake connection config is [this code](https://github.com/TobikoData/sqlmesh/blob/7acaef31817a4481666e7a109daa0c1c019f61b8/sqlmesh/core/model/meta.py#L471) completely ignores what you did, strips the quotes and re-normalizes it in the source dialect. The normalized identifier in the source dialect is not compatible with Snowflake.

The reason why we dont just update `model.fqn` to fix the catalog is because `model.fqn` is used as a key all over the place in internal dictionaries and potentially the state

eakmanrq: I would like to hear more about what the ideal fix would be for `model.fqn`. Are you suggesting mixing the reference dialect used for normalization so that catalog is from target and db/table is from source? Or it all uses target instead?

erindru: Yes that's right. So the model `fqn` is currently `<catalog>.<schema>.<table>`, all normalized and quoted in the source dialect.

The proposal would be to leabe `<schema>.<table>` normalized and quoted in the source dialect, but change `<catalog>` to be normalized and quoted in the _target_ dialect, but ONLY if `<catalog>` came from the gateway config. If `<catalog>` was manually specified on the model by the user, then we of don't touch it

erindru: Ok so after talking to @izeigerman, and going down a massive rabbit hole trying to normalize the `<catalog>` part to the target dialect and the rest to the source dialect, we determined it was easier to just do the catalog replacement within the adapter (like in this PR) because the change is minimal and also self contained to the adapter.

If anyone's interested, fixing `model.fqn` to normalize the `<catalog>` part  separately from the rest is harder than it sounds. Even if you fix all the tests that make assumptions about the format of the catalog you're still left with broken dependencies.

This is because `model.depends_on` [runs the model query and extracts the table names](https://github.com/TobikoData/sqlmesh/blob/e27daf54c19ba1c3d0fe78a86f5b928b3795630e/sqlmesh/core/model/definition.py#L846) (which [of course](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/renderer.py#L236) are [fully normalized in the source dialect](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/renderer.py#L295), [including the catalog](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/dialect.py#L1079)). Due to this, it then fails to match them against the models loaded into the Context (because `model.fqn` doesnt match against what was extracted from the query), so the dependency tree is broken.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -326,12 +345,44 @@ def _get_data_objects(
...
+            # Snowflake connection config. This is because the catalog present on the model gets normalized and quoted to match
+            # the source dialect, which isnt always compatible with Snowflake
+            expression = expression.transform(catalog_rewriter)
+
+        return super()._to_sql(expression=expression, quote=quote, **kwargs)
```
izeigerman, line 380: So we do traverse the tree twice

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -280,6 +292,11 @@ def _get_data_objects(
 
         schema = to_schema(schema_name)
         catalog_name = schema.catalog or self.get_current_catalog()
+        if catalog_name:
+            catalog_name = normalize_identifiers(catalog_name, dialect=self.dialect).sql(
```
tobymao, line 296: is this problematic if the schema.catalog is duckdb?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -326,12 +345,44 @@ def _get_data_objects(
...
+        self.execute(exp.Use(kind="SCHEMA", this=to_schema(schema)))
+
+    def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.Any) -> str:
+        # note: important to use self._default_catalog instead of the self.default_catalog property
+        # otherwise we get RecursionError: maximum recursion depth exceeded
+        # because it calls get_current_catalog(), which executes a query, which needs the default catalog, which calls get_current_catalog()... etc
+        if self._default_catalog:
+
+            def unquote_and_lower(identifier: str) -> str:
+                return exp.parse_identifier(identifier).name.lower()
```
tobymao, line 360: why are we doing this?

sqlmesh/core/engine_adapter/snowflake.py
erindru: In trying to answer your question, I realised this was no longer needed. The tests pass without it, which means I probably added it to get a certain test to pass and then made a wider change to get other tests to pass which made this redundant.

I've removed it

sqlmesh/core/engine_adapter/snowflake.py
erindru: to "fuzzy match" the catalog set on the query vs the default catalog set in the config. The idea is to distinguish catalogs the user has deliberately set vs when the default catalog has been used.

For example, let's say the source dialect is `duckdb` and the target is `snowflake`. The query is normalized in the source dialect and might contain a reference to `"erin"."schema"."table"`. However, my Snowflake gateway is configured with `database: ERIN`, so this is wrong, we need to change the reference to `ERIN."schema"."table"` or ideally `"ERIN"."schema"."table"` to be truly unambiguous,

Since `ERIN` is not the same as `"erin"`, I cant do a direct string comparison. But if I remove the quotes and lowercase both sides, `erin`==`erin` so I can assume that the default catalog is in use and can replace the catalog reference in the query with the _normalized_ version from the gateway config.

If the user has specified some other catalog manually, eg the model fqn is `"external"."schema"."table"` then it wont match because `erin` != `external` so the catalog is not touched.

Doing it like this means it'll also work if I do something crazy like use a case sensitive catalog in my connection config, eg `database: '"EriN"'`.

Since `unquote_and_lower('"ErIN"') == unquote_and_lower('"erin"')` I know to replace the catalog

tobymao: perfect. can we add a comment or change the name such that the intention of the fuzzy match is clear?

Issue #2869, opened 2024-07-04
Table diff error not found in axis


achicoine-coveo:
Context: 
Running sqlmesh==0.109.2
The project dialect is configured as Snowflake and it runs on a duckdb gateway.

When trying to use the table diff feature, I get the error "['abc'] not found in axis".
Looking at the error, it seemed strange to me that the column name is lowercase since I know a lot of database connectors will return the column names in uppercase. So I tried with a quoted identifier in lowercase and it works for that one.

Quoted lowercase works:
<img width="1486" alt="Screenshot 2024-07-04 at 3 43 41 PM" src="https://github.com/TobikoData/sqlmesh/assets/110614540/819eb287-0123-49c0-8087-fb9372ec4572">

Quoted uppercase didn't work:
<img width="1466" alt="Screenshot 2024-07-04 at 3 53 23 PM" src="https://github.com/TobikoData/sqlmesh/assets/110614540/d5a27232-6752-4a25-8e95-ca6d0c4f92ec">

Unquoted uppercase didn't work:
<img width="1470" alt="Screenshot 2024-07-04 at 3 45 13 PM" src="https://github.com/TobikoData/sqlmesh/assets/110614540/7f651c21-2392-48c8-a68c-9e5b3af8244a">

Unquoted lowercase didn't work:
<img width="1467" alt="Screenshot 2024-07-04 at 3 48 04 PM" src="https://github.com/TobikoData/sqlmesh/assets/110614540/2c5fd226-04f3-4cef-88ea-815da03cc0a8">

Code that could be relevant:
https://github.com/TobikoData/sqlmesh/blob/f3cb2fa5cbc8a504514d333f5407abeba0172787/sqlmesh/core/table_diff.py#L179 the identifiers are normalized here.
Attempt at dropping the index happens here: https://github.com/TobikoData/sqlmesh/blob/f3cb2fa5cbc8a504514d333f5407abeba0172787/sqlmesh/core/table_diff.py#L345

I'm not sure what's happening in the code, but my guess would be the column names are uppercase in the pandas dataframe and it's trying to drop the column name in lowercase. A lot of connectors will return the column names in uppercase in the dataframe even if the column names were lowercase in the sql.

Traceback:
Traceback (most recent call last):
  File "/Users/alexislocal/.virtualenvs/dae-cost-data-model-prototype/lib/python3.9/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
[...]

PR #2872, opened 2024-07-05
Fix: normalization issue in table diff

Themiscodes:
This addresses an issue encountered when running table diff, fixes: #2869 
The proposed fix is to use the project dialect for normalization instead of the gateway dialect and also account for quoted names. `test_table_diff` has been adapted to test for both quoted and unquoted identifiers in snowflake.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1271,7 +1271,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         self,
         source: str,
         target: str,
-        on: t.List[str] | exp.Condition | None = None,
+        on: t.List[str | exp.Identifier] | exp.Condition | None = None,
         model_or_snapshot: t.Optional[ModelOrSnapshot] = None,
         where: t.Optional[str | exp.Condition] = None,
         limit: int = 20,
@@ -1299,6 +1299,10 @@ class GenericContext(BaseContext, t.Generic[C]):
             The TableDiff object containing schema and summary differences.
         """
         source_alias, target_alias = source, target
+
+        if on and not isinstance(on, exp.Condition):
+            on = [exp.to_identifier(key, quoted=True) for key in on]
+
         if model_or_snapshot:
             model = self.get_model(model_or_snapshot, raise_if_missing=True)
             source_env = self.state_reader.get_environment(source)
@@ -1321,7 +1325,13 @@ class GenericContext(BaseContext, t.Generic[C]):
             if not on:
                 for ref in model.all_references:
                     if ref.unique:
-                        on = ref.columns
+                        on = (
+                            [key.this for key in ref.expression.expressions]
+                            if isinstance(ref.expression, exp.Tuple)
+                            else [ref.expression.this.this]
+                            if isinstance(ref.expression.this.this, str)
+                            else [ref.expression.this.this.this]
+                        )
 
         if not on:
             raise SQLMeshError(
@@ -1337,6 +1347,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             source_alias=source_alias,
             target_alias=target_alias,
             model_name=model.name if model_or_snapshot else None,
+            model_dialect=model.dialect if model_or_snapshot else None,
             limit=limit,
             decimals=decimals,
         )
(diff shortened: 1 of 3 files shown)


sqlmesh/core/context.py
```diff
@@ -1268,7 +1268,7 @@ def table_diff(
         self,
         source: str,
         target: str,
-        on: t.List[str] | exp.Condition | None = None,
+        on: t.List[str | exp.Identifier] | exp.Condition | None = None,
```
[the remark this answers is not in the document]
Themiscodes: This was to account for quoted identifiers, since when I ran a few tests with quoted identifiers in the grain, like lowercase "id" in Snowflake, these were not handled

The information if they're quoted or not is saved here: [L1326](https://github.com/TobikoData/sqlmesh/blob/ad8193a6f30cdcf1656af0a33ff678649e84be47/sqlmesh/core/context.py#L1326)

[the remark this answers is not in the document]
tobymao: why don't you just have quoted strings?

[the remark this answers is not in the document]
Themiscodes: ok I'll try this instead

sqlmesh/core/context.py
```diff
@@ -1318,7 +1319,14 @@ def table_diff(
             if not on:
                 for ref in model.all_references:
                     if ref.unique:
-                        on = ref.columns
+                        expr = ref.expression
+
+                        if isinstance(expr, exp.Tuple):
+                            on = [key.this.sql() for key in expr.expressions]
```
tobymao, line 1325: is this right? you're generated sqlglot dialect, why are we converting this?

tobymao (approved):

sqlmesh/core/context.py
```diff
@@ -1318,7 +1319,14 @@ def table_diff(
...
+                        elif isinstance(expr, exp.Paren):
+                            on = [expr.this.this.sql()]
```
georgesittas, line 1327: Should this be just `expr.this.sql()`?

sqlmesh/core/context.py
```diff
@@ -1318,7 +1319,14 @@ def table_diff(
...
+                        else:
+                            on = [expr.this.sql()]
```
georgesittas, line 1329: What case does this handle?

georgesittas (approved):
Couple of small comments

sqlmesh/core/context.py
Themiscodes: Yes didn't realise it you're right I'm changing this one now thanks

sqlmesh/core/context.py
```diff
@@ -1321,7 +1322,12 @@ def table_diff(
             if not on:
                 for ref in model.all_references:
                     if ref.unique:
-                        on = ref.columns
+                        expr = ref.expression
+
+                        if isinstance(expr, exp.Tuple):
+                            on = [key.this.sql() for key in expr.expressions]
+                        else:
+                            on = [expr.this.sql()]
```
georgesittas, line 1330: Let's add a test to cover this code path.

Are we sure the only alternative here is an `exp.Paren`? If that's the case, I'd add an `assert isinstance(expr, exp.Paren)` here to enforce this and make it clear what case this `else` is handling.

Themiscodes: it's not only `exp.Paren` the grain might also be defined without a parenthesis so in that case an `exp.Column`, but in both cases `.sql()` will transform them into the correct sql, this is why I removed the conditional. 

This code path is tested in: `tests/integrations/jupyter/test_magics.py::test_table_diff`, but I can also add another test explicitly for that

georgesittas: > it's not only exp.Paren the grain might also be defined without a parenthesis so in that case an exp.Column, but in both cases .sql() will transform them into the correct sql, this is why I removed the conditional.

Sounds good.

> This code path is tested in: tests/integrations/jupyter/test_magics.py::test_table_diff, but I can also add another test explicitly for that

Gotcha, should be fine then. I see now that we diff `sushi.top_waiters` which has a single column for its grain.

georgesittas (approved):
Much cleaner 👍 

Nit: maybe consider adding a small comment on the `else` branch so that it's clear we're dealing with either a `Paren` or a single `Column` there?

PR #2874, opened 2024-07-05
Feat: Support dbt model versioning

izeigerman:
Introduces support for the dbt model versioning feature as described here: https://docs.getdbt.com/docs/collaborate/govern/model-versions

### DIFF
--- a/sqlmesh/dbt/context.py
+++ b/sqlmesh/dbt/context.py
@@ -185,8 +185,11 @@ class DbtContext:
             for model in t.cast(
                 t.Dict[str, t.Union[ModelConfig, SeedConfig]], {**self._seeds, **self._models}
             ).values():
-                self._refs[model.name] = model
+                self._refs[model.alias or model.name] = model
                 self._refs[model.config_name] = model
+                if model.version is not None and model.version == model.latest_version:
+                    self._refs[model.name] = model
+                    self._refs[f"{model.package_name}.{model.name}"] = model
         return self._refs
 
     @property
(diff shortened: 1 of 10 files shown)


tobymao (approved):
what is the consequence of a dbt model version? does it create a different physical table?

izeigerman:
@tobymao for all intents and purposes each version is treated as a separate model. 

This is also consistent with dbt docs:
> How is this different from just creating a new model?
Honestly, it's only a little bit different! There isn't much magic here, and that's by design.

sungchun12:
@tobymao it creates a whole new physical table specific to that version. 


```shell
dbt.exceptions.AmbiguousAliasError: Compilation Error
  dbt found two resources with the database representation "database.schema.model_name".
  dbt cannot create two resources with identical database representations. To fix this,
  change the configuration of one of these resources:
  - model.project_name.model_name.v1 (models/.../model_name.sql)
  - model.project_name.model_name.v2 (models/.../model_name_v2.sql)
```

georgesittas (approved):

sqlmesh/dbt/context.py
```diff
@@ -185,8 +185,11 @@ def refs(self) -> t.Dict[str, t.Union[ModelConfig, SeedConfig]]:
             for model in t.cast(
                 t.Dict[str, t.Union[ModelConfig, SeedConfig]], {**self._seeds, **self._models}
             ).values():
-                self._refs[model.name] = model
+                self._refs[model.alias or model.name] = model
```
github-christophe-oudar, line 188: I think this change introduce a regression:
`ref` can only be used with the model name AFAIK (I tried to use alias and it fails with dbt).
Moving back to `self._refs[model.name] = model` fixes this issue on my end.

izeigerman: What issue are you seeing?

PR #2879, opened 2024-07-08
Feat: add support for parameterization of model names in unit tests

georgesittas:
cc @notgerry

### DIFF
--- a/sqlmesh/utils/yaml.py
+++ b/sqlmesh/utils/yaml.py
@@ -34,8 +34,12 @@ def load(
             source = file.read()
 
     if render_jinja:
-        JINJA_METHODS[VAR] = create_var(variables or {})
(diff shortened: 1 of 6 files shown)


sqlmesh/utils/yaml.py
```diff
@@ -32,6 +34,7 @@ def load(
             source = file.read()
 
     if render_jinja:
```
tobymao, line 36: why are you mutating the constant?

tobymao: perhaps a new dict copy?

georgesittas: Ah whoops, yeah I'll explode the constant within a new dict. Thanks for catching.

tobymao (approved):

PR #2843, opened 2024-07-01
Feat: add mysql operator for airflow scheduler

kaijie-qin:

PR #2881, opened 2024-07-08
fix: allow mysql to not have default catalog behind

eakmanrq:
Fix for this PR: https://github.com/TobikoData/sqlmesh/pull/2843

The issue is that MySQL doesn't support default catalogs and prior to this PR Airflow did require. This now lets users say that they don't want a default catalog which should only apply to MySQL.


docs/integrations/engines/mysql.md
```diff
@@ -37,9 +37,11 @@ By default, the connection ID is set to `mysql_default`, but can be overridden u
 ```python linenums="1"
 sqlmesh_airflow = SQLMeshAirflow(
     "mysql",
-    default_catalog="<database name>",
+    default_catalog="NO_DEFAULT_CATALOG",
```
izeigerman, line 40: Let's make this a named constant instead

izeigerman: I mean let's make sure users use it as well

izeigerman: Maybe let's expose it in `airflow/__init__.py`

sqlmesh/schedulers/airflow/common.py
```diff
@@ -30,6 +30,7 @@
 SNAPSHOT_CLEANUP_COMMAND_XCOM_KEY = "snapshot_cleanup_command"
 
 DEFAULT_CATALOG_VARIABLE_NAME = "sqlmesh_default_catalog"
+NO_DEFAULT_CATALOG_VALUE = "NO_DEFAULT_CATALOG"
```
izeigerman, line 33: Let's not add `_VALUE`

sqlmesh/schedulers/airflow/plugin.py
```diff
@@ -28,14 +31,16 @@ def on_load(cls, *args: t.Any, **kwargs: t.Any) -> None:
...
+        if default_catalog == NO_DEFAULT_CATALOG_VALUE:
+            default_catalog = None
 
         with util.scoped_state_sync() as state_sync:
             try:
                 # If default catalog is not defined then we want to raise an error unless
                 # this is a fresh install since we know nothing needs to be migrated and
                 # the client will prevent making any changes until the default catalog is set.
-                if not default_catalog:
+                if default_catalog == "MISSING":
```
izeigerman, line 43: Do we need `MISSING`? Should we just do 
```
if default_catalog == NO_DEFAULT_CATALOG_VALUE:
            default_catalog = None
```
right after this check?

eakmanrq: I believe the issue would be in the try/except. If we did that then you could end up at `versions = state_sync.get_versions(validate=False)` and throw an exception which would then end up migrating with a default catalog value that is not correct. Therefore you would need to duplicate that logic to convert the no default catalog to none.

izeigerman: ok, let's come up with a more distinct placeholder then and maybe leave a comment. This is a good gotcha and it wasn't obvious to me at all. I can totally see someone using `MISSING` as a catalog name lol.

eakmanrq: Let me know what you think: https://github.com/TobikoData/sqlmesh/pull/2881/commits/2e578b965e88c3ef94210e8dbf72d84d4f2067d8

izeigerman (approved):

PR #2884, opened 2024-07-09
Fix: Invalidate the cached model's rendered query after updating the mapping schema

izeigerman:
This update moves away from caching the unoptimized rendered model query on the file system, as it can lead to caching an invalid or broken query that contains macros referring to schemas of upstream models (e.g., `@STAR`).

Rather than caching an invalid query, we now cache a set of upstream dependencies, which is the only attribute derived from the unoptimized rendered query that is used as part of the loader to construct the DAG before the mapping schemas can be updated and the optimized query cache subsequently refreshed.

### DIFF
--- a/sqlmesh/core/model/cache.py
+++ b/sqlmesh/core/model/cache.py
@@ -59,12 +59,13 @@ class ModelCache:
         return loaded_model
 
 
-class OptimizedQueryCacheEntry(PydanticModel):
-    optimized_rendered_query: exp.Expression
+class RenderedQueryCacheEntry(PydanticModel):
+    optimized_query: t.Optional[exp.Expression]
+    rendered_query: exp.Expression
 
 
-class OptimizedQueryCache:
-    """File-based cache implementation for optimized model queries.
+class RenderedQueryCache:
+    """File-based cache implementation for rendered and optimized model queries.
 
     Args:
         path: The path to the cache folder.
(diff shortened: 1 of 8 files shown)


sqlmesh/core/model/cache.py
```diff
@@ -59,49 +59,55 @@ def get_or_load(self, name: str, entry_id: str = "", *, loader: t.Callable[[], M
         return loaded_model
 
 
-class OptimizedQueryCacheEntry(PydanticModel):
-    optimized_rendered_query: exp.Expression
+class RenderedQueryCacheEntry(PydanticModel):
+    optimized_query: t.Optional[exp.Expression]
```
tobymao, line 63: why is this optional?

sqlmesh/core/model/cache.py
```diff
@@ -14,7 +14,7 @@
 
 class SqlModelCacheEntry(PydanticModel):
     model: SqlModel
-    rendered_query: t.Optional[exp.Expression] = None
+    full_depends_on: t.Set[str]
```
tobymao, line 17: quick note, i've had issues with sets in pydantic because they are unserializable in to_dict()...

sqlmesh/utils/cache.py
```diff
@@ -119,7 +119,7 @@ def put(self, name: str, entry_id: str = "", *, value: T) -> None:
             raise SQLMeshError(f"Cache path '{self._path}' is not a directory.")
 
         with gzip.open(self._cache_entry_path(name, entry_id), "wb", compresslevel=1) as fd:
-            pickle.dump(value.dict(), fd)
+            pickle.dump(value.dict(exclude_none=False), fd)
```
tobymao, line 122: what is this

sqlmesh/core/model/cache.py
izeigerman: Only if it's a complex type. It's fine for primitives.

tobymao (approved):

sqlmesh/utils/cache.py
izeigerman: Without this `OptimizedQueryCacheEntry(optimized_rendered_query=None)` will result in `{}` which somehow messes up deserialization later.

PR #2891, opened 2024-07-10
Fix: control python model loading with enabled flag

Themiscodes:
This fix ensures that models are not loaded when the 'enabled' flag in a Python model is false, mirroring the behaviour of sql models.


georgesittas (approved):
Shold we add a test similar to disabled.sql in sushi for this? Fine either way.

georgesittas (approved):

izeigerman (approved):
Thank you, @Themiscodes ! My oversight

PR #2892, opened 2024-07-10
Chore: improve error reporting when loading seeds

georgesittas:


tobymao (approved):

sqlmesh/core/model/definition.py
```diff
@@ -1663,12 +1663,11 @@ def load_sql_based_model(
...
                 **common_kwargs,
             )
-        except Exception:
-            raise_config_error(
-                "The model definition must either have a SELECT query, a JINJA_QUERY block, or a valid Seed kind",
-                path,
+        except Exception as ex:
+            raise ConfigError(
+                f"The model definition must either have a SELECT query, a JINJA_QUERY block, "
+                f"or a valid Seed kind{' at ' + str(path) if path else ''}. {ex}."
```
izeigerman, line 1669: Why not just include `{ex}` in `raise_config_error`?

sqlmesh/core/model/definition.py
```diff
@@ -1663,12 +1663,11 @@ def load_sql_based_model(
...
             )
```
izeigerman, line 1670: If we're doing `raise ConfigError` then you can also add `from ex`

izeigerman (approved):

PR #2898, opened 2024-07-12
Fix: override python env's variables with unit test vars when present

georgesittas:


sqlmesh/core/renderer.py
```diff
@@ -125,6 +127,13 @@ def _render(
             **kwargs,
         }
 
+        if variables:
+            if c.SQLMESH_VARS not in self._python_env:
+                self._python_env[c.SQLMESH_VARS] = Executable.value({})
+
+            existing_variables = ast.literal_eval(self._python_env[c.SQLMESH_VARS].payload)
+            self._python_env[c.SQLMESH_VARS] = Executable.value(existing_variables | variables)
```
tobymao, line 135: why do you need them in the python env? can't you set locals like we do on line 180? macro_evaluator.locals.update(...)

georgesittas: python env is propagated in both jinja env and the macro evaluator where we later process the variables and add them in locals like you say, so i chose this approach to cover both of 'em at once

tobymao: this looks way hacky-er than just putting them in both variables, can we put them in render_kwargs instead?

georgesittas: Refactored - https://github.com/TobikoData/sqlmesh/pull/2898/commits/04429825c1d858c52c0ef9c133e5cf388e11f3a3

tobymao (approved):

PR #2899, opened 2024-07-12
Fix: Make sure that views are not recreated unnecessarily in the face of concurrent plan application

izeigerman:
This fix ensures that the snapshot view doesn't get recreated unnecessarily when concurrent plans evaluate the same snapshot.

Even though the recreation of a view is not a significant issue for most engines, it leads to negative consequences in Postgres due to the lack of support for late schema binding. As a result, the only way to replace a view in Postgres is through the `DROP CASCADE` statement, which drops all dependent views downstream. When the same view snapshot is evaluated concurrently in different plans, the environment views can be unintentionally dropped (and not recreated).

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -1530,22 +1530,18 @@ class ViewStrategy(PromotableStrategy):
         snapshots = kwargs["snapshots"]
         if (
             (
-                (
-                    isinstance(query_or_df, exp.Expression)
-                    and snapshot.is_materialized_view
-                    and deployability_index.is_deployable(snapshot)
-                    and model.render_query(
-                        snapshots=snapshots,
-                        deployability_index=deployability_index,
-                        engine_adapter=self.adapter,
-                    )
-                    == query_or_df
+                isinstance(query_or_df, exp.Expression)
+                and snapshot.is_materialized_view
+                and deployability_index.is_deployable(snapshot)
+                and model.render_query(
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1577,10 +1574,28 @@ def create(
...
+        if not is_snapshot_deployable and is_table_deployable:
+            # If the snapshot is not deployable, the query may contain references to non-deployable tables or views.
+            # Therefore, we postpone the creation of the deployable view until the snapshot is deployed to production.
+            logger.info(
+                "Skipping creation of the deployable view '%s' for the non-deployable snapshot",
+                table_name,
+            )
+            return
+
+        if self.adapter.table_exists(table_name):
```
tobymao, line 1587: won't this break the previous fix back in november where we always force recreation of views?

izeigerman: The only reason we forced recreation before is because we created a view with wrong (dev) references in the first place. Now when we defer the creation of the deployable view it's no longer a problem.

izeigerman: If I'm not mistaken with have an integration test for this.

tobymao (approved):

sqlmesh/core/snapshot/evaluator.py
izeigerman: Actually it doesn't look like we do. I'll make sure to add one.

PR #2901, opened 2024-07-15
Feat: Add support for defaults in user audits

alexischicoinedatadeveloper:
I was writing an audit and I figured it would be nice to have defaults for the variables.
I tried it and it almost already works, except it was loading the values as strings instead of expressions.

The goal is to be able to define an audit with defaults like this:
audit (
    name generic_sums_add_up,
    defaults (
        percentage_absolute_difference_accepted=0.01,
    )
);

Since defaults has to be a dictionary of string to Expression, it seems to make sense to parse the string values into Expression.

defaults: t.Dict[str, exp.Expression]

I tried the code and it works, but there are some points I'm not sure about.
1. Is it possible to support not needing a trailing comma when there's only one key in the dictionary?
2. Is sqlglot.parse the right way to do it? How would we get the audit's configured dialect and the project's dialect to pass them in to parse?
3. Is it correct to take out the first element of what parse returns? I tried with a default value that was a list like (0.01, 0.02) and it kept the whole list so that seemed fine.

The formatting of the code is probably not right. I saw there's a make file but not sure how to run it is there a page on that you can point me to? Anything on standards for the codebase would also help.

I could also provide a draft for documentation (I know where the docs file is) and some tests (not sure where I'd put them).

### DIFF
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -7,7 +7,6 @@ from functools import cached_property
 from pathlib import Path
 
 from pydantic import Field
-from pydantic_core.core_schema import ValidationInfo
 from sqlglot import exp
 from sqlglot.optimizer.qualify_columns import quote_identifiers
 from sqlglot.optimizer.simplify import gen
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -7,6 +7,7 @@ from functools import cached_property
 from pathlib import Path
 
 from pydantic import Field
+from pydantic_core.core_schema import ValidationInfo
 from sqlglot import exp
 from sqlglot.optimizer.qualify_columns import quote_identifiers
 from sqlglot.optimizer.simplify import gen
@@ -148,7 +149,7 @@ def audit_string_validator(cls: t.Type, v: t.Any) -> t.Optional[str]:
 
 
 @field_validator("defaults", mode="before", check_fields=False)
-def audit_map_validator(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+def audit_map_validator(cls: t.Type, v: t.Any, values: ValidationInfo) -> t.Dict[str, t.Any]:
     if isinstance(v, exp.Paren):
         return dict([_maybe_parse_arg_pair(v.unnest())])
     if isinstance(v, (exp.Tuple, exp.Array)):
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -149,13 +149,13 @@ def audit_string_validator(cls: t.Type, v: t.Any) -> t.Optional[str]:
 
 
 @field_validator("defaults", mode="before", check_fields=False)
-def audit_map_validator(cls: t.Type, v: t.Any, values: ValidationInfo) -> t.Dict[str, t.Any]:
+def audit_map_validator(cls: t.Type, v: t.Any, values: t.Any) -> t.Dict[str, t.Any]:
     if isinstance(v, exp.Paren):
         return dict([_maybe_parse_arg_pair(v.unnest())])
     if isinstance(v, (exp.Tuple, exp.Array)):
         return dict(map(_maybe_parse_arg_pair, v.expressions))
     elif isinstance(v, dict):
-        dialect = values.data.get("dialect")
+        dialect = (values if isinstance(values, dict) else values.data).get("dialect")
         return {
             key: value
             if isinstance(value, exp.Expression)
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -34,6 +34,7 @@ from sqlmesh.utils.metaprogramming import Executable
 from sqlmesh.utils.pydantic import (
     PydanticModel,
     field_validator,
+    get_dialect,
     model_validator,
     model_validator_v1_args,
 )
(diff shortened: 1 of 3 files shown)


sqlmesh/core/audit/definition.py
```diff
@@ -152,7 +152,8 @@ def audit_map_validator(cls: t.Type, v: t.Any) -> t.Dict[str, t.Any]:
     if isinstance(v, (exp.Tuple, exp.Array)):
         return dict(map(_maybe_parse_arg_pair, v.expressions))
     elif isinstance(v, dict):
-        return v
+        return {key: parse(str(value))[0] if not isinstance(value, exp.Expression) else value for key, value in
```
tobymao, line 155:
```suggestion
        return {key: value if isinstance(value, exp.Expression) else d.parse_one(str(value), dialect=values.get("dialect")) for key, value in
```

sqlmesh/core/audit/definition.py
```diff
@@ -152,7 +152,8 @@ def audit_map_validator(cls: t.Type, v: t.Any) -> t.Dict[str, t.Any]:
     if isinstance(v, (exp.Tuple, exp.Array)):
         return dict(map(_maybe_parse_arg_pair, v.expressions))
     elif isinstance(v, dict):
-        return v
```
tobymao, line 155: we need to pass in the pydantic values dictionary, look at how model/meta.py is done, and let's double check that it's passed through. the audit definition should have the model dialect implicitly passed in and then when do values.get("dialect")

achicoine-coveo: ok thanks for the hint I'll look at how that's done

alexischicoinedatadeveloper:
I added the change to parse non expressions in the branch for Tuple and Array, but when I passed in a single element with no trailing comma it came back as a Paren so maybe that's not needed?

tobymao:
well do you need to add the case for Paren?

sqlmesh/core/audit/definition.py
```diff
@@ -603,9 +608,9 @@ def _raise_config_error(msg: str, path: pathlib.Path) -> None:
 
 # mypy doesn't realize raise_config_error raises an exception
 @t.no_type_check
-def _maybe_parse_arg_pair(e: exp.Expression) -> t.Tuple[str, exp.Expression]:
+def _maybe_parse_arg_pair(e: exp.Expression, dialect: str ) -> t.Tuple[str, exp.Expression]:
     if isinstance(e, exp.EQ):
-        return e.left.name, e.right
+        return e.left.name, e.right if isinstance(e.right, exp.Expression) else d.parse_one(str(e.right), dialect=dialect)
```
tobymao, line 613: i'm a bit confused, how is this possible? e.right should never be a non expression, are you sure this needed to be changed?

alexischicoinedatadeveloper: yes you're right I removed this change and the test still passes.

alexischicoinedatadeveloper:
> well do you need to add the case for Paren?

Without addressing the Paren case it throws this error when the trailing comma isn't there.
 sqlmesh.utils.errors.AuditConfigError: Defaults must be a tuple of exp.EQ or a dict at '.'

alexischicoinedatadeveloper:
I saw another part of the code get the dialect like this does it make sense here as well?
https://github.com/TobikoData/sqlmesh/blob/98d18093dae793bc1c0c90b61cf7351d17cada19/sqlmesh/utils/pydantic.py#L78

tobymao:
> I saw another part of the code get the dialect like this does it make sense here as well?
> 
> https://github.com/TobikoData/sqlmesh/blob/98d18093dae793bc1c0c90b61cf7351d17cada19/sqlmesh/utils/pydantic.py#L78

yea i think so, can you use it?

achicoine-coveo:
> > I saw another part of the code get the dialect like this does it make sense here as well?
> > https://github.com/TobikoData/sqlmesh/blob/98d18093dae793bc1c0c90b61cf7351d17cada19/sqlmesh/utils/pydantic.py#L78
> 
> yea i think so, can you use it?

To use the function directly?
from sqlmesh.core.model import model

    dialect = (values if isinstance(values, dict) else values.data).get("dialect")
    return model._dialect if dialect is None else dialect

I'm just not sure what this model part does. Would that part also make sense for this usage? Maybe that would be different than the dialect defined in the audit?

tobymao:
> > > I saw another part of the code get the dialect like this does it make sense here as well?
> > > https://github.com/TobikoData/sqlmesh/blob/98d18093dae793bc1c0c90b61cf7351d17cada19/sqlmesh/utils/pydantic.py#L78
> > 
> > 
> > yea i think so, can you use it?
> 
> To use the function directly?
> 
> ```python
>     from sqlmesh.core.model import model
> 
>     dialect = (values if isinstance(values, dict) else values.data).get("dialect")
>     return model._dialect if dialect is None else dialect
> ```
> 
> I'm just not sure what this model part does. Would that part also make sense for this usage? Maybe that would be different than the dialect defined in the audit?

yea can we try this?

achicoine-coveo:
> > > > I saw another part of the code get the dialect like this does it make sense here as well?
> > > > https://github.com/TobikoData/sqlmesh/blob/98d18093dae793bc1c0c90b61cf7351d17cada19/sqlmesh/utils/pydantic.py#L78
> > > 
> > > 
> > > yea i think so, can you use it?
> > 
> > 
> > To use the function directly?
> > ```python
> >     from sqlmesh.core.model import model
> > 
> >     dialect = (values if isinstance(values, dict) else values.data).get("dialect")
> >     return model._dialect if dialect is None else dialect
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm just not sure what this model part does. Would that part also make sense for this usage? Maybe that would be different than the dialect defined in the audit?
> 
> yea can we try this?

Ok I'll put it in the code. I'll try to see if I can see the value of dialect from model and what happens if it's different from the one in audit.

achicoine-coveo:
There's a test that fails, but I don't see how it's related to the changes.

PR #2903, opened 2024-07-15
Fix: Using macros with IGNORE NULLS

### LINKED ISSUE #2902: Using Macros in Navigation Functions with IGNORE NULLS
Hello there,

I can use a SQLMesh Macro with no issue in a navigation function like so:
```sql
@DEF(test, x -> x);

SELECT FIRST_VALUE(@test(x)) OVER (ORDER BY y) AS column_test
   , x
   , y
FROM test;
```

However, adding `IGNORE NULLS` after the Macro call:
```sql
@DEF(test, x -> x);

SELECT FIRST_VALUE(@test(x) IGNORE NULLS) OVER (ORDER BY y) AS column_test
   , x
   , y
FROM test;
```

I get a `Error: Macro variable '' is undefined.` error.

izeigerman:
Closes #2902

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -241,7 +241,7 @@ def _parse_macro(self: Parser, keyword_macro: str = "") -> t.Optional[exp.Expres
             return field
         return self.expression(MacroVar, this=field.this, comments=comments)
 
-    if isinstance(field, exp.Window):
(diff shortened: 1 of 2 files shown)


erindru (approved):

erindru:
Oh, I think you need to `make py-style` and push again to get the tests to pass

tobymao (approved):

PR #2905, opened 2024-07-15
feat: support upstream/downstream git selector

eakmanrq:
Prior to this change the git branch selector didn't support upstream/downstream selection (using `+`) so this PR adds that functionality. This is part of work to allow the Github Action CI/CD bot to be able to compare against git main instead of prod. Slack context for that change: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1720792390212429

### DIFF
--- a/tests/core/test_selector.py
+++ b/tests/core/test_selector.py
@@ -392,6 +392,7 @@ def test_model_selection_normalized(mocker: MockerFixture, make_snapshot):
     [
         (["git:main"], {'"test_model_a"', '"test_model_c"'}),
         (["git:main & +*model_c"], {'"test_model_c"'}),
+        (["git:main+"], {'"test_model_a"', '"test_model_c"', '"test_model_d"'}),
     ],
 )
 def test_expand_git_selection(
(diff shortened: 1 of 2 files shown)


tests/core/test_selector.py
```diff
@@ -392,6 +392,7 @@ def test_model_selection_normalized(mocker: MockerFixture, make_snapshot):
     [
         (["git:main"], {'"test_model_a"', '"test_model_c"'}),
         (["git:main & +*model_c"], {'"test_model_c"'}),
+        (["git:main+"], {'"test_model_a"', '"test_model_c"', '"test_model_d"'}),
```
izeigerman, line 395: Do we need a test for the leading `+` too?

izeigerman (approved):

tests/core/test_selector.py
eakmanrq: Good call that actually caught a bug

eakmanrq: Addressed: https://github.com/TobikoData/sqlmesh/pull/2905/commits/a5a208158fb7dd116d6d53ed3be6fa1db20ca3f7

PR #2907, opened 2024-07-15
Fix: Quote columns when getting column level lineage

vchan:
When the UI requests a column's lineage, if the column contains unsafe characters, the request fails because the column is not quoted. For example, the column `@foo` would be parsed as a MacroVar in this model query:`SELECT col AS "@foo" FROM my_table`. This PR quotes the specified column when generating the lineage so that it's properly parsed into an identifier.


izeigerman:
What's the context for this? Please add in the description.

tobymao (approved):

tobymao:
there was a bug for the ui on this, but ya, please add a description

vchan:
Description updated.

PR #2908, opened 2024-07-16
Fix!: Ensure Python models are backed by tables and not views

erindru:


tests/core/test_model.py
```diff
@@ -1764,10 +1764,10 @@ def a_model(context):
         path=Path("."),
     )
 
-    assert isinstance(python_model.kind, ViewKind)
+    assert isinstance(python_model.kind, FullKind)
 
     # string kind name specified
-    @model("kind_string", kind="full", columns={'"COL"': "int"})
+    @model("kind_string", kind="external", columns={'"COL"': "int"})
```
izeigerman, line 1770: What's this test? I don't think users suppose to ever set the `external` kind themselves.

erindru: It was more because `full` was the default, so I needed to do another kind to show that it was actually parsing the value and wasn't just falling back to the default.

I first chose `incremental_by_time_range` but that has other mandatory parameters like `time_column` which I couldn't figure out how to represent in the string, so I settled on `external` because that one is simple.

tests/core/test_integration.py
```diff
@@ -1586,17 +1590,7 @@ def _validate_apply(context):
 @pytest.mark.parametrize(
     "from_, to",
     [
-        (ModelKindName.INCREMENTAL_BY_TIME_RANGE, ModelKindName.VIEW),
-        (ModelKindName.INCREMENTAL_BY_TIME_RANGE, ModelKindName.EMBEDDED),
         (ModelKindName.INCREMENTAL_BY_TIME_RANGE, ModelKindName.FULL),
-        (ModelKindName.VIEW, ModelKindName.EMBEDDED),
```
erindru, line 1592: All of these transitions became invalid for this test now that the `FULL` and `EMBEDDED` kinds are disabled for Python models.

This test is tightly coupled with the `sushi.items` model which is a Python model

izeigerman (approved):

PR #2910, opened 2024-07-16
chore: refactor plan to be more maintainable

tobymao:


tests/core/test_plan.py
```diff
@@ -1548,7 +1548,7 @@ def test_indirectly_modified_forward_only_model(make_snapshot, mocker: MockerFix
 
     assert updated_snapshot_a.change_category == SnapshotChangeCategory.BREAKING
     assert updated_snapshot_b.change_category == SnapshotChangeCategory.FORWARD_ONLY
-    assert updated_snapshot_c.change_category == SnapshotChangeCategory.INDIRECT_BREAKING
+    assert updated_snapshot_c.change_category == SnapshotChangeCategory.FORWARD_ONLY
```
izeigerman, line 1551: Let's also make test in which C points at both A and B, in which case, the C's category should indeed be `INDIRECT_BREAKING`.

sqlmesh/core/plan/builder.py
```diff
@@ -180,9 +181,15 @@ def set_choice(self, snapshot: Snapshot, choice: SnapshotChangeCategory) -> Plan
             snapshot: The target snapshot.
             choice: The user decision on how to version the target snapshot and its children.
         """
-        plan = self.build()
-        self._set_choice(snapshot, choice, plan.directly_modified, plan.indirectly_modified)
-        self._adjust_new_snapshot_intervals()
+        if self._forward_only:
```
izeigerman, line 184: Do we also need to check that the category is not assigned to an indirectly modified snapshot?

tobymao: i don't understand this comment, can you explain it to me?

tests/core/test_plan.py
tobymao: done

sqlmesh/utils/dag.py
```diff
@@ -30,7 +31,8 @@ def add(self, node: T, dependencies: t.Optional[t.Iterable[T]] = None) -> None:
             node: The node to add.
             dependencies: Optional dependencies to add to the node.
         """
-        self._sorted = None
+        self._sorted = []
```
izeigerman, line 34: IMHO I preferred the explicit `None` way. This way we have a clear delineation between an empty DAG and missing cache.

sqlmesh/utils/dag.py
```diff
@@ -30,7 +31,8 @@ def add(self, node: T, dependencies: t.Optional[t.Iterable[T]] = None) -> None:
...
+        self._upstream = {}
```
izeigerman, line 35: Ditto

Issue #2829, opened 2024-06-26
Session Parameters - query tag and timezone


achicoine-coveo:
My company is using query tags in Snowflake to do cost reporting. I was trying to find a way to set query tags at the session level, but it looks like that's not supported at the moment. It could also be useful to set them at the model level.

Looking at the python session for Snowflake: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#label-snowflake-connector-methods-connect

timezone can be set in the session.
From doing code with this in the past I believe the query tag has to be set once the session is obtained using an alter session.

Would all session parameters need to be listed and documented or could you do it in a generic way running alter session set key = value for all the extra ones? There's a lot of parameters in Snowflake and even more in spark and I wouldn't expect sqlmesh to support all of them explicitly and with documentation, but passing them through would be great.
https://docs.snowflake.com/en/sql-reference/parameters

Docs on what Sqlmesh supports currently for Snowflake:
https://sqlmesh.readthedocs.io/en/stable/integrations/engines/snowflake/#installation

I believe this is the relevant code:
https://github.com/TobikoData/sqlmesh/blob/9087309155b1eae7641dc9359a35b4632fb143ec/sqlmesh/core/config/connection.py#L317

In the code for dbt query tag is seen, but I believe it falls through the cracks and isn't used.
https://github.com/TobikoData/sqlmesh/blob/c685b18b03f17b27042357662e272b4c72fb413f/sqlmesh/dbt/target.py#L253

Workaround: for now I can set the timezone in the models that need it with a pre-statement, but setting the query tag and other parameters in every model would get annoying and it would be easy to miss one. I would also need the query tags to be on every query in the sessions and not just the ones directly in the models.

izeigerman:
Fixed in #2912

PR #2912, opened 2024-07-17
Feat: pass session parameters for Snowflake

alexischicoinedatadeveloper:
This PR is to implement the features in this issue:
https://github.com/TobikoData/sqlmesh/issues/2829

Looking at the code it didn't seem like much was missing.
Tried it with this in my config:
            session_parameters:
              timezone: UTC
              query_tag: state_test_tag

Not sure how to test it, but I checked in my query history that the queries are getting tagged.
image

From what I understand of the code this should get passed straight through to the connector.
https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect
This would add support for all session parameters supported by Snowflake.
https://docs.snowflake.com/en/sql-reference/parameters

I see here it passes everything defined in the connection kwarg keys.
https://github.com/TobikoData/sqlmesh/blob/7c1ed838295a898da759e59379cdc361b687a22c/sqlmesh/core/config/connection.py#L90
Then that gets passed to Snowflake's connector function and it takes in session_parameters.

I put in a basic test with the others would that be enough?

### DIFF
--- a/docs/integrations/engines/snowflake.md
+++ b/docs/integrations/engines/snowflake.md
@@ -24,6 +24,7 @@ pip install "sqlmesh[snowflake]"
 | `private_key`            | The optional private key to use for authentication. Key can be Base64-encoded DER format (representing the key bytes), a plain-text PEM format, or bytes (Python config only). | string |    N     |
 | `private_key_path`       | The optional path to the private key to use for authentication. This would be used instead of `private_key`.                                                                   | string |    N     |
 | `private_key_passphrase` | The optional passphrase to use to decrypt `private_key` (if in PEM format) or `private_key_path`. Keys can be created without encryption so only provide this if needed.       | string |    N     |
+| `session_parameters`     | The optional session parameters to set for the connection.                                                                                                                     | dict   |    N     |
 
 
 #### Lowercase object names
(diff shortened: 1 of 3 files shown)


docs/integrations/engines/snowflake.md
```diff
@@ -24,6 +24,7 @@ pip install "sqlmesh[snowflake]"
 | `private_key`            | The optional private key to use for authentication. Key can be Base64-encoded DER format (representing the key bytes), a plain-text PEM format, or bytes (Python config only). | string |    N     |
 | `private_key_path`       | The optional path to the private key to use for authentication. This would be used instead of `private_key`.                                                                   | string |    N     |
 | `private_key_passphrase` | The optional passphrase to use to decrypt `private_key` (if in PEM format) or `private_key_path`. Keys can be created without encryption so only provide this if needed.       | string |    N     |
+ | `session_parameters`    | The optional session parameters to set for the connection.                                                                                                                     | dict   |    N     |
```
izeigerman, line 27:
```suggestion
| `session_parameters`    | The optional session parameters to set for the connection.                                                                                                                      | dict   |    N     |
```

sqlmesh/core/config/connection.py
```diff
@@ -333,6 +333,7 @@ class SnowflakeConnectionConfig(ConnectionConfig):
         private_key_passphrase: The optional passphrase to use to decrypt `private_key` or `private_key_path`. Keys can be created without encryption so only provide this if needed.
         register_comments: Whether or not to register model comments with the SQL engine.
         pre_ping: Whether or not to pre-ping the connection before starting a new transaction to ensure it is still alive.
+        session_parameters: The optional session parameters to set for the connection.
```
izeigerman, line 336: I suggest we call this `session_properties`, since this is a terminology that we already use in the model definition.

izeigerman: Ah, this should match the name expected by the snowflake connector. Disregard my comment.

izeigerman:
@alexischicoinedatadeveloper , thanks a lot for the contribution! Make sure to run `make style` and this should be good to go.

izeigerman (approved):
Looks great, thank you!

PR #2916, opened 2024-07-17
Fix: allow overriding built-in datetime vars in tests

### LINKED ISSUE #2914: Broken unit tests when upgrading from 0.110.0 to 0.111.0
I have some unit tests that work in 0.110.0 and not in 0.111.0.

The issue seems related to changes that were made around the handling of time variables.

Here's a model and a test to reproduce.
```sql
MODEL (
  name source.serverless_task_history,
    time_column (start_time)
  )
);

SELECT
  DATE_TRUNC('DAY', start_time)::TIMESTAMPTZ AS record_date,
  start_time::TIMESTAMPTZ,
  TASK_ID::BIGINT,
  task_name::VARCHAR,
  database_name::VARCHAR,
  schema_name::VARCHAR,
  credits_used::DOUBLE
FROM snowflake.account_usage.serverless_task_history
WHERE
  start_time BETWEEN @start_tstz AND @end_tstz
```
Test
``` yaml
test_serverless:
  model: source.serverless_task_history
  inputs:
    snowflake.account_usage.serverless_task_history:
      query: |
        select
        '2024-06-01 01:05:00 +0000'::timestamptz as start_time,
        null as task_id,
        null as task_name,
        null as database_name,
        null as schema_name,
        null as credits_used
  outputs:
    partial: true
    query:
      partial: true
      rows:
      - start_time: 2024-06-01 01:05:00 +0000
        record_date: 2024-06-01 00:00:00 +0000

  vars:
    start_tstz: 2024-06-01 00:00:00+0000
    end_tstz: 2024-06-02 00:00:00+0000
```

Running it with sqlmesh 0.110.0
image

Running it with sqlmesh 0.111.0
image

This seems to come from a different issue with partial where it will give this error when no rows make it through the filter.

Let's bypass the partial issue by adding the columns to the outputs.

The test becomes this:
``` yaml
test_serverless:
  model: source.serverless_task_history
  inputs:
    snowflake.account_usage.serverless_task_history:
      query: |
        select
        '2024-06-01 01:05:00 +0000'::timestamptz as start_time,
        null as task_id,
        null as task_name,
        null as database_name,
        null as schema_name,
        null as credits_used
  outputs:
    partial: true
    query:
      partial: true
      rows:
      - start_time: 2024-06-01 01:05:00 +0000
        record_date: 2024-06-01 00:00:00 +0000
        task_id: null
        task_name: null
        database_name: null
        schema_name: null
        credits_used: null


  vars:
    start_tstz: 2024-06-01 00:00:00+0000
    end_tstz: 2024-06-02 00:00:00+0000
```

I get this error about missing rows:
image

Go back to sqlmesh 0.110.0
image

treysp:
This fixes a bug in #2898 where:
- Not all built-in time macro variables were included in `TIME_KWARG_KEYS`
- All variables in `TIME_KWARG_KEYS` that were NOT overridden by the user were set to a value of `None`

It also fixes a bug in the `partial` setting that allows omitting columns from the test output rows. The current implementation skips subsetting the output DF if there are no ROWS in the intersected input/ouput DFs. This PR corrects it to skip subsetting if there are no COLUMNS in the intersected DFs.

Closes #2914

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -23,7 +23,7 @@ from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.macros import RuntimeStage
 from sqlmesh.core.model import Model, PythonModel, SqlModel
 from sqlmesh.utils import UniqueKeyDict, random_id, type_is_known, yaml
-from sqlmesh.utils.date import pandas_timestamp_to_pydatetime
+from sqlmesh.utils.date import date_dict, pandas_timestamp_to_pydatetime
 from sqlmesh.utils.errors import ConfigError, TestError
 from sqlmesh.utils.yaml import load as yaml_load
 
@@ -32,7 +32,14 @@ if t.TYPE_CHECKING:
 
     Row = t.Dict[str, t.Any]
 
-TIME_KWARG_KEYS = {"start", "end", "execution_time", "latest"}
+TIME_KWARG_KEYS = {
+    "start",
+    "end",
+    "execution_time",
+    "latest",
+    # all built-in datetime macro var names
+    *date_dict(execution_time="1970-01-01", start="1970-01-01", end="1970-01-01").keys(),
+}
 
 
 class ModelTest(unittest.TestCase):
@@ -177,7 +184,7 @@ class ModelTest(unittest.TestCase):
         """Compare two DataFrames"""
         if partial:
             intersection = actual[actual.columns.intersection(expected.columns)]
-            if not intersection.empty:
+            if len(intersection.columns) > 0:
                 actual = intersection
 
         # Two astypes are necessary, pandas converts strings to times as NS,
@@ -548,7 +555,8 @@ class SqlModelTest(ModelTest):
 
     def _render_model_query(self) -> exp.Query:
         variables = self.body.get("vars", {}).copy()
-        time_kwargs = {key: variables.pop(key, None) for key in TIME_KWARG_KEYS}
+        variables_keys = variables.copy().keys()
+        time_kwargs = {key: variables.pop(key) for key in variables_keys if key in TIME_KWARG_KEYS}
 
         query = self.model.render_query_or_raise(
             **time_kwargs,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -177,7 +184,7 @@ def assert_equal(
         """Compare two DataFrames"""
         if partial:
             intersection = actual[actual.columns.intersection(expected.columns)]
-            if not intersection.empty:
+            if len(intersection.columns) > 0:
```
tobymao, line 187: if intersection.columns

sqlmesh/core/test/definition.py
```diff
@@ -548,7 +555,8 @@ def runTest(self) -> None:
 
     def _render_model_query(self) -> exp.Query:
         variables = self.body.get("vars", {}).copy()
-        time_kwargs = {key: variables.pop(key, None) for key in TIME_KWARG_KEYS}
+        variables_keys = variables.copy().keys()
```
tobymao, line 558: you probably don't need the keys() part for x in dict is the same right?

tobymao: why copy this at all? can you just do for key in TIME_KWARG_KEYS?

sqlmesh/core/test/definition.py
treysp: Pandas doesn't like treating `.columns` as truthy:

```
>>> import pandas as pd
>>> temp = pd.DataFrame({})
>>> if temp.columns:
...     print("truthy")
... 
  File "<stdin>", line 1, in <module>
  File "/Users/trey/tobiko/sqlmesh/.venv/lib/python3.11/site-packages/pandas/core/indexes/base.py", line 3170, in __nonzero__
    raise ValueError(
ValueError: The truth value of a RangeIndex is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

Should I use `intersection.columns.any()` instead?

tobymao: nah this is fine, didn't realize it was pandas

sqlmesh/core/test/definition.py
treysp: If we do `{key: variables.pop(key, None) for key in TIME_KWARG_KEYS}`, then every macro variable in it that is NOT overridden gets assigned a value of `None` that's passed to the renderer. 

If a model query uses time variables that were not overridden, they then render with `None` instead of whatever value they would normally have.

treysp: > you probably don't need the keys() part for x in dict is the same right?

yes, fixed

tobymao: should be in {key: variables.pop(key) for key in TIME if key in variables}

treysp: Ahhh - thank you. 

I initially did something similar with the for/if reversed. It hit this error, which I incorrectly assumed would happen if I conditioned directly on `variables` in any way.

```
    time_kwargs = {key: variables.pop(key) for key in variables if key in TIME_KWARG_KEYS}
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: dictionary changed size during iteration
```

tobymao (approved):

georgesittas (approved):

PR #2919, opened 2024-07-18
Feat: warn when state connection uses a non-recommended engine

treysp:


treysp:
@eakmanrq Is this a reasonable description of the mysql/duckdb databases?

`Other state engines with fast and reliable database transactions (less tested than the recommended engines):`

izeigerman (approved):

PR #2920, opened 2024-07-18
Fix: Support unordered columns in test data generation query

Themiscodes:
Currently, the order of columns in input data generated from a SQL query must be identical to the model definition. This update lifts that restriction, to allow to skip or declare columns in random order without issues. This is addressed by adapting the `_add_missing_columns` function, so that aside from filling with nulls the missing values, to maintain the correct order as well.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -503,24 +503,22 @@ class ModelTest(unittest.TestCase):
         if not all_columns or query.is_star:
             return query
 
-        query_columns = set(query.named_selects)
-        missing_columns = [col for col in all_columns if col not in query_columns]
-        if missing_columns:
-            if isinstance(query, exp.Union):
-                if isinstance(query.left, exp.Query):
-                    self._add_missing_columns(query.left, all_columns)
-                if isinstance(query.right, exp.Query):
-                    self._add_missing_columns(query.right, all_columns)
-            else:
-                selects = {select.alias: select for select in query.selects}
-                query.select(
-                    *[
-                        exp.null().as_(col) if col not in query_columns else selects[col]
-                        for col in all_columns
-                    ],
-                    append=False,
-                    copy=False,
-                )
+        if isinstance(query, exp.Union):
+            if isinstance(query.left, exp.Query):
+                self._add_missing_columns(query.left, all_columns)
+            if isinstance(query.right, exp.Query):
+                self._add_missing_columns(query.right, all_columns)
+        else:
+            query_columns = set(query.named_selects)
+            selects = {select.alias: select for select in query.selects}
+            query.select(
+                *[
+                    exp.null().as_(col) if col not in query_columns else selects[col]
+                    for col in all_columns
+                ],
+                append=False,
+                copy=False,
+            )
 
         return query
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -503,10 +503,22 @@ def _add_missing_columns(
         if not all_columns or query.is_star:
             return query
 
-        query_columns = set(query.named_selects)
-        missing_columns = [col for col in all_columns if col not in query_columns]
-        if missing_columns:
-            query.select(*[exp.null().as_(col) for col in missing_columns], copy=False)
+        if isinstance(query, exp.Union):
```
izeigerman, line 506: I'm not sure this would work with multiple nested unions

tests/core/test_test.py
```diff
@@ -1340,18 +1340,22 @@ def test_generate_input_data_using_sql(mocker: MockerFixture, tmp_path: Path) ->
...
       query: |
-        SELECT 1 AS id, 1 AS item_id
+        SELECT 1 AS item_id, 1 AS id
         UNION ALL
-        SELECT 2 AS id, 1 AS item_id
+        SELECT 1 AS item_id, 2 AS id
         UNION ALL
         SELECT 3 AS id, 2 AS item_id
+        UNION ALL
+        SELECT 3 as item_id
```
izeigerman, line 1349: Can we also add:
```
UNION ALL
SELECT 4 as item_id
```

izeigerman: At the same time, I don't get it. This looks like a badly formed SQL. Why support this at all?

Themiscodes: Yes, that's true this is a poorly formed SQL in this test, but it was meant to ensure all select arrangements work. For example, currently if we do `SELECT 1 as item_id` this would generate `SELECT 1 as id, null as item_id,` so this fix aims to address this and produce the correct sql when the order is random.

So the main purpose is to support skipping columns, when a table has many and the user only needs to test a few, so that they don't have to explicitly define all columns in the right order.

sqlmesh/core/test/definition.py
Themiscodes: I'll try to see if it works with nested unions or adjust the logic

izeigerman: I feel like we don't need to traverse the Union but instead check the schema of one of its queries and then just put the whole thing into the subquery.

Additionally, I don't understand why do we need to do this at the query level at all and not just add columns to a dataframe we get after evaluating it.

tests/core/test_test.py
georgesittas: Yeah not sure if we want to support this pattern, it may make the implementation more complex. Let's just deal with reasonable inputs. I think one fair expectation is that if you skip some columns in one query of a union, you'll skip it in all of them.

sqlmesh/core/test/definition.py
georgesittas: > why do we need to do this at the query level at all instead of just adding columns to a dataframe we get after evaluating it.

+1 to this if we can do it, seems simpler to implement using existing logic.

Themiscodes: Fair points I adapted the logic to use the previous one for filling the nulls and handling the order during the creation of the view

tests/core/test_test.py
Themiscodes: Ok adapted the logic to have this expectation from the user

Themiscodes:
Addressed comments, reverted to the previous logic for `add_missing_columns` and adjusted the columns in the dictionary used for the create_view. This allows the columns to be in random order or skipped entirely, with the assumption that if a column is skipped in one query it will be skipped in all of them.

izeigerman (approved):

PR #2922, opened 2024-07-18
Chore!: Update managed table docs

erindru:
Clarify the downside of using normal tables for dev previews of managed tables

### DIFF
--- a/docs/concepts/models/managed_models.md
+++ b/docs/concepts/models/managed_models.md
@@ -10,7 +10,7 @@ Due to this, managed models would typically be built off an [External Model](./e
 ## Difference from materialized views
 The difference between an Managed model and a materialized view is down to semantics and in some engines there is no difference.
 
-SQLMesh has support for [materialized views](./model_kinds#materialized-views) already. However, depending on the engine, these are subject to some limitations, such as:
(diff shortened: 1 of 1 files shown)


docs/concepts/models/managed_models.md
```diff
@@ -34,6 +34,11 @@ However, there is usually extra vendor-imposed costs associated with Managed mod
 
 Therefore, we try to not create managed tables unnecessarily. For example, in [forward-only plans](../plans.md#forward-only-change) we just create a normal table to preview the changes and only re-create the managed table on deployment to prod.
 
+!!! warning
+    Due to the use of normal tables for dev previews, it is possible to write a query that uses features that are available to normal tables in the target engine but not managed tables. This could result in a scenario where a plan works in a virtual environment but fails when deployed to production.
```
eakmanrq, line 38: "plan works in a virtual environment"

Shouldn't this be "plan works in a dev environment" since all environments are "virtual"?

erindru: Oh yeah, good point. I tend to think of prod as a "the actual environment" but you're right, its just another virtual environment

eakmanrq (approved):

PR #2926, opened 2024-07-22
Chore!: Refactor how engine adapter integration tests are run

erindru:
This PR does the following:

- Split out the docker engine integration tests from one big task into multiple tasks (one per engine) and enable them for all PR's. They take less time to run than `style_and_slow_tests` and running them on all PR's should help increase confidence that changes haven't broken any of the docker engines
- Publish the pytest results to CircleCI
- Make it easy to run the tests for a specific engine locally (eg you can `git clone` sqlmesh and then immediately run `make spark-test` and have things Just Work™)
- Re-enable cloud database tests (BigQuery, Databricks, Snowflake, Redshift) for `main`. Note that these rely on some utility scripts to set up / tear down databases for the tests to run in, because we cant just spin up fresh new instances of cloud database engines like we can for the docker engines
- Address some flakiness

Here is an example workflow run that shows the cloud database tests running: https://app.circleci.com/pipelines/gh/TobikoData/sqlmesh/14490/workflows/9d6bbb49-6fca-46d4-a4b5-6e8e6de84f61
(since I had to add the `main` branch filter back in for this PR)


.circleci/continue_config.yml
```diff
@@ -196,32 +196,32 @@ jobs:
...
-          name: Bring up Dockerized Engines
-          command: make engine-up
-      - run:
-          name: Make sure DBs are ready
-          command: sleep 60
-      - run:
-          name: Run tests
-          command: make engine-docker-test
-          no_output_timeout: 30m
+  # engine_adapter_docker_tests:
```
izeigerman, line 199: Do we want to leave this commented out?

izeigerman: Sorry, I ignored the fact that it was WIP

erindru: Eventually there wont be big commented blocks but i'm still messing around with stuff at the moment :)

tests/core/engine_adapter/test_integration.py
```diff
@@ -856,18 +867,8 @@ def test_ctas(ctx: TestContext):
         table_description = ctx.get_table_comment(table.db, "test_table")
         column_comments = ctx.get_column_comments(table.db, "test_table")
 
-        # Trino on Hive COMMENT permissions are separate from standard SQL object permissions.
```
erindru, line 859: @treysp this didnt seem to be a problem anymore in Trino 452, the column comments were returned as expected

tests/core/engine_adapter/test_integration.py
```diff
@@ -2183,30 +2175,14 @@ def test_dialects(ctx: TestContext):
...
-                #
-                # They are adding a non-aware version TIMESTAMP_NTZ that's currently in public preview -
-                # you have to specify a table option to use it:
-                # "Feature support is enabled automatically when you create a new Delta table with a column of
-                # TIMESTAMP_NTZ type. It is not enabled automatically when you add a column of
-                # TIMESTAMP_NTZ type to an existing table.
-                # To enable support for TIMESTAMP_NTZ columns, support for the feature must be explicitly enabled for
-                # the existing table."
-                # https://docs.databricks.com/en/sql/language-manual/data-types/timestamp-ntz-type.html
-                "databricks": pd.Timestamp("2020-01-01 00:00:00+00:00"),
```
erindru, line 2208: @eakmanrq on the Databricks Runtime 15.3 (which is the runtime i'm using for CircleCI to run the tests on), this override didnt seem to be needed anymore.

The default `{"default": pd.Timestamp("2020-01-01 00:00:00")}` seems to work now.

However, have I inadvertently targeted a Databricks runtime that we dont support?

tests/core/engine_adapter/docker/trino/catalog/datalake.properties
```diff
@@ -6,10 +6,6 @@ hive.s3.aws-access-key=minio
 hive.s3.aws-secret-key=minio123
 hive.metastore-cache-ttl=0s
 hive.metastore-refresh-interval=5s
-hive.metastore-timeout=10s
-hive.allow-drop-table=true
-hive.allow-add-column=true
-hive.allow-drop-column=true
-hive.allow-rename-column=true
-hive.allow-rename-table=true
+hive.metastore.thrift.client.connect-timeout=10s
```
erindru, line 9: This just fixes some properties that were dropped / renamed between Trino 428 and Trino 452

setup.py
```diff
@@ -54,10 +54,7 @@
             "google-cloud-bigquery[pandas]",
             "google-cloud-bigquery-storage",
         ],
-        "databricks": [
-            "databricks-sql-connector",
-            "databricks-cli",
-        ],
+        "databricks": ["databricks-sql-connector", "databricks-connect"],
```
erindru, line 57: @eakmanrq I removed `databricks-cli` because its [no longer supported by Databricks](https://docs.databricks.com/en/dev-tools/cli/index.html#information-for-legacy-databricks-cli-users) and I needed the new version to use from the utility scripts.

Also, do you see an issue adding `databricks-connect` here? It was required by the integration tests because they deal with DataFrame's

eakmanrq: You can't include `databricks-connect`. The reason is that the user needs to use a specific version of `databricks-connect` based on their databricks runtime. For example if they are targetting a cluster with runtime 13.3 then they need to use `databricks-connect==13.3.*`

tests/core/engine_adapter/test_integration.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         test_type: str,
         engine_adapter: EngineAdapter,
         gateway: str,
+        is_cloud_database: bool = False,
```
eakmanrq, line 48: The parameters are already marked as being `remote` or `local`. Could `remote` mark be used instead of adding this?

tests/core/engine_adapter/test_integration.py
eakmanrq: I think it is fine then to remove this override

.circleci/manage-test-db.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
```
eakmanrq, line 1: I'm in the process of reviewing all the changes, but what happens if I run the integration tests directly? Does it depend on running this script first?

erindru: Nope, when you run directly it doesn't create a test database for you, that's just for CircleCI.

It'll use whatever you set for the `<engine>_DATABASE` environment variable (assuming you're using the default config, otherwise you can of course override it to whatever you need).

For the docker databases, it does however run the `wait-for-db.sh` script which relies on `docker` and `netcat` being installed.

setup.py
erindru: Thanks, I thought something like that would be the case. I've shifted the installation of `databricks-connect` to the `install-prerequisites.sh` script that CircleCI runs before triggering tests

tests/core/engine_adapter/test_integration.py
erindru: This was actually powered by the `remote` mark already. It was more convienient to make it a property of the `TestContext` vs injecting the `request` fixture everywhere it was needed and then calling `request.node.get_closest_marker("remote") is not None` to figure out if the `remote` mark was set.

However, i've renamed the property on `TestContext` from `is_cloud_database` to `is_remote`, is that better?

.circleci/manage-test-db.sh
erindru: btw - the way I was doing this locally is create a file `.env` at the root of the repo containing the env vars:

```
SNOWFLAKE_ACCOUNT=<redacted>
SNOWFLAKE_WAREHOUSE=small
SNOWFLAKE_DATABASE=erin
SNOWFLAKE_USER=<redacted>
SNOWFLAKE_PASSWORD=<redacted>

DATABRICKS_CATALOG=erin
DATABRICKS_SERVER_HOSTNAME=<redacted>
DATABRICKS_HTTP_PATH=<redacted>
DATABRICKS_ACCESS_TOKEN=<redacted>

REDSHIFT_HOST=<redacted>
REDSHIFT_PORT=5439
REDSHIFT_USER=<redacted>
REDSHIFT_PASSWORD=<redacted>
REDSHIFT_DATABASE=erin

BIGQUERY_KEYFILE=<redacted>
```

VSCode picks this up automatically so no extra config needed if running tests from VSCode.

To run the tests for eg Snowflake, you can do the following:

```
$ export $(cat .env | xargs)
$ make snowflake-test
```

For the docker databases it's even easier, just run:

```
$ make trino-test
```

Which will ensure the correct dependencies are installed, the containers are started and available and then kick off the tests

eakmanrq (approved):

PR #2927, opened 2024-07-22
Docs: clarify macro variables docs, add examples

treysp:
And revert change from mkdocs `tabbed` to `blocks` extension (which broke tabbed code blocks)


z3z1ma:
Recently learned `@this_model` works in hooks in SQL models. This is useful for cases when you are calling an macro. This should be documented. 

```python
MODEL ();

SELECT 1;

@do_something(@this_model);
```

Passing in the model name instead would both not be dry and also it's parsed into an exp.Column by the parser which means the model name is not swapped with the physical name. We should document this common use case. It's easier than resolving the physical name every time in the macro body.

PR #2930, opened 2024-07-23
chore: docs expand on state connection example

eakmanrq:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1721764730631289


eakmanrq:
Using this instead: https://github.com/TobikoData/sqlmesh/pull/2931

PR #2938, opened 2024-07-24
Chore: add persistent duckdb gateway to sushi and wursthall examples

treysp:
These changes allow wursthall to use a persistent duckdb database with the CLI `--gateway duckdb_persistent` option.

They also modify sushi so its persistent duckdb config is specified with the same `--gateway duckdb_persistent` option (rather than the current `--config local_config` option).

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -62,12 +62,6 @@ test_config = Config(
     model_defaults=model_defaults,
 )
 
-# A stateful DuckDB config.
-local_config = Config(
-    default_connection=DuckDBConnectionConfig(database=f"{DATA_DIR}/local.duckdb"),
-    model_defaults=model_defaults,
-)
-
 airflow_config = Config(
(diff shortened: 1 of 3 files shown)


examples/sushi/config.py
```diff
@@ -29,9 +29,17 @@
 model_defaults_iceberg = ModelDefaultsConfig(**defaults, storage_format="iceberg")
 
 
-# An in memory DuckDB config.
+# A DuckDB config, in-memory by default.
```
tobymao, line 32: there's already a local_config, should we remove that or if not, why do have it here too?

treysp: good catch - removed in [7505ace](https://github.com/TobikoData/sqlmesh/pull/2938/commits/7505ace0827269ae4293e369c4eb4236ba315705)

tobymao (approved):

PR #2940, opened 2024-07-24
Chore: Add Missing ConnectionConfigs to core.config init

kam-cheng-emis:
This commit adds two missing ConnectionConfig subclasses to the sqlmesh.core.config __init__.py file:
- BaseDuckDBConnectionConfig
- TrinoConnectionConfig

This allows for easier referencing of the above, and is consistent with the manner in which other ConnectionConfig subclasses are imported.


izeigerman (approved):
Thanks!

PR #2942, opened 2024-07-24
Feat!: Expose Snowflake dynamic tables through the DBT adapter

erindru:
The initial implementation of Managed Models was for sqlmesh projects only.

This maps `materialized: dynamic_table` for Snowflake DBT projects to a managed model so dynamic tables can be used in DBT projects

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -19,6 +19,7 @@ from sqlmesh.core.model import (
     ModelKind,
     SCDType2ByColumnKind,
     ViewKind,
+    ManagedKind,
     create_sql_model,
 )
 from sqlmesh.core.model.kind import SCDType2ByTimeKind, OnDestructiveChange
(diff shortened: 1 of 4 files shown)


sqlmesh/dbt/model.py
```diff
@@ -399,8 +407,20 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
+            if self.snowflake_warehouse is not None:
+                model_kwargs["session_properties"] = {"warehouse": self.snowflake_warehouse}
+
+            if self.model_materialization == Materialization.DYNAMIC_TABLE:
+                if not self.snowflake_warehouse:
+                    raise ConfigError("`snowflake_warehouse` must be set for dynamic tables")
+                if not self.target_lag:
+                    raise ConfigError("`target_lag` must be set for dynamic tables")
+
+                model_kwargs["physical_properties"] = {
```
izeigerman, line 420: Don't we override here what was set on line 408?

erindru: No? Line 408 only runs `if context.target.dialect == "bigquery"`

izeigerman (approved):

PR #2943, opened 2024-07-24
Update sqlmesh_macros.md

nickmuoh:
Updated python macro example with a slight caveat when passing strings with spaces to the macro.

### DIFF
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -1594,11 +1594,30 @@ Usage in SQLMesh:
 
 ```sql linenums="1"
 SELECT
-  @repeat_string('SQLMesh ', 3) as repeated_string
+  @repeat_string('SQLMesh_', 3) as repeated_string
 FROM some_table;
 ```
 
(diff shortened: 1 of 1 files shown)


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1594,12 +1594,31 @@ Usage in SQLMesh:
...
 ```sql linenums="1"
 SELECT
-  @repeat_string('SQLMesh ', 3) as repeated_string
+  @repeat_string('SQLMesh_', 3) as repeated_string
 FROM some_table;
 ```
 
 This macro takes two arguments: `text` of type `str` and `count` of type `int`, and it returns a string. Without type hints, the inputs to the macro would have been two `exp.Literal` objects you would have had to convert to strings and integers manually.
 
+If the string you pass to the macro contains spaces then you need to wrap it in a `exp.Literal` string:
```
tobymao, line 1603: let's delete this whole paragraph and just change the macro to use exp.Literal.string

nickmuoh:
Closing this PR has it has already been addressed and merged here https://github.com/TobikoData/sqlmesh/pull/2946

PR #2944, opened 2024-07-25
feat: add GenericNotificationTarget to NotificationTargets

kam-cheng-emis:
The GenericNotificationTarget object is a subclass of BaseNotificationTarget, and will offer users greater freedom in creating their own custom notifications.

This will be useful for users who do want to use the currently available methods (smtp, slack_webhook, slack_api).

### DIFF
--- a/sqlmesh/core/notification_target.py
+++ b/sqlmesh/core/notification_target.py
@@ -394,9 +394,36 @@ class BasicSMTPNotificationTarget(BaseTextBasedNotificationTarget):
         return all((self.host, self.user, self.password, self.sender))
 
 
+class GenericNotificationTarget(BaseNotificationTarget):
+    """A generic notification target that can be used to create custom notification targets.
+
+    This target is not meant to be used directly, but rather as a base class for custom notification targets.
+
+    The `send` method should be overridden to provide the actual notification functionality.
+
+    Example:
+    ```python
+    class MyCustomNotificationTarget(GenericNotificationTarget):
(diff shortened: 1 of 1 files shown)


sqlmesh/core/notification_target.py
```diff
@@ -396,12 +398,13 @@ def is_configured(self) -> bool:
 
 NotificationTarget = Annotated[
     t.Union[
+        BaseNotificationTarget,
         BasicSMTPNotificationTarget,
         ConsoleNotificationTarget,
         SlackApiNotificationTarget,
         SlackWebhookNotificationTarget,
     ],
-    Field(discriminator="type_"),
```
kam-cheng-emis, line 404: [Pydantic Docs on Union/discriminators](https://docs.pydantic.dev/latest/concepts/unions/#union-validation-errors)

Changing from discriminator to union_mode "left_to_right" allows us to add BaseNotificationTarget and remove the error: `TypeError: Value 'smtp' for discriminator 'type_' mapped to multiple choices`

vchan:
Thank you for the PR, @kam-cheng-emis! How about adding a generic NotificationTarget class that subclasses BaseNotificationTarget?
```python
class GenericNotificationTarget(BaseNotificationTarget):
    type_: Literal["generic"] = Field(alias="type", default="generic")
```
You can call it simple or generic or whatever makes sense. Then we don't need to change the discriminator.

vchan (approved):
Thanks for the contribution!

Issue #2567, opened 2024-05-06
Enable audits in model defaults


AronsonDan:
The scenario goes as follows:

1. table which is defined as an external source was ingested and serves as a base table
2. that table has some meta fields that i would like to drag along downstream the models
3. Would love to test all across the transformations in that model that the columns exists and to fail a plan in case it dosent

suggestion:
enable model level audits 
```yaml
model_defaults:
  dialect: snowflake
  audits:
    - assert_base_columns
```

PR #2947, opened 2024-07-25
Feat: Support enabling audits in model defaults

Themiscodes:
This update enables audits in model defaults and to be applied globally. The syntax for each respective audit is similar to the model definition, but is used in the YAML file:

```yaml
model_defaults:
    audits: 
        - assert_positive_order_ids
        - does_not_exceed_threshold(column := id, threshold := 1000)
```

fixes: #2567

### DIFF
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -69,7 +69,6 @@ class ModelDefaultsConfig(BaseConfig):
                 kwargs[arg.left.name.lower()] = arg.right
             return func.lower(), kwargs
 
-        breakpoint()
         if isinstance(v, list):
             return [extract(parse(audit)[0]) for audit in v]
 
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import typing as t
 
-from sqlglot import exp
-from sqlmesh.core.dialect import parse
+from sqlmesh.core.dialect import parse, extract_audit
 from sqlmesh.core.config.base import BaseConfig
 from sqlmesh.core.model.kind import (
     ModelKind,
@@ -14,7 +13,6 @@ from sqlmesh.core.model.kind import (
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.core.model.meta import AuditReference
 from sqlmesh.utils.pydantic import field_validator
-from sqlmesh.utils.errors import ConfigError
 
 
 class ModelDefaultsConfig(BaseConfig):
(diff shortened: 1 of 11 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -461,7 +461,11 @@ def ctas_query(self, **render_kwarg: t.Any) -> exp.Query:
             )
         return query
 
-    def referenced_audits(self, audits: t.Dict[str, ModelAudit]) -> t.List[ModelAudit]:
+    def referenced_audits(
+        self,
+        audits: t.Dict[str, ModelAudit],
+        config_audits: t.Optional[t.List[AuditReference]] = None,
```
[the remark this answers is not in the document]
Themiscodes: Right `default_audits` is a more fitting name thank you!

sqlmesh/core/model/definition.py
```diff
@@ -461,7 +461,11 @@ def ctas_query(self, **render_kwarg: t.Any) -> exp.Query:
             )
         return query
 
-    def referenced_audits(self, audits: t.Dict[str, ModelAudit]) -> t.List[ModelAudit]:
+    def referenced_audits(
+        self,
+        audits: t.Dict[str, ModelAudit],
+        default_audits: t.Optional[t.List[AuditReference]] = None,
```
erindru, line 467: Small nit: instead of making this default to `None`, is there any reason why it couldn't default to `[]` instead?

Then below where you have:
```
self.audits + default_audits if default_audits else self.audits
```

It could become:
```
self.audits + default_audits
```

Themiscodes: yes this makes sense changed it

z3z1ma:
I would love to apply a global audit to all `grains` that asserts uniqueness, highly useful and far less repetition...
Can this be used to do that? I'd imagine you could pass `@this_model` into a python macro call which returns SQL for the uniqueness test? Not "first-class" persay but feasible?

erindru (approved):

sqlmesh/core/config/model.py
```diff
@@ -37,6 +40,14 @@ class ModelDefaultsConfig(BaseConfig):
...
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
+    audits: t.Optional[t.List[AuditReference]] = None
 
     _model_kind_validator = model_kind_validator
     _on_destructive_change_validator = on_destructive_change_validator
+
+    @field_validator("audits", mode="before")
+    def _audits_validator(cls, v: t.Any) -> t.Any:
+        if isinstance(v, list):
+            return [extract_audit(parse(audit)[0]) for audit in v]
```
tobymao, line 51: should you use parse_one here?

Themiscodes: yes adapted this to use `parse_one`

tobymao (approved):

Themiscodes:
> I would love to apply a global audit to all `grains` that asserts uniqueness, highly useful and far less repetition... Can this be used to do that? 

yes, I think that's possible @z3z1ma, though I haven't tried it myself yet. essentially, this allows you to enable an audit to run globally, so including an audit defined in the `audits` directory that asserts uniqueness will work the same way as if you referenced it in each separate model of your project

PR #2960, opened 2024-07-29
More SQLMesh Native Macros

sungchun12:
I want all these quality of life ergonomics to be built into the SQLMesh experience. When a user is like, “darn, I’m too lazy to write this out. I miss how dbt-utils made me feel.” 

We can reply so elegantly, “here you go: `@deduplicate(demo.table, [user_id, cast(timestamp as date)], ['timestamp desc', 'status asc'])` ” and give a little nod of alignment, have a little smirk and then move on with our lives.

I’ll likely make a company blog about it after the ones I make are released in a new version. 

**dbt-utils macros to port to SQLMesh:**
**deduplicate**: https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#deduplicate-source


sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,92 @@ def var(
...
+        >>> from sqlglot.schema import MappingSchema
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "@deduplicate(demo.table, [user_id, cast(timestamp as date)], ['timestamp desc', 'status asc'])"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        'SELECT * FROM demo.table QUALIFY ROW_NUMBER() OVER (PARTITION BY user_id, CAST(timestamp AS DATE) ORDER BY timestamp DESC, status ASC NULLS LAST) = 1'
+    """
+
+    partition_clause = exp.Tuple(
+        expressions=[
+            col if not isinstance(col, exp.Cast) else exp.Cast(this=col.this, to=col.args.get("to"))
```
tobymao, line 1128: always use builder methods when possible,

exp.func("cast", ...)

exp.column(...) etc

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,92 @@ def var(
...
+        partition_by: column names, or expressions to use to identify a window of rows out of which to select one as the deduplicated row
+        order_by: A list of strings representing the ORDER BY clause
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlglot.schema import MappingSchema
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "@deduplicate(demo.table, [user_id, cast(timestamp as date)], ['timestamp desc', 'status asc'])"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        'SELECT * FROM demo.table QUALIFY ROW_NUMBER() OVER (PARTITION BY user_id, CAST(timestamp AS DATE) ORDER BY timestamp DESC, status ASC NULLS LAST) = 1'
```
tobymao, line 1123: i don't really see the value of this macro, can't you write the SQL as is? it's relatively simple.

sungchun12: Fair question. Its goal is to be empathetic to those migrating from dbt to SQLMesh and want a similar UX to the dbt-utils macro system they're used to. It prevents people from having to google window function syntax each time.

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,94 @@ def var(
...
+    order_clause = exp.Order(expressions=order_expressions)
+
+    window_function = exp.Window(
+        this=exp.RowNumber(), partition_by=partition_clause, order=order_clause
+    )
+
+    first_unique_row = exp.EQ(this=window_function, expression=exp.Literal.number(1))
+
+    query = (
+        exp.Select(expressions=[exp.Star()])
```
tobymao, line 1163: exp.select("*")

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,94 @@ def var(
...
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlglot.schema import MappingSchema
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "@deduplicate(demo.table, [user_id, cast(timestamp as date)], ['timestamp desc', 'status asc'])"
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        'SELECT * FROM demo.table QUALIFY ROW_NUMBER() OVER (PARTITION BY user_id, CAST(timestamp AS DATE) ORDER BY timestamp DESC, status ASC NULLS LAST) = 1'
+    """
+
+    partition_clause = exp.Tuple(
```
tobymao, line 1126: exp.tuple_

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,94 @@ def var(
...
+            col
+            if not isinstance(col, exp.Cast)
+            else exp.func("cast", this=col.this, to=col.args.get("to"))
+            for col in partition_by
+        ]
+    )
+
+    order_expressions = []
+    for order_item in order_by:
+        parts = order_item.split()
```
tobymao, line 1137: delete everything  here, it should just be evaluator.parse_one(order, into=exp.Ordered)

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,94 @@ def var(
...
+    if not order_expressions:
+        raise SQLMeshError("At least one order_by expression is required: '<column> <asc|desc>'")
+
+    order_clause = exp.Order(expressions=order_expressions)
+
+    window_function = exp.Window(
+        this=exp.RowNumber(), partition_by=partition_clause, order=order_clause
+    )
+
+    first_unique_row = exp.EQ(this=window_function, expression=exp.Literal.number(1))
```
tobymao, line 1160: window_function.eq(1)

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,94 @@ def var(
...
+    query = (
+        exp.Select(expressions=[exp.Star()])
+        .from_(relation, dialect=evaluator.dialect)
+        .qualify(first_unique_row)
+    )
+
+    return query
+
+
+@macro("date_spine")
```
tobymao, line 1171: let's delete these unimplement ones and only do one at a time

sungchun12: so 3 prs in total?

tobymao: yes, that's ideal

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
sungchun12: updated

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -901,6 +901,42 @@ FROM rides
...
+would be rendered as:
+
+```sql linenums="1"
+WITH "raw_data" AS (                                                                           
+  SELECT                                                                                       
+    *                                                                                          
+  FROM "my_table" AS "my_table"                                                                
+  QUALIFY                                                                                      
+    ROW_NUMBER() OVER (PARTITION BY "id", CAST("event_date" AS DATE) ORDER BY "event_date"     
+DESC, "status" ASC) = 1
```
georgesittas, line 932: nit: is this the actual render output? Formatting looks a bit weird.

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,65 @@ def var(
...
+        )
+
+    if not isinstance(order_by, list):
+        raise SQLMeshError("order_by must be a list of strings: ['<column> <asc|desc>']")
+
+    partition_clause = exp.tuple_(
+        *[
+            col
+            if not isinstance(col, exp.Cast)
+            else exp.func("cast", this=col.this, to=col.args.get("to"))
```
georgesittas, line 1137: Isn't this `func("cast", ...)` redundant given that we know `col` is a cast instance already? I think we can just use `partition_by` as is when creating the tuple.

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,65 @@ def var(
...
+            for col in partition_by
+        ]
+    )
+
+    order_expressions = [
+        evaluator.parse_one(order_item, into=exp.Ordered) for order_item in order_by
```
georgesittas, line 1143: I think we need to set dialect here because if there's no `ASC` or `DESC`, we'll use the wrong default. Different dialects specify different default orders (nulls first/last is also affected)

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,65 @@ def var(
...
+        >>> MacroEvaluator().transform(parse_one(sql)).sql()
+        'SELECT * FROM demo.table QUALIFY ROW_NUMBER() OVER (PARTITION BY user_id, CAST(timestamp AS DATE) ORDER BY timestamp DESC, status ASC) = 1'
+    """
+    if not isinstance(partition_by, list):
+        raise SQLMeshError(
+            "partition_by must be a list of columns: [<column>, cast(<column> as <type>)]"
+        )
+
+    if not isinstance(order_by, list):
+        raise SQLMeshError("order_by must be a list of strings: ['<column> <asc|desc>']")
```
georgesittas, line 1131: what about null ordering here? can you specify nulls first/last it in the qualify clause? if so let's maybe update the error message here

tests/core/test_macros.py
```diff
@@ -663,3 +663,169 @@ def test_macro_first_value_ignore_respect_nulls(assert_exp_eq) -> None:
...
+        ) _t
+        WHERE _w = 1
+        """,
+        ),
+    ],
+)
+def test_deduplicate(assert_exp_eq, dialect, sql, expected_sql):
+    schema = MappingSchema({}, dialect=dialect)
+    evaluator = MacroEvaluator(schema=schema, dialect=dialect)
+    assert_exp_eq(evaluator.transform(parse_one(sql)), expected_sql, dialect=dialect)
```
georgesittas, line 808:
```suggestion
    assert_exp_eq(evaluator.transform(parse_one(sql, dialect=dialect)), expected_sql, dialect=dialect)
```

tests/core/test_macros.py
```diff
@@ -663,3 +663,169 @@ def test_macro_first_value_ignore_respect_nulls(assert_exp_eq) -> None:
         "SELECT FIRST_VALUE(@test(x) RESPECT NULLS) OVER (ORDER BY y) AS column_test"
     )
     assert_exp_eq(evaluator.transform(actual_expr), expected_sql, dialect="duckdb")
+
+
+@pytest.mark.parametrize(
```
georgesittas, line 668: There are basically two expected outputs in this test - those that include `QUALIFY` and those that don't. I'd consider refactoring it to compact it a bit because there's a lot of duplication now. We can partition by dialect that supports `QUALIFY` vs not to control the expected output.

docs/concepts/macros/sqlmesh_macros.md
sungchun12: yes it does when I render an example sql model.
image

sqlmesh/core/macros.py
sungchun12: alright, I'll remove it.

docs/concepts/macros/sqlmesh_macros.md
georgesittas: Perhaps it happens due to the terminal width. To be clear, I'm talking about the `"event_date" DESC, 1223) = 1` being in a separate line.

sungchun12: ahh yes, it's due to terminal width. here it is with a wider terminal

image

sungchun12: I'll update the example in this markdown so people don't get confused with line wrapping.

sqlmesh/core/macros.py
sungchun12: will do

sqlmesh/core/macros.py
sungchun12: I updated this so it's nulls first ordering by default to mirror the dbt-utils macro: https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/deduplicate.sql

this is done by default in the Ordered Class
```python
class Ordered(Expression):
    arg_types = {"this": True, "desc": False, "nulls_first": True, "with_fill": False}
```

docs/concepts/macros/sqlmesh_macros.md
georgesittas: sgtm 👍

sqlmesh/core/macros.py
sungchun12: updated the error message

tests/core/test_macros.py
sungchun12: updated with my own commit

tests/core/test_macros.py
sungchun12: done

sungchun12: redshift is separated out due to its quoting

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,61 @@ def var(
...
+
+    order_clause = exp.Order(expressions=order_expressions)
+
+    window_function = exp.Window(
+        this=exp.RowNumber(), partition_by=partition_clause, order=order_clause
+    )
+
+    first_unique_row = window_function.eq(1)
+
+    query = exp.select("*").from_(relation, dialect=evaluator.dialect).qualify(first_unique_row)
```
georgesittas, line 1153: Nit: since `relation` is already an expression, `dialect` is not necessary here:

```suggestion
    query = exp.select("*").from_(relation).qualify(first_unique_row)
```

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,61 @@ def var(
...
+        'SELECT * FROM demo.table QUALIFY ROW_NUMBER() OVER (PARTITION BY user_id, CAST(timestamp AS DATE) ORDER BY timestamp DESC, status ASC) = 1'
+    """
+    if not isinstance(partition_by, list):
+        raise SQLMeshError(
+            "partition_by must be a list of columns: [<column>, cast(<column> as <type>)]"
+        )
+
+    if not isinstance(order_by, list):
+        raise SQLMeshError(
+            "order_by must be a list of strings, null values in columns ordered first: ['<column> <asc|desc>']"
```
georgesittas, line 1132: See other comment re: `<nulls first|nulls last>`

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,61 @@ def var(
...
+
+    partition_clause = exp.tuple_(*[col for col in partition_by])
+
+    order_expressions = [
+        evaluator.transform(parse_one(order_item, into=exp.Ordered, dialect=evaluator.dialect))
+        for order_item in order_by
+    ]
+
+    if not order_expressions:
+        raise SQLMeshError("At least one order_by expression is required: '<column> <asc|desc>'")
```
georgesittas, line 1143: Ditto

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
```diff
@@ -1100,6 +1100,61 @@ def var(
...
+
+    if not isinstance(order_by, list):
+        raise SQLMeshError(
+            "order_by must be a list of strings, null values in columns ordered first: ['<column> <asc|desc>']"
+        )
+
+    partition_clause = exp.tuple_(*[col for col in partition_by])
+
+    order_expressions = [
+        evaluator.transform(parse_one(order_item, into=exp.Ordered, dialect=evaluator.dialect))
```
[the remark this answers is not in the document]
sungchun12: following the convention of the docs tests in the other macro definitions. This should also identify macros passed as args for this deduplicate macro itself right?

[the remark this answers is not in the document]
georgesittas: Ah because you're actually passing SQL strings that are parsed. Seems legit.

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -901,6 +901,40 @@ FROM rides
...
 
+### @DEDUPLICATE
+
+`@DEDUPLICATE` is used to deduplicate rows in a table based on the specified partition and order columns with a window function.
+
+It supports the following arguments, in this order:
+
+- `relation`: The table or CTE name to deduplicate
+- `partition_by`: column names, or expressions to use to identify a window of rows out of which to select one as the deduplicated row
+- `order_by`: A list of strings representing the ORDER BY clause, null values in columns ordered first
```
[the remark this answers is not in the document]
sungchun12: Ahhh I did misunderstand what you were saying. I was trying to avoid giving more order by configs to the user to avoid bloat, but in the error message it's good to include them. Updated my error messages and tests to include nulls ordering

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
sungchun12: updated

sqlmesh/core/macros.py
sungchun12: updated the error message

georgesittas (approved):

Issue #2948, opened 2024-07-26
Should python macro functions work in audits?


patricksurry:
In the example project, I added macros/constants.py containing 
```python
from sqlmesh import macro
from sqlmesh.core.macros import MacroEvaluator

@macro()
def min_item(
    evaluator: MacroEvaluator
) -> int:
    return 0
```

I modified models/incremental_model.sql to reference it like this:
```sql
...
WHERE
  event_date BETWEEN @start_date AND @end_date
  and item_id > @min_item()
``` 

which works just fine.  But if I also modify audits/assert_positive_order_ids.sql to reference it:
```sql
...
SELECT
  *
FROM @this_model
WHERE
  item_id < @min_item()
```

then it fails to find the macro:
```sh
sqlmesh-example % sqlmesh audit                                   
Error: Macro 'min_item' does not exist.

sqlmesh-example % sqlmesh --version
0.113.0
```

izeigerman:
I believe we don't scan audits for macro references when constructing the model's python environment. This is why the macro can't be found.

@georgesittas , @Themiscodes can either of you please have a look when you get a chance.

Themiscodes:
Sure, I will look into it

PR #2969, opened 2024-07-30
Feat: Support macros within audits

Themiscodes:
This update addresses an issue, where macros can't be resolved when called inside audits, fixes: #2948 

It involves scanning the audits as well for macro references and including these macros when constructing the model's Python environment.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2271,11 +2271,12 @@ def _extract_audit_expressions(
     audit_expressions = []
     if audit_names and audits:
         for audit_name in audit_names:
-            if audit_name in audits:
-                audit_expressions.append(audits[audit_name].query)
+            audit = audits.get(audit_name)
+            if audit:
+                audit_expressions.append(audit.query)
     if inline_audits:
-        for audit in inline_audits:
-            audit_expressions.append(inline_audits[audit].query)
+        for audit_name in inline_audits:
+            audit_expressions.append(inline_audits[audit_name].query)
 
     return audit_expressions
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -20,6 +20,7 @@ from sqlglot.optimizer.simplify import gen
 from sqlglot.schema import MappingSchema, nested_set
 from sqlglot.time import format_time
 
+from sqlmesh.core.dialect import extract_audit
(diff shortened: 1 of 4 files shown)


erindru (approved):

sqlmesh/core/model/definition.py
```diff
@@ -1736,10 +1746,11 @@ def create_sql_model(
 
     pre_statements = pre_statements or []
     post_statements = post_statements or []
+    audit_expressions = audit_expressions or []
 
     if not python_env:
         python_env = _python_env(
-            [*pre_statements, query, *post_statements],
+            [*pre_statements, query, *post_statements, *audit_expressions],
```
tobymao, line 1753: is this a breaking change? are we going to need a migration? will this change data hashes?

Themiscodes: I believe since by scanning the audit expressions and adding its macros to the model's `python_env` will only result in changes when an audit contains macro references, this shouldn't affect the data hash or introduce any breaking changes. as previously, if the user defined a macro within an audit it wouldn't be resolved and would cause the plan to fail

sqlmesh/core/model/definition.py
```diff
@@ -2236,6 +2249,37 @@ def _list_of_calls_to_exp(value: t.List[t.Tuple[str, t.Dict[str, t.Any]]]) -> ex
...
+
+    if default_audits:
+        for audit_name, _ in default_audits:
+            audit_names.append(audit_name)
+
+    audit_expressions = []
+    if audit_names and audits:
+        for audit_name in audit_names:
+            if audit_name in audits:
+                audit_expressions.append(audits[audit_name].query)
```
georgesittas, line 2275:
```suggestion
            audit = audits.get(audit_name)
            if audit:
                audit_expressions.append(audit.query)
```

georgesittas, line 2275:
```suggestion
            audit = audits.get(audit_name)
            if audit:
                audit_expressions.append(audit.query)
```

georgesittas (approved):
Let's also test for default audits?

sqlmesh/core/model/definition.py
Themiscodes: Thanks refactored this and added test for default audits

georgesittas (approved):

PR #2975, opened 2024-07-31
Fix!: Improve handling of the `--where` option in table_diff

erindru:
Prior to this change, running `sqlmesh table_diff` with something like `--where "foo = 'bar'"` would throw an error like:

```
duckdb.duckdb.BinderException: Binder Error: Ambiguous reference to column name "foo" (use: "t.foo" or "s.foo")
```

This was because the where clause was injected verbatim without automatically qualifying the columns. The user had to know that table_diff used a subselect internally and supply `--where "s.foo = 'bar' and t.foo = 'bar'"` in order to get this to work.

This PR refactors the generated table_diff query to use CTE's to build the source / target / stats datasets so the `--where` clause can still be injected verbatim but it's now scoped to each dataset so there are no more ambiguous column references

### DIFF
--- a/sqlmesh/core/table_diff.py
+++ b/sqlmesh/core/table_diff.py
@@ -272,7 +272,22 @@ class TableDiff:
             def name(e: exp.Expression) -> str:
                 return e.args["alias"].sql(identify=True)
 
-            query = (
+            source_query = (
+                exp.select(*(exp.column(c) for c in source_schema.keys()))
+                .from_(self.source)
+                .where(self.where)
+            )
+            target_query = (
+                exp.select(*(exp.column(t) for t in target_schema.keys()))
+                .from_(self.target)
+                .where(self.where)
+            )
+
+            source_identifier = exp.to_identifier("__source")
+            target_identifier = exp.to_identifier("__target")
+            stats_identifier = exp.to_identifier("__stats")
+
+            stats_query = (
                 exp.select(
                     *s_selects.values(),
                     *t_selects.values(),
@@ -313,14 +328,16 @@ class TableDiff:
                     ).as_("null_grain"),
                     *comparisons,
                 )
-                .from_(exp.alias_(self.source, "s"))
-                .join(
-                    self.target,
-                    on=self.on,
-                    join_type="FULL",
-                    join_alias="t",
-                )
-                .where(self.where)
+                .from_(exp.alias_(source_identifier, "s"))
+                .join(exp.alias_(target_identifier, "t"), on=self.on, join_type="FULL")
+            )
+
+            ctes = exp.With(
+                expressions=[
+                    exp.CTE(this=source_query, alias=exp.TableAlias(this=source_identifier)),
+                    exp.CTE(this=target_query, alias=exp.TableAlias(this=target_identifier)),
+                    exp.CTE(this=stats_query, alias=exp.TableAlias(this=stats_identifier)),
+                ]
             )
 
             query = exp.select(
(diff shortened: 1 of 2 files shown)


sqlmesh/core/table_diff.py
```diff
@@ -272,7 +272,22 @@ def _column_expr(name: str, table: str) -> exp.Expression:
             def name(e: exp.Expression) -> str:
                 return e.args["alias"].sql(identify=True)
 
-            query = (
+            source_query = (
+                exp.select(*(exp.column(c) for c in source_schema.keys()))
```
tobymao, line 276: you can just pass c, if it’s a string

you don’t need keys(), dict is iterable

sqlmesh/core/table_diff.py
```diff
@@ -272,7 +272,22 @@ def _column_expr(name: str, table: str) -> exp.Expression:
...
+                .from_(self.source)
+                .where(self.where)
+            )
+            target_query = (
+                exp.select(*(exp.column(t) for t in target_schema.keys()))
```
tobymao, line 281: ditto

sqlmesh/core/table_diff.py
```diff
@@ -313,14 +328,16 @@ def name(e: exp.Expression) -> str:
...
                 )
-                .from_(exp.alias_(self.source, "s"))
-                .join(
-                    self.target,
-                    on=self.on,
-                    join_type="FULL",
-                    join_alias="t",
-                )
-                .where(self.where)
+                .from_(exp.alias_(source_identifier, "s"))
```
tobymao, line 331: doesn’t seem right since it’s an identifier, is there no error here?

sqlmesh/core/table_diff.py
```diff
@@ -313,14 +328,16 @@ def name(e: exp.Expression) -> str:
...
+                .join(exp.alias_(target_identifier, "t"), on=self.on, join_type="FULL")
+            )
+
```
tobymao, line 334: i don’t like manually creating with, can you use a builder api or use subqueries?

sqlmesh/core/table_diff.py
erindru:
```
>>> exp.alias_(exp.to_identifier("__source"), "s").sql()
'__source AS s'
```

It got the correct result. What should I have done?

sqlmesh/core/table_diff.py
erindru: Is the builder API documented anywhere? There didnt seem to be a `.with(..)`

sqlmesh/core/table_diff.py
tobymao: exp.tables(…).as_()

sqlmesh/core/table_diff.py
erindru: It looks like I still need to use `exp.column()` due to this effect:

```
>>> exp.select("country code").sql()
'SELECT country AS code'

>>> exp.select(exp.column("country code")).sql()
'SELECT "country code"'
```

however i've removed `.keys()`

sqlmesh/core/table_diff.py
erindru: ah, I think I found it. There appears to be no top-level `with_()` like there is for `select()` but doing `exp.Select().with_(...)` gets me there.

Is that better?

tobymao: yea

georgesittas (approved):

Issue #2866, opened 2024-07-04
Support pre/post-statements in Python models that yield DataFrames


georgesittas:
Context: https://github.com/TobikoData/sqlmesh/discussions/2539

Summary: it's [possible](https://sqlmesh.readthedocs.io/en/latest/concepts/models/python_models/#optional-prepost-statements) to execute pre/post-statements in Python models, but only at evaluation time. This can be costly, so the item for this issue is to enable setting `pre_statements` and `post_statements` similar to `SqlBasedModel`s, so that they can be executed at table creation time and gated using `runtime_stage`.

PR #2977, opened 2024-08-01
Feat!: Support pre-post statements in python models at creation time

Themiscodes:
This update introduces the ability to enable pre- and post-statements in python models that can be executed during the table creation phase, fixes: #2866 

The SqlBased model renderers has been moved to the `_Model` class, so that the logic can be reused by python models. Also to support macro calls in the pre- and post-statements, the python environment is updated with the referenced macros (similarly to sql models).

### DIFF
--- a/sqlmesh/core/model/decorator.py
+++ b/sqlmesh/core/model/decorator.py
@@ -8,6 +8,8 @@ import inspect
 from sqlglot import exp
 from sqlglot.dialects.dialect import DialectType
 
+from sqlmesh.core.macros import MacroRegistry
+from sqlmesh.utils.jinja import JinjaMacroRegistry
 from sqlmesh.core import constants as c
 from sqlmesh.core.dialect import MacroFunc
 from sqlmesh.core.model.definition import (
@@ -75,6 +77,8 @@ class model(registry_decorator):
         module_path: Path,
         path: Path,
         defaults: t.Optional[t.Dict[str, t.Any]] = None,
+        macros: t.Optional[MacroRegistry] = None,
+        jinja_macros: t.Optional[JinjaMacroRegistry] = None,
         dialect: t.Optional[str] = None,
         time_column_format: str = c.DEFAULT_TIME_COLUMN_FORMAT,
         physical_schema_override: t.Optional[t.Dict[str, str]] = None,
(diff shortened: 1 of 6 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1877,6 +1852,21 @@ def create_python_model(
...
     # Find dependencies for python models by parsing code if they are not explicitly defined
     # Also remove self-references that are found
+
+    pre_statements = kwargs.get("pre_statements", None) or []
+    post_statements = kwargs.get("post_statements", None) or []
+
+    python_env.update(
+        _python_env(
+            expressions=[*pre_statements, *post_statements],
+            jinja_macro_references=None,
```
georgesittas, line 1862: Let's make sure jinja can be used for pre/post-statements in python models too, otherwise the functionality is unnecessarily crippled, right? This section may be useful: https://github.com/TobikoData/sqlmesh/blob/6e71e5fd60ad68062124510b4dc97dc83332facb/sqlmesh/core/model/definition.py#L1627

sqlmesh/core/model/definition.py
```diff
@@ -1853,8 +1825,11 @@ def create_seed_model(
 def create_python_model(
     name: str,
     entrypoint: str,
+    module_path: Path,
```
georgesittas, line 1828: Why was this added as a positional argument? This makes the PR breaking right? Let's double check this, these API changes are a bit risky.

sqlmesh/core/model/decorator.py
```diff
@@ -123,7 +127,7 @@ def model(
         for key in ("pre_statements", "post_statements"):
             statements = common_kwargs.get(key)
             if statements:
-                common_kwargs[key] = [exp.maybe_parse(s, dialect=dialect) for s in statements]
+                common_kwargs[key] = [parse_one(s, dialect=dialect) for s in statements]
```
georgesittas, line 130: Can `statements` ever contain actual expression instances instead of strings? This will crash if this happens. Perhaps that's why we were using `exp.maybe_parse` here in the first place?

Themiscodes: right, I didn't consider this. I've adjusted it to allow the use of actual expression instances

georgesittas (approved):

georgesittas (approved):

PR #2979, opened 2024-08-01
Fix!: Send notifications for failing non-blocking audits from the builtin scheduler

crericha:
Currently, the builtin-scheduler only sends notifications when the `SnapshotEvaluator` raises an `AuditError` exception, which only happens for failing, blocking audits.

This PR addresses two things. First, it delays when the `AuditError` is raised, so that all of the snapshot's audits execute first and second, it sends notifications for all audits that failed, blocking or non-blocking.

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -227,7 +227,7 @@ class Scheduler:
         )
 
         audit_error_to_raise: t.Optional[AuditError] = None
-        for audit_result in [result for result in audit_results if result.count]:
+        for audit_result in (result for result in audit_results if result.count):
             error = AuditError(
                 audit_name=audit_result.audit.name,
                 model=snapshot.model_or_none,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/scheduler.py
```diff
@@ -212,25 +214,38 @@ def evaluate(
...
+            execution_time=execution_time,
+            raise_exception=False,
+            snapshots=snapshots,
+            deployability_index=deployability_index,
+            wap_id=wap_id,
+            **kwargs,
+        )
+
+        audit_error_to_raise: t.Optional[AuditError] = None
+        for audit_result in [result for result in audit_results if result.count]:
```
georgesittas, line 230: Let's make this a generator?

```suggestion
        for audit_result in (result for result in audit_results if result.count):
```

georgesittas (approved):

PR #2980, opened 2024-08-02
Fix: ensure Context.close doesn't unnecessarily create a state sync db

georgesittas:
@treysp noticed that `sqlmesh info` triggers all migrations in a fresh project (e.g. `sqlmesh init duckdb`), which is weird. This happens because the CLI calls `close()` on the instantiated context and in turn that results in doing `self.state_sync` within that method, which creates a new state sync and populates it. We only need to do that if `_state_sync` actually exists.


treysp (approved):
Thank you!

PR #2981, opened 2024-08-02
Docs: add snowflake connection quickstart

treysp:
And make minor edits to the Databricks connection quickstart


docs/integrations/engines/snowflake.md
```diff
@@ -1,5 +1,296 @@
...
+      type: duckdb
+      database: db.db
+  snowflake:
+    connection:
+      type: snowflake
+
+default_gateway: local
+
+model_defaults:
+  dialect: duckdb
```
erindru, line 156: Should the example default dialect be `dialect: snowflake`?

Otherwise it feels like we are recommending cross-dialect projects in the first instance which have a tendency to trigger all the Snowflake normalization edge cases

treysp: I had the same thought, but:
- Since we have both the DuckDB and snowflake gateways present, cross-dialect execution is implied either way
- I don't want to show changing the default dialect when we change engines because that's explicitly wrong (in the sense that the dialect of the models is constant regardless of the engine you use)

What do you think about me deleting the DuckDB gateway first thing so we can set the model dialect to snowflake without implying cross-database usage? I think we'd probably still want a brief note about what default dialect is for/how it works?

erindru: Yeah either that or ignore duckdb entirely in the Snowflake quickstart (to keep the Snowflake quickstart focused on just Snowflake) and just `sqlmesh init snowflake` in the Prerequisites section instead of `sqlmesh init duckdb`

PR #2983, opened 2024-08-04
Fix!: Regression in Snowflake Python models using Snowpark DataFrame's

erindru:
Prior to this change, attempting to create a Python model on Snowflake utilising a Snowpark DataFrame would produce the following error:

```
snowflake.connector.errors.ProgrammingError: SQL compilation error:  Object found is of type 'VIEW', not specified type 'TABLE'.
```

This is because for Snowpark, the DataFrame is made available via `CREATE TEMPORARY VIEW` but the cleanup logic in the Snowflake engine adapter assumed a Table (because Pandas DataFrames use Tables) and tried to issue `DROP TABLE` instead of `DROP VIEW`.

Now the cleanup logic will issue `DROP VIEW` if a Snowpark DataFrame is in use.

In order to add a test for this, I had to re-introduce a dependency on `snowflake-snowpark-python` (first introduced in #2666 and then removed in #2761)

Since Snowpark doesnt work on Python 3.12, I also added some guards to only install it on Python < 3.12 and also skip the test if Snowpark is not available

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -204,9 +204,11 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin):
             target_table or "pandas", quoted=False
         )  # write_pandas() re-quotes everything without checking if its already quoted
 
+        is_snowpark_dataframe = snowpark and isinstance(df, snowpark.dataframe.DataFrame)
+
         def query_factory() -> Query:
-            if snowpark and isinstance(df, snowpark.dataframe.DataFrame):
-                df.createOrReplaceTempView(temp_table.sql(dialect=self.dialect, identify=True))
+            if is_snowpark_dataframe:
+                df.createOrReplaceTempView(temp_table.sql(dialect=self.dialect, identify=True))  # type: ignore
             elif isinstance(df, pd.DataFrame):
                 from snowflake.connector.pandas_tools import write_pandas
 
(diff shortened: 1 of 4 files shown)


tests/core/test_snapshot.py
```diff
@@ -371,7 +371,7 @@ def test_missing_intervals_end_bounded_with_ignore_cron(make_snapshot):
 
 
 def test_missing_intervals_past_end_date_with_lookback(make_snapshot):
-    snapshot: Snapshot = make_snapshot(
+    snapshot: Snapshot = make_snapshot(  # type: ignore
```
erindru, line 374: mypy was issuing a warning here so I silenced it

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -204,9 +204,11 @@ def _df_to_source_queries(
             target_table or "pandas", quoted=False
         )  # write_pandas() re-quotes everything without checking if its already quoted
 
+        is_snowpark_dataframe = snowpark and isinstance(df, snowpark.dataframe.DataFrame)
```
tobymao, line 207: does this force 3.11 or is the snowpark check enough to allow python 3.12 if they don't use snowpark

tobymao (approved):

sqlmesh/core/engine_adapter/snowflake.py
erindru: The latter. `snowpark` is defined as `optional_import("snowflake.snowpark")` which will return `False` on Python 3.12 so `is_snowpark_dataframe` will be False

PR #2984, opened 2024-08-05
fix: remove error when deleting expired external model snapshots

eakmanrq:
Prior to this change we always did a check on the table's database against the physical schema as a safety check to avoid deleting tables outside of where we expect to delete tables. This logic though assumes the following `evaluation_strategy.delete` call actually does a delete. This isn't true for symbolic evaluations. Therefore we had a bug where users could get errors when their symbolic models would expire and it would say it was going to delete outside of the scope of the physical db but that wasn't really the case since the actual delete was a no-op. 

Since the Evaluation Strategy is responsible for determining what deletes are actually happening, I push this check down to the strategy implementation layer. Now for deletes that do nothing there will not be a check and therefore there will no longer be false errors.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -799,12 +799,11 @@ class SnapshotEvaluator:
         evaluation_strategy = _evaluation_strategy(snapshot, self.adapter)
 
         for is_table_deployable, table_name in table_names:
-            table = exp.to_table(table_name)
-            if table.db != snapshot.physical_schema:
-                raise SQLMeshError(
-                    f"Table '{table_name}' is not a part of the physical schema '{snapshot.physical_schema}' and so can't be dropped."
-                )
-            evaluation_strategy.delete(table_name, is_table_deployable=is_table_deployable)
+            evaluation_strategy.delete(
+                table_name,
+                is_table_deployable=is_table_deployable,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1825,3 +1826,11 @@ def _check_destructive_schema_change(
         raise SQLMeshError(
             f"{warning_msg} To allow this, change the model's `on_destructive_change` setting to `warn` or `allow` or include it in the plan's `--allow-destructive-model` option."
         )
+
+
+def _check_table_db_is_physical_schema(table_name: str, physical_schema: str) -> None:
```
erindru, line 1831: `is_physical_schema` or `is_in_physical_schema`?

eakmanrq: I think `is_physical_schema` is correct right? We are checking if the physical schema matches the database name. One thing that might cause confusion is that we are mixing the terms `schema` and `db` to mean the same thing (because SQLGlot refers to it as `db` while SQLMesh refers to it here as `schema`).

erindru: Actually yeah, you're right. This throws me off all the time, some things are `catalog.db.table`, some are `catalog.schema.table` and elsewhere might be closer to `db.schema.table`

erindru (approved):

PR #2990, opened 2024-08-08
Feat: implement builtin date spine macro

sungchun12:
The goal of this PR is to bring a lot of UX familiarity from dbt-utils into SQLMesh native macros similar to the deduplicate macro I just contributed. 

dbt-utils version of this macro: https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/date_spine.sql

This renders to SQL to generate a date spine table. It's typically used to join in unique, hard-coded, date ranges to join with other tables/views so people don't have to constantly adjust date ranges in `where` clauses everywhere.
- From @afzaljasani : "When I worked with a lot of e-commerce companies they would have daily aggregate metrics from different sources like Facebook ads, Google ads, Shopify, etc. because most of these sources can’t be joined to each other they would join to a date spine table to get a 360 view of the business"
- Do analysis specific to this range of promotional dates
- I talked with someone at Modern Animal who said they don't like how dbt does it natively, so they built their own date_spine macro in house. 

**Design Highlights**
- Large upstream changes in SQLGlot had to happen due to `generate_date_array` transpiling and behaving very differently across query engines. Huge thanks to @georgesittas to making this PR way easier for me. He's a real one. 
- Verified it works with live query manual testing so far: snowflake, postgres, duckdb, databricks, spark, redshift, bigquery. I intentionally test for these as they have the most common usage.
- Verified it works with a development version of SQLGlot: `25.11.2.dev1`

**Differences from dbt-utils' version**
- The spine will include the `start_date` (if it is aligned to the datepart), AND it will include the `end_date`
- You can input macros AND plain string dates instead of casting them: `@date_spine('quarter', '2022-01-01', '2024-12-16')`
- Instead of using cross joins to arbitrarily generate date intervals, this relies on native `generate_series`-like sql functions to be faster and cheaper for the majority of query engines. If those functions don't exist for a particular engine (ex: redshift), recursion will be used. 

Other examples of ad hoc queries working across engines:
BigQuery
image

Redshift
image

Snowflake
image


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -909,7 +909,7 @@ It supports the following arguments, in this order:
 
 - `relation`: The table or CTE name to deduplicate
 - `partition_by`: column names, or expressions to use to identify a window of rows out of which to select one as the deduplicated row
-- `order_by`: A list of strings representing the ORDER BY clause, optional - nulls ordering: ['<column> <asc|desc> nulls <first|last>']
+- `order_by`: A list of strings representing the ORDER BY clause, optional - you can add nulls ordering like this: ['status desc nulls last']
```
georgesittas, line 912:
```suggestion
- `order_by`: A list of strings representing the ORDER BY clause, optional - you can add null ordering like this: ['status desc nulls last']
```

Q: what is "status" here?

sqlmesh/core/macros.py
```diff
@@ -1157,6 +1158,151 @@ def deduplicate(
...
+        datepart: The datepart to use for the date spine: day, week, month, year
+        start_date: The start date for the date spine in format YYYY-MM-DD
+        end_date: The end date for the date spine in format YYYY-MM-DD
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlglot.schema import MappingSchema
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "@date_spine('week', '2022-01-20', '2024-12-16')"
+        >>> MacroEvaluator(dialect='duckdb').transform(parse_one(sql)).sql()
```
georgesittas, line 1180: Let's also use `dialect="duckdb"` in both `parse_one` and `sql`. Or we could get rid of it in the `MacroEvaluator` instance?

sqlmesh/core/macros.py
```diff
@@ -1157,6 +1158,151 @@ def deduplicate(
...
+        'SELECT UNNEST(GENERATE_SERIES(CAST(\'2022-01-20\' AS DATE), CAST(\'2024-12-16\' AS DATE), (7 *INTERVAL \'1\' DAY))) AS "date_week"'
+
+    """
+    if datepart.name not in ("day", "week", "month", "year"):
```
georgesittas, line 1184: I'd consider assigning `datepart.name` to a variable to avoid invoking it repeatedly.

sqlmesh/core/macros.py
```diff
@@ -1157,6 +1158,151 @@ def deduplicate(
...
+        raise SQLMeshError(
+            f"Invalid date range - start_date '{start_date.name}' is after end_date '{end_date.name}'."
+        )
+
+    alias_name = f"date_{datepart.name}"
+    start_date_column = exp.cast(start_date, "DATE", dialect=evaluator.dialect)
+    end_date_column = exp.cast(end_date, "DATE", dialect=evaluator.dialect)
+    date_interval = exp.Interval(this=exp.Literal.number(1), unit=datepart.name)
+
+    if evaluator.dialect == "bigquery":
```
georgesittas, line 1207: I don't like this "switching" approach. Why can't we just take a stab at transpiling these array generations?

sungchun12: I did, but `generate_date_array` doesn't transpile to any of these other database engines. And you'll notice in bigquery, the alias must happen in the from clause vs. the select clause to work correctly. What do you suggest instead given the above?

georgesittas: I'll need to investigate more to answer this, but I'd push for handling this in SQLGlot vs here, because it introduces a pattern that we're generally trying to avoid. I'm pretty confident that we handle many of the nuances already, just need to find the correct AST representation for this concept.

If you have already documented how date spines can be generated in the supported dialects, it would be very helpful if you opened a ticket in SQLGlot. Or we can continue the discussion in Slack and figure out the details -- me and @VaggelisD can help take this to the finish line by improving the transpilation part.

sungchun12: Alright we can take this back to slack then, since it sounds like the switching approach is a no go

georgesittas: Given https://github.com/tobymao/sqlglot/pull/3899, I think we should be able to do something like this now, more or less:

```python
@macro()
def date_spine(evaluator, start, end, date_part):
    generate_date_array = exp.func(
        "GENERATE_DATE_ARRAY",
        exp.cast(start, "date"),
        exp.cast(end, "date"),
        f"INTERVAL 1 {date_part}"
    )
    exploded = exp.Unnest(expressions=generate_date_array)
    return exp.select(f"date_{date_part}").from_(exploded.as_(f"_dates(date_{date_part}"))
```

georgesittas: Note: redshift and mysql use the recursive CTE approach to produce the date spine. We can test / benchmark this and if it turns out too slow there's another idea which is to statically expand the date spine by doing a datediff in python and replacing it with a query that produces static data.

sqlmesh/core/macros.py
sungchun12: I got rid of it

docs/concepts/macros/sqlmesh_macros.md
sungchun12: updated to this for clarity: `['<column_name> desc nulls last']`

sqlmesh/core/macros.py
sungchun12: done

sungchun12:
I signed the CLA with the correct github username. Commits under my personal name is likely due to using my personal laptop for commits. For all purposes, both commit signatures are the same person.

tests/core/test_macros.py
```diff
@@ -754,3 +754,223 @@ def test_deduplicate_error_handling(macro_evaluator):
         str(e.value.__cause__)
         == "order_by must be a list of strings, optional - nulls ordering: ['<column> <asc|desc> nulls <first|last>']"
     )
+
+
+@pytest.mark.parametrize(
+    "dialect, date_part",
```
georgesittas, line 760: Should we add the expected SQL in this parameterize? Or just parameterize on the date part since we're already checking all dialects in that nested if-elif-else block?

georgesittas (approved):
Sick how simple the implementation is now 🚀

The PR looks good, just a few comments to consider re: docs and test refactoring.

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -935,6 +935,39 @@ SELECT
...
+WITH "discount_promotion_dates" AS (                                                                             
+  SELECT                                                                                                         
+    "_exploded"."date_day" AS "date_day"                                                                         
+  FROM UNNEST(CAST(GENERATE_SERIES(CAST('2024-01-01' AS DATE), CAST('2024-01-16' AS DATE), INTERVAL '1' DAY) AS  
+DATE[])) AS "_exploded"("date_day")                                                                              
+)                                                                                                                
+SELECT                                                                                                           
+  "discount_promotion_dates"."date_day" AS "date_day"                                                            
+FROM "discount_promotion_dates" AS "discount_promotion_dates"
+```
```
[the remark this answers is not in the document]
sungchun12: updated

tests/core/test_macros.py
```diff
@@ -754,3 +754,223 @@ def test_deduplicate_error_handling(macro_evaluator):
...
+            '{date_part}',
+            '2022-01-01',
+            '2024-12-31'
+        )
+        """
+    schema = MappingSchema({}, dialect=dialect)
+    evaluator = MacroEvaluator(schema=schema, dialect=dialect)
+
+    # Generate the expected SQL based on the dialect and date_part
+    if dialect == "duckdb":
```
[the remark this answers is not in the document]
sungchun12: whoops yes, I didn't update this for the interval arg after making that conditional. It's now DRYer.

PR #2991, opened 2024-08-08
create snowflake comments in bulk

ananis25:
Snowflake adapter creates comments on tables/views one by one (the default behavior) which can get really slow when working with wide tables (upwards of 5 mins for the update step, which took 20s after the change). 

The `ALTER TABLE` syntax in snowflake supports adding multiple column comments at once, which we use here. The `ALTER VIEW` syntax doc doesn't mention it, but it works there too. 

1 - https://docs.snowflake.com/en/sql-reference/sql/alter-table-column#syntax
2 - https://docs.snowflake.com/en/sql-reference/sql/alter-view#syntax

Please let me know if this needs more testing. I tested this works in my project, and by manually running the sql generated in the added test routine.

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import logging
 import typing as t
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


georgesittas (approved):
Thanks for the PR @ananis25! LGTM, cc @treysp can you also take a look?

treysp (approved):
Thanks for the PR!

PR #2994, opened 2024-08-08
Chore: fix sushi waiters default_catalog capitalization

treysp:
And address pandas warnings by resetting DF indexes in sushi python models

Context:
- Snowflake normalizes unquoted object names to uppercase, which can cause case mismatches with default_catalog due to sqlmesh's default normalization behavior. 
- SQLMesh addresses this by rewriting the default catalog name on the fly in snowflake `_to_sql()`. 
- Sushi's waiters.py model code manually extracts default_catalog name prior to the rewriting, so we must manually lowercase the default_catalog if the parent catalog name is an uppercase version of the default catalog name.

### DIFF
--- a/examples/sushi/models/waiters.py
+++ b/examples/sushi/models/waiters.py
@@ -41,14 +41,27 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
         # There are tests which force not having a default catalog so we check here if one is defined
         # and add it to the name if it is
         default_catalog = evaluator.default_catalog
+        parent_snapshot_name = parent_snapshots[0].name
+        parent_snapshot_catalog = parent_snapshot_name.split(".")[0].strip('"')
+
         if default_catalog:
             # make sure we don't double quote the default catalog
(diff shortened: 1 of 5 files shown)


examples/sushi/models/waiters.py
```diff
@@ -41,14 +41,27 @@ def entrypoint(evaluator: MacroEvaluator) -> exp.Select:
...
         if default_catalog:
             # make sure we don't double quote the default catalog
             default_catalog = default_catalog.strip('"')
+
+            # Snowflake normalizes unquoted names to uppercase, which can cause case mismatches with
+            # default_catalog due to sqlmesh's default normalization behavior. SQLMesh addresses this
+            # by rewriting the default catalog name on the fly in snowflake `_to_sql()`. This model
+            # code manually extracts default_catalog name, so we manually lowercase the default_catalog
+            # if the parent catalog name is an uppercase version of the default catalog name.
+            default_catalog = (
```
erindru, line 56: test vs reality strikes again!

erindru (approved):

Issue #2988, opened 2024-08-07
SQLMesh+dbt+duckdb: Unable to read source declared in dbt


cnstlungu:
I'm using sqlmesh+dbt+duckdb. I have declared my sources as follows:

![image](https://github.com/user-attachments/assets/03ed0a7b-8d43-4c60-8f86-99788103d1d0)

In my model, upon using the source jinja function, I get an error:

![image](https://github.com/user-attachments/assets/e5dc8423-c9ee-4d41-a1e0-030b495c4998)

```
Traceback (most recent call last):
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/web/server/api/endpoints/commands.py", line 69, in evaluate
    df = context.evaluate(
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/analytics/__init__.py", line 99, in wrapper
    return func(*args, **kwargs)
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/context.py", line 802, in evaluate
    df = self.snapshot_evaluator.evaluate_and_fetch(
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/snapshot/evaluator.py", line 164, in evaluate_and_fetch
    result = self._evaluate_snapshot(
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/snapshot/evaluator.py", line 578, in _evaluate_snapshot
    return self.adapter._fetch_native_df(query_or_df.limit(limit))
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/engine_adapter/base.py", line 1697, in _fetch_native_df
    self.execute(query, quote_identifiers=quote_identifiers)
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/engine_adapter/base.py", line 1831, in execute
    self._execute(sql, **kwargs)
  File "/home/costin/repos/portable-data-stack-sqlmesh/.venv/lib/python3.10/site-packages/sqlmesh/core/engine_adapter/base.py", line 1837, in _execute
    self.cursor.execute(sql, **kwargs)
[...]

PR #2999, opened 2024-08-12
Feat: Add support for dbt sources external location defined in meta

Themiscodes:
An update to the dbt adapter to resolve sources' tables from an external location when specified in the source metadata, rather than using schema and table names, fixes: #2988

### DIFF
--- a/sqlmesh/dbt/source.py
+++ b/sqlmesh/dbt/source.py
@@ -44,6 +44,7 @@ class SourceConfig(GeneralConfig):
     loaded_at_field: t.Optional[str] = None
     quoting: t.Dict[str, t.Optional[bool]] = {}
     external: t.Optional[t.Dict[str, t.Any]] = {}
+    source_meta: t.Optional[t.Dict[str, t.Any]] = {}
     columns: t.Dict[str, ColumnConfig] = {}
 
     _canonical_name: t.Optional[str] = None
@@ -93,5 +94,7 @@ class SourceConfig(GeneralConfig):
                 "identifier": self.table_name,
                 "type": RelationType.External.value,
                 "quote_policy": AttributeDict(self.quoting),
+                "external": self.external,
+                "source_meta": self.source_meta,
             }
         )
--- a/sqlmesh/dbt/source.py
+++ b/sqlmesh/dbt/source.py
@@ -87,6 +87,15 @@ class SourceConfig(GeneralConfig):
 
     @property
     def relation_info(self) -> AttributeDict:
+        external_location = (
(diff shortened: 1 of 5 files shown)


sqlmesh/dbt/source.py
```diff
@@ -44,6 +44,7 @@ class SourceConfig(GeneralConfig):
     loaded_at_field: t.Optional[str] = None
     quoting: t.Dict[str, t.Optional[bool]] = {}
     external: t.Optional[t.Dict[str, t.Any]] = {}
+    source_meta: t.Optional[t.Dict[str, t.Any]] = {}
```
erindru, line 47: Is there a reason why you had to do this vs using the existing `meta` field on the `GeneralConfig` base class?

sqlmesh/dbt/source.py
```diff
@@ -86,12 +87,22 @@ def canonical_name(self, context: DbtContext) -> str:
...
     @property
     def relation_info(self) -> AttributeDict:
+        external_location = (
+            self.source_meta.get("external_location", None) if self.source_meta else None
+        )
+        external = (
+            {"external": external_location.replace("{name}", self.table_name)}
+            if external_location
+            else {}
+        )
```
crericha, line 97: I suggest we call this extras, so the dict can cover other cases. I'm adding additional databricks support, which is similar to this, but different fields from a different dbt config dict and imo it makes sense to use the same dict var for both.

sqlmesh/dbt/source.py
Themiscodes: yes i considered this, but I noticed that in `dbt` they have both fields: [dbt source definition](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/artifacts/resources/v1/source_definition.py#L65).

where `source_meta` is the top-level meta key. for example when we define a source as:

```yaml
sources:
  - name: parquet
    meta:
      external_location: "s3://{name}.parquet"
    tables:
      - name: table name
        meta: 
          owner: "@Alice"
```

in dbt's `SourceDefinition` we would get:

```python
meta = {'owner': '@alice'}
source_meta = {'external_location': "s3://{name}.parquet"}
```

so I wanted to ensure that both these fields could be mapped in our `SourceConfig` to align with their definition

sqlmesh/dbt/source.py
Themiscodes: yes changed this to `extras` this does make more sense

sqlmesh/dbt/source.py
erindru: Oooh I see, of course. Makes sense!

crericha (approved):

PR #3000, opened 2024-08-12
Docs: Update python pre/post statements example

Themiscodes:
Update docs with the added functionality of pre/post statements for python models

### DIFF
--- a/docs/concepts/models/python_models.md
+++ b/docs/concepts/models/python_models.md
@@ -100,7 +100,7 @@ You can set the `pre_statements` and `post_statements` arguments to a list of SQ
         "id": "int",
         "name": "text",
     },
-    pre_statements=[exp.Cache(this=exp.Table(this=exp.Identifier(this="x")), expression=exp.Select(expressions=exp.Literal(this='1',is_string=False)))]
+    pre_statements=["SET GLOBAL parameter = 'value';", exp.Cache(this=exp.Table(this=exp.Identifier(this="x")), expression=exp.Select(expressions=exp.Literal(this='1',is_string=False)))]
     post_statements=["@CREATE_INDEX(@this_model, id)"],
 )
 def execute(
--- a/docs/concepts/models/python_models.md
+++ b/docs/concepts/models/python_models.md
@@ -120,6 +120,7 @@ def create_index(
 ):
     if evaluator.runtime_stage == "creating":
         return f"CREATE INDEX idx ON {model_name}({column});"
+    return None
 ```
 
 Alternatively, pre- and post-statements can be issued with the SQLMesh [`fetchdf` method](../../reference/cli.md#fetchdf) [described above](#execution-context).
--- a/docs/concepts/models/python_models.md
+++ b/docs/concepts/models/python_models.md
@@ -117,6 +117,37 @@ def execute(
     context.fetchdf("CREATE INDEX idx ON example.pre_post_statements (id);")
 ```
 
+Alternatively, you can set the `pre_statements` and `post_statements` arguments to a list of SQL strings and/or SQLGlot expressions to define the model's pre/post-statements.
+
+``` python linenums="1" hl_lines="6-7"
+def execute(
+    context: ExecutionContext,
+    start: datetime,
+    end: datetime,
+    execution_time: datetime,
+    pre_statements=["SET GLOBAL parameter = 'value';"],
+    post_statements=["@CREATE_INDEX('example.pre_post_statements', id)"],
+    **kwargs: t.Any,
+) -> pd.DataFrame:
+
+    return pd.DataFrame([
+        {"id": 1, "name": "name"}
+    ])
+
+```
+
+Within these pre/post-statements, you can also use SQLMesh or Jinja macros to achieve more nuanced control. For example, a macro that creates an index based on the runtime stage:
+
+``` python linenums="1"
+@macro()
+def create_index(
+    evaluator: MacroEvaluator,
+    model_name: str,
+    column: str,
+):
+    if evaluator.runtime_stage == "creating":
+        return f"CREATE INDEX idx ON {model_name}({column});"
+```
 
 ## Dependencies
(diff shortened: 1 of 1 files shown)


docs/concepts/models/python_models.md
```diff
@@ -117,6 +117,37 @@ def execute(
...
+    **kwargs: t.Any,
+) -> pd.DataFrame:
+
+    return pd.DataFrame([
+        {"id": 1, "name": "name"}
+    ])
+
+```
+
+Within these pre/post-statements, you can also use SQLMesh or Jinja macros to achieve more nuanced control. For example, a macro that creates an index based on the runtime stage:
```
georgesittas, line 139: Nit: would personally remove this bit
```suggestion
Within these pre/post-statements, you can also use SQLMesh or Jinja macros. For example, a macro that creates an index based on the runtime stage:
```

georgesittas (approved):
Nice, should we put the new section at the top of the "Optional pre/post-statements" and move the fetchdf approach to the bottom?

docs/concepts/models/python_models.md
Themiscodes: sure adapting these

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,39 @@ Optional pre/post-statements allow you to execute SQL commands before and after
...
+
+``` python linenums="1"
+@macro()
+def create_index(
+    evaluator: MacroEvaluator,
+    model_name: str,
+    column: str,
+):
+    if evaluator.runtime_stage == "creating":
+        return f"CREATE INDEX idx ON {model_name}({column});"
```
georgesittas, line 122: I just noticed we don't have a fallback `return` statement here. Let's add it to avoid any confusions, even though `None` is implicitly returned in Python.
```suggestion
    if evaluator.runtime_stage == "creating":
        return f"CREATE INDEX idx ON {model_name}({column});"
    return None
```

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,40 @@ Optional pre/post-statements allow you to execute SQL commands before and after
 
 For example, pre/post-statements might modify settings or create indexes. However, be careful not to run any statement that could conflict with the execution of another statement if models run concurrently, such as creating a physical table.
 
-Pre- and post-statements are issued with the SQLMesh [`fetchdf` method](../../reference/cli.md#fetchdf) [described above](#execution-context).
+You can set the `pre_statements` and `post_statements` arguments to a list of SQL strings and/or SQLGlot expressions to define the model's pre/post-statements.
```
treysp, line 93: "to a list of SQL strings and/or SQLGlot expressions" --> "to a list of SQL strings, SQLGlot expressions, or macro calls"

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,40 @@ Optional pre/post-statements allow you to execute SQL commands before and after
...
+
+``` python linenums="1" hl_lines="6-7"
+def execute(
+    context: ExecutionContext,
+    start: datetime,
+    end: datetime,
+    execution_time: datetime,
+    pre_statements=["SET GLOBAL parameter = 'value';"],
```
treysp, line 101: Add sqlglot expression

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,40 @@ Optional pre/post-statements allow you to execute SQL commands before and after
...
+    post_statements=["@CREATE_INDEX('example.pre_post_statements', id)"],
```
treysp, line 102: This would need to call against the physical table, right? 

Can a user pass `@this_model` instead of model name so it gets automatically resolved to the physical table? 

If not, I think the `@CREATE_INDEX` definition below needs to extract the physical table [from the snapshot](https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#accessing-snapshots), although I don't know the specific code for that off the top of my head.

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,40 @@ Optional pre/post-statements allow you to execute SQL commands before and after
 
 For example, pre/post-statements might modify settings or create indexes. However, be careful not to run any statement that could conflict with the execution of another statement if models run concurrently, such as creating a physical table.
 
-Pre- and post-statements are issued with the SQLMesh [`fetchdf` method](../../reference/cli.md#fetchdf) [described above](#execution-context).
+You can set the `pre_statements` and `post_statements` arguments to a list of SQL strings and/or SQLGlot expressions to define the model's pre/post-statements.
+
+``` python linenums="1" hl_lines="6-7"
+def execute(
```
treysp, line 96: Include @model decorator in example

docs/concepts/models/python_models.md
```diff
@@ -90,7 +90,40 @@ Optional pre/post-statements allow you to execute SQL commands before and after
...
+    **kwargs: t.Any,
+) -> pd.DataFrame:
+
+    return pd.DataFrame([
+        {"id": 1, "name": "name"}
+    ])
+
+```
+
+Within these pre/post-statements, you can also use SQLMesh or Jinja macros. For example, a macro that creates an index based on the runtime stage:
```
treysp, line 112: Replace with:

The previous example's `post_statements` called user-defined SQLMesh macro `@CREATE_INDEX('example.pre_post_statements', id)`.

We could define the `CREATE_INDEX` macro in the project's `macros` directory like this. The macro creates a table index on a single column, conditional on the [runtime stage](../../macros/macro_variables.md#runtime-variables) being `creating` (table creation time).

treysp (approved):

docs/concepts/models/python_models.md
Themiscodes: yes they can and it will be resolved. so I updated the example to use `@this_model` instead to show this

docs/concepts/models/python_models.md
Themiscodes: right updated this good catch it was in the wrong arguments before

PR #3001, opened 2024-08-12
feat: add Databricks Serverless support

eakmanrq:
Adds support for using Serverless compute from a notebook. I tested by plan/applying using a Serverless environment. 

Two key changes:
* Serverless does not support global temp views but does support session temp views. Global is at the application level while session is at the spark session level. I don't believe we support any sharing of state across sessions so changing this to session scoped views should be fine.
* Serverless threw a unique exception when trying to access SparkContext when it was not available so handle this exception too (we already had logic in place for supporting databricks connect). 

Note: This PR is build on the current constraints that the Python SQL Connector cannot query serverless. If that changes in the future, then the fix in the PR of having serverless use session temp views would become a problem if you want to mix SQL Connector and Databricks Connect. It might be that the global temp constraint is lifted by then but if not then the hybrid mode of using both would likely need to be disabled when using serverless.


sqlmesh/core/engine_adapter/spark.py
```diff
@@ -324,21 +323,27 @@ def _get_data_objects(
...
-                    if "Type: VIEW" in row["information"]
-                    else DataObjectType.TABLE
-                ),
+        data_objects = []
+        catalog = self.get_current_catalog()
+        for row in results:  # type: ignore
+            row_dict = row.asDict() if not isinstance(row, dict) else row
+            schema = row_dict.get("namespace") or row_dict.get("database")
+            # if schema is not present then that means it is a session temp view and can be skipped
+            if not schema:
```
izeigerman, line 332: How can this be the case if we run `SHOW TABLE EXTENDED IN {schema_name} LIKE '{pattern}` above?

eakmanrq: It must include temp objects when you run that command.

izeigerman (approved):

sqlmesh/core/engine_adapter/spark.py
eakmanrq: Yep verified this. Testing out using `isTemporary` field instead of checking for null schema.

eakmanrq:
Found out that when using Databricks SQL Connector + Databricks connect then you need the global temp views since they are trying to share temp objects across sessions (but still within the same application). Therefore now I just use the global temp views in all cases except when using databricks serverless.

Also added serverless support to databricks-connect.

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -260,8 +264,14 @@ def _df_to_source_queries(
 
         def query_factory() -> Query:
             temp_table = self._get_temp_table(target_table or "spark", table_only=True)
-            df.createOrReplaceGlobalTempView(temp_table.sql(dialect=self.dialect))  # type: ignore
-            temp_table.set("db", "global_temp")
+            if self.use_serverless:
```
izeigerman, line 267: This looks like databricks specific. Should we just override `query_factory` it in the corresponding adapter?

izeigerman: EG.:
```
def query_factory() -> Query:
            temp_table = self._get_temp_table(target_table or "spark", table_only=True)
            if self.use_serverless:
                ...
            else:
                super()....
```

eakmanrq: Hmm I think I will just completely override it since the `_use_spark_session` is also databricks specific. 

Change: https://github.com/TobikoData/sqlmesh/pull/3001/commits/91cb2a16e058fa297f1e7985d81fadd1f104a627

izeigerman (approved):

PR #3006, opened 2024-08-13
Feat: make MacroEvaluator.columns_to_types more lenient

georgesittas:
This PR makes the `columns_to_types` method more lenient by leveraging any existing snapshots in the evaluator's context as a fallback to fetch the target model's schema.

The motivation can be found [here](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1723568953070039?thread_ts=1723562831.081849&cid=C044BRE5W4S): one real-world example is executing `ALTER` statements for the physical table of a model after it's created, for which we need to know the model's schema. This is not possible today, because the model's own schema is not included in its `mapping_schema` -- it only includes its dependencies' schemas.

I also added a guard that protects from returning the dummy schema unless we're at the loading runtime stage, to avoid cryptic errors such as "Column `__schema_unavailable_at_load__` does not exist!" when queries are executed during table creation or model evaluation.

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -405,7 +405,11 @@ class MacroEvaluator:
 
     def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, exp.DataType]:
         """Returns the columns-to-types mapping corresponding to the specified model."""
-        if self._schema is None or self._schema.empty:
+
+        # We only return this dummy schema at load time, because if we don't actually know the
+        # target model's schema at creation/evaluation time, returning a dummy schema could lead
+        # to unintelligible errors when the query is executed
+        if (self._schema is None or self._schema.empty) and self.runtime_stage == "loading":
             self.columns_to_types_called = True
             return {"__schema_unavailable_at_load__": exp.DataType.build("unknown")}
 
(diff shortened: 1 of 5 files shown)


sqlmesh/core/macros.py
```diff
@@ -414,9 +418,16 @@ def columns_to_types(self, model_name: TableName | exp.Column) -> t.Dict[str, ex
...
             dialect=self.dialect,
         )
-        columns_to_types = self._schema.find(
-            exp.to_table(normalized_model_name), ensure_data_types=True
+        model_name = exp.to_table(normalized_model_name)
+
+        columns_to_types = (
+            self._schema.find(model_name, ensure_data_types=True) if self._schema else None
         )
+        if columns_to_types is None:
```
izeigerman, line 426: I'd also add an explicit unit test for this and not just rely on the sushi project to test this.

georgesittas: Good idea, will add shortly. Thanks for reviewing 👍

georgesittas: Added - https://github.com/TobikoData/sqlmesh/pull/3006/commits/c21c23b3aae6860221ecafeadf4d1e5ce285da89

PR #3008, opened 2024-08-14
fix!: Toby/perf

tobymao:


sqlmesh/core/model/definition.py
```diff
@@ -1202,6 +1236,17 @@ class SeedModel(_SqlBasedModel):
...
 
+    def __getstate__(self) -> t.Dict[t.Any, t.Any]:
+        state = super().__getstate__()
+        state["__dict__"] = state["__dict__"].copy()
+        state["__dict__"].pop("_reader", None)
+        return state
+
+    def copy(self, **kwargs: t.Any) -> Self:
+        model = super().copy(**kwargs)
+        model.__dict__.pop("_reader", None)
```
izeigerman, line 1247: What's `_reader`?

sqlmesh/core/context.py
```diff
@@ -427,12 +426,20 @@ def upsert_model(self, model: t.Union[str, Model], **kwargs: t.Any) -> Model:
             raise SQLMeshError(f"The disabled model '{model.name}' cannot be upserted")
         path = model._path
 
-        # model.copy() can't be used here due to a cached state that can be a part of a model instance.
-        model = t.cast(Model, type(model)(**{**t.cast(Model, model).dict(), **kwargs}))
+        model = model.copy(update=kwargs)
+        model._full_depends_on = None  # the query can be updated to change deps
```
izeigerman, line 430: Why do we not do this in the model's `copy()`?

sqlmesh/core/model/definition.py
tobymao: seed reader / csv reader

sqlmesh/core/context.py
tobymao: because most of the times you want to cache _full_depends_on, it's just in context.upsert (like in tests), we mutate the query which can change depends_on

izeigerman: I think this may lead to too many gotchas. I'd check in `copy` whether `update` include `depends_on` or `query`. And then invalidate based on that. Otherwise this is too error prone.

izeigerman: Also I don't see us resetting renderers on copy

sqlmesh/core/model/cache.py
```diff
@@ -44,22 +42,20 @@ def get_or_load(self, name: str, entry_id: str = "", *, loader: t.Callable[[], M
...
-        loaded_model = loader()
-        if isinstance(loaded_model, SqlModel):
-            new_entry = SqlModelCacheEntry(
-                model=loaded_model, full_depends_on=loaded_model.full_depends_on
-            )
-            self._file_cache.put(name, entry_id, value=new_entry)
+        model = loader()
+        if isinstance(model, SqlModel):
+            # make sure we preload full_depends_on
+            model.full_depends_on
```
izeigerman, line 51: Such a hack :) This is why `full_depends_on` was a separate attribute - to explicitly store it and not rely on internal mechanics of materializing the cache.

sqlmesh/core/model/definition.py
```diff
@@ -698,7 +711,7 @@ def is_python(self) -> bool:
     def is_seed(self) -> bool:
         return False
 
-    @cached_property
+    @property
```
izeigerman, line 714: Why this changed?

sqlmesh/core/model/definition.py
```diff
@@ -995,6 +1012,23 @@ class SqlModel(_SqlBasedModel):
...
+        # query renderer is very expensive to serialize
+        state["__dict__"].pop("_query_renderer", None)
+        state["__dict__"].pop("column_descriptions", None)
+        private = state[PRIVATE_FIELDS]
+        private["_columns_to_types"] = None
+        return state
+
+    def copy(self, **kwargs: t.Any) -> Self:
+        model = super().copy(**kwargs)
+        model.__dict__.pop("_query_renderer", None)
```
izeigerman, line 1027: Ah, found it. Why not reset it using the attribute directly? Why pop it?

sqlmesh/core/node.py
```diff
@@ -193,6 +200,8 @@ class _Node(PydanticModel):
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
     _path: Path = Path()
+    _data_hash: t.Optional[str] = None
```
izeigerman, line 203: Why is this here and not in the `_Model`? I think it only applies there.

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1425,6 +1427,7 @@ def _snapshot_to_json(snapshot: Snapshot) -> str:
 
 def parse_snapshot(
     model_cache: ModelCache,
+    optimized_query_cache: OptimizedQueryCache,
```
izeigerman, line 1430: Let's not do this here. I'm going to work on a more comprehensive snapshot caching next.

sqlmesh/core/model/definition.py
tobymao: because it's a cached_property and it's behavior is special

sqlmesh/core/node.py
tobymao: no, audits also have these properties

sqlmesh/core/model/definition.py
tobymao: because it's annoying to deal with cached_properties in cleanup

sqlmesh/core/context.py
tobymao: happens when you to context.upsert_model, used in tests, and notebooks

sqlmesh/core/model/definition.py
```diff
@@ -995,6 +1012,25 @@ class SqlModel(_SqlBasedModel):
...
+        model.__dict__.pop("column_descriptions", None)
+        model._columns_to_types = None
+        if "query" in kwargs.get("update", {}):
```
izeigerman, line 1030: We should check for `depends_on` here too.

izeigerman (approved):

PR #3010, opened 2024-08-14
feat: automatically load signals from the `signals/` subfolder

plaflamme:
This is a follow-up to https://github.com/TobikoData/sqlmesh/pull/2822 to allow loading a `signal_factory` from the sqlmesh project structure instead of having to specify it to the `Context`. This allows using built-in signals without any extra configuration in `config.py` or similar.

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -102,7 +102,7 @@ def signal_factory(f: SignalFactory) -> None:
     global _registered_signal_factory
 
     if _registered_signal_factory is not None:
-        raise RuntimeError("only one function may be decorated with @signal_factory")
+        raise SQLMeshError("Only one function may be decorated with @signal_factory")
 
     _registered_signal_factory = f
 
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -78,6 +78,33 @@ class Signal(abc.ABC):
 
 
 SignalFactory = t.Callable[[t.Dict[str, t.Union[str, int, float, bool]]], Signal]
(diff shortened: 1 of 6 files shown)


sqlmesh/core/scheduler.py
```diff
@@ -78,6 +78,33 @@ def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
...
+
+        @signal_factory
+        def my_signal_factory(signal_metadata: t.Dict[str, t.Union[str, int, float, bool]]) -> Signal:
+            return AlwaysReadySignal()
+    """
+
+    global _registered_signal_factory
+
+    if _registered_signal_factory is not None:
+        raise RuntimeError("only one function may be decorated with @signal_factory")
```
izeigerman, line 105:
```suggestion
        raise SQLMeshError("Only one function may be decorated with @signal_factory")
```

sqlmesh/core/scheduler.py
```diff
@@ -78,6 +78,33 @@ def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
...
+        class AlwaysReadySignal(Signal):
+            def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
+                return True
+
+        @signal_factory
+        def my_signal_factory(signal_metadata: t.Dict[str, t.Union[str, int, float, bool]]) -> Signal:
+            return AlwaysReadySignal()
+    """
+
+    global _registered_signal_factory
```
izeigerman, line 102: Let's add a test for this in `test_scheduler.py`.

izeigerman (approved):
Awesome, thank you!

PR #3013, opened 2024-08-15
Chore!: Set interval end per model in plan

izeigerman:
Previously we were using one end date per plan for all models that are part of the plan. This approach has proven problematic because, at times, SQLMesh derived incorrect missing intervals for some models when making changes in dev, especially when the cron expression didn't align with day / hour boundaries.

This update ensures that each individual model receives its own end date value, ensuring that SQLMesh correctly backfills changes up to what is already available in the target environment.

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -22,7 +22,7 @@ from sqlmesh.schedulers.airflow.client import AirflowClient
 from sqlmesh.schedulers.airflow.mwaa_client import MWAAClient
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.hashing import md5
-from sqlmesh.utils.pydantic import model_validator, model_validator_v1_args
+from sqlmesh.utils.pydantic import model_validator, model_validator_v1_args, field_validator
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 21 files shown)


sqlmesh/core/config/scheduler.py
```diff
@@ -266,7 +278,7 @@ class CloudComposerSchedulerConfig(_BaseAirflowSchedulerConfig, BaseConfig, extr
             whether a DAG has been created.
         backfill_concurrent_tasks: The number of concurrent tasks used for model backfilling during plan application.
         ddl_concurrent_tasks: The number of concurrent tasks used for DDL operations (table / view creation, deletion, etc).
-        max_snapshot_ids_per_request: The maximum number of snapshot IDs that can be sent in a single HTTP GET request to the Airflow Webserver.
+        max_snapshot_ids_per_request: The maximum number of snapshot IDs that can be sent in a single HTTP GET request to the Airflow Webserver (Deprecated).
```
tobymao, line 281: should we just remove this? without the deprecation warning

izeigerman: This will break existing users who set this property. I know for a fact there are people who use it.

tobymao (approved):

PR #3016, opened 2024-08-16
Fix: Exponentially multiplying removed interval records

### LINKED ISSUE #3015: Restating an incremental model can cause exponential growth in the `_intervals` state table
This script will reproduce the issue:

```shell
cd examples/sushi
rm -f data/duckdb.db
sqlmesh --gateway=duckdb_persistent plan --no-prompts --auto-apply
while true;
do
  sqlmesh --gateway=duckdb_persistent plan --restate-model sushi.waiter_revenue_by_day --start 2024-08-14 --end 2024-08-14 --no-prompts --auto-apply &> /dev/null
  python -c 'import duckdb;c=duckdb.connect("data/duckdb.db");print(c.execute("select count(*) from sqlmesh._intervals").fetchone()[0]);'
done
```

Output for me was:

```
27
45
99
261
747
2205
6579
19701
59067
177165
531459
1594341
```

izeigerman:
Closes #3015

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -657,6 +657,7 @@ class EngineAdapterStateSync(CommonStateSyncMixin, StateSync):
                             exp.select("name", "identifier", "version")
                             .from_(self.intervals_table)
                             .where(where)
+                            .distinct()
                         )
                     ]
                 )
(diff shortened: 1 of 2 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -657,6 +657,7 @@ def remove_interval(
                             exp.select("name", "identifier", "version")
                             .from_(self.intervals_table)
                             .where(where)
+                            .distinct()
```
tobymao, line 660: why were there ever duplicates?

tobymao (approved):

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Wdym? There can be multiple interval records per snapshot which is completely normal and expected

tobymao: ah i see, this is just fetching if any interval exists, you don't actually care about the interval

PR #3018, opened 2024-08-16
Chore!: Introduce the snapshot cache to the state sync

izeigerman:
This update introduces a file cache for snapshots fetched from the state, which should significantly speed up plan generation time.

### DIFF
--- a/sqlmesh/core/snapshot/cache.py
+++ b/sqlmesh/core/snapshot/cache.py
@@ -44,6 +44,7 @@ class SnapshotCache:
         if snapshot_ids_to_load:
             loaded_snapshots = loader(snapshot_ids_to_load)
             for snapshot in loaded_snapshots:
+                self._update_node_hash_cache(snapshot)
                 self.put(snapshot)
                 snapshots[snapshot.snapshot_id] = snapshot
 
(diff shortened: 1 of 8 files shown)


sqlmesh/core/snapshot/cache.py
```diff
@@ -0,0 +1,69 @@
...
+
+import typing as t
+
+from pathlib import Path
+from sqlmesh.core.model.cache import OptimizedQueryCache
+from sqlmesh.core.snapshot.definition import Snapshot, SnapshotId
+from sqlmesh.utils.cache import FileCache
+
+
+class SnapshotCache:
```
tobymao, line 11: should this not be in the state sync cache?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -212,6 +213,8 @@ def push_snapshots(self, snapshots: t.Iterable[Snapshot]) -> None:
 
         if snapshots:
             self._push_snapshots(snapshots)
+            for snapshot in snapshots:
+                self._snapshot_cache.put(snapshot)
```
tobymao, line 217: is this going to reload the optimized query cache even if it's not necessary? we should make sure with_optimized_query only loads when necessary

sqlmesh/core/snapshot/cache.py
izeigerman: I think this is pretty independent from the context in which it's used.

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Wdym by "reload"? Reading from the file system?

tobymao (approved):

PR #3019, opened 2024-08-16
Feat!: add clickhouse engine adapter

treysp:
Current limitations:
- Models requiring data type inference may need manually specified column types
- SCD Type 2 models not supported
- dbt adapter not supported

Main changes impacting base:
- `_drop_tablelike_object` and similar methods have been generalized to `_drop_object`
- Clickhouse does not allow schema definitions in views, so a new `SUPPORTS_VIEW_SCHEMA` flag has been added

New concept "engine run mode":
- CH has three "modes": standalone, cluster, and cloud.
    - If cluster, we have to inject the cluster name into create/drop/alter statements. 
    - SQLMesh can generally only use a single cluster to create schemas, but theoretically someone could want individual models created on other clusters. For now, we only allow a single cluster specified in the connection `cluster` arg.
- Represented by an engine adapter property in the `EngineRunMode` enum. Values:
  - SINGLE_MODE_ENGINE (for engines whose behavior is constant across modes)
  - STANDALONE
  - CLUSTER
  - CLOUD
- Base engine adapter returns `SINGLE_MODE_ENGINE`, only overridden in clickhouse


tests/core/engine_adapter/docker/clickhouse/full.xml
```diff
@@ -0,0 +1,1690 @@
+<!-- TODO: I dumped this from the docker image so I could see the full extent of the default config to decide what needed to be overridden -->
+<!-- It needs to be removed at some point -->
```
eakmanrq, line 2: By at some point, do you mean before merging this PR or sometime in the future?

sqlmesh/core/config/connection.py
```diff
@@ -1350,6 +1350,50 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
...
+    Clickhouse Connection Configuration.
+
+    Property reference: https://clickhouse.com/docs/en/integrations/python#client-initialization
+    """
+
+    host: str
+    username: str
+    password: t.Optional[str] = None
+    port: t.Optional[int] = None
+    default_cluster: str = "default"
```
eakmanrq, line 1364: Just confirming: This is the default name Clickhouse assigns if I don't provide a name? Basically just making sure `default` is a reasonable default to use or if we should require this field.

treysp: My understanding is that if you stand up a zookeeper instance for managing the cluster, it will automatically create a cluster named `default`.

@erindru Is that right?

tests/core/engine_adapter/docker/clickhouse/full.xml
treysp: @erindru

eakmanrq: Mostly asking because if we plan to merge this in I don't really even know from the message what needs to be changed in order to remove this.

erindru: Yeah this can be removed now, I was just referring to it in order to create the other xml files

sqlmesh/core/config/connection.py
erindru: Not quite. Basically there's always a cluster called "default" if you do `select * from system.clusters` but typically you'd give your cluster an actual name when you configure one. This results in two entries in `system.clusters`, one for "default" and one for your actual cluster.

Defaulting this to "default" would make queries work but it also might be surprising for users who were expecting it to pick their configured cluster. Perhaps leaving it unset by default and then throwing an exception if it's needed at runtime (ie, when running in cluster mode) to prompt the user to set it would be better?

Or if it's unset, querying `system.clusters` and if there's only one use that, otherwise raise an exception asking the user to set a default one?

treysp: > Perhaps leaving it unset by default and then throwing an exception if it's needed at runtime (ie, when running in cluster mode) to prompt the user to set it would be better?

That sounds good

treysp: Oh yeah - it also occurred to me that someone might have a cluster available but want to run in standalone mode. Currently. we will automatically use the cluster if it's detected.

What do you think about adding a `force_standalone_mode` arg to the connection so people can bypass the cluster if needed?

erindru: I guess since I've not used Clickhouse before I'm not familiar with the different ways people use it. 

Another way to do this could be to assume standalone mode if the `cluster` property isn't set, rather than throwing an error or having another property to force standalone mode.

Rather than trying to guess all the different usage combinations up front we could also just let them naturally come out of the woodwork when people start using this feature

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1052,11 +1052,18 @@ def _replace_query_for_model(self, model: Model, name: str, query_or_df: QueryOr
...
             storage_format=model.storage_format,
             partitioned_by=model.partitioned_by,
+            partitioned_by_user_cols=model.partitioned_by_user_columns,
             partition_interval_unit=model.interval_unit,
             clustered_by=model.clustered_by,
             table_properties=model.physical_properties,
             table_description=model.description,
             column_descriptions=model.column_descriptions,
+            ordered_by=[
+                *(model.grains if model.grains else []),
```
tobymao, line 1062: why are you using grain here?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,465 @@
...
+    ) -> t.List[SourceQuery]:
+        temp_table = self._get_temp_table(target_table, **kwargs)
+
+        def query_factory() -> Query:
+            # It is possible for the factory to be called multiple times and if so then the temp table will already
+            # be created so we skip creating again. This means we are assuming the first call is the same result
+            # as later calls.
+            if not self.table_exists(temp_table):
+                self.create_table(
+                    temp_table, columns_to_types, storage_format=exp.Var(this="MergeTree"), **kwargs
```
tobymao, line 156: exp.var("MergeTree")

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,465 @@
...
+        if storage_format:
+            table_engine = (
+                storage_format.this if isinstance(storage_format, exp.Var) else storage_format  # type: ignore
+            )
+        properties.append(exp.EngineProperty(this=table_engine))
+
+        # copy of table_properties so we can pop items off below then consume the rest later
+        table_properties_copy = {}
+        if table_properties:
+            table_properties_copy = table_properties.copy()
```
tobymao, line 314: why do you need to copy it? you already copy it on the next line 315

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,465 @@
...
+        # `view_properties`` is model.virtual_properties during view promotion but model.physical_properties elsewhen
+        # - in the create_view promotion call, the cluster won't be in `view_properties`` so we pass the
+        #     model.physical_properties["CLUSTER"] value to the "physical_cluster" kwarg
+        if self.engine_run_mode.is_cluster:
+            on_cluster = (
+                view_properties_copy.pop("CLUSTER", None)
+                or kwargs.pop("physical_cluster", None)
+                or self.cluster
+            )
+            properties.append(exp.OnCluster(this=exp.Var(this=on_cluster)))
```
tobymao, line 415: var

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1811,7 +1840,7 @@ def _intervals(snapshot: Snapshot, deployability_index: DeployabilityIndex) -> I
 
 def _check_destructive_schema_change(
     snapshot: Snapshot,
-    alter_expressions: t.List[exp.AlterTable],
+    alter_expressions: t.List[exp.Alter],
```
tobymao, line 1843: why was this changed? isn't this already in main a while ago? did you not rebase?

treysp: it's not yet in main #3003

.circleci/continue_config.yml
```diff
@@ -328,10 +329,11 @@ workflows:
                 - databricks
                 - redshift
                 - bigquery
-          filters:
-            branches:
-              only:
-                - main
+                - clickhouse-cloud
+          # filters:
```
erindru, line 333: Probably want to un-do this prior to merge

treysp: that's my intention - should have put "TODO" on there

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,456 @@
...
+            exists=ignore_if_not_exists,
+            kind="DATABASE",
+            cascade=None,
+            cluster=exp.OnCluster(this=exp.to_identifier(self.cluster))
+            if self.engine_run_mode.is_cluster
+            else None,
+            **drop_args,
+        )
+
+    # TODO: `RENAME` is valid SQL, but `EXCHANGE` is an atomic swap
```
erindru, line 117: Should this be removed if its unused?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,456 @@
...
+        """
+        target_table = exp.to_table(target_table_name)
+        source_table = exp.to_table(source_table_name)
+        on_cluster_sql = kwargs.get("ON_CLUSTER", None) or (
+            self.cluster if self.engine_run_mode.is_cluster else None
+        )
+        on_cluster_sql = (
+            f" ON CLUSTER {exp.to_identifier(on_cluster_sql)} " if on_cluster_sql else " "
+        )
+        create_sql = f"CREATE TABLE{' IF NOT EXISTS' if exists else ''} {target_table}{on_cluster_sql}AS {source_table}"
```
erindru, line 220: Should this be done with sqlglot builders instead of strings?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,456 @@
...
+        table_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        table_description: t.Optional[str] = None,
+        table_kind: t.Optional[str] = None,
+        empty_ctas: bool = False,
+    ) -> t.Optional[exp.Properties]:
+        properties: t.List[exp.Expression] = []
+
+        # table engine, default `MergeTree`
+        table_engine = "MergeTree"
```
erindru, line 290: Should this be a class level constant?

sqlmesh/core/engine_adapter/shared.py
```diff
@@ -192,6 +192,29 @@ def is_multi_catalog_supported(self) -> bool:
...
 
 
+class EngineRunMode(Enum):
+    SINGLE_MODE_ENGINE = 1
+    STANDALONE = 2
+    CLUSTER = 3
+    CLOUD = 4
+
+    @property
+    def is_single_mode_engine(self) -> bool:
```
erindru, line 202: What's the difference between single mode engine and standalone?

sqlmesh/core/model/meta.py
```diff
@@ -354,8 +354,9 @@ def unique_key(self) -> t.List[exp.Expression]:
 
     @property
     def partitioned_by(self) -> t.List[exp.Expression]:
+        """Columns to partition the model by, including the time column if it is not already included."""
         if self.time_column and self.time_column.column not in [
-            col for col in self._partition_by_columns
+            col for col in self.partitioned_by_user_columns
```
erindru, line 359: Ah I see, this explains `partition_by_user_columns`.

Previously `_partition_by_columns` existed but was private, you made it public and exposed it as `partitioned_by_user_columns` to differentiate between columns the user set vs columns sqlmesh injected.

I wonder if this can be cleaned up in future by having engine adapters indicating whether or not its appropriate to include the time_column in the list of columns to partition by. Then we might be able to use a single property again

tests/core/engine_adapter/docker/clickhouse/full.xml
erindru, line 1: This file needs to be removed

sqlmesh/core/engine_adapter/shared.py
treysp: I was thinking that there are some engines that can run standalone or in a cluster-like configuration without any changes in behavior. SQLMesh likely wouldn't even know which way it's configured in, so it's not really accurate to say that it's "in" any particular mode.

So the idea is to distinguish engines whose behavior is constant regardless of mode from engines whose mode is relevant for sqlmesh behavior.

The mode enum isn't used outside clickhouse right now, though, so it kind of doesn't matter what value it's set to for other engines.

sqlmesh/core/model/meta.py
treysp: Ah, that's a cleaner solution for sure - should be easy to add

sqlmesh/core/engine_adapter/shared.py
treysp: The name "single mode engine" is definitely not consistent with my explanation, though.

If we keep that value at all, maybe change to `NO_MODE_BEHAVIOR_CHANGES` or similar?

sqlmesh/core/engine_adapter/clickhouse.py
treysp: It's not currently parseable/generatable with sqlglot - i'll add it later this week and update

sqlmesh/core/model/meta.py
treysp: Hmm...meta.py doesn't have access to the engine adapter here

treysp: Another approach would be to convert partitioned_by into a dict, with user columns and time columns stored separately.

But that would mean collapsing it into a list everywhere it's used.

erindru: Yeah I think your solution is correct given the current design :+1:

sqlmesh/core/engine_adapter/shared.py
erindru: Hmm, it still feels like a Clickhouse abstraction leak. Is it actually needed at all outside the internals of the Clickhouse adapter? I didn't notice any checks for `engine_run_mode` in the evaluator

treysp: no, it's just internal to the CH adapter

erindru (approved):

sqlmesh/core/engine_adapter/base.py
```diff
@@ -842,20 +848,24 @@ def drop_managed_table(self, table_name: TableName, exists: bool = True) -> None
...
+        name: TableName | SchemaName,
+        exists: bool = True,
+        kind: str = "TABLE",
+        **drop_args: t.Any,
     ) -> None:
-        """Drops a "tablelike object".
+        """Drops an object.
 
-        A "tablelike object" could be a TABLE or a DYNAMIC TABLE or a TEMPORARY TABLE etc depending on the context.
+        A an object could be a DATABASE, SCHEMA, VIEW, TABLE, DYNAMIC TABLE, TEMPORARY TABLE etc depending on the :kind.
```
georgesittas, line 860:
```suggestion
        An object could be a DATABASE, SCHEMA, VIEW, TABLE, DYNAMIC TABLE, TEMPORARY TABLE etc depending on the :kind.
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1002,34 +1016,53 @@ def create_schema(
...
-                kind="SCHEMA",
-                exists=ignore_if_not_exists,
-                cascade=cascade,
-            )
+        return self._drop_object(
+            name=schema_name,
+            exists=ignore_if_not_exists,
+            kind="SCHEMA",
+            cascade=cascade,
         )
```
georgesittas, line 1066: Should we propagate `drop_args` to `_drop_object` here? Or was it left out intentionally?

tests/core/engine_adapter/test_integration.py
```diff
@@ -2208,6 +2235,15 @@ def test_to_time_column(
...
         pytest.skip("Time column tests only need to run for query")
 
+    if ctx.dialect == "clickhouse" and time_column_type.is_type(exp.DataType.Type.TIMESTAMPTZ):
+        # Clickhouse does not have natively timezone-aware types and does not accept timestrings
+        #   with UTC offset "+XX:XX". Therefore, we remove the timezone offset and set a timezone-
+        #   specific data type to validate what is returned.
+        import re
+
+        time_column = re.match("^(.*?)\+", time_column).group(1)
+        time_column_type = exp.DataType.build("TIMESTAMP('UTC')", dialect="clickhouse")
```
georgesittas, line 2245: @erindru will we still need this after we merge this change https://github.com/tobymao/sqlglot/pull/3938/files#diff-f2821f7be91ef6edbb6ed1683c3fa2ef31f23201961ba7757a4c37687f2ad896R120-R126? If not, let's leave a TODO to clean it up later.

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -0,0 +1,461 @@
...
+                        ordered_by_cols.append(col)
+                # python list of columns
+                elif isinstance(ordered_by_raw, list):
+                    ordered_by_cols.extend(ordered_by_raw)
+
+            ordered_by_expressions = (
+                exp.Tuple(expressions=[exp.to_column(k) for k in ordered_by_cols])  # type: ignore
+                if ordered_by_cols
+                # default () if no columns provided
+                else exp.Literal(this="()", is_string=False)
```
georgesittas, line 319: Why are we constructing `()` like this? Can't we just create a `Tuple` with an empty expression list? So there's no need for a branch.

georgesittas (approved):

sqlmesh/core/engine_adapter/base.py
treysp: good catch

ravenac95:
I don't know if you guys have run into this for these set of changes, but I was testing with a model: https://github.com/opensource-observer/oso/pull/1981/files#diff-8d421f6bf18815bdb75c0ee1801fbe9f024df45015d11c8845eb077312b59356R13

I had to set the types here because by default, the `CREATE` command that was generated if I didn't specify the types for these columns was nullable. I actually attempted to make this so it was written with `postgres` syntax (just as a test) like this:

```
  columns (
    bucket_day DATE NOT NULL, 
    event_source VARCHAR NOT NULL,
    to_artifact_id VARCHAR NOT NULL,
    metric VARCHAR NOT NULL,
    amount BIGINT
  )
```

However, this too produced nullable columns when transpiled to clickhouse sql. For now, I'll set it this way so it's not blocking us but I figured I'd mention it. We were able to successfully deploy our simple sqlmesh project on clickhouse.

treysp:
Hi @ravenac95 - yeah, nullability makes transpilation to Clickhouse painful. 

Usually nullability is a constraint on a column in a specific object, while in Clickhouse it's a property of a data type independent of any object.

So, conceptually, in non-Clickhouse dialects all data types are nullable because they aren't tied to a specific object. Therefore, transpilation always generates nullable Clickhouse columns.

In a SQLMesh model you are specifying types for a specific object, though, so transpiling the `NOT NULL` constraint to a non-nullable Clickhouse column makes sense and may be doable. I'll think more about this.

treysp:
@ravenac95 To clarify the issue you ran into, in your initial test was the model dialect postgres? Or the model dialect was clickhouse and you used the postgres `DATE NOT NULL` syntax in the `columns()` block?

ravenac95:
> @ravenac95 To clarify the issue you ran into, in your initial test was the model dialect postgres? Or the model dialect was clickhouse and you used the postgres `DATE NOT NULL` syntax in the `columns()` block?

I set it to postgres if memory serves correctly. Let me double check!

ravenac95:
Can confirm, it is failing if I set the dialect to `postgres`

This is what I see. I have this in my model:

```
  columns (
    bucket_day DATE NOT NULL, 
    event_source VARCHAR NOT NULL,
    to_artifact_id VARCHAR NOT NULL,
    metric VARCHAR NOT NULL,
    amount BIGINT
  )
``` 

I see this SQL being generated by `sqlmesh.core.engine_adapter.base.EngineAdapter#_build_create_table_exp`:

```
CREATE TABLE IF NOT EXISTS
sqlmesh__metrics.metrics__int_timeseries_metrics_by_artifact_developers__3352340965_
_temp (bucket_day Nullable(DATE), event_source Nullable(String), to_artifact_id
Nullable(String), metric Nullable(String), amount Nullable(Int64)) ENGINE=MergeTree
ORDER BY () PARTITION BY ("bucket_day")
```

And I get this error from clickhouse:

```
Error: Failed processing SnapshotId<"metrics"."int_timeseries_metrics_by_artifact_developers": 2616927852>. HTTPDriver for https://xxxxxx.xxxxx.gcp.clickhouse.cloud:8443 returned response code 400)
 Code: 44. DB::Exception: Partition key contains nullable columns, but merge tree setting `allow_nullable_key` is disabled. (ILLEGAL_COLUMN) (version 24.5.1.22957 (official build))

```

treysp:
And unless you specify `columns` the model you linked to fails with the same nullable key error, even with dialect set to clickhouse?

ravenac95:
> And unless you explicitly specify `columns`, the model you linked to fails with the same nullable key error, even with dialect set to clickhouse?

Yes. Though I get a slightly different error:

This is the generated SQL:
```
CREATE TABLE IF NOT EXISTS
sqlmesh__metrics.metrics__int_timeseries_metrics_by_artifact_developers__3033528082_
_temp (bucket_day Nullable, to_artifact_id Nullable(String), event_source
Nullable(String), metric Nullable(String), amount Nullable(Int64)) ENGINE=MergeTree
ORDER BY ()
```

It doesn't seem to get the correct column type for the derived column `bucket_day` from the upstream model. 

And I get this error:

```
Error: Failed processing SnapshotId<"metrics"."int_timeseries_metrics_by_artifact_developers": 3131413473>. HTTPDriver for https://w541fomho5.us-central1.gcp.clickhouse.cloud:8443 returned response code 500)
 Code: 42. DB::Exception: Nullable data type family must have exactly one argument - nested type. (NUMBER_OF_ARGUMENTS_DOESNT_MATCH) (version 24.5.1.22957 (official build))
```

georgesittas:
This is the culprit: `bucket_day Nullable`. @treysp do you see anything here that would cause this?

owensilk:
@treysp thanks so much for this work to add support for Clickhouse. Do you know if there is any plans for supporting the dbt integration?

treysp:
Hi @owensilk - this PR was to get the core functionality in place. I'm currently working on adding the sqlmesh features not included in this PR and the dbt adapter.

Would you like me to let you know when the PR including the dbt adapter goes up?

owensilk:
> Hi @owensilk - this PR was to get the core functionality in place. I'm currently working on adding the sqlmesh features not included in this PR and the dbt adapter.
> 
> Would you like me to let you know when the PR including the dbt adapter goes up?

Amazing thanks @treysp for the reply. Yes would love to hear when this is released. SQL mesh is a great project and keen to try and adopt. Thank you

PR #3020, opened 2024-08-16
fix: slack user warning :: the top level `text` argument is missing in the request payload

zzztimbo:
Each time the SQLMesh CI bot runs, we see these warnings:
```
/home/runner/.cache/pypoetry/virtualenvs/pipe-dwh-Jd5RneRz-py3.11/lib/python3.11/site-packages/slack_sdk/web/internal_utils.py:288: UserWarning: The top-level `text` argument is missing in the request payload for a chat.postMessage call - It's a best practice to always provide a `text` argument when posting a message. The `text` argument is used in places where content cannot be rendered such as: system push notifications, assistive technology such as screen readers, etc.
  warnings.warn(missing_text_message, UserWarning)
/home/runner/.cache/pypoetry/virtualenvs/pipe-dwh-Jd5RneRz-py3.11/lib/python3.11/site-packages/slack_sdk/web/internal_utils.py:289: UserWarning: Additionally, the attachment-level `fallback` argument is missing in the request payload for a chat.postMessage call - To avoid this warning, it is recommended to always provide a top-level `text` argument when posting a message. Alternatively you can provide an attachment-level `fallback` argument, though this is now considered a legacy field (see https://api.slack.com/reference/messaging/attachments#legacy_fields for more details).
  warnings.warn(missing_fallback_message, UserWarning)
```

Here is an issue that describes the problem: https://github.com/slackapi/bolt-js/issues/1249

I tested both the Slack Webhook and Slack API notifications and they seem to still work without issues. I'm not quite sure how to verify what the `text` field actually looks like though.

### DIFF
--- a/sqlmesh/core/notification_target.py
+++ b/sqlmesh/core/notification_target.py
@@ -283,6 +283,8 @@ class BaseSlackNotificationTarget(BaseNotificationTarget):
             ),
         )
 
+        composed.add_text(msg)
+
         self._send_slack_message(
             composed=composed.slack_message,
         )
@@ -318,6 +320,7 @@ class SlackWebhookNotificationTarget(BaseSlackNotificationTarget):
 
     def _send_slack_message(self, composed: slack.TSlackMessage) -> None:
         self.client.send(
+            text=composed["text"],
             blocks=composed["blocks"],
             attachments=composed["attachments"],  # type: ignore
         )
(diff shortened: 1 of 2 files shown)


tobymao:
thanks tim, can you fix the build? run make style

izeigerman (approved):

PR #3028, opened 2024-08-21
Fix: introspect function defaults to extract global refs for serialization

georgesittas:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1724232161253179

In current main, having a macro like the following doesn't work because we don't walk the function parameters to extract global references, such as modules and the like.

```
from sqlglot import exp
from sqlmesh import macro

@macro()
def test_default_param(evaluator, boolean_var: exp.Boolean = exp.true()):
  return boolean_var
```

### DIFF
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -76,6 +76,9 @@ def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
 
         func_args = next(node for node in ast.walk(root_node) if isinstance(node, ast.arguments))
         arg_defaults = (d for d in func_args.defaults + func_args.kw_defaults if d is not None)
+
+        # ast.Name corresponds to variable references, such as foo or x.foo. The former is
+        # represented as Name(id=foo), and the latter as Attribute(value=Name(id=x) attr=foo)
         arg_globals = [
             n.id for default in arg_defaults for n in ast.walk(default) if isinstance(n, ast.Name)
         ]
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -72,11 +72,18 @@ def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
     variables = {}
 
     if hasattr(func, "__code__"):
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/metaprogramming.py
```diff
@@ -177,9 +184,9 @@ def _decorator_name(decorator: ast.expr) -> str:
     return ""
 
 
-def decorators(func: t.Callable) -> t.List[str]:
+def decorators(func: t.Callable, root_node: t.Optional[ast.Module] = None) -> t.List[str]:
```
georgesittas, line 187: Note: added this to avoid repeating the same `parse_source(func)`.

sqlmesh/utils/metaprogramming.py
```diff
@@ -72,11 +72,18 @@ def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
...
 
     if hasattr(func, "__code__"):
-        code = func.__code__
+        root_node = parse_source(func)
+
+        func_args = next(node for node in ast.walk(root_node) if isinstance(node, ast.arguments))
+        arg_defaults = (d for d in func_args.defaults + func_args.kw_defaults if d is not None)
+        arg_globals = [
+            n.id for default in arg_defaults for n in ast.walk(default) if isinstance(n, ast.Name)
+        ]
```
georgesittas, line 81: Here `Name` corresponds to variable references, such as `foo` or `x.foo`. In the former case, you'd have a single `Name` node, whereas in the latter you'd have an `Attribute(value=Name(), attr=...)` node.

There's currently a known bug where if you had:

```python
from sqlmesh import macro

class a:
    class b:
        value = 5

@macro()
def foo(evaluator, arg=a.b.value):
    pass
```

This wouldn't work, because of the order with which we're hydrating the python env.: we're first executing the function definition payload, and then the class definition, so the former one doesn't have the class readily available and crashes with a `NameError: a is not defined`.

georgesittas: I'll probably add a comment for this.

tobymao (approved):

Issue #2698, opened 2024-05-29
Seed column & model column mismatch due to normalization


georgesittas:
To reproduce the bug:

1. Create new SQLMesh project using `sqlmesh init postgres --template empty` - Postgres is the default dialect because its normalization strategy is `LOWERCASE`, i.e. it treats quoted identifiers as case-sensitive, and lowercases unquoted ones.
2. Create the following seed model:

```sql
MODEL (
    name sqlmesh_example.demo_seed,
    kind SEED (
        path '../seeds/demo_seed.csv'
    ),
    columns(
        "camelCaseColumn" TEXT,
    ),
);
```

3. Create the following seed file under `seeds/`, named `demo_seed.csv`:
```csv
"camelCaseColumn"
foo
```

4. Run `sqlmesh plan`.

This currently produces the following error:

```
  File "sqlmesh/core/model/definition.py", line 1182, in render_seed
    df.loc[:, string_columns] = df[string_columns].mask(
  File "sqlmesh/.venv/lib/python3.10/site-packages/pandas/core/frame.py", line 3899, in __getitem__
    indexer = self.columns._get_indexer_strict(key, "columns")[1]
  File "sqlmesh/.venv/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6115, in _get_indexer_strict
    self._raise_if_missing(keyarr, indexer, axis_name)
  File "sqlmesh/.venv/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6176, in _raise_if_missing
    raise KeyError(f"None of [{key}] are in the [{axis_name}]")
KeyError: "None of [Index(['camelCaseColumn'], dtype='object')] are in the [columns]"
```

The reason this happens is because we [normalize](https://github.com/TobikoData/sqlmesh/blob/aa3ee94e1d41c18d41d95565dc1b7d0d4e331921/sqlmesh/core/model/seed.py#L87-L92) the seed file's column by lowercasing it, whereas the `column` property lists it with quotes, making it case-sensitive. We're essentially always treating CSV columns as case-insensitive, since there's no way to quote them and preserve those quotes in the dataframe returned by `pd.read_csv`.

Some ideas to fix this:

[...]

PR #3036, opened 2024-08-22
Fix: Seed and model column mismatch due to normalization

Themiscodes:
This update aims to resolve a lookup issue in `render_seed` by renaming the columns that are looked up and can't be resolved (due to being [normalised](https://github.com/TobikoData/sqlmesh/blob/aa3ee94e1d41c18d41d95565dc1b7d0d4e331921/sqlmesh/core/model/seed.py#L78C9-L78C16)) with the names from the model definition columns, fixes: #2698

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1266,12 +1266,14 @@ class SeedModel(_SqlBasedModel):
     def render_seed(self) -> t.Iterator[QueryOrDF]:
         self._ensure_hydrated()
 
+        type_columns = []
         date_columns = []
         datetime_columns = []
         bool_columns = []
         string_columns = []
 
         for name, tpe in (self.columns_to_types_ or {}).items():
+            type_columns.append(name)
             if tpe.this in (exp.DataType.Type.DATE, exp.DataType.Type.DATE32):
                 date_columns.append(name)
             elif tpe.this in exp.DataType.TEMPORAL_TYPES:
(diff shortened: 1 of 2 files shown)


erindru:
Can you please write a test that demonstrates this bug is fixed?

sqlmesh/core/model/definition.py
```diff
@@ -1265,12 +1266,14 @@ def render(
     def render_seed(self) -> t.Iterator[QueryOrDF]:
         self._ensure_hydrated()
 
+        type_columns = []
```
georgesittas, line 1269: Why do we need the "type" prefix here?

sqlmesh/core/model/definition.py
```diff
@@ -1281,6 +1284,12 @@ def render_seed(self) -> t.Iterator[QueryOrDF]:
                 string_columns.append(name)
 
         for df in self._reader.read(batch_size=self.kind.batch_size):
+            for column in type_columns:
+                if column not in df:
+                    normalized_name = normalize_identifiers(column, dialect=self.dialect).name
+                    if normalized_name in df:
+                        df.rename(columns={normalized_name: column}, inplace=True)
```
georgesittas, line 1291: Should we do a single `rename` call here? We can do it in one go after the loop - only need to populate the dict within the loop.

Themiscodes:
Thank you @erindru and @georgesittas  for the review! I've addressed the comments and added a unit test

georgesittas (approved):

erindru (approved):

tests/core/test_model.py
```diff
@@ -786,6 +786,63 @@ def test_seed_provided_columns():
...
+            name db.seed,
+            dialect postgres,
+            kind SEED (
+              path '{str(model_csv_path)}',
+            ),
+            columns (
+              'camelCaseId' int,
+              'camelCaseBool' boolean,
+              'camelCaseString' text,
+              'camelCaseTimestamp' timestamp
```
georgesittas, line 811: These quotes denote strings, we should use double quotes instead.

PR #3039, opened 2024-08-23
fix!: rename renderer cache so it's not serialized

tobymao:


sqlmesh/core/model/definition.py
```diff
@@ -119,7 +119,7 @@ class _Model(ModelMeta, frozen=True):
     mapping_schema: t.Dict[str, t.Any] = {}
 
     _full_depends_on: t.Optional[t.Set[str]] = None
-    __statement_renderers: t.Dict[int, ExpressionRenderer] = {}
+    _renderer_cache: t.Dict[int, ExpressionRenderer] = {}
```
izeigerman, line 122: Let's clarify that this is not a query cache.

izeigerman (approved):

PR #3041, opened 2024-08-23
Fix: Don't cache snapshot intervals

izeigerman:
Intervals should never be cached so that they can be hydrated correctly after snapshots are loaded from cache.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -1207,6 +1207,14 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
         if not self.version:
             raise SQLMeshError(f"Snapshot {self.snapshot_id} has not been versioned yet.")
 
+    def __getstate__(self) -> t.Dict[t.Any, t.Any]:
+        state = super().__getstate__()
+        state["__dict__"] = state["__dict__"].copy()
+        # Don't store intervals.
+        state["__dict__"]["intervals"] = []
+        state["__dict__"]["dev_intervals"] = []
+        return state
+
 
 class SnapshotTableCleanupTask(PydanticModel):
     snapshot: SnapshotTableInfo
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -1207,6 +1207,14 @@ def _ensure_categorized(self) -> None:
         if not self.version:
             raise SQLMeshError(f"Snapshot {self.snapshot_id} has not been versioned yet.")
 
+    def __getstate__(self) -> t.Dict[t.Any, t.Any]:
+        state = super().__getstate__()
+        state["__dict__"] = state["__dict__"].copy()
+        # Don't store intervals.
+        state["__dict__"]["intervals"] = []
```
tobymao, line 1214: should we do this for copy too or no? we replicate these in models

tests/core/test_snapshot.py
```diff
@@ -1957,10 +1960,30 @@ def _loader(snapshot_ids: t.Set[SnapshotId]) -> t.Collection[Snapshot]:
...
+            query=parse_one("SELECT 1"),
+        )
+    )
+    snapshot.add_interval("2023-01-01", "2023-01-02")
+    snapshot.add_interval("2023-01-01", "2023-01-02", is_dev=True)
+
+    assert len(snapshot.intervals) > 0
+    assert len(snapshot.dev_intervals) > 0
+
+    loaded_snapshot = pickle.loads(pickle.dumps(snapshot))
```
tobymao, line 1987: you can also assert that the snapshot obj hasn't lost it's intervals

tobymao (approved):

sqlmesh/core/snapshot/definition.py
izeigerman: I've confirmed that we rely on the fact that intervals are preserved after copying. This is only relevant when pickling.

tests/core/test_snapshot.py
izeigerman: added

PR #3043, opened 2024-08-26
Fix: Heuristics that determine the default start date of the forward-only preview plan

izeigerman:
It should use the minimum value among the maximum interval end values to make a more accurate estimate for models whose cron is not aligned with a day boundary.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -40,7 +40,7 @@ import time
 import traceback
 import typing as t
 import unittest.result
-from datetime import timedelta
+from datetime import date, timedelta
 from functools import cached_property
 from io import StringIO
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


sqlmesh/core/context.py
```diff
@@ -1153,10 +1154,9 @@ def plan_builder(
             )
             if max_interval_end_per_model:
                 default_end = max(max_interval_end_per_model.values())
-        else:
-            default_end = None
-
-        default_start = to_date(default_end) - timedelta(days=1) if default_end and is_dev else None
+                default_start = to_date(min(max_interval_end_per_model.values())) - timedelta(
```
tobymao, line 1157: why don’t you need the else branch anymore?

tobymao (approved):

sqlmesh/core/context.py
izeigerman: Because we initialize this value to None at the top. No need to do set None again.

PR #3046, opened 2024-08-26
Fix: Handle missing source target aliases in sample data diff

Themiscodes:
This update adds placeholder environment source and target names to the joined sample (when `source_alias` and `target_alias` are not defined during the construction of the TableDiff) in order to differentiate the output results in row_diff and display results when calling `show_row_diff`.

### DIFF
--- a/tests/core/test_table_diff.py
+++ b/tests/core/test_table_diff.py
@@ -6,6 +6,7 @@ from sqlmesh.core import dialect as d
 from sqlmesh.core.context import Context
 from sqlmesh.core.config import AutoCategorizationMode, CategorizerConfig
 from sqlmesh.core.model import SqlModel, load_sql_based_model
+from sqlmesh.core.table_diff import TableDiff
 
 
 @pytest.mark.slow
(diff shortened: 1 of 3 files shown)


georgesittas (commented):
Let's add a test please

Themiscodes:
Thanks @georgesittas added unit test and addressed comments

tests/core/test_table_diff.py
```diff
@@ -134,8 +135,27 @@ def test_data_diff_decimals(sushi_context_fixed_date):
...
+    assert row_diff.full_match_count == 2
+    assert row_diff.partial_match_count == 1
+    assert "s__value" in joined_sample_columns
+    assert "t__value" in joined_sample_columns
+
+    table_diff = TableDiff(
+        adapter=engine_adapter,
+        source="table_diff_source",
+        target="table_diff_target",
+        source_alias="dev",
```
erindru, line 150: I thought this PR was for fixing the sample table when `source_alias` and `target_alias` aren't defined, but the test defines them here?

erindru: Ah right, I missed that :) Thanks for clarifying!

georgesittas (approved):

PR #3049, opened 2024-08-26
Fix: allow omitting the MODEL (...) block when name inference is activated

georgesittas:
Since model name inference is now [supported](https://sqlmesh.readthedocs.io/en/stable/guides/configuration/?h=#model-naming), the `MODEL` meta may not be needed when defining models. This PR relaxes some constraints in order to support models that only contain a query and optional pre/post-statements.

Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1724422066768009

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -604,7 +604,7 @@ def format_model_expressions(
     Returns:
         A string representing the formatted model.
     """
-    if len(expressions) == 1:
+    if len(expressions) == 1 and isinstance(expressions[0], (Audit, Metric, Model)):
         return expressions[0].sql(pretty=True, dialect=dialect)
 
     *statements, query = expressions
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -604,7 +604,7 @@ def format_model_expressions(
     Returns:
         A string representing the formatted model.
     """
-    if len(expressions) == 1 and isinstance(expressions[0], (Audit, Metric, Model)):
+    if len(expressions) == 1 and is_meta_expression(expressions[0]):
         return expressions[0].sql(pretty=True, dialect=dialect)
 
     *statements, query = expressions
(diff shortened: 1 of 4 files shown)


tests/core/test_model.py
```diff
@@ -334,19 +334,32 @@ def test_partitioned_by(
...
-    """
-    )
-
+def test_no_model_statement(tmp_path: Path):
+    # No name inference => MODEL (...) is required
+    expressions = d.parse("SELECT 1 AS x")
     with pytest.raises(
         ConfigError,
-        match="MODEL statement is required as the first statement in the definition at '.",
+        match="The MODEL statement is required as the first statement in the definition, unless model name inference is activated. at '.'",
```
izeigerman, line 342:
```suggestion
        match="The MODEL statement is required as the first statement in the definition, unless model name inference is enabled. at '.'",
```

sqlmesh/core/model/definition.py
```diff
@@ -1563,10 +1563,14 @@ def load_sql_based_model(
...
     meta = expressions[0]
     if not isinstance(meta, d.Model):
-        raise_config_error(
-            "MODEL statement is required as the first statement in the definition",
-            path,
-        )
+        if not infer_names:
+            raise_config_error(
+                "The MODEL statement is required as the first statement in the definition, "
+                "unless model name inference is activated.",
```
izeigerman, line 1569:
```suggestion
                "unless model name inference is enabled.",
```

sqlmesh/core/dialect.py
```diff
@@ -604,7 +604,7 @@ def format_model_expressions(
     Returns:
         A string representing the formatted model.
     """
-    if len(expressions) == 1:
+    if len(expressions) == 1 and isinstance(expressions[0], (Audit, Metric, Model)):
```
izeigerman, line 607: Looks like something that should be a helper function. Provided its usage in `context.py`.

georgesittas: Thanks for the suggestion, refactored it into an `is_meta_expression` helper under `dialect.py`.

izeigerman (approved):

PR #3061, opened 2024-08-27
Fix!: Retain indexes when creating state table backups

izeigerman:
Previously, we created backup state tables using the CTAS statement, which didn't retain any table indexes or constraints. 

This change fixes the issue and also adds a migration script that recreates the original indexes for those who may have restored from a backup in the past.

### DIFF
--- a/sqlmesh/migrations/v0056_restore_table_indexes.py
+++ b/sqlmesh/migrations/v0056_restore_table_indexes.py
@@ -1,0 +1,116 @@
+"""Readds indexes and primary keys in case tables were restored from a backup."""
+
+from sqlglot import exp
+from sqlmesh.utils import random_id
+from sqlmesh.utils.migration import index_text_type
+
+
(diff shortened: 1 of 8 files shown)


sqlmesh/migrations/v0056_restore_table_indexes.py
```diff
@@ -0,0 +1,116 @@
...
+    engine_adapter.create_table(
+        new_environments_table, environments_columns_to_types, primary_key=("name",)
+    )
+    engine_adapter.insert_append(
+        new_environments_table,
+        exp.select("*").from_(environments_table),
+        columns_to_types=environments_columns_to_types,
+    )
+
+    # Recreate the itnervals table and its indexes.
```
tobymao, line 93: intervals

tobymao (approved):
you were able to test the rollback script on mysql and postgres?

izeigerman:
@tobymao postgres. Still working on mysql

PR #3071, opened 2024-08-30
fix: always provide feedback during plan execution

plaflamme:
This fixes 2 issues:

* when a plan fails to apply for any reason, the bot provided no feedback and would appear running forever; this ensures that we always provide feedback by cacthing all exceptions
* when a plan fails to to a conflict, the user can typically solve this by simply `/deploy`ing again, so this handles that case by treating it like a `skipped` plan which is less scary than a failure.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -72,7 +72,7 @@ from sqlmesh.core.state_sync.common import transactional
 from sqlmesh.utils import major_minor, random_id, unique
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.date import TimeLike, now, now_timestamp, time_like_to_str, to_timestamp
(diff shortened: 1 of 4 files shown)


plaflamme:
@eakmanrq added some tests in https://github.com/TobikoData/sqlmesh/pull/3071/commits/7b85c8812810054e9d7c1aa9ea9a7be2e91c4862 not sure if that's what you had in mind, PTAL

eakmanrq:
> @eakmanrq added some tests in [7b85c88](https://github.com/TobikoData/sqlmesh/commit/7b85c8812810054e9d7c1aa9ea9a7be2e91c4862) not sure if that's what you had in mind, PTAL

Yeah looks good just need to fix the tests

plaflamme:
@eakmanrq sorry about that, everything looks good now

eakmanrq (approved):

PR #3073, opened 2024-08-30
feat: format check

etonlels:
Adds a `--check` flag to `sqlmesh format`

### DIFF
--- a/docs/reference/notebook.md
+++ b/docs/reference/notebook.md
@@ -437,3 +437,32 @@ options:
   --read READ    The input dialect of the sql string.
   --write WRITE  The output dialect of the sql string.
 ```
+
+#### format
+```
+%format [--transpile TRANSPILE] [--append-newline] [--normalize]
+              [--pad PAD] [--indent INDENT]
+              [--normalize-functions NORMALIZE_FUNCTIONS] [--leading-comma]
+              [--max-text-width MAX_TEXT_WIDTH] [--check]
+
+Format all SQL models and audits.
+
+options:
+  --transpile TRANSPILE, -t TRANSPILE
+                        Transpile project models to the specified dialect.
+  --append-newline      Whether or not to append a newline to the end of the
+                        file.
+  --normalize           Whether or not to normalize identifiers to lowercase.
+  --pad PAD             Determines the pad size in a formatted string.
+  --indent INDENT       Determines the indentation size in a formatted string.
+  --normalize-functions NORMALIZE_FUNCTIONS
+                        Whether or not to normalize all function names.
+                        Possible values are: 'upper', 'lower'
+  --leading-comma       Determines whether or not the comma is leading or
+                        trailing in select expressions. Default is trailing.
+  --max-text-width MAX_TEXT_WIDTH
+                        The max number of characters in a segment before
+                        creating new lines in pretty mode.
+  --check               Whether or not to check formatting (but not actually
+                        format anything).
+```
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -860,17 +860,18 @@ class GenericContext(BaseContext, t.Generic[C]):
         self,
         transpile: t.Optional[str] = None,
         append_newline: t.Optional[bool] = None,
+        *,
+        check: t.Optional[bool] = None,
         **kwargs: t.Any,
-    ) -> None:
+    ) -> bool:
         """Format all SQL models and audits."""
         format_targets = {**self._models, **self._audits}
         for target in format_targets.values():
             if not target._path.suffix == ".sql":
                 continue
             with open(target._path, "r+", encoding="utf-8") as file:
-                expressions = parse(
-                    file.read(), default_dialect=self.config_for_node(target).dialect
-                )
+                before = file.read()
+                expressions = parse(before, default_dialect=self.config_for_node(target).dialect)
                 if transpile and is_meta_expression(expressions[0]):
                     for prop in expressions[0].expressions:
                         if prop.name.lower() == "dialect":
@@ -882,15 +883,18 @@ class GenericContext(BaseContext, t.Generic[C]):
                             )
                 format = self.config_for_node(target).format
                 opts = {**format.generator_options, **kwargs}
-                file.seek(0)
-                file.write(
-                    format_model_expressions(expressions, transpile or target.dialect, **opts)
-                )
+                after = format_model_expressions(expressions, transpile or target.dialect, **opts)
                 if append_newline is None:
                     append_newline = format.append_newline
                 if append_newline:
-                    file.write("\n")
-                file.truncate()
+                    after += "\n"
+                if not check:
+                    file.seek(0)
+                    file.write(after)
+                    file.truncate()
+                elif before != after:
+                    return False
+        return True
 
     @python_api_analytics
     def plan(
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -184,6 +184,8 @@ Options:
                               trailing.
   --max-text-width INTEGER    The max number of characters in a segment before
                               creating new lines in pretty mode.
+  --check                     Whether or not to check formatting (but not
+                              actually format anything).
   --help                      Show this message and exit.
 ```
 
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -255,12 +255,19 @@ def evaluate(
     type=int,
     help="The max number of characters in a segment before creating new lines in pretty mode.",
 )
+@click.option(
+    "--check",
+    is_flag=True,
+    help="Whether or not to check formatting (but not actually format anything).",
+    default=None,
+)
 @click.pass_context
 @error_handler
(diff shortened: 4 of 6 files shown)


sqlmesh/core/context.py
```diff
@@ -855,22 +856,39 @@ def evaluate(
...
+    @overload
+    def format(
+        self,
+        *,
+        check: Literal[True],
+        **kwargs: t.Any,
+    ) -> bool: ...
+
+    @overload
+    def format(
```
izeigerman, line 868: Why do we need these overloaded methods?

etonlels: For typing purposes. When `True` is passed as the `check` argument, the return value is a boolean indicating whether everything is correctly formatted or not. When `None` or `False` are passed, the return value is `None` (i.e. the same as it is currently). This preserves the current "API" of this method when not performing a check. If you prefer, I could just make the return value a boolean always.

izeigerman: Ah, got it, makes sense. Didn't notice the return value.

FWIW, I think it's fine to always return a bool.

etonlels:
@izeigerman I noticed that the `format` method is also being referenced [here](https://github.com/TobikoData/sqlmesh/blob/ae6f97ff3d3cd74e43283570cb8d6e23e34ac399/sqlmesh/magics.py#L692-L742), but I'm not familiar with what that's for exactly. Should a change be made there as well?

sqlmesh/core/context.py
etonlels: fixed

georgesittas:
Hey 👋

> @izeigerman I noticed that the `format` method is also being referenced [here](https://github.com/TobikoData/sqlmesh/blob/ae6f97ff3d3cd74e43283570cb8d6e23e34ac399/sqlmesh/magics.py#L692-L742), but I'm not familiar with what that's for exactly. Should a change be made there as well?
 
Yeah, that's the notebook interface. We generally try to keep the CLI and notebook APIs in sync, so it makes sense to reflect your changes in there as well.

georgesittas (approved):
LGTM, should be good to go once the notebook API is updated.

georgesittas:
Let's update the documentation as well while we're at it (note that `format` is currently missing from notebook docs, but we should add that):

- https://sqlmesh.readthedocs.io/en/stable/reference/cli/
- https://sqlmesh.readthedocs.io/en/stable/reference/notebook/

The way I usually update these pages is I copy-paste the `--help` version of the command of interest in the corresponding markdown file.

etonlels:
@georgesittas I updated the CLI docs, but when I go to view the docs for `%format` in a Jupyter notebook by pressing `Shift+Tab`, I get the docs for Python's builtin `format` function, i.e. the following:
```
Signature: format(value, format_spec='', /)
Docstring:
Return value.__format__(format_spec)

format_spec defaults to the empty string.
See the Format Specification Mini-Language section of help('FORMATTING') for
details.
Type:      builtin_function_or_method
```
I'm guessing that's why it's not currently documented. I'm not an expert with Jupyter notebooks (I never use them). Does it make sense to override this behavior? Is that even possible?

georgesittas:
Interesting. In any case, we can also just fill in the docs by hand, mimicking the other magic command descriptions. Can also take this to the finish line myself later, so no worries.

@izeigerman do you want to take another look?

sqlmesh/core/analytics/__init__.py
etonlels, line 1: FYI I took the liberty of improving the typing in this file since I noticed it was a bit off while looking at something else. If you prefer, I can break this out into its own PR.

etonlels (commented):
> In any case, we can also just fill in the docs by hand, mimicking the other magic command descriptions.

I can take a stab at that.

etonlels:
@georgesittas @izeigerman I think I've addressed everything so far. PTAL when you get the chance. Thanks!

docs/reference/notebook.md
```diff
@@ -437,3 +437,32 @@ options:
...
+                        Whether or not to normalize all function names.
+                        Possible values are: 'upper', 'lower'
+  --leading-comma       Determines whether or not the comma is leading or
+                        trailing in select expressions. Default is trailing.
+  --max-text-width MAX_TEXT_WIDTH
+                        The max number of characters in a segment before
+                        creating new lines in pretty mode.
+  --check               Whether or not to check formatting (but not actually
+                        format anything).
+```
```
etonlels, line 468: > In any case, we can also just fill in the docs by hand, mimicking the other magic command descriptions.

@georgesittas in case you need to do this again in the future, I ended up temporarily changing the name of the `format` method to something like `formut` to avoid a naming collision, and then I copy-pasted from the `Shift+Tab` completion in a Jupyter notebook.

etonlels: Never mind. It's better to just execute a cell with `%format?` in it and copy-paste that. For some reason the `Shift-Tab` completion is off though...

sqlmesh/core/analytics/__init__.py
```diff
@@ -31,9 +31,13 @@ def disable_analytics() -> None:
     collector = AnalyticsCollector(dispatcher=NoopEventDispatcher())
 
 
-def cli_analytics(func: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
+P = t.ParamSpec("P")
+T = t.TypeVar("T")
+
+
+def cli_analytics(func: t.Callable[P, T]) -> t.Callable[P, T]:
```
georgesittas, line 38: It looks like there's an [alternative form](https://docs.python.org/3/library/typing.html#typing.ParamSpec) we may be able to use here:

```python
def add_logging[T, **P](f: Callable[P, T]) -> Callable[P, T]:
    '''A type-safe decorator to add logging to a function.'''
    def inner(*args: P.args, **kwargs: P.kwargs) -> T:
        logging.info(f'{f.__name__} was called')
        return f(*args, **kwargs)
    return inner
```

Haven't encountered this before, but perhaps we can get rid of the `P`, `T` definitions in favor of the more compact version? Ditto for `python_api_analytics`.

sqlmesh/core/context.py
```diff
@@ -860,17 +860,18 @@ def format(
         self,
         transpile: t.Optional[str] = None,
         append_newline: t.Optional[bool] = None,
+        *,
+        check: t.Optional[bool] = None,
```
georgesittas, line 864: Why do we need `*` here?

georgesittas (commented):
LGTM, thanks for the contribution @etonlels!

sqlmesh/core/analytics/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 import atexit
 import os
-import typing as t
+import typing_extensions as t
```
izeigerman, line 5: Is this a correct import? Somehow I was sure that these extension were promoted to `typing` in later versions of Python

tobymao: we should revert this change

sqlmesh/core/analytics/__init__.py
tobymao: none of these changes seem relevant, we should revert

sqlmesh/core/context.py
etonlels: I originally added it (in part) to help with the typing `@overload`s that I had (but later got the ok to remove). It's not techincally "needed" per se, but I think it's advisable. The `*` is a special parameter, which you can read about [here](https://docs.python.org/3/tutorial/controlflow.html#special-parameters). Essentially it ensures that the `check` argument is always used as a keyword argument and never a positional argument. This is helpful because something like `format(check=True)` is way more clear than just `format(True)`. Technically it would probably be even better for the `*` to be moved up 2 lines, but I was trying to avoid messing with your APIs too much.

sqlmesh/core/analytics/__init__.py
etonlels: > It looks like there's an [alternative form](https://docs.python.org/3/library/typing.html#typing.ParamSpec) we may be able to use here

I too would prefer that syntax, but AFAIK it's only supported in Python 3.12+. Based on your test matrix, it looks like you're targeting Python 3.8+, so unfortunately we need to use the old syntax for backwards compatibility. If you look at that [same page but for Python 3.11](https://docs.python.org/3.11/library/typing.html#typing.ParamSpec), you'll see that the example is different and matches what I have here.

> none of these changes seem relevant, we should revert

Sure thing 👍 @tobymao would you like me to make a separate PR for it?

georgesittas: Ah I see, I didn't notice this was introduced in later versions.

sqlmesh/core/analytics/__init__.py
etonlels: reverted

sqlmesh/core/analytics/__init__.py
tobymao: maybe? i'm not that interested in it if it's not 3.8+ compatable. but for sure let's remove it from this PR

PR #3076, opened 2024-08-31
Chore: Isolate web tests

vchan:
This PR attempts to fix flaky web/UI lineage tests (when run in parallel) by reseting FastAPI app dependency overrides in between tests and creating a new TestClient per test.

### DIFF
--- a/tests/web/conftest.py
+++ b/tests/web/conftest.py
@@ -1,7 +1,4 @@
-from __future__ import annotations
-
 from pathlib import Path
-import typing as t
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/web/conftest.py
```diff
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 from pathlib import Path
+import typing as t
```
tobymao, line 4: we don’t need types in tests

tests/web/conftest.py
```diff
@@ -8,7 +11,7 @@
 
 
 @pytest.fixture
-def project_tmp_path(tmp_path: Path) -> Path:
+def project_tmp_path(tmp_path: Path) -> t.Generator[Path, None, None]:
```
tobymao, line 14: this can just be like iterator right? or just remove the types

vchan: I'll remove the types.

PR #3077, opened 2024-09-02
feat: use multiprocessing to speed up loading

tobymao:


sqlmesh/core/loader.py
```diff
@@ -42,6 +45,24 @@
...
+
+def _initializer(optimized_query_cache: OptimizedQueryCache) -> None:
+    global _cache
+    _cache = optimized_query_cache
+
+
+def _load_cache(model: Model) -> t.Tuple[str, t.Optional[str]]:
+    assert _cache
+    if isinstance(model, SqlModel):
+        entry_name = _cache._entry_name(model)
```
izeigerman, line 59: Why do you need this? Why not just call put without an argument and let it compute the entry name?

sqlmesh/core/loader.py
```diff
@@ -42,6 +45,24 @@
 logger = logging.getLogger(__name__)
 
 
+_cache = None
```
izeigerman, line 48: Why do you need this global state? The cache instance should be very lightweight and it should be ok to just create an instance when loading.

sqlmesh/core/loader.py
izeigerman: Or can we just pass cache as an argument instead?

tobymao: because computing the hash entry isn't free, it calls gen which is a full traversal of ast

sqlmesh/core/loader.py
tobymao: every time?, this makes it so that every process initializes the cache once

sqlmesh/core/loader.py
izeigerman: Then I suggest making the `_entry_name` method public.

sqlmesh/core/loader.py
izeigerman: What is there to initialize?

tobymao: in order for each process to have access to the state, the pattern i've found online is to use a global

izeigerman (approved):

PR #3082, opened 2024-09-03
Feat: Support BigQuery nested column additions without dropping

### LINKED ISSUE #2321: Support mutating BigQuery nested types without dropping
This is a feature request for the BigQuery engine.

Currently, sqlmesh supports nested types (i.e.: `STRUCT`) within models, e.g.:

```sql
SELECT STRUCT(foo, bar) AS my_struct FROM underlying
```

It also supports changes to these, e.g.:

```sql
SELECT STRUCT(foo, bar, new_field) AS my_struct FROM underlying
```

But because BigQuery [does not support mutating nested types using SQL](https://cloud.google.com/bigquery/docs/managing-table-schemas#sql_1), the way this is implemented in sqlmesh requires dropping the column and creating a new one (which truncates the data).

The feature request is to fallback on using the [BigQuery API](https://cloud.google.com/bigquery/docs/managing-table-schemas#api_1) directly for doing these mutations. Specifically, it should be possible to add columns to a nested type without dropping it.

The caveat is that adding columns to a nested type can only be done "at the end" (i.e.: `STRUCT(foo, new_field, bar)` is not possible, only `STRUCT(foo, bar, new_field)`).

Ideally, sqlmesh would allow **appending** fields to a `STRUCT` as long as the query also matches the resulting schema.

Themiscodes:
This update adds support for adding fields to a nested type in BigQuery without dropping the existing column by using the ([BigQuery API](https://cloud.google.com/bigquery/docs/managing-table-schemas#api_1)), fixes #2321

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import typing as t
+from collections import defaultdict
 
 import pandas as pd
 from sqlglot import exp
@@ -233,10 +234,14 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin):
         and uses the API to add columns to structs, where SQL is not supported.
         """
 
-        alter_expressions = self.__add_nested_columns(alter_expressions)
-        if alter_expressions:
+        nested_fields, non_nested_expressions = self._split_alter_expressions(alter_expressions)
+
+        if nested_fields:
+            self._update_table_schema(nested_fields, alter_expressions[0].this)
+
+        if non_nested_expressions:
             with self.transaction():
-                for alter_expression in alter_expressions:
+                for alter_expression in non_nested_expressions:
                     self.execute(alter_expression)
 
     def fetchone(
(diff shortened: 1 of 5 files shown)


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -261,6 +276,63 @@ def fetchall(
...
+        for alter_expression in alter_expressions:
+            action = alter_expression.args["actions"][0]
+
+            if (
+                isinstance(action, exp.ColumnDef)
+                and isinstance(action.this, exp.Dot)
+                and isinstance(action.kind, exp.DataType)
+            ):
+                record = action.this.this.sql()
+                field = action.this.expression.sql()
```
izeigerman, line 299: Why are we not passing a dialect here?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -261,6 +276,63 @@ def fetchall(
...
+                data_type = action.kind.sql(dialect="bigquery")
+                fields_to_add.setdefault(record, []).append((data_type, field))
```
izeigerman, line 301:
```
fields_to_add: t.Dict[str, list[t.Tuple[exp.DataType, str]]] = defaultdict(list)
field_to_add[record].append((data_type, field))
```

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -261,6 +276,63 @@ def fetchall(
         )
         return list(self._query_data)
 
+    def __add_nested_columns(
+        self,
+        alter_expressions: t.List[exp.Alter],
+    ) -> t.List[exp.Alter]:
+        """
+        Add nested columns for the given alter expressions using the BigQuery API directly.
```
izeigerman, line 284: Took me a minute to figure out what does the return value suppose to represent. Perhaps worth adding in the docstring. 

Alternatively, it could be a good idea to split nested and non-nested expressions upstream.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -261,6 +276,63 @@ def fetchall(
         )
         return list(self._query_data)
 
+    def __add_nested_columns(
```
izeigerman, line 279: Why double underscore?

sqlmesh/core/schema_diff.py
```diff
@@ -523,25 +526,33 @@ def _alter_operation(
...
         if self.support_nested_operations:
             if new_type.this == current_type.this == exp.DataType.Type.STRUCT:
-                return self._get_operations(
+                operations = self._get_operations(
                     columns,
                     current_type,
                     new_type,
                     root_struct,
                 )
+                if not operations[0].op == TableAlterOperationType.DROP or self.support_nested_drop:
```
izeigerman, line 535: I'm a little concerned that we rely on a specific order in which operations are added to the list in `_get_operations`. Should we check the entire list instead?

sqlmesh/core/engine_adapter/bigquery.py
Themiscodes: Right my omission changing this

sqlmesh/core/engine_adapter/bigquery.py
Themiscodes: My bad meant to have a single underscore

sqlmesh/core/engine_adapter/bigquery.py
Themiscodes: Right, factored it out to a function that splits the expressions beforehand

sqlmesh/core/schema_diff.py
Themiscodes: Yes, added a function to perform this check and moved it before `_get_operations` since I noticed that it mutates the current_type

Themiscodes:
Thanks for the review @izeigerman refactored and addressed comments

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -261,6 +281,66 @@ def fetchall(
...
+                record = action.this.this.sql(dialect="bigquery")
+                field = action.this.expression.sql(dialect="bigquery")
+                data_type = action.kind.sql(dialect="bigquery")
+                fields_to_add[record].append((data_type, field))
+            else:
+                non_nested_expressions.append(alter_expression)
+
+        return fields_to_add, non_nested_expressions
+
+    def _update_table_schema(
```
izeigerman, line 310: [Nit] Should we clarify that this method is only for nested fields?

izeigerman (approved):
One cosmetic comment, LGTM otherwise. Thanks for addressing comments!

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -224,6 +225,25 @@ def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
...
+
+        nested_fields, non_nested_expressions = self._split_alter_expressions(alter_expressions)
+
+        if nested_fields:
+            self._update_table_schema(nested_fields, alter_expressions[0].this)
+
+        if non_nested_expressions:
+            with self.transaction():
+                for alter_expression in non_nested_expressions:
+                    self.execute(alter_expression)
```
georgesittas, line 245: [Nit]: should we fallback to the base class `alter_table` for these?

plaflamme:
This is fantastic, thanks!

Scanned through quickly, couldn't say for certain: will this succeed at issuing schema change **only if** the column is appended or does it also "re-order" things so it can make the change?

Themiscodes:
Thanks @plaflamme, currently it is designed to succeed both when fields are appended at the end and when they're not, in which case the columns in the struct are added to the end to accommodate the change as you say. But a 're-order' will occur only if a nested field is removed or replaced with another, as this will trigger a `Drop` operation and a construction of the struct from scratch, as it did till now.

georgesittas (approved):

PR #3087, opened 2024-09-04
Fix python sql explicit column declaration

ravenac95:
We encountered this while making a python model that had `is_sql=True`. We send some generative things to our clickhouse instance and because we use the clickhouse dialect this seems to cause some issues when attempting to transpile to other dialects like `duckdb` due to the `Nullable` type appearing as a derived type for a specific column of the final result. So to solve this we decided to just explicitly set the columns in the python sql model. However we encountered a bug in doing this.


georgesittas (approved):
LGTM - thanks for fixing this!

PR #3098, opened 2024-09-05
Feat: clickhouse dbt adapter

treysp:
SQLMesh's dbt adapter does not implement all functionality supported by dbt-clickhouse, including (but not limited to) materialized views and Clickhouse dictionaries.

### DIFF
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -167,7 +167,7 @@ class DbtLoader(Loader):
 
                 self._projects.append(project)
 
-                if project.context.target.database != (self._context.default_catalog or ""):
+                if project.context.target.database != self._context.default_catalog:
                     raise ConfigError(
                         "Project default catalog does not match context default catalog"
                     )
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -167,7 +167,7 @@ class DbtLoader(Loader):
 
                 self._projects.append(project)
 
-                if project.context.target.database != self._context.default_catalog:
+                if project.context.target.database != (self._context.default_catalog or ""):
                     raise ConfigError(
                         "Project default catalog does not match context default catalog"
                     )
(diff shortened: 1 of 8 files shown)


erindru (approved):

sqlmesh/dbt/loader.py
```diff
@@ -167,7 +167,7 @@ def _load_projects(self) -> t.List[Project]:
 
                 self._projects.append(project)
 
-                if project.context.target.database != self._context.default_catalog:
+                if project.context.target.database != (self._context.default_catalog or ""):
```
crericha, line 170: What does an empty catalog imply?

treysp: Clickhouse doesn't have catalogs (just 2-level object naming), so the catalog should always be empty

sqlmesh/dbt/target.py
```diff
@@ -844,7 +847,100 @@ def to_sqlmesh(self, **kwargs: t.Any) -> ConnectionConfig:
...
+    retries: int = 1
+    database_engine: t.Optional[str] = None
+    cluster_mode: bool = False
+    sync_request_timeout: int = 5
+    compress_block_size: int = 1048576
+    check_exchange: bool = True
+    use_lw_deletes: bool = False
+    allow_automatic_deduplication: bool = False
+    tcp_keepalive: t.Union[bool, t.Tuple[int, int, int], t.List[int]] = False
+    database: str = ""
```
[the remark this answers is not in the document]
treysp: In Clickhouse's 2-level naming, the second level is called "database". So in this context `database` is just an alias for `schema`. 

dbt-clickhouse errors if you specify database and its value doesn't match `schema` https://github.com/ClickHouse/dbt-clickhouse/blob/44d26308ea6a3c8ead25c280164aa88191f05f47/dbt/adapters/clickhouse/credentials.py#L49

crericha (approved):

Issue #2496, opened 2024-04-23
Don't create temp tables at plan time unless they're actually needed


georgesittas:
We currently create temporary tables at plan time preemptively, even though they may not necessarily be needed. This behavior was introduced before we started to create tables on demand (i.e., if they're missing), but we can move away from it now.

For instance, the initial plan application for the example project produces the following tables in duckdb:

```
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__full_model__3031530868(item_id INTEGER, num_orders BIGINT);
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__full_model__3031530868__temp(item_id INTEGER, num_orders BIGINT);
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__incremental_model__141311533(id INTEGER, item_id INTEGER, event_date DATE);
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__incremental_model__141311533__temp(id INTEGER, item_id INTEGER, event_date DATE);
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__seed_model__1248844913(id INTEGER, item_id INTEGER, event_date DATE);
CREATE TABLE sqlmesh__sqlmesh_example.sqlmesh_example__seed_model__1248844913__temp(id INTEGER, item_id INTEGER, event_date DATE);
```

erindru:
This would also help with having less tables for the janitor to clean up.

Recent changes have made the [actual snapshot deployability](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/evaluator.py#L685)  available in the strategy `create()` method (previously, only a doctored DeployabilityIndex was available that always returned "not deployable").

So we could use this to skip creating the dev preview ("__temp") tables if the snapshot being created is deployable

PR #3099, opened 2024-09-05
Fix!: Dont clone managed tables to create dev previews

erindru:
Prior to this change, making a forward-only change to a managed model only worked if you did it via a dev environment first.

Applying it directly to prod resulted in the following error on Snowflake: `SQL compilation error: Invalid operation Cannot clone from an dynamic table.` because in order to create the dev preview table (which [always gets created](https://github.com/TobikoData/sqlmesh/issues/2496) regardless of the environment), SQLMesh triggers a clone of the current prod table.

This PR does two things:
 - Allows the evaluation strategy to decide if `clone()` is suitable for creating dev preview tables
 - Implements `migrate()` for the `EngineManagedStrategy` now that I better understand the conditions that can trigger `migrate()` in the first place.

The usual function of `migrate()` is to mutate the schema of an existing table to match a new definition without re-creating that table (which can be expensive). However:

- Managed models are more like views in that you cant just mutate the schema, you have to replace the entire query backing the model
- So, if the new model definition results in schema changes, we throw an error because it has to be rebuilt which defeats the purpose of foward-only if we were to treat it as forward-only
- If there are no schema changes (eg an unrelated macro that doesn't affect the model schema got updated), `migrate()` is a no-op

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -1794,7 +1794,7 @@ class EngineManagedStrategy(MaterializableStrategy):
         if len(potential_alter_expressions) > 0:
             # this can happen if a user changes a managed model and deliberately overrides a plan to be forward only, eg `sqlmesh plan --forward-only`
             raise SQLMeshError(
-                f"Managed table '{target_table_name}' cannot be treated as forward only because its schema cannot be changed without a full rebuild"
+                f"The schema of the managed model '{target_table_name}' cannot be updated in a forward-only fashion."
             )
 
     def delete(self, name: str, **kwargs: t.Any) -> None:
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -667,6 +667,7 @@ class SnapshotEvaluator:
                 and snapshot.is_materialized
                 and snapshot.previous_versions
                 and self.adapter.SUPPORTS_CLONING
+                and evaluation_strategy.allow_clone_for_dev_preview(snapshot)
             ):
                 target_table_name = snapshot.table_name(is_deployable=False)
                 tmp_table_name = f"{target_table_name}__schema_migration_source"
@@ -1058,6 +1059,14 @@ class EvaluationStrategy(abc.ABC):
             table_description=model.description,
             column_descriptions=model.column_descriptions,
         )
+
+    def allow_clone_for_dev_preview(self, snapshot: Snapshot) -> bool:
+        """
+        When creating tables to preview forward-only changes in dev environments,
+        should clones of existing tables be used (if the adapter supports it),
+        or should new tables be created
+        """
+        return True
 
 
 class SymbolicStrategy(EvaluationStrategy):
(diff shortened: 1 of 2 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1059,6 +1060,14 @@ def _replace_query_for_model(self, model: Model, name: str, query_or_df: QueryOr
             column_descriptions=model.column_descriptions,
         )
 
+    def allow_clone_for_dev_preview(self, snapshot: Snapshot) -> bool:
```
erindru, line 1063: note: the `snapshot` parameter is currently unused but I left it there incase we wanted evaluation strategies to be able to make more fine-grained decisions in future

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1786,9 +1795,30 @@ def migrate(
...
+            # in the normal case, the current prod snapshot is mutated via ALTER statements to match the schema of the dev preview table
+            # However, since we cant change the schema of a managed model, we have to create a new one instead
+            logger.info(
+                "Creating new managed table: %s to migrate %s to the schema of %s",
+                target_table_name,
+                snapshot.model.name,
+                source_table_name,
+            )
+
+            self.create(
```
izeigerman, line 1812: Can't dynamic tables be altered? I'm concerned that with this approach all prior data will be wiped out and the table will have to start from scratch which defeats the purpose of a forward-only change.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -667,6 +667,7 @@ def _create_snapshot(
                 and snapshot.is_materialized
                 and snapshot.previous_versions
                 and self.adapter.SUPPORTS_CLONING
+                and evaluation_strategy.allow_clone_for_dev_preview(snapshot)
```
izeigerman, line 670: I'm a little concerned that we're adding this abstraction too soon. For example, do we definitely know that dynamic / managed tables can be cloned universally across all engines? If some engines do support this, then this should be a property of the adapter and not the strategy.

Should we not generalize prematurely and just check the model type here?

sqlmesh/core/snapshot/evaluator.py
erindru: Yes, that's true. A dynamic table is essentially a materialized view where the database keeps it up to date internally, so you cant `ALTER TABLE ADD COLUMN` to it, to change the schema you have to replace the query backing the view which recreates it.

Would it be better to keep the original behavior of just throwing an error (although only do it if there are schema changes to be made) and declaring that forward-only changes aren't supported on managed models?

sqlmesh/core/snapshot/evaluator.py
erindru: The attempted clone here is just for creating the "dev preview" table which gets thrown away on deployment to prod. We deliberately use normal tables instead of dynamic tables for dev preview tables, so I couldnt see a situation where we would ever need to clone the production dynamic table (even if the underlying engine supports it).

The reason why I added a method on the evaluation strategy  instead of checking the model type directly within the snapshot evaluator was because I thought the whole point of evaluation strategies was to contain these checks so that the model types didn't have to leak into the snapshot evaluator.

Would it be better to adjust the check to:
```
and snapshot.model.kind != ManagedKind
```
?

sqlmesh/core/snapshot/evaluator.py
erindru: by the way - a side-effect of recreating managed models declared as forward-only is that this provides a pattern for users to update managed models without having a bunch of old snapshot versions lying around (that can incur compute cost).

I believe Nick was [asking about this recently](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1721419318871229) and resorted to writing his own cleanup logic

sqlmesh/core/snapshot/evaluator.py
izeigerman: well the fact that we have `snapshot.is_materialized` in there means that this leakage already occurred :) So if anything adding `snapshot.is_managed` there will make it consistent.

sqlmesh/core/snapshot/evaluator.py
izeigerman: > by the way - a side-effect of recreating managed models declared as forward-only is that this provides a pattern for users to update 

I can see that. However, this goes against the very idea of forward-only which suppose to prevent the **rebuild**. That's why I like the 

> Would it be better to keep the original behavior of just throwing an error (although only do it if there are schema changes to be made) and declaring that forward-only changes aren't supported on managed models?

suggestion better.

erindru: Ok, if there are no schema changes then `migrate()` is a no-op which should resolve the issue of an unrelated macro change tagging the model as modified and hitting `migrate()`  codepath.

If there are schema changes, an error is thrown advising the user that managed models cannot be forward only

sqlmesh/core/snapshot/evaluator.py
erindru: Ok, i've updated the check to use the `is_managed` property (which I forgot existed) and added a comment explaining why

izeigerman (approved):
One small comment, LGTM otherwise. Thanks for addressing comments!

PR #3103, opened 2024-09-06
fix: Allow macros in set operations

### LINKED ISSUE #3102: Unable to parse macro call in UNION ALL
[Public slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1725557364459159)

Repro:

```SQL
@macro()
def union_macro(evaluator):
 return "SELECT 2 as col"

MODEL (...)

SELECT 1 as col UNION ALL @union_macro()
```

In `_parse_set_operations()` we'll attempt to `_parse_select()` for the RHS which doesn't have a path for parsing placeholders yet; This means that we'll return `None` instead of parsing the macro, causing the error in the discussion.

VaggelisD:
Fixes #3102

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -314,6 +314,31 @@ def _parse_join(
     return macro
 
 
+def _parse_select(
+    self: Parser,
+    nested: bool = False,
+    table: bool = False,
+    parse_subquery_alias: bool = True,
+    parse_set_operation: bool = True,
+) -> t.Optional[exp.Expression]:
+    select = self.__parse_select(  # type: ignore
+        nested=nested,
+        table=table,
+        parse_subquery_alias=parse_subquery_alias,
+        parse_set_operation=parse_set_operation,
+    )
+
+    if (
+        not select
+        and not parse_set_operation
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -314,6 +314,31 @@ def _parse_join(
...
+
+    if (
+        not select
+        and not parse_set_operation
+        and self._match_pair(TokenType.PARAMETER, TokenType.VAR, advance=False)
+    ):
+        self._advance()
+        return _parse_macro(self)
+
+    return select
```
georgesittas, line 339: Let's try to simplify this, maybe this works too:

```python
    return select or self._parse_placeholder()
```

VaggelisD: This was my original implementation but it broke several tests, e.g:

```
E   SELECT *
E   FROM @this_model
E   WHERE @AND(@column NOT IN @is_in, @condition)
```

So I tried to shield it off such that we attempt to parse macros only during set operations.

georgesittas: I see, that's ok then. Perhaps being explicit by checking the `PARAMETER` and `VAR` pair is better to avoid parsing edge cases. Please disregard my suggestion 👍

georgesittas (approved):

PR #3106, opened 2024-09-06
Fix: allow overriding blocking setting of builtin audits

georgesittas:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1725579711884559

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -559,16 +559,28 @@ class _Model(ModelMeta, frozen=True):
 
         referenced_audits = []
 
-        for audit_name, _ in self.audits + default_audits:
+        for audit_name, audit_args in self.audits + default_audits:
             if audit_name in self.inline_audits:
                 referenced_audits.append(self.inline_audits[audit_name])
             elif audit_name in audits:
                 referenced_audits.append(audits[audit_name])
-            elif audit_name not in BUILT_IN_AUDITS:
-                raise_config_error(
-                    f"Unknown audit '{audit_name}' referenced in model '{self.name}'",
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -559,16 +559,28 @@ def referenced_audits(
...
+                        self._path,
+                    )
+
+                # Builtin audits are generally not included in order to reduce fingerprint size,
+                # but those that override `blocking` need to be included because otherwise doing
+                # `audit.blocking` will always return the builtin audit's default value
+                blocking = audit_args.get("blocking")
+                if blocking:
+                    referenced_audits.append(
+                        audit.copy(update={"blocking": blocking == exp.true()})  # type: ignore
```
izeigerman, line 581: Is `blocking` rendered at this point? What if the value was a macro / var call?

tests/core/test_context.py
```diff
@@ -292,6 +292,37 @@ def test_plan_execution_time():
...
+    context = Context(config=Config())
+    context.upsert_model(
+        load_sql_based_model(
+            parse(
+                """
+                MODEL(
+                    name db.x,
+                    kind FULL,
+                    audits (
+                        not_null(columns := [c], blocking := false),
```
izeigerman, line 305: Can we have a test with a macro call too for blocking

izeigerman (approved):

sqlmesh/core/model/definition.py
georgesittas: I think it should be, it's rendered in `load_sql_based_model` using the `meta_renderer`.

PR #3107, opened 2024-09-06
Fix: Make sure signals are loaded when running models

### LINKED ISSUE #3104: Signals for the built-in scheduler are not being loaded on `sqlmesh run`s
When files are present under `signals/`, they should be loaded to initialize the `signal_factory`. Currently, `sqlmesh run` avoids loading the local project altogether leading to the factory not being registered, leading to signals not being evaluated during `sqlmesh run`.

Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1725634479113459

izeigerman:
Normally, we don't load the project on `sqlmesh run` command. However, we still need to load user-defined signals in order for the run to work correctly.

This update ensures that SQLMesh loads just signals when running models.

Closes #3104

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -327,6 +327,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         self._metrics: UniqueKeyDict[str, Metric] = UniqueKeyDict("metrics")
         self._jinja_macros = JinjaMacroRegistry()
         self._default_catalog: t.Optional[str] = None
+        self._loaded: bool = False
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/context.py
```diff
@@ -586,6 +588,11 @@ def run(
             state_sync_type=self.state_sync.state_type(),
         )
 
+        if not self._loaded:
+            # Signals should be loade to run correctly.
```
georgesittas, line 592:
```suggestion
            # Signals should be loaded to run correctly.
```

georgesittas (approved):

tobymao (approved):

PR #3108, opened 2024-09-09
Feat!: Add physical_schema_mapping property and deprecate physical_schema_override

erindru:
This PR essentially upgrades `physical_schema_override` config property to allow the use of regex's, which brings it to feature parity with the existing `environment_catalog_mapping` property.

However, `physical_schema_override` couldn't have its semantics changed without breaking backwards compatibility, so it's been deprecated in favour of a new property `physical_schema_mapping` which has the regex sematics.

Note that this change only affects the project config / loader. The models themselves retain their existing `physical_schema_override` property which is a string containing the actual schema override relevant for that model and not the mapping from the config.


sqlmesh/core/config/root.py
```diff
@@ -171,16 +176,32 @@ def _normalize_and_validate_fields(cls, values: t.Dict[str, t.Any]) -> t.Dict[st
                     f"The `{plan_deprecated}` config is deprecated. Please use the `plan.{plan_deprecated}` config instead."
                 )
 
+        if values.get("physical_schema_override") and values.get("physical_schema_mapping"):
```
izeigerman, line 179: Should we check for key existence instead? Eg. this check will still pass if one of the dicts is empty.

sqlmesh/core/config/root.py
```diff
@@ -171,16 +176,32 @@ def _normalize_and_validate_fields(cls, values: t.Dict[str, t.Any]) -> t.Dict[st
...
-        }
+
+        def _normalize_identifiers(key: str) -> None:
+            values[key] = {
+                k: normalize_identifiers(v, dialect=dialect).name
+                for k, v in values.get(key, {}).items()
+            }
+
+        _normalize_identifiers("environment_catalog_mapping")
+        _normalize_identifiers("physical_schema_mapping")
```
izeigerman, line 198: Was it a bug that we didn't o this previously?

sqlmesh/core/config/root.py
izeigerman: Unless keys are always present in this validator, of course

sqlmesh/core/config/root.py
```diff
@@ -171,16 +176,32 @@ def _normalize_and_validate_fields(cls, values: t.Dict[str, t.Any]) -> t.Dict[st
...
+
+        if values.get("physical_schema_override"):
```
izeigerman, line 200: Should we keep this kind of validation in one place and move the check into the `before` validator? Right after we check for both keys.

sqlmesh/core/config/root.py
```diff
@@ -100,6 +104,7 @@ class Config(BaseConfig):
     env_vars: t.Dict[str, str] = {}
     username: str = ""
     physical_schema_override: t.Dict[str, str] = {}
+    physical_schema_mapping: t.Dict[re.Pattern, str] = {}
```
izeigerman, line 107: Instead of having both properties can we just automatically convert `physical_schema_override` into `physical_schema_mapping` by wrapping every key into `^` / `$`? This way we can just always pass the mapping and avoid extra complexity with resolved.

izeigerman: I guess resolver is kind of helpful, but we can still get rid of an unnecessary field.

erindru: Yes, that's a much better idea. I've implemented it

sqlmesh/core/config/root.py
erindru: I think so. I can see that the `physical_schema_override` property was quite old and perhaps pre-dated the discovery of all the normalization issues

sqlmesh/core/model/definition.py
```diff
@@ -2034,14 +2035,22 @@ def _create_model(
...
-    physical_schema_override = physical_schema_override or {}
+    physical_schema_mapping = physical_schema_mapping or {}
 
     raw_kind = kwargs.pop("kind", None)
     if raw_kind:
         kwargs["kind"] = create_model_kind(raw_kind, dialect, defaults or {})
 
     defaults = {k: v for k, v in (defaults or {}).items() if k in klass.all_fields()}
 
+    def _resolve_physical_schema_override(schema_name: str) -> t.Optional[str]:
```
erindru, line 2046: I just moved the resolving logic to here now that it only depends on a single property

sqlmesh/core/config/root.py
erindru: Yes, good point. I poked around in the debugger and the only keys that are present are the ones that get passed in when the `Config` object is created, so checking for key existence makes sense

sqlmesh/core/config/root.py
erindru: Yes, makes sense. I've moved it

sqlmesh/core/model/definition.py
erindru: And then I changed my mind and removed the function altogether to just do this in line

izeigerman (approved):
Let's remove the obsolete line from the docstring and this is good to go

Issue #3109, opened 2024-09-09
Sqlmesh UI on constant refresh


RN-nguyen:
Loaded up wursthall and sushi_dbt in sqlmesh ui and getting a refresh no matter what page I am on. 

Here is a video showing what's happening on my end. I am not doing anything when the page is refreshing here:

https://github.com/user-attachments/assets/4dae62d3-9b49-41fe-b00c-1fedcd1ee642

tobymao:
@mykalmax can you take a look?

RN-nguyen:
Verified branch addresses refreshing issue related to file watcher.
Also able to create, update, delete new files and folders.

PR #3113, opened 2024-09-09
fix: switch from pooling to default method to detect changes

mykalmax:
fixing https://github.com/TobikoData/sqlmesh/issues/3109

so when using `pooling` (and pooling is not default option for wathfiles). It seems to treat some files (like symlinks for example) as deleted and just keeps emitting deletion change. And that results in UI keep triggering updates and API calls (an infinite loop)

### DIFF
--- a/web/client/src/library/components/fileExplorer/context.tsx
+++ b/web/client/src/library/components/fileExplorer/context.tsx
@@ -158,7 +158,7 @@ export default function FileExplorerProvider({
   }
 
   function renameArtifact(artifact: ModelArtifact, newName?: string): void {
-    newName = newName?.trim()
+    newName = newName?.replaceAll('/', '').trim()
 
     const parentArtifacts = artifact.parent?.artifacts ?? []
     const isDuplicate = parentArtifacts.some(a => a.name === newName)
@@ -300,7 +300,11 @@ export default function FileExplorerProvider({
         artifact.rename(artifact.copyName())
       }
 
-      const new_path = [target.path, artifact.name].join('/')
+      let new_path = [target.path, artifact.name].join('/')
+
+      while (new_path.startsWith('/')) {
+        new_path = new_path.slice(1).trim()
+      }
 
       if (artifact instanceof ModelDirectory) {
         moveArtifactCallbacks.push(() => {
--- a/web/client/src/library/pages/root/Navigation.tsx
+++ b/web/client/src/library/pages/root/Navigation.tsx
@@ -14,7 +14,7 @@ export default function PageNavigation(): JSX.Element {
   return (
     <div
       className={clsx(
-        'relative min-w-[10rem] px-2 min-h-8 max-h-8 w-full flex items-center',
+        'relative min-w-[10rem] px-2 min-h-[2rem] max-h-[2rem] w-full flex items-center',
         modules.showHistoryNavigation ? 'justify-between' : 'justify-end',
         isFetchingModels && 'overflow-hidden',
       )}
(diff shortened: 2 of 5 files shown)


web/server/watcher.py
```diff
@@ -28,33 +28,41 @@ async def watch_project() -> None:
         (settings.project_path / c.METRICS).resolve(),
         (settings.project_path / c.SEEDS).resolve(),
     ]
-    ignore_entity_patterns = context.config.ignore_patterns if context else c.IGNORE_PATTERNS
-    ignore_entity_patterns.append("^\\.DS_Store$")
-    ignore_entity_patterns.append("^.*\\.db(\\.wal)?$")
-    ignore_paths = [str((settings.project_path / c.CACHE).resolve())]
-    watch_filter = DefaultFilter(
-        ignore_paths=ignore_paths, ignore_entity_patterns=ignore_entity_patterns
+    watch_filter = DefaultFilter()
```
mykalmax, line 31: here i wanted to extend defaults

web/client/src/library/components/fileExplorer/context.tsx
```diff
@@ -158,7 +158,7 @@ export default function FileExplorerProvider({
   }
```
mykalmax, line 159: path should not contain leading  `/` and name should not contain `/` at all

web/client/src/library/pages/root/Navigation.tsx
```diff
@@ -14,7 +14,7 @@ export default function PageNavigation(): JSX.Element {
   return (
     <div
       className={clsx(
-        'relative min-w-[10rem] px-2 min-h-8 max-h-8 w-full flex items-center',
+        'relative min-w-[10rem] px-2 min-h-[2rem] max-h-[2rem] w-full flex items-center',
```
mykalmax, line 17: fixing small layout shift when showing `loading` message

web/server/watcher.py
vchan: I personally would have done it the following way:
```python
DefaultFilter(
    ignore_paths=DefaultFilter.ignore_paths + ignore_paths,
    ignore_entity_patterns=DefaultFilter.ignore_entity_patterns + ignore_entity_patterns,
)
```

vchan (approved):

web/server/watcher.py
```diff
@@ -28,33 +28,41 @@ async def watch_project() -> None:
...
-                elif change == Change.deleted or not path.exists():
+                if change == Change.deleted or not path.exists():
                     changes.append(
                         models.ArtifactChange(
                             change=Change.deleted,
                             path=str(relative_path),
                         )
                     )
-                elif change == Change.modified and path.is_file():
+                if change == Change.added:
```
vchan, line 59: Not `elif`?

web/server/watcher.py
```diff
@@ -28,33 +28,40 @@ async def watch_project() -> None:
         (settings.project_path / c.METRICS).resolve(),
         (settings.project_path / c.SEEDS).resolve(),
     ]
+    ignore_dirs = ensure_list(".env")
```
vchan, line 31: This doesn't seem necessary when you can explicitly make this a list, i.e. `ignore_dirs = [".env"]`.

web/server/utils.py
```diff
@@ -104,3 +104,22 @@ def is_relative_to(path: PurePath, other: PurePath | str) -> bool:
         return True
     except ValueError:
         return False
+
+
+def ensure_list(value: t.Optional[t.Any] = None) -> t.List:
```
vchan, line 109: Can we use ensure_list from from sqlglot.helper? I see it used in various places in our codebase.

vchan (approved):

PR #3114, opened 2024-09-09
Chore: Use factory method to create builtin scheduler

crericha:
Introduce a common factory method to create the builtin scheduler instead of directly constructing the `Scheduler` class.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -479,6 +479,17 @@ class GenericContext(BaseContext, t.Generic[C]):
         if not snapshots:
             raise ConfigError("No models were found")
 
+        return self._create_scheduler(snapshots)
+
+    def _create_scheduler(self, snapshots: t.Iterable[Snapshot]) -> Scheduler:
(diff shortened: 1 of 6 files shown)


sqlmesh/core/context.py
```diff
@@ -479,6 +479,17 @@ def scheduler(self, environment: t.Optional[str] = None) -> Scheduler:
         if not snapshots:
             raise ConfigError("No models were found")
 
+        return self._create_scheduler(snapshots)
+
+    def _create_scheduler(self, snapshots: t.Iterable[Snapshot]) -> Scheduler:
```
izeigerman, line 484: Should we just make it public (no `_`)?

izeigerman (approved):
Naming comment, LGTM otherwise.

sqlmesh/core/context.py
crericha: Yeah, we can. My only concern was having it and `def scheduler` method both be public, it could be confusing. I don't have a strong opinion though. Do you?

PR #3117, opened 2024-09-10
feat: materialized view support clustered by and partitioned by

eakmanrq:
Currently if a user tries to use partitioned_by or clustered_by with materialized views then we will throw an error despite it being supported on some engines (like BQ). 

This PR allows setting these fields if you have a view set to be materialized. 

Note that during model validation we optimistically assume if a user is using partitioned by/clustered by on materialized model that the underlying engine will support it. If not then we will either throw an error or possibly SQLGlot will remove on transpilation but this is untested.


sqlmesh/core/engine_adapter/base.py
```diff
@@ -955,6 +960,23 @@ def create_view(
         if not self.SUPPORTS_VIEW_SCHEMA and isinstance(schema, exp.Schema):
             schema = schema.this
 
+        if materialized_properties:
+            partitioned_by = materialized_properties.pop("partitioned_by", None)
```
izeigerman, line 964: Why do we not do this just for BQ? Do other engines support this in materialized views?  I bet postgres doesn't. So far our strategy has been ignoring these for engines that don't support them.

Eg. see the `_build_table_properties_exp` method. Btw, any chance we can reuse it to populate the properties?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2015,6 +2037,24 @@ def _table_or_view_properties_to_expressions(
...
+    def _build_partitioned_by_exp(
+        self,
+        partitioned_by: t.List[exp.Expression],
+        *,
+        partition_interval_unit: t.Optional[IntervalUnit] = None,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        catalog_name: t.Optional[str] = None,
+        **kwargs: t.Any,
+    ) -> t.Union[exp.PartitionedByProperty, exp.Property]:
+        raise NotImplementedError("This engine does not support partitioning.")
```
izeigerman, line 2049: This changes semantics from ignoring unsupported properties to failing. I don't think this aligns well with ability to run the same project across different engines.

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -116,6 +116,37 @@ class HiveMetastoreTablePropertiesMixin(EngineAdapter):
...
+            # In addition, partition column transform expressions like `day(col)` or `bucket(col, 5)` are allowed
+            # Also, column names and transforms need to be strings and supplied as an ARRAY[varchar]
+            # ref: https://trino.io/docs/current/connector/iceberg.html#table-properties
+            return exp.Property(
+                this=exp.var("PARTITIONING"),
+                value=exp.array(
+                    *(exp.Literal.string(e.sql(dialect=self.dialect)) for e in partitioned_by)
+                ),
+            )
+        else:
```
georgesittas, line 140: [Nit] we can remove this `else` and dedent the following block.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -955,6 +960,23 @@ def create_view(
...
+        if materialized_properties:
+            partitioned_by = materialized_properties.pop("partitioned_by", None)
+            clustered_by = materialized_properties.pop("clustered_by", None)
+            if partitioned_by:
+                materialized_properties["catalog_name"] = exp.to_table(view_name).catalog
+                properties.append(
+                    "expressions",
+                    self._build_partitioned_by_exp(
+                        partitioned_by, **(materialized_properties or {})
+                    ),
```
georgesittas, line 972:
```suggestion
                    self._build_partitioned_by_exp(partitioned_by, **materialized_properties),
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -955,6 +960,23 @@ def create_view(
...
+                )
+            if clustered_by:
+                properties.append(
+                    "expressions",
+                    self._build_clustered_by_exp(clustered_by, **(materialized_properties or {})),
```
georgesittas, line 977:
```suggestion
                    self._build_clustered_by_exp(clustered_by, **materialized_properties),
```

georgesittas (commented):
It will be nice if we can simplify this using Iaroslav's [suggestion](https://github.com/TobikoData/sqlmesh/pull/3117/files#r1752946394). LGTM otherwise, just a few minor comments.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Snowflake supports cluster by: https://docs.snowflake.com/en/sql-reference/sql/create-materialized-view

Databricks supports partition by: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-materialized-view.html

So I don't think it is correct to say that this is a BQ only feature.

eakmanrq: For the second part about `_build_table_properties_exp`, I actually initial did that: https://github.com/TobikoData/sqlmesh/pull/3117/commits/daf02da3fecc87b1337424a114ca2914028f8fc6

I then realized that the contract with that method is that it build table properties for a given engine therefore someone could include default table properties that wouldn't apply to a view. Clickhouse had some errors that made me realize that this wasn't the right pattern. Therefore I split specifically building cluster by/partition by since they were shared across tables and views.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Ok will update to return None instead and handle None

sqlmesh/core/engine_adapter/base.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/3117/commits/37ab52c49a691e28a287f86815b1a04275851f21

sqlmesh/core/engine_adapter/base.py
eakmanrq: Returns None now: https://github.com/TobikoData/sqlmesh/pull/3117/commits/37ab52c49a691e28a287f86815b1a04275851f21

izeigerman (approved):

PR #3122, opened 2024-09-12
Feat: Support mixed multi-repo SQLMesh and dbt projects

Themiscodes:
This update adds support for multiple repos of mixed **SQLMesh** and **dbt** projects. This allows managing and sourcing tables from either repo within the same project and facilitates a gradual migration from **dbt** to **SQLMesh**.

Bootstrapping the project to integrate projects of both types can be done similar to the native [multi-repo setup](https://sqlmesh.readthedocs.io/en/stable/guides/multi_repo/):

```
sqlmesh -p examples/multi_hybrid/dbt_repo -p examples/multi_hybrid/sqlmesh_repo plan
```


sqlmesh/core/context.py
```diff
@@ -331,6 +331,24 @@ def __init__(
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
 
+        # Separate loader and configurations for different project types (dbt, native, hybrid)
+        self._dbt_loader: t.Optional[Loader] = None
+        self._sqlmesh_loader: t.Optional[Loader] = None
+        self.dbt_configs: t.Dict[Path, C] = {}
+        self.sqlmesh_configs: t.Dict[Path, C] = {}
+        for path, config in self.configs.items():
+            project_type = c.DBT if config.loader.__name__.lower().startswith(c.DBT) else c.NATIVE
```
erindru, line 340: this seems fairly convoluted, was there a reason you didn't do something like `isinstance(config.loader, DbtLoader)` ? Or make `project_type` a property of the loader?

sqlmesh/core/context.py
```diff
@@ -601,8 +641,13 @@ def run(
 
         if not self._loaded:
             # Signals should be loaded to run correctly.
-            with sys_path(*self.configs):
-                self._loader.load_signals(self)
+            if self._sqlmesh_loader:
+                with sys_path(*self.sqlmesh_configs):
```
erindru, line 645: There's a bunch of these `with sys_path...` conditions that appear to do the same thing (call a known method on the loader that is the same between loaders).

I wonder if there's a way to make this more concise to reduce the duplication

tests/core/analytics/test_collector.py
```diff
@@ -46,6 +42,7 @@ def test_on_project_loaded(collector: AnalyticsCollector, mocker: MockerFixture)
 
     from dbt.version import __version__ as dbt_version
 
+    version = ', "dbt_version": "' + dbt_version + '"' if project_type != "native" else ""
```
erindru, line 45: Should this use the project type constant for `"native"` ?

sqlmesh/core/context.py
Themiscodes: `isinstance` wouldn't work because `config.loader` would resolve as type `abc.ABC`. Changed this to be an attribute of the loader for now

sqlmesh/core/context.py
Themiscodes: right i'll try to think of how to refactor these without making it too unclear

sqlmesh/core/context.py
```diff
@@ -517,41 +532,68 @@ def state_reader(self) -> StateReader:
...
+            project
+            for loader, configs in [
+                (self._dbt_loader, self.dbt_configs),
+                (self._sqlmesh_loader, self.sqlmesh_configs),
+            ]
+            for project in [
+                self._load_factory("project", loader, configs, update_schemas=update_schemas)
+            ]
+            if isinstance(project, LoadedProject)
+        ]
```
georgesittas, line 554: Is it correct that this list can contain at most two `LoadedProject` objects? If so, I'd consider inlining their instantiation here because the double loop within the comprehension feels needlessly complicated.

sqlmesh/core/context.py
```diff
@@ -2042,6 +2084,21 @@ def _register_notification_targets(self) -> None:
...
         )
 
+    def _load_factory(
+        self,
+        method: str,
+        loader: t.Optional[Loader] = None,
+        configs: t.Optional[t.Dict[Path, C]] = None,
+        **kwargs: t.Any,
+    ) -> LoadedProject | None:
+        if loader and configs:
```
georgesittas, line 2094: I don't think you can ever have a truthy `loader` and a falsey `configs` here, so the latter may be unnecessary.

Themiscodes: yes that's true this is only there to pass a mypy check

sqlmesh/core/context.py
```diff
@@ -517,41 +532,68 @@ def state_reader(self) -> StateReader:
...
+                self._load_factory("project", loader, configs, update_schemas=update_schemas)
+            ]
+            if isinstance(project, LoadedProject)
+        ]
+
+        self._standalone_audits.clear()
+        self._audits.clear()
+        self._macros.clear()
+        self._models.clear()
+        self._metrics.clear()
```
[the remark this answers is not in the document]
Themiscodes: Right it's because before when we had one project we were assigning to `_models`, `_metrics` etc, so if it already had values they were overwritten. Here since I chose to use update I had to clear the morels, macros and metrics before so that they were empty

sqlmesh/core/context.py
```diff
@@ -517,41 +532,68 @@ def state_reader(self) -> StateReader:
...
+            self._jinja_macros = self._jinja_macros.merge(project.jinja_macros)
+            self._macros.update(
+                {key: value for key, value in project.macros.items() if key not in self._macros}
+            )
+            self._models.update(
+                {key: value for key, value in project.models.items() if key not in self._models}
+            )
+            self._metrics.update(
+                {key: value for key, value in project.metrics.items() if key not in self._metrics}
+            )
```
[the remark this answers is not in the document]
Themiscodes: right my bad these checks are indeed unnecessary removing them

sqlmesh/core/context.py
```diff
@@ -331,6 +331,23 @@ def __init__(
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
 
+        # Separate loader and configurations for different project types (dbt, native, hybrid)
+        self._dbt_loader: t.Optional[Loader] = None
+        self._sqlmesh_loader: t.Optional[Loader] = None
+        self.dbt_configs: t.Dict[Path, C] = {}
+        self.sqlmesh_configs: t.Dict[Path, C] = {}
+        for path, config in self.configs.items():
+            project_type = config.loader.PROJECT_TYPE
```
georgesittas, line 340: Another way to go about this is:

```python
project_type = c.DBT if issubclass(config.loader, DbtLoader) else c.NATIVE
```

So you won't have to define a new constant in the loader. The reason [you can't do](https://github.com/TobikoData/sqlmesh/pull/3122#discussion_r1757654453) `isinstance` is because `config.loader` is a loader _class_, not an object.

Themiscodes: Yes altering to use `issubclass` it really seems like the more elegant solution for this, thanks!

sqlmesh/core/context.py
```diff
@@ -331,6 +332,23 @@ def __init__(
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
 
+        # Separate loader and configurations for different project types (dbt, native, hybrid)
+        self._dbt_loader: t.Optional[Loader] = None
```
tobymao, line 336: do we really need to separate this out? can we just have a generic dict of loaders?

sqlmesh/core/context.py
```diff
@@ -517,41 +533,59 @@ def state_reader(self) -> StateReader:
 
     def refresh(self) -> None:
         """Refresh all models that have been updated."""
-        if self._loader.reload_needed():
+        if (self._dbt_loader and self._dbt_loader.reload_needed()) or (
```
tobymao, line 536: this is a pattern i'd like to avoid, i'd rather iterate through loaders

sqlmesh/core/context.py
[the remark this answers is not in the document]
tobymao: sorry i'm not sure i get it, why do you need to clear it?

sqlmesh/core/context.py
```diff
@@ -601,8 +635,12 @@ def run(
 
         if not self._loaded:
             # Signals should be loaded to run correctly.
-            with sys_path(*self.configs):
-                self._loader.load_signals(self)
+            if self._sqlmesh_loader:
+                with sys_path(*self.sqlmesh_configs):
+                    self._sqlmesh_loader.load_signals(self)
+            if self._dbt_loader:
```
tobymao, line 641: let's refactor all these places where do if sqlmesh_loader, if dbt_loader

sqlmesh/core/context.py
Themiscodes: refactored this to use dict of loaders and all other parts that use it, it indeed made the code and logic clearer

sqlmesh/core/context.py
```diff
@@ -517,41 +526,53 @@ def state_reader(self) -> StateReader:
...
-            duplicates = set(self._models) & set(self._standalone_audits)
-            if duplicates:
-                raise ConfigError(
-                    f"Models and Standalone audits cannot have the same name: {duplicates}"
-                )
+        self.dag = (
+            DAG({**projects[0].dag.graph, **projects[1].dag.graph})
+            if len(projects) == 2
+            else projects[0].dag
+        )
```
georgesittas, line 562: Should we just do a loop over `projects` here and generalize this logic instead of indexing `[0]` and `[1]`? So, `self.dag = <expand all `dag.graph` project attributes>` or something.

sqlmesh/core/loader.py
```diff
@@ -321,7 +321,7 @@ def _load_sql_models(
     ) -> UniqueKeyDict[str, Model]:
         """Loads the sql models into a Dict"""
         models: UniqueKeyDict[str, Model] = UniqueKeyDict("models")
-        for context_path, config in self._context.configs.items():
+        for context_path, config in self._context._loaders[c.NATIVE][1].items():
```
georgesittas, line 324: I'm a little concerned it will be annoying to remember what each item in the `_loaders` pairs is, i.e. that `[1]` here refers to the configs dict. Just something to think about in case there's a cleaner way to implement it.

georgesittas (approved):
LGTM once we reach a conclusion on the dict `clear()`-ing being discussed [here](https://github.com/TobikoData/sqlmesh/pull/3122#discussion_r1759840029).

sqlmesh/core/loader.py
Themiscodes: was wondering the same, do you think a dictionary instead of tuple would be better with loader and configs keys to distinguish it better?
so doing then something like: `self._context._loaders[c.NATIVE].configs`

georgesittas: Yes, this makes sense to me 👍

sqlmesh/core/context.py
Themiscodes: right missed this thanks!

PR #3125, opened 2024-09-12
Feat!: finalize clickhouse engine adapter

treysp:
This PR finalizes the Clickhouse engine adapter:
- Adds SCD model kind support
- Adds ability to pass arbitrary settings to client connection 
- Adds docs

### Context and implementation details

Joins and NULLs
  - Clickhouse defaults to filling empty cells with a datatype-specific default (e.g., 0 for integer columns). 
  - The SCD and table diff queries SQLMesh builds require that we change that behavior to fill with NULLs.
  - We do that by injecting `SETTINGS join_use_nulls = 1` into the query
  - SCD detail: 
    - The original user query is embedded in a CTE. 
    - Query settings are dynamically scoped, so our setting on the outer query will apply to the user query CTE. 
    - If `join_use_nulls` =  0 on the CH server, we inject `join_use_nulls = 0` into the CTE to preserve behavior expected by user.

Connection settings
-  Following dbt-clickhouse (maintained by Clickhouse), we pass these settings to the connection:
   - Always  
     - mutations_sync = "2"
     - insert_distributed_sync" = "1"
   - When running in cluster or cloud modes
     - database_replicated_enforce_synchronous_settings = "1"
     - insert_quorum = "auto"

Storage format
- Each CH table must have a "table engine" (`MergeTree` by default)
- Users specify a table engine in the MODEL DDL `storage_format` key
- Specification may be a function call, so we now generate the value's SQL if the value is an Expression other than Literal or Identifier
  - Generated Literal/Identifier SQL may include quotes, and we want to defer normalization until later

### DIFF
--- a/docs/integrations/engines/clickhouse.md
+++ b/docs/integrations/engines/clickhouse.md
@@ -9,7 +9,7 @@ This page describes SQLMesh support for the Clickhouse engine, including configu
 
 [Clickhouse](https://clickhouse.com/) is a distributed, column-oriented SQL engine designed to rapidly execute analytical workloads.
 
-It provides users find-grained control of its behavior, but that control comes at the cost of complex configuration.
+It provides users fine-grained control of its behavior, but that control comes at the cost of complex configuration.
 
 This section provides background information about Clickhouse, providing context for how to use SQLMesh with the Clickhouse engine.
 
--- a/docs/integrations/engines/clickhouse.md
+++ b/docs/integrations/engines/clickhouse.md
@@ -1,0 +1,295 @@
+# Clickhouse
+
+This page describes SQLMesh support for the Clickhouse engine, including configuration options specific to Clickhouse.
+
+!!! note
+    Clickhouse may not be used for the SQLMesh [state connection](../../reference/configuration.md#connections).
+
+## Background
+
+[Clickhouse](https://clickhouse.com/) is a distributed, column-oriented SQL engine designed to rapidly execute analytical workloads.
+
+It provides users find-grained control of its behavior, but that control comes at the cost of complex configuration.
+
+This section provides background information about Clickhouse, providing context for how to use SQLMesh with the Clickhouse engine.
+
+### Object naming
+
+Most SQL engines use a three-level hierarchical naming scheme: tables/views are nested within _schemas_, and schemas are nested within _catalogs_. For example, the full name of a table might be `my_catalog.my_schema.my_table`.
+
+Clickhouse instead uses a two-level hierarchical naming scheme that has no counterpart to _catalog_. In addition, it calls the second level in the hierarchy "databases." SQLMesh and its documentation refer to this second level as "schemas."
+
+SQLMesh fully supports Clickhouse's two-level naming scheme without user action.
+
+### Table engines
+
+Every Clickhouse table is created with a ["table engine" that controls how the table's data is stored and queried](https://clickhouse.com/docs/en/engines/table-engines). Clickhouse's (and SQLMesh's) default table engine is `MergeTree`.
+
+The [`MergeTree` engine family](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree) requires that every table be created with an `ORDER BY` clause.
+
+SQLMesh will automatically inject an empty `ORDER BY` clause into every `MergeTree` family table's `CREATE` statement, or you can specify the columns/expressions by which the table should be ordered.
+
+### Clickhouse modes of operation
+
+Conceptually, it may be helpful to view Clickhouse as having three modes of operation: single server, cluster, and Clickhouse Cloud. SQLMesh supports all three modes.
+
+#### Single server mode
+
+Single server mode is similar to other SQL engines: aside from choosing each table's engine, you do not need to worry about how computations are executed. You issue standard SQL commands/queries, and Clickhouse executes them.
+
+#### Cluster mode
+
+Cluster mode allows you to scale your Clickhouse engine to any number of networked servers. This enables massive workloads, but requires that you specify how computations are executed by the networked servers.
+
+Clickhouse coordinates the computations on the networked servers with [Clickhouse Keeper](https://clickhouse.com/docs/en/architecture/horizontal-scaling) (it also supports [Apache ZooKeeper](https://zookeeper.apache.org/)).
+
+You specify named virtual clusters of servers in the Keeper configuration, and those clusters provide namespaces for data objects and computations. For example, you might include all networked servers in the cluster you name `MyCluster`.
+
+In general, you must be connected to a Clickhouse server to execute commands. By default, each command you execute runs in single-server mode on the server you are connected to.
+
+To associate an object with a cluster, DDL commands that create or modify it must include the text `ON CLUSTER [your cluster name]`.
+
+If you provide a cluster name in your SQLMesh connection configuration, SQLMesh will automatically inject the `ON CLUSTER` statement into the DDL commands for all objects created while executing the project. We provide more information about clusters in SQLMesh [below](#cluster-specification).
+
+#### Clickhouse Cloud mode
+
+[Clickhouse Cloud](https://clickhouse.com/cloud) is a managed Clickhouse platform. It allows you to scale Clickhouse without administering a cluster yourself or modifying your SQL commands to run on the cluster.
+
+Clickhouse Cloud automates Clickhouse's cluster controls, which sometimes constrains Clickhouse's flexibility or how you execute SQL commands. For example, creating a table with a `SELECT` command must [occur in two steps on Clickhouse Cloud](https://clickhouse.com/docs/en/sql-reference/statements/create/table#from-select-query). SQLMesh handles this limitation for you.
+
+Aside from those constraints, Clickhouse Cloud mode is similar to single server mode - you run standard SQL commands/queries, and Clickhouse Cloud executes them.
+
+## Cluster specification
+
+A Clickhouse cluster allows multiple networked Clickhouse servers to operate on the same data object. Every cluster must be named in the Clickhouse configuration files, and that name is passed to a table's DDL statements in the `ON CLUSTER` clause.
+
+For example, we could create a table `my_schema.my_table` on cluster `TheCluster` like this: `CREATE TABLE my_schema.my_table ON CLUSTER TheCluster (col1 Int8)`.
+
+To create SQLMesh objects on a cluster, provide the cluster name to the `cluster` key in the SQLMesh connection definition (see all connection parameters [below](#localbuilt-in-scheduler)).
+
+SQLMesh will automatically inject the `ON CLUSTER` clause and cluster name you provide into all project DDL statements.
+
+## Model definition
+
+This section describes how you control a table's engine and other Clickhouse-specific functionality in SQLMesh models.
+
+### Table engine
+
+SQLMesh uses the `MergeTree` table engine with an empty `ORDER BY` clause by default.
+
+Specify a different table engine by passing the table engine definition to the model DDL's `storage_format` parameter. For example, you could specify the `Log` table engine like this:
+
+``` sql linenums="1" hl_lines="4"
+MODEL (
+    name my_schema.my_log_table,
+    kind full,
+    storage_format Log,
+);
+
+select
+    *
+from other_schema.other_table;
+```
+
+You may also specify more complex table engine definitions. For example:
+
+``` sql linenums="1" hl_lines="4"
+MODEL (
+    name my_schema.my_rep_table,
+    kind full,
+    storage_format ReplicatedMergeTree('/clickhouse/tables/{shard}/table_name', '{replica}', ver),
+);
+
+select
+    *
+from other_schema.other_table;
+```
+
+#### ORDER BY
+
+`MergeTree` family engines require that a table's `CREATE` statement include the `ORDER BY` clause.
+
+SQLMesh will automatically inject an empty `ORDER BY ()` when creating a table with an engine in the `MergeTree` family. This creates the table without any ordering.
+
+You may specify columns/expressions to `ORDER BY` by passing them to the model `physical_properties` dictionary's `order_by` key.
+
+For example, you could order by columns `col1` and `col2` like this:
+
+``` sql linenums="1" hl_lines="4-6"
+MODEL (
+    name my_schema.my_log_table,
+    kind full,
+    physical_properties (
+        order_by = (col1, col2)
+    )
+);
+
+select
+    *
+from other_schema.other_table;
+```
+
+Note that there is an `=` between the `order_by` key name and value `(col1, col2)`.
+
+Complex `ORDER BY` expressions may need to be passed in single quotes, with interior single quotes escaped by the `\` character.
+
+#### PRIMARY KEY
+
+Table engines may also accept a `PRIMARY KEY` specification. Similar to `ORDER BY`, specify a primary key in the model DDL's `physical_properties` dictionary. For example:
+
+``` sql linenums="1" hl_lines="6"
+MODEL (
+    name my_schema.my_log_table,
+    kind full,
+    physical_properties (
+        order_by = (col1, col2),
+        primary_key = col1
+    )
+);
+
+select
+    *
+from other_schema.other_table;
+```
+
+Note that there is an `=` between the `primary_key` key name and value `col1`.
+
+### Partitioning
+
+Some Clickhouse table engines support partitioning. Specify the partitioning columns/expressions in the model DDL's `partitioned_by` key.
+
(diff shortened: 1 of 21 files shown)


docs/integrations/engines/clickhouse.md
```diff
@@ -0,0 +1,295 @@
...
+This page describes SQLMesh support for the Clickhouse engine, including configuration options specific to Clickhouse.
+
+!!! note
+    Clickhouse may not be used for the SQLMesh [state connection](../../reference/configuration.md#connections).
+
+## Background
+
+[Clickhouse](https://clickhouse.com/) is a distributed, column-oriented SQL engine designed to rapidly execute analytical workloads.
+
+It provides users find-grained control of its behavior, but that control comes at the cost of complex configuration.
```
erindru, line 12: fine-grained?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2291,6 +2305,15 @@ def _rename_table(
     ) -> None:
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
+    def inject_query_setting(
+        self,
+        query: Query,
+        setting_name: str = "join_use_nulls",
```
erindru, line 2311: Should the base level have clickhouse-specific defaults?

tests/core/engine_adapter/test_clickhouse.py
```diff
@@ -333,3 +346,443 @@ def test_create_table_properties(make_mocked_engine_adapter: t.Callable, mocker)
...
+            "name": exp.DataType.build("VARCHAR"),
+            "price": exp.DataType.build("DOUBLE"),
+            "test_UPDATED_at": exp.DataType.build("TIMESTAMP"),
+            "test_valid_from": exp.DataType.build("TIMESTAMP"),
+            "test_valid_to": exp.DataType.build("TIMESTAMP"),
+        },
+        execution_time=datetime(2020, 1, 1, 0, 0, 0),
+    )
+
+    assert adapter.cursor.execute.call_args_list[4][0][0] == parse_one(
```
erindru, line 432: I'm curious - did `to_sql_calls(adapter)` not work here?

erindru:
Awesome work! I bet you learned way more than you wanted to about the Clickhouse internals

tests/core/engine_adapter/test_clickhouse.py
treysp: It does - oversight on my part

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +293,70 @@ def _build_partitioned_by_exp(
             this=exp.Schema(expressions=partitioned_by),
         )
 
+    def inject_query_setting(
+        self,
+        query: Query,
+        setting_name: t.Optional[str] = "join_use_nulls",
+        setting_value: t.Optional[str] = "1",
```
georgesittas, line 300: I don't think it's good practice to change the prototypes of methods, even if it's default values. Let's keep the definition consistent and do something like `setting_name = setting_name or "join_use_nulls` at the beginning of this method.

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +293,70 @@ def _build_partitioned_by_exp(
...
+        #      for their query.
+        #  - To prevent this, we:
+        #     - Determine if the user query already sets `join_use_nulls` - if so, we leave it as-is
+        #     - If the user query does not set `join_use_nulls`, we query the server for the current setting. If the
+        #         current setting is `0`, we add `join_use_nulls = 0` to the user query.
+        #  - We specify our setting on the outermost scope, so it applies to all CTEs/subqueries other than the user's
+        if setting_name and setting_value:
+            inject_setting = True
+            if check_server_default:
+                user_settings = query.args.get("settings")
```
georgesittas, line 319: Are we searching the whole AST in places where this method is called? I'm thinking that the `SETTINGS` clause applies to the whole query, so if it can be nested deeper than the top-level query we may miss it.

In this case, perhaps a `(select.args.get("settings") for select in find_all(Select))` generator could be used. Though there's some overhead to fetch every `Select` which is crappy...

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +293,70 @@ def _build_partitioned_by_exp(
...
+                if user_settings and [
+                    isinstance(setting, exp.EQ) and setting.this.this == setting_name
```
georgesittas, line 321: I think we can do:
```suggestion
                    isinstance(setting, exp.EQ) and setting.name == setting_name
```

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +293,70 @@ def _build_partitioned_by_exp(
...
+                query.set(
+                    "settings",
+                    (query.args.get("settings") or [])
+                    + [
+                        exp.EQ(
+                            this=exp.var(setting_name),
+                            expression=exp.Literal(this=setting_value, is_string=False),
+                        )
+                    ],
+                )
```
georgesittas, line 342: Consider refactoring this into:
```python
query.append("settings", exp.var(setting_name).eq_(exp.Literal.number(setting_value)))
```

sqlmesh/core/model/meta.py
```diff
@@ -152,7 +152,12 @@ def _normalize(value: t.Any) -> t.Any:
         return v
 
     @field_validator("storage_format", mode="before")
-    def _storage_format_validator(cls, v: t.Any) -> t.Optional[str]:
+    @field_validator_v1_args
+    def _storage_format_validator(cls, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[str]:
+        if isinstance(v, exp.Expression) and not (
+            isinstance(v, exp.Literal) or isinstance(v, exp.Identifier)
```
georgesittas, line 158:
```suggestion
            isinstance(v, (exp.Literal, exp.Identifier))
```

sqlmesh/utils/date.py
```diff
@@ -353,16 +353,16 @@ def to_time_column(
...
     if time_column_type.is_type(exp.DataType.Type.DATE):
         return exp.cast(exp.Literal.string(to_ds(time_column)), to="date")
-    if time_column_type.this in TEMPORAL_TZ_TYPES:
-        return exp.cast(exp.Literal.string(to_tstz(time_column)), to=time_column_type.this)
-    if time_column_type.this in exp.DataType.TEMPORAL_TYPES:
-        return exp.cast(exp.Literal.string(to_ts(time_column)), to=time_column_type.this)
+    if time_column_type.is_type(*TEMPORAL_TZ_TYPES):
+        return exp.cast(exp.Literal.string(to_tstz(time_column)), to=time_column_type)
+    if time_column_type.is_type(*exp.DataType.TEMPORAL_TYPES):
+        return exp.cast(exp.Literal.string(to_ts(time_column)), to=time_column_type)
```
georgesittas, line 359: Did these change due to the `Nullable` refactor?

georgesittas (commented):
Great work @treysp!

docs/integrations/engines/clickhouse.md
```diff
@@ -0,0 +1,295 @@
...
+        order_by = (col1, col2)
+    )
+);
+
+select
+    *
+from other_schema.other_table;
+```
+
+Note that there is an `=` between the `order_by` key name and value `(col1, col2)`.
```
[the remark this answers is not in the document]
treysp: yep, that's how the other properties look as well https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/?h=physical#physical_properties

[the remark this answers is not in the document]
georgesittas: Gotcha, forgot how these looked like. Thanks for clarifying 👍

docs/integrations/engines/clickhouse.md
```diff
@@ -0,0 +1,295 @@
...
+The SCD model definition query is embedded as a CTE in the full SQLMesh-generated query. If run alone, the model definition query would use the Clickhouse server's current `join_use_nulls` value.
+
+If that value is not `1`, the SQLMesh setting on the outer query would override the server value and produce incorrect results.
+
+Therefore, SQLMesh uses the following procedure to ensure the model definition query runs with the correct `join_use_nulls` value:
+
+- If the model query sets `join_use_nulls` itself, do nothing
+- If the model query does not set `join_use_nulls` and the current server `join_use_nulls` value is `1`, do nothing
+- If the model query does not set `join_use_nulls` and the current server `join_use_nulls` value is `0`, add `SETTINGS join_use_nulls = 0` to the CTE model query
+    - All other CTEs and the outer query will still execute with a `join_use_nulls` value of `1`
```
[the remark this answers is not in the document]
treysp: Are you asking about NULL not being used for missing rows in the user query only, or hypothetically about what would happen if we didn't inject the setting?

Assuming the former, they should work - every comparison they make is one of:
- Are these two non-NULL values equal?
- Is one of these values NULL and the other isn't?
- Are these both NULL?

If the user query does not use NULL for missing rows, all the missing rows their query returns will have values, so our comparisons are of two non-NULL values.

We could have an issue if the user query had a join, and one of the non-key columns in their join was used in the SCD primary key:
  - If there was no match for a row in their query, that column would get a default value instead of NULL.
  - SCD would see that default value as a "real" key
  - If the row subsequently had a match in their query, the key would modify
  - SCD would see that as a new key and treat it as a new row, rather than as a row update

I can add that to the docs.

Can you think of any other problematic scenarios?

sqlmesh/core/engine_adapter/clickhouse.py
treysp: Thank you - that felt wrong, but I couldn't quite see the alternative

sqlmesh/core/engine_adapter/clickhouse.py
treysp: Hmm...a user query could contain a subquery with a different setting, but that setting would have to be explicit (in which case our setting wouldn't overwrite it).

My logic is:
- If a scope sets its own value, our value will not override
- Any scope without its own value will search "outward" to determine its value
- Therefore, any nested scopes will search until they either hit a user-specified value or the outermost user scope where we add the server default. 

For example:
```
SELECT 
  * 
FROM (
  SELECT 
    * 
  FROM inner_left 
  LEFT JOIN inner_right 
  ON inner_left.id = inner_right.id 
  SETTINGS join_use_nulls = 1
) as outer_left
LEFT JOIN outer_right
ON outer_left.id = outer_right.id
```

Assume the server value is join_use_nulls=0, so we need to inject that value to preserve this query's outer scope behavior.

Adding join_use_nulls=0 to the user's outer query won't override the 1 in the subquery, so they will see the expected behavior.

sqlmesh/utils/date.py
treysp: They were initially made to accommodate the old behavior, but the rest of those changes were reverted after the nullable refactor.

I left these changes in because `.is_type(*LIST)` seems more correct and safer than `.this in LIST`. (It seems like we should avoid reaching into an node's `.this` unless there's no alternative.)

sqlmesh/core/engine_adapter/clickhouse.py
treysp: I didn't know about `append` - thank you!

docs/integrations/engines/clickhouse.md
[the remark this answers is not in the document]
georgesittas: I didn't have an exact example in mind, I think I was under the assumption that SCD2 required `NULL`s instead of values for missing rows. So then I thought "if the user's _not_ using `NULL`s for missing rows, will the SCD2 logic work correctly? Given that we won't set `join_use_nulls` ourselves in this case to respect the model query's semantics. Does this make sense?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1466,23 +1466,29 @@ def remove_managed_columns(
         if check_columns:
             row_check_conditions = []
             for col in check_columns:
-                t_col = col.copy()
+                col_qualified = col.copy()
+                col_qualified.set("table", exp.parse_identifier("joined"))
```
izeigerman, line 1470: just curious: shouldn't this just be `exp.to_identifier` since there isn't much parsing to speak of

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2293,6 +2307,15 @@ def _rename_table(
     ) -> None:
         self.execute(exp.rename_table(old_table_name, new_table_name))
 
+    def inject_query_setting(
```
izeigerman, line 2310: [Nit] shouldn't this be `_inject_query_setting`?

izeigerman: Ah, I see it's also used in table_diff

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2293,6 +2307,15 @@ def _rename_table(
...
+        self,
+        query: Query,
+        setting_name: t.Optional[str] = None,
```
izeigerman, line 2313: Looks like `setting_name` and `setting_value` are never used?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +299,64 @@ def _build_partitioned_by_exp(
...
+        #     - If the user query does not set `join_use_nulls`, we query the server for the current setting. If the
+        #         current setting is `0`, we add `join_use_nulls = 0` to the user query.
+        #  - We specify our setting on the outermost scope, so it applies to all CTEs/subqueries other than the user's
+        setting_name = setting_name if setting_name and setting_value else "join_use_nulls"
+        setting_value = setting_value if setting_name and setting_value else "1"
+        assert setting_name and setting_value
+
+        inject_setting = True
+        if check_server_default:
+            user_settings = query.args.get("settings")
```
izeigerman, line 328: Do we need to check nested queries too?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +299,64 @@ def _build_partitioned_by_exp(
...
+            if user_settings and [
+                isinstance(setting, exp.EQ) and setting.name == setting_name
+                for setting in user_settings
+            ]:
+                inject_setting = False
+            else:
+                server_setting = self.fetchone(
+                    f"SELECT value FROM system.settings WHERE name = '{setting_name}'"
```
izeigerman, line 336: I suggest using sqlglot builder for this:
```
exp.select("value").from_(...).where(exp.col("name").eq(exp.Literal.string(setting_name))
```

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +299,64 @@ def _build_partitioned_by_exp(
...
+                )[0]
+                inject_setting = setting_value != server_setting
```
izeigerman, line 338: Do we even need to check the value in `system.settings`? What's the problem with just always setting it explicitly?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +299,64 @@ def _build_partitioned_by_exp(
             this=exp.Schema(expressions=partitioned_by),
         )
 
+    def inject_query_setting(
```
izeigerman, line 302: This method seems too generic, and just by glancing at it, it's impossible to tell when I should be calling this vs not. Nowhere in the name / signature of this method there's any indication that it's only for SCD2 / table diffing / other cases that involve joins and rely on null values.

I understand the motivation to make it more abstract, but in this case it seems to be causing more confusion than clarity.

Can we perhaps make it purpose-driven for now, until a better abstraction emerges (if it ever does). Eg. something like:
```
def ensure_nulls_after_join
```
or something similar that would indicate the purpose and the outcome.

treysp: yep - that was my initial inclination, but i wasn't sure if it leaked too much CH-specific stuff into the base adapter

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -296,6 +299,73 @@ def _build_partitioned_by_exp(
             this=exp.Schema(expressions=partitioned_by),
         )
 
+    def add_nulls_after_join_setting(
```
izeigerman, line 302: [Nit] Now the name is too prescriptive and specific to clickhouse. For example, I can see how the same behavior can be achieved using query modifications other than a setting potentially.


A more descriptive name could be something like `ensure_nulls_for_unmatched_after_join`

izeigerman (approved):

PR #3130, opened 2024-09-13
Docs: clarify that time_column and macro vars are UTC

treysp:


georgesittas (approved):
Perhaps we should get rid of the "you can set a time zone [...]" bit in the [testing docs](https://sqlmesh.readthedocs.io/en/stable/concepts/tests/#freezing-time) to be consistent about discouraging altering the time zone? Not sure what's best here.

izeigerman (approved):

PR #3131, opened 2024-09-13
fix: types

etonlels:
We've started building off the Python APIs and noticed that `UniqueKeyDict` has a number of methods that are not type-safe, including `__getitem__`, `values`, and `items`. This PR attempts to correct that. Correcting `UniqueKeyDict` made it necessary to make some changes elsewhere to satisfy `mypy`. Let me know if I did those correctly or not.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -54,7 +54,7 @@ from sqlglot.lineage import GraphHTML
 from sqlmesh.core import analytics
 from sqlmesh.core import constants as c
 from sqlmesh.core.analytics import python_api_analytics
-from sqlmesh.core.audit import Audit, StandaloneAudit
+from sqlmesh.core.audit import Audit, ModelAudit, StandaloneAudit
 from sqlmesh.core.config import CategorizerConfig, Config, load_configs
 from sqlmesh.core.config.loader import C
 from sqlmesh.core.console import Console, get_console
@@ -319,7 +319,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         )
         self.dag: DAG[str] = DAG()
         self._models: UniqueKeyDict[str, Model] = UniqueKeyDict("models")
-        self._audits: UniqueKeyDict[str, Audit] = UniqueKeyDict("audits")
+        self._audits: UniqueKeyDict[str, ModelAudit] = UniqueKeyDict("audits")
         self._standalone_audits: UniqueKeyDict[str, StandaloneAudit] = UniqueKeyDict(
             "standaloneaudits"
         )
@@ -743,7 +743,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 pass
         return self.config
 
-    def config_for_node(self, node: str | Model | StandaloneAudit) -> Config:
+    def config_for_node(self, node: str | Model | Audit) -> Config:
         if isinstance(node, str):
             return self.config_for_path(self.get_snapshot(node, raise_if_missing=True).node._path)  # type: ignore
         return self.config_for_path(node._path)  # type: ignore
@@ -888,7 +888,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         """Format all SQL models and audits."""
         format_targets = {**self._models, **self._audits}
         for target in format_targets.values():
-            if not target._path.suffix == ".sql":
+            if target._path is None or target._path.suffix != ".sql":
                 continue
             with open(target._path, "r+", encoding="utf-8") as file:
                 before = file.read()
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2332,7 +2332,7 @@ def _list_of_calls_to_exp(value: t.List[t.Tuple[str, t.Dict[str, t.Any]]]) -> ex
 
 def _extract_audit_expressions(
     audits: t.Optional[dict[str, Audit]] = None,
-    inline_audits: t.Optional[dict[str, Audit]] = None,
+    inline_audits: t.Optional[t.Mapping[str, Audit]] = None,
     model_audits: t.Optional[t.Any] = None,
     default_audits: t.Optional[t.List[AuditReference]] = None,
 ) -> list:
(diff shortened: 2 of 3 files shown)


sqlmesh/core/context.py
```diff
@@ -743,7 +743,7 @@ def config_for_path(self, path: Path) -> Config:
                 pass
         return self.config
 
-    def config_for_node(self, node: str | Model | StandaloneAudit) -> Config:
+    def config_for_node(self, node: str | Model | Audit) -> Config:
```
tobymao, line 746: why was this one changed? i don't think regular audits should be passed in here right?

etonlels: So I changed this because otherwise I get the following error:
```
sqlmesh/core/context.py:895: error: Argument 1 to "config_for_node" of "GenericContext" has incompatible type "Union[SqlModel, SeedModel, PythonModel, ExternalModel, ModelAudit, StandaloneAudit]"; expected "Union[str, SqlModel, SeedModel, PythonModel, ExternalModel, StandaloneAudit]"  [arg-type]
sqlmesh/core/context.py:905: error: Argument 1 to "config_for_node" of "GenericContext" has incompatible type "Union[SqlModel, SeedModel, PythonModel, ExternalModel, ModelAudit, StandaloneAudit]"; expected "Union[str, SqlModel, SeedModel, PythonModel, ExternalModel, StandaloneAudit]"  [arg-type]
tests/core/test_snapshot_evaluator.py:2676: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Found 2 errors in 1 file (checked 414 source files)
```
But maybe the correct way to address this would be changing [this line](https://github.com/TobikoData/sqlmesh/pull/3131/files#diff-02c2689e9e8516b4f905e12fa4dac9b0fff86efae7964c29fb9337055f6f3d94R889) to
```python
        format_targets = {**self._models, **self._standalone_audits}
```
WDYT?

sqlmesh/core/context.py
```diff
@@ -1914,7 +1914,9 @@ def _snapshots(
 
         local_nodes = {**(models_override or self._models), **self._standalone_audits}
         nodes = local_nodes.copy()
-        audits = self._audits.copy()
+        audits = {
+            name: audit for name, audit in self._audits.items() if isinstance(audit, ModelAudit)
+        }
```
etonlels, line 1919: Would it be helpful to add a `_model_audits` attribute to `GenericContext` (like `_standalone_audits`) and use that here instead?

etonlels: Actually, it seems like maybe the type annotation for `_audits` should just be changed to `UniqueKeyDict[str, ModelAudit]`?

sqlmesh/core/context.py
tobymao: how are you getting this error? are you using mypy?

etonlels: Yes, `mypy`. Are you not able to recreate on your end?

etonlels: To be clear, I'm just running `make style` locally.

etonlels: Let me push with the change reverted and see if it shows up in your CI checks.

etonlels: Yep. Same error in your CI now:
Screenshot 2024-09-16 at 10 13 06

sqlmesh/utils/__init__.py
```diff
@@ -65,7 +65,7 @@ def random_id(short: bool = False) -> str:
     return uuid.uuid4().hex
 
 
-class UniqueKeyDict(dict, t.Mapping[KEY, VALUE]):
+class UniqueKeyDict(t.Dict[KEY, VALUE]):
```
tobymao, line 68: is this correct?

etonlels: Yes, this is the correct way to subclass `dict` without losing type information in Python 3.8. If you only had to worry about Python 3.9+, then it could just be `dict[KEY, VALUE]` instead. Happy to prove that this is correct, however, via tests (or through whatever means you deem necessary).

sqlmesh/core/context.py
tobymao: yea this one is not correct, i think you can just change _audits to be ModelAudit

sqlmesh/core/model/definition.py
```diff
@@ -2332,7 +2332,7 @@ def _list_of_calls_to_exp(value: t.List[t.Tuple[str, t.Dict[str, t.Any]]]) -> ex
 
 def _extract_audit_expressions(
     audits: t.Optional[dict[str, Audit]] = None,
-    inline_audits: t.Optional[dict[str, Audit]] = None,
```
tobymao, line 2335: why does your change make it so this one has to change? this seems like a regression

sqlmesh/core/context.py
tobymao: ok let's add it back

sqlmesh/core/context.py
etonlels: changed

sqlmesh/core/context.py
etonlels: restored

sqlmesh/core/model/definition.py
etonlels: Yeah I understand it may seem like that at first, but it has to do with Python's type system. The reason this needs to be changed is that Python `dict`s are _invariant_ in their value (and key) types. There's a good explanation as to why that is [here](https://stackoverflow.com/questions/78537075/why-is-dictint-int-incompatible-with-dictint-int-str/78537409#78537409). This means you can't pass a `dict[str, ModelAudit]` as a `dict[str, Audit]` argument. However, `Mapping` is a superclass of `dict` and is _covariant_ in value types. Therefore, changing this type to `Mapping[str, Audit]` satisfies the type constraints (and `inline_audits` isn't mutated at all in the function body, so we don't need a mutable type here).

PR #3138, opened 2024-09-16
feat: Add SQLMesh warning when an exp.Command is parsed

VaggelisD:
[Public slack context](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1726478874076549)

This PR overrides SQLGlot's `_warn_unsupported()` to suggest the usage of Jinja on the occasion that the (unsupported) statement is attempting to reference the physical table.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -318,9 +318,11 @@ def _parse_join(
 
 
 def _warn_unsupported(self: Parser) -> None:
-    self.__warn_unsupported()  # type: ignore
+    sql = self._find_sql(self._tokens[0], self._tokens[-1])[: self.error_message_context]
+
     logger.warning(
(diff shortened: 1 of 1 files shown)


sqlmesh/core/dialect.py
```diff
@@ -314,6 +317,13 @@ def _parse_join(
     return macro
 
 
+def _warn_unsupported(self: Parser) -> None:
+    self.__warn_unsupported()  # type: ignore
```
tobymao, line 321: i don't think we need to call super

VaggelisD: I kept it there to maintain the diagnostic message of which syntax is not supported; The following sqlmesh warning will not suffice for that matter. Wdyt, is it not needed anymore?

tobymao: you can make the current message more robust

VaggelisD: Not sure how to word this best for the average user who won't understand SQLGlot's `exp.Command`, SQLMesh's physical table references etc. Happy to change the message to whatever you prefer.

PR #3140, opened 2024-09-16
Fixed daemon detection to determine forking behavior

sammywachtel:
Introduced the `is_daemon_process` function in `utils/system.py` to determine if the current process is a daemon. Updated the logic in `constants.py` to use this new utility function for determining forking behavior.


izeigerman:
Hey @sammywachtel ! Thanks a lot for your submission. Can you please provide a little more context around the issue you ran into with the existing implementation and how the new approach is better.

sqlmesh/utils/system.py
```diff
@@ -0,0 +1,13 @@
...
+
+def is_daemon_process() -> bool:
+    """
+    Checks if the current process is a daemon process.
+
+    :return: True if the process is a daemon process, False otherwise.
+    :rtype: bool
+    """
+
+    return not os.isatty(sys.stdout.fileno())
```
izeigerman, line 13: Can you just do:
```
sys.stdout.isatty()
```
?

sammywachtel: Hey @izeigerman. I believe there's a risk of getting an AttributeError in some cases when the underlying file descriptor doesn't implement isatty(). Or if the stdout is in a wrapper that doesn't correctly implement isatty(). Anyway, it feels like best practice. It would work most of the time with just sys.stdout.isatty(), though.

izeigerman (approved):

sqlmesh/utils/system.py
sammywachtel: As for your question about why the change: I am working with the airflow integration and after one of the recent new versions (I was on 0.18.0 I think) I started getting errors surrounding the worker settings. I noticed the code in constants.py had a recent change that was supposed restrict the number of workers to 1 when in an airflow environment, but the check for the process being within a daemon was using the multiprocessing package which would only return a "true" if the daemon were a multiprocessing package daemon. It doesn't appear to be that way with airflow, but rather, it's using a standard system daemon for which there's no built-in check. I hope that helps!

tobymao:
@sammywachtel please sign the cla, thank you!

sammywachtel:
> @sammywachtel please sign the cla, thank you!

It is signed, but for some reason still shows up as pending.

PR #3141, opened 2024-09-16
fix: correctly generate when_matched SQL

eakmanrq:
Prior to this PR if a user had variables in their model definition and used when_matched then they would get a parsing error. This is because after we replace variables we generate SQL with the model def and reparse it which doesn't involve the Pydantic model validators. As a result we have to generate the SQL in a format that matches what the user would have originally put in which is with the comma separating it. 

Updated the test case to include a variable to make sure this is tested going forward.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -579,7 +579,12 @@ def _props_sql(self: Generator, expressions: t.List[exp.Expression]) -> str:
     size = len(expressions)
 
     for i, prop in enumerate(expressions):
-        sql = self.indent(f"{prop.name} {self.sql(prop, 'value')}")
+        value = prop.args.get("value")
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -579,7 +579,12 @@ def _props_sql(self: Generator, expressions: t.List[exp.Expression]) -> str:
     size = len(expressions)
 
     for i, prop in enumerate(expressions):
-        sql = self.indent(f"{prop.name} {self.sql(prop, 'value')}")
+        value = prop.args.get("value")
+        if prop.name == "when_matched" and isinstance(value, list):
+            output_value = ", ".join([self.sql(v) for v in value])
```
tobymao, line 584: don't need the [] the generator works fine

tobymao (approved):

PR #3143, opened 2024-09-17
Allows setting a clickhouse connection pool

ravenac95:
We have actually been finding some issues in general with clickhouse (I'll be putting together some notes soon). However, one of the important things we found was the clickhouse performance can be improved by allowing the creation of a connection pool. 

I figured the easiest thing to do would be to allow the user to just create the pool manager directly rather than attempt to expose the necessary config on the `ClickhouseConnectionConfig` class

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1427,17 +1427,6 @@ class ClickhouseConnectionConfig(ConnectionConfig):
 
     @property
     def _connection_factory(self) -> t.Callable:
-        """Returns a clickhouse connection. If the pool_manager_factory is set
-        the connect function is returned with the `pool_mgr` value set. The
-        factory should be something like:
-
-            def pool_manager_factory(config: ClickhouseConnectionConfig):
-                from clickhouse_connect.driver import httputil
-
-                return httputil.get_pool_manager(
-                    num_pools=config.concurrent_tasks
-                )
-        """
         from clickhouse_connect.dbapi import connect  # type: ignore
         from clickhouse_connect.driver import httputil  # type: ignore
         from functools import partial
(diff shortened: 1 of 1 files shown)


izeigerman:
Hey @ravenac95 , thanks for the submission! Why make it optional? Why wouldn't we want to always have a connection pool and just configure it unconditionally? Additionally, the suggested approach doesn't work with the yaml config format.

ravenac95:
@izeigerman alright, I made an update so the config just accepts a dict that can be passed into the pool_mgr factory from clickhouse.

sqlmesh/core/config/connection.py
```diff
@@ -1422,8 +1428,19 @@ def _engine_adapter(self) -> t.Type[EngineAdapter]:
     @property
     def _connection_factory(self) -> t.Callable:
         from clickhouse_connect.dbapi import connect  # type: ignore
+        from clickhouse_connect.driver import httputil  # type: ignore
+        from functools import partial
 
-        return connect
+        pool_manager_options: t.Dict[str, t.Any] = dict(
+            num_pools=self.concurrent_tasks,
```
eakmanrq, line 1435: As I understand from the docs the number of pools is equal to number of hosts or a max of 10. I don't see how it would correlate with the concept of `concurrent_tasks`.

ravenac95: Some of this is definitely ignorance on my part. Let me review the docs more closely but I'm fine setting any value here really.

ravenac95: Definitely correct, I'll make an update

ravenac95:
I'm going to run some tests! I'll get back to you. If you'd all like, we can close this for now and I can reopen if this is indeed necesary.

eakmanrq:
Yeah no problem @ravenac95. Lets go ahead and close it for now but let us know what you find and if we should take another look at this.

Issue #1315, opened 2023-08-15
Support Athena


nicor88:
Support for AWS Athena as adapter.

Happy to propose a PR - given some guidelines.

tobymao:
We don't have plans for this right now, but we're happy to accept a PR.

Come join [tobikodata.com/slack](url) to chat with us about it.

Going to close this in favor of https://github.com/TobikoData/sqlmesh/issues/124, but again, happy to help guide you through a pr.

nicor88:
@tobymao Athena is based on trino/presto - but there are some internal differences that make the difference. 
see for example https://github.com/dbt-athena/dbt-athena and https://github.com/starburstdata/dbt-trino
Most of the SQL commands, but how the connection is created and how the result is return back from athena is quite different from trino - I still suggest to keep this as another "issue" ;)

tobymao:
ok sounds good

alvaro-ponce:
hi! I am here just wondering how is this feature going on, I am currently giving SQLMesh a try and the Athena adapter is something I would definitively love

Thank you in advance!

pixie79:
As someone that builds Greenfield data estates we often start with Athena then switch to other tooling later to manage our data estate when there is volume. It would be great to see Athena supported here like it is with DBT core

georgesittas:
@erindru should we close this given #3154?

nicor88:
@erindru before closing this, I would like to know if:
* partitioned tables are supported
* table properties for iceberg specifically are supported (see [this](https://github.com/TobikoData/sqlmesh/pull/3154#discussion_r1775103036))

If not, those are relevant features that I've see used quite a lot in dbt-athena, and might be relevant to have those features in.

erindru:
With the initial implementation:
- Partitioned Iceberg tables are supported
- Partitioned Hive tables are partially supported. They can be created but operations that need to delete data will currently fail
- Table properties are supported, any `physical_properties` that don't have a special meaning to the adapter are passed through

Follow-up PR's will add:
- Full support for partitioned Hive tables (in so far as emulating the DBT method for incremental models of identifying affected partitions, looking up their location in the metastore, deleting the objects from S3 and then dropping the partition)
- Support for loading `dbt-athena` projects

So @georgesittas this isnt quite finished yet

nicor88:
@erindru Nice, those are the type of features that I was expecting, great work, I will give it a try soon:)

[...]

PR #3154, opened 2024-09-19
Feat: Athena adapter

erindru:
Initial implementation of an Athena adapter. Addresses #1315

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -326,15 +326,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-          filters:
-            branches:
-              only:
-                - main
+                #- snowflake
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
+                - athena
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -326,16 +326,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                #- snowflake
-                #- databricks
-                #- redshift
-                #- bigquery
-                #- clickhouse-cloud
+                - snowflake
+                - databricks
+                - redshift
+                - bigquery
+                - clickhouse-cloud
                 - athena
-          #filters:
-          #  branches:
-          #    only:
-          #      - main
+          filters:
+           branches:
+             only:
+               - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -18,6 +18,7 @@ from sqlmesh.core.config.base import BaseConfig
 from sqlmesh.core.config.common import (
     concurrent_tasks_validator,
     http_headers_validator,
+    validate_s3_location,
 )
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.utils.errors import ConfigError
@@ -1498,7 +1499,9 @@ class AthenaConnectionConfig(ConnectionConfig):
     # SQLMesh options
     s3_warehouse_location: t.Optional[str] = None
     concurrent_tasks: int = 4
-    register_comments: bool = False  # because Athena doesnt support comments in most cases
+    register_comments: Literal[False] = (
+        False  # because Athena doesnt support comments in most cases
+    )
     pre_ping: Literal[False] = False
 
     type_: Literal["athena"] = Field(alias="type", default="athena")
@@ -1508,9 +1511,18 @@ class AthenaConnectionConfig(ConnectionConfig):
     def _root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
         work_group = values.get("work_group")
         s3_staging_dir = values.get("s3_staging_dir")
+        s3_warehouse_location = values.get("s3_warehouse_location")
 
         if not work_group and not s3_staging_dir:
             raise ConfigError("At least one of work_group or s3_staging_dir must be set")
+
+        if s3_staging_dir:
+            values["s3_staging_dir"] = validate_s3_location(s3_staging_dir, error_type=ConfigError)
+
+        if s3_warehouse_location:
+            values["s3_warehouse_location"] = validate_s3_location(
+                s3_warehouse_location, error_type=ConfigError
+            )
 
         return values
 
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1483,6 +1483,66 @@ class ClickhouseConnectionConfig(ConnectionConfig):
         return {"compress": compress, "client_name": f"SQLMesh/{__version__}", **settings}
 
 
+class AthenaConnectionConfig(ConnectionConfig):
+    # PyAthena connection options
+    aws_access_key_id: t.Optional[str] = None
+    aws_secret_access_key: t.Optional[str] = None
+    role_arn: t.Optional[str] = None
+    role_session_name: t.Optional[str] = None
+    region_name: t.Optional[str] = None
+    work_group: t.Optional[str] = None
+    s3_staging_dir: t.Optional[str] = None
+    schema_name: t.Optional[str] = None
+    catalog_name: t.Optional[str] = None
+
+    # SQLMesh options
+    s3_warehouse_location: t.Optional[str] = None
+    concurrent_tasks: int = 4
+    register_comments: bool = False  # because Athena doesnt support comments in most cases
+    pre_ping: Literal[False] = False
+
+    type_: Literal["athena"] = Field(alias="type", default="athena")
+
+    @model_validator(mode="after")
+    @model_validator_v1_args
+    def _root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
+        work_group = values.get("work_group")
+        s3_staging_dir = values.get("s3_staging_dir")
+
+        if not work_group and not s3_staging_dir:
+            raise ConfigError("At least one of work_group or s3_staging_dir must be set")
+
+        return values
+
+    @property
+    def _connection_kwargs_keys(self) -> t.Set[str]:
+        return {
+            "aws_access_key_id",
+            "aws_secret_access_key",
+            "role_arn",
+            "role_session_name",
+            "region_name",
+            "work_group",
+            "s3_staging_dir",
+            "schema_name",
+            "catalog_name",
+        }
+
+    @property
+    def _engine_adapter(self) -> t.Type[EngineAdapter]:
+        return engine_adapter.AthenaEngineAdapter
+
+    @property
+    def _extra_engine_config(self) -> t.Dict[str, t.Any]:
+        return {"s3_warehouse_location": self.s3_warehouse_location}
+
+    @property
+    def _connection_factory(self) -> t.Callable:
+        from pyathena import connect  # type: ignore
+
+        return connect
+
+
 CONNECTION_CONFIG_TO_TYPE = {
     # Map all subclasses of ConnectionConfig to the value of their `type_` field.
     tpe.all_field_infos()["type_"].default: tpe
--- a/sqlmesh/core/engine_adapter/athena.py
+++ b/sqlmesh/core/engine_adapter/athena.py
@@ -4,6 +4,7 @@ import typing as t
 import logging
 from sqlglot import exp
 from sqlmesh.core.dialect import to_schema
+from sqlmesh.core.config.common import validate_s3_location
 from sqlmesh.core.engine_adapter.mixins import PandasNativeFetchDFSupportMixin
 from sqlmesh.core.engine_adapter.trino import TrinoEngineAdapter
 from sqlmesh.core.node import IntervalUnit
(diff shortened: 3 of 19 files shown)


georgesittas:
Released v25.22.0, should be able to upgrade and get the CI working now.

.circleci/continue_config.yml
```diff
@@ -326,15 +326,16 @@ workflows:
...
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-          filters:
-            branches:
-              only:
-                - main
+                #- snowflake
```
erindru, line 329: Note: i'll uncomment these immediately prior to merging. It's helpful to be able to run the Athena integration tests for this PR

sqlmesh/core/config/connection.py
```diff
@@ -1467,6 +1467,66 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
...
+    region_name: t.Optional[str] = None
+    work_group: t.Optional[str] = None
+    s3_staging_dir: t.Optional[str] = None
+    schema_name: t.Optional[str] = None
+    catalog_name: t.Optional[str] = None
+
+    # SQLMesh options
+    s3_warehouse_location: t.Optional[str] = None
+    concurrent_tasks: int = 4
+    register_comments: bool = False  # because Athena doesnt support comments in most cases
```
izeigerman, line 1485: Should we even let users set true here then?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+)
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core._typing import SchemaName, TableName
+
+logger = logging.getLogger(__name__)
+
+
+def _ensure_valid_location(value: str) -> str:
+    if not value.startswith("s3://"):
```
izeigerman, line 28: Shouldn't this be validated in the connection config instead?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+        """
+        Returns all the data objects that exist in the given schema and optionally catalog.
+        """
+        schema_name = to_schema(schema_name)
+        schema = schema_name.db
+        query = (
+            exp.select(
+                exp.case()
+                .when(
+                    # 'awsdatacatalog' is the default catalog that is invisible for all intents and purposes
```
izeigerman, line 119: This doesn't quite explain why do we set catalog to NULL of the actual value is `awsdatacatalog`. What does "invisible" actually mean here?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+                .when(
+                    exp.column("table_type", table="t").eq("BASE TABLE"),
+                    exp.Literal.string("table"),
+                )
+                .else_(exp.column("table_type", table="t"))
+                .as_("type"),
+            )
+            .from_(exp.to_table("information_schema.tables", alias="t"))
+            .where(
+                exp.and_(
```
izeigerman, line 138: `AND` seems redundant here

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+                "s3_base_location"
+            )  # pop because it's handled differently and we dont want it to end up in the TBLPROPERTIES clause
+            if isinstance(s3_base_location_property, exp.Expression):
+                base_uri = s3_base_location_property.name
+            else:
+                base_uri = s3_base_location_property
+
+        elif self.s3_warehouse_location:
+            # If the user has set `s3_warehouse_location` in the connection config, the base URI is <s3_warehouse_location>/<catalog>/<schema>/
+            catalog_name = table.catalog if hasattr(table, "catalog") else None
```
izeigerman, line 387: Looking at signatures we always expect `table` to be of type `exp.Table`. How could `catalog` and `db` attributes be missing, provided they are property methods in that class?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+            schema_name = table.db if hasattr(table, "db") else None
+            base_uri = os.path.join(
+                self.s3_warehouse_location, catalog_name or "", schema_name or ""
+            )
+        else:
+            # Assume the user has set a default location for this schema in the metastore
+            return None
+
+        table_name = table.name if hasattr(table, "name") else None
```
izeigerman, line 396: ditto

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+            partitioned_by=partitioned_by,
+            table_properties=table_properties,
+            table_description=table_description,
+            table_kind=table_kind,
+            **kwargs,
+        )
+
+        is_hive = self._table_type(table_properties) == "hive"
+
+        # Filter any PARTITIONED BY properties from the main column list since they cant be specified in both places
```
izeigerman, line 232: good old hive

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+        self.execute(f"DELETE FROM {table.sql(dialect=self.dialect, identify=True)}")
+
+    def _table_type(
+        self, table_properties: t.Optional[t.Dict[str, exp.Expression]] = None
+    ) -> t.Union[t.Literal["hive"], t.Literal["iceberg"]]:
+        """
+        Use the user-specified table_properties to figure out of this is a Hive or an Iceberg table
+        """
+        # if table_type is not defined or is not set to "iceberg", this is a Hive table
+        if table_properties and (table_type := table_properties.get("table_type", None)):
```
izeigerman, line 346: Why are we not using `storage_format` for this instead?

izeigerman: Basically any value that is not `iceberg` is treated as `hive`.

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -0,0 +1,441 @@
...
+            properties.append(location)
+
+        for name, value in table_properties.items():
+            properties.append(exp.Property(this=exp.var(name), value=value))
+
+        if is_iceberg and expression:
+            # To make a CTAS expression persist as iceberg, alongside setting `table_type=iceberg` (which the user has already
+            # supplied in physical_properties and is thus set above), you also need to set:
+            #  - is_external=false
+            #  - table_location='s3://<path>'
```
izeigerman, line 325: Should we ensure somehow that the location has already been set at this point?

sqlmesh/core/config/connection.py
erindru: How do we force it to false and make it unsettable? Is that the `register_comments: t.Literal[False] = False` syntax?

sqlmesh/core/engine_adapter/athena.py
erindru: It's also used in `_table_location` to ensure that final location is valid since that is assembled at runtime and relies on user input from `physical_properties`

sqlmesh/core/engine_adapter/athena.py
erindru: It's because the integration test `test_get_data_objects` expects that if it passes a schema like `test_schema_x` (as opposed to a catalog-schema combo like `test_catalog.test_schema_x`) to `get_data_objects()`, the resulting data objects should have None set on the catalog property.

I'll amend the comment

sqlmesh/core/engine_adapter/athena.py
erindru: Good catch, this didnt get updated when I tightened up the signature

sqlmesh/core/engine_adapter/athena.py
erindru: I thought about `storage_format` but decided not to use it because it describes a different concept. Both Hive and Iceberg tables support different storage formats.

For example, a Hive table can be `STORED AS PARQUET` or `STORED AS ORC` or if you really dont like your colleagues `STORED AS TEXTFILE`.

Same for Iceberg, the internal `storage_format` can be set to `parquet` or `orc` or whatever the engine supports.

So `storage_format=hive` / `storage_format=parquet` doesn't make sense because they're _table_ formats. We dont have a top-level `table_format` property and I didnt want to add one just for Athena

sqlmesh/core/engine_adapter/athena.py
erindru: Oh nice catch!

erindru: Actually, the location will be set already a few lines up if the user supplied it (or `s3_warehouse_location` was set in the config).

But ive just done some tests and unlike Trino, it looks like Athena will not automatically generate table locations for you if the schema the table is in was created with a location set.

I created a schema using `CREATE SCHEMA foo LOCATION 's3://path'` and then tried to create both Hive and Iceberg tables in that schema without setting a location explicitly. Both times it failed with an error asking to set the location.

So i'll tighten this up and throw an error if SQLMesh cant figure out the table location

sqlmesh/core/config/connection.py
izeigerman: Yes, exactly

sqlmesh/core/engine_adapter/athena.py
izeigerman: I see. I'd still add validation to the config to fail faster.

sqlmesh/core/engine_adapter/athena.py
izeigerman: FWIW, when used with spark `iceberg` is provided through `storage_format` because the SparkSQL syntax looks like: 
```
CREATE TABLE ... USING [iceberg|parquet|etc]
```
Should we be consistent?

erindru: If you did that, how would you specify "Hive + Parquet", "Hive + ORC", "Iceberg + Parquet", "Iceberg + ORC" etc?

The table format is independent of the storage format.

`storage_format` is [currently defined](https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/?h=storage#storage_format) as:
```
Storage format is a property for engines such as Spark or Hive that support storage formats such as parquet and orc.
```
in the docs which I think makes sense. It mentions the actual format of the files on disk, eg Parquet or ORC. Not the type of table that is managing those files. 

The [Hive syntax](https://iceberg.apache.org/docs/latest/hive/#non-partitioned-tables) is:
```
CREATE TABLE x (i int) STORED BY ICEBERG STORED AS ORC;
```

Which differentiates `STORED BY` (table format) and `STORED AS` (storage format). `STORED AS ICEBERG` is invalid.

I'm not super familiar with Spark but it looks like `USING` [triggers a "Data Source"](https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-datasource.html#parameters) and then if Iceberg is available on the classpath, you can `USING iceberg` and then [set the storage format](https://iceberg.apache.org/docs/latest/spark-ddl/#create-table) in `TBLPROPERTIES` (or in the Iceberg catalog config). So it definitely blurs these concepts compared to other engines

erindru: I guess if we wanted to represent both concepts using the `storage_format` property, we would need to overload the syntax and do some basic parsing.

Something like:

`storage_format iceberg` = iceberg + whatever default is configured in the engine
`storage_format iceberg[parquet]` = iceberg + parquet
`storage_format iceberg[orc]` = iceberg + orc

etc

sqlmesh/core/config/connection.py
erindru: Ok, done

sqlmesh/core/engine_adapter/athena.py
erindru: Ok, i've moved it to a common location and call it from a Pydantic validator as well as within the Athena adapter. I also added some tests to show it was working

sqlmesh/core/engine_adapter/athena.py
izeigerman: Honestly, I haven't seen anyone using iceberg with anything other than parquet, so values `parquet`, `orc` automatically assume Hive, while `iceberg` just assumes iceberg with parquet as per CREATE TABLE syntax in spark: https://iceberg.apache.org/docs/nightly/spark-ddl/#create-table

izeigerman: IMHO, I wouldn't worry too much about ORC, at least in the context of Iceberg.

erindru: I think the bigger picture here is that table formats and storage formats are two separate concepts and we should not conflate the two.

We also shouldn't just assume that people will only ever want to use parquet and hardcode it, isnt the point of `physical_properties` to expose features of the underlying database system?

I can already imagine someone with an established data lake that chose ORC for whatever reason and has downstream consumers that expect Iceberg/ORC tables, and now they cant easily use SQLMesh because SQLMesh will only create Iceberg/Parquet tables.

We also need to add DBT support and even DBT [clearly separates these concepts](https://github.com/dbt-labs/dbt-athena/tree/main?tab=readme-ov-file#iceberg)

izeigerman (approved):

sqlmesh/core/engine_adapter/athena.py
nicor88: @erindru @izeigerman few comments from my side about this, after maintaining dbt-athena for 2 years.

People want to be in control of the iceberg tables that they use.
What @izeigerman raised 
> I haven't seen anyone using iceberg with anything other than parquet, so values
it's correct, but as a user I want to be in control of the data format used by Iceberg, therefore they might team that want to use ORC, other that want to use parquet, but the user must be in control, and have the possibility to decide.

This is valid for also for table properties, full list is available [here](https://docs.aws.amazon.com/athena/latest/ug/create-table-as.html#ctas-table-properties).
There are certain property that can be templated of course, but some other properties like vacuum_min_snapshots_to_keep (just to mention one), where the final user must be in control.

erindru: Thanks Nicola, appreciate your insight!

PR #3159, opened 2024-09-19
Docs: clarify how to set table location for trino on aws glue

treysp:


erindru:
Interesting, there's some overlap with Athena: https://github.com/TobikoData/sqlmesh/pull/3154/files#diff-810f5d6dd3e7fda3367cac3598147973d1db787a6026543dc9461dc9ac2f4872R29

docs/integrations/engines/trino.md
```diff
@@ -60,6 +60,57 @@ hive.metastore.uri=thrift://example.net:9083
...
+
+If a model should not use the default location, you may specify its location in the model DDL's [`physical_properties` dictionary](../../concepts/models/overview.md#physical_properties). This value will be passed to the `LOCATION` property in the model's `CREATE` statements.
+
+For example:
+
+``` sql linenums="1"
+MODEL (
+  name my_schema.my_model,
+  physical_properties (
+    location = 's3://my-bucket/'
```
erindru, line 86: One problem I found setting this at the model level is that it the path doesn't automatically get adjusted for each snapshot. So you have the situation where SQLMesh creates multiple snapshots for the model in the same location and they clobber each other

treysp: That is definitely a problem! Would it make sense to pull that out into a mixin that we use in both Athena and Trino, or is it too specific?

erindru: I think it could work as a mixin, it would essentially need to detect `s3_base_location` and rewrite it to `location` (injecting the snapshot table name) before handing the properties off to the adapter for further processing.

Perhaps a follow-up PR once the Athena one has been reviewed?

PR #3161, opened 2024-09-20
fix: support all model methods in dbt projects

tobymao:


sqlmesh/dbt/manifest.py
```diff
@@ -398,6 +399,9 @@ def _extra_dependencies(self, target: str, package: str) -> Dependencies:
         for call_name, node in extract_call_names(target, cache=self._calls):
             if call_name[0] == "config":
                 continue
+            elif isinstance(node, jinja2.nodes.Getattr):
+                if call_name[0] == "model":
+                    dependencies.macros.append(MacroReference(name=".".join(call_name)))
```
izeigerman, line 404: Repurposing macro references for this feels like a hack. Can we have a separate set in `Dependencies` instead?

sqlmesh/dbt/basemodel.py
```diff
@@ -289,10 +294,27 @@ def sqlmesh_model_kwargs(self, context: DbtContext) -> t.Dict[str, t.Any]:
...
             self.dependencies.macros, package=self.package_name
         )
+
+        model_attrs = {
+            macro.name.split(".")[1]
+            for macro in self.dependencies.macros
+            if macro.name.startswith("model.")
+        }
+
+        model: AttributeDict[str, t.Any] = AttributeDict(
```
izeigerman, line 304: [Nit] Let's clarify it refers to a node in the manifest, eg. `model_node` or `model_manifest_node`.

sqlmesh/dbt/basemodel.py
```diff
@@ -289,10 +294,21 @@ def sqlmesh_model_kwargs(self, context: DbtContext) -> t.Dict[str, t.Any]:
         jinja_macros = model_context.jinja_macros.trim(
             self.dependencies.macros, package=self.package_name
         )
+
+        model_node: AttributeDict[str, t.Any] = AttributeDict(
```
izeigerman, line 298: btw, I don't think you need to convert to `AttributeDict` since the validator within `JinjaMacros` should take of that.

sqlmesh/dbt/manifest.py
```diff
@@ -398,6 +399,9 @@ def _extra_dependencies(self, target: str, package: str) -> Dependencies:
         for call_name, node in extract_call_names(target, cache=self._calls):
             if call_name[0] == "config":
                 continue
+            elif isinstance(node, jinja2.nodes.Getattr):
```
izeigerman, line 402: We already have a similar logic couple lines above for `adapter`:
```
 if call_name[0] != "adapter":
```
can we consolidate?

tobymao: no this is done for type checking, because getattr doesn't have the same argument thing as Call

tests/dbt/test_manifest.py
```diff
@@ -31,6 +31,7 @@ def test_manifest_helper(caplog):
     assert models["top_waiters"].dependencies == Dependencies(
         refs={"sushi.waiter_revenue_by_day", "waiter_revenue_by_day"},
         variables={"top_waiters:revenue", "top_waiters:limit"},
+        model_attrs={"columns"},
```
izeigerman, line 34: can we also check some other attribute, eg. `config`?

izeigerman (approved):

Issue #3163, opened 2024-09-23
Support the 'column_types' attribute for dbt seeds


izeigerman:
According to [docs](https://docs.getdbt.com/reference/resource-configs/column_types), the `column_types` attribute is the idiomatic way to provide column types for seeds in dbt but it's currently not supported in SQLMesh.

Currently, SQLMesh supports only the [columns](https://docs.getdbt.com/reference/resource-properties/columns) attribute and the data types provided there, which can be specified for both models and seeds.

We have to make sure that we correctly reconcile the `column_types` and the `columns` attributes if both a provided for a given seed.

PR #3172, opened 2024-09-24
Feat: Support column types for dbt seeds

Themiscodes:
Add support for the `column_types` attribute for seeds in dbt, fixes: #3163 

If either or both `column_types` and `columns` are defined, they are reconciled to determine the seed configs. Additionally, support for the `quote_columns` property has been added. They can be defined in the `dbt_project.yml` file:

```yaml
seeds:
  <seed name>:
    +quote_columns: true
    +column_types:
      <column name>: <SQL data type>
```

or in the seed properties configuration file:

```yaml
seeds:
  - name: <seed name>
    config:
      quote_columns: true
      column_types:
        <column name>: <SQL data type>
    columns:
        - name: <column name>
          description: "column description"
```

[dbt docs](https://docs.getdbt.com/reference/resource-configs/column_types)

### DIFF
--- a/docs/integrations/dbt.md
+++ b/docs/integrations/dbt.md
@@ -202,19 +202,7 @@ SQLMesh parses seed CSV files using [Panda's `read_csv` utility](https://pandas.
 
 dbt parses seed CSV files using [agate's csv reader](https://agate.readthedocs.io/en/latest/api/csv.html#csv-reader-and-writer) and [customizes agate's default type inference](https://github.com/dbt-labs/dbt-common/blob/ae8ffe082926fdb3ef2a15486588f40c7739aea9/dbt_common/clients/agate_helper.py#L59).
 
-If SQLMesh and dbt infer different column types for a seed CSV file, you may specify your desired data types in a [seed properties configuration file](https://docs.getdbt.com/reference/seed-properties).
-
-Specify a column's SQL data type in its `data_type` key, as shown below. The file must list all columns present in the CSV file; SQLMesh's default type inference will be used for columns that do not specify the `data_type` key.
-
-``` yaml
-seeds:
-  - name: <seed name>
-    columns:
-      - name: <column name>
-        data_type: <SQL data type>
-```
-
-Alternatively, you can specify a [column_types](https://docs.getdbt.com/reference/resource-configs/column_types) dictionary in your `dbt_project.yml` file, where the keys define the column names and the values the data types.
+If SQLMesh and dbt infer different column types for a seed CSV file, you may specify a [column_types](https://docs.getdbt.com/reference/resource-configs/column_types) dictionary in your `dbt_project.yml` file, where the keys define the column names and the values the data types.
 
 ``` yaml
 seeds:
@@ -223,7 +211,7 @@ seeds:
       <column name>: <SQL data type>
 ```
 
-You can also define this dictionary in the seed properties configuration file.
+Alternatively, you can define this dictionary in the seed [seed properties configuration file](https://docs.getdbt.com/reference/seed-properties).
 
 ``` yaml
 seeds:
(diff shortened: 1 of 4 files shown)


tests/dbt/test_transformation.py
```diff
@@ -410,6 +410,40 @@ def test_seed_columns():
...
+            "zipcode": "text",
+        },
+        columns={
+            "zipcode": ColumnConfig(name="zipcode", description="Business zipcode"),
+        },
+        quote_columns=True,
+    )
+
+    expected_column_types = {
+        "address": parse_one("text", into=exp.DataType),
```
erindru, line 429: Open question: In terms of using SQLGlot, is the preferred way to construct `DataType` objects vs `exp.DataType.build("text")`?

georgesittas: We should use `exp.DataType.build`, that's the prevalent pattern in the codebase.

docs/integrations/dbt.md
```diff
@@ -214,6 +214,25 @@ seeds:
         data_type: <SQL data type>
 ```
 
+Alternatively, you can specify a [column_types](https://docs.getdbt.com/reference/resource-configs/column_types) dictionary in your `dbt_project.yml` file, where the keys define the column names and the values the data types.
+
+``` yaml
+seeds:
+  <seed name>
+    +column_types:
+        <column name> : <SQL data type>
```
georgesittas, line 223: [Nit] let's be consistent and use 2 space indentation and no space between `<column name>` and `:` (ditto for the section below).

georgesittas (commented):
LGTM but lacking some dbt adapter context, @crericha wdyt?

tests/dbt/test_transformation.py
Themiscodes: sure I will change these and make sure to use this way from now on

docs/integrations/dbt.md
```diff
@@ -214,6 +214,25 @@ seeds:
         data_type: <SQL data type>
 ```
 
+Alternatively, you can specify a [column_types](https://docs.getdbt.com/reference/resource-configs/column_types) dictionary in your `dbt_project.yml` file, where the keys define the column names and the values the data types.
```
izeigerman, line 217: I think this should actually be a primary way to specify column types for seeds

sqlmesh/dbt/seed.py
```diff
@@ -31,10 +32,18 @@ class SeedConfig(BaseModelConfig):
...
     delimiter: str = ","
+    column_types: t.Optional[t.Dict[str, str]] = None
+    quote_columns: t.Optional[bool] = False
 
     def to_sqlmesh(self, context: DbtContext) -> Model:
         """Converts the dbt seed into a SQLMesh model."""
         seed_path = self.path.absolute().as_posix()
+
+        for name, data_type in (self.column_types or {}).items():
+            column = self.columns.setdefault(name, ColumnConfig(name=name))
```
izeigerman, line 43: Can we somehow avoid mutating the SeedConfig object and copy the `columns` instead? Otherwise, it appears that `to_sqlmesh` now has a side-effect which is not expected

Themiscodes: Yes it might not be the perfect solution, but to avoid calling `column_types_to_sqlmesh` twice I chose to copy the columns and after doing the translation in `sqlmesh_model_kwargs` (which uses `self.columns`) reset them back to the original

docs/integrations/dbt.md
Themiscodes: Right realigned this part

sqlmesh/dbt/seed.py
Themiscodes: sure refactored this to accept this optional dictionary

izeigerman (approved):

PR #3173, opened 2024-09-24
Feat: allow formatter to use CAST over :: syntax

georgesittas:
Allows users to change the formatter so that it always produces the `CAST(foo AS type)` syntax over `::`. This may come in handy for users that want to easily copy-paste their model queries without having to switch back the `::` manually.

Context: https://tobiko-data.slack.com/archives/C07HJH22U9Z/p1727216482891999

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -172,8 +172,8 @@ Options:
   -t, --transpile TEXT        Transpile project models to the specified
                               dialect.
   --append-newline            Include a newline at the end of each file.
-  --cast-function             Use the standard CAST function over the ::
-                              syntax.
+  --no-rewrite-casts          Preserve the existing casts, without rewriting
+                              them to use the :: syntax.
   --normalize                 Whether or not to normalize identifiers to
                               lowercase.
   --pad INTEGER               Determines the pad size in a formatted string.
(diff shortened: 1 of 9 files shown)


docs/reference/cli.md
```diff
@@ -172,6 +172,8 @@ Options:
   -t, --transpile TEXT        Transpile project models to the specified
                               dialect.
   --append-newline            Include a newline at the end of each file.
+  --cast-function             Use the standard CAST function over the ::
```
erindru, line 175: Hmm, not sure on this naming. `--cast-function` makes it sounds like it's for casting functions.

What do you think of `--no-rewrite-casts` ?

georgesittas: Yep, seems better. I'll amend, thanks!

georgesittas: Done

sqlmesh/core/context.py
```diff
@@ -900,6 +900,7 @@ def evaluate(
     def format(
         self,
         transpile: t.Optional[str] = None,
+        no_rewrite_casts: t.Optional[bool] = None,
```
erindru, line 903: This API is now a bit inverted internally.

Would `rewrite_casts: bool = True` make more sense? Then the CLI can set it to `False` if `--no-rewrite-casts` was set.

georgesittas: Done

tobymao (approved):

PR #3175, opened 2024-09-25
Feat!: Add new table_format property alongside storage_format

erindru:
Context: https://github.com/TobikoData/sqlmesh/pull/3154#discussion_r1772042796

Currently, the `storage_format` property is a bit overloaded. It's defined as setting the on-disk file format (eg `parquet`, `orc`) which makes sense but it's also been overloaded in various places to include _table_ formats such as `iceberg`.

This PR introduces a `table_format` property to clearly separate table formats from storage formats.

The problem with the current setup of a single property is that it makes it difficult to describe concepts like "Iceberg + ORC", "Iceberg + Parquet", "Hive + Ion" etc.

Since `orc` and `parquet` are already valid values for `storage_format`,  `table_format` is a new property to hold the table format (eg `hive` or `iceberg`). To keep backwards compatibility, it's opt-in, so engine adapters need to explicitly take advantage of it.

Implementing this is a precursor for compatibility with models created using [dbt-athena](https://github.com/dbt-labs/dbt-athena/tree/main?tab=readme-ov-file#iceberg) which cleanly separates these concepts.

It also means that any users that may have gone "all in" on a format like Iceberg/ORC can use SQLMesh to create tables that are compatible with their existing downstream consumers. Currently, SQLMesh assumes Parquet when `storage_format=iceberg`


sqlmesh/core/engine_adapter/athena.py
```diff
@@ -66,10 +66,8 @@ def create_state_table(
             table_name,
             columns_to_types,
             primary_key=primary_key,
-            table_properties={
-                # it's painfully slow, but it works
-                "table_type": exp.Literal.string("iceberg")
-            },
+            # it's painfully slow, but it works
+            table_format="iceberg",
```
erindru, line 70: Promoting it to a top-level property makes it explicit and consistent across adapters

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -297,15 +301,12 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+        self, table_type: t.Optional[str] = None
     ) -> t.Union[t.Literal["hive"], t.Literal["iceberg"]]:
-        """
-        Use the user-specified table_properties to figure out of this is a Hive or an Iceberg table
-        """
+        if table_type and table_type.lower() == "iceberg":
+            return "iceberg"
+
         # if we cant detect any indication of Iceberg, this is a Hive table
-        if table_properties and (table_type := table_properties.get("table_type", None)):
```
erindru, line 306: This became a lot simpler, we now dont need to try and parse the table type out of a `physical_properties` entry

sqlmesh/migrations/v0057_add_table_format.py
```diff
@@ -0,0 +1,5 @@
+"""Add table_format to the model top-level properties"""
```
erindru, line 1: I think this is correct? The goal is to rewrite the snapshots because their fingerprints all change with the addition of the property

tests/core/engine_adapter/test_athena.py
```diff
@@ -149,9 +149,9 @@ def test_create_table_iceberg(adapter: AthenaEngineAdapter) -> None:
             name test_table,
             kind FULL,
             partitioned_by (colc, bucket(16, cola)),
+            table_format iceberg,
             storage_format parquet,
             physical_properties (
```
erindru, line 154: ^ this is what it looks like inside a model definition

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -160,8 +161,14 @@ def _build_table_properties_exp(
...
         properties: t.List[exp.Expression] = []
 
-        if storage_format:
-            properties.append(exp.FileFormatProperty(this=exp.Var(this=storage_format)))
+        if table_format and storage_format and self.dialect == "spark":
+            properties.append(exp.FileFormatProperty(this=exp.Var(this=table_format)))
+            properties.append(
+                exp.Property(this="write.format.default", value=exp.Literal.string(storage_format))
+            )
+        else:
```
izeigerman, line 169: `elif storage_format`?

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -160,8 +161,14 @@ def _build_table_properties_exp(
     ) -> t.Optional[exp.Properties]:
         properties: t.List[exp.Expression] = []
 
-        if storage_format:
-            properties.append(exp.FileFormatProperty(this=exp.Var(this=storage_format)))
+        if table_format and storage_format and self.dialect == "spark":
```
izeigerman, line 164: What if I only have `table_format` but not `storage_format`? Technically if I specify `iceberg` in `table_format` I may want to use a whatever default storage format it comes with (i.e. parquet)

sqlmesh/core/model/meta.py
```diff
@@ -151,7 +152,7 @@ def _normalize(value: t.Any) -> t.Any:
 
         return v
 
-    @field_validator("storage_format", mode="before")
+    @field_validator("table_format", "storage_format", mode="before")
     @field_validator_v1_args
     def _storage_format_validator(cls, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[str]:
```
izeigerman, line 157: [Nit] The name is no longer accurate

sqlmesh/migrations/v0057_add_table_format.py
izeigerman: I'm guessing we'll continue supporting the old way of things. Should we deprecate usage of `iceberg` in `storage_format` and emit a warning?

tests/core/engine_adapter/test_trino.py
```diff
@@ -400,3 +400,47 @@ def test_delta_timestamps(make_mocked_engine_adapter: t.Callable):
...
+        columns_to_types=model.columns_to_types_or_raise,
+        table_format=model.table_format,
+        storage_format=model.storage_format,
+    )
+
+    # Trino needs to ignore the `table_format` property because to create Iceberg tables, you target an Iceberg catalog
+    # rather than explicitly telling it to create an Iceberg table
+    assert to_sql_calls(adapter) == [
+        'CREATE TABLE IF NOT EXISTS "iceberg"."test_table" ("cola" TIMESTAMP, "colb" VARCHAR, "colc" VARCHAR) WITH (FORMAT=\'ORC\')',
+        'CREATE TABLE IF NOT EXISTS "iceberg"."test_table" WITH (FORMAT=\'ORC\') AS SELECT CAST("cola" AS TIMESTAMP) AS "cola", CAST("colb" AS VARCHAR) AS "colb", CAST("colc" AS VARCHAR) AS "colc" FROM (SELECT CAST(1 AS TIMESTAMP) AS "cola", CAST(2 AS VARCHAR) AS "colb", \'foo\' AS "colc") AS "_subquery"',
```
izeigerman, line 445: I don't see `iceberg` being mentioned anywhere. Is that expected?

erindru: So in Trino, you dont tell it to create an Iceberg table.

You configure an Iceberg catalog and every table created in there is an Iceberg table. Same for Hive, you configure a Hive catalog and every table created in there is a Hive table.

So `table_format` is actually a no-op on Trino and `storage_format` chooses the on-disk format as usual. This test is checking that `table_format` isnt propagating into the table properties (eg its ignored instead of being output as `FORMAT='ICEBERG'` which would be invalid)

sqlmesh/core/engine_adapter/mixins.py
erindru: This logic was for backwards compatibility for people who might have `storage_format=iceberg` in their models already. It takes `table_format` into account only if both are defined.

But I can improve this

sqlmesh/migrations/v0057_add_table_format.py
erindru: > What about migrating existing snapshots that currently specify iceberg in storage_format?
Wouldn't that lead to a big diff on the next plan because the model files wont have `table_format iceberg` in them?

> Should we deprecate usage of iceberg in storage_format and emit a warning?
Yes, that's a good idea

sqlmesh/core/engine_adapter/mixins.py
izeigerman: Yeah, I get that.  So far I see 2 scenario being covered:
1. Only `storage_format` is specified - this is for backward compatibilty
2. Both `table_format` and `storage_format` are specified - this is new

What about a case when there is `table_format` but no `storage_format`?

erindru: You're right, that was missing. I've refactored the logic and added tests to cover the following scenarios for Spark:

- both `table_format` and `storage_format` defined
- just `table_format` defined
- just `storage_format` defined

sqlmesh/core/model/meta.py
erindru: Adjusted

.circleci/continue_config.yml
```diff
@@ -332,10 +332,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
```
erindru, line 335: Reminder to self: Uncomment these prior to merging

sqlmesh/migrations/v0057_add_table_format.py
erindru: I've added a warning in the model `after` validator

izeigerman (approved):

PR #3176, opened 2024-09-25
fix!: typing imports were not working

tobymao:


sqlmesh/core/macros.py
```diff
@@ -72,6 +72,13 @@ class MacroStrTemplate(Template):
 EXPRESSIONS_NAME_MAP = {}
 SQL = t.NewType("SQL", str)
 
+SUPPORTED_TYPES = {
+    "t": t,
```
eakmanrq, line 76: Does this work if they do 'typing.List'? If not want to support that too?

tobymao: added it

eakmanrq (approved):

PR #3183, opened 2024-09-25
Chore: Include MSSQL into the list of engines recommended for storing SQLMesh state

izeigerman:
SQL server is a proper OLTP database and can easily be used to store the SQLMesh state.


tobymao (approved):

eakmanrq (commented):
Should update this too right? https://sqlmesh.readthedocs.io/en/stable/guides/configuration/?h=recommended#state-connection

Issue #3184, opened 2024-09-25
BigQuery alter of nested types does not support complex types


plaflamme:
The [recent change](https://github.com/TobikoData/sqlmesh/pull/3082) to support altering nested types in BigQuery does not properly handle complex types. That is, if an existing `STRUCT` field is being altered, sqlmesh fails to do so correctly if the field being appended is **itself** a complex type (nested or array).

More formally:

* altering `STRUCT(a INT64)` to `STRUCT(a INT64, b STRING)` works
* altering `STRUCT(a INT64)` to `STRUCT(a INT64, b ARRAY<STRING>)` or `STRUCT(a INT64, b STRUCT<c STRING>)` does not work

The issue is around the construction of `SchemaField` [here](https://github.com/TobikoData/sqlmesh/blob/e2829a13f20feb41bdd051aec535ef551d0c43e3/sqlmesh/core/engine_adapter/bigquery.py#L324). If `new_field[0]` is a complex type, the resulting string will be `ARRAY<STRING>` or similar which is incorrect. Properly handling nested types when constructing `SchemaField` is more involved.

PR #3185, opened 2024-09-25
fix: properly convert nested types to BQ schema

plaflamme:
This is related to https://github.com/TobikoData/sqlmesh/issues/3184 but does not fix it. What it provides is a general function for converting a sqlglot schema to a bigquery schema. This new function is not in the codepath of altering schemas, but it should be possible to use it there to actually fix the issue.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -370,20 +370,50 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin):
         )
         return self._db_call(job.result)
 
+    def __get_bq_schemafield(self, name: str, tpe: exp.DataType) -> bigquery.SchemaField:
+        from google.cloud import bigquery
+
+        mode = "NULLABLE"
+        if tpe.is_type(exp.DataType.Type.ARRAY):
+            mode = "REPEATED"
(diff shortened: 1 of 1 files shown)


izeigerman:
Thanks a lot for this, @plaflamme . Can this logic be tested at all?

Themiscodes:
Thank you @plaflamme! You're right when constructing the SchemaField we needed to traverse recursively to handle deeper levels of nesting correctly. And as you suggest changing:
https://github.com/TobikoData/sqlmesh/blob/9833a61dc589d51b81948d5fa82f3897fa38d05a/sqlmesh/core/engine_adapter/bigquery.py#L324
to use your `__get_bq_schemafield` function:
```py
self.__get_bq_schemafield(new_field[1], exp.DataType.build(new_field[0]))
```
does seem to resolve the issue when altering schemas, based on the examples I've tested so far.

plaflamme:
@izeigerman I didn't add tests since this is a private method, but we could extract it and test it on different `exp.DataType`s, yes. As for the codepath that uses this, probably under test in the bigquery integration tests, but I'm not sure.

@Themiscodes great! I wasn't confident applying this new function was going to be sufficient to fix the issue as there are other things going on in when altering a schema. Glad to hear it's working out.

izeigerman (approved):
Great fix, @plaflamme , thank you!

PR #3189, opened 2024-09-26
fix: more perf by offloading mork work to cores

tobymao:


sqlmesh/core/loader.py
```diff
@@ -560,14 +562,14 @@ def update_model_schemas(
...
-        _update_model_schemas_parallel(dag, models, schema, optimized_query_cache)
+        _update_model_schemas_parallel(dag, models, audits, schema, optimized_query_cache)
 
 
 def _update_schema_with_model(schema: MappingSchema, model: Model) -> None:
     columns_to_types = model.columns_to_types
     if columns_to_types:
         try:
-            schema.add_table(model.fqn, columns_to_types, dialect=model.dialect, normalize=False)
+            schema.add_table(model.fqn, columns_to_types, dialect=model.dialect)
```
tobymao, line 572: this isn't needed because the schema is already set with normalize=False

sqlmesh/core/model/cache.py
```diff
@@ -133,47 +138,65 @@ def _entry_name(model: SqlModel) -> str:
...
+def load_optimized_query_and_mapping(
+    model: Model, mapping: t.Dict
+) -> t.Tuple[str, t.Optional[str], str, str, t.Dict]:
+    assert _optimized_query_cache
+    assert _audits is not None
 
+    schema = MappingSchema(normalize=False)
+    for parent, columns_to_types in mapping.items():
+        schema.add_table(parent, columns_to_types, dialect=model.dialect)
+    model.update_schema(schema)
```
izeigerman, line 185: I'm not sure I follow: what's the point of doing the update schema for individual model on the executor side?

sqlmesh/core/model/cache.py
```diff
@@ -133,47 +138,65 @@ def _entry_name(model: SqlModel) -> str:
         return f"{model.name}_{crc32(hash_data)}"
 
 
-def optimized_query_cache_pool(optimized_query_cache: OptimizedQueryCache) -> ProcessPoolExecutor:
+def optimized_query_cache_pool(*args: t.Any) -> ProcessPoolExecutor:
```
izeigerman, line 141: Why can't we enumerate the arguments properly?

sqlmesh/core/model/cache.py
```diff
@@ -133,47 +138,65 @@ def _entry_name(model: SqlModel) -> str:
...
+_audits: t.Optional[t.Dict[str, ModelAudit]] = None
 
 
-@t.overload
-def load_optimized_query_cache(model_or_tuple: Model) -> t.Tuple[str, t.Optional[str]]: ...
+def _init_optimized_query_cache(
+    optimized_query_cache: t.Optional[OptimizedQueryCache] = None,
+    audits: t.Optional[t.Dict[str, ModelAudit]] = None,
+) -> None:
+    global _optimized_query_cache, _nodes, _audits
```
izeigerman, line 158: What's `_nodes`?

sqlmesh/core/loader.py
```diff
@@ -617,19 +620,29 @@ def process_models(completed_model: t.Optional[Model] = None) -> None:
...
                 futures.remove(future)
-                fqn, entry_name = future.result()
+                fqn, entry_name, data_hash, metadata_hash, mapping_schema = future.result()
                 model = models[fqn]
-                if entry_name:
-                    optimized_query_cache.with_optimized_query(model, entry_name)
-
+                model._data_hash = data_hash
+                model._metadata_hash = metadata_hash
+                model.update_schema(mapping_schema)
```
izeigerman, line 645: We wouldn't have to do this twice, if we were updating the schema prior to submitting a task like we used to

sqlmesh/core/model/cache.py
tobymao: update_schema is an expensive operation it's 

Models X Upstreams per Model X Columns Per upstream

sqlmesh/core/loader.py
tobymao: we need to do it here on the computed mapping because it was done in a different process. we do it in the process so that we can leverage cores

sqlmesh/core/model/cache.py
tobymao: deleting

sqlmesh/core/model/cache.py
tobymao: it's done on a different data type, so it's free, the expensive work is done in cores, then we can just set it in teh main process

izeigerman (approved):

PR #3192, opened 2024-09-27
fix!: correctly use execution time on new rows for scd type 2 by column

eakmanrq:
This corrects the SCD Type 2 By Column default behavior to use execution time for new rows after the table is initially loaded. This matches what the documentation example says it should behave. This does mean there is a behavior change for current users but the current behavior seems to be a bug. There is not a way for a user to maintain old behavior if they wanted it since we consider it a bug. 

From reviewing the code, `truncate` is a great way to identify if a table is being loaded for the first time. This also means that if a user issues a restatement on the table and wants to rebuild from scratch then it would also be set to `truncate` and would mimic loading the rows for the first time.

### DIFF
--- a/docs/concepts/models/model_kinds.md
+++ b/docs/concepts/models/model_kinds.md
@@ -736,10 +736,10 @@ This is the most accurate representation of the menu based on the source data pr
 
 ### SCD Type 2 By Column Configuration Options
 
-| Name                         | Description                                                                                                                                                                   | Type                      |
-|------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------|
-| columns                      | The name of the columns to check for changes. `*` to represent that all columns should be checked.                                                                            | List of strings or string |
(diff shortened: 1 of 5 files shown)


tests/core/engine_adapter/test_integration.py
```diff
@@ -1364,6 +1364,7 @@ def test_scd_type_2_by_time(ctx: TestContext):
         execution_time="2023-01-01 00:00:00",
         updated_at_as_valid_from=False,
         columns_to_types=input_schema,
+        truncate=True,
```
eakmanrq, line 1367: This properly mimics how the adapter will be used by truncating on first load and then not truncating on follow up load.

tests/core/engine_adapter/test_integration.py
```diff
@@ -1633,7 +1637,7 @@ def test_scd_type_2_by_column(ctx: TestContext):
                     "id": 5,
                     "name": "e",
                     "status": "inactive",
-                    "valid_from": "1970-01-01 00:00:00",
+                    "valid_from": "2023-01-05 00:00:00",
```
eakmanrq, line 1640: The prior value was a bug and this is it being fixed.

tobymao (approved):

PR #3201, opened 2024-09-30
Feat: Support INCREMENTAL_BY_TIME_RANGE models on Athena/Hive

erindru:
ref: issue #1315

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -326,16 +326,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
+                #- snowflake
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
+          # branches:
+          #   only:
+          #     - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -326,16 +326,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                #- snowflake
-                #- databricks
-                #- redshift
-                #- bigquery
-                #- clickhouse-cloud
+                - snowflake
+                - databricks
+                - redshift
+                - bigquery
+                - clickhouse-cloud
                 - athena
-          #filters:
-          # branches:
-          #   only:
-          #     - main
+          filters:
+           branches:
+             only:
+               - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/sqlmesh/core/engine_adapter/athena.py
+++ b/sqlmesh/core/engine_adapter/athena.py
@@ -3,8 +3,9 @@ from functools import lru_cache
 import typing as t
 import logging
 from sqlglot import exp
+from sqlglot.optimizer.qualify_columns import quote_identifiers
 from sqlmesh.core.dialect import to_schema
-from sqlmesh.core.config.common import validate_s3_location
+from sqlmesh.utils.aws import validate_s3_uri, parse_s3_uri
 from sqlmesh.core.engine_adapter.mixins import PandasNativeFetchDFSupportMixin
 from sqlmesh.core.engine_adapter.trino import TrinoEngineAdapter
 from sqlmesh.core.node import IntervalUnit
@@ -16,6 +17,8 @@ from sqlmesh.core.engine_adapter.shared import (
     DataObjectType,
     CommentCreationTable,
     CommentCreationView,
+    SourceQuery,
+    InsertOverwriteStrategy,
 )
 
 if t.TYPE_CHECKING:
@@ -53,8 +56,16 @@ class AthenaEngineAdapter(PandasNativeFetchDFSupportMixin):
     @s3_warehouse_location.setter
     def s3_warehouse_location(self, value: t.Optional[str]) -> None:
         if value:
-            value = validate_s3_location(value)
+            value = validate_s3_uri(value, base=True)
         self._s3_warehouse_location = value
+
+    @property
+    def s3_warehouse_location_or_raise(self) -> str:
+        # this makes tests easier to write without extra null checks to keep mypy happy
+        if location := self.s3_warehouse_location:
+            return location
+
+        raise SQLMeshError("s3_warehouse_location was expected to be populated; it isnt")
 
     def create_state_table(
         self,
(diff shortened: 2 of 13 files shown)


.circleci/continue_config.yml
erindru, line 1: I'll revert these prior to merge.

As an aside, we need a better way of allowing cloud test runs to be triggered on the relevant PR's.

CircleCI appears to be fairly limited in this regard but we might be able to do something with GitHub Actions triggering a CircleCI workflow in response to a PR comment

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -296,14 +307,28 @@ def _build_table_properties_exp(
...
+            table_name = exp.to_table(table_name)
+
+        # Truncating an Iceberg table is just DELETE FROM <table>
+        if self._query_table_type(table_name) == "iceberg":
+            return self.delete_from(table_name, exp.true())
+
+        # Truncating a partitioned Hive table is dropping all partitions and deleting the data from S3
+        if self._is_hive_partitioned_table(table_name):
+            self._clear_partition_data(table_name, exp.true())
+        else:
```
tobymao, line 320: elif s3_location := self._query...

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -317,6 +342,8 @@ def _query_table_type(
         Hit the DB to check if this is a Hive or an Iceberg table
         """
         table_name = exp.to_table(table_name)
+        quote_identifiers(table_name)
```
tobymao, line 345: why was this necessary?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -326,6 +353,15 @@ def _query_table_type(
                 return "hive"
         return "iceberg"
 
+    def _is_hive_partitioned_table(self, table_name: exp.Table) -> bool:
```
tobymao, line 356: if this takes a Table, maybe just call this arg table

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -373,6 +409,101 @@ def _find_matching_columns(
...
+        table_type = self._query_table_type(table_name)
+
+        if table_type == "iceberg":
+            # Iceberg tables work as expected, we can use the default behaviour
+            return super()._insert_overwrite_by_time_partition(
+                table_name, source_queries, columns_to_types, where, **kwargs
+            )
+
+        # For Hive tables, we need to drop all the partitions covered by the query and delete the data from S3
+        self._clear_partition_data(table_name, where)
```
tobymao, line 429: one shitty thing about this is we lose transactions, if something goes wrong here you lose data right

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -296,14 +307,28 @@ def _build_table_properties_exp(
         return None
 
     def _truncate_table(self, table_name: TableName) -> None:
-        table = exp.to_table(table_name)
-        # Athena doesnt support TRUNCATE TABLE. The closest thing is DELETE FROM <table> but it only works on Iceberg
-        self.execute(f"DELETE FROM {table.sql(dialect=self.dialect, identify=True)}")
+        if isinstance(table_name, str):
+            table_name = exp.to_table(table_name)
```
tobymao, line 311: i don't think this one is necessary, because you handle it it in all the other functions right?

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -373,6 +409,101 @@ def _find_matching_columns(
...
+        return super()._insert_overwrite_by_time_partition(
+            table_name,
+            source_queries,
+            columns_to_types,
+            where,
+            insert_overwrite_strategy_override=InsertOverwriteStrategy.INTO_IS_OVERWRITE,  # since we already cleared the data
+            **kwargs,
+        )
+
+    def _clear_partition_data(self, table: TableName, where: t.Optional[exp.Condition]) -> None:
```
tobymao, line 441: table -> table_name

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -373,6 +409,101 @@ def _find_matching_columns(
...
+            exp.to_identifier(f"{table.name}$partitions", quoted=True)
+        )
+
+        query = exp.select("*").from_(partition_table_name).where(where)
+        if limit:
+            query = query.limit(limit)
+
+        partition_values = [list(r) for r in self.fetchall(query, quote_identifiers=True)]
+
+        if len(partition_values) > 0:
```
tobymao, line 477: if partition_values

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -373,6 +409,101 @@ def _find_matching_columns(
...
+                TableName=table.name,
+                PartitionsToGet=[{"Values": v} for v in partition_values],
+            )
+            return sorted(
+                [(p["Values"], p["StorageDescriptor"]["Location"]) for p in response["Partitions"]]
+            )
+
+        return []
+
+    def _query_table_s3_location(self, table_name: exp.Table) -> str:
```
tobymao, line 489: table

sqlmesh/core/engine_adapter/athena.py
erindru: That's right, but Athena doesn't do transactions anyway.

I think we can say that if you want atomic deletes then you'll need to use Iceberg

sqlmesh/core/engine_adapter/athena.py
erindru: It's because the query its used in,  `SHOW TBLPROPERTIES {table}` is just a string and not an Expression.

since we are string building, we need to quote the identifiers manually. I couldn't figure out how to make it an Expression because sqlglot just turns it into an `exp.Command` string literal anyway

sqlmesh/core/engine_adapter/athena.py
erindru: `_truncate_table` needs this signature (or mypy will fail) because it's defined in the EngineAdapter superclass and im just overriding it

sqlmesh/core/engine_adapter/athena.py
erindru: > or what you should do is clean up so all internal methods take in exp.Table instead

Unfortunately a bunch are defined in the superclass which means they cant be easily cleaned up. I'll clean up the ones that only exist in this adapter class though

sqlmesh/core/engine_adapter/athena.py
tobymao: ok, we can remove this and on line 349 do sql(dialect='hive', identify=True)

erindru: Oh cool, forgot you could do that!

sqlmesh/core/engine_adapter/athena.py
erindru: Updated, yeah I kept switching between `TableName` and `exp.Table` because the parent classes use `TableName` but checking for strings all the time was getting tedious.

The general pattern throughout the codebase of making all the methods take a `str` or an `exp.Expression` is nice in some ways but annoying in others

tobymao (approved):

PR #3205, opened 2024-10-01
Feat: Snowflake: Handle forward_only changes to 'clustered_by'

erindru:
This change allows you to update `clustered_by` on a Snowflake model, trigger a forward-only plan and have the cluster key on the table be altered.

Previously, it would be a no-op which was a bit misleading

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -454,28 +454,29 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin, ClusteredByMixi
         )
 
         if current_table_info and target_table_info:
-            if any([current_table_info.is_clustered, target_table_info.is_clustered]):
-                if target_table_info.clustering_key:
-                    if current_table_info.clustering_key != target_table_info.clustering_key:
-                        # Note: If you create a table with eg `CLUSTER BY (c2, c1)` and read the info back from information_schema,
-                        # it gets returned as a string like "LINEAR(c2, c1)" which we need to parse back into a list of columns
-                        parsed_cluster_key = parse_one(
-                            target_table_info.clustering_key, dialect=self.dialect
-                        )
-                        additional_expressions.append(
-                            exp.Alter(
-                                this=current_table,
-                                kind="TABLE",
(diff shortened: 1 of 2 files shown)


georgesittas (approved):

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -422,3 +433,49 @@ def _create_column_comments(
                 f"Column comments for table '{table.alias_or_name}' not registered - this may be due to limited permissions.",
                 exc_info=True,
             )
+
+    def get_alter_expressions(
```
izeigerman, line 437: Is this something we can only do for snowflake?

erindru: For now, I didn't investigate the possibility or syntax on other databases

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -422,3 +433,49 @@ def _create_column_comments(
...
+                            target_table_info.clustering_key, dialect=self.dialect
+                        )
+                        additional_expressions.append(
+                            exp.Alter(
+                                this=current_table,
+                                kind="TABLE",
+                                actions=[exp.Cluster(expressions=parsed_cluster_key.expressions)],
+                            )
+                        )
+                else:
```
izeigerman, line 472: I feel like we can do `elif current_table_info.is_clustered:` here and get rid of the `any` block

izeigerman (approved):

sqlmesh/core/engine_adapter/snowflake.py
izeigerman: Can we please follow-up with the same support for BigQuery

sqlmesh/core/engine_adapter/snowflake.py
erindru: Fair call, i've updated it

sqlmesh/core/engine_adapter/snowflake.py
erindru: https://github.com/TobikoData/sqlmesh/pull/3231

PR #3211, opened 2024-10-02
Fix: avoid parsing column names into qualified columns in InsertOverwriteWithMergeMixin

georgesittas:
There was an issue with the mssql adapter where we'd get back a dotted column from its `columns` method, e.g. `foo.bar.baz`, and then we'd turn that into a qualified column instead of treating it as a single identifier with literal dot characters. This in turn caused problems in the context of the generated `MERGE` statement, because qualified columns are syntactically invalid within the `INSERT` clause.

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -89,7 +89,7 @@ class InsertOverwriteWithMergeMixin(EngineAdapter):
         for source_query in source_queries:
             with source_query as query:
                 query = self._order_projections_and_filter(query, columns_to_types, where=where)
-                columns = [exp.to_column(col) for col in columns_to_types]
+                columns = [exp.column(col) for col in columns_to_types]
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -89,7 +89,7 @@ def _insert_overwrite_by_condition(
         for source_query in source_queries:
             with source_query as query:
                 query = self._order_projections_and_filter(query, columns_to_types, where=where)
-                columns = [exp.to_column(col) for col in columns_to_types]
+                columns = [exp.column(exp.to_identifier(col)) for col in columns_to_types]
```
tobymao, line 92: why do you need the to_identifier?

georgesittas: we don't, I forgot to switch this thanks

tobymao (approved):

Issue #3070, opened 2024-08-29
Snowflake Secure View Support


isaiahgm:
[Secure views](https://docs.snowflake.com/en/user-guide/views-secure#overview-of-secure-views) are feature in Snowflake that allows engineers to obfuscate the view query code and must be used when sharing views across Snowflake Account using [Snowflake Data Sharing](https://docs.snowflake.com/en/user-guide/data-sharing-provider#general-data-sharing-considerations-and-usage).

Because SQL Mesh only surfaces data for end users as views, we are unable to share data created by SQL Mesh with other Snowflake Accounts, which limits SQL Mesh's potential use cases.

I propose adding this functionality either as another option in the model configuration or as a key pair in the virtual properties config option.

BenKentzer:
Hello!
+1 for this feature, but there is a slightly messy work around...
https://docs.snowflake.com/en/user-guide/data-sharing-views

Secure views as an option will e gratefully received here too!

benfdking:
This was released in the latest version and is documented below. 

https://sqlmesh.readthedocs.io/en/stable/integrations/engines/snowflake/#secure-views

PR #3212, opened 2024-10-02
feat: adding secure views for snowflake

benfdking:
Related to #3070 

NOTE I know I need to revert a number of unnecessary format changes. 

A few thoughts: 

- [ ] I don't think this is tested enough because I haven't run anything against Snowflake
- [ ] I am not certain about the way I create the views, it's a bit ugly
- [ ] More generally, I am slightly nervous about the approach of just adding properties to the root model configuration that are specific to individual engines, as it feels like a way to get way too many confusing configuration points


tobymao:
@benfdking is it possible to make this more generic? we've also had a request for transient tabels, i wonder if we can make it easy to specify any type of create table kind

benfdking:
I believe this approach is more versatile and can be applied to both `TABLE` and `VIEW` types. However, there is a conflict between the new `creatable_properties` and `materialize` properties in the case of Snowflake view. This is because, the properties control the same part of the query and according to the [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/sql/create-view), the `SECURE` property has to come first.

The issue can be easily resolved by checking the properties and moving `SECURE` to the beginning if it is present but I was wondering if we should use this as an opportunity to remove the `materialize` property?

tobymao:
@benfdking what would removing it entail?

PR #3215, opened 2024-10-03
Fix: Infer column types when data type is omitted in dbt seeds

Themiscodes:
This fix addresses an issue with dbt seeds where columns defined solely by a description, or where the data type was specified for only some columns, did not trigger type inference for the rest of the columns. It ensures that explicitly declared data types are set from the declaration while the types of the remaining columns are inferred.

### DIFF
--- a/sqlmesh/dbt/seed.py
+++ b/sqlmesh/dbt/seed.py
@@ -41,25 +41,33 @@ class SeedConfig(BaseModelConfig):
         seed_path = self.path.absolute().as_posix()
 
         if column_types := self.column_types:
-            columns = copy.deepcopy(self.columns)
+            column_types_override = copy.deepcopy(self.columns)
             for name, data_type in column_types.items():
-                column = columns.setdefault(name, ColumnConfig(name=name))
+                column = column_types_override.setdefault(name, ColumnConfig(name=name))
                 column.data_type = data_type
                 column.quote = self.quote_columns or column.quote
-                kwargs = self.sqlmesh_model_kwargs(context, columns)
+                kwargs = self.sqlmesh_model_kwargs(context, column_types_override)
         else:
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/seed.py
```diff
@@ -41,25 +41,33 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
                 column.quote = self.quote_columns or column.quote
-                kwargs = self.sqlmesh_model_kwargs(context, columns)
+                kwargs = self.sqlmesh_model_kwargs(context, column_types_override)
         else:
             kwargs = self.sqlmesh_model_kwargs(context)
 
-        if kwargs.get("columns") is None:
+        columns = kwargs.get("columns") or {}
+        missing_types = (
+            set((kwargs.get("column_descriptions") or {}).keys()) | set(self.columns.keys())
```
tobymao, line 55: you don't need to use keys()

sqlmesh/dbt/seed.py
```diff
@@ -41,25 +41,33 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
                 agate_helper.from_csv(seed_path, [], delimiter=self.delimiter)
                 if SUPPORTS_DELIMITER
                 else agate_helper.from_csv(seed_path, [])
             )
-            kwargs["columns"] = {
+            inferred_types = {
                 name: AGATE_TYPE_MAPPING[tpe.__class__]
                 for name, tpe in zip(agate_table.column_names, agate_table.column_types)
             }
+            columns.update(
```
tobymao, line 67: for key in missing_types or inferred_types:
  columns[key] = inferred_types[key]

Themiscodes:
Thanks for the review @tobymao addressed comments

tobymao (approved):

PR #3216, opened 2024-10-03
Fix: load custom materializations on run

z3z1ma:
No one can use custom materializations currently as far as I can surmise because of this. 
So perhaps I am kicking in the gears before others here. But we have a nifty custom materialization use cases that requires this fix.


z3z1ma:
Any chance this can make it into a `0.125.2`?

izeigerman (approved):

georgesittas:
> Any chance this can make it into a `0.125.2`?

Yeah, planning to deploy v0.125.2 soon.

PR #3218, opened 2024-10-03
Feat: Support SQLMesh project generation from dlt pipeline

Themiscodes:
WIP: This update introduces the ability to generate a SQLMesh project from a [dlt](https://github.com/dlt-hub/dlt) pipeline. It creates the project directory scaffolding, inspects the schema and automatically generates incremental and full models from the tables and sets up the `config.yaml` connection configurations using the pipeline's credentials.

**Importing a DLT project:**

To import a dlt project into SQLMesh, ensure the dlt pipeline has been run or restored locally. Then, in the pipeline directory, use the `init` command specifying its name:

```bash
sqlmesh init -t dlt --dlt-pipeline "pipeline_name" dialect
```

The resulting SQLMesh project can be executed as usual with `sqlmesh plan`. The incremental models are designed to process data using the unique `_dlt_load_id` key present in each dlt table, but the model definitions can be modified as needed.


sqlmesh/cli/example_project.py
```diff
@@ -216,8 +245,8 @@ def _create_audits(audits_path: Path) -> None:
     _write_file(audits_path / "assert_positive_order_ids.sql", EXAMPLE_AUDIT)
 
 
-def _create_models(models_path: Path) -> None:
-    for model_name, model_def in [
+def _create_models(models_path: Path, models: t.Optional[t.List[t.Tuple[str, str]]] = None) -> None:
```
tobymao, line 248: models should be a set if you're going to do an in statement

sqlmesh/integrations/dlt.py
```diff
@@ -0,0 +1,168 @@
...
+        for name, table in schema.tables.items()
+        if (has_table_seen_data(table) and not name.startswith(schema._dlt_tables_prefix))
+        or name == schema.loads_table_name
+    }
+
+    sqlmesh_models = []
+    SQLMESH_SCHEMA_NAME = f"{dataset}_sqlmesh"
+    for table_name, table in dlt_tables.items():
+        dlt_columns = {
+            c["name"]: parse_one(str(c["data_type"]), into=exp.DataType, dialect=dialect)
```
tobymao, line 50: DataType.build(...)

sqlmesh/integrations/dlt.py
```diff
@@ -0,0 +1,168 @@
...
+
+    dlt_tables = {
+        name: table
+        for name, table in schema.tables.items()
+        if (has_table_seen_data(table) and not name.startswith(schema._dlt_tables_prefix))
+        or name == schema.loads_table_name
+    }
+
+    sqlmesh_models = []
+    SQLMESH_SCHEMA_NAME = f"{dataset}_sqlmesh"
```
tobymao, line 47: don't use constants in methods

tobymao:
can we get an example dlt project and have sqlmesh run on it? we'll also need to update the docs

sqlmesh/integrations/dlt.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+
+    import dlt
+    from dlt.common.schema.utils import has_table_seen_data, is_complete_column
+    from dlt.pipeline.exceptions import CannotRestorePipelineException
+
+    try:
+        pipeline = dlt.attach(pipeline_name=pipeline_name)
+    except CannotRestorePipelineException:
+        raise click.ClickException("Could not attach to pipeline {pipeline_name}")
```
izeigerman, line 22: Missing `f""`. The string won't be interpolated

sqlmesh/integrations/dlt.py
```diff
@@ -0,0 +1,169 @@
...
+def generate_incremental_model(
+    model_name: str, model_def_columns: str, select_columns: str, grain: str, from_table: str
+) -> str:
+    """Generate the SQL definition for an incremental model."""
+
+    key = "load_id" if from_table.endswith("_dlt_loads") else "_dlt_id"
+
+    return f"""MODEL (
+  name {model_name},
+  kind INCREMENTAL_BY_UNIQUE_KEY (
```
izeigerman, line 92: Is it always unique by key?

sqlmesh/integrations/dlt.py
```diff
@@ -0,0 +1,169 @@
...
+            for c in filter(is_complete_column, table["columns"].values())
+            if c.get("primary_key") and c.get("name")
+        ]
+        model_def_columns = format_columns(dlt_columns, dialect)
+        select_columns = format_columns_for_select(dlt_columns)
+        grain = format_grain(primary_key)
+        incremental_model_name = f"{sqlmesh_schema_name}.incremental_{table_name}"
+        full_model_name = f"{sqlmesh_schema_name}.full_{table_name}"
+
+        incremental_model_sql = generate_incremental_model(
```
izeigerman, line 65: Why do we produce both incremental and full models at the same time?

sqlmesh/integrations/dlt.py
Themiscodes: Thanks my bad missed this

sqlmesh/integrations/dlt.py
Themiscodes: We don't have to. My thinking was to create a project similar to the example project and from each dlt table to create an equivalent incremental and full model. So from the full model the user can remove the columns they don't need or remove them according to their use case

But we can simply generate one type of model that loads from the dlt tables and leave it to the user if they want then to have other models to source from them

sqlmesh/integrations/dlt.py
Themiscodes: I thought this was the simpler way since each dlt table has a unique load id column and use this as its key. And if someone wanted to implement some other kind of table they could adjust these later. But I can change it if you feel another type should serve better as a default generated table

sqlmesh/integrations/dlt.py
tobymao: we should just generate one, what does their dbt generator do?

Themiscodes: the dbt generator creates staging and dimensional models under corresponding staging and marts directories for each dlt table. but I suppose yes we don't have to follow that logic, so I can modify the code and simply generate one type of model. and in that case should I keep the incremental kind to source from the dlt table?

tobymao: yes, i think it should be incremental, but what does it mean creates staging and dimension, and is that analogous to you do both incremental and full?

Themiscodes: no the staging and dimension in this case is for dbt to do the incremental processing but it isn't analogous. I simply followed the example project  that we have with both incremental and full kinds, so yes I can remove the full models

docs/integrations/dlt.md
```diff
@@ -0,0 +1,87 @@
...
+
+SQLMesh enables efforless project generation using data ingested through [dlt](https://github.com/dlt-hub/dlt). This involves creating a baseline project scaffolding, generating incremental models to process the data from the pipeline's tables by inspecting its schema and configuring the gateway connection using the pipeline's credentials.
+
+## Getting started
+### Reading from a dlt pipeline
+
+To load data from a dlt pipeline into SQLMesh, ensure the dlt pipeline has been run or restored locally. Then simply execute the sqlmesh `init` command *within the dlt project root directory* using the `dlt` template option and specifying the pipeline's name with the `dlt-pipeline` option:
+
+```bash
+$ sqlmesh init -t dlt --dlt-pipeline <pipeline-name> dialect
```
tobymao, line 11: i think you'll probably also want to define a start date or no?

sqlmesh/integrations/dlt.py
tobymao: shouldn't you filter by dates here? otherwise you're always doing a full scan on from_table

docs/integrations/dlt.md
izeigerman: It is generated automatically

sqlmesh/integrations/dlt.py
Themiscodes: Altered the model kinds to `INCREMENTAL_BY_TIME_RANGE` by using the load time (converting the load ids which are unix timestamps) as time_column

docs/integrations/dlt.md
Themiscodes: Yes also added a function to extract the start date from the pipeline directly to be set in the sqlmesh `config.yaml`

setup.py
```diff
@@ -105,6 +106,9 @@
         "dbt": [
             "dbt-core<2",
         ],
+        "dlt": [
+            "dlt>=1.1.0",
```
georgesittas, line 110: Why do we need to constrain this?

setup.py
```diff
@@ -67,6 +67,7 @@
             "dbt-duckdb>=1.7.1",
             "dbt-snowflake",
             "dbt-bigquery",
+            "dlt",
```
izeigerman, line 70: I don't think we want this in 2 places. Let's just add the `dlt` target in Makefile where we need it.

izeigerman (approved):
One small comment, LGTM otherwise

setup.py
Themiscodes: Sure, removed it from the `dev` requirements and added it the makefile in `install-dev` and `install-cicd-test` since `test_cli` is the test that requires it

PR #3222, opened 2024-10-03
Feat: Support DBT Athena adapter

erindru:
Initial support for [dbt-athena](https://github.com/dbt-labs/dbt-athena)

ref issue #1315

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ setup(
             "pytz",
             "snowflake-connector-python[pandas,secure-local-storage]>=3.0.2",
             "sqlalchemy-stubs",
-            "tenacity==8.1.0",
+            "tenacity",
             "types-croniter",
             "types-dateparser",
             "types-python-dateutil",
@@ -100,6 +100,7 @@ setup(
             "dbt-redshift",
             "dbt-sqlserver>=1.7.0",
             "dbt-trino",
+            "dbt-athena-community",
         ],
         "dbt": [
             "dbt-core<2",
--- a/sqlmesh/core/engine_adapter/athena.py
+++ b/sqlmesh/core/engine_adapter/athena.py
@@ -22,6 +22,9 @@ from sqlmesh.core.engine_adapter.shared import (
 
 if t.TYPE_CHECKING:
     from sqlmesh.core._typing import SchemaName, TableName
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -86,7 +86,7 @@
             "pytz",
             "snowflake-connector-python[pandas,secure-local-storage]>=3.0.2",
             "sqlalchemy-stubs",
-            "tenacity==8.1.0",
+            "tenacity",
```
erindru, line 89: Note: this was pinned in [this commit](https://github.com/TobikoData/sqlmesh/commit/0ab79bbb3e87d4827391a404aa90f7fe27472332) back when we were using Airflow 2.3.3.

Pinning it was causing an ancient `dbt-athena-community` to be resolved. Since we've moved on to newer Airflow in the meantime, i've removed the pin

tobymao (approved):

sqlmesh/core/engine_adapter/athena.py
```diff
@@ -90,14 +97,7 @@ def _get_data_objects(
...
             exp.select(
-                exp.case()
-                .when(
-                    # calling code expects data objects in the default catalog to have their catalog set to None
-                    exp.column("table_catalog", table="t").eq("awsdatacatalog"),
-                    exp.Null(),
-                )
-                .else_(exp.column("table_catalog"))
-                .as_("catalog"),
+                exp.column("table_catalog").as_("catalog"),
```
izeigerman, line 100: Why do we no longer need this?

erindru: I hit [this problem](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/dbt/loader.py#L170) which triggered me to analyse catalog support some more.

Athena has the concept of catalogs, just not setting them or querying the current one after a connection is established. So I changed `CATALOG_SUPPORT` to `CatalogSupport.SINGLE_CATALOG_ONLY` and implemented `get_current_catalog()` in terms of reading the catalog off the PyAthena connection.

Since the connection catalog defaults to `awsdatacatalog` when it's not set, now `get_default_catalog()` returns eg `awsdatacatalog` instead of `None` so I needed to change `get_data_objects()` to return `awsdatacatalog` instead of `None` so that code checking if the "current" catalog = the default catalog would continue to work

PR #3224, opened 2024-10-04
Feat: DuckDB extension config to support community extensions

rubenvereecken:
This now supports alternative DuckDB extension repositories like the built-in `community`.

```yaml
gateways:
  bq:
    connection:
      type: duckdb
      extensions:
        - name: bigquery
          repository: community
          force_install: true
      catalogs:
        mem: ':memory:'
        bq:
          type: bigquery
          path: 'project=project'
          read_only: true
```

This generates
```
FORCE INSTALL 'bigquery' FROM community;
```

I think this event supports local repositories, but they might need double quotes or something in the yaml config, so you can get to something like (note the quotes!)
```
INSTALL 'something-local' FROM 'https://some-other-repo';
```

Basically, blame @Kayrnt for setting me on this path.


Kayrnt:
Well it looks like a mandatory feature 😏 

For the reviewers:
- [the documentation about force install](https://duckdb.org/docs/extensions/working_with_extensions.html#force-installing-to-upgrade-extensions)
- [the documentation regarding community extension installation](https://community-extensions.duckdb.org/index.html#how-to-use-a-community-extension)

Kayrnt:
Also pinging @hafenkran for visibility (probably something I'll try myself too when I find some free time!)

sqlmesh/core/config/connection.py
```diff
@@ -174,11 +174,21 @@ def _cursor_init(self) -> t.Optional[t.Callable[[t.Any], None]]:
...
             for extension in self.extensions:
+                extension = extension if isinstance(extension, dict) else {"name": extension}
+
+                install_command = f"INSTALL {extension['name']}"
+
+                if extension.get("repository"):
+                    install_command = f"{install_command} FROM {extension['repository']}"
+
+                if extension.get("force_install"):
+                    install_command = f"FORCE {install_command}"
```
izeigerman, line 185: Shouldn't there be space before `FORCE`?

sqlmesh/core/config/connection.py
```diff
@@ -149,7 +149,7 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
         pre_ping: Whether or not to pre-ping the connection before starting a new transaction to ensure it is still alive.
     """
 
-    extensions: t.List[str] = []
+    extensions: t.List[str | t.Dict[str, t.Any]] = []
```
izeigerman, line 152: Can you please use `t.Union` instead to support older Python versions

izeigerman:
This looks great, thank you! Any chance you can also update docs: https://github.com/TobikoData/sqlmesh/blob/main/docs/integrations/engines/duckdb.md and include an example there?

Kayrnt:
@izeigerman one problem that was identified is, apparently, [the way SQLMesh tries to access `INFORMATION_SCHEMA`](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1728054728741209?thread_ts=1727966914.081469&cid=C044BRE5W4S) as BigQuery is mounted but AFAIK the DuckDB storage API doesn't support it yet. Do you have some ideas on how to workaround it? (using SQL only and not a Python model)
Else this PR is still useful for other kind of extensions (e.g. functions like advanced pivot) and maybe to some materialization too (that wouldn't need `INFORMATION_SCHEMA`).

izeigerman:
@Kayrnt, no, unfortunately, querying `INFORMATION_SCHEMA` is a big part of the existing evaluation flow. I agree that this change is still useful.

hafenkran:
hey @rubenvereecken @Kayrnt - thanks for the ping, and apologies for the delayed response.

I think the idea is pretty cool, and theoretically, it should work (with restrictions). I’ve spent some time looking into it and managed to get initial interaction of it running. However, there were a few bugs I had to fix. For instance, sqlmesh uses the USE statement, which wasn’t compatible with the BigQuery extension. I also had to set a default dataset because `main` doesn’t exist by default. All these changes are now available in the Community Extension repository since this morning and can be tried out.

The `INFORMATION_SCHEMA` actually doesn't make any problems here and integrates well with the BigQuery extension, which made this approach plausible to me. However, I ran into issues during the initial setup of the [migrations](https://github.com/TobikoData/sqlmesh/tree/main/sqlmesh/migrations). Theoretically, they work, but there’s a practical limitation with BigQuery: it struggles when renaming tables that have data recently inserted into them, because BigQuery uses a streaming buffer (see [here](https://cloud.google.com/knowledge/kb/cannot-update-or-delete-over-bigquery-streaming-tables-000004334)). Renaming isn’t allowed until the buffer is flushed, which can sometimes take up to 90 minutes. This is what the migrations try. This, unfortunately, breaks the process.

However, the value of the PR remains pretty valuable, but given the constraints mentioned, the use case of using BigQuery is currently impractical (at least with the current chain of migrations). 

Note that this is just to the best of my knowledge and current research. It's also possible that I missed something or that I don't see the whole picture. I didn't spend too much time on it.

github-christophe-oudar:
@hafenkran Thanks for deep diving into that topic 🙌 

I don't know well the workflow from SQLMesh but maybe @izeigerman knows if there's a difference between DuckDB & BigQuery in the way that it is designed so that it would work with BQ direct integration vs DuckDB. However it's indeed pretty promising because then maybe using a custom materialization (similar to the BQ one?), it would work!

Combining DuckDB & BigQuery looks great on the paper!

izeigerman:
@hafenkran 

> However, I ran into issues during the initial setup of the [migrations](https://github.com/TobikoData/sqlmesh/tree/main/sqlmesh/migrations).

Migrations are only relevant in the context of SQLMesh state and I believe nobody should ever use the BigQuery extension to store the SQLMesh state :)  So I think this shouldn't be a big deal. Normally, sqlmesh doesn't rename any tables outside migrations. 

If this is the only blocker I think we can disregard it and just mention somewhere in docs not to use the extension to store the state.

> knows if there's a difference between DuckDB & BigQuery in the way that it is designed so that it would work with BQ direct integration vs DuckDB

@github-christophe-oudar , unfortunately, this can only be definitively determined through comprehensive testing.

Just to reiterate, I'm on board with merging this as soon as the PR is green.

sqlmesh/core/config/connection.py
tobymao, line 152:
```suggestion
    extensions: t.List[t.Union[str, t.Dict[str, t.Any]]] = []
```

PR #3226, opened 2024-10-04
Feat(clickhouse): use table/partition swap for insert-overwrite

treysp:
The current Clickhouse adapter uses the delete-insert loading approach. Clickhouse is not designed for deletes/updates, so its performance may be unusably slow.

This PR replaces delete-insert with a table/partition swap approach. At a high level, it operates by creating a temp table that contains new records to insert, adding existing records that should be retained to the temp table, then swapping the temp table for the current table.

To improve performance, the swapping occurs at the partition level for partitioned tables (potentially processing much less data). Incremental by time models now automatically partition the underlying table by calendar week (`toMonday(timecol)`).

Implementation note:
- Almost all of the implementation is in the `_insert_overwrite_by_condition()` adapter method, which is called by other methods like `replace_by_unique_key()`.

### DIFF
--- a/sqlmesh/core/engine_adapter/clickhouse.py
+++ b/sqlmesh/core/engine_adapter/clickhouse.py
@@ -296,7 +296,7 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
                 partitions_to_replace = self._get_partition_ids(temp_table)
 
                 # drop affected partitions that have no records in temp_table
-                #   - NOTE: `all_affected_partitions` will be empty when is_inc_by_partition=True
+                #   - NOTE: `all_affected_partitions` will be empty when keep_existing_partition_rows=False
                 #      because previous code block is skipped
                 partitions_to_drop = all_affected_partitions - partitions_to_replace
 
@@ -347,7 +347,7 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
         )
 
         # limit existing records insert expression WHERE to affected target table partitions
-        #   by adding `AND _partition_id IN (affected partition IDs)`
+        #   by adding `AND _partition_id IN (SELECT partition_id FROM partitions_temp_table)`
         existing_records_insert_exp.set(
             "expression",
             existing_records_insert_exp.expression.where(
--- a/sqlmesh/core/engine_adapter/clickhouse.py
+++ b/sqlmesh/core/engine_adapter/clickhouse.py
@@ -202,7 +202,7 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
                 dynamic_key_exp: Expression to build key (replace by key only)
                 dynamic_key_unique: Whether more than one record can exist per key value (replace by key only)
 
-                is_inc_by_partition: Boolean of whether the call is only overwriting full partitions (incremental by partition only)
+                is_inc_by_partition: Whether to overwrite partitions with only new records (incremental by partition only)
 
         Returns:
             Side effects only: execution of insert-overwrite operation.
@@ -223,7 +223,7 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
             # insert new records into temp table
             for source_query in source_queries:
                 with source_query as query:
-                    # REPLACE BY KEY: if unique key, DISTINCTify by key value so only one row is present per key
+                    # REPLACE BY KEY: if unique key, DISTINCTify by key columns so only one row is present per key
                     if dynamic_key and dynamic_key_unique:
                         query = query.distinct(*dynamic_key)  # type: ignore
 
@@ -235,7 +235,7 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
                         order_projections=False,
                     )
 
-            # REPLACE BY KEY: build `where` as "key IN (new rows' key values)"
+            # REPLACE BY KEY: build `where` expression as "key IN (new rows' key values)"
             if dynamic_key:
                 key_query = exp.select(dynamic_key_exp).from_(temp_table)
                 if not dynamic_key_unique:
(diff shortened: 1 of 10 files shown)


sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -174,6 +175,198 @@ def create_schema(
...
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        where: t.Optional[exp.Condition] = None,
+        insert_overwrite_strategy_override: t.Optional[InsertOverwriteStrategy] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        target_table = exp.to_table(table_name)
+        columns_to_types = columns_to_types or self.columns(target_table)
+
+        temp_table = self._get_temp_table(target_table)
+        self._create_table_like(temp_table, target_table)
```
tobymao, line 191: you're doing a full copy?

treysp: No, this creates an empty table with the same structure (columns, data types, partitioning, etc.) as target

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -174,6 +175,289 @@ def create_schema(
...
+            source_queries: Source queries returning records to insert
+            columns_to_types: Column names and data types of target table
+            where: SQLGlot expression determining which target table rows should be overwritten
+            insert_overwrite_strategy_override: Not used by Clickhouse
+            kwargs:
+                dynamic_key: Key columns (replace by key only)
+                dynamic_key_exp: Expression to build key (replace by key only)
+                dynamic_key_unique: Whether more than one record can exist per key value (replace by key only)
+
+                is_inc_by_partition: Whether to overwrite partitions with only new records (incremental by partition only)
```
erindru, line 205: Would a better name for this be a boolean `replace` ?

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -174,6 +175,289 @@ def create_schema(
...
+
+            # REPLACE BY KEY: build `where` expression as "key IN (new rows' key values)"
+            if dynamic_key:
+                key_query = exp.select(dynamic_key_exp).from_(temp_table)
+                if not dynamic_key_unique:
+                    key_query = key_query.distinct()
+                where = dynamic_key_exp.isin(query=key_query)
+
+            # get target table partition key to confirm it's actually partitioned
+            table_partition_exp = self.fetchdf(
```
erindru, line 246: Is this more like a `fetchone()` ? Since you just grab `[0]` and expect it to exist

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -174,6 +175,289 @@ def create_schema(
...
+        temp_table: exp.Table,
+        partitions_to_replace: t.Set[str],
+        partitions_to_drop: t.Set[str],
+    ) -> exp.Alter:
+        alter_expr = exp.Alter(this=target_table, kind="TABLE")
+
+        for partition in partitions_to_replace:
+            alter_expr.append(
+                "actions",
+                exp.ReplacePartition(
```
erindru, line 363: man we really need to work on SQLGlot's builder API's

sqlmesh/core/engine_adapter/clickhouse.py
```diff
@@ -174,6 +175,289 @@ def create_schema(
...
+        where: t.Optional[exp.Condition] = None,
+        limit: t.Optional[int] = None,
+    ) -> t.Set[t.Any]:
+        """List partition IDs present in table"""
+        partitions_query = exp.select(partition_col_name).distinct().from_(table)
+        if where:
+            partitions_query = partitions_query.where(where)
+        if limit:
+            partitions_query = partitions_query.limit(limit)
+        partitions_df = self.fetchdf(partitions_query)
```
erindru, line 457: I noticed that the engine adapter exposes `fetchall()` as well so we can avoid DataFrame's and the associated type mangling if all we are doing is converting resultsets to python sets/lists

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -750,7 +750,7 @@ def test_insert_overwrite_by_time_partition(ctx: TestContext):
     results = ctx.get_metadata_results()
     assert len(results.views) == 0
     assert len(results.materialized_views) == 0
-    assert len(results.tables) == len(results.non_temp_tables) == 1
+    assert len(results.tables) == 2 if ctx.dialect == "clickhouse" else 1
```
erindru, line 753: What's the extra table produced by Clickhouse that the other adapters dont have?

tests/core/engine_adapter/integration/test_integration_clickhouse.py
```diff
@@ -0,0 +1,522 @@
...
+
+@pytest.fixture(
+    params=[
+        pytest.param(
+            "clickhouse",
+            marks=[
+                pytest.mark.docker,
+                pytest.mark.engine,
+                pytest.mark.clickhouse,
+                pytest.mark.xdist_group("engine_integration_clickhouse"),
```
erindru, line 16: As of [e8576e9a](https://github.com/TobikoData/sqlmesh/commit/e8576e9aab71d3237a80b360d8ae1ed417849540) the `xdist_group`'s arent required anymore

tests/core/engine_adapter/integration/test_integration_clickhouse.py
```diff
@@ -0,0 +1,522 @@
...
+        existing_table_name.sql(),
+        ctx.input_data(existing_data, columns_to_types),
+        columns_to_types,
+        partitioned_by=partitioned_by,
+    )
+    return existing_table_name
+
+
+def test_insert_overwrite_by_condition_replace_partitioned(ctx: TestContext):
+    ctx.init()
```
erindru, line 95: `ctx.init()` can also be removed, it happens automatically in the fixture function now

tests/core/engine_adapter/integration/test_integration_clickhouse.py
```diff
@@ -0,0 +1,522 @@
...
+    ctx.compare_with_current(
+        existing_table_name,
+        pd.DataFrame(
+            [
+                {"id": 5, "ds": pd.Timestamp("2024-02-29")},
+                {"id": 6, "ds": pd.Timestamp("2024-04-01")},
+            ]
+        ),
+    )
+    ctx.engine_adapter.drop_table(existing_table_name.sql())
```
erindru, line 130: As of very recently, any table created with `ctx.table(...)` will automatically be cleaned up. `TestContext` keeps track of them and its `cleanup()` is invoked on fixture teardown

tests/core/engine_adapter/integration/test_integration_clickhouse.py
erindru, line 1: Nice work!

tests/core/engine_adapter/integration/test_integration_clickhouse.py
treysp: Nice - i see some additional ones in the dbt tests, should I put up a PR that just removes those?

tests/core/engine_adapter/integration/test_integration_clickhouse.py
treysp: Thanks!

sqlmesh/core/engine_adapter/clickhouse.py
erindru: Ahh I see what you're saying. To me, "replace" is more about the data (vs eg "merge" or "append").

Technically if you replace a partition with an exact copy of itself, it would be "replaced" but nothing has really changed, so i'd consider it a no-op rather than a replace.

Anyway, I think `keep_existing_partition_rows` is still a more descriptive name than `is_inc_by_partition`

tests/core/engine_adapter/integration/test_integration_clickhouse.py
erindru: Oh, good point. I was focusing on the integration tests in `test_inetgration.py` when I removed the `xdist_group`'s. 

Probably can leave the DBT ones for now, there isn't a speed issue with those tests

erindru (approved):

PR #3228, opened 2024-10-07
Adding markdown feature to model description

Harmuth94:
This feature adds markdown capabilities to the description field enabling among other things:
* Multiline
* Code blocks and code (triple and single backtick enclosing)
* Bolding

Markdown features that are not currently working with this:
* Headers
* Ordered and unordered lists
* Italicizing

### DIFF
--- a/web/client/package-lock.json
+++ b/web/client/package-lock.json
@@ -35,6 +35,7 @@
         "react-dnd": "^16.0.1",
         "react-dnd-html5-backend": "^16.0.1",
         "react-dom": "^18.2.0",
+        "react-markdown": "^9.0.1",
         "react-router-dom": "^6.15.0",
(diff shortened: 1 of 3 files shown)


mykalmax:
Hey @Harmuth94 thanks for your help, could you please provide more details on what use case markdown support is solving other then respecting multiline formatting? Having an example scenario would be great.

Harmuth94:
The primary use case is codeblocks. 

For example if I wanted to document in the description how to use the table/model I would use

```sql
/* My advanced model
This model is advanced and here’s an example query of how to join in

```sql
select a.id
from my.model
join b
on b.id=a.id
` ``
… model definition
```

mykalmax (approved):

PR #3231, opened 2024-10-08
Feat: BigQuery - Handle forward_only changes to clustered_by

erindru:
Follow-up PR from https://github.com/TobikoData/sqlmesh/pull/3205

Add the ability to change `clustered_by` in BigQuery tables in a forward_only fashion. Previously, we would silently succeed

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -5,7 +5,7 @@ import typing as t
 from collections import defaultdict
 
 import pandas as pd
-from sqlglot import exp
+from sqlglot import exp, parse_one
 from sqlglot.transforms import remove_precision_parameterized_types
 
 from sqlmesh.core.dialect import to_schema
@@ -39,6 +39,9 @@ logger = logging.getLogger(__name__)
 
 NestedField = t.Tuple[str, str, t.List[str]]
 NestedFieldsDict = t.Dict[str, t.List[NestedField]]
+
+# used to tag AST nodes to be specially handled in alter_table()
+_CLUSTERING_META_KEY = "__sqlmesh_update_table_clustering"
 
 
 @set_catalog()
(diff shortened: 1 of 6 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -863,30 +864,42 @@ def _drop_object(
         """
         self.execute(exp.Drop(this=exp.to_table(name), kind=kind, exists=exists, **drop_args))
 
-    def get_alter_expressions(
+    def get_alter_operations(
         self,
         current_table_name: TableName,
         target_table_name: TableName,
-    ) -> t.List[exp.Alter]:
+    ) -> AlterTableOperations:
```
erindru, line 871: Since the alter operations for BigQuery cannot be expressed in SQL, I piggybacked `SourceQuery` to allow arbitrary side-effects like making API calls

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -886,10 +927,41 @@ def _get_data_objects(
...
+            rendered_columns = [c.sql(dialect=self.dialect) for c in cluster_by]
+            bq_table.clustering_fields = rendered_columns
+
+            self._db_call(self.client.update_table, table=bq_table, fields=["clustering_fields"])
+
+            # BigQuery only applies new clustering going forward, so this rewrites the columns to apply the new clustering to historical data
+            # ref: https://cloud.google.com/bigquery/docs/creating-clustered-tables#modifying-cluster-spec
+            self.execute(exp.update(table, {c: c for c in cluster_by}, where=exp.true()))
+
+            return exp.select("1")  # dummy query to satisfy the SourceQuery interface
```
erindru, line 950: A downside of using `SourceQuery` is that the interface requires a `Query` (SELECT) to be returned.

I'm wondering if a new `SourceCommand` interface should be introduced instead (and `SourceQuery` changed to extend `SourceCommand`)

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -337,3 +338,53 @@ def _build_clustered_by_exp(
         **kwargs: t.Any,
     ) -> t.Optional[exp.Cluster]:
         return exp.Cluster(expressions=[exp.column(col) for col in clustered_by])
+
+    def _parse_clustering_key(self, clustering_key: t.Optional[str]) -> t.List[exp.Expression]:
```
erindru, line 342: I moved the cluster key changing logic out of the Snowflake adapter and into `ClusteredByMixin` so the BigQuery adapter could utilise it

izeigerman (approved):
Great!

PR #3236, opened 2024-10-09
Chore: Reintroduce parallelism in integration tests

erindru:
The goal of this PR is to ensure that we are running tests efficiently to reduce the impact of adding more. 

Running tests in parallel also helps expose race conditions and encourages writing tests that are self-contained.

- Remove the xdist_group's that were causing the engine adapter integration tests to be pinned to a single worker. These are no longer needed now we use CircleCI to run each engine in its own workflow
- Remove `concurrent_tasks=1` from the yaml config. This means that the connection pools are created in multithreaded mode by default and `concurrent_tasks=1` is only set for the tests that need it (essentially any test that uses a SQLMesh context to plan/apply and has duckdb configured for the state connection)
- Add the ability to configure which schema the example project and sushi runs against. This allows parallel runs on cloud engines where we cant create a unique database per test run (BigQuery, Athena)

I also investigated the long-standing issues with Trino's metadata caching which causes the tests to be flaky. The tl;dr is that:
  - Only the Hive connector is affected, the Iceberg / Delta connectors have no issues
  - The caching options set in the Trino catalog config are the same as the defaults so I removed them
  - I spent a bunch of time trying to narrow the correctness issues down to Trino or the Metastore. After upgrading both to the latest available versions, I concluded that the issue is in Trino and metadata updates don't seem to immediately propagate between its threads internally
  - I was able to gain a decent reliability improvement by, when a table is created, immediately blocking until `table_exists()` returns True. This allowed comment registration to not silently fail and improved the reliability of `test_ctas()` in particular


erindru:
The main speedups are [before](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/15791/workflows/c89e4018-01d2-4ada-86d7-850b87ed3ece) -> [after](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/15795/workflows/166e2bd2-4b15-48cb-91da-98aa4e92997e):

| test | before | after |
|-------|-----------|---------|
| cloud_engine_snowflake | 13m29s | 6m14s |
| cloud_engine_databricks | 25m44s | 9m33s |
| cloud_engine_bigquery | 16m28s | 8m0s |
| cloud_engine_athena | 39m14s | 14m29s |

sqlmesh/core/engine_adapter/trino.py
```diff
@@ -271,3 +272,44 @@ def _to_delta_ts(
...
+        )
+
+        # extract the table name
+        if isinstance(table_name_or_schema, exp.Schema):
+            table_name = table_name_or_schema.this
+            assert isinstance(table_name, exp.Table)
+        else:
+            table_name = table_name_or_schema
+
+        if self.current_catalog_type == "hive":
```
erindru, line 311: Doing this relatively simple change reduced a bunch of flakiness

tests/core/engine_adapter/integration/__init__.py
```diff
@@ -379,7 +385,7 @@ def get_column_comments(
                     AND pgc.relkind = '{'v' if table_kind == "VIEW" else 'r'}'
                 ;
             """
-        elif self.dialect in ["mysql", "snowflake"]:
+        elif self.dialect in ["mysql", "snowflake", "trino"]:
```
erindru, line 388: Trino column comments do appear in `information_schema` now

tests/core/engine_adapter/integration/docker/trino/catalog/datalake.properties
```diff
@@ -1,11 +1,13 @@
...
-hive.s3.path-style-access=true
-hive.s3.aws-access-key=minio
-hive.s3.aws-secret-key=minio123
-hive.metastore-cache-ttl=0s
-hive.metastore-refresh-interval=5s
 hive.metastore.thrift.client.connect-timeout=10s
 hive.metastore.thrift.client.read-timeout=10s
 hive.storage-format=PARQUET
+
+fs.native-s3.enabled=true
```
erindru, line 8: Trino as of [Release 458](https://trino.io/docs/current/release/release-458.html) disabled `hive.s3` ("legacy s3") and requires the use of "native s3".

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -1328,31 +1306,39 @@ def test_sushi(ctx: TestContext):
...
 
-    context = Context(paths="./examples/sushi", config=config, gateway=ctx.gateway)
+    # Copy sushi example to tmpdir
+    shutil.copytree(pathlib.Path("./examples/sushi"), tmp_path, dirs_exist_ok=True)
+
+    # Rewrite schema references to test schema references
+    # Note that we deliberately do it at the filesystem level instead of messing with the Context to ensure
+    # that we are testing an actual Context rather than a doctored one
+    extensions = ["*.sql", "*.yaml", "*.py"]
+    replacements = {
```
erindru, line 1327: I'm open to a better way of making the `sushi` schema dynamic instead of using text-based find/replace.

I didn't want to modify the example to use vars in all the model names just for these tests because I think there is value in keeping our examples reasonably simple

eakmanrq: Yeah this is annoying because of our decision to use the examples as our test fixtures. To be fair we have already made decisions in the past that adds complexity to sushi just for test sake but I get how you could feel this would cross a line.

eakmanrq:
> Remove the xdist_group's that were causing the engine adapter integration tests to be pinned to a single worker. These are no longer needed now we use CircleCI to run each engine in its own workflow

What would happen if I was running all the tests locally and not within CircleCI? Or if we updated CircleCI to run all the engine adapter tests together? It should be fine right?

erindru:
> What would happen if I was running all the tests locally and not within CircleCI? Or if we updated CircleCI to run all the engine adapter tests together? It should be fine right?

Yep, this still works as expected with `make engine-up` followed by `make engine-docker-test`.

Cloud engines are `make engine-remote-test` because they require specific environment variables to be set with the credentials. The environment variables to set are indicated by the `guard-*` clauses in the `Makefile` and also in the [config.yaml](https://github.com/TobikoData/sqlmesh/blob/main/tests/core/engine_adapter/integration/config.yaml#L105)

tests/core/engine_adapter/integration/test_integration.py
erindru: Well if we don't mind adding more complexity to Sushi I could use it to demonstrate (and test I guess) dynamic model names.

For example, configuring a global variable `env: prod`, changing the model names to eg `sushi_@{env}.customer_total_revenue` and then in tests setting `env` to the test_id.

Do you think that's a better solution?

.circleci/continue_config.yml
```diff
@@ -341,10 +341,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
```
erindru, line 344: Note to self: dont forget to uncomment this

tests/core/engine_adapter/integration/test_integration.py
eakmanrq: I agree with your original assessment that readability/simplicity of the example is better than making this more robust. Therefore I am good with what you have. I was pointing out though that in other cases we have decided to add complexity to sushi so we are inconsistent.

eakmanrq (approved):

PR #3240, opened 2024-10-10
Fix: use parentheses for databricks' CLUSTER BY clause

georgesittas:
Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1728573925367219?thread_ts=1728490239.247289&cid=C044BRE5W4S

### DIFF
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -280,7 +280,10 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
             table_kind=table_kind,
         )
         if clustered_by:
-            clustered_by_exp = exp.Cluster(expressions=[exp.column(col) for col in clustered_by])
+            cluster_key = exp.maybe_parse(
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -280,7 +280,10 @@ def _build_table_properties_exp(
             table_kind=table_kind,
         )
         if clustered_by:
-            clustered_by_exp = exp.Cluster(expressions=[exp.column(col) for col in clustered_by])
+            cluster_key = exp.maybe_parse(
+                f"({','.join(clustered_by)})", into=exp.Ordered, dialect="databricks"
```
erindru, line 284: nit: you could use `dialect=self.dialect` here instead of hardcoding `"databricks"`

georgesittas: yeah thanks for calling it out, it was just me being lazy :P

erindru (approved):

PR #3247, opened 2024-10-12
Feat: Add Singlestore Engine Adapter

nicosuave:
Very WIP now but is functional. Had some issues with typing imports not resolving hence the `Any`. 

TODO:
- [ ] Add tests
- [ ] Add SQLGlot dialect to support physical properties for `SORT` & `SHARD` keys


erindru:
Hey @nicosuave , thanks for your PR.

We can give this a closer review once the SQLGlot dialect is available and you can base the `SingleStoreEngineAdapter` off it and get the tests passing. So far it looks like you're on the right track!

Also, for ongoing maintenance, we will need access to a SingleStore instance to run the integration tests against and to test new features. Is that something you can arrange?

nicosuave:
Thanks @erindru

> We can give this a closer review once the SQLGlot dialect is available and you can base the `SingleStoreEngineAdapter` off it and get the tests passing. So far it looks like you're on the right track!

Could you or someone on the team point me in the right direction as to where to add a new key / index type in a SQLGlot dialect? I saw the token mapping dict, but that didn't seem like quite enough.

> Also, for ongoing maintenance, we will need access to a SingleStore instance to run the integration tests against and to test new features. Is that something you can arrange?

Do you know the size of the test dataset? I think a Singlestore Helios cloud trial account should suffice https://www.singlestore.com/cloud-trial/

nicosuave:
If running locally in CI with Docker is preferred, I can also provide a license key (at least for some time).

erindru:
> Could you or someone on the team point me in the right direction as to where to add a new key / index type in a SQLGlot dialect? I saw the token mapping dict, but that didn't seem like quite enough.

@VaggelisD or @georgesittas are you able to assist here?

> Do you know the size of the test dataset? I think a Singlestore Helios cloud trial account should suffice https://www.singlestore.com/cloud-trial/

The test datasets are very small, we mostly focus on correctness at this stage. So a test run will issue hundreds / thousands of individual queries but will not load large amounts of data. I havent analyzed the limitations of your free tier but if its just based on data size then we are unlikely to exceed it

> If running locally in CI with Docker is preferred, I can also provide a license key (at least for some time)

We support a bunch of cloud-only databases (BigQuery, Snowflake, Redshift, Athena etc) so adding another one is not a big deal. If a license key is required, then there is little benefit to running it locally in Docker vs just running against a hosted cloud instance

georgesittas:
Hey 👋 

> Could you or someone on the team point me in the right direction as to where to add a new key / index type in a SQLGlot dialect? I saw the token mapping dict, but that didn't seem like quite enough.

Can you elaborate on what you mean by "key / index type" here? What's an example of that?

tobymao:
it's just custom single store ddl, it's like create table ( x key ...)

georgesittas:
I see, so if I understand correctly, it's this bit in the [syntax docs](https://docs.singlestore.com/cloud/reference/sql-reference/data-definition-language-ddl/create-table/):

```
<create_definition>:
...
| { INDEX | KEY } [<index_name>] [<index_type>] (<index_col_name>,...)
```

If that's the case, I think it's already supported in sqlglot's parser (based on some tests I'm seeing). Happy to provide more detailed assistance, though I think I'll need some more context on what exactly needs to be implemented here so I can guide accordingly.

nicosuave:
Thanks @georgesittas I'll take a closer look at the processing of physical properties; I assume that's the next spot to integrate

nicosuave:
I have made some progress; was just struggling with the test suite a bit

tobymao:
closing due to inactivity

PR #3249, opened 2024-10-13
Ensure using project instead of execution project for temp table as default

Kayrnt:
When doing a table_diff, the catalog is not specified which leads BigQuery to use the execution_project (as it's provided as "default project" to the BQ driver). Yet, we don't want the execution_project to be used for storage (as catalog) but the regular "project" which is here overriden at adapter level to ensure it's set on tables that don't have it and thus fixing the problem.


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -966,6 +967,22 @@ def _update_clustering_key(
...
+    @contextlib.contextmanager
+    def temp_table(
+        self,
+        query_or_df: QueryOrDF,
+        name: TableName = "diff",
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        **kwargs: t.Any,
+    ) -> t.Iterator[exp.Table]:
+        # ensure that we use default catalog if no catalog is not specified else it would use the execution project as default
+        if isinstance(name, exp.Table) and not name.catalog:
```
tobymao, line 979: what happens if name is not a table? should probably call to_table()

tobymao: should this be done in the base class, and not just bigquery?

erindru: Yeah, I cant think of a reason not to do this in the base class. It would make all references fully qualified / unambiguous

Kayrnt: `name` is a `TableName` which means it's either a `exp.Table` or a `str`. Then if it's a `str`, I assume it's either a `table` (in which case, it could be relevant but I'm not sure in which case it is?) or `catalog.db.table` which means that it might not be relevant to override the provided catalog by the default one.
However I see what you mean by expecting the `to_table()` wrapping but I wasn't it wouldn't mess up with something else.

So would you the following workflow:
- if name is an str => convert it to table
- Check if that exp.table has a catalog
- if not add the default one else keep existing one
- call super implementation

 works for you?

tobymao: yep thanks, and let's move this logic to the base class?

Kayrnt: Oh ok then I'll move it back to the base class 👍

Kayrnt: I wanted to use: 
```
        if isinstance(name, str):
            name = exp.to_table(name)
        # ensure that we use default catalog if none is not specified
        if not name.catalog:
            name.set("catalog", value=self.default_catalog)
```

However as you can see it leads to issues to:
- postgres (and I don't really understand why but it fails 1 test with the catalog change and 2 tests if I add the to_table)
- spark (I'm not sure why it changes the name of the table when I set the catalog?)
- table_diff.test_generated_sql (I'm not sure how it changes from a CREATE to a DROP 🤯)

I tried with only: 
```
 if not name.catalog:
            name.set("catalog", value=self.default_catalog)
```

but it's definitely that part that leads to most regressions. 

Do you really want to try to explore that approach? It looks like much harder to fix that way though.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2040,6 +2040,12 @@ def temp_table(
         Yields:
             The table expression
         """
+        if isinstance(name, str):
```
tobymao, line 2043: you can just unconditionally do table = exp.to_table(name)

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2040,6 +2040,12 @@ def temp_table(
...
+            name = exp.to_table(name)
+        # ensure that we use default catalog if none is not specified
+        if isinstance(name, exp.Table) and not name.catalog:
```
tobymao, line 2046: you may want to check db as well, if table.db, otherwise you'll add a catalog if there's no db...

sqlmesh/core/engine_adapter/base.py
Kayrnt: you mean `name = exp.to_table(name)` right?
As we have `table = self._get_temp_table(name)` later

sqlmesh/core/engine_adapter/base.py
Kayrnt: Ok I'm adding it but I would have expected that it would be projected only if both were defined. Isn't it checked at "print" level?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2040,10 +2040,9 @@ def temp_table(
         Yields:
             The table expression
         """
-        if isinstance(name, str):
-            name = exp.to_table(name)
+        name = exp.to_table(name)
         # ensure that we use default catalog if none is not specified
-        if isinstance(name, exp.Table) and not name.catalog:
+        if isinstance(name, exp.Table) and not name.catalog and not name.db:
```
tobymao, line 2045: it should be and name.db

PR #3253, opened 2024-10-15
feat: adding ability to specify table/view type

benfdking:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -988,7 +1008,11 @@ def create_view(
         )
         if create_view_properties:
             for view_property in create_view_properties.expressions:
-                properties.append("expressions", view_property)
+                # Small hack to make sure SECURE goes at the beginning before materialized as required by Snowflake
+                if isinstance(view_property, exp.SecureProperty):
+                    properties.expressions.insert(0, view_property)
```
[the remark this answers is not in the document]
benfdking: 🙏  for feedback, good to know. And I've modified that.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +408,20 @@ def create_index(
         )
         self.execute(expression)
 
+    class _KeyForType(Enum):
+        TABLE = "TABLE_TYPE"
+        VIEW = "VIEW_TYPE"
```
georgesittas: (Also discussed internally, sharing here for context)

I don't think we should associate the database object's type with the `*_properties` keys. For example, the physical layer can also contain views, i.e. if the model kind is `VIEW`, so setting `table_type` will be inconsistent.

Let's use a single common key within `physical_properties` and `virtual_properties`, e.g. `creatable_type = TEMPORARY`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +408,20 @@ def create_index(
...
+
+    def _parse_type_out_of_properties(
+        self, properties: t.Dict[str, exp.Expression], type_key: _KeyForType
+    ) -> t.Optional[exp.Expression]:
+        """Parses out the type from the properties dictionary and returns it if it exists. It does also remove the key"""
+        for key in list(properties.keys()):
+            upper_key = key.upper()
+            if upper_key == type_key.value:
+                return properties.pop(key)
+        return None
```
georgesittas: [Nit] should this be named "extract_creatable_type_from_properties"? The verb "parse" is a misnomer.
```suggestion
        return next(properties.pop(key) for key in properties if key.upper() = type_key.value, None)
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -988,7 +1008,11 @@ def create_view(
         )
         if create_view_properties:
             for view_property in create_view_properties.expressions:
-                properties.append("expressions", view_property)
+                # Small hack to make sure SECURE goes at the beginning before materialized as required by Snowflake
```
georgesittas, line 1008: I guess consolidating the `materialized` key may be a bit more involved, and would probably lead us to eventually deprecate it, but did we conclude that we don't want to allow folks to set it using this new property?

tests/core/engine_adapter/test_duckdb.py
```diff
@@ -75,3 +75,19 @@ def test_set_current_catalog(make_mocked_engine_adapter: t.Callable, duck_conn):
     assert to_sql_calls(adapter) == [
         'USE "test_catalog"',
     ]
+
+
+def test_temporary_table(make_mocked_engine_adapter: t.Callable, duck_conn):
```
georgesittas, line 80: Let's make sure to test this using a SQL example as well, besides calling the Python API methods.

georgesittas: Also, we should test using a list as well, e.g. `"creatable_type": [exp.TemporaryProperty(), ...]`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2115,6 +2137,11 @@ def _build_table_properties_exp(
                 )
             )
 
+        if table_properties:
+            table_type = self._extract_creatable_type_from_properties(table_properties)
+            if table_type:
+                properties.append(table_type)
```
georgesittas, line 2143: I think we may need to do something like this here instead (no need for branching):

```python
properties.extend(ensure_list(extract(...)))
```

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +409,24 @@ def create_index(
         )
         self.execute(expression)
 
+    @staticmethod
```
georgesittas, line 412: [Nit] What's the rationale behind using `@staticmethod` here? Should we just make this a normal function helper?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +409,24 @@ def create_index(
...
+    def _extract_creatable_type_from_properties(
+        properties: t.Dict[str, exp.Expression],
+    ) -> t.Optional[exp.Expression]:
+        """Parses out the type from the properties dictionary and returns it if it exists. It does also remove the key.
```
georgesittas, line 416: Let's make sure to the "Parses" bit here and in any other places too.

sqlmesh/core/engine_adapter/base.py
izeigerman: Yes. Materialized has special handling besides just setting a modified. For example, when deleting a materialized view we need to include the modifier as well. Also, in some engines materialized views need to be manually refreshed.

tests/core/engine_adapter/test_duckdb.py
benfdking: Due to some complexity, we agreed to push this slightly later.

sqlmesh/core/engine_adapter/base.py
benfdking: I renamed this slightly to be closer to Python conventions.

sqlmesh/core/engine_adapter/base.py
benfdking: Should be a test for this now and it raises an exception.

sqlmesh/core/engine_adapter/base.py
benfdking: Avoided the commit as introduced a little more inline complexity.

sqlmesh/core/engine_adapter/base.py
benfdking: I think this is done.

georgesittas (approved):

benfdking:
Two issues:

* [ ] Not working in duckdb with 

```
MODEL (
    name sqlmesh_example.full_model,
    kind FULL,
    cron '@daily',
    grain item_id,
    audits (assert_positive_order_ids),
    physical_properties (
      creatable_type = TEMP
    )
  );

  SELECT
    item_id,
    COUNT(DISTINCT id) AS num_orders,
  FROM
    sqlmesh_example.incremental_model
  GROUP BY item_id
  ```

* [ ] Change is marked as non-breaking

**Comment by benfdking** [comment]
```
MODEL (
    name sqlmesh_example.full_model,
    kind FULL,
    cron '@daily',
    grain item_id,
    audits (assert_positive_order_ids),
    virtual_properties (
      creatable_type = SECURE
    ),
    physical_properties (
      creatable_type = TRANSIENT
    )
  );

  SELECT
    item_id,
    COUNT(DISTINCT id) AS num_orders,
  FROM
    sqlmesh_example.incremental_model
  GROUP BY item_id
  ```

In Snowflake gives you a secure view and a transient physical layer.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +409,39 @@ def create_index(
...
+                        raise SQLMeshError(
+                            f"Cannot use {value.this.this} as a creatable_type as it conflicts with the `materialize` parameter."
+                        )
+                    # TODO Improve this
+                    if value_as_str == "SECURE":
+                        return exp.SecureProperty()
+                    if value_as_str == "TRANSIENT":
+                        return exp.TransientProperty()
+                    if value_as_str == "TEMPORARY" or value_as_str == "TEMP":
+                        return exp.TemporaryProperty()
```
georgesittas, line 439: I think we want to do something like:

```python
value = properties.pop(key).name
properties = exp.maybe_parse(value, into=exp.Properties, dialect=self.dialect)

property, *others = properties.expressions
if others:
    # Error - this is unsupported today, can look into it in the future if needed

if isinstance(property, exp.MaterializedProperty):
    # Error, can't use due to `materialize` parameter

return property

File: sqlmesh/core/engine_adapter/base.py [hunk 10]
```diff
@@ -951,6 +986,11 @@ def create_view(
         if not properties:
             properties = exp.Properties(expressions=[])
 
+        if view_properties:
+            table_type = self._pop_creatable_type_from_properties(view_properties)
+            if table_type:
+                properties.append("expressions", table_type)
```
georgesittas, line 992: Let's  be consistent and either do the `extend(ensure_list(...))` trick here or change the other sections.

tests/core/engine_adapter/test_duckdb.py
```diff
@@ -75,3 +75,17 @@ def test_set_current_catalog(make_mocked_engine_adapter: t.Callable, duck_conn):
...
+    adapter = make_mocked_engine_adapter(DuckDBEngineAdapter)
+
+    adapter.create_table(
+        "test_table",
+        {"a": exp.DataType.build("INT"), "b": exp.DataType.build("INT")},
+        table_properties={"creatable_type": exp.Column(this=exp.Identifier(this="Temporary"))},
```
georgesittas, line 86: You can simply do `exp.column("temporary")` instead of manually building the tree. Same for other tests.

tests/core/engine_adapter/test_snowflake.py
```diff
@@ -449,3 +449,198 @@ def test_materialized_view_properties(make_mocked_engine_adapter: t.Callable):
...
+def test_creatable_type_secure_materialized_view(make_mocked_engine_adapter: t.Callable):
+    adapter = make_mocked_engine_adapter(SnowflakeEngineAdapter)
+
+    adapter.create_view(
+        "test_table",
+        parse_one("SELECT 1"),
+        secure=True,
+        materialized=True,
+        view_properties={
+            "creatable_type": exp.Column(this=exp.Identifier(this="secure")),
```
georgesittas, line 481: Why do we set both `view_properties` and `secure=True` here? Is the latter correct?

sqlmesh/core/engine_adapter/base.py
benfdking: Much nicer

tests/core/engine_adapter/test_duckdb.py
benfdking: Addressed

tests/core/engine_adapter/test_snowflake.py
benfdking: That is legacy from my first attempt 😛

sqlmesh/core/engine_adapter/base.py
```diff
@@ -407,6 +409,44 @@ def create_index(
...
+        """
+        for key in list(properties.keys()):
+            upper_key = key.upper()
+            if upper_key == KEY_FOR_CREATABLE_TYPE:
+                value = properties.pop(key)
+                if (
+                    isinstance(value, exp.Column)
+                    and isinstance(value.this, exp.Identifier)
+                    and isinstance(value.this.this, str)
+                ):
```
georgesittas, line 427: I don't think we're ever _not_ going to have a `Column` here, hence why I went directly for `...pop(key).name` in that other [comment](https://github.com/TobikoData/sqlmesh/pull/3253#discussion_r1811270890).

georgesittas (approved):
Nice work 🎉

PR #3258, opened 2024-10-16
Feat!: Create temp tables only when needed

Themiscodes:
This aims to minimise unnecessary table creation (fixes: #2496 ) by skipping the creation of dev `__temp tables` when not needed, ie if snapshot `is_deployable` and doesn't reuse previous version (`forward_only`, `indirect_non_breaking`, `metadata`). In that case only prod table is created and the dry run is performed on this.

### DIFF
--- a/tests/core/test_snapshot_evaluator.py
+++ b/tests/core/test_snapshot_evaluator.py
@@ -181,19 +181,11 @@ def test_evaluate(mocker: MockerFixture, adapter_mock, make_snapshot):
         table_description=None,
     )
 
-    adapter_mock.create_table.assert_has_calls(
-        [
-            call(
-                snapshot.table_name(is_deployable=False),
-                column_descriptions=None,
-                **common_kwargs,
-            ),
-            call(
-                snapshot.table_name(),
-                column_descriptions={},
-                **common_kwargs,
-            ),
-        ]
+    # Create will be called once and only prod table will be created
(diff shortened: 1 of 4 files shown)


tests/core/test_snapshot_evaluator.py
```diff
@@ -181,19 +181,11 @@ def x(evaluator, y=None) -> None:
...
-                column_descriptions=None,
-                **common_kwargs,
-            ),
-            call(
-                snapshot.table_name(),
-                column_descriptions={},
-                **common_kwargs,
-            ),
-        ]
+    # Create will be called once and only prod table will be created
```
izeigerman, line 184: I've noticed we've modified all these tests, but have we added any which ensures that a dev table actually gets created?

Themiscodes: I added the `test_migrate_snapshot` test which validates a dev table been created for a forward only model but I will try to think of other use cases as well and add more

izeigerman (approved):

tests/core/test_snapshot_evaluator.py
Themiscodes: Added unit tests for each case of creating dev preview tables, if you want to take another look @izeigerman

izeigerman: Looks good, thanks 👍

PR #3261, opened 2024-10-16
Fix(clickhouse): remove fractional seconds when time column is datetime/timestamp type

treysp:
In some scenarios, the end of SQLMesh's incremental by time filter is created by subtracting 1 microsecond from a date, resulting in fractional seconds. For example, `2024-01-02` minus 1 microsecond = `2024-01-01 23:59:59.999999`.

Clickhouse's `DateTime` data type (and alias timestamp) only support second precision, so an error is thrown when it tries to cast fractional seconds to `DateTime`.

This PR prevents that error by removing the fractional seconds when the time column is `DateTime` or timestamp types.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -629,11 +629,9 @@ class _Model(ModelMeta, frozen=True):
                 time,
                 time_column_type,
                 self.time_column.format,
-                include_fractional_seconds=not (
-                    self.dialect in TIME_COLUMN_NO_FRACTIONAL_SECOND_TYPES
-                    and time_column_type.is_type(
-                        *TIME_COLUMN_NO_FRACTIONAL_SECOND_TYPES[self.dialect]
-                    )
+                include_microseconds=not (
+                    self.dialect in NO_MICROSECONDS
+                    and time_column_type.is_type(*NO_MICROSECONDS[self.dialect])
                 ),
             )
         return exp.convert(time)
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/date.py
```diff
@@ -256,14 +256,20 @@ def to_ds(obj: TimeLike) -> str:
     return to_ts(obj)[0:10]
 
 
-def to_ts(obj: TimeLike) -> str:
+def to_ts(obj: TimeLike, include_fractional_seconds: bool = True) -> str:
```
izeigerman, line 259: This feels like a mouthful. Can we just call this `include_microseconds`?

tobymao: what if it's precision?

tobymao: fine with include_microseconds as well

sqlmesh/core/model/definition.py
```diff
@@ -2440,3 +2450,8 @@ def _meta_renderer(
 def get_model_name(path: Path) -> str:
     path_parts = list(path.parts[path.parts.index("models") + 1 : -1]) + [path.stem]
     return ".".join(path_parts[-3:])
+
+
+TIME_COLUMN_NO_FRACTIONAL_SECOND_TYPES = {
```
izeigerman, line 2455: `NO_MICROSECOND`

sqlmesh/core/model/definition.py
```diff
@@ -2440,3 +2448,6 @@ def _meta_renderer(
 def get_model_name(path: Path) -> str:
     path_parts = list(path.parts[path.parts.index("models") + 1 : -1]) + [path.stem]
     return ".".join(path_parts[-3:])
+
+
+NO_MICROSECONDS = {"clickhouse": (exp.DataType.Type.DATETIME, exp.DataType.Type.TIMESTAMP)}
```
izeigerman, line 2453: I'm sorry, I think I didn't express myself clearly. I meant
```
TIME_COLUMN_NO_MICROSECONDS_TYPES
```
suggesting to replace the `FRACTION_SECONDS` part only. Apologies about the confusion

izeigerman (approved):
Cosmetic suggestion. Good to go otherwise

PR #3265, opened 2024-10-17
fix: handle missing signals in built in scheduler closes #3094

### LINKED ISSUE #3094: Built-in scheduler should consider changes to missing intervals due to signals
Assuming the following:

```sql
MODEL(name full_with_signal, kind FULL, cron '@daily', signals=[(kind=SomeSignal)]);
select * from external_table
```
and
```sql
MODEL(name other_full, kind FULL, cron '@daily');
select * from full_with_signal
```

When the scheduler runs and both models have a missing interval, then they will correctly execute in sequence, allowing `other_full` to take into account the data in `full_with_signal`.

If the signal on `full_with_signal` determines that the interval is not yet ready, it will correctly be skipped by the scheduler, but the interval for `other_full` **will still** be computed in that run. In a subsequent run, the signal becomes ready and the interval for `full_with_signal` is computed, but now we're in a situation where they have run in the reverse order, making `other_full` **not** take data from `full_with_signal` into account.

The built-in scheduler needs to consider changes to missing intervals made by signals.

tobymao:
if a signal marks an interval as 'unready', then all downstream jobs should also consider than interval as unready.

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -23,7 +23,7 @@ from sqlmesh.core.snapshot import (
     missing_intervals,
 )
 from sqlmesh.core.snapshot.definition import Interval as SnapshotInterval
-from sqlmesh.core.snapshot.definition import SnapshotId
+from sqlmesh.core.snapshot.definition import Intervals, SnapshotId, merge_intervals
 from sqlmesh.core.state_sync import StateSync
 from sqlmesh.utils import format_exception
 from sqlmesh.utils.concurrency import concurrent_apply_to_dag, NodeExecutionFailedError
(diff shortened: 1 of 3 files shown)


tobymao:
cc @plaflamme

plaflamme (approved):
Nice, this is simpler than I thought. LGTM!

tobymao:
i need t still handle mixed time granularities

sqlmesh/core/scheduler.py
```diff
@@ -549,19 +558,33 @@ def compute_interval_params(
...
                     signal=signal_factory(signal),
                     intervals=intervals,
                 )
+            unready -= set(intervals)
+        else:
+            unready = set()
+
+        for parent in snapshot.model.depends_on:
+            if parent in all_unready_intervals:
+                unready.update(all_unready_intervals[parent])
```
izeigerman, line 576: The batches can be completely misaligned between the parent and the child. It's not just cron granularity but also a different batch size.

izeigerman: Never mind the batch_size. It comes in later

izeigerman (approved):

PR #3268, opened 2024-10-18
Fix(clickhouse): use DateTime64(6) type in incremental time filter

treysp:
### Clickhouse background
Clickhouse has two datetime types:
- `DateTime`: fixed one second precision
- `DateTime64`: user-specified precision (default of 3/millisecond when bare type passed)

Clickhouse errors if you try to cast a string containing fractional seconds to `DateTime`. 

### Current PR
SQLMesh time intervals are sometimes calculated as [datetime - 1 microsecond], returning values containing fractional seconds.

This PR prevents fractional second casting errors by always using `DateTime64` in incremental time filters. It uses a default precision of 6 unless the user explicitly passed `DateTime64`.

The returned type is non-nullable by default. In SCD models, however, we generate statements like `CAST(NULL AS DateTime64)` that require a nullable type. Therefore, we force nullable types when calling `to_time_column()` in the SCD adapter method.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -330,21 +330,21 @@ workflows:
           name: cloud_engine_<< matrix.engine >>
           context:
             - sqlmesh_cloud_database_integration
-          requires:
-            - engine_tests_docker
+          # requires:
+          #   - engine_tests_docker
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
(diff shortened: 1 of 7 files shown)


erindru:
LGTM, were you meant to leave this as Draft?

treysp:
I thought this could be affected by how we go in https://github.com/tobymao/sqlglot/pull/4269, so was holding it until that is resolved

PR #3273, opened 2024-10-21
Fix: pass dialect to MD5 func builder in generate_surrogate_key macro

georgesittas:
Today we get the first behavior when using this macro in BigQuery, which is incorrect.

```python
>>> from sqlglot import exp
>>> exp.func("MD5", exp.Literal.string(f"Original value")).sql("bigquery")
"TO_HEX(MD5('Original value'))"
>>> exp.func("MD5", exp.Literal.string(f"Original value"), dialect="bigquery").sql("bigquery")
"MD5('Original value')"
```

Context: https://tobiko-data.slack.com/archives/C0448SFS3PF/p1729531496678999

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -895,15 +895,15 @@ def star(
 
 
 @macro()
-def generate_surrogate_key(_: MacroEvaluator, *fields: exp.Expression) -> exp.Func:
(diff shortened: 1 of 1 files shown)


mr-mcox:
Thank you for addressing this - it was causing me a bit of confusion!

georgesittas:
Sure thing, appreciate the report!

PR #3278, opened 2024-10-22
Docs: add secure and transient details for snowflake

benfdking:


georgesittas (approved):
LGTM, although I was wondering whether it makes sense to document the `creatable_type` in the model config section instead and just link to that here?

benfdking:
Good shout will add!

PR #3283, opened 2024-10-23
Docs: adding glossary term for virtual and physical layer

benfdking:


docs/concepts/glossary.md
```diff
@@ -57,6 +57,9 @@ Combining data from various sources (such as from a data warehouse) into one uni
 ## Lineage
 The lineage of your data is a visualization of the life cycle of your data as it flows from data sources downstream to consumption.
 
+## Physical Layer
+The physical layer is where SQLMesh actually stores and manages data in database tables and materialized views. It represents the concrete data storage layer of the system, in contrast to the virtual layer's abstract view definitions. While SQLMesh handles the management and maintenance of this layer automatically, shielding users from the underlying complexity, it's where all persistent data of models ultimately resides.
```
treysp, line 61: The physical layer is where SQLMesh stores and manages data in database tables and materialized views. It is the concrete data storage layer of the SQL engine, in contrast to the [SQLMesh virtual layer's](#virtual-layer) views. SQLMesh handles the management and maintenance of the physical layer automatically, and users should rarely interact with it directly.

docs/concepts/glossary.md
```diff
@@ -78,6 +81,9 @@ A view is the result of a SQL query on a database.
 ## Virtual Environments
 SQLMesh's unique approach to environment that allows it to provide both environment isolation and the ability to share tables across environments. This is done in a way to ensure data consistency and accuracy. See [plan application](plans.md#plan-application) for more information.
 
+## Virtual Layer
+The virtual layer is a conceptual framework managed by SQLMesh that provides an abstraction layer over the physical layer and physical data storage. Unlike the physical layer, where data is actually stored in tables and views, the virtual layer exists purely as a set of views, derived from the model definitions and environments, that expose the underlying physical layer. While typically transparent to end users, the virtual layer is the layer that users interact with when querying data.
```
treysp, line 85: The virtual layer is SQLMesh's abstraction layer over the [physical layer and physical data storage](#physical-layer). While the physical layer consists of tables where data is actually stored, the virtual layer consists of views that expose tables in the underlying physical layer. When a `plan` updates the models in an environment, SQLMesh automatically updates the environment's virtual layer views to point to the correct underlying physical table. Most users should only interact with the virtual layer when building models or querying data.

benfdking: Interesting. I guess in my definition, it was the more abstract layer that encompasses all the environments' virtual part whereas in yours, the virtual layer is part of an environment.

treysp: That's also how I think of it, but I thought it might be useful to clarify that each model name can have one view per environment, where the views point to different underlying physical tables. I'm now doubting that and think the environment-specific aspect is too detailed for a glossary definition. 

I updated my comment above to remove that sentence and I think it reads better - wdyt?

treysp (approved):

docs/concepts/glossary.md
benfdking: I like it!

docs/concepts/glossary.md
benfdking: Definitely better!

PR #3286, opened 2024-10-23
Fix: Adjustments to upgrade docs

mykalmax:


caiters (approved):
nothing stood out to me as alarming,  but since I'm less familiar with this code base would def recommend having Vincent take a look as well, unless you are confident :)

RN-nguyen:
Ran sqlmesh ui tests and looks good to me. https://app.qase.io/public/report/b22453ec7f3d58f36e46dbbf59b2900bf6002139

PR #3290, opened 2024-10-24
Docs: use github README as docs frontpage

treysp:


sungchun12 (approved):
Can you make sure the gif renders correctly? Besides that LGTM.

treysp:
It renders correctly locally, but I don't think we can test it on readthedocs without actually pushing it

PR #3295, opened 2024-10-25
Feat: Treat models with unaligned start as non-deployable insted of ignoring them

izeigerman:
Previously, SQLMesh ignored models that depend on past when the model's start date was before the specified plan start. This logic was introduced before we had the concepts of deployability and the deployability index. In practice, it caused a lot of confusion since it wasn't always obvious why a particular model wasn't evaluated when it was supposed to.

This update removes the concept of 'ignored' models and instead treats models with an invalid plan start date as non-deployable. This means that the model will still be evaluated, and its results will be available for preview in dev but will not be reused when deploying to prod.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -1308,7 +1308,9 @@ class DeployabilityIndex(PydanticModel, frozen=True):
 
     @classmethod
     def create(
-        cls, snapshots: t.Dict[SnapshotId, Snapshot] | t.Collection[Snapshot]
+        cls,
+        snapshots: t.Dict[SnapshotId, Snapshot] | t.Collection[Snapshot],
+        start: t.Optional[TimeLike] = None,
     ) -> DeployabilityIndex:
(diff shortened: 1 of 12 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -1318,6 +1320,8 @@ def create(
         deployability_mapping: t.Dict[SnapshotId, bool] = {}
         representative_shared_version_ids: t.Set[SnapshotId] = set()
 
+        start_date_cache: t.Optional[t.Dict[str, datetime]] = {}
```
tobymao, line 1323: this uses a new cache for ever snapshot, can we reuse this across all snapshots? i'm afraid this could be expensive

tobymao (approved):

sqlmesh/core/snapshot/definition.py
izeigerman: The cache is shared between all snapshots that are part of the index

eakmanrq (approved):

PR #3296, opened 2024-10-25
Feat(clickhouse): fall back to RENAME if EXCHANGE TABLES errors

treysp:
Clickhouse's atomic `EXCHANGE TABLES` command can fail on certain OSs, older versions of Clickhouse, or when exchanging incompatible tables.

This PR adds a fallback to the non-atomic equivalent `RENAME TABLES` if the exchange operation fails.


setup.py
```diff
@@ -94,6 +94,7 @@
             "types-requests==2.28.8",
             "typing-extensions",
             "custom-materializations",
+            "clickhouse-connect",
```
tobymao, line 97: move this into alphabeticalorder

tobymao (approved):

PR #3299, opened 2024-10-26
feat: add sqlmesh.lock and fix plans with only req changes

tobymao:


sqlmesh/core/constants.py
```diff
@@ -70,6 +70,7 @@
 
 EXTERNAL_MODELS_YAML = "external_models.yaml"
 EXTERNAL_MODELS_DEPRECATED_YAML = "schema.yaml"
+REQUIREMENTS = "sqlmesh.lock"
```
izeigerman, line 73: Not sure I like this name. In context of SQLMesh it's pretty ambiguous. Why not `requirements.sqlmesh.txt`? Or something like that.

sqlmesh/core/context_diff.py
```diff
@@ -59,6 +70,10 @@ class ContextDiff(PydanticModel):
     """Snapshot IDs that were promoted by the previous plan."""
     previous_finalized_snapshots: t.Optional[t.List[SnapshotTableInfo]]
     """Snapshots from the previous finalized state."""
+    previous_requirements: t.Dict[str, str] = {}
+    """Previous requirements."""
+    fixed_requirements: t.Dict[str, str] = {}
```
izeigerman, line 75: Let's call this `provided` instead of `fixed` maybe

sqlmesh/core/constants.py
tobymao: i'll make it sqlmesh-requirements.lock because it's not a requirements file, it's a lock file. the syntax is different

sqlmesh/core/context.py
```diff
@@ -338,6 +339,7 @@ def __init__(
                     loader=(loader or config.loader)(**config.loader_kwargs), configs={}
                 )
             self._loaders[project_type].configs[path] = config
+            self._load_requirements(path)
```
izeigerman, line 342: Shouldn't this happen in the loader?

sqlmesh/core/context_diff.py
```diff
@@ -207,14 +226,23 @@ def create_no_diff(cls, environment: str, state_reader: StateReader) -> ContextD
             previous_plan_id=env.plan_id,
             previously_promoted_snapshot_ids={s.snapshot_id for s in env.promoted_snapshots},
             previous_finalized_snapshots=env.previous_finalized_snapshots,
+            previous_requirements=env.requirements,
+            provided_requirements=env.requirements,
```
izeigerman, line 230: Is this a duplicated line?

izeigerman (approved):
One comment, LGTM otherwise

PR #3304, opened 2024-10-28
Fix: Skip creation call for tables that already exist

Themiscodes:
This update aims to eliminate unnecessary calls to create schemas and tables when the required deployable or non-deployable versions already exist, by identifying if the tables that need to be created are missing, instead of checking for both versions.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -263,9 +263,7 @@ class SnapshotEvaluator:
         for snapshot in target_snapshots:
             if not snapshot.is_model or snapshot.is_symbolic:
                 continue
-            deployability_flags = (
-                [True] if (not snapshot.is_managed or not snapshot.reuses_previous_version) else []
-            )
+            deployability_flags = [True]
             if (
                 snapshot.reuses_previous_version
                 or snapshot.is_managed
@@ -277,7 +275,7 @@ class SnapshotEvaluator:
                     snapshot.table_name(is_deployable), dialect=snapshot.model.dialect
                 )
                 snapshots_with_table_names[snapshot].add(table.name)
-                table_deployability[snapshot.name + table.name] = is_deployable
+                table_deployability[table.name] = is_deployable
                 tables_by_schema[d.schema_(table.db, catalog=table.catalog)].add(table.name)
 
         def _get_data_objects(schema: exp.Table) -> t.Set[str]:
@@ -295,16 +293,14 @@ class SnapshotEvaluator:
             }
 
         snapshots_to_create = []
-        versions_to_create: t.Dict[str, t.List[bool]] = defaultdict(list)
+        target_deployability_flags: t.Dict[str, t.List[bool]] = defaultdict(list)
         for snapshot, table_names in snapshots_with_table_names.items():
             missing_versions = table_names - existing_objects
             if missing_versions or (snapshot.is_seed and not snapshot.intervals):
                 snapshots_to_create.append(snapshot)
                 for version in missing_versions or table_names:
-                    versions_to_create[snapshot.name].append(
-                        table_deployability[snapshot.name + version]
-                    )
-                versions_to_create[snapshot.name].sort(key=lambda x: x)
+                    target_deployability_flags[snapshot.name].append(table_deployability[version])
+                target_deployability_flags[snapshot.name].sort()
             elif on_complete:
                 on_complete(snapshot)
 
@@ -318,7 +314,7 @@ class SnapshotEvaluator:
                 lambda s: self._create_snapshot(
                     s,
                     snapshots,
-                    versions_to_create,
+                    target_deployability_flags,
                     deployability_index,
                     on_complete,
                     allow_destructive_snapshots,
@@ -648,7 +644,7 @@ class SnapshotEvaluator:
         self,
         snapshot: Snapshot,
         snapshots: t.Dict[SnapshotId, Snapshot],
-        versions_to_create: t.Dict[str, t.List[bool]],
+        deployability_flags: t.Dict[str, t.List[bool]],
         deployability_index: t.Optional[DeployabilityIndex],
         on_complete: t.Optional[t.Callable[[SnapshotInfoLike], None]],
         allow_destructive_snapshots: t.Set[str],
@@ -721,8 +717,8 @@ class SnapshotEvaluator:
                 finally:
                     self.adapter.drop_table(tmp_table_name)
             else:
-                dry_run = not (len(versions_to_create[snapshot.name]) > 1)
-                for is_table_deployable in versions_to_create[snapshot.name]:
+                dry_run = not (len(deployability_flags[snapshot.name]) > 1)
+                for is_table_deployable in deployability_flags[snapshot.name]:
                     evaluation_strategy.create(
                         table_name=snapshot.table_name(is_deployable=is_table_deployable),
                         model=snapshot.model,
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -262,7 +262,10 @@ class SnapshotEvaluator:
         for snapshot in target_snapshots:
             if not snapshot.is_model or snapshot.is_symbolic:
                 continue
-            for is_deployable in (True, False):
+            is_snapshot_deployable = (
+                deployability_index.is_deployable(snapshot) if deployability_index else True
+            )
+            for is_deployable in self._table_deployability_flags(snapshot, is_snapshot_deployable):
                 table = exp.to_table(
                     snapshot.table_name(is_deployable), dialect=snapshot.model.dialect
                 )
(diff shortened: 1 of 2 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -898,6 +891,23 @@ def _create_schemas(self, tables: t.Iterable[t.Union[exp.Table, str]]) -> None:
...
 
+    def _table_deployability_flags(
+        self, snapshot: Snapshot, is_snapshot_deployable: bool
+    ) -> t.List[bool]:
+        flags = (
+            []
+            if (
+                is_snapshot_deployable
+                and not snapshot.reuses_previous_version
+                and not snapshot.is_managed
```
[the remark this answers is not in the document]
izeigerman: no, sorry, I got confused

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -262,7 +262,10 @@ def create(
         for snapshot in target_snapshots:
             if not snapshot.is_model or snapshot.is_symbolic:
                 continue
-            for is_deployable in (True, False):
+            is_snapshot_deployable = (
+                deployability_index.is_deployable(snapshot) if deployability_index else True
+            )
+            for is_deployable in self._table_deployability_flags(snapshot, is_snapshot_deployable):
```
izeigerman, line 268: I think we always want to check for the deployable table here and only sometimes for non-deployable too. Otherwise, this update strips sqlmesh of its self-healing properties, i.e. automatically recreating the table if one was deleted.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -898,6 +891,23 @@ def _create_schemas(self, tables: t.Iterable[t.Union[exp.Table, str]]) -> None:
...
+            []
+            if (
+                is_snapshot_deployable
+                and not snapshot.reuses_previous_version
+                and not snapshot.is_managed
+            )
+            else [False]
+        )
+        if not snapshot.reuses_previous_version:
+            flags.append(True)
```
izeigerman, line 907: To reiterate: we need to make sure we recreate the deployable table if it's actually missing.

sqlmesh/core/snapshot/evaluator.py
Themiscodes: Right I hadn't considered this possibility. I've updated to always check for deployable versions and when necessary for non-deployable versions. But I now realise this information of missing tables also needs to be passed into the `_create_snapshot` function so looking into how to address this there as well

sqlmesh/core/snapshot/evaluator.py
Themiscodes: Refactored this to always check if the deployable table is missing and moved logic of identifying tables to be created to the `create` function. Then after getting the missing tables that need to be built, passing these table `versions_to_create` (based on deployability) to the `_create_snapshot` function to avoid duplicating logic

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -259,14 +259,25 @@ def create(
         """
         snapshots_with_table_names = defaultdict(set)
         tables_by_schema = defaultdict(set)
+        table_deployability: t.Dict[str, bool] = {}
         for snapshot in target_snapshots:
             if not snapshot.is_model or snapshot.is_symbolic:
                 continue
-            for is_deployable in (True, False):
+            deployability_flags = (
+                [True] if (not snapshot.is_managed or not snapshot.reuses_previous_version) else []
```
izeigerman, line 267: I suggest we always have `True` here. Unconditionally

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -284,9 +295,16 @@ def _get_data_objects(schema: exp.Table) -> t.Set[str]:
...
         for snapshot, table_names in snapshots_with_table_names.items():
-            if table_names - existing_objects or (snapshot.is_seed and not snapshot.intervals):
+            missing_versions = table_names - existing_objects
+            if missing_versions or (snapshot.is_seed and not snapshot.intervals):
                 snapshots_to_create.append(snapshot)
+                for version in missing_versions or table_names:
+                    versions_to_create[snapshot.name].append(
+                        table_deployability[snapshot.name + version]
+                    )
+                versions_to_create[snapshot.name].sort(key=lambda x: x)
```
izeigerman, line 307: Why do we sort?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -259,14 +259,25 @@ def create(
...
+                or snapshot.is_managed
+                or (deployability_index and not deployability_index.is_deployable(snapshot))
+            ):
+                deployability_flags.append(False)
+            for is_deployable in deployability_flags:
                 table = exp.to_table(
                     snapshot.table_name(is_deployable), dialect=snapshot.model.dialect
                 )
                 snapshots_with_table_names[snapshot].add(table.name)
+                table_deployability[snapshot.name + table.name] = is_deployable
```
izeigerman, line 280: I think table names are unique enough so you can just use them as a key

sqlmesh/core/snapshot/evaluator.py
Themiscodes: This wa for the test assertions that sometimes failed due to a different order of the create in the case of both deployable and non-deployable versions, so this way it would always be in the order [False, True]

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -625,6 +648,7 @@ def _create_snapshot(
         self,
         snapshot: Snapshot,
         snapshots: t.Dict[SnapshotId, Snapshot],
+        versions_to_create: t.Dict[str, t.List[bool]],
```
izeigerman, line 651: Feels like we can just pass `t.List[bool]` here, since we only need a record for the provided `snapshot`

sqlmesh/core/snapshot/evaluator.py
izeigerman: Also I think the `key` argument is redundant

sqlmesh/core/snapshot/evaluator.py
Themiscodes: This was only for managed tables to retain the logic. Which if I'm not mistaken since they don't have an existing deployable table it would lead to appear as missing, if you can confirm @erindru this as well?

Then in the case of having to create a deployable table it would lead in calling `create_managed_table` which we wouldn't want for a snapshot that reuses a previous version I think, But I can adapt the logic for managed tables in the `_create_snapshot` function so that is avoided and then always have `True` in the flags here

sqlmesh/core/snapshot/evaluator.py
izeigerman: Also `versions` seems a little overloaded here. Let's be explicit and call it `deployability_flags` or `deployability_flags_to_create`.

sqlmesh/core/snapshot/evaluator.py
Themiscodes: Oh right will change this

sqlmesh/core/snapshot/evaluator.py
Themiscodes: Yes I was unsure of what to call it to make sense in this context, will alter this back to `deployability_flags`

sqlmesh/core/snapshot/evaluator.py
izeigerman: Wdym? This is just for checking the table existence, right? 

>  which we wouldn't want for a snapshot that reuses a previous version I think

If it reuses the previous version than the deployable table **should** exist unless it was dropped externally, right?

Themiscodes: I see yes, I misunderstood how managed tables work because I was focusing on the test:
`test_create_managed_forward_only_with_previous_version_doesnt_clone_for_dev_preview`
because `get_data_objects` wasn't returning any existing deployable table. So I will patch this function in this test and then change to always check for missing deployable versions as you suggested

sqlmesh/core/snapshot/evaluator.py
Themiscodes: I'm not certain though of how to use the list of the flags for the specific snapshot instead of the whole dict since the function is called in `concurrent_apply_to_snapshots`

izeigerman: The function is actually called `_create_snapshot` (singular), no? It just gets passed into `concurrent_apply_to_snapshots` for application over a collection of snapshots

Themiscodes: Yes my oversight changed this to use the specific snapshot list

Themiscodes:
Thank you for the review @izeigerman addressed comments

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -284,9 +293,14 @@ def _get_data_objects(schema: exp.Table) -> t.Set[str]:
             }
 
         snapshots_to_create = []
+        target_deployability_flags: t.Dict[str, t.List[bool]] = defaultdict(list)
         for snapshot, table_names in snapshots_with_table_names.items():
-            if table_names - existing_objects or (snapshot.is_seed and not snapshot.intervals):
+            missing_versions = table_names - existing_objects
```
izeigerman, line 298: [Nit] `missing_tables`

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -697,22 +717,8 @@ def _create_snapshot(
...
-                    )
-                    else [False]
-                )
-                if not snapshot.reuses_previous_version:
-                    table_deployability_flags.append(True)
-                    if len(table_deployability_flags) > 1:
-                        dry_run = False
-
-                for is_table_deployable in table_deployability_flags:
+                dry_run = not (len(deployability_flags) > 1)
```
izeigerman, line 720: [Nit]:
```
dry_run =  len(deployability_flags) == 1
```

izeigerman (approved):
A few nitpicks, LGTM otherwise. Thanks a lot for addressing comments

PR #3311, opened 2024-10-30
feat: add temp_schema option for table_diff

Kayrnt:
So far, table_diff will create a temp table in `sqlmesh_temp` schema.
However in some environments, the user might want to use an existing schema for that because he either don't want to create and manage another schema or he doesn't have the rights to do so.

The PR introduces a CLI parameter `--temp-schema <catalog>.<schema>` or `--temp-schema <schema>` to write the temp tables inside that catalog and/or schema.

To do list:
- [x] Check that works as expected
- [ ] Ensure it works with API Endpoint
- [x] Add related documentation


web/server/api/endpoints/table_diff.py
```diff
@@ -34,7 +35,7 @@ def get_table_diff(
         show=False,
     )
     _schema_diff = diff.schema_diff()
-    _row_diff = diff.row_diff()
+    _row_diff = diff.row_diff(temp_schema=temp_schema)
```
Kayrnt, line 38: I'm not sure I'm not breaking something here, I guess it's fine.

georgesittas (approved):
Seems good, thanks for the PR @Kayrnt 👍

Kayrnt:
I forgot to delete a small part I added in the doc (following a Copilot suggestion 🙈). 
I clean it up, I think it should be good to go 👍

PR #3321, opened 2024-10-31
fix: decorator type erasure

etonlels:
Functions decorated with `cli_analytics` or `python_api_analytics` are not type-safe because the decorators are essentially stripping type information from the functions they decorate. This PR fixes that.

### DIFF
--- a/sqlmesh/core/analytics/__init__.py
+++ b/sqlmesh/core/analytics/__init__.py
@@ -9,6 +9,17 @@ from sqlmesh.core.analytics.collector import AnalyticsCollector
 from sqlmesh.core.analytics.dispatcher import AsyncEventDispatcher, NoopEventDispatcher
 from sqlmesh.utils import str_to_bool
 
+if t.TYPE_CHECKING:
+    import sys
+
+    if sys.version_info >= (3, 10):
+        from typing import ParamSpec
+    else:
+        from typing_extensions import ParamSpec
+
+    _P = ParamSpec("_P")
+    _T = t.TypeVar("_T")
+
 
 def init_collector() -> AnalyticsCollector:
     dispatcher = (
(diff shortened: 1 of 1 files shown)


sqlmesh/core/analytics/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 import atexit
 import os
-import typing as t
+import typing_extensions as t
```
georgesittas, line 5: Should we put this and the two globals under a `t.TYPE_CHECKING` branch?

etonlels: We can. I changed it. Let me know if it's what you had in mind.

sqlmesh/core/analytics/__init__.py
```diff
@@ -9,6 +9,12 @@
 from sqlmesh.core.analytics.dispatcher import AsyncEventDispatcher, NoopEventDispatcher
 from sqlmesh.utils import str_to_bool
 
+if t.TYPE_CHECKING:
+    from typing_extensions import ParamSpec
```
izeigerman, line 13: I think we need something like
```
if sys.version_info >= (3, 9):
    from typing import ParamSpec
else:
    from typing_extensions import ParamSpec
```

etonlels: What's the rationale for doing that? The `typing_extensions` module already has logic to determine whether it should just import the corresponding class in the `typing` module vs. recreating the type itself. It seems best to let it handle that instead of trying to redo it.

etonlels: @izeigerman I pushed a change similar to what you suggested. Hopefully that unblocks this.

izeigerman (approved):

Issue #3317, opened 2024-10-31
Cannot generate external model for BigQuery information schema


ncbkr:
We have a model that is querying the [BigQuery information schema](https://cloud.google.com/bigquery/docs/information-schema-intro#syntax). In contrast to regular tables living in datasets within projects, the `INFORMATION_SCHEMA` follows a different [syntax](https://cloud.google.com/bigquery/docs/information-schema-intro#syntax) that does not have to comply with the usual `<project>`.`<dataset>`.`<table>` format. E.g. this would be valid identifier for an information schema table: 

```
myProject.`region-us`.INFORMATION_SCHEMA.TABLES` 
```

We have a query like the following:

```
SELECT *
FROM `region-eu`.INFORMATION_SCHEMA.JOBS
```

When running `sqlmesh create_external_models`, we receive the following warning

> 2024-10-31 15:08:10,605 - ThreadPoolExecutor-0_0 - sqlmesh.core.schema_loader - WARNING - Unable to get schema for '"region-eu"."INFORMATION_SCHEMA"."JOBS"': '404 GET https://bigquery.googleapis.com/bigquery/v2/projects/region-eu/datasets/INFORMATION_SCHEMA/tables/JOBS?prettyPrint=false: Project region-eu is not found. Make sure it references valid GCP project that hasn't been deleted.; Project id: region-eu'. (schema_loader.py:62)

The warning indicates that SQLMesh is interpreting `region-eu` as the project name, which is not true for information schema tables.

PR #3324, opened 2024-11-01
Fix: enable fetching schema for models querying INFORMATION_SCHEMA

georgesittas:
This is related to #3317 but does not completely fix it. There are two problems at hand here:

1. We're currently parsing queries like `SELECT * FROM dataset.INFORMATION_SCHEMA.SOME_VIEW` incorrectly: the `Table` node's `catalog` field gets set to `dataset`, but the correct representation is to keep it in `db` and create a `Dot` node for the `INFORMATION_SCHEMA.SOME_VIEW` part. I'm planning to look into this issue separately, as it's gonna be more involved and will probably need some new logic in SQLGlot to detect references to `INFORMATION_SCHEMA` views.
2. Even if the table was fully-qualified, i.e. `SELECT * FROM project.dataset.INFORMATION_SCHEMA.SOME_VIEW`, we would still fail because the BigQuery client's `get_table` [method](https://github.com/TobikoData/sqlmesh/blob/feb96867c7a3242eab2b6d23e1f78dcca9c9c411/sqlmesh/core/engine_adapter/bigquery.py#L590-L596) doesn't work with table references that have more than 3 identifiers or "parts" in the referenced database object. Here's the [source code](https://github.com/googleapis/python-bigquery/blob/fef8b886bc86d355c7745585fc53dc8a5a019ab1/google/cloud/bigquery/client.py#L1132) in case anyone wants to verify this, the bottom line is that this [method](https://github.com/googleapis/python-bigquery/blob/fef8b886bc86d355c7745585fc53dc8a5a019ab1/google/cloud/bigquery/_helpers.py#L945) is called with a 4-part reference and fails:

```
ValueError: table_id must be a fully-qualified ID in standard SQL format, e.g., "project.dataset.table_id", got project.dataset.INFORMATION_SCHEMA.TABLES
```

This PR addresses (2), while https://github.com/tobymao/sqlglot/pull/4336 addresses (1).

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -223,17 +223,35 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 return "JSON"
             return kind.name
 
-        table = self._get_table(table_name)
-        columns = {
-            field.name: exp.DataType.build(
-                dtype_to_sql(field.to_standard_sql().type), dialect=self.dialect
-            )
-            for field in table.schema
-        }
-        if include_pseudo_columns and table.time_partitioning and not table.time_partitioning.field:
-            columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP")
-            if table.time_partitioning.type_ == "DAY":
-                columns["_PARTITIONDATE"] = exp.DataType.build("DATE")
+        def create_mapping_schema(
+            schema: t.Sequence[bigquery.SchemaField],
+        ) -> t.Dict[str, exp.DataType]:
+            return {
+                field.name: exp.DataType.build(
+                    dtype_to_sql(field.to_standard_sql().type), dialect=self.dialect
+                )
+                for field in schema
+            }
+
+        table = exp.to_table(table_name)
+        if len(table.parts) > 3:
+            # The client's `get_table` method can't handle paths with >3 identifiers
+            self.execute(exp.select("*").from_(table))
+            query_results = self._query_job._query_results
+            columns = create_mapping_schema(query_results.schema)
+        else:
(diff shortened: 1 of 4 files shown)


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -223,17 +223,35 @@ def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
...
+                field.name: exp.DataType.build(
+                    dtype_to_sql(field.to_standard_sql().type), dialect=self.dialect
+                )
+                for field in schema
+            }
+
+        table = exp.to_table(table_name)
+        if len(table.parts) > 3:
+            # The client's `get_table` method can't handle paths with >3 identifiers
+            self.execute(exp.select("*").from_(table))
```
erindru, line 239: Do we want to `LIMIT 1` on this?

tests/core/engine_adapter/integration/test_integration_bigquery.py
```diff
@@ -178,3 +178,43 @@ def _get_data_object(table: exp.Table) -> DataObject:
 
     metadata = _get_data_object(target_table_1)
     assert not metadata.is_clustered
+
+
+def test_fetch_schema_of_information_schema_tables(
+    ctx: TestContext, engine_adapter: BigQueryEngineAdapter
+):
+    # We produce Table(this=Dot(this=INFORMATION_SCHEMA, expression=TABLES)) here,
+    # otherwise `db` or `catalog` wo3uld be set, which is not the right representation
```
erindru, line 187: nit: wo3uld

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: Ah, good catch. Yeah, possibly.

erindru (approved):

PR #3328, opened 2024-11-04
Chore: Refactor context run and _run

crericha:


sqlmesh/core/context.py
```diff
@@ -627,30 +629,75 @@ def run(
...
+                    circuit_breaker=_has_environment_changed,
+                )
+                done = True
+            except CircuitBreakerError:
+                logger.warning(
+                    "Environment '%s' has been modified while running. Restarting the run...",
+                    environment,
+                )
+                if exit_on_env_update:
+                    interrupted = True
```
izeigerman, line 691: Let's just set `done = True` here and keep only `done` in the while loop

PR #3330, opened 2024-11-05
Feat: Add multiple gateway support for DuckDB and Postgres

Themiscodes:
This update enables sourcing data from multiple PostgreSQL and DuckDB gateways, allowing references to the databases defined in the config file's `gateways` by setting the `cross_gateway` key to `true` :

```yaml
gateways:
    main_db:
        connection:
            type: duckdb
            database: main_db.db
    postgres_db:
        connection:
            type: postgres
            host: 127.0.0.1
            port: 5432
            user: admin
            password: 1234
            database: postgres_db
    secondary_db:
        connection:
            type: duckdb
            database: secondary_db.duckdb
            
default_gateway: main_db
cross_gateway: true
```

Then in the model definitions, data can be sourced from any of the gateways, allowing for operations like joining tables across them:

```sql
MODEL (
  name main_db.sales_summary,
  kind FULL
);

SELECT
  s.customer_id,
  s.order_id,
  s.order_date,
  p.product_name
FROM 
  postgres_db.public.orders AS s
JOIN 
  secondary_db.products AS p
ON 
  s.product_id = p.product_id;
```

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -57,6 +57,11 @@ from sqlmesh.core import constants as c
 from sqlmesh.core.analytics import python_api_analytics
 from sqlmesh.core.audit import Audit, ModelAudit, StandaloneAudit
 from sqlmesh.core.config import CategorizerConfig, Config, load_configs
+from sqlmesh.core.config.connection import (
+    DuckDBAttachOptions,
+    DuckDBConnectionConfig,
+    PostgresConnectionConfig,
+)
 from sqlmesh.core.config.loader import C
 from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context_diff import ContextDiff
@@ -362,6 +367,10 @@ class GenericContext(BaseContext, t.Generic[C]):
 
         self._connection_config = self.config.get_connection(self.gateway)
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
+
+        if self.config.cross_gateway:
+            self._load_multiple_gateways()
+
         self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
         self.console = console or get_console(dialect=self._engine_adapter.dialect)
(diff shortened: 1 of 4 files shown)


sqlmesh/core/context.py
```diff
@@ -2148,6 +2157,36 @@ def _load_requirements(self, path: Path) -> None:
...
+            if isinstance(config.connection, PostgresConnectionConfig):
+                self._connection_config.catalogs[gateway_name] = DuckDBAttachOptions(
+                    type=config.connection.type_, path=config.connection._libpq_string
+                )
+            elif (
+                isinstance(config.connection, DuckDBConnectionConfig)
+                and config.connection.database
+                and config.connection.database != self._connection_config.database
+            ):
+                self._connection_config.catalogs[config.connection.database.split(".")[0]] = (
```
erindru, line 2184: Are you trying to parse the catalog out of here? It might be best to use a sqlglot parsing function, i'm thinking `exp.to_table()` but @georgesittas will know the most appropriate one

Themiscodes: Yes it was to get for example from "main.db" the database name so "main", but maybe I could use the gateway name there instead directly assuming it is the same as the catalog name

sqlmesh/core/context.py
```diff
@@ -2148,6 +2157,36 @@ def _load_requirements(self, path: Path) -> None:
...
 
+    def _load_multiple_gateways(self) -> None:
+        # Validate there are multiple gateways and the default is duckdb
+        if (
+            len(self.config.gateways) < 2
+            or not isinstance(self._connection_config, DuckDBConnectionConfig)
+            or not self._connection_config.database
+            or self._connection_config.catalogs
+        ):
+            return
```
georgesittas, line 2168: Should we actually validate more strictly here? If the cross-gateway conditions aren't met, then it seems like we should just fail and emit a warning, otherwise the error handling is deferred to post-load time. Unless I'm missing something?

sqlmesh/core/context.py
```diff
@@ -2148,6 +2157,36 @@ def _load_requirements(self, path: Path) -> None:
...
+
+        # Set the attach options for the main and additional connection gateways (Currently supported DuckDB and PostgreSQL)
+        self._connection_config.catalogs = {
+            self._connection_config.database.split(".")[0]: self._connection_config.database
+        }
```
georgesittas, line 2173: Instead of splitting and indexing on `[0]`, can we instead use `Path.stem`?

sqlmesh/core/context.py
```diff
@@ -2148,6 +2157,36 @@ def _load_requirements(self, path: Path) -> None:
...
+        for gateway_name, config in self.config.gateways.items():
+            if isinstance(config.connection, PostgresConnectionConfig):
+                self._connection_config.catalogs[gateway_name] = DuckDBAttachOptions(
+                    type=config.connection.type_, path=config.connection._libpq_string
+                )
```
georgesittas, line 2178: I'm not sure about mutating the connections' `catalogs` attribute here.. Curious what others think, but my hunch is we may wanna push this down to the config validators or something.

tests/core/test_config.py
```diff
@@ -702,3 +703,72 @@ def test_scheduler_config(tmp_path_factory):
     assert isinstance(config.get_gateway("airflow_gateway").scheduler, AirflowSchedulerConfig)
     assert isinstance(config.get_gateway("mwaa_gateway").scheduler, MWAASchedulerConfig)
     assert isinstance(config.get_gateway("builtin_gateway").scheduler, BuiltInSchedulerConfig)
+
+
+def test_load_multiple_gateways(tmp_path):
```
georgesittas, line 708: Let's include an end-to-end integration test as well.

sqlmesh/core/context.py
georgesittas: > but maybe I could use the gateway name there instead directly assuming it is the same as the catalog name

This assumption looks incorrect, catalogs and gateway names are orthogonal concepts. See `stem` suggestion regarding the string splitting bit.

sqlmesh/core/context.py
Themiscodes: I thought in the rare case where the cross_gateway is set to true but the user only has one gateway, or maybe a different type than the supported ones, not to fail, but simply continue with the default gateway as normal. But I'm happy to make the validation stricter and emitting an error when the conditions aren't met

sqlmesh/core/context.py
Themiscodes: Good idea, this makes more sense I'll work on moving it there

PR #3331, opened 2024-11-05
Feat: allow formatting specific models via path posargs for more typi…

z3z1ma:
…cal and robust cli behavior.

This allows `sqlmesh format` to be used much more prescriptively. Furthermore it can be used in pre-commit hooks correctly this way too. The all or nothing was very limiting so we get this discernment almost for free given small change.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -971,12 +971,15 @@ class GenericContext(BaseContext, t.Generic[C]):
         append_newline: t.Optional[bool] = None,
         *,
         check: t.Optional[bool] = None,
+        paths: t.Optional[t.Tuple[t.Union[str, Path], ...]] = None,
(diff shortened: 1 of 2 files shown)


z3z1ma:
I think a diagnostic print of `print(f"Formatting {target._path}", file=sys.stderr)` would make the execution less obscure to the end user too but I'll let you weigh in.

sqlmesh/core/context.py
```diff
@@ -971,13 +971,16 @@ def format(
         append_newline: t.Optional[bool] = None,
         *,
         check: t.Optional[bool] = None,
+        paths: t.Tuple[t.Union[str, Path], ...] = (),
```
izeigerman, line 974: [Nit] can we make it optional instead

izeigerman (approved):

PR #3332, opened 2024-11-05
Fix!: bump sqlglot to v25.29.0, fix info schema view handling in bigquery

### LINKED ISSUE #3317: Cannot generate external model for BigQuery information schema

georgesittas:
Fixes #3317

The goal of this PR is to enable SQLMesh to correctly handle information schema view references in BigQuery. The main problem with those until now was that, in their fully-qualified form, they comprised 4 identifiers:

```sql
project.dataset_or_region.INFORMATION_SCHEMA.SOME_VIEW
```

This means that we'd end up with `Table` references of mixed nesting, e.g. model names comprise 3 identifiers:

```sql
project.dataset.model_name
```

Mixing multiple nesting levels in table references is prohibited by SQLGlot's schema module [[1](https://github.com/tobymao/sqlglot/blob/87ab8fe9cc4d6d060d8fe8a9c3faf8c47c2c9ed6/sqlglot/schema.py#L289-L293), [2](https://github.com/tobymao/sqlglot/blob/87ab8fe9cc4d6d060d8fe8a9c3faf8c47c2c9ed6/sqlglot/schema.py#L149)], in order to avoid issues related to ambiguity. So, one workaround for that was to represent information schema views using 3 identifiers at parse time, only for BigQuery. Other engines don't allow >3 identifiers in their table references based on my investigation.

I went with this approach because making the schema module more lenient, i.e. allowing multiple nesting depths, was quite complex. We rely on the invariant that the depth is the same in several places and the scope of the current approach seemed way smaller in comparison.

I guess one thing we'll need to be careful about is that parsing BigQuery's information schema views without specifying the dialect can result in an incorrect AST representation, because we represent the first example with the 4 identifiers using a `Dot` instead of merging the last two parts into a single `Identifier`, as done in BigQuery's parser.

For additional context, please refer to:
- [Fix: enable fetching schema for models querying INFORMATION_SCHEMA](https://github.com/TobikoData/sqlmesh/commit/c866f838d339d0e45360cb79673af2573c564d5c)
- [Fix(bigquery)!: parse information schema views into a single identifier](https://github.com/tobymao/sqlglot/commit/84f78aafd5d7e74da407167cd394d2bff0718cfb)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         "requests",
         "rich[jupyter]",
         "ruamel.yaml",
-        "sqlglot[rs]~=25.28.0",
+        "sqlglot[rs]~=25.29.0",
         "tenacity",
     ],
     extras_require={
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -234,7 +234,7 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
             }
 
         table = exp.to_table(table_name)
-        if len(table.parts) > 3:
+        if len(table.parts) == 3 and "." in table.name:
             # The client's `get_table` method can't handle paths with >3 identifiers
             self.execute(exp.select("*").from_(table).limit(1))
             query_results = self._query_job._query_results
--- a/tests/core/engine_adapter/integration/__init__.py
+++ b/tests/core/engine_adapter/integration/__init__.py
@@ -217,7 +217,7 @@ class TestContext:
         schema = self.add_test_suffix(schema)
         self._schemas.append(schema)
 
-        table = exp.to_table(table_name)
+        table = exp.to_table(table_name, dialect=self.dialect)
         table.set("db", exp.parse_identifier(schema, dialect=self.dialect))
 
         return exp.to_table(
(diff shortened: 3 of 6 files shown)


erindru (approved):
I'm.. amazed this works

izeigerman (approved):

PR #3334, opened 2024-11-05
Chore: Fix build

erindru:


erindru:
The migration test will always fail because the code in `LAST_TAG` (`v0.130.2`) contains the issue

izeigerman (approved):

PR #3340, opened 2024-11-07
Feat: Add dlt_refresh command to update models from a dlt pipeline

Themiscodes:
As part of the dlt integration, this update introduces the `dlt_refresh` command to attach to a dlt pipeline and update the SQLMesh models, either by specifying the tables to be generated or updating with all the new tables.
### Usage:

- **Generate all missing tables**:
  
```bash
$ sqlmesh dlt_refresh <pipeline-name>
```

- **Generate all missing tables and overwrite existing ones** (use with `--force` or `-f`):
  
```bash
$ sqlmesh dlt_refresh <pipeline-name> --force
```

- **Generate specific dlt tables** (using `--table` or `-t`):

```bash
$ sqlmesh dlt_refresh <pipeline-name> --table <dlt-table>
```

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -916,3 +916,38 @@ def clean(obj: Context) -> None:
 def table_name(obj: Context, model_name: str, dev: bool) -> None:
     """Prints the name of the physical table for the given model."""
     print(obj.table_name(model_name, dev))
+
+
+@cli.command("dlt")
+@click.argument("pipeline", required=True)
+@click.option(
+    "-u",
+    "--update",
+    type=str,
+    multiple=True,
+    is_flag=False,
+    flag_value="UpdateAllFlag",
+    help="The dlt tables to update in the SQLMesh models. When none specified, all missing tables will be added.",
+)
+@click.option(
+    "-f",
+    "--force",
+    is_flag=True,
+    default=False,
+    help="If set it will overwrite the existing models with the new dlt tables.",
+)
+@click.pass_context
+@error_handler
+@cli_analytics
+def dlt(
+    ctx: click.Context,
+    pipeline: str,
+    force: bool,
+    update: t.List[str] = [],
+) -> None:
+    """Attaches to a DLT pipeline with the option to update specific or all models in the SQLMesh project."""
+    from sqlmesh.integrations.dlt import update_dlt_models
+
+    if update:
+        update_tables = list(update) if update[0] != "UpdateAllFlag" else []
+        ctx.obj.console.log_success(update_dlt_models(ctx.obj, pipeline, update_tables, force))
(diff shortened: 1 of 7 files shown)


sqlmesh/cli/main.py
```diff
@@ -916,3 +916,38 @@ def clean(obj: Context) -> None:
 def table_name(obj: Context, model_name: str, dev: bool) -> None:
     """Prints the name of the physical table for the given model."""
     print(obj.table_name(model_name, dev))
+
+
+@cli.command("dlt")
+@click.argument("pipeline", required=True)
+@click.option(
+    "-u",
+    "--update",
```
izeigerman, line 925: Should this be `--table`?

sqlmesh/cli/main.py
```diff
@@ -916,3 +916,38 @@ def clean(obj: Context) -> None:
...
+def dlt(
+    ctx: click.Context,
+    pipeline: str,
+    force: bool,
+    update: t.List[str] = [],
+) -> None:
+    """Attaches to a DLT pipeline with the option to update specific or all models in the SQLMesh project."""
+    from sqlmesh.integrations.dlt import update_dlt_models
+
+    if update:
```
izeigerman, line 951: The interface doesn't seem to be intuitive. Basically you **always** need the `--update` flag otherwise the command does nothing.

sqlmesh/integrations/dlt.py
```diff
@@ -85,6 +89,29 @@ def generate_dlt_models_and_settings(
...
+    )
+
+    if not update_tables and not force:
+        existing_models = [m.name for m in context.models.values()]
+        sqlmesh_models = {model for model in sqlmesh_models if model[0] not in existing_models}
+
+    if sqlmesh_models:
+        _create_models(models_path=context.path / "models", models=sqlmesh_models)
+        model_names = "\n".join([f"- {model[0]}" for model in sqlmesh_models])
+        return f"Updated SQLMesh models:\n{model_names}"
```
izeigerman, line 110: I think the better API would be returning a set of model names here and then handling actual printing upstream. Alternatively, you can just print into console here directly and not return anything.

Themiscodes: Right I’ll refactor this to return the model names instead.

sqlmesh/cli/main.py
Themiscodes: I have it like that because I have another minor update in progress that will work alongside this `dlt` command but with a different option. But I agree it looks unintuitive so I'm happy to adjust it in the meantime, if it seems redundant

Themiscodes:
Thank you for the review @izeigerman addressed comments and revised command names

izeigerman (approved):
Nice 👍

PR #3342, opened 2024-11-07
Feat(clickhouse): add clickhouse dbt adapter

treysp:


sqlmesh/dbt/model.py
```diff
@@ -423,6 +439,80 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
+                    # dictionary materialization
+                    raise ConfigError(msg)
+                else:
+                    logger.warning(f"{msg} Falling back to the '{fallback[1]}' materialization.")
+                    self.materialized = fallback[1]
+
+            if self.model_materialization == Materialization.INCREMENTAL:
+                if self.incremental_strategy in ["delete+insert", "append", "insert_overwrite"]:
+                    logger.warning(
+                        f"The '{self.incremental_strategy}' incremental strategy is not supported - SQLMesh uses the temp table/partition swap strategy for all incremental models."
```
izeigerman, line 462: How about the `INCREMENTAL_UNMANAGED` type? Can't we support `append` and `insert_overwrite` with it?

treysp: Excellent point

sqlmesh/dbt/model.py
```diff
@@ -373,13 +384,19 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
-                if isinstance(self.partition_by, list)
-                else self._big_query_partition_by_expr(context)
-            )
+            partitioned_by = []
+            if isinstance(self.partition_by, list):
+                for p in self.partition_by:
+                    try:
+                        partitioned_by.append(d.parse_one(p, dialect=model_dialect))
+                    except SqlglotError as e:
+                        raise ConfigError(f"Failed to parse partition_by field '{p}': {e}") from e
```
izeigerman, line 396: Let's include the path / model this error came from

sqlmesh/dbt/model.py
```diff
@@ -423,6 +439,81 @@ def to_sqlmesh(self, context: DbtContext) -> Model:
...
+            ]
+            if self.materialized in unsupported_materializations:
+                fallback = self.materialized.split("_")
+                msg = f"SQLMesh does not support the '{self.materialized}' model materialization."
+                if len(fallback) == 1:
+                    # dictionary materialization
+                    raise ConfigError(msg)
+                else:
+                    logger.warning(f"{msg} Falling back to the '{fallback[1]}' materialization.")
+                    self.materialized = fallback[1]
```
izeigerman, line 457: Not sure I like the idea of mutating this object during the conversion.

sqlmesh/dbt/target.py
```diff
@@ -860,6 +863,97 @@ def to_sqlmesh(self, **kwargs: t.Any) -> ConnectionConfig:
...
+    allow_automatic_deduplication: bool = False
+    tcp_keepalive: t.Union[bool, t.Tuple[int, ...], t.List[int]] = False
+    database: str = ""
+    local_suffix: str = "local"
+    local_db_prefix: str = ""
+
+    type: Literal["clickhouse"] = "clickhouse"
+
+    def default_incremental_strategy(self, kind: IncrementalKind) -> str:
+        return "legacy"
```
izeigerman, line 927: What's `legacy`?

sqlmesh/dbt/model.py
treysp: Agreed, but we can't wait until later because we call the `materialization` property that passes the `materialized` string to an enum on the next line.

Should I add a materialized validator and do it there?

sqlmesh/dbt/target.py
treysp: It's the dbt-clickhouse version of temp table swap.

I wasn't sure whether this method should mimic what dbt-clickhouse would return or our sqlmesh interpretation/version.

izeigerman: I see, thanks

izeigerman: I don't see us handling it in the conversion code. Is it expected?

sqlmesh/dbt/model.py
izeigerman: Validator makes sense to me. Otherwise `model_materialization` is not usable at all with clickhouse

treysp:
Feedback addressed and ready for you to take another pass - thanks!

izeigerman (approved):

PR #3346, opened 2024-11-08
suggestion: cleaner errors on failure

Ben-Epstein:
I've recently started using sqlmesh and am really enjoying it! One place I've found it to be a bit difficult is parsing error messages. I've noticed that the error is frequently in the middle of a lot of other logs and it's been a bit verbose. 

This is by no means a PR to merge, but rather a suggestion/discussion opener for a potentially cleaner error log output. 

Here is the current state when I write a bug, for example yielding None
image
I zoomed out so you can see the full trace.

Here it is with the double logs commented out
image

Or something even more extreme, just to show the point
image

We could also pick non-red to make it even more obvious, just wanted to start the discussion.

Thanks for the awesome tool!

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -371,12 +371,12 @@ class Scheduler:
         for error in errors:
             if isinstance(error.__cause__, CircuitBreakerError):
                 raise error.__cause__
-            sid = error.node[0]
-            formatted_exception = "".join(format_exception(error.__cause__ or error))
-            log_message = f"FAILED processing snapshot {sid}\n{formatted_exception}"
-            self.console.log_error(log_message)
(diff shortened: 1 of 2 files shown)


treysp:
Thanks for raising this. The error traces can definitely be hard to unwind, especially when the cause is a database error like invalid sql.

We've talked about how to address this, but since we can be passing errors up the stack from so many different places we didn't identify a clear way to ensure we're only suppressing noise (and not the actual error you're looking for).

Ben-Epstein:
@treysp that makes sense. Is it worth experimenting with perhaps the highlighting I showed in the final screenshot? Then we can monitor how users interact with errors over time and get a clearer understanding of how much "noise" is being logged

PR #3356, opened 2024-11-10
feat: add connection info to  command on the CLI

Kayrnt:
# Description

Adds connection details to help users to ensure that the project is configured as expected.
The output looks like the following:
```
sqlmesh -p ./sqlmesh_project info
Models: 1
Macros: 0


Connection:
  concurrent_tasks: 1
  execution_project: my-great-project
  job_retries: 1
  keyfile: '****'
  location: US
  method: oauth
  pre_ping: false
  priority: interactive
  project: my-great-project
  register_comments: true
  scopes:
  - https://www.googleapis.com/auth/bigquery
  type: bigquery




Test Connection:
  concurrent_tasks: 1
  execution_project: my-great-project
  job_retries: 1
  keyfile: '****'
  location: US
  method: oauth
  pre_ping: false
  priority: interactive
  project: my-great-project
  register_comments: true
  scopes:
  - https://www.googleapis.com/auth/bigquery
  type: bigquery




State Connection:
  concurrent_tasks: 1
  connector_config: {}
  database: local_state.db
  extensions: []
  pre_ping: false
  register_comments: true
  type: duckdb


Data warehouse connection succeeded
State backend connection succeeded
```


sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
                 strict=strict,
             )
 
+    def get_sensitive_fields(self) -> Set[str]:
+        """Returns a set of common sensitive field names"""
+        return {
```
izeigerman, line 1853: There's already a similar list here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/config/scheduler.py#L110-L125. Can we unify them?

sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
                 strict=strict,
             )
 
+    def get_sensitive_fields(self) -> Set[str]:
```
izeigerman, line 1851: Shouldn't be a method

sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
...
+            "access_token",
+            "refresh_token",
+            "auth_token",
+            "credentials",
+            "client_secret",
+            "certificate",
+            "ssh_key",
+        }
+
+    def is_sensitive_field(self, field_name: str, sensitive_fields: Set[str]) -> bool:
```
izeigerman, line 1869: Shouldn't be a method. Let's move this to utils along with the list of sensitive fields.

sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
...
+        """
+        Check if a field name contains any sensitive keywords
+        """
+        field_lower = field_name.lower()
+        return any(sensitive in field_lower for sensitive in sensitive_fields)
+
+    def mask_sensitive_value(self, value: Any) -> str:
```
izeigerman, line 1876: Ditto

sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
...
+        """
+        Mask sensitive values with a placeholder
+        Returns '****' for non-empty values and '' for empty ones
+        """
+        if value and str(value).strip():
+            return "*" * len(str(value))
+        return "None"
+
+    def print_config(self, config: ConnectionConfig | None, console: Any) -> None:
```
izeigerman, line 1885: There's already `self` available, why pass `console` as an argument? Also let's make this private (prefix with `_`). We should be careful about adding public methods to the Context since it's a gateway into SQLMesh Python API.

sqlmesh/core/context.py
```diff
@@ -1846,15 +1848,77 @@ def create_external_models(self, strict: bool = False) -> None:
...
+            else:
+                console.log_status_update(f"{field_name}: {value}")
+
     @python_api_analytics
     def print_info(self, skip_connection: bool = False) -> None:
         """Prints information about connections, models, macros, etc. to the console."""
         self.console.log_status_update(f"Models: {len(self.models)}")
         self.console.log_status_update(f"Macros: {len(self._macros) - len(macro.get_registry())}")
 
+        self.console.log_status_update("\n\nConnection:")
```
izeigerman, line 1911: Should we make the headers more pronounced? Otherwise they blend in with individual connection attributes.

sqlmesh/core/context.py
izeigerman: Could also be a candidate for utils

sqlmesh/core/context.py
Kayrnt: Great idea! Thanks for pointing out!

sqlmesh/core/context.py
Kayrnt: Should it be a just a value then?

sqlmesh/core/context.py
Kayrnt: Ok, do you prefer that I create a dedicated utils file for that?

sqlmesh/core/context.py
Kayrnt: How would you do that?

Kayrnt:
@izeigerman, I moved the code but I didn't make any change yet regarding the template for the info output.

sqlmesh/utils/info.py
```diff
@@ -0,0 +1,71 @@
+from typing import Any, Optional, Set
```
izeigerman, line 1: Let's call make this module more descriptive by calling it `config.py` since everything here is related to configuration.

sqlmesh/core/context.py
izeigerman: You can make it bold by adding appropriate tags, eg:
```
self.console.log_status_update("\n\n[bold]Connection:[/bold]")
```
Ideally we also want to indent everything that follows under the configuration. This way we may not even need to do the highlighting. 

Actually, can we just put this into dict and serialize as YAML instead? Eg.:
```
config = {"Test Connection": { ... }}
```

sqlmesh/utils/info.py
Kayrnt: Alright! I updated the PR.

sqlmesh/core/context.py
Kayrnt: Ok I can try that

sqlmesh/utils/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Args:
+        config: Pydantic model containing configuration
+        console: Console object with log_status_update method
+    """
+    if not config:
+        console.log_status_update("No connection configuration found.")
+        return
+
+    configDict = config.dict(mode="json")
```
izeigerman, line 67:
```suggestion
    config_dict = config.dict(mode="json")
```

sqlmesh/utils/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    for field_name in configDict:
```
izeigerman, line 69:
```suggestion
    for field_name in config_dict:
```

sqlmesh/utils/config.py
```diff
@@ -0,0 +1,77 @@
+from typing import Any, Optional, Set
+
+from sqlmesh.core.config.connection import ConnectionConfig
+import yaml
```
izeigerman, line 4: from sqlmesh.utils import yaml

setup.py
```diff
@@ -96,6 +96,7 @@
             "types-dateparser",
             "types-python-dateutil",
             "types-pytz",
+            "types-PyYAML",
```
izeigerman, line 99: This is not needed. We use ruamel in SQLmesh

Kayrnt: Ok, I guess it was only needed because of the import

izeigerman (approved):

github-christophe-oudar:
Thank you for the review and merge 🎉

PR #3357, opened 2024-11-11
Fix(clickhouse): support non-datetime time column partitioning

treysp:
The Clickhouse adapter automatically partitions incremental by time models if the time column is not included in any partitioning expression.

Clickhouse strongly recommends against fine-grained/small partitions. Therefore, the automatic partitioning "floors" the time column to weekly granularity with the `toMonday()` function.

`toMonday()` only accepts date/datetime input types, so we currently error if the time column is string type.

This PR allows non-date/datetime time columns by:
- If time column type known and date/datetime, pass column directly
- If time column type known and NOT date/datetime, cast time column to `DateTime64` before passing to `toMonday()` then cast output back to original time column type
- If time column type not known, cast time column to `DateTime64` before passing to `toMonday()`

### Implementation note

This PR moves the `partitioned_by` property from the `ModelMeta` to `_Model` class so it can access the `_Model` `column_to_types` property.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2473,11 +2473,13 @@ def get_model_name(path: Path) -> str:
 
 # function applied to time column when automatically used for partitioning in INCREMENTAL_BY_TIME_RANGE models
 def clickhouse_partition_func(
-    column: exp.Expression, columns_to_types: t.Dict[str, exp.DataType]
+    column: exp.Expression, columns_to_types: t.Optional[t.Dict[str, exp.DataType]]
 ) -> exp.Expression:
     # `toMonday()` function accepts a Date or DateTime type column
 
-    col_type = columns_to_types.get(column.name) or exp.DataType.build("UNKNOWN")
+    col_type = (columns_to_types and columns_to_types.get(column.name)) or exp.DataType.build(
+        "UNKNOWN"
+    )
     col_type_is_conformable = col_type.is_type(
         exp.DataType.Type.DATE,
         exp.DataType.Type.DATE32,
@@ -2487,15 +2489,14 @@ def clickhouse_partition_func(
 
     #  if input column is already a conformable type, just pass the column
     if col_type_is_conformable:
-        return exp.func("toMonday", column)
+        return exp.func("toMonday", column, dialect="clickhouse")
 
     # if input column type is not known, cast input to DateTime64
-    if not col_type_present or (
-        col_type_present and columns_to_types[column.name].is_type(exp.DataType.Type.UNKNOWN)
-    ):
+    if col_type.is_type(exp.DataType.Type.UNKNOWN):
         return exp.func(
             "toMonday",
             exp.cast(column, exp.DataType.build("DateTime64('UTC')", dialect="clickhouse")),
+            dialect="clickhouse",
         )
 
     # if input column type is known but not conformable, cast input to DateTime64 and cast output back to original type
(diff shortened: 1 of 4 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -2455,3 +2469,43 @@ def _meta_renderer(
...
+) -> exp.Expression:
+    # `toMonday()` function accepts a Date or DateTime type column
+
+    col_type_present = columns_to_types and column.name in columns_to_types
+    col_type_is_conformable = col_type_present and columns_to_types[column.name].is_type(
+        exp.DataType.Type.DATE,
+        exp.DataType.Type.DATE32,
+        exp.DataType.Type.DATETIME,
+        exp.DataType.Type.DATETIME64,
+    )
```
georgesittas, line 2486:
```suggestion
    col_type = columns_to_types.get(column.name) or exp.DataType.build("UNKNOWN")
    col_type_is_conformable = col_type.is_type(
        exp.DataType.Type.DATE,
        exp.DataType.Type.DATE32,
        exp.DataType.Type.DATETIME,
        exp.DataType.Type.DATETIME64,
    )
```

georgesittas: You can then use `col_type` directly in L2493-2494 and L2507.

sqlmesh/core/model/definition.py
```diff
@@ -2455,3 +2469,43 @@ def _meta_renderer(
...
+
+    #  if input column is already a conformable type, just pass the column
+    if col_type_is_conformable:
+        return exp.func("toMonday", column)
```
georgesittas, line 2490: I'd do `dialect="clickhouse"` here just to avoid any dialect-specific parsing shenanigans.

sqlmesh/core/model/definition.py
```diff
@@ -2455,3 +2469,43 @@ def _meta_renderer(
...
+            "toMonday",
+            exp.cast(column, exp.DataType.build("DateTime64('UTC')", dialect="clickhouse")),
+        )
+
+    # if input column type is known but not conformable, cast input to DateTime64 and cast output back to original type
+    return exp.cast(
+        exp.func(
+            "toMonday",
+            exp.cast(column, exp.DataType.build("DateTime64('UTC')", dialect="clickhouse")),
+        ),
```
georgesittas, line 2506: Ditto about `dialect="clickhouse"`.

sqlmesh/core/model/definition.py
```diff
@@ -2455,3 +2469,43 @@ def _meta_renderer(
...
+        columns_to_types[column.name],
+    )
+
+
+TIME_COL_PARTITION_FUNC = {"clickhouse": lambda x, y: clickhouse_partition_func(x, y)}
```
georgesittas, line 2511:
```suggestion
TIME_COL_PARTITION_FUNC = {"clickhouse": clickhouse_partition_func}
```

georgesittas (approved):
Nice, thanks for addressing comments 👍

swt30:
Hey @treysp what version of Clickhouse are you using? I am experimenting with sqlmesh + clickhouse and finding that this causes an error because `DateTime64('UTC')` needs `scale` as the first parameter, at least in the latest Clickhouse. I'm new to Clickhouse so perhaps I'm missing something here? 

Adding a scale of 6 [here](https://github.com/TobikoData/sqlmesh/pull/3357/files#diff-87d4f5b42e1c683ecf7db16c0c8bbb8c4dcd45b039cae1e87fbf863b598cd223R2498) and just below fixes the problem. But the choice of 6 feels a bit arbitrary. What if the column were a nanosecond-precision string-formatted timestamp?

treysp:
Thanks for reporting - that looks like a bug. 

Precision is essentially arbitrary, so my inclination would be to go for nanosecond to prevent truncation. Do you see any issues with that? Pandas handles nanosecond timestamps, so conversion to Python should be ok.

treysp:
Addressing this in #3582

Issue #3360, opened 2024-11-12
sql model seems to call macro even when `@IF` conditional is false


Ben-Epstein:
I have the following macro

```py
from sqlmesh import macro
from sqlmesh.core.macros import MacroEvaluator

@macro(metadata_only=True)
def test_macro(evaluator: MacroEvaluator):
    print("EVAL STAGE", evaluator.runtime_stage)
    return "1"
```

and the following sql model
```sql
MODEL (
    name sqlmesh_example.foo,
    kind INCREMENTAL_BY_TIME_RANGE(
        time_column ts
    ),
    cron '@daily',
    grain (id, ts),
    table_format "delta",
  );

  SELECT id, ts, mean(value)  mean_value
  FROM delta_scan('./ext_table1')
  GROUP BY id, ts
  ORDER BY id, ts;

  @IF(
    @runtime_stage = 'evaluating',
    @test_macro()
  );
```

When I run any command, for example `sqlmesh fetchdf 'select * from sqlmesh_example.foo'`, I see the print
```
EVAL STAGE loading
```

It seems like bug to me, but am I missing something or making a very silly mistake?

Thanks!

z3z1ma:
I don't think SQLMesh has short circuiting during the rendering phase where these macros are expanded. I think it could be hand rolled for @IF but would not be trivial (I think). 

The alternative currently is to move the if statement to the body of the macro in Python.

tobymao:
the if statement is only used for the return value, so test_macro() is called, but it won't be returned as a part of the sql. i'd just move the if statement to the macro as @z3z1ma suggested

Ben-Epstein:
@tobymao can you update the documentation [here](https://sqlmesh.readthedocs.io/en/stable/concepts/models/sql_models/#optional-prepost-statements) to reflect that? This section implies that the macro shouldn't run

<img width="726" alt="image" src="https://github.com/user-attachments/assets/b5162b5c-1243-4b73-9569-d07aba1832a2">

georgesittas:
I think this section is fine. The statement _does_ run only once, it won't be sent to the engine at load time. I can see how there's some ambiguity, but I believe we should address it by saying that `@IF` doesn't short-circuit, as Alex mentioned.

PR #3361, opened 2024-11-12
Chore: update IF macro docs to say that it does not short-circuit

georgesittas:
Context: https://github.com/TobikoData/sqlmesh/issues/3360#issuecomment-2470524434


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -498,6 +498,11 @@ SELECT
 FROM table
 ```
 
+!!! note
+
+    The `@IF` macro does not short-circuit based on the condition's value; its arguments are always rendered. This
```
treysp, line 503: Hmm...this is a tough one - might be too much to fit into a note. Wdyt?

Possible text:

SQLMesh macro arguments can contain placeholder values, such as local variables. 

Before a macro can evaluated, SQLMesh must substitute the actual values into these placeholders - that process is called "rendering." Rendering is always the first step in SQLMesh's macro processing.

Rendering occurs **before** the `@IF` condition is evaluated. For example, SQLMesh cannot determine whether `my_column > @my_value` until it has first substituted the number `@my_value` represents.

Your macro might do things besides returning a value, such as printing a message or executing a statement (i.e., the macro "has side effects"). The side effect code will always run during the rendering step. To prevent, modify the macro code to condition the side effects on the evaluation stage.

georgesittas: Fair enough, I think I over-condensed it. I'd keep it shorter, though. Don't we talk about rendering elsewhere within the SQLMesh macros page? I think I'd keep the last two paragraphs in your suggestion.

treysp: Good call on the last two paragraphs. How about this?

----------

[Macro rendering](#sqlmesh-macro-approach) occurs before the `@IF` condition is evaluated. For example, SQLMesh cannot determine whether `my_column > @my_value` until it has first substituted the number `@my_value` represents.

Your macro might do things besides returning a value, such as printing a message or executing a statement (i.e., the macro "has side effects"). The side effect code will always run during the rendering step. To prevent this, modify the macro code to condition the side effects on the evaluation stage.

PR #3367, opened 2024-11-13
Feat: include dropped column name in destructive change message

treysp:


georgesittas (approved):

sqlmesh/core/schema_diff.py
```diff
@@ -687,5 +687,14 @@ def has_drop_alteration(alter_expressions: t.List[exp.Alter]) -> bool:
     )
 
 
+def get_dropped_column_name(alter_expressions: t.List[exp.Alter]) -> t.Optional[str]:
```
izeigerman, line 690: Should we print all impacted columns at once?

treysp: Definitely, I should have thought of that

izeigerman (approved):
thanks!

PR #3370, opened 2024-11-13
Fix: Normalise column names in schema differ based on engine

Themiscodes:
This update aims to take account of case sensitivity of columns by storing the engine's dialect in the `SchemaDiffer` and normalising the column names based on it during the comparison of the schemas.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -118,6 +118,7 @@ class EngineAdapter:
         **kwargs: t.Any,
     ):
         self.dialect = dialect.lower() or self.DIALECT
+        self.SCHEMA_DIFFER.dialect = self.dialect
         self._connection_pool = create_connection_pool(
             connection_factory, multithreaded, cursor_kwargs=cursor_kwargs, cursor_init=cursor_init
         )
--- a/sqlmesh/core/schema_diff.py
+++ b/sqlmesh/core/schema_diff.py
@@ -298,6 +298,7 @@ class SchemaDiffer(PydanticModel):
     1. Support column moves. Databricks Delta supports moves and would allow exact matches.
 
     Args:
+        dialect: The dialect of the engine to normalise the columns accordingly to the case sensitivity.
         support_positional_add: Whether the engine for which the diff is being computed supports adding columns in a
             specific position in the set of existing columns.
         support_nested_operations: Whether the engine for which the diff is being computed supports modifications to
--- a/sqlmesh/core/schema_diff.py
+++ b/sqlmesh/core/schema_diff.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from enum import Enum, auto
 from sqlglot import exp
 from sqlglot.helper import ensure_list, seq_get
+from sqlglot.optimizer.normalize_identifiers import normalize_identifiers
 
 from sqlmesh.utils import columns_to_types_to_struct
 from sqlmesh.utils.pydantic import PydanticModel
@@ -321,6 +322,7 @@ class SchemaDiffer(PydanticModel):
             parameterized type can ALTER to its unlimited length version, along with different types in some engines.
     """
 
+    dialect: str = ""
     support_positional_add: bool = False
     support_nested_operations: bool = False
     support_nested_drop: bool = False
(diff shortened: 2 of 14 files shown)


tests/core/test_schema_diff.py
```diff
@@ -45,6 +45,44 @@ def test_schema_diff_calculate():
...
+    [
+        ("", []),
+        (
+            "mysql",
+            [
+                "ALTER TABLE apply_to_table DROP COLUMN id",
+                "ALTER TABLE apply_to_table ADD COLUMN Id INT",
+            ],
+        ),
+        ("duckdb", []),
```
izeigerman, line 59: Can we add bigquery too please

izeigerman (approved):

sqlmesh/core/engine_adapter/base.py
```diff
@@ -118,6 +118,7 @@ def __init__(
         **kwargs: t.Any,
     ):
         self.dialect = dialect.lower() or self.DIALECT
+        self.SCHEMA_DIFFER.dialect = self.dialect
```
georgesittas, line 121: Is this actually needed?

Themiscodes: I was thinking initially it was needed since the `SchemaDiffer` is initialised without a dialect, but you're right can remove it

sqlmesh/core/schema_diff.py
```diff
@@ -655,7 +658,10 @@ def compare_structs(
         """
         return [
             op.expression(table_name, self.array_element_selector)
-            for op in self._from_structs(current, new)
+            for op in self._from_structs(
+                normalize_identifiers(current, dialect=self.dialect),
+                normalize_identifiers(new, dialect=self.dialect),
```
georgesittas, line 663: I'm skeptical about this section, I think we can end up normalizing identifiers twice: `current` and `new` are built off of columns-to-types mappings [here](https://github.com/TobikoData/sqlmesh/blob/63e2dbd54430ca16682749ca1c2e5c5f71312303/sqlmesh/core/schema_diff.py#L677-L679). For reference, [here](https://github.com/TobikoData/sqlmesh/blob/63e2dbd54430ca16682749ca1c2e5c5f71312303/sqlmesh/core/plan/builder.py#L454-L464) is one place where `compare_columns` is called. The problem I'm seeing is that the `columns_to_types` property in models doesn't preserve quotes, it simply stores their normalized names.

For example, if I change the example project's seed model to:

```sql
MODEL (
    kind SEED (
      path '../seeds/seed_data.csv'
    ),
    columns (
      id INTEGER,
      "ItemId" INTEGER,
      event_date DATE
    ),
    grain (id, event_date)
  );
```

Then observe how `columns_to_types` looks:

```python
>>> ctx.models['"db"."sqlmesh_example"."seed_model"'].columns_to_types
{'id': DataType(this=Type.INT, nested=False), 'ItemId': DataType(this=Type.INT, nested=False), 'event_date': DataType(this=Type.DATE, nested=False)}
```

Notice how there are no quotes around `ItemId` in the mapping. If I'm reading the control flow correctly, you'll force lowercase/uppercase names that are already normalized here, even when you shouldn't, which can cause its own problems.

Let me know if I'm missing something, but I'd double check this before merging, and make sure to test this scenario.

sqlmesh/core/engine_adapter/base.py
izeigerman: I think it's safer to have it? If any downstream adapter doesn't override it

georgesittas: I see, fair enough. Do we need to set the dialect in each adapter then? If we choose to mutate `dialect` here, perhaps doing `SchemaDiffer(dialect=DIALECT)` elsewhere is redundant.

Themiscodes: I was thinking to be on the safe side and cover a case of a SchemaDiffer used by getting it directly of a specific adapter, for example in some tests:
https://github.com/TobikoData/sqlmesh/blob/63e2dbd54430ca16682749ca1c2e5c5f71312303/tests/core/test_plan.py#L1312

sqlmesh/core/schema_diff.py
Themiscodes: Yes you're right I tested with various engined and the columns_to_types are already normalised based on case sensitivity. So using `normalize_identifiers` is incorrectly forcing normalization when it shouldn't here. Also, in the other section that `compare_columns` is used: 

https://github.com/TobikoData/sqlmesh/blob/63e2dbd54430ca16682749ca1c2e5c5f71312303/sqlmesh/core/engine_adapter/base.py#L904-L907

the columns are fetched from the database, which again include the correct case sensitivity (so if a column for example "ItemId" is quoted, it will be retrieved as ItemId). So the SchemaDiffer does account for case sensitivity, operating under the assumption that identifiers are already correctly normalized when the columns are passed to `compare_columns`

Issue #3372, opened 2024-11-14
--paths flag doesn't support wildcards


Harmuth94:
Currently, the sqlmesh plan command requires explicit specification of each project path using the `-p` flag. This limitation poses challenges for us managing multiple SQLMesh projects (one for each source system) with interdependencies, including downstream aggregation of sources, shared utility projects containing macros and tests.

## Solution
Introduce support for wildcard patterns in the `-p` flag of the sqlmesh plan command. This enhancement would allow users to specify multiple projects efficiently. For example, using `-p all_projects/*` would include all projects within the `all_projects` directory.

georgesittas:
Fixed by linked PR.

PR #3373, opened 2024-11-14
add wildcard

Harmuth94:
closing #3372

### DIFF
--- a/sqlmesh/core/config/loader.py
+++ b/sqlmesh/core/config/loader.py
@@ -27,7 +27,7 @@ def load_configs(
     config = config or "config"
 
     absolute_paths = [
-        Path(t.cast(t.Union[str, Path], path)).absolute() for path in ensure_list(paths)
+        p.absolute() for path in ensure_list(paths) for p in Path().glob(str(t.cast(t.Union[str, Path], path)))
     ]
 
     if not isinstance(config, str):


Harmuth94:
To use wildcards enclose path in '' or ""

sqlmesh/core/config/loader.py
```diff
@@ -27,7 +28,9 @@ def load_configs(
     config = config or "config"
 
     absolute_paths = [
-        Path(t.cast(t.Union[str, Path], path)).absolute() for path in ensure_list(paths)
+        Path(t.cast(t.Union[str, Path], p)).absolute()
+        for path in ensure_list(paths)
+        for p in (glob.glob(str(path)) or [str(path)])
```
Harmuth94, line 33: `glob.glob()` returns `None` for absolute paths

Harmuth94: Did some digging and it was not the absolute paths.

The issue is that glob.glob returns an empty list if the pattern doesn't match anything (This was the case for a few pytest cases). The `load_configs` function would return an unexpected empty dictionary and causing downstream issues. 

Adding `or [str(path)]` will the correct downstream invalidation of the path/config.

georgesittas: I see, seems reasonable.

georgesittas (approved):

erindru (approved):

Issue #3381, opened 2024-11-16
STAR macro not working with Snowflake dialect


mpcarter:
When using the `@STAR` macro with a snowflake model, it mismatches the case of the quoted alias used for the table (relation) and those used in the expansion of the columns. 

```sql
MODEL (
    name sqlmesh_example.star_model
);

SELECT
@STAR(sqlmesh_example.incremental_model)
from sqlmesh_example.incremental_model
```

Using snowflake dialect it render as
```sql
SELECT
  CAST("incremental_model"."ID" AS INT) AS "ID",
  CAST("incremental_model"."ITEM_ID" AS INT) AS "ITEM_ID",
  CAST("incremental_model"."EVENT_DATE" AS DATE) AS "EVENT_DATE"
FROM "SQLMESH"."SQLMESH_EXAMPLE"."INCREMENTAL_MODEL" AS "INCREMENTAL_MODEL"
```
The table alias is `"INCREMENTAL_MODEL"`, but the table alias used with the columns is lowercase and quoted, `"incremental_model"`. So it fails to deploy.

Using duckdb it renders as
```sql
SELECT
  CAST("incremental_model"."id" AS INT) AS "id",
  CAST("incremental_model"."item_id" AS INT) AS "item_id",
  CAST("incremental_model"."event_date" AS DATE) AS "event_date"
FROM "sqlmesh"."sqlmesh_example"."incremental_model" AS "incremental_model"
```

mpcarter:
I believe I found a fix. In the star macro, the `table_identifier` needs to be normalized.

```python
    excluded_names = {
        normalize_identifiers(excluded, dialect=evaluator.dialect).name
        for excluded in exclude.expressions
    }
    quoted = quote_identifiers.this
    table_identifier = normalize_identifiers(alias or relation, dialect=evaluator.dialect).name # use this instead
    # table_identifier = alias.name or relation.name

    columns_to_types = {
        k: v for k, v in evaluator.columns_to_types(relation).items() if k not in excluded_names
    }
```

I copied the code for `@STAR` into a user-defined macro to test it and it rendered my example like this for snowflake dialect

```sql
SELECT
  CAST("INCREMENTAL_MODEL"."ID" AS INT) AS "ID",
  CAST("INCREMENTAL_MODEL"."ITEM_ID" AS INT) AS "ITEM_ID",
  CAST("INCREMENTAL_MODEL"."EVENT_DATE" AS DATE) AS "EVENT_DATE"
[...]

PR #3382, opened 2024-11-16
Fix: normalize table identifier in star macro

mpcarter:
The star macro mismatches the case of the table alias when used with snowflake dialect.
The table alias will be uppercase while the alias used with columns remains lowercase.
Since the identifiers are all quoted this causes a problem for snowflake.

Solution is to apply normalization function to the table_identifier.

Related to https://github.com/TobikoData/sqlmesh/issues/3381


sqlmesh/core/macros.py
```diff
@@ -872,7 +872,7 @@ def star(
         for excluded in exclude.expressions or except_.expressions
     }
     quoted = quote_identifiers.this
-    table_identifier = alias.name or relation.name
+    table_identifier = normalize_identifiers(alias or relation, dialect=evaluator.dialect).name
```
georgesittas, line 875: I think you need to get rid of the `.name` here and it should work?

georgesittas (commented):
Thanks for the PR - let's please test this & make sure CI passes. Feel free to ping me if you need help debugging this.

sqlmesh/core/macros.py
georgesittas, line 875: Trying this out real quick...

```suggestion
    table_identifier = normalize_identifiers(alias or relation, dialect=evaluator.dialect)
```

georgesittas: Yeah, it looks like my suggestion wasn't right after all, sorry about that.

> Not sure how to run the actual test suite though.

Did you already try `make fast-test`? I think you can also do `make doc-test` if the failing test is embedded in a docstring.

mpcarter: I found the problem `alias or relation` was returning empty alias because even when alias not provided it is not None. So it needed `alias.name or relation.name` along with `normalize_identifiers().name`. 

Attempted to add a test too. Ran test_macros.py with success.

georgesittas (approved):
LGTM, thanks for the fix @mpcarter! Should be good to go once you run `make style` to fix CI.

sqlmesh/core/macros.py
```diff
@@ -872,7 +872,7 @@ def star(
         for excluded in exclude.expressions or except_.expressions
     }
     quoted = quote_identifiers.this
-    table_identifier = alias.name or relation.name
+    table_identifier = normalize_identifiers(alias.name or relation.name, dialect=evaluator.dialect).name
```
georgesittas, line 875: Actually, I thought about this again and we do need to refactor this a bit. If a user passes a quoted name as an arg, you'll need to preserve them in the objects passed to the `normalize_identifiers` call, otherwise you risk normalizing a case-sensitive identifier.

I think this is what we should do instead:

```python
    table_identifier = normalize_identifiers(
        alias if alias.name else relation, dialect=evaluator.dialect
    ).name
```

mpcarter: Made the suggested change and added a couple more test cases to cover those scenarios. Seems to be working!

I put this in draft because I was thinking that might avoid the CI checks for now. Was having a few hurdles getting the `make style` to run. But it didn't.

If it doesn't pass CI this time, I will get that fixed later today. Appreciate the guidance!

mpcarter:
Finally got the styles to run, and checks passed.

georgesittas (approved):

PR #3383, opened 2024-11-16
Refactor!: support complex expressions in clustered_by property

georgesittas:
Snowflake supports [complex expressions](https://docs.snowflake.com/en/user-guide/tables-clustering-keys#defining-a-clustering-key-for-a-table) for its clustering keys, e.g.:

```
CREATE OR REPLACE TABLE t2 (c1 timestamp, c2 STRING, c3 NUMBER) CLUSTER BY (TO_DATE(C1), substring(c2, 0, 10));
```

Today, we're only expecting columns. This PR changes that, so now `clustered_by` is treated similar to `partitioned_by`.


georgesittas:
@izeigerman does it look like we'll need a migration script for this? If a sqlglot bump suffices, I've opened a [PR](https://github.com/TobikoData/sqlmesh/pull/3384) for that.

tobymao (approved):

sqlmesh/core/model/meta.py
```diff
@@ -348,13 +343,13 @@ def _root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
...
         if kind:
-            for field in ("partitioned_by_", "clustered_by"):
+            for field in ("partitioned_by_", "clustered_by_"):
                 if (
                     values.get(field)
                     and not kind.is_materialized
                     and not (kind.is_view and kind.materialized)
                 ):
-                    raise ValueError(f"{field} field cannot be set for {kind} models")
+                    raise ValueError(f"{field[:-1]} field cannot be set for {kind} models")
```
izeigerman, line 352: Why -1?

izeigerman (approved):

PR #3385, opened 2024-11-16
Feat: expose a resolve_table helper in Jinja to get physical model names

georgesittas:
Check out the added test to see what this looks like.

Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1731683617203209


tobymao:
do we have a similar method in the evaluator already?

georgesittas:
We have this: https://github.com/TobikoData/sqlmesh/blob/2b9d1c0d700df0efc481fc3ced97328b29a388f2/sqlmesh/core/renderer.py#L162-L171

It's not quite the same though, in Jinja you want to be able to resolve specific table names, most likely because a statement is not parsable -- if it was, the resolution would happen automatically.

erindru (approved):
+1 for escape hatches. I think this will also help with when we suggest Jinja as a solution in situations where SQLGlot cant understand the DDL statements natively and just parses them into `exp.Command` (which means normal substitution wont occur)

georgesittas:
Exactly, that's what motivated this PR in the first place.

PR #3389, opened 2024-11-18
Refactor: make resolve_table consistent across evaluation contexts

georgesittas:
This is a followup to https://github.com/TobikoData/sqlmesh/commit/6a324e7ea3bba49b7e1d7feb7e3814683bb65e47. Until now, we had:

- `ExecutionContext.table` -- resolve a model's name into its physical table
- `MacroEvaluator.resolve_tables` -- same, but this was mostly used to resolved all references within query (it transforms the whole thing)
- A `resolve_table` function registered in the Jinja environment, which can be used on a single table name

This PR exposes the `resolve_table` helper to the macro evaluator and the execution context, so that all of these execution environments support it, while also deprecating `Context.table`.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -104,17 +104,12 @@ class BaseExpressionRenderer:
             return self._cache
 
         if self._model_fqn and "this_model" not in kwargs:
-            kwargs["this_model"] = exp.to_table(
-                self._to_table_mapping(
-                    (
-                        [snapshots[self._model_fqn]]
-                        if snapshots and self._model_fqn in snapshots
-                        else []
-                    ),
-                    deployability_index,
-                ).get(self._model_fqn, self._model_fqn),
-                dialect=self._dialect,
-            ).sql(dialect=self._dialect, identify=True)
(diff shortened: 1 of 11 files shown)


sqlmesh/core/renderer.py
```diff
@@ -168,6 +160,10 @@ def _render(
             jinja_env=jinja_env,
             schema=self.schema,
             runtime_stage=runtime_stage,
+            resolve_table=t.cast(
+                t.Callable[[str | exp.Expression], str],
+                jinja_env.globals["resolve_table"],
```
izeigerman, line 165: Why are we doing this? I think we should just pass `self._resolve_table` directly

georgesittas: Because the lambda in the jinja env closes around `table_mapping` as well which is not currently propagated to the macro evaluator. We do something similar for `resolve_tables=lambda e: ...` in the line below this.

izeigerman: Yeah, my question was rather whether we should do the same lambda thing here instead of assuming that something is available in the jinja env.

izeigerman (approved):

PR #3393, opened 2024-11-19
Feat!: Support pinning of the physical table version for forward-only models

izeigerman:
This update introduces the following improvements:
* When a new forward-only model is added, and multiple versions of it are created concurrently without either being promoted to production, we ensure that both versions share the same physical table. This ensures that the name of the physical table remains deterministic and predictable, which is especially important when the model's data is manually backfilled outside of SQLMesh.
* Added the ability to override and pin the version of the physical table, but only for forward-only models. This is the ultimate way for users to ensure that a model's physical table remains unchanged.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -798,8 +798,15 @@ class _Model(ModelMeta, frozen=True):
             # TODO: would this sort of logic be better off moved into the Kind?
             if self.dialect == "snowflake" and "target_lag" not in self.physical_properties:
                 raise_config_error(
-                    "Snowflake managed tables must specify the 'target_lag' physical property"
+                    "Snowflake managed tables must specify the 'target_lag' physical property",
+                    self._path,
                 )
+
+        if self.physical_version is not None and not self.forward_only:
+            raise_config_error(
+                "Pinning a physical version is only supported for forward only models",
+                self._path,
+            )
 
     def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
         """Determines whether this model is a breaking change in relation to the `previous` model.
@@ -839,6 +846,7 @@ class _Model(ModelMeta, frozen=True):
             *(gen(expr) for expr in (self.clustered_by or [])),
             self.stamp,
             self.physical_schema,
+            self.physical_version,
             self.interval_unit.value if self.interval_unit is not None else None,
         ]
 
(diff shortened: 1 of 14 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -839,6 +846,7 @@ def _data_hash_values(self) -> t.List[str]:
             *(gen(expr) for expr in (self.clustered_by or [])),
             self.stamp,
             self.physical_schema,
+            self.physical_version,
```
tobymao, line 849: this makes it a breaking change "!"

sqlmesh/core/snapshot/definition.py
```diff
@@ -928,6 +931,9 @@ def categorize_as(self, category: SnapshotChangeCategory) -> None:
                     previous_version.data_version.temp_version
                     or previous_version.fingerprint.to_version()
                 )
+        elif self.is_model and self.model.forward_only and not self.previous_version:
+            # If this is a new model then use a deterministic version, independent of the fingerprint.
+            self.version = hash_data([self.name, *self.model.kind.data_hash_values])
```
tobymao, line 936: so the only way this would be problematic is if two people made new forward only models that aren't compatible

sqlmesh/core/model/definition.py
izeigerman: yes, thanks

sqlmesh/core/snapshot/definition.py
izeigerman: Depends on what you mean by being incompatible?

tobymao (approved):

sqlmesh/core/snapshot/definition.py
tobymao: i make a branch model A with column x

you make a branch with model b with column y

this now crashes right?

izeigerman: Yes, I've just thought of this too. It actually shouldn't, because the table migration should take care of this but it only kicks in for `FORWARD_ONLY` categories. However, both new model versions will be categorized as `BREAKING`. So I need to make sure that the migration is invoked, and we should be good even in this scenario.

Good catch.

izeigerman: Going to add an integration test for this

izeigerman: Added an integration test for this

PR #3394, opened 2024-11-19
Feat!: Add multiple engine project support

Themiscodes:
**WIP** : This is an update to enable the use of multiple engines within a single SQLMesh project. The engine for each model will be determined by the gateway specified in the model definition:

```sql
MODEL (
  name main_db.sales,
  kind FULL,
  gateway <gateway_name>
);
```

If no gateway is specified the default gateway will be used and the virtual layer will be using the default_gateway for the views.

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -123,6 +123,7 @@ class BuiltInSchedulerConfig(_EngineAdapterStateSyncSchedulerConfig, BaseConfig)
         return BuiltInPlanEvaluator(
             state_sync=context.state_sync,
             snapshot_evaluator=context.snapshot_evaluator,
+            snapshot_evaluators=context._snapshot_evaluators,
             create_scheduler=context.create_scheduler,
             default_catalog=self.get_default_catalog(context),
             console=context.console,
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -386,23 +386,24 @@ class GenericContext(BaseContext, t.Generic[C]):
             "test_configs"
         )
 
-        for gateway_name in self.config.gateways:
-            connection = self.config.get_connection(gateway_name)
-            adapter = (
-                connection.create_engine_adapter()
-                if gateway_name != self.config.default_gateway
-                else self._engine_adapter
-            )
-            evaluator = SnapshotEvaluator(
-                adapter.with_log_level(logging.INFO),
-                ddl_concurrent_tasks=connection.concurrent_tasks,
-            )
-            self._snapshot_evaluators[gateway_name] = evaluator
-            if self.config.default_gateway == gateway_name:
-                self._default_evaluator = evaluator
-            self._test_connection_configs[gateway_name] = self.config.get_test_connection(
-                gateway_name, adapter.default_catalog, default_catalog_dialect=adapter.DIALECT
-            )
+        if not gateway:
+            for gateway_name in self.config.gateways:
+                connection = self.config.get_connection(gateway_name)
+                adapter = (
+                    connection.create_engine_adapter()
+                    if gateway_name != self.config.default_gateway
+                    else self._engine_adapter
+                )
+                evaluator = SnapshotEvaluator(
+                    adapter.with_log_level(logging.INFO),
+                    ddl_concurrent_tasks=connection.concurrent_tasks,
+                )
+                self._snapshot_evaluators[gateway_name] = evaluator
+                if self.config.default_gateway == gateway_name:
+                    self._default_evaluator = evaluator
+                self._test_connection_configs[gateway_name] = self.config.get_test_connection(
+                    gateway_name, adapter.default_catalog, default_catalog_dialect=adapter.DIALECT
+                )
 
         self.console = console or get_console(dialect=self._engine_adapter.dialect)
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -61,6 +61,7 @@ from sqlmesh.core.config import (
     Config,
     load_configs,
 )
+from sqlmesh.core.config.connection import ConnectionConfig
 from sqlmesh.core.config.loader import C
 from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context_diff import ContextDiff
@@ -95,6 +96,7 @@ from sqlmesh.core.snapshot import (
     SnapshotFingerprint,
     to_table_mapping,
 )
+from sqlmesh.core.snapshot.definition import SnapshotTableCleanupTask
 from sqlmesh.core.state_sync import (
     CachingStateSync,
     StateReader,
@@ -379,22 +381,34 @@ class GenericContext(BaseContext, t.Generic[C]):
         self._snapshot_evaluators: UniqueKeyDict[str, SnapshotEvaluator] = UniqueKeyDict(
             "evaluators"
         )
+        self._default_evaluator: t.Optional[SnapshotEvaluator] = None
+        self._test_connection_configs: UniqueKeyDict[str, ConnectionConfig] = UniqueKeyDict(
+            "test_configs"
+        )
+
         for gateway_name in self.config.gateways:
             connection = self.config.get_connection(gateway_name)
-            adapter = connection.create_engine_adapter()
+            adapter = (
+                connection.create_engine_adapter()
+                if gateway_name != self.config.default_gateway
+                else self._engine_adapter
+            )
             evaluator = SnapshotEvaluator(
                 adapter.with_log_level(logging.INFO),
                 ddl_concurrent_tasks=connection.concurrent_tasks,
             )
             self._snapshot_evaluators[gateway_name] = evaluator
+            if self.config.default_gateway == gateway_name:
+                self._default_evaluator = evaluator
+            self._test_connection_configs[gateway_name] = self.config.get_test_connection(
+                gateway_name, adapter.default_catalog, default_catalog_dialect=adapter.DIALECT
+            )
 
         self.console = console or get_console(dialect=self._engine_adapter.dialect)
 
         self._test_connection_config = self.config.get_test_connection(
             self.gateway, self.default_catalog, default_catalog_dialect=self.engine_adapter.DIALECT
         )
-
-        self._snapshot_evaluator: t.Optional[SnapshotEvaluator] = None
 
         self._provided_state_sync: t.Optional[StateSync] = state_sync
         self._state_sync: t.Optional[StateSync] = None
@@ -427,12 +441,12 @@ class GenericContext(BaseContext, t.Generic[C]):
 
     @property
     def snapshot_evaluator(self) -> SnapshotEvaluator:
-        if not self._snapshot_evaluator:
-            self._snapshot_evaluator = SnapshotEvaluator(
+        if not self._default_evaluator:
+            self._default_evaluator = SnapshotEvaluator(
                 self.engine_adapter.with_log_level(logging.INFO),
                 ddl_concurrent_tasks=self.concurrent_tasks,
             )
-        return self._snapshot_evaluator
+        return self._default_evaluator
 
     def execution_context(
         self, deployability_index: t.Optional[DeployabilityIndex] = None
@@ -973,8 +987,8 @@ class GenericContext(BaseContext, t.Generic[C]):
             limit: A limit applied to the model.
         """
         snapshot = self.get_snapshot(model_or_snapshot, raise_if_missing=True)
-
-        df = self.snapshot_evaluator.evaluate_and_fetch(
+        evaluator = self._snapshot_evaluators.get(snapshot.model.gateway, self.snapshot_evaluator)
+        df = evaluator.evaluate_and_fetch(
             snapshot,
             start=start,
             end=end,
@@ -1642,11 +1656,13 @@ class GenericContext(BaseContext, t.Generic[C]):
         }
 
         try:
-            test_adapter = self._test_connection_config.create_engine_adapter(
-                register_comments_override=False
+            model_to_test = self.get_model(model, raise_if_missing=True)
+            connection_config = self._test_connection_configs.get(
+                model_to_test.gateway, self._test_connection_config
             )
+            test_adapter = connection_config.create_engine_adapter(register_comments_override=False)
             generate_test(
-                model=self.get_model(model, raise_if_missing=True),
+                model=model_to_test,
                 input_queries=input_queries,
                 models=self._models,
                 engine_adapter=self._engine_adapter,
@@ -1748,7 +1764,10 @@ class GenericContext(BaseContext, t.Generic[C]):
         errors = []
         skipped_count = 0
         for snapshot in snapshots:
-            for audit_result in self.snapshot_evaluator.audit(
+            evaluator = self._snapshot_evaluators.get(
+                snapshot.model.gateway, self.snapshot_evaluator
+            )
+            for audit_result in evaluator.audit(
                 snapshot=snapshot,
                 start=start,
                 end=end,
@@ -1893,8 +1912,9 @@ class GenericContext(BaseContext, t.Generic[C]):
 
     def close(self) -> None:
         """Releases all resources allocated by this context."""
-        if self._snapshot_evaluator:
-            self._snapshot_evaluator.close()
+        if self._snapshot_evaluators:
+            for evaluator in self._snapshot_evaluators.values():
+                evaluator.close()
         if self._state_sync:
             self._state_sync.close()
 
@@ -2106,9 +2126,18 @@ class GenericContext(BaseContext, t.Generic[C]):
     def _run_janitor(self, ignore_ttl: bool = False) -> None:
         self._cleanup_environments()
         expired_snapshots = self.state_sync.delete_expired_snapshots(ignore_ttl=ignore_ttl)
-        self.snapshot_evaluator.cleanup(
-            expired_snapshots, on_complete=self.console.update_cleanup_progress
-        )
+
+        gateways: t.Dict[str, t.List[SnapshotTableCleanupTask]] = {}
+        for snapshot_cleanup_task in expired_snapshots:
+            snapshot_name = snapshot_cleanup_task.snapshot.name
+            gateway = self.snapshots[snapshot_name].model.gateway or "default"
+            gateways.setdefault(gateway, []).append(snapshot_cleanup_task)
+
+        for gateway, snapshots_cleanup_tasks in gateways.items():
+            evaluator = self._snapshot_evaluators.get(gateway, self.snapshot_evaluator)
+            evaluator.cleanup(
+                snapshots_cleanup_tasks, on_complete=self.console.update_cleanup_progress
+            )
 
         self.state_sync.compact_intervals()
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -56,7 +56,11 @@ from sqlmesh.core import analytics
 from sqlmesh.core import constants as c
 from sqlmesh.core.analytics import python_api_analytics
 from sqlmesh.core.audit import Audit, ModelAudit, StandaloneAudit
-from sqlmesh.core.config import CategorizerConfig, Config, load_configs
+from sqlmesh.core.config import (
+    CategorizerConfig,
+    Config,
+    load_configs,
+)
 from sqlmesh.core.config.loader import C
 from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context_diff import ContextDiff
@@ -372,6 +376,18 @@ class GenericContext(BaseContext, t.Generic[C]):
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
         self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
+        self._snapshot_evaluators: UniqueKeyDict[str, SnapshotEvaluator] = UniqueKeyDict(
+            "evaluators"
+        )
+        for gateway_name in self.config.gateways:
+            connection = self.config.get_connection(gateway_name)
+            adapter = connection.create_engine_adapter()
+            evaluator = SnapshotEvaluator(
+                adapter.with_log_level(logging.INFO),
+                ddl_concurrent_tasks=connection.concurrent_tasks,
+            )
+            self._snapshot_evaluators[gateway_name] = evaluator
+
         self.console = console or get_console(dialect=self._engine_adapter.dialect)
 
         self._test_connection_config = self.config.get_test_connection(
@@ -516,6 +532,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             max_workers=self.concurrent_tasks,
             console=self.console,
             notification_target_manager=self.notification_target_manager,
+            snapshot_evaluators=self._snapshot_evaluators,
         )
 
     @property
--- a/sqlmesh/core/model/meta.py
+++ b/sqlmesh/core/model/meta.py
@@ -77,6 +77,7 @@ class ModelMeta(_Node):
     allow_partials: bool = False
     signals: t.List[exp.Tuple] = []
     enabled: bool = True
+    gateway: t.Optional[str] = None
 
     _bool_validator = bool_validator
     _model_kind_validator = model_kind_validator
@@ -160,6 +161,13 @@ class ModelMeta(_Node):
         # so this ensures they'll be stored as lowercase
         dialect = str_or_exp_to_str(v)
         return dialect and dialect.lower()
+
+    @field_validator("gateway", mode="before")
+    def _gateway_validator(cls, v: t.Any) -> t.Optional[str]:
+        if v is None:
+            return None
+        gateway = str_or_exp_to_str(v)
+        return gateway and gateway.lower()
 
     @field_validator("partitioned_by_", "clustered_by_", mode="before")
     @field_validator_v1_args
--- a/tests/core/test_context.py
+++ b/tests/core/test_context.py
@@ -313,6 +313,15 @@ def test_evaluate_limit():
     assert context.evaluate("without_limit", "2020-01-01", "2020-01-02", "2020-01-02", 2).size == 2
 
 
+def test_gateway_specific_evaluator(copy_to_temp_path):
+    path = copy_to_temp_path("examples/sushi")
+    ctx = Context(paths=path, config="isolated_systems_config", gateway="prod")
+    assert not ctx._snapshot_evaluators
+
+    ctx = Context(paths=path, config="isolated_systems_config")
+    assert len(ctx._snapshot_evaluators) == 3
+
+
 def test_multiple_gateways(tmp_path: Path):
     db_path = str(tmp_path / "db.db")
     gateways = {
@@ -335,9 +344,6 @@ def test_multiple_gateways(tmp_path: Path):
     assert gateway_model.gateway == "staging"
     context.upsert_model(gateway_model)
     assert context.evaluate("staging.stg_model", "2020-01-01", "2020-01-02", "2020-01-02").size == 5
-    assert (
-        context.evaluate("staging.stg_model", "2020-01-01", "2020-01-02", "2020-01-02", 2).size == 2
-    )
 
     default_model = load_sql_based_model(
         parse(
--- a/tests/core/test_context.py
+++ b/tests/core/test_context.py
@@ -14,6 +14,7 @@ from pytest_mock.plugin import MockerFixture
 from sqlglot import exp, parse_one
 from sqlglot.errors import SchemaError
 
+from sqlmesh.core.config.gateway import GatewayConfig
 import sqlmesh.core.constants
 import sqlmesh.core.dialect as d
 from sqlmesh.core.config import (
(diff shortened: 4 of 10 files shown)


sqlmesh/core/context.py
```diff
@@ -372,14 +378,39 @@ def __init__(
...
+
+        if not gateway:
+            for gateway_name in self.config.gateways:
+                connection = self.config.get_connection(gateway_name)
+                adapter = (
+                    connection.create_engine_adapter()
+                    if gateway_name != self.config.default_gateway
+                    else self._engine_adapter
+                )
+                evaluator = SnapshotEvaluator(
```
erindru, line 397: Hmm, I think we'd still want to manage all the snapshots within a single evaluator but direct them to the relevant engine_adapter's as required.

I can see why youve done this though. The problem here though is that the `engine_adapter` is an instance property of the `SnapshotEvaluator`, so the `SnapshotEvaluator` currently assumes a single gateway / engine.

However, I dont see anything engine specific within `SnapshotEvaluator`  so perhaps it could take a function that returns an EngineAdapter based on the supplied gateway name and use that to acquire the correct engine adapter when it needs to materialize a snapshot.

@izeigerman i'd be keen to hear your thoughts on this

sqlmesh/core/config/scheduler.py
```diff
@@ -123,6 +123,7 @@ def create_plan_evaluator(self, context: GenericContext) -> PlanEvaluator:
         return BuiltInPlanEvaluator(
             state_sync=context.state_sync,
             snapshot_evaluator=context.snapshot_evaluator,
```
tobymao, line 125: do you need snapshot_evaluator anymore? should snapshot_evaluators suffice?

sqlmesh/core/context.py
tobymao: i think it could be cleaner if there was only one snapshot evaluator with many adapters, because the rest of the code base only deals with one evaluator

sqlmesh/core/config/scheduler.py
tobymao: i don't think we should have multiple evaluators

sqlmesh/core/context.py
Themiscodes: right i will change then to use multiple engine adapters in a single evaluator. I went back and forth on the best approach, especially considering parts involving concurrency or areas that would require engine specific logic such as during schema creation that required more precise changes in the evaluator. but overall it does seem that using multiple engines would be the cleaner solution, so currently refactoring the code accordingly

sqlmesh/core/context.py
```diff
@@ -372,13 +381,35 @@ def __init__(
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
         self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
-        self.console = console or get_console(dialect=self._engine_adapter.dialect)
+        self._engine_adapters: UniqueKeyDict[str, EngineAdapter] = UniqueKeyDict("engine_adapters")
```
tobymao, line 384: you don't need a uniquekeydict, it's only for user inputs, from yamls

sqlmesh/core/context.py
```diff
@@ -1625,11 +1656,14 @@ def create_test(
         }
 
         try:
-            test_adapter = self._test_connection_config.create_engine_adapter(
-                register_comments_override=False
+            model_to_test = self.get_model(model, raise_if_missing=True)
+            connection_config = self._test_connection_configs.get(
+                model_to_test.gateway or self.config.default_gateway
             )
+            assert connection_config
```
tobymao, line 1663: just do configs[gateway] if you're going to assert anyways

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -277,17 +286,24 @@ def create(
                 snapshots_with_table_names[snapshot].add(table.name)
                 table_deployability[table.name] = is_deployable
                 tables_by_schema[d.schema_(table.db, catalog=table.catalog)].add(table.name)
```
tobymao, line 288: make the schema a variable since it's used twice

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -277,17 +286,24 @@ def create(
...
+                gateway_per_schema[d.schema_(table.db, catalog=table.catalog)] = (
```
tobymao, line 289: this doesn't seem right, why is there a a gateway for each schema? isn't the gateway dependent on the model?

Themiscodes: yes, but this is used in `_get_data_objects` and `_create_schemas` to ensure that the appropriate engine adapter is used there too. so when the corresponding model has an explicit gateway set, mapping this to the schemas that are used in these schema-specific queries as well

sqlmesh/core/context.py
```diff
@@ -237,12 +243,14 @@ def __init__(
         deployability_index: t.Optional[DeployabilityIndex] = None,
         default_dialect: t.Optional[str] = None,
         default_catalog: t.Optional[str] = None,
+        engine_adapters: t.Optional[t.Dict[str, EngineAdapter]] = None,
         variables: t.Optional[t.Dict[str, t.Any]] = None,
     ):
         self.snapshots = snapshots
         self.deployability_index = deployability_index
         self._engine_adapter = engine_adapter
```
tobymao, line 251: do we need this anymore?

Themiscodes: no thank you I will remove this

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -337,16 +352,27 @@ def migrate(
             snapshots: Mapping of snapshot ID to snapshot.
             allow_destructive_snapshots: Set of snapshots that are allowed to have destructive schema changes.
         """
+
+        snapshot_to_adapter = {
+            t.name: self._get_engine_adapter(t.model.gateway)
+            if t.model_or_none and t.model.gateway
```
tobymao, line 358: if t.is_model

tobymao: make t s instead or snapshot

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -90,11 +90,18 @@ class SnapshotEvaluator:
         adapter: The adapter that interfaces with the execution engine.
         ddl_concurrent_tasks: The number of concurrent tasks used for DDL
             operations (table / view creation, deletion, etc). Default: 1.
+        engine_adapters: Optional dictionary of multiple engine adapters.
     """
 
-    def __init__(self, adapter: EngineAdapter, ddl_concurrent_tasks: int = 1):
+    def __init__(
+        self,
+        adapter: EngineAdapter,
```
tobymao, line 98: can we get rid of adapter and instead make it

adapters: EngineAdapter | t.Dict[str, EngineAdapter]

that way there's only one property to deal with

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -777,9 +820,9 @@ def _promote_snapshot(
         if snapshot.is_model:
             table_name = snapshot.table_name(deployability_index.is_representative(snapshot))
             view_name = snapshot.qualified_view_name.for_environment(
-                environment_naming_info, dialect=self.adapter.dialect
+                environment_naming_info, dialect=self.adapter().dialect
```
georgesittas, line 823: Should we just call `adapter()` once?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -796,9 +839,9 @@ def _demote_snapshot(
         on_complete: t.Optional[t.Callable[[SnapshotInfoLike], None]],
     ) -> None:
         view_name = snapshot.qualified_view_name.for_environment(
-            environment_naming_info, dialect=self.adapter.dialect
+            environment_naming_info, dialect=self.adapter().dialect
```
georgesittas, line 842: Ditto about calling `adapter()` once.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -861,18 +905,20 @@ def _audit(
         blocking = audit_args.pop("blocking", None)
         blocking = blocking == exp.true() if blocking else audit.blocking
 
+        adapter = self.adapter(snapshot.model.gateway if snapshot.is_model else None)
```
georgesittas, line 908: This is also repeated a lot, may be worth making a new method on snapshots that does this and do `snapshot.gateway` instead. Not sure which class in the snapshot hierarchy makes the most sense to have this.

sqlmesh/core/context.py
```diff
@@ -370,15 +376,37 @@ def __init__(
 
         self._connection_config = self.config.get_connection(self.gateway)
         self.concurrent_tasks = concurrent_tasks or self._connection_config.concurrent_tasks
-        self._engine_adapter = engine_adapter or self._connection_config.create_engine_adapter()
 
-        self.console = console or get_console(dialect=self._engine_adapter.dialect)
+        self._engine_adapters: t.Dict[str, EngineAdapter] = {
+            self.config.default_gateway_name: engine_adapter
```
tobymao, line 381: do you still need engine_adapter?

sqlmesh/core/context.py
```diff
@@ -1465,8 +1497,11 @@ def table_diff(
         """
         source_alias, target_alias = source, target
 
+        adapter = self.engine_adapter
```
tobymao, line 1500: you can make engine_adapter a property that fetches from engine_adapters

sqlmesh/core/context.py
```diff
@@ -1465,8 +1497,11 @@ def table_diff(
...
         if model_or_snapshot:
             model = self.get_model(model_or_snapshot, raise_if_missing=True)
+            if model and model.gateway:
```
tobymao, line 1503: or do 

self.engine_adapter(model.gate_way)

sqlmesh/core/context.py
```diff
@@ -1625,14 +1660,19 @@ def create_test(
...
         try:
-            test_adapter = self._test_connection_config.create_engine_adapter(
-                register_comments_override=False
-            )
+            model_to_test = self.get_model(model, raise_if_missing=True)
+            connection_config = self._test_connection_configs[
+                model_to_test.gateway or self.config.default_gateway_name
+            ]
+            adapter = self._engine_adapters[
+                model_to_test.gateway or self.config.default_gateway_name
```
tobymao, line 1668: ditto, this should be a function

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -358,11 +384,24 @@ def cleanup(
         snapshots_to_dev_table_only = {
             t.snapshot.snapshot_id: t.dev_table_only for t in target_snapshots
         }
+
+        snapshot_to_adapter = {
```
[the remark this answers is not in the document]
tobymao: you still don't really need this dictionary i think

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -87,15 +87,25 @@ class SnapshotEvaluator:
...
+        adapters: EngineAdapter | t.Dict[str, EngineAdapter],
+        ddl_concurrent_tasks: int = 1,
+    ):
+        self.adapters = adapters if isinstance(adapters, t.Dict) else {"": adapters}
         self.ddl_concurrent_tasks = ddl_concurrent_tasks
 
+    def adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
+        """Returns the adapter for the specified gateway or the default adapter if none is provided."""
+        default_gateway = next(iter(self.adapters))
+        return self.adapters.get(gateway or default_gateway, self.adapters[default_gateway])
```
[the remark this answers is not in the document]
izeigerman: should we fail if the gateway was specified but is missing in the mapping?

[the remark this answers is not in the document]
Themiscodes: > Is the first gateway in self.adapters always the default gateway? 

yes because in the context, the adapters are set by placing the default gateway as the first item in the dictionary.

> What if the config has a default_gateway, different than the first one?

this could in theory occur if someone initialised a SnapshotEvaluator but didn't set the first item in the dictionary of adapters as the default. so the assumption is that when a dictionary of adapters is passed to the evaluator, the first item is always considered the default gateway. since the default engine adapter is no longer passed explicitly to the evaluator, I was unsure if there might be a better way to handle this

> should we fail if the gateway was specified but is missing in the mapping?

yes, I was unsure whether to fail or not, but I suppose it would be wiser to raise an error rather than default to the adapter

sqlmesh/core/snapshot/evaluator.py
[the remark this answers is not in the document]
Themiscodes: right removing these

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -276,18 +288,22 @@ def create(
                 )
                 snapshots_with_table_names[snapshot].add(table.name)
                 table_deployability[table.name] = is_deployable
-                tables_by_schema[d.schema_(table.db, catalog=table.catalog)].add(table.name)
+                table_schema = d.schema_(table.db, catalog=table.catalog)
+                tables_by_schema[table_schema].add(table.name)
+                gateway_by_schema[table_schema] = snapshot.model.gateway or ""
```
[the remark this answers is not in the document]
Themiscodes: I'm wondering if this is a problem though since this dictionary is used for creating the schemas

themisvaltinos:
Addressed comments and updated docs

tests/core/test_context.py
```diff
@@ -312,6 +313,81 @@ def test_evaluate_limit():
...
+        assert ctx._get_engine_adapter("dev")
+
+    ctx = Context(paths=path, config="isolated_systems_config")
+    assert len(ctx._engine_adapters) == 3
+    assert ctx.engine_adapter == ctx._engine_adapters["dev"]
+    assert ctx._get_engine_adapter("prod") == ctx._engine_adapters["prod"]
+    assert ctx._get_engine_adapter("test") == ctx._engine_adapters["test"]
+
+
+def test_multiple_gateways(tmp_path: Path):
```
georgesittas, line 331: Besides checking the gateway values, what's the value of the remaining assertions in this test?

tests/core/test_context.py
```diff
@@ -372,7 +448,7 @@ def test_override_builtin_audit_blocking_mode():
             call(
                 "Audit 'not_null' for model 'db.x' failed.\n"
                 "Got 1 results, expected 0.\n"
-                'SELECT * FROM (SELECT * FROM "sqlmesh__db"."db__x__2457047842" AS "db__x__2457047842") AS "_q_0" WHERE "c" IS NULL AND TRUE\n'
+                'SELECT * FROM (SELECT * FROM "sqlmesh__db"."db__x__1949820056" AS "db__x__1949820056") AS "_q_0" WHERE "c" IS NULL AND TRUE\n'
```
georgesittas, line 451: Should we fix this test by actually interpolating the corresponding `snapshot.fingerprint.to_version()` instead of hard-coding it?

tests/core/test_integration.py
```diff
@@ -2337,12 +2337,6 @@ def test_multi(mocker):
     assert len(plan.new_snapshots) == 4
     context.apply(plan)
 
-    context = Context(
-        paths=["examples/multi/repo_1"],
-        engine_adapter=context.engine_adapter,
-        state_sync=context.state_sync,
-        gateway="memory",
-    )
```
georgesittas, line 2345: Why was this removed?

tests/core/test_state_sync.py
```diff
@@ -2225,12 +2225,12 @@ def test_snapshot_batching(state_sync, mocker, make_snapshot):
         call(
             exp.to_table("sqlmesh._snapshots"),
             where=parse_one(
-                f"(name, identifier) in (('\"a\"', '{snapshot_b.identifier}'), ('\"a\"', '{snapshot_c.identifier}'))"
+                f"(name, identifier) in (('\"a\"', '{snapshot_b.identifier}'), ('\"a\"', '{snapshot_a.identifier}'))"
```
georgesittas, line 2228: Why did this test change?

georgesittas (commented):
Thanks for addressing comments @themisvaltinos, I did another pass. Quite a few of them were minor / nits, but thought I'd mention them anyway so they're visible when this is ready for a clean up pass.

tests/core/test_state_sync.py
themisvaltinos: Because of the new hash values, so now the hash of Snapshot A is greater than that of Snapshot C and so in the first batch when they're ordered it is going to be A instead.

docs/guides/multi_engine.md
```diff
@@ -0,0 +1,112 @@
+# Multi-Engine guide
+
+Organizations typically connect to a data warehouse through a single engine to ensure data consistency. However, there are cases where the processing capabilities of one engine may be better suited to specific tasks than another.
+
+By decoupling storage from compute and with growing support for open table formats like Apache Iceberg and Hive, different engines can now interact with the same data.
+
+With SQLMesh's new multi-engine feature, users can leverage multiple engine adapters within a single project, offering the flexibility to choose the best engine for each task.
+
+This feature allows you to run each model on a specified engine, provided the data catalog is shared and the engines support read/write operations on it.
```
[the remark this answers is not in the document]
themisvaltinos: sure so add maybe that it is possible to use a setup where the catalog is not shared, when the engine provides support or an integration for federated queries, but it would be advisable to proceed with caution in such cases?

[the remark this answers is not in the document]
georgesittas: Perhaps, yeah. Let's see what others think about this use case.

sqlmesh/core/context.py
```diff
@@ -367,18 +371,38 @@ def __init__(
...
+        self._test_connection_configs: t.Dict[str, ConnectionConfig] = {
+            self.default_gateway: self.config.get_test_connection(
+                self.gateway,
+                self.default_catalog,
+                default_catalog_dialect=self.engine_adapter.DIALECT,
+            )
+        }
 
-        self._snapshot_evaluator: t.Optional[SnapshotEvaluator] = None
+        if not gateway:
```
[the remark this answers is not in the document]
themisvaltinos: my thinking was that we use the gateway to "control" if we want the project to be using the specific engine only, so when it is provided to only have this engine adapter. also since some of our integration tests run in parallel across different engines, and we use this gateway to control which engine to use, without it, these tests hang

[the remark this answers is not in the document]
georgesittas: > my thinking was that we use the gateway to "control" if we want the project to be using the specific engine only, so when it is provided to only have this engine adapter.

Hmm, I'm thinking it's more like "choose this over the default gateway", not as much "I want **only** this adapter".

For example, you could configure a multi-gateway project in a way such that when switching your gateway to a non-default one, you'd create your db objects in a different, "staging" schema or catalog (if you made either a function of the current gateway, for example). Does this make sense?

sqlmesh/core/context.py
```diff
@@ -289,7 +295,6 @@ class GenericContext(BaseContext, t.Generic[C]):
     """Encapsulates a SQLMesh environment supplying convenient functions to perform various tasks.
 
     Args:
-        engine_adapter: The default engine adapter to use.
```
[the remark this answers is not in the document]
tobymao: i don't think it's needed

sqlmesh/core/context.py
```diff
@@ -367,18 +371,38 @@ def __init__(
...
-            self.gateway, self.default_catalog, default_catalog_dialect=self.engine_adapter.DIALECT
-        )
+        self.console = console or get_console(dialect=self.engine_adapter.dialect)
+        self._test_connection_configs: t.Dict[str, ConnectionConfig] = {
+            self.default_gateway: self.config.get_test_connection(
+                self.gateway,
+                self.default_catalog,
+                default_catalog_dialect=self.engine_adapter.DIALECT,
+            )
+        }
```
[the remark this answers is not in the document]
tobymao: yea, maybe just control it in the test

sqlmesh/core/context.py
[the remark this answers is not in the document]
tobymao: i don't understand what that means

sqlmesh/core/context.py
```diff
@@ -228,6 +233,7 @@ class ExecutionContext(BaseContext):
         engine_adapter: The engine adapter to execute queries against.
         snapshots: All upstream snapshots (by model name) to use for expansion and mapping of physical locations.
         deployability_index: Determines snapshots that are deployable in the context of this evaluation.
+        engine_adapters: Optional dictionary of multiple engine adapters.
```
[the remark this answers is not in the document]
georgesittas: Note: discussed about implications of this offline. We can't just deprecate `engine_adapter` in `ExecutionContext`, because it's used in python models to do dynamic db calls. Same holds for the macro evaluator's `engine_adapter` attribute, which needs to match the corresponding model's gateway.

sqlmesh/core/context.py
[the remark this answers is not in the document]
georgesittas: @tobymao you mean my comment? I mean if you had a model like this in a multi-gateway project:

```sql
MODEL (
  name @IF(@gateway = ..., X, Y),
  ...
); ...
```

and you wanted to do things like run `sqlmesh --gateway foo plan`, you'd still want your project to work with multiple gateways. Currently, if the `--gateway` option is passed, we won't initialize the other connections / engine adapters in the constructor, and so this use case wouldn't be supported.

sqlmesh/core/context.py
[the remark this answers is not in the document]
georgesittas: +1 to this ^, I think the most likely scenario is that folks will want to use the default gateway's testing connection. For example, if someone set it to spark to use it throughout their multi-gateway project, then we could end up silently testing the models that use a different gateway using duckdb, unless they reused the same testing configs for spark in the other gateways.

tests/core/test_context.py
themisvaltinos: it was to test `_get_engine_adapter` returns the correct engine adapter but will clean this up a bit so it checks each case only once

tests/core/test_context.py
themisvaltinos: sure I can change that, I thought these were left as reminders for breaking changes, to identify when the data hashes change and create a migration script

sqlmesh/core/context.py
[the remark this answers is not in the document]
themisvaltinos: right revised back to the old logic with one test_connection_config

tests/core/test_integration.py
themisvaltinos: revised it to take the previous engine_adapter and added it back

sqlmesh/core/context.py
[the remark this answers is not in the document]
tobymao: you need the property but you don’t need to pass it in right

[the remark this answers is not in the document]
georgesittas: Why? How will the execution context know which adapter to use then for a python model that specifies a different gateway than the default?

tobymao (approved):

sqlmesh/core/context.py
```diff
@@ -1867,7 +1883,7 @@ def print_info(self, skip_connection: bool = False, verbose: bool = False) -> No
                 self.config.get_state_connection(self.gateway), self.console, "State Connection"
             )
 
-        self._try_connection("data warehouse", self._engine_adapter.ping)
+        self._try_connection("data warehouse", self.engine_adapter.ping)
```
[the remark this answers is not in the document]
themisvaltinos: sure, changed it to ping all available adapters

docs/guides/multi_engine.md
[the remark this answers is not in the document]
erindru: > What about use cases where the catalog's not shared but an engine provides native support / plugins for federated queries? 

Isnt this the existing single-gateway model, just against a federated query engine? Trino falls into this as well. You write your SQLMesh project against Trino and the fact that some catalogs are backed by another DB engine is an implementation detail. It's all the same SQL dialect, the query federation provides the compatibility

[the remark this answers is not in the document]
georgesittas: I see, yeah I think your point is reasonable. Feel free to ignore this @themisvaltinos, perhaps we can even remove the final section in the markdown where we say "take caution [...]" and just leave the example there?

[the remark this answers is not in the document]
themisvaltinos: ok removed this section

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -208,6 +208,7 @@ def create_view(
         underlying table without dropping the view first. This is a problem for us since we want to be able to
         swap tables out from under views. Therefore, we create the view as non-binding.
         """
+        create_kwargs.pop("no_schema_binding", None)
```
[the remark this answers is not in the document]
izeigerman: We should at least warn that this attribute is deprecated and has no effect.

sqlmesh/core/context.py
```diff
@@ -1867,10 +1883,13 @@ def print_info(self, skip_connection: bool = False, verbose: bool = False) -> No
                 self.config.get_state_connection(self.gateway), self.console, "State Connection"
             )
 
-        self._try_connection("data warehouse", self._engine_adapter.ping)
+        self._create_engine_adapters()
```
izeigerman, line 1886: Why do we call this here? Didn't we already call this in constructor?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -88,15 +88,30 @@ class SnapshotEvaluator:
...
 
-    def __init__(self, adapter: EngineAdapter, ddl_concurrent_tasks: int = 1):
-        self.adapter = adapter
+    def __init__(
+        self,
+        adapters: EngineAdapter | t.Dict[str, EngineAdapter],
+        ddl_concurrent_tasks: int = 1,
+    ):
+        self.adapters = adapters if isinstance(adapters, t.Dict) else {"": adapters}
+        self.adapter = next(iter(self.adapters.values()))
```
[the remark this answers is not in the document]
georgesittas: I'd made a similar comment previously, but it [seems like](https://github.com/TobikoData/sqlmesh/pull/3394#discussion_r1854337841) the first one is always gonna be the primary adapter because that's set in the [dictionary first](https://github.com/TobikoData/sqlmesh/pull/3394/files#diff-02c2689e9e8516b4f905e12fa4dac9b0fff86efae7964c29fb9337055f6f3d94R377-R379).

sqlmesh/core/context.py
themisvaltinos: this is to test all engine adapter connections in the `config.yaml` when calling `sqlmesh info`. so the engine adapters wouldn't be in the context if not calling the create_engine_adapters first

sqlmesh/core/snapshot/evaluator.py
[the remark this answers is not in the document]
themisvaltinos: one alternative would be to pass again the default_adapter? or maybe the `gateway` or `default_gateway` should be passed as a string to the evaluator when initialising?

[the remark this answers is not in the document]
themisvaltinos: Went with the latter and using as a fallback the first adapter if the gateway isn't passed on to the evaluator

sqlmesh/core/context.py
themisvaltinos: So the `create_engine_adapters `is called here to create and test **all** engine adapters, while before it is called to create only the ones that are needed for the project (ie the default adapter and the ones that are explicitly set in a model)

themisvaltinos:
Thanks @izeigerman addressed comments if you want to take another look

sqlmesh/core/model/meta.py
```diff
@@ -78,6 +78,7 @@ class ModelMeta(_Node):
     signals: t.List[exp.Tuple] = []
     enabled: bool = True
     physical_version: t.Optional[str] = None
+    gateway: t.Optional[str] = None
```
izeigerman, line 81: Should this be a part of the data hash?

themisvaltinos: Yes I think so I already have it in the data hash

sqlmesh/core/context.py
izeigerman: Can we just load all available adapters once on load? I'm a little concerned with the control flow here. If I'm using Context in the notebook or a script, the `self. _engine_adapters` changes underneath me depending on which method I call.

sqlmesh/core/context.py
```diff
@@ -1867,10 +1884,13 @@ def print_info(self, skip_connection: bool = False, verbose: bool = False) -> No
...
+        self._try_connection(
+            "data warehouse", [adapter.ping for adapter in self._engine_adapters.values()]
```
izeigerman, line 1889: On a second thought, I don't think this is valid. Eg. the multiple gateways may have different state and test connections but we don't test those. I'd say when using `sqlmesh info` users are expected to specify which specific gateway they want info for, like `sqlmesh --gateway <target gateway> info`.

izeigerman: Additionally, if we remove this logic from here, the `_create_engine_adapters` method will only be called from `def snapshot_evaluator`, which means that we no longer need to mutate the `_engine_adapters` state within the context and just return the dict from the method itself.

izeigerman: If we do want to keep this logic, I'd suggest extracting the actual gateways from the models and only testing those.

sqlmesh/core/context.py
```diff
@@ -367,19 +369,22 @@ def __init__(
         self.environment_ttl = self.config.environment_ttl
         self.pinned_environments = Environment.sanitize_names(self.config.pinned_environments)
         self.auto_categorize_changes = self.config.plan.auto_categorize_changes
+        self.default_gateway = gateway or self.config.default_gateway_name
```
izeigerman, line 372: I suggest to not overload the `default_gateway` here. Got me really confused. Default gateway is the one specified in the configuration. `gateway` is the override user provides in eg. CLI `sqlmesh --gateway <gateway>`. We need a better way to name this variable.

izeigerman: Some ideas: `selected_gateway`

sqlmesh/core/context.py
```diff
@@ -406,24 +411,32 @@ def default_dialect(self) -> t.Optional[str]:
...
+            if self._snapshot_gateways:
+                self._create_engine_adapters(set(self._snapshot_gateways.values()))
             self._snapshot_evaluator = SnapshotEvaluator(
-                self.engine_adapter.with_log_level(logging.INFO),
+                {
+                    gateway: adapter.with_log_level(logging.INFO)
+                    for gateway, adapter in self._engine_adapters.items()
+                },
                 ddl_concurrent_tasks=self.concurrent_tasks,
+                default_gateway=self.default_gateway,
```
izeigerman, line 428: Ditto, `default_gateway` means the default gateway configured by a user. This is not it.

sqlmesh/core/context.py
themisvaltinos: Yes actually that's much better thank you changed all these

sqlmesh/core/context.py
themisvaltinos: > On a second thought, I don't think this is valid. Eg. the multiple gateways may have different state and test connections but we don't test those. I'd say when using sqlmesh info users are expected to specify which specific gateway they want info for, like sqlmesh --gateway <target gateway> info.

Right this use is actually more clear to test the specified engine when using info with a specific gateway, so removing this logic to revert to how it was originally

georgesittas: Ah, nice. I think this was a misleading suggestion on my end, the `--gateway` idea is indeed much better 👍

izeigerman (approved):

PR #3396, opened 2024-11-19
Feat: Add support for YC managed Airflow

petrikoro:
# **Add Support for YC Managed Airflow**

## **Description**
This PR introduces a new configuration for YC Managed Airflow, enabling the use of both `username:password` credentials and an IAM token for authentication.

Documentation for the YC Managed Airflow: https://yandex.cloud/en/docs/managed-airflow/operations/af-interfaces

## **Changes Made**
- Added the `YCAirflowSchedulerConfig` class to support YC Managed Airflow configuration.

### DIFF
--- a/docs/integrations/airflow.md
+++ b/docs/integrations/airflow.md
@@ -155,3 +155,7 @@ default_scheduler:
   type: mwaa
   environment: <The MWAA Environment Name>
 ```
+
+### Yandex Cloud
+
+Yandex Cloud offers [managed Airflow instances](https://yandex.cloud/en/services/managed-airflow), and SQLMesh fully supports Airflow hosted by Yandex Cloud.
--- a/docs/integrations/airflow.md
+++ b/docs/integrations/airflow.md
@@ -156,6 +156,6 @@ default_scheduler:
   environment: <The MWAA Environment Name>
 ```
 
-### Yandex Cloud
+### YC Airflow
 
-Yandex Cloud offers [managed Airflow instances](https://yandex.cloud/en/services/managed-airflow), and SQLMesh fully supports Airflow hosted by Yandex Cloud.
+SQLMesh fully supports Airflow hosted on Yandex [managed Airflow instances](https://yandex.cloud/en/services/managed-airflow) - see the [configuration reference page](../reference/configuration.md#yc-airflow) for more information.
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -279,10 +279,6 @@ class YCAirflowSchedulerConfig(_BaseAirflowSchedulerConfig, BaseConfig):
     dag_creation_poll_interval_secs: int = 30
     dag_creation_max_retry_attempts: int = 10
 
-    backfill_concurrent_tasks: int = 4
-    ddl_concurrent_tasks: int = 4
-
-    max_snapshot_ids_per_request: t.Optional[int] = None
     use_state_connection: bool = False
 
     default_catalog_override: t.Optional[str] = None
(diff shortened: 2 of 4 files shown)


izeigerman:
It appears that the only difference from the vanilla `AirflowSchedulerConfig` configuration is a different name of the authorization header? Can we instead extend `AirflowSchedulerConfig` with ability to override it:
```
class AirflowSchedulerConfig(_BaseAirflowSchedulerConfig, BaseConfig):
    authorization_header_name: str = "Authorization"
```

It's fine if we still want a custom configuration for YC, but its implementation should become much simpler:
```
class YCAirflowSchedulerConfig(AirflowSchedulerConfig):
     authorization_header_name: str = "X-Cloud-Authorization"
```

Also does it need both the token **and** username / password?

petrikoro:
Yes, both the token and username:password are required, as stated in the documentation: https://yandex.cloud/en/docs/managed-airflow/operations/af-interfaces#rest-api.

This is the primary reason I prefer to implement it as a separate class, rather than modifying the vanilla `AirflowSchedulerConfig`, since I don't want to override the authentication method (now it's either username:password or a token).

sqlmesh/core/config/scheduler.py
```diff
@@ -251,6 +251,60 @@ def get_client(self, console: t.Optional[Console] = None) -> AirflowClient:
...
+    password: str
+    token: str
+    dag_run_poll_interval_secs: int = 10
+    dag_creation_poll_interval_secs: int = 30
+    dag_creation_max_retry_attempts: int = 10
+
+    backfill_concurrent_tasks: int = 4
+    ddl_concurrent_tasks: int = 4
+
+    max_snapshot_ids_per_request: t.Optional[int] = None
```
izeigerman, line 285: This attribute is deprecated. Let's remove it from the new config.

izeigerman:
@petrikoro any chance you can add a section on YC AIrflow support to https://github.com/TobikoData/sqlmesh/blob/main/docs/integrations/airflow.md?

sqlmesh/core/config/scheduler.py
petrikoro: Is only `max_snapshot_ids_per_request` deprecated, or are all of them? I see that `backfill_concurrent_tasks` and `ddl_concurrent_tasks` are still in use, even though the validator has `check_field=False`: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/config/common.py#L49

petrikoro:
> @petrikoro any chance you can add a section on YC AIrflow support to https://github.com/TobikoData/sqlmesh/blob/main/docs/integrations/airflow.md?

Updated docs, could you please take a look?

docs/integrations/airflow.md
```diff
@@ -155,3 +155,7 @@ default_scheduler:
   type: mwaa
   environment: <The MWAA Environment Name>
 ```
+
+### Yandex Cloud
+
+Yandex Cloud offers [managed Airflow instances](https://yandex.cloud/en/services/managed-airflow), and SQLMesh fully supports Airflow hosted by Yandex Cloud.
```
izeigerman, line 161: Sorry about the back and forth but it'd be helpful to mention which scheduler type to specify to use this. Eg. here's how we document this for the Cloud Composer: https://github.com/TobikoData/sqlmesh/blob/13cf28ab60972ef043fbe6c65ddc11908f949631/docs/reference/configuration.md#cloud-composer. We can then include a link to the config reference here.

izeigerman:
One last comment, should be good to go otherwise.

docs/integrations/airflow.md
petrikoro: Thank you! Somehow I missed the part about the configuration :D

I added information about the configuration in the latest commit and slightly adjusted the wording in `docs/integrations/airflow.md`

izeigerman (approved):

izeigerman:

PR #3405, opened 2024-11-20
chore!: refactor compute interval params to do more at runtime

eakmanrq:
Prior to this PR, `compute_interval_params` was written to determine the absolute set of intervals to be processed. This meant taking into account depends on past limitations and signals. In addition it would then batch these intervals based on the model's batch size.

This PR changes this function's behavior to instead return a superset of what may actually need to run just based on the intervals that are missing. When actually running the final set of intervals are calculated based on depends on past, signals, batch size, ect. Therefore what is actually run will be a subset of what is provided.

The overall functionality of SQLMesh is unchanged in this PR just this internal detail has been changed.

Two key other changes in the PR: 
* Prior to this PR we had `Intervals` defined as tuple of datetimes in some places and tuples of ints in other places. Now `Intervals` is consistently ints and the functions consistently exchange ints. I couldn't see a good reason why this different existed.
* Airflow was "insulated" from these changes by making of copy of existing functionality into the Airflow codebase and then also moving over the datetime tuple to there too.

### DIFF
--- a/docs/guides/signals.md
+++ b/docs/guides/signals.md
@@ -46,7 +46,9 @@ The method returns `True` (indicating that all intervals are ready for evaluatio
 ```python linenums="1"
 import random
 import typing as t
-from sqlmesh.core.scheduler import signal_factory, Batch, Signal
+from sqlmesh.core.scheduler import signal_factory, Signal
+from sqlmesh.utils.date import DatetimeRanges
+
 
 class RandomSignal(Signal):
     def __init__(
@@ -55,7 +57,7 @@ class RandomSignal(Signal):
     ):
         self.signal_metadata = signal_metadata
 
-    def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
+    def check_intervals(self, batch: DatetimeRanges) -> t.Union[bool, DatetimeRanges]:
         threshold = self.signal_metadata["threshold"]
         return random.random() > threshold
 ```
@@ -69,7 +71,9 @@ We can now add a factory function that returns the `RandomSignal` sub-class. Not
 ```python linenums="1" hl_lines="8-10"
 import random
 import typing as t
-from sqlmesh.core.scheduler import signal_factory, Batch, Signal
+from sqlmesh.core.scheduler import signal_factory, Signal
+from sqlmesh.utils.date import DatetimeRanges
+
 
 class RandomSignal(Signal):
     def __init__(
@@ -78,9 +82,10 @@ class RandomSignal(Signal):
     ):
         self.signal_metadata = signal_metadata
 
-    def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
+    def check_intervals(self, batch: DatetimeRanges) -> t.Union[bool, DatetimeRanges]:
         threshold = self.signal_metadata["threshold"]
         return random.random() > threshold
+
 
 @signal_factory
 def my_signal_factory(signal_metadata: t.Dict[str, t.Union[str, int, float, bool]]) -> Signal:
@@ -122,13 +127,13 @@ In this example, there are two signals.
 import typing as t
 from datetime import datetime
 
-from sqlmesh.core.scheduler import signal_factory, Batch, Signal
-from sqlmesh.utils.date import to_datetime
+from sqlmesh.core.scheduler import signal_factory, Signal
+from sqlmesh.utils.date import DatetimeRanges, to_datetime
 
 
 class AlwaysReady(Signal):
     # signal that indicates every interval is always ready
-    def check_intervals(self, batch: Batch) -> t.Union[bool, Batch]:
+    def check_intervals(self, batch: DatetimeRanges) -> t.Union[bool, DatetimeRanges]:
         return True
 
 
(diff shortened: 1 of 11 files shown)


sqlmesh/core/scheduler.py
```diff
@@ -39,17 +39,17 @@
 from sqlmesh.utils.errors import AuditError, CircuitBreakerError, SQLMeshError
 
 logger = logging.getLogger(__name__)
-Interval = t.Tuple[datetime, datetime]
-Batch = t.List[Interval]
-SnapshotToBatches = t.Dict[Snapshot, Batch]
+Interval = t.Tuple[int, int]
```
tobymao, line 42: this should continue to be datetime because users implement check_intervals method

sqlmesh/core/scheduler.py
```diff
@@ -4,8 +4,8 @@
 import logging
 import traceback
 import typing as t
-from datetime import datetime
 
+from more_itertools import flatten
```
tobymao, line 8: this seems like a transitive dep, either don't use it or add it explicitly to setup.py

sqlmesh/core/scheduler.py
```diff
@@ -35,21 +35,22 @@
...
     validate_date_range,
+    DatetimeRanges,
 )
 from sqlmesh.utils.errors import AuditError, CircuitBreakerError, SQLMeshError
 
 logger = logging.getLogger(__name__)
-Interval = t.Tuple[datetime, datetime]
-Batch = t.List[Interval]
-SnapshotToBatches = t.Dict[Snapshot, Batch]
+Interval = t.Tuple[int, int]
```
tobymao, line 43: why does this exist? can you import interval from snapshot definition?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1642,6 +1642,33 @@ def missing_intervals(
...
+            else:
+                timestamps.append(end_ts)
+            break
+
+        timestamps.append(ts)
+    return timestamps
+
+
+@lru_cache(maxsize=None)
+def expand_range_as_interval(
```
tobymao, line 1665: this can just be done inline in scheduler and doesn't need to be cached, the expensive part is croniter

tests/core/test_scheduler.py
```diff
@@ -59,14 +58,10 @@ def test_interval_params(scheduler: Scheduler, sushi_context_fixed_date: Context
 
     assert compute_interval_params([orders, waiter_revenue], start=start_ds, end=end_ds) == {
         orders: [
-            (to_datetime(start_ds), to_datetime("2022-01-31")),
```
tobymao, line 62: do we have a test that checks batching anymore?

tests/conftest.py
```diff
@@ -485,3 +486,19 @@ def _make_function(table_name: str, random_id: str) -> exp.Table:
         return temp_table
 
     return _make_function
+
+
+@pytest.fixture
+def get_batched_missing_intervals() -> (
```
tobymao, line 492: let's put this in test_scheduler because that's the only place that uses it

sqlmesh/core/scheduler.py
```diff
@@ -380,26 +374,104 @@ def run(
 
         return not errors
 
-    def run_batches(
+    def _batch_intervals(
```
tobymao, line 377: this method can be public, should be fine

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -325,7 +325,7 @@ def _fetch_native_df(
             # pandas will fail (Ex: `SHOW TERSE OBJECTS IN SCHEMA`). Therefore we manually convert
             # the result into a DataFrame when this happens.
             rows = self.cursor.fetchall()
-            columns = self.cursor._result_set.batches[0].column_names
+            columns = self.cursor._result_set.merged_missing_intervals[0].column_names
```
tobymao, line 328: errr are you sure this is right?

eakmanrq: good catch

docs/guides/signals.md
```diff
@@ -46,7 +46,9 @@ The method returns `True` (indicating that all intervals are ready for evaluatio
 ```python linenums="1"
 import random
 import typing as t
-from sqlmesh.core.scheduler import signal_factory, Batch, Signal
+from sqlmesh.core.scheduler import signal_factory, Signal
+from sqlmesh.utils.date import DatetimeRanges
```
izeigerman, line 50: So this is a breaking change technically?

eakmanrq: Yes the import path does change but the functionality doesn't. Changed title to be breaking.

sqlmesh/core/scheduler.py
```diff
@@ -380,26 +374,110 @@ def run(
...
+            return [(values[i], values[i + 1]) for i in range(len(values) - 1)]
+
+        dag = DAG[str]()
+
+        for snapshot in merged_intervals:
+            dag.add(snapshot.name, [p.name for p in snapshot.parents])
+
+        snapshot_intervals = {
+            snapshot: list(
+                flatten(
```
izeigerman, line 397: Do we really need this dependency? Doesn't python do this natively?
```
i
for interval in intervals
for in expand_range_as_interval(*interval, snapshot.node.interval_unit)
```

tobymao: yea i agree, we should get rid of it

eakmanrq: removed

izeigerman (approved):

PR #3412, opened 2024-11-22
Fix: allow :: casting for JSONExtract[Scalar] functions

### LINKED ISSUE #3406: `sqlmesh format` is rewritting `::` to `CAST`, but only in some cases
Bumping from `0.129.0` to `0.133.0` causes the following diffs in our models:

```sql
SELECT
  foo::STRING,
  JSON_VALUE(j, '$.bar')::NUIMERIC AS bar
FROM ...
```

rewritten to:

```sql
SELECT
  foo::STRING,
  CAST(JSON_VALUE(j, '$.bar') AS NUIMERIC) AS bar
FROM ...
```

Can we avoid this? We much prefer the shorthand casts than the normal BigQuery syntax.

georgesittas:
Fixes #3406

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -13,7 +13,7 @@ from functools import lru_cache
 import pandas as pd
 from sqlglot import Dialect, Generator, ParseError, Parser, Tokenizer, TokenType, exp
 from sqlglot.dialects.dialect import DialectType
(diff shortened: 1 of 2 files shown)


tobymao (approved):

georgesittas:
Discussed internally and concluded that the complexity to address this is not worth it.

PR #3417, opened 2024-11-22
Fix: Support coercion of BIGNUMERIC values into FLOAT64 for BigQuery

izeigerman:
Even though `BIGNUMERIC` is not compatible with `FLOAT64` when it comes to column alteration, `BIGNUMERIC` values can still be safely inserted into `FLOAT64` columns. This update ensures that SQLMesh can do this without needing to drop and recreate a column.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -78,6 +78,11 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 exp.DataType.build("DATETIME", dialect=DIALECT),
             },
         },
+        coerceable_types={
+            exp.DataType.build("FLOAT64", dialect=DIALECT): {
+                exp.DataType.build("BIGNUMERIC", dialect=DIALECT),
+            },
+        },
         support_coercing_compatible_types=True,
         parameterized_type_defaults={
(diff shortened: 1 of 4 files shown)


tobymao (approved):

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -78,6 +78,11 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 exp.DataType.build("DATETIME", dialect=DIALECT),
             },
         },
+        coerceable_types={
```
eakmanrq, line 81: Should we call this `coerceable_type_overrides`?

eakmanrq: Trying to better communicate there is already an established way to define coerceable types

tobymao: i think it's fine as is

izeigerman: This doesn't override though. `support_coercing_compatible_types` still applies

eakmanrq: Ah right thanks it just updates the set if overlapping keys. Thanks.

eakmanrq (approved):

PR #3418, opened 2024-11-22
Tobiko cloud docs structure

mesmith027:
Added the structure for the cloud documentation. 

Everything is in the docs/cloud directory: 
- cloud_index.md: overview page of the differences and benefits of Tobiko cloud 
- quickstart.md: Quickstart onboarding guide
- features
    - alerts_notifications.md 
    - dag_debugger.md
    - data_catalog.md
    - model_freshness.md
    - scheduler.md 
- observability 
    - monitoring.md
    - measures_dashboards.md


mkdocs.yml
```diff
@@ -98,6 +98,18 @@ nav:
     - Configuration:
       - reference/configuration.md
       - reference/model_configuration.md
+#  - Tobiko Cloud:
```
izeigerman, line 101: Do we not need to expose this at the top level?

izeigerman (approved):

mkdocs.yml
mesmith027: Hey, not sure I follow your question, but here is that I think you were asking below:

While we build the docs, I have added the structure to the mkdocs.yml file and commented it out just for the time being. Trey had a brilliant idea of not exposing the menu option to many "empty" files. They will be there for us to add to and edit but not accessible from the top bar menu until they have useful content for the community. 
Of course, anyone can find them by directly putting the slug in, and be redirected to a page that says the content is coming soon, but most users don't navigate the docs like this!

mkdocs.yml
```diff
@@ -162,3 +174,5 @@ extra:
   analytics:
     provider: google
     property: G-JXQ1R227VS
+#  navigation:
```
treysp, line 177: I think this would need to go under the `theme: features:` keys (like [here](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-expansion)).

But it looks like it doesn't support scoping to specific sections, so it's sitewide or none :(

PR #3427, opened 2024-11-26
Fix: Add CASCADE modified when dropping columns in Postgres / Redshift

izeigerman:
Otherwise, the schema migration fails because of a view in the virtual layer which still points at the table that is being modified.

### DIFF
--- a/sqlmesh/core/engine_adapter/postgres.py
+++ b/sqlmesh/core/engine_adapter/postgres.py
@@ -61,6 +61,7 @@ class PostgresEngineAdapter(
                 exp.DataType.build("BPCHAR", dialect=DIALECT).this
             },
         },
+        drop_cascade=True,
     )
 
     def _fetch_native_df(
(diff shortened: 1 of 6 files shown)


erindru:
> the schema migration fails because of a view in the virtual layer which still points at the table that is being modified

In that case, isn't the real issue that the view was created with schema binding?

themisvaltinos:
> In that case, isn't the real issue that the view was created with schema binding?

@erindru in the case of redshift I'm not sure if it is enough since we do create with no schema binding for the virtual layer as the following line is `False` 

https://github.com/TobikoData/sqlmesh/blob/d1ffac05e3d7cbea741bc56d20b3cf4c6ad69317/sqlmesh/core/engine_adapter/redshift.py#L220

only when the call to create_view is from when we build a `ctas` statement:

https://github.com/TobikoData/sqlmesh/blob/d1ffac05e3d7cbea741bc56d20b3cf4c6ad69317/sqlmesh/core/engine_adapter/mixins.py#L303

though for the ctas case I still had to modify and always create with no schema binding for redshift in my pr so it worked for views to external tables.

izeigerman:
@erindru additionally, there can be external views created on top of SQLMesh views that would prevent this operation from succeeding.

tobymao (approved):

erindru (approved):

PR #3430, opened 2024-11-27
Fix: Enable concurrency in DuckDB

erindru:
The motivation for this PR is to help prevent people from getting tripped up when using DuckDB for state. The most common problem is related to SQLMesh freezing.

Prior to this PR:
 - DuckDB `concurrent_tasks` was hardcoded to 1
 - This ensured that its connection pool was always in single threaded mode
 - If the main warehouse connection had `concurrent_tasks: 8` or something set, this causes those threads to hit the single DuckDB cursor
 - DuckDB deadlocks

In the time since DuckDB support was originally added, support for [multiple thread concurrency](https://duckdb.org/docs/guides/python/multiple_threads.html) has been made available. This is based on using threadlocal cursors, which our multithreaded connection pool already does.

So this PR:
 - Removes the hardcoding on `concurrent_tasks` (but still defaults to `1`)
 - Automatically sets `concurrent_tasks` based on the warehouse connection `concurrent_tasks` to ensure that the DuckDB connection pool runs in multithreaded mode when required
 - Removes DuckDB from the list of recommended production state sync engines since at the end of the day its still single user / flat file and there is no path to production usage

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -346,10 +346,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          #filters:
-          # branches:
-          #   only:
-          #     - main
+          filters:
+           branches:
+             only:
+               - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -346,10 +346,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
+          # branches:
+          #   only:
+          #     - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -166,13 +166,6 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
     def _connection_factory(self) -> t.Callable:
         import duckdb
 
-        if self.concurrent_tasks > 1:
-            # ensures a single connection instance is used across threads
-            # rather than a new connection being established per thread
-            # this is in line with https://duckdb.org/docs/guides/python/multiple_threads.html
-            # the important thing is that the *cursor*'s are per thread, but the connection should be shared
-            return lru_cache(duckdb.connect)
-
         return duckdb.connect
 
     @property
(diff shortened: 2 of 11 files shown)


tobymao (approved):

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -1314,8 +1314,6 @@ def test_sushi(ctx: TestContext, tmp_path_factory: pytest.TempPathFactory):
 
     current_gateway_config = config.gateways[ctx.gateway]
     current_gateway_config.state_schema = sushi_state_schema
-    if (sc := current_gateway_config.state_connection) and sc.type_ == "duckdb":
```
erindru, line 1317: This is the main test for this PR.  These guards were introduced originally to prevent the integration tests from hanging

tests/core/engine_adapter/integration/test_integration_duckdb.py
```diff
@@ -0,0 +1,72 @@
...
+from sqlglot import exp
+
+from sqlmesh.core.config.connection import DuckDBConnectionConfig
+from sqlmesh.utils.connection_pool import ThreadLocalConnectionPool
+
+pytestmark = [pytest.mark.duckdb, pytest.mark.engine, pytest.mark.slow]
+
+
+@pytest.mark.parametrize("database", [None, "db.db"])
+def test_multithread_concurrency(tmp_path, database: t.Optional[str]):
```
erindru, line 14: This is the test I wrote to explore DuckDB's concurrency. It's based on the example from their website.

It originally took _minutes_ to complete until I started sharing the same connection object across all threads, then it dropped down to a few seconds

erindru:
Update: Still working on this approach as I can still trigger hangs locally (although its vastly improved / a lot harder now)

.circleci/continue_config.yml
```diff
@@ -346,10 +346,10 @@ workflows:
...
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
+          # branches:
+          #   only:
+          #     - main
```
erindru, line 352: TODO: revert this. I just wanted to check that the cloud engines are running fine as well

erindru:
Ok, i'm happy with the stability of this PR now.
 - I cant trigger a deadlock locally anymore
 - I ran the full suite of tests (including cloud) 5 times in a row on CircleCI with zero failures or lockups
 
 @tobymao its changed a bit since you first looked at it, the key part being the DuckDB connection factory making `connection.close()` a no-op when running in Multithreaded mode

sqlmesh/core/config/connection.py
```diff
@@ -39,7 +40,7 @@
 
 logger = logging.getLogger(__name__)
 
-RECOMMENDED_STATE_SYNC_ENGINES = {"postgres", "gcp_postgres", "mysql", "duckdb", "mssql"}
+RECOMMENDED_STATE_SYNC_ENGINES = {"postgres", "gcp_postgres", "mysql", "mssql"}
```
georgesittas, line 43: IIRC, DuckDB was included in this set for a more "friendly" experience for new users. If we remove it, then the example project and all simple/dummy POC projects using DuckDB for state will log a warning:

> WARNING - The duckdb engine is not recommended for storing SQLMesh state in production deployments. Please see https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#state-connection for a list of recommended engines and more information.

Just wanted to double-check whether we concluded this change is fine after all, because it seems somewhat important from a UX perspective.

georgesittas (approved):

sqlmesh/core/config/connection.py
erindru: Yeah, I did wonder about that. The thing is, it drives the wrong behaviour and users always end up going "Why cant I use DuckDB for state?"

At least this way we are up front about it.

I also thought that perhaps we could not show the warning if the user was using DuckDB for the warehouse connection as well. If its DuckDB for warehouse and DuckDB for state, its probably a POC / example project so we could just not log any warnings.

If they change to a proper warehouse then they will get a warning until they choose better state, and if DuckDB is suitable for their warehouse needs then they will continue to not get warnings.

What do you think?

erindru: Cool, i've adjusted the state sync warning to only show if the user isnt using DuckDB for both the warehouse and state

PR #3432, opened 2024-11-27
fix: round columns percentage stats to 9th decimal

### LINKED ISSUE #3431: `table_diff` rounds only at 1 decimal instead of 9 in column comparaison stats
Percent matching on column comparaison stats is rounded at first decimal even though it shows 9 decimals, which makes it even less precise than the full match count.

For example, even thought the partial match at row level shows 98.75%, it shows at 98.8% at column level:
```
Row Counts:
├──  FULL MATCH: 79 rows (98.75%)
└──  PARTIAL MATCH: 1 rows (1.25%)

                 pct_match
commentaire  100.000000000
my_int        98.800000000
```

I discovered that because although I had partially matching rows, I could not see what column was the culprit...:
```
Row Counts:
├──  FULL MATCH: 96279 rows (100.0%)
└──  PARTIAL MATCH: 2 rows (0.0%)

                        pct_match
partially_non_matching_column  100.000000000
```

esciara:
fixes #3431


georgesittas:
Thanks for the PR @esciara, can you please add a test for this?

esciara:
Was looking into it... now sure how to do it. can you point me to somewhere?

esciara:
Actually I was more looking on [tests/core/test_table_diff.py](https://github.com/TobikoData/sqlmesh/blob/943cc309c5c1cdc900f3f9dd7c6e9467c2577592/tests/core/test_table_diff.py) , since the fix is integration agnostic...

esciara:
@georgesittas is the test I added (or rather changed) good enough?

georgesittas (approved):
It should be good, yeah. Thanks!

erindru:
This looks fine, if I understand correctly its just for display purposes and not related to the existing `--decimals` [option](https://sqlmesh.readthedocs.io/en/stable/reference/cli/#table_diff)

PR #3441, opened 2024-11-29
Fix(athena): Fix error replacing partitions in Hive tables where the partition column is a date / timestamp type

erindru:
Prior to this PR, replacing an existing partition in a Hive table where the partition column was defined as `DATE` or `TIMESTAMP` resulted in the following error:

```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter PartitionsToGet[0].Values[0], value: 2024-11-21, type: <class 'datetime.date'>, valid types: <class 'str'>
```

The cause was the query to lookup partition values was returning typed python objects but the boto3 method to delete partitions from the metastore needs strings. This PR converts the values to strings adds some tests to show it works

Note that Iceberg tables are unaffected

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -340,16 +340,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
+                #- snowflake
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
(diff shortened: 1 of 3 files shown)


.circleci/continue_config.yml
```diff
@@ -340,16 +340,16 @@ workflows:
...
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
+                #- snowflake
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
                 - athena
```
erindru, line 348: This is temporary, I wanted to run the full athena test suite for this PR

tobymao (approved):

PR #3442, opened 2024-11-29
Fix(mysql): Use LONGTEXT data type when storing large blobs of text

### LINKED ISSUE #3438: Environment snapshots truncated in MySQL
We're currently using:-
* SQLMesh 0.125.6
* SQLGlot 25.24.5
* Schema Version 57
* State management connector - MySQL

After promoting a plan I then got a Pydantic validation error message when reading the state back out because the snapshots stored against the environment was truncated.
I believe this limitation still exists in the latest version after looking at code and issues that have been raised.
For reference we're up to 102 models and 32 external models in the project now.

environment.snapshots is currently implemented as [a text field](https://github.com/TobikoData/sqlmesh/blob/v0.137.1/sqlmesh/core/state_sync/engine_adapter.py#L142) which translates to TEXT in MySQL and is capped at 65,535 characters.
MySQL has different types for TEXT, MEDIUMTEXT and LONGTEXT whereas for Postgres there's just a single TEXT data type which doesn't have different sizes.

I'm considering the following options.
1. If the data types for MySQL for any snapshot field in environments could be transpiled to LONGTEXT somehow whilst not impacting the other engines that can store state.
2. If we should consider migrating our state backend from MySQL to Postgres in the real short term.
3. If the data stored in snapshots for the environment can be less verbose. From what I can tell it looks like the whole of the snapshot information is being duplicated against the environment whereas realistically we only need the primary keys to get the snapshot information from the snapshot table. I can appreciate that this would introduce some complexity and time with needing to join across tables then.

MikeWallis42:
fixes #3438

I found some code that @izeigerman committed [here](https://github.com/TobikoData/sqlmesh/commit/911c745b9e616d3018e08b0de6213a3df7e0a87e) that does what we're looking for.

Since migration v0033 the migration helper `blob_text_type` hasn't been used so I can only assume that this migration was reverted.
See attached integration testing after building the wheel and running `sqlmesh migrate` against a MySQL target.
Screenshot 2024-11-29 at 12 20 36

I got a bit lost looking at how to prevent this from happening in the future in unit/integration testing, any recommendations would be welcome.

### DIFF
--- a/sqlmesh/migrations/v0060_mysql_fix_blob_text_type.py
+++ b/sqlmesh/migrations/v0060_mysql_fix_blob_text_type.py
@@ -1,4 +1,4 @@
-"""Use LONGTEXT type for blob fields in MySQL."""
+"""Duplicate of v0033, Use LONGTEXT type for blob fields in MySQL."""
 
 from sqlglot import exp
 
@@ -24,6 +24,8 @@ def migrate(state_sync, **kwargs):  # type: ignore
 
     targets = [
         (environments_table, "snapshots"),
+        (environments_table, "previous_finalized_snapshots"),
+        (environments_table, "requirements"),
         (snapshots_table, "snapshot"),
         (seeds_table, "content"),
         (plan_dags_table, "dag_spec"),
--- a/sqlmesh/migrations/v0060_mysql_fix_blob_text_type.py
+++ b/sqlmesh/migrations/v0060_mysql_fix_blob_text_type.py
@@ -1,4 +1,8 @@
-"""Duplicate of v0033, Use LONGTEXT type for blob fields in MySQL."""
+"""Duplicate of v0033, Use LONGTEXT type for blob fields in MySQL.
+
+Seeds table has since been dropped.
+Environments table now has a requirements column.
+"""
 
 from sqlglot import exp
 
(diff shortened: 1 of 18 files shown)


sqlmesh/migrations/v0060_mysql_fix_blob_text_type.py
```diff
@@ -0,0 +1,48 @@
...
+    plan_dags_table = "_plan_dags"
+
+    if schema:
+        environments_table = f"{schema}.{environments_table}"
+        snapshots_table = f"{schema}.{snapshots_table}"
+        plan_dags_table = f"{schema}.{plan_dags_table}"
+
+    targets = [
+        (environments_table, "snapshots"),
+        (environments_table, "previous_finalized_snapshots"),
```
izeigerman, line 29: I suggest to do the same for `promoted_snapshot_ids`, since it can grow in size quite a bit too

izeigerman: Here and in other scripts.

izeigerman:
@MikeWallis42 this is a great catch and a fix. Please see my suggestion about the `promoted_snapshot_ids` column. LGTM otherwise. Thank you!

MikeWallis42:
Thanks for the feedback @izeigerman, this will be up shortly

izeigerman:
@MikeWallis42 I think you might need to rebase for the spark tests to pass

izeigerman (approved):

PR #3446, opened 2024-12-02
fix: consider sqlmesh major/minor for migrate and rollback

### LINKED ISSUE #3445: Allow rollbacks for SQLMesh updates
As part of our go-live readiness checks we simulated a migration and rollback from SQLMesh 0.125.6 to 0.126.2 and then was unable to rollback.

The migration process seems to only create backup tables if there's a schema version change or if there's a sqlglot major/minor change.
If SQLMesh has been updated without one of these 2 things also being satisfied then the backup tables are not created.
Because the backup tables don't exist then SQLMesh throws an exception when trying to rollback.
We then cannot proceed without manual intervention.

I believe I've found a simple one line fix, PR to follow.

MikeWallis42:
Fixes #3445

I was considering on top of this if it would be sensible to extend/copy the circle ci migration test.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,6 +106,7 @@ setup(
             "dbt-databricks",
             "dbt-redshift",
             "dbt-sqlserver>=1.7.0",
+            "dbt-fabric!=1.8.8",  # Broken API, https://github.com/microsoft/dbt-fabric/compare/v1.8.7...v1.8.8 L183
             "dbt-trino",
         ],
         "dbt": [
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,6 @@ setup(
             "dbt-databricks",
             "dbt-redshift",
             "dbt-sqlserver>=1.7.0",
-            "dbt-fabric!=1.8.8",  # Broken API, https://github.com/microsoft/dbt-fabric/compare/v1.8.7...v1.8.8 L183
             "dbt-trino",
         ],
         "dbt": [
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -1308,7 +1308,13 @@ class EngineAdapterStateSync(StateSync):
         versions = self.get_versions(validate=False)
         migrations = MIGRATIONS[versions.schema_version :]
 
-        migrate_rows = migrations or major_minor(SQLGLOT_VERSION) != versions.minor_sqlglot_version
+        migrate_rows = any(
+            [
+                migrations,
+                major_minor(SQLGLOT_VERSION) != versions.minor_sqlglot_version,
+                major_minor(SQLMESH_VERSION) != versions.minor_sqlmesh_version,
+            ]
+        )
         if not skip_backup and migrate_rows:
             self._backup_state()
 
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -1223,18 +1223,15 @@ class EngineAdapterStateSync(StateSync):
         migration_start_ts = time.perf_counter()
 
         try:
-            migrate_snapshots_and_environments = self._apply_migrations(
-                default_catalog, skip_backup
-            )
+            migrate_rows = self._apply_migrations(default_catalog, skip_backup)
 
-            if not migrate_snapshots_and_environments:
-                if major_minor(SQLMESH_VERSION) != versions.minor_sqlmesh_version:
-                    self._update_versions()
+            if not migrate_rows and major_minor(SQLMESH_VERSION) == versions.minor_sqlmesh_version:
                 return
 
-            self._migrate_rows(promoted_snapshots_only)
-            # Cleanup plan DAGs since we currently don't migrate snapshot records that are in there.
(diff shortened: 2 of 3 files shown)


MikeWallis42:
Looks like dbt-fabric published a new "patch" version and renamed a public function
https://github.com/microsoft/dbt-fabric/compare/v1.8.7...v1.8.8#diff-c1b0256b44c042e86c431c5151b6c0812c05e8c047c1d4f9b12b0b9538cbdb55R183
this is a requirement for dbt-sqlserver which we're installing https://github.com/dbt-msft/dbt-sqlserver/blob/master/setup.py#L69C1-L69C36

I'm attempting to exclude this patch version for the moment.

MikeWallis42:
Looks like PySpark Pandas has an issue that will be fixed in v4 because of the removal of distutils from Python 3.12

https://github.com/apache/spark/blob/v3.5.3/python/pyspark/sql/pandas/utils.py#L24
https://github.com/apache/spark/pull/43192
https://issues.apache.org/jira/browse/SPARK-45390

Still working out how to handle this.

MikeWallis42:
This was when the Bitnami image moved to Python 3.12
https://github.com/bitnami/containers/commit/ef4c745e602abf8548516cc939e67af41a09b22c

Looks like we should be able to use this image https://hub.docker.com/layers/bitnami/spark/3.5.3-debian-12-r0/images/sha256-0c045caae9ec3498fa2ddbe8d34ecd7d96178113502cac4feb2a58e837af2f6c?context=explore
until such time as v4 is released.

tests/core/engine_adapter/integration/docker/spark/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM docker.io/bitnami/spark:3.5
+# Pinned to a specific version due to PySpark Pandas requiring distutils which dissappears in python 3.12
+# Should be fixed in v4 https://issues.apache.org/jira/browse/SPARK-45390
+FROM docker.io/bitnami/spark:3.5.2-debian-12-r0
```
MikeWallis42, line 3: I assume that the Circle CI config is referenced from the main trunk rather than this branch. If not then please let me know!

MikeWallis42:
Going to wait on feedback/suggestions now with the PySpark failure.
As it's still failing then it must be the image that Circle CI is running the tests with that's the complication and I don't want to just wind that back to a version of Ubuntu that has Python 3.11 packaged within it.
From a quick search apparently you can get an older version of Python in a couple of ways https://askubuntu.com/questions/1512005/python3-11-install-on-ubuntu-24-04

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1305,7 +1305,13 @@ def _apply_migrations(
         versions = self.get_versions(validate=False)
         migrations = MIGRATIONS[versions.schema_version :]
 
-        migrate_rows = migrations or major_minor(SQLGLOT_VERSION) != versions.minor_sqlglot_version
+        migrate_rows = any(
+            [
+                migrations,
+                major_minor(SQLGLOT_VERSION) != versions.minor_sqlglot_version,
+                major_minor(SQLMESH_VERSION) != versions.minor_sqlmesh_version,
```
izeigerman, line 1312: What is the motivation for this change? The sqlmesh minor version was omitted intentionally to avoid migration when it's not needed even if the minor version was bumped. Generally there are only 2 reasons to migrate: if there's a new migration script (the `migrations` collection) or if the sqlglot version changed.

izeigerman: I read the issue carefully and now I understand the motivation better. However, `migrate_rows` not only controls whether the backups are made but also whether the migration should be performed. I suggested decoupling the 2 and only do backups on sqlmesh minor versions changes without affecting the `migrate_rows` flag.

MikeWallis42: After reading your comments and digging in to the code a bit more I believe I can see what you're talking about.

Hopefully I've addressed your concerns by only performing a table copy when it's a sqlmesh minor change without impacting the flag that drives the complex snapshot and environment migration process.

treysp:
Hello - we've [fixed the Pyspark issue](#3450), so you should be good to revert the Dockerfile change and rebase

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Yes, this looks great, thank you! However, I'm not sure why the refactor was performed in the other method (`def migrate`)? Is that change necessary? If not, can we please revert it.

setup.py
```diff
@@ -106,6 +106,7 @@
             "dbt-databricks",
             "dbt-redshift",
             "dbt-sqlserver>=1.7.0",
+            "dbt-fabric!=1.8.8",  # Broken API, https://github.com/microsoft/dbt-fabric/compare/v1.8.7...v1.8.8 L183
```
izeigerman, line 109: Could this change be causing the build failure?

izeigerman (approved):
LGTM once green. Thanks for addressing comments!

setup.py
MikeWallis42: Yep looks like dbt-sqlserver has published a release to account for the breaking change in dbt-fabric

Issue #3443, opened 2024-11-29
Systematically add the gateway info to table models in `external_tables.yaml`


esciara:
Given the context where we have two gateways with the `config.yaml` as follows

```yaml
gateways:
  local:
    connection:
      type: duckdb
      database: my_transfo.db
  bigquery:
    connection:
      type: bigquery
      project: my-project
      location: europe-west1
    state_connection:
      type: duckdb
      database: bigquery_state.duckdb

default_gateway: duckdb

model_defaults:
  dialect: bigquery
  start: 2024-10-30
```

Currently when creating external models through `sqlmesh create_external_models`, gateway information is not added to the models. Hence the information in the `external_models.yaml` file looks something like this:

```yaml
- name: '`my_transfo`.`dataset_or_schema`.`table_name`'
  columns:
    col_1: STRING
    col_2: STRING
```

When creating the  external models for the `bigquery` gateway through `sqlmesh --gateway bigquery create_external_models`, the external models are changed like this:

```yaml
- name: '`my_transfo`.`dataset_or_schema`.`table_name`'
  columns:
    col_1: STRING
    col_2: STRING
- name: '`my-project`.`dataset_or_schema`.`table_name`'
  columns:
    col_1: STRING
    col_2: STRING
  gateway: bigquery
```

Say I change my default gateway to `bigquery` in `config.yaml`:

```yaml
default_gateway: bigquery
```

If then I add a new column to the table, then it is not the table defined for the `bigquery` gateway that is changed, but the one for the `default` gateway. Hence the file is changed like this:

```yaml
- name: '`my_transfo`.`dataset_or_schema`.`table_name`'
  columns:
    col_1: STRING
    col_2: STRING
    col_3: STRING
- name: '`my-project`.`dataset_or_schema`.`table_name`'
  columns:
    col_1: STRING
    col_2: STRING
  gateway: bigquery
```

But when I do a `sqlmesh plan`, even though the default gateway has been changed to `bigquery`, it is not the default definition of the external table that is used, but the one labelled for the gateway `bigquery`. Hence `sqlmesh` throws an error saying it does not find `col_3`.

The simple solution to this would be to add the gateway information to all external models, regardless of whether it is the default gateway or not.

erindru:
> The simple solution to this would be to add the gateway information to all external models, regardless of whether it is the default gateway or not.

[...]

PR #3453, opened 2024-12-03
Fix!: Make `create_external_models` behave like other SQLMesh commands with respect to `--gateway`

erindru:
Addresses #3443 

Prior to this PR, assuming a config with:
```
gateways:
  local:
     connection:
        type: duckdb
default_gateway: local
```

Running `sqlmesh create_external_models` would produce different results than `sqlmesh --gateway local create_external_models`.

The first command, `sqlmesh create_external_models` would produce model definitions that looked like:
```
- name: foo
  columns: ..etc
- name: bar
  columns: ...etc
```

Whereas the second command, `sqlmesh --gateway local create_external_models` would produce:
```
- name: foo
  columns: ...etc
  gateway: local
- name: bar
  columns: ...etc
  gateway: local
```

That is, it would stamp the `gateway:` key on each model. The problem is that you can get into a situation where you have a file that looks like:

```
- name: foo
  columns: { "col_a": "int }
- name: foo
  columns: { "col_a": "int", "col_b": "string" }
  gateway: local
```

That is, the same model defined differently, one with a `gateway:` set and one without. 
- `sqlmesh create_external_models` would operate on the first entry because it assumes no explicit gateway means `sqlmesh --gateway None create_external_models`
- but `sqlmesh plan` would operate on the second entry because `sqlmesh plan` with `default_gateway: local` is the same as `sqlmesh --gateway local plan`.

So after this PR:
 - `sqlmesh create_external_models` behaves the same as `sqlmesh --gateway local create_external_models`
 - This means that the `gateway:` will be stamped on the external model entries next time users run the command
 - The existing load order remains for compatibility. That is, models with no `gateway:` defined are loaded first, overlayed with models for the current gateway
 - If the wrong model is getting loaded because it has two versions (one with a gateway and one without), users will need to **manually edit the file** to remove the conflicting entry. We cant do it automatically because we cant assume that the model with no `gateway:` set is the one that should be removed, it's perhaps still relevant on other gateways
 - There is no need for a migration because running `create_external_models` is a manual action and the user can decide afterwards if they want to run a forward-only plan or just a normal plan (because there may have been legitimate upstream changes, not just the addition of the `gateway:` flag)

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1830,6 +1830,14 @@ class GenericContext(BaseContext, t.Generic[C]):
             external_models_yaml = (
                 path / c.EXTERNAL_MODELS_YAML if not deprecated_yaml.exists() else deprecated_yaml
             )
+
+            external_models_gateway: t.Optional[str] = self.gateway or self.config.default_gateway
+            if not external_models_gateway:
+                # can happen if there was no --gateway defined and the default_gateway is ''
+                # which means that the single gateway syntax is being used which means there is
+                # no named gateway which means we shuld not put `gateway: ` on the external models
+                external_models_gateway = None
+
             create_external_models_file(
                 path=external_models_yaml,
                 models=UniqueKeyDict(
@@ -1843,7 +1851,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 adapter=self._engine_adapter,
                 state_reader=self.state_reader,
                 dialect=config.model_defaults.dialect,
-                gateway=self.gateway,
+                gateway=external_models_gateway,
                 max_workers=self.concurrent_tasks,
                 strict=strict,
             )
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1835,7 +1835,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             if not external_models_gateway:
                 # can happen if there was no --gateway defined and the default_gateway is ''
                 # which means that the single gateway syntax is being used which means there is
-                # no named gateway which means we shuld not put `gateway: ` on the external models
+                # no named gateway which means we should not stamp `gateway:` on the external models
                 external_models_gateway = None
 
             create_external_models_file(
--- a/tests/core/test_schema_loader.py
+++ b/tests/core/test_schema_loader.py
@@ -16,13 +16,13 @@ from sqlmesh.core.model import SqlModel, create_external_model, load_sql_based_m
 from sqlmesh.core.model.definition import ExternalModel
 from sqlmesh.core.schema_loader import create_external_models_file
 from sqlmesh.core.snapshot import SnapshotChangeCategory
-from sqlmesh.utils.yaml import YAML
+from sqlmesh.utils import yaml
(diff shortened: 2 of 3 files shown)


tests/core/test_schema_loader.py
```diff
@@ -16,13 +16,13 @@
 from sqlmesh.core.model.definition import ExternalModel
 from sqlmesh.core.schema_loader import create_external_models_file
 from sqlmesh.core.snapshot import SnapshotChangeCategory
-from sqlmesh.utils.yaml import YAML
+from sqlmesh.utils import yaml
```
erindru, line 19: I decided to clean this up a bit and stop using the internal `YAML` class

tests/core/test_schema_loader.py
```diff
@@ -107,20 +107,20 @@ def test_create_external_models(tmpdir, assert_exp_eq):
     assert context.models['"memory"."sushi"."raw_fruits"'].gateway is None
 
 
-def test_gateway_specific_external_models(tmpdir: Path):
+def test_gateway_specific_external_models(tmp_path: Path):
```
erindru, line 110: The `tmpdir` fixture returns some kind of Pytest internal Path object which didnt play nice with `yaml.load()` which wants a `pathlib.Path` object.

So I switched to the `tmp_path` fixture

georgesittas (approved):
Do we need to flag this as breaking?

sqlmesh/core/context.py
```diff
@@ -1830,6 +1830,14 @@ def create_external_models(self, strict: bool = False) -> None:
             external_models_yaml = (
                 path / c.EXTERNAL_MODELS_YAML if not deprecated_yaml.exists() else deprecated_yaml
             )
+
+            external_models_gateway: t.Optional[str] = self.gateway or self.config.default_gateway
```
izeigerman, line 1834: Should there be any specific logic around handling existing records with empty gateway?

erindru:
> Do we need to flag this as breaking?

I guess so. It is a behavior change, although it doesn't affect state when a user immediately upgrades SQLMesh, only when they run `create_external_models` again

sqlmesh/core/context.py
erindru: I couldn't come up with a test case that needed anything beyond [what already exists](https://github.com/TobikoData/sqlmesh/blob/e2b106fba8a0b17860a61bc38a3868d38c72e4a4/sqlmesh/core/schema_loader.py#L91)

This is the write path, so existing records with an empty gateway will be untouched unless the user is in [single gateway mode](https://github.com/TobikoData/sqlmesh/blob/a7e75dc98455f1e36a4bb9dde5bd71d0b91c54f9/sqlmesh/core/config/root.py#L89) where the gateway is named `''`and then they will be overwritten.

On the read path, records with no gateway will be loaded and then merged with records for the current gateway based on the `name`. If the resulting model is the wrong one (and only the user can determine this), then the user needs to manually edit the file to set `gateway:` on the record with no gateway and then remove the record with the gateway set.

Alternatively, the user can delete the file entirely and start from scratch which will ensure that all records are recreated with the proper `gateway:` set

izeigerman (approved):

PR #3454, opened 2024-12-03
Fix: import for workers

mykalmax:
Fixing case where urls has `//` and breaking import
include css files and workers into js output instead of having separate files

### DIFF
--- a/web/client/src/api/instance.ts
+++ b/web/client/src/api/instance.ts
@@ -42,10 +42,7 @@ export async function fetchAPI<T = any, B extends object = any>(
     config
   const hasSearchParams = Object.keys({ ...params }).length > 0
   const fullUrl = url.replace(/([^:]\/)\/+/g, '$1')
-  const input = new URL(
-    `${getUrlPrefix()}${fullUrl}`.replaceAll('//', '/'),
-    window.location.origin,
-  )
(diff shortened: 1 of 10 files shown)


RN-nguyen:
Verified sqlmesh ui tests, various mode, plan and CLL look good in qa.

web/client/src/api/instance.ts
```diff
@@ -126,8 +123,16 @@ function toRequestBody(obj: unknown): BodyInit {
...
-export function getUrlPrefix(): string {
-  return isStringEmptyOrNil(window.__BASE_URL__) ? '/' : window.__BASE_URL__
+export function getUrlWithPrefix(url: string = '/'): string {
+  let urlWithPrefix = `${
+    isStringEmptyOrNil(window.__BASE_URL__) ? '/' : window.__BASE_URL__
+  }/${url}`
+
+  while (urlWithPrefix.includes('//')) {
+    urlWithPrefix = urlWithPrefix.replaceAll('//', '/')
+  }
```
caiters, line 133: What input are you expecting where you might need a while loop here? or is it just in case?

mykalmax: it is combining `base url` with argument `url` and  both may have trailing/leading `/` or actually be a root path `/` and in that case string becomes `///`

caiters (approved):

Issue #3461, opened 2024-12-03
Change in audit and audit failure not captured in subsequent plan


thenaturalist:
I am puzzled and surprised by a pattern of changing code in an audit which is not caught by a plan command. 

Situation:

1. I have an audit called `is_positive`, defined as:
```
AUDIT (
    name is_positive,
  );

  SELECT * FROM @this_model
  WHERE @column < 0;
```

2. I have a model where I call the audit like so:
```
    [...]
    audits (
        is_positive(column := foobar),
    [...]
```

3. I change my audit to inverse the operator from `WHERE @column < 0;` to `WHERE @column > 0;`

4. I run `sqlmesh plan dev`, results in `No differences when compared to 'dev'`.
**Expected behaviour**: 
A) The change in the audit affecting the model is never happening silently, but explicitly. 
B) Immediately during this plan, the updated audit is run against the data and fails as expected.

5. I remove the line `is_positive(column := foobar),` from my model, rerun `sqlmesh plan dev`, forcing a metadata update:
```
Summary of differences against `dev`:
Models:
└── Metadata Updated:
    └── foo.bar
Apply - Virtual Update [y/n]: y
```

6. I re-add my `is_positive(column := foobar),` audit to my model, and rerun `sqlmesh plan dev`, forcing a metadata update again:
```
Summary of differences against `dev`:
Models:
└── Metadata Updated:
    └── foo.bar
Apply - Virtual Update [y/n]: y
```
The audit **still does not fail, despite the changed definition**?

7. I explicitly audit my model with `sqlmesh audit --model foo.bar --start 2024-10-01 --end 2024-10-03` and only then does the audit fail. 
```
MainThread - sqlmesh.core.snapshot.evaluator - WARNING - Audit 'is_positive' for model 'foo.bar' failed.
Got 3448 results, expected 0.
SELECT * FROM (SELECT * FROM "db"."sqlmesh__foo"."bar__3202718574" AS "bar__3202718574" WHERE "time_col" BETWEEN '2024-10-01' AND '2024-10-03') AS "_q_0" WHERE "column" > 0
Audit is warn only so proceeding with execution. (evaluator.py:898)
is_positive on model foo.bar ❌ FAIL [3448].
```

Only now does an audit fail which basically changed two plan runs ago?

georgesittas:
Hey, thanks for the report. We'll look into this shortly.

tobymao:
[...]

PR #3466, opened 2024-12-04
Fix!: always control error handling of audits in scheduler

georgesittas:
While working on https://github.com/TobikoData/sqlmesh/issues/3461, I noticed something strange: even though the audit is blocking, after running plan I got this warning:

```
2024-12-04 19:55:59,002 - MainThread - sqlmesh.core.snapshot.evaluator - WARNING - Audit 'is_positive' for model 'test_model' failed.
Got 1 results, expected 0.
SELECT * FROM (SELECT * FROM "db"."sqlmesh__default"."test_model__3395822052" AS "test_model__3395822052") AS "_q_0" WHERE "foobar" < 0
Audit is warn only so proceeding with execution. (evaluator.py:941)
```

After digging a bit, I realized this happened due to the `raise_exception=False` flag blocking a `raise` statement. Then, I noticed that in most execution paths where we reach `_audit`, this flag should be set to `False`.

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import abc
 import logging
-import traceback
 import typing as t
 
 from sqlglot import exp
@@ -245,7 +244,6 @@ class Scheduler:
             start=start,
             end=end,
             execution_time=execution_time,
-            raise_exception=False,
             snapshots=snapshots,
             deployability_index=deployability_index,
             wap_id=wap_id,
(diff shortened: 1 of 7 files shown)


tobymao (approved):

sqlmesh/core/scheduler.py
```diff
@@ -268,9 +266,10 @@ def evaluate(
                 )
             if audit_result.blocking:
                 audit_error_to_raise = error
+            else:
+                logger.warning(f"{error}\nAudit is warn only so proceeding with execution.")
```
izeigerman, line 270:
```suggestion
                logger.warning(f"{error}\nAudit is non-blocking only so proceeding with execution.")
```

PR #3468, opened 2024-12-04
Chore!: deprecate python 3.8, use public API for magics arg parsing

georgesittas:


setup.py
```diff
@@ -70,7 +70,7 @@
             "databricks-sql-connector",
             "dbt-bigquery",
             "dbt-core",
-            "dbt-duckdb>=1.7.1",
+            "dbt-duckdb>=1.7.1,!=1.9.1",
```
georgesittas, line 73: dbt-duckdb broke for python 3.8 after https://github.com/duckdb/dbt-duckdb/pull/482/files

georgesittas: OTOH they don't test python 3.8, it seems like their support starts at ≥3.9. So we may have to fix the tests instead so they won't run for 3.8...

georgesittas: Opened a ticket [upstream](https://github.com/duckdb/dbt-duckdb/issues/488) to track this.

sqlmesh/magics.py
```diff
@@ -60,8 +61,14 @@ def wrapper(self: SQLMeshMagics, *args: t.Any, **kwargs: t.Any) -> None:
         if bound_method:
             args_split = arg_split(args[0])
             parser = bound_method.parser
-            # Calling the private method to bypass setting of defaults.
-            parsed_args, _ = parser._parse_known_args(args_split, Namespace())
+
+            # Temporarily supress default values, otherwise any missing arg would be set and affect analytics
+            original_parser_actions = deepcopy(parser._actions)
```
izeigerman, line 66: Do we also need to clear `_defaults`?

izeigerman (approved):

sqlmesh/magics.py
georgesittas: I remember peeking at `_defaults` and it was empty, I'll revisit and cross-check with the current implementation.

georgesittas: I cleared it to be safe, since there seems to be a [code path](https://github.com/python/cpython/blob/main/Lib/argparse.py#L1942-L1945) where the namespace can be mutated if defaults are present.

georgesittas:
@tobymao @izeigerman a couple of interesting commits I found that could be refactored as well:

1. `add custom source parser because it's broken in < 3.9` -- https://github.com/TobikoData/sqlmesh/commit/d6f3c7cb3178ab7efc632f20cd63b7501abfce22
2. `allow python env building to use relative paths` -- https://github.com/TobikoData/sqlmesh/commit/92242e2de7bee0c7e89de1cfc7d956e18b1d339c

I think we've extended the logic since then ([example](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/metaprogramming.py#L37-L38)), but if you think it's worth it I'll try to remove them.

georgesittas:
Also - all checks have passed, I'm not sure why `ci/circleci: style_and_slow_tests-3.8` shows up - I got rid of it.

izeigerman:
> a couple of interesting commits I found that could be refactored as well:

@georgesittas  can we create an issue to refactor these and other places that are guarded by `< (3, 9)`

izeigerman:
@georgesittas , removed python 3.8 from mandatory checks

georgesittas:
> > a couple of interesting commits I found that could be refactored as well:
> 
> @georgesittas can we create an issue to refactor these and other places that are guarded by `< (3, 9)`

I've already removed everything that checked for `(3, 8)` and `>= (3, 9)`. I think those two are the last remaining pieces.

PR #3470, opened 2024-12-05
Fix: UI task position

### LINKED ISSUE #3469: UI tasks are in the wrong position
I've noticed when using the UI that there are a couple of points where the tasks are communicated in the wrong order.

As you can see in the screenshot 57 of 3 tasks
Screenshot 2024-12-04 at 15 49 19

The same is true when it comes to promoting tasks.
I believe I've seen where the issue is, PR to follow.

MikeWallis42:
Fixes #3469


georgesittas (approved):
Thanks for the fix @MikeWallis42, appreciate the PRs!

MikeWallis42:
@georgesittas, thanks for the speedy reviews.
I'm glad I can give back a little as we take so much from open source.

PR #3476, opened 2024-12-05
Fix: use python 3.9 for airflow dockerfile template

georgesittas:
Quick followup to the python 3.8 deprecation, I missed this one spot it seems..


benfdking (approved):

georgesittas:
Verified image builds and airflow tests pass locally when I run the CI check.

PR #3487, opened 2024-12-07
Fix: Prevent unintentional table schema changes during evaluation

izeigerman:
When restating forward-only models, SQLMesh uses the `CREATE OR REPLACE TABLE` semantics to recreate the underlying table. 

While performing this operation, SQLMesh used `columns_to_types` mapping provided by the user, if the user explicitly mapped all columns and their types in their model definition.

However, in case of forward-only models, the order of columns supplied by the user may not match the actual order of columns in the underlying table after applying schema changes. This is because very few engines support positional column alterations.

As a result, the schema of the table after restatement can be different from what it was before. This behavior is problematic because of the following reason:
1. The restatement only suppose to change rows but not the schema, which makes this behavior counterintuitive to users.
2. These unintentional schema changes are not reflected in the virtual layer, making the existing views outdated. This can result in subsequent failures when querying the views downstream.

### DIFF
--- a/sqlmesh/utils/pandas.py
+++ b/sqlmesh/utils/pandas.py
@@ -34,7 +34,11 @@ PANDAS_TYPE_MAPPINGS = {
 def columns_to_types_from_df(df: pd.DataFrame) -> t.Dict[str, exp.DataType]:
     result = {}
     for column_name, column_type in df.dtypes.items():
-        exp_type = PANDAS_TYPE_MAPPINGS.get(column_type)
+        exp_type: t.Optional[exp.DataType] = None
+        if hasattr(pd, "DatetimeTZDtype") and isinstance(column_type, pd.DatetimeTZDtype):
+            exp_type = exp.DataType.build("timestamp")
+        else:
+            exp_type = PANDAS_TYPE_MAPPINGS.get(column_type)
         if not exp_type:
             raise ValueError(f"Unsupported pandas type '{column_type}'")
         result[str(column_name)] = exp_type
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/pandas.py
```diff
@@ -34,7 +34,11 @@
 def columns_to_types_from_df(df: pd.DataFrame) -> t.Dict[str, exp.DataType]:
     result = {}
     for column_name, column_type in df.dtypes.items():
-        exp_type = PANDAS_TYPE_MAPPINGS.get(column_type)
+        exp_type: t.Optional[exp.DataType] = None
+        if hasattr(pd, "DatetimeTZDtype") and isinstance(column_type, pd.DatetimeTZDtype):
+            exp_type = exp.DataType.build("timestamp")
```
georgesittas, line 39: Why are we constructing a `Type.TIMESTAMP` instead of `Type.TIMESTAMPTZ`, given that the `column_type` is a timezone-aware datetime? Also, should we be `build`ing this `DataType` in a dialect-agnostic way?

izeigerman: This code was written before `DataType.build` required the dialect. This change is not actually necessary for my PR. Let me revert this and address this separately.

izeigerman: Actually, it looks like the dialect is only used to parse the input argument. Provided my values are hardcoded and correspond to the default dialect anyway I don't think I need an explicit dialect in here. I will switch to using TIMESTAMPZ

izeigerman: Fixed this

erindru (approved):
I have nothing to add to @georgesittas 's review, LGTM

georgesittas (approved):

PR #3492, opened 2024-12-09
Add Snowflake Tracking

sungchun12:
See how people are using SQLMesh with Snowflake. Should look and feel like this in the query history.
image

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -407,6 +407,7 @@ class SnowflakeConnectionConfig(ConnectionConfig):
     role: t.Optional[str] = None
     authenticator: t.Optional[str] = None
     token: t.Optional[str] = None
+    application: t.ClassVar[str] = "Tobiko_SQLMesh"
 
     # Private Key Auth
     private_key: t.Optional[t.Union[str, bytes]] = None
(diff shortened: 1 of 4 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -407,6 +408,7 @@ class SnowflakeConnectionConfig(ConnectionConfig):
     role: t.Optional[str] = None
     authenticator: t.Optional[str] = None
     token: t.Optional[str] = None
+    application: str = 'Tobiko_SQLMesh'
```
tobymao, line 411: this shouldn’t be configurable

sungchun12: made it a property

sungchun12: Nevermind, made it a classvar instead. Let me know if you want more protection on it.

sqlmesh/core/config/connection.py
```diff
@@ -407,6 +407,7 @@ class SnowflakeConnectionConfig(ConnectionConfig):
     role: t.Optional[str] = None
     authenticator: t.Optional[str] = None
     token: t.Optional[str] = None
+    application: t.ClassVar[str] = "Tobiko_SQLMesh"
```
georgesittas, line 410: I think doing `t.Literal["Tobiko_SQLMesh"] = "Tobiko_SQLMesh"` is more common in the codebase, but not sure if `t.ClassVar` semantics match the use case better.

Does this guard from the field being mutated, though? Perhaps we should add it to `_FIELD_UPDATE_STRATEGY` with the `IMMUTABLE` strategy? cc @tobymao

sungchun12: I'll assume Iaroslav's 👍 means it's worth changing this.

tobymao (approved):

PR #3493, opened 2024-12-10
feat(cli): ability to set config CLI params via env vars

kelsin:
Allows `SQLMESH_GATEWAY`, `SQLMESH_IGNORE_WARNINGS` and `SQLMESH_DEBUG` global options to be set via environment variables.

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -40,11 +40,13 @@ def _sqlmesh_version() -> str:
     "--gateway",
     type=str,
     help="The name of the gateway.",
+    envvar="SQLMESH_GATEWAY",
 )
 @click.option(
     "--ignore-warnings",
     is_flag=True,
     help="Ignore warnings.",
+    envvar="SQLMESH_IGNORE_WARNINGS",
 )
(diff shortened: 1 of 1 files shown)


sqlmesh/cli/main.py
```diff
@@ -40,16 +40,19 @@ def _sqlmesh_version() -> str:
...
     "--ignore-warnings",
     is_flag=True,
     help="Ignore warnings.",
+    envvar="SQLMESH_IGNORE_WARNINGS",
 )
 @click.option(
     "--debug",
     is_flag=True,
     help="Enable debug mode.",
+    envvar="SQLMESH_DEBUG",
```
izeigerman, line 55: Just a heads up that we already check it here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/__init__.py#L218

izeigerman (approved):

sqlmesh/cli/main.py
kelsin: Ok so that one isn't needed here, will unset :)

PR #3494, opened 2024-12-10
Fix: quote the table produced by _resolve_table

georgesittas:
During the recent refactor, where `this_model` changed from a `str` into a `Table` instance, I missed the fact that we were previously [quoting](https://github.com/TobikoData/sqlmesh/commit/b00d7d40b42ff759243b161567a81547c1002486#diff-fddedea4eda285dbd51b83821f3cd54dc42d3946c0056760acdc29efc3e565afL242) it, so I didn't match that behavior.

We discovered that this led to an issue where the resolved table for `this_model` would be normalized twice, if e.g. it was interpolated in a string, and hence generated as a table with unquoted identifiers.

This PR fixes the above issue by having `_resolve_table` quote the table after resolving it, to match the semantics of `_resolve_tables`.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -240,7 +240,7 @@ class BaseExpressionRenderer:
         table_mapping: t.Optional[t.Dict[str, str]] = None,
         deployability_index: t.Optional[DeployabilityIndex] = None,
     ) -> exp.Table:
-        return exp.replace_tables(
+        table = exp.replace_tables(
             exp.maybe_parse(table_name, into=exp.Table, dialect=self._dialect),
             {
                 **self._to_table_mapping((snapshots or {}).values(), deployability_index),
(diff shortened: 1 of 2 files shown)


sqlmesh/core/renderer.py
```diff
@@ -249,6 +249,11 @@ def _resolve_table(
             dialect=self._dialect,
             copy=False,
         )
+        # We quote the table here to mimic the behavior of _resolve_tables, otherwise we may end
+        # up normalizing twice, because _to_table_mapping returns the mapped names unquoted.
```
izeigerman, line 253: Is there any test we can come up with which results in unexpected behavior if the table was normalized twice?

georgesittas: Perhaps we could have a test where `@this_model` is rendered in the context of a Snowflake model and the schema becomes `"sqlmesh_FOO"`. Without this fix we'd get `"SQLMESH_FOO"`.

georgesittas: (I'm planning to implement the test soon)

georgesittas: Aight I gave it a whirl, see https://github.com/TobikoData/sqlmesh/pull/3494/commits/5c05c4d73cf798527b0d8b7158813aed2783fecb

PR #3496, opened 2024-12-10
feat: make log output more human-friendly

plaflamme:
The current output looks like this:

```
Inserting batch (1733702400000, 1733788800000) into ...
```

This will output something like:

```
Inserting batch (2024-12-09 00:00:00, 2024-12-10 00:00:00) into ...
```

This is what the output used to look like prior to https://github.com/TobikoData/sqlmesh/pull/3405


eakmanrq (approved):
Thanks!

georgesittas (approved):

PR #3497, opened 2024-12-10
Refactor!: make when_matched syntax compatible with merge syntax

georgesittas:
Before this PR, the `when_matched` attribute couldn't be parsed consistently. E.g., a user [reported](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1733343291111989) that, even though the following definition worked, switching the two clauses so that `THEN DELETE` is first would result in a parse error.

```sql
MODEL (
  name @{macro_val}.test,
    unique_key purchase_order_id,
    when_matched
    WHEN MATCHED AND source._operation <> 1 THEN UPDATE SET target.purchase_order_id = 1
    WHEN MATCHED AND source._operation = 1 THEN DELETE
  )
);

select purchase_order_id from @{macro_val}.upstream;
```

This happened because we relied on some implicit behavior around how the clauses are consumed ([example](https://github.com/TobikoData/sqlmesh/pull/3497#discussion_r1878883456)).

This PR refactors the `when_matched` syntax, so that parentheses are used as explicit delimiters. It also makes a few changes to the API (triggered by the relevant SQLGlot [changes](https://github.com/tobymao/sqlglot/commit/051c6f085e426dbcdda3fbc324a180ada268355f)), in order to represent multiple `WHEN [NOT] MATCHED` clauses with a single `Whens` expression, instead of a list of `When` expressions.

- [x] Fix MSSQL test related to `DATETIME` and `DATETIME2` ([context](https://github.com/tobymao/sqlglot/commit/ab108518c53173ddf71ac1dfd9e45df6ac621b81))
- [x] Check if there are additional side-effects caused by the above SQLGlot change
- [x] Add migration script

### DIFF
--- a/docs/concepts/models/model_kinds.md
+++ b/docs/concepts/models/model_kinds.md
@@ -320,7 +320,9 @@ MODEL (
   name db.employees,
   kind INCREMENTAL_BY_UNIQUE_KEY (
     unique_key name,
-    when_matched WHEN MATCHED THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary)
+    when_matched (
+      WHEN MATCHED THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary)
+    )
   )
 );
 ```
@@ -334,8 +336,10 @@ MODEL (
   name db.employees,
   kind INCREMENTAL_BY_UNIQUE_KEY (
     unique_key name,
-    when_matched WHEN MATCHED AND source.value IS NULL THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary),
-    WHEN MATCHED THEN UPDATE SET target.title = COALESCE(source.title, target.title)
+    when_matched (
+      WHEN MATCHED AND source.value IS NULL THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary)
+      WHEN MATCHED THEN UPDATE SET target.title = COALESCE(source.title, target.title)
+    )
   )
 );
 ```
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
         "rich[jupyter]",
         "ruamel.yaml",
         "setuptools; python_version>='3.12'",
-        "sqlglot[rs]~=25.34.1",
+        "sqlglot[rs]~=26.0.0",
         "tenacity",
     ],
     extras_require={
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -409,13 +409,12 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
         return None
 
     name = key.name.lower()
-    if name == "when_matched":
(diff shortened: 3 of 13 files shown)


setup.py
```diff
@@ -48,7 +48,7 @@
         "rich[jupyter]",
         "ruamel.yaml",
         "setuptools; python_version>='3.12'",
-        "sqlglot[rs]~=25.34.1",
+        "sqlglot[rs]~=26.0.0",
```
erindru, line 51: I think this has created a regression in mssql, datetimes being truncated to milliseconds

georgesittas: Yeah, most likely. I'm not sure yet if this is a regression or if we simply need to update the SQLMesh side to match the updated SQLGlot logic. Either way, will triage soon and patch as needed.

cc @VaggelisD

docs/concepts/models/model_kinds.md
```diff
@@ -334,8 +334,10 @@ MODEL (
   name db.employees,
   kind INCREMENTAL_BY_UNIQUE_KEY (
     unique_key name,
-    when_matched WHEN MATCHED AND source.value IS NULL THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary),
-    WHEN MATCHED THEN UPDATE SET target.title = COALESCE(source.title, target.title)
+    when_matched (
```
tobymao, line 337: do you need the parenthesis?

georgesittas: Yeah, I think we do. For example:

```sql
MODEL (
  name db.employees,
    when_matched WHEN MATCHED THEN UPDATE SET target.salary = COALESCE(source.salary, target.salary),
    unique_key name
  )
);

SELECT 'name' AS name, 1 AS salary;
```

This fails to parse because the comma is consumed while parsing the `UPDATE SET` [clause](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py#L7047-L7051), so you end up consuming `unique_key` and `name` looks like a stray, unknown property afterwards:

```
[Property(
  this=when_matched,
  value=[
    When(
      matched=True,
      then=Update(
        expressions=[
          EQ(
            this=Column(
              this=Identifier(this=salary, quoted=False),
              table=Identifier(this=target, quoted=False)),
            expression=Coalesce(
              this=Column(
                this=Identifier(this=salary, quoted=False),
                table=Identifier(this=source, quoted=False)),
              expressions=[
                Column(
                  this=Identifier(this=salary, quoted=False),
                  table=Identifier(this=target, quoted=False))])),
          Column(
            this=Identifier(this=unique_key, quoted=False))]))])] <~~~
```

I think models that use `when_matched` must've been working by coincidence until now, e.g. by having it placed at the end of the model kind's attributes.

The parentheses solve this issue by making the delimiters around the `WHEN MATCHED` clauses explicit.

setup.py
georgesittas: Fixed the test as we discussed @erindru.

izeigerman (approved):

PR #3498, opened 2024-12-10
Fix(bigquery): Pass catalog when checking for clustering key changes

erindru:


.circleci/continue_config.yml
```diff
@@ -340,15 +340,15 @@ workflows:
...
                 - bigquery
-                - clickhouse-cloud
-                - athena
-          filters:
-           branches:
-             only:
-               - main
+                #- clickhouse-cloud
+                #- athena
+          #filters:
```
erindru, line 348: todo: revert

note that Snowflake also inherits `ClusteredByMixin` so ive enabled Snowflake tests for this PR too

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -356,10 +357,15 @@ def get_alter_expressions(
         current_table = exp.to_table(current_table_name)
         target_table = exp.to_table(target_table_name)
 
+        current_table_schema = schema_(current_table.db, catalog=current_table.catalog)
```
erindru, line 360: I observed that this was the function that the `SnapshotEvaluator` uses when it calls `get_data_objects()` to build a schema-with-catalog Table object.

The `Table` object cant be passed directly or `to_schema()` produces a `db.table` value instead of a `catalog.db` value

izeigerman (approved):

PR #3501, opened 2024-12-11
Chore: Doc outlines for tcloud observability

crericha:
- Added observability feature broken down by where it is used
- Added screenshot
- Bulleted outline of model freshness and observability
- List of observability functionality for each location without tcloud

@treysp @mesmith027 let me know how this looks at a starting point. Is there more that needs to be fleshed out for you guys to convert into documentation?


crericha:
@treysp to take on next iteration before merging. Closing for now.

PR #3511, opened 2024-12-13
fix: ensure that restatements in prod also trigger restatements in dev

erindru:


sqlmesh/core/plan/evaluator.py
```diff
@@ -358,9 +359,27 @@ def _restate(self, plan: EvaluatablePlan, snapshots_by_name: t.Dict[str, Snapsho
         if not plan.restatements:
             return
 
+        snapshots_to_restate = {
+            snapshots_by_name[name].table_info for name in plan.restatements.keys()
```
georgesittas, line 363:
```suggestion
            snapshots_by_name[name].table_info for name in plan.restatements
```

georgesittas (approved):
Nice, LGTM 👍

sqlmesh/core/plan/evaluator.py
```diff
@@ -358,9 +359,27 @@ def _restate(self, plan: EvaluatablePlan, snapshots_by_name: t.Dict[str, Snapsho
...
+            # restating intervals on prod plans should mean that the intervals are cleared across all environments,
+            # not just the version currently in prod
+            # this ensures that work done in dev environments can still be promoted to prod by forcing dev environments to
+            # re-run intervals that changed in prod
+            # without this rule, its possible that promoting a dev table to prod will introduce old data to prod
+            all_envs = self.state_sync.get_environments()
+            all_matching_snapshots = [
+                s
+                for s in itertools.chain(*[e.snapshots for e in all_envs])
+                if s.name in plan.restatements
```
izeigerman, line 376: We should check not only if it's in restatements but also if any of its parents are in there. Then you need to add it there. Example:

I have models A <- B <- C in prod and I restate A.
In dev I added 2 new models D and E so that my DAG looks like A <- B <- C <- D <- E. We need to make sure that D and E are also restated despite not being a part of prod.

There's, however, an open question of what intervals to restate for these. I don't think we can figure this out without fetching full snapshot payloads. The good news is that it shouldn't happen that frequently and we only need to fetch if the snapshot name is missing in the list but its parent is there

tobymao: do we need to do this in batch?

sqlmesh/core/plan/evaluator.py
erindru: This gets me every time :) thanks for flagging

sqlmesh/core/plan/evaluator.py
erindru: Oh man, I see. Ok i'll add a test case for that and have a go at solving it

> There's, however, an open question of what intervals to restate for these. 
Wouldn't it be the same intervals that are being restated in the upstream table?

> do we need to do this in batch?
Do you mean fetching the full snapshots? From what I can tell, the EngineAdapterStateSync [automatically batches](https://github.com/TobikoData/sqlmesh/blob/2ac88947cf5767932661826aeac0eb828901334c/sqlmesh/core/state_sync/engine_adapter.py#L1556) the fetch requests if you give it a big long list of snapshot ids

izeigerman: > Wouldn't it be the same intervals that are being restated in the upstream table?

Not necessarily. For example, if a model that is being restated is hourly, but the downstream (D or E) is daily you'd need to extend the restated interval potentially. Or perhaps if a partial interval is removed, SQLMesh already does the right thing, but it needs to be verified.

tests/core/test_integration.py
```diff
@@ -1651,6 +1651,572 @@ def test_restatement_plan_ignores_changes(init_and_plan_context: t.Callable):
     context.apply(plan)
 
 
+def test_restatement_plan_hourly_with_downstream_daily_restates_correct_intervals(tmp_path: Path):
```
erindru, line 1654: This test is for an existing bug where restating in the same environment still doesn't work properly if the interval being restated is smaller than the interval_unit of one of the models

sqlmesh/core/plan/definition.py
```diff
@@ -284,6 +284,10 @@ def is_selected_for_backfill(self, model_fqn: str) -> bool:
     def plan_id(self) -> str:
         return self.environment.plan_id
 
+    @property
+    def is_prod(self) -> bool:
```
tobymao, line 288: nit: seems unecessary

erindru: The double negative of `not self._is_dev` all over the place to indicate `prod` was bothering me. But I can revert this if you prefer

sqlmesh/core/plan/evaluator.py
erindru: > if a partial interval is removed, SQLMesh already does the right thing

It kind of did. Within the restatement plan it didnt (it would just skip the snapshots with partial intervals), but if you immediately do a cadence run after the restatement plan it would process the correct intervals.

I've added some code to widen the intervals for affected snapshots as part of the restatement plan and a test to show it working

sqlmesh/core/plan/definition.py
tobymao: meh up to you

izeigerman (approved):
Fantastic work 🚀

Issue #3479, opened 2024-12-06
create model option and defaults to not optimize


tobymao:
optimizing takes time and can increase sql payload size.

make a model option to control the level of optimization per model and at a project level.

cc @VaggelisD

esciara:
Hi.

Could you explain or point to the definition (doc?) of "optimize" that is used here?

I am asking because as this optimization is certainly [not for size of the sql](https://github.com/TobikoData/sqlmesh/issues/3472#issuecomment-2521799118), nor for speed as I could observe that it could take quite a while to generate the "optimized" sql.

Thx

georgesittas:
> Hi.
> 
> Could you explain or point to the definition (doc?) of "optimize" that is used here?
> 
> I am asking because as this optimization is certainly [not for size of the sql](https://github.com/TobikoData/sqlmesh/issues/3472#issuecomment-2521799118), nor for speed as I could observe that it could take quite a while to generate the "optimized" sql.
> 
> Thx

There are some transformations applied to the query that simplify it, e.g. by folding constants. These transformations are part of [SQLGlot's optimizer](https://github.com/tobymao/sqlglot/tree/main/sqlglot/optimizer), hence the term.

izeigerman:
I don't think the default should be not optimziing

georgesittas:
Yep, I don't think it is, @VaggelisD made it so that it's only opt-out by setting the flag (or model default) to false.

PR #3512, opened 2024-12-13
Feat!: Support 'optimize' flag in model defs

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3479


docs/concepts/models/overview.md
```diff
@@ -354,6 +354,11 @@ Learn more about these properties and their default values in the [model configu
 ### gateway
 :   Specifies the gateway to use for the execution of this model. When not specified, the default gateway is used.
 
+### optimize
+:   Whether the model's SQL should be optimized. This attribute is `true` by default. Setting it
+to `false` causes SQLMesh to disable query canonicalization & simplification. This should be turned off only if the optimized query leads to errors such as surpassing text limit.
```
georgesittas, line 359: @tobymao this feature is probably gonna be niche, but I wonder if we can optimize at load time to build the mapping schemas and then only switch off optimization when rendering a query to execute it. The motivation being that setting `optimize false` also impacts the column-level lineage functionality, so trying to see if there's a way to avoid it.

sqlmesh/core/config/model.py
```diff
@@ -45,6 +46,7 @@ class ModelDefaultsConfig(BaseConfig):
     on_destructive_change: t.Optional[OnDestructiveChange] = None
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
     audits: t.Optional[t.List[FunctionCall]] = None
+    optimize: bool = True
```
georgesittas, line 49: I think we wanna do `t.Optional[bool] = None`, otherwise you're always including bool in the payload unnecessarily. So `None` and `True` should lead to optimization, whereas an explicit `False` should not.

sqlmesh/core/model/meta.py
```diff
@@ -78,6 +78,7 @@ class ModelMeta(_Node):
     enabled: bool = True
     physical_version: t.Optional[str] = None
     gateway: t.Optional[str] = None
+    optimize: t.Optional[bool] = None
```
georgesittas, line 81: Should this be part of the model's data hash?

docs/concepts/models/overview.md
tobymao: i think that's ok, if they want lineage, they can fully qualify. i think the point of this flag is not not optimize at all, even at load time

tobymao: let's put a note saying here that doing this will disable column level lineage unless you manually qualify and expand all stars

sqlmesh/core/model/meta.py
VaggelisD: Does the metadata hash include it automatically or should I include it manually?

tobymao: you need to include it manually

docs/concepts/models/overview.md
georgesittas: Yeah I can see what you mean, my hunch is also that this won't really be used _that_ much in practice, anyway. 

Though my understanding was that the issue is related to the payload's size when sending queries to the engine, so that wouldn't affect the loading stage.

sqlmesh/core/model/definition.py
```diff
@@ -1949,6 +1951,11 @@ def _create_model(
     model_schema_name = exp.to_table(name, dialect=dialect).db
     physical_schema_override: t.Optional[str] = None
 
+    if not defaults:
+        # Set query optimization to "True" by default, unless specified otherwise by
+        # the user in the Model flags
+        kwargs["optimize"] = kwargs.get("optimize", True)
```
tobymao, line 1957: why was this added here? is this the proper way to do config defaults? why aren't there other places that do this pattern?

VaggelisD: It doesn't look like it's needed yeah, will remove

docs/concepts/models/overview.md
```diff
@@ -354,6 +354,12 @@ Learn more about these properties and their default values in the [model configu
...
+
+  NOTE: Turning off the optimizer will also disable column-level lineage, unless the query is manually qualified and all `SELECT` stars are expanded.
```
georgesittas, line 361:
```suggestion
!!! warning
    Turning off the optimizer may prevent column-level lineage from working for the affected model and its descendants, unless all columns in the model's query are qualified and it contains no star projections (e.g. `SELECT *`).
```

sqlmesh/core/model/definition.py
```diff
@@ -934,6 +935,7 @@ def metadata_hash(self) -> str:
                 self.project,
                 str(self.allow_partials),
                 gen(self.session_properties_) if self.session_properties_ else None,
+                str(self.optimize) if self.optimize is not None else None,
```
georgesittas, line 938: Not sure why we need the branch here:
```suggestion
                str(self.optimize),
```

sqlmesh/core/renderer.py
```diff
@@ -65,6 +66,7 @@ def __init__(
         self.update_schema({} if schema is None else schema)
         self._cache: t.List[t.Optional[exp.Expression]] = []
         self._model_fqn = model_fqn
+        self._optimize = True if optimize is None else optimize
```
georgesittas, line 69:
```suggestion
        self._optimize = optimize is not False
```

sqlmesh/core/model/definition.py
VaggelisD: Reverted that entirely

sqlmesh/core/model/definition.py
```diff
@@ -881,6 +882,7 @@ def _data_hash_values(self) -> t.List[str]:
             self.physical_version,
             self.gateway,
             self.interval_unit.value if self.interval_unit is not None else None,
+            self.optimize,
```
georgesittas, line 885: This needs to be converted into a string right? The items in this list are supposed to be strings, why does it work if this is a bool?

VaggelisD: Right, on the metadata hash there was an explicit error regarding the conversion, here it wasn't apparent. Will switch to `str(self.optimize)`

georgesittas (approved):
Let's test this end-to-end as well, i.e. the whole plan-apply flow.

tests/core/test_model.py
```diff
@@ -4186,7 +4186,7 @@ def test_default_catalog_sql(assert_exp_eq):
 
 
 def test_default_catalog_python():
-    HASH_WITH_CATALOG = "663490914"
+    HASH_WITH_CATALOG = "2071559843"
```
tobymao, line 4189: i know you didn't write these but could we fix these hardcoded hash tests so they aren't hardcoded anymore

VaggelisD: It seems to me that these particular cases test exactly that, so we (1) can't know the hash/version before to do the assertion or (2) we do not want to propagate it as it could silently change. 

Though, I think we could modify the following:

- `tests/core/test_snapshot_evaluator.py`: Add `f"{snapshot.version}"` to the physical names

- `tests/core/test_state_sync.py`: The ordering here is affected by the values, so we could sort & replace them from least to greatest

tobymao: ok

georgesittas (approved):
This needs to be marked as breaking because a migration's gonna be needed.

sqlmesh/core/config/model.py
```diff
@@ -45,6 +46,7 @@ class ModelDefaultsConfig(BaseConfig):
     on_destructive_change: t.Optional[OnDestructiveChange] = None
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
     audits: t.Optional[t.List[FunctionCall]] = None
+    optimize: t.Optional[bool] = None
```
izeigerman, line 49: Isn't this only applicable to SQL model? Should we make sure this is not set for models of other kinds?

sqlmesh/core/model/definition.py
```diff
@@ -881,6 +882,7 @@ def _data_hash_values(self) -> t.List[str]:
             self.physical_version,
             self.gateway,
             self.interval_unit.value if self.interval_unit is not None else None,
+            str(self.optimize),
```
izeigerman, line 885:
```suggestion
            self.optimize if self.optimize is not None else None,
```

sqlmesh/core/config/model.py
VaggelisD: Yep, it does make sense for it to be exclusive to `SQLModel`. Moving it there would add some (small) complexity such as the need to override `data_hash_values` afaict; Is this fine or would it be better to leave it and instead emit a warning/error?

georgesittas: Should be fine, we already do this for other attributes exclusive to it.

VaggelisD: Ended up leaving it in `ModelMeta` and added a validation check

tobymao (approved):

sqlmesh/core/model/definition.py
```diff
@@ -840,6 +841,12 @@ def validate_definition(self) -> None:
                 self._path,
             )
 
+        if not self.is_sql and self.optimize_query is not None:
+            raise_config_error(
+                "SQLMesh query optimizer can only be enabled/disabled for SQLModels",
```
izeigerman, line 846:
```suggestion
                "SQLMesh query optimizer can only be enabled/disabled for SQL models",
```

PR #3514, opened 2024-12-13
Fix: store max file modification time for dbt tests

georgesittas:
Similar to: https://github.com/TobikoData/sqlmesh/blob/434be409ab4ced4bfa39fea19a5f60ba85a36b2b/sqlmesh/core/loader.py#L497-L526

### DIFF
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -71,6 +71,8 @@ class DbtLoader(Loader):
     def __init__(self) -> None:
         self._projects: t.List[Project] = []
         self._macros_max_mtime: t.Optional[float] = None
+        self._audits_max_mtime: t.Optional[float] = None
+
         super().__init__()
 
     def load(self, context: GenericContext, update_schemas: bool = True) -> LoadedProject:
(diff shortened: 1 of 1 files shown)


tobymao (approved):

izeigerman:
I don't think this is enough. Shouldn't you use the new value when invalidating the cache here: https://github.com/TobikoData/sqlmesh/blob/fa045ec235055eca406b81041dfd8fe1058fac03/sqlmesh/dbt/loader.py#L302?

georgesittas:
> I don't think this is enough. Shouldn't you use the new value when invalidating the cache here:
> 
> https://github.com/TobikoData/sqlmesh/blob/fa045ec235055eca406b81041dfd8fe1058fac03/sqlmesh/dbt/loader.py#L302
> ?

Thanks for the heads up. 👍
This was supposed to be a draft but forgot to mark as such; I haven't tested it yet.

georgesittas:
It looks like we [treat dbt custom generic tests as macros](https://github.com/TobikoData/sqlmesh/blob/fa045ec235055eca406b81041dfd8fe1058fac03/sqlmesh/dbt/manifest.py#L152-L153), which is possible because they also do something similar (*) in their manifest.json file, so cache invalidation should already work in main branch: the `_macros_max_mtime` will be updated when a test is changed.

I believe this makes this PR redundant (tested locally and theory checks out), so I'll go ahead and close it for now.

(*) they basically replace their raw code to a call that delegates to a `test_`-prefixed macro with the original body.

PR #3516, opened 2024-12-13
Chore: switch tests from freezegun to time-machine

treysp:
Libraries used for freezing time in the sqlmesh unit tests.

Change motivated by freezegun's inability to handle timezones.

Notes:
- Default behavior
  - freezegun default is for time to be totally frozen inside the context, such that repeatedly capturing `now()` would give the same value every time
  - time_machine default is for time to continue inside the context
  - in the tests where it's relevant, we specify `tick=False` to retain freezegun's default behavior
 - Time zones
   - time_machine infers tz as UTC for objects that are not tz-aware
   - If a tz-aware object is present without an explicit tz, the machine's local tz will be used
   - We manually specify UTC everywhere to avoid mixing tz-aware and non-tz-aware objects

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ setup(
         "croniter",
         "duckdb!=0.10.3",
         "dateparser",
-        "freezegun",
         "hyperscript>=0.1.0",
         "importlib-metadata; python_version<'3.12'",
         "ipywidgets",
@@ -50,6 +49,7 @@ setup(
         "setuptools; python_version>='3.12'",
         "sqlglot[rs]~=26.0.0",
         "tenacity",
+        "time-machine",
     ],
     extras_require={
         "athena": ["PyAthena[Pandas]"],
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
         "croniter",
         "duckdb!=0.10.3",
         "dateparser",
+        "freezegun",
         "hyperscript>=0.1.0",
         "importlib-metadata; python_version<'3.12'",
         "ipywidgets",
@@ -49,7 +50,6 @@ setup(
         "setuptools; python_version>='3.12'",
         "sqlglot[rs]~=26.0.0",
         "tenacity",
-        "time-machine",
     ],
     extras_require={
         "athena": ["PyAthena[Pandas]"],
@@ -94,6 +94,7 @@ setup(
             "ruff~=0.7.0",
             "snowflake-connector-python[pandas,secure-local-storage]>=3.0.2",
             "sqlalchemy-stubs",
+            "time-machine",
             "types-croniter",
             "types-dateparser",
             "types-python-dateutil",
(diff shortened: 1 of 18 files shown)


tests/cli/test_cli.py
```diff
@@ -692,11 +692,12 @@ def test_run_cron_elapsed(runner, tmp_path):
     create_example_project(tmp_path)
 
     # Create and backfill `prod` environment
-    with freeze_time("2023-01-01 23:59:00"):
+    with time_machine.travel("2023-01-01 23:59:00 UTC", tick=False) as traveler:
```
eakmanrq, line 695: Why was `UTC` needed in this test but not in other places?

tests/core/test_integration.py
```diff
@@ -505,7 +505,7 @@ def test_hourly_model_with_lookback_no_backfill_in_dev(init_and_plan_context: t.
         ]
 
 
-@freeze_time("2023-01-08 00:00:00")
+@time_machine.travel("2023-01-08 00:00:00 UTC", tick=False)
```
eakmanrq, line 508: Why was tick required here but not other tests?

tests/cli/test_cli.py
treysp: Based on this from the docs, I had assumed it wouldn't be needed anywhere.

After some testing, my inference is that if nothing is passed to a `datetime`/`time` method anywhere in the call stack everything will automatically be UTC.

If a tz-aware object is passed, the local timezone can come into play. When that happens, we sometimes need to manually specify the UTC tz.

```
A datetime.datetime. If it is naive, it will be assumed to have the UTC timezone. If it has tzinfo set to a [zoneinfo.ZoneInfo instance](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfo), the current timezone will also be mocked.
```

tests/core/test_integration.py
treysp: In some tests, we only capture time at one point - so the result is the same whether or not time keeps going

treysp: But if we capture time twice, we'll only get the same value if `tick=False`.

The default in freezegun was `tick=False`, but it's `True` in time-machine.

setup.py
```diff
@@ -94,6 +94,7 @@
             "ruff~=0.7.0",
             "snowflake-connector-python[pandas,secure-local-storage]>=3.0.2",
             "sqlalchemy-stubs",
+            "time-machine",
```
erindru, line 97: This adds `time_machine`, do we need to remove `freezegun` from `install_requires` as well or is it still used somewhere?

treysp: It's used in core/tests/definition.py so people can freeze time in their SQLMesh tests.

By default, the tests are run locally in DuckDB. 

freezegun seems to be correctly interoperating with DuckDB's default tz handling, so it seems dangerous to switch to something fully tz-aware (which will therefore require explicit tz handling, possibly by the user).

eakmanrq: So it might make sense to update that too to time-machine but yeah it could be a breaking change and if so makes sense to leave out-of-scope of this PR.

eakmanrq (approved):

PR #3518, opened 2024-12-15
Check that test_adapter exist before closing

Kayrnt:
It looks like that
```
model_to_test = self.get_model(model, raise_if_missing=True)
```

got introduced https://github.com/TobikoData/sqlmesh/pull/3394 but it might fail and result in an exception that would prevent the creation of the `test_adapter` that we are trying to close.

(it happened on my end as I was tinkering on my project)


erindru (approved):
Thanks!

PR #3524, opened 2024-12-16
Feat!: Add support for virtual statements to be executed post update

themisvaltinos:
This update introduces statements that are executed after a virtual update. These can be used for example to grant privileges on views of the virtual layer. 

These expressions should be defined within a `ON_VIRTUAL_UPDATE_BEGIN; ...; ON_VIRTUAL_UPDATE_END;` block. For example:

```sql
MODEL (
  name db.customers 
);

SELECT 
  item_id
FROM 
  db.seed_model;

ON_VIRTUAL_UPDATE_BEGIN; 

GRANT SELECT ON VIEW @this_model TO ROLE role_name;

INJA_STATEMENT_BEGIN;     
GRANT REFERENCES, SELECT ON FUTURE VIEWS IN DATABASE {{ db }} TO ROLE owner_name;  
JINJA_END;  

ON_VIRTUAL_UPDATE_END;

```

Jinja expressions are also supported within this block by properly nesting the Jinja block as shown in the example above.

In these statements table resolution occurs at the virtual level, meaning that qualified view names (including `@this_model`) are used instead of the physical table names. For example, when running the plan in an environment named `dev`, the `db.customers` as well as `this_model` would resolve to `db__dev.customers` rather than the physical table.

For python models these statements are defined in the `@model` [decorator](https://wiki.python.org/moin/PythonDecorators):

  ```python
@model( 
	"db.customers",  
	on_virtual_update=["GRANT SELECT ON VIEW @this_model TO ROLE role_name;"]
)
```

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -808,7 +808,7 @@ def _is_virtual_statement_end(tokens: t.List[Token], pos: int) -> bool:
     return _is_command_statement(ON_VIRTUAL_UPDATE_END, tokens, pos)
 
 
-def virtual_statement(statement: exp.Expression) -> VirtualUpdateStatement:
+def virtual_statement(statement: t.List[exp.Expression]) -> VirtualUpdateStatement:
     return VirtualUpdateStatement(this=statement)
 
 
@@ -861,9 +861,10 @@ def parse(
     while pos < total:
         token = tokens[pos]
         if _is_virtual_statement_end(tokens, pos):
-            pos += 2
+            chunks[-1][0].append(token)
             virtual = False
             chunks.append(([], ChunkType.SQL))
+            pos += 2
         elif _is_jinja_end(tokens, pos) or (
             chunks[-1][1] == ChunkType.SQL
             and token.token_type == TokenType.SEMICOLON
@@ -875,17 +876,14 @@ def parse(
                 # Jinja end statement
                 chunks[-1][0].append(token)
                 pos += 2
-            if virtual and tokens[pos] != ON_VIRTUAL_UPDATE_END:
-                # This is required for nested Jinja statements that precede
-                # SQL statements within an ON_VIRTUAL_UPDATE block
-                chunks.append(
-                    (
-                        [Token(TokenType.VAR, text=ON_VIRTUAL_UPDATE_BEGIN)],
-                        ChunkType.VIRTUAL_STATEMENT,
-                    )
+            chunks.append(
+                (
+                    [],
+                    ChunkType.VIRTUAL_STATEMENT
+                    if virtual and tokens[pos] != ON_VIRTUAL_UPDATE_END
+                    else ChunkType.SQL,
                 )
-            else:
-                chunks.append(([], ChunkType.SQL))
+            )
         elif _is_jinja_query_begin(tokens, pos):
             chunks.append(([token], ChunkType.JINJA_QUERY))
             pos += 2
@@ -893,7 +891,7 @@ def parse(
             chunks.append(
                 (
                     [token],
-                    ChunkType.VIRTUAL_JINJA_STATEMENT if virtual else ChunkType.JINJA_STATEMENT,
+                    ChunkType.JINJA_STATEMENT,
                 )
             )
             pos += 2
@@ -908,32 +906,65 @@ def parse(
     parser = dialect.parser()
     expressions: t.List[exp.Expression] = []
 
-    for chunk, chunk_type in chunks:
-        if chunk_type == ChunkType.SQL:
-            parsed_expressions: t.List[t.Optional[exp.Expression]] = (
-                parser.parse(chunk, sql) if into is None else parser.parse_into(into, chunk, sql)
-            )
-            for expression in parsed_expressions:
-                if expression:
-                    expression.meta["sql"] = parser._find_sql(chunk[0], chunk[-1])
-                    expressions.append(expression)
-        elif chunk_type == ChunkType.VIRTUAL_STATEMENT:
-            sql_chunk = chunk[1:-1]
-            for expression in parser.parse(sql_chunk, sql):
-                if expression:
-                    expression.meta["sql"] = expression.sql(dialect=dialect)
-                    expressions.append(virtual_statement(expression))
+    def parse_sql_chunk(chunk: t.List[Token]) -> t.List[exp.Expression]:
+        parsed_expressions: t.List[t.Optional[exp.Expression]] = (
+            parser.parse(chunk, sql) if into is None else parser.parse_into(into, chunk, sql)
+        )
+        expressions = []
+        for expression in parsed_expressions:
+            if expression:
+                expression.meta["sql"] = parser._find_sql(chunk[0], chunk[-1])
+                expressions.append(expression)
+        return expressions
+
+    def parse_jinja_chunk(chunk: t.List[Token]) -> exp.Expression:
+        start, *_, end = chunk
+        segment = sql[start.end + 2 : end.start - 1]
+        factory = jinja_query if chunk_type == ChunkType.JINJA_QUERY else jinja_statement
+        expression = factory(segment.strip())
+        meta_sql = sql[start.start : end.end + 1]
+        expression.meta["sql"] = meta_sql
+        return expression
+
+    def parse_virtual_statement(
+        chunks: t.List[t.Tuple[t.List[Token], ChunkType]], pos: int
+    ) -> t.Tuple[t.List[exp.Expression], int]:
+        # For virtual statements we need to handle both SQL and Jinja nested blocks within the chunk
+        virtual_update_statements = []
+        start = chunks[pos][0][0].start
+
+        while chunks[pos - 1][0] == [] or (
+            chunks[pos - 1][0] and chunks[pos - 1][0][-1].text != ON_VIRTUAL_UPDATE_END
+        ):
+            chunk, chunk_type = chunks[pos]
+            if chunk_type == ChunkType.JINJA_STATEMENT:
+                virtual_update_statements.append(parse_jinja_chunk(chunk))
+            else:
+                virtual_update_statements.extend(
+                    parse_sql_chunk(chunk[int(chunk[0].text == "ON_VIRTUAL_UPDATE_BEGIN") : -1])
+                )
+            pos += 1
+
+        if virtual_update_statements:
+            statement = virtual_statement(virtual_update_statements)
+            end = chunk[-1].end + 1
+            statement.meta["sql"] = sql[start:end]
+            return [statement], pos
+
+        return [], pos
+
+    pos = 0
+    total_chunks = len(chunks)
+    while pos < total_chunks:
+        chunk, chunk_type = chunks[pos]
+        if chunk_type == ChunkType.VIRTUAL_STATEMENT:
+            virtual_expression, pos = parse_virtual_statement(chunks, pos)
+            expressions.extend(virtual_expression)
+        elif chunk_type == ChunkType.SQL:
+            expressions.extend(parse_sql_chunk(chunk))
         else:
-            start, *_, end = chunk
-            segment = sql[start.end + 2 : end.start - 1]
-            factory = jinja_query if chunk_type == ChunkType.JINJA_QUERY else jinja_statement
-            expression = factory(segment.strip())
-            expression.meta["sql"] = sql[start.start : end.end + 1]
-            expressions.append(
-                virtual_statement(expression)
-                if chunk_type == ChunkType.VIRTUAL_JINJA_STATEMENT
-                else expression
-            )
+            expressions.append(parse_jinja_chunk(chunk))
+        pos += 1
 
     return expressions
 
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -929,14 +929,16 @@ def parse(
         virtual_update_statements = []
         start = chunks[pos][0][0].start
 
-        while chunks[pos - 1][0] == [] or chunks[pos - 1][0][-1].text != ON_VIRTUAL_UPDATE_END:
+        while (
+            chunks[pos - 1][0] == [] or chunks[pos - 1][0][-1].text.upper() != ON_VIRTUAL_UPDATE_END
+        ):
             chunk, chunk_type = chunks[pos]
             if chunk_type == ChunkType.JINJA_STATEMENT:
                 virtual_update_statements.append(parse_jinja_chunk(chunk, False))
             else:
                 virtual_update_statements.extend(
                     parse_sql_chunk(
-                        chunk[int(chunk[0].text == "ON_VIRTUAL_UPDATE_BEGIN") : -1], False
+                        chunk[int(chunk[0].text.upper() == ON_VIRTUAL_UPDATE_BEGIN) : -1], False
                     ),
                 )
             pos += 1
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -62,7 +62,7 @@ class JinjaStatement(Jinja):
 
 
 class VirtualUpdateStatement(exp.Expression):
-    pass
+    arg_types = {"expressions": True}
 
 
 class ModelKind(exp.Expression):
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: sqlmesh/core/snapshot/evaluator.py [hunk 1]
```diff
@@ -985,6 +986,47 @@ def _get_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
...
+            snapshots.values(),
+            environment_naming_info,
+            default_catalog=default_catalog,
+            dialect=self.adapter.dialect,
+        )
+
+        for snapshot in target_snapshots:
+            adapter = self._get_adapter(snapshot.model_gateway)
+            snapshot_deps = {snapshots[p_sid].name: snapshots[p_sid] for p_sid in snapshot.parents}
+            snapshot_deps[snapshot.name] = snapshot
```
georgesittas, line 1015: Where is this used / needed?

tests/core/test_model.py
```diff
@@ -6266,3 +6266,174 @@ def assert_metadata_only():
...
+            "JINJA_STATEMENT_BEGIN; GRANT SELECT ON VIEW {{this_model}} TO ROLE admin; JINJA_END;"
+        ),
+        *d.parse(
+            "GRANT REFERENCES, SELECT ON FUTURE VIEWS IN DATABASE demo_db TO ROLE owner_name;"
+        ),
+        *d.parse("@resolve_parent_name('parent')"),
+        *d.parse(
+            "GRANT SELECT ON VIEW demo_db.table /* sqlglot.meta replace=false */ TO ROLE admin;"
+        ),
+    ]
```
georgesittas, line 6339: We can do this in a single `d.parse` call, right? It produces a list of expressions, one for each statement, so you also don't have to expand it but just use it directly.

tests/core/test_model.py
```diff
@@ -6266,3 +6266,174 @@ def assert_metadata_only():
...
+
+    assert model.on_virtual_update == expected_virtual_statements
+
+    assert parent.on_virtual_update == [
+        *d.parse(
+            "JINJA_STATEMENT_BEGIN; GRANT SELECT ON VIEW {{this_model}} TO ROLE admin; JINJA_END;"
+        )
+    ]
```
georgesittas, line 6347: No need to expand the result here, it's already a list.

tests/core/test_model.py
```diff
@@ -6266,3 +6266,174 @@ def assert_metadata_only():
     model = load_sql_based_model(expressions, signal_definitions=signal.get_registry())
     model.signals.clear()
     assert_metadata_only()
+
+
+def test_model_on_virtual_update(make_snapshot: t.Callable):
```
georgesittas, line 6271: Should we add an end-to-end test for this?

sqlmesh/core/model/definition.py
```diff
@@ -2086,7 +2104,9 @@ def _split_sql_model_statements(
             loaded_audit = load_audit([expr, expressions[idx + 1]], dialect=dialect)
             assert isinstance(loaded_audit, ModelAudit)
             inline_audits[loaded_audit.name] = loaded_audit
-            idx += 2
+            idx += 1
```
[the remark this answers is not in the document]
themisvaltinos: so it doesn't have to be added one more time for the virtual statements case when increasing the index. it does the same thing so I can add it back how it was

sqlmesh/core/dialect.py
```diff
@@ -61,6 +61,10 @@ class JinjaStatement(Jinja):
     pass
 
 
+class VirtualStatement(exp.Expression):
+    pass
```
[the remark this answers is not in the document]
themisvaltinos: sure, I didn't want the name to be lengthy but can change it to `VirtualUpdateStatement`

sqlmesh/core/plan/evaluator.py
```diff
@@ -323,6 +325,17 @@ def _update_views(
...
+
+            if promoted_snapshots := [
+                s for s in added_snapshots if s.is_model and not s.is_symbolic
+            ]:
+                self._virtual_statements(
+                    plan,
+                    promoted_snapshots,
+                    snapshots,
+                    deployability_index,
+                )
```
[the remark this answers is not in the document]
themisvaltinos: right will inline it instead

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: thank you for noticing my bad this was supposed to be passed in the kwargs instead of `snapshots`

sqlmesh/core/dialect.py
```diff
@@ -850,13 +894,23 @@ def parse(
                 if expression:
                     expression.meta["sql"] = parser._find_sql(chunk[0], chunk[-1])
                     expressions.append(expression)
+        elif chunk_type == ChunkType.VIRTUAL_STATEMENT:
+            sql_chunk = chunk[1:-1]
+            for expression in parser.parse(sql_chunk, sql):
+                if expression:
+                    expression.meta["sql"] = expression.sql(dialect=dialect)
```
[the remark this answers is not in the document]
themisvaltinos: yes didn't consider this refactoring to preserve the original text instead

sqlmesh/core/dialect.py
```diff
@@ -831,36 +857,95 @@ def parse(
...
             chunks.append(([token], ChunkType.JINJA_QUERY))
             pos += 2
         elif _is_jinja_statement_begin(tokens, pos):
-            chunks.append(([token], ChunkType.JINJA_STATEMENT))
+            chunks.append(
+                (
+                    [token],
+                    ChunkType.JINJA_STATEMENT,
+                )
+            )
```
georgesittas, line 877: [Nit - opinion] I'd reformat this to the single-line form.

sqlmesh/core/dialect.py
```diff
@@ -831,36 +857,95 @@ def parse(
...
+
+    def parse_virtual_statement(
+        chunks: t.List[t.Tuple[t.List[Token], ChunkType]], pos: int
+    ) -> t.Tuple[t.List[exp.Expression], int]:
+        # For virtual statements we need to handle both SQL and Jinja nested blocks within the chunk
+        virtual_update_statements = []
+        start = chunks[pos][0][0].start
+
+        while chunks[pos - 1][0] == [] or (
+            chunks[pos - 1][0] and chunks[pos - 1][0][-1].text != ON_VIRTUAL_UPDATE_END
```
georgesittas, line 918: Don't we know that `chunks[pos - 1][0]` is a non-empty list because of the first condition?
```suggestion
            chunks[pos - 1][0][-1].text != ON_VIRTUAL_UPDATE_END
```

sqlmesh/core/dialect.py
```diff
@@ -831,36 +857,95 @@ def parse(
...
+        ):
+            chunk, chunk_type = chunks[pos]
+            if chunk_type == ChunkType.JINJA_STATEMENT:
+                virtual_update_statements.append(parse_jinja_chunk(chunk))
+            else:
+                virtual_update_statements.extend(
+                    parse_sql_chunk(chunk[int(chunk[0].text == "ON_VIRTUAL_UPDATE_BEGIN") : -1])
```
georgesittas, line 925: It's more readable if you do `chunk_start = 1 if chunk[0].text == ... else 0`. Current style is _too_ C-like. :)

sqlmesh/core/dialect.py
```diff
@@ -775,10 +781,24 @@ def jinja_statement(statement: str) -> JinjaStatement:
...
+def _is_virtual_statement_begin(tokens: t.List[Token], pos: int) -> bool:
+    return _is_command_statement(ON_VIRTUAL_UPDATE_BEGIN, tokens, pos)
+
+
+def _is_virtual_statement_end(tokens: t.List[Token], pos: int) -> bool:
+    return _is_command_statement(ON_VIRTUAL_UPDATE_END, tokens, pos)
+
+
+def virtual_statement(statement: t.List[exp.Expression]) -> VirtualUpdateStatement:
+    return VirtualUpdateStatement(this=statement)
```
georgesittas, line 793: If you're using `VirtualUpdateStatement` as a wrapper around a list of `Expression`s, then you should override `arg_types` so it only contains `"expressions": True`. Using `this` to store a list is misleading, because it's commonly used to store a single `Expression` instance.

sqlmesh/core/model/definition.py
```diff
@@ -1064,6 +1075,7 @@ class SqlModel(_Model):
         query: The main query representing the model.
         pre_statements: The list of SQL statements that precede the model's query.
         post_statements: The list of SQL statements that follow after the model's query.
+        on_virtual_update: The list of SQL statements to be executed after virtual update.
```
georgesittas, line 1078:
```suggestion
        on_virtual_update: The list of SQL statements to be executed after the virtual update.
```

sqlmesh/core/snapshot/evaluator.py
georgesittas, line 1015: I wonder if we wanna make a helper for this, there are a few places where we do the same thing in the codebase:

- https://github.com/TobikoData/sqlmesh/blob/2adf5a1ba71deb6d0db63c8304f350b0a6d00c7e/sqlmesh/core/snapshot/evaluator.py#L722-L725
- https://github.com/TobikoData/sqlmesh/blob/2adf5a1ba71deb6d0db63c8304f350b0a6d00c7e/sqlmesh/core/snapshot/evaluator.py#L821-L824
- https://github.com/TobikoData/sqlmesh/blob/2adf5a1ba71deb6d0db63c8304f350b0a6d00c7e/sqlmesh/core/scheduler.py#L164-L167

cc @izeigerman

sqlmesh/core/dialect.py
themisvaltinos: Right thank you revising this

izeigerman:
Can we please make sure this feature is documented before we merge.

sqlmesh/core/model/definition.py
```diff
@@ -914,7 +925,7 @@ def _data_hash_values(self) -> t.List[str]:
             data.append(key)
             data.append(gen(value))
 
-        for statement in (*self.pre_statements, *self.post_statements):
+        for statement in (*self.pre_statements, *self.post_statements, *self.on_virtual_update):
```
izeigerman, line 928: I don't think this should impact the data hash

sqlmesh/core/model/definition.py
```diff
@@ -1007,7 +1018,7 @@ def _additional_metadata(self) -> t.List[str]:
         if metadata_only_macros:
             additional_metadata.append(str(metadata_only_macros))
 
-        for statement in (*self.pre_statements, *self.post_statements):
+        for statement in (*self.pre_statements, *self.post_statements, *self.on_virtual_update):
```
izeigerman, line 1021: We should include `on_virtaul_update` expressions into the metadata hash, but here it won't be included because of the `self._is_metadata_statement` check. We should just do something like:
```
for statement in self.on_virtual_update:
    additional_metadata.append(gen(statement))
 ```

File: sqlmesh/core/plan/evaluator.py [hunk 16]
```diff
@@ -322,6 +324,21 @@ def _update_views(
                     promotion_result.removed_environment_naming_info,
                     on_complete=lambda s: self.console.update_promotion_progress(s, False),
                 )
+
+            if promoted_snapshots := [
+                s for s in added_snapshots if s.is_model and not s.is_symbolic
+            ]:
+                self.snapshot_evaluator._execute_virtual_statements(
```
izeigerman, line 331: Why are we doing this here and not here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/evaluator.py#L1240?

themisvaltinos: There is a check for `is_model` before calling promote here: 
https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/evaluator.py#L845
and my thinking behind the check that it is not `is_symbolic` was that we wouldn't have virtual statements for external or embedded models. Should I only check for `is_model` instead?

sqlmesh/core/model/definition.py
themisvaltinos: thank you I missed the is_metadata_statement check revising this

sqlmesh/core/plan/evaluator.py
izeigerman: What I'm saying that the whole `_execute_virtual_statements` logic should happen as part of the `PromotableStrategy`. This strategy won't be invoked for symbolic models.

themisvaltinos: Ok I understand now, I’ll move the logic there then so that it happens after its view is updated

sqlmesh/core/dialect.py
```diff
@@ -850,36 +876,91 @@ def parse(
...
+        start = chunks[pos][0][0].start
+
+        while chunks[pos - 1][0] == [] or chunks[pos - 1][0][-1].text != ON_VIRTUAL_UPDATE_END:
+            chunk, chunk_type = chunks[pos]
+            if chunk_type == ChunkType.JINJA_STATEMENT:
+                virtual_update_statements.append(parse_jinja_chunk(chunk, False))
+            else:
+                virtual_update_statements.extend(
+                    parse_sql_chunk(
+                        chunk[int(chunk[0].text == "ON_VIRTUAL_UPDATE_BEGIN") : -1], False
```
georgesittas, line 939: Can `chunk[0].text` ever be lowercase? e.g. `on_virtual_update_begin`. We may need to uppercase this first.

georgesittas (approved):

sqlmesh/core/dialect.py
themisvaltinos: Yes it can be lowercase my oversight. I've adapted the code by uppercasing to account for it in the check. Thanks for noticing!

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1247,13 +1243,18 @@ def promote(
...
+        adapter.create_view(
             view_name,
             exp.select("*").from_(table_name, dialect=self.adapter.dialect),
             table_description=model.description if is_prod else None,
             column_descriptions=model.column_descriptions if is_prod else None,
             view_properties=model.virtual_properties,
         )
+        if on_virtual_update := model.on_virtual_update:
+            adapter.execute(
+                model._render_statements(on_virtual_update, engine_adapter=adapter, **kwargs)
```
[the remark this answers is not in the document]
izeigerman: Also I don't think `kwargs` contains anything meaningful here. This is mostly for future expansions, since this interface is public. No need to pass in into renderer I don't think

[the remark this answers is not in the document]
themisvaltinos: Right changing it to use a public method instead similar to pre/post statements

Regarding the `kwargs` I am passing the table mapping of qualified view names, which is then propagated down to `render` and used here:
https://github.com/TobikoData/sqlmesh/blob/c1d117e4105fbd6a730e46164f849246f308522d/sqlmesh/core/renderer.py#L382
so it is needed to resolve model names to their corresponding `view` names of the current environment, so I'm not sure without passing them how I could do this

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -853,7 +859,18 @@ def _promote_snapshot(
                 view_name=view_name,
                 model=snapshot.model,
                 environment=environment_naming_info.name,
+                deployability_index=deployability_index,
```
izeigerman, line 862: Do we still need to pass deployability_index here?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1662,6 +1662,21 @@ def to_table_mapping(
     }
 
 
+def to_view_mapping(
+    snapshots: t.Iterable[Snapshot],
+    environment_naming_info: EnvironmentNamingInfo,
+    default_catalog: t.Optional[str] = None,
+    dialect: t.Optional[str] = None,
+) -> t.Dict[str, str]:
+    return {
```
izeigerman, line 1671: [Nit] I guess we can return empty dict here if the environment is prod

izeigerman (approved):
This looks great, thank you!

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: oh right we don't will remove this

PR #3528, opened 2024-12-17
Feat: add this_model property in the macro evaluator to return a string

georgesittas:
Provides an easier path to migrate away from old code that used `locals["this_model"]`.


eakmanrq (approved):

tobymao (approved):

izeigerman (approved):
thank you!

PR #3529, opened 2024-12-17
Feat: Add support for auto-restatements

izeigerman:
It’s prevalent among dbt users to nuke the tables of their incremental models occasionally to ensure correctness as well as to capture late arriving data. This is because dbt models lack an idempotent way to re-process arbitrary intervals of data. 

When transitioning to SQLMesh, these users are forced to use restatement plans in order to replicate workflows they are used to in dbt. However, the restatement plans are highly invasive, as they remove intervals for all versions of the model across all environments. This negatively impacts users iterating in development environments, as they are suddenly forced to rerun models they haven’t modified.

To address this issue, we are introducing first-class support for this use case and call it Auto-Restatements. This allows users to configure a cadence for full (or partial) refreshes of incremental models, which will be automatically applied as part of the `sqlmesh run` command when appropriate.

### DIFF
--- a/docs/concepts/models/overview.md
+++ b/docs/concepts/models/overview.md
@@ -402,6 +402,49 @@ Some properties are only available in specific model kinds - see the [model conf
 ### disable_restatement
 :   Set this to true to indicate that [data restatement](../plans.md#restatement-plans) is disabled for this model.
 
+### auto_restatement_cron
+:   A cron expression that determines when SQLMesh should automatically restate this model. Restatement means re-evaluating either a number of last intervals (controlled by [`auto_restatement_intervals`](#auto_restatement_intervals)) for model kinds that support it or the entire model for model kinds that don't. Downstream models that depend on this model will also be restated. The auto-restatement is only applied when running the `sqlmesh run` command against the production environment.
+
+    A common use case for auto-restatement is to periodically re-evaluate a model (less frequently than the model's cron) to account for late-arriving data or dimension changes. However, relying on this feature is generally not recommended, as it often indicates an underlying issue with the data model or dependency chain. Instead, users should prefer setting the [`lookback`](#lookback) property to handle late-arriving data more effectively.
+
+    Unlike the [`lookback`](#lookback) property, which only controls the time range of data scanned, auto-restatement rewrites all previously processed data for this model in the target table.
+
+    For model kinds that don't support [`auto_restatement_intervals`](#auto_restatement_intervals) the table will be re-created from scratch.
+
+    Models with [`disable_restatement`](#disable_restatement) set to `true` will not be restated automatically even if this property is set.
+
+    **NOTE**: Models with this property set can only be [previewed](../plans.md#data-preview-for-forward-only-changes) in development environments, which means that the data computed in those environments will not be reused in production.
+
+    ```sql linenums="1" hl_lines="6"
+    MODEL (
+      name test_db.national_holidays,
+      cron '@daily',
+      kind INCREMENTAL_BY_UNIQUE_KEY (
+        unique_key key,
+        auto_restatement_cron '@weekly',
+      )
+    );
+    ```
+
+### auto_restatement_intervals
+:   The number of last intervals to restate automatically. This is only applied in conjunction with [`auto_restatement_cron`](#auto_restatement_cron).
+
+    If not specified, the entire model will be restated.
+
+    This property is only supported for the `INCREMENTAL_BY_TIME_RANGE` model kind.
+
+    ```sql linenums="1" hl_lines="7"
+    MODEL (
+      name test_db.national_holidays,
+      cron '@daily',
+      kind INCREMENTAL_BY_TIME_RANGE (
+        time_column event_ts,
+        auto_restatement_cron '@weekly',
+        auto_restatement_intervals 7, -- automatically restate the last 7 days of data
+      )
+    );
+    ```
+
 ## Macros
 Macros can be used for passing in parameterized arguments such as dates, as well as for making SQL less repetitive. By default, SQLMesh provides several predefined macro variables that can be used. Macros are used by prefixing with the `@` symbol. For more information, refer to [macros](../macros/overview.md).
 
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+import sys
 import typing as t
 from collections import defaultdict
 from datetime import datetime, timedelta
 from enum import IntEnum
+import logging
 from functools import cached_property, lru_cache
 from pathlib import Path
 
@@ -27,6 +29,7 @@ from sqlmesh.utils.date import (
     make_exclusive,
     now,
     now_timestamp,
+    time_like_to_str,
     to_date,
     to_datetime,
     to_ds,
@@ -49,6 +52,9 @@ Interval = t.Tuple[int, int]
 Intervals = t.List[Interval]
 
 Node = t.Annotated[t.Union[Model, StandaloneAudit], Field(discriminator="source_type")]
+
+
+logger = logging.getLogger(__name__)
 
 
 class SnapshotChangeCategory(IntEnum):
@@ -163,6 +169,7 @@ class SnapshotIntervals(PydanticModel, frozen=True):
     version: str
     intervals: Intervals
     dev_intervals: Intervals
+    pending_restatement_intervals: Intervals = []
 
     @property
     def snapshot_id(self) -> SnapshotId:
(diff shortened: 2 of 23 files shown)


docs/concepts/models/overview.md
```diff
@@ -395,6 +395,47 @@ Some properties are only available in specific model kinds - see the [model conf
...
 
+### auto_restatement_cron
+:   A cron expression that determines when SQLMesh should automatically restate this model. Restatement means re-evaluating either a number of last intervals (controlled by [`auto_restatement_intervals`](#auto_restatement_intervals)) for model kinds that support it or the entire model for model kinds that don't. Downstream models that depend on this model will also be restated. The auto-restatement is only applied when running the `sqlmesh run` command against the production environment.
+
+    A use case for auto-restatement is to re-evaluate a model periodically (less frequently than the model's cron) to account for late-arriving data and / or dimension changes. Instead of setting the [`lookback`](#lookback) property, which applies during each model evaluation, users can leverage auto-restatements to process all late-arriving data in one go, frontloading the cost.
+
+    Unlike the [`lookback`](#lookback) property, which only controls the time range of data scanned, auto-restatement rewrites all previously processed data for this model in the target table.
```
tobymao, line 403: we should strongly discourage people from using this as it has negative tradeoffs

sungchun12: Would it be fair to say, "Use this as a last resort. If you're on the fence using this feature, please reach out on our slack community https://tobikodata.com/slack"?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1865,6 +1919,86 @@ def snapshots_to_dag(snapshots: t.Collection[Snapshot]) -> DAG[SnapshotId]:
...
+                time_like_to_str(next_auto_restated_interval[0]),
+                time_like_to_str(next_auto_restated_interval[1]),
+                snapshot.snapshot_id,
+            )
+            auto_restated_intervals.append(next_auto_restated_interval)
+
+        normalized_auto_restated_intervals = []
+        for interval in merge_intervals(auto_restated_intervals):
+            interval_start = snapshot.node.interval_unit.cron_floor(interval[0])
+            interval_end = snapshot.node.interval_unit.cron_floor(interval[1])
```
tobymao, line 1964: what is this section doing?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1149,29 +1226,62 @@ def _get_snapshot_intervals(
...
+                        # Remove all pending restatement intervals recorded before the current interval has been added
+                        pending_restatement_intervals[(name, version)] = remove_interval(
+                            pending_restatement_intervals[(name, version)], start, end
+                        )
+                        last_seen_identifier_per_version[(name, version)] = identifier
 
             for name, identifier, version in {**intervals, **dev_intervals}:
+                # Only set the pending restatement intervals for the last seen snapshot per version
+                # FIXME: Remove this logic once intervals are stored per (name, version)
+                pending_restatement_intervals_for_snapshot = (
```
tobymao, line 1272: are there any issues with only storing this on the latest one?

sqlmesh/migrations/v0065_add_auto_restatements.py
```diff
@@ -0,0 +1,46 @@
...
+
+from sqlglot import exp
+
+from sqlmesh.utils.migration import index_text_type
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
+    engine_adapter = state_sync.engine_adapter
+    schema = state_sync.schema
+    auto_restatements_table = "_auto_restatements"
```
tobymao, line 11: this entire table is only going to store auto restatements? i could see us storing other things on the snapshot version level

docs/concepts/models/overview.md
izeigerman: How do you suggest expressing this?

sqlmesh/core/snapshot/definition.py
izeigerman: If the upstream model is hourly and the downstream is daily, this expands the interval to restate a complete interval for the downstream model.
If it's `01/02 10:00:00` right now, and we restate 24 intervals of the hourly model we need to make sure we restate the whole `01/01` for the daily model. This handles the partial interval

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Haven't found any yet. This is because we add these to each snapshot that share the same version. So the identifier doesn't really matter

sqlmesh/migrations/v0065_add_auto_restatements.py
izeigerman: Yes, I can see that too. When the time comes we can rename it. Today, I don't have other use cases for it.

docs/concepts/models/overview.md
tobymao: Using this indicates a problem with data modeling. You should not being doing joins early in your pipelines because dimensions change, keep things normalized until as late as possible.

In terms of late arriving data, you should use lookback.

When you opt into this feature, you lose deployability, meaning no promotion from dev to prod. This is because the data is no longer consistent.

sqlmesh/core/snapshot/definition.py
```diff
@@ -1033,6 +1020,64 @@ def needs_destructive_check(
...
+        num_intervals_to_restate = self.model.auto_restatement_intervals
+        if num_intervals_to_restate is None:
+            return (self.intervals[0][0], self.intervals[-1][1])
+
+        auto_restatement_end_ts = to_timestamp(
+            self.node.interval_unit.cron_floor(next_auto_restatement_ts)
+        )
+        auto_restatement_start_ts = (
+            auto_restatement_end_ts
+            - num_intervals_to_restate * self.node.interval_unit.milliseconds
```
tobymao, line 1056: i'm not sure we use this shortcut anywhere. wonder if this has issues with irregular crons

sqlmesh/core/snapshot/definition.py
```diff
@@ -1865,6 +1919,86 @@ def snapshots_to_dag(snapshots: t.Collection[Snapshot]) -> DAG[SnapshotId]:
     return dag
 
 
+def apply_auto_restatements(
+    snapshots: t.Dict[SnapshotId, Snapshot], execution_time: TimeLike
+) -> t.List[SnapshotIntervals]:
```
tobymao, line 1924: can we check that this method doesn't have performance issues?, afraid of the cron calls etc

sqlmesh/core/snapshot/definition.py
```diff
@@ -1865,6 +1919,86 @@ def snapshots_to_dag(snapshots: t.Collection[Snapshot]) -> DAG[SnapshotId]:
...
+            for interval in auto_restated_intervals_per_snapshot[parent_s_id]
+        ]
+        if next_auto_restated_interval:
+            logger.info(
+                "Calculated the next auto restated interval (%s, %s) for snapshot %s",
+                time_like_to_str(next_auto_restated_interval[0]),
+                time_like_to_str(next_auto_restated_interval[1]),
+                snapshot.snapshot_id,
+            )
+            auto_restated_intervals.append(next_auto_restated_interval)
```
tobymao, line 1959: i'm not sure i understand this logic.

so you find the auto restated intervals of your parents, but then you add self next auto restated_intervals?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1033,6 +1020,64 @@ def needs_destructive_check(
...
+        )
+        if execution_time_ts < next_auto_restatement_ts:
+            return None
+
+        num_intervals_to_restate = self.model.auto_restatement_intervals
+        if num_intervals_to_restate is None:
+            return (self.intervals[0][0], self.intervals[-1][1])
+
+        auto_restatement_end_ts = to_timestamp(
+            self.node.interval_unit.cron_floor(next_auto_restatement_ts)
```
tobymao, line 1052: check this for edge case of not running for a while. need to respect now

tobymao (approved):

sqlmesh/migrations/v0065_add_auto_restatements.py
georgesittas: Just a heads up that there's another 0065 script [here](https://github.com/TobikoData/sqlmesh/pull/3512).

docs/concepts/models/overview.md
izeigerman: reworded

sqlmesh/migrations/v0065_add_auto_restatements.py
izeigerman: thanks!

tobymao (approved):

PR #3531, opened 2024-12-17
Fix: Add adapters for commands that are not using the snapshot evaluator

themisvaltinos:
This fixes an omission on the latest multi-engine feature, to create the engine adapters in cases when they're not generated in the snapshot_evaluator property, ie for the commands `render`, `table_diff` and `create_test`

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2023,6 +2023,12 @@ class GenericContext(BaseContext, t.Generic[C]):
             if snapshot.is_model and snapshot.model.gateway
         }
 
+    @cached_property
+    def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
+        """Returns all engine adapters for the gateways defined in the configs."""
+        self._create_engine_adapters()
+        return self._engine_adapters
+
     def _create_engine_adapters(self, gateways: t.Optional[t.Set] = None) -> None:
         """Create engine adapters for the gateways, when none provided include all defined in the configs."""
 
@@ -2037,8 +2043,7 @@ class GenericContext(BaseContext, t.Generic[C]):
 
     def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
         if gateway:
-            self._create_engine_adapters()
-            if adapter := self._engine_adapters.get(gateway):
+            if adapter := self.engine_adapters.get(gateway):
                 return adapter
             raise SQLMeshError(f"Gateway '{gateway}' not found in the available engine adapters.")
         return self.engine_adapter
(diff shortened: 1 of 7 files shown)


erindru:
No existing tests picked up this regression, are you able to add a test?

sqlmesh/core/context.py
```diff
@@ -2032,6 +2034,7 @@ def _create_engine_adapters(self, gateways: t.Optional[t.Set] = None) -> None:
 
     def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
         if gateway:
+            self._create_engine_adapters()
```
georgesittas, line 2037: Is this recomputed every time? Should we cache this?

themisvaltinos: Yes, adapted this

sqlmesh/core/context.py
```diff
@@ -2018,6 +2020,12 @@ def _snapshot_gateways(self) -> t.Dict[str, str]:
             if snapshot.is_model and snapshot.model.gateway
         }
 
+    @cached_property
+    def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
+        """Returns all engine adapters for the gateways defined in the configs."""
+        self._create_engine_adapters()
+        return self._engine_adapters
```
georgesittas, line 2027: Should we move the logic in `_create_engine_adapters` here?

cc @izeigerman does this look safe to cache?

themisvaltinos: I considered this, only thing was that `_create_engine_adapters` takes the `gateways` as an argument and for the snapshot_evaluator we were only creating the ones used in the project rather than all.

themisvaltinos: it was only used in the `snapshot_evaluator` before, when initialising the SnapshotEvaluator

georgesittas (approved):

izeigerman (approved):

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -1731,6 +1733,8 @@ def _normalize_snowflake(name: str, prefix_regex: str = "(sqlmesh__)(.*)"):
     if config.model_defaults.dialect != ctx.dialect:
         config.model_defaults = config.model_defaults.copy(update={"dialect": ctx.dialect})
 
+    # To enable parallelism in integration tests
```
izeigerman, line 1736: Why does this impact parallelism?

izeigerman (approved):

tests/core/engine_adapter/integration/test_integration.py
themisvaltinos: To isolate the connections per test because otherwise they will encounter concurrency issues, as they will end up using the same connections. Since all the engine adapters are created upfront and this particular file's tests are using for the context the same config: https://github.com/TobikoData/sqlmesh/blob/main/tests/core/engine_adapter/integration/config.yaml

PR #3533, opened 2024-12-18
Refactor: remove freezegun dependency in favor of time-machine

### LINKED ISSUE #3522: Investigate switching sqlmesh core from freezegun to time-machine
The `freezegun` library allows you to "freeze time" in Python, which is valuable for reproducibility in unit tests.

SQLMesh users can specify an execution time (or other datetime variable) in their SQLMesh tests. Core implements this functionality using `freezegun`.

`freeezegun` has limited support for timezone handling, but the `time-machine` library has full support. We switched the sqlmesh unit test suite to `time-machine` in #3516 for this reason.

That PR did not switch sqlmesh core to `time-machine`, but we will want to at some point. The actual change is simple, the real task will be understanding the implications for DuckDB interoperability, etc.

georgesittas:
Closes #3522

### DIFF
--- a/tests/core/test_test.py
+++ b/tests/core/test_test.py
@@ -1236,12 +1236,12 @@ test_foo:
 
     @model("py_model", columns={"ts1": "timestamptz", "ts2": "timestamptz"})
     def execute(context, start, end, execution_time, **kwargs):
-        datetime_now = datetime.datetime.now()
+        datetime_now_utc = datetime.datetime.now(tz=datetime.timezone.utc)
 
         context.engine_adapter.execute(exp.select("CURRENT_TIMESTAMP"))
         current_timestamp = context.engine_adapter.cursor.fetchone()[0]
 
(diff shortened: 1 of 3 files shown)


tests/core/test_test.py
```diff
@@ -1236,12 +1236,12 @@ def test_freeze_time(mocker: MockerFixture) -> None:
 
     @model("py_model", columns={"ts1": "timestamptz", "ts2": "timestamptz"})
     def execute(context, start, end, execution_time, **kwargs):
-        datetime_now = datetime.datetime.now()
+        datetime_now_utc = datetime.datetime.now(datetime.UTC)
```
georgesittas, line 1239: The `datetime.now` function [produces a local datetime by default](https://docs.python.org/3/library/datetime.html#datetime.datetime.now), so the python model test here was failing with `time-travel`, whereas `freezegun` worked because it patched it to `FakeDatetime(2023, 1, 1, 10, 5, 3)`.

cc @treysp

treysp (approved):

PR #3534, opened 2024-12-18
Fix!: reject string model names

### LINKED ISSUE #3517: More explicit errors on string literal usage of model definitions
Coming out of this discussion: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1734109818343679

It would be really helpful to have more explicit errors on single quote usage (string literals) in sql models, as the arising errors from misuse are very confusing.

### Case 1
I had 2 sql models defined, one as
```
MODEL (
    name sqlmesh_example.runescape_mapping,
    ...
```
and the other as
```
MODEL (
    name 'sqlmesh_example.hourly_scrape'
    ...
```

and was getting the error
```
❯ uv run sqlmesh plan dev
2024-12-13 12:08:32,038 - MainThread - sqlmesh.core.loader - ERROR - SQLMesh requires all model names and references to have the same level of nesting. (loader.py:634)
Error: Table "db"."sqlmesh_example"."runescape_mapping" must match the schema's nesting level: 1.
```

The issue was the single quotes breaking the schema's nest level (one dot) but that was really difficult to debug

### Use case 2
In that model, the `kind` was defined as
```
MODEL (
    name sqlmesh_example.hourly_scrape,
        time_column 'ts'
    ),
    -- kind FULL,
    start '2024-12-12',
    cron '@hourly',
    grain (item_id, ts),
...
```

with the time column defined in single quotes. On running, I was getting the error
```
duckdb.duckdb.ConversionException: Conversion Error: timestamp field value out of range: "ts", expected format is (YYYY-MM-DD 
HH:MM:SS[.US][±HH:MM| ZONE])
LINE 1: ...ample__hourly_scrape__189843799" where 'ts' >= TIMESTAMP '2024-12-12 00:00:00+...
```

I changed the model kind to `FULL` and stopped getting the error, but it was unclear that the issue was simply that the timestamp column was literally `'ts'`. 

Thank you!

georgesittas:
Fixes #3517 -- only "case 1", I didn't prioritize fixing the `time_column` one because I thought there are many cases where we expected columns/identifiers in the model meta, so that's probably a bigger effort. The model name was low-hanging fruit to avoid inconsistent schema nesting issues.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -116,6 +116,7 @@ def _parse_statement(self: Parser) -> t.Optional[exp.Expression]:
         return None
 
     parser = PARSERS.get(self._curr.text.upper())
+    error_msg = None
 
     if parser:
         # Capture any available description in the form of a comment
@@ -125,7 +126,8 @@ def _parse_statement(self: Parser) -> t.Optional[exp.Expression]:
         try:
             self._advance()
             meta = self._parse_wrapped(lambda: t.cast(t.Callable, parser)(self))
-        except ParseError:
+        except ParseError as parse_error:
+            error_msg = parse_error.args[0]
             self._retreat(index)
 
         # Only return the DDL expression if we actually managed to parse one. This is
@@ -135,7 +137,12 @@ def _parse_statement(self: Parser) -> t.Optional[exp.Expression]:
             meta.comments = comments
             return meta
 
-    return self.__parse_statement()  # type: ignore
+    try:
+        return self.__parse_statement()  # type: ignore
+    except ParseError:
+        if error_msg:
+            raise ParseError(error_msg)
+        raise
 
 
 def _parse_lambda(self: Parser, alias: bool = False) -> t.Optional[exp.Expression]:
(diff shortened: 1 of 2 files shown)


tests/core/test_dialect.py
```diff
@@ -571,7 +571,7 @@ def test_model_normalization_multiple_serde(
     expressions = parse(
         f"""
         MODEL (
-            name {table},
+            name {exp.maybe_parse(table, into=exp.Table).sql(dialect=normalization_dialect)},
```
georgesittas, line 574: This needed to be changed because when the normalization dialect is `mysql`, we produce `db."TABLE"` and double quotes actually denote strings. So we first parse using the base dialect and generate the proper identifier quotes per dialect.

sqlmesh/core/dialect.py
```diff
@@ -555,6 +562,11 @@ def parse(self: Parser) -> t.Optional[exp.Expression]:
 
             if key in table_keys:
                 value = self._parse_table_parts()
+                if value and self._prev.token_type == TokenType.STRING:
+                    self.raise_error(
+                        f"'{key}' property cannot be a string value: {value}. "
+                        "Please use the identifier syntax instead, e.g. foo.bar instead of 'foo.bar'"
```
erindru, line 568: Nice, this will save a bunch of user confusion

erindru (approved):

tobymao (approved):

izeigerman:
This is a breaking change, isn't it?

PR #3535, opened 2024-12-19
feat!: exclusions in selector powered by full parsing

tobymao:


tobymao:
need to add a test for exclusions

sqlmesh/core/selector.py
```diff
@@ -162,163 +166,183 @@ def expand_model_selections(
...
+        models = models or self._models
+        models_by_tags: t.Dict[str, t.Set[str]] = {}
+
+        for fqn, model in models.items():
+            for tag in model.tags:
+                tag = tag.lower()
+                models_by_tags.setdefault(tag, set())
+                models_by_tags[tag].add(model.fqn)
+
+        print(node)
```
izeigerman, line 181: Ah, refined debugging techniques!

sqlmesh/core/selector.py
```diff
@@ -162,163 +166,183 @@ def expand_model_selections(
...
+                return {fqn} if fqn in models else set()
+            if isinstance(node, exp.And):
+                return evaluate(node.left) & evaluate(node.right)
+            if isinstance(node, exp.Or):
+                return evaluate(node.left) | evaluate(node.right)
+            if isinstance(node, exp.Paren):
+                return evaluate(node.this)
+            if isinstance(node, exp.Not):
+                return set(models) - evaluate(node.this)
+            if isinstance(node, exp.Anonymous):
```
izeigerman, line 202: Why are we re-using the `Anonymous` node? Should we instead have a custom node with explicit enum instead?

sqlmesh/core/selector.py
```diff
@@ -162,163 +166,183 @@ def expand_model_selections(
...
+                        *self._git_client.list_uncommitted_changed_files(),
+                        *self._git_client.list_committed_changed_files(target_branch=target_branch),
+                    }
+                    return {m.fqn for m in self._models.values() if m._path in git_modified_files}
+
+                selected = set()
+
+                for model_name in evaluate(node.expressions[0]):
+                    selected.add(model_name)
+                    if "U" in modifier:
```
izeigerman, line 230: yes, I definitely vote for a custom node with the first-class modifier support

sqlmesh/core/selector.py
tobymao: eh it's a bit more effort to do this for not much gain, main reason would be custom serialization, but this is all internal stuff anyways

izeigerman (approved):

PR #3539, opened 2024-12-19
Feat: do recursive glob matching for ignore_patterns

### LINKED ISSUE #3507: `ignore_patterns` in `config.yaml` does not glob recursively
> [!WARNING]
>
> This might be a design choice. If it is, please close this 😋 

## Summary

The [`ignore_patterns` configuration option in the `config.yaml` file](https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#projects) takes a list of glob patterns.

However, the globbing isn't recursive: the pattern `/**/` does not match any number of directories.

This might be a design choice, but it took me by surprise, as I'm used to `/**/` matching zero or more directories.

## Minimal working example

Consider a minimal SQLMesh project with the following structure:

```
config.yaml
models/
    models.sql
    domain/
        domain.sql
        user/
          user.sql
```

I've checked how SQLMesh ignores files by running `sqlmesh info` and checking the `Models` count in the console.

With `ignore_patterns: []`, SQLMesh finds 3 models as expected.

With `ignore_patterns: ["models/*.sql"]`, SQLMesh finds 2 models as expected.

However, with `ignore_patterns: ["models/**/*.sql"]`, SQLMesh still finds 2 models, whereas I was expecting it to find 0 models.

georgesittas:
Fixes #3507

### DIFF
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+import glob
 import linecache
 import logging
 import os
@@ -262,12 +263,18 @@ class Loader(abc.ABC):
         ignore_patterns = ignore_patterns or []
         extension = extension or ""
 
+        # We try to match both ignore_pattern itself and every file returned by glob,
+        # so that we will always ignore file names that do not appear in the latter.
+        ignored_filepaths = set(ignore_patterns) | {
+            ignored_path
+            for ignore_pattern in ignore_patterns
+            for ignored_path in glob.glob(str(self._context.path / ignore_pattern), recursive=True)
+        }
         for filepath in path.glob(f"**/*{extension}"):
-            for ignore_pattern in ignore_patterns:
-                if filepath.match(ignore_pattern):
-                    break
-            else:
-                yield filepath
+            if any(filepath.match(ignored_filepath) for ignored_filepath in ignored_filepaths):
+                continue
+
+            yield filepath
 
     def _add_model_to_dag(self, model: Model) -> None:
         self._dag.add(model.fqn, model.depends_on)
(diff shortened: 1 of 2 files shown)


erindru (approved):

sqlmesh/core/loader.py
```diff
@@ -264,7 +265,14 @@ def _glob_paths(
 
         for filepath in path.glob(f"**/*{extension}"):
             for ignore_pattern in ignore_patterns:
-                if filepath.match(ignore_pattern):
+                # We try to match both ignore_pattern itself and every file returned by glob,
+                # so that we will always ignore file names that do not appear in the latter.
+                if filepath.match(ignore_pattern) or any(
```
tobymao, line 270: is th ere any performance problem here?

georgesittas: I haven't benchmarked this, but based on their [docs](https://docs.python.org/3/library/glob.html#glob.glob):

> If recursive is true, the pattern “**” will match any files and zero or more directories, subdirectories and symbolic links to directories. If the pattern is followed by an [os.sep](https://docs.python.org/3/library/os.html#os.sep) or [os.altsep](https://docs.python.org/3/library/os.html#os.altsep) then files will not match.

Given that `**` patterns weren't working until now, I believe it's fair to assume that existing `ignore_patterns` lists only contain absolute/relative file paths or patterns with a single `*` in them. My intuition is that `glob`ing those patterns shouldn't incur any significant performance hits compared to the existing  `filename.match(...)` call.

sqlmesh/core/loader.py
```diff
@@ -262,12 +263,18 @@ def _glob_paths(
...
         for filepath in path.glob(f"**/*{extension}"):
-            for ignore_pattern in ignore_patterns:
-                if filepath.match(ignore_pattern):
-                    break
-            else:
-                yield filepath
+            if any(filepath.match(ignored_filepath) for ignored_filepath in ignored_filepaths):
+                continue
+
+            yield filepath
```
georgesittas, line 277: Refactored to pre-compute the set of ignored file paths.

tobymao (approved):

PR #3540, opened 2024-12-19
Feat!: Add support for merge_filter and dbt incremental_predicates for Incremental By Unique Key

themisvaltinos:
This update introduces `merge_filters` in `INCREMENTAL_BY_UNIQUE_KEY` kind models and also adds support for dbt projects with `incremental_predicates`. It updates the incremental logic for engines that support the `MERGE` strategy, by incorporating ability to filter data during the merge operation.

When using this feature, both **source** and **target** aliases can be used to distinguish between the source and target columns similar to `WHEN MATCHED`.

### Example Model

```sql
MODEL (
    name db.test,
        unique_key [purchase_order_id],
        merge_filters (
            source._operation <> 1 AND 
            target.start_date > DATEADD(day, -7, CURRENT_DATE)
        )
    )
);
```

The above predicates are then used into SQLMesh’s `MERGE` statements:

```sql
MERGE INTO <existing_table> AS "__MERGE_TARGET__"
    USING <temp_table> AS "__MERGE_SOURCE__"
    ON
        -- Unique key condition
        -- Custom incremental predicates to limit the data scan
        AND "__MERGE_SOURCE__"._operation <> 1
        AND "__MERGE_TARGET__".start_date > DATEADD(day, -7, CURRENT_DATE)
    WHEN MATCHED THEN 
        UPDATE ...
    WHEN NOT MATCHED THEN 
        INSERT ...
```

This support extends to an existing dbt project to be converted to the equivalent sqlmesh INCREMENTAL_BY_UNIQUE_KEY model, as it aligns with dbt’s **incremental predicates**.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -1264,7 +1264,7 @@ def is_meta_expression(v: t.Any) -> bool:
     return isinstance(v, (Audit, Metric, Model))
 
 
-def replace_table_references(expression: exp.Expression) -> exp.Expression:
+def replace_merge_table_aliases(expression: exp.Expression) -> exp.Expression:
     """
     Resolves references from the "source" and "target" tables (or their DBT equivalents)
     with the corresponding SQLMesh merge aliases (MERGE_SOURCE_ALIAS and MERGE_TARGET_ALIAS)
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -429,10 +429,8 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
             lambda: _parse_macro_or_clause(self, self._parse_when_matched),
             optional=True,
         )
-    elif name == "merge_filters":
-        value = self._parse_wrapped(
-            lambda: _parse_macro_or_clause(self, self._parse_conjunction), optional=True
-        )
+    elif name == "merge_filter":
+        value = self._parse_conjunction()
     elif self._match(TokenType.L_PAREN):
         value = self.expression(exp.Tuple, expressions=self._parse_csv(self._parse_equality))
         self._match_r_paren()
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -429,6 +429,10 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
             lambda: _parse_macro_or_clause(self, self._parse_when_matched),
             optional=True,
         )
+    elif name == "incremental_predicates":
+        value = self._parse_wrapped(
+            lambda: _parse_macro_or_clause(self, self._parse_conjunction), optional=True
+        )
     elif self._match(TokenType.L_PAREN):
         value = self.expression(exp.Tuple, expressions=self._parse_csv(self._parse_equality))
         self._match_r_paren()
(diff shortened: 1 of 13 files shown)


sqlmesh/core/dialect.py
```diff
@@ -429,6 +429,10 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
             lambda: _parse_macro_or_clause(self, self._parse_when_matched),
             optional=True,
         )
+    elif name == "incremental_predicates":
+        value = self._parse_wrapped(
+            lambda: _parse_macro_or_clause(self, self._parse_conjunction), optional=True
+        )
```
georgesittas, line 435: If this is only applicable to `MERGE`, should we rename it appropriately? Current naming feels too vague to me.

themisvaltinos: I was considering this as well, for now I thought since dbt uses the same [incremental_predicates](https://docs.getdbt.com/docs/build/incremental-strategy#about-incremental_predicates) name I think it will be easier for people that migrate off dbt to immediately know where it is used, but can rename it to something else

tests/core/test_model.py
```diff
@@ -5908,6 +5909,116 @@ def test_model_kind_to_expression():
...
 
 
+def test_incremental_predicates():
+    expressions = d.parse(
+        """
+        MODEL (
+          name db.employees,
+          kind INCREMENTAL_BY_UNIQUE_KEY (
+            unique_key name,
+            incremental_predicates source.salary > 0
```
izeigerman, line 5919: Does this support `@start_*` and `@end_*` macros? Can we add a test for this?

izeigerman: We need to make sure they are resolved at evaluation time and not load time

sqlmesh/core/dialect.py
themisvaltinos: As a follow up to our internal discussion revised the name to `merge_filters`

sqlmesh/core/dialect.py
```diff
@@ -429,6 +429,10 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
             lambda: _parse_macro_or_clause(self, self._parse_when_matched),
             optional=True,
         )
+    elif name == "merge_filters":
+        value = self._parse_wrapped(
+            lambda: _parse_macro_or_clause(self, self._parse_conjunction), optional=True
+        )
```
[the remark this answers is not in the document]
themisvaltinos: no we don't I just left them as an optional in case a user had it inside parenthesis but can remove them

sqlmesh/core/dialect.py
```diff
@@ -429,6 +429,10 @@ def _parse_props(self: Parser) -> t.Optional[exp.Expression]:
             lambda: _parse_macro_or_clause(self, self._parse_when_matched),
             optional=True,
         )
+    elif name == "merge_filters":
```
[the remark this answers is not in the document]
themisvaltinos: yes I'm fine with either, the plural wa if there were many conditions in the conjunction, but merge_filter works too

sqlmesh/core/dialect.py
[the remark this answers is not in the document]
themisvaltinos: Oh right then removing this

izeigerman (approved):

PR #3544, opened 2024-12-20
Feat: Add ability to exclude dependencies in the requirements lock file

izeigerman:
This adds ability to opt-out some dependencies and prevent them from being included by the automatic dependency detection logic.

### DIFF
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -74,8 +74,8 @@ class ContextDiff(PydanticModel):
     """Snapshots from the previous finalized state."""
     previous_requirements: t.Dict[str, str] = {}
     """Previous requirements."""
-    provided_requirements: t.Dict[str, str] = {}
-    """Requirements from lock file."""
+    requirements: t.Dict[str, str] = {}
+    """Python dependencies."""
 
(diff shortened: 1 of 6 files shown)


erindru (approved):
LGTM

sqlmesh/core/context_diff.py
```diff
@@ -406,3 +394,33 @@ def text_diff(self, name: str) -> str:
         except SQLMeshError as e:
             logger.warning("Failed to diff model '%s': %s", name, str(e))
             return ""
+
+
+def _build_requirements(
```
tobymao, line 399: why was this moved out?

tobymao (approved):

georgesittas (approved):

sqlmesh/core/context_diff.py
izeigerman: Cause context_diff should only care about the final set of requirements. Because of lazy computation I had to pass both `provided_requirements` and `excluded_requirements` into the instance itself just so that the actual value is computed later. This presented additional complications when it came to creating the `no_diff` instance, since I had to pass the `excluded_requirements` even though requirements were sourced from the target environment.

Issue #3545, opened 2024-12-20
`dlt_load_id` in Generated SQL Doesn't Match dlt Tables


dat-a-man:
I’m working on generating SQLmesh SQL code from a dlt (BlueSky) pipeline using the command:  

```bash
sqlmesh init -t dlt --dlt-pipeline bluesky duckdb
```  
However, I noticed an issue. SQLmesh generates SQL that includes `dlt_load_id` in **all four tables**, but in the dlt pipeline, `dlt_load_id` is only present in **one table**.

**Steps to Reproduce**  
1. After running the dlt pipeline, run the command:  

   ```bash
   sqlmesh init -t dlt --dlt-pipeline bluesky duckdb
   ```  
2. Inspect the generated SQL code for the tables.  

**Expected Behavior**  
`dlt_load_id` should only be present in the table where it exists in the dlt pipeline.  

**Actual Behavior**  
`dlt_load_id` is included in all tables generated by SQLmesh.

**References**  
- Tutorial used: [SQLMesh dlt Pipeline Tutorial](https://davidsj.substack.com/p/sqlmesh-init-t-dlt-dlt-pipeline-bluesky?r=125hnz)
- Related blog post: [dlt-SQLMesh Metadata Handover](https://dlthub.com/blog/sqlmesh-dlt-handover) 
- Documentation: [SQLMesh dlt Integration](https://sqlmesh.readthedocs.io/en/stable/integrations/dlt/?h=dlt)

**Additional Notes**  
This issue seems to relate to improper handling of child tables during SQL generation.  

---

In the database in table, followers__labels
<img width="435" alt="Screenshot 2024-12-20 at 8 49 51 AM" src="https://github.com/user-attachments/assets/7ac4eedb-049b-45b2-9047-c175eec4818d" />

In code there is `dlt_load_id`, originally not in DB
<img width="970" alt="Screenshot 2024-12-20 at 8 50 52 AM" src="https://github.com/user-attachments/assets/ea22ceb3-47ee-45b8-8127-c2e759014d80" />

themisvaltinos:
Thanks for the detailed description in the issue @dat-a-man opened a pr with support for nested dlt tables

PR #3547, opened 2024-12-20
Feat(dlt): Add support for generation of nested dlt tables

themisvaltinos:
This is an update to the dlt integration and adds support for [nested DLT tables](https://dlthub.com/docs/general-usage/destination-tables#nested-tables), fixes: #3545 

- Each row in all (root and nested) data tables created by dlt contains a unique column named _dlt_id (row key), but only parent tables contain the `_dlt_load_id`. 
- Since each dlt nested table contains a column named `_dlt_parent_id` referencing a particular row (`_dlt_id`) of a parent table (parent key), this is used to join and get the parent `_dlt_load_id` to be used for the incremental model.

### DIFF
--- a/sqlmesh/integrations/dlt.py
+++ b/sqlmesh/integrations/dlt.py
@@ -62,12 +62,28 @@ def generate_dlt_models_and_settings(
                 primary_key.append(str(col["name"]))
 
         load_id = next(
-            (col for col in ["_dlt_load_id", "load_id", "_dlt_parent_id"] if col in dlt_columns),
+            (col for col in ["_dlt_load_id", "load_id"] if col in dlt_columns),
             None,
         )
+        load_key = "c." + load_id if load_id else ""
+        parent_table = None
+
+        # Handling for nested tables: https://dlthub.com/docs/general-usage/destination-tables#nested-tables
+        if not load_id:
+            if (
+                "_dlt_parent_id" in dlt_columns
+                and (parent_table := table["parent"])
+                and parent_table in dlt_tables
+            ):
+                load_key = "p._dlt_load_id"
+                parent_table = dataset + "." + parent_table
+            else:
+                break
 
         column_types = [
-            exp.cast("c." + column, data_type, dialect=dialect).as_(column).sql(dialect=dialect)
+            exp.cast(exp.column(column, table="c"), data_type, dialect=dialect)
+            .as_(column)
+            .sql(dialect=dialect)
             for column, data_type in dlt_columns.items()
             if isinstance(column, str)
         ]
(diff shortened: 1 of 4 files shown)


sqlmesh/integrations/dlt.py
```diff
@@ -61,8 +61,13 @@ def generate_dlt_models_and_settings(
...
                 primary_key.append(str(col["name"]))
 
+        load_id = next(
+            (col for col in ["_dlt_load_id", "load_id", "_dlt_parent_id"] if col in dlt_columns),
+            None,
+        )
+
         column_types = [
-            exp.cast(column, data_type, dialect=dialect).as_(column).sql(dialect=dialect)
+            exp.cast("c." + column, data_type, dialect=dialect).as_(column).sql(dialect=dialect)
```
georgesittas, line 70: Let's user `exp.column` to add the `c` table.

sqlmesh/integrations/dlt.py
```diff
@@ -73,15 +78,32 @@ def generate_dlt_models_and_settings(
...
-        )
-
-        sqlmesh_models.add((incremental_model_name, incremental_model_sql))
+        if load_id:
+            load_key = "c." + load_id
+            parent_table = None
+
+            # Handling for nested tables: https://dlthub.com/docs/general-usage/destination-tables#nested-tables
+            if load_id == "_dlt_parent_id" and "_dlt_parent_id" in dlt_columns:
+                if parent_table := table["parent"]:
```
georgesittas, line 87: Can we move the `parent_table := table["parent"]` condition in the outer branch?

sqlmesh/integrations/dlt.py
```diff
@@ -73,15 +78,32 @@ def generate_dlt_models_and_settings(
...
+                    if (
+                        parent_table in dlt_tables
+                        and "_dlt_id" in dlt_tables[parent_table]["columns"]
+                    ):
+                        load_key = "p._dlt_load_id"
```
georgesittas, line 92: Is it correct that this assumes the parent is always gonna be the root table? Can we have more than two levels of nesting?

sqlmesh/integrations/dlt.py
```diff
@@ -73,15 +78,32 @@ def generate_dlt_models_and_settings(
...
+                        parent_table = dataset + "." + parent_table
+                    else:
+                        break
```
georgesittas, line 95: [Nit] consider flipping the branches so that you can dedent L93 and reduce nesting a bit.

sqlmesh/integrations/dlt.py
```diff
@@ -130,7 +162,7 @@ def generate_incremental_model(
 {select_columns},
   {time_column} as _dlt_load_time
 FROM
-  {from_table}
+  {from_table} as c{join}
```
georgesittas, line 165: [Nit] should we remove the leading `\n` from L146 and just put a newline here instead?

sqlmesh/integrations/dlt.py
themisvaltinos: yes we can have more than two levels of nesting but the parent is always going to be the root table. I will extend the example I have in the test to make it clear that we cover arbitrary levels of nesting

georgesittas: Thanks for clarifying 👍

themisvaltinos:
Thanks for the review @georgesittas addressed comments and revised the code if you want to have another look

georgesittas (approved):

PR #3548, opened 2024-12-20
Chore!: deprecate pydantic v1

georgesittas:
Planning to refactor v1 artifacts in a followup PR.


setup.py
```diff
@@ -42,7 +42,7 @@
         "ipywidgets",
         "jinja2",
         "pandas",
-        "pydantic",
+        "pydantic>2.0.0",
```
eakmanrq, line 45: >= right? Unless there is a reason to not want 2.0.0.

georgesittas: Ah, good catch, this was an oversight.

tobymao (approved):

eakmanrq:
FYI there are some comments saying we can deprecate code as a result of Pydantic 1 support being dropped. Here is one example: https://github.com/TobikoData/sqlmesh/blob/756eb705d0b4f33e8d2df9a2681f4678f312b9a6/sqlmesh/core/snapshot/definition.py#L428

I get though if you would prefer to not make a code change like this in the PR.

georgesittas:
> FYI there are some comments saying we can deprecate code as a result of Pydantic 1 support being dropped. Here is one example:
> 
> https://github.com/TobikoData/sqlmesh/blob/756eb705d0b4f33e8d2df9a2681f4678f312b9a6/sqlmesh/core/snapshot/definition.py#L428
> 
> I get though if you would prefer to not make a code change like this in the PR.

Yep, I'm planning to clean all those sections up. There are several refactors that can happen after this deprecation.

izeigerman (approved):
🎉

PR #3561, opened 2024-12-23
Fix: plan physical tables progress bar

treysp:
The `plan` command's physical table creation progress bar sometimes appears even when there are no physical tables to create, with an incorrectly large number of tables listed.

This PR ensures that it only appears if there is a table to create and that the count of tables to create is correct.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -218,30 +218,36 @@ class BuiltInPlanEvaluator(PlanEvaluator):
             plan: The plan to source snapshots from.
             deployability_index: Indicates which snapshots are deployable in the context of this creation.
         """
-        snapshots_to_create = [
+        snapshots_to_maybe_create = [
             s
             for s in snapshots.values()
             if s.is_model and not s.is_symbolic and plan.is_selected_for_backfill(s.name)
         ]
-        snapshots_to_create_count = len(snapshots_to_create)
 
-        if snapshots_to_create_count > 0:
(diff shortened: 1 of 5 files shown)


sqlmesh/core/plan/evaluator.py
```diff
@@ -218,30 +218,36 @@ def _push(
...
         ]
-        snapshots_to_create_count = len(snapshots_to_create)
 
-        if snapshots_to_create_count > 0:
+        snapshots_to_create = self.snapshot_evaluator.get_snapshots_to_create(
+            snapshots_to_maybe_create,
+            deployability_index=deployability_index,
+        )
+
+        if snapshots_to_create["snapshots_to_create"]:
```
erindru, line 232: I'm not familiar with `TypedDict` but do you still have to access its members using strings? If so it seems like an object would be more appropriate

treysp: Ah, good point - I converted it to a pydantic model now. Does that resolve the issue?

erindru (approved):

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -256,23 +264,12 @@ def demote(
                 self.ddl_concurrent_tasks,
             )
 
-    def create(
+    def get_snapshots_to_create(
```
izeigerman, line 267: I don't like this interface. It reveals too much implementation details of the snapshot evaluator and makes the plan evaluator (or any other calling code) more complex . I suggest instead passing another callback into the `create` method which will be called once the number of snapshots needing creation is known.

PR #3568, opened 2024-12-27
Chore!: Disable date prompts by default when running the plan command

izeigerman:
The default settings for the start and the end date are good enough that it should satisfy the vast majority of use cases. This should simplify the UX since it's one less thing for users to think about.

In addition to being `true` by default, the `no_prompts` flag has the following behavioral changes:
* Change categories are now printed
* Expected backfill dates are now printed
* SQLMesh prompts for confirmation to apply the plan unless the `--auto-apply` flag is specified. Previously, `no_prompts` also suppressed the confirmation prompt

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -692,10 +692,9 @@ class TerminalConsole(Console):
             default_catalog=default_catalog,
         )
 
-        if not no_prompts:
-            self._show_options_after_categorization(
-                plan_builder, auto_apply, default_catalog=default_catalog
-            )
+        self._show_options_after_categorization(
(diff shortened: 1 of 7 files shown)


sqlmesh/core/console.py
```diff
@@ -933,6 +944,9 @@ def _prompt_backfill(
             if not plan_builder.override_end:
                 if plan.provided_end:
                     blank_meaning = f"'{time_like_to_str(plan.provided_end)}'"
+                elif plan.interval_end_per_model:
+                    max_end = max(plan.interval_end_per_model.values())
+                    blank_meaning = f"'{time_like_to_str(max_end)}'"
```
sungchun12, line 949: This is a nice quality of life improvement. Very juicy

tobymao (approved):

PR #3577, opened 2025-01-03
Fix!: Models with allow_partials set to true should respect cron

izeigerman:
This update makes it so that models for which `allow_partials` is set to true still respect their cron schedules. 

When running `sqlmesh run` the model won't run unless its time to run has come even if the `allow_partials` flag is set for it. However, if the `--ignore-cron` flag is specified in the `run` command, the model will run every time.

Additionally, this change has the following implications:
- Execution time is no longer rounded down to a minute
- The previous point also results in the `valid_to` column in SCD2 models being populated with millisecond precision


tobymao (approved):

sungchun12:
Amazing, can't wait to unpause some stuff in tcloud :)

PR #3578, opened 2025-01-03
Fix: plan physical tables progress bar

treysp:
The plan command's physical table creation progress bar sometimes appears even when there are no physical tables to create, with an incorrectly large number of tables listed.

This PR ensures that it only appears if there is a table to create and that the count of tables to create is correct.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -223,12 +223,6 @@ class BuiltInPlanEvaluator(PlanEvaluator):
             for s in snapshots.values()
             if s.is_model and not s.is_symbolic and plan.is_selected_for_backfill(s.name)
         ]
-        snapshots_to_create_count = len(snapshots_to_create)
-
-        if snapshots_to_create_count > 0:
-            self.console.start_creation_progress(
-                snapshots_to_create_count, plan.environment, self.default_catalog
-            )
 
         completed = False
         try:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -270,6 +271,7 @@ def create(
             target_snapshots: Target snapshots.
             snapshots: Mapping of snapshot ID to snapshot.
             deployability_index: Determines snapshots that are deployable in the context of this creation.
+            on_start: A callback to initialize the snapshot creation progress bar.
```
izeigerman, line 274: [Nit] The creation progress bar is just a specific use case and this is a generic callback. I'd rather describe what input arguments for the callback are.

sqlmesh/core/plan/evaluator.py
```diff
@@ -237,6 +231,9 @@ def _push(
                 snapshots,
                 allow_destructive_snapshots=plan.allow_destructive_models,
                 deployability_index=deployability_index,
+                on_start=lambda x: self.console.start_creation_progress(
```
izeigerman, line 234: This is already fine, but another idea that occurred to me is instead of passing 2 callbacks, have an abstract interface like:
```
class SnapshotCreationObserver(abc.ABC):

    def on_snapshot_list_ready(self, snapshots: ...) -> None:
    
    def on_snapshot_created(self, snapshot: ...) -> None:
```

izeigerman (approved):

sqlmesh/core/plan/evaluator.py
izeigerman: I think this will make the interface cleaner, and potentially easier to extend in future. So I'd prefer it.

PR #3579, opened 2025-01-03
Fix: Move rendering for merge_filter to evaluation time

themisvaltinos:
This update moves the rendering of `merge_filter` to evaluation time instead of load time, ensuring that any temporal macros (e.g. `@start_ds`) are resolved based on the evaluation time objects.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -1277,9 +1277,9 @@ def replace_merge_table_aliases(expression: exp.Expression) -> exp.Expression:
     from sqlmesh.core.engine_adapter.base import MERGE_SOURCE_ALIAS, MERGE_TARGET_ALIAS
 
     if isinstance(expression, exp.Column):
-        if expression.table.lower() in ("target", "dbt_internal_dest"):
+        if expression.table.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
             expression.set("table", exp.to_identifier(MERGE_TARGET_ALIAS))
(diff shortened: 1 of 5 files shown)


sqlmesh/core/dialect.py
```diff
@@ -1277,9 +1277,9 @@ def replace_merge_table_aliases(expression: exp.Expression) -> exp.Expression:
     from sqlmesh.core.engine_adapter.base import MERGE_SOURCE_ALIAS, MERGE_TARGET_ALIAS
 
     if isinstance(expression, exp.Column):
-        if expression.table.lower() in ("target", "dbt_internal_dest"):
+        if expression.table.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
```
izeigerman, line 1280: Why did you have to change this?

izeigerman (approved):

sqlmesh/core/dialect.py
themisvaltinos: This was to ensure consistency in case insensitive dialects because the `replace_merge_table_aliases` function has already been called in the validator:
https://github.com/TobikoData/sqlmesh/blob/bc14266cfadc9aeea93ef108535f52852b0cfb46/sqlmesh/core/model/kind.py#L482
so the source and target table references are replaced, except for those within macros. During rendering, aliases inside macros are also replaced, but the ones that were already replaced are lowercase in case-insensitive dialects. So I added this to align all aliases and avoid confusion

PR #3582, opened 2025-01-03
Fix(clickhouse): specify DateTime64 precision for model partitioning expression

treysp:
SQLMesh automatically partitions Clickhouse incremental by time range models, aggregating timestamps to week. If the timestamp data type is not a datetime variant, SQLMesh automatically casts it to DateTime64 in the aggregation expression.

In Clickhouse, casting to DateTime64 type requires specifying a numeric precision.

This PR specifies a precision of 9/nanoseconds (using max precision to prevent accidental truncation of timestamp digits).

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2290,7 +2290,7 @@ def clickhouse_partition_func(
     if col_type.is_type(exp.DataType.Type.UNKNOWN):
         return exp.func(
             "toMonday",
-            exp.cast(column, exp.DataType.build("DateTime64('UTC')", dialect="clickhouse")),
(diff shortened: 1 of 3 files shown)


izeigerman (approved):

swt30:
:+1: only caveat I can see is that the [maximum date representable in a DateTime64(9)](https://clickhouse.com/docs/en/sql-reference/data-types/datetime64) is `2262-04-11 23:47:16`. But I won't be around to see the Y2K bug caused by that :laughing: so using the maximum supported precision feels like the way to go. 

This only applies when you pass a non-conformable type like a string or an integer anyway, right?

treysp:
> This only applies when you pass a non-conformable type like a string or an integer anyway, right?

Yep, if time column is non-conformable AND you don't include the time column in the model partitioning expression

Issue #3573, opened 2025-01-02
SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION var is considered as an input for Config


lafirmm:
Got this error when I disabled sqlmesh state migration using this env var SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Config
Extra inputs are not permitted [type=extra_forbidden, input_value={'disable_state_migration': '1'}, input_type=dict]
```

I guess we should either rename the SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION env var or tweak the load_config_from_env function to ignore this variable.

Also I think its good to provide an option to set load_from_env as False when we create sqlmesh context object.

lafirmm:
@tobymao,  is it okay if I create a PR to ignore `SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION` when we load configs from env using load_config_from_env function? or please let me know if you've anything else in your mind. Thanks :)


```
def load_config_from_env() -> t.Dict[str, t.Any]:
    config_dict: t.Dict[str, t.Any] = {}
    for key, value in os.environ.items():
        key = key.lower()
        # TODO DISABLE_STATE_MIGRATION_KEY will be stored in constants
        DISABLE_STATE_MIGRATION_KEY = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
        if key.startswith(f"{c.SQLMESH}__") and key != DISABLE_STATE_MIGRATION_KEY:
            segments = key.split("__")[1:]
            if not segments or not segments[-1]:
                raise ConfigError(f"Invalid SQLMesh configuration variable '{key}'.")

            target_dict = config_dict
            for config_key in segments[:-1]:
                if config_key not in target_dict:
                    target_dict[config_key] = {}
                target_dict = target_dict[config_key]
            target_dict[segments[-1]] = value

    return config_dict
```

tobymao:
@izeigerman

PR #3592, opened 2025-01-07
Fix: change SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION var name

lafirmm:
fixes https://github.com/TobikoData/sqlmesh/issues/3573

since `load_config_from_env` method assumes that the sqlmesh config vars starts with `"SQLMESH__"`, the `SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION` var name is misinterpreted as an input for config and it throws below error,

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Config
Extra inputs are not permitted [type=extra_forbidden, input_value={'disable_state_migration': '1'}, input_type=dict]
```
hence the variable name has been renamed.

I believe there are more than one solution to solve this issue, but I feel this is more sensible and a simple change. In fact, I've created a custom class for the SQLMesh generic context with an arg to not load env vars as config as a workaround for my current project. However, I really appreciate any feedbacks or suggestions and happy to contribute here. Thanks :)


izeigerman:
@lafirmm this would be a breaking change. Can you describe how you run into this issue? I don't think that SQLMesh configuration is expected to be loaded on the Airflow side

lafirmm:
> @lafirmm this is a breaking change. Can you describe how you ran into this issue? I don't think that SQLMesh configuration is expected to be loaded on the Airflow side

I want to handle the SQLMesh state migration by myself and not done by this [SqlmeshAirflowPlugin](https://github.com/TobikoData/sqlmesh/blob/9d7c3856fd77e3a1be86f04e0b994c6186e70f39/sqlmesh/schedulers/airflow/plugin.py#L59) initaited by this [setup](https://github.com/TobikoData/sqlmesh/blob/9d7c3856fd77e3a1be86f04e0b994c6186e70f39/setup.py#L25).
Hence I created an environment variable in airflow containers (managed by our team) called `SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION` with value `1` as described and introduced here https://github.com/TobikoData/sqlmesh/pull/2921, and the environment variable I created `SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION` in airflow containers is misinterpreted as an input for config and it throws above error.

izeigerman:
> And, the SQLMesh config is loaded on the airflow side because we don't use built-in dag generators (SQLMeshAirflow class) but a custom method.

I think this is a problem then.

> Also can you please explain how this is a breaking change since this env var name SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION is used nowhere except [here](https://github.com/TobikoData/sqlmesh/blob/9d7c3856fd77e3a1be86f04e0b994c6186e70f39/sqlmesh/schedulers/airflow/plugin.py#L34).

Because existing users are relying on it

izeigerman:
To ensure that we don't break existing users I'd suggest excluding this variable from the config loader, even though it's a hackier approach

lafirmm:
> To ensure that we don't break existing users I'd suggest excluding this variable from the config loader, even though it's a hackier approach

@izeigerman I agree and made the corresponding changes, I'd really appreciate your feedback

izeigerman (approved):
Thanks for addressing my suggestion!

Issue #3588, opened 2025-01-06
SQL format button in UI star exclude issue


simonprydden:
In the SQLMesh UI, if you click the format button it replaces EXCLUDE with EXCEPT , which is not valid sql.

```
WITH base AS (
  SELECT
    1 AS a,
    2 AS b
)
SELECT
  * EXCLUDE (b)
FROM base
```

```
 .venv ❯ sqlmesh --version
0.142.0
```
```
>>> import duckdb
>>> duckdb.__version__
'1.1.3'
```

izeigerman:
Fixed in `0.145.0`

PR #3596, opened 2025-01-07
Fix: Properly include dialect when formatting

mykalmax:
fixing https://github.com/TobikoData/sqlmesh/issues/3588


tobymao (approved):

RN-nguyen:
Verified format button in sqlmesh ui formats dialect correctly.

PR #3598, opened 2025-01-08
Docs: Airflow facade edits

treysp:


docs/cloud/features/scheduler/airflow.md
```diff
@@ -1,141 +1,227 @@
...
-You also need to add an Airflow [connection](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#creating-a-connection-with-the-ui) containing your Tobiko Cloud credentials:
+### Connect Airflow to Tobiko Cloud
 
-```yaml
-id: tobiko_cloud
-type: HTTP
-host: https://tobiko/cloud/base/url
-password: <tobiko cloud API token>
-```
+Next, add an Airflow [connection](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#creating-a-connection-with-the-ui) containing your Tobiko Cloud credentials. The connection can be specified in YAML or in the Airflow UI:
```
sungchun12, line 25: You don't specify in YAML (this looks like template code to ease the reader), only through the UI, I recommend splitting this out like in the original as the entrypoint is the UI. There is not either/or

sungchun12 (approved):
make that quick split out change with the connections. Overall, good guide improvements.

docs/cloud/features/scheduler/airflow.md
treysp: To clarify, is this correct?

- The only way to add a connection is via UI
- The "code" is just an example list of values that you might put into the UI

erindru: Partially correct :)

> The "code" is just an example list of values that you might put into the UI
Correct

> The only way to add a connection is via UI
Not quite. The easiest way to get started is to add a connection with the UI. However, you can also add connections in environment variables and some companies use a custom secrets backend like AWS Secrets Manager

PR #3604, opened 2025-01-09
Fix: use dialect when freezing current date/time values in tests

### LINKED ISSUE #3601: Support freezing current_timestamp with execution_time when testing in BigQuery
# What's the problem
freezing `CURRENT_TIMESTAMP` when testing in bq does not work as expected.
It works as expected on duckdb.

# What happens
On bq, the query tries to do `CAST('2024-01-01 00:00:00+00:00' AS DATETIME)` which is illegal. it should be `CAST('2024-01-01' AS ..)` or `CAST'(2024-01-01 00:00:00 AS ..)'`

# How to reproduce

model.sql
```sql
MODEL (
  name sqlmesh_poc.example_datetime,
  kind FULL,
);

SELECT
  user_id,
  TIMESTAMP_DIFF(CURRENT_TIMESTAMP(), signup_timestamp, DAY) = 0 AS is_d0
FROM
  sqlmesh_poc.source_table
```

test.yaml
```
test_example_datetime:
  model: example_datetime
  inputs:
    sqlmesh_poc.source_table:
      columns:
        user_id: string
        signup_timestamp: DATETIME
      rows:
        - user_id: user_1
          signup_timestamp: 2024-01-01
        - user_id: user_2
          signup_timestamp: 2023-01-01
  outputs:
    query:
    - user_id: user_1
      is_d0: True
    - user_id: user_2
      is_d0: False
  vars:
    execution_time: 2024-01-01
```

georgesittas:
Fixes #3601

@serkef I think either the test or the model in your example is incorrect, because setting the `DATETIME` cast issue aside, the `TIMESTAMPDIFF` crashes when you're trying to compare a `TIMESTAMP` (current timestamp) and a `DATETIME`.

This PR addresses the issue related to the `CAST`.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -114,10 +114,18 @@ class ModelTest(unittest.TestCase):
             exec_time = exp.Literal.string(self._execution_time)
             self._transforms = {
                 **self._transforms,
-                exp.CurrentDate: lambda self, _: self.sql(exp.cast(exec_time, "date")),
-                exp.CurrentDatetime: lambda self, _: self.sql(exp.cast(exec_time, "datetime")),
-                exp.CurrentTime: lambda self, _: self.sql(exp.cast(exec_time, "time")),
-                exp.CurrentTimestamp: lambda self, _: self.sql(exp.cast(exec_time, "timestamp")),
+                exp.CurrentDate: lambda self, _: self.sql(
+                    exp.cast(exec_time, "date", dialect=dialect)
+                ),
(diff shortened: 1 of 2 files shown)


serkef:
you are right, `signup_timestamp` on my example https://github.com/TobikoData/sqlmesh/issues/3601 should be `TIMESTAMP` and not `DATETIME`.
I tried your solution from this PR and it solves the issue with cast! Thanks :)

tobymao (approved):

Issue #3316, opened 2024-10-31
Support MERGE in Redshift


dmat:
Redshift supports MERGE INTO, but Sqlmesh is not currently supporting MERGE in Redshift, instead using temp tables for merge.

PR #3607, opened 2025-01-09
Feat(redshift): Add native Merge support for Redshift

themisvaltinos:
This update adds native `MERGE` support for Incremental by Unique Key models in Redshift (fixes: #3316)

This required adjustments to the generated statement to accommodate for the lack of alias support for the target table and limitations of the when matched clause [Redshift MERGE Documentation](https://docs.aws.amazon.com/redshift/latest/dg/r_MERGE.html)

### DIFF
--- a/sqlmesh/core/engine_adapter/redshift.py
+++ b/sqlmesh/core/engine_adapter/redshift.py
@@ -346,8 +346,12 @@ class RedshiftEngineAdapter(
                 expression.set("table", exp.to_table(target_table))
             return expression
 
-        # Redshift does not support multiple "WHEN MATCHED" clauses.
-        if len(whens.expressions) != 2:
+        # Ensure that there is exactly one "WHEN MATCHED" and one "WHEN NOT MATCHED" clause.
+        # Since Redshift does not support multiple "WHEN MATCHED" clauses.
+        if (
+            len(whens.expressions) != 2
+            or whens.expressions[0].args["matched"] == whens.expressions[1].args["matched"]
+        ):
             raise SQLMeshError(
                 "Redshift only supports a single WHEN MATCHED and WHEN NOT MATCHED clause"
             )
--- a/sqlmesh/core/engine_adapter/redshift.py
+++ b/sqlmesh/core/engine_adapter/redshift.py
@@ -7,10 +7,10 @@ import pandas as pd
 from sqlglot import exp
 
 from sqlmesh.core.dialect import to_schema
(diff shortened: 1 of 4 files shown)


erindru:
This theoretically looks good but are you able to show the Redshift integration tests still passing by commenting out the branch filter for the cloud tests in [.circleci/continue_config.yml](https://github.com/TobikoData/sqlmesh/blob/main/.circleci/continue_config.yml#L316)? (and re-instating it prior to merge)

LogicalMergeMixin had less syntax limitations so I wonder if "upgrading" to Redshift's limited version of MERGE will break things

themisvaltinos:
> This theoretically looks good but are you able to show the Redshift integration tests still passing by commenting out the branch filter for the cloud tests in [.circleci/continue_config.yml](https://github.com/TobikoData/sqlmesh/blob/main/.circleci/continue_config.yml#L316)? (and re-instating it prior to merge)

Thanks for this suggestion I tested it and it appears to be working as expected

> LogicalMergeMixin had less syntax limitations so I wonder if "upgrading" to Redshift's limited version of MERGE will break things

I was wondering of this as well, but given that logical merge does not support `when_matched` or `merge_filter` I was thinking it should be fine in that aspect. But can you think of any other potential risks we should consider?

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -329,6 +329,41 @@ def _get_data_objects(
...
+        def resolve_target_table(expression: exp.Expression) -> exp.Expression:
+            if (
+                isinstance(expression, exp.Column)
+                and expression.table.upper() == MERGE_TARGET_ALIAS
+            ):
+                expression.set("table", exp.to_table(target_table))
+            return expression
+
+        # Redshift does not support multiple "WHEN MATCHED" clauses.
+        if len(whens.expressions) != 2:
```
erindru, line 350: The `!=2` check makes some assumptions about how this code is called.

Would counting the `exp.When` objects where `matched=True` and `matched=False` and erroring if there is more than 1 of each be more robust?

erindru:
> I tested it and it appears to be working as expected
Thanks, looks good!

> But can you think of any other potential risks we should consider?
Every change carries risk :) but in this case, our tests pass so I think we are good. If this change breaks someones workflow, its an opportunity to improve our test suite

sqlmesh/core/engine_adapter/redshift.py
themisvaltinos: Yes good point also more future proof in case we change how we construct `when_matched`. Revised it to check that there are only one `exp.When` with `matched=True` and one with `matched=False`

erindru (approved):

PR #3609, opened 2025-01-09
Fix: The interval unit can be larger than cron perid of allow_partials is set to true

izeigerman:


sqlmesh/core/node.py
```diff
@@ -260,7 +260,7 @@ def _interval_unit_validator(cls, v: t.Any) -> t.Optional[t.Union[IntervalUnit,
     @model_validator_v1_args
     def _node_root_validator(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
         interval_unit = values.get("interval_unit_")
-        if interval_unit:
+        if interval_unit and not values.get("allow_partials", False):
```
tobymao, line 263:
```suggestion
        if interval_unit and not values.get("allow_partials"):
```

tobymao (approved):

PR #3611, opened 2025-01-09
Fix: this_model macro usage in the on_virtual_update statements

izeigerman:
Without this fix `this_model` is resolved to the physical table name when rendering `on_virtual_update` statements.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -433,6 +433,8 @@ class _Model(ModelMeta, frozen=True):
         engine_adapter: t.Optional[EngineAdapter] = None,
         **kwargs: t.Any,
     ) -> t.List[exp.Expression]:
+        if "this_model" not in kwargs:
+            kwargs["this_model"] = self.fully_qualified_table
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -433,6 +433,8 @@ def render_on_virtual_update(
         engine_adapter: t.Optional[EngineAdapter] = None,
         **kwargs: t.Any,
     ) -> t.List[exp.Expression]:
+        if "this_model" not in kwargs:
+            kwargs["this_model"] = exp.maybe_parse(self.fqn, into=exp.Table, dialect=self.dialect)
```
tobymao, line 437: self.fqn is always normalized already, i don't think you want dialect here?

tobymao: exp.to_table(self.fqn)

tobymao: it should be

self.fully_qualified_table

tobymao (approved):

PR #3613, opened 2025-01-10
Docs: clarify interval_unit definition and use in lookback/batch size

treysp:


tobymao (approved):

sungchun12 (changes requested):

docs/concepts/models/overview.md
```diff
@@ -227,10 +227,43 @@ Learn more about these properties and their default values in the [model configu
...
+      name sqlmesh_example.demo,
+      kind INCREMENTAL_BY_TIME_RANGE (
+        time_column date_column,
+        lookback 2, -- 2 days of late-arriving data to backfill
+      ),
+      start '2024-11-01',
+      cron '@hourly', --tells sqlmesh to run hourly, not tied to the interval_unit
+      allow_partials true, -- allow partial hourly intervals to be written to the table while still respecting the interval_unit
+      interval_unit 'day', -- lowest granularity of data to be time bucketed
+    );
```
sungchun12, line 266: looks like code suggestions doesn't like duplicate ```, you can add that in so it renders correctly

docs/concepts/models/overview.md
```diff
@@ -227,10 +227,84 @@ Learn more about these properties and their default values in the [model configu
...
+        kind INCREMENTAL_BY_TIME_RANGE (
+          time_column date_column,
+        ),
+        start '2024-11-01',
+        cron '30 7 * * *', -- cron expression for "every day at 7:30am"
+        interval_unit 'hour', -- backfill up until the most recently completed hour (rather than day)
+      );
+    ```
+    <br></br>
+    **NOTE: the example below is an advanced use case. Use the `allow_partials` configuration option with caution.**
```
tobymao, line 281: allow partials is bad, can elaborate more 

https://tobikodata.com/data-completeness.html

1. you don't know if missing data is due to a partial run, or if there was a problem
2. if you ever miss running things, you could end up with incomplete data forever

in generally, we recommend not doing allow_partials

tobymao (approved):

docs/concepts/models/overview.md
```diff
@@ -227,10 +227,84 @@ Learn more about these properties and their default values in the [model configu
 :   Tags are one or more labels used to organize your models.
 
 ### cron
-:   Cron is used to schedule your model to process or refresh at a certain interval. It accepts a [cron expression](https://en.wikipedia.org/wiki/Cron) or any of `@hourly`, `@daily`, `@weekly`, or `@monthly`.
+:   Cron is used to schedule your model to process or refresh at a certain cadence. It accepts a [cron expression](https://en.wikipedia.org/wiki/Cron) or any of `@hourly`, `@daily`, `@weekly`, or `@monthly`.
```
sungchun12, line 230:
```suggestion
:   Cron is used to schedule your model to process or refresh at a certain cadence. It accepts a [cron expression](https://en.wikipedia.org/wiki/Cron) or any of `@hourly`, `@daily`, `@weekly`, or `@monthly`. Must be in `UTC` time zone.
```

docs/concepts/models/overview.md
sungchun12, line 281:
```suggestion
    **NOTE: the example below is an advanced use case. Use the `allow_partials` configuration option with caution. One potential issue is you will NOT know if missing data is due to a partial run or if there was a problem processing today's data such as a true data gap or anomaly. Overall, you risk sharing incomplete/incorrect data even when SQLMesh runs successfully. Read more: [here](https://tobikodata.com/data-completeness.html)**
```

sungchun12 (approved):
Made a couple suggestions, and this is a welcome improvement to these config definitions. Great use of examples.

Issue #3322, opened 2024-10-31
Better compile time checks if the column exists or not


sungchun12:
Make SQLMesh smarter when running compile time checks by validating a column exists or not BEFORE running queries in the query engine. 

Scenario: I have a SQL model below where I intentionally add a fake column that does not exist. The upstream model `tcloud_demo.seed_raw_payments` has columns defined explicitly. When I run a new plan, it will fail as expected at the query engine level. I want it to fail BEFORE that happens. This makes developing in SQLMesh much faster as users don't have to wait for the query engine to fix obvious errors, and it incentivizes them to explicitly define their columns. 

Note: this will not solve for models defined as `select *` which is a fair expectation

```shell
(.venv) ➜  tobiko-cloud-demo git:(demo-sung) ✗ sqlmesh plan dev
======================================================================
Successfully Ran 1 tests against duckdb
----------------------------------------------------------------------
Summary of differences against `dev`:
Models:
├── Directly Modified:
│   └── tcloud_demo__dev.stg_payments
└── Indirectly Modified:
    ├── tcloud_demo__dev.orders
    └── tcloud_demo__dev.customers
---                                                                                        
                                                                                           
+++                                                                                        
                                                                                           
@@ -16,5 +16,6 @@                                                                          
                                                                                           
   order_id,                                                                               
   payment_method,                                                                         
   amount / 100 AS amount,                                                                 
-  'example_column' AS example_column                                                      
+  'example_column' AS example_column,                                                     
+  fake_column                                                                             
 FROM tcloud_demo.seed_raw_payments                                                        
Directly Modified: tcloud_demo__dev.stg_payments (Breaking)
└── Indirectly Modified Children:
    ├── tcloud_demo__dev.customers (Indirect Breaking)
    └── tcloud_demo__dev.orders (Indirect Breaking)
Models needing backfill (missing dates):
├── tcloud_demo__dev.customers: 2024-10-30 - 2024-10-30
├── tcloud_demo__dev.orders: 2024-10-30 - 2024-10-30
└── tcloud_demo__dev.stg_payments: 2024-10-30 - 2024-10-30
Enter the backfill start date (eg. '1 year', '2020-01-01') or blank to backfill from the 
beginning of history: 
Enter the backfill end date (eg. '1 month ago', '2020-01-01') or blank to backfill up until
'2024-10-31 00:00:00': 
Apply - Backfill Tables [y/n]: y
[...]

Issue #3139, opened 2024-09-16
`select * from` prevention and auto-resolution based on upstream raw schema and/or model definition


sungchun12:
Everyone understands that it's best practice to avoid `select * from` by default, but it's the easiest way for users to keep moving fast. In plain terms, people will bias towards this whether we like it or not. This will also mess with impact analysis as non-breaking changes won't be taken advantage of in sql models. The goal of this is to make the developer experience enjoyable for the boring and tedious bits.

It'd be nice to have a config in `config.yaml` that inserts a behavior before running `sqlmesh plan <dev env>` that prompts the user to write out the columns by hand OR if possible, ask for permission to overwrite `*` with all the columns and then simply overwrites it using the `@STAR` macro behind the scenes: https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/?h=@star#star

```yaml
gateways:
    tobiko_cloud: # this will use the config in tcloud.yaml for state_connection
        connection:
            type: bigquery
            method: service-account-json
            concurrent_tasks: 5
            register_comments: true
            keyfile_json: {{ env_var('GOOGLE_SQLMESH_CREDENTIALS') }}
            project: sqlmesh-public-demo

default_gateway: tobiko_cloud

prevent_select_star: true # a config like this
```

tobymao:
this is kind of a duplicate of #3322 . if we implement a strict mode, this would be part of it

PR #3614, opened 2025-01-10
Feat!: Add 'validate_query' model attribute for compile-time checks

VaggelisD:
Inspired by https://github.com/TobikoData/sqlmesh/issues/3322 and https://github.com/TobikoData/sqlmesh/issues/3139

This PR introduces `validate_query` model flag which will raise an error (instead of emitting warnings) if query optimization fails. 

The approach taken largely follows [the optimize_query PR](https://github.com/TobikoData/sqlmesh/pull/3512), with the addition that the optimized query cache is invalidated if `validate_query` is altered for better UX. 

cc: @sungchun12


sqlmesh/core/model/cache.py
```diff
@@ -135,6 +135,7 @@ def _entry_name(model: SqlModel) -> str:
         hash_data.append(str([gen(d) for d in model.macro_definitions]))
         hash_data.append(str([(k, v) for k, v in model.sorted_python_env]))
         hash_data.extend(model.jinja_macros.data_hash_values)
+        hash_data.extend(str(model.validate_query))
```
izeigerman, line 138: nice 👍

sqlmesh/core/renderer.py
```diff
@@ -526,12 +530,17 @@ def _optimize_query(self, query: exp.Query, all_deps: t.Set[str]) -> exp.Query:
...
-            logger.warning(
+            warning = (
                 f"SELECT * cannot be expanded due to missing schema(s) for model(s): {deps}. "
                 "Run `sqlmesh create_external_models` and / or make sure that the model "
-                f"'{self._model_fqn}' can be rendered at parse time.",
+                f"'{self._model_fqn}' can be rendered at parse time."
             )
 
+            if strict:
+                raise_config_error(warning, self._path)
```
izeigerman, line 540: Isn't it a little too harsh? If I enable the validation globally, I'd expect SQLMesh to do its best to detect invalid columns given the information available to it. I wouldn't expect it to fail because it's missing the schema for some models.

tobymao:
this is a breaking change and requires a migration

sqlmesh/migrations/v0067_add_validate_query.py
```diff
@@ -0,0 +1,5 @@
+"""Add the validate_query model attribute."""
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
```
tobymao, line 4: this isn't necessary if we bump sqlglot

sqlmesh/core/renderer.py
tobymao: maybe in that case they should do validate = False

VaggelisD: My thinking & understanding of the `validate_query` flag is that it'll generally be turned off by default and should be used when the user wants to enforce a "strict mode" that abides by SQLGlot's rules (the terms & conditions apply here). 

I'm not sure how we could make it less harsh, there's also `logger.error` but it doesn't stop execution afaict. I can see how this can get annoying though, open to suggestions

izeigerman: What i've seen in the past, is that for some models the schema is available while for others - it's not. I can see someone enabling this globally just to do the validation for those models for which schema is available.

sqlmesh/migrations/v0067_add_validate_query.py
izeigerman: Then you would want to upgrade sqlglot in this PR. Otherwise, the PR must be green to be merged

docs/concepts/models/overview.md
```diff
@@ -361,6 +361,9 @@ to `false` causes SQLMesh to disable query canonicalization & simplification. Th
 !!! warning
     Turning off the optimizer may prevent column-level lineage from working for the affected model and its descendants, unless all columns in the model's query are qualified and it contains no star projections (e.g. `SELECT *`).
 
+### validate_query
+:   Whether the model's query will be validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings.
```
sungchun12, line 365:
```suggestion
:   Whether the model's query will be validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings. This will display invalid columns in your SQL statements along with models containing `SELECT *` that cannot be automatically expanded to list out all columns. This ensures SQL is verified locally before time and money are spent running the SQL in your data warehouse.
```

docs/reference/model_configuration.md
```diff
@@ -39,7 +39,7 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
...
 | `optimize_query`             | Whether the model's query should be optimized. This attribute is `true` by default. Setting it to `false` causes SQLMesh to disable query canonicalization & simplification. This should be turned off only if the optimized query leads to errors such as surpassing text limit.                                                                                                                                                                                                      |       bool        |    N     |
-
+| `validate_query`             | Whether the model's query will be strictly validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings.                                                                                                                                                                                                |       bool        |    N     |
```
sungchun12, line 42:
```suggestion
| `validate_query`             | Whether the model's query will be strictly validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings. This will display invalid columns in your SQL statements along with models containing `SELECT *` that cannot be automatically expanded to list out all columns.                                                                                                                                                                                              |       bool        |    N     |
```

sungchun12:
Can you make this serve all invalid column errors vs. one at a time?

It'll be better UX for someone to see all errors at once, fix them, and then run this again. If someone had 10 errors in a row, this will feel annoying.

image

tests/core/test_model.py
```diff
@@ -6918,3 +6918,106 @@ def model_with_virtual_statements(context, **kwargs):
...
+
+    snapshots = list(plan.snapshots.values())
+    assert len(snapshots) == 1
+
+    snapshot = snapshots[0]
+    assert len(snapshot.previous_versions) == 1
+    assert snapshot.change_category == SnapshotChangeCategory.METADATA
+
+    # Ensure non-SQLModels raise if strict mode is set
+    with pytest.raises(
```
sungchun12, line 7008: Can you add a test that it doesn't raise this error if the model config is updated?

The below still raises the error locally.

```sql
MODEL (
  kind SEED (
    path '../seeds/seed_data.csv'
  ),
  columns (
    id INTEGER,
    item_id INTEGER,
    event_date DATE
  ),
  grain (id, event_date),
  validate_query False
);

```

sqlmesh/core/renderer.py
```diff
@@ -526,12 +530,17 @@ def _optimize_query(self, query: exp.Query, all_deps: t.Set[str]) -> exp.Query:
         if self._model_fqn and not should_optimize and any(s.is_star for s in query.selects):
             deps = ", ".join(f"'{dep}'" for dep in sorted(missing_deps))
 
-            logger.warning(
+            warning = (
```
sungchun12, line 533: All I see is a warning if you can't expand `select *`. For models where it can be expanded, can you overwrite the contents of the SQL model and replace `select *` with the rendered query? There could even be a prompt: "Do you want to overwrite `model xyz` to list out all columns in your select statement? [y/n]"

sqlmesh/core/renderer.py
tobymao: the issue with that is you won't know whether or not your model is being validated or not. because if any parent is missing stuff, it'll happily chug along. if you turn on validation for your project, i think it should probably be opt out on the model

sqlmesh/migrations/v0067_add_validate_query.py
VaggelisD: Will keep the migration script if that's okay, is their use generally discouraged?

VaggelisD:
> Can you make this serve all invalid column errors vs. one at a time?
> 
> It'll be better UX for someone to see all errors at once, fix them, and then run this again. If someone had 10 errors in a row, this will feel annoying.
> 
> image


I think this is required sadly i.e if we have 2 invalid columns back to back, the optimizer will keep raising on the first one it comes across (e.g `"fake_column"` on your example).

If I'm not mistaken, this is what happens with `main` today too, e.g rendering this standalone model:

```
MODEL (
  name test,
);

SELECT foo, bar
```

Will first yield a single warning for `"foo"` and only if that's fix will we get a warning for `"bar"`.

tests/core/test_model.py
VaggelisD: My intention was that `validate_query` should be defined on SQLModels only, otherwise it won't have an effect as there's no query to be validated which is the case for `SEED` models too (?)

sqlmesh/core/renderer.py
VaggelisD: Hmm, do you have a concrete example? Afaict this already happens through the optimizer, right? E.g:

```Python3
>>> query = d.parse(
    """
    MODEL (
        name test,
    );

    SELECT * FROM (SELECT 1 AS col)
    """
)

>>> model = load_sql_based_model(query)
>>> print(model.render_query())

SELECT "_q_0"."col" AS "col" FROM (SELECT 1 AS "col") AS "_q_0"
```

If optimization is instead turned off:

```Python3
>>> query = d.parse(
    """
    MODEL (
        name test,
        optimize_query False
    );

    SELECT * FROM (SELECT 1 AS col)
    """
)


>>> model = load_sql_based_model(query)
>>> print(model.render_query())

SELECT * FROM (SELECT 1 AS "col") AS "_q_0"
```

Also, since that idea enables a certain optimization, would it be best suited for a follow-up PR?

sqlmesh/migrations/v0067_add_validate_query.py
georgesittas: > is their use generally discouraged?

No, it's not discouraged. It's simply a matter of timing, i.e. if we wanna bump sqlglot to a new minor or major anyway, then you can omit an empty migration script, which would be preferable.

tests/core/test_model.py
```diff
@@ -6918,3 +6918,106 @@ def model_with_virtual_statements(context, **kwargs):
...
+        validate_query True,
+    );
+
+    SELECT foo
+    """
+    )
+
+    with pytest.raises(
+        ConfigError,
+        match=r"""Column '"foo"' could not be resolved for model.*""",
```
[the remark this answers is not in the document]
VaggelisD: Yep, the warning/error messages should have remained the same, I wildcard-matched the rest of the text as it's long:

```
Column '"foo"' could not be resolved for model '"test"', the column may not exist or is ambiguous
```

tests/core/test_model.py
sungchun12: I understand the intention, but it's still erroring out locally for me. I expect users to update this globally in `config.yaml` by default.

image

sqlmesh/core/renderer.py
sungchun12: I understand the rendered query can be printed out, but the main problem is to make sure no `select *` is versioned . All columns are explicit as a strong opinion we want to enforce if a user opts into this config.

That's fine as a follow-up PR

sungchun12:
> > Can you make this serve all invalid column errors vs. one at a time?
> > It'll be better UX for someone to see all errors at once, fix them, and then run this again. If someone had 10 errors in a row, this will feel annoying.
> > image
> 
> I think this is required sadly i.e if we have 2 invalid columns back to back, the optimizer will keep raising on the first one it comes across (e.g `"fake_column"` on your example).
> 
> If I'm not mistaken, this is what happens with `main` today too, e.g rendering this standalone model:
> 
> ```
> MODEL (
>   name test,
> );
> 
> SELECT foo, bar
> ```
> 
> Will first yield a single warning for `"foo"` and only if that's fixed will we get a warning for `"bar"`.

For this PR, we can ignore displaying all errors at once. For a future PR, is it feasible to concurrently render and then print out a full batch of errors? Is that an architectural limitation that requires a larger conversation?

tests/core/test_model.py
VaggelisD: Good catch, I'd raise if `validate_query` was non-`None` instead of non-`True`

VaggelisD:
> For this PR, we can ignore displaying all errors at once. For a future PR, is it feasible to concurrently render and then print out a full batch of errors? Is that an architectural limitation that requires a larger conversation?

I think it's more of an architectural limitation, even if we silence all but the last exception, I'm not sure what are the side effects of forcing the optimizer to look past the first error. I can play with it and see if it's hallunicating or not

sungchun12 (approved):

PR #3615, opened 2025-01-10
Chore!: remove pydantic v1 validator arg helpers

### LINKED ISSUE #3566: Refactor tech debt related to Pydantic v1
This [commit](https://github.com/TobikoData/sqlmesh/commit/f0d2e0f65e20ea49c2b14b0b852a212c2168c09b) deprecated support for Pydantic v1, so it's now possible to simplify code that was needed to bridge v1 and v2.

georgesittas:
Fixes #3566

This PR mainly aims to remove `model_validator_v1_args` and `field_validator_v1_args`. It's a WIP, I'm not sure yet if this is the right way to do it, just wanted some first feedback.

### DIFF
--- a/sqlmesh/core/model/common.py
+++ b/sqlmesh/core/model/common.py
@@ -14,7 +14,7 @@ from sqlmesh.core.macros import MacroRegistry, MacroStrTemplate
 from sqlmesh.utils import str_to_bool
 from sqlmesh.utils.errors import ConfigError, SQLMeshError, raise_config_error
 from sqlmesh.utils.metaprogramming import Executable, build_env, prepare_env, serialize_env
-from sqlmesh.utils.pydantic import field_validator, field_validator_v1_args
+from sqlmesh.utils.pydantic import ValidationInfo, field_validator
 
 if t.TYPE_CHECKING:
     from sqlmesh.utils.jinja import MacroReference
@@ -176,11 +176,10 @@ def single_value_or_tuple(values: t.Sequence) -> exp.Identifier | exp.Tuple:
     )
 
 
-@field_validator_v1_args
 def parse_expression(
     cls: t.Type,
     v: t.Union[t.List[str], t.List[exp.Expression], str, exp.Expression, t.Callable, None],
-    values: t.Dict[str, t.Any],
+    info: t.Optional[ValidationInfo] = None,
 ) -> t.List[exp.Expression] | exp.Expression | t.Callable | None:
     """Helper method to deserialize SQLGlot expressions in Pydantic Models."""
     if v is None:
@@ -189,7 +188,7 @@ def parse_expression(
     if callable(v):
         return v
 
-    dialect = values.get("dialect")
+    dialect = info.data.get("dialect") if info else ""
 
     if isinstance(v, list):
         return [
@@ -213,12 +212,14 @@ def parse_bool(v: t.Any) -> bool:
     return str_to_bool(str(v or ""))
 
 
-@field_validator_v1_args
-def parse_properties(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[exp.Tuple]:
+def parse_properties(
+    cls: t.Type, v: t.Any, info: t.Optional[ValidationInfo] = None
+) -> t.Optional[exp.Tuple]:
     if v is None:
         return v
 
-    dialect = values.get("dialect")
+    dialect = info.data.get("dialect") if info else ""
+
     if isinstance(v, str):
         v = d.parse_one(v, dialect=dialect)
     if isinstance(v, (exp.Array, exp.Paren, exp.Tuple)):
@@ -254,10 +255,9 @@ def default_catalog(cls: t.Type, v: t.Any) -> t.Optional[str]:
     return str(v)
 
 
-@field_validator_v1_args
-def depends_on(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[t.Set[str]]:
-    dialect = values.get("dialect")
-    default_catalog = values.get("default_catalog")
+def depends_on(cls: t.Type, v: t.Any, info: ValidationInfo) -> t.Optional[t.Set[str]]:
+    dialect = info.data.get("dialect")
+    default_catalog = info.data.get("default_catalog")
 
     if isinstance(v, exp.Paren):
         v = v.unnest()
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2,12 +2,12 @@ from __future__ import annotations
 
 import json
 import logging
-import sys
 import types
 import re
 import typing as t
 from functools import cached_property
 from pathlib import Path
+from typing_extensions import Self
 
 import pandas as pd
 import numpy as np
@@ -59,10 +59,6 @@ if t.TYPE_CHECKING:
     from sqlmesh.core.snapshot import DeployabilityIndex, Node, Snapshot
     from sqlmesh.utils.jinja import MacroReference
 
-    if sys.version_info >= (3, 11):
-        from typing import Self
-    else:
-        from typing_extensions import Self
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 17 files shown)


sqlmesh/core/model/common.py
```diff
@@ -176,11 +176,10 @@ def single_value_or_tuple(values: t.Sequence) -> exp.Identifier | exp.Tuple:
     )
 
 
-@field_validator_v1_args
 def parse_expression(
     cls: t.Type,
     v: t.Union[t.List[str], t.List[exp.Expression], str, exp.Expression, t.Callable, None],
-    values: t.Dict[str, t.Any],
+    info: t.Optional[ValidationInfo] = None,
```
izeigerman, line 182: Why is this optional?

sqlmesh/core/model/common.py
```diff
@@ -213,12 +212,14 @@ def parse_bool(v: t.Any) -> bool:
     return str_to_bool(str(v or ""))
 
 
-@field_validator_v1_args
-def parse_properties(cls: t.Type, v: t.Any, values: t.Dict[str, t.Any]) -> t.Optional[exp.Tuple]:
+def parse_properties(
+    cls: t.Type, v: t.Any, info: t.Optional[ValidationInfo] = None
```
izeigerman, line 216: ditto

sqlmesh/core/model/definition.py
```diff
@@ -59,10 +59,6 @@
     from sqlmesh.core.snapshot import DeployabilityIndex, Node, Snapshot
     from sqlmesh.utils.jinja import MacroReference
 
-    if sys.version_info >= (3, 11):
```
izeigerman, line 62: Why was this removed?

sqlmesh/core/model/common.py
georgesittas: Made it optional to allow the tests to use `parse_expression`. Did a quick check to see if I can mock `info` but wasn't sure how to do it, can revisit if needed.

The `parse_properties` validator is also invoked in `model/kind.py` without an info argument, so it being optional makes more sense there from a functionality standpoint:

```python
            tuple_exp = parse_properties(cls, v)
```

sqlmesh/utils/pydantic.py
```diff
@@ -27,14 +27,10 @@
 
 
 def field_validator(*args: t.Any, **kwargs: t.Any) -> t.Callable[[t.Any], t.Any]:
```
georgesittas, line 29: I tried to import these validators instead of overriding them, which resulted in several mypy errors that were previously silenced due to the `t.Any` hints. For this reason I decided to keep them, at least temporarily. Fixing the mypy issue was a little annoying.

sqlmesh/core/_typing.py
```diff
@@ -9,3 +10,8 @@
     SchemaName = t.Union[str, exp.Table]
     SessionProperties = t.Dict[str, t.Union[exp.Expression, str, int, float, bool]]
     CustomMaterializationProperties = t.Dict[str, t.Union[exp.Expression, str, int, float, bool]]
+
+if sys.version_info >= (3, 11):
+    from typing import Self as Self
```
izeigerman, line 15: [Nit] Why do we need `as Self`?

georgesittas: Because it gets removed on formatting for some reason. May be able to use `# noqa`, too, didn't try it.

sqlmesh/core/model/common.py
izeigerman: So can we not make it optional?

sqlmesh/core/model/kind.py
```diff
@@ -407,7 +362,42 @@ class IncrementalByTimeRangeKind(_IncrementalBy):
     time_column: TimeColumn
     auto_restatement_intervals: t.Optional[SQLGlotPositiveInt] = None
 
-    _time_column_validator = TimeColumn.validator()
+    @field_validator("time_column", mode="before")
+    @classmethod
+    def _time_column_validator(cls, v: t.Any, values: t.Any) -> TimeColumn:
```
izeigerman, line 367: I think it was intentionally defined as part of the `TimeColumn` class to keep it closer to the actual model definition. Can we keep it there?

izeigerman: Why is it using `values: t.Any` and not `ValidationInfo`?

sqlmesh/core/model/kind.py
```diff
@@ -407,7 +362,42 @@ class IncrementalByTimeRangeKind(_IncrementalBy):
...
+        dialect = get_dialect(values)
```
izeigerman, line 368: Shouldn't this be `values.data`? The fact that our tests didn't catch this is kind of scary

sqlmesh/dbt/target.py
```diff
@@ -257,17 +255,17 @@ class SnowflakeConfig(TargetConfig):
...
-        if (
-            values.get("password")
-            or values.get("authenticator")
-            or values.get("private_key")
-            or values.get("private_key_path")
-        ):
-            return values
+    @classmethod
+    def validate_authentication(cls, data: t.Any) -> t.Any:
+        if isinstance(data, dict):
```
izeigerman, line 260: [Nit] Can this be a part of a single `if` block?

sqlmesh/dbt/target.py
```diff
@@ -401,14 +401,13 @@ class RedshiftConfig(TargetConfig):
...
-    def validate_database(
-        cls, values: t.Dict[str, t.Union[t.Tuple[str, ...], t.Optional[str], t.Dict[str, t.Any]]]
-    ) -> t.Dict[str, t.Union[t.Tuple[str, ...], t.Optional[str], t.Dict[str, t.Any]]]:
-        values["database"] = values.get("database") or values.get("dbname")
-        if not values["database"]:
-            raise ConfigError("Either database or dbname must be set")
-        return values
+    @classmethod
+    def validate_database(cls, data: t.Any) -> t.Any:
+        if isinstance(data, dict):
```
izeigerman, line 406: Let's follow the same style of early return

sqlmesh/integrations/github/cicd/config.py
```diff
@@ -30,13 +30,15 @@ class GithubCICDBotConfig(BaseConfig):
...
-    @model_validator_v1_args
-    def _validate(cls, values: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
-        if values.get("enable_deploy_command") and not values.get("merge_method"):
-            raise ValueError("merge_method must be set if enable_deploy_command is True")
-        if values.get("command_namespace") and not values.get("enable_deploy_command"):
-            raise ValueError("enable_deploy_command must be set if command_namespace is set")
-        return values
+    @classmethod
+    def _validate(cls, data: t.Any) -> t.Any:
+        if isinstance(data, dict):
```
izeigerman, line 35: ditto

sqlmesh/core/model/kind.py
georgesittas: It seems like the `get_dialect` implementation accounts for both `ValidationInfo` and `dict`, see [here](https://github.com/TobikoData/sqlmesh/blob/568b265355af6d35f0055831cdec0f70a75fff82/sqlmesh/utils/pydantic.py#L45-L57).

georgesittas: Changed it anyway though.

sqlmesh/core/model/kind.py
georgesittas: Updated.

sqlmesh/dbt/target.py
georgesittas: Done

sqlmesh/dbt/target.py
georgesittas: Yep, done everywhere

sqlmesh/core/model/common.py
georgesittas: I took another look, I'm not sure how this can be done because `ValidationInfo` is a protocol, so I can't instantiate a dummy instance for it in the test, or in the above `parse_properties` call.

One alternative is to extract the logic into a separate helper function that expects a dialect argument, and then call that instead so we can bypass the validation info argument altogether in other call sites. What do you think?

izeigerman: ok, never mind

PR #3617, opened 2025-01-10
Docs: Clarify "sqlmesh init" usage

techieshark:
Why:

Because if you follow the guide as is, you will get:

```
sqlmesh init
Error: Default SQL dialect is a required argument for SQLMesh projects
```

and it's not immediately clear what the list of supported values should be.


tobymao:
thank you

PR #3620, opened 2025-01-11
Feat: implement support for bigframes

z3z1ma:
I PR'ed the upstream lib so that they have no upper bound on their SQLGlot pin. This means the library is now fully compatible with SQLMesh. This PR adds support. Instead of bundling the extra with bigquery, I have temporarily added it to its own extra so it is opt-in. Once they cut a version release, we can add it with a proper `>=` version specifier. 

This enables many novel uses such as using Gemini / LLM capabilities at scale more effectively, using ML models (training , predicting, etc), and so on leveraging BigQuery's compute through a dataframe interface wrapped intuitively by `sqlmesh`'s python model system.

https://cloud.google.com/bigquery/docs/use-bigquery-dataframes#pandas-examples

https://github.com/googleapis/python-bigquery-dataframes/blob/main/notebooks/getting_started/ml_fundamentals_bq_dataframes.ipynb

I have tested this locally. I was able to crunch through 35G in 15 seconds with no local computation. Its very nice.

The following example requires no seeds or pre-loaded data as it leverages BQ's open datasets. You can run this with or without the remote function. 

```python
"""Working example to get your feet wet"""
import typing as t
from datetime import datetime

from bigframes.pandas import DataFrame

from sqlmesh import ExecutionContext, model


def get_bucket(num: int):
    if not num:
        return "NA"
    boundary = 10
    return "at_or_above_10" if num >= boundary else "below_10"


@model(
    "mart.wiki",
    columns={
        "title": "text",
        "views": "int",
        "bucket": "text",
    },
)
def execute(
    context: ExecutionContext,
    start: datetime,
    end: datetime,
    execution_time: datetime,
    **kwargs: t.Any,
) -> DataFrame:
    # Create a remote function to be used in the Bigframe DataFrame
    remote_get_bucket = context.bigframe.remote_function([int], str)(get_bucket)

    # Returns the Bigframe DataFrame handle, no data is computed locally
    df = context.bigframe.read_gbq("bigquery-samples.wikipedia_pageviews.200809h")

    df = (
        # This runs entirely on the BigQuery engine lazily
        df[df.title.str.contains(r"[Gg]oogle")]
        .groupby(["title"], as_index=False)["views"]
        .sum(numeric_only=True)
        .sort_values("views", ascending=False)
    )

    return df.assign(bucket=df["views"].apply(remote_get_bucket))
```

Support was simple given the way the `query_factory` works and similar art with Snowpark. Very straightforward. Good work on that.

You can even do interesting things like include a read from a gcs bucket in BQ engine:

```python
filepath_or_buffer = "gs://cloud-samples-data/bigquery/us-states/us-states.csv"
df_from_gcs = bpd.read_csv(filepath_or_buffer)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ setup(
             "google-cloud-bigquery[pandas]",
             "google-cloud-bigquery-storage",
         ],
+        "bigframes": ["bigframes>=1.32.0"],
         "clickhouse": ["clickhouse-connect"],
         "databricks": ["databricks-sql-connector"],
         "dev": [
--- a/sqlmesh/core/engine_adapter/_typing.py
+++ b/sqlmesh/core/engine_adapter/_typing.py
@@ -8,6 +8,8 @@ from sqlmesh.utils import optional_import
 if t.TYPE_CHECKING:
     import pyspark
     import pyspark.sql.connect.dataframe
+    from bigframes.session import Session as BigframeSession  # noqa
+    from bigframes.dataframe import DataFrame as BigframeDataFrame
 
     snowpark = optional_import("snowflake.snowpark")
 
@@ -23,6 +25,7 @@ if t.TYPE_CHECKING:
         pd.DataFrame,
         pyspark.sql.DataFrame,
         pyspark.sql.connect.dataframe.DataFrame,
+        BigframeDataFrame,
         SnowparkDataFrame,
     ]
 
--- a/docs/concepts/models/python_models.md
+++ b/docs/concepts/models/python_models.md
@@ -33,7 +33,7 @@ The `execute` function is wrapped with the `@model` [decorator](https://wiki.pyt
 
 Because SQLMesh creates tables before evaluating models, the schema of the output DataFrame is a required argument. The `@model` argument `columns` contains a dictionary of column names to types.
 
-The function takes an `ExecutionContext` that is able to run queries and to retrieve the current time interval that is being processed, along with arbitrary key-value arguments passed in at runtime. The function can either return a Pandas, PySpark, or Snowpark Dataframe instance.
+The function takes an `ExecutionContext` that is able to run queries and to retrieve the current time interval that is being processed, along with arbitrary key-value arguments passed in at runtime. The function can either return a Pandas, PySpark, Bigframe, or Snowpark Dataframe instance.
 
 If the function output is too large, it can also be returned in chunks using Python generators.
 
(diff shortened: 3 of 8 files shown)


sqlmesh/core/engine_adapter/_typing.py
```diff
@@ -8,6 +8,8 @@
 if t.TYPE_CHECKING:
     import pyspark
     import pyspark.sql.connect.dataframe
+    from bigframes.session import Session as BigframeSession  # noqa
+    from bigframes.dataframe import DataFrame as BigframeDataFrame
```
georgesittas, line 12: Are these imports safe? What if you don't have `bigframes` installed?

georgesittas (commented):
Awesome, thanks for driving this! LGTM, just one question to make sure the imports are safe. Will let others chime in here too.

sqlmesh/core/engine_adapter/_typing.py
z3z1ma: It works the same way as SQLMesh does without Snowpark in Python 3.12

The if t.TYPE_CHECKING means the code is never hit at runtime, its just for the type checker. And the type checker is okay if an import doesn't exist. At worst it complains. Hence why the whole test suit passes even though bigframes is not installed. 

cc @georgesittas

georgesittas: Gotcha, makes sense.

georgesittas (approved):

setup.py
```diff
@@ -58,6 +58,9 @@
             "google-cloud-bigquery[pandas]",
             "google-cloud-bigquery-storage",
         ],
+        "bigframes": [
+            "bigframes @ git+https://github.com/googleapis/python-bigquery-dataframes.git@c71ec093314409cd4c7a52a713dbd6164fbbd792"
```
tobymao, line 62: why is this pinned to git and not regular big frames?

z3z1ma: Waiting on a version release upstream. 

https://github.com/googleapis/python-bigquery-dataframes/pull/1278 (merged)
https://github.com/googleapis/python-bigquery-dataframes/pull/1283 (followup on googles side)

tobymao: maybe let's wait til a release?

tobymao (approved):

z3z1ma:
FYI @tobymao this PR is good to merge whenever you all want to, the test failure is completely unrelated I think.

georgesittas:
> FYI @tobymao this PR is good to merge whenever you all want to, the test failure is completely unrelated I think.

Yeah, this test is flaky for some reason, no need to worry about it.

PR #3622, opened 2025-01-11
fix: update default value for `invalidate_hard_deletes` to match docs

zzztimbo:
https://sqlmesh.readthedocs.io/en/stable/concepts/models/model_kinds/?h=invalidate+hard+deletes#deletes

> If invalidate_hard_deletes is set to true (default):
>  - valid_to column will be set to the time when the SQLMesh run started that detected the missing record (called execution_time).
>  - If the record is added back, then the valid_to column will remain unchanged.


tobymao:
@eakmanrq

PR #3627, opened 2025-01-14
Fix: Reduce parsing overhead and improve performance when unpausing snapshots deployed to prod

izeigerman:
Unpausing snapshots that are deployed to prod involves fetching all snapshots that share the same version as the ones being deployed. 

Before this change, fetching snapshots with the same versions involved full parsing of snapshot objects, including their models and queries. This caused a significant performance issue for projects with a high number of models, especially forward-only models.

With this update, SQLMesh will perform only a partial parsing when fetching snapshots with the same version, falling back to parsing the full snapshot only when absolutely necessary (which should be rare).

Additionally, this should also improve the performance of deleting expired snapshots, which relies on the same method to fetch snapshots with the same version.

### DIFF
--- a/sqlmesh/core/state_sync/engine_adapter.py
+++ b/sqlmesh/core/state_sync/engine_adapter.py
@@ -27,6 +27,7 @@ from pathlib import Path
 from datetime import datetime
 
 import pandas as pd
+from pydantic import Field
 from sqlglot import __version__ as SQLGLOT_VERSION
 from sqlglot import exp
 from sqlglot.helper import seq_get
@@ -37,10 +38,12 @@ from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.environment import Environment
 from sqlmesh.core.model import ModelKindName, SeedModel
+from sqlmesh.core.node import IntervalUnit
 from sqlmesh.core.snapshot import (
     Intervals,
     Node,
     Snapshot,
+    SnapshotChangeCategory,
     SnapshotFingerprint,
     SnapshotId,
     SnapshotIdLike,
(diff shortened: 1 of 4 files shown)


sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1942,3 +1963,94 @@ def __getitem__(self, snapshot_id: SnapshotId) -> Snapshot:
         if snapshot is None:
             raise KeyError(snapshot_id)
         return snapshot
+
+
+class SharedVersionSnapshot(PydanticModel):
```
tobymao, line 1968: should this go in snapshot definition?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1942,3 +1963,94 @@ def __getitem__(self, snapshot_id: SnapshotId) -> Snapshot:
...
+    @property
+    def snapshot_id(self) -> SnapshotId:
+        return SnapshotId(name=self.name, identifier=self.identifier)
+
+    @property
+    def is_forward_only(self) -> bool:
+        return self.change_category == SnapshotChangeCategory.FORWARD_ONLY
+
+    @property
+    def normalized_effective_from_ts(self) -> t.Optional[int]:
```
tobymao, line 1996: should we make a mixin for these shared methods or no, not worth it?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -464,8 +472,14 @@ def unpause_snapshots(
                         snapshot.snapshot_id,
                         target_snapshot.snapshot_id,
                     )
+                    full_snapshot = snapshot.full_snapshot
```
tobymao, line 475: do you think it's worth hitting the cache here, or allowing full_snapshot to take in a cache?

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: It's very specific to the State Sync implementation and is only used internally

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: It's the only one. Didn't want to bother for now, but I'm open to reconsider

sqlmesh/core/state_sync/engine_adapter.py
izeigerman: Maybe as a follow-up. It should happen rarely enough to not matter.

PR #3628, opened 2025-01-14
Fix: The inference of the default plan start

izeigerman:
The previous logic was overly simplistic and worked well only for daily models, while yielding poor results for models with a coarser cadence (weekly, monthly).

The new logic is based on interval units of models that constitute a plan.


tobymao:
so if end is 1/2 00:00, then start will be 1/1, but if end is 1/1 15:00 then start will be 12/31. i guess that's correct

tobymao (approved):

PR #3633, opened 2025-01-14
Fix: parse python model column types with dialect

treysp:
Currently, python model column data types are parsed with the default sqlglot dialect instead of the explicit model dialect or project default dialect. 

This PR updates it to use the explicit model dialect when present and default project dialect when not.

### DIFF
--- a/sqlmesh/core/model/decorator.py
+++ b/sqlmesh/core/model/decorator.py
@@ -71,7 +71,9 @@ class model(registry_decorator):
             column_name: (
                 column_type
                 if isinstance(column_type, exp.DataType)
-                else exp.DataType.build(str(column_type))
+                else exp.DataType.build(
+                    str(column_type), dialect=self.kwargs.get("dialect", self._dialect)
+                )
             )
             for column_name, column_type in self.kwargs.pop("columns", {}).items()
         }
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/decorator.py
```diff
@@ -71,7 +71,9 @@ def __init__(self, name: t.Optional[str] = None, is_sql: bool = False, **kwargs:
             column_name: (
                 column_type
                 if isinstance(column_type, exp.DataType)
-                else exp.DataType.build(str(column_type))
+                else exp.DataType.build(
+                    str(column_type), dialect=self.kwargs.get("dialect", self._dialect)
+                )
```
georgesittas, line 76: Where is `_dialect` set? It seems like that this is being manually set in the tests. Is the default dialect being propagated here somehow? Was this tested in a simple project?

treysp: It's set in the python model loader https://github.com/TobikoData/sqlmesh/blob/295d37a21ae5ea77d02382bf0819c701f3c23edf/sqlmesh/core/loader.py#L471

Historical context: https://github.com/TobikoData/sqlmesh/pull/2527

treysp: I tested it locally in a modified quickstart project - is that what you're asking?

georgesittas: Yep, and thanks for sharing that snippet 👍

georgesittas (approved):

erindru (approved):

PR #3635, opened 2025-01-15
Feat: pass model tags to the airflow ui

petrikoro:
This pull request enhances the Airflow integration with SQLMesh by adding support for model tags in Airflow UI.


izeigerman (approved):
Thanks!

PR #3643, opened 2025-01-16
Fix: ensure strict defaults dont break all the models to which they a…

z3z1ma:
…re not applicable.

Currently using a basic setting like this: 

```py
config = sqlmesh.Config(
    gateways={"prod": prod, "dev": dev},
    default_gateway="dev",
    model_defaults=ModelDefaultsConfig(
        dialect="bigquery",
        start="2020-01-01",
        validate_query=True,  # NOTE: it is bugged -- it is applying itself to SEEDs and 100s of EXTERNAL models indiscriminately and the throwing error in `validate_definition`
    )
)
```

Causes a project to be un-loadable as all models including seeds and external models would receive this default argument, then raise an error in the [validate_definition](https://github.com/z3z1ma/sqlmesh/blob/fb943d25ef387dc4c0358e010e7e5e86eb4779db/sqlmesh/core/model/definition.py#L919-L929). Considering folks can have many seeds and hundreds of external models, explicitly setting validate_query false on them when you set a project wide setting seems suboptimal compared to fixing it here. :) 

I ensured the test case I added fails without this change and passes with it.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2087,6 +2087,9 @@ def _create_model(
         kwargs["kind"] = create_model_kind(raw_kind, dialect, defaults or {})
 
     defaults = {k: v for k, v in (defaults or {}).items() if k in klass.all_fields()}
+    if not issubclass(klass, SqlModel):
+        defaults.pop("optimize_query", None)
+        defaults.pop("validate_query", None)
 
     statements = []
 
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -7032,6 +7032,17 @@ col_a,col_b,col_c
     context.upsert_model(model)
     context.plan(auto_apply=True, no_prompts=True)
 
+    # Ensure strict defaults don't break all non SQL models to which they weren't applicable in the first place
+    seed_strict_defaults = create_seed_model(
+        "test_db.test_seed_model", model_kind, defaults=strict_default
+    )
+    external_strict_defaults = create_external_model(
+        "test_db.test_external_model", columns={"a": "int", "limit": "int"}, defaults=strict_default
+    )
+    context.upsert_model(seed_strict_defaults)
+    context.upsert_model(external_strict_defaults)
+    context.plan(auto_apply=True, no_prompts=True)
+
 
 def test_partition_interval_unit():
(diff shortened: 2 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -2082,11 +2082,15 @@ def _create_model(
...
+    defaults = {k: v for k, v in (defaults or {}).items() if k in klass.all_fields()}
+
     raw_kind = kwargs.pop("kind", None)
     if raw_kind:
-        kwargs["kind"] = create_model_kind(raw_kind, dialect, defaults or {})
-
-    defaults = {k: v for k, v in (defaults or {}).items() if k in klass.all_fields()}
+        parsed_kind = create_model_kind(raw_kind, dialect, defaults or {})
+        kwargs["kind"] = parsed_kind
+        if parsed_kind.is_external or parsed_kind.is_seed or issubclass(klass, PythonModel):
```
georgesittas, line 2091: Should this be `if klass is not SqlModel` or something?

georgesittas (commented):
Thanks for catching this!

sqlmesh/core/model/definition.py
z3z1ma: Ah yeah maybe so, can simplify by not relying on kind at all I suppose, 1 sec

tobymao: hey is this the best place to put this?, what about in the validator or something

z3z1ma: Yeah @tobymao , by the time you are in the context of a validator -- the defaults have already been mixed in with explicit arguments. We can see it happening just below my change
https://github.com/z3z1ma/sqlmesh/blob/927d4cba9d87f219e543f737fc470e257dd33ec6/sqlmesh/core/model/definition.py#L2120-L2130

The behavior my change gives us (as can be seen in the unmutated test cases above my new one) is that we still raise if a user explicitly adds `validate_query TRUE` to a SEED or EXTERNAL model in the definition but do NOT raise by injecting a default since we filter it appropriately before it reaches the validators.  

Also latest change is nice and doesn't even look at kind, its more correct now I think and a simple change

georgesittas (approved):

PR #3644, opened 2025-01-16
Fix: no models ready to run message shouldn't error 

treysp:
When no models are ready to run, the `sqlmesh run` command issues a message stating when the next model will be ready to run.

Some model kinds, like `seed` and `symbolic`, do not have a "ready" time. Currently, we error if a project only contains these kinds of models.

This PR updates the code to handle this situation, omitting the time when next model ready from the message if it cannot be calculated.


georgesittas (approved):
Thanks for checking this out! FYI @benfdking.

Issue #3624, opened 2025-01-13
Does end_dt provide sufficient precision for a Microsoft SQL TIMESTAMP2 column?


martinburch:
With a model defined like
```
MODEL (
    kind INCREMENTAL_BY_TIME_RANGE (
        time_column EventCreatedAtUtc
    )
);

SELECT * FROM t WHERE
t.CreatedAtUtc BETWEEN @start_dt and @end_dt
```

I'm seeing a query in the logs like
```
[createdatutc] BETWEEN CAST('2024-07-01 00:00:00+00:00' AS DATETIMEOFFSET) AT TIME ZONE 'UTC'
AND CAST('2025-01-12 23:59:59.999999+00:00' AS DATETIMEOFFSET) AT TIME ZONE 'UTC')
```

CreatedAtUtc is defined as `[datetime2](7) NOT NULL`

Won't SQLMesh's query fail to capture records created between 59.999999**0** and 59.999999**9**? How can I account for this loss of precision?

martinburch:
On further inspection of the T-SQL docs, I'm going to call this issue a bug: SQLMesh doesn't provide sufficient precision for default DATETIMEOFFSET(7) in T-SQL dialect when building `_dt` variables.
https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver16

tobymao:
sqlmesh defaults to 6 decimals of precision. if you want full precision, it may be better to write a macro to convert end_dt to the next day and then do an inclusive exclusive operation

so for example

to find the next end_dt, you can write a python macro or something that bumps the date

created_at >= start_dt and created_at < end_dt_next

i'm going to close this for now. but happy to discuss, i don't think there's an easy fix for this because python/most sql dialects only support 6 digits, also the impact of this is relatively small and there are work arounds like moving to inclusive/exclusive date ranges.

martinburch:
I think there could be a dialect-specific fix for mssql/t-sql precision, as seen in the clickhouse dialect-specific fix: https://github.com/TobikoData/sqlmesh/blob/b04cc4b215b9d2b289073d0ed2afa1e0e5c4a354/sqlmesh/utils/date.py#L360

[...]

PR #3648, opened 2025-01-17
Fix!: Support full precision for TSQL date types

themisvaltinos:
This update aims to enhance the support for MSSQL/T-SQL full precision when building the `end_dt` macro and date ranges, specifically for [date types](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-ver16) with up to 100 nano seconds precision, fixes: #3624


sqlmesh/core/snapshot/definition.py
```diff
@@ -987,20 +985,6 @@ def is_valid_start(
                 return earliest_interval >= start_ts
         return True
 
-    def get_latest(self, default: t.Optional[TimeLike] = None) -> t.Optional[TimeLike]:
```
themisvaltinos, line 990: Note: This was removed because this function appears to not be used anymore.

tobymao: this could be used by clients in the api, it's a useful method, i think we should keep it

sqlmesh/utils/date.py
```diff
@@ -239,14 +244,16 @@ def date_dict(
     ]
 
     if start is not None:
-        prefixes.append(("start", to_datetime(start)))
+        prefixes.append(("start", start if dialect == "tsql" else to_datetime(start)))
```
tobymao, line 247: why do you need to change start? for tsql?

sqlmesh/utils/date.py
```diff
@@ -308,6 +318,14 @@ def make_inclusive(start: TimeLike, end: TimeLike) -> DatetimeRange:
     Returns:
         A tuple of inclusive datetime objects.
     """
+
+    # For nanosecond precision in T-SQL; Python's datetime supports up to microsecond
+    if dialect == "tsql":
+        end_ts = to_utc_timestamp(end)
+        if is_date(end):
+            end_ts += pd.Timedelta(1, unit="day")
+        return (to_utc_timestamp(start), end_ts - pd.Timedelta(100, unit="ns"))
```
tobymao, line 327: how does this work? we're no longer returning datetimes, now we're returning pandas objects?

sqlmesh/core/snapshot/definition.py
themisvaltinos: right I noted this in case there was some external usage I wasn't aware of. I'll revert the changes to keep this as is then

sqlmesh/utils/date.py
themisvaltinos: actually it's not needed so I will change it back. i had it for concistency so that both start_dt and end_dt were panda timestamps

sqlmesh/utils/date.py
themisvaltinos: revised it to return pandas `Timestamp` to handle the nanosecond precision, and as it extends Python's `datetime` it allows us to use the same methods we were using with `datetime`

sqlmesh/utils/date.py
```diff
@@ -295,33 +297,40 @@ def make_inclusive(start: TimeLike, end: TimeLike) -> DatetimeRange:
...
+
     return (to_datetime(start), make_inclusive_end(end))
 
 
 def make_inclusive_end(end: TimeLike) -> datetime:
-    return make_exclusive(end) - timedelta(microseconds=1)
+    return make_exclusive(end) - pd.Timedelta(1, unit="ns")
 
 
 def make_exclusive(time: TimeLike) -> datetime:
```
tobymao, line 321: is this type still right is datetime and pd.Timedelta a subclass?

sqlmesh/utils/date.py
```diff
@@ -295,33 +297,40 @@ def make_inclusive(start: TimeLike, end: TimeLike) -> DatetimeRange:
...
-    dt = to_datetime(time)
+    dt = to_utc_timestamp(to_datetime(time))
     if is_date(time):
-        dt = dt + timedelta(days=1)
+        dt = dt + pd.Timedelta(1, unit="day")
```
tobymao, line 324: does time delta no longer work?

tobymao:
do we need a migration for this? cc @georgesittas

sqlmesh/utils/date.py
themisvaltinos: good point, adding a one day `timedelta` works here so no need for pandas in this

sqlmesh/utils/date.py
georgesittas: yep, for example:

```python
>>> pd.Timestamp("2020-01-01") + pd.Timedelta(1, unit="day")
Timestamp('2020-01-02 00:00:00')
>>> issubclass(pd.Timestamp, datetime.datetime)
True
```

tobymao (approved):

georgesittas:
> do we need a migration for this? cc @georgesittas

There's definitely a diff between main and this PR. E.g. if I change the filter in the example incremental model to:

```sql
-- instead of `@start_date` and `@end_date`
event_date BETWEEN @start_dt AND @end_dt
```

Then I see this after running first plan, switching to this branch, wiping out cache and re-running plan:

```
Models:
├── Directly Modified:
│   └── sqlmesh_example.incremental_model
└── Indirectly Modified:

Directly Modified: sqlmesh_example.incremental_model (Non-breaking)
└── Indirectly Modified Children:
    └── sqlmesh_example.full_model (Indirect Non-breaking)
└── sqlmesh_example.incremental_model: 2020-01-01 - 2025-01-21
```

tobymao (approved):

PR #3657, opened 2025-01-18
Fix: Defer creation of the deploable table for a new forward-only model

izeigerman:
After adding a new forward-only model in a dev environment, the user may later iterate on it by adding or removing columns. 

If `on_destructive_change` is set to `fail`, the user might encounter an error when removing a previously existing column.

This behavior is generally expected and correct, assuming the model was previously deployed to prod. However, if this is a new model that only exists within the current dev environment, failing due to destructive changes is counterintuitive. Nothing will actually be destroyed as a result of the change, since no production copy exists yet.

This fix addresses the issue by deferring the creation of the deployable physical table for a new forward-only model until it is deployed to prod.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -762,6 +762,8 @@ class SnapshotEvaluator:
                 and adapter.SUPPORTS_CLONING
                 # managed models cannot have their schema mutated because theyre based on queries, so clone + alter wont work
                 and not snapshot.is_managed
+                # If the deployable table is missing we can't clone it
+                and True not in deployability_flags
             ):
                 target_table_name = snapshot.table_name(is_deployable=False)
                 tmp_table_name = f"{target_table_name}__schema_migration_source"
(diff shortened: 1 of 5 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -762,6 +762,8 @@ def _create_snapshot(
                 and adapter.SUPPORTS_CLONING
                 # managed models cannot have their schema mutated because theyre based on queries, so clone + alter wont work
                 and not snapshot.is_managed
+                # If the deployable table is missing we can't clone it
+                and True not in deployability_flags
```
tobymao, line 766: why was this added?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -797,6 +799,19 @@ def _create_snapshot(
             else:
                 dry_run = len(deployability_flags) == 1
                 for is_table_deployable in deployability_flags:
+                    if (
+                        is_table_deployable
+                        and snapshot.model.forward_only
+                        and not is_snapshot_representative
```
tobymao, line 805: how can a table be deployable but not represenatitve?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -829,15 +844,47 @@ def _migrate_snapshot(
...
+
         target_table_name = snapshot.table_name()
-        _evaluation_strategy(snapshot, adapter).migrate(
-            target_table_name=target_table_name,
-            source_table_name=tmp_table_name,
-            snapshot=snapshot,
-            snapshots=parent_snapshots_by_name(snapshot, snapshots),
-            allow_destructive_snapshots=allow_destructive_snapshots,
-        )
+        if adapter.table_exists(target_table_name):
```
tobymao, line 850: so this will only be true when it's been deployed to prod, because otherwise only the temp table will exist?

tobymao (approved):

sqlmesh/core/snapshot/evaluator.py
izeigerman: It's a deployable table for a non-deployable snapshot. That can be the case

sqlmesh/core/snapshot/evaluator.py
izeigerman: Exactly right

PR #3667, opened 2025-01-20
Fix!: render audits at runtime

georgesittas:
The motivation behind this PR is that, given this model:

```sql
MODEL (
  name test_model,
    unique_key (a, d)
  ),
  audits (
    unique_combination_of_columns(
      columns = (a, d),
      condition = d between @start_date and @end_date
    )
  ),
);

SELECT
  1 AS a,
  CAST('2025-01-20' AS DATE) AS d
```

The audit condition renders as follows at runtime today:

```sql
"d" BETWEEN CAST('1970-01-01' AS DATE) AND CAST('1970-01-01' AS DATE)
```

The reason this happens is that the model meta is rendered at load time and we store the properties to the corresponding pydantic fields, e.g. `audits`, even though `start_date` is a variable that changes depending on the runtime stage.

Approach is similar to https://github.com/TobikoData/sqlmesh/commit/e10449184cea6369c0ac2dcd1e6b8736d2a612fb.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1765,11 +1765,13 @@ def load_sql_based_model(
 
     unrendered_merge_filter = None
     unrendered_signals = None
+    unrendered_audits = None
 
     for prop in meta.expressions:
         if prop.name.lower() == "signals":
             unrendered_signals = prop.args.get("value")
-
+        if prop.name.lower() == "audits":
+            unrendered_audits = prop.args.get("value")
         if (
             prop.name.lower() == "kind"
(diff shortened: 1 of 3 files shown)


erindru (approved):
LGTM

Would this change the fingerprint of existing models that had audits?

tobymao (approved):

georgesittas:
> Would this change the fingerprint of existing models that had audits?

It does, but I didn't notice it during testing because the old, rendered version of `audits` had been cached locally and so there was no metadata diff! Great find, thanks.

Going to include an empty migration script.

Issue #3477, opened 2024-12-05
Investigate possible refactors after deprecating Python 3.8


georgesittas:
Support for Python 3.8 was dropped in https://github.com/TobikoData/sqlmesh/commit/7a7a9f61c31be23ff26f95903e9106f456dbfa8a, which means there may be an opportunity to refactor a few places:

1. https://github.com/TobikoData/sqlmesh/commit/d6f3c7cb3178ab7efc632f20cd63b7501abfce22: `add custom source parser because it's broken in < 3.9` – simplify the definition of `_is_relative_to` [1]
2. https://github.com/TobikoData/sqlmesh/commit/92242e2de7bee0c7e89de1cfc7d956e18b1d339c: `allow python env building to use relative paths` – simplify the python source parser
3. https://github.com/TobikoData/sqlmesh/commit/b4dbb4587edaf7eb78103479c449d915542f0a51: `Fix: Remove intervals instead missing intervals` – address the TODO in `dag.py`

[1]: we may not be able to get rid of this completely, because it's been [extended](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/metaprogramming.py#L37-L38) since the first implementation.

georgesittas:
Did some quick investigation, if we want we can address (2) using the linked PR in the future, the other two don't seem high ROI for now.

PR #3677, opened 2025-01-21
Chore!: remove custom source parsing

georgesittas:
Addresses item (2) in https://github.com/TobikoData/sqlmesh/issues/3477.

### DIFF
--- a/tests/utils/test_metaprogramming.py
+++ b/tests/utils/test_metaprogramming.py
@@ -207,7 +207,8 @@ def test_serialize_env() -> None:
     env = serialize_env(env, path=path)  # type: ignore
 
     assert prepare_env(env)
-    assert env == {
+
+    expected_env = {
         "MAIN": Executable(
             name="main_func",
             alias="MAIN",
@@ -315,5 +316,6 @@ def test_context_manager():
             payload="from tests.utils.test_metaprogramming import test_context_manager",
             kind=ExecutableKind.IMPORT,
         ),
-        "wraps": Executable(payload="from functools import wraps", kind=ExecutableKind.IMPORT),
     }
+
+    assert env == expected_env
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -4,7 +4,6 @@ import ast
 import dis
 import importlib
 import inspect
-import linecache
 import os
 import re
 import sys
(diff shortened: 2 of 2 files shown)


tests/utils/test_metaprogramming.py
```diff
@@ -315,5 +316,6 @@ def test_context_manager():
             payload="from tests.utils.test_metaprogramming import test_context_manager",
             kind=ExecutableKind.IMPORT,
         ),
-        "wraps": Executable(payload="from functools import wraps", kind=ExecutableKind.IMPORT),
```
georgesittas, line 318: FYI, I checked the before vs after by putting a breakpoint [here](https://github.com/TobikoData/sqlmesh/blob/c96d3c3edd4b5bd6c63594c2696ac073f7ad302e/sqlmesh/utils/metaprogramming.py#L282-L284), and this is what I saw:

```python
# main
(Pdb) p obj
<function test_context_manager at 0x13f7e00e0>
(Pdb) func_globals(obj).items()
dict_items([('_GeneratorContextManager', <class 'contextlib._GeneratorContextManager'>), ('wraps', <function wraps at 0x102e14040>), ('func', <function test_context_manager at 0x13f7e0040>)])


# jo/remove_custom_source_parsing
(Pdb) p obj
<function test_context_manager at 0x12d173e20>
(Pdb) func_globals(obj).items()
dict_items([('_GeneratorContextManager', <class 'contextlib._GeneratorContextManager'>), ('contextmanager', <function contextmanager at 0x100691da0>), ('func', <function test_context_manager at 0x12d173d80>)])
```

It looks like `inspect.getsource` brings back `contextmanager` directly instead of `wraps`, which we used to get with the custom source parser.

----

On a separate note, I think the payload changing here means we could see a diff in projects due to this change. Not sure how we should / can deal with this, though. Any thoughts?

tobymao: can you migrate it or no

georgesittas: Not sure how we could migrate it. It would involve doing surgical removes / additions in the payload which is a little sketchy.

tobymao: do you see a diff or no? i’ll let @izeigerman decide if we want this or not

georgesittas: Yes, I see a diff. Given the above model (see dropdown), re-running plan after switching to this branch results in:

```

Models:
└── Directly Modified:
    └── default.foo
---

+++

@@ -18,7 +18,6 @@

 import sqlglot
 from models.model import test_context_manager
 from sqlglot.expressions import to_table
-from functools import wraps
 KLASS_X = 1
 KLASS_Y = 2
 KLASS_Z = 3
Directly Modified: default.foo (Breaking)
└── default.foo: 2025-01-20 - 2025-01-21
Creating physical table ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 1/1 • 0:00:00

Model versions created successfully

z
[1/1] default.foo evaluated in 0.05s
Evaluating models ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 1/1 • 0:00:00


Model batches executed successfully

Virtually Updating 'prod' ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 0:00:00

Target environment updated successfully
```

georgesittas: It should be noted that `from functools import wraps` is a transitive dependency, because `contextmanager` (actually used in the model) is implemented through `functool.wraps`. We would previously include this in the serialized environment, but it looks like the plan succeeds without it.

izeigerman: @georgesittas if we can't safely migrate this then we shouldn't do this

izeigerman: or at least make it a metadata change

georgesittas:
Closing this for now to avoid dealing with unexpected side-effects (e.g. migration issues). Can revisit in the future if needed.

PR #3687, opened 2025-01-22
Fix!: include names of decorator argument references when building python env

### LINKED ISSUE #3640: SQLMesh decorator processing fails with AttributeError for imported decorators
## Description
When using imported decorators in SQLMesh models, the processing fails with `AttributeError: 'Attribute' object has no attribute 'id'`. Possible cause: SQLMesh's AST processing assumes all decorators are `Name` nodes, but they can also be `Attribute` nodes depending on the import style.

### Environment
- Python 3.12
- SQLMesh latest
- tenacity (or other decorators?)

### Steps
1. Create a new file `test_model.py`:
```python
from tenacity import retry, stop_after_attempt
from sqlmesh import model
import pandas as pd
from datetime import datetime
from sqlmesh import ExecutionContext
import typing as t


@retry(stop=stop_after_attempt(3))
def fetch_data():
    return "test data"

@model(
    name='test.raw_data',
    description='Test model with retry decorator',
    columns={
        "data": "text",
        "_sqlmesh__extracted_at": "timestamp"
    },
)
def execute(
    context: ExecutionContext,
    start: datetime,
    end: datetime,
    execution_time: datetime,
    **kwargs: t.Any,
) -> pd.DataFrame:
    data = fetch_data()
    return pd.DataFrame([{
        "data": data,
        "_sqlmesh__extracted_at": execution_time
    }])
```

2. Run SQLMesh:
```bash
sqlmesh plan
```

### Error
```
AttributeError: 'Attribute' object has no attribute 'id'
```

### Expected Behavior
The model should process successfully, handling the `@retry` decorator.

georgesittas:
Fixes #3640

The main difficulty in solving the linked issue was making SQLMesh detect the `stop_after_attempt` reference, in order to extract and inject it into the python environment so it can be (de)serialized:

```python
@retry(stop=stop_after_attempt(3))
def fetch_data():
    return "test data"
```

The decorator names themselves were already being picked up in the previous `decorators` helper. In this PR I expand that functionality by also searching for the decorator call arguments' references.

I implemented a new visitor class in order to limit the search to only the decorator sub-trees, instead of traversing the whole function tree. Without this, we'd have to do a nested `walk` within the root node traversal loop, leading to unnecessarily revisiting nodes under the decorator sub-trees.

Other than that, I also made sure to exclude callable instances of classes, because they can't be serialized. One such example is the `tenacity.Retrying` class ([source](https://github.com/jd/tenacity/blob/0d40e76f7d06d631fb127e1ec58c8bd776e70d49/tenacity/__init__.py#L652-L654)). Letting these **instances** into the python environment doesn't "just work" unfortunately, because an error is raised once we reach this [section](https://github.com/TobikoData/sqlmesh/blob/75468112ca4796adada3b6d47ea737d926e2c29d/sqlmesh/utils/metaprogramming.py#L357-L359), since there's no `__name__` attribute in them.

I verified that this fix works for a project with the model of interest, as well as the example project we have under `test_metaprogramming.py`.

### DIFF
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -58,7 +58,7 @@ def _code_globals(code: types.CodeType) -> t.Dict[str, None]:
     return variables
 
 
-def func_globals(func: t.Callable, path: t.Optional[Path] = None) -> t.Dict[str, t.Any]:
+def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
     """Finds all global references and closures in a function and nested functions.
 
     This function treats closures as global variables, which could cause problems in the future.
@@ -91,16 +91,9 @@ def func_globals(func: t.Callable, path: t.Optional[Path] = None) -> t.Dict[str,
             if ref:
                 variables[var] = ref
 
-        # Extract values visible to a closure (i.e., its "globals") only if it's defined in a
-        # file relative to the project's path. This is meant to block traversing third-party
-        # library closures, which could result in extracting unnecessary global references.
-        if func.__closure__ and _is_relative_to(func.__globals__.get("__file__"), path):
+        if func.__closure__:
             for var, value in zip(code.co_freevars, func.__closure__):
                 variables[var] = value.cell_contents
-
-        # If we can access the decorated function directly, then extract its globals too
-        if hasattr(func, "__wrapped__"):
-            variables.update(func_globals(func.__wrapped__, path=path))
 
     return variables
 
@@ -319,7 +312,7 @@ def build_env(
                     else:
                         build_env(v, env=env, name=v.__name__, path=path)
         elif callable(obj):
-            for k, v in func_globals(obj, path=path).items():
+            for k, v in func_globals(obj).items():
                 build_env(v, env=env, name=k, path=path)
 
     if name not in env:
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -83,7 +83,9 @@ def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
         ]
 
         code = func.__code__
-        for var in arg_globals + list(_code_globals(code)) + decorators(func, root_node=root_node):
+        for var in (
+            arg_globals + list(_code_globals(code)) + decorator_vars(func, root_node=root_node)
+        ):
             ref = func.__globals__.get(var)
             if ref:
                 variables[var] = ref
@@ -127,6 +129,38 @@ class _ClassFinder(ast.NodeVisitor):
             raise ClassFoundException(line_number)
         self.generic_visit(node)
         self.stack.pop()
+
+
+class _DecoratorDependencyFinder(ast.NodeVisitor):
+    def __init__(self, dependencies: t.List[str]) -> None:
+        self.dependencies = dependencies
+
+    def _extract_dependencies(self, node: ast.ClassDef | ast.FunctionDef) -> None:
+        for decorator in node.decorator_list:
+            dependencies: t.List[str] = []
+            for n in ast.walk(decorator):
+                if isinstance(n, ast.Attribute):
+                    dep = n.attr
+                elif isinstance(n, ast.Name):
+                    dep = n.id
+                else:
+                    continue
+
+                if dep in IGNORE_DECORATORS:
+                    dependencies = []
+                    break
+
+                dependencies.append(dep)
+
+            self.dependencies.extend(dependencies)
+
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        self._extract_dependencies(node)
+
+    def visit_ClassDef(self, node: ast.ClassDef) -> None:
+        self._extract_dependencies(node)
+
+    visit_AsyncFunctionDef = visit_FunctionDef  # type: ignore
 
 
 def getsource(obj: t.Any) -> str:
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/metaprogramming.py
```diff
@@ -179,25 +213,22 @@ def parse_source(func: t.Callable) -> ast.Module:
...
-    decorators = []
-
-    for node in ast.walk(root_node):
-        if isinstance(node, (ast.FunctionDef, ast.ClassDef)):
-            for decorator in node.decorator_list:
-                name = _decorator_name(decorator)
-                if name not in IGNORE_DECORATORS:
-                    decorators.append(name)
-    return unique(decorators)
+    _DecoratorDependencyFinder(names).visit(root_node)
```
izeigerman, line 230: [Nit] Rather than passing the `names` reference to be mutated inside, I think I bit more obvious control flow would be:
```
finder = _DecoratorDependencyFinder()
finder.visit(roo_node)
return unique(finder.dependencies)
```

tests/utils/test_metaprogramming.py
```diff
@@ -316,4 +326,25 @@ def test_context_manager():
             kind=ExecutableKind.IMPORT,
         ),
         "wraps": Executable(payload="from functools import wraps", kind=ExecutableKind.IMPORT),
+        "functools": Executable(payload="import functools", kind=ExecutableKind.IMPORT),
+        "retry": Executable(payload="from tenacity import retry", kind=ExecutableKind.IMPORT),
+        "stop_after_attempt": Executable(
+            payload="from tenacity.stop import stop_after_attempt", kind=ExecutableKind.IMPORT
+        ),
+        "wrapped_f": Executable(
```
izeigerman, line 334: Why is this import included twice?

tests/utils/test_metaprogramming.py
```diff
@@ -316,4 +326,25 @@ def test_context_manager():
             kind=ExecutableKind.IMPORT,
         ),
         "wraps": Executable(payload="from functools import wraps", kind=ExecutableKind.IMPORT),
+        "functools": Executable(payload="import functools", kind=ExecutableKind.IMPORT),
```
izeigerman, line 329: Why do we even capture t his import?

sqlmesh/utils/metaprogramming.py
georgesittas: Ah, yeah that's cleaner, good call.

tests/utils/test_metaprogramming.py
georgesittas: Because it's included in argument list of `fetch_data`'s decorator:

```
(Pdb) p decorator_vars(func, root_node=root_node)
['wraps', 'f', 'functools', 'WRAPPER_ASSIGNMENTS']
```

Due to [this](https://github.com/jd/tenacity/blob/0d40e76f7d06d631fb127e1ec58c8bd776e70d49/tenacity/__init__.py#L329) line, specifically.

tests/utils/test_metaprogramming.py
georgesittas: This [section](https://github.com/TobikoData/sqlmesh/blob/b71bd97b9746e03812e7c77a61dfb9bcb368e95f/sqlmesh/utils/metaprogramming.py#L91-L93) brings in the additional "weird" references:

```python
(Pdb) p list(zip(code.co_freevars, func.__closure__))
[('f', <cell at 0x143ae7430: function object at 0x143b1dc60>), ('self', <cell at 0x143ae73a0: Retrying object at 0x143ae6600>), ('wrapped_f', <cell at 0x143ae73d0: function object at 0x143b1e160>)]
```

This checks out with tenacity's decorator implementation (notice the `f` and `wrapped_f` definitions): https://github.com/jd/tenacity/blob/0d40e76f7d06d631fb127e1ec58c8bd776e70d49/tenacity/__init__.py#L322-L346.

So, regarding the two imports:

- The first one corresponds to the `ref` extracted using `_code_globals(code)` on the `execute` function (it brings in the names `['fetch_data', 'pd']`)
- The second one corresponds to `wrapped_f`.

I have an idea on how to fix this.

tests/utils/test_metaprogramming.py
```diff
@@ -316,4 +326,25 @@ def test_context_manager():
...
+            payload="from tests.utils.test_metaprogramming import fetch_data",
+            kind=ExecutableKind.IMPORT,
+        ),
+        "f": Executable(
+            payload="""@retry(stop=stop_after_attempt(3))
+def fetch_data():
+    return 'test data'""",
+            name="fetch_data",
+            path="test_metaprogramming.py",
+            alias="f",
```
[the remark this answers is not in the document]
georgesittas: I think [this reply](https://github.com/TobikoData/sqlmesh/pull/3687#discussion_r1928995768) answers it.

tests/utils/test_metaprogramming.py
georgesittas: @izeigerman @tobymao thinking out loud: what about only extracting closure variables for functions that are defined within the module path corresponding to the active sqlmesh context? Something along the lines of:

```diff
-        if func.__closure__:
-            for var, value in zip(code.co_freevars, func.__closure__):
-                variables[var] = value.cell_contents
+        if func.__closure__ and _is_relative_to(func.__globals__.get("__file__"), path):
+                for var, value in zip(code.co_freevars, func.__closure__):
+                    variables[var] = value.cell_contents
+
+        if hasattr(func, "__wrapped__"):
+            variables.update(func_globals(func.__wrapped__, path=path))
```

Given these changes I'm seeing two failed tests:

1. `y` needs to be excluded [here](https://github.com/TobikoData/sqlmesh/blob/b71bd97b9746e03812e7c77a61dfb9bcb368e95f/tests/utils/test_metaprogramming.py#L163)
2. `'f'`, `'func'` and `'wrapped_f'` need to be excluded from the `test_serialize_env` test in this PR (`functools` is still being picked up by `decorator_vars` above this section)

I added the branch with `__wrapped__` because without it the `retry` and `stop_after_attempt` dependencies aren't included in the serialized env, which I believe is incorrect?

izeigerman (approved):

tests/utils/test_metaprogramming.py
georgesittas: The first failure I mentioned was due to not updating the `test_func_globals`, after all. After including the `path` argument, it works as expected.

georgesittas: I ended up reverting this last change, because it could lead to diffs (in this case, due to `"func"` [missing](https://github.com/TobikoData/sqlmesh/pull/3687/commits/879a08fbca8689c70c04e9320f3e1e6bfe1178be#diff-396dbfdae476e8cfa1dac0201675484b84bdec45b725a721acb9e1fca570f672L289)):

image

PR #3688, opened 2025-01-22
Docs: update incremental by partition concepts

treysp:
Adds info box "should you use this model kind?" explaining table partitioning vs. loading by partition and that other model kinds are usually a better option.

Adds warning box that all restatements fully rebuild the table and should be done with caution.

### DIFF
--- a/docs/concepts/models/model_kinds.md
+++ b/docs/concepts/models/model_kinds.md
@@ -148,7 +148,7 @@ Models of the `INCREMENTAL_BY_PARTITION` kind are computed incrementally based o
 
     `INCREMENTAL_BY_PARTITION` models are inherently [non-idempotent](../glossary.md#idempotency), so restatements and other actions can cause data loss. This makes them more complex to manage than other model kinds.
 
-    In most scenarios, another model kind can meet your needs and will be easier to manage. The `INCREMENTAL_BY_PARTITION` model kind should only be used when the data must be loaded by partition (usually for performance reasons).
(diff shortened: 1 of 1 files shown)


docs/concepts/models/model_kinds.md
```diff
@@ -142,11 +142,25 @@ Depending on the target engine, models of the `INCREMENTAL_BY_TIME_RANGE` kind a
...
+    Any model kind can use a partitioned table by specifying the [`partitioned_by` key](../models/overview.md#partitioned_by) in the `MODEL` DDL. The "partition" in `INCREMENTAL_BY_PARTITION` is about how the data is **loaded** when the model runs.
 
-This kind is a good fit for datasets that have the following traits:
+    `INCREMENTAL_BY_PARTITION` models are inherently [non-idempotent](../glossary.md#idempotency), so restatements and other actions can cause data loss. This makes them more complex to manage than other model kinds.
+
+    In most scenarios, another model kind can meet your needs and will be easier to manage. The `INCREMENTAL_BY_PARTITION` model kind should only be used when the data must be loaded by partition (usually for performance reasons).
```
tobymao, line 151: most likely the kind is iscremental by time range

treysp: LIke this?

> In most scenarios, an `INCREMENTAL_BY_TIME_RANGE` model can meet your needs and will be easier to manage. The `INCREMENTAL_BY_PARTITION` model kind should only be used when the data must be loaded by partition (usually for performance reasons).

PR #3689, opened 2025-01-22
Fix: Make sure that full / view models only backfilled once

izeigerman:
When there are gaps in missing intervals like in the following example:
```
[1, 2) [3, 4)
```
SQLMesh ends up backfilling both `[1, 2)` and `[3, 4)` intervals even for non-incremental model kinds like `FULL` and `VIEW`. This behavior is correct for incremental models but doesn't make much sense for non-incremental ones.

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -294,7 +294,7 @@ def test_plan_dev_start_date(runner, tmp_path):
         input="\ny\n",
     )
     assert_plan_success(result, "dev")
-    assert "sqlmesh_example__dev.full_model: 2023-01-01" in result.output
+    assert "sqlmesh_example__dev.full_model: full refresh" in result.output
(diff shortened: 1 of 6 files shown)


tests/cli/test_cli.py
```diff
@@ -294,7 +294,7 @@ def test_plan_dev_start_date(runner, tmp_path):
         input="\ny\n",
     )
     assert_plan_success(result, "dev")
-    assert "sqlmesh_example__dev.full_model: 2023-01-01" in result.output
+    assert "sqlmesh_example__dev.full_model" in result.output
```
erindru, line 297: This is a substring of what was previously asserted, so if there is a regression, wouldn't this test still pass? eg:

```
assert "sqlmesh_example__dev.full_model" in result.output
````
would still be true if the output contained `sqlmesh_example__dev.full_model: 2023-01-01` right?

erindru (approved):

Issue #1254, opened 2023-08-01
Add an option to `sqlmesh plan` to use rendered model queries for diffing


izeigerman:
This should help investigate issues which occur due to non-deterministic nature of some macros which result in unexpected creation of new model snapshots.

crericha:
Should we automatically diff the rendered models if there is no diff between the raw models? And say something like raw models match, showing rendered diff.

tobymao:
@VaggelisD this seems pretty easy.

we need to add an option to sqlmesh plan

in render_definition.py, we need the option to propagate to call render_query() and render_pre/post_statements

PR #3691, opened 2025-01-23
Feat: Add plan option to show rendered model diff

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/1254

The default behavior of plan diff is to compare raw models i.e the original queries & statements:

```SQL
Differences from the `dev` environment:

Models:
└── Directly Modified:
    └── default__dev.test

--- 

+++ 

@@ -4,13 +4,13 @@

 CREATE TABLE IF NOT EXISTS foo AS
 (
   SELECT
-    @OR(FALSE, TRUE)
+    @AND(TRUE, NULL)
 )
 SELECT
-  4 + 2
+  5 + 2
 CREATE TABLE IF NOT EXISTS foo2 AS
 (
   SELECT
-    @AND(TRUE, FALSE)
+    @OR(TRUE, NULL)
 )
-DROP VIEW @this_model
+DROP VIEW IF EXISTS @this_model

```

This PR introduces the `sqlmesh plan --rendered-model-diff` which instead compares rendered queries/statements:

```SQL
@@ -4,13 +4,13 @@

 CREATE TABLE IF NOT EXISTS "foo" AS
 (
   SELECT
-    FALSE OR TRUE
+    TRUE
 )
 SELECT
-  6 AS "_col_0"
+  7 AS "_col_0"
 CREATE TABLE IF NOT EXISTS "foo2" AS
 (
   SELECT
-    TRUE AND FALSE
+    TRUE
 )
-DROP VIEW "test"
+DROP VIEW IF EXISTS "test"
```

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -335,6 +335,7 @@ Options:
                                   application (prod environment only).
   --enable-preview                Enable preview for forward-only models when
                                   targeting a development environment.
+  --rendered-model-diff           Output text differences for the rendered versions of models
   -v, --verbose                   Verbose output.
   --help                          Show this message and exit.
 ```
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -335,7 +335,7 @@ Options:
                                   application (prod environment only).
   --enable-preview                Enable preview for forward-only models when
                                   targeting a development environment.
-  --rendered-model-diff           Output text differences for the rendered versions of models
+  --diff-rendered                 Output text differences for the rendered versions of models and standalone audits
   -v, --verbose                   Verbose output.
   --help                          Show this message and exit.
 ```
--- a/docs/reference/notebook.md
+++ b/docs/reference/notebook.md
@@ -103,7 +103,7 @@ options:
             [--no-auto-categorization] [--include-unmodified]
             [--select-model [SELECT_MODEL ...]]
             [--backfill-model [BACKFILL_MODEL ...]] [--no-diff] [--run]
-            [environment]
+            [environment] [--rendered-model-diff]
 
 Goes through a set of prompts to both establish a plan and apply it
 
@@ -153,6 +153,8 @@ options:
   --no-diff             Hide text differences for changed models.
   --run                 Run latest intervals as part of the plan application
                         (prod environment only).
+  --rendered-model-diff Output text differences for the rendered versions of models
+
 ```
 
 #### run_dag
--- a/docs/reference/notebook.md
+++ b/docs/reference/notebook.md
@@ -103,7 +103,7 @@ options:
             [--no-auto-categorization] [--include-unmodified]
             [--select-model [SELECT_MODEL ...]]
             [--backfill-model [BACKFILL_MODEL ...]] [--no-diff] [--run]
-            [environment] [--rendered-model-diff]
+            [environment] [--diff-rendered]
 
 Goes through a set of prompts to both establish a plan and apply it
 
@@ -153,7 +153,7 @@ options:
   --no-diff             Hide text differences for changed models.
   --run                 Run latest intervals as part of the plan application
                         (prod environment only).
-  --rendered-model-diff Output text differences for the rendered versions of models
+  --diff-rendered       Output text differences for the rendered versions of models and standalone audits
 
 ```
 
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -283,11 +283,12 @@ class StandaloneAudit(_Node, AuditMixin):
             self._metadata_hash = hash_data(data)
         return self._metadata_hash
 
-    def text_diff(self, other: Node, rendered_model_diff: bool = False) -> str:
+    def text_diff(self, other: Node, rendered: bool = False) -> str:
         """Produce a text diff against another node.
 
         Args:
             other: The node to diff against.
+            rendered: Whether the diff should be between raw vs rendered nodes
 
         Returns:
             A unified text diff showing additions and deletions.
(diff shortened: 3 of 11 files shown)


docs/reference/notebook.md
```diff
@@ -153,6 +153,8 @@ options:
   --no-diff             Hide text differences for changed models.
   --run                 Run latest intervals as part of the plan application
                         (prod environment only).
+  --rendered-model-diff Output text differences for the rendered versions of models
```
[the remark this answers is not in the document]
izeigerman: Can we call the argument just `--rendered-diff` so that we can apply this beyond just models if needed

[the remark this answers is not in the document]
izeigerman: or better `--diff-rendered`

sqlmesh/core/audit/definition.py
```diff
@@ -283,7 +283,7 @@ def metadata_hash(self) -> str:
             self._metadata_hash = hash_data(data)
         return self._metadata_hash
 
-    def text_diff(self, other: Node) -> str:
+    def text_diff(self, other: Node, rendered_model_diff: bool = False) -> str:
```
izeigerman, line 286: Don't we want to apply the same approach to audits?

sqlmesh/core/context_diff.py
```diff
@@ -76,6 +76,8 @@ class ContextDiff(PydanticModel):
     """Previous requirements."""
     requirements: t.Dict[str, str] = {}
     """Python dependencies."""
+    rendered_model_diff: t.Optional[bool] = None
```
izeigerman, line 79: I think it's fine to just have it bool here

sqlmesh/core/context.py
```diff
@@ -1172,6 +1175,7 @@ def plan_builder(
         categorizer_config: t.Optional[CategorizerConfig] = None,
         enable_preview: t.Optional[bool] = None,
         run: bool = False,
+        rendered_model_diff: t.Optional[bool] = None,
```
izeigerman, line 1178: Since we currently don't expose this in config having this to be just bool with default `False` is fine.

sqlmesh/core/model/definition.py
```diff
@@ -187,7 +187,10 @@ def render(
         )
 
     def render_definition(
-        self, include_python: bool = True, include_defaults: bool = False
+        self,
+        include_python: bool = True,
+        include_defaults: bool = False,
+        render_model: bool = False,
```
izeigerman, line 193: `render_query`

sqlmesh/core/model/definition.py
```diff
@@ -1217,15 +1224,26 @@ def render_query(
...
             include_python=include_python, include_defaults=include_defaults
         )
-        result.extend(self.pre_statements)
-        result.append(self.query)
-        result.extend(self.post_statements)
-        result.extend(self.on_virtual_update)
+
+        if render_model:
+            result.extend(self.render_pre_statements())
+            result.append(self.render_query() or self.query)
```
izeigerman, line 1238: Should we fail if we can't render the query?

izeigerman: Probably not

sqlmesh/core/model/definition.py
```diff
@@ -674,11 +677,12 @@ def ctas_query(self, **render_kwarg: t.Any) -> exp.Query:
             )
         return query
 
-    def text_diff(self, other: Node) -> str:
+    def text_diff(self, other: Node, rendered_model_diff: bool = False) -> str:
```
izeigerman, line 680: `rendered_model_diff` -> `diff_rendered` or just `rendered`

sqlmesh/core/audit/definition.py
VaggelisD: Wasn't sure if it's applicable but can propagate it there too, the main target were SQL models as far as I understood

sqlmesh/core/model/definition.py
VaggelisD: I thought `render_model` was more fitting because we'll also render it's statements along with the query, but don't have strong feelings about it

sqlmesh/core/model/definition.py
VaggelisD: Yeah I also thought we shouldn't fail here, it's the same approach we also take elsewhere e.g on `data_hash_values`

sqlmesh/core/audit/definition.py
VaggelisD: Added it on audits too + a relevant test

sqlmesh/core/model/definition.py
georgesittas: ^ makes sense to me.

VaggelisD: Pinging @izeigerman because this was resolved before, any preference between the two?

docs/reference/cli.md
```diff
@@ -335,6 +335,7 @@ Options:
                                   application (prod environment only).
   --enable-preview                Enable preview for forward-only models when
                                   targeting a development environment.
+  --rendered-model-diff           Output text differences for the rendered versions of models
```
georgesittas, line 338: I think these docs are outdated? Isn't the option `diff-rendered` after the recent renaming? Standalone audits can be diffed as well.

VaggelisD: Good catch, updated the docs, open to suggestions for the wording

sqlmesh/core/model/definition.py
```diff
@@ -674,11 +677,12 @@ def ctas_query(self, **render_kwarg: t.Any) -> exp.Query:
             )
         return query
 
-    def text_diff(self, other: Node) -> str:
+    def text_diff(self, other: Node, rendered: bool = False) -> str:
         """Produce a text diff against another node.
 
         Args:
             other: The node to diff against.
+            diff_rendered: Whether the diff should compare raw vs rendered models
```
georgesittas, line 685:
```suggestion
            rendered: Whether the diff should compare raw vs rendered models
```

sqlmesh/core/model/definition.py
VaggelisD: Sounds good, thanks!

izeigerman (approved):

Issue #3599, opened 2025-01-08
Support all top-level properties in `model_defaults`


eakmanrq:
Currently some top level properties are supported but it would be ideal to support all of them by default. One thing to consider when adding this is that users need a way to "unset" a default value. So for example set it to "None" and disable the `model_defaults` value.

PR #3696, opened 2025-01-23
Feat: Support top level properties in model_defaults

themisvaltinos:
This update extends the support for top-level properties to be enabled project wide, fixes: #3599 

- when both project-wide and model-specific properties are defined, they are combined. 
- if a property is set at the model level, it takes precedence over the project-wide setting. 
- if a user wants to unset a project-wide property on a specific model, they can do so by explicitly setting it to None.

Don't know if it makes sense to include more, as the others seem quite model-specific, but I’m happy to add them if anyone feels they’re necessary.

### DIFF
--- a/docs/reference/model_configuration.md
+++ b/docs/reference/model_configuration.md
@@ -52,10 +52,15 @@ The SQLMesh project-level `model_defaults` key supports the following options, d
 - cron
 - owner
 - start
+- table_format
 - storage_format
+- physical_properties
+- virtual_properties
 - session_properties (on per key basis)
 - on_destructive_change (described [below](#incremental-models))
 - audits (described [here](../concepts/audits.md#generic-audits))
+- optimize_query
+- validate_query
 
 
 ### Model Naming
--- a/docs/reference/model_configuration.md
+++ b/docs/reference/model_configuration.md
@@ -45,6 +45,69 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
 
 The SQLMesh project-level configuration must contain the `model_defaults` key and must specify a value for its `dialect` key. Other values are set automatically unless explicitly overridden in the model definition. Learn more about project-level configuration in the [configuration guide](../guides/configuration.md).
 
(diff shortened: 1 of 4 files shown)


docs/reference/model_configuration.md
```diff
@@ -40,6 +40,7 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
...
 | `optimize_query`             | Whether the model's query should be optimized. This attribute is `true` by default. Setting it to `false` causes SQLMesh to disable query canonicalization & simplification. This should be turned off only if the optimized query leads to errors such as surpassing text limit.                                                                                                                                                                                                      |       bool        |    N     |
 | `validate_query`             | `validate_query`             | Whether the model's query will be strictly validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings. This will display invalid columns in your SQL statements along with models containing `SELECT *` that cannot be automatically expanded to list out all columns.                                                                                                                                                                                              |       bool        |    N     |
+
 ### Model defaults
```
erindru, line 44: Might want to add a note somewhere in the docs about how to _un_-set a default at the model level.

It looks like this is supported by defining the property as `None` at the model level

themisvaltinos: Yes good point, added an explanation and an example to make it clear how to unset a property

sqlmesh/core/model/definition.py
```diff
@@ -2270,25 +2269,27 @@ def _split_sql_model_statements(
...
     else:
-        session_properties = {}
+        properties = {}
 
     for k, v in (default or {}).items():
-        if k not in session_properties:
-            session_properties[k] = exp.Literal.string(k).eq(v)
+        if k not in properties:
+            properties[k] = exp.Literal.string(k).eq(v)
+        elif properties[k].expression.sql().lower() in {"none", "null"}:
```
georgesittas, line 2288: Is matching `null` here safe? Can that be an actual property value?

georgesittas (commented):
LGTM, just one comment.

sqlmesh/core/model/definition.py
themisvaltinos: This is to match for python models properties where `None` will be `null` here

georgesittas (approved):

PR #3706, opened 2025-01-24
Feat: improve CLI node evaluation error display

treysp:
CLI currently prints duplicate/unnecessary error information when a a node evaluation error occurs.

This PR removes the duplication, the unnecessary information, and improves error message formatting.

**Background**

We currently use multiple approaches to print errors to CLI:
1. `console.log_error()` is a method we defined to print red text to CLI
2. `logger.error()` is a standard method that prints to both CLI output and log
3. Standard Python error printing with `raise` and/or use of `Click.ClickException()`

Calls to these are spread out, can be duplicative, and do not format the displayed error.

**Changes**

This PR resolves this by separating these three actions:
1. Print only formatted errors to CLI: 
    - All formatting in console methods `log_failed_models()` and `log_skipped_models()`
2. Print full error traceback to logger: `logger.info()` so it doesn't also print to CLI
3. Creates new error class `PythonModelEvalError` to differentiate from SQL eval errors


sqlmesh/core/scheduler.py
```diff
@@ -340,20 +340,34 @@ def run(
...
-            if isinstance(error.__cause__, CircuitBreakerError):
-                raise error.__cause__
-            sid = error.node[0]
-            formatted_exception = "".join(format_exception(error.__cause__ or error))
-            log_message = f"FAILED processing snapshot {sid}\n{formatted_exception}"
-            self.console.log_error(log_message)
-            # Log with INFO level to prevent duplicate messages in the console.
-            logger.info(log_message)
+        if skipped_snapshots:
+            skipped_message = "\n[dark_orange3]Skipped models[/dark_orange3]\n\n"
```
izeigerman, line 344: All the formatting must happen within the `console` instance. Please note that we don't always have rich available.

sqlmesh/core/scheduler.py
```diff
@@ -340,20 +340,34 @@ def run(
...
+            for skipped in skipped_snapshots:
+                skipped_name = skipped
+                for delim in ["'", '"', "[", "]", "`"]:
+                    skipped_name = skipped_name.replace(delim, "")
+                skipped_message += f"  {skipped_name}\n"
+
+            self.console.log_status_update(skipped_message + "\n")
```
izeigerman, line 351: We might need a special method to report skipped and a different method to report failed

sqlmesh/core/scheduler.py
```diff
@@ -340,20 +340,34 @@ def run(
...
+                err_msg.append(str(error))
+
+                if error.__cause__:
+                    cause_msg = str(error.__cause__).replace("\n", "\n  ")
+                    err_msg.append("  " + cause_msg)
+
+                full_exception_msg = "\n".join(format_exception(error.full_exception))
+                logger.info(f"EXECUTION ERROR\n{full_exception_msg}\n")
+
+            self.console.log_status_update("[red]Failed models[/red]\n")
```
izeigerman, line 369: Ditto

sqlmesh/utils/concurrency.py
```diff
@@ -13,9 +13,30 @@
...
+        error_msg = str(ex)
+        if not isinstance(ex, PythonModelEvalError):
+            error_class = str(ex.__class__).replace("<class '", "").replace("'>", "")
+            error_msg = error_msg.replace("\n", "\n      ")
+            error_msg = f"  {error_class}:\n    {error_msg}"
+
+        for delim in ["'", '"', "[", "]", "`"]:
+            node_name = node_name.replace(delim, "")
+
+        node_name = f"[red]{node_name}[/red]\n\n  "
```
izeigerman, line 36: This is rich specific formatting and should not be here

sqlmesh/utils/concurrency.py
```diff
@@ -226,11 +246,10 @@ def sequential_apply_to_dag(
         try:
             fn(node)
         except Exception as ex:
-            if raise_on_error:
-                raise NodeExecutionFailedError(node) from ex
+            error = NodeExecutionFailedError(ex, node)
```
izeigerman, line 249: Why can't we rely on `__cause__`?

sqlmesh/core/plan/evaluator.py
```diff
@@ -203,7 +204,8 @@ def _backfill(
             interval_end_per_model=plan.interval_end_per_model,
         )
         if completion_status.is_failure:
-            raise PlanError("Plan application failed.")
+            self.console.log_status_update("[red]Error: Plan application failed.[/red]")
```
izeigerman, line 207: Ditto. Rich specific formatting should be contained to the corresponding console implementation

sqlmesh/core/plan/evaluator.py
```diff
@@ -238,8 +241,21 @@ def _push(
...
             )
             completed = True
+        except NodeExecutionFailedError as ex:
+            self.console.stop_creation_progress(success=False)
+            progress_stopped = True
+
+            full_exception_msg = "\n".join(format_exception(ex.full_exception))
+            logger.info(f"EXECUTION ERROR\n{full_exception_msg}\n")
+
+            self.console.log_status_update("\n[red]Failed models[/red]\n")
```
izeigerman, line 251: Same here

sqlmesh/core/console.py
```diff
@@ -231,6 +231,14 @@ def show_sql(self, sql: str) -> None:
     def log_status_update(self, message: str) -> None:
         """Display general status update to the user."""
 
+    @abc.abstractmethod
+    def log_skipped_models(self, message: str) -> None:
```
izeigerman, line 235: Can we document arguments for this and `log_failed_models`

sqlmesh/core/scheduler.py
```diff
@@ -326,20 +326,31 @@ def run(
...
+            for i, error in enumerate(errors):
+                if isinstance(error.__cause__, CircuitBreakerError):
+                    raise error.__cause__
+
+                err_msg_dict[error.node_name] = str(error)
+
+                exception_msg = (
+                    "\n".join(format_exception(error.__cause__)) if error.__cause__ else str(error)
+                )
+                logger.info(f"EXECUTION ERROR\n{exception_msg}\n")
```
izeigerman, line 351: Can we log the snapshot ID before the exception like we did before

sqlmesh/utils/concurrency.py
```diff
@@ -13,9 +13,32 @@
...
-        super().__init__(f"Execution failed for node {node}")
+
+        node_name = ""
+        if isinstance(node, SnapshotId):
+            node_name = node.name
+        elif isinstance(node, tuple):
+            node_name = node[0]
+        self.node_name = node_name
+
+        # MSSQL returns a `tuple([int] error code, [bytes] error message)`.
```
izeigerman, line 26: I feel like if it's MSSQL specific, then it should be handled in the corresponding engine and not here.

sqlmesh/utils/concurrency.py
```diff
@@ -13,9 +13,32 @@
...
+        error_msg = None
+        for arg in ex.args:
+            if not error_msg and isinstance(arg, (str, bytes)):
+                error_msg = arg.decode(errors="replace") if isinstance(arg, bytes) else arg
+        error_msg = str(ex) if not error_msg else error_msg
+
+        if not isinstance(ex, PythonModelEvalError):
+            error_class = str(ex.__class__).replace("<class '", "").replace("'>", "")
+            error_msg = "  " + error_msg.replace("\n", "\n  ")
+            error_msg = f"  {error_class}:\n{error_msg}"
```
izeigerman, line 37: Shouldn't this formatting be handled in `PythonModelEvalError`?

sqlmesh/utils/concurrency.py
```diff
@@ -13,9 +13,32 @@
...
+        # TODO: verify return types of other engines
+        error_msg = None
+        for arg in ex.args:
+            if not error_msg and isinstance(arg, (str, bytes)):
+                error_msg = arg.decode(errors="replace") if isinstance(arg, bytes) else arg
+        error_msg = str(ex) if not error_msg else error_msg
+
+        if not isinstance(ex, PythonModelEvalError):
+            error_class = str(ex.__class__).replace("<class '", "").replace("'>", "")
+            error_msg = "  " + error_msg.replace("\n", "\n  ")
```
izeigerman, line 36: What this indentation for? If it's to format nicely in the console then this logic should be inside the console.

sqlmesh/core/context.py
```diff
@@ -715,7 +715,7 @@ def _has_environment_changed() -> bool:
                 self.notification_target_manager.notify(
                     NotificationEvent.RUN_FAILURE, traceback.format_exc()
                 )
-                logger.error(f"Run Failure: {traceback.format_exc()}")
+                logger.info("Run failed.\n" + traceback.format_exc())
```
izeigerman, line 718: Would anything print that the run has failed into the console?

sqlmesh/core/context.py
```diff
@@ -2037,9 +2036,10 @@ def _run_plan_tests(
                     self._test_connection_config._engine_adapter.DIALECT,
                 )
             if not result.wasSuccessful():
-                raise PlanError(
-                    "Cannot generate plan due to failing test(s). Fix test(s) and run again"
+                self.console.log_error(
+                    "Error: Cannot generate plan due to failing test(s). Fix test(s) and run again"
                 )
+                raise PlanError
```
izeigerman, line 2042: Let's not raise anything without a message

sqlmesh/core/context.py
```diff
@@ -2037,9 +2036,10 @@ def _run_plan_tests(
                     self._test_connection_config._engine_adapter.DIALECT,
                 )
             if not result.wasSuccessful():
-                raise PlanError(
-                    "Cannot generate plan due to failing test(s). Fix test(s) and run again"
+                self.console.log_error(
```
izeigerman, line 2039: Wouldn't the `PlanError` be intercepted in the cli's error handler and we end up printing the error twice? Why not just raise and let it print there?

sqlmesh/core/scheduler.py
```diff
@@ -326,20 +325,16 @@ def run(
...
         for error in errors:
             if isinstance(error.__cause__, CircuitBreakerError):
                 raise error.__cause__
-            sid = error.node[0]
-            formatted_exception = "".join(format_exception(error.__cause__ or error))
-            log_message = f"FAILED processing snapshot {sid}\n{formatted_exception}"
-            self.console.log_error(log_message)
-            # Log with INFO level to prevent duplicate messages in the console.
-            logger.info(log_message)
+            logger.info(str(error))
```
izeigerman, line 335: This won't print a stacktrace in logs, which I believe we need. You can include the exception by passing `exc_info=error`

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -468,7 +469,8 @@ def get_plan_summary(self, plan: Plan) -> str:
                 return "No changes to apply."
             return f"{difference_summary}\n{missing_dates}"
         except PlanError as e:
-            return f"Plan failed to generate. Check for pending or unresolved changes. Error: {e}"
+            err_str = f" Error: {e}" if not isinstance(e, PlanApplyError) else ""
```
izeigerman, line 472: The `Error:` prefix doesn't really look useful. The first sentence already indicates that an error occurred

sqlmesh/utils/metaprogramming.py
```diff
@@ -440,26 +439,34 @@ def print_exception(
...
         exception: The exception to print the stack trace for.
         python_env: The environment containing stringified python code.
-        out: The output stream to write to.
     """
     tb: t.List[str] = []
+    indent = ""
 
     for error_line in format_exception(exception):
-        match = re.search('File "<string>", line (.*), in (.*)', error_line)
+        traceback_match = error_line.startswith("Traceback (most recent call last):")
```
izeigerman, line 447: [Nit] We don't need to match ll of these at once since we'd only use a handful. A better approach would be:
```
traceback_match = error_line.startswith("Traceback (most recent call last):")
model_def_match = re.search('File ".*?core/model/definition.py', error_line)
if traceback_match or model_def_match:
    continue
    
error_match = re.search("^.*?Error: ", error_line)
if error_match:
    tb.append(f"{indent*2}  {error_line}")
    continue
...
```

sqlmesh/core/plan/evaluator.py
```diff
@@ -480,7 +492,10 @@ def evaluate(
                 self.dag_run_poll_interval_secs,
             )
             if not plan_application_succeeded:
-                raise PlanError("Plan application failed.")
+                msg = "Error: Plan application failed."
+                self.console.log_error(msg)
+                logger.info(msg)
+                raise PlanApplyError
```
izeigerman, line 498: [Ditto] Let's not raise exceptions without messages. Also why can't this be handled by the click's exception handler

sqlmesh/core/plan/evaluator.py
```diff
@@ -238,8 +240,18 @@ def _push(
...
             completed = True
+        except NodeExecutionFailedError as ex:
+            self.console.stop_creation_progress(success=False)
+            progress_stopped = True
+
+            logger.info(str(ex))
+            self.console.log_failed_models([ex])
+            self.console.log_error("Error: Plan application failed.")
+
+            raise PlanApplyError
```
izeigerman, line 251: ditto

sqlmesh/core/plan/evaluator.py
```diff
@@ -203,7 +203,8 @@ def _backfill(
             interval_end_per_model=plan.interval_end_per_model,
         )
         if completion_status.is_failure:
-            raise PlanError("Plan application failed.")
+            self.console.log_error("Error: Plan application failed.")
+            raise PlanApplyError
```
izeigerman, line 207: ditto

sqlmesh/cli/__init__.py
```diff
@@ -38,11 +37,11 @@ def _default_exception_handler(
...
         return func()
-    except NodeExecutionFailedError as ex:
-        cause = ex.__cause__
-        raise click.ClickException(f"Failed processing {ex.node}. {cause}")
+    except PlanApplyError:
+        exit(1)
     except (SQLMeshError, SqlglotError, ValueError) as ex:
-        raise click.ClickException(str(ex))
+        click.echo(click.style(f"\nError: {str(ex)}", fg="red"))
+        exit(1)
```
izeigerman, line 44: This means that we don't get into the `finally` block and therefore the context is not being closed gracefully. Let's make sure the `final` block executes.

treysp: Oh, I saw this in [the Python docs](https://docs.python.org/3/library/sys.html#sys.exit):

"
Since [exit()](https://docs.python.org/3/library/constants.html#exit) ultimately “only” raises an exception, it will only exit the process when called from the main thread, and the exception is not intercepted. 

Cleanup actions specified by finally clauses of [try](https://docs.python.org/3/reference/compound_stmts.html#try) statements are honored, and it is possible to intercept the exit attempt at an outer level.
"

izeigerman: TIL! Thanks for checking

sqlmesh/cli/__init__.py
```diff
@@ -53,6 +52,8 @@ def _debug_exception_handler(
 ) -> DECORATOR_RETURN_TYPE:
     try:
         return func()
+    except PlanApplyError:
+        exit()
```
izeigerman, line 56: Why are we exiting with an error code?

sqlmesh/core/plan/evaluator.py
```diff
@@ -480,7 +490,9 @@ def evaluate(
                 self.dag_run_poll_interval_secs,
             )
             if not plan_application_succeeded:
-                raise PlanError("Plan application failed.")
+                msg = "Plan application failed."
+                logger.info(msg)
+                raise PlanError(msg)
```
izeigerman, line 495: Shouldn't this be `PlanApplyError`?

sqlmesh/core/context.py
```diff
@@ -1426,23 +1425,21 @@ def apply(
...
+            tb = traceback.format_exc()
             self.notification_target_manager.notify(
                 NotificationEvent.APPLY_FAILURE,
                 environment=plan.environment_naming_info.name,
                 plan_id=plan.plan_id,
-                exc=traceback.format_exc(),
+                exc=tb,
             )
-            self.log_failure_error(method="Apply")
+            logger.info("Plan application failed.\n" + tb)
```
izeigerman, line 1435: Can't we just use native logging capabilities to log this? Like `logger.inf(..., exc_info=e)`?

sqlmesh/core/context.py
```diff
@@ -724,10 +724,9 @@ def _has_environment_changed() -> bool:
                     interrupted = True
                     done = True
             except Exception as e:
-                self.notification_target_manager.notify(
-                    NotificationEvent.RUN_FAILURE, traceback.format_exc()
-                )
-                self.log_failure_error("Run")
+                tb = traceback.format_exc()
+                self.notification_target_manager.notify(NotificationEvent.RUN_FAILURE, tb)
+                logger.info("Run failed.\n" + tb)
```
izeigerman, line 729: ditto. why don't we just use logger to format the stacktrace

sqlmesh/cli/__init__.py
```diff
@@ -40,7 +41,16 @@ def _default_exception_handler(
         return func()
     except NodeExecutionFailedError as ex:
         cause = ex.__cause__
-        raise click.ClickException(f"Failed processing {ex.node}. {cause}")
+
```
izeigerman, line 44: In which case would we even end up here now?

sqlmesh/core/console.py
```diff
@@ -2147,3 +2199,48 @@ def _format_missing_intervals(snapshot: Snapshot, missing: SnapshotIntervals) ->
...
+def _format_node_errors(errors: t.List[NodeExecutionFailedError]) -> t.Dict[str, str]:
+    """Formats a list of node execution errors for display."""
+
+    def _format_node_error(ex: NodeExecutionFailedError) -> str:
+        cause = ex.__cause__ if ex.__cause__ else ex
+
+        error_msg = str(cause)
+
+        if not isinstance(cause, (NodeExecutionFailedError, PythonModelEvalError)):
+            error_class = str(cause.__class__).replace("<class '", "").replace("'>", "")
```
izeigerman, line 2213: ```cause.__class__.__name__```

izeigerman (approved):

PR #3710, opened 2025-01-24
Fix: Include unrendered query into the metadata hash

izeigerman:


sqlmesh/core/model/definition.py
```diff
@@ -1386,6 +1386,10 @@ def _data_hash_values(self) -> t.List[str]:
         data.extend(self.jinja_macros.data_hash_values)
         return data
 
+    @property
+    def _additional_metadata(self) -> t.List[str]:
+        return [*super()._additional_metadata, gen(self.query)]
```
tobymao, line 1391: nit, but this could be double counted in 1385 in certain cases, any way we can cache that so we don't pay a perf penalty

i'm a little bit afraid this is going to kill performance

tobymao (approved):

PR #3719, opened 2025-01-27
better onboarding instructions

sungchun12:
I've consistently encountered friction installing and using the right installation for `tcloud`/`sqlmesh`. This calls it out proactively in the tcloud docs and prevents that friction in the main readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,11 +139,15 @@ Install SQLMesh through [pypi](https://pypi.org/project/sqlmesh/) by running:
 ```bash
 mkdir sqlmesh-example
 cd sqlmesh-example
-python -m venv .env
-source .env/bin/activate
+python -m venv .venv
+source .venv/bin/activate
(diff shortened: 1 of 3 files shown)


treysp (approved):

README.md
```diff
@@ -139,12 +139,16 @@ Install SQLMesh through [pypi](https://pypi.org/project/sqlmesh/) by running:
 ```bash
 mkdir sqlmesh-example
 cd sqlmesh-example
-python -m venv .env
-source .env/bin/activate
+python -m venv .venv
```
tobymao, line 142: should we change this to use UV to match cloud?

sungchun12: Literally clicked merge when you commented. I recommend no given `uv` is still a new player and may be a intimidating to first-timers on our page. I'm okay with it being in tcloud docs as that requires a live explainer from someone.

PR #3723, opened 2025-01-27
Feat(trino): Add the ability to control schema locations

erindru:
Currently, there is no way to instruct SQLMesh to create schemas with a specific `LOCATION` property set for Trino.

This PR introduces a **connection** property called `schema_location_mapping` that allows a user to use regular expressions to map schemas to their corresponding physical locations.

- It's a connection property because it's specific to the Trino adapter
- It follows the convention of our other mapping properties (`physical_schema_mapping` and `environment_catalog_mapping`)
- It takes catalog into account if one exists, because a schema is really a `catalog.schema` combination most of the time

### DIFF
--- a/docs/integrations/engines/trino.md
+++ b/docs/integrations/engines/trino.md
@@ -95,6 +95,112 @@ hive.metastore.glue.default-warehouse-dir=s3://my-bucket/
 | `retries`            | Number of retries to attempt when a request fails. Default: `3`                                                                                                           |  int   |    N     |
 | `timezone`           | Timezone to use for the connection. Default: client-side local timezone                                                                                                   | string |    N     |
 
+## Table and Schema locations
+
+When using connectors that are decoupled from their storage (such as the Iceberg, Hive or Delta connectors), when creating new tables Trino needs to know the location in the physical storage it should write the table data to.
+
+This location gets stored against the table in the metastore so that any engine trying to read the data knows where to look.
+
+### Default behaviour
+
+Trino allows you to optionally configure a `default-warehouse-dir` property at the [Metastore](https://trino.io/docs/current/object-storage/metastores.html) level. When creating objects, Trino will infer schema locations to be `<default warehouse dir>/<schema name>` and table locations to be `<default warehouse dir>/<schema name>/<table name>`.
+
+However, if you dont set this property, Trino can still infer table locations if a *schema* location is explicitly set.
+
+For example, if you specify the `LOCATION` property when creating a schema like so:
+
+```sql
+CREATE SCHEMA staging_data
+WITH (LOCATION = 's3://warehouse/production/staging_data')
+```
+
+Then any tables created under that schema will have their location inferred as `<schema location>/<table name>`.
+
+If you specify neither a `default-warehouse-dir` in the metastore config nor a schema location when creating the schema, you must specify an explicit table location when creating the table or Trino will produce an error.
+
+Creating a table in a specific location is very similar to creating a schema in a specific location:
+
+```sql
+CREATE TABLE staging_data.customers (customer_id INT)
+WITH (LOCATION = 's3://warehouse/production/staging_data/customers')
+```
+
+### Configuring in SQLMesh
+
+Within SQLMesh, you can configure the value to use for the `LOCATION` property when SQLMesh creates tables and schemas. This overrides what Trino would have inferred based on the cluster configuration.
+
+#### Schemas
+
+To configure the `LOCATION` property that SQLMesh will specify when issuing `CREATE SCHEMA` statements, you can use the `schema_location_mapping` connection property. This applies to all schemas that SQLMesh creates, including its internal ones.
+
+The simplest example is to emulate a `default-warehouse-dir`:
+
+```yaml title="config.yaml"
+gateways:
+  trino:
+    connection:
+      type: trino
+      ...
+      schema_location_mapping:
+        '.*': 's3://warehouse/production/@{schema_name}'
+```
+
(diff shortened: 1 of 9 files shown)


docs/integrations/engines/trino.md
```diff
@@ -95,6 +95,112 @@ hive.metastore.glue.default-warehouse-dir=s3://my-bucket/
...
+
+The simplest example is to emulate a `default-warehouse-dir`:
+
+```yaml title="config.yaml"
+gateways:
+  trino:
+    connection:
+      type: trino
+      ...
+      schema_location_mapping:
```
erindru, line 144: For better UX in a common case, I also considered adding a `warehouse_location` config property (like `s3_warehouse_location` in the [Athena](https://sqlmesh.readthedocs.io/en/stable/integrations/engines/athena/#s3-locations) adapter that in the background would be translated to a `schema_location_mapping` similar to this.

The benefit would be it would free up the user from having to think about regex

docs/integrations/engines/trino.md
```diff
@@ -95,6 +95,112 @@ hive.metastore.glue.default-warehouse-dir=s3://my-bucket/
...
+Often, you dont need to configure an explicit table location because if you have configured explicit schema locations, table locations are automatically inferred by Trino to be a subdirectory under the schema location.
+
+However, if you need to, you can configure an explicit table location by adding a `location` property to the model `physical_properties`:
+
+```
+MODEL (
+  name staging.customers,
+  kind FULL,
+  physical_properties (
+    location = 's3://warehouse/staging/customers'
```
erindru, line 195: I've just realised that this example is a bad one because the table name eg `customers` here actually needs to be dynamic and substituted with the actual snapshot table name.

The way this is set up, all snapshots for this model will be written to the same location and clobber each other

sqlmesh/core/config/root.py
```diff
@@ -158,13 +158,7 @@ def _gateways_ensure_dict(cls, value: t.Dict[str, t.Any]) -> t.Dict[str, t.Any]:
     def _validate_regex_keys(
         cls, value: t.Dict[str | re.Pattern, t.Any]
     ) -> t.Dict[re.Pattern, t.Any]:
-        compiled_regexes = {}
```
erindru, line 161: I broke this out into a utility function so I could reference it here and in the Trino connection config

docs/integrations/engines/trino.md
erindru: Based on internal discussion, using the `@this_model` macro here is non trivial. I'll need to address this in a follow-up PR

docs/integrations/engines/trino.md
```diff
@@ -95,6 +95,112 @@ hive.metastore.glue.default-warehouse-dir=s3://my-bucket/
...
+- a schema called `utils` would be mapped to the location `s3://utils-bucket/utils` because it directly matches the `^utils$` pattern
+- a schema called `transactions` in a catalog called `landing` would be mapped to the location `s3://raw-data/landing/transactions` because the string `landing.transactions` matches the `^landing\..*$` pattern
+- schemas called `staging_customers` and `staging_accounts` would be mapped to the locations `s3://bucket/staging_customers_dev` and `s3://bucket/staging_accounts_dev` respectively because they match the `^staging.*$` pattern
+- a schema called `accounts` in a catalog called `staging` would be mapped to the location `s3://bucket/accounts_dev` because the string `staging.accounts` matches the `^staging.*$` pattern
+- schemas called `sqlmesh__staging_customers` and `sqlmesh__staging_utils` would be mapped to the locations `s3://sqlmesh-internal/dev/sqlmesh__staging_customers` and `s3://sqlmesh-internal/dev/sqlmesh__staging_utils` respectively because they match the `^sqlmesh.*$` pattern
+
+!!! info "Placeholders"
+    You may use the `@catalog_name` and `@schema_name` placeholders in the mapping value.
```
erindru, line 179: I decided to use this `@variable` syntax to be consistent with our macro syntax.

However, it may be confusing because this isn't an actual macro evaluation context, you cant use `@gateway` or call your own macros here. It just does `str.replace()` which I figured was fine for an initial implementation

docs/integrations/engines/trino.md
```diff
@@ -95,6 +95,112 @@ hive.metastore.glue.default-warehouse-dir=s3://my-bucket/
...
+
+```yaml title="config.yaml"
+gateways:
+  trino:
+    connection:
+      type: trino
+      ...
+      schema_location_mapping:
+        '^utils$': 's3://utils-bucket/@schema_name'
+        '^landing\..*$': 's3://raw-data/@catalog_name/@schema_name'
```
izeigerman, line 164: What if I want to add a prefix / postfix to `@schema_name`, eg. `@{schema_name}_bla`? What would be the syntax then?

erindru: That's covered on the next line:
```
'^staging.*$': 's3://bucket/@schema_name_dev'
```

It works because this uses basic string replace currently, like so:

```
>>> 's3://bucket/@schema_name_dev'.replace("@schema_name", "staging")
's3://bucket/staging_dev'
```

izeigerman: I'd go with a more explicit pattern like `@{schema_name}` which is also consistent with how we interpolate macros. But up to you

izeigerman (approved):

docs/integrations/engines/trino.md
erindru: That's fair and would also make it forwards-compatible if we ever decide to upgrade from string replacement to running these through the macro evaluator.

I've updated the syntax to mimic our macro explicit variable interpolation syntax instead

PR #3730, opened 2025-01-28
Chore: lift pydantic <2.6.0 dev version pin

georgesittas:
If we hadn't constrained the pydantic version, then the problem solved by [this PR](https://github.com/TobikoData/sqlmesh/pull/3698) would've been caught in CI: `test_json_serde` already detects it for ≥v2.8.0.

I intentionally re-introduced the initialization of `info` to `None` in the first commit of this PR, in order to demonstrate how the test [fails](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/17857/workflows/7ce6c86f-6d28-4d28-8d5d-c81a3dba0a6f/jobs/159989). The second commit reverts that change.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -281,10 +281,10 @@ workflows:
       - airflow_docker_tests:
           requires:
             - style_and_slow_tests
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
+          #   branches:
+          #     only:
+          #       - main
       - engine_tests_docker:
           name: engine_<< matrix.engine >>
           matrix:
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -281,10 +281,10 @@ workflows:
       - airflow_docker_tests:
           requires:
             - style_and_slow_tests
-          # filters:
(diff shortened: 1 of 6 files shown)


georgesittas:
@eakmanrq can you take a look please? It looks like this was introduced some time ago as a package constraint, and then it was moved under the `"dev"` dependencies list [here](https://github.com/TobikoData/sqlmesh/commit/82135eb7df589c8faf165b2a1ac0a3468c2f2c44). I ran `make airflow-docker-test-with-env` locally in this branch and it looks like it succeeded; do you recall if there was something else that caused problems with this?

eakmanrq:
>The airflow tests [ran for this PR](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/17863/workflows/92c18cb8-5788-4a44-98bc-ebaacb8bac5a/jobs/160082) and it looks like they all succeeded; do you recall if there was something else that caused problems with this?

Based on my comment it looks like it was just left due to Airflow not supporting 2.6+. So if the tests pass now then we should be good to go.

eakmanrq (approved):

Issue #3459, opened 2024-12-03
`sqlmesh init <database>` should dynamically generate a specific template config


sungchun12:
When I initialize a new sqlmesh project, I want to see template configs beyond the standard duckdb one. It should be specific to the database specified and should be commented out yaml lines. It makes it easier as I have one less browser tab to open in docs to update the config as needed. 

[Internal slack thread](https://tobikodata.slack.com/archives/C04DTU8RF4J/p1733159755768409)

```shell
sqlmesh init bigquery
bat config.yaml 
───────┬───────────────────────────────────────────────────────────────────────────
       │ File: config.yaml
───────┼───────────────────────────────────────────────────────────────────────────
   1   │ gateways:
   2   │   local:
   3   │     connection:
   4   │       type: duckdb
   5   │       database: db.db
   6   │ 
   7   │ default_gateway: local
   8   │ 
   9   │ model_defaults:
  10   │   dialect: bigquery
  11   │   start: 2024-12-01
```

Future Output

```yaml
  gateways:
      local:
          type: bigquery
          method: service-account-json
          concurrent_tasks: 5
          register_comments: true
          keyfile_json: {{ env_var('GOOGLE_SQLMESH_CREDENTIALS') }} # it'll be nice to even include env var placeholders as I've seen users accidentally include plain text passwords in their configs
          project: your-project

  default_gateway: local

  model_defaults:
      dialect: bigquery

```

PR #3733, opened 2025-01-28
Feat: Generate per-dialect options on init project

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3459

Dynamically generate the engine-appropriate connection options through Pydantic's `model_fields`. The following conventions are applied:
- If the `Field` is not required, it's commented out 
- If the `Field` has a (literal) default value it's appended to the RHS

### DIFF
--- a/sqlmesh/cli/example_project.py
+++ b/sqlmesh/cli/example_project.py
@@ -4,9 +4,38 @@ from pathlib import Path
 from dataclasses import dataclass
 
 import click
+
 from sqlglot import Dialect
 from sqlmesh.integrations.dlt import generate_dlt_models_and_settings
 from sqlmesh.utils.date import yesterday_ds
+
+from sqlmesh.core.config.connection import (
+    AthenaConnectionConfig,
+    BigQueryConnectionConfig,
+    ClickhouseConnectionConfig,
+    ConnectionConfig,
+    DatabricksConnectionConfig,
+    DuckDBConnectionConfig,
+    MSSQLConnectionConfig,
+    PostgresConnectionConfig,
+    RedshiftConnectionConfig,
+    SnowflakeConnectionConfig,
+    TrinoConnectionConfig,
+)
+
+DIALECT_TO_CONFIG_CLASS: t.Dict[str, t.Type[ConnectionConfig]] = {
+    "databricks": DatabricksConnectionConfig,
+    "duckdb": DuckDBConnectionConfig,
+    "postgres": PostgresConnectionConfig,
+    "redshift": RedshiftConnectionConfig,
+    "snowflake": SnowflakeConnectionConfig,
+    "bigquery": BigQueryConnectionConfig,
+    "sqlserver": MSSQLConnectionConfig,
+    "tsql": MSSQLConnectionConfig,
+    "trino": TrinoConnectionConfig,
+    "athena": AthenaConnectionConfig,
+    "clickhouse": ClickhouseConnectionConfig,
+}
 
 
 class ProjectTemplate(Enum):
@@ -23,11 +52,37 @@ def _gen_config(
     start: t.Optional[str],
     template: ProjectTemplate,
 ) -> str:
-    connection_settings = (
-        settings
-        or """      type: duckdb
+    if not settings:
+        if dialect in DIALECT_TO_CONFIG_CLASS:
+            required_fields = []
+            non_required_fields = []
+
+            for name, field in DIALECT_TO_CONFIG_CLASS[dialect].model_fields.items():
+                field_name = field.alias or name
+                default_value = field.get_default()
+
+                default_value = (
+                    default_value if isinstance(default_value, (str, int, bool)) else None
+                )
+
+                required = field.is_required() or field_name == "type"
+                option_str = (
+                    f"      {'# ' if not required else ''}{field_name}: {default_value or ''}\n"
+                )
+
+                if required:
+                    required_fields.append(option_str)
+                else:
+                    non_required_fields.append(option_str)
+
+            connection_settings = "".join(required_fields + non_required_fields)
+
+    else:
+        connection_settings = (
+            settings
+            or """      type: duckdb
       database: db.db"""
-    )
+        )
 
     default_configs = {
         ProjectTemplate.DEFAULT: f"""gateways:
@@ -44,7 +99,7 @@ model_defaults:
         ProjectTemplate.AIRFLOW: f"""gateways:
   local:
     connection:
-      {connection_settings}
+{connection_settings}
 
 default_gateway: local
 
--- a/sqlmesh/cli/example_project.py
+++ b/sqlmesh/cli/example_project.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from dataclasses import dataclass
 
 import click
-
 from sqlglot import Dialect
 from sqlmesh.integrations.dlt import generate_dlt_models_and_settings
 from sqlmesh.utils.date import yesterday_ds
@@ -27,6 +26,9 @@ def _gen_config(
     template: ProjectTemplate,
 ) -> str:
     if not settings:
+        connection_settings = """      type: duckdb
+      database: db.db"""
+
         if dialect in CONNECTION_CONFIG_TO_TYPE:
             required_fields = []
             non_required_fields = []
@@ -35,9 +37,10 @@ def _gen_config(
                 field_name = field.alias or name
                 default_value = field.get_default()
 
-                default_value = (
-                    default_value if isinstance(default_value, (str, int, bool)) else None
-                )
+                if isinstance(default_value, Enum):
+                    default_value = default_value.value
+                elif not isinstance(default_value, (str, int, bool)):
+                    default_value = None
 
                 required = field.is_required() or field_name == "type"
                 option_str = (
@@ -52,30 +55,26 @@ def _gen_config(
             connection_settings = "".join(required_fields + non_required_fields)
 
     else:
-        connection_settings = (
-            settings
-            or """      type: duckdb
-      database: db.db"""
-        )
+        connection_settings = settings
 
     default_configs = {
         ProjectTemplate.DEFAULT: f"""gateways:
-  local:
+  dev:
     connection:
 {connection_settings}
 
-default_gateway: local
+default_gateway: dev
 
 model_defaults:
   dialect: {dialect}
   start: {start or yesterday_ds()}
 """,
         ProjectTemplate.AIRFLOW: f"""gateways:
-  local:
+  dev:
     connection:
 {connection_settings}
 
-default_gateway: local
+default_gateway: dev
 
 default_scheduler:
   type: airflow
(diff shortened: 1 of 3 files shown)


sqlmesh/cli/example_project.py
```diff
@@ -4,10 +4,39 @@
...
+    DatabricksConnectionConfig,
+    DuckDBConnectionConfig,
+    MSSQLConnectionConfig,
+    PostgresConnectionConfig,
+    RedshiftConnectionConfig,
+    SnowflakeConnectionConfig,
+    TrinoConnectionConfig,
+)
+
+DIALECT_TO_CONFIG_CLASS: t.Dict[str, t.Type[ConnectionConfig]] = {
```
georgesittas, line 26: This mapping seems unnecessary, it's already defined in `connection.py`:

```python
> /<path>/sqlmesh/core/config/connection.py(1748)parse_connection_config()
-> return CONNECTION_CONFIG_TO_TYPE[connection_type](**v)
(Pdb) CONNECTION_CONFIG_TO_TYPE
{'athena': <class 'sqlmesh.core.config.connection.AthenaConnectionConfig'>, 'azuresql': <class 'sqlmesh.core.config.connection.AzureSQLConnectionConfig'>, 'bigquery': <class 'sqlmesh.core.config.connection.BigQueryConnectionConfig'>, 'clickhouse': <class 'sqlmesh.core.config.connection.ClickhouseConnectionConfig'>, 'databricks': <class 'sqlmesh.core.config.connection.DatabricksConnectionConfig'>, 'duckdb': <class 'sqlmesh.core.config.connection.DuckDBConnectionConfig'>, 'gcp_postgres': <class 'sqlmesh.core.config.connection.GCPPostgresConnectionConfig'>, 'mssql': <class 'sqlmesh.core.config.connection.MSSQLConnectionConfig'>, 'motherduck': <class 'sqlmesh.core.config.connection.MotherDuckConnectionConfig'>, 'mysql': <class 'sqlmesh.core.config.connection.MySQLConnectionConfig'>, 'postgres': <class 'sqlmesh.core.config.connection.PostgresConnectionConfig'>, 'redshift': <class 'sqlmesh.core.config.connection.RedshiftConnectionConfig'>, 'snowflake': <class 'sqlmesh.core.config.connection.SnowflakeConnectionConfig'>, 'spark': <class 'sqlmesh.core.config.connection.SparkConnectionConfig'>, 'trino': <class 'sqlmesh.core.config.connection.TrinoConnectionConfig'>}
```

sqlmesh/cli/example_project.py
```diff
@@ -44,7 +99,7 @@ def _gen_config(
         ProjectTemplate.AIRFLOW: f"""gateways:
   local:
     connection:
-      {connection_settings}
+{connection_settings}
```
georgesittas, line 102: [Nit] should we default to a name other than `local`? It's a misnomer when it comes to cloud engines, so perhaps a more neutral noun will be preferable here.

Don't feel strongly in any case, just thought I'd point it out. I'd see what others think as well before changing.

tests/cli/test_cli.py
```diff
@@ -944,3 +944,25 @@ def test_plan_dlt(runner, tmp_path):
...
+def test_init_project_dialects(runner, tmp_path):
+    init_example_project(tmp_path, "redshift")
+    with open(tmp_path / "config.yaml") as file:
+        redshift_config = file.read()
+
+        assert (
+            redshift_config
+            == "gateways:\n  local:\n    connection:\n      type: redshift\n      # concurrent_tasks: 4\n      # register_comments: True\n      # pre_ping: \n      # pretty_sql: \n      # user: \n      # password: \n      # database: \n      # host: \n      # port: \n      # source_address: \n      # unix_sock: \n      # ssl: \n      # sslmode: \n      # timeout: \n      # tcp_keepalive: \n      # application_name: \n      # preferred_role: \n      # principal_arn: \n      # credentials_provider: \n      # region: \n      # cluster_identifier: \n      # iam: \n      # is_serverless: \n      # serverless_acct_id: \n      # serverless_work_group: \n\n\ndefault_gateway: local\n\nmodel_defaults:\n  dialect: redshift\n  start: 2025-01-27\n"
+        )
+        remove(tmp_path / "config.yaml")
```
georgesittas, line 958: Can we rewrite these tests so that they're properly formatted? This format is not easily readable. Consider instead:

```python
assert redshift_config == (
    "gateways:\n"
    "  local:\n"
    "    connection:\n"
    ...
)
```

Can also just do the typical triple-quoted version as an alternative, I just have a slight personal preference for this the implicit string concatenation syntax sometimes.

sqlmesh/cli/example_project.py
```diff
@@ -23,11 +52,37 @@ def _gen_config(
     start: t.Optional[str],
     template: ProjectTemplate,
 ) -> str:
-    connection_settings = (
-        settings
-        or """      type: duckdb
+    if not settings:
+        if dialect in DIALECT_TO_CONFIG_CLASS:
```
georgesittas, line 56: How can `dialect` ever not be in `DIALECT_TO_CONFIG_CLASS`? I think `get_or_raise` should've already validated this:

```python
def _create_config(...) -> None:
    if dialect:
        Dialect.get_or_raise(dialect)
```

sqlmesh/cli/example_project.py
VaggelisD: Ah good catch, didn't understand that it's dark magic ends up doing that

sqlmesh/cli/example_project.py
VaggelisD: Yep I see what you mean, I'm open to suggestions

tests/cli/test_cli.py
VaggelisD: I get what you're saying, I also started off with multi-line strings but the test will become too long if I add in a couple other dialects too. If real estate is not an issue I can switch to multi-line again

sqlmesh/cli/example_project.py
VaggelisD: This isn't a dialect check per se, it's that the dialect is in the "focused" ones for this PR.

This was an idea that we had discussed privately with Sung, we wanted to concentrate on a few dialects at first (by revisiting/defining their default values etc) but if we switch to `CONNECTION_CONFIG_TO_TYPE` then we'll support all configs out of the bat.

VaggelisD: Also, `dialect` is an optional parameter for this function, should we make it default to `duckdb` so we can always hit the dictionary (unless `settings` is provided, of course)?

macros/__init__.py
georgesittas, line 1: oversight?

sqlmesh/core/config/connection.py
```diff
@@ -1734,6 +1734,8 @@ def get_catalog(self) -> t.Optional[str]:
         exclude=(ConnectionConfig, BaseDuckDBConnectionConfig),
     )
 }
+CONNECTION_CONFIG_TO_TYPE["tsql"] = CONNECTION_CONFIG_TO_TYPE["mssql"]
+CONNECTION_CONFIG_TO_TYPE["sqlserver"] = CONNECTION_CONFIG_TO_TYPE["mssql"]
```
georgesittas, line 1738: These should probably be reverted, given [this](https://github.com/tobymao/sqlglot/pull/4666) was closed.

macros/__init__.py
VaggelisD: Woops, classic local testing

sqlmesh/cli/example_project.py
sungchun12: Let's call it `dev`. Easy to understand.

sungchun12:
I tested this out locally and realized it's not helpful unless you explain in plain terms what each config means paired with an example. I recommend you update the template with this look and feel. You can leverage our existing docs to fill out the config descriptions.

This mainly needs to be done for: snowflake, databricks, bigquery, postgres, duckdb, redshift

What will also help is that most users if they're serious will want to read our docs to verify what connection config settings matter to them. Let's save them the extra step and print it to the terminal: https://sqlmesh.readthedocs.io/en/stable/integrations/overview/#execution-engines

```yaml
gateways:
  local:
    connection:
      type: bigquery
      # concurrent_tasks: 1
      # register_comments: True
      # pre_ping: 
      # pretty_sql: 
      # method: BigQueryConnectionMethod.OAUTH
      # project: 
      # execution_project: # The name of the GCP project to bill for the execution of the models. If not set, the project associated with the model will be used. type: string, ex: my-project
      # quota_project: 
      # location: 
      keyfile: # Path to the keyfile to be used with service-account method # type: string, ex: '/source/keyfile.json'
      # keyfile_json: 
      # token: 
      # refresh_token: 
      # client_id: 
      # client_secret: 
      # token_uri: 
      # scopes: 
      # job_creation_timeout_seconds: 
      # job_execution_timeout_seconds: 
      # job_retries: 1
      # job_retry_deadline_seconds: 
      # priority: 
      # maximum_bytes_billed: 
```

VaggelisD:
> Actually, as I was writing this. The "nice to have" path may be the best of both worlds where you add in the commented out configs and then dynamically print the link to the relevant engine as it's much easier to onboard in that flow vs. cluttering the yaml with elongated comments. Let me know what you think about the nice to have path being the need to have one.

This was what I was about to recommend to, I think the dynamic generation is needed in order to be up to date with the latest options and outputting the documentation as in "check here for more details" should do the trick.

sqlmesh/cli/example_project.py
```diff
@@ -23,30 +25,66 @@ def _gen_config(
...
-        settings
-        or """      type: duckdb
+    if not settings:
+        connection_settings = """      type: duckdb
       database: db.db"""
-    )
+
+        doc_link = "# Visit https://sqlmesh.readthedocs.io/en/stable/integrations/engines{dialect_link} for more information on configuring the connection to your execution engine."
+
+        dialect_alias = "mssql" if dialect == "tsql" else dialect
```
georgesittas, line 34: [Nit] Should this be `engine` instead? I don't think `mssql` refers to the _dialect_ in this case; that would be t-sql.

sqlmesh/cli/example_project.py
```diff
@@ -23,30 +25,66 @@ def _gen_config(
...
+            required_fields = []
+            non_required_fields = []
+
+            for name, field in CONNECTION_CONFIG_TO_TYPE[dialect_alias].model_fields.items():
+                field_name = field.alias or name
+                default_value = field.get_default()
+
+                if isinstance(default_value, Enum):
+                    default_value = default_value.value
+                elif not isinstance(default_value, (str, int, bool)):
```
georgesittas, line 47: Should we include `float` here? Perhaps worth considering a constant: `PRIMITIVES = (str, int, bool, float)`.

georgesittas (approved):
LGTM. I think we may need to update a few existing doc pages if this is merged, to avoid showing the outdated config template.

tobymao (approved):

PR #3737, opened 2025-01-29
fix!: allow multi repos to load macros with same module name closes #…

tobymao:
…3734


blecourt-private:
What will happen if macros with the same name exist in multiple projects? E.g. if `repo_1` and `repo_2` both define the macro `@one()`, possibly having different definitions. Will models from `repo_1` use the macro `@one()` that is defined in `repo_1/macros/` and will models from `repo_2` use the macro `@one()` that is defined in `repo_2/macros/`?

sqlmesh/core/context.py
```diff
@@ -589,8 +599,10 @@ def load(self, update_schemas: bool = True) -> GenericContext[C]:
             self.default_dialect or ""
         }
 
+        project_types = {loader.config.loader.__name__.lower() for loader in self._loaders}
+
         analytics.collector.on_project_loaded(
-            project_type=self.project_type,
+            project_type=c.HYBRID if len(project_types) > 1 else first(project_types),
```
izeigerman, line 605: I don't think this is correct. The expected values are: hybrid, dbt, and native. This will return a class name instead

sqlmesh/core/context.py
```diff
@@ -2243,9 +2255,9 @@ def _register_notification_targets(self) -> None:
 
     def _load_materializations_and_signals(self) -> None:
         if not self._loaded:
-            for context_loader in self._loaders.values():
-                with sys_path(*context_loader.configs):
-                    context_loader.loader.load_materializations(self)
+            for loader in self._loaders:
+                with sys_path(loader.context_path):
```
izeigerman, line 2259: [Ditto] Consider moving into the loader

sqlmesh/core/loader.py
```diff
@@ -54,42 +53,53 @@ class LoadedProject:
...
 
 class Loader(abc.ABC):
     """Abstract base class to load macros and models for a context"""
 
-    def __init__(self) -> None:
+    def __init__(self, context: GenericContext, path: Path) -> None:
         self._path_mtimes: t.Dict[Path, float] = {}
-        self._dag: DAG[str] = DAG()
+        self.context = context
+        self.context_path = path
```
izeigerman, line 66: IMHO `context_path` is confusing, since there's one context but multiple paths. Shouldn't this be `project_path` or `config_path`?

sqlmesh/core/context.py
```diff
@@ -2243,9 +2255,9 @@ def _register_notification_targets(self) -> None:
...
+                    loader._load_materializations(self)
```
izeigerman, line 2260: Let's not call private methods

izeigerman: Also you pass `context` as the first argument which is no longer in the function signature. I hope we have tests that cover this

sqlmesh/core/loader.py
```diff
@@ -139,22 +136,11 @@ def load(self, context: GenericContext, update_schemas: bool = True) -> LoadedPr
...
             excluded_requirements=excluded_requirements,
         )
         return project
 
-    def load_signals(self, context: GenericContext) -> UniqueKeyDict[str, signal]:
-        """Loads signals for the built-in scheduler."""
-        self._context = context
-        return self._load_signals()
-
-    def load_materializations(self, context: GenericContext) -> None:
```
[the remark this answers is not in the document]
tobymao: i made this a public method now, it wasn't doing anything except setting context. signals is unused since i made signals serializable

izeigerman (approved):

themisvaltinos:
Regarding the failing `tests/core/test_integration.py::test_multi_hybrid`, the issue arises because the Jinja macro defined in the `dbt` project is being referenced in the `sqlmesh` repo. From what I tested locally on `sqlmesh` multi-repo project as well the macros are now isolated to each repo and cannot be shared across them, so just noting this in case it isn't an intended outcome of the change

tobymao:
@themisvaltinos i'm not sure how the previous behavior ever worked, it seems problematic so i'll get rid of the test

tobymao:
@blecourt-private if two macros have the same name, it's fine. each project is loaded individually and stand alone so there is no conflict. there is a test for that.

you are not allowed to reference macros defined outside, multi-repos should be standalone. if you want to share macros, you can publish a 3rd party library or use git sub-modules to share a folder.

themisvaltinos (approved):

PR #3740, opened 2025-01-29
Feat: Allow macros in python model properties

themisvaltinos:
This update enables macros to be passed in the properties for Python models, aside from `audits`, `signals`, `merge filter` and `physical_properties` which are rendered at evaluation time (the latter will be addressed on a subsequent pr). This allows to dynamically control for example the `enabled` flag based on the gateway:

```python
@model(
  ...,
  enabled="@IF(@gateway = 'local', True, False)",
)
```

### DIFF
--- a/sqlmesh/core/constants.py
+++ b/sqlmesh/core/constants.py
@@ -97,7 +97,5 @@ RUNTIME_RENDERED_MODEL_FIELDS = {
     "description",
     "cron",
     "physical_properties",
-    "virtual_properties",
-    "session_properties",
     "merge_filter",
 }
--- a/sqlmesh/core/constants.py
+++ b/sqlmesh/core/constants.py
@@ -90,12 +90,3 @@ NATIVE = "native"
 HYBRID = "hybrid"
 
 DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
-
-RUNTIME_RENDERED_MODEL_FIELDS = {
-    "audits",
-    "signals",
-    "description",
-    "cron",
-    "physical_properties",
-    "merge_filter",
-}
--- a/sqlmesh/core/constants.py
+++ b/sqlmesh/core/constants.py
@@ -90,3 +90,13 @@ NATIVE = "native"
 HYBRID = "hybrid"
 
 DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
+
+RUNTIME_RENDERED_MODEL_FIELDS = {
+    "audits",
+    "signals",
+    "cron",
+    "physical_properties",
+    "virtual_properties",
+    "session_properties",
+    "merge_filter",
+}
--- a/sqlmesh/core/constants.py
+++ b/sqlmesh/core/constants.py
@@ -94,6 +94,7 @@ DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
 RUNTIME_RENDERED_MODEL_FIELDS = {
     "audits",
     "signals",
+    "description",
     "cron",
     "physical_properties",
     "virtual_properties",
--- a/sqlmesh/core/model/decorator.py
+++ b/sqlmesh/core/model/decorator.py
@@ -15,10 +15,10 @@ from sqlmesh.core.dialect import SQLMESH_MACRO_PREFIX, MacroFunc, parse_one
 from sqlmesh.core import dialect as d
 from sqlmesh.core.model.definition import (
     Model,
-    _meta_renderer,
     create_python_model,
     create_sql_model,
     get_model_name,
+    render_meta,
 )
 from sqlmesh.core.model.kind import ModelKindName, _ModelKind
 from sqlmesh.utils import registry_decorator
@@ -121,48 +121,53 @@ class model(registry_decorator):
         build_env(self.func, env=env, name=entrypoint, path=module_path)
 
         # Properties to be rendered at model creation time
-        expressions = [exp.Property(this="name", value=exp.to_table(self.name, dialect=dialect))]
+        expressions = [
+            exp.Property(this=exp.var("name"), value=exp.to_table(self.name, dialect=dialect))
+        ]
         for field_name in {"enabled", "start", "end"}:
-            if (
-                (field_value := self.kwargs.get(field_name))
-                and isinstance(field_value, exp.Expression)
+            if (field_value := self.kwargs.get(field_name)) and (
+                isinstance(field_value, exp.Expression)
                 or (isinstance(field_value, str) and field_value.startswith(SQLMESH_MACRO_PREFIX))
             ):
                 expressions.append(
                     exp.Property(
-                        this=field_name,
-                        value=parse_one(field_value, dialect=dialect)
-                        if isinstance(field_value, str)
-                        else field_value,
+                        this=exp.var(field_name),
+                        value=exp.maybe_parse(field_value, dialect=dialect),
                     )
                 )
 
-        renderer_kwargs = {
-            "module_path": module_path,
-            "macros": macros,
-            "jinja_macros": jinja_macros,
-            "variables": variables,
-            "path": path,
-            "dialect": dialect,
-            "default_catalog": default_catalog,
-        }
-        meta_renderer = _meta_renderer(
-            expression=d.Model(expressions=expressions),
-            **renderer_kwargs,  # type: ignore
-        )
-        if rendered := meta_renderer.render():
-            for prop in rendered[0].expressions:
-                self.kwargs[prop.this] = prop.args.get("value").sql(dialect=dialect)
+        if expressions and (
+            rendered_meta := render_meta(
+                expression=d.Model(expressions=expressions),
+                module_path=module_path,
+                macros=macros,
+                jinja_macros=jinja_macros,
+                variables=variables,
+                path=path,
+                dialect=dialect,
+                default_catalog=default_catalog,
+            )
+        ):
+            for prop in rendered_meta[0].expressions:
+                self.kwargs[prop.this.name if isinstance(prop.this, exp.Var) else prop.this] = (
+                    prop.args.get("value").sql(dialect=dialect)
+                )
 
         common_kwargs = {
             "defaults": defaults,
+            "path": path,
             "time_column_format": time_column_format,
             "python_env": serialize_env(env, path=module_path),
             "physical_schema_mapping": physical_schema_mapping,
             "project": project,
+            "default_catalog": default_catalog,
+            "variables": variables,
+            "dialect": dialect,
             "columns": self.columns if self.columns else None,
+            "module_path": module_path,
+            "macros": macros,
+            "jinja_macros": jinja_macros,
             "audit_definitions": audit_definitions,
-            **renderer_kwargs,
             **self.kwargs,
         }
 
--- a/sqlmesh/core/model/decorator.py
+++ b/sqlmesh/core/model/decorator.py
@@ -11,14 +11,13 @@ from sqlglot.dialects.dialect import DialectType
 from sqlmesh.core.macros import MacroRegistry
 from sqlmesh.utils.jinja import JinjaMacroRegistry
 from sqlmesh.core import constants as c
-from sqlmesh.core.dialect import SQLMESH_MACRO_PREFIX, MacroFunc, parse_one
-from sqlmesh.core import dialect as d
+from sqlmesh.core.dialect import MacroFunc, parse_one
 from sqlmesh.core.model.definition import (
     Model,
     create_python_model,
     create_sql_model,
     get_model_name,
-    render_meta,
+    render_meta_fields,
 )
 from sqlmesh.core.model.kind import ModelKindName, _ModelKind
 from sqlmesh.utils import registry_decorator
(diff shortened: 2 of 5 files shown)


sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
...
+                expressions.append(
+                    exp.Property(
+                        this=field_name,
+                        value=parse_one(field_value, dialect=dialect)
+                        if isinstance(field_value, str)
+                        else field_value,
+                    )
+                )
+
+        renderer_kwargs = {
```
izeigerman, line 141: How was this set compiled? Can we have this set shared somehow with the one in `load_sql_based_model`? I don't want to update one place and then forget about the other one

sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
...
+            "module_path": module_path,
+            "macros": macros,
+            "jinja_macros": jinja_macros,
+            "variables": variables,
+            "path": path,
+            "dialect": dialect,
+            "default_catalog": default_catalog,
+        }
+        meta_renderer = _meta_renderer(
```
izeigerman, line 150: Perhaps, instead of kwargs, the `_meta_renderer` function should take all the arguments above by name explicitly (no default values).

sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
 
         build_env(self.func, env=env, name=entrypoint, path=module_path)
 
+        # Properties to be rendered at model creation time
+        expressions = [exp.Property(this="name", value=exp.to_table(self.name, dialect=dialect))]
+        for field_name in {"enabled", "start", "end"}:
```
izeigerman, line 126: Why only these properties? Shouldn't we instead take all fields from the `ModelMeta` schema with few exceptions like `signals`?

sqlmesh/core/model/decorator.py
themisvaltinos: right I will see how to either share or refactor to have a function with the same logic in both `load_sql_based_model` and here

sqlmesh/core/model/decorator.py
themisvaltinos: ok I will add all fields of `ModelMeta` instead, aside from these that shouldn't be rendered at creation time. it will require specific handling for each property, but it can be done having something similar to `META_FIELD_CONVERTER` like in `render_definition`

sqlmesh/core/model/decorator.py
```diff
@@ -12,9 +12,11 @@
 from sqlmesh.core.macros import MacroRegistry
 from sqlmesh.utils.jinja import JinjaMacroRegistry
 from sqlmesh.core import constants as c
-from sqlmesh.core.dialect import MacroFunc, parse_one
+from sqlmesh.core.dialect import SQLMESH_MACRO_PREFIX, MacroFunc, parse_one
+from sqlmesh.core import dialect as d
 from sqlmesh.core.model.definition import (
     Model,
+    _meta_renderer,
```
georgesittas, line 19: Should we rename this to get rid of the leading `_`? It's not a private helper anymore if we reference it here.

sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
 
         build_env(self.func, env=env, name=entrypoint, path=module_path)
 
+        # Properties to be rendered at model creation time
+        expressions = [exp.Property(this="name", value=exp.to_table(self.name, dialect=dialect))]
+        for field_name in {"enabled", "start", "end"}:
+            if (
+                (field_value := self.kwargs.get(field_name))
+                and isinstance(field_value, exp.Expression)
+                or (isinstance(field_value, str) and field_value.startswith(SQLMESH_MACRO_PREFIX))
```
georgesittas, line 130: Let's include parentheses so that operator precedence can be easily understood.

sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
...
+            ):
+                expressions.append(
+                    exp.Property(
+                        this=field_name,
+                        value=parse_one(field_value, dialect=dialect)
+                        if isinstance(field_value, str)
+                        else field_value,
+                    )
+                )
```
georgesittas, line 139:
```suggestion
                expressions.append(
                    exp.Property(
                        this=field_name,
                        value=exp.maybe_parse(field_value, dialect=dialect),
                    )
                )
```

sqlmesh/core/model/decorator.py
```diff
@@ -119,21 +121,49 @@ def model(
 
         build_env(self.func, env=env, name=entrypoint, path=module_path)
 
+        # Properties to be rendered at model creation time
+        expressions = [exp.Property(this="name", value=exp.to_table(self.name, dialect=dialect))]
```
tobymao, line 125: "name" should be a var, otherwise the ast is invalid

themisvaltinos:
Thanks for the review @izeigerman , @georgesittas , @tobymao  addressed comments and revised logic to handle all model properties, except for those that need to be rendered at evaluation time

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
     return exp.Tuple(expressions=values)
 
 
-def _meta_renderer(
+def render_meta_fields(
+    fields: t.Dict[str, t.Any],
+    module_path: Path,
+    path: Path,
+    jinja_macros: t.Optional[JinjaMacroRegistry] = None,
```
izeigerman, line 2358: I suggest we don't have default values for any fields here, so that we know to explicitly provide them everywhere we call this

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                    module_path=module_path,
+                    macros=macros,
+                    jinja_macros=jinja_macros,
+                    variables=variables,
+                    path=path,
+                    dialect=dialect,
+                    default_catalog=default_catalog,
+                )
+                assert rendered_expr and len(rendered_expr) == 1
+                return rendered_expr[0].sql(dialect=dialect)
```
izeigerman, line 2380: This doesn't look right. If the input is expression, we return a string instead?

izeigerman: Shouldn't we always return an expression?

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+            except Exception:
+                pass
```
izeigerman, line 2382: Why ignore?

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+
+        return value
+
+    for field_name, _ in ModelMeta.all_field_infos().items():
+        field_name = field_name[:-1] if field_name.endswith("_") else field_name
```
izeigerman, line 2387: No need to do this. You can get the alias from the `FieldInfo` object in the value of the dict that you're iterating over.

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+        if (field_value := fields.get(field_name)) and field_name not in {
+            "audits",
```
izeigerman, line 2389: Should we set these in a named constant like `RUNTIME_RENDERED_MODEL_FIELDS`

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+            "signals",
+            "physical_properties",
+            "virtual_properties",
+            "session_properties",
+        }:
+            if isinstance(field_value, t.Dict):
```
izeigerman, line 2395: Hm, not sure whether it's a good practice to check for instance of `t.Dict`. I believe we just check for `dict` in other places.

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                for key, value in field_value.items():
+                    if key != "merge_filter":
```
izeigerman, line 2397: Can be a part of the same `RUNTIME_RENDERED_MODEL_FIELDS`

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                        fields[field_name][key] = render_field_value(value)
+            elif isinstance(field_value, t.List):
```
izeigerman, line 2399: Ditto: `isinstance(..., list)`

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,59 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                    expression=exp.maybe_parse(value, dialect=dialect),
+                    module_path=module_path,
+                    macros=macros,
+                    jinja_macros=jinja_macros,
+                    variables=variables,
+                    path=path,
+                    dialect=dialect,
+                    default_catalog=default_catalog,
+                )
+                assert rendered_expr and len(rendered_expr) == 1
```
izeigerman, line 2379: Should we inform the user about the problem here?

sqlmesh/core/model/definition.py
themisvaltinos: The issue is they flow in as optional from model's arguments: 
https://github.com/TobikoData/sqlmesh/blob/dc613d05e9eb599fe564598143dd41429a8f72be/sqlmesh/core/model/decorator.py#L87-L90

sqlmesh/core/model/definition.py
themisvaltinos: Yes revised it, this was needed for the model's name, but adjusted it for this field only

sqlmesh/core/model/definition.py
themisvaltinos: This was for cron that it was for example trying to render and failing to find `@daily` macro, but revised it and simply added cron in the  `RUNTIME_RENDERED_MODEL_FIELDS` as well.

sqlmesh/core/model/definition.py
themisvaltinos: Right adapted it to follow instead the same pattern we have in the rest of the places

sqlmesh/core/constants.py
```diff
@@ -90,3 +90,12 @@
 HYBRID = "hybrid"
 
 DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
+
+RUNTIME_RENDERED_MODEL_FIELDS = {
+    "audits",
+    "signals",
+    "description",
+    "cron",
```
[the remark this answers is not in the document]
themisvaltinos: For description I was thinking we'd want to retain how it was unrendered, similarly to how we handle it here: https://github.com/TobikoData/sqlmesh/blob/c24ee34ff9cca73dcaba90b3e8780ac466c3c5d7/sqlmesh/core/model/definition.py#L201-L202
For cron since the user may pass a cron keyword like: `@hourly`, `@daily` trying to render it will fail since it will be considered a macro and unable to resolve it

sqlmesh/core/constants.py
```diff
@@ -90,3 +90,12 @@
 HYBRID = "hybrid"
 
 DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
+
+RUNTIME_RENDERED_MODEL_FIELDS = {
```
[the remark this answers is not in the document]
themisvaltinos: Right it's only referenced there for now so moved it

sqlmesh/core/model/definition.py
```diff
@@ -2361,7 +2351,52 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+
+        return value
+
+    for field_name, field_info in ModelMeta.all_field_infos().items():
+        field = field_info.alias or field_name
+        if (field_value := fields.get(field)) and field not in c.RUNTIME_RENDERED_MODEL_FIELDS:
+            if isinstance(field_value, dict):
+                for key, value in field_value.items():
+                    if key not in c.RUNTIME_RENDERED_MODEL_FIELDS:
+                        fields[field][key] = render_field_value(value)
```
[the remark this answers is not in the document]
themisvaltinos: Haven't considered this thanks revised it

sqlmesh/core/model/definition.py
```diff
@@ -2382,7 +2381,62 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                jinja_macros=jinja_macros,
+                variables=variables,
+                path=path,
+                dialect=dialect,
+                default_catalog=default_catalog,
+            )
+            if rendered_expr is None:
+                raise SQLMeshError(
+                    f"Failed to render model: `{fields['name']}` at `{path}`\n"
+                    f"\tExpected rendering `{expression.sql(dialect=dialect)}` to return an expression"
```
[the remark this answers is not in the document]
themisvaltinos: Thanks for catching it my bad

sqlmesh/core/model/definition.py
izeigerman: That's fine. I'm not suggesting making them non-optional. I suggest removing default arguments so that each caller consciously provide some value for every argument and doesn't miss them by accident.

sqlmesh/core/model/definition.py
```diff
@@ -2382,7 +2381,61 @@ def _refs_to_sql(values: t.Any) -> exp.Expression:
...
+                macros=macros,
+                jinja_macros=jinja_macros,
+                variables=variables,
+                path=path,
+                dialect=dialect,
+                default_catalog=default_catalog,
+            )
+            if rendered_expr is None:
+                raise SQLMeshError(
+                    f"Failed to render model: `{fields['name']}` at `{path}`\n"
```
izeigerman, line 2411:
```suggestion
                    f"Failed to render model attribute `{fields['name']}` at `{path}`\n"
```

izeigerman (approved):
A few smaller comments. Good to go otherwise!

sqlmesh/core/model/definition.py
themisvaltinos: Oh I see I misunderstood revising this to remove the defaults

PR #3763, opened 2025-01-31
Chore: warn instead of raising if unit test references unknown model

georgesittas:
Motivation: when a model is disabled and there's a unit test for it, running it will result in a crash. I opted for a warning because not doing anything at all can result in masking typos in enabled models that _should_ be tested.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -310,7 +310,7 @@ class ModelTest(unittest.TestCase):
         path: Path | None,
         preserve_fixtures: bool = False,
         default_catalog: str | None = None,
-    ) -> ModelTest:
+    ) -> t.Optional[ModelTest]:
         """Create a SqlModelTest or a PythonModelTest.
 
         Args:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -329,7 +329,8 @@ def create_test(
         name = normalize_model_name(name, default_catalog=default_catalog, dialect=dialect)
         model = models.get(name)
         if not model:
-            _raise_error(f"Model '{name}' was not found", path)
+            logger.warning(f"Model '{name}' was not found{' at ' + str(path) if path else ''}")
```
tobymao, line 332: logger.error maybe

georgesittas: is that so it'll be less likely to be filtered out? or are the semantics of `error` different?

treysp: Just FYI, `logger.error()` also prints to stdout in red text with the full log identifier info (like `0:00:002025-01-29 12:18:23,587 - MainThread - sqlmesh.core.renderer -`).

That caused much of the duplicative CLI error output people were seeing, so anywhere we add `logger.error()` we need to check how it does/doesn't interfere with other stdout content.

georgesittas: Sounds good. Warning is probably fine, AFAICT errors should be generally logged on fatal failures and this isn't one.

PR #3765, opened 2025-01-31
Fix decouple logging from console

izeigerman:
This PR decouples information that we present to the user in the console from the logging specifically for warnings and errors.

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -2362,7 +2362,7 @@ class EngineAdapter:
             self.execute(self._build_create_comment_table_exp(table, table_comment, table_kind))
         except Exception:
             logger.warning(
-                f"Table comment for '{table.alias_or_name}' not registered - this may be due to limited permissions.",
+                f"Table comment for '{table.alias_or_name}' not registered - this may be due to limited permissions",
                 exc_info=True,
             )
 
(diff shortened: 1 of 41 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -2362,7 +2362,7 @@ def _create_table_comment(
             self.execute(self._build_create_comment_table_exp(table, table_comment, table_kind))
         except Exception:
             logger.warning(
```
treysp, line 2364: Should this one be `get_console().log_warning()`?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2389,7 +2389,7 @@ def _create_column_comments(
                 self.execute(self._build_create_comment_column_exp(table, col, comment, table_kind))
             except Exception:
                 logger.warning(
```
treysp, line 2391: Should this one be `get_console().log_warning()`?

tobymao (approved):

sqlmesh/core/engine_adapter/base.py
izeigerman: Nope. This happens at runtime

sqlmesh/core/engine_adapter/base.py
izeigerman: same

PR #3769, opened 2025-02-01
Fix: set default catalog in jinja resolve table

z3z1ma:
Currently, if you have use model names like `mart_common.dim_date`, the jinja resolve_table will not resolve it to the models physical table unless you do `{{ resolve_table(default_catalog ~ 'mart_common.dim_date' }}`


izeigerman:
Nice, thank you! Any chance we can add a test for this? CC: @georgesittas

z3z1ma:
Sure done, I added a test which fails without this change.

izeigerman:
@z3z1ma thanks! Can you please rebase for the build to succeed.

izeigerman:
Actually, looks like I can do this myself

PR #3771, opened 2025-02-02
Feat: Enable previews by default for native projects and dbt projects if the target engine suports cloning

izeigerman:
Previously, the previews were disabled by default to prevent accidental full refreshes when making changes to forward-only models in dev environments.

This behavior was overly conservative and is only dangerous when using the dbt incremental style that relies on the maximum timestamp value in the target table. Even then, it is only problematic if the target engine does not support cloning, in which case the target table will be empty, leading to a full refresh of the dbt incremental model.

This PR enables previews for forward-only changes by default, except when the target project is dbt and the target engine does not support cloning.


docs/reference/configuration.md
```diff
@@ -61,15 +61,15 @@ Global variable values may be any of the data types in the table below or lists
...
+| `auto_apply`              | Indicates whether to automatically apply a new plan after creation (Default: False)                                                                                                                                                                     | boolean              | N        |
+| `forward_only`            | Indicates whether the plan should be [forward-only](../concepts/plans.md#forward-only-plans) (Default: False)                                                                                                                                           | boolean              | N        |
+| `enable_preview`          | Indicates whether to enable [data preview](../concepts/plans.md#data-preview) for forward-only models when targeting a development environment (Default: False for dbt projects if the target engine doesn't support cloning, True otherwise)           | boolean              | N        |
```
tobymao, line 70: Default: True unless edge case.

sqlmesh/core/model/kind.py
```diff
@@ -120,7 +120,7 @@ def is_symbolic(self) -> bool:
 
     @property
     def is_materialized(self) -> bool:
-        return not (self.is_symbolic or self.is_view)
+        return self.model_kind_name is not None and not (self.is_symbolic or self.is_view)
```
tobymao, line 123: why did this change?

tobymao (approved):

docs/reference/configuration.md
izeigerman: You sure? This feels very vague

sqlmesh/core/model/kind.py
izeigerman: Enabling previews exposed a failure in tests when we attempted to restate standalone audit snapshots. The reason was the fact that `is_materialized` returned true for this kind of snapshots. `model_kind_name` is None in this case , but since the condition was a negation it returned true.

docs/reference/configuration.md
tobymao: i meant the ordering not the actual content, put the true one first then describe the edge case

izeigerman: aaaah, sorry, I misunderstood

sungchun12:
Such a great idea. Thanks for doing this! Makes talking about forward only incrementals that much easier.

PR #3772, opened 2025-02-03
Feat!: Adjust physical_properties evaluation and add macro to resolve physical table names

erindru:
This PR:
 - Moves evaluation of `physical_properties` to run time instead of load time
   - This enables the usage of things that only exist at run time, such as `@this_model`
 - Adds a macro called `@physical_location` that allows a user to create arbitrary `exp.Literal` or `exp.Table` objects in the AST based on the *parts* of the physical table name

The motivation is twofold:
 - Right now, there is no way to set custom locations for tables on engines that separate storage and compute (Athena, Trino, Spark etc). This is true even if you write a custom macro and reference it from `physical_properties`, because `physical_properties` is currently rendered at load time which means no snapshot info is available
- Right now, `@this_model` [returns a fully rendered FQN string](https://github.com/TobikoData/sqlmesh/blob/a85a8b25be6c943a75ad282fbe84efa6ff9a0ae2/sqlmesh/core/macros.py#L452) which makes it difficult to work with. We can't change this without breaking backwards compatibility, but we can introduce something else that allows users to build new strings based on the *components* of the FQN, rather than the FQN itself

### DIFF
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -1000,6 +1000,46 @@ Note: This is DuckDB SQL and other dialects will be transpiled accordingly.
 - Recursive CTEs (common table expressions) will be used for `Redshift / MySQL / MSSQL`.
 - For `MSSQL` in particular, there's a recursion limit of approximately 100. If this becomes a problem, you can add an `OPTION (MAXRECURSION 0)` clause after the date spine macro logic to remove the limit. This applies for long date ranges.
 
+### @PHYSICAL_LOCATION
+
+`@PHYSICAL_LOCATION` is a helper macro intended to be used in situations where you need to gain access to the *components* of the physical table name. It's intended for use in the following situations:
+
+- Providing explicit control over table locations on a per-model basis for engines that decouple storage and compute (such as Athena, Trino, Spark etc)
+- Generating references to engine-specific metadata tables that are derived from the physical table name, such as the [`<table>$properties`](https://trino.io/docs/current/connector/iceberg.html#metadata-tables) metadata table in Trino.
+
+Under the hood, it uses the `@this_model` variable so it can only be used during the `creating` and `evaluation` [runtime stages](./macro_variables.md#runtime-variables).
+
+The `@PHYSICAL_LOCATION` supports the following arguments:
+
+ - `template` - The string template to render into
+ - `mode` - What to return after rendering the template. Valid values are `literal` or `table`. Defaults to `literal`.
+
+The `template` can contain the following placeholders that will be substituted:
+
+  - `@{catalog_name}` - The name of the catalog, eg `datalake`
+  - `@{schema_name}` - The name of the physical schema that SQLMesh is using for the model version table, eg `sqlmesh__landing`
+  - `@{table_name}` - The name of the physical table that SQLMesh is using for the model version, eg `landing__customers__2517971505`
+
+It can be used in a `MODEL` block:
+
+```sql linenums="1" hl_lines="5"
+MODEL (
+  name datalake.landing.customers,
+  ...
+  physical_properties (
+    location = @physical_location('s3://warehouse-data/@{catalog_name}/prod/@{schema_name}/@{table_name}')
+  )
+);
+-- CREATE TABLE "datalake"."sqlmesh__landing"."landing__customers__2517971505" ...
+-- WITH (location = 's3://warehouse-data/datalake/prod/sqlmesh__landing/landing__customers__2517971505')
+```
+
+And also within a query, using `mode := 'table'`:
+
+```sql linenums="1"
+SELECT * FROM @physical_location('@{catalog_name}.@{schema_name}.@{table_name}$properties', mode := 'table')
+-- SELECT * FROM "datalake"."sqlmesh__landing"."landing__customers__2517971505$properties"
+```
 
 ### @AND
 
--- a/docs/concepts/macros/sqlmesh_macros.md
+++ b/docs/concepts/macros/sqlmesh_macros.md
@@ -1000,19 +1000,19 @@ Note: This is DuckDB SQL and other dialects will be transpiled accordingly.
 - Recursive CTEs (common table expressions) will be used for `Redshift / MySQL / MSSQL`.
 - For `MSSQL` in particular, there's a recursion limit of approximately 100. If this becomes a problem, you can add an `OPTION (MAXRECURSION 0)` clause after the date spine macro logic to remove the limit. This applies for long date ranges.
 
-### @PHYSICAL_LOCATION
+### @RESOLVE_TEMPLATE
 
-`@PHYSICAL_LOCATION` is a helper macro intended to be used in situations where you need to gain access to the *components* of the physical table name. It's intended for use in the following situations:
+`@resolve_template` is a helper macro intended to be used in situations where you need to gain access to the *components* of the physical object name. It's intended for use in the following situations:
 
 - Providing explicit control over table locations on a per-model basis for engines that decouple storage and compute (such as Athena, Trino, Spark etc)
 - Generating references to engine-specific metadata tables that are derived from the physical table name, such as the [`<table>$properties`](https://trino.io/docs/current/connector/iceberg.html#metadata-tables) metadata table in Trino.
 
-Under the hood, it uses the `@this_model` variable so it can only be used during the `creating` and `evaluation` [runtime stages](./macro_variables.md#runtime-variables).
+Under the hood, it uses the `@this_model` variable so it can only be used during the `creating` and `evaluation` [runtime stages](./macro_variables.md#runtime-variables). Attempting to use it at the `loading` runtime stage will result in a no-op.
 
-The `@PHYSICAL_LOCATION` supports the following arguments:
+The `@resolve_template` macro supports the following arguments:
 
- - `template` - The string template to render into
- - `mode` - What to return after rendering the template. Valid values are `literal` or `table`. Defaults to `literal`.
+ - `template` - The string template to render into an AST node
+ - `mode` - What type of SQLGlot AST node to return after rendering the template. Valid values are `literal` or `table`. Defaults to `literal`.
 
 The `template` can contain the following placeholders that will be substituted:
 
@@ -1027,7 +1027,7 @@ MODEL (
   name datalake.landing.customers,
   ...
   physical_properties (
-    location = @physical_location('s3://warehouse-data/@{catalog_name}/prod/@{schema_name}/@{table_name}')
+    location = @resolve_template('s3://warehouse-data/@{catalog_name}/prod/@{schema_name}/@{table_name}')
   )
 );
 -- CREATE TABLE "datalake"."sqlmesh__landing"."landing__customers__2517971505" ...
@@ -1037,7 +1037,7 @@ MODEL (
 And also within a query, using `mode := 'table'`:
 
 ```sql linenums="1"
-SELECT * FROM @physical_location('@{catalog_name}.@{schema_name}.@{table_name}$properties', mode := 'table')
+SELECT * FROM @resolve_template('@{catalog_name}.@{schema_name}.@{table_name}$properties', mode := 'table')
 -- SELECT * FROM "datalake"."sqlmesh__landing"."landing__customers__2517971505$properties"
 ```
 
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1184,7 +1184,7 @@ def date_spine(
 
 
 @macro()
-def physical_location(
+def resolve_template(
     evaluator: MacroEvaluator,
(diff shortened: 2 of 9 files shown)


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1000,6 +1000,46 @@ Note: This is DuckDB SQL and other dialects will be transpiled accordingly.
 - Recursive CTEs (common table expressions) will be used for `Redshift / MySQL / MSSQL`.
 - For `MSSQL` in particular, there's a recursion limit of approximately 100. If this becomes a problem, you can add an `OPTION (MAXRECURSION 0)` clause after the date spine macro logic to remove the limit. This applies for long date ranges.
 
+### @PHYSICAL_LOCATION
+
+`@PHYSICAL_LOCATION` is a helper macro intended to be used in situations where you need to gain acccess to the *components* of the physical table name. It's intended for use in the following situations:
```
erindru, line 1005: As always, naming things is one of the hardest problems.

- `@physical_table` - the object being created is not always a table, such as `kind: view`
- `@this_snapshot` - uses the terminology `snapshot` which I can see in some places is being eschewed for "model version"
- `@this_model_version` - kind of long winded

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -590,6 +590,28 @@ def _evaluate_snapshot(
...
+        )
+
+        render_statements_kwargs = dict(
+            engine_adapter=adapter,
+            snapshots=snapshots,
+            deployability_index=deployability_index,
+            **common_render_kwargs,
+        )
+
+        rendered_physical_properties = snapshot.model.render_physical_properties(
```
themisvaltinos, line 611: Perhaps this could be moved inside `apply`, so it doesn't occur when limit is `None` where the properties aren't needed. Though this doesn't seem a particularly expensive render, so it's likely not that significant

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
...
 
 
+@macro()
+def physical_location(
+    evaluator: MacroEvaluator,
+    template: exp.Literal,
+    mode: exp.Literal = exp.Literal.string("literal"),
+) -> t.Union[exp.Literal, exp.Table]:
+    """
+    Generates a either a String literal or an exp.Table representing a physical table location, based on rendering the provided template String literal.
```
themisvaltinos, line 1193: Small typo: `Generates either a String literal...`

sqlmesh/core/snapshot/evaluator.py
erindru: Oh, the reason I had it outside `apply` was so that we didnt occur the hit for every `index` being applied since we only need to render it once.

I see what you're saying though. Perhaps it could be rendered right after the limit check and passed to `apply` instead? The problem is, `apply` is called in a bunch of places so all the call sites would need to be updated

sqlmesh/core/macros.py
erindru: Thanks!

sqlmesh/core/model/definition.py
```diff
@@ -630,6 +630,26 @@ def render_merge_filter(
             raise SQLMeshError(f"Expected one expression but got {len(rendered_exprs)}")
         return rendered_exprs[0].transform(d.replace_merge_table_aliases)
 
+    def render_physical_properties(self, **render_kwargs: t.Any) -> t.Dict[str, exp.Expression]:
```
izeigerman, line 633: Don't we want the same behavior for virtual properties?

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
...
+            @{schema_name} -> replaced with the schema of the exp.Table returned from @this_model
+            @{table_name} -> replaced with the name of the exp.Table returned from @this_model
+        mode: What to return.
+            'literal' -> return an exp.Literal string
+            'table' -> return an exp.Table
+
+    Example:
+        >>> from sqlglot import parse_one, exp
+        >>> from sqlmesh.core.macros import MacroEvaluator, RuntimeStage
+        >>> sql = "@physical_location('s3://data-bucket/prod/@{catalog_name}/@{schema_name}/@{table_name}')"
```
izeigerman, line 1211: @tobymao does this API look good to you?

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
     return exp.select(alias_name).from_(exploded)
 
 
+@macro()
+def physical_location(
```
izeigerman, line 1187: IMHO, the name is somewhat confusing and doesn't accurately reflect what it does. We already have `resolve_table` can this be `resolve_template`? Or perhaps we can extend `resolve_table` with an optional template instead? If it's provided then we follow the template, if not we return fqn.

erindru: Yep, I don't like the current name either.

The problem with extending `resolve_table` is that it's not a macro per se - it's a lambda that gets passed into the template context. And it's also available directly in Python from Python models and is defined to return a `str` instead of an AST node so I think changing it will cause breakage.

I like `@resolve_template` - since it's the physical object being resolved which might not always be a table, it could be a view for `kind: VIEW` models. And you need to pass a template for it to render.

sqlmesh/core/model/definition.py
erindru: Recording the result of our internal discussion here: maybe, but I couldnt think of a use-case and was trying to keep this PR small.

Since the load time rendering context is a subset of the runtime rendering context, if we were to later find a use-case to render virtual_properties at runtime then we could make the switch then without breaking backwards compatibility

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: Yes you're right, I hadn't considered that. Maybe then something like this:
```python
rendered_physical_properties = snapshot.model.render_physical_properties(
    **render_statements_kwargs
) if limit is None else None
```
But arguably I don't think it is that much of a slowdown, so the way you have it now should be fine

sqlmesh/core/model/definition.py
```diff
@@ -1839,12 +1859,9 @@ def load_sql_based_model(
...
 
-    # signals, audits and merge_filter must remain unrendered, so that they can be rendered later at evaluation runtime
-    if unrendered_signals:
-        meta_fields["signals"] = unrendered_signals
-
-    if unrendered_audits:
-        meta_fields["audits"] = unrendered_audits
+    # Discard the potentially half-rendered versions of these properties and replace them with the
+    # original unrendered versions. They will get rendered properly at evaluation time
+    meta_fields.update(unrendered_properties)
```
georgesittas, line 1864: I think this is breaking because it impacts the data hash. If someone used `@gateway` or other load-time-renderable variables within `physical_properties`, then they should get a diff due to this change.

This is similar to the recent `when_matched` refactor, IIRC.

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
...
+        this_model = exp.to_table(evaluator.locals["this_model"])
+        template_str: str = template.this
+        result = (
+            template_str.replace("@{catalog_name}", this_model.catalog)
+            .replace("@{schema_name}", this_model.db)
+            .replace("@{table_name}", this_model.name)
+        )
+
+        if mode.this.lower() == "table":
+            return exp.to_table(result)
```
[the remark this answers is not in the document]
erindru: Man this trap for young players always gets me :facepalm: 

Thanks for noticing it, will fix

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
...
+    Example:
+        >>> from sqlglot import parse_one, exp
+        >>> from sqlmesh.core.macros import MacroEvaluator, RuntimeStage
+        >>> sql = "@physical_location('s3://data-bucket/prod/@{catalog_name}/@{schema_name}/@{table_name}')"
+        >>> evaluator = MacroEvaluator(runtime_stage=RuntimeStage.CREATING)
+        >>> evaluator.locals.update({"this_model": exp.to_table("test_catalog.sqlmesh__test.test__test_model__2517971505")})
+        >>> evaluator.transform(parse_one(sql)).sql()
+        "'s3://data-bucket/prod/test_catalog/sqlmesh__test/test__test_model__2517971505'"
+    """
+    if evaluator.runtime_stage != "loading":
```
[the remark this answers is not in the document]
erindru: Good call, i'll write a test to verify the behaviour during unit tests.

Theoretically it should be ignored entirely

docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1000,6 +1000,46 @@ Note: This is DuckDB SQL and other dialects will be transpiled accordingly.
 - Recursive CTEs (common table expressions) will be used for `Redshift / MySQL / MSSQL`.
 - For `MSSQL` in particular, there's a recursion limit of approximately 100. If this becomes a problem, you can add an `OPTION (MAXRECURSION 0)` clause after the date spine macro logic to remove the limit. This applies for long date ranges.
 
+### @RESOLVE_TEMPLATE
+
+`@resolve_template` is a helper macro intended to be used in situations where you need to gain access to the *components* of the physical object name. It's intended for use in the following situations:
```
tobymao, line 1005: why template? i'm not sure i understand the name, seems a bit too generic

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
...
+        if "this_model" not in evaluator.locals:
+            raise SQLMeshError(
+                "@this_model must be present in the macro evaluation context in order to use @resolve_template"
+            )
+
+        this_model = exp.to_table(evaluator.locals["this_model"])
+        template_str: str = template.this
+        result = (
+            template_str.replace("@{catalog_name}", this_model.catalog)
```
tobymao, line 1226: is this just doing string replacement?, is this safe?

sqlmesh/core/macros.py
```diff
@@ -1183,6 +1183,59 @@ def date_spine(
     return exp.select(alias_name).from_(exploded)
 
 
+@macro()
+def resolve_template(
+    evaluator: MacroEvaluator,
+    template: exp.Literal,
+    mode: exp.Literal = exp.Literal.string("literal"),
```
tobymao, line 1190: this can just be a literal string, the type convert will handle it

mode: str

sqlmesh/core/macros.py
tobymao: ok this is fine

sqlmesh/core/macros.py
erindru: In short, yes. It's doing the same thing that [this](https://github.com/TobikoData/sqlmesh/blob/c24ee34ff9cca73dcaba90b3e8780ac466c3c5d7/sqlmesh/core/engine_adapter/trino.py#L361) is doing to keep consistency. It deliberately mimics our macro syntax, is that what you meant by safety?

I was holding off "upgrading" it to actually run the template string through the macro evaluator via `evaluator.evaluate()`, figuring that could be easily added later if the need arose

docs/concepts/macros/sqlmesh_macros.md
erindru: maybe `@render_template` would be more appropriate?

It's essentially rendering an in-line template with `@this_model` broken out into its components and made available as `@{catalog_name}`, `@{schema_name}` and `@{table_name}` so the user can recombine those as required

My [original implementation](https://github.com/TobikoData/sqlmesh/pull/3747/files#diff-fddedea4eda285dbd51b83821f3cd54dc42d3946c0056760acdc29efc3e565afR163) added these as variables to the main evaluation context so users could access them directly, something like:
```
physical_properties (
  location = @"s3://bucket/@{catalog_name}/@{schema_name}/@{table_name}"
)
```

but there was a preference to create a macro instead?

sqlmesh/core/macros.py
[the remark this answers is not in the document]
erindru: I've adjusted the logic to be dependent on the presence of `@this_model` instead and written a test to show it works from unit tests

sqlmesh/core/model/definition.py
erindru: I spoke with @izeigerman and there isnt a way to migrate this automatically.

The problem is that the state would contain already-rendered macros and we wouldnt know what the unrendered version would look like without the project files.

When a migration occurs, access to the project files is not guaranteed. So we are just hoping that not many people were using macros / variables in `physical_properties`.

The workaround if this triggers a bunch of metadata changes and tries to rebuild everything would be to run a `--forward-only` plan to update state but not rebuild any tables

georgesittas: Sounds good– sorry for not getting back to this thread earlier.

PR #3778, opened 2025-02-03
fix: load multi repo models earlier to ensure schema is correct

tobymao:


sqlmesh/core/context.py
```diff
@@ -2105,63 +2116,39 @@ def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
...
             fingerprint_cache: t.Dict[str, SnapshotFingerprint] = {}
 
             for node in nodes.values():
-                if node.fqn not in local_nodes and node.fqn in remote_snapshots:
-                    ttl = remote_snapshots[node.fqn].ttl
-                else:
-                    config = self.config_for_node(node)
-                    ttl = config.snapshot_ttl
+                kwargs = {}
+                if node._path:
```
izeigerman, line 2125: oh, that's a bit dangerous. Is there a more explicit signal for us to know that the node was loaded from the local file system?

izeigerman: Perhaps there's an explicit signal that tells that it wasn't

sqlmesh/core/context.py
```diff
@@ -2105,63 +2116,39 @@ def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
...
+                    kwargs["ttl"] = self.config_for_node(node).snapshot_ttl
```
izeigerman, line 2126: Why do we no longer need the `else` branch? Don't we want to derive the value from the existing snapshot if available?

tobymao: because we override it from the snapshots anyway when we get the snapshot from disk

sqlmesh/core/context.py
```diff
@@ -574,15 +575,38 @@ def load(self, update_schemas: bool = True) -> GenericContext[C]:
...
+                model = self._models.get(fqn)  # type: ignore
+
+                if not model or not model._data_hash:
+                    continue
+
+                # make a copy of remote models that depend on local models or in the downstream chain
+                # without this, a SELECT * FROM local will not propogate properly because the downstream
+                # model will get mutated (schema changes) but the object is the same as the remote cache
+                if any(
+                    not self._models[dep]._data_hash
```
izeigerman, line 601: Does missing `_data_hash` signals that the model wasn't loaded from cache or that it's a local model? If it's the latter can we check by project name instead?

tobymao: project is not enough, you need the tree of things that are invalidated

sqlmesh/core/model/definition.py
```diff
@@ -1255,8 +1265,12 @@ def columns_to_types(self) -> t.Optional[t.Dict[str, exp.DataType]]:
             if query is None:
                 return None
 
+            unknown = exp.DataType.build("unknown")
+
             self._columns_to_types = {
-                select.output_name: select.type or exp.DataType.build("unknown")
+                # copy data type because it is used in the engine to build CTAS and other queries
+                # this can change the parent which will miss up the diffing algo
+                select.output_name: (select.type or unknown).copy()
```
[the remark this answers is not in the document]
tobymao: this is needed or else it will break since i got rid of copy in diff

izeigerman (approved):

PR #3780, opened 2025-02-03
Give a primer for new tcloud users

sungchun12:
A video helps people get comfortable before meeting with SAs, so they don't feel surprised during the live setup meeting.

It also inherently sets up the native scheduler.


afzaljasani (approved):
lgtm

PR #3781, opened 2025-02-03
feat: don't force db connect if using serverless

eakmanrq:
Prior to this PR, if a user said they wanted to use Serverless for databricks-connect then it forced the use of databricks-connect and therefore one could not use the python sql connector. In addition the documentation said that the SQL connector did not support Databricks Serverless Compute which was misleading - although it doesn't support the workspace side Serverless, typically used by Notebooks and Jobs, it does support SQL Warehouse Serverless compute. 

Therefore a user could have wanted to use serverless across their stack - Serverless compute for jobs that require PySpark DataFrame and SQL Warehouse Serverless for their SQL queries. This PR now enables this pattern. 

One key limitation it works around was temporary objects - since serverless doesn't support global temporary objects, and instead requires session temporary objects, there was an issue mixing databricks-connect and Python SQL connector across the serverless products since they couldn't share this state. This PR resolves this by recording in session connection metadata if a temporary objects was made in a databricks-connect session and if so it will force using databricks-connect for the remainder of the session.

This PR also adds improvements to documentation, removes excess log output in the console, and improves error message if the user has different default catalogs across their SQL and databricks-connect sessions. 

Initial PR that added serverless support for context: https://github.com/TobikoData/sqlmesh/pull/3001

### DIFF
--- a/docs/integrations/engines/databricks.md
+++ b/docs/integrations/engines/databricks.md
@@ -14,9 +14,9 @@ SQLMesh connects to Databricks with the [Databricks SQL Connector](https://docs.
 
 The SQL Connector is bundled with SQLMesh and automatically installed when you include the `databricks` extra in the command `pip install "sqlmesh[databricks]"`.
 
-The SQL Connector has all the functionality needed for SQLMesh to execute SQL models on Databricks and Python models locally (the default SQLMesh approach).
+The SQL Connector has all the functionality needed for SQLMesh to execute SQL models on Databricks and Python models that do not return PySpark DataFrames.
 
-The SQL Connector does not support Databricks Serverless Compute. If you require Serverless Compute then you must use the Databricks Connect library.
+If you have Python models returning PySpark DataFrames, check out the [Databricks Connect](#databricks-connect-1) section.
 
 ### Databricks Connect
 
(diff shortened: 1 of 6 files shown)


docs/integrations/engines/databricks.md
```diff
@@ -14,9 +14,9 @@ SQLMesh connects to Databricks with the [Databricks SQL Connector](https://docs.
 
 The SQL Connector is bundled with SQLMesh and automatically installed when you include the `databricks` extra in the command `pip install "sqlmesh[databricks]"`.
 
-The SQL Connector has all the functionality needed for SQLMesh to execute SQL models on Databricks and Python models locally (the default SQLMesh approach).
+The SQL Connector has all the functionality needed for SQLMesh to execute SQL models on Databricks and Python models that do not return PySpark DataFrames.
 
-The SQL Connector does not support Databricks Serverless Compute. If you require Serverless Compute then you must use the Databricks Connect library.
+If you have Python models returning PySpark DataFrames, checkout the [Databricks Connect](#databricks-connect-1) section.
```
treysp, line 19:
```suggestion
If you have Python models returning PySpark DataFrames, check out the [Databricks Connect](#databricks-connect-1) section.
```

docs/integrations/engines/databricks.md
```diff
@@ -229,7 +229,9 @@ If you want Databricks to process PySpark DataFrames in SQLMesh Python models, t
 
 SQLMesh **DOES NOT** include/bundle the Databricks Connect library. You must [install the version of Databricks Connect](https://docs.databricks.com/en/dev-tools/databricks-connect/python/install.html) that matches the Databricks Runtime used in your Databricks cluster.
 
-SQLMesh's Databricks Connect implementation supports Databricks Runtime 13.0 or higher. If SQLMesh detects that you have Databricks Connect installed, then it will use it for all Python models (both Pandas and PySpark DataFrames).
+If SQLMesh detects that you have Databricks Connect installed, then it will automatically configure the connection and use it for all Python models that return a Pandas or PySpark DataFrame.
+
+To have databricks-connect installed but ignored by SQLMesh, set `disable_databricks_connect` to `true` in the connection configuration.
```
treysp, line 234: When/why would someone want to have it installed but ignored? 

Just needing installed in the env for a non-sqlmesh reason, or people would switch back and forth between using it in sqlmesh and not?

treysp (approved):

docs/integrations/engines/databricks.md
eakmanrq: > Just needing installed in the env for a non-sqlmesh reason

Yeah this is what I am thinking of. They use a single python environment for their work and therefore they don't want to uninstall it just so SQLMesh behaves as they expect.

PR #3782, opened 2025-02-04
Chore: Address Clickhouse Cloud test flakiness

erindru:
This PR attempts to debug the following issue noticed in the failing clickhouse cloud tests on main:
```
 http.client.RemoteDisconnected: Remote end closed connection without response
```

It looks like Clickhouse Cloud can now go into an `idle` state. I added a pre-ping to ensure the cicd instance is running before triggering the tests.

However, the issue still occurred. So I set:
```
connect_timeout: 30
```
on the connection config (previously it was using the default value of 10).

This actually did fix the connection issue but it raised another issue:

```
DB::Exception: Too many databases. The limit (server configuration parameter `max_database_num_to_throw`) is set to 100, the current number of databases is 272.
```

So it looks like Clickhouse Cloud is now limiting the number of "databases" (schemas) that you can have on a single instance. I cleared out a bunch of test schemas left over from failed runs and now the tests can run successfully again.

After this, I still experienced some flakiness. I reduced the pytest concurrency to 1 and increased the number of HTTP retries on the Clickhouse connection pool. I also removed the `-x` option because I notice it conflicts with `pytest-retry` and results in `pytest-retry` not retrying tests (this is a problem for the other engine adapter integration tests too, however that can be addressed in a different PR)

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -308,15 +308,15 @@ workflows:
             parameters:
               engine:
                 - snowflake
-                - databricks
-                - redshift
-                - bigquery
+                #- databricks
+                #- redshift
+                #- bigquery
                 - clickhouse-cloud
-                - athena
-          filters:
-           branches:
-             only:
-               - main
+                #- athena
+          #filters:
+          # branches:
+          #   only:
+          #     - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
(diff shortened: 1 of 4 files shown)


.circleci/continue_config.yml
```diff
@@ -308,15 +308,15 @@ workflows:
...
-                - athena
-          filters:
-           branches:
-             only:
-               - main
+                #- athena
+          #filters:
+          # branches:
+          #   only:
+          #     - main
```
erindru, line 319: TODO: reinstate prior to merge

tobymao (approved):

georgesittas (approved):

PR #3791, opened 2025-02-05
Feat: Add 'auditing' and 'promoting' runtime stages

izeigerman:
We render queries when executing audits as well as `on_virtual_update` statements, however neither has a runtime stage associated with it. This PR addresses the gap.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -918,6 +918,7 @@ class SnapshotEvaluator:
                 snapshots=snapshots,
                 deployability_index=deployability_index,
                 table_mapping=table_mapping,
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -918,6 +918,7 @@ def _promote_snapshot(
                 snapshots=snapshots,
                 deployability_index=deployability_index,
                 table_mapping=table_mapping,
+                runtime_stage=RuntimeStage.PROMOTING,
```
izeigerman, line 921: Even though, on_virtual_update statements are defined in a separate block, having a dedicate runtime stage still makes sense for cases when the same macro is reused across multiple stages.

tobymao (approved):

PR #3799, opened 2025-02-06
fix: db properly support `with_log_level`

eakmanrq:
Follow up to: https://github.com/TobikoData/sqlmesh/pull/3793

The key thing that was missing in prior PR was properly supporting the `with_log_level` method. It creates a copy of the engine adapter but since it doesn't want to create a new connection is creates it with a placeholder connection. The issue is that databricks engine adapter tries to set the default engine adapter after creation and then when it tries to do that with the placeholder connection it gets an error.

I moved the set_current_catalog to the `_begin_session` method so although now it will run multiple times it does more lazily run it which gets around the `with_log_level` issue. 

I then ran the integration tests with hybrid, no databricks connect, and only databricks connect and fixed issues that came up. I got some errors related to converting Pandas to SQL if we don't have databricks-connect but that should be unrelated to this PR.

### DIFF
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -21,7 +21,7 @@ from sqlmesh.engines.spark.db_api.spark_session import connection, SparkSessionC
 from sqlmesh.utils.errors import SQLMeshError, MissingDefaultCatalogError
 
 if t.TYPE_CHECKING:
-    from sqlmesh.core._typing import SchemaName, TableName
+    from sqlmesh.core._typing import SchemaName, TableName, SessionProperties
     from sqlmesh.core.engine_adapter._typing import DF, PySparkSession, Query
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -48,9 +48,13 @@ class DatabricksEngineAdapter(SparkEngineAdapter):
         },
     )
 
-    def __init__(self, *args: t.Any, **kwargs: t.Any):
-        super().__init__(*args, **kwargs)
+    def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
+        super().__init__(*args, **{**kwargs, "skip_connection_post_init": True})
         self._set_spark_engine_adapter_if_needed()
+        if not kwargs.get("skip_connection_post_init"):
```
izeigerman, line 54: Why not just let the base class call `_connection_post_init`?

eakmanrq: Because the spark engine adapter would not be set by that time.

izeigerman (approved):

PR #3803, opened 2025-02-07
Chore: Use 'dev' suffix instead of 'temp' for non-deployable physical tables

izeigerman:
This PR ensures consistent terminology in the codebase for referring to non-deployable physical tables. The non-deployable physical tables are going to be called as "dev" tables going forward.

Additionally, this updates changes the suffix for non-deployable table names from `__temp` to `__dev` for the following reasons:
* These tables are not truly temporary; their lifespan matches that of the snapshots they are associated with. As long as a snapshot exists, its corresponding dev table persists
* The existing suffix confuses users who don't like seeing something that looks like a temporary table in their warehouse
* `__temp` is used elsewhere in the codebase to denote actual temporary tables, whose lifespan does not extend beyond the duration of the user session

At the same time, this update ensures that existing dev tables with the old suffix are not lost or orphaned. Existing snapshots will continue to follow the old naming pattern. All new snapshots will follow the new naming pattern.

### DIFF
--- a/sqlmesh/migrations/v0069_update_dev_table_suffix.py
+++ b/sqlmesh/migrations/v0069_update_dev_table_suffix.py
@@ -1,0 +1,165 @@
+"""Update the dev table suffix to be 'dev'. Rename temp_version to dev_version."""
+
+import json
+
+import pandas as pd
+from sqlglot import exp
+
+from sqlmesh.utils.migration import index_text_type, blob_text_type
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
+    engine_adapter = state_sync.engine_adapter
+    schema = state_sync.schema
+    snapshots_table = "_snapshots"
+    environments_table = "_environments"
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+        environments_table = f"{schema}.{environments_table}"
+
+    index_type = index_text_type(engine_adapter.dialect)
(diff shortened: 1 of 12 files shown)


sqlmesh/migrations/v0069_update_dev_table_suffix.py
```diff
@@ -0,0 +1,165 @@
...
+        "promoted_snapshot_ids": exp.DataType.build(blob_type),
+        "suffix_target": exp.DataType.build("text"),
+        "catalog_name_override": exp.DataType.build("text"),
+        "previous_finalized_snapshots": exp.DataType.build(blob_type),
+        "normalize_name": exp.DataType.build("boolean"),
+        "requirements": exp.DataType.build(blob_type),
+    }
+
+    new_snapshots = []
+    for (
```
erindru, line 51: I think we need to come up with safer ways to write migrations, I see a lot of duplication / potential for issues between the migration scripts

erindru (approved):

sqlmesh/migrations/v0069_update_dev_table_suffix.py
izeigerman: Duplication is by design, since we can't reuse anything in the current code base, nor can we reuse anything from previous migrations. Each script must be self-contained and isolated.

PR #3805, opened 2025-02-08
Feat: include alter statements in destructive change error message

treysp:
When someone gets the destructive change error message, they want to understand more about the problem than just the name of the affected column(s).

SQLMesh has not actually run any ALTER statements when this error occurs, so the statements are not printed to the log. That makes it difficult to figure out what SQLMesh "wanted to do" that it decided it shouldn't.

This PR adds those ALTER statements to the destructive error/warning message.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -1220,10 +1220,7 @@ class TerminalConsole(Console):
                 )
             )
         else:
-            self._print(
-                f"[yellow]WARNING: Plan requires a destructive change to forward-only model '{snapshot_name}'s schema[/yellow]"
-            )
-            logger.warning(
+            self.log_warning(
                 format_destructive_change_msg(
                     snapshot_name, dropped_column_names, alter_expressions, dialect, error
                 )
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -35,11 +35,12 @@ from sqlmesh.core.test import ModelTest
 from sqlmesh.utils import rich as srich
 from sqlmesh.utils.concurrency import NodeExecutionFailedError
 from sqlmesh.utils.date import time_like_to_str, to_date, yesterday_ds
-from sqlmesh.utils.errors import PythonModelEvalError
+from sqlmesh.utils.errors import PythonModelEvalError, format_destructive_change_msg
 
 if t.TYPE_CHECKING:
     import ipywidgets as widgets
 
+    from sqlglot import exp
     from sqlglot.dialects.dialect import DialectType
     from sqlmesh.core.context_diff import ContextDiff
     from sqlmesh.core.plan import Plan, EvaluatablePlan, PlanBuilder, SnapshotIntervals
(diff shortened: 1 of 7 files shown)


sqlmesh/core/console.py
```diff
@@ -1183,6 +1205,30 @@ def log_failed_models(self, errors: t.List[NodeExecutionFailedError]) -> None:
...
+                format_destructive_change_msg(
+                    snapshot_name, dropped_column_names, alter_expressions, dialect
+                )
+            )
+        else:
+            self._print(
+                f"[yellow]WARNING: Plan requires a destructive change to forward-only model '{snapshot_name}'s schema[/yellow]"
+            )
+            logger.warning(
+                format_destructive_change_msg(
```
izeigerman, line 1227: Why do we not print this into console?

sqlmesh/core/plan/builder.py
```diff
@@ -482,21 +482,24 @@ def _check_destructive_changes(self, directly_modified: t.Set[SnapshotId]) -> No
...
-                        f" that drops column{'s' if dropped_column_names and len(dropped_column_names) > 1 else ''} '{dropped_column_str}'"
-                        if dropped_column_str
-                        else ""
-                    )
-                    warning_msg = f"Plan results in a destructive change to forward-only model '{snapshot.name}'s schema{dropped_column_msg}."
+                    model_dialect = snapshot.model.dialect
+
                     if snapshot.model.on_destructive_change.is_warn:
-                        get_console().log_warning(warning_msg)
+                        get_console().log_destructive_change(
```
izeigerman, line 490: We don't need to duplicate this in 2 branches:
```
 get_console().log_destructive_change(
  snapshot_name,
      dropped_column_names,
      schema_diff,
      model_dialect,
      error=not snapshot.model.on_destructive_change.is_warn,
 )
```

sqlmesh/core/plan/evaluator.py
```diff
@@ -288,11 +288,18 @@ def _promote(
...
-            )
+            try:
+                self.snapshot_evaluator.migrate(
+                    [s for s in snapshots.values() if s.is_paused],
+                    snapshots,
+                    plan.allow_destructive_models,
+                )
+            except NodeExecutionFailedError as ex:
+                logger.info(str(ex), exc_info=ex)
+                self.console.log_failed_models([ex])
```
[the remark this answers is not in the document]
treysp: My understanding of the flow we set up in the previous error handling PR was:
- In snapshot evaluator, raise NodeExecutionFailedError
- When return from snapshot evaluator, identify/detect NodeExecutionFailedError
- If NodeExecutionFailedError present, log it to console and logger
- Raise PlanError

So the `except` here was the way I implemented the "intercept and log NodeExecutionFailedError and raise PlanError" pattern.

Without it, the NodeExecutionFailedError raises all the way up to console. (Which may be fine, that's just not the approach we're currently using.)

sqlmesh/core/console.py
treysp: Since it's an inline warning, I figured we'd want one line of information rather than everything. Will update.

sqlmesh/core/plan/evaluator.py
```diff
@@ -288,11 +288,15 @@ def _promote(
...
-                plan.allow_destructive_models,
-            )
+            try:
+                self.snapshot_evaluator.migrate(
+                    [s for s in snapshots.values() if s.is_paused],
+                    snapshots,
+                    plan.allow_destructive_models,
+                )
+            except Exception as ex:
+                raise PlanError(str(ex.__cause__))
```
izeigerman, line 298: Are we sure that we always get a cause here? Should we then pattern match on `NodeEvaluation..` rather than `Exception`?

treysp: Ah, my bad. Should we also guard against no cause like this?

```
           except NodeExecutionFailedError as ex:
                raise PlanError(str(ex.__cause__) if ex.__cause__ else str(ex))
```

izeigerman (approved):

Issue #3702, opened 2025-01-24
Link (possibly links) broken in python API docs


ebremstedt:
If i click on `edit on Github`  on this link: 
https://sqlmesh.readthedocs.io/en/stable/_readthedocs/html/sqlmesh/core/lineage.html

I am sent to https://github.com/TobikoData/sqlmesh/core/lineage.py (which doesnt work)

I believe this should be https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/lineage.py

treysp:
Thanks for letting us know - much appreciated!

PR #3807, opened 2025-02-08
fix: update github links for pdoc api docs

gerrykou:
On api docs generated by pdoc, the edit github link is broken as describe din #3702
Tested locally, it works.
Better to be tested so it is sure that is not braking another functionality.
Fixes #3802


treysp:
Thanks!

PR #3808, opened 2025-02-08
Replace pymssql with pyodbc

mattiasthalen:
Replace pymssql in mssql in connectors with pyodbc.

This opens the ability to use other kind of authorisations, e.g., service principal. It will also be needed in the future in order to integrate with Microsoft Fabric.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1248,6 +1248,7 @@ class MySQLConnectionConfig(ConnectionConfig):
 
 class MSSQLConnectionConfig(ConnectionConfig):
     host: str
+    authentication: t.Optional[str] = None
     user: t.Optional[str] = None
     password: t.Optional[str] = None
     database: t.Optional[str] = ""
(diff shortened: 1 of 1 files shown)


mattiasthalen:
Failing ci due to pyodbc not being in the requirements. I've tried finding where to put it, but no luck. Any hints?

erindru:
> Failing ci due to pyodbc not being in the requirements. 

Check setup.py, `extras_require` for `mssql`

However - I don't think we want to _replace_ pymssql, because that will break SQLMesh for every project that is currently using MSSQL the second they upgrade SQLMesh (by introducing a requirement to have ODBC set up and ODBC drivers installed)

Have you considered implementing this as an option instead? Something like adding a `driver` parameter the config. It could default to `driver=pymssql` and retain the current behaviour, or could be set to `driver=odbc` to trigger the new behaviour

tobymao:
closing this for now as there's been no progress. happy to reopen when it's ready

PR #3809, opened 2025-02-09
Feat!: Run audits as non-blocking on dev previews

erindru:
Currently, audits on dev preview tables are skipped entirely on the basis that auditing a preview of the data is kind of pointless vs auditing the actual data.

However, this can be quite surprising to users that:
 - Use `forward_only` incremental models and change them in a development environment (creating a dev preview)
 - Have CI/CD pipelines set up and expect PR's to run the full set of audits on the dev preview in the development environment before allowing merge to prod

So, this PR enables audits on dev preview tables. However, it's likely that `row_count` audits in particular will fail if they are only run on a subset of the data. So the tradeoff is forcing all audits to non-blocking so that users can decide for themselves whether or not something should be merged.


izeigerman (approved):
nice

PR #3826, opened 2025-02-12
Feat: add [WARNING] to console warning messages

treysp:


sqlmesh/core/console.py
```diff
@@ -267,7 +270,7 @@ def log_error(self, message: str) -> None:
         """Display error info to the user."""
 
     @abc.abstractmethod
-    def log_warning(self, message: str) -> None:
+    def log_warning(self, message: str, add_warning_label: bool = True) -> None:
```
izeigerman, line 273: Why not add the label to all warning messages?

treysp: Because the Markdown console output must start with "```" and it calls this method

I guess the alternative approaches are:
- Do nothing and have `[WARNING]` appear before the code block opens with "```"
- Expand the lstrip block to also handle "```"

izeigerman: Can we just override it and not call `super()`?

izeigerman (approved):

Issue #3830, opened 2025-02-13
cicd-test fails


lafirm:
after the recent [audit log message PR](https://github.com/TobikoData/sqlmesh/pull/3818/files#diff-5af59db99ac14709aa0e02acc49ee767af92605439b2f23c567b4c7d0618b2b1R1371-R1374) , cicd test fails and I think that's due to the false row count.

<img width="1476" alt="Image" src="https://github.com/user-attachments/assets/a4886088-19d8-4cb2-afcb-144048c85050" />

PR #3831, opened 2025-02-13
fix:cicd test

lafirm:
should fix https://github.com/TobikoData/sqlmesh/issues/3830


lafirm:
apologies! didn't check this https://github.com/TobikoData/sqlmesh/pull/3828

georgesittas:
Thanks for the PR regardless :)
Let's use Erin's instead for this one.

PR #3832, opened 2025-02-13
Feat: Extend support of project wide model properties

themisvaltinos:
This update extends the model properties that are supported to be enabled across the entire project through the configuration to also include:  `allow_partials`, `enabled`, `interval_unit`.

### DIFF
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -42,13 +42,6 @@ class ModelDefaultsConfig(BaseConfig):
         allow_partials: Whether the models can process partial (incomplete) data intervals.
         enabled: Whether the models are enabled.
         interval_unit: The temporal granularity of the models data intervals. By default computed from cron.
-        column_descriptions: A key-value mapping of column names to their descriptions.
-        description: Description of the models. Automatically registered in the SQL engine's table COMMENT field or equivalent.
-        grains: The column(s) whose combination uniquely identifies each row in the models.
-        references: The model column(s) used to join to models' grains.
-        clustered_by: The column(s) used to cluster the models' physical table.
-        partitioned_by: The column(s) used to define the models' partitioning key.
-        tags: Arbitrary strings used to organize or classify the models.
 
     """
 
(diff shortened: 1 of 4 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1885,6 +1885,8 @@ def load_sql_based_model(
         "description": (
             "\n".join(comment.strip() for comment in rendered_meta.comments)
             if rendered_meta.comments
+            else description
+            if defaults and (description := defaults.get("description"))
```
georgesittas, line 1889: Do we use `description` anywhere else? If not, I'd probably go for `else defaults.get(...)` instead of `:=` here.

themisvaltinos: Thanks! good point we don't so revised it

sqlmesh/core/config/model.py
```diff
@@ -53,6 +66,16 @@ class ModelDefaultsConfig(BaseConfig):
     audits: t.Optional[t.List[FunctionCall]] = None
     optimize_query: t.Optional[bool] = None
     validate_query: t.Optional[bool] = None
+    allow_partials: t.Optional[bool] = None
+    interval_unit: t.Optional[IntervalUnit] = None
+    enabled: t.Optional[bool] = None
+    description: t.Optional[str] = None
+    column_descriptions: t.Optional[t.Dict[str, str]] = None
+    grains: t.Optional[t.List[str]] = None
+    references: t.Optional[t.List[str]] = None
```
izeigerman, line 75: ditto

sqlmesh/core/config/model.py
```diff
@@ -53,6 +66,16 @@ class ModelDefaultsConfig(BaseConfig):
     audits: t.Optional[t.List[FunctionCall]] = None
     optimize_query: t.Optional[bool] = None
     validate_query: t.Optional[bool] = None
+    allow_partials: t.Optional[bool] = None
+    interval_unit: t.Optional[IntervalUnit] = None
+    enabled: t.Optional[bool] = None
+    description: t.Optional[str] = None
+    column_descriptions: t.Optional[t.Dict[str, str]] = None
```
izeigerman, line 73: ditto. Do we expect models to have same columns?

themisvaltinos: yes I agree that this along with the others you mentioned make sense as a model-specified property. It was only in the rare case where a project uses the same column (e.g. id) as the primary key for most or all models and the user doesn't have to redefine them in each model. I can remove these though

georgesittas (approved):

PR #3836, opened 2025-02-13
Feat: allow different warning messages for logger and console

treysp:
Currently, the same warning message is print both to file and console. However, we sometimes have more information for the user than is sensible to print to the console. 

This PR adds the ability to print separate messages to file and console. If separate messages are passed, the console message will automatically include a link to the log file.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -270,7 +270,7 @@ class Console(abc.ABC):
         """Display error info to the user."""
 
     @abc.abstractmethod
-    def log_warning(self, message: str) -> None:
+    def log_warning(self, message: str, logger_message: t.Optional[str] = None) -> None:
         """Display warning info to the user."""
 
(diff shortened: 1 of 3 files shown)


sqlmesh/core/console.py
```diff
@@ -270,8 +270,13 @@ def log_error(self, message: str) -> None:
         """Display error info to the user."""
 
     @abc.abstractmethod
-    def log_warning(self, message: str) -> None:
-        """Display warning info to the user."""
+    def log_warning(self, message: str, logger_message: t.Optional[str] = None) -> None:
```
izeigerman, line 273: The fact that the console uses a logger is an implementation detail and shouldn't leak into the interface. I suggest calling this `short_message` and `long_message` instead. The long_message may have applications in future other than just printing into the log file

izeigerman (approved):

PR #3837, opened 2025-02-14
Fix: Streamline execution of pre- / post- statements when creating a physical table

izeigerman:
Previously we'd only run pre- / post- statements for the non-dev tables, completely ignoring the dev tables. This update makes it so that pre- and post- statements run consistently for **every** table SQLMesh creates.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -757,25 +757,14 @@ class SnapshotEvaluator:
         deployability_index = deployability_index or DeployabilityIndex.all_deployable()
 
         adapter = self._get_adapter(snapshot.model.gateway)
-        common_render_kwargs: t.Dict[str, t.Any] = dict(
+        create_render_kwargs: t.Dict[str, t.Any] = dict(
             engine_adapter=adapter,
             snapshots=parent_snapshots_by_name(snapshot, snapshots),
             runtime_stage=RuntimeStage.CREATING,
+            deployability_index=deployability_index,
         )
-        pre_post_render_kwargs = dict(
-            **common_render_kwargs,
-            deployability_index=deployability_index.with_deployable(snapshot),
-        )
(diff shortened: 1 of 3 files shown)


tests/core/test_snapshot_evaluator.py
```diff
@@ -2826,7 +2826,7 @@ def test_create_post_statements_use_deployable_table(
     snapshot = make_snapshot(model)
     snapshot.categorize_as(SnapshotChangeCategory.BREAKING)
 
-    expected_call = f'CREATE INDEX IF NOT EXISTS "test_idx" ON "sqlmesh__test_schema"."test_schema__test_model__{snapshot.version}" /* test_schema.test_model */("a" NULLS FIRST)'
+    expected_call = f'CREATE INDEX IF NOT EXISTS "test_idx" ON "sqlmesh__test_schema"."test_schema__test_model__{snapshot.version}__dev" /* test_schema.test_model */("a" NULLS FIRST)'
```
erindru, line 2829: This test is `test_create_post_statements_use_deployable_table`

But it shows it using the dev / non deployable table?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1085,6 +1068,40 @@ def _get_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
             raise SQLMeshError(f"Gateway '{gateway}' not found in the available engine adapters.")
         return self.adapter
 
+    def _execute_create(
```
erindru, line 1071: Nice one cleaning this up!

tests/core/test_snapshot_evaluator.py
izeigerman: Ah, the test name is obsolete. It's actually creating a non-deployable table and expects post-statements to still reference the deployable one. This is logic is no longer expected so the test is wrong. Will rename it

PR #3838, opened 2025-02-14
fix: respect disable_restatement remove intervals across env

eakmanrq:
Prior to this PR, when we removed intervals across environments, introduced in https://github.com/TobikoData/sqlmesh/pull/3511, we would also remove intervals for downstream snapshots even if they had `disable_restatement` enabled. One tricky part is that the model, in most cases, would not be backfilled in that plan and this would not be communicated in console output. A case where it would be shown is if that downstream model was also an upstream model of one of the restated models or it's downstream models. I get that was confusing - lets give this example:

A -> B -> C

If `B` had `disable_restatement`, and `A` was being restated, then `B` would both have it's intervals removed and it would be backfilled in the plan since it is upstream of `C`. But if `C` had `disable_restatement` then it would have it's intervals removed but would not be backfilled in the plan since it was not upstream of anything else. 

Going through the effort of explaining this in case someone needs to try to reproduce this in the future because I kept trying patterns where just `C` was disabled and couldn't repro.

This PR fixes the issue by doing the following:
* EvaluatablePlan now includes `disabled_restatement_models`
   * This is needed since `EvaluatablePlan` doesn't include any Snapshots, just `SnapshotTableInfo`, and `SnapshotTableInfo` just includes the model kind type but not addition info for the given kind. 
* When finding downstream snapshots across environments, we now filter out the snapshots that have `disabled_restatement_models`

We are basing what is disabled or not on what is in prod but it seems the worst case scenario for this is that a dev snapshot could be restated when it wasn't intended but that seems fine since it is just in development.

### DIFF
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -255,6 +255,11 @@ class Plan(PydanticModel, frozen=True):
             models_to_backfill=self.models_to_backfill,
             interval_end_per_model=self.interval_end_per_model,
             execution_time=self.execution_time,
+            disabled_restatement_models={
+                s.name
+                for s in self.snapshots.values()
+                if s.is_model and s.model.disable_restatement
+            },
         )
 
     @cached_property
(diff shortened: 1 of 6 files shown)


sqlmesh/core/plan/definition.py
```diff
@@ -255,6 +255,11 @@ def to_evaluatable(self) -> EvaluatablePlan:
             models_to_backfill=self.models_to_backfill,
             interval_end_per_model=self.interval_end_per_model,
             execution_time=self.execution_time,
+            disabled_restatement_snapshots={
```
izeigerman, line 258: [Nit] It looks like the convention is to use "models" every time we reference by name and use "snapshot" every time we reference by SnapshotId / Snapshot. Let's make it `models`.

izeigerman (approved):

PR #3841, opened 2025-02-14
fix: signals that return an empty list are considered ready

plaflamme:
This fixes the test which was not asserting, adds a test to cover this case (which fails) and fixes the actual problem by de-denting a line (the line is skipped when the returned list is empty because it skips the iteration).


tobymao (approved):
thank you

PR #3842, opened 2025-02-14
feat: improve gcp postgres connection config options

eakmanrq:
Prior to this PR, it was assumed that customers would be connecting to GCP Postgres using the public IP. This is most likely the case but they could also connect using private if on a VPN. 

Also noticed that `timeout` was an argument that could be passed into the GCP Postgres connector class but we didn't include it and just passed into the resulting connection class. Also noticed some config options weren't documented or tested which was also fixed.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -987,6 +987,7 @@ class GCPPostgresConnectionConfig(ConnectionConfig):
     password: t.Optional[str] = None
     enable_iam_auth: t.Optional[bool] = None
     db: str
+    ip_type: t.Union[t.Literal["public"], t.Literal["private"], t.Literal["psc"]] = "public"
     # Keyfile Auth
     keyfile: t.Optional[str] = None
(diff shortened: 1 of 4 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -1052,7 +1053,15 @@ def _connection_factory(self) -> t.Callable:
...
             )
 
-        return Connector(credentials=creds).connect
+        kwargs = {
+            "credentials": creds,
+            "ip_type": self.ip_type,
+        }
+
+        if self.timeout:
+            kwargs["timeout"] = self.timeout
```
izeigerman, line 1062: Should we remove it from `_connection_kwargs_keys` then?

eakmanrq: My assumption was that it applies then to both. So we are passing it into the GCP connector class and then the connector that is created for pg8000. I'm not sure how the two interact but this at least keeps them aligned.

izeigerman (approved):

julian-purelei:
❤️

PR #3844, opened 2025-02-14
Fix: Inference of python model names from the file system

izeigerman:
The issue is that when name inference is enabled, all Python models receive the same name in the model registry, causing only one model to be picked up. The solution is to infer the name before registering the model in the registry.


georgesittas (approved):

ebremst3dt:
👏 👏 

Well done!

PR #3848, opened 2025-02-15
Feat!: add model blueprinting

georgesittas:
This PR adds a new feature: the ability to implement model blueprints based on a single template.

This feature is useful when multiple models need to share the same logic, but parameterized differently. Today, this is generally possible by either duplicating model definitions, thus violating the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), or creating a custom loader, which involves knowing the inner workings of the codebase.

A simple example that demonstrates the syntax to achieve model blueprinting is shown below:

```sql
MODEL (
  name @blueprint.some_table,
  blueprints (
    (blueprint := customer1),
    (blueprint := customer2),
    (blueprint := customer3),
  ),
);

SELECT
  *
FROM @blueprint.some_source
```

This definition essentially expands into three different models, each one sharing the same logic but using a different value for the `@blueprint` variable. A blueprint can be further customized by specifying additional variables:

```sql
MODEL (
  name @customer.some_table,
  blueprints (
    (customer := customer1, field_a := x, field_b := y),
    (customer := customer2, field_a := z, field_b := w),
  )
);

SELECT
  @field_a,
  @field_b
FROM @customer.some_source
```

Similarly, this will expand into two different models, where each one will contain the variables specified in the corresponding `blueprints` tuple. For example, the first model would look like:

```sql
MODEL (
  name customer1.some_table,
);

SELECT
  x,
  y
FROM customer1.some_source
```

The blueprint variable mappings can be combined with custom macros to implement even more complicated patterns, e.g., where the model logic itself changes completely depending on a variable's value.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -548,8 +548,6 @@ def _create_parser(expression_type: t.Type[exp.Expression], table_keys: t.List[s
         from sqlmesh.core.model.kind import ModelKindName
 
         expressions: t.List[exp.Expression] = []
-        gateway: t.Optional[exp.Expression] = None
-        blueprints: t.List[exp.Expression] = []
 
         while True:
             prev_property = seq_get(expressions, -1)
@@ -613,36 +611,12 @@ def _create_parser(expression_type: t.Type[exp.Expression], table_keys: t.List[s
             else:
                 value = self._parse_bracket(self._parse_field(any_token=True))
 
-                if key == "gateway":
-                    gateway = value
-                elif key == "blueprints":
-                    if isinstance(value, exp.Paren):
-                        blueprints = [value.this]
-                    elif isinstance(value, (exp.Tuple, exp.Array)):
-                        blueprints = value.expressions
-                    else:
-                        raise ConfigError(
-                            "The 'blueprints' values need to be enclosed in "
-                            f"parentheses or brackets, got {value} instead."
-                        )
-
-                    # We don't want to include blueprints in the property list
-                    continue
-
             if isinstance(value, exp.Expression):
                 value.meta["sql"] = self._find_sql(start, self._prev)
 
             expressions.append(self.expression(exp.Property, this=key, value=value))
 
-        expression = self.expression(expression_type, expressions=expressions)
-
-        # We store these properties in the meta to provide quick access at load time
-        if blueprints:
-            expression.meta["blueprints"] = blueprints
-            if gateway:
-                expression.meta["gateway"] = gateway
-
-        return expression
+        return self.expression(expression_type, expressions=expressions)
 
     return parse
 
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -543,11 +543,13 @@ def _parse_if(self: Parser) -> t.Optional[exp.Expression]:
         return exp.Anonymous(this="IF", expressions=[cond, stmt])
 
 
-def _create_parser(parser_type: t.Type[exp.Expression], table_keys: t.List[str]) -> t.Callable:
+def _create_parser(expression_type: t.Type[exp.Expression], table_keys: t.List[str]) -> t.Callable:
     def parse(self: Parser) -> t.Optional[exp.Expression]:
         from sqlmesh.core.model.kind import ModelKindName
 
         expressions: t.List[exp.Expression] = []
+        gateway: t.Optional[exp.Expression] = None
+        blueprints: t.List[exp.Expression] = []
 
         while True:
             prev_property = seq_get(expressions, -1)
@@ -611,12 +613,36 @@ def _create_parser(parser_type: t.Type[exp.Expression], table_keys: t.List[str])
             else:
                 value = self._parse_bracket(self._parse_field(any_token=True))
 
+                if key == "gateway":
+                    gateway = value
+                elif key == "blueprints":
+                    if isinstance(value, exp.Paren):
+                        blueprints = [value.this]
+                    elif isinstance(value, (exp.Tuple, exp.Array)):
+                        blueprints = value.expressions
+                    else:
+                        raise ConfigError(
+                            "The 'blueprints' values need to be enclosed in "
+                            f"parentheses or brackets, got {value} instead."
+                        )
+
+                    # We don't want to include blueprints in the property list
+                    continue
+
             if isinstance(value, exp.Expression):
                 value.meta["sql"] = self._find_sql(start, self._prev)
 
             expressions.append(self.expression(exp.Property, this=key, value=value))
 
-        return self.expression(parser_type, expressions=expressions)
+        expression = self.expression(expression_type, expressions=expressions)
+
+        # We store these properties in the meta to provide quick access at load time
+        if blueprints:
+            expression.meta["blueprints"] = blueprints
+            if gateway:
+                expression.meta["gateway"] = gateway
+
+        return expression
 
     return parse
 
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -10,14 +10,12 @@ from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
 
-from sqlglot import exp
 from sqlglot.errors import SqlglotError
-from sqlglot.helper import seq_get
 
 from sqlmesh.core import constants as c
 from sqlmesh.core.audit import Audit, ModelAudit, StandaloneAudit, load_multiple_audits
-from sqlmesh.core.dialect import parse, Model as ModelMeta
-from sqlmesh.core.macros import MacroRegistry, MacroVar, macro
+from sqlmesh.core.dialect import parse
+from sqlmesh.core.macros import MacroRegistry, macro
 from sqlmesh.core.metric import Metric, MetricMeta, expand_metrics, load_metric_ddl
 from sqlmesh.core.model import (
     Model,
@@ -25,7 +23,7 @@ from sqlmesh.core.model import (
     ModelCache,
     SeedModel,
     create_external_model,
-    load_sql_based_model,
+    load_sql_based_models,
 )
 from sqlmesh.core.model import model as model_registry
 from sqlmesh.core.signal import signal
@@ -294,7 +292,9 @@ class Loader(abc.ABC):
         """Project file to track for modifications"""
         self._path_mtimes[path] = path.stat().st_mtime
 
-    def _get_variables(self, gateway_name: str) -> t.Dict[str, t.Any]:
+    def _get_variables(self, gateway_name: t.Optional[str] = None) -> t.Dict[str, t.Any]:
+        gateway_name = gateway_name or self.context.selected_gateway
+
         if gateway_name not in self._variables_by_gateway:
             try:
                 gateway = self.config.get_gateway(gateway_name)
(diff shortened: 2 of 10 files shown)


sqlmesh/core/dialect.py
```diff
@@ -611,12 +613,36 @@ def parse(self: Parser) -> t.Optional[exp.Expression]:
...
                 value.meta["sql"] = self._find_sql(start, self._prev)
 
             expressions.append(self.expression(exp.Property, this=key, value=value))
 
-        return self.expression(parser_type, expressions=expressions)
+        expression = self.expression(expression_type, expressions=expressions)
+
+        # We store these properties in the meta to provide quick access at load time
+        if blueprints:
+            expression.meta["blueprints"] = blueprints
```
izeigerman, line 641: Not sure I like using the meta for this. It's pretty non-obvious and implicit. Plus it introduces even more coupling between sql parsing and model definition in a pretty opaque way. I suggest handling this in the `load_sql_based_model` instead.

sqlmesh/core/loader.py
```diff
@@ -404,32 +411,64 @@ def _load() -> Model:
...
-                    default_catalog=self.context.default_catalog,
-                    variables=self._variables,
-                    infer_names=self.config.model_naming.infer_names,
-                    signal_definitions=signals,
-                )
+                model_meta = seq_get(expressions, 0)
+                blueprints = isinstance(model_meta, ModelMeta) and model_meta.meta.get("blueprints")
+
+                loaded_models = []
+                for blueprint in blueprints or [None]:
```
izeigerman, line 418: This again leaks the details of model handling to the loader. I don't think the loader should worry about this. `load_sql_based_model` should encapsulate this logic instead.

sqlmesh/core/loader.py
```diff
@@ -461,23 +500,46 @@ def _load_python_models(
...
-                        audit_definitions=audits,
-                    )
-                    if model.enabled:
-                        models[model.fqn] = model
+                    registered_entrypoint = registry[name]
+
+                    gateway = registered_entrypoint.kwargs.get("gateway") or ""
+                    blueprints = registered_entrypoint.kwargs.pop("blueprints", None)
+
+                    for blueprint in blueprints or [None]:
```
izeigerman, line 508: We need to consolidate the logic that parses blueprints with the one above.

sqlmesh/core/loader.py
```diff
@@ -404,32 +411,64 @@ def _load() -> Model:
...
+                model_meta = seq_get(expressions, 0)
+                blueprints = isinstance(model_meta, ModelMeta) and model_meta.meta.get("blueprints")
+
+                loaded_models = []
+                for blueprint in blueprints or [None]:
+                    if not blueprint:
+                        blueprint_variables = {}
+                    elif isinstance(blueprint, (exp.Tuple, exp.Array)):
+                        blueprint_variables = {
+                            # Assumes the syntax: (k := v [, ...]) or [k := v [, ...]]
```
izeigerman, line 423: What is this syntax?

izeigerman: We need to discuss this. I don't like how inconsistent the 2 alternative ways of doing this are. I think we should stick with either one or the other. The key value one looks not bad

sqlmesh/core/loader.py
```diff
@@ -307,6 +294,26 @@ def _track_file(self, path: Path) -> None:
...
 
+    def _get_variables(self, gateway_name: str) -> t.Dict[str, t.Any]:
+        if gateway_name not in self._variables_by_gateway:
+            try:
+                gateway = self.config.get_gateway(gateway_name)
+            except ConfigError:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
+                    f"Gateway '{gateway_name}' not found in project '{self.config.project}'."
```
tobymao, line 305: why allow this? why not just raise?

sqlmesh/core/loader.py
tobymao: there is no need for the blueprint syntax, we only need the kv syntax, it's strictly better

sqlmesh/core/loader.py
georgesittas: This logic exited before this PR, I just moved it into a helper. I'm not sure what the original motivation was.

sqlmesh/core/loader.py
georgesittas: I guess the advantage of using `blueprints (a, b, c, d)` is that it's more concise. The alternative would be: `blueprints ((blueprint := a), (blueprint := b), (blueprint := c), (blueprint := d))`, which is kinda repetitive.

sqlmesh/core/loader.py
georgesittas: Implemented a `load_sql_based_models` helper and refactored the previous approach by pushing the logic to the `model/definition.py` module. Please take a look and let me know if you had something else in mind.

sqlmesh/core/loader.py
georgesittas: Done, both SQL and Python model blueprinting now happens within `create_model_blueprints`.

sqlmesh/core/loader.py
georgesittas: Got rid of the `blueprints (a, b, c, d)` syntax. Only the key-value one is supported now.

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+            variables = blueprint
+        else:
+            # TODO: put docs link here
+            raise ConfigError(
+                f"Invalid blueprint value: {blueprint} at '{path}', please refer to <link>."
+            )
+
+        if isinstance(gateway, d.MacroVar):
+            gateway_name = variables.get(gateway.name)
+        elif isinstance(gateway, str) and gateway.startswith("@"):
```
tobymao, line 1829: why is gateway expected to be @? can someone pass in a normal gateway?

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+    gateway: t.Optional[str | exp.Expression],
+    blueprints: t.Any,
+    get_variables: t.Callable[[t.Optional[str]], t.Dict[str, str]],
+    loader: t.Callable[..., Model],
+    **loader_kwargs: t.Any,
+) -> t.List[Model]:
+    path = loader_kwargs.get("path")
+    dialect = loader_kwargs.get("dialect")
+
+    if not blueprints:
```
izeigerman, line 1798: This entire block can be a separate function dedicated to resolving the blueprint value

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
     blocking: bool = True
 
 
+def create_model_blueprints(
```
izeigerman, line 1788: This is more like `create_models_from_blueprints` rather that creating blueprints

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+            raise ConfigError(
+                f"Invalid blueprint value: {blueprint} at '{path}', please refer to <link>."
+            )
+
+        if isinstance(gateway, d.MacroVar):
+            gateway_name = variables.get(gateway.name)
+        elif isinstance(gateway, str) and gateway.startswith("@"):
+            gateway_name = variables.get(gateway[1:])
+        else:
+            gateway_name = None
```
izeigerman, line 1832: What if the gateway is a macro call and not a variable?

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+        blueprints = blueprints.expressions
+    elif not isinstance(blueprints, list):
+        # TODO: put docs link here
+        raise ConfigError(
+            f"Invalid 'blueprints' property '{blueprints}' at '{path}', please refer to <link>."
+        )
+
+    model_blueprints: t.List[Model] = []
+    for blueprint in blueprints:
+        if not blueprint:
```
izeigerman, line 1812: This also looks like something that belongs to a function

sqlmesh/core/model/definition.py
georgesittas: They can, this is only supposed to handle the case where someone does `gateway @blueprint_value`, in order to fetch the variables of the corresponding gateway.

sqlmesh/core/model/definition.py
georgesittas: Good question. I guess that would result in a failed lookup because of the `(` and `)` in the call text, granted though that this is somewhat implicit. Other than that, I think it should cover the use case we're interested in, i.e. rendering blueprint variables.

sqlmesh/core/model/definition.py
georgesittas: Yep, the body just seemed small/simple enough to me, so I avoided creating more helpers for this. Happy either way, let me know what's your preference

sqlmesh/core/model/definition.py
georgesittas: I guess my other [reply](https://github.com/TobikoData/sqlmesh/pull/3848#discussion_r1959518379) applies here as well.

sqlmesh/core/model/definition.py
georgesittas: I guess we want this logic to kick in anyway so that the appropriate variables are passed to the model?

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+    blueprints: t.Any,
+    get_variables: t.Callable[[t.Optional[str]], t.Dict[str, str]],
+    loader: t.Callable[..., Model],
+    **loader_kwargs: t.Any,
+) -> t.List[Model]:
+    path = loader_kwargs.get("path")
+    dialect = loader_kwargs.get("dialect")
+
+    if not blueprints:
+        blueprints = [None]
```
tobymao, line 1799: why do you do blueprints = [None]? the iteration of not blueprint -> variables = {} seems a bit hacky

sqlmesh/core/model/definition.py
tobymao: can gateway be a function call? @get_gateway()

georgesittas: Yes, that is valid today. It's also valid in this PR, it simply won't be picked up in the `variables` dict lookup, see other [thread](https://github.com/TobikoData/sqlmesh/pull/3848#discussion_r1959510837).

sqlmesh/core/model/definition.py
georgesittas: It's a simple way to handle both blueprints being present in the meta and not.

sqlmesh/core/model/definition.py
```diff
@@ -1784,6 +1785,84 @@ class AuditResult(PydanticModel):
...
+    gateway: t.Optional[exp.Expression] = None
+    blueprints: t.Optional[t.List[t.Optional[exp.Expression]]] = None
+
+    model_meta = seq_get(expressions, 0)
+    for prop in (isinstance(model_meta, d.Model) and model_meta.expressions) or []:
+        if prop.name == "gateway":
+            gateway = prop.args["value"]
+        elif prop.name == "blueprints":
+            # We pop the blueprints property here because it shouldn't be part of the model
+            blueprints = prop.pop().args["value"]
```
tobymao, line 1855: does the mutate the ast node? seems a bit off

sqlmesh/core/model/definition.py
izeigerman: I think this will make the method simpler to comprehend. The purpose of this method is to get the list of blueprints and iterate over them. Instead it:
1. Parses blueprints
2. Extracts blueprint variables
3. Parses gateway
4. Iterates over blueprints and loads models.

I think it should only be doing (4) with everything else delegated to subroutines.

georgesittas: Got it, sounds good.

sqlmesh/core/model/definition.py
tobymao: should that be done in the meta loader and not popping the ast? is the best place?

tobymao (approved):

sqlmesh/core/model/definition.py
georgesittas: can look into that and see what it looks like 👍

georgesittas: Moved it to `_create_model` right before validating the model's definition. Does that work for you?

sqlmesh/core/model/definition.py
georgesittas: Addressed, please see updated logic & test.

sqlmesh/core/model/definition.py
georgesittas: Addressed. Refactored both this section and the blueprint variable extraction out into helpers.

sqlmesh/core/model/definition.py
tobymao: yea better

sqlmesh/core/model/definition.py
tobymao: looks good

serkef:
omg this feature is AMAZING 💯 
Will try it soon

PR #3849, opened 2025-02-15
fix!: normalize catalog override name

eakmanrq:
Prior to this PR, if you were using Snowflake and the `environment_catalog_mapping` config it would result in lowercase catalog names. 

This PR updates the config to normalize according to the dialect. Tested against Snowflake and confirmed to be working as expected. 

Note: The PR doesn't attempt to address a potential issue if someone wanted to quote their catalog name. I think we are missing a `to_identifier` or `parse_identifier` somewhere but it is not the problem I fixing here.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -222,13 +222,20 @@ class QualifiedViewName(PydanticModel, frozen=True):
         return exp.table_(
             self.table_name_for_environment(environment_naming_info, dialect=dialect),
             db=self.schema_for_environment(environment_naming_info, dialect=dialect),
-            catalog=self.catalog_for_environment(environment_naming_info),
+            catalog=self.catalog_for_environment(environment_naming_info, dialect=dialect),
         )
 
     def catalog_for_environment(
-        self, environment_naming_info: EnvironmentNamingInfo
+        self, environment_naming_info: EnvironmentNamingInfo, dialect: DialectType = None
     ) -> t.Optional[str]:
-        return environment_naming_info.catalog_name_override or self.catalog
+        if environment_naming_info.catalog_name_override:
+            catalog_name = environment_naming_info.catalog_name_override
+            return (
+                normalize_identifiers(catalog_name, dialect=dialect).name
+                if environment_naming_info.normalize_name
+                else catalog_name
+            )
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -222,13 +222,20 @@ def table_for_environment(
...
     def catalog_for_environment(
-        self, environment_naming_info: EnvironmentNamingInfo
+        self, environment_naming_info: EnvironmentNamingInfo, dialect: DialectType = None
     ) -> t.Optional[str]:
-        return environment_naming_info.catalog_name_override or self.catalog
+        if environment_naming_info.catalog_name_override:
+            catalog_name = environment_naming_info.catalog_name_override
+            return (
+                normalize_identifiers(catalog_name, dialect=dialect).name
+                if environment_naming_info.normalize_name
```
izeigerman, line 235: Not sure I understand why the catalog is affected by the **environment name** normalization rules. To me the two are orthogonal

eakmanrq: The environment here is representing the virtual environment. What affects the catalog is the `environment_catalog_mapping`. So this all seems related to me. If we are saying the environment needs to be normalized then that includes the full environment path which includes the catalog. It was a bug before that catalog was unaffected.

georgesittas:
> Note: The PR doesn't attempt to address a potential issue if someone wanted to quote their catalog name. I think we are missing a to_identifier or parse_identifier somewhere but it is not the problem I fixing here.

Did you verify this issue exists?

sqlmesh/core/snapshot/definition.py
georgesittas: I think it makes sense: if a user does `'^dev.*': dev`, then I think it's reasonable to expect that using an unquoted `dev` in their queries' FQNs would work => we should normalize as uppercase. If it's in lower case, this won't work.

georgesittas (approved):

eakmanrq:
> > Note: The PR doesn't attempt to address a potential issue if someone wanted to quote their catalog name. I think we are missing a to_identifier or parse_identifier somewhere but it is not the problem I fixing here.
> 
> Did you verify this issue exists?

No just looking at the code it seemed like it could.

PR #3850, opened 2025-02-16
Fix: Warn when SQLMesh automatically adjusts a restatement range to cover the whole model

erindru:
Prior to this change, if you tried to restate eg 1 day of a model that is [full_history_restatement_only](https://github.com/TobikoData/sqlmesh/blob/ab6b2502549d12d2a23d4af2d976c180ca6ef1f7/sqlmesh/core/model/kind.py#L131), for example an `INCREMENTAL_BY_UNIQUE_KEY` model:

```
$ sqlmesh plan --restate-model sqlmesh_example.unique_key_model --start 2020-01-01 --end 2020-01-02

No changes to plan: project files match the `prod` environment

Models needing backfill:
└── sqlmesh_example.unique_key_model: [2020-01-01 - 2020-01-10]
```

SQLMesh would automatically adjust the backfill intervals to cover the entire model without making it obvious to the user.

Even if the user notices that the end date in this example was changed to `2020-01-10` they may be confused as to _why_ SQLMesh ignored their specified end date of `2020-01-02`.

This PR adds a warning like so:
```
$ sqlmesh plan --restate-model sqlmesh_example.unique_key_model --start 2020-01-01 --end 2020-01-02

[WARNING] Model 'sqlmesh_example.unique_key_model' is 'INCREMENTAL_BY_UNIQUE_KEY' which does not support partial restatement.
Expanding the requested restatement intervals from [2020-01-01 - 2020-01-02] to [2020-01-01 - 2020-01-10] in order to fully restate the model.

No changes to plan: project files match the `prod` environment

Models needing backfill:
└── sqlmesh_example.unique_key_model: [2020-01-01 - 2020-01-10]
```

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -715,9 +715,29 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
                 a run.
         """
         end = execution_time or now_timestamp() if self.depends_on_past else end
+        removal_interval = self.inclusive_exclusive(start, end, strict)
+
         if not is_preview and self.full_history_restatement_only and self.intervals:
-            start = self.intervals[0][0]
-        return self.inclusive_exclusive(start, end, strict)
+            expanded_removal_interval = self.inclusive_exclusive(self.intervals[0][0], end, strict)
+            requested_start, requested_end = removal_interval
+            expanded_start, expanded_end = expanded_removal_interval
+
+            is_subset = requested_start > expanded_start or requested_end < expanded_end
+
+            # only warn if the requested removal interval was a subset of the actual model intervals and was automatically expanded
+            # if the requested interval was the same or wider than the actual model intervals, no need to warn
+            if removal_interval != expanded_removal_interval and is_subset:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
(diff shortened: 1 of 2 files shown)


erindru:
The test failures are related to https://github.com/TobikoData/sqlmesh/pull/3851

sqlmesh/core/snapshot/definition.py
```diff
@@ -715,9 +715,29 @@ def get_removal_interval(
...
-        return self.inclusive_exclusive(start, end, strict)
+            expanded_removal_interval = self.inclusive_exclusive(self.intervals[0][0], end, strict)
+            requested_start, requested_end = removal_interval
+            expanded_start, expanded_end = expanded_removal_interval
+
+            is_subset = requested_start > expanded_start or requested_end < expanded_end
+
+            # only warn if the requested removal interval was a subset of the actual model intervals and was automatically expanded
+            # if the requested interval was the same or wider than the actual model intervals, no need to warn
+            if removal_interval != expanded_removal_interval and is_subset:
```
georgesittas, line 729: Doesn't `is_subset` imply `remove_interval != expanded_removal_interval`? Just wondering if we can simplify this condition.

georgesittas (approved):

sqlmesh/core/snapshot/definition.py
erindru: Yeah, I guess it does. I've simplified the condition, thanks!

PR #3852, opened 2025-02-17
fix: stop from get loaded context in async

benfdking:
**Problem**

In enterprise, we are seeing the above error with tasks. 

image

**Solution**

Move the work out of the separate executor which solves the above conflict. 

The problem with that change alone is that it make the call slower because in async await context it's now blocking. To avoid that I've made the dependency injection sync which means FastAPI runs it in itsthreadpool, which means it doesn't slow other calls down. Once thing that needed to be done is to then move the lock protecting the work into a thread.

### DIFF
--- a/web/server/settings.py
+++ b/web/server/settings.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
-import asyncio
+import contextlib
 import logging
 import os
+import threading
 import typing as t
-from contextlib import asynccontextmanager
 from functools import lru_cache
 from pathlib import Path
 
@@ -19,7 +19,8 @@ from web.server.exceptions import ApiException
 from web.server.models import Mode
 
 logger = logging.getLogger(__name__)
-get_context_lock = asyncio.Lock()
+
+get_loaded_context_lock = threading.Lock()
 
 MODE_TO_MODULES = {
     models.Mode.IDE: {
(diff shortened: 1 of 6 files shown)


web/server/settings.py
```diff
@@ -73,7 +73,10 @@ def _get_context(path: str | Path, config: str, gateway: str) -> Context:
 @lru_cache()
 def _get_loaded_context(path: str | Path, config: str, gateway: str) -> Context:
     context = _get_context(path, config, gateway)
-    context.load()
+    try:
+        context.load()
+    except Exception:
+        logger.exception(f"Error loading context for {path}")
```
mykalmax, line 79: if loading failed this won't send any errors to UI

izeigerman (approved):

tobymao:
should we merge this?

benfdking:
@tobymao My understanding is yes but I am yet to get a good answer on why we introduced in the first place and yet to dig into whether the asyncio lock provides value. 

It does fix sqlmesh UI upstream in enterprise.

web/server/watcher.py
```diff
@@ -21,7 +21,7 @@
 
 async def watch_project() -> None:
     settings = get_settings()
-    context = await get_context(settings)
+    context = get_context(settings)
```
izeigerman, line 24: This function is still async

izeigerman: So we need to call this in a threadpool / executor

izeigerman: or make it non-async

PR #3856, opened 2025-02-18
Chore: fix audit doc typos

georgesittas:
We do `foo_audit = ModelAudit("foo", ...)` in `builtin.py`, so my guess is the variable name was copied over inadvertently.

### DIFF
--- a/docs/concepts/audits.md
+++ b/docs/concepts/audits.md
@@ -378,7 +378,7 @@ MODEL (
 );
 ```
 
-#### string_length_equal_audit, string_length_equal_audit_non_blocking
+#### string_length_equal, string_length_equal_non_blocking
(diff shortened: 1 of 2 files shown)


themisvaltinos (commented):
Maybe fixing the comments with the examples of the builtin.py file as well for consistency:

https://github.com/TobikoData/sqlmesh/blob/8327fefdfac6bd4affe58c5d71f8e9b72a5f9864/sqlmesh/core/audit/builtin.py#L439

georgesittas:
Ah, yes, I didn't notice we did this in the comments as well. Thanks for catching! Pushing a followup soon.

themisvaltinos (approved):

PR #3858, opened 2025-02-18
Feat!: Add support for project before_all and after_all statements

### LINKED ISSUE #3259: On run end
Add feature similar to dbt's `on-run-end` and `on-run-start` functionality.
https://docs.getdbt.com/reference/project-configs/on-run-start-on-run-end

I think ideally it would allow some way to run macros or macro-enabled sql at the start and/or end of `sqlmesh plan` or `sqlmesh run`. 
The most common use case to run any set up or wrap up code that is not specific to a single model. 

E.g. granting usage to schemas.
UDFs used in multiple models.

themisvaltinos:
This update adds support for `before_all` and `after_all` statements to be executed at the start and end of the `sqlmesh plan` and `run` commands.

These statements can be defined in the configuration file under the `before_all`, `after_all` keys as SQL statements or alternatively using sqlmesh macros:

```yaml
before_all:
  - CREATE TABLE IF NOT EXISTS analytics (table VARCHAR, eval_time VARCHAR)
after_all:
  - "@grant_select_privileges()"
  - "@grant_schema_usage()"
```

The introduction of these statements allows for actions that need to be run before all individual model statements execute or after all respectively and also can help simplify tasks such as granting privileges on the physical tables or views ( fixes #3259 )

For example instead of using an `on_virtual_update` statement in each model to grant privileges on the views of the virtual layer, this can be reduced to a single macro to be executed at the end of the plan:

```python
from sqlmesh.core.macros import macro
from sqlmesh.core.snapshot.definition import to_view_mapping

@macro()
def grant_select_privileges(evaluator):
    if evaluator._environment_naming_info:
        mapping = to_view_mapping(
            evaluator._snapshots.values(), evaluator._environment_naming_info
        )
        return [
            f"GRANT SELECT ON VIEW {view_name} TO ROLE admin_role;"
	    for view_name in mapping.values()
        ]
```

Similarly, to grant schema privileges you can define a macro as follows:

```python
from sqlmesh import macro

@macro()
def grant_schema_usage(evaluator):
    if evaluator._environment_naming_info:
        schemas = {
            snapshot.qualified_view_name.schema_for_environment(
                evaluator._environment_naming_info
            )
            for snapshot in evaluator._snapshots.values()
            if snapshot.is_model
        }
        return [
            f"GRANT USAGE ON SCHEMA {schema} TO admin_role;"
            for schema in schemas
        ]
```


themisvaltinos:
> Would there be a way to restrict actions to certain virtual environments? For example, if you want to apply select privileges to views in the virtual layer but only when `plan` or `run` is executed in prod environment.

Yes you can definitely achieve this by adding a condition in the examples I provided to check if the environment is `prod` so that select privileges would only be applied when this condition is met.

As part of this PR, I exposed the environment naming info in the macro evaluator and added the `this_env` macro to provide the environment name, which should assist with more advanced use cases that require control over these statements' behaviour.

sqlmesh/core/config/plan.py
```diff
@@ -30,3 +30,5 @@ class PlanConfig(BaseConfig):
     no_prompts: bool = True
     auto_apply: bool = False
     use_finalized_state: bool = False
+    before_all: t.Optional[t.List[str]] = None
```
izeigerman, line 33: Why only plan? How about run? Should this be a top-level config instead?

sqlmesh/core/loader.py
```diff
@@ -594,6 +626,25 @@ def _load_metrics(self) -> UniqueKeyDict[str, MetricMeta]:
...
+            "before_all": self.config.plan.before_all or [],
+            "after_all": self.config.plan.after_all or [],
+        }
+
+        python_env = make_python_env(
+            [exp.maybe_parse(stmt) for stmts in statements.values() for stmt in stmts],
+            module_path=self.config_path,
+            jinja_macro_references=None,
+            macros=macros,
+            variables=self._get_variables(),
```
izeigerman, line 642: Why aren't we taking a gateway into account?

izeigerman: Never mind. I see that we fallback to a selected gateway if not specified

sqlmesh/core/plan/evaluator.py
```diff
@@ -120,6 +120,18 @@ def evaluate(
                 after_promote_snapshots = all_names - before_promote_snapshots
                 deployability_index_for_evaluation = DeployabilityIndex.all_deployable()
 
+            self.state_sync.update_project_statements(plan)
+            self.snapshot_evaluator._execute_project_statements(
```
izeigerman, line 124: We should not be calling private methods

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1102,6 +1105,37 @@ def _execute_create(
         )
         adapter.execute(snapshot.model.render_post_statements(**create_render_kwargs))
 
+    def _execute_project_statements(
```
izeigerman, line 1108: Why is this in the snapshot evaluator? It has nothing to do with individual snapshots

sqlmesh/core/loader.py
```diff
@@ -40,6 +43,23 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class ProjectStatements:
```
izeigerman, line 47: I don't think this belongs to the loader. Isn't this `EnvironmentStatements`? In which case this belongs to `environment.py`.

sqlmesh/core/state_sync/base.py
```diff
@@ -177,6 +179,22 @@ def update_auto_restatements(
...
+    @abc.abstractmethod
+    def get_project_statements(self, environment: str) -> t.List[ProjectStatements]:
+        """Fetches project statements from the project_statements table.
+
+        Returns:
+            A tuple of (before_all, after_all) statements.
+        """
+
+    @abc.abstractmethod
+    def update_project_statements(self, plan: EvaluatablePlan) -> None:
```
izeigerman, line 191: We should only pass what we are storing and not the whole plan.

sqlmesh/core/config/plan.py
themisvaltinos: Right I had it under PlanConfig, since in my initial iteration they were separate statement between the plan and run commands. But since we'll be using the same statements for both, it makes more sense to move them to top level

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: yes I was skeptical about if I should have this here as well i will move this

sqlmesh/core/loader.py
themisvaltinos: right will rename this and move all relating logic to environment.py so it's consistent

themisvaltinos:
Thanks for the review @izeigerman addressed comments if you want to take another look

sqlmesh/core/environment.py
```diff
@@ -118,6 +126,7 @@ class Environment(EnvironmentNamingInfo):
         default=None, alias="previous_finalized_snapshots"
     )
     requirements: t.Dict[str, str] = {}
+    statements: t.Optional[t.List[t.Any]] = None
```
izeigerman, line 129: Btw, making statements a part of the `Environment` instance is not what I meant last time. I meant extending the `promote` method with the additional `statements` argument. 

The problem with the existing approach is that if I construct the environment object with statements, I expect to fetch it with statements too. Which is not what happens today.

sqlmesh/migrations/v0071_add_environment_statements.py
```diff
@@ -0,0 +1,27 @@
+"""Add the environment statements table."""
```
izeigerman, line 1: Please note that `v0071` has been merged already

themisvaltinos: Thank you yes I updated it locally to v0072 will push this along with the latest changes

sqlmesh/core/context_diff.py
```diff
@@ -258,12 +271,24 @@ def has_changes(self) -> bool:
...
+            or self.has_environment_statements_changes
         )
 
     @property
     def has_requirement_changes(self) -> bool:
         return self.previous_requirements != self.requirements
 
+    @property
+    def has_environment_statements_changes(self) -> bool:
+        if (
```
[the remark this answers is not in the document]
themisvaltinos: If there are any differences in the previous environment statements that exist in the state and the new in the project. But the first conditional is simply to cover the case that there are no environment statements in the state table, and the project doesn't have environment statements as well so it doesn't pick it up as having changes. I should probably simplify this check though

sqlmesh/migrations/v0071_add_environment_statements.py
georgesittas: cc @VaggelisD let's make sure to coordinate, I recall you also had a migration script in your linter PR.

sqlmesh/core/environment.py
themisvaltinos: I understand now, I updated the code to extend the `promote` method instead and pass in the statements to update the corresponding ones to the environment

sqlmesh/core/environment.py
```diff
@@ -11,12 +13,18 @@
...
 from sqlmesh.utils import word_characters_only
 from sqlmesh.utils.date import TimeLike, now_timestamp
+from sqlmesh.utils.metaprogramming import Executable
 from sqlmesh.utils.pydantic import PydanticModel, field_validator
 
 T = t.TypeVar("T", bound="EnvironmentNamingInfo")
 PydanticType = t.TypeVar("PydanticType", bound="PydanticModel")
 
 
+class ExecutionStage(Enum):
```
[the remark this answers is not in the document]
themisvaltinos: Yes good point, removed it and added these to `RuntimeStage`

sqlmesh/core/state_sync/base.py
```diff
@@ -345,6 +353,7 @@ def promote(
         self,
         environment: Environment,
         no_gaps_snapshot_names: t.Optional[t.Set[str]] = None,
+        environment_statements: t.Optional[str] = None,
```
izeigerman, line 356: Why is this a string? What happened to the type?

sqlmesh/core/plan/evaluator.py
```diff
@@ -286,6 +321,7 @@ def _promote(
         promotion_result = self.state_sync.promote(
             plan.environment,
             no_gaps_snapshot_names=no_gaps_snapshot_names if plan.no_gaps else set(),
+            environment_statements=plan.json(include={"environment_statements"}),
```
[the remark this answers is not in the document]
izeigerman: Why are we not passing the original type?

sqlmesh/core/state_sync/base.py
themisvaltinos: I was thinking of doing the serialisation before passing it here, will move it instead to the state sync

sqlmesh/core/plan/evaluator.py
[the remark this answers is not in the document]
themisvaltinos: right will revise this

sqlmesh/core/scheduler.py
```diff
@@ -470,8 +492,53 @@ def evaluate_node(node: SchedulingUnit) -> None:
...
+        runtime_stage: RuntimeStage,
+        environment_naming_info: EnvironmentNamingInfo,
+        snapshots: t.Optional[t.Dict[str, Snapshot]] = None,
+        start: t.Optional[TimeLike] = None,
+        end: t.Optional[TimeLike] = None,
+        execution_time: t.Optional[TimeLike] = None,
+    ) -> None:
+        adapter = self.snapshot_evaluator.adapter
+        if (
+            environment_statements := self.state_sync.get_environment_statements(
```
izeigerman, line 518: Does this mean that we fetch the same thing twice?

sqlmesh/core/scheduler.py
```diff
@@ -470,8 +492,53 @@ def evaluate_node(node: SchedulingUnit) -> None:
...
+                    environment_naming_info=environment_naming_info,
+                    snapshots=snapshots_by_name,
+                    start=start,
+                    end=end,
+                    execution_time=execution_time,
+                )
+
             self.state_sync.recycle()
 
+    def execute_environment_statements(
```
izeigerman, line 507: I feel like this function is a more generalized version of the same function from `plan/evaluator.py`. Can we make it generic and reuse it in the plan evaluator? The only missing piece here is passing statements as an argument instead of fetching it here in place.

sqlmesh/core/scheduler.py
themisvaltinos: yes I should move it to do it once before

sqlmesh/core/scheduler.py
themisvaltinos: yes had it like that as one function before when I had it in the snapshot evaluator to reuse it for both the plan evaluator and the scheduler. I can make it one generic function but were should I keep it in the `scheduler` and then import it in the ` plan/evaluator`?

themisvaltinos:
Thanks @izeigerman addressed comments and revised code to align with the new state sync structure

sqlmesh/core/scheduler.py
izeigerman: To me this looks like something that belongs to `environment.py` but I don't feel strongly.

sqlmesh/core/environment.py
```diff
@@ -208,3 +211,27 @@ def _convert_list_to_dicts(self, value: t.Optional[t.List[t.Any]]) -> t.List[dic
         if not value:
             return []
         return value if isinstance(value[0], dict) else [v.dict() for v in value]
+
+
+@dataclass
+class EnvironmentStatements:
```
izeigerman, line 217: Why not use pydantic model?

sqlmesh/core/context.py
```diff
@@ -346,6 +346,7 @@ def __init__(
         self._metrics: UniqueKeyDict[str, Metric] = UniqueKeyDict("metrics")
         self._jinja_macros = JinjaMacroRegistry()
         self._requirements: t.Dict[str, str] = {}
+        self._environment_statements: t.List[EnvironmentStatements] = []
```
izeigerman, line 349: Why is this a list?

izeigerman: ah, multi project. I get it. But then should we always have one `EnvironmentStatements` and merge `before_all` and `afer_all` lists instead?

izeigerman: Ah, python_envs can be different. I get it too

sqlmesh/core/context_diff.py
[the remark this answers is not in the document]
izeigerman: > as well so it doesn't pick it up as having changes

I don't get it how come
```
return self.environment_statements != self.previous_environment_statements
```
doesn't cover this scenario.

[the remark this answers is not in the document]
izeigerman: > nd the project doesn't have environment statements 

if the project doesn't have them, but the state does have them, then we should sync the 2 and remove the statements right?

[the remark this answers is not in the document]
themisvaltinos: Yes, revised this to only have this condition 
```
return self.environment_statements != self.previous_environment_statements
```
(by not keeping empty environment statements in a project that doesn't contain one, which was why I was overcomplicating this condition before. I realise now this was a bad design decision)

>  if the project doesn't have them, but the state does have them, then we should sync the 2 and remove the statements from the target environment, right?
 
 right adapted the state sync to align the statements table when this happens and remove them

sqlmesh/core/scheduler.py
themisvaltinos: Moved it to the `environment.py`

sqlmesh/core/environment.py
themisvaltinos: Yes good point made it a pedantic model, this also removed the need to have a custom `parse_obj`

themisvaltinos:
thanks @izeigerman addressed the latest comments and updated docs

sqlmesh/core/environment.py
```diff
@@ -208,3 +212,47 @@ def _convert_list_to_dicts(self, value: t.Optional[t.List[t.Any]]) -> t.List[dic
...
             return []
         return value if isinstance(value[0], dict) else [v.dict() for v in value]
+
+
+class EnvironmentStatements(PydanticModel):
+    before_all: t.List[str]
+    after_all: t.List[str]
+    python_env: t.Dict[str, Executable]
+
+    def to_dict(self) -> t.Dict[str, t.Any]:
```
izeigerman, line 222: Why do we still needs this? Can't we use the pydantic model?

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -224,6 +225,10 @@ def promote(
             added_table_infos -= set(existing_environment.promoted_snapshots)
 
         self.environment_state.update_environment(environment)
+        if environment_statements is not None:
```
izeigerman, line 228: Shouldn't we still delete the existing record though?

sqlmesh/core/state_sync/db/environment.py
```diff
@@ -254,3 +334,19 @@ def _environment_to_df(environment: Environment) -> pd.DataFrame:
...
+def _environment_statements_to_df(
+    environment_name: str, plan_id: str, environment_statements: t.List[EnvironmentStatements]
+) -> pd.DataFrame:
+    return pd.DataFrame(
+        [
+            {
+                "environment_name": environment_name,
+                "plan_id": plan_id,
+                "environment_statements": json.dumps(
+                    {"environment_statements": [e.to_dict() for e in environment_statements]}
```
izeigerman, line 348: Why can't we just put the json list directly?

izeigerman (approved):
A few last comments for your consideration. Appreciate your patience

sqlmesh/core/state_sync/db/facade.py
themisvaltinos: Yes this is why I check for none, an empty list is passed when needing to delete the statements from the table. Updated to only delete in that case the existing record from the `_environment_statements` table and added the test `test_promote_environment_without_statements` to demonstrate this behaviour.

sqlmesh/core/state_sync/db/environment.py
themisvaltinos: Thanks, good point I’ve refactored this to pass the list and use pydantic dict() as you suggested too which simplified this

sqlmesh/core/context.py
```diff
@@ -1974,6 +1979,7 @@ def _run(
         select_models: t.Optional[t.Collection[str]],
         circuit_breaker: t.Optional[t.Callable[[], bool]],
         no_auto_upstream: bool,
+        is_run_command: bool = False,
```
izeigerman, line 1982: `_run` is only ever called by the `run` command. We don't need to pass this argument here.

sqlmesh/core/scheduler.py
```diff
@@ -246,6 +247,7 @@ def run(
         circuit_breaker: t.Optional[t.Callable[[], bool]] = None,
         deployability_index: t.Optional[DeployabilityIndex] = None,
         auto_restatement_enabled: bool = False,
+        is_run_command: bool = False,
```
izeigerman, line 250: The variable names feels like an abstraction leak. Shouldn't this be `run_environment_statements: bool`

PR #3863, opened 2025-02-19
Feat: Allow CustomKind subclasses for custom materializations

erindru:
The motivation behind this pull request is to allow custom materializations to fill a role of "behaves similar to an existing materialization - except with some requirements relaxed or changed".

Right now this is either difficult / requires copy+pasting a bunch of code or is just not possible (eg if your custom materialization needs to leverage the project level `time_column_format` config property).

The current custom materializations also dont play nicely with the builtin audits because there is no way of exposing `model.kind.time_column` (which this PR addresses)

### DIFF
--- a/examples/custom_materializations/custom_materializations/custom_kind.py
+++ b/examples/custom_materializations/custom_materializations/custom_kind.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import typing as t
+
+from sqlmesh import CustomMaterialization, CustomKind, Model
+from sqlmesh.utils.pydantic import validate_string
+from pydantic import field_validator
+
+if t.TYPE_CHECKING:
+    from sqlmesh import QueryOrDF
+
+
+class CustomCustomKind(CustomKind):
+    custom_property: t.Optional[str] = None
+
+    @field_validator("custom_property", mode="before")
+    @classmethod
+    def _validate_custom_property(cls, v: t.Any) -> str:
+        return validate_string(v)
+
+
+class CustomFullWithCustomKindMaterialization(CustomMaterialization[CustomCustomKind]):
+    NAME = "custom_full_with_custom_kind"
+
+    def insert(
+        self,
+        table_name: str,
+        query_or_df: QueryOrDF,
+        model: Model,
+        is_first_insert: bool,
+        **kwargs: t.Any,
+    ) -> None:
+        assert type(model.kind).__name__ == "CustomCustomKind"
+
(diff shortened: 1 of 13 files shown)


sqlmesh/core/model/kind.py
```diff
@@ -975,6 +975,16 @@ def create_model_kind(v: t.Any, dialect: str, defaults: t.Dict[str, t.Any]) -> M
         ):
             props["on_destructive_change"] = defaults.get("on_destructive_change")
 
+        if kind_type == CustomKind:
+            # load the custom materialization class and check if it uses a custom kind type
+            from sqlmesh.core.snapshot.evaluator import get_custom_materialization_type
+
+            actual_kind_type, _ = get_custom_materialization_type(
+                validate_string(props.get("materialization"))
```
izeigerman, line 983: How clear is the error when materialization doesn't exist or is not a string?

examples/custom_materializations/custom_materializations/custom_kind.py
```diff
@@ -0,0 +1,35 @@
...
+
+from sqlmesh import CustomMaterialization, CustomKind, Model
+from sqlmesh.utils.pydantic import validate_string
+from pydantic import field_validator
+
+if t.TYPE_CHECKING:
+    from sqlmesh import QueryOrDF
+
+
+class CustomCustomKind(CustomKind):
```
izeigerman, line 13: what a name lol

izeigerman (approved):

sqlmesh/core/model/kind.py
erindru: Good question.

Not specifying `materialization` at all on `main`:
```
$ sqlmesh plan
Error: 1 validation error for CustomKind
materialization
  Field required [type=missing, input_value={'name': 'CUSTOM'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
```

Not specifying `materialization` at all in this PR:
```
$ sqlmesh plan
Error: Materialization strategy with name 'None' was not found.
```

It's because we cant instantiate `CustomKind` up front since if the user has defined extra properties intended for a subclass, Pydantic will throw an error.

I'll update the check to output something like:
```
$ sqlmesh plan
Error: The 'materialization' property is required for models of the CUSTOM kind
```

For completeness, specifying a `materialization` that doesnt exist is the same in both:
```
$ sqlmesh plan
Error: Materialization strategy with name 'non_existent' was not found.
```

PR #3865, opened 2025-02-19
Feat: make date_spine macro less strict to allow dynamic behavior

### LINKED ISSUE #3864: Feature Request: Dynamic Date Spines
`@DATE_SPINE` enables data devs to build [date spines](https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#date_spine), however it requires hard coded date ranges.

It would be useful if data devs can build dynamic date spines, e.g.
```sql
WITH discount_promotion_dates AS (
  @date_spine('day', current_date - 90, current_date)
)

SELECT * FROM discount_promotion_dates
```

For now, I've been able to achieve the same result with the following BQ logic:
```
WITH _spine AS (
  SELECT
    spine_date
  FROM UNNEST(GENERATE_DATE_ARRAY(CURRENT_DATE - 90, CURRENT_DATE, INTERVAL '1' DAY)) AS spine_date
)
```

georgesittas:
Fixes #3864

The validations we do in `date_spine` today are too strict. This PR relaxes them a bit so that users can supply arbitrary expressions as arguments, besides just string literals– take a look at the new test to see why that makes sense.

### DIFF
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -5,6 +5,7 @@ from datetime import date, datetime
 from pathlib import Path
 from unittest.mock import patch, PropertyMock
 
+import time_machine
 import pandas as pd
 import pytest
 from pytest_mock.plugin import MockerFixture
@@ -7719,6 +7720,7 @@ def entrypoint(evaluator):
     )
 
 
+@time_machine.travel("2020-01-01 00:00:00 UTC")
 def test_dynamic_date_spine_model(assert_exp_eq):
     @macro()
     def get_current_date(evaluator):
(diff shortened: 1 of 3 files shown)


tests/core/test_model.py
```diff
@@ -7717,3 +7717,39 @@ def entrypoint(evaluator):
...
+        """
+    )
+    model = load_sql_based_model(expressions)
+    assert_exp_eq(
+        model.render_query(),
+        """
+        WITH "discount_promotion_dates" AS (
+          SELECT
+            "_exploded"."date_day" AS "date_day"
+          FROM UNNEST(CAST(GENERATE_SERIES(CAST('2025-02-19' AS DATE) - 90, CAST('2025-02-19' AS DATE), INTERVAL '1' DAY) AS DATE[])) AS "_exploded"("date_day")
```
z3z1ma, line 7749: You may want to freeze time for this test if it is not the case already, so `now` is deterministic

georgesittas: Definitely, good catch. Thanks!

tobymao (approved):

PR #3872, opened 2025-02-19
Feat: add support for na_values and keep_default_na in csv_settings

georgesittas:
Pandas' `read_csv` coerces certain values into `NaN` by default. This PR introduces support for the `na_values` and `keep_default_na` settings, which can be conditionally specified in order to override the coercion behavior.

Reference: https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html


sqlmesh/core/model/kind.py
```diff
@@ -640,9 +640,10 @@ def to_expression(
 
     @property
     def data_hash_values(self) -> t.List[t.Optional[str]]:
+        csv_setting_values = (self.csv_settings or CsvSettings()).dict().values()
         return [
             *super().data_hash_values,
-            *(self.csv_settings or CsvSettings()).dict().values(),
+            *(v if isinstance(v, (str, type(None))) else str(v) for v in csv_setting_values),
```
izeigerman, line 646: Does it make this change breaking?

georgesittas: I don't think so. Moreover, it looks like this also fixes a bug: the previous form didn't take `bool` [values](https://github.com/TobikoData/sqlmesh/blob/503d13ff4af0aaac0fbfcc06921565e2f8132d82/sqlmesh/core/model/seed.py#L23) in `CsvSettings` into account and so we would end up crashing if somebody used them:

```
(Pdb) model.kind
SeedKind<path: ../seeds/waiter_names.csv, batch_size: 100, csv_settings: CsvSettings<quotechar: ', doublequote: False, escapechar: \>>
(Pdb) model.kind.data_hash_values
['SEED', "'", False, '\\']
(Pdb) model.data_hash
*** TypeError: sequence item 3: expected str instance, bool found
```

This is because we concatenate data hash values and then [join them together](https://github.com/TobikoData/sqlmesh/blob/503d13ff4af0aaac0fbfcc06921565e2f8132d82/sqlmesh/utils/hashing.py#L20-L21), so it fails at the `join` method call.

izeigerman: We did add more values to the data hash value list. I can't wrap my head around why it doesn't impact the fingerprints.

izeigerman: Good catch with that bug

georgesittas: I think they're just filtered out because they're `None`? I will do a sanity check.

izeigerman: I believe Nones should be replaced with empty strings

georgesittas: Update: the default values are excluded in the `dict()` call by pydantic, so this should be backwards-compatible.

izeigerman (approved):

PR #3874, opened 2025-02-20
Chore: consolidate `make install-*`

erindru:
Makefile updates:
- consolidate `install-cicd-test` into `install-dev`
- remove `install-engine-test` as it was unused
- add the `./examples/custom_materializations` package to `make install-dev` so that `make fast-test` works out of the box


setup.py
```diff
@@ -68,12 +68,17 @@
             "beautifulsoup4",
             "clickhouse-connect",
             "cryptography",
-            "custom-materializations",
             "databricks-sql-connector",
             "dbt-bigquery",
             "dbt-core",
             "dbt-duckdb>=1.7.1",
             "dbt-snowflake",
```
erindru, line 75: I notice there were already a bunch of DBT engine libraries under `[dev]` so it seemed arbitrary to have some under `[dev]` and some under `[cicdtest]`

setup.py
```diff
@@ -68,12 +68,17 @@
             "beautifulsoup4",
             "clickhouse-connect",
             "cryptography",
-            "custom-materializations",
```
erindru, line 71: ^ it's unclear how to update this package if it changes. Installing the local version means that it always works and is always up to date

tobymao (approved):

PR #3875, opened 2025-02-20
Fix: Mask MotherDuck connection token in the logs

themisvaltinos:
Mask the MotherDuck connection token in logs, displaying it as asterisks (`md:db?motherduck_token=************`) instead of revealing the actual value.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -303,13 +303,13 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
             key = data_file if isinstance(data_file, str) else data_file.path
             if adapter := BaseDuckDBConnectionConfig._data_file_to_adapter.get(key):
                 logger.info(
-                    f"Using existing DuckDB adapter due to overlapping data file: {self._mask_motherduck_token(key)}"
+                    f"Using existing DuckDB adapter due to overlapping data file: {_mask_motherduck_token(key)}"
                 )
                 return adapter
 
         if data_files:
             masked_files = {
-                self._mask_motherduck_token(file if isinstance(file, str) else file.path)
+                _mask_motherduck_token(file if isinstance(file, str) else str(file))
                 for file in data_files
             }
             logger.info(f"Creating new DuckDB adapter for data files: {masked_files}")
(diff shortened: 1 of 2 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -323,6 +329,16 @@ def get_catalog(self) -> t.Optional[str]:
             return list(self.catalogs)[0]
         return None
 
+    def _mask_motherduck_token(self, path: str) -> str:
+        token_regex = r"(\?motherduck_token=)(\S*)"
```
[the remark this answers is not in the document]
themisvaltinos: Thanks wasn't aware of this, revised it to use pattern instead

izeigerman:
I believe there's another place where we leak tokens into logs: https://github.com/TobikoData/sqlmesh/blob/170df723bfbddbb608b3da545a7f535f6879f93b/sqlmesh/core/config/connection.py#L265-L273

The ATTACH statement contains a token and this is statement is happily logged by the engine adapter.

themisvaltinos:
> I believe there's another place where we leak tokens into logs:
> https://github.com/TobikoData/sqlmesh/blob/170df723bfbddbb608b3da545a7f535f6879f93b/sqlmesh/core/config/connection.py#L271-L273
> 
> The ATTACH statement contains a token and this is statement is happily logged by the engine adapter.

If I'm not mistaken we do not log this since this is the DuckDBPyConnection's `execute` method. Or is this being logged somewhere internally?

izeigerman:
@themisvaltinos ah, great point. I think you're right

sqlmesh/core/config/connection.py
```diff
@@ -323,6 +329,16 @@ def get_catalog(self) -> t.Optional[str]:
...
 
+    def _mask_motherduck_token(self, path: str) -> str:
+        MOTHERDUCK_TOKEN_REGEX = re.compile(r"(\?motherduck_token=)(\S*)")
+        if token_match := MOTHERDUCK_TOKEN_REGEX.search(path):
+            path = re.sub(
+                MOTHERDUCK_TOKEN_REGEX,
+                r"\1" + "*" * len(token_match.group(2)),
+                path,
+            )
+        return path
```
georgesittas, line 340: Cross-posting here from Slack: should we override `__str__` and `__repr__` and include this masking there to avoid this leaking due to interpolation in the future?

I believe one concern was that a path can also include the token. Perhaps we could capture both the `path` and `token` cases by using a single regex?

themisvaltinos: Right revised it to mask in DuckDBAttachOptions as well

georgesittas:
Update: I realized this is inconsistent with the other config classes, where `password` etc are not masked in `__str__` and `__repr__`. Apologies for the back and forth, that probably wasn't a great suggestion.

I think we can just revert the last commit and simply factor out the regex into the top of the file so it's compiled once. Then it should be good to go.

georgesittas (approved):

PR #3879, opened 2025-02-21
fix: prevent get_removal_interval from returning invalid interval

tobymao:
if your partial restate a daily model -> monthly, (2024-02-15, 2024-02-20), get_removal_interval returned an interval where start > end. this ensures get_removal_interval only returns valid intervals.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 import re
-import sys
 import typing as t
 from collections import defaultdict
 from functools import cached_property
@@ -27,7 +26,14 @@ from sqlmesh.core.snapshot.categorizer import categorize_change
 from sqlmesh.core.snapshot.definition import Interval, SnapshotId
 from sqlmesh.utils import columns_to_types_all_known, random_id
 from sqlmesh.utils.dag import DAG
-from sqlmesh.utils.date import TimeLike, now, to_datetime, yesterday_ds, to_timestamp
+from sqlmesh.utils.date import (
+    TimeLike,
+    now,
+    to_datetime,
+    yesterday_ds,
+    to_timestamp,
+    time_like_to_str,
+)
 from sqlmesh.utils.errors import NoChangesPlanError, PlanError, SQLMeshError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -322,56 +328,64 @@ def is_restateable_snapshot(snapshot: Snapshot) -> bool:
...
-                if self._context_diff.snapshots[s].is_incremental
-            ] + [interval]
+            restating_parents = [
+                self._context_diff.snapshots[s] for s in snapshot.parents if s in restatements
+            ]
+
+            if not restating_parents and snapshot.name not in restate_models:
+                continue
+            if not removal_interval:
+                self._console.log_error(
```
izeigerman, line 379: If this is an error should we raise? Otherwise should we `log_warning` instead?

sqlmesh/core/plan/builder.py
```diff
@@ -322,56 +328,64 @@ def is_restateable_snapshot(snapshot: Snapshot) -> bool:
...
-                for s in snapshot_dependencies
-                if self._context_diff.snapshots[s].is_incremental
-            ] + [interval]
+            restating_parents = [
+                self._context_diff.snapshots[s] for s in snapshot.parents if s in restatements
+            ]
+
+            if not restating_parents and snapshot.name not in restate_models:
+                continue
+            if not removal_interval:
```
izeigerman, line 378: This doesn't look correct. What if `removal_interval` is empty, but `restating_parents` is not?

sqlmesh/core/plan/builder.py
```diff
@@ -322,56 +328,64 @@ def is_restateable_snapshot(snapshot: Snapshot) -> bool:
...
-            possible_intervals = [
-                restatements.get(s, dummy_interval)
-                for s in snapshot_dependencies
-                if self._context_diff.snapshots[s].is_incremental
-            ] + [interval]
+            restating_parents = [
+                self._context_diff.snapshots[s] for s in snapshot.parents if s in restatements
+            ]
+
+            if not restating_parents and snapshot.name not in restate_models:
```
izeigerman, line 376: Should we do
```
removal_interval = snapshot.get_removal_interval(
```
after this check and not before?

sqlmesh/core/plan/builder.py
izeigerman: Shouldn't this be
```
if not possible_intervals
```
instead?

sqlmesh/core/plan/builder.py
izeigerman: Btw, this check doesn't seem useful, since `restating_parents` doesn't filter out non-incremental models. So `restating_models` might be non-empty, but then `possible_intervals` will be empty.

sqlmesh/core/plan/builder.py
tobymao: should be a warning

sqlmesh/core/plan/builder.py
tobymao: this used to be true before interval expansion, because we wanted to have daily parents restate, but not the monthly. i think this doesn't make sense anymore

tobymao: updated

sqlmesh/core/plan/builder.py
tobymao: possible_intervals can be empty because parents are full, but get_removal_interval should always exist with expansion

izeigerman (approved):

PR #3882, opened 2025-02-22
Add optional pyodbc driver for mssql and azuresql

mattiasthalen:
Adding the possibility to use pyodbc as the driver for MSSQL (and Azure SQL).

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1270,6 +1270,7 @@ class MySQLConnectionConfig(ConnectionConfig):
 
 class MSSQLConnectionConfig(ConnectionConfig):
     host: str
+    authentication: t.Optional[str] = None
     user: t.Optional[str] = None
     password: t.Optional[str] = None
     database: t.Optional[str] = ""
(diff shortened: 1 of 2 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -1285,13 +1286,16 @@ class MSSQLConnectionConfig(ConnectionConfig):
...
+    
+    driver: t.Optional[str] = "pymssql"
 
     type_: t.Literal["mssql"] = Field(alias="type", default="mssql")
 
     @property
     def _connection_kwargs_keys(self) -> t.Set[str]:
         return {
             "host",
+            "authentication",
```
erindru, line 1298: I know this PR is still in draft, but instead of defining `authentication` it might be  worth defining a dict of arbitrary `odbc_properties` instead, since there [are a bunch](https://learn.microsoft.com/en-us/sql/connect/odbc/dsn-connection-string-attribute?view=sql-server-ver16) that can be specified.

That way it's clear that these only work in ODBC mode, and theyre passed verbatim to the ODBC driver

tobymao:
closing this for now due to inactivity

PR #3884, opened 2025-02-24
Fix!: Improve interval compaction and cleanup

izeigerman:
Previously. intervals were compacted per snapshot ID. Previously, intervals were compacted based on snapshot IDs. While this typically works fine, it becomes problematic when multiple snapshots share the same version. In such cases, we must retain interval records from long-expired snapshots to maintain a complete history of backfilled intervals. 

If the model is forward-only, SQLMesh preserves interval records for all snapshots in its history. This becomes problematic, as the number of records required to compute intervals for a snapshot grows indefinitely without an upper bound.

This update enhances compaction and cleanup logic by introducing the following behavior upon snapshot expiration:
* If the version of the expired snapshot is not shared by any other snapshot, then all interval records associated with this version can be safely deleted
* If the **dev** version of the expired snapshot is not shared by any other snapshot, then all **dev** interval records associated with this dev version can be safely delete
* If the version of the expired snapshot is shared by other snapshots, the `identifier` and `dev_version` columns are set to `NULL` for non-dev interval records associated with this snapshot. This improves compaction by rolling up records of **all** expired snapshots to the `(name, version)` tuple, rather than `(name, identifier)`.
* If the **dev** version of the expired snapshot is shared by other snapshots, only the `identifier` column is set to `NULL` for **dev**` interval records associated with this snapshot. This improves compaction by rolling up dev records of **all** expired snapshots to the `(name, dev_version)` tuple, rather than `(name, identifier)`.

To support the above behavior, the `dev_version` column has been added to the interval tables. A corresponding migration script has been added to fill this new column with correct values, as well as to delete obsolete interval records following the logic outlined above.


sqlmesh/core/plan/evaluator.py
```diff
@@ -176,6 +176,7 @@ def _backfill(
                         name=snapshot.name,
                         identifier=snapshot.identifier,
                         version=snapshot.version,
+                        dev_version=snapshot.dev_version,
```
tobymao, line 179: i don't fully understand why dev version is needed, can you explain that?

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1198,17 +1203,152 @@ def recycle(self) -> None:
...
+
+        for where in self._snapshot_id_filter(snapshot_ids, alias=None):
+            # Nullify the identifier for dev intervals
+            # Set is_compacted to False so that it's compacted during the next compaction
+            self.engine_adapter.update_table(
+                self.intervals_table,
+                {"identifier": None, "is_compacted": False},
+                where=where.and_(exp.column("is_dev")),
+            )
+            # Nullify bote identifier and dev version for non-dev intervals
```
tobymao, line 1237: typo bote

tobymao:
breaking change because of migration

sqlmesh/core/state_sync/engine_adapter.py
```diff
@@ -1198,17 +1203,152 @@ def recycle(self) -> None:
...
+        ]
+        if not dev_keys_to_delete:
+            return
+
+        for where in self._snapshot_name_version_filter(
+            dev_keys_to_delete, version_column_name="dev_version", alias=None
+        ):
+            self.engine_adapter.delete_from(self.intervals_table, where.and_(exp.column("is_dev")))
+
+    def _delete_interlals_by_version(self, targets: t.List[SnapshotTableCleanupTask]) -> None:
```
tobymao, line 1260: typo

tobymao (approved):

PR #3886, opened 2025-02-24
Fix: Allow deserialization of models using custom materializations 

erindru:
Allow deserialization of models using custom materializations _without the custom materialization classes being present_.

This essentially moves the `kind` validation for custom materializations into `Model.validate_definition()`, which gets called after everything is loaded.

This means that components that just need to coerce Snapshots into JSON and back without doing anything special can continue to do so without needing the custom materialization classes loaded (since these would typically exist in an external library).

Only the components that evaluate the snapshots will need the custom materialization classes available in the Python environment.

This PR also rolls back the ability for CustomKind subclasses to define persistent properties outside of `materialization_properties`. The problem with that approach was that it forced the CustomKind subclasses (arbitrary user code) to be available in the Python environment of any context that touches a Snapshot, including state migration because it can affect how model fingerprints are calculated.

The downside of reverting this is that once again leveraging the project-level `time_column_format` property within a custom materialization becomes a problem. I've solved it by saving the value onto the `Model` in a private field and then when `validate_definition()` is called, re-applying it by calling `self.set_time_format()`

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -32,7 +32,14 @@ from sqlmesh.core.model.common import (
     single_value_or_tuple,
 )
 from sqlmesh.core.model.meta import ModelMeta, FunctionCall
-from sqlmesh.core.model.kind import ModelKindName, SeedKind, ModelKind, FullKind, create_model_kind
+from sqlmesh.core.model.kind import (
+    ModelKindName,
+    SeedKind,
+    ModelKind,
+    FullKind,
+    create_model_kind,
+    CustomKind,
+)
 from sqlmesh.core.model.seed import CsvSeedReader, Seed, create_seed
 from sqlmesh.core.renderer import ExpressionRenderer, QueryRenderer
 from sqlmesh.core.signal import SignalRegistry
@@ -978,6 +985,12 @@ class _Model(ModelMeta, frozen=True):
                     "Query validation can only be enabled for SQL models",
                     self._path,
                 )
+
+        if isinstance(self.kind, CustomKind):
+            from sqlmesh.core.snapshot.evaluator import get_custom_materialization_type_or_raise
+
+            # Will raise if the custom materialization points to an invalid class
+            get_custom_materialization_type_or_raise(self.kind.materialization)
 
     def is_breaking_change(self, previous: Model) -> t.Optional[bool]:
         """Determines whether this model is a breaking change in relation to the `previous` model.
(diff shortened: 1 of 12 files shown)


erindru:
Note: the `migration_test` failure is expected since i've un-done a feature. Since it was an advanced feature that has only been present for a few days, I'm taking a chance that nobody was using it.

sqlmesh/core/model/kind.py
```diff
@@ -976,25 +946,40 @@ def create_model_kind(v: t.Any, dialect: str, defaults: t.Dict[str, t.Any]) -> M
...
-                )
+    name = (v.name if isinstance(v, exp.Expression) else str(v)).upper()
+    return model_kind_type_from_name(name)(name=name)  # type: ignore
 
-            actual_kind_type, _ = get_custom_materialization_type(
-                validate_string(props.get("materialization"))
-            )
 
-            return actual_kind_type(**props)
+def create_time_column(v: t.Any, dialect: str) -> TimeColumn:
```
izeigerman, line 955: Can we make this a classmethod of the `TimeColumn` class for consistency? Like `def create(...)`

sqlmesh/core/model/definition.py
```diff
@@ -979,6 +989,22 @@ def validate_definition(self) -> None:
...
+            # Validate custom kinds here instead of as part of deserialization
+            # This allows the python types to be used in contexts where the custom materialization classes are not available
+            from sqlmesh.core.snapshot.evaluator import get_custom_materialization_type
+
+            # The below call fails if a materialization with the given name doesn't exist.
+            kind_subclass, _ = get_custom_materialization_type(self.kind.materialization)
+
+            if kind_subclass != CustomKind:
+                # hack because the Pydantic model is meant to be immutable
+                self.__dict__["kind"] = kind_subclass(**self.kind.dict(), dialect=self.kind.dialect)
```
izeigerman, line 1002: I wouldn't do it like this. I'd still attempt to deserialize into a correct type in the kind validator and fallback to `CustomKind` with a warning if the strategy is not available. Here we should only validate without coercing.

sqlmesh/core/model/kind.py
erindru: Sure, done

sqlmesh/core/model/definition.py
erindru: You're right, that's cleaner and I couldn't think of a downside since the evaluator attempts to re-fetch the evaluation strategy before using it which can raise an error if the class cant be found.

I've switched to back to using the kind subclass in `create_model_kind` if its available, otherwise falling back to `CustomKind` which is now possible because custom top-level properties are no longer supported.

However, i've made `Model.validate_definition()` still raise an error if the custom materialization class is not present

izeigerman (approved):

PR #3887, opened 2025-02-24
Feat(BigQuery): Add support for column descriptions in nested fields

### LINKED ISSUE #3854: Support nested column descriptions in BigQuery
I would like to add nested column descriptions to SQLMesh models. If running on BigQuery then the data type would be a RECORD type.

For example I don't just want to document the `my_record` column, I want to document its nested fields `my_record.field_1` and `my_record.field_2`. All 3 must be described, not just the top level.

This is especially helpful for wide tables that are pre-joined for performance reasons at query time.

I expect the syntax may look like this:

``` 
model (
     name my_dataset.my_table,
     description "Contains rows.",
     column_descriptions (
         my_record = "Description of the my_record struct."
         my_record.field_1 = "Description of field 1."
         my_record.field_2 = "Description of field 2."
     )
 );
```
This functionality is present in dbt but is not available with sqlmesh. See below for how this is documented in a schema.yml file:

```
      - name: my_record
         description: Description of the my_record struct
         data_type: struct
       - name: my_record.field_1
         description: Description of field 1.
         data_type: string # example
       - name: my_record.field_2
         description: Description of field 2.
         data_type: array<string>  # example
```

This feature should work regardless if the column is a RECORD (aka STRUCT) or a REPEATED RECORD (aka ARRAY of STRUCTs).

Also see: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1739773563724809?thread_ts=1739773563.724809&cid=C044BRE5W4S

themisvaltinos:
This PR introduces support for column descriptions in nested types in BigQuery, closes #3854 

The syntax follows BigQuery's conventions for nested and repeated columns, using dots to indicate the appropriate level of nesting:

```sql
MODEL (
  name schema.record_model,
  column_descriptions(
    id = "Unique identifier for the event",
    repeated_record.sub_repeated_record.sub_field = "Sub-field within nested repeated record",
    record.nested_record.field = "Field within nested record",
    record.nested_record.array = "Array of integers",
    repeated_record.sub_repeated_record = "Nested repeated record",
    record.nested_record = "Nested record"
  )
);

SELECT
  id,
  [STRUCT([STRUCT(sub_field AS sub_field)] AS sub_repeated_record)] AS repeated_record,
  STRUCT(
    STRUCT([2, 3] AS array, field AS field) AS nested_record
  ) AS record
FROM
  schema.seed_model;
```

The column definitions are built using the helper method `_build_column_def`, which is applied recursively in the BigQuery adapter to handle the RECORD and REPEATED RECORD columns. For views which are registered in `_create_column_comments` we traverse the fields and apply comments at each appropriate level.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -308,16 +308,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                # - snowflake
-                # - databricks
-                # - redshift
+                - snowflake
+                - databricks
+                - redshift
                 - bigquery
-                # - clickhouse-cloud
-                # - athena
-          # filters:
-          #  branches:
-          #    only:
-          #      - main
+                - clickhouse-cloud
+                - athena
+          filters:
+           branches:
+             only:
+               - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -308,16 +308,16 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
+                # - snowflake
+                # - databricks
+                # - redshift
                 - bigquery
-                - clickhouse-cloud
-                - athena
-          filters:
-           branches:
-             only:
-               - main
+                # - clickhouse-cloud
+                # - athena
+          # filters:
+          #  branches:
+          #    only:
+          #      - main
       - trigger_private_tests:
           requires:
             - style_and_slow_tests
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -796,29 +796,44 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
         col_type: t.Optional[exp.DATA_TYPE] = None,
         nested_names: t.List[str] = [],
     ) -> exp.ColumnDef:
-        # For BigQuery's nested fields we have to recursively build the column descriptions
-        if (
-            col_type
-            and isinstance(col_type, exp.DataType)
-            and col_type.is_type(exp.DataType.Type.STRUCT)
-            and (column_defs := col_type.expressions)
-        ):
+        # Helper function to build column definitions with column descriptions
+        def build_nested_column_defs(
+            col_type: exp.DataType,
+            nested_names: t.List[str],
+        ) -> exp.DataType:
             column_expressions = []
-            for column_def in column_defs:
+            for column_def in col_type.expressions:
                 if isinstance(column_def, exp.ColumnDef):
                     column = self._build_column_def(
                         col_name=column_def.name,
                         column_descriptions=column_descriptions,
                         engine_supports_schema_comments=engine_supports_schema_comments,
                         col_type=column_def.kind,
-                        nested_names=nested_names + [col_name],
+                        nested_names=nested_names,
                     )
                 else:
                     column = column_def
                 column_expressions.append(column)
-            col_type = exp.DataType(
-                this=exp.DataType.Type.STRUCT, expressions=column_expressions, nested=True
-            )
+            return exp.DataType(this=col_type.this, expressions=column_expressions, nested=True)
+
+        # Recursively build column definitions for BigQuery's RECORDs (struct) and REPEATED RECORDs (array of struct)
+        if (
+            col_type
+            and isinstance(col_type, exp.DataType)
+            and (expressions := col_type.expressions)
+        ):
+            if col_type.is_type(exp.DataType.Type.STRUCT):
+                col_type = build_nested_column_defs(col_type, nested_names + [col_name])
+            elif col_type.is_type(exp.DataType.Type.ARRAY) and expressions[0].is_type(
+                exp.DataType.Type.STRUCT
+            ):
+                col_type = exp.DataType(
+                    this=exp.DataType.Type.ARRAY,
+                    expressions=[
+                        build_nested_column_defs(col_type.expressions[0], nested_names + [col_name])
+                    ],
+                    nested=True,
+                )
 
         return exp.ColumnDef(
             this=exp.to_identifier(col_name),
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -669,18 +669,17 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
             for column, comment in column_comments.items():
                 fields = table_def["schema"]["fields"]
                 field_names = column.split(".")
+                last_index = len(field_names) - 1
 
                 # Traverse the fields with nested fields down to leaf level
-                for name in field_names:
-                    if fields and (
-                        field := next((field for field in fields if field["name"] == name), None)
-                    ):
-                        if name == field_names[-1]:
+                for idx, name in enumerate(field_names):
+                    if field := next((field for field in fields if field["name"] == name), None):
+                        if idx == last_index:
                             field["description"] = self._truncate_comment(
                                 comment, self.MAX_COLUMN_COMMENT_LENGTH
                             )
                         else:
-                            fields = field.get("fields", None)
+                            fields = field.get("fields", [])
 
             # An "etag" is BQ versioning metadata that changes when an object is updated/modified. `update_table`
             # compares the etags of the table object passed to it and the remote table, erroring if the etags
@@ -797,7 +796,7 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
         nested_names: t.List[str] = [],
     ) -> exp.ColumnDef:
         # Helper function to build column definitions with column descriptions
-        def build_nested_column_defs(
+        def _build_struct_with_descriptions(
             col_type: exp.DataType,
             nested_names: t.List[str],
         ) -> exp.DataType:
@@ -817,20 +816,18 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
             return exp.DataType(this=col_type.this, expressions=column_expressions, nested=True)
 
         # Recursively build column definitions for BigQuery's RECORDs (struct) and REPEATED RECORDs (array of struct)
-        if (
-            col_type
-            and isinstance(col_type, exp.DataType)
-            and (expressions := col_type.expressions)
-        ):
+        if isinstance(col_type, exp.DataType) and (expressions := col_type.expressions):
             if col_type.is_type(exp.DataType.Type.STRUCT):
-                col_type = build_nested_column_defs(col_type, nested_names + [col_name])
+                col_type = _build_struct_with_descriptions(col_type, nested_names + [col_name])
             elif col_type.is_type(exp.DataType.Type.ARRAY) and expressions[0].is_type(
                 exp.DataType.Type.STRUCT
             ):
                 col_type = exp.DataType(
                     this=exp.DataType.Type.ARRAY,
                     expressions=[
-                        build_nested_column_defs(col_type.expressions[0], nested_names + [col_name])
+                        _build_struct_with_descriptions(
+                            col_type.expressions[0], nested_names + [col_name]
+                        )
                     ],
                     nested=True,
                 )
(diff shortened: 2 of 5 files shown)


.circleci/continue_config.yml
```diff
@@ -307,16 +307,16 @@ workflows:
...
                 - bigquery
-                - clickhouse-cloud
-                - athena
-          filters:
-           branches:
-             only:
-               - main
+                # - clickhouse-cloud
+                # - athena
+          # filters:
```
themisvaltinos, line 316: Note: to reinstate prior to merging.

sqlmesh/core/model/meta.py
```diff
@@ -238,7 +238,7 @@ def _column_descriptions_validator(
             vs = vs.expressions
 
         raw_col_descriptions = (
-            vs if isinstance(vs, dict) else {v.this.name: v.expression.name for v in vs}
+            vs if isinstance(vs, dict) else {v.this.sql(dialect): v.expression.name for v in vs}
```
georgesittas, line 241: Why did this change?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -779,6 +788,65 @@ def _build_table_properties_exp(
...
+    def _build_column_def(
+        self,
+        col_name: str,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        engine_supports_schema_comments: bool = False,
+        col_type: t.Optional[exp.DATA_TYPE] = None,
+        nested_names: t.List[str] = [],
+    ) -> exp.ColumnDef:
+        # Helper function to build column definitions with column descriptions
+        def build_nested_column_defs(
```
georgesittas, line 800: Should this be renamed to `_build_struct_with_descriptions` or something?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -779,6 +788,65 @@ def _build_table_properties_exp(
...
+            col_type: exp.DataType,
+            nested_names: t.List[str],
+        ) -> exp.DataType:
+            column_expressions = []
+            for column_def in col_type.expressions:
+                if isinstance(column_def, exp.ColumnDef):
```
georgesittas, line 806: Can this condition actually be false?

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,106 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
...
+    adapter = make_mocked_engine_adapter(BigQueryEngineAdapter)
+    allowed_column_comment_length = BigQueryEngineAdapter.MAX_COLUMN_COMMENT_LENGTH
+
+    execute_mock = mocker.patch(
+        "sqlmesh.core.engine_adapter.bigquery.BigQueryEngineAdapter.execute"
+    )
+
+    nested_columns_to_types = {
+        "record": exp.DataType.build(
+            "STRUCT<'int_field': INT, 'record_field': STRUCT<'sub_record_field': STRUCT<'nest_array' ARRAY<INT64>>>"
```
georgesittas, line 683: Looks like a typo:
```suggestion
            "STRUCT<'int_field': INT, 'record_field': STRUCT<'sub_record_field': STRUCT<'nest_array': ARRAY<INT64>>>"
```

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,106 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
...
+        ),
+        "repeated_record": exp.DataType.build(
+            "ARRAY<STRUCT<'nested_repeated_record' ARRAY<STRUCT<'int_field': INT, 'array_field': ARRAY<INT64>, 'struct_field' STRUCT<'nested_field': INT>>>>>"
```
georgesittas, line 686: Ditto:
```suggestion
            "ARRAY<STRUCT<'nested_repeated_record': ARRAY<STRUCT<'int_field': INT, 'array_field': ARRAY<INT64>, 'struct_field': STRUCT<'nested_field': INT>>>>>"
```

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,106 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
     ]
 
 
+def test_nested_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture):
+    adapter = make_mocked_engine_adapter(BigQueryEngineAdapter)
+    allowed_column_comment_length = BigQueryEngineAdapter.MAX_COLUMN_COMMENT_LENGTH
```
georgesittas, line 675: Do we actually need to involve the max comment length in this test? Should we just use dummy / short comments that are easier to eyeball? If you're testing truncation, I think it'd suffice to only do it for one field of a struct, not all of them.

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,106 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
...
+        ),
+    }
+
+    long_column_descriptions = {
+        "record": "1" * (allowed_column_comment_length * 2),
+        "record.int_field": "2" * (allowed_column_comment_length * 2),
+        "record.record_field": "3" * (allowed_column_comment_length * 2),
+        "record.record_field.sub_record_field": "4" * (allowed_column_comment_length * 2),
+        "record.record_field.sub_record_field.nest_array": "5"
+        * (allowed_column_comment_length * 2),
```
georgesittas, line 696: For instance this bit is awkward to read. Not sure what the multiplication offers in this test.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -665,13 +665,22 @@ def _create_column_comments(
...
+                for name in field_names:
+                    if fields and (
+                        field := next((field for field in fields if field["name"] == name), None)
+                    ):
+                        if name == field_names[-1]:
+                            field["description"] = self._truncate_comment(
+                                comment, self.MAX_COLUMN_COMMENT_LENGTH
+                            )
+                        else:
+                            fields = field.get("fields", None)
```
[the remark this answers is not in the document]
themisvaltinos: Right good point adapting this to use a list

tests/core/engine_adapter/test_bigquery.py
themisvaltinos: I wanted to ensure that the truncation happens in all nested comments, but you're right that it is a bit of an eyesore. Since we have the test above to test the truncation adapted this test to use shorter descriptions instead to make it easier to read

georgesittas: You can still do it a for a nested field, but not necessary for all of them.

sqlmesh/core/engine_adapter/bigquery.py
themisvaltinos: I don't think so but I wanted to be cautious to be on the safe side

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -665,13 +665,22 @@ def _create_column_comments(
...
+            for column, comment in column_comments.items():
+                fields = table_def["schema"]["fields"]
+                field_names = column.split(".")
+
+                # Traverse the fields with nested fields down to leaf level
+                for name in field_names:
+                    if fields and (
+                        field := next((field for field in fields if field["name"] == name), None)
+                    ):
+                        if name == field_names[-1]:
```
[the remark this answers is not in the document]
themisvaltinos: Excellent point didn't consider the edge case of nested fields with identical names, adapted it as you suggested thanks for catching this!

sqlmesh/core/model/meta.py
themisvaltinos: To retain all parts for later when we add the comments, if the column in `column_descriptions` is `record.nested_record.field`, since using `this.name` would only keep the field part. This way we can retain the original with the nested fields provided by the user to be able to support this

georgesittas: I see, but I'm not sure if this will work correctly if any of the parts are quoted. The quotes will be included in the string, e.g. ``` '`foo`' ```, then the dict lookup will fail downstream because keys don't include quotes.

So I would test the following cases:
- ```foo.`bar`.baz```
- ```foo.`bar with spaces`.baz```

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -665,13 +665,21 @@ def _create_column_comments(
...
+
+                # Traverse the fields with nested fields down to leaf level
+                for idx, name in enumerate(field_names):
+                    if field := next((field for field in fields if field["name"] == name), None):
+                        if idx == last_index:
+                            field["description"] = self._truncate_comment(
+                                comment, self.MAX_COLUMN_COMMENT_LENGTH
+                            )
+                        else:
+                            fields = field.get("fields", [])
```
georgesittas, line 682:
```suggestion
                            fields = field.get("fields") or []
```

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -665,13 +665,21 @@ def _create_column_comments(
...
+            # Set column descriptions, supporting nested fields (e.g. record.field.nested_field)
+            for column, comment in column_comments.items():
+                fields = table_def["schema"]["fields"]
+                field_names = column.split(".")
+                last_index = len(field_names) - 1
+
+                # Traverse the fields with nested fields down to leaf level
+                for idx, name in enumerate(field_names):
+                    if field := next((field for field in fields if field["name"] == name), None):
+                        if idx == last_index:
```
georgesittas, line 677: Let's also add a test for the `a.b.b` case.

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: Ok, I guess being defensive in this case doesn't hurt. The alternative of crashing at runtime because of some weird corner case is probably not good. My concern had to do with reasoning about the alternative case if somehow reads this code in the future, perhaps leave a comment explaining this?

sqlmesh/core/model/meta.py
themisvaltinos: Revised it to cover the cases of columns that are quoted and contain spaces between and added test for this

sqlmesh/core/engine_adapter/bigquery.py
themisvaltinos: Sure added test for this case as well

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,139 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
...
+        "record with space.record_field.sub_record_field": "Record Nested Records Subfield",
+        "record with space.record_field.sub_record_field.nest_array": "Record Nested Records Nested Array",
+        "repeated_record": "Top Repeated Record",
+        "repeated_record.nested_repeated_record": "Nested Repeated Record",
+        "repeated_record.nested_repeated_record.int_field": "Nested Repeated Record Int Field",
+        "repeated_record.nested_repeated_record.array_field": "Nested Repeated Array Field",
+        "repeated_record.nested_repeated_record.struct_field with space": "Nested Repeated Struct Field",
+        "repeated_record.nested_repeated_record.struct_field with space.nested_field": "Nested Repeated Record Nested Field",
+        "same_name_": "Level 1",
+        "same_name_.same_name_": "Level 2",
```
georgesittas, line 724: Maybe let's omit the comment in `same_name_.same_name_` to show that it won't get one anyway.

georgesittas (commented):
A couple more comments, but LGTM otherwise.

tests/core/engine_adapter/test_bigquery.py
themisvaltinos: Right adapted it

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -670,6 +670,139 @@ def test_comments(make_mocked_engine_adapter: t.Callable, mocker: MockerFixture)
...
+    allowed_column_comment_length = BigQueryEngineAdapter.MAX_COLUMN_COMMENT_LENGTH
+
+    execute_mock = mocker.patch(
+        "sqlmesh.core.engine_adapter.bigquery.BigQueryEngineAdapter.execute"
+    )
+
+    nested_columns_to_types = {
+        "record with space": exp.DataType.build(
+            "STRUCT<"
+            "'int_field': INT, "
```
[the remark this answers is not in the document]
themisvaltinos: Thanks yes you're right revised it. And made them one liner strings I hope it's more readable now

georgesittas (approved):

PR #3888, opened 2025-02-24
Feat: Add model defaults support per gateway configuration

VaggelisD:
Users can employ different gateways on the same project:

```YAML
gateways:
  redshift:
      connection:
        type: redshift
        ...

  snowflake:
    connection:
      type: snowflake
      ...

default_gateway: snowflake

model_defaults:
  dialect: snowflake
  start: 2025-02-05
```

However, the models will be normalized at the `model_defaults.dialect` even if the `default_gateway` is set to a different engine, which can lead to issues / inefficiencies e.g Redshift's `INFORMATION_SCHEMA` stores the tables as lower-case, so with Snowflake's upper-case normalization we'll fail to find the tables/views. 

This PR adds support for `model_defaults` under gateways which override the global dialect, e.g:

```YAML
gateways:
  redshift:
      connection:
        type: redshift
        ...

      model_defaults:
        dialect: snowflake, normalization_strategy=case_insensitive
```

This enables the following behavior:

```Python3
# Snowflake dialect, Redshift gateway without model_defaults
❯ sqlmesh render vaggelis.test_model
[WARNING] The redshift engine is not recommended for storing SQLMesh state in production deployments. Please see 
https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#state-connection for a list of recommended engines and more information.
SELECT                                                                                                                                              
  "COL" AS "COL"                                                                                                                                    
FROM "TBL" AS "TBL"  

# Snowflake dialect, Redshift gateway with model_defaults.dialect: snowflake, normalization_strategy=case_insensitive
❯ sqlmesh render vaggelis.test_model
[WARNING] The redshift engine is not recommended for storing SQLMesh state in production deployments. Please see 
https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#state-connection for a list of recommended engines and more information.
SELECT                                                                                                                                              
  "col" AS "col"                                                                                                                                    
FROM "tbl" AS "tbl" 
```

### DIFF
--- a/sqlmesh/core/config/root.py
+++ b/sqlmesh/core/config/root.py
@@ -287,6 +287,11 @@ class Config(BaseConfig):
 
     @property
     def dialect(self) -> t.Optional[str]:
+        if self.default_gateway:
+            gateway_config = self.gateways[self.default_gateway]
+            if gateway_config.model_defaults:
+                return gateway_config.model_defaults.dialect
+
         return self.model_defaults.dialect
 
     @property
--- a/sqlmesh/core/config/root.py
+++ b/sqlmesh/core/config/root.py
@@ -287,11 +287,6 @@ class Config(BaseConfig):
 
     @property
     def dialect(self) -> t.Optional[str]:
-        if self.default_gateway:
-            gateway_config = self.gateways[self.default_gateway]
-            if gateway_config.model_defaults:
-                return gateway_config.model_defaults.dialect
-
         return self.model_defaults.dialect
 
     @property
--- a/tests/core/test_context.py
+++ b/tests/core/test_context.py
@@ -11,7 +11,7 @@ import pytest
 import pandas as pd
 from pathlib import Path
 from pytest_mock.plugin import MockerFixture
-from sqlglot import exp, parse_one
+from sqlglot import exp, parse_one, Dialect
 from sqlglot.errors import SchemaError
 
 from sqlmesh.core.config.gateway import GatewayConfig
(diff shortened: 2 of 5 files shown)


sqlmesh/core/config/root.py
```diff
@@ -287,6 +287,11 @@ def default_gateway_name(self) -> str:
 
     @property
     def dialect(self) -> t.Optional[str]:
+        if self.default_gateway:
+            gateway_config = self.gateways[self.default_gateway]
+            if gateway_config.model_defaults:
+                return gateway_config.model_defaults.dialect
+
```
georgesittas, line 294: I'm not sure if this is the right way, why just do it for the dialect? Folks may want to choose different defaults per gateway in general.

In an ideal world, imo `Config.model_defaults` should be the result of merging the base / global defaults with the gateway's defaults, prioritizing the latter when there's a key collision.

@tobymao thoughts?

tobymao: right i agree

tests/core/test_context.py
```diff
@@ -1103,6 +1103,35 @@ def test_override_dialect_normalization_strategy():
...
+    assert Snowflake.NORMALIZATION_STRATEGY == NormalizationStrategy.UPPERCASE
+
+    ctx = Context(paths=tmp_path, config=config, gateway="duckdb")
+
+    dialect = Dialect.get_or_raise(ctx.config.dialect)
+
+    assert dialect == "snowflake"
+    assert Snowflake.NORMALIZATION_STRATEGY == NormalizationStrategy.CASE_INSENSITIVE
+
+    Snowflake.NORMALIZATION_STRATEGY = NormalizationStrategy.CASE_INSENSITIVE
```
georgesittas, line 1132:
```suggestion
    Snowflake.NORMALIZATION_STRATEGY = NormalizationStrategy.UPPERCASE
```

georgesittas (approved):

tests/core/test_context.py
georgesittas: This is probably the reason why tests are failing, FYI.

PR #3890, opened 2025-02-24
fix!: raise instead of warn on catalog mismatch

eakmanrq:
For engines that are single catalog only, like Redshift and Postgres, currently we will warn if we detect they seem to be doing a cross-catalog operation. This code is old and hasn't been an issue, but we hit a scenario where a user ran janitor and the janitor deleted something unexpectedly due to this being a warning.

Therefore changing this to raise instead.

### DIFF
--- a/sqlmesh/core/engine_adapter/shared.py
+++ b/sqlmesh/core/engine_adapter/shared.py
@@ -11,7 +11,7 @@ from pydantic import Field
 from sqlglot import exp
 
 from sqlmesh.core.dialect import to_schema
-from sqlmesh.utils.errors import UnsupportedCatalogOperationError
+from sqlmesh.utils.errors import UnsupportedCatalogOperationError, SQLMeshError
 from sqlmesh.utils.pydantic import PydanticModel
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/engine_adapter/shared.py
```diff
@@ -328,7 +328,7 @@ def internal_wrapper(*args: t.Any, **kwargs: t.Any) -> t.Any:
             container[key] = expression  # type: ignore
             if catalog_support.is_single_catalog_only:
                 if catalog_name != engine_adapter._default_catalog:
-                    logger.warning(
+                    raise SQLMeshError(
                         f"{engine_adapter.dialect} requires that all catalog operations be against a single catalog: {engine_adapter._default_catalog}. Ignoring catalog: {catalog_name}"
```
vchan, line 332: Should we delete the "Ignoring catalog: ..." part?

vchan (approved):

sqlmesh/core/engine_adapter/shared.py
eakmanrq: thanks updated

PR #3892, opened 2025-02-25
Fix: Merge source and target aliases resolution for multi part fqns

themisvaltinos:
This update fixes an issue in `when_matched` and `merge_filter` when dealing with multi-part fqns (e.g. `source.second_part.column` ), where the aliases for source and target weren't replaced. The `replace_merge_table_aliases` function now is checking for the first part of the column instead of the column's table and replaces that. This allows for cases like BigQuery's nested fields, which can have arbitrary lengths of parts, by resolving the first part of the name.

Related slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1740423499786879

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -1361,10 +1361,10 @@ def replace_merge_table_aliases(expression: exp.Expression) -> exp.Expression:
     """
     from sqlmesh.core.engine_adapter.base import MERGE_SOURCE_ALIAS, MERGE_TARGET_ALIAS
 
-    if isinstance(expression, exp.Column):
-        if expression.table.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
-            expression.set("table", exp.to_identifier(MERGE_TARGET_ALIAS))
-        elif expression.table.lower() in ("source", "dbt_internal_source", "__merge_source__"):
-            expression.set("table", exp.to_identifier(MERGE_SOURCE_ALIAS))
+    if isinstance(expression, exp.Column) and (first_part := expression.parts[0]):
+        if first_part.this.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
+            first_part.replace(exp.to_identifier(MERGE_TARGET_ALIAS))
+        elif first_part.this.lower() in ("source", "dbt_internal_source", "__merge_source__"):
+            first_part.replace(exp.to_identifier(MERGE_SOURCE_ALIAS))
 
     return expression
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -1361,10 +1361,10 @@ def replace_merge_table_aliases(expression: exp.Expression) -> exp.Expression:
     """
     from sqlmesh.core.engine_adapter.base import MERGE_SOURCE_ALIAS, MERGE_TARGET_ALIAS
 
-    if isinstance(expression, exp.Column):
-        if expression.table.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
-            expression.set("table", exp.to_identifier(MERGE_TARGET_ALIAS))
-        elif expression.table.lower() in ("source", "dbt_internal_source", "__merge_source__"):
-            expression.set("table", exp.to_identifier(MERGE_SOURCE_ALIAS))
+    if isinstance(expression, exp.Column) and (first_part := expression.parts[0]):
```
tobymao, line 1364: i'm a bit confused why this change was needed, expression.table should work for multi part fqns like catalog.db.table, what's going on here?

themisvaltinos: This issue arises because for instance in BigQuery if a record field contains a nested field and it’s used in a condition like `WHEN MATCHED AND source.record.field = 1`, we end up comparing the table `record` instead of the `source` which in this case would be the db in the Column expression:
```
Column(
    this=Identifier(this=field, quoted=False),
    table=Identifier(this=record, quoted=False),
    db=Identifier(this=source, quoted=False),
)
```

To prevent this from happening I used the `parts` approach to always replace the first part if that is using the aliases `source` or `target` since we never know in which column part it is going to be in those cases.

tobymao: what happens then if they do 

catalog.db.source.field ? wouldn't you change the catalog by accident?

themisvaltinos: I'm not sure if this is allowed in the syntax for [when matched and merge filter](https://sqlmesh.readthedocs.io/en/stable/concepts/models/model_kinds/#when-matched-expression) at least not to be replaced with the table names used in the merge into:
```sql
MERGE INTO <existing_table> AS "__MERGE_TARGET__"
    USING <temp_table> AS "__MERGE_SOURCE__"
    ON
        AND "__MERGE_SOURCE__".record.operation <> 1
```
since we use `source` and `target` to refer to the tables in the merge like in this example and replace these with the aliases. But correct me if I'm wrong @georgesittas , @eakmanrq  if this is allowed as well?

georgesittas: Seems reasonable; I believe `__MERGE_SOURCE__` and `__MERGE_TARGET__` are expected to appear as the leftmost part, since they're just aliases for the `MERGE`'s `source` and `target` sources.

tobymao (approved):

PR #3895, opened 2025-02-25
Fix(Redshift): Introduce flag to control the use of redshifts native operation or logical merge

themisvaltinos:
This update modifies the default behaviour for the incremental_by_unique_key model to use SQLMesh's logical merge for Redshift, unless the user opts in to use the native Redshift MERGE operation.

To enable the native MERGE statement in Redshift, the user must set the merge_operation flag to true in the connection configuration. By default it is disabled.

Example configuration:
```yaml
gateways:
  redshift:
    connection:
      type: redshift
      merge_operation: true
```

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1104,6 +1104,7 @@ class RedshiftConnectionConfig(ConnectionConfig):
         serverless_acct_id: The account ID of the serverless. Default value None
         serverless_work_group: The name of work group for serverless end point. Default value None.
         pre_ping: Whether or not to pre-ping the connection before starting a new transaction to ensure it is still alive.
+        merge_operation: Whether to use the Redshift merge operation instead of the SQLMesh logical merge.
     """
 
     user: t.Optional[str] = None
@@ -1127,6 +1128,7 @@ class RedshiftConnectionConfig(ConnectionConfig):
     is_serverless: t.Optional[bool] = None
     serverless_acct_id: t.Optional[str] = None
     serverless_work_group: t.Optional[str] = None
+    merge_operation: t.Optional[bool] = None
 
     concurrent_tasks: int = 4
     register_comments: bool = True
@@ -1169,6 +1171,10 @@ class RedshiftConnectionConfig(ConnectionConfig):
         from redshift_connector import connect
 
         return connect
+
+    @property
+    def _extra_engine_config(self) -> t.Dict[str, t.Any]:
+        return {"merge_operation": self.merge_operation}
 
 
 class PostgresConnectionConfig(ConnectionConfig):
--- a/sqlmesh/core/engine_adapter/redshift.py
+++ b/sqlmesh/core/engine_adapter/redshift.py
@@ -14,6 +14,7 @@ from sqlmesh.core.engine_adapter.mixins import (
     NonTransactionalTruncateMixin,
     VarcharSizeWorkaroundMixin,
     RowDiffMixin,
+    logical_merge,
 )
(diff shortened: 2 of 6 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -1127,6 +1128,7 @@ class RedshiftConnectionConfig(ConnectionConfig):
     is_serverless: t.Optional[bool] = None
     serverless_acct_id: t.Optional[str] = None
     serverless_work_group: t.Optional[str] = None
+    merge_operation: t.Optional[bool] = None
```
izeigerman, line 1131: `use_merge` or `enable_merge`

themisvaltinos: Ok sound better revised it to `enable_merge`

tests/conftest.py
```diff
@@ -465,6 +466,11 @@ def _make_function(
                 "sqlmesh.core.engine_adapter.mssql.MSSQLEngineAdapter.catalog_support",
                 new_callable=PropertyMock(return_value=CatalogSupport.REQUIRES_SET_CATALOG),
             )
+        if isinstance(adapter, RedshiftEngineAdapter):
+            mocker.patch(
+                "sqlmesh.core.engine_adapter.redshift.RedshiftEngineAdapter.enable_merge",
+                new_callable=PropertyMock(return_value=True),
```
erindru, line 472: Is there a reason why we set this to `True` in tests instead of leaving its default value of `False`?

erindru:
I think there is a missing test. Unless i'm mistaken, nothing is testing that the engine adapter actually issues a `MERGE` statement vs a logical merge if the new property is set

tests/conftest.py
themisvaltinos: yes this was to opt in and  use a native MERGE statement for these tests: https://github.com/TobikoData/sqlmesh/blob/main/tests/core/engine_adapter/test_redshift.py#L328

themisvaltinos: you're right though there isn't one for the logical merge adding a test now

erindru: This hasnt opted in though, its enabled merge for every single test that uses a `RedshiftEngineAdapter` created via `make_mocked_engine_adapter()`

Wouldnt it be better to just turn on this option for the tests that need it?

themisvaltinos: Oh I see what you're saying yes you're right

erindru (approved):
Looks good to me!

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -45,6 +46,7 @@ class RedshiftEngineAdapter(
     # Redshift doesn't support comments for VIEWs WITH NO SCHEMA BINDING (which we always use)
     COMMENT_CREATION_VIEW = CommentCreationView.UNSUPPORTED
     SUPPORTS_REPLACE_TABLE = False
+    MERGE_OPERATION = False
```
georgesittas, line 49: Is this used anywhere besides L129?

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -120,6 +122,12 @@ def build_var_length_col(
             for column_name, data_type in columns
         }
 
+    @property
+    def enable_merge(self) -> bool:
+        # Redshift supports the MERGE operation but we use the logical merge
+        # unless the user has opted in by setting enable_merge in the connection.
+        return self._extra_config.get("enable_merge") or self.MERGE_OPERATION
```
georgesittas, line 129: Should we do `bool(self._extra...)` instead? See other comment too.

PR #3896, opened 2025-02-25
Fix MotherDuck connection details in docs

danthelion:


themisvaltinos (commented):
Is there more context on this and was this removed? The documentation above refers to this example using catalogs option instead so it is inconsistent now. If this is going to be added, I feel the docs should include both examples

docs/integrations/engines/motherduck.md
```diff
@@ -43,12 +43,10 @@ gateways:
     motherduck:  
         connection:   
             type: motherduck  
-            catalogs:
```
themisvaltinos, line 46: Why was this removed?

danthelion:
Sorry, didn't actually mean to open this targeting ya'lls repo lol

PR #3902, opened 2025-02-26
Fix: Selectively widen restatement intervals if they affect a downstream model that cannot be partially restated

erindru:
Prior to this PR, if you had something like `A [incremental_by_time_range] <- B[incremental_by_unique_key]` and tried to restate a part of A, B would only be fully restated if it was in `prod`.

If B was in a dev environment, then it would be partially restated even though it's a model type that doesn't support partial restatement. If B was later promoted to `prod`, then this could lead to incorrect data in `prod`.

This PR adjusts the logic to, when calculating restatement intervals based on `SnapshotTableInfo`, if any of the affected snapshots are "full restatement only" then the full snapshot records are fetched from state so that the restatement intervals can be widened correctly.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -462,7 +462,7 @@ class BuiltInPlanEvaluator(PlanEvaluator):
         if not prod_restatements:
             return set()
 
-        snapshots_to_restate: t.Set[t.Tuple[SnapshotTableInfo, Interval]] = set()
+        snapshots_to_restate: t.Dict[SnapshotId, t.Tuple[SnapshotTableInfo, Interval]] = {}
 
         for env in self.state_sync.get_environments():
             keyed_snapshots = {s.name: s.table_info for s in env.snapshots}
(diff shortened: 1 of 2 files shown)


sqlmesh/core/plan/evaluator.py
```diff
@@ -453,10 +453,31 @@ def _restatement_intervals_across_all_environments(
...
+                        for a in affected_snapshot_names
+                    }
                 )
 
-        return snapshots_to_restate
+        # for any affected full_history_restatement_only snapshots, we need to widen the intervals being restated to
+        # include the whole time range for that snapshot. This requires a call to state to load the full snapshot record,
+        # so we only do it if necessary
+        if full_history_restatement_snapshot_ids := [
+            s_id for s_id, s in snapshots_to_restate.items() if s[0].full_history_restatement_only
```
izeigerman, line 466: Do we worry about the `depends_on_past` case: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/definition.py#L763 ?

sqlmesh/core/plan/evaluator.py
```diff
@@ -453,10 +453,31 @@ def _restatement_intervals_across_all_environments(
...
+        ]:
+            full_snapshots = self.state_sync.get_snapshots(full_history_restatement_snapshot_ids)
```
izeigerman, line 468: We should also check whether the snapshot IDs are different from the current plan so that we don't fetch those. I expect that many snapshots will be shared and so there's no point in fetching them again

sqlmesh/core/plan/evaluator.py
erindru: To summarise the results of the offline discussion: we will put a `//fixme` here because right now `depends_on_past` requires a rendered query and is lazily evaluated which exposes a can of worms with regards to the StateSync and running user code

sqlmesh/core/plan/evaluator.py
erindru: Makes sense, i've updated the code to pass down the list of already loaded snapshots.

Additional snapshots are only loaded if they're not part of this list

izeigerman (approved):

sqlmesh/core/plan/evaluator.py
```diff
@@ -438,7 +438,9 @@ def _restate(self, plan: EvaluatablePlan, snapshots_by_name: t.Dict[str, Snapsho
         # Without this rule, its possible that promoting a dev table to prod will introduce old data to prod
         snapshot_intervals_to_restate.update(
             self._restatement_intervals_across_all_environments(
-                plan.restatements, plan.disabled_restatement_models
+                prod_restatements=plan.restatements,
+                disable_restatement_models=plan.disabled_restatement_models,
+                loaded_snapshots={s.snapshot_id: s for s in snapshots_by_name.values()},
```
izeigerman, line 443: We already have this mapping upstream. Should we just pass it all the way?

erindru: The downside of that is that it creates two sources of truth. I will leave it for now unless it turns out there is a performance issue

PR #3917, opened 2025-02-27
Fix: make dialect regex more robust

### LINKED ISSUE #3916: If the name of the model contains the string `DIALECT`, the model will fail to execute.
The procedure for reproducing the error is as follows:

Create a model file as follows.
```
% cat models/incremental_items_lower.sql 
MODEL (
  name DIALECT_CHECK_RAW_sqlmesh.incremental_items_lower,
    time_column _dlt_load_time,
  ),
);

SELECT
  CAST(c.id AS BIGINT) AS id,
  CAST(c.name AS TEXT) AS name,
  CAST(c._dlt_load_id AS TEXT) AS _dlt_load_id,
  CAST(c._dlt_id AS TEXT) AS _dlt_id,
  TO_TIMESTAMP(CAST(c._dlt_load_id AS DOUBLE)) as _dlt_load_time
FROM
  DIALECT_CHECK_RAW.items_lower as c
WHERE
  TO_TIMESTAMP(CAST(c._dlt_load_id AS DOUBLE)) BETWEEN @start_ds AND @end_ds

```

Next, attempt to format it. But it fails.
```
% pdm run sqlmesh format
Error: Failed to load model definition at '/path/to/project/root/models/incremental_items_lower.sql'.
Unknown dialect 'CHECK'.
```

Perhaps the string `DIALECT` in the model name is mistakenly recognized as the model's attribute key, `dialect`.

The version of sqlmesh is as follows:
```
% pdm run sqlmesh --version 
0.162.0
```

georgesittas:
Fixes #3916

The new regex says: "If I see a `dialect`, ensure there's at least one whitespace, then [a-zA-Z]*, optionally wrapped in single quotes to allow stuff like `'duckdb'` (we have some tests for this ...).

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -761,7 +761,7 @@ def text_diff(
 
 
 DIALECT_PATTERN = re.compile(
-    r"(model|audit).*?\(.*?dialect[^a-z,]+([a-z]*|,)", re.IGNORECASE | re.DOTALL
+    r"(model|audit).*?\(.*?dialect\s+'?([a-z]*)", re.IGNORECASE | re.DOTALL
 )
 
 
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -4071,6 +4071,9 @@ def test_model_dialect_name():
     )
     assert "name `project-1`.`db`.`tbl1`" in model.render_definition()[0].sql(dialect="bigquery")
 
+    # This used to fail due to the dialect regex picking up `DIALECT_TEST` as the model's dialect
+    expressions = d.parse("MODEL(name DIALECT_TEST.foo); SELECT 1")
+
 
 def test_model_allow_partials():
     expressions = d.parse(


georgesittas:
Looks like CodeQL found an issue with the pattern matching: https://github.com/TobikoData/sqlmesh/pull/3917/checks?check_run_id=37937812882.

sqlmesh/core/dialect.py
```diff
@@ -761,7 +761,7 @@ def text_diff(
 
 
 DIALECT_PATTERN = re.compile(
-    r"(model|audit).*?\(.*?dialect[^a-z,]+([a-z]*|,)", re.IGNORECASE | re.DOTALL
+    r"(model|audit).*?\(.*?dialect\s+'?([a-z]*)'?", re.IGNORECASE | re.DOTALL
```
tobymao, line 764: does this have performance implications?

georgesittas: Maybe, but I'm not sure how else we can express this. Any ideas?

tobymao (approved):

PR #3925, opened 2025-02-28
IPython display import from core deprecated

simon-pactum:
> DeprecationWarning: Importing display from IPython.core.display is deprecated since IPython 7.14, please import from IPython.display

Does as the deprecation warning suggest, and puts lower bound in ipython version.

---

This turned into an error in IPython 9:

>   File "/opt/venv/lib/python3.13/site-packages/sqlmesh/magics.py", line 11, in <module>
    from IPython.core.display import display
ImportError: cannot import name 'display' from 'IPython.core.display' (/opt/venv/lib/python3.13/site-packages/IPython/core/display.py)

PR #3927, opened 2025-02-28
fix: fix ipython import for new versions

benfdking:


martinburch:
this also needs to come with the version requirement change in setup.py https://github.com/TobikoData/sqlmesh/pull/3925/files

benfdking:
> this also needs to come with the version requirement change in setup.py https://github.com/TobikoData/sqlmesh/pull/3925/files

@martinburch 🙏

georgesittas (approved):

PR #3933, opened 2025-03-04
Feat: Adds support for macros in model_defaults and conditional properties

themisvaltinos:
This update adds: 

- Macro support in `model_defaults` attributes.
- The ability to conditionally enable these, as well as for `physical_properties`, `session_properties`, `virtual_properties`.  
- For `session_properties`, `virtual_properties` rendering is now moved to runtime to align with `physical_properties`. 
- Introduces the `model_kind_name` macro variable, which holds the name of the current model kind.

This macro is intended for example to disable from `physical_properties` the `creatable_type` for your project's VIEW-type models and set it to `TRANSIENT` for the rest of the model kinds, by having in `model_defaults`:

```yaml
model_defaults:
  dialect: snowflake
  start: 2022-01-01
  physical_properties:
    creatable_type: "@IF(@model_kind_name != 'VIEW', 'TRANSIENT', NULL)"
```

### DIFF
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -10,8 +10,6 @@ from sqlmesh.core.model.kind import (
     model_kind_validator,
     on_destructive_change_validator,
 )
-from sqlmesh.core.node import IntervalUnit
-from sqlmesh.utils.date import TimeLike
 from sqlmesh.core.model.meta import FunctionCall
 from sqlmesh.utils.pydantic import field_validator
 
@@ -48,7 +46,7 @@ class ModelDefaultsConfig(BaseConfig):
     dialect: t.Optional[str] = None
     cron: t.Optional[str] = None
     owner: t.Optional[str] = None
-    start: t.Optional[TimeLike] = None
+    start: t.Optional[t.Any] = None
     table_format: t.Optional[str] = None
     storage_format: t.Optional[str] = None
     on_destructive_change: t.Optional[OnDestructiveChange] = None
@@ -56,10 +54,10 @@ class ModelDefaultsConfig(BaseConfig):
     virtual_properties: t.Optional[t.Dict[str, t.Any]] = None
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
     audits: t.Optional[t.List[FunctionCall]] = None
-    optimize_query: t.Optional[bool] = None
-    allow_partials: t.Optional[bool] = None
-    interval_unit: t.Optional[IntervalUnit] = None
-    enabled: t.Optional[bool] = None
+    optimize_query: t.Optional[t.Any] = None
+    allow_partials: t.Optional[t.Any] = None
+    interval_unit: t.Optional[t.Any] = None
+    enabled: t.Optional[t.Any] = None
 
     _model_kind_validator = model_kind_validator
     _on_destructive_change_validator = on_destructive_change_validator
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -11,6 +11,7 @@ from sqlmesh.core.model.kind import (
     on_destructive_change_validator,
 )
 from sqlmesh.core.model.meta import FunctionCall
+from sqlmesh.core.node import IntervalUnit
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.pydantic import field_validator
 
@@ -55,10 +56,10 @@ class ModelDefaultsConfig(BaseConfig):
     virtual_properties: t.Optional[t.Dict[str, t.Any]] = None
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
     audits: t.Optional[t.List[FunctionCall]] = None
-    optimize_query: t.Optional[t.Any] = None
-    allow_partials: t.Optional[t.Any] = None
-    interval_unit: t.Optional[t.Any] = None
-    enabled: t.Optional[t.Any] = None
+    optimize_query: t.Optional[t.Union[str, bool]] = None
+    allow_partials: t.Optional[t.Union[str, bool]] = None
+    interval_unit: t.Optional[t.Union[str, IntervalUnit]] = None
+    enabled: t.Optional[t.Union[str, bool]] = None
 
     _model_kind_validator = model_kind_validator
     _on_destructive_change_validator = on_destructive_change_validator
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -11,6 +11,7 @@ from sqlmesh.core.model.kind import (
     on_destructive_change_validator,
 )
 from sqlmesh.core.model.meta import FunctionCall
+from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.pydantic import field_validator
 
 
@@ -46,7 +47,7 @@ class ModelDefaultsConfig(BaseConfig):
     dialect: t.Optional[str] = None
     cron: t.Optional[str] = None
     owner: t.Optional[str] = None
-    start: t.Optional[t.Any] = None
+    start: t.Optional[TimeLike] = None
     table_format: t.Optional[str] = None
     storage_format: t.Optional[str] = None
     on_destructive_change: t.Optional[OnDestructiveChange] = None
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2278,7 +2278,7 @@ def create_python_model(
         }
 
     used_variables = {k: v for k, v in (variables or {}).items() if k in referenced_variables}
-    if variables:
+    if used_variables:
         python_env[c.SQLMESH_VARS] = Executable.value(used_variables)
 
     return _create_model(
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -661,7 +661,6 @@ class _Model(ModelMeta, frozen=True):
         def _render(expression: exp.Expression) -> exp.Expression | None:
             # note: we use the _statement_renderer instead of _create_renderer because it sets model_fqn which
             # in turn makes @this_model available in the evaluation context
-            render_kwargs["model_kind_name"] = self.kind.name
             rendered_exprs = self._statement_renderer(
                 exp.maybe_parse(expression.this)
                 if isinstance(expression, exp.Literal)
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2012,9 +2012,10 @@ def load_sql_based_model(
     unrendered_merge_filter = None
 
     for prop in meta.expressions:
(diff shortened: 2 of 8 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -657,15 +657,23 @@ def render_merge_filter(
...
     def render_physical_properties(self, **render_kwargs: t.Any) -> t.Dict[str, exp.Expression]:
-        def _render(expression: exp.Expression) -> exp.Expression:
+        def _render(expression: exp.Expression) -> exp.Expression | None:
             # note: we use the _statement_renderer instead of _create_renderer because it sets model_fqn which
             # in turn makes @this_model available in the evaluation context
-            rendered_exprs = self._statement_renderer(expression).render(**render_kwargs)
-
-            if not rendered_exprs:
-                raise SQLMeshError(
+            render_kwargs["model_kind_name"] = self.kind.name
```
izeigerman, line 663: We should make this available not just for physical properties, but more generally too. Can we add this where we add `this_model`?

izeigerman: Looks like we already do this below. Why do we need this here?

sqlmesh/core/model/definition.py
```diff
@@ -657,15 +657,23 @@ def render_merge_filter(
...
+            rendered_exprs = self._statement_renderer(
+                exp.maybe_parse(expression.this)
+                if isinstance(expression, exp.Literal)
+                and (d.SQLMESH_MACRO_PREFIX in expression.this)
+                else expression
+            ).render(**render_kwargs)
+
+            # Warn instead of raising for cases where a property is conditionally assigned
+            if not rendered_exprs or rendered_exprs[0].sql().lower() in {"none", "null"}:
```
izeigerman, line 672: I don't get it. Isn't NULL an expression too?

tests/core/test_model.py
```diff
@@ -3512,6 +3524,60 @@ def test_project_level_properties(sushi_context):
     assert model_2.cron == "@daily"
 
 
+def test_conditional_physical_properties(sushi_context):
+    model_defaults = ModelDefaultsConfig(
+        physical_properties={
+            "creatable_type": "@IF(@model_kind_name != 'VIEW', 'TRANSIENT', NULL)"
```
izeigerman, line 3530: While we're at it can we add a test which calls a macro in some other attributes of model defaults? Eg. `virtual_proeprties` and `enabled`

sqlmesh/core/model/definition.py
themisvaltinos: Yes, for the cases where it's Null expression, ie when returning None from the render this check would suffice:
```
isinstance(properties[k].expression, exp.Null)
```
but if a user wants to unset in a SQL model in the ddl a certain top-level property and does
```
virtual_properties (
  creatable_type = None
)
```
here we'd have a column where the identifier is None. So this would have to be written to cover both cases as:
```
isinstance(properties[k].expression, exp.Null) or properties[k].expression.this.this == 'None'
```
or something similar. This is why I opted for that check for both cases as being simpler, but im happy to change if there's some better option

tests/core/test_model.py
themisvaltinos: These attributes aren’t rendered in the `model_defaults` only if they were defined in the sql or python models the macros are resolved. Sure should I extend support to render these from model_defaults as well as part of this PR?

izeigerman: Yes. Otherwise this introduces inconsistent behavior

themisvaltinos: Yes makes sense will extend the support for the rest as well

sqlmesh/core/model/definition.py
izeigerman: I think I get it now, thanks

themisvaltinos:
Thanks @izeigerman addressed comments and generalised logic for all model defaults and rendering of the three kind of properties during runtime for consistency, if you want to have another look

sqlmesh/core/model/definition.py
```diff
@@ -674,7 +679,20 @@ def _render(expression: exp.Expression) -> exp.Expression:
...
+        return {
+            k: rendered
+            for k, v in properties.items()
+            if (rendered := (_render(v) if isinstance(v, exp.Expression) else v))
+        }
+
+    def render_physical_properties(self, **render_kwargs: t.Any) -> t.Dict[str, t.Any]:
+        return self.render_properties(properties=self.physical_properties, **render_kwargs)
+
+    def render_virtual_properties(self, **render_kwargs: t.Any) -> t.Dict[str, t.Any]:
```
izeigerman, line 691: @erindru didn't you argue that virtual properties should be rendered at load time?

sqlmesh/core/config/model.py
```diff
@@ -49,19 +47,19 @@ class ModelDefaultsConfig(BaseConfig):
...
     physical_properties: t.Optional[t.Dict[str, t.Any]] = None
     virtual_properties: t.Optional[t.Dict[str, t.Any]] = None
     session_properties: t.Optional[t.Dict[str, t.Any]] = None
     audits: t.Optional[t.List[FunctionCall]] = None
-    optimize_query: t.Optional[bool] = None
-    validate_query: t.Optional[bool] = None
-    allow_partials: t.Optional[bool] = None
-    interval_unit: t.Optional[IntervalUnit] = None
-    enabled: t.Optional[bool] = None
+    optimize_query: t.Optional[t.Any] = None
```
izeigerman, line 58: Why did we erase the types entirely? Shouldn't they be `t.Union[str, bool]` etc?

izeigerman: When do we validate that the value passed here is correct?

sqlmesh/core/model/definition.py
```diff
@@ -2604,17 +2646,62 @@ def render_field_value(value: t.Any) -> t.Any:
...
+        jinja_macros=jinja_macros,
+        variables=variables,
+        path=path,
+        dialect=dialect,
+        default_catalog=default_catalog,
+    )
+
+    for prop in PROPERTIES:
+        if default_properties := rendered_defaults.get(prop):
+            for key, value in default_properties.items():
```
izeigerman, line 2698: I'm having a hard time understanding what this block does

sqlmesh/core/model/definition.py
themisvaltinos: I think we were discussing that for virtual properties there wasn't a use case for it yet like a specific use case for physical_properties so they don't have to be rendered at runtime. Unless you think @erindru that indeed virtual properties should be rendered at load time?

sqlmesh/core/config/model.py
themisvaltinos: Yes you're right will need to adjust the type and add validation somewhere. I suppose after the rendering since it can't occur before resolving the macros

sqlmesh/core/model/definition.py
themisvaltinos: it prepares the properties if they contain macros and they're string so that when they're rendered they're expressions. in hindsight though this is a bad place for it it should be moved inside the render_properties

sqlmesh/core/model/definition.py
erindru: I dont have a strong opinion. I needed `physical_properties` to be rendered at creation/evaluation time because I needed access to `@this_model` for a specific use case, but I didn't touch `virtual_properties` at the time because there was no clear reason to.

themisvaltinos: Right don't have a strong opinion either, this allows the use of `this_model` or `model_kind_name` in virtual properties as well, but at the same time I don't know how this could be helpful. So happy to move it back to load time

sqlmesh/core/model/definition.py
themisvaltinos: Since it had to occur during loading and not runtime later in `render_properties` so that we populate the python env so I factored it out to a separate function with a more descriptive name for its purpose

sqlmesh/core/config/model.py
themisvaltinos: Revised it and added validation so that after rendering the values to raise if the values are not of the correct type

themisvaltinos:
Addressed comments @izeigerman  regarding `virtual_properties` and `session_properties` let me know if you'd prefer to move them back to loading time

sqlmesh/core/model/definition.py
```diff
@@ -657,16 +658,20 @@ def render_merge_filter(
             raise SQLMeshError(f"Expected one expression but got {len(rendered_exprs)}")
         return rendered_exprs[0].transform(d.replace_merge_table_aliases)
 
-    def render_physical_properties(self, **render_kwargs: t.Any) -> t.Dict[str, exp.Expression]:
-        def _render(expression: exp.Expression) -> exp.Expression:
+    def render_properties(
```
izeigerman, line 661: Shouldn't this be private?

sqlmesh/core/model/definition.py
```diff
@@ -2587,15 +2631,15 @@ def render_field_value(value: t.Any) -> t.Any:
                 dialect=dialect,
                 default_catalog=default_catalog,
             )
-            if rendered_expr is None:
-                raise SQLMeshError(
-                    f"Failed to render model attribute `{fields['name']}` at `{path}`\n"
-                    f"'{expression.sql(dialect=dialect)}' must return an expression"
+            # Warn instead of raising for cases where an attribute is conditionally assigned
```
izeigerman, line 2634: I don't think we support this behavior for all meta fields, are we not? Eg. if the macro returns NULL for something like `allow_partials` we do want to fail, don't we? Should we then distinguish between macro not returning anything and macro returning `NULL`? The former could be a bug, the latter seems intentional.

izeigerman: Looks like this validation happens later in `render_model_defaults`. Please correct me if I'm wrong

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1371,7 +1378,7 @@ def promote(
             exp.select("*").from_(table_name, dialect=self.adapter.dialect),
             table_description=model.description if is_prod else None,
             column_descriptions=model.column_descriptions if is_prod else None,
-            view_properties=model.virtual_properties,
+            view_properties=model.render_virtual_properties(**kwargs),
```
izeigerman, line 1381: Let's not use `kwargs`. Let's make an explicit dict `render_kwargs` instead.

izeigerman (approved):
Please address the remaining comments. LGTM otherwise

sqlmesh/core/model/definition.py
themisvaltinos: Right the validation happens in render_model_defaults and it will be a bug if the rendering didn't return anything at all here, so revised to error out in that case and not for the intentional set to null by the user

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: Sure revised this

PR #3935, opened 2025-03-04
perf: query tables with limit 0 when retrieving schema for tables wit…

lenare1K5:
…h >3 identifiers


tobymao:
@lenare1K5 can you please sign the cla?

lenare1K5:
@tobymao I signed the cla 👍 

Should I also write some tests for my change?

georgesittas:
> Should I also write some tests for my change?

There's already a test for this here: https://github.com/TobikoData/sqlmesh/blob/5ec13d29946b3903ae49de8bef650c0b37bac316/tests/core/engine_adapter/integration/test_integration_bigquery.py#L187

As long as you've verified that the `LIMIT 0` approach works, I expect that this will pass.

georgesittas (approved):

PR #3941, opened 2025-03-04
Feat!: Allow opting out of the time_column being automatically added to the partitioned_by columns

erindru:
Currently, if you have a model something like:

```
   time_column ds
)
```

Then the `ds` column would be automatically added to the `partitioned_by` columns whether you wanted it to or not.

This PR allows you to opt-out with the following syntax:

```
   time_column ds,
   partition_by_time_column false
)
```

(the default is `true` to keep the existing behaviour)

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1180,12 +1180,18 @@ class _Model(ModelMeta, frozen=True):
     def partitioned_by(self) -> t.List[exp.Expression]:
         """Columns to partition the model by, including the time column if it is not already included."""
         if self.time_column and not self._is_time_column_in_partitioned_by:
-            return [
-                TIME_COL_PARTITION_FUNC.get(self.dialect, lambda x, y: x)(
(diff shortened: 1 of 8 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1180,12 +1180,18 @@ def full_depends_on(self) -> t.Set[str]:
...
-            return [
-                TIME_COL_PARTITION_FUNC.get(self.dialect, lambda x, y: x)(
-                    self.time_column.column, self.columns_to_types
-                ),
-                *self.partitioned_by_,
-            ]
+            # This allows the user to opt out of automatic time_column injection
+            # by setting `partition_by_time_column false` on the model kind
+            if (
+                hasattr(self.kind, "partition_by_time_column")
```
erindru, line 1186: Note that this is deliberately a `hasattr` check so I can utilize this property on a custom materialization

izeigerman (approved):

Issue #3686, opened 2025-01-22
Detect and warn when regular model name also in external models


treysp:
A project has a model named `example.model`. Someone accidentally adds `example.model` to the external models config file.

Currently, this generates unclear output - sqlmesh should detect and warn so users can resolve.

izeigerman:
@treysp what output is generated today?

georgesittas:
So given this entry in `external_models.yaml` of the example project:

```yaml
- name: sqlmesh_example.full_model
  start: 1 week ago
  columns:
    customer_id: int
    zip: text
```

The resulting error is:

```
sqlmesh.utils.errors.TestError: Model '"db"."sqlmesh_example"."full_model"' is an unsupported model type for testing at /<path>/playground/tests/test_full_model.yaml
```

I believe this happens because we're using `update` in this section: https://github.com/TobikoData/sqlmesh/blob/cc0653c351232d917c103212aa497d094b9a0001/sqlmesh/core/loader.py#L387-L389

So the SQL kind is overwritten by the external kind, which can't be unit-tested, hence that cryptic error message. Is the usage of `update` in the linked section intentional? I'd assume that we want to enforce the unique name constraint for models across all kinds.

lafirm:
@tobymao, guess, it can be closed now :)

georgesittas:
Appreciate your working on this @lafirm!

PR #3945, opened 2025-03-05
Fix: handle duplicate model names across sql, python and external models

lafirm:
Recently, I encountered an issue where I mistakenly named my Python model the same as an SQL model, resulting in the SQL model being overwritten by the Python model. I believe we should not allow users to have multiple models with the same name across all types – SQL, Python, and external models (currently, duplicate names are only handled within the same type, i.e., you can't create two or more SQL models with the same name or two or more Python models with the same name because of UniqueKeyDict Class).

I found that this issue has already been raised [here](https://github.com/TobikoData/sqlmesh/issues/3686), so I made some assumptions and implemented a simple fix. Please review and let me know your feedback. Thanks!


izeigerman (approved):

sqlmesh/core/loader.py
```diff
@@ -386,11 +386,18 @@ def _load_models(
...
-        models.update(self._load_python_models(macros, jinja_macros, audits, signals))
+        sql_models = self._load_sql_models(macros, jinja_macros, audits, signals)
+        external_models = self._load_external_models(audits, gateway)
+        python_models = self._load_python_models(macros, jinja_macros, audits, signals)
 
-        return models
+        all_model_names = (
+            list(sql_models.keys()) + list(external_models.keys()) + list(python_models.keys())
+        )
+        duplicates = [name for name, count in t.Counter(all_model_names).items() if count > 1]
```
georgesittas, line 396: I believe we should be importing `Counter` from `collections`, because `typing.Counter` is [deprecated](https://docs.python.org/3/library/typing.html#typing.Counter).

georgesittas (commented):
Let's add a test for this.

tests/core/test_loader.py
```diff
@@ -0,0 +1,101 @@
...
+    context, _ = init_and_plan_context("examples/sushi")
+
+    write_model(model["contents"], context.path / model["path"])
+    write_model(model["contents"], context.path / duplicate_model_path(model["path"]))
+
+    with pytest.raises(
+        ValueError,
+        match=r'Duplicate key \'"memory"."sushi"."test_model"\' found in UniqueKeyDict<models>. Call dict.update\(\.\.\.\) if this is intentional.',
+    ):
+        context.load()
```
lafirm, line 76: unit test fails because we are not getting the error expected for the python models. but locally, we are getting the same error when we create 2 python models with the same name. @georgesittas , @izeigerman please let me know if you have any idea on why it behaves like this.

sqlmesh/core/loader.py
```diff
@@ -396,11 +396,18 @@ def _load_models(
...
-        models.update(self._load_external_models(audits, gateway))
-        models.update(self._load_python_models(macros, jinja_macros, audits, signals))
+        sql_models = self._load_sql_models(macros, jinja_macros, audits, signals)
+        external_models = self._load_external_models(audits, gateway)
+        python_models = self._load_python_models(macros, jinja_macros, audits, signals)
 
-        return models
+        all_model_names = (
+            list(sql_models.keys()) + list(external_models.keys()) + list(python_models.keys())
+        )
```
georgesittas, line 405: I believe these `keys()` calls are not necessary here:

```suggestion
        all_model_names = list(sql_models) + list(external_models) + list(python_models)
```

georgesittas (approved):

lafirm:
@georgesittas, made the changes recommended and renamed slow to cicd tests in circleci which I felt irrelevant. Please let me know what you think. Thanks :)

lafirm:
> Let's revert that change for now please.
alright, consider it done. Thanks and apologies for that 😊

PR #3946, opened 2025-03-05
Feat: separate BQ user auth from account impersonation

treysp:
Authenticating a user's credentials and impersonating a service account are two separate actions that the initial implementation incorrectly bundled

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -912,6 +912,7 @@ class BigQueryConnectionConfig(ConnectionConfig):
     def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         """The static connection kwargs for this connection"""
         import google.auth
+        from google.auth import impersonated_credentials
         from google.api_core import client_info, client_options
         from google.oauth2 import credentials, service_account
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -948,6 +937,15 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         else:
             raise ConfigError("Invalid BigQuery Connection Method")
 
+        if self.impersonated_service_account:
+            from google.auth import impersonated_credentials
```
eakmanrq, line 941: Might as well move this up with the other google imports. Feels strange having it be imported when needed when the others are imported at the top of the method

sqlmesh/core/config/connection.py
```diff
@@ -948,6 +937,15 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
...
+
+            creds = impersonated_credentials.Credentials(
+                source_credentials=creds,
+                target_principal=self.impersonated_service_account,
+                target_scopes=self.scopes,
```
eakmanrq, line 946: Is it correct to assume that the scopes provided apply to both the authentication and then the impersonation? I really don't know - just wanting to make sure we thought about this. The alternative would be to add a impersonated_service_account_scopes but we should only do that if it makes sense. If not sure we can leave it as is.

eakmanrq (approved):

sqlmesh/core/config/connection.py
treysp: Did some research, and i'm not sure - will leave as is for now

PR #3787, opened 2025-02-05
Feat!: Introduce Model linting

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3753

This PR introduces a new interface for a general purpose `Model` linter:
- Each `Rule` checks for a specific lint error; If a violation is found, a `RuleViolation` is created with the proper context
- Rules are grouped in `RuleSet`s; Each `RuleSet` is a mapping from `Rule` class name -> `Rule` class.
- The `Linter` applies the `RuleSets` on the model and outputs warnings / raises errors depending on their severity

The linter configuration along with the default values is the following:

```YAML
gateways:
 ...

default_gateway: dev

model_defaults:
  ...

linter:
  enabled: False
  
  # Can be overridden with "ALL" or ["rule0", "rule1", ...., "ruleN"]
  rules: []
  warn_rules:
  ignored_rules: ["noselectstar", "ambiguousorinvalidcolumn", "invalidselectstarexpansion"]
```

This is to preserve the behavior of `QueryRenderer` today; For this matter, `validate_query` is deprecated and instead the corresponding checks are implemented as rules in the Linter.

On a model granularity, rules can be ignored by the new `ignored_rules` attribute:

```SQL
MODEL (
  name sqlmesh_example.test_model,
  kind FULL,
  ignored_rules ["noselectstar"]
);

SELECT ...;
```

Users can implement their own lint rules in a `linter/` directory by subclassing `Rule`; These will be automatically picked up at load time:

```Python3
# linter/user.py

import typing as t

from sqlmesh.core.linter.rule import Rule, RuleViolation, RuleSet
from sqlmesh.core.model import Model

class NoMissingOwner(Rule):
    """Model owner always should be specified."""

    def check_model(self, model: Model) -> t.Optional[RuleViolation]:
        return self.violation() if not model.owner else None 
```

PR #3950, opened 2025-03-06
Chore: Add Linter documentation

VaggelisD:
Documentation of the linter project as it's being worked on here https://github.com/TobikoData/sqlmesh/pull/3787


docs/concepts/linter.md
```diff
@@ -0,0 +1,83 @@
...
+        """A summary of what this rule checks for."""
+        return self.__doc__ or ""
+
+    def violation(self, violation_msg: t.Optional[str] = None) -> RuleViolation:
+        """Create a RuleViolation instance for this rule"""
+        return RuleViolation(rule=self, violation_msg=violation_msg or self.summary)
+
+```
+
+Thus, each `Rule` can be broken down to it's vital components:
```
sungchun12, line 30:
```suggestion
Thus, each `Rule` can be broken down to its vital components:
```

docs/concepts/linter.md
```diff
@@ -0,0 +1,83 @@
...
+- The name (or code) of the rule is it's class name in lowercase.
```
sungchun12, line 31:
```suggestion
- The name (or code) of the rule is its class name in lowercase.
```

docs/concepts/models/overview.md
```diff
@@ -443,8 +443,8 @@ to `false` causes SQLMesh to disable query canonicalization & simplification. Th
 !!! warning
     Turning off the optimizer may prevent column-level lineage from working for the affected model and its descendants, unless all columns in the model's query are qualified and it contains no star projections (e.g. `SELECT *`).
 
-### validate_query
-:   Whether the model's query will be validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings. This will display invalid columns in your SQL statements along with models containing `SELECT *` that cannot be automatically expanded to list out all columns. This ensures SQL is verified locally before time and money are spent running the SQL in your data warehouse.
+### ignored_rules
```
sungchun12, line 446: did you deprecate `validate_query`? If yes, can you add a disclaimer that it doesn't work as of xyz version?

docs/guides/configuration.md
```diff
@@ -1109,6 +1109,63 @@ def grant_schema_usage(evaluator):
...
+
+=== "Python"
+
+    ```python linenums="1"
+    from sqlmesh.core.config import Config, LinterConfig
+
+    config = Config(
+         linter=LinterConfig(
+            enabled=True,
+            rules=["rule1", "rule2"],
```
sungchun12, line 1137: Update these with the actual rule names

docs/guides/configuration.md
```diff
@@ -1109,6 +1109,63 @@ def grant_schema_usage(evaluator):
...
+
+It can be configured under the `linter` key, with the rules being defined as lists of rule names or `"ALL"`. An example:
+
+=== "YAML"
+
+    ```yaml linenums="1"
+    linter:
+        enabled: True
+
+        rules: ["rule1", "rule2"]
```
sungchun12, line 1124: Update these with the actual rule names

docs/guides/configuration.md
```diff
@@ -1109,6 +1109,63 @@ def grant_schema_usage(evaluator):
...
+
+
+=== "YAML"
+
+    ```yaml linenums="1"
+    linter:
+        enabled: True
+
+        rules: "ALL"
+        ignored_rules: ["rule4"]
```
sungchun12, line 1163: Update these with the actual rule names

sungchun12 (changes requested):
Overall good enough.

docs/concepts/models/overview.md
VaggelisD: Yeah, `validate_query` was essentially raising the 2 renderer rules for ambiguous/invalid columns and invalid select star expansion. 

Now, they are 2 distinct built-in rules.

docs/reference/model_configuration.md
```diff
@@ -39,7 +39,7 @@ Configuration options for SQLMesh model properties. Supported by all model kinds
...
 | `optimize_query`             | Whether the model's query should be optimized. This attribute is `true` by default. Setting it to `false` causes SQLMesh to disable query canonicalization & simplification. This should be turned off only if the optimized query leads to errors such as surpassing text limit.                                                                                                                                                                                                      |       bool        |    N     |
-| `validate_query`             | Whether the model's query will be strictly validated at compile time. This attribute is `false` by default. Setting it to `true` causes SQLMesh to raise an error instead of emitting warnings. This will display invalid columns in your SQL statements along with models containing `SELECT *` that cannot be automatically expanded to list out all columns.                                                                                                                                                                                              |       bool        |    N     |
```
themisvaltinos, line 42: `validate_query` can be removed from the model defaults in line 154 as well

docs/concepts/linter.md
```diff
@@ -0,0 +1,118 @@
+# Linter
+
+Linting enables you to validate the model definition, ensuring it adheres to best practices. When a new [plan](plans.md) is created, each model is checked against a set of built-in and user-defined rules to verify that its attributes comply with the defined standards; This improves code quality, consistency, and helps to detect issues early in the development cycle.
+
+For more information regarding linter configuration visit the relevant [guide here](../guides/configuration.md).
+
+# Rules
+
+Each rule is responsible for detecting a specific issue or pattern within a model. Rules are defined as classes that implement the logic for validation by subclassing `Rule` (redacted form):
```
georgesittas, line 9: What does "redacted form" mean here?

```suggestion
Each rule is responsible for detecting a specific issue or pattern in a model. Rules are defined as classes that implement the logic for validation by subclassing `Rule` (redacted form):
```

docs/concepts/linter.md
```diff
@@ -0,0 +1,118 @@
...
+Thus, each `Rule` can be broken down to its vital components:
+- The name (or code) of the rule is defined as its class name in lowercase.
+- The core logic is implemented in `Rule::check_model(...)` which can analyze any `Model` attribute.
+- If an issue is detected, a `RuleViolation` object should be returned with the proper context. This can be created manually or through the `Rule::violation()` helper.
+- A short explanation of the rule's purpose should be added in the form of a class docstring or by subclassing `Rule::summary`.
+
+
+
+## Built-in
+SQLMesh comes with a set of predefined rules which check for potential SQL errors or enforce stylistic opinions. An example of the latter the `NoSelectStar` rule which prohibits users from writing `SELECT *` in the outer-most select:
```
georgesittas, line 39:
```suggestion
SQLMesh comes with a set of predefined rules which check for potential SQL errors or enforce stylistic opinions. An example of the latter is the `NoSelectStar` rule, which prohibits users from writing `SELECT *` in the outer-most select:
```

docs/concepts/linter.md
```diff
@@ -0,0 +1,118 @@
...
+
+| Name                                 | Check       | Explanation
+|--------------------------------------|----------------------------------------------------------------------------------------------------------------------|
+| ambiguousorinvalidcolumns            | Correctness | The optimizer was unable to trace or found duplicate columns                                           |
+| invalidselectstarexpansion           | Correctness | The optimizer was unable to expand the top-level `SELECT *`                                            |
+| noselectstar                         | Stylistic   | The query's top-level selection should not be `SELECT *`, even if it can be expanded by the optimizer |
+
+
+## User defined rules
+Users can implement their own custom rules in a similar fashion. For that matter, SQLMesh will load any subclass of `Rule` under the `linter/` directory.
```
georgesittas, line 65:
```suggestion
Users can implement their own custom rules in a similar fashion. SQLMesh will load any subclass of `Rule` under the `linter/` directory.
```

docs/concepts/linter.md
```diff
@@ -0,0 +1,118 @@
...
+
+For instance, if an organization wanted to ensure all models are owned by an engineer, one solution would be to implement the following check:
```
georgesittas, line 67:
```suggestion
For instance, if an organization wanted to ensure all models have an owner, one solution would be to implement the following check:
```

docs/concepts/linter.md
```diff
@@ -0,0 +1,118 @@
+# Linter
+
+Linting enables you to validate the model definition, ensuring it adheres to best practices. When a new [plan](plans.md) is created, each model is checked against a set of built-in and user-defined rules to verify that its attributes comply with the defined standards; This improves code quality, consistency, and helps to detect issues early in the development cycle.
```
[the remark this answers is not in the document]
VaggelisD: > This isn't plan-specific, right? Linting happens whenever the project is loaded.

Yeah, I also mention elsewhere that this happens during loading but I suspect that most users associate that with `sqlmesh plan` since it's the main command they'd run; Not sure if `loading` as a stage is brought up as often.

docs/concepts/linter.md
VaggelisD: > What does "redacted form" mean here?

I wanted to provide the general gist of `Rule` without copy-pasting code per se, since a lot of the code would be implementation details that are bound to change. If some users want to take that a step further, they can visit the source code.

docs/concepts/linter.md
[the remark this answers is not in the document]
georgesittas: You could phrase this simply as "when a project is loaded in SQLMesh" to avoid ambiguities and terms that may not be clear.

@sungchun12 thoughts?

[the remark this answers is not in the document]
VaggelisD: > You could phrase this simply as "when a project is loaded in SQLMesh" to avoid ambiguities and terms that may not be clear.

That looks cool yeah, still wondering if the stages are documented enough for the casual user, Sung's input would be helpful here.

docs/concepts/linter.md
sungchun12: If that's the case, replace it with "quick snippets" and add a hyperlink to see the full source code.

georgesittas: I think "(redacted form)" is unnecessary here, I was confused when I read it.

Another way to phrase this could be:

"Rules are defined as classes that implement the logic for validation. Each rule needs to inherit from the base `Rule` class: [...]"

georgesittas: Even better yeah, +1 to hyperlinking so that we don't risk this section getting stale if it's an interface that users need to follow.

sungchun12 (approved):
amazing

PR #3951, opened 2025-03-06
test: reproduce scheduler issue

plaflamme:


plaflamme:
The bug is real, though this repro is wrong... The issue stems from [this iteration](https://github.com/TobikoData/sqlmesh/blob/f91a9cab42adb40772f41b11002beec46256e876/sqlmesh/core/scheduler.py#L403C36-L403C62) not going in topological order which leads to processing children before their parents in some cases, missing some "unready" intervals on [this line](https://github.com/TobikoData/sqlmesh/blob/f91a9cab42adb40772f41b11002beec46256e876/sqlmesh/core/scheduler.py#L420)

PR #3953, opened 2025-03-06
Chore!: Switch from setup.py to pyproject.toml

erindru:
`pip install` has been warning us for months about things like:

```
DEPRECATION: Legacy editable install of sqlmesh is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to add a pyproject.toml
```

It's only a matter of time before this causes major problems at an inopportune moment, so to head it off i've added a `pyproject.toml` and removed the `setup.py` config.

The packages can now be built with `python -m build` and the build occurs in an isolated environment. I verified that the resulting wheel and sdist contained the same contents as the direct setuptools version, in particular that the wheel only contained the compiled UI assets and not the UI source code (which makes it ~5mb instead of ~10mb).

I also ported `sqlmesh-tests` because it was running the parent sqlmesh `setup.py` in order to extract dependencies. I changed it to use the `toml` library to load the sqlmesh `pyproject.toml` and just grab the dependencies that way.

Note that I havent moved any of the mypy config from `setup.cfg` and into pyproject.toml, I was just focusing on packaging.


pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+    "ipywidgets",
+    "jinja2",
+    "pandas",
+    "pydantic>=2.0.0",
+    "requests",
+    "rich[jupyter]",
+    "ruamel.yaml",
+    "sqlglot[rs]~=26.8.0",
+    "tenacity",
+    "time-machine",
```
izeigerman, line 26: Shouldn't this be a dev dependency?

georgesittas: It's used in the testing module: https://github.com/TobikoData/sqlmesh/blob/e100fbf126761a86004fd50d235798e64977c562/sqlmesh/core/test/definition.py#L678-L680

izeigerman (approved):

PR #3954, opened 2025-03-07
Fix: State sync snapshot count() quoting

erindru:
PR #3903 introduced a subtle bug where the snapshot count queries stopped being quoted. This meant that on engines like Snowflake, `_snapshots` was interpreted as `_SNAPSHOTS` which caused migration to fail.

This PR fixes the issue and adds a test to detect this in future

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -314,10 +314,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
+          # branches:
+          #   only:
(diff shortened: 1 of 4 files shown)


.circleci/continue_config.yml
```diff
@@ -314,10 +314,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
```
erindru, line 317: I'll uncomment this prior to merge, want to see if the new test picks up any issues on the other cloud engines

tobymao (approved):

PR #3956, opened 2025-03-07
Fix: Batch intervals in DAG order to make sure that parent snapshot signals are respected by their children

izeigerman:
Previously, we were batching intervals in an arbitrary order, which could randomly result in child intervals running even when the same interval was unavailable in its parent due to the signal condition not being met.

This fix ensures that batches are computed in DAG order.

Huge kudos to @plaflamme for providing a test case which reproduces the issue (#3951)


plaflamme (approved):
thanks

georgesittas (approved):

PR #3967, opened 2025-03-10
add nomissingaudits rule

sungchun12:
I added a linting rule for a quick win like making sure audits are configured. This will serve as an example PR for others to contribute built-in rules that the community deems worthy for the masses.


docs/guides/linter.md
```diff
@@ -235,9 +237,9 @@ You may specify that a rule's violation should not error and only log a warning
...
             # but only warn if "invalidselectstarexpansion" is violated
-            warning_rules=["invalidselectstarexpansion"],
+            warn_rules=["invalidselectstarexpansion"],
         )
     )
     ```
 
-SQLMesh will raise an error if the same rule is included in more than one of the `rules`, `warning_rules`, and `ignored_rules` keys since they should be mutually exclusive.
\ No newline at end of file
+SQLMesh will raise an error if the same rule is included in more than one of the `rules`, `warn_rules`, and `ignored_rules` keys since they should be mutually exclusive.
```
sungchun12, line 245: fixed this. `warning_rules` doesn't exist

docs/guides/linter.md
```diff
@@ -1,5 +1,7 @@
 # Linter guide
 
+![Linter](linter_example.png)
```
sungchun12, line 3: get people excited to see this image in their own terminal

tests/core/test_context.py
```diff
@@ -1761,6 +1763,7 @@ def assert_cached_violations_exist(cache: OptimizedQueryCache, model: Model):
     def model3_entrypoint(evaluator: MacroEvaluator) -> str:
         return "select * from model1"
 
+    # breakpoint()
```
treysp, line 1766: A stray found its way in!

treysp (approved):
Great improvements - thank you!

PR #3968, opened 2025-03-11
Fix: Fail the plan command if the selector expression doesn't return any models

izeigerman:
To prevent unexpected behavior and confusing errors we should fail the plan immediately if we discover that the selector expressions provided by the user don't return any models.


erindru (approved):

tobymao (approved):
you didn't want to do this at the selector level and check each node?

izeigerman:
> you didn't want to do this at the selector level and check each node?

@tobymao, I think we want to warn at the selector but decision to fail should be made downstream IMHO

PR #3975, opened 2025-03-11
Fix: Direct breaking category should only impact direct children snapshot

izeigerman:
In a DAG that looks like `A -> B -> C`, if the user makes a `Breaking` change to `A` and a `Non-breaking` change to `B`, then `C` should be categorized as `Indirect Non-Breaking` rather than `Indirect Breaking`, as it was before this change.

At the same time, if `A` was categorized as `Forward-Only` and `B` as `Non-Breaking`, SQLMesh should prefer the `Forward-only` category for `C` over the `Indirect Non-Breaking` one.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -578,21 +578,25 @@ class PlanBuilder:
                     if mode == AutoCategorizationMode.FULL:
                         snapshot.categorize_as(SnapshotChangeCategory.BREAKING)
         elif self._context_diff.indirectly_modified(snapshot.name):
-            categories = []
+            all_upstream_categories = []
+            direct_parent_categories = []
 
             for p_id in dag.upstream(s_id):
                 parent = self._context_diff.snapshots.get(p_id)
 
                 if parent and self._is_new_snapshot(parent):
-                    categories.append(parent.change_category)
+                    all_upstream_categories.append(parent.change_category)
+                    if p_id in snapshot.parents:
+                        direct_parent_categories.append(parent.change_category)
 
-            if not categories or any(
+            if not direct_parent_categories or any(
                 category.is_breaking or category.is_indirect_breaking
-                for category in categories
+                for category in direct_parent_categories
                 if category
             ):
                 snapshot.categorize_as(SnapshotChangeCategory.INDIRECT_BREAKING)
-            elif any(category.is_forward_only for category in categories if category):
+            elif any(category.is_forward_only for category in all_upstream_categories if category):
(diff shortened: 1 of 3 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -578,21 +578,25 @@ def _categorize_snapshot(
                     if mode == AutoCategorizationMode.FULL:
                         snapshot.categorize_as(SnapshotChangeCategory.BREAKING)
         elif self._context_diff.indirectly_modified(snapshot.name):
-            categories = []
+            all_upstream_categories = []
```
tobymao, line 581: this should be sets right?

sqlmesh/core/plan/builder.py
```diff
@@ -578,21 +578,25 @@ def _categorize_snapshot(
...
                 parent = self._context_diff.snapshots.get(p_id)
 
                 if parent and self._is_new_snapshot(parent):
-                    categories.append(parent.change_category)
+                    all_upstream_categories.append(parent.change_category)
+                    if p_id in snapshot.parents:
+                        direct_parent_categories.append(parent.change_category)
 
-            if not categories or any(
+            if not direct_parent_categories or any(
```
tobymao, line 592: if there are no direct parent categories, but then grandparent A is forward only, is this still correct?

sqlmesh/core/plan/builder.py
```diff
@@ -578,21 +578,25 @@ def _categorize_snapshot(
...
                 category.is_breaking or category.is_indirect_breaking
-                for category in categories
+                for category in direct_parent_categories
                 if category
             ):
                 snapshot.categorize_as(SnapshotChangeCategory.INDIRECT_BREAKING)
-            elif any(category.is_forward_only for category in categories if category):
+            elif any(category.is_forward_only for category in all_upstream_categories if category):
```
tobymao, line 598: we can just do set comparison, if ChangeCategory.FORWARD_ONLY in all_upstream_categories

sqlmesh/core/plan/builder.py
izeigerman: How can there be no direct parent categories?

izeigerman: Note: it's direct **parent** not direct **category**

izeigerman: Sorry, let me rephrase the question. How can there be a grandparent category but not a parent category?

tobymao: i’m not sure, is there a case of a revert or something where it’s possible?

Issue #3978, opened 2025-03-12
BigQuery `RANGE` type is inferred to be `JSON` in different places


plaflamme:
The 2 cases where this comes up are:

In a model definition without an explicit cast:
```sql
SELECT RANGE(start::DATE, end::DATE) AS my_range
```

In this case `my_range` will be inferred as `JSON` (it should be `RANGE<DATE>` in this case).

Similarly, when an external table has a `RANGE` column, the `external_models.yml` file will have `JSON` for that column's type.

PR #3981, opened 2025-03-12
Fix!: Type inference for BQ engine adapter

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/3978

In the BigQuery engine adapter, type inference for the fields happens through `to_standard_sql()`; However, it seems like the types that do not belong to stanard SQL are not handled properly (e.g RANGE or JSON, see https://github.com/TobikoData/sqlmesh/commit/4951ff9b539c42242050617869127764d120cf6d) and are instead marked as `TYPE_KIND_UNSPECIFIED`. 

This PR adds a fallback for these types and instead uses `field.field_type`, as was used [before](https://github.com/TobikoData/sqlmesh/pull/1089/files):

```SQL
MODEL (name vaggelis.test);

SELECT 
 RANGE('01-01-1900'::DATE, '01-01-1902'::DATE) AS my_range, 
 JSON '{"id": 10}' AS foo;
```

```Python3
>>> bq_adapter.columns("vaggelis.test");
{'my_range': DataType(
  this=Type.RANGE,
  expressions=[
    DataType(this=Type.DATE, nested=False)],
  nested=True), 'foo': DataType(this=Type.JSON, nested=False)}
```

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -231,7 +231,9 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
     ) -> t.Dict[str, exp.DataType]:
         """Fetches column names and types for the target table."""
 
-        def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
+        def dtype_to_sql(
+            dtype: t.Optional[StandardSqlDataType], field: t.Optional[bigquery.SchemaField] = None
+        ) -> str:
             assert dtype
 
             kind = dtype.type_kind
@@ -249,7 +251,16 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 )
                 return f"STRUCT<{fields}>"
             if kind.name == "TYPE_KIND_UNSPECIFIED":
-                return "JSON"
+                if not field:
+                    return "UNKNOWN"
+
+                field_type = field.field_type
+
+                if field_type == "RANGE":
+                    return f"RANGE<{field.range_element_type.element_type}>"
+
+                return field_type
+
             return kind.name
 
         def create_mapping_schema(
@@ -257,7 +268,7 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
         ) -> t.Dict[str, exp.DataType]:
             return {
                 field.name: exp.DataType.build(
-                    dtype_to_sql(field.to_standard_sql().type), dialect=self.dialect
+                    dtype_to_sql(field.to_standard_sql().type, field), dialect=self.dialect
                 )
                 for field in schema
             }
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -232,9 +232,10 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
         """Fetches column names and types for the target table."""
 
         def dtype_to_sql(
-            dtype: t.Optional[StandardSqlDataType], field: t.Optional[bigquery.SchemaField] = None
+            dtype: t.Optional[StandardSqlDataType], field: bigquery.SchemaField
         ) -> str:
             assert dtype
+            assert field
 
             kind = dtype.type_kind
             assert kind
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -231,7 +231,9 @@ def columns(
     ) -> t.Dict[str, exp.DataType]:
         """Fetches column names and types for the target table."""
 
-        def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
+        def dtype_to_sql(
+            dtype: t.Optional[StandardSqlDataType], field: t.Optional[bigquery.SchemaField] = None
```
georgesittas, line 235: What happens if you have nested data structures, like array of `RANGE` or `JSON`? Don't we need to include `field` in the recursive calls to `dtype_to_sql`?

Not sure if `field` should be optional here.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -249,15 +251,24 @@ def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
                 )
                 return f"STRUCT<{fields}>"
             if kind.name == "TYPE_KIND_UNSPECIFIED":
-                return "JSON"
+                if not field:
+                    return "UNKNOWN"
```
georgesittas, line 255: Should this raise instead? Or just preserve previous behavior by defaulting to `"JSON"`.

cc @tobymao

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -249,15 +251,24 @@ def dtype_to_sql(dtype: t.Optional[StandardSqlDataType]) -> str:
...
+
+                field_type = field.field_type
+
+                if field_type == "RANGE":
+                    return f"RANGE<{field.range_element_type.element_type}>"
```
georgesittas, line 260: Is `RANGE` always a parameterized type, i.e. is `field.range_element_type.element_type` here safe?

VaggelisD: Yeah, from bigquery's `schema.py`:

```
    @property
    def range_element_type(self):
        """Optional[FieldElementType]: The subtype of the RANGE, if the
        type of this field is RANGE.

        Must be set when ``type`` is `"RANGE"`. Must be one of `"DATE"`,
        `"DATETIME"` or `"TIMESTAMP"`.
        """
        if self._properties.get("rangeElementType"):
            ret = self._properties.get("rangeElementType")
            return FieldElementType.from_api_repr(ret)
```

georgesittas: Great, thanks for checking. May be worth leaving a brief comment about this– up to you.

sqlmesh/core/engine_adapter/bigquery.py
VaggelisD: Yeah we can do that. From what I saw:
- When we have a `STRUCT`, the `field` is a `RECORD` type and contains the nested fields in `field.fields`
- When we have an `ARRAY`, the `field` is the array actual field

So we can recurse into `dtype_to_sql` with `field` always set. For instance, take this model:

```
SELECT STRUCT([JSON '{"id": 10}'] as arr) AS col;
```

This will need 3 recursive `dtype_to_sql` calls to unwrap the `STRUCT<ARRAY<JSON>>`, and at each call the field is:

```
At STRUCT: field SchemaField('col', 'RECORD', 'NULLABLE', None, None, (SchemaField('arr', 'JSON', 'REPEATED', None, None, (), None),), None)
At ARRAY: field SchemaField('arr', 'JSON', 'REPEATED', None, None, (), None)
At JSON: field SchemaField('arr', 'JSON', 'REPEATED', None, None, (), None)
```

I tried to replicate this with `RANGE` but I couldn't get it to work if it's nested.

PS: This example actually works on `main` "by accident", since we default to returning "JSON". On my PR, we properly infer `JSON` through `field`

sqlmesh/core/engine_adapter/bigquery.py
VaggelisD: The latest commit always has `field` set, this should not be relevant anymore.

georgesittas (approved):
LGTM, let's test these changes.

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: Seems ok, let's just make sure a few examples with weird nesting are tested. Like struct of struct of array, struct of array, array of struct...

tobymao (approved):

sqlmesh/core/engine_adapter/bigquery.py
VaggelisD: I added some tests but ensuring deeply nested structures are built properly is a bit of a PITA, since e.g with structs we lose the ability of comparing the entire column with `exp.DataType.build(...)`.

Let me know if you have any other test in mind

PR #3982, opened 2025-03-12
Feat: add second level of verbosity

treysp:
By default, the `plan` command CLI output:
- Condenses long lists of indirectly modified models
- Only lists individual models for the model evaluation progress bar

If the `--verbose` flag is passed the CLI output:
- Lists all indirectly modified models
- Lists individual models for both table creation and view promotion progress bars

This PR introduces levels of verbosity, allowing us to unbundle list condensation from progress bar behavior. There are now three levels:
- Default: current default behavior
- `--verbose` flag: list models for all progress bars (but condense indirect model lists)
- `--very-verbose` flag: display full indirectly modified lists, list models for all progress bars


sqlmesh/cli/main.py
```diff
@@ -428,19 +429,27 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
...
 @opt.verbose
+@opt.very_verbose
 @click.pass_context
 @error_handler
 @cli_analytics
 def plan(
-    ctx: click.Context, verbose: bool, environment: t.Optional[str] = None, **kwargs: t.Any
+    ctx: click.Context,
+    verbose: bool,
+    very_verbose: bool,
```
izeigerman, line 439: Let's not do this.  See a better example here: https://click.palletsprojects.com/en/stable/options/#counting. We want one argument, which when repeated results in higher verbosity levels.

sqlmesh/core/console.py
```diff
@@ -1993,7 +1996,10 @@ def show_model_difference_summary(
             self._print("\n**Added Models:**")
             added_models = sorted(added_snapshot_models)
             list_length = len(added_models)
-            if not self.verbose and list_length > self.INDIRECTLY_MODIFIED_DISPLAY_THRESHOLD:
+            if (
+                self.verbosity < Verbosity.VERY_VERBOSE
```
izeigerman, line 2000: Sorry, I don't get it. So if it's very verbose we won't print this? Why not?

izeigerman: Here and below

sqlmesh/core/constants.py
```diff
@@ -91,3 +92,9 @@
 HYBRID = "hybrid"
 
 DISABLE_SQLMESH_STATE_MIGRATION = "SQLMESH__AIRFLOW__DISABLE_STATE_MIGRATION"
+
+
+class Verbosity(IntEnum):
```
izeigerman, line 97: Feel free to put this into `utils/verbosity.py` or `utils/__init__.py`

sqlmesh/core/console.py
treysp: If verbosity is lower than very verbose (is default or verbose), hit this branch and print the abbreviated model list

If verbosity is very verbose, go to the else branch and print every model name

izeigerman: ah, I get it now, thanks

izeigerman (approved):

PR #3991, opened 2025-03-13
Perf: use table name reference directly instead of select * subquery …

### LINKED ISSUE #3990: Audits expanding @this_model into a select * subquery is potentially inefficient
## Context 

We noticed the unique values audit has `@this_model` expanded to a subquery as can be seen below. 
The issue with this, is that using SELECT * in the subquery has the potential to be highly inefficient if the underlying database engine optimizer does not prune the projection. There seems to be little reason to using a subquery instead of directly referencing the table. Can you help us understand if this is an issue and if so how we might go about optimizing this to be less reliant on the DB optimizer. 

### Audit Code

```python
# unique_values(columns=(column_1, column_2))
unique_values_audit = ModelAudit(
    name="unique_values",
    defaults={"condition": exp.true()},
    query="""
SELECT *
FROM (
  SELECT
    @EACH(
      @columns,
      c -> row_number() OVER (PARTITION BY c ORDER BY c) AS rank_@c
    )
  FROM @this_model
  WHERE @condition
)
WHERE @REDUCE(
  @EACH(
    @columns,
    c -> rank_@c > 1
  ),
  (l, r) -> l OR r
)
    """,
)

```

### Expanded Example

```sql
SELECT
  COUNT(*)
FROM (
  SELECT
    *
  FROM (
    SELECT
      ROW_NUMBER() OVER (PARTITION BY "id" ORDER BY "id" NULLS FIRST ) AS "rank_id"
    FROM (
      SELECT
        *
      FROM
        "db"."sqlmesh__mart_common"."mart_common__table__931970220" AS "mart_common__table__931970220") AS "_q_0"
    WHERE
      $1) AS "_q_1"
  WHERE
    "rank_id" > $2) AS "audit"
```

z3z1ma:
…in audit wo where.

Striaghtforward change which allows audits without where conditions added by the snapshot containing a `time_column` to reference the table directly which reduces reliance on black box engine optimizer behavior. 

Resolves #3990

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -527,7 +527,9 @@ class _Model(ModelMeta, frozen=True):
             deployability_index=deployability_index,
             **{
                 **audit.defaults,
-                "this_model": exp.select("*").from_(quoted_model_name).where(where).subquery(),
+                "this_model": exp.select("*").from_(quoted_model_name).where(where).subquery()
+                if where is not None
+                else quoted_model_name,
                 **kwargs,
             },  # type: ignore
         )
--- a/tests/core/test_audit.py
+++ b/tests/core/test_audit.py
@@ -901,7 +901,7 @@ def test_audit_query_normalization():
     )
     assert (
         rendered_audit_query.sql("snowflake")
-        == """SELECT * FROM (SELECT * FROM "DB"."TEST_MODEL" AS "TEST_MODEL") AS "_Q_0" WHERE "A" IS NULL AND TRUE"""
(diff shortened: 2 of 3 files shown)


tobymao:
looks good, let's fix the build

izeigerman:
Looks like this change causes an unexpected metadata update. Probably because we use the rendered audit query to compute a metadata hash. @z3z1ma an empty migration script here https://github.com/TobikoData/sqlmesh/tree/main/sqlmesh/migrations will help the test pass

tobymao:
once i merge mine it’ll have a migration, so don’t worry about it

PR #4000, opened 2025-03-14
Feat: improve CLI evaluation progress bar

treysp:
This PR updates the CLI `plan` model evaluation progress bar output and modifies all progress bar labels.

**Evaluation progress bar**

The evaluation progress bar prints one line of output for each executed batch. 

Currently, it includes the node's batch number, node's name, and batch run time:
`[3/7] sushi.customer_revenue_lifetime evaluated in 0.02s`

The new output includes more information for each batch and is vertically aligned across batches. The column widths are static and specified in the code. If a model's information string is longer than the column width, the information is truncated with 3 dots (dots on left for node names so you see table, right for annotation so you see SQLMesh action).

This annotated screenshot below shows evaluation progress bar output from a modified version of the sushi project. The output includes an annotation of what SQLMesh did for each node:
  1. `raw.demographics` is an external model, and SQLMesh ran its audits
  2. `sushi.added_model` is an hourly incremental by time model processing an interval <24h, so its interval start/end times are displayed
  3. `sushi.raw_marketing` is a full model, so SQLMesh fully refreshed the table
  4. `sushi.waiter_names` is a seed model, so SQLMesh inserted from file
  5. `assert_item_price_above_zero` is a standalone audit, which SQLMesh ran
  6. `sushi.top_waiters` is a view, so SQLMesh recreated the view
  7. If a model has audits, we display the number that passed with a check mark
  8. If a model has failed non-blocking audits, we issue a warning and display the number that failed with a red X

Untitled

Model kind annotations not in screenshot:
- `INCREMENTAL_BY_UNIQUE_KEY`: [insert or update rows]
- `INCREMENTAL_BY_PARTITION`: [insert partitions]

All other model kinds display the processed interval start/ends, including time if the interval is <24h long.

**Progress bar language**

- Creation progress bar
  - Bar label: `Creating physical tables` --> `Updating physical layer`
  - Completion message: `Model versions created successfully` --> `✔ Physical layer updated`
-  Evaluation progress bar
   - Bar label: `Evaluating models` --> `Executing model batches`
   - Completion message: `Model batches executed successfully` --> `✔ Model batches executed`
- Promotion progress bar
  - Bar label: `Virtually Updating 'prod'` --> `Updating virtual layer`
  - Completion message: `Target environment updated successfully` --> `✔ Virtual layer updated`

**Default catalog in display names**

Currently, we only display a model's catalog if it was explicitly specified (and not if SQLMesh used the default catalog). 

This PR updates very verbose mode (`sqlmesh plan -vv`) to include default catalog in all model names.

Screenshot 2025-03-27 at 6 02 38 PM

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -516,9 +516,13 @@ class NoopConsole(Console):
         pass
 
 
-def make_progress_bar(message: str, console: t.Optional[RichConsole] = None) -> Progress:
+def make_progress_bar(
+    message: str,
+    console: t.Optional[RichConsole] = None,
+    justify: t.Literal["default", "left", "center", "right", "full"] = "right",
+) -> Progress:
     return Progress(
-        TextColumn(f"[bold blue]{message}", justify="right"),
+        TextColumn(f"[bold blue]{message}", justify=justify),
         BarColumn(bar_width=PROGRESS_BAR_WIDTH),
         "[progress.percentage]{task.percentage:>3.1f}%",
         "•",
@@ -696,7 +700,7 @@ class TerminalConsole(Console):
                 )
 
                 self.evaluation_progress_live.console.print(
-                    f"{GREEN_CHECK_MARK} {batch}{display_name}{annotation} {duration}"
+                    f"{batch}{display_name}{annotation} {duration}"
                 )
 
             self.evaluation_total_progress.update(
@@ -749,7 +753,7 @@ class TerminalConsole(Console):
         if self.creation_progress is not None and self.creation_task is not None:
             if self.verbosity >= Verbosity.VERBOSE:
                 self.creation_progress.live.console.print(
-                    f"{GREEN_CHECK_MARK} {snapshot.display_name(self.environment_naming_info, self.default_catalog if self.verbosity < Verbosity.VERY_VERBOSE else None, dialect=self.dialect).ljust(self.PROGRESS_BAR_COLUMN_WIDTHS['name'])} [green]created[/green]"
+                    f"{snapshot.display_name(self.environment_naming_info, self.default_catalog if self.verbosity < Verbosity.VERY_VERBOSE else None, dialect=self.dialect).ljust(self.PROGRESS_BAR_COLUMN_WIDTHS['name'])} [green]created[/green]"
                 )
             self.creation_progress.update(self.creation_task, refresh=True, advance=1)
 
@@ -799,16 +803,8 @@ class TerminalConsole(Console):
     ) -> None:
         """Indicates that a new snapshot promotion progress has begun."""
         if self.promotion_progress is None:
-            self.promotion_progress = Progress(
-                TextColumn(
-                    "[bold blue]Updating virtual layer ",  # space to align with other progress bars
-                    justify="right",
-                ),
-                BarColumn(bar_width=PROGRESS_BAR_WIDTH),
-                "[progress.percentage]{task.percentage:>3.1f}%",
-                "•",
-                TimeElapsedColumn(),
-                console=self.console,
+            self.promotion_progress = make_progress_bar(
+                "Updating virtual layer ", self.console, justify="left"
             )
 
             self.promotion_progress.start()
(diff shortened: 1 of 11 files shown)


sqlmesh/core/scheduler.py
```diff
@@ -205,6 +207,7 @@ def evaluate(
             wap_id=wap_id,
             **kwargs,
         )
+        get_console().store_evaluation_audit_results(snapshot, audit_results)
```
erindru, line 210: It's almost like `evaluate()` needs to take a callback to populate the audit results. That way the caller can do something with them before any `AuditError`'s break the control flow below.

It means we wouldn't need to use the Console object like a global variable here

treysp: Ahh...so the console would still define a method for ingesting the audit results, but the method call would be passed down as a callback (rather than retrieving the console in place)?

treysp: ok, switched to using a callback and converted the other `get_console()` to a callback as well

erindru: Nice! This is a cleaner approach imo

sqlmesh/core/scheduler.py
```diff
@@ -165,8 +166,10 @@ def evaluate(
         execution_time: TimeLike,
         deployability_index: DeployabilityIndex,
         batch_index: int,
+        environment_naming_info: EnvironmentNamingInfo,
```
izeigerman, line 169: I don't like passing this conceptually, cause evaluation should not depend on environment in any way. This opens up an easy path for this invariant to break.

izeigerman: I'm ok with printing the warning upstream

sqlmesh/core/scheduler.py
```diff
@@ -466,18 +478,28 @@ def evaluate_node(node: SchedulingUnit) -> None:
             try:
                 assert execution_time  # mypy
                 assert deployability_index  # mypy
-                self.evaluate(
+                audit_results = []  # so it exists for finally if `evaluate` raises
```
izeigerman, line 481: This needs to be set before `try:` to guarantee that it's available in the `finally` block

izeigerman (approved):
Please consider my suggestions. Looks good otherwise

sqlmesh/core/scheduler.py
treysp: Thanks - I didn't fully understand/appreciate the leakage this introduced. Will keep an eye out to avoid this in the future.

sqlmesh/core/scheduler.py
treysp: Gah, my bad - even with my incorrect implementation it could obviously not exist if one of the assertions failed...

sqlmesh/core/console.py
```diff
@@ -2643,3 +2736,73 @@ def _format_audits_errors(error: NodeAuditsErrors) -> str:
...
+    if snapshot.model.kind.is_seed:
+        return "insert from seed file"
+    if snapshot.model.kind.is_full:
+        return "full refresh"
+    if snapshot.model.kind.is_view:
+        return "recreate view"
+    if snapshot.model.kind.is_incremental_by_unique_key:
+        return "insert or update rows"
+    if snapshot.model.kind.is_incremental_by_partition:
+        return "insert partition"
```
sungchun12, line 2806:
```suggestion
        return "insert partitions"
```
don't want to confuse people that we only insert a single partition

sqlmesh/core/console.py
```diff
@@ -2643,3 +2736,73 @@ def _format_audits_errors(error: NodeAuditsErrors) -> str:
...
+    if snapshot.is_model and snapshot.model.kind.is_external:
+        return "run external model audits"
+    if snapshot.model.kind.is_seed:
+        return "insert from seed file"
+    if snapshot.model.kind.is_full:
+        return "full refresh"
+    if snapshot.model.kind.is_view:
+        return "recreate view"
+    if snapshot.model.kind.is_incremental_by_unique_key:
+        return "insert or update rows"
```
sungchun12, line 2804:
```suggestion
        return "insert/update rows"
```

save space

sqlmesh/core/console.py
```diff
@@ -2643,3 +2736,73 @@ def _format_audits_errors(error: NodeAuditsErrors) -> str:
...
+        else text[: (trunc_length - 3)] + "..."
+    )
+    return f"{prefix}{truncated_text}{suffix}"
+
+
+def _create_evaluation_model_annotation(snapshot: Snapshot, interval_info: t.Optional[str]) -> str:
+    if snapshot.is_audit:
+        return "run standalone audit"
+    if snapshot.is_model and snapshot.model.kind.is_external:
+        return "run external model audits"
```
sungchun12, line 2796:
```suggestion
        return "run external audits"
```
save space

sqlmesh/core/console.py
```diff
@@ -756,7 +822,7 @@ def update_promotion_progress(self, snapshot: SnapshotInfoLike, promoted: bool)
             if self.verbosity >= Verbosity.VERBOSE:
                 action_str = "[green]promoted[/green]" if promoted else "[yellow]demoted[/yellow]"
                 self.promotion_progress.live.console.print(
-                    f"{snapshot.display_name(self.environment_naming_info, self.default_catalog, dialect=self.dialect)} {action_str}"
+                    f"{snapshot.display_name(self.environment_naming_info, self.default_catalog if self.verbosity < Verbosity.VERY_VERBOSE else None, dialect=self.dialect).ljust(self.PROGRESS_BAR_COLUMN_WIDTHS['name'])} {action_str}"
```
sungchun12, line 825: Is there a way to left align the promoted/demoted column?

image

sungchun12 (changes requested):

sungchun12 (approved):

PR #4001, opened 2025-03-15
Fix MotherDuck multi-catalog configs

naoyak:
Creating a MotherDuck connection config with the `catalogs` argument (multiple attached databases / data sources) rather than specifying `database` (single db) was failing because the MotherDuck connection was never authenticated.

The reason is because the `motherduck_token` has to be sent in the initial connection string when instantiating the connection factory (or otherwise made available via env var) - TIL that `ATTACH md:{database_name}?motherduck_token={token_string}` is actually not valid, so the MD token can't be passed in on a per-catalog basis. In any case the same token (=MD account) should be shared throughout the gateway config.


sqlmesh/core/config/connection.py
```diff
@@ -269,16 +269,18 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
                 except BinderException as e:
                     # If a user tries to create a catalog pointing at `:memory:` and with the name `memory`
                     # then we don't want to raise since this happens by default. They are just doing this to
                     # set it as the default catalog.
+                    # If a user tried to attach a MotherDuck database/share which has already by attached via
+                    # `ATTACH 'md:'`, then we don't want to raise since this is expected.
                     if not (
                         'database with name "memory" already exists' in str(e)
                         and path_options == ":memory:"
+                    ) and not (
```
naoyak, line 282: Currently, the `ATTACH` commands for the individual MotherDuck databases included in `catalogs` will raise an error since they have been automatically attached by the catch all `md:` in the connection string.

We could either:
1. Ignore the errors (this way)
2. Skip `ATTACH`ing the MD catalogs and assume they exist
3. Skip ATTACHing the catalogs but verify they are already attached and available

sqlmesh/core/config/connection.py
```diff
@@ -377,8 +376,7 @@ def to_sql(self, alias: str) -> str:
             f" AS {alias}" if not (self.type == "motherduck" or self.path.startswith("md:")) else ""
         )
         options_sql = f" ({', '.join(options)})" if options else ""
-        token_sql = "?motherduck_token=" + self.token if self.token else ""
```
naoyak, line 380: removing because it doesn't actually work

sqlmesh/core/config/connection.py
```diff
@@ -269,17 +269,17 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
                         if not path_options.startswith("md:"):
                             query += f" AS {alias}"
-                        elif self.token:
-                            query += f"?motherduck_token={self.token}"
                     cursor.execute(query)
                 except BinderException as e:
                     # If a user tries to create a catalog pointing at `:memory:` and with the name `memory`
                     # then we don't want to raise since this happens by default. They are just doing this to
                     # set it as the default catalog.
+                    # If a user tried to attach a MotherDuck database/share which has already by attached via
```
izeigerman, line 277:
```suggestion
                    # If a user tried to attach a MotherDuck database/share which has already been attached via
```

izeigerman: Why is this not an error? Why would anyone specify the same MD catalog more than once?

sqlmesh/core/config/connection.py
```diff
@@ -294,7 +294,7 @@ def create_engine_adapter(self, register_comments_override: bool = False) -> Eng
         if self.database:
             if isinstance(self, MotherDuckConnectionConfig):
                 data_files.add(
-                    f"md:{self.database}"
+                    f"md:{self.database.replace('md:', '')}"
```
izeigerman, line 297: What case do we account for here with this replacement?

sqlmesh/core/config/connection.py
```diff
@@ -349,19 +351,19 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         from sqlmesh import __version__
 
         custom_user_agent_config = {"custom_user_agent": f"SQLMesh/{__version__}"}
-        if not self.database:
-            return {"config": custom_user_agent_config}
-        connection_str = f"md:{self.database or ''}"
+        connection_str = "md:"
+        if self.database:
+            # Attach MD database in single mode to block accessing other databases
```
izeigerman, line 356: Why do we block accessing other databases here?

sqlmesh/core/config/connection.py
naoyak: This would be a friendly enhancement to allow the user to specify `database` with the `md:` prefix as well as without

sqlmesh/core/config/connection.py
naoyak: so if the motherduck connection is created with the `database` argument instead of `catalogs`, that implies we are only using a single MotherDuck catalog/db and there shouldn't be any reason to allow attaching another MD database (see [docs](https://motherduck.com/docs/key-tasks/authenticating-and-connecting-to-motherduck/authenticating-to-motherduck/#using-attach-mode)). Do you think that makes sense?

sqlmesh/core/config/connection.py
naoyak: So the way the implementation for MotherDuckConnectionConfig works right now, the db adapter uses `md:` as the connection string for duckdb, implicitly adding all the databases/shares available to the token account. if you then issue an `ATTACH` command to a specific db `md:some_db`, duckdb will emit an error.

per my comment (https://github.com/TobikoData/sqlmesh/pull/4001?notification_referrer_id=NT_kwDOADILGbMxNTMxOTIzNzM0MTozMjc5NjQx#discussion_r1996645657) above, we could either skip the ATTACH trusting that the db actually exists on the account, or we could check that it's there.

izeigerman: I see. This makes sense, thank you

sqlmesh/core/config/connection.py
izeigerman: IMHO, i'd prefer to have a consistent format what database is. `md:` is a URI and not a database name

sqlmesh/core/config/connection.py
izeigerman: I see. Is there any reason to be this limiting? Alternatively, I might suggest to ALWAYS connect with `attach_model=single` in which case users need to manually ATTACH catalogs and so we no longer need to check that error message. Or am I thinking about this wrong?

naoyak: so i guess I was understanding this functionality wrong. this doesn't *block* you from attaching other databases but rather forces you to invoke `ATTACH 'md:some_other_db'` if you end up wanting to access another catalog.

> Alternatively, I might suggest to ALWAYS connect with attach_model=single in which case users need to manually ATTACH catalogs and so we no longer need to check that error message. 

`attach_mode=single` is only available in the connection string passed when launching duckdb - when the connection is figured with `database`, the user should specify a catalog name, whereas with the `catalogs:` option we invoke duckdb with the string `'md:'` so we don't have a catalog name to pass to the `attach_mode=single` option.

We could possibly:
- arbitrarily pick the first MD catalog under `catalogs` and pass it to the connection string
- or alternatively, launch duckdb with a blank connection string (i.e. no MD connection until we `ATTACH` the listed catalogs). maybe this makes the most sense?

izeigerman: Apologies about the delay.

> or alternatively, launch duckdb with a blank connection string 

This makes sense to me

naoyak: > or alternatively, launch duckdb with a blank connection string (i.e. no MD connection until we ATTACH the listed catalogs). maybe this makes the most sense?

Hmm, after looking into this apparently this won't work either.
If you initialize the DuckDB connection without an active MotherDuck token (via envvar or connection string), then you can't connect to MD via the `ATTACH 'md:some_db?motherduck_token=some_token'` syntax, instead you have to invoke `.open 'md:some_db?motherduck_token=some_token'`. Since `DuckDBAttachOptions` uses the `ATTACH` command this doesn't work as currently implemented.

I think we should pick one of the solutions from https://github.com/TobikoData/sqlmesh/pull/4001#discussion_r1996645657 :

> 1. Ignore the errors (this way)
2. Skip ATTACHing the MD catalogs and assume they exist
3. Skip ATTACHing the catalogs but verify they are already attached and available

tobymao:
@izeigerman can you take another look?

izeigerman (approved):

naoyak:
Thanks for the feedback @izeigerman!

PR #4006, opened 2025-03-17
Fix: Enable schema binding when creating views for queries with recursive CTEs in redshift

izeigerman:
We should always include NO SCHEMA BINDING when creating views in Redshift. Otherwise view creation fails when used with something like external tables.

At the same, views with NO SCHEMA BINDING can't be created for models that rely on recursive CTEs in their queries. This update makes it so that NO SCHEMA BINDING is omitted only when recursive CTES are used.

### DIFF
--- a/sqlmesh/core/engine_adapter/mixins.py
+++ b/sqlmesh/core/engine_adapter/mixins.py
@@ -306,15 +306,7 @@ class VarcharSizeWorkaroundMixin(EngineAdapter):
 
             temp_view_name = self._get_temp_table("ctas")
 
-            is_recursive_cte = any(
-                w.args.get("recursive", False) for w in select_statement.find_all(exp.With)
-            )
-            self.create_view(
-                temp_view_name,
-                select_statement,
-                replace=False,
-                no_schema_binding=not is_recursive_cte,
-            )
+            self.create_view(temp_view_name, select_statement, replace=False)
             try:
                 columns_to_types_from_view = self._default_precision_to_max(
                     self.columns(temp_view_name)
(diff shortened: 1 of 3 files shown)


erindru (approved):

sqlmesh/core/engine_adapter/mixins.py
```diff
@@ -306,8 +306,14 @@ def _build_create_table_exp(
 
             temp_view_name = self._get_temp_table("ctas")
 
+            is_recursive_cte = any(
```
themisvaltinos, line 309: I'm thinking whether this logic should be only for `redshift` specifically? so maybe moved inside `create_view`

izeigerman: aah, excelent point. I agree

themisvaltinos: This should fix the failing `mssql` tests as well since now this adds `no schema binding` when it does not contain recursive ctes for other engines

izeigerman: @themisvaltinos fixed. let me know

themisvaltinos (approved):

sqlmesh/core/engine_adapter/redshift.py
```diff
@@ -217,11 +217,13 @@ def create_view(
...
-
-        if create_kwargs.pop("no_schema_binding", None) is False:
-            logger.warning(
-                "The 'no_schema_binding' attribute is deprecated. Views in Redshift are created as non-binding."
+        no_schema_binding = True
+        if isinstance(query_or_df, exp.Expression):
+            # We can't include NO SCHEMA BINDING if the query has a recursive CTE
+            has_recursive_cte = any(
+                w.args.get("recursive", False) for w in query_or_df.find_all(exp.With)
             )
```
georgesittas, line 225: [Nit] I think you can do `With.recursive` and that gets you a `bool` directly.

georgesittas (approved):

PR #4025, opened 2025-03-24
Fix: Always support start_* / end_* macros in pre- / post- statements

izeigerman:
Previously, whether the `start_*` / `end_*` macros were supported in pre- or post-statements depended on the model kind. For example, they were not supported if the model was of kind VIEW or FULL. This change ensures that these macros are always available in pre- and post-statements


tobymao (approved):
why do this make sense in pre and post statements?

izeigerman:
@tobymao side effects like measurement recording, reporting, etc.

PR #4027, opened 2025-03-25
Feat: Add CICD bot check for linting

VaggelisD:
This PR enhances the CI/CD bot with a new check specific to the linter. 

For the new check, we have the following cases:
- If linting is disabled, the check does not run at all
- If linting is fully successful, the check passes with `Linter Success`
- If linting shows error/warning outputs, the summary is as shown below:

- **Linter errors**
linter_errors

<br />
<br />

- **Linter warnings**
linter_warnings


kelsin (approved):

sqlmesh/core/console.py
```diff
@@ -2181,6 +2180,17 @@ def log_failed_models(self, errors: t.List[NodeExecutionFailedError]) -> None:
...
+    def show_linter_violations(
+        self, violations: t.List[RuleViolation], model: Model, is_error: bool = False
+    ) -> None:
+        severity = "errors" if is_error else "warnings"
+        violations_msg = "\n".join(f" - {violation}" for violation in violations)
+        msg = f"\nLinter {severity} for `{model._path}`:\n{violations_msg}\n"
+
+        self._print(msg)
+        if is_error:
+            self._errors.append(msg)
```
georgesittas, line 2192: Just FMI– why do we both `_print` and append to `_errors`? What does the latter achieve? Could that lead to duplicating information?

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -326,10 +327,13 @@ def __init__(
             if review.state.lower() == "approved"
         }
         logger.debug(f"Approvers: {', '.join(self._approvers)}")
-        self._context: Context = Context(
-            paths=self._paths,
-            config=self.config,
-        )
+
+        self._context: Context = Context(paths=self._paths, config=self.config, load=False)
+        self._console.clear_captured_outputs()
```
georgesittas, line 332: Also FMI– why is this required?

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -654,6 +658,40 @@ def _update_check_handler(
...
+        conclusion: t.Optional[GithubCheckConclusion] = None,
+    ) -> None:
+        if not self._context.config.linter.enabled:
+            # Linter is not enabled, skip check
+            return
+
+        def conclusion_handler(
+            conclusion: GithubCheckConclusion,
+        ) -> t.Tuple[GithubCheckConclusion, str, t.Optional[str]]:
+            test_summary = (
```
georgesittas, line 673: Is `test_summary` the right name for this? Seems like it could've been copied from a unit testing handler?

georgesittas (approved):
Changes look reasonable, although I'm not super familiar with the CICD bot's code in general. Left a few questions to understand things a bit better.

sqlmesh/integrations/github/cicd/controller.py
VaggelisD: My thinking here is that we don't have a way of isolating the linter output from the rest; So, from `Context::__init__()` up until `Context::load()` we might have aggregated prints which are not derived from the linter. 

By clearing it here, we have (at the very least) isolated only the load-time outputs, which should be more exclusive to the linter.

The previous point about `_errors` here is that it further isolates the load-time _errors_ from the rest of the output, which should be specific enough. 

However, now that I think about it, that separates the linter warnings from the errors (the linter result could be mixed, depending on the rule sets), so  we may have to revert back to using all the load-time output.

VaggelisD: Note: In my testing I haven't observed any interim prints making their way in the linter summary, but they could silently make their way in.

VaggelisD: > If we can't know that load will only result in linter-specific output, doesn't this kind of defeat the purpose of this call?

It clears any previous output but yeah, it might have mixed output if other prints are added during load. I guess one solution would be to add `linter_output` in `MarkdownConsole` and only print that, was skeptical of it but it looks like the only way to isolate them.

sqlmesh/core/console.py
georgesittas: Where is `log_error` used in this case?

VaggelisD: It's used in the parent `show_linter_violations` and in many other output helpers; In this case, however, its output is not really fit for the multiple violations so I chose to override it and simplify it.

VaggelisD:
Note: https://github.com/TobikoData/sqlmesh/pull/4032 actually solves the initial problem, this PR can be rebased on top of it & simply divert the output to the linter check/job.

tests/integrations/github/cicd/test_integration.py
```diff
@@ -52,6 +52,143 @@ def get_columns(
...
+    PR with a non-breaking change and auto-categorization will be backfilled, merged, and deployed to prod
+
+    Scenario:
+    - PR is not merged
+    - PR has been approved by a required reviewer
+    - Tests passed
+    - PR Merge Method defined
+    - Delete environment is disabled
+    - Changes made in PR with auto-categorization
+    """
```
eakmanrq, line 75: Probably want to include the fact the linter is enabled as part of this comment. Also double check this is accurate with what you ended up doing in the test.

eakmanrq (commented):
Please update docs to document the new `linter` bot output and put a screenshot of the linter output from this PR in the examples sections so users can see that this functionality is included.

tests/integrations/github/cicd/test_integration.py
VaggelisD: Ah this should have been linter specific imo, changed it in the latest commit.

PR #4029, opened 2025-03-25
feat: cache external models for faster loading

tobymao:


sqlmesh/core/loader.py
```diff
@@ -620,6 +577,59 @@ def _load_metrics(self) -> UniqueKeyDict[str, MetricMeta]:
 
         return metrics
 
+    def _load_external_models(
```
georgesittas, line 580: It looks like the dbt loader [calls this method](https://github.com/TobikoData/sqlmesh/blob/75b7270a5c949ae0c8304f05bad6bf1ec8cc5f83/sqlmesh/dbt/loader.py#L148), so it feels like moving this under `SqlMeshLoader` loader isn't right?

sqlmesh/core/loader.py
```diff
@@ -69,6 +76,10 @@ def __init__(self, context: GenericContext, path: Path) -> None:
         self.config = self.context.configs[self.config_path]
         self._variables_by_gateway: t.Dict[str, t.Dict[str, t.Any]] = {}
 
+    @abc.abstractmethod
+    def _cache(self, **kwargs: t.Any) -> CacheBase:
+        """Returns an instance of a CacheBase."""
```
izeigerman, line 81: Not very helpful description lol. `kwargs` for future extensions?

izeigerman: [Nit] Let's place this at the end of the class with other `_` methods

sqlmesh/core/loader.py
```diff
@@ -208,32 +219,41 @@ def _load_external_models(
...
+                                "default_catalog": self.context.default_catalog,
+                                **row,
+                            },
+                        )
+                        for row in YAML().load(file.read())
+                    ]
+            except Exception as ex:
+                raise ConfigError(f"Failed to load model definition at '{path}'.\n{ex}")
+
+        cache = self._cache()
```
izeigerman, line 242: I don't understand how this works. Doesn't dbt's cache require these arguments?
```
                project=project,
                macros_max_mtime=macros_max_mtime,
                yaml_max_mtimes=yaml_max_mtimes,
                ```

Comment (tobymao) [reply sqlmesh/core/loader.py:242]: you're right, i need to change the api one more time

File: sqlmesh/core/loader.py [hunk 4]
```diff
@@ -413,10 +429,10 @@ def _load_sql_models(
         jinja_macros: JinjaMacroRegistry,
         audits: UniqueKeyDict[str, ModelAudit],
         signals: UniqueKeyDict[str, signal],
+        cache: CacheBase,
```
izeigerman, line 432: Should we consider making this a non-breaking change by making this optional?

sqlmesh/core/loader.py
```diff
@@ -192,6 +199,7 @@ def _load_metrics(self) -> UniqueKeyDict[str, MetricMeta]:
     def _load_external_models(
         self,
         audits: UniqueKeyDict[str, ModelAudit],
+        cache: CacheBase,
```
izeigerman, line 202: Ditto: we could consider making this non-breaking

izeigerman (approved):

sqlmesh/core/loader.py
tobymao: discussed offline, chances are low folks override this, additionally, it will break unless they pre-emptively added kwargs to their override

georgesittas (approved):

PR #4032, opened 2025-03-25
Fix!: Move linting from load to plan creation

VaggelisD:
Currently, linting happens during load time which has proven to cause the following issues:
- It's triggered on every load (perf. overhead) and on commands such as `sqlmesh info` which don't always warrant linting
- It's blocking the implementation of a `sqlmesh lint` command, since the `Context` is already loaded (and linted) before the CLI command has been registered
- It's output is hidden from CI/CD since only the plan errors are shown, meaning that if `LinterError` is raised the CI/CD bot does not have the full output.

This PR moves the linting process to plan time which solves the aforementioned issues.


sqlmesh/core/console.py
```diff
@@ -2194,6 +2194,16 @@ def log_failed_models(self, errors: t.List[NodeExecutionFailedError]) -> None:
 
             self._print("```\n")
 
+    def show_linter_violations(
```
georgesittas, line 2197: Do we still need this?

tests/core/test_context.py
```diff
@@ -1731,8 +1733,10 @@ def assert_cached_violations_exist(cache: OptimizedQueryCache, model: Model):
                 assert_cached_violations_exist(cache, model2)
 
             ctx.upsert_model(error_model)
+            ctx.plan(environment="dev", auto_apply=True, no_prompts=True)
+
             assert (
-                """Column '"col"' could not be resolved for model '"test"'"""
+                """noselectstar: Query should not contain SELECT * on its outer most projections"""
```
georgesittas, line 1739: How come this test changed?

georgesittas (commented):
Imo it makes sense to add a `--skip-lints` flag to the plan command, similar to `--skip-tests`, because you may just wanna iterate in dev and not have the linter yell at you.

LGTM, otherwise 👍

sqlmesh/core/console.py
VaggelisD: Yeah, this will still be used to format the output.

In this PR, the output will actually appear in the "Prod Plan Preview" check of the GH actions (since it's now a part of the plan), and the follow up CICD PR will move it to it's own check, but the output will still be the same.

tests/core/test_context.py
VaggelisD: Because the former test would yield `PlanError` on plan application since it crashes on the engine due to the column resolution.

Thus, I changed it to something that simply triggers the linter without causing further issues.

VaggelisD:
> Imo it makes sense to add a --skip-lints flag to the plan command, similar to --skip-tests, because you may just wanna iterate in dev and not have the linter yell at you.

Added it in the latest commit

georgesittas (approved):

sqlmesh/cli/main.py
```diff
@@ -339,6 +339,11 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     is_flag=True,
     help="Skip tests prior to generating the plan if they are defined.",
 )
+@click.option(
+    "--skip-lints",
```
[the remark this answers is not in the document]
VaggelisD: `skip-linter` seems good to me

izeigerman (approved):

PR #4034, opened 2025-03-26
Feat: Introduce `sqlmesh lint` command

VaggelisD:


docs/reference/cli.md
```diff
@@ -522,3 +523,14 @@ Options:
...
   --help                          Show this message and exit.
 ```
+
+## lint
+```
+Usage: sqlmesh lint [OPTIONS]
+  Run linter for the target model(s).
+
+Options:
+  --model TEXT           A model to lint. Multiple models can be lint.
```
georgesittas, line 533: Let's include that if no models are specified, then every model will be linted.

sqlmesh/cli/main.py
```diff
@@ -1007,3 +1007,22 @@ def dlt_refresh(
...
+@click.pass_obj
+@error_handler
+@cli_analytics
+def lint(
+    obj: Context,
+    models: t.Iterator[str],
+) -> None:
+    """Run the linter for the target model(s)."""
+    model_list = list(obj.get_model(model) for model in models) if models else obj.models.values()
+    obj.lint_models(*model_list)
```
georgesittas, line 1028: Why are we constructing the model list here instead of `lint_models`? If you do this you'll have to repeat the logic in the magics, etc.

docs/reference/cli.md
izeigerman, line 533:
```suggestion
  --model TEXT           A model to lint. Multiple models can be linted.
```
`linted` or `checked`

izeigerman (approved):

PR #4038, opened 2025-03-26
Feat: state import/export

erindru:
This PR implements the ability to export the state database to a file and import it back.

The state export file format is a `json` file. I tried to implement a streaming interface via the `StateStream` abstraction and the use of the [json_stream](https://pypi.org/project/json-stream/) library. The goal is to be able to dump large projects without loading everything into memory and crashing with an OOM. 

In terms of version compatibility, there is a hard requirement to use the same version of SQLMesh to load the state as was used to dump it. This greatly simplifies the implementation and ensures our Pydantic model definitions will always be compatible. Guidance is included in the documentation on how to upgrade an older state file to be compatible with a new version of SQLMesh.

State export:
Screenshot From 2025-03-31 15-43-59

State import:
Screenshot From 2025-03-31 15-43-33


sqlmesh/cli/main.py
```diff
@@ -1030,3 +1031,53 @@ def lint(
 ) -> None:
     """Run the linter for the target model(s)."""
     obj.lint_models(models)
+
+
+@cli.group(no_args_is_help=True)
+def state() -> None:
```
erindru, line 1037: This is a slight departure from our existing structure but I decided to group `state` operations to keep things open for more operations in future (such as perhaps being able to query state).

So the CLI syntax is `sqlmesh state dump` or `sqlmesh state load` vs something like `sqlmesh state_dump` or `sqlmesh state_load`.

@treysp i'd be keen to know if this is a direction you were planning to head in given the recent CLI refactoring work

sqlmesh/core/context.py
```diff
@@ -2074,6 +2077,56 @@ def clear_caches(self) -> None:
...
+        )
+
+        def _confirm_fn() -> bool:
+            if confirm:
+                self.console.log_status_update("")
+                self.console.log_warning(
+                    f"This [b]destructive[/b] operation will delete all existing state against the '{self.selected_gateway}' gateway \n"
+                    f"and replace it with what's in the '{input_file.as_posix()}' file."
+                )
+                if isinstance(self.console, TerminalConsole):
```
erindru, line 2123: I had some trouble figuring out how to handle the confirmations since they need to go before the load or dump progress even starts.

I'm not super happy with this implementation, open to ideas

sqlmesh/core/state_sync/base.py
```diff
@@ -459,6 +464,16 @@ def add_interval(
         )
         self.add_snapshots_intervals([snapshot_intervals])
 
+    @abc.abstractmethod
+    def load(self, stream: StateStream, clear: bool = True) -> None:
```
erindru, line 468: I decided to make dump/load first class citizens on `StateSync` rather than trying to coordinate everything over the public interface.

The reason is to allow different `StateSync` implementations to perform local optimizations or call internal methods without exposing them publicly and also helps with being able to wrap things in transactions.

sqlmesh/core/state_sync/common.py
```diff
@@ -3,13 +3,20 @@
...
+import itertools
+import abc
 
 from sqlmesh.core.console import Console
 from sqlmesh.core.dialect import schema_
 from sqlmesh.core.environment import Environment
+from sqlmesh.core.snapshot import Snapshot
+from sqlmesh.core.snapshot import SnapshotNameVersion
+
+AutoRestatement = t.NewType("AutoRestatement", t.Tuple[SnapshotNameVersion, t.Optional[int]])
```
erindru, line 15: I felt that `AutoRestatement` needed a type because I was getting lost in typehinting `t.Tuple[SnapshotNameVersion, t.Optional[int]]` everywhere

sqlmesh/core/state_sync/dump_load.py
```diff
@@ -0,0 +1,300 @@
...
+        console.update_state_load_info(versions=stream.versions)
+
+        state_sync.load(stream)
+
+    def _dump_pydantic_model(self, model: PydanticModel) -> t.Dict[str, t.Any]:
+        dump_args: t.Dict[str, t.Any] = PYDANTIC_DEFAULT_ARGS
+        return model.model_dump(mode="json", **dump_args)
+
+
+class SQLMeshJSONStreamEncoder(JSONStreamEncoder):
```
erindru, line 207: This is how the `json_stream` library allows the standard library `json.dump` to work in a streaming fashion. It works together with the `@streamable_list` and `@streamable_dict` decorators

sqlmesh/core/context.py
```diff
@@ -2074,6 +2077,56 @@ def clear_caches(self) -> None:
         for path in self.configs:
             rmtree(path / c.CACHE)
 
+    def dump_state(self, output_file: Path, confirm: bool = True) -> None:
```
izeigerman, line 2080: What about the following options:
- Dump local state only
- Only dump a specific environment(s)

sqlmesh/core/context.py
```diff
@@ -2074,6 +2077,56 @@ def clear_caches(self) -> None:
...
+
+        self.console.log_status_update(f"[b]Gateway[/b]: [green]{self.selected_gateway}[/green]")
+        self.console.print_connection_config(
+            self._state_connection_config, title="State Connection"
+        )
+
+        # trigger a connection to the StateSync so we can fail early if there is a problem
+        self.state_sync.get_versions(validate=True)
+
+        if confirm and isinstance(self.console, TerminalConsole):
```
izeigerman, line 2096: This logic should be a part of the console implementation. I don't like breaking the abstraction here.

sqlmesh/core/context.py
```diff
@@ -2074,6 +2077,56 @@ def clear_caches(self) -> None:
...
+            if not self.console._confirm("\nContinue?"):
+                return
+
+        self.console.log_status_update("")
+
+        dump_state(self.state_sync, output_file, self.console)
+
+    def load_state(self, input_file: Path, confirm: bool = True) -> None:
```
izeigerman, line 2104: I'd prefer `export` / `import` over `dump` / `load`..

sqlmesh/core/state_sync/dump_load.py
```diff
@@ -0,0 +1,300 @@
...
+            state_sync: The StateSync that the user has requested to dump state from
+            data: A factory function that produces new streaming JSON reader attached to the file we are loading state from.
+                This is so each section of the file can have its own reader which allows it to be read in isolation / out-of-order
+                This puts less reliance on downstream consumers performing operations in a certain order
+            console: A Console instance to print progress to
+        """
+        pass
+
+
+class V1StateSerDe(StateSerDe):
```
izeigerman, line 70: Why do we need `V1`?

sqlmesh/core/state_sync/dump_load.py
```diff
@@ -0,0 +1,300 @@
...
+            yield "versions", versions
+            console.update_state_dump_versions(versions)
+
+            yield "snapshots", _dump_snapshots(state_stream.snapshots)
+            console.update_state_dump_snapshots(complete=True)
+
+            yield "environments", _dump_environments(state_stream.environments)
+            console.update_state_dump_environments(complete=True)
+
+            yield "auto_restatements", _dump_auto_restatements(state_stream.auto_restatements)
```
izeigerman, line 128: I don't know if we should literally dump our tables 1-to-1. This is way too low level. For example a complete snapshot instance is assembled using data from `_snapshots`, `_intervals` and `_auto_restatements` tables. I don't think we want to expose user to all the internals. Instead, I believe it should just be `environments`, `snapshots`, `versions`.

izeigerman: Additionally, the format should be compatible with export of the local state.

georgesittas (commented):
Did a quick first pass, agree with Iaroslav's comments. Nice 👍

docs/concepts/state.md
```diff
@@ -0,0 +1,235 @@
+# State
+
+SQLMesh stores information about your project in a state database separate from your main warehouse.
```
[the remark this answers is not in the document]
erindru: Well, I agree that technically correct is the best kind of correct :)

I guess it's about what we want to encourage. The point is that state is a different workload type to a warehouse workload so for the best experience it needs to be stored in a suitable database type.

Of course this falls down if your warehouse is an OLTP database (PostgreSQL, MySQL, MSSQL) in which case storing state in your warehouse is perfectly fine.

sqlmesh/core/context.py
erindru: I was planning to add those in follow-up PR's to prevent this one from becoming too big.

Also, what is local state? Sounds like something that isn't even present in the StateSync, is it a ContextDiff against prod?

sqlmesh/core/state_sync/dump_load.py
erindru: I guess we don't anymore. This was written before the decision to tightly couple the state export format with the version of SQLMesh used to create it

sqlmesh/core/state_sync/dump_load.py
erindru: regarding AutoRestatements, I could see that the `_auto_restatements` table is joined in when calling `get_snapshots()` but I couldn't see how it was being populated if the snapshots were written back via `push_snapshots()`.

But I guess part of the load could be to extract the auto restatement information from the Snapshot records themselves and call `update_auto_restatements()` to create the auto restatement records

sqlmesh/core/context.py
erindru: Local state is `context.snapshots` reflecting local changes that have not yet been persisted to the StateSync.

To support it, we need some extra metadata in the state file to "taint" the file as non importable if it contains local state

sqlmesh/core/context.py
erindru: I've refactored how the console interactions work and have moved it

sqlmesh/core/context.py
erindru: I've refactored how the console interactions work and have moved it

sqlmesh/core/context.py
erindru: I've switched the terminology

sqlmesh/core/state_sync/dump_load.py
erindru: This is now removed

sqlmesh/core/state_sync/dump_load.py
erindru: I've improved the import implementation to keep track of the auto restatements as the snapshots are being inserted and then insert them at the end.

This means the auto restatements table no longer needs to be written to the state file

docs/concepts/state.md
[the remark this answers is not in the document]
erindru: I've tweaked this wording

sqlmesh/core/context.py
erindru: Local state export is now available via:
```sh
$ sqlmesh state export --local -o local_state.json
```

And specific environments can now be exported like:
```sh
$ sqlmesh state export --environment foo --environment bar -o specific_environment_state.json
```

sqlmesh/cli/main.py
```diff
@@ -1030,3 +1031,84 @@ def lint(
...
+
+@state.command("import")
+@click.option(
+    "-i",
+    "--input-file",
+    help="Path to the state file",
+    required=True,
+    type=click.Path(exists=True, dir_okay=False, readable=True, path_type=Path),
+)
+@click.option(
```
erindru, line 1098: This effectively makes the default import strategy "merge" instead of "replace", so if the user wants to wipe all state they need to pass `--replace` specifically.

@izeigerman do you have a preference of what the default should be?

izeigerman: I think this is fine. I can make an argument for either option to be the default.

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -439,6 +444,100 @@ def rollback(self) -> None:
...
+            snapshot_iterator, intervals_iterator, auto_restatments_iterator = itertools.tee(
+                snapshot_chunk, 3
+            )
+            overwrite_existing_snapshots = (
+                not clear
+            )  # if clear=True, all existing snapshjots were dropped anyway
+            self.snapshot_state.push_snapshots(
+                snapshot_iterator, overwrite=overwrite_existing_snapshots
+            )
+            self.add_snapshots_intervals((s.snapshot_intervals for s in intervals_iterator))
```
izeigerman, line 522: Note that doing it like this will not restore the interval records to their original state at the time of the dump. When we return snapshots from the state we merge intervals by the `(name, version)` tuple, effectively losing information about which `identifier` each interval record was associated with.  In theory it shouldn't matter, but in practice there's a high likelihood of edge cases.

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -439,6 +444,100 @@ def rollback(self) -> None:
...
+                    if s.next_auto_restatement_ts
+                }
+            )
+
+        existing_environments = set(self.get_environments_summary().keys()) if not clear else set()
+        for environment in stream.environments:
+            if not clear and environment.name in existing_environments:
+                self.environment_state.update_environment(environment)
+            else:
+                self.promote(environment)
```
izeigerman, line 537: I don't follow this. Why not just always update?

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -439,6 +444,100 @@ def rollback(self) -> None:
...
+                    s.name_version: s.next_auto_restatement_ts
+                    for s in auto_restatments_iterator
+                    if s.next_auto_restatement_ts
+                }
+            )
+
+        existing_environments = set(self.get_environments_summary().keys()) if not clear else set()
+        for environment in stream.environments:
+            if not clear and environment.name in existing_environments:
+                self.environment_state.update_environment(environment)
```
izeigerman, line 535: This doesn't update the "environment statements" (before_all / after_all)

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "sqlglot[rs]~=26.12.0",
     "tenacity",
     "time-machine",
+    "json-stream"
```
izeigerman, line 27: This doesn't look particularly alive: https://github.com/daggaz/json-stream. Are we sure we want to depend on this?

sqlmesh/core/state_sync/export_import.py
```diff
@@ -0,0 +1,229 @@
...
+        if not metadata.get("importable", False):
+            # this can happen if the state file was created from local unversioned snapshots that were not sourced from the project state database
+            raise SQLMeshError("State file is marked as not importable. Aborting")
+
+    handles: t.List[t.TextIO] = []
+
+    def _new_handle() -> StreamingJSONObject:
+        handle = input_file.open("r", encoding="utf8")
+        handles.append(handle)
+        stream = json_stream.load(handle)
```
izeigerman, line 221: Should we use `persistent=True` and hence avoid creating a separate file handle per top-level field? Our state in raw JSON format doesn't use that much memory (correct me if I'm wrong).

pyproject.toml
erindru: Last release was Jan 12th 2025: https://pypi.org/project/json-stream/

It seemed more alive than some of the other stuff we depend on

sqlmesh/core/state_sync/export_import.py
erindru: That effectively disables streaming though? I didn't see a problem with creating multiple read only file handles, it allows the file to be read out of order and for different parts of the code to be able to focus on their bit without having to be aware of what other parts are doing.

The point of streaming the file was to avoid the "load everything into memory" problem which always rears its head when you least want it to.

I get that the current StateSync interface doesnt allow streaming loads for anything except Snapshots but I wanted to keep things flexible at this layer since I anticipate StateSync being improved in future

sqlmesh/core/state_sync/db/facade.py
erindru: What is the correct way to handle this?

The reason I took this approach is because I observed that it's [what](https://github.com/TobikoData/sqlmesh/blob/b540146ad2619fcfdb2d1d298d37761ae72811d7/sqlmesh/core/plan/evaluator.py#L137) the evaluator [seems to do](https://github.com/TobikoData/sqlmesh/blob/b540146ad2619fcfdb2d1d298d37761ae72811d7/sqlmesh/core/plan/evaluator.py#L739)

My assumption was the fact that `EngineAdapterStateSync` splits out intervals from the snapshots to store them separately and just joins them on load was just an implementation detail

sqlmesh/core/state_sync/db/facade.py
erindru: So originally I was using `promote()` to write environments back.

But then I needed to support the "merge" strategy which fails with a plan ID error if you try to overwrite an existing environment using `promote()`, so I switched to using `update_environment()` for those.

But it turns out you're quite right, `update_environment()` works in both cases so I have simplified this

sqlmesh/core/state_sync/db/facade.py
erindru: Oh, thanks for spotting. I've adjusted it to take the `EnvironmentStatements` into account too.

This prompted a change to the `environments` dict in the state file to now look like:
```
"environments": {
  "prod": {
      "environment": {
          /* Environment object */
       },
       "statements": [
          /* list of EnvironmentStatements objects for the environment */
        ]
   },
   "dev": {
     /*...*/
   }
}
```

sqlmesh/core/state_sync/db/facade.py
izeigerman: yeah, it's a bit more complex than this unfortunately. I'm sure there are corner cases, but I’m struggling to come up with one right now. Let's keep this as is for now

izeigerman (approved):

PR #4039, opened 2025-03-27
Fix(athena): Fix SCD_TYPE_2 when using table_format=iceberg

erindru:
The `SCDType2` strategy was not passing the `table_format` property down to the engine adapter.

When `table_format=iceberg` was set on the model, this caused the temporary / intermediate tables to be created as Hive, creating data type mismatches on the timestamp columns between the Iceberg source table and the Hive temporary table.

This PR adjusts the `SCDType2` evaluation strategy to pass down the table format so that the intermediate / temporary tables are created as Iceberg if it is specified.

This wasn't picked up in existing tests because the current integration tests bypass the evaluation strategy altogether and just call the `scd_*` methods directly on the engine adapter, correctly passing the table format.


.circleci/continue_config.yml
```diff
@@ -308,16 +308,16 @@ workflows:
...
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
+                #- snowflake
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
```
erindru, line 315: todo: revert prior to merge

tobymao (approved):

PR #4040, opened 2025-03-27
feat: cache templating to improve performance

tobymao:


georgesittas (approved):

sqlmesh/core/macros.py
```diff
@@ -1378,3 +1369,30 @@ def _coerce(
...
+    if not isinstance(v, str):
+        try:
+            v = exp.convert(v)
+        # we use bare Exception instead of ValueError because there's
+        # a recursive error with MagicMock.
+        except Exception:
+            pass
+
+        if isinstance(v, exp.Expression):
+            v = v.sql(dialect=dialect)
```
georgesittas, line 1392: Should we dedent this by 1 level? The original code had it at the same level as the str instance check:

```
            if not isinstance(v, str):
                try:
                    v = exp.convert(v)
                except Exception:
                    pass

            if isinstance(v, exp.Expression):
                v = v.sql(dialect=self.dialect)
            mapping[k] = v
```

tobymao: thanks

PR #4044, opened 2025-03-27
Feat(dbt): Add support for on-run-start and on-run-end hooks

themisvaltinos:
This update adds support for dbt's [`on-run-start` and `on-run-end` hooks](https://docs.getdbt.com/reference/project-configs/on-run-start-on-run-end) which are defined in `dbt_project.yml`:

```yaml
on-run-start: sql-statement | [sql-statement]  
on-run-end: sql-statement | [sql-statement]  
```  

Internally, these are mapped to SQLMesh's [`before_all` and `after_all` statements](https://sqlmesh.readthedocs.io/en/stable/guides/configuration/?h=before#before_all-and-after_all-statements) 

---

As part of this update, the [`schemas`](https://docs.getdbt.com/reference/dbt-jinja-functions/schemas) variable is also supported, enabling to grant privileges as you would in dbt with a macro:  

```jinja
{% macro grant_usage_to_schemas(schemas, user) %}
  {% for schema in schemas %}
    GRANT USAGE ON SCHEMA {{ schema }} TO {{ user }};
  {% endfor %}
{% endmacro %}
```  

This macro can then be used in `on-run-end`:  

```yaml
on-run-end:
  - "{{ grant_usage_to_schemas(schemas, 'user') }}"
```  

---

This is also supported in native SQLMesh projects for `before_all` and `after_all` statements:  

```yaml
after_all:
  - "@grant_usage_to_role(@schemas, admin)"
```  

And by defining the corresponding macro:  

```python
from sqlmesh import macro

@macro()
def grant_usage_to_role(evaluator, schemas, role):
  if evaluator._environment_naming_info:
    return [
      f"GRANT USAGE ON SCHEMA {schema} TO {role};"
      for schema in schemas
    ]
```

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -110,8 +110,11 @@ class BaseExpressionRenderer:
             if snapshots and (
                 schemas := set(
                     [
-                        s.qualified_view_name.schema_for_environment(environment_naming_info)
+                        s.qualified_view_name.schema_for_environment(
+                            environment_naming_info, dialect=self._dialect
+                        )
                         for s in snapshots.values()
+                        if s.is_model and not s.is_symbolic
                     ]
                 )
             ):
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -107,6 +107,15 @@ class BaseExpressionRenderer:
 
         if environment_naming_info := kwargs.get("environment_naming_info", None):
             kwargs["this_env"] = getattr(environment_naming_info, "name")
+            if snapshots and (
+                schemas := set(
+                    [
+                        s.qualified_view_name.schema_for_environment(environment_naming_info)
+                        for s in snapshots.values()
+                    ]
+                )
+            ):
+                kwargs["schemas"] = list(schemas)
 
         this_model = kwargs.pop("this_model", None)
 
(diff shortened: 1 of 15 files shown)


sqlmesh/core/renderer.py
```diff
@@ -411,19 +420,21 @@ def render(
 
 def render_statements(
     statements: t.List[str],
-    dialect: DialectType = None,
+    dialect: str,
```
erindru, line 423: Curious as to why this was switched from `DialectType` to `str`?

themisvaltinos: So `dialect` was always a `str` to begin with, since it comes from the `engine_adapter`'s `dialect` property. I had it as `DialectType` which is a union that contains `str` (`t.Union[str, Dialect, t.Type[Dialect], None]`) because `sqlglot.parse` expected that and MyPy raised an error. Now that I've switched to sqlmesh's `parse`, which expects a string, I can define it as `str` which is also more accurate of what is passed in.

erindru: Ok, fair enough.

I did think we were moving away from passing the dialect around as strings because it makes it harder to propagate setting overrides but like you say `sqlmesh.core.dialect.parse()` takes a string so refactoring that is outside the scope of this PR

erindru (approved):
Looks good to me!

I'm not 100% on how the `JinjaMacroRegistry` is meant to be constructed but the tests look like they're exercising the right things

themisvaltinos:
> Looks good to me!
> I'm not 100% on how the JinjaMacroRegistry is meant to be constructed but the tests look like they're exercising the right things

Thanks for drawing my attention to it! I took a closer look and extended the implementation to also account for multiple projects when constructing and adding the jinja macros to the registry

sqlmesh/dbt/loader.py
```diff
@@ -230,6 +238,60 @@ def _load_requirements(self) -> t.Tuple[t.Dict[str, str], t.Set[str]]:
...
+
+        on_run_start: t.List[str] = []
+        on_run_end: t.List[str] = []
+        jinja_root_macros: t.Dict[str, MacroInfo] = {}
+        variables: t.Dict[str, t.Any] = self._get_variables()
+        dialect = self.config.dialect
+        for project in self._load_projects():
+            context = project.context.copy()
+            if manifest := context._manifest:
+                on_run_start.extend(manifest._on_run_start or [])
```
izeigerman, line 252: If we access these outside the manifest instance, we shouldn't mark them as private. Or at the very least add a public property method to expose these.

sqlmesh/dbt/loader.py
```diff
@@ -230,6 +238,60 @@ def _load_requirements(self) -> t.Tuple[t.Dict[str, str], t.Set[str]]:
...
+                    jinja_root_macros |= root_macros
+                context.set_and_render_variables(context.variables, root_package)
+                variables |= context.variables
+
+        if statements := on_run_start + on_run_end:
+            jinja_macro_references, used_variables = extract_macro_references_and_variables(
+                *(gen(stmt) for stmt in statements)
+            )
+            jinja_macros = context.jinja_macros
+            jinja_macros.root_macros = jinja_root_macros
```
izeigerman, line 266: Why do we do this gymnastics with root macros?

PR #4047, opened 2025-03-27
Feat!: Make model tests run concurrently

VaggelisD:
Currently, model tests run sequentially leveraging the `unittest` framework. This PR adds concurrency to further speed up the testing process by splitting the work between threads, each of whom runs a single test. Then, the results are aggregated into one and shown back to the user.

For this process, some of the internals of the `TestTextRunner` are borrowed in order to generate the final report; This is because we now have N runners which would output a full report for each test.


sqlmesh/core/test/__init__.py
```diff
@@ -20,6 +26,66 @@
     from sqlmesh.core.config.loader import C
 
 
+class ModelTextTestRunner(unittest.TextTestRunner):
```
izeigerman, line 29: [Nit] We should probably move all of this out of `__init__.py` and into `runner.py`.

sqlmesh/core/test/__init__.py
```diff
@@ -14,12 +21,57 @@
...
+    stream.writeln(unittest.TextTestResult.separator2)
+    stream.writeln(
+        f'Ran {tests_run} {"tests" if tests_run > 1 else "test"} in {test_duration:.3f}s \n'
+    )
+    stream.writeln(
+        f'{"OK" if is_success else "FAILED"}{" (" + ", ".join(infos) + ")" if infos else ""}'
+    )
+
+
 def run_tests(
```
izeigerman, line 75: IMHO this should also go into `runner.py` but up to you

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,66 @@
...
+        from io import StringIO
+
+        super().__init__(
+            stream=StringIO(),
+            resultclass=ModelTextTestResult,
+            **kwargs,
+        )
+
+
+def log_test_report(results: ModelTextTestResult, test_duration: float) -> None:
```
izeigerman, line 25: Just a heads up: I meant moving stuff related to running things like `run_tests` , not logging.

sqlmesh/core/test/result.py
```diff
@@ -49,8 +49,21 @@ def addFailure(self, test: unittest.TestCase, err: ErrorType) -> None:
...
+    def addError(self, test: unittest.TestCase, err: ErrorType) -> None:
+        """Called when the test case test signals an error.
+
+        The traceback is suppressed because it is redundant and not useful.
+
+        Args:
+            test: The test case.
+            err: A tuple of the form returned by sys.exc_info(), i.e., (type, value, traceback).
+        """
+        self.err = (test, err)
```
izeigerman, line 64: Is it always going to be a single error?

izeigerman: We should consider a better name for this. Perhaps `raw_error` or `original_error` or something

sqlmesh/core/test/__init__.py
```diff
@@ -43,46 +54,107 @@ def run_tests(
...
             ModelTextTestResult,
-            unittest.TextTestRunner(
-                stream=stream,
-                verbosity=2 if verbosity >= Verbosity.VERBOSE else 1,
-                resultclass=ModelTextTestResult,
-            ).run(unittest.TestSuite(tests)),
+            ModelTextTestRunner().run(t.cast(unittest.TestCase, test)),
         )
+
+        with lock:
```
izeigerman, line 120: Do we really need this? Can't we just collect all results using futures and then combine them? This way we don't need any coordination.

izeigerman (approved):
A few comments to consider, looks good otherwise

sqlmesh/core/test/__init__.py
```diff
@@ -43,46 +54,107 @@ def run_tests(
...
+
+    lock = threading.Lock()
+
+    combined_results = ModelTextTestResult(
+        stream=unittest.runner._WritelnDecorator(stream or sys.stderr),  # type: ignore
+        verbosity=2 if verbosity >= Verbosity.VERBOSE else 1,
+        descriptions=None,
+    )
+
+    worker_payload = []
```
georgesittas, line 71: Don't feel strongly about this, but I think I'd consider extracting the logic that follows out into a helper function that constructs the worker payloads. Right now, `run_tests` spans L34-L157 and it feels a bit loaded because it does various things.

Also, a nit, but I'd rename to plural form.

sqlmesh/core/test/__init__.py
```diff
@@ -43,46 +54,107 @@ def run_tests(
...
+            ]
+
+            for future in concurrent.futures.as_completed(futures):
+                test_results.append(future.result())
     finally:
-        for testing_engine_adapter in testing_adapter_by_gateway.values():
-            testing_engine_adapter.close()
+        for _, engine_adapter in worker_payload:
+            if engine_adapter:
+                engine_adapter.close()
```
georgesittas, line 149: The same adapter object can appear more than once in this list. What happens with `close()` in that case?

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,66 @@
...
+    for test_case, err in failures:
+        stream.writeln(unittest.TextTestResult.separator1)
+        stream.writeln(f"FAIL: {test_case}")
+        stream.writeln(unittest.TextTestResult.separator2)
+        stream.writeln(err)
+
+    for error in errors:
+        stream.writeln(unittest.TextTestResult.separator1)
+        stream.writeln(f"ERROR: {error[1]}")
+        stream.writeln(unittest.TextTestResult.separator2)
```
georgesittas, line 57: Why are handling the logging manually and not deferring to `unittest`?

sqlmesh/core/test/result.py
VaggelisD: > Is it always going to be a single error?


I presume so, we now run a single test per thread and each error raises an exception so I'd assume the first one to be raised is immediately returned.

sqlmesh/core/test/__init__.py
VaggelisD: I used it because the logging logic (e.g `.` / `F` / `E` etc) is abstracted away in `addSuccess`, `addFailure` etc and should happen in real time; If we collect all results at the end, we'll get an empty CLI for the 99% of the runtime.

However, I guess we could remove it by:
- Creating a dummy `TextResult` in each worker that'll call these functions as is happening now, to simulate the unittest real time logging
- Each worker will return a tuple in the form of `(success / fail / error code, original_err)` since that's what's really used in `log_test_reports`

Wdyt?

sqlmesh/core/test/__init__.py
```diff
@@ -43,46 +54,107 @@ def run_tests(
...
+    default_test_connection = config.get_test_connection(
+        gateway_name=default_gateway,
+        default_catalog=default_catalog,
+        default_catalog_dialect=default_catalog_dialect,
+    )
+
+    lock = threading.Lock()
+
+    combined_results = ModelTextTestResult(
+        stream=unittest.runner._WritelnDecorator(stream or sys.stderr),  # type: ignore
```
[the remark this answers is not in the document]
VaggelisD: I don't think so sadly, the test runner assumes it's stream has `stream.writeln` which it does so by manually appending it with `_WritelnDecorator` internally.

For our case, we "mute" the and handle the logging on our own, but doing so we miss the `_WritelnDecorator` abstraction.

sqlmesh/core/test/__init__.py
```diff
@@ -43,46 +54,107 @@ def run_tests(
...
+        verbosity=2 if verbosity >= Verbosity.VERBOSE else 1,
+        descriptions=None,
```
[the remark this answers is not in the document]
VaggelisD: The `TextTestResult` requires it in it's constructor and uses it in the following function, which I don't think we call anywhere so left it `None` for now:

```Python3
class TextTestResult(result.TestResult):
    ...

    def getDescription(self, test):
        doc_first_line = test.shortDescription()
        if self.descriptions and doc_first_line:
            return '\n'.join((str(test), doc_first_line))
        else:
            return str(test)
```

sqlmesh/core/test/__init__.py
VaggelisD: I think it should be a no-op but looking at `close()` again I can't immediately prove it, will add a lookup

sqlmesh/core/test/runner.py
VaggelisD: Because `unittest` produces this report at the end of a runner, but in our case we create a single runner running a single test per worker; If we didn't silence the runners, we'd get a full report for each test case.

Thus, I had to replicate the important internals at a point which we could control.

sqlmesh/core/test/runner.py
VaggelisD: Actually, would you prefer moving whatever logging logic to a separate file as in `logging.py` or even back to `__init__.py`? 

It's mostly `log_test_report` that sticks out here, but maybe in the future it can be extended with more information.

izeigerman: > Actually, would you prefer moving whatever logging logic to a separate file as in logging.py or even back to __init__.py?

looks like it can be a part of `result.py` since it's directly related to it

sqlmesh/core/test/__init__.py
izeigerman: > etc and should happen in real time;

ah, that's what I was missing. Sounds good

sqlmesh/core/test/runner.py
VaggelisD: Sounds good

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,241 @@
...
+
+    # Ensure workers are not greater than the number of tests
+    num_workers = min(len(model_test_metadata) or 1, default_test_connection.concurrent_tasks)
+
+    start_time = time.perf_counter()
+    try:
+        with ThreadPoolExecutor(max_workers=num_workers) as pool:
+            futures = [
+                pool.submit(_run_single_test, metadata=metadata, engine_adapter=engine_adapter)
+                for metadata, engine_adapter in zip(model_test_metadata, engine_adapters)
```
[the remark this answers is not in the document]
georgesittas: Actually, why do we even need an `engine_adapters` list here? Can't we continue constructing the adapter by gateway mapping, or perhaps an adapter factory by gateway mapping so that you can also construct new duckdb connections?

sqlmesh/core/test/result.py
```diff
@@ -49,8 +49,21 @@ def addFailure(self, test: unittest.TestCase, err: ErrorType) -> None:
             err: A tuple of the form returned by sys.exc_info(), i.e., (type, value, traceback).
         """
         exctype, value, tb = err
+        self.original_err = (test, err)
         return super().addFailure(test, (exctype, value, None))  # type: ignore
 
+    def addError(self, test: unittest.TestCase, err: ErrorType) -> None:
+        """Called when the test case test signals an error.
+
+        The traceback is suppressed because it is redundant and not useful.
```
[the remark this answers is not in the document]
VaggelisD: Looks like the second line is not as true, we don't suppress the traceback, we cache it for the final result. Will change it

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,241 @@
...
+        stream=unittest.runner._WritelnDecorator(stream or sys.stderr),  # type: ignore
+        verbosity=2 if verbosity >= Verbosity.VERBOSE else 1,
+        descriptions=True,
+    )
+
+    engine_adapters = create_test_engine_adapters(
+        model_test_metadata,
+        config,
+        default_gateway,
+        testing_adapter_by_gateway,
```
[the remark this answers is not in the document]
VaggelisD: I guess we can move this inside `create_test_engine_adapters`, it's not really required anymore on the top level function.

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,241 @@
...
+                test_results.append(future.result())
+    finally:
+        closed_adapters: t.Set[int] = set()
+
+        for engine_adapter in engine_adapters:
+            # The engine adapters list might have duplicates, so we ensure that we close each adapter once
+            hashed_adapter = hash(engine_adapter)
+            if engine_adapter and hashed_adapter not in closed_adapters:
+                engine_adapter.close()
+                closed_adapters.add(hashed_adapter)
```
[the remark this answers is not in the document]
VaggelisD: Right, but there's a separation now between the shared adapters and the DuckDB ones which is odd to handle; My thinking is that I'd like to maintain a single point where the `close()` logic will take place which will accommodate both. 

`testing_adapter_by_gateway` only contains the shared instances for that matter, so it's not a complete solution. 

Wdyt?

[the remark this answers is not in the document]
georgesittas: If you had the `ModelTestMetadata -> EngineAdapter` mapping, perhaps you could create a set out of the adapters, something like `set(adapter_by_test.values())` and then iterate over these unique adapters?

sqlmesh/core/test/runner.py
```diff
@@ -0,0 +1,236 @@
...
+        from io import StringIO
+
+        super().__init__(
+            stream=StringIO(),
+            resultclass=ModelTextTestResult,
+            **kwargs,
+        )
+
+
+def create_test_engine_adapters_for_tests(
```
georgesittas, line 48: [Nit]
```suggestion
def create_testing_engine_adapters(
```

georgesittas (approved):
Much cleaner! 🚀 🎉

sqlmesh/core/test/runner.py
VaggelisD: Done 👍

PR #4052, opened 2025-03-28
fix!: fix macros with spaces to make them more consistent

tobymao:
this could break existing macro usage with variables that have spaces

### DIFF
--- a/tests/core/test_macros.py
+++ b/tests/core/test_macros.py
@@ -1082,3 +1082,24 @@ def test_resolve_template_table():
         evaluator.transform(parsed_sql).sql(identify=True)
         == 'SELECT * FROM "test_catalog"."sqlmesh__test"."test__test_model__2517971505$partitions"'
     )
+
+
+def test_macro_with_spaces():
+    evaluator = MacroEvaluator()
+    evaluator.evaluate(d.parse_one(""" @DEF(x, "a b") """))
+    evaluator.evaluate(d.parse_one(""" @DEF(y, 'a b') """))
+    evaluator.evaluate(d.parse_one(""" @DEF(z, a."b c") """))
+
+    for sql, expected in (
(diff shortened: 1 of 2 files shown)


izeigerman (approved):

tests/core/test_macros.py
```diff
@@ -1082,3 +1082,38 @@ def test_resolve_template_table():
...
+          @{x},
+          a_@x,
+          a.@x,
+          @y,
+          @{y},
+          a_@y,
+          a.@y,
+        FROM a.@x, a."@{x}", foo(@'@x')
+        """
+    )
```
georgesittas, line 1101: Is this dead code or are you implicitly testing that we can parse all of these? If it's the latter, I'd leave a comment to ensure this doesn't get inadvertently removed.

tests/core/test_macros.py
```diff
@@ -1082,3 +1082,38 @@ def test_resolve_template_table():
...
+    evaluator = MacroEvaluator()
+    evaluator.evaluate(d.parse_one(""" @DEF(x, "a b") """))
+    evaluator.evaluate(d.parse_one(""" @DEF(y, 'a b') """))
+    evaluator.evaluate(d.parse_one(""" @DEF(z, a."b c") """))
+
+    for sql, expected in (
```
georgesittas, line 1107: Consider testing with config variables as well besides `@DEF` ones. For example:

```yaml
# config.yaml
variables:
  v1: "a b"
  v2: '"a b"'
  v3: "'a b'"
```

georgesittas (approved):

tests/core/test_macros.py
tobymao: i'm not sure if config variables iare in the scope of these tests, that's probably something to go with your changes

PR #4061, opened 2025-03-31
Fix!: serialize blueprint variables separately to leverage AST

### LINKED ISSUE #4017: SQL Model Blueprinting bug?
Hi all,

This issue references a [discussion](https://tobiko-data.slack.com/archives/C044BRE5W4S/p1741920812676169) in the Tobiko Slack.  @izeigerman and @georgesittas were super helpful in getting me up to speed on using the new SQL model blueprinting features.  This new feature has already reduced my lines of code by an order of magnitude; so cool!

However, I did run into a couple of issues for which @georgesittas suggested I open an issue.  So, I've attached a zipped sqlmesh project folder that tries to illustrate the two problems.

First, as shown in `models/test_1.sql`, in the list of table names in the blueprints section of the model definition, the table in line 7 includes spaces in the table name, and the table in line 8 uses a reserved word in SQL (e.g., "transaction").

I've tried putting single quotes and double quotes around those table names, but to no avail.  I get errors like the following:

`'name' property cannot be a string value: raw."my table space". Please use the identifier syntax instead, e.g. foo.bar instead of 'foo.bar'. Line 2, Col: 28.`

Second, as shown in `models/test_2.sql`, in the list of table names in the blueprints section of the model definition, one of the variables includes values with a colon - e.g., `range_value := "A1:B4"` in line 5.

In this case, sqlmesh throws the following error:
`a BinderException: Binder Error: Invalid range ''J1:M3261''`
Again, I tried both single and double quotes to no avail.

Hopefully, the attached example (using just sqlmesh and DuckDB) is enough to reproduce the problems.

[sqlmesh_bug.zip](https://github.com/user-attachments/files/19376978/sqlmesh_bug.zip)

georgesittas:
Fixes #4017

Given a model like this:

```sql
MODEL (
  name m.@{foo},
  blueprints (
    (foo := "s p a c e s"),
    (foo := 'c'),
  ),
);

SELECT
  @foo AS col
```

The blueprint variables that contain quotes are incorrectly rendered today:

```python
from sqlmesh import Context
ctx = Context()
for fqn, model in ctx.models.items():
    print(f"Fully-qualified name: {fqn}, name: {model.name}, query: {model.render_query().sql()}")

# Fully-qualified name: "db"."m"."""s p a c e s""", name: m."""s p a c e s""", query: SELECT '"s p a c e s"' AS "col"
# Fully-qualified name: "db"."m"."'c'", name: m."'c'", query: SELECT '''c''' AS "col"
```

This PR addresses this bug with the help of [this commit](https://github.com/TobikoData/sqlmesh/commit/d0154902e646d3ef8bc758660018afc233108578), by changing the representation of the blueprint values: their values are now the parsed ASTs, and not the corresponding SQL strings. This behavior is more intuitive, because for example quoted identifiers will be correctly interpolated in contexts such as model names, without dropping the quotes.

Since SQLGlot expressions aren't serialized today, making this fix wasn't as simple as pushing the blueprint variables to the `SQLMESH_VARS` dict, since that only contains Python literals that can be `eval`'d on the fly without additional context, when hydrating the Python environment.

Additionally, we need to serialize blueprint values because if, e.g., a model query references them, then the raw SQL with the variable reference is serialized and we need to have the variable's payload at deserialization time to ensure the query can be properly rendered.

This led to a new dictionary in the python env. that will contain only the blueprint variables, so that we can unambiguously serialize them as strings that are produced by `sql(dialect=...)` and then deserialize them in the macro evaluator by parsing with the same dialect.

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -41,7 +41,7 @@ from sqlmesh.utils import (
 from sqlmesh.utils.date import DatetimeRanges, to_datetime, to_date
 from sqlmesh.utils.errors import MacroEvalError, SQLMeshError
 from sqlmesh.utils.jinja import JinjaMacroRegistry, has_jinja
-from sqlmesh.utils.metaprogramming import Executable, prepare_env, print_exception
+from sqlmesh.utils.metaprogramming import Executable, SqlValue, prepare_env, print_exception
 
 if t.TYPE_CHECKING:
     from sqlglot.dialects.dialect import DialectType
@@ -173,14 +173,15 @@ class MacroEvaluator:
             "MacroEvaluator": MacroEvaluator,
         }
         self.python_env = python_env or {}
-        self._jinja_env: t.Optional[Environment] = jinja_env
         self.macros = {normalize_macro_name(k): v.func for k, v in macro.get_registry().items()}
+        self.columns_to_types_called = False
+        self.default_catalog = default_catalog
+
+        self._jinja_env: t.Optional[Environment] = jinja_env
         self._schema = schema
         self._resolve_table = resolve_table
         self._resolve_tables = resolve_tables
-        self.columns_to_types_called = False
         self._snapshots = snapshots if snapshots is not None else {}
-        self.default_catalog = default_catalog
         self._path = path
         self._environment_naming_info = environment_naming_info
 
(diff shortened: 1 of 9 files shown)


sqlmesh/core/macros.py
```diff
@@ -191,7 +192,13 @@ def __init__(
             elif v.is_import and getattr(self.env.get(k), c.SQLMESH_MACRO, None):
                 self.macros[normalize_macro_name(k)] = self.env[k]
             elif v.is_value:
-                self.locals[k] = self.env[k]
+                if k == c.SQLMESH_BLUEPRINT_VARS:
```
izeigerman, line 195: Seems like a hack. Should we make a special `Executable` type instead?

tobymao: exactly what i was thinking

georgesittas: I considered this, but the issue was that `c.SQLMESH_VARS` maps to a dict of primitives in the python environment. When we hydrate it from state, we just call `eval` on that dict and everything's fine because primitives can be evaluated. If it had an `Executable` value though wouldn't it crash?

georgesittas: I've refactored this and moved blueprint vars under `c.SQLMESH_VARS` in the python env., discussing in Slack...

PR #4063, opened 2025-03-31
Chore: Fix airflow tests

erindru:
The airflow tests are currently failing on main because the ancient version of pip was having trouble resolving dependencies and hanging.

Simply upgrading pip prior to setting up the python env seems to fix the issue

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -284,10 +284,10 @@ workflows:
       - airflow_docker_tests:
           requires:
             - style_and_cicd_tests
-          #filters:
-          #  branches:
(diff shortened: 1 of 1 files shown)


.circleci/continue_config.yml
```diff
@@ -281,10 +284,10 @@ workflows:
       - airflow_docker_tests:
           requires:
             - style_and_cicd_tests
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru, line 287: TODO: uncomment prior to merge

themisvaltinos (approved):

PR #4064, opened 2025-03-31
Feat: improve CLI virtual layer progress bar

treysp:
Improves virtual layer progress bar by:
- Only displaying models that have virtual views
- Differentiating created, updated, and dropped views

Currently, nodes without virtual layer views are included in the progress bar output - they will no longer appear.

Screenshot 2025-03-31 at 5 17 11 PM

The screenshots below use the quickstart project, running an initial plan then making these modifications:  add `add_model`, modify `incremental_model`, drop `full_model`. 

Currently, the first two are "promoted" and last "demoted":

Screenshot 2025-03-31 at 5 16 25 PM

Now we differentiate new from updated views, and views are "dropped" instead of "demoted":

Screenshot 2025-03-31 at 5 17 50 PM

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -246,11 +246,7 @@ class Console(LinterConsole, StateExporterConsole, StateImporterConsole, Janitor
         """Indicates that a new snapshot promotion progress has begun."""
 
     @abc.abstractmethod
-    def update_promotion_progress(
-        self,
-        snapshot: SnapshotInfoLike,
-        promoted: bool,
-    ) -> None:
+    def update_promotion_progress(self, snapshot: SnapshotInfoLike, promoted: bool) -> None:
         """Update the snapshot promotion progress."""
 
     @abc.abstractmethod
@@ -478,11 +474,7 @@ class NoopConsole(Console):
     ) -> None:
         pass
 
-    def update_promotion_progress(
-        self,
-        snapshot: SnapshotInfoLike,
-        promoted: bool,
-    ) -> None:
+    def update_promotion_progress(self, snapshot: SnapshotInfoLike, promoted: bool) -> None:
         pass
 
     def stop_promotion_progress(self, success: bool = True) -> None:
@@ -998,11 +990,7 @@ class TerminalConsole(Console):
             self.environment_naming_info = environment_naming_info
             self.default_catalog = default_catalog
 
-    def update_promotion_progress(
-        self,
-        snapshot: SnapshotInfoLike,
-        promoted: bool,
-    ) -> None:
+    def update_promotion_progress(self, snapshot: SnapshotInfoLike, promoted: bool) -> None:
         """Update the snapshot promotion progress."""
         if self.promotion_progress is not None and self.promotion_task is not None:
             if self.verbosity >= Verbosity.VERBOSE:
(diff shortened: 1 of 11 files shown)


sungchun12 (approved):

sqlmesh/core/console.py
```diff
@@ -816,15 +827,29 @@ def start_promotion_progress(
...
+        snapshot: SnapshotInfoLike,
+        promoted: bool,
+        snapshots_with_virtual_views: t.List[SnapshotId],
+    ) -> None:
         """Update the snapshot promotion progress."""
-        if self.promotion_progress is not None and self.promotion_task is not None:
+        if (
+            self.promotion_progress is not None
+            and self.promotion_task is not None
+            and snapshot.snapshot_id in snapshots_with_virtual_views
```
izeigerman, line 840: So if the given `snapshot` is not in `snapshots_with_virtual_views` then why call `update_promotion_progress` at all? Shouldn't we do this check upstream?

sqlmesh/core/console.py
```diff
@@ -172,7 +172,11 @@ def start_promotion_progress(
         """Indicates that a new snapshot promotion progress has begun."""
 
     @abc.abstractmethod
-    def update_promotion_progress(self, snapshot: SnapshotInfoLike, promoted: bool) -> None:
+    def update_promotion_progress(
```
izeigerman, line 175: I assume these changes are not necessary any more

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -944,22 +950,31 @@ def _promote_snapshot(
             )
             adapter.execute(snapshot.model.render_on_virtual_update(**render_kwargs))
 
-        if on_complete is not None:
-            on_complete(snapshot)
+            if (
+                on_complete is not None
+                and snapshots_with_virtual_views
```
izeigerman, line 955: Can we just do `snapshots_with_virtual_views or set()` upstream and avoid this check here? Also consider using a set instead of a list

izeigerman (approved):
A few remaining comments. LGTM once addressed

sqlmesh/core/console.py
```diff
@@ -560,6 +560,7 @@ def __init__(
         self.evaluation_total_task: t.Optional[TaskID] = None
         self.evaluation_model_progress: t.Optional[Progress] = None
         self.evaluation_model_tasks: t.Dict[str, TaskID] = {}
+        self.evaluation_model_batch_sizes: t.Dict[Snapshot, int] = {}
```
izeigerman, line 563: is this used?

treysp: Yes, in the eval progress bars - forgot to add it here in the last PR

It's created in the prog bar start method and reset in the prog bar end method, and I guess only used there (which is why there wasn't an error from me forgetting it in the init)?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -944,8 +944,8 @@ def _promote_snapshot(
             )
             adapter.execute(snapshot.model.render_on_virtual_update(**render_kwargs))
 
-        if on_complete is not None:
-            on_complete(snapshot)
+            if on_complete is not None:
```
izeigerman, line 947: I think we should report completion even if we didn't do anything.

tobymao:
@treysp what's the hold up on this?

izeigerman (approved):

PR #4072, opened 2025-04-02
Fix: Make jinja_macros optional for environment statements

themisvaltinos:
Make jinja_macros optional for environment statements to ensure they don't cause extra diffs when running plan in latest versions of pydantic ( > 2.10 ).

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -437,7 +437,7 @@ def render_statements(
                     expression,
                     dialect,
                     [],
-                    jinja_macro_registry=jinja_macros or JinjaMacroRegistry(),
+                    jinja_macro_registry=jinja_macros or None,
                     python_env=python_env,
                     default_catalog=default_catalog,
                     quote_identifiers=False,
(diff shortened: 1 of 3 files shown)


sqlmesh/core/renderer.py
```diff
@@ -437,7 +437,7 @@ def render_statements(
                     expression,
                     dialect,
                     [],
-                    jinja_macro_registry=jinja_macros or JinjaMacroRegistry(),
+                    jinja_macro_registry=jinja_macros or None,
```
erindru, line 440: This... seems suspect. Why did it need to change?

themisvaltinos: it didn't need to change. the reason I also added this was that before I didn't have it as None because it would always have a jinja macro registry even for native sqlmesh projects which was not needed. since jinja macros is optional now I thought there was no need to carry an empty jinja macro registry dict for these in state so only have it for the dbt projects that actually use them

erindru: Ok, I get it. But isnt `jinja_macros or None` the same as `jinja_macros` because the parameter is already defined as `t.Optional[JinjaMacroRegistry]` ?

themisvaltinos: oh you're right. I should have had it like that from the beginning since this was already handled inside the Renderer

erindru (approved):

PR #4077, opened 2025-04-03
Fix: Account for array types when showing sample in table diff

themisvaltinos:
This update improves the filtering approach in table diff when displaying the diff samples for a column, which failed to properly handle array types before.

As part of this also decimal values are now correctly aligned by propagating the decimal precision in row diff to properly display them in the sample. Also added helper methods (`capture_console_output` and `create_test_console`) so we can verify in tests the exact output of console methods from now on.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -2044,8 +2044,6 @@ class TerminalConsole(Console):
 
 def _cells_match(x: t.Any, y: t.Any) -> bool:
     """Helper function to compare two cells and returns true if they're equal, handling array objects."""
-    if x is None or y is None:
-        return x == y
 
     # Convert array-like objects to list for consistent comparison
     def _normalize(val: t.Any) -> t.Any:
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -1905,32 +1905,10 @@ class TerminalConsole(Console):
                     # Create a table with the joined keys and comparison columns
                     column_table = row_diff.joined_sample[keys + [source_column, target_column]]
 
-                    def compare_cells(x: t.Any, y: t.Any) -> bool:
-                        """Compare two cells and returns true if they're not equal, handling array objects."""
-                        if x is None or y is None:
-                            return x != y
-
-                        # Convert any array-like object to list for consistent comparison
-                        def to_list(val: t.Any) -> t.Any:
-                            return (
-                                list(val)
-                                if isinstance(val, (pd.Series, np.ndarray, list, tuple, set))
-                                else val
-                            )
-
-                        x = to_list(x)
-                        y = to_list(y)
-                        if isinstance(x, list) and isinstance(y, list):
-                            if len(x) != len(y):
-                                return True
-                            return any(a != b for a, b in zip(x, y))
-
-                        return x != y
-
                     # Filter to retain non identical-valued rows
                     column_table = column_table[
                         column_table.apply(
-                            lambda row: compare_cells(row[source_column], row[target_column]),
+                            lambda row: _compare_df_cells(row[source_column], row[target_column]),
                             axis=1,
                         )
                     ]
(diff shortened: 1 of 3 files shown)


sqlmesh/core/console.py
```diff
@@ -1904,10 +1905,37 @@ def show_row_diff(
                     # Create a table with the joined keys and comparison columns
                     column_table = row_diff.joined_sample[keys + [source_column, target_column]]
 
-                    # Filter out identical-valued rows
+                    def compare_cells(x: t.Any, y: t.Any) -> bool:
```
izeigerman, line 1908: This is too many level of nestedness. Let's make this top level with `_`

izeigerman (approved):

sqlmesh/core/console.py
georgesittas: @themisvaltinos I'd double-check if we already have similar logic in `tests/definition.py`. My expectation is we're probably normalizing there as well because we deal with dataframes, so perhaps there's room for refactoring here?

sqlmesh/core/console.py
```diff
@@ -1904,10 +1905,37 @@ def show_row_diff(
...
+                        if x is None or y is None:
+                            return x != y
+
+                        # Convert any array-like object to list for consistent comparison
+                        def to_list(val: t.Any) -> t.Any:
+                            return (
+                                list(val)
+                                if isinstance(val, (pd.Series, np.ndarray, list, tuple, set))
+                                else val
+                            )
```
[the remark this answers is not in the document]
themisvaltinos: The logic here is simply to filter out rows for a more readable sample, as the table is already created before, so there was no similar function in `tests/definition.py` that was value specific. I checked and the dictionaries are handled with the simple comparison, so no extra handling (e.g., for DuckDB in normalize_df) is needed, but extended the test with dicts to show this

themisvaltinos:
Thanks for the review addressed comments

sqlmesh/core/console.py
```diff
@@ -2027,6 +2042,25 @@ def show_linter_violations(
...
 
 
+def _compare_df_cells(x: t.Any, y: t.Any) -> bool:
+    """Helper function to compare two cells and returns true if they're not equal, handling array objects."""
+    if x is None or y is None:
+        return x != y
+
+    # Convert any array-like object to list for consistent comparison
+    def to_list(val: t.Any) -> t.Any:
+        return list(val) if isinstance(val, (pd.Series, np.ndarray, list, tuple, set)) else val
```
georgesittas, line 2052: The `set` here looks incorrect– the ordering in sets is non-deterministic, so converting a set into a list could result in inconsistent results when displaying the table diff.

Additional feedback for this: consider factoring this tuple out into a constant, e.g. `LIST_TYPES` or something.

sqlmesh/core/console.py
```diff
@@ -2027,6 +2042,25 @@ def show_linter_violations(
             self.log_warning(msg)
 
 
+def _compare_df_cells(x: t.Any, y: t.Any) -> bool:
```
georgesittas, line 2045: [Nit] consider naming this differently to reflect its purpose better, e.g. `_cells_match` (so you'd flip the return values to reflect this).

Should be a low-lift to make it so you read L1911 and understand what's going on without having to look at this.

sqlmesh/core/console.py
```diff
@@ -2027,6 +2042,25 @@ def show_linter_violations(
...
+    # Convert any array-like object to list for consistent comparison
+    def to_list(val: t.Any) -> t.Any:
+        return list(val) if isinstance(val, (pd.Series, np.ndarray, list, tuple, set)) else val
+
+    x = to_list(x)
+    y = to_list(y)
+    if isinstance(x, list) and isinstance(y, list):
+        if len(x) != len(y):
+            return True
+        return any(a != b for a, b in zip(x, y))
```
georgesittas, line 2059: Why do we have special logic for this? List comparison should work out of the box, no need to manually do this. So you'd just do `return to_list(x) != to_list(y)`, or even better `_normalize(x) != _normalize(y)`.

themisvaltinos: Yes excellent point this was left as I originally was comparing lumpy arrays but with lists there is no need anymore

georgesittas (approved):

PR #4080, opened 2025-04-04
Fix!: make metadata_only a transitive property in Python objects

georgesittas:
Take this example project:

- test_model.sql
```
MODEL (
  name test_model,
  kind FULL,
);

@parse_one_macro();

SELECT
  1 AS c
```

- test_macro.py:
```python
from sqlglot import parse_one
from sqlmesh import macro


@macro(metadata_only=True)
def parse_one_macro(evaluator):
    if evaluator.runtime_stage != "evaluating":
        return

    evaluator.engine_adapter.execute("SELECT 1")  # Change the argument here to `parse_one("SELECT 1")`
```

If we run plan, then modify the macro as explained in the comment and then run plan again, we get prompted for a backfill, even though the macro is marked as "metadata only":

```
...
Directly Modified: default.test_model (Non-breaking)
Models needing backfill:
└── default.test_model: [full refresh]
Apply - Backfill Tables [y/n]:
```

This happens because the new version of the macro depends on the `parse_one` global, but it's not included as a "metadata only" `Executable` in the Python env, because we don't propagate this property transitively today.

This PR addresses the problem by making it transitive.

Note that there is a nuance: if a dependency is marked as "metadata only" transitively, but it's also used in a non-"metadata only" context, then the latter version of the object takes precedence because the Python env is a flat namespace and we have to prioritize the most conservative categorization in this case, otherwise we could miss a required backfill.


georgesittas:
I think we may need a migration for this one @tobymao, I was thinking of deserializing all snapshots, walking their Python envs and then storing the newly serialized version with the `is_metadata` flags set appropriately. Any thoughts?

sqlmesh/utils/metaprogramming.py
```diff
@@ -283,13 +283,25 @@ def build_env(
...
+
+        if name_missing_from_env or (
+            not is_metadata and env[name] == obj and getattr(env[name], c.SQLMESH_METADATA, None)
+        ):
+            if not name_missing_from_env:
+                # The existing object in the env is "metadata only" but we're walking it again as a
+                # non-"metadata only" dependency, so we update this flag to ensure all transitive
+                # dependencies are also not marked as "metadata only"
+                is_metadata = False
+                if hasattr(obj, c.SQLMESH_METADATA):
```
tobymao, line 302: if it hits this branch, doesn't mean it has to have this attribute?

georgesittas:
Just a heads up, I discovered another edge case...

```python
from sqlglot.expressions import SQLGLOT_META
from sqlmesh import macro


@macro(metadata_only=True)
def macro(evaluator):
    print(SQLGLOT_META)
    return None
```

This will crash because we need to add `SQLGLOT_META` in the python env and we eventually call `setattr` on it, which fails because it's a string literal (this is true for other primitives as well, e.g. ints, tuples, etc).

This points out a more general problem, though: the current approach of setting an attribute to detect "metadata only" objects is not enough, because we can't call `setattr` on certain objects as explained above.

Instead, we may need to refactor our approach altogether and store a dictionary in the python env that contains the name of every "metadata only" dependency in it.

georgesittas:
> Just a heads up, I discovered another edge case... [...]
>
> Instead, we may need to refactor our approach altogether and store a dictionary in the python env that contains the name of every "metadata only" dependency in it.

Addressed this [here](https://github.com/TobikoData/sqlmesh/pull/4080/commits/a9c0f3f72e299d03992f75429451005d3ce7cc16).

sqlmesh/utils/metaprogramming.py
```diff
@@ -267,6 +267,7 @@ def build_env(
     obj: t.Any,
     *,
     env: t.Dict[str, t.Any],
+    env_metadata: t.Set[str],
```
izeigerman, line 270: Do we need a new attribute for this? Can we make the `env`'s value a tuple maybe?

sqlmesh/utils/metaprogramming.py
```diff
@@ -395,24 +413,31 @@ def is_value(self) -> bool:
...
 
     Args:
         env: Dictionary to store the env.
+        env_metadata: Keys that correspond to "metadata only" objects in the env.
         path: The root path to seralize. Other modules will not be walked and treated as imports.
     """
     serialized = {}
 
     for k, v in env.items():
+        is_metadata = True if k in env_metadata else None
```
izeigerman, line 437: None and not False?

izeigerman (approved):

sqlmesh/utils/metaprogramming.py
georgesittas: So you don't have redundant `False` values in the serialized `Executable`, `None` is dropped when doing `json()`.

sqlmesh/utils/metaprogramming.py
georgesittas: Thanks for spotting this, refactored and it looks cleaner now :)

georgesittas:
> > Just a heads up, I discovered another edge case... [...]
> > Instead, we may need to refactor our approach altogether and store a dictionary in the python env that contains the name of every "metadata only" dependency in it.
> 
> Addressed this [here](https://github.com/TobikoData/sqlmesh/pull/4080/commits/a9c0f3f72e299d03992f75429451005d3ce7cc16). Will look into implementing a migration script next.

I'm not sure if we can (easily) migrate this after all. I started working on a script that looks like this:

```python
        ...
        parsed_snapshot = json.loads(snapshot)
        node = parsed_snapshot["node"]
        raw_python_env = node.get("python_env")

        if isinstance(raw_python_env, dict):
            type_adapter = TypeAdapter(dict[str, Executable])
            serialized_python_env = type_adapter.validate_python(raw_python_env)
            hydrated_python_env = prepare_env(serialized_python_env)
```

The idea was to walk the serialized objects in order to reconstruct the python env with the metadata flag properly propagated. The problem though is that the resulting hydrated environment has _all_ dependencies of a model, and there’s no signal in state for what objects in the python env are top-level & user-defined, so we'll also walk transitive dependencies as if they were top-level ones (e.g. macros), and possibly reset their `is_metadata` flag to `None`.

tobymao (approved):

sqlmesh/migrations/v0078_detect_diff_caused_py_metadata_flag_propagation_and_warn.py
```diff
@@ -0,0 +1,491 @@
...
+                                "which in turn relies on non-metadata objects. This means that the next "
+                                "plan command may detect unexpected changes and prompt about backfilling "
+                                "this model. If this is a concern, consider running a forward-only plan "
+                                "instead: https://sqlmesh.readthedocs.io/en/stable/concepts/plans/#forward-only-plans.\n"
+                            )
+                            break
+        except Exception:
+            pass
+
+        new_snapshots.append(
```
tobymao, line 149: do you mutate snapshots at all? why are you reinserting them?

georgesittas: good point, I had started out with the goal of mutating them but I forgot to change this 👍

sqlmesh/core/model/common.py
```diff
@@ -42,53 +48,80 @@ def make_python_env(
...
+        for macro_func_or_var in expression.find_all(d.MacroFunc, d.MacroVar, exp.Identifier):
+            if macro_func_or_var.__class__ is d.MacroFunc:
+                name = macro_func_or_var.this.name.lower()
+                if name not in macros:
+                    continue
+
+                # If this macro has been seen before as a non-metadata macro, prioritize that
+                used_macros[name] = (
+                    macros[name],
+                    (used_macros.get(name) or (None, is_metadata))[1],
```
izeigerman, line 77:
```suggestion
                    used_macros.get(name, (None, is_metadata))[1],
```

sqlmesh/core/model/common.py
```diff
@@ -42,53 +48,80 @@ def make_python_env(
...
+                            path,
+                        )
+                    used_variables.add(args[0].this.lower())
+            elif macro_func_or_var.__class__ is d.MacroVar:
+                name = macro_func_or_var.name.lower()
+                if name in macros:
+                    # If this macro has been seen before as a non-metadata macro, prioritize that
+                    used_macros[name] = (
+                        macros[name],
+                        (used_macros.get(name) or (None, is_metadata))[1],
```
izeigerman, line 95:
```suggestion
                    used_macros.get(name, (None, is_metadata))[1],
```

sqlmesh/core/model/common.py
```diff
@@ -42,53 +48,80 @@ def make_python_env(
...
             python_env[name] = used_macro
         elif not hasattr(used_macro, c.SQLMESH_BUILTIN) and name not in python_env:
+            used_macro_func = used_macro.func
+            previous_is_metadata = getattr(used_macro_func, c.SQLMESH_METADATA, None)
+
+            if is_metadata:
+                setattr(used_macro_func, c.SQLMESH_METADATA, is_metadata)
+
             build_env(used_macro.func, env=env, name=name, path=module_path)
+            setattr(used_macro_func, c.SQLMESH_METADATA, previous_is_metadata)
```
izeigerman, line 124: Why do we set it back? Is it because it will be used by other models and we want to reset the attribute value? This seems like a hack to me. Shouldn't we instead pass the explicit value into `build_env` if needed and only fallback on `getattr(used_macro_func, c.SQLMESH_METADATA, None)` if the explicit value is None?

sqlmesh/migrations/v0078_warn_if_non_migratable_python_env_and_make_audits_and_signals_macros_metadata.py
```diff
@@ -0,0 +1,139 @@
...
+    snapshots_table = "_snapshots"
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    warning = (
+        "SQLMesh detected that it may not be able to fully migrate the state database. This should not impact "
+        "the migration process, but may result in unexpected changes being reported by the next `sqlmesh plan` "
+        "command. Please run the `sqlmesh diff` (https://sqlmesh.readthedocs.io/en/stable/reference/cli/?h=cli#diff) "
+        "command after the migration has completed, before making any new changes. If any unexpected changes are reported, "
+        "consider running a forward-only plan (https://sqlmesh.readthedocs.io/en/stable/concepts/plans/#forward-only-plans) "
```
izeigerman, line 39:
```suggestion
        "consider running a forward-only plan to apply these changes and avoid unnecessary backfills: `sqlmesh plan prod --forward-only`. See https://sqlmesh.readthedocs.io/en/stable/concepts/plans/#forward-only-plans for more details. "
```

izeigerman (approved):

sqlmesh/utils/metaprogramming.py
```diff
@@ -279,17 +280,26 @@ def build_env(
         env: Dictionary to store the env.
         name: Name of the object in the env.
         path: The module path to serialize. Other modules will not be walked and treated as imports.
+        is_metadata_obj: An optional flag that determins whether the input object is metadata-only.
```
georgesittas, line 283:
```suggestion
        is_metadata_obj: An optional flag that determines whether the input object is metadata-only.
```

PR #4085, opened 2025-04-04
Fix: retain DBX SQL execution log level when updating spark engine adapter

treysp:


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -121,6 +121,7 @@ def _set_spark_engine_adapter_if_needed(self) -> None:
         self._spark_engine_adapter = SparkEngineAdapter(
             partial(connection, spark=spark, catalog=catalog),
             default_catalog=catalog,
+            execute_log_level=logger.getEffectiveLevel(),
```
izeigerman, line 124: This should be `self._execute_log_level`?

treysp: yes - updated

izeigerman (approved):

eakmanrq (approved):

PR #4087, opened 2025-04-05
Fix: Make dialect in dbt init noncompulsory again

themisvaltinos:
This fixes a regression with the init command for dbt projects which didn't work without a dialect.


georgesittas (approved):
Let's test this?

themisvaltinos:
> Let's test this?

Yes good point added test for dbt so we can catch this type of issues in the future

PR #4093, opened 2025-04-07
Feat/bq session properties

ncbkr:
We want to make session properties available in BigQuery. The change is built upon this piece of information from the [google docs](https://cloud.google.com/bigquery/docs/adding-labels#adding-label-to-session). The underlying assumption is that each model is evaluated in its own session.

The result looks as shown in the attached screenshot.

image


tobymao:
not sure i understand the purpose of this since session properties is never filled.

PR #4101, opened 2025-04-08
Feat!: Add support for multiple virtual layers controlled by gateway

themisvaltinos:
This update adds support for managing a virtual layer per gateway, so a `gateway` when specified in the model will be responsible to create the views of the virtual layer as well as the physical tables. This is controlled by setting the provisionally called `gateway_managed_virtual_layer` flag to true in the config.

```yaml
gateways:
  local:
    connection:
      type: postgres
      ...

  bigquery_gw:
    connection:
      type: bigquery
    variables:
      overriden_var: 'bigquery'

  snowflake_gw:
    connection:
      type: snowflake
      database: db1
      ...
    variables:
      overriden_var: 'snowflake'

default_gateway: postgres

gateway_managed_virtual_layer: true
```

In this case the virtual layer will be created by the specific gateway instead of using the default gateway, which means that the catalog doesn't need to be shared between the gateways. Also the gateway specific catalog will be set for each corresponding model and the gateway specific variables will take precedence over global ones for the gateway-specified in the model.

The default behaviour remains unchanged when the flag is set to false or omitted from the config.


sqlmesh/core/environment.py
```diff
@@ -32,12 +32,15 @@ class EnvironmentNamingInfo(PydanticModel):
...
             `dev`, then it will become `DEV` when targeting Snowflake.
+        gateway_managed_virtual_layer: Determines whether the virtual layer's views are created by the model-specific
+            gateways, otherwise the default gateway is used. Default: False.
     """
 
     name: str = c.PROD
     suffix_target: EnvironmentSuffixTarget = Field(default=EnvironmentSuffixTarget.SCHEMA)
     catalog_name_override: t.Optional[str] = None
     normalize_name: bool = True
+    gateway_managed_virtual_layer: bool = False
```
izeigerman, line 43: In the context of environment we already know that it concerns the virtual layer, so we can call this just `gateway_managed`

sqlmesh/core/environment.py
```diff
@@ -49,6 +52,11 @@ def _sanitize_name(cls, v: str) -> str:
     def _validate_normalize_name(cls, v: t.Any) -> bool:
         return True if v is None else bool(v)
 
+    @field_validator("gateway_managed_virtual_layer", mode="before")
+    @classmethod
+    def _validate_gateway_managed_virtual_layer(cls, v: t.Any) -> bool:
```
izeigerman, line 57: We should reuse the `_validate_normalize_name` validator and just rename it

sqlmesh/core/environment.py
izeigerman: or `model_gateway_managed` for clarity

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -953,7 +963,12 @@ def _demote_snapshot(
         environment_naming_info: EnvironmentNamingInfo,
         on_complete: t.Optional[t.Callable[[SnapshotInfoLike], None]],
     ) -> None:
-        adapter = self.adapter
+        adapter = (
+            self._get_adapter(snapshot.model_gateway)
+            if environment_naming_info.gateway_managed_virtual_layer
+            and isinstance(snapshot, Snapshot)
```
izeigerman, line 969: I don't think I like this approach. Let's not implicitly change the behavior depending on which type was passed, it never ends well. I'd rather extend `SnapshotTableInfo` and `SnapshotInfoLike` interface to support this flag.

sqlmesh/core/state_sync/common.py
```diff
@@ -21,21 +21,36 @@
...
         environment for environment in environments if environment.suffix_target.is_table
     ]
-    for expired_catalog, expired_schema in {
+
+    # Drop the schemas for the expired environments
+    # Note: We have to use the corresponding adapter if it is a gateway managed virtual layer
+    for engine_adapter, expired_catalog, expired_schema in {
         (
+            (
+                engine_adapter := (
```
izeigerman, line 41: This construct is really hard to comprehend

sqlmesh/core/context.py
```diff
@@ -2287,22 +2290,45 @@ def _context_diff(
...
-        cleanup_expired_views(self.engine_adapter, expired_environments, console=self.console)
+
+        environment_snapshot_adapters: t.Dict[str, t.Dict[str, EngineAdapter]] = {}
+        for environment in expired_environments:
+            snapshot_adapters: t.Dict[str, EngineAdapter] = {}
+            if environment.gateway_managed_virtual_layer:
+                snapshots = self.state_sync.get_snapshots(environment.snapshots)
+                for snapshot_id, snapshot in snapshots.items():
+                    if snapshot.is_model and not snapshot.is_symbolic:
+                        snapshot_adapters[snapshot_id.name] = self._get_engine_adapter(
```
izeigerman, line 2321: If we included `model_gateway` into `SnapshotTableInfo` we wouldn't have to do this.

sqlmesh/core/environment.py
izeigerman: Plus `model_gateway` too to simplify `cleanup_expired_views`

sqlmesh/core/state_sync/common.py
```diff
@@ -44,27 +59,40 @@ def cleanup_expired_views(
...
             ) from e
-    for expired_view in {
-        snapshot.qualified_view_name.for_environment(
-            environment.naming_info, dialect=adapter.dialect
+
+    # Drop the views for the expired environments
+    for engine_adapter, expired_view in {
+        (
+            (
+                engine_adapter := (
```
izeigerman, line 78: This looks just like 41-47 and is equally hard to comprehend

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -204,6 +204,7 @@ def delete_expired_snapshots(
         Returns:
             A tuple of expired snapshot IDs and cleanup targets.
         """
+
```
izeigerman, line 207: Why did this file change?

sqlmesh/migrations/v0056_restore_table_indexes.py
```diff
@@ -1,4 +1,4 @@
-"""Readds indexes and primary keys in case tables were restored from a backup."""
+"""Reads indexes and primary keys in case tables were restored from a backup."""
```
izeigerman, line 1: Why did this change?

sqlmesh/core/environment.py
themisvaltinos: yes makes sense revised it

sqlmesh/core/environment.py
themisvaltinos: refactored it into one validator

sqlmesh/core/state_sync/common.py
themisvaltinos: revised these to be readable and more easy to understand

sqlmesh/core/state_sync/db/snapshot.py
themisvaltinos: yes my bad reverted it

themisvaltinos:
Thanks for the review @izeigerman addressed comments and updated `SnapshotTableInfo` to keep the model gateway as you suggested which simplified the code significantly, when you have time, if you can take another look

sqlmesh/core/model/definition.py
```diff
@@ -1907,6 +1907,13 @@ def create_models_from_blueprints(
         else:
             gateway_name = None
 
+        if (
+            (catalogs := loader_kwargs.pop("catalogs", None))
+            and gateway_name
+            and (catalog := catalogs.get(gateway_name))
+        ):
+            loader_kwargs["default_catalog"] = catalog
```
georgesittas, line 1915: Why are we popping the catalogs here?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -226,15 +226,23 @@ def promote(
...
+                    else self.adapter.dialect,
                 )
-                for s in target_snapshots
-                if s.is_model and not s.is_symbolic
-            ]
-        )
+                tables.append(table)
+                if environment_naming_info.gateway_managed:
+                    table_schema = d.schema_(table.db, catalog=table.catalog)
+                    gateway_by_schema[table_schema] = snapshot.model_gateway or ""
```
georgesittas, line 243: Now that the catalogs don't have to be shared, can't we get a key collision in this section?

For example, let's say you have two snapshots evaluated in different engines and the corresponding table's `catalog` and `db` are the same. Is this possible now? Could it lead to problems, by e.g. overwriting a gateway?

sqlmesh/migrations/v0078_add_gateway_managed_property.py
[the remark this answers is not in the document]
themisvaltinos: yes thanks for the heads up will rename

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: Yes, this is possible assuming both the schema and database are the same but overwriting a gateway isn't an issue since this is for creating schemas for the virtual layer. So, in this hypothetical scenario let’s say Snapshot 1’s model specifies the gateway as engine_2, while Snapshot 2’s model uses the default gateway, which is engine_1. Since they share the same database, we can assume they have access to the same catalog, so we don't have to create the same schema twice. Which engine actually performs the creation doesn't matter, since they both have access to it

sqlmesh/core/model/definition.py
themisvaltinos: Good question, I originally had a comment here in case it was unclear so reinstated it. It is because after getting from the catalogs the corresponding catalog for this model in the case it uses the gateway property, to remove the dict so it doesn't raise Pydantic validation error since catalogs isn't a model property.

sqlmesh/core/environment.py
```diff
@@ -32,22 +32,27 @@ class EnvironmentNamingInfo(PydanticModel):
...
         return word_characters_only(v).lower()
 
-    @field_validator("normalize_name", mode="before")
+    @field_validator("normalize_name", "gateway_managed", mode="before")
     @classmethod
-    def _validate_normalize_name(cls, v: t.Any) -> bool:
-        return True if v is None else bool(v)
+    def _validate_boolean_field(cls, v: t.Any, info: ValidationInfo) -> bool:
+        if v is None:
+            return info.field_name == "normalize_name"
```
izeigerman, line 54: Ah, now I see why they were in different validators. Sorry, I missed this before

sqlmesh/core/context.py
```diff
@@ -2227,6 +2219,17 @@ def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
                 self._engine_adapters[gateway_name] = adapter
         return self._engine_adapters
 
+    @cached_property
+    def catalogs(self) -> t.Dict[str, str]:
```
izeigerman, line 2223: shouldn't this be `default_catalogs`? Or `default_catalog_per_gateway`?

sqlmesh/core/loader.py
```diff
@@ -468,6 +468,7 @@ def _load() -> t.List[Model]:
                         default_catalog=self.context.default_catalog,
                         infer_names=self.config.model_naming.infer_names,
                         signal_definitions=signals,
+                        catalogs=self.context.catalogs,
```
izeigerman, line 471: ditto: `default_catalogs` or `default_catalogs_per_gateway`

sqlmesh/core/model/definition.py
```diff
@@ -1907,6 +1907,14 @@ def create_models_from_blueprints(
         else:
             gateway_name = None
 
+        # We pop to avoid pydantic validation issues since catalogs is not a model property
+        if (
+            (catalogs := loader_kwargs.pop("catalogs", None))
```
izeigerman, line 1912: Instead of doing this can we just make it a first class argument?

izeigerman: Also name

sqlmesh/core/plan/evaluator.py
```diff
@@ -424,7 +424,9 @@ def _demote_snapshots(
         on_complete: t.Optional[t.Callable[[SnapshotInfoLike], None]] = None,
     ) -> None:
         self.snapshot_evaluator.demote(
-            target_snapshots, environment_naming_info, on_complete=on_complete
+            target_snapshots,
```
izeigerman, line 427: Does this file ned to be changed?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -226,15 +226,23 @@ def promote(
             deployability_index: Determines snapshots that are deployable in the context of this promotion.
             on_complete: A callback to call on each successfully promoted snapshot.
         """
-        self._create_schemas(
-            [
-                s.qualified_view_name.table_for_environment(
-                    environment_naming_info, dialect=self.adapter.dialect
+
+        gateway_by_schema: t.Dict[t.Any, str] = {}
```
izeigerman, line 230: Why key is `t.Any`?

sqlmesh/core/snapshot/evaluator.py
izeigerman: > Since they share the same database, we can assume they have access to the same catalog, 

I guess the question is what if they dont?

izeigerman, line 243: Shouldn't the fallback value be `None` instead of `""`?

sqlmesh/core/state_sync/common.py
```diff
@@ -21,21 +21,33 @@
 
 
 def cleanup_expired_views(
-    adapter: EngineAdapter, environments: t.List[Environment], console: t.Optional[Console] = None
+    adapter: EngineAdapter,
+    environments: t.List[Environment],
+    console: t.Optional[Console] = None,
+    engine_adapters: t.Optional[t.Dict[str, EngineAdapter]] = None,
```
izeigerman, line 27: Why is it optional? Also I think it should go before `console`.

sqlmesh/core/state_sync/common.py
```diff
@@ -21,21 +21,33 @@
...
+    # We have to use the corresponding adapter if the virtual layer is gateway managed
+    def get_adapter(gateway_managed: bool, gateway: t.Optional[str] = None) -> EngineAdapter:
+        if gateway_managed and gateway:
+            return (engine_adapters or {}).get(gateway, adapter)
+        return adapter
+
+    # Drop the schemas for the expired environments
+    for engine_adapter, expired_catalog, expired_schema in {
         (
+            (engine_adapter := get_adapter(environment.gateway_managed, snapshot.model_gateway)),
```
izeigerman, line 45: why do we assign twice? Shouldn't this be just
```
get_adapter(environment.gateway_managed, snapshot.model_gateway),
```?

File: sqlmesh/core/state_sync/db/facade.py [hunk 19]
```diff
@@ -199,7 +199,10 @@ def promote(
                 )
                 != table_infos[name].qualified_view_name.for_environment(environment.naming_info)
             }
-            if not existing_environment.expired:
+            if (
+                not existing_environment.expired
+                and existing_environment.gateway_managed == environment.gateway_managed
```
izeigerman, line 204: I don't think this is right place for this. We need this check here instead: https://github.com/TobikoData/sqlmesh/blob/c59288bb53e10211e927224e35c01f99b9cfca2f/sqlmesh/core/state_sync/db/facade.py#L228-L232. 

The goal is to *recreate* all the views. The current block has nothing to do with that

sqlmesh/core/snapshot/evaluator.py
izeigerman: Anyway, I believe we already have this problem when creating tables in the physical layer

sqlmesh/core/state_sync/common.py
```diff
@@ -21,21 +21,33 @@
 
 
 def cleanup_expired_views(
-    adapter: EngineAdapter, environments: t.List[Environment], console: t.Optional[Console] = None
+    adapter: EngineAdapter,
```
izeigerman, line 24: This should now be `default_adapter` maybe?

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: > I guess the question is what if they dont?

If they don't then there won't be a collision in this, since the table schema contains the db info as well, so they will be under a different key

sqlmesh/core/context.py
themisvaltinos: Yes revising this to the latter for clarity

sqlmesh/core/plan/evaluator.py
themisvaltinos: ah no it changed from other changes I reverted and then running make style so will reinstate to one line how it was

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: actually I talked to George and now I see what he meant that the unfortunate case that the catalog has identical name in two data warehouses. will revise this to cover this case

themisvaltinos: yes we want it to be `none` thank you will change it

sqlmesh/core/state_sync/common.py
themisvaltinos: because we need the adapter before using it as well to grab the corresponding adapter's dialect:
https://github.com/TobikoData/sqlmesh/blob/c59288bb53e10211e927224e35c01f99b9cfca2f/sqlmesh/core/state_sync/common.py#L47-L52

sqlmesh/core/context.py
```diff
@@ -2227,6 +2219,17 @@ def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
                 self._engine_adapters[gateway_name] = adapter
         return self._engine_adapters
 
+    @cached_property
+    def default_catalog_per_gateway(self) -> t.Dict[str, str]:
+        """Returns the catalogs for each engine adapter in a multi virtual layer setup when the catalog isn't shared."""
```
izeigerman, line 2224: > in a multi virtual layer setup

wait, why is it only virtual layer? Shouldn't we apply the model gateway's default catalog to the model specifying that gateway?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -316,24 +324,32 @@ def create(
...
-                obj
-                for objs in concurrent_apply_to_values(
-                    list(tables_by_schema),
-                    lambda s: _get_data_objects(s, gateway_by_schema[s]),
-                    self.ddl_concurrent_tasks,
-                )
-                for obj in objs
-            }
+            existing_objects: t.Set[str] = set()
+            for gateway, tables_by_schema in tables_by_gateway_and_schema.items():
```
izeigerman, line 340: Awesome. Let's leave a comment for why we're doing this gymnastics (same schema can be shared across multiple gateways)

sqlmesh/core/context.py
```diff
@@ -2227,6 +2219,17 @@ def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
...
+        if self.gateway_managed_virtual_layer:
```
izeigerman, line 2225: Shouldn't we always have default catalogs? Shouldn't this be 
```
if self._default_catalog_per_gateway is not None:
```
instead?

sqlmesh/core/context_diff.py
```diff
@@ -118,7 +121,7 @@ def create(
         env = state_reader.get_environment(environment)
 
         create_from_env_exists = False
-        if env is None or env.expired:
+        if env is None or env.expired or env.gateway_managed != gateway_managed_virtual_layer:
```
izeigerman, line 124: Shouldn't we also reflect this change in https://github.com/TobikoData/sqlmesh/blob/b31c898d6834cc8147ec8a506ec2dc563ae2a733/sqlmesh/core/context_diff.py#L273? 

Otherwise, if the flag is the only thing that changed the planner won't detect a difference.

sqlmesh/core/context.py
themisvaltinos: yes I think you're right, I thought it didn't apply to projects with one virtual layer since the catalog should be the same unless overriden by the user in a model, but since this is anyway in a multi - gateway setting even without the flag it makes sense to set it correctly

tests/core/test_integration.py
```diff
@@ -4494,6 +4496,128 @@ def test_multi(mocker):
     ]
 
 
+@use_terminal_console
+def test_multi_virtual_layer(mocker):
+    context = Context(paths=["tests/fixtures/multi_virtual_layer"])
+
+    local_db = "db.duckdb"
+    if os.path.exists(local_db):
```
izeigerman, line 4504: To avoid problems with this state as well as concurrent test runs I suggest copying the project folder into a temporary location (use the `tmp_path` fixture)

izeigerman (approved):
very nice 👍

georgesittas (approved):
Great work!

Issue #4068, opened 2025-04-01
Testing module not accessing gateway variables


asos-davidstafford:
It appears that the test cannot access the variable defined in the gateway.  I have followed the jinja documentation define here
[testing](https://sqlmesh.readthedocs.io/en/stable/concepts/tests/?h=test#parameterized-model-names), however the SCHEMA name in this case is ignored.

```
MODEL (
  name @SILVER_SCHEMA.mytable,
  kind FULL,
  cron '@daily',
....
....
)

SELECT * FROM `@EXTERNAL_CATALOG.bronze.sourcetable
```

```
test:
  description: "Test"
  model: {{ var('SILVER_SCHEMA') }}.mytable
  inputs: 
    mysourcedb.bronze.sourcetable:
``` 

When trying to run the test i get

```shell
[WARNING] Model '"sandbox_dev"."none"."mytable"' was not found at 
C:\git\POVs\sqlmesh-pov\technology\tests\test_ref_country_region_v2.yaml
``` 

If this should be supported by inputs, then this is also broken in the same way

PR #4102, opened 2025-04-09
Fix!: Propagate gateway variables to model tests

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/4068

This PR addresses the following items:

- [Fix] https://github.com/TobikoData/sqlmesh/pull/4047 did not properly handle cases where the model doesn't exist, the errors should be caught

- [Fix] The testing module only takes `config.variables` into account; This is now extended to also use the selected gateway's variables

- [Refactor] The APIs for loading & running tests are fragmented into multiple points e.g `run_model_tests` and `run_tests`. To address that, this PR attempts to unify the test loading+running logic.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2465,13 +2465,13 @@ class GenericContext(BaseContext, t.Generic[C]):
     def load_model_tests(
         self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
     ) -> t.List[ModelTestMetadata]:
-        # If a set of tests is provided, use a single loader to load them
-        # Otherwise, gather all tests from all loaders/repos
+        # If a set of specific test path(s) are provided, we can use a single loader
+        # since it's not required to walk every tests/ folder in each repo
         loaders = [self._loaders[0]] if tests else self._loaders
 
         model_tests = []
         for loader in loaders:
-            model_tests.extend(loader._load_model_tests(tests=tests, patterns=patterns))
+            model_tests.extend(loader.load_model_tests(tests=tests, patterns=patterns))
 
         return model_tests
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1785,7 +1785,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         if verbosity >= Verbosity.VERBOSE:
             pd.set_option("display.max_columns", None)
 
-        test_meta = self._load_model_tests(tests=tests, patterns=match_patterns)
+        test_meta = self.load_model_tests(tests=tests, patterns=match_patterns)
 
         return run_tests(
             model_test_metadata=test_meta,
@@ -2462,7 +2462,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 "Linter detected errors in the code. Please fix them before proceeding."
             )
 
-    def _load_model_tests(
+    def load_model_tests(
         self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
     ) -> t.List[ModelTestMetadata]:
         # If a set of tests is provided, use a single loader to load them
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -364,6 +364,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         self._excluded_requirements: t.Set[str] = set()
         self._default_catalog: t.Optional[str] = None
         self._linters: t.Dict[str, Linter] = {}
+        self._variables_by_project_gateway: t.Dict[t.Tuple[str, str], t.Dict[str, t.Any]] = {}
         self._loaded: bool = False
 
         self.path, self.config = t.cast(t.Tuple[Path, C], next(iter(self.configs.items())))
@@ -1786,9 +1787,10 @@ class GenericContext(BaseContext, t.Generic[C]):
             pd.set_option("display.max_columns", None)
 
         test_meta = load_model_tests(
-            loaders=self._loaders,
+            configs=self.configs,
             tests=tests,
             patterns=match_patterns,
+            get_variables=self._get_variables,
         )
 
         return run_tests(
@@ -2465,6 +2467,32 @@ class GenericContext(BaseContext, t.Generic[C]):
             raise LinterError(
                 "Linter detected errors in the code. Please fix them before proceeding."
             )
+
+    def _get_variables(
+        self, config: t.Optional[C] = None, gateway_name: t.Optional[str] = None
+    ) -> t.Dict[str, t.Any]:
+        config = config or self.config
+        gateway_name = gateway_name or self.selected_gateway
+
+        key = (config.project, gateway_name)
+        if key not in self._variables_by_project_gateway:
+            try:
+                gateway = config.get_gateway(gateway_name)
+            except ConfigError:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
+                    f"Gateway '{gateway_name}' not found in project '{config.project}'."
+                )
+                gateway = None
+
+            self._variables_by_project_gateway[key] = {
+                **config.variables,
+                **(gateway.variables if gateway else {}),
+                c.GATEWAY: gateway_name,
+            }
+
+        return self._variables_by_project_gateway[key]
 
 
 class Context(GenericContext[Config]):
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1785,17 +1785,10 @@ class GenericContext(BaseContext, t.Generic[C]):
         if verbosity >= Verbosity.VERBOSE:
             pd.set_option("display.max_columns", None)
 
-        # Merge the root variables with the gateway's variables
-        variables = {
-            **self.config.variables,
-            **self.config.get_gateway(self.selected_gateway).variables,
-        }
-
         test_meta = load_model_tests(
-            configs=self.configs,
+            loaders=self._loaders,
             tests=tests,
             patterns=match_patterns,
-            variables=variables,
         )
 
         return run_tests(
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2465,9 +2465,12 @@ class GenericContext(BaseContext, t.Generic[C]):
     def _load_model_tests(
         self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
     ) -> t.List[ModelTestMetadata]:
-        model_tests = []
+        # If a set of tests is provided, use a single loader to load them
+        # Otherwise, gather all tests from all loaders/repos
+        loaders = [self._loaders[0]] if tests else self._loaders
 
-        for loader in self._loaders:
+        model_tests = []
+        for loader in loaders:
             model_tests.extend(loader._load_model_tests(tests=tests, patterns=patterns))
 
         return model_tests
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1785,10 +1785,11 @@ class GenericContext(BaseContext, t.Generic[C]):
         if verbosity >= Verbosity.VERBOSE:
             pd.set_option("display.max_columns", None)
 
-        default_gateway = self.gateway or self.config.default_gateway_name
-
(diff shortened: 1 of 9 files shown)


sqlmesh/core/context.py
```diff
@@ -1781,47 +1780,30 @@ def test(
...
-                gateway=self.gateway,
-                dialect=self.default_dialect,
-                verbosity=verbosity,
-                preserve_fixtures=preserve_fixtures,
-                stream=stream,
-                default_catalog=self.default_catalog,
-                default_catalog_dialect=self.engine_adapter.DIALECT,
-            )
+        # Merge the root variables with the gateway's variables
+        variables = {**self.config.variables, **self.config.get_gateway(default_gateway).variables}
```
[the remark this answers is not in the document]
VaggelisD: Right, that complicates things a tiny bit. I ended up not going with the `_get_variables` before but I may have to, thanks

sqlmesh/core/test/runner.py
```diff
@@ -128,34 +125,40 @@ def run_tests(
...
-            models=models,
-            engine_adapter=engine_adapter,
-            dialect=dialect,
-            path=metadata.path,
-            default_catalog=default_catalog,
-            preserve_fixtures=preserve_fixtures,
-        )
+    ) -> t.Optional[ModelTextTestResult]:
+        result: t.Optional[ModelTextTestResult] = None
+        try:
```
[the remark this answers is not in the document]
VaggelisD: I remembered that the code pre-concurrency was wrapping everything with `try/except`, but I took another look at it and it actually seems to guard against this case by checking if the test was created successfully (which won't be the case if the model does not exist).

I'll go ahead and remove the try/catch and instead return `None` early if `test = ModelTest.create_test(...)`  is also `None`.

sqlmesh/core/context.py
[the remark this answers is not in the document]
VaggelisD: I've been taking a look at this and I'm not sure how easy it is to implement; That is because `gateway` can come _after_ it's used in a test.

Consider this example:

```
test_example_full_model:
  model: {{ var('foo') }}.full_model
  gateway: other_gateway
  ...
```

We'll parse this individual test using `load_model_test_file()` which uses `yaml_load()` under the hood, and that requires the `variables` to be supplied already. If we wanted to make this work, I think we'd have to:
- Search for the `gateway` key in the yaml
- Load the variables from that gateway
- Call `load_model_test_file()` again


Wdyt?

sqlmesh/core/test/discovery.py
```diff
@@ -113,3 +117,46 @@ def get_all_model_tests(
...
+    configs: dict[pathlib.Path, C],
+    tests: t.Optional[t.List[str]] = None,
+    patterns: list[str] | None = None,
+    variables: dict[str, t.Any] | None = None,
+) -> list[ModelTestMetadata]:
+    """Load model tests into a list of ModelTestMetadata which will be propagated to the test runner.
+
+    Args:
+        tests: A list of tests to load; If not specified, all tests are loaded
+        patterns: A list of patterns to match against.
```
izeigerman, line 132: Match what?

VaggelisD: Hm, I guess it'd be better if the argument was named `ignore_patterns` and the comment was  `A list of patterns used to filter tests` or something similar, since that flows into `filter_tests_by_patterns(...)`

VaggelisD: Something confusing here is that `get_all_model_tests(...)` (pre existing helper) actually has both `patterns` and `ignore_patterns` which seem to do the same filename filtering, if I'm not missing anything. 

I'll check if that's true and simplify this

sqlmesh/core/context.py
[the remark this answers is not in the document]
georgesittas: Ah, I see... shoot. This is indeed more complicated than originally anticipated. I'd rather avoid loading the document twice just to extract the gateway name...

Perhaps we could match `gateway: <name>` using a regex and then see if the extracted gateway exists in the config to extract its variables? My hunch is that this could be easier. I'd consider propagating a `_get_variables` callable to the YAML loader so the variables can be fetched before rendering the Jinja string.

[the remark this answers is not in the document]
VaggelisD: Yeah that seemed to do the job 👍

sqlmesh/core/context.py
```diff
@@ -1781,47 +1780,24 @@ def test(
...
-                preserve_fixtures=preserve_fixtures,
-                stream=stream,
-                default_catalog=self.default_catalog,
-                default_catalog_dialect=self.engine_adapter.DIALECT,
-            )
+        test_meta = load_model_tests(
+            loaders=self._loaders,
+            tests=tests,
+            patterns=match_patterns,
+        )
```
georgesittas, line 1787: Passing `loaders` here crosses a logical boundary by mixing two unrelated modules, so it seems like a smell. I don't think we should be passing `_loaders` to tests, and also we shouldn't be accessing an internal attribute of the loader (`_get_variables`) to fix the variable situation in tests.

Instead, since `_get_variables` isn't inherently tied to the loading logic, I'd consider bubbling it up as an internal method of the context, and then propagating it as a callback to the loaders and then the testing module.

sqlmesh/core/test/discovery.py
```diff
@@ -99,17 +104,45 @@ def filter_tests_by_patterns(
...
+            test_file = load_model_test_file(
+                pathlib.Path(filename), get_variables=loaders[0]._get_variables
+            )
+            if test_name:
+                test_meta.append(test_file[test_name])
+            else:
+                test_meta.extend(test_file.values())
+    else:
+        for loader in loaders:
+            test_meta.extend(
```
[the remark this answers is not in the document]
VaggelisD: Yeah, the `loader` internally has it's matching `config`, but based on https://github.com/TobikoData/sqlmesh/pull/4102#discussion_r2037181636 I think I'll revert this change to iterate again over `self.config.items()`

sqlmesh/core/context.py
VaggelisD: Moved `_get_variables` up to `Context`

sqlmesh/core/loader.py
```diff
@@ -329,26 +328,7 @@ def _track_file(self, path: Path) -> None:
...
-                gateway = None
-
-            self._variables_by_gateway[gateway_name] = {
-                **self.config.variables,
-                **(gateway.variables if gateway else {}),
-                c.GATEWAY: gateway_name,
-            }
-
-        return self._variables_by_gateway[gateway_name]
+        return self.context._get_variables(config=self.config, gateway_name=gateway_name)
```
izeigerman, line 331: I don't know if I like this approach. Shouldn't test loading go into the loader instead?

georgesittas: Yeah good point, I wonder why they were separated in the first place.

VaggelisD: Sounds good, moved the discovery/load related helpers to `loader.py`

sqlmesh/core/context.py
```diff
@@ -2485,6 +2457,19 @@ def lint_models(
...
+    def _load_model_tests(
+        self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
+    ) -> t.List[ModelTestMetadata]:
+        # If a set of tests is provided, use a single loader to load them
+        # Otherwise, gather all tests from all loaders/repos
+        loaders = [self._loaders[0]] if tests else self._loaders
+
+        model_tests = []
+        for loader in loaders:
+            model_tests.extend(loader._load_model_tests(tests=tests, patterns=patterns))
```
izeigerman, line 2469: It's ok to make `load_model_tests` a public method.

sqlmesh/utils/yaml.py
```diff
@@ -36,15 +40,25 @@ def load(
...
     if isinstance(source, Path):
         path = source
         with open(source, "r", encoding="utf-8") as file:
             source = file.read()
 
+    if get_variables:
+        # If the user has specified a quoted/escaped gateway (e.g. "gateway: 'ma\tin'"), we need to
+        # parse it as YAML to match the gateway name stored in the config
+        gateway_line = GATEWAY_PATTERN.search(source)
+        gateway = yaml.load(gateway_line.group(0))["gateway"] if gateway_line else None
```
izeigerman, line 58: Sorry, I've just noticed this: why do we pass a model specific implementation detail into the generic YAML parser?

VaggelisD: These lines solve this comment: https://github.com/TobikoData/sqlmesh/pull/4102#discussion_r2035502814

sqlmesh/core/loader.py
```diff
@@ -678,6 +687,53 @@ def _load_linting_rules(self) -> RuleSet:
...
+        self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
+    ) -> t.List[ModelTestMetadata]:
+        """Loads YAML-based model tests"""
+        test_meta: t.List[ModelTestMetadata] = []
+
+        if tests:
+            for test in tests:
+                filename, test_name = test.split("::", maxsplit=1) if "::" in test else (test, "")
+
+                test_file = self._load_model_test_file(Path(filename))
```
georgesittas, line 712: Consider renaming this to `test_meta` or something similar, `test_file` is slightly misleading.

sqlmesh/core/context.py
```diff
@@ -2485,6 +2457,19 @@ def lint_models(
                 "Linter detected errors in the code. Please fix them before proceeding."
             )
 
+    def load_model_tests(
+        self, tests: t.Optional[t.List[str]] = None, patterns: list[str] | None = None
+    ) -> t.List[ModelTestMetadata]:
+        # If a set of tests is provided, use a single loader to load them
+        # Otherwise, gather all tests from all loaders/repos
+        loaders = [self._loaders[0]] if tests else self._loaders
```
georgesittas, line 2465: We could improve the comment here, explaining why any old loader does the trick if `tests` is provided. The reason is that the test path is part of the argument, so there's no need to walk every `tests/` folder in each repo, and we just need an arbitrary loader only for the loading logic.

sqlmesh/core/context.py
```diff
@@ -2485,6 +2457,19 @@ def lint_models(
...
+
+        model_tests = []
+        for loader in loaders:
+            model_tests.extend(loader._load_model_tests(tests=tests, patterns=patterns))
```
georgesittas, line 2469: Let's not rely on an internal method here, we don't do this elsewhere in context w.r.t. loading stuff. Perhaps you can follow the `load_materializations` approach?

tests/core/test_test.py
```diff
@@ -1560,36 +1562,28 @@ def execute(context, start, end, execution_time, **kwargs):
...
     )
-    child = t.cast(SqlModel, context.upsert_model(child))
 
-    test_file = tmp_path / "tests" / "test_parameterized_model_names.yaml"
-    test_file.write_text(
-        """
+    test_text = """
 test_parameterized_model_names:
-  model: {{ var('gold') }}.sch.a
+  model: {{{{ var('gold') }}}}.sch.a {gateway}
```
georgesittas, line 1580: Why did we add another pair of braces here and elsewhere?

tests/core/test_test.py
```diff
@@ -2203,3 +2258,117 @@ def test_test_output(tmp_path: Path) -> None:
...
+def test_number_of_tests_found(tmp_path: Path) -> None:
+    init_example_project(tmp_path, dialect="duckdb")
+
+    # Example project contains 1 test and we add a new file with 2 tests
+    test_file = tmp_path / "tests" / "test_new.yaml"
+    test_file.write_text(
+        """
+test_example_full_model1:
+  model: sqlmesh_example.full_model
+  description: This is an invalid test
```
georgesittas, line 2322: Why are these new tests invalid?

VaggelisD: Oops, probably copied the format from the test above and didn't remove the description

tests/core/test_test.py
VaggelisD: Had to escape these since I used a string format variable

georgesittas (approved):

PR #4116, opened 2025-04-10
Feat: Improve the cli before all after all diff and include python env

themisvaltinos:
This update adds the Python dependencies to the `before all` / `after all` diff and aims to improve the formatting using the Syntax library for consistency with the rest of the diffs. This adds syntax highlighting for Python code in the dependencies section and accordingly for SQL in the statements.

Screenshot 2025-04-14 at 12 23 24

### DIFF
--- a/sqlmesh/core/context_diff.py
+++ b/sqlmesh/core/context_diff.py
@@ -16,6 +16,7 @@ import sys
 import typing as t
 from difflib import ndiff
 from functools import cached_property
+from rich.syntax import Syntax
 from sqlmesh.core import constants as c
 from sqlmesh.core.console import get_console
 from sqlmesh.core.macros import RuntimeStage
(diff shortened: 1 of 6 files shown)


sqlmesh/core/context_diff.py
```diff
@@ -330,22 +331,59 @@ def requirements_diff(self) -> str:
             )
         )
 
-    def environment_statements_diff(self) -> str:
+    def environment_statements_diff(self) -> t.List[Syntax]:
+        PYTHON_ENV = "python_env"
```
georgesittas, line 335: We don't need a constant for this.

sqlmesh/core/context_diff.py
```diff
@@ -330,22 +331,59 @@ def requirements_diff(self) -> str:
...
+
+        def extract_python_env(python_env: t.Dict[str, Executable]) -> t.List[str]:
+            return [
+                v.payload if v.is_import or v.is_definition else f"{k} = {v.payload}"
+                for k, v in python_env.items()
+            ]
```
georgesittas, line 341: Let's not copy-paste code like this, I'd rather extract out the logic common to `render_definition` and this method into a helper under `model/common.py` or something.

sqlmesh/core/model/common.py
```diff
@@ -352,6 +352,14 @@ def depends_on(cls: t.Type, v: t.Any, info: ValidationInfo) -> t.Optional[t.Set[
     return v
 
 
+def python_env_payloads(sorted_python_env: t.List[t.Tuple[str, Executable]]) -> t.List[str]:
+    """Returns the payloads of the sorted python env."""
```
georgesittas, line 356: If this is intended to be used with a sorted python env, then I'd rather make it explicit by sorting it here. Otherwise the naming / comment should change, to make this more generic, so it'll be the user's responsibility to sort it.

tests/core/test_plan.py
```diff
@@ -2904,20 +2906,34 @@ def test_plan_environment_statements_diff(make_snapshot):
...
+    stripped = strip_ansi_codes(output)
+    expected_output = (
+        "=== before_all ===\n"
+        "  + CREATE OR REPLACE TABLE table_1 AS SELECT 1\n"
+        "  + @test_macro()\n\n"
+        "=== after_all ===\n"
+        "  + CREATE OR REPLACE TABLE table_2 AS SELECT 2\n\n"
+        "=== dependencies ===\n"
+        "  + def test_macro(evaluator):\n"
+        "    return 'one'"
```
georgesittas, line 2936: Just a quick heads up that this format is different compared to what we show for the usual model diffing. May be ok, but figured I'd point out anyway in case others feel more strongly about this.

sqlmesh/core/context_diff.py
```diff
@@ -330,22 +331,48 @@ def requirements_diff(self) -> str:
...
-            return (
-                f"  {runtime_stage}:\n    " + "\n    ".join(ndiff(previous, current)) + "\n"
-                if previous or current
-                else ""
+            return [
+                expr
+                for statement in statements
+                for expr in (
+                    python_env_payloads(
+                        sorted(statement.python_env.items(), key=lambda x: (x[1].kind, x[0]))
```
georgesittas, line 341: This is also a good candidate for factoring out, i.e. the python env sorting.

themisvaltinos:
Thanks for the review @georgesittas  addressed comments if you want to have another look

sqlmesh/core/context_diff.py
```diff
@@ -330,22 +331,48 @@ def requirements_diff(self) -> str:
...
+                    )
+                    if attr == "python_env"
+                    else getattr(statement, attr)
+                )
+            ]
+
+        def format_diff(attribute: str) -> t.Optional[t.Tuple[str, str]]:
```
georgesittas, line 348: Should we just compute the raw text diff here and send it unformatted to the console, so that it handles all formatting logic? You could send a mapping of "attribute" to diff for that attribute so you don't lose the context of "this diff corresponds to `before_all`", etc.

Do we format anywhere else in the context diff module?

sqlmesh/core/context_diff.py
```diff
@@ -330,22 +331,59 @@ def requirements_diff(self) -> str:
...
+                expr
+                for statement in statements
+                for expr in (
+                    extract_python_env(statement.python_env)
+                    if attr == PYTHON_ENV
+                    else getattr(statement, attr)
+                )
+            ]
+
+        def format_diff(attribute: str) -> t.Optional[Syntax]:
```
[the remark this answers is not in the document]
themisvaltinos: yes my bad excellent point

sqlmesh/core/model/common.py
themisvaltinos: you're right revised it to sort it here since it is always used with a sorted python env

tests/core/test_plan.py
themisvaltinos: I wanted to better separate the sections, but you're right I removed these to be in line with the usual diffing format we have that doesn't use markers

sqlmesh/core/context_diff.py
themisvaltinos: Good point added a helper for this

sqlmesh/core/context_diff.py
themisvaltinos: I guess the name is unfortunate now since the formatting is handled by syntax in console, here it is only the diff that is calculated and simply the shared logic of adding the label that is used for MarkdownConsole as well. So I renamed it to compute instead

georgesittas (approved):

Issue #4096, opened 2025-04-07
MotherDuck: Concurrency


jameswinegar:
MotherDuck supports concurrency on transformations because they have WAL logs per table instead of at the database level. SQLMesh is currently setting concurrency to 1 for all DuckDB instances. If MotherDuck is part of the gateway connection it should allow threading. 

`WARNING:sqlmesh.core.console:DuckDB does not support concurrency - setting threads to 1.`

This is from a dbt -> sqlmesh migration we are working on, and I haven't dug into the code to validate if this is an artifact from that.

jameswinegar:
Confirmed that sqlmesh.dbt is enforcing threads = 1 through some misdirection.

Confirmed that MotherDuck connections do not actually support `concurrent_tasks`. Threads are not handled correctly, example test case I'm working through below. 

```
def test_motherduck_multithreaded_connection_factory(make_config):
    from sqlmesh.core.engine_adapter import DuckDBEngineAdapter
    from sqlmesh.utils.connection_pool import ThreadLocalConnectionPool
    from threading import Thread

    config = make_config(type="motherduck", token="1234", database="thatone")

    # defaults to 1, no issue
    assert config.concurrent_tasks == 1

    # check that the connection factory always returns the same connection in multithreaded mode
    # this sounds counter-intuitive but that's what DuckDB recommends here: https://duckdb.org/docs/guides/python/multiple_threads.html
    config = make_config(type="motherduck", token="1234", database="thatone", concurrent_tasks=8)
    adapter = config.create_engine_adapter()
    assert isinstance(adapter, DuckDBEngineAdapter)
    assert isinstance(adapter._connection_pool, ThreadLocalConnectionPool)

    threads = []
    connection_objects = []

    def _thread_connection():
        connection_objects.append(adapter.connection)

    for _ in range(8):
        threads.append(Thread(target=_thread_connection))

    for thread in threads:
        thread.start()

    for thread in threads:
        thread.join()

    assert len(connection_objects) == 8
    assert len(set(connection_objects)) == 1  # they should all be the same object

    # test that recycling the pool means we dont end up with unusable connections (eg check we havent cached a closed connection)
    assert adapter.fetchone("select 1") == (1,)
    adapter.recycle()
    assert adapter.fetchone("select 1") == (1,)
```

```
tests/core/test_connection_config.py::test_motherduck_multithreaded_connection_factory
[...]

PR #4117, opened 2025-04-10
Fix!: MotherDuck Concurrency 

jameswinegar:
Addresses #4096: Since `custom_user_agent_config` is a dict it is unhashable and `ThreadLocalConnectionPool` fails when instantiating connections. This only happens when `concurrent_tasks > 1`.

This is not a regression test for this included in the PR because `ThreadLocalConnectionPool` is forcing authentication to MotherDuck.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -9,6 +9,7 @@ import re
 import typing as t
 from enum import Enum
 from functools import partial, lru_cache
+from urllib.parse import urlencode
 
 import pydantic
 from pydantic import Field
@@ -383,16 +384,16 @@ class MotherDuckConnectionConfig(BaseDuckDBConnectionConfig):
 
         connection_str = "md:"
 
-        query_params = []
-        query_params.append(f"custom_user_agent=SQLMesh/{__version__}")
+        query_params = {}
+        query_params["custom_user_agent"] = f"SQLMesh/{__version__}"
         if self.database:
             # Attach single MD database instead of all databases on the account
             connection_str += self.database
-            query_params.append("attach_mode=single")
+            query_params["attach_mode"] = "single"
         if self.token:
-            query_params.append(f"motherduck_token={self.token}")
+            query_params["motherduck_token"] = self.token
 
-        connection_str += f"?{'&'.join(query_params)}"
+        connection_str += f"?{urlencode(query_params)}"
 
         return {"database": connection_str}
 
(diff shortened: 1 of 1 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -379,16 +379,14 @@ def _connection_kwargs_keys(self) -> t.Set[str]:
     @property
     def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         """kwargs that are for execution config only"""
-        from sqlmesh import __version__
 
-        custom_user_agent_config = {"custom_user_agent": f"SQLMesh/{__version__}"}
```
tobymao, line 384: this can't be removed, so i don't believe this is a proper fix

jameswinegar: See the change to make it a query parameter

sqlmesh/core/config/connection.py
```diff
@@ -381,14 +381,20 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         """kwargs that are for execution config only"""
         from sqlmesh import __version__
 
-        custom_user_agent_config = {"custom_user_agent": f"SQLMesh/{__version__}"}
         connection_str = "md:"
+
+        query_params = []
```
tobymao, line 386: you can use urlencode

from urllib import urlencode

urlencode(...) 

instead of manually doing string concat

tobymao (approved):

sqlmesh/core/config/connection.py
```diff
@@ -381,14 +382,20 @@ def _static_connection_kwargs(self) -> t.Dict[str, t.Any]:
         """kwargs that are for execution config only"""
         from sqlmesh import __version__
 
-        custom_user_agent_config = {"custom_user_agent": f"SQLMesh/{__version__}"}
         connection_str = "md:"
+
+        query_params = {}
+        query_params["custom_user_agent"] = f"SQLMesh/{__version__}"
```
izeigerman, line 388: Are you sure that the `custom_user_agent` can be passed like this? Judging by https://github.com/duckdb/duckdb/blob/main/tools/pythonpkg/src/pyconnection.cpp#L2102-L2105 I don't see it being extracted from the URI

izeigerman:
I think we're solving the wrong thing here. Isn't this: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/config/connection.py#L216-L250 a primary issue?

tobymao:
@izeigerman passing it in like that does work and is in their docs https://motherduck.com/docs/integrations/how-to-integrate/

sqlmesh/core/config/connection.py
tobymao: https://motherduck.com/docs/integrations/how-to-integrate/

db, err := sql.Open("duckdb", "md:my_database?custom_user_agent=INTEGRATION_NAME")

jameswinegar:
See #4124 for real solution.

PR #4122, opened 2025-04-10
Docs: Trey edits to hybrid deployment docs

treysp:


docs/cloud/features/scheduler/hybrid_executors/hybrid-executors_standard-hybrid-deployment.png
eakmanrq, line 1: When I read the chart the thing I got stuck on was "Where did the executor go on the Standard Deployment"? Would it make sense to include it above the gateway? That way both graphs have the same blocks (just arranged differently).

docs/cloud/features/scheduler/hybrid_executors_overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+![executors](./scheduler/executors.png)
+
+We strongly recommend setting up a monitoring system for the executor containers to ensure they run smoothly and to help troubleshoot issues. Monitoring should include logs and system metrics like memory and CPU usage.
+
+#### .env file
+
+As mentioned above, `TCLOUD` environment variables **must** be set on the executor's Docker container. However, other environment variables can be set later because they are only needed after the executor is connected to Tobiko Cloud.
+
+Instead of setting a variable on the container, you can use a `.env` file. This can be useful for environment variables that require frequent updating, such as API tokens.
```
eakmanrq, line 118: "short-lived API tokens". For many people their API tokens are long lived and therefore not an issue. Don't want them adding this complexity when it isn't needed.

docs/cloud/features/scheduler/hybrid_executors_overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Note**: the `TCLOUD_ENV_FILE` environment variable **must** be set on the executor's Docker container, just like the `TCLOUD_URL`, `TCLOUD_CLIENT_ID`, and `TCLOUD_CLIENT_SECRET` variables.
+
+### Network Configuration
+
+Tobiko Cloud never pushes information into your network, so it doesn't need inbound access.
+
+Instead, the executors you host make outbound requests polling Tobiko Cloud for work tasks and other information.
+
+Inside your network, executors have no direct connection to local user environments. They are only required to connect to Tobiko Cloud and your SQL engine, with all information flowing via outbound requests.
```
eakmanrq, line 132: I'm not sure what "local user environments" mean. Really executors they are hosting have access to whatever the network they are running in which is their implementation detail. So maybe remove this first sentence.

eakmanrq (commented):
One thing that would be nice to mention, that I wasn't sure where to fit it in, is that in the Helm deployment the chart will configure all the environment variables for them automatically. They just need to fill out the `values.yaml`. This is what we want users doing if they are doing a k8s deployment since it means it will be done correctly. So it would be great if we had something that pushed them over to the helm docs rather then reading this and thinking they can configure it themselves. Also if they are thinking they won't be deploying with helm that is fine they can still use it to generate the correct environment variables.

docs/cloud/features/scheduler/hybrid_executors_overview.md
treysp: Updated - also added a note that people should only use a .env if they need to

docs/cloud/features/scheduler/hybrid_executors_overview.md
treysp: Yeah, sloppy language on my part. What I'm trying to communicate is "if you want to internally partition your network that's ok because users don't need access to the executors"

But the second sentence conveys that, so will delete

PR #4123, opened 2025-04-10
feat: add execution context to signals

tobymao:


sqlmesh/core/scheduler.py
```diff
@@ -369,7 +374,20 @@ def batch_intervals(self, merged_intervals: SnapshotToIntervals) -> t.Dict[Snaps
...
             unready = set(intervals)
-            intervals = snapshot.check_ready_intervals(intervals)
+
+            from sqlmesh.core.context import ExecutionContext
+
+            adapter = self.snapshot_evaluator.get_adapter(snapshot.model_gateway)
+
+            context = ExecutionContext(
+                adapter,
+                self.snapshots_by_name,
```
izeigerman, line 384: [Nit] we should probably pass the copy here

sqlmesh/core/macros.py
```diff
@@ -1286,11 +1286,17 @@ def call_macro(
     func: t.Callable,
     dialect: DialectType,
     path: Path,
-    *args: t.Any,
-    **kwargs: t.Any,
+    args: t.Tuple[t.Any, ...],
+    kwargs: t.Dict[str, t.Any],
+    **optional_kwargs: t.Any,
```
izeigerman, line 1291: let's make this a dict also for consistency

tobymao: i don't want to do that because then i'll have to pass in optional_kwargs empty dict even if i don't need it

izeigerman (approved):

PR #4124, opened 2025-04-11
Fix: Properly support engines that can share a single connection instance across threads

izeigerman:
Previously there was a workaround to achieve this with duckdb. It resulted in duckdb connection never closing among other issues.


tobymao (approved):

jameswinegar:
@izeigerman I pulled this down and tested locally with concurrent_tasks. `sqlmesh plan` was successful. I also confirmed with MotherDuck team that the custom_user_agent is actually being injected (`duckdb/v1.2.2(osx_arm64) python/3.12 SQLMesh/0.171.3.dev3`)

izeigerman:
Thanks for confirming @jameswinegar ! There's one more improvement I'd like to make before getting this in. Specifically, move the connection sharing logic to the connection pool. Should be done soon

tobymao:
looks good

PR #4125, opened 2025-04-11
Chore: Refactor the state stream interface

erindru:
Prior to this PR, the `StateStream` interface as a whole didn't lend itself well to streaming. The properties it exposed were streams but it itself was not a stream. This made it difficult to send over a HTTP boundary and reassemble on the other side.

This PR changes the `StateStream` interface to be an iterator over its components instead. This makes it much easier to send over HTTP as the sender controls the order in which the components are emitted, not the caller.

The caller just iterates over the stream and uses `isinstance` checks to identify which component its dealing with. I chose `isinstance` checks over checking a "type" property because they play nicely with MyPy

### DIFF
--- a/sqlmesh/core/state_sync/db/facade.py
+++ b/sqlmesh/core/state_sync/db/facade.py
@@ -19,7 +19,6 @@ from __future__ import annotations
 import contextlib
 import logging
 import typing as t
-import itertools
 from pathlib import Path
 from datetime import datetime
 
@@ -450,7 +449,9 @@ class EngineAdapterStateSync(StateSync):
 
     @transactional()
     def export(self, environment_names: t.Optional[t.List[str]] = None) -> StateStream:
-        state_sync = self
+        versions = self.get_versions(
+            validate=True
+        )  # will throw if the state db hasnt been created or there is a version mismatch
 
         snapshot_ids_to_export: t.Set[SnapshotId] = set()
         selected_environments: t.List[Environment] = []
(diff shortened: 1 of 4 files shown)


sqlmesh/core/state_sync/db/facade.py
```diff
@@ -466,80 +469,88 @@ def _include_snapshot(s_id: SnapshotId) -> bool:
...
+                if _include_snapshot(s.snapshot_id)
+            }
+            for chunk in chunk_iterable(all_snapshot_ids, SnapshotState.SNAPSHOT_BATCH_SIZE):
+                yield from state_sync.get_snapshots(chunk).values()
 
-            @property
-            def environments(self) -> t.Iterable[EnvironmentWithStatements]:
-                envs = selected_environments if environment_names else state_sync.get_environments()
+        def _export_environments() -> t.Iterator[EnvironmentWithStatements]:
+            envs = selected_environments if environment_names else state_sync.get_environments()
```
izeigerman, line 483: ideally we should stream environments as well. They are not exactly light weight objects

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -466,80 +469,88 @@ def _include_snapshot(s_id: SnapshotId) -> bool:
...
-                    for e in state_sync.get_environments()
-                    for s in e.snapshots
-                    if _include_snapshot(s.snapshot_id)
-                }
-                for chunk in chunk_iterable(all_snapshot_ids, SnapshotState.SNAPSHOT_BATCH_SIZE):
-                    yield from state_sync.get_snapshots(chunk).values()
+        def _export_snapshots() -> t.Iterator[Snapshot]:
+            all_snapshot_ids = {
+                s.snapshot_id
+                for e in state_sync.get_environments()
```
izeigerman, line 475: We fetch environments twice?

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -466,80 +469,88 @@ def _include_snapshot(s_id: SnapshotId) -> bool:
...
-                {
-                    s.name_version: s.next_auto_restatement_ts
-                    for s in auto_restatments_iterator
-                    if s.next_auto_restatement_ts
-                }
-            )
+                for snapshot_chunk in chunk_iterable(
+                    state_chunk, SnapshotState.SNAPSHOT_BATCH_SIZE
+                ):
+                    snapshot_iterator, intervals_iterator, auto_restatments_iterator = (
```
izeigerman, line 524: It took me a loooong minute to realize that these 3 iterators are all in fact snapshots. Why can't we just materialize it once instead?

sqlmesh/core/state_sync/db/facade.py
erindru: ...how did I miss this :facepalm: 

I've refactored it, thanks for picking it up

sqlmesh/core/state_sync/db/facade.py
erindru: We do stream them in the sense theyre yielded one by one from the generator. Also, if the user specifies a list of environment names, we fetch them one by one using `get_environment()` rather than fetching them all at once using `get_environments()`.

The thing is, in order to know which snapshots to export, we need to load all the specified environments because the mapping of "which snapshot belongs to which environment" only exists in the json blob right?

sqlmesh/core/state_sync/db/facade.py
erindru: It was done like this because `snapshot_state.push_snapshots` consumes the iterator which means there are no items available for `add_snapshots_intervals` or `auto_restatements`.

However, I guess under the hood `itertools.tee` would materialize the whole list anyway because the iterators are being consumed sequentially. So i've simplified this by materializing it once as you suggested

izeigerman (approved):

PR #4126, opened 2025-04-11
chore: refactor out targets as early as possible

benfdking:
- slight improvement that is then a launch pad to 'multi proces'
  formatting of docs to make it faster


benfdking:
@tobymao what's best practice here to parrallelise?

georgesittas (approved):

PR #4127, opened 2025-04-11
feat: introduce lsp

benfdking:
- supports formatting whole documents


sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
```
georgesittas, line 1: Do we need this?

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,98 @@
...
+from lsprotocol import types
+from pygls.server import LanguageServer
+from pygls.workspace import TextDocument
+from sqlmesh._version import __version__
+from sqlmesh.core.model import Model
+
+logger = logging.getLogger(__name__)
+
+CONTEXTS: t.Dict[str, Context] = {}
+"""A mapping of workspace paths to SQLMesh contexts."""
```
georgesittas, line 19: Why do we need one context per path? Was this done to support [multiple repos](https://sqlmesh.readthedocs.io/en/stable/guides/multi_repo), or something else? Keep in mind that even for multi-repo projects, there is a single context.

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,97 @@
...
+PATHS_TO_MODELS: t.Dict[str, t.Tuple[Context, Model]] = {}
+"""A mapping of file paths to SQLMesh (context, model) tuples."""
+
+server = LanguageServer("sqlmesh_lsp", __version__)
+
+_CACHE: t.Set[str] = set()
+"""A cache of URIs for which we have already ensured a context exists."""
+
+
+def ensure_context_for_document(document: TextDocument) -> TextDocument:
```
georgesittas, line 30: I suggest adding some line breaks here and in `formatting`, the continuous block of text is hard to read.

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,97 @@
...
+            if config_path.exists():
+                with suppress(Exception):
+                    handle = Context(paths=[f"{path}"])
+                    CONTEXTS[str(path)] = handle
+                    PATHS_TO_MODELS.update(
+                        {
+                            str(model._path.resolve()): (handle, model)
+                            for model in handle.models.values()
+                        }
+                    )
```
georgesittas, line 57: Just FMI– is this pattern of defining & mutating globals common when working with LSPs?

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,97 @@
...
+                    break
+        path = path.parent
+    return document
+
+
+@server.feature(types.TEXT_DOCUMENT_FORMATTING)
+def formatting(
+    ls: LanguageServer, params: types.DocumentFormattingParams
+) -> t.List[types.TextEdit]:
+    """Format the document based using SQLMesh format_model_expressions."""
```
georgesittas, line 69:
```suggestion
    """Format the document using SQLMesh format_model_expressions."""
```

sqlmesh/lsp/main.py
benfdking: This is done to support multiple repos primarily for the purpose of if you have multiple unconnected repos.

sqlmesh/lsp/main.py
benfdking: It's not uncomon. For me initially it's just simple and as we add complexity we can improve.

sqlmesh/lsp/main.py
benfdking: We don't ... but it allows for using running the script like 

```shell
./sqlmesh/lsp/main.py
```

sqlmesh/lsp/main.py
izeigerman: @benfdking I'm not at all sure that we support multiple `Context` instances per process today. I'm pretty sure there are quite a few singleton objects out there, including (but not limited to) macro registry.

I think we would want to launch a separate LSP server per project instead.

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@server.feature(types.TEXT_DOCUMENT_FORMATTING)
+def formatting(
+    ls: LanguageServer, params: types.DocumentFormattingParams
+) -> t.List[types.TextEdit]:
+    """Format the document using SQLMesh format_model_expressions."""
+    try:
+        document = ensure_context_for_document(ls.workspace.get_document(params.text_document.uri))
+        context, _ = PATHS_TO_MODELS.get(document.path, (None, None))
```
izeigerman, line 72: (None, None) doesn't see to match the type signature of `PATHS_TO_MODELS`. Are we sure we're running mypy for this?

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,113 @@
...
+] = None
+
+
+server = LanguageServer("sqlmesh_lsp", __version__)
+
+
+def ensure_context_for_document(document: TextDocument) -> TextDocument:
+    """Ensure that a context exists for the given document if applicable by searching for a config.py or config.yml file in the parent directories."""
+    # If the context is already loaded, return the document, if it is part of the same context
+    global GLOBAL_CONTEXT
```
tobymao, line 33: why are we using global variables here with lazy initialization?

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,113 @@
...
+from lsprotocol import types
+from pygls.server import LanguageServer
+from pygls.workspace import TextDocument
+from sqlmesh._version import __version__
+from sqlmesh.core.dialect import Audit
+from sqlmesh.core.model import Model
+
+logger = logging.getLogger(__name__)
+
+GLOBAL_CONTEXT: t.Optional[
```
tobymao, line 19: why is this a tuple, would it be easier for this to be two separate variables?

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,113 @@
...
+    path = Path(document.path).resolve()
+    if path.suffix not in (".sql", ".py"):
+        return document
+    loaded = False
+    while path.parents and not loaded:
+        for ext in ("py", "yml", "yaml"):
+            config_path = path / f"config.{ext}"
+            if config_path.exists():
+                with suppress(Exception):
+                    handle = Context(paths=[path])
```
tobymao, line 71: this doesn't seem right, since it won't handle the multi repo use case

izeigerman: i don't know if we should be worrying about this in the 1st iteration. 

@benfdking correct me if I'm wrong , but right now there's no simple way of know whether 2 repos are part of the same project or are completely independent.

tobymao: i'd like to understand what this code is doing, if we can easily get multi repo to work, i don't see why we shouldn't

themisvaltinos: on first read I think this is going from a subdirectory backwards till it finds a config. instead from the root `document.path` folder then if this check fails `if config_path.exists()` we should look for the configs in the subfolders (for the multi repo case( and then use all of these config paths to load one `Context(paths=[path_1, path_2, ..])`

sqlmesh/lsp/main.py
themisvaltinos: yes an issue with this down the line is lineage, since you might have multiple repos but a model in `repo_1` reference a model in `repo_2`

sqlmesh/lsp/main.py
izeigerman: @themisvaltinos I think the question is: how do we even know if the 2 configs refer to the same SQLMesh project or not? Or because it's VSCode we can assume that all repos added to the workspace belong to the same SQLMesh project?

sqlmesh/lsp/main.py
izeigerman: I agree with this. Why can't we have a proper class with a state instead?

sqlmesh/lsp/main.py
benfdking: This should be now somewhat improved.

sqlmesh/lsp/main.py
benfdking: That's if you don't get it no? It was correct but also no longer relevant.

sqlmesh/lsp/main.py
benfdking: Changed, it's mainly where I moved from, I kind of like the aspect of not having a empty dict before you can have Context.

sqlmesh/lsp/main.py
benfdking: To me for this to work, we will want to have an override which will probably just be an environment variable for the lsp. If it's set, we read the array of paths in that variable and then just set the path to it. 

Then we need to do something to make it possible in vscode but there are many config options.

georgesittas: Multi-repo projects are usually structured like this:

```
➜  multi_repo tree .
.
├── repo_1
│   └── config.yaml
└── repo_2
    └── config.yaml
```

If we want to be smart and infer the project's paths, we would have to start at the root directory (`multi_repo`) and search for configs in its sub-directories (`repo_1`, `repo_2`), as Themis suggested.

Otherwise, I would also approach this by allowing one to set an environment variable or VSCode config variable to point to the different paths explicitly, like you said Ben.

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,102 @@
...
+                path = audit._path.resolve()
+                if path == document.path:
+                    FILE_MAP[document.uri] = audit
+                    return document
+            return document
+
+    # If there is no context, load the context and then call this function again
+    path = Path(document.path).resolve()
+    if path.suffix not in (".sql", ".py"):
+        return document
```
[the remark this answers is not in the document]
benfdking: For me, I prefer the initial breakdown of whether context is loaded/not before we do anything.

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,102 @@
...
+
+def ensure_context_for_document(document: TextDocument) -> TextDocument:
+    """Ensure that a context exists for the given document if applicable by searching for a config.py or config.yml file in the parent directories."""
+    # If the context is already loaded, return the document, if it is part of the same context
+    global GLOBAL_CONTEXT, FILE_MAP
+    if GLOBAL_CONTEXT is not None:
+        GLOBAL_CONTEXT.load()
+        if document.uri in FILE_MAP:
+            return document
+        else:
```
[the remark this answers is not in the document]
georgesittas: I'd be happy to, I'm pretty sure we consistently suggest refactoring superfluous `else` blocks, might as well remove this burden from the reviewing cycles.

I'll let Toby / Iaroslav decide.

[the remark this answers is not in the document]
tobymao: sure we can institutionalize it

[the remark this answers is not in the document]
benfdking: https://github.com/TobikoData/sqlmesh/pull/4135

sqlmesh/lsp/main.py
```diff
@@ -0,0 +1,97 @@
...
+                    ),
+                    new_text=file.read(),
+                )
+            ]
+    except Exception as e:
+        ls.show_message(f"Error formatting SQL: {e}", types.MessageType.Error)
+        return []
+
+
+def main() -> None:
```
benfdking, line 91: We need to be able to pass in the context to support enterprise

sqlmesh/lsp/main.py
z3z1ma: Maybe multirepo needs explicit linkage in general rather than implicit linkage that is only ever represented by user input of multiple -p flags. 

A top level file called .sqlmeshrc could store something like 


```
# specific paths via flag defaults merged together
-p subdir/one
-p subdir/two

# another default flag
--log-file-dir my/dir

# specify command specific flag defaults
audit -s "1 month ago"

format --no-rewrite-casts

run --skip-janitor
```


Through this mechanism, you not only make multi repo explicit, you add a powerful mechanism for users to define behavior explicitly and centrally without implicit knowledge of cli flags with distributed usage. 

If you're interested in something like this, let me know. I'll convert it to a top level issue and take a spike at it.

izeigerman (approved):

PR #4132, opened 2025-04-11
Fix: qualify columns in source / target CTEs created for table diffing

georgesittas:
The current `table_diff` logic generate CTEs for the source & target tables that look like this:

```sql
WITH __source AS (
    SELECT <unqualified columns> FROM <source table>
), __target AS (
    SELECT <unqualified columns> FROM <target table>
), ...
```

There's an edge case where if (a) the target dwh is BigQuery and (b) either the source or target table's name matches one of its columns' name, then the corresponding projection for that column will instead be treated as a record of type `STRUCT`, because BigQuery treats it as a reference _to the table itself_.

The following demonstrates this behavior:

```sql
CREATE TABLE test-project.george.foo AS SELECT 1 AS foo, 2 AS y;
SELECT foo FROM test-project.george.foo;
```

This returns the following, which is basically a `STRUCT` that BigQuery auto-expands in its UI for UX reasons.

```
ROW | foo.foo | foo.y
  1 |       1 |     1
```

This causes problems because we're comparing those columns with each other and `=` isn't defined for `STRUCT`s.


tests/core/engine_adapter/integration/test_integration_bigquery.py
```diff
@@ -403,3 +403,31 @@ def test_column_types(ctx: TestContext):
...
+    ctx.engine_adapter.ctas(target_table, query)
+
+    table_diff = TableDiff(
+        adapter=ctx.engine_adapter,
+        source=exp.table_name(src_table),
+        target=exp.table_name(target_table),
+        on=["id"],
+    )
+
+    row_diff = table_diff.row_diff()
```
georgesittas, line 430: This statement in particular fails in main:

```
>       row_diff = table_diff.row_diff()
...
E               google.api_core.exceptions.BadRequest: 400 No matching signature for operator = for argument types: STRUCT<id INT64, source INT64, target INT64>, INT64
E                 Signature: T1 = T1
E                   Unable to find common supertype for templated argument <T1>
E                     Input types for <T1>: {INT64, STRUCT<id INT64, source INT64, target INT64>} at [1:1094]; reason: invalidQuery, location: query, message: No matching signature for operator = for argument types: STRUCT<id INT64, source INT64, target INT64>, INT64
E                 Signature: T1 = T1
E                   Unable to find common supertype for templated argument <T1>
E                     Input types for <T1>: {INT64, STRUCT<id INT64, source INT64, target INT64>} at [1:1094]```

**Review by tobymao — APPROVED** [review]
nice find

File: sqlmesh/core/table_diff.py [hunk 2]
```diff
@@ -317,18 +317,24 @@ def name(e: exp.Expression) -> str:
...
                     self.target_key_expression.as_(SQLMESH_JOIN_KEY_COL),
                 )
-                .from_(self.target_table)
+                .from_(self.target_table.as_("t"))
                 .where(self.where)
             )
 
+            # Ensure every column is qualified with the alias in the source and target queries
+            for col in source_query.find_all(exp.Column):
+                col.set("table", exp.to_identifier("s"))
```
tobymao, line 334: can you do this in the builder instead in the exp.column statement?

georgesittas: I originally did that but realized it's not as straightforward because we also [include a more complex expression](https://github.com/TobikoData/sqlmesh/blob/43655b285994e43ee23447f9c46adffeac491564/sqlmesh/core/table_diff.py#L318) that's [constructed elsewhere](https://github.com/TobikoData/sqlmesh/blob/43655b285994e43ee23447f9c46adffeac491564/sqlmesh/core/table_diff.py#L227-L246).

georgesittas: Update– it seems like we don't need to worry about the source/target key expressions, because these are properly aliased and so shouldn't clash with any of these tables' columns. I'll apply Toby's suggestion.

PR #4136, opened 2025-04-14
Docs: add custom loader example

treysp:


docs/guides/customizing_sqlmesh.md
```diff
@@ -0,0 +1,74 @@
...
+
+SQLMesh's `SqlMeshLoader` class handles the loading process - customize it by subclassing it and overriding its methods.
+
+!!! note "Python configuration only"
+
+    Custom loaders require using the [Python configuration format](./configuration.md#python) (YAML is not supported).
+
+### Modify every model
+
+One reason to customize the loading process is to do something to or verify something about every model. For example, you might want to add a post-statement to every model or verify that every model's `owner` field is populated.
```
tobymao, line 27: this is not a good use case, it's better to use the linter for that

docs/guides/customizing_sqlmesh.md
```diff
@@ -0,0 +1,74 @@
...
+        new_models = {}
+        # Loop through the existing model names/objects
+        for model_name, model in models.items():
+            # Create list of existing and new post-statements
+            new_post_statements = [
+                # Existing post-statements from model object
+                *model.post_statements,
+                # New post-statement is raw SQL, so we parse it with SQLGlot's `parse_one` function.
+                # Make sure to specify the SQL dialect.
+                parse_one(f"UNLOAD ...", dialect="redshift"),
```
izeigerman, line 62: I don't think this is a real statement. I suggest we put something more realistic in here

PR #4139, opened 2025-04-14
feat: introduce minimal extension

benfdking:
firt pass of the vscode extension that includes three features: 
1. lsp which can be used for formatting files
2. command to format whole project
3. ability to restart servers/react to interpreters being changed

# To test

open branch

run the following to set up your vscode environment with right settings (be mindful this might override yours)

```bash
make vscode_settings
```

launch the debugger task 'Run extension'

open a SQL file and in the command palette run `Format document` to document single file

run `Format SQLMesh project` to format whole project

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,3 +209,7 @@ clickhouse-cloud-test: guard-CLICKHOUSE_CLOUD_HOST guard-CLICKHOUSE_CLOUD_USERNA
 
 athena-test: guard-AWS_ACCESS_KEY_ID guard-AWS_SECRET_ACCESS_KEY guard-ATHENA_S3_WAREHOUSE_LOCATION engine-athena-install
 	pytest -n auto -x -m "athena" --retries 3 --retry-delay 10 --junitxml=test-results/junit-athena.xml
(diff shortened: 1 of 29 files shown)


vscode/extension/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2024 Tobiko Data Inc.
```
tobymao, line 189: date, also why does this need to be here, the vscode extension should be under the same license as the regular repo, it's in the same place

benfdking: You need it but just copying over at build time now.

tobymao (approved):

PR #4144, opened 2025-04-15
Fix: Handle macros in model properties when formatting

themisvaltinos:
This fixes an issue in formatting [model meta properties that have macro functions](https://github.com/TobikoData/sqlmesh/pull/2574) by handling them in `_props_sql` to account for macro funcs. 

Relevant slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1744698810151999?thread_ts=1741243847.726999&cid=C044BRE5W4S

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -635,7 +635,7 @@ def _props_sql(self: Generator, expressions: t.List[exp.Expression]) -> str:
 
     for i, prop in enumerate(expressions):
         if isinstance(prop, MacroFunc):
-            sql = self.indent(f"{self.sql(prop, comment=False)}")
+            sql = self.indent(self.sql(prop, comment=False))
         else:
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -634,7 +634,10 @@ def _props_sql(self: Generator, expressions: t.List[exp.Expression]) -> str:
     size = len(expressions)
 
     for i, prop in enumerate(expressions):
-        sql = self.indent(f"{prop.name} {self.sql(prop, 'value')}")
+        if isinstance(prop, MacroFunc):
+            sql = self.indent(f"{self.sql(prop, comment=False)}")
```
georgesittas, line 638: No need for f-string

georgesittas (approved):

sqlmesh/core/dialect.py
themisvaltinos: oh right good point

PR #4149, opened 2025-04-15
Feat: fit eval progress bar column widths to content

treysp:
Currently, the eval progress bar has static column widths.

With this PR, column widths will be based on the maximum length of information to display.

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -760,39 +760,9 @@ class TerminalConsole(Console):
             )
 
             # determine column widths
-            annotation_width = 0
-            for snapshot, intervals in batched_intervals.items():
-                interval_str_len = 0
-                for interval in intervals:
-                    interval_str_len = max(
-                        interval_str_len,
-                        len(
-                            _create_evaluation_model_annotation(
-                                snapshot, _format_evaluation_model_interval(snapshot, interval)
-                            )
-                        ),
-                    )
-
-                # The annotation includes audit results. We cannot build the audits result string
-                # until after evaluation occurs, but we must determine the annotation column width here.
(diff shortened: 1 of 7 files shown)


sqlmesh/core/console.py
```diff
@@ -754,10 +750,62 @@ def start_evaluation_progress(
...
+                                snapshot, _format_evaluation_model_interval(snapshot, interval)
+                            )
+                        ),
+                    )
+
+                # The annotation includes audit results. We cannot build the audits result string
+                # until after evaluation occurs, but we must determine the annotation column width here.
+                # Therefore, we add enough padding for the longest possible audits result string.
+                audit_str_len = 0
+                if snapshot.is_model and snapshot.model.audits:
```
izeigerman, line 778: Is there any opportunity to break this down into smaller functions?

izeigerman (approved):

sqlmesh/core/console.py
```diff
@@ -815,27 +830,21 @@ def update_snapshot_evaluation_progress(
...
-                annotation = _justify_evaluation_model_info(
-                    annotation + audits_str,
-                    annotation_width
-                    if not num_audits_failed
-                    else annotation_width - 1,  # -1 for RED_X_MARK's extra space
-                    dots_side="right",
-                    prefix=" \\[",
-                    suffix="]",
+                annotation_len = self.evaluation_column_widths["annotation"]
+                annotation = f"\[{annotation + audits_str}]".ljust(
```
themisvaltinos, line 834: I think we should escape the backslash here as we did because \[ is not a valid escape sequence in Python strings and causes a syntax error

PR #4152, opened 2025-04-16
feat: publish linter errors through lsp

benfdking:


sqlmesh/core/linter/rule.py
```diff
@@ -50,6 +50,19 @@ def __repr__(self) -> str:
         return f"{self.rule.name}: {self.violation_msg}"
 
 
+class RuleViolationWithModelAndType(RuleViolation):
```
georgesittas, line 53: Should we just shove the new logic in `RuleViolation`? Model and violation type seems useful in general?

sqlmesh/lsp/main.py
```diff
@@ -80,8 +152,16 @@ def ensure_context_for_document(self, document: TextDocument) -> TextDocument:
...
+        if self.context_and_reverse_map is not None:
+            context, _ = self.context_and_reverse_map
+            context.load()  # Reload or refresh context
+            self.context_and_reverse_map = (
+                context,
+                {
+                    f"file://{Path(model._path).resolve().as_posix()}": model.name
+                    for model in context._models.values()
+                },
+            )
```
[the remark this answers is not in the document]
benfdking: My understanding is that we only really need one because the linter will only really look at one file. 

If you pass many it will just look at the same file over and over again?

sqlmesh/lsp/main.py
```diff
@@ -31,14 +32,85 @@ def __init__(
         """
         self.server = LanguageServer(server_name, version)
         self.context_class = context_class
-        self.context: t.Optional[Context] = None
+        # A tuple of (context, reverse_map) where the reverse_map is uri to model name
+        self.context_and_reverse_map: t.Optional[t.Tuple[Context, t.Dict[str, str]]] = None
```
izeigerman, line 36: Any reason this shouldn't be 2 variables?

sqlmesh/lsp/main.py
```diff
@@ -31,14 +32,85 @@ def __init__(
...
 
+        @self.server.feature(types.TEXT_DOCUMENT_DID_OPEN)
+        def did_open(ls: LanguageServer, params: types.DidOpenTextDocumentParams) -> None:
+            if self.context_and_reverse_map is None:
+                self.ensure_context_for_document(
+                    ls.workspace.get_document(params.text_document.uri)
+                )
+            if self.context_and_reverse_map is None:
+                raise RuntimeError("No context found")
+            model = self.context_and_reverse_map[1][params.text_document.uri]
```
izeigerman, line 53: Yes, this is not very readable. Let's split them

sqlmesh/lsp/main.py
```diff
@@ -31,14 +32,85 @@ def __init__(
...
+            )
+
+        @self.server.feature(types.TEXT_DOCUMENT_DID_CHANGE)
+        def did_change(ls: LanguageServer, params: types.DidChangeTextDocumentParams) -> None:
+            if self.context_and_reverse_map is None:
+                self.ensure_context_for_document(
+                    ls.workspace.get_document(params.text_document.uri)
+                )
+            if self.context_and_reverse_map is None:
+                raise RuntimeError("No context found")
```
izeigerman, line 74: This looks like an exact copy paste of 46-51. I feel like this should be handled with something like:
```
def context_get_or_load()
```

sqlmesh/lsp/main.py
```diff
@@ -31,14 +32,85 @@ def __init__(
...
+            self.lint_cache.pop(params.text_document.uri, None)
+
+        @self.server.feature(types.TEXT_DOCUMENT_DID_SAVE)
+        def did_save(ls: LanguageServer, params: types.DidSaveTextDocumentParams) -> None:
+            if self.context_and_reverse_map is None:
+                self.ensure_context_for_document(
+                    ls.workspace.get_document(params.text_document.uri)
+                )
+            if self.context_and_reverse_map is None:
+                raise RuntimeError("No context found")
```
izeigerman, line 100: ditto

sqlmesh/lsp/main.py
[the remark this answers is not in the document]
izeigerman: as long as we don't use this mapping for anything other than linting

sqlmesh/lsp/main.py
```diff
@@ -80,8 +152,16 @@ def ensure_context_for_document(self, document: TextDocument) -> TextDocument:
         for a config.py or config.yml file in the parent directories.
         """
         # If the context is already loaded, check if this document belongs to it.
```
izeigerman, line 154: Is this comment even accurate? How do we check if the document belongs here?

sqlmesh/lsp/main.py
```diff
@@ -97,7 +177,15 @@ def ensure_context_for_document(self, document: TextDocument) -> TextDocument:
                 if config_path.exists():
                     try:
                         # Use user-provided instantiator to build the context
-                        self.context = self.context_class(paths=[path])
+                        context = self.context_class(paths=[path])
+                        self.context_and_reverse_map = (
+                            context,
+                            {
+                                f"file://{Path(model._path).resolve().as_posix()}": model.name
+                                for model in context._models.values()
```
izeigerman, line 185: This looks like an exact same thing as 158-164. I think there should be only one way to populate the map.

izeigerman, line 185: Any reason you're using `_models` instead of `_models`? I suggest to not rely on the protected context state.

sqlmesh/lsp/main.py
```diff
@@ -31,14 +32,85 @@ def __init__(
...
+            model = self.context_and_reverse_map[1][params.text_document.uri]
+            if model is None:
+                return
+            self.lint_cache[params.text_document.uri] = self.context_and_reverse_map[0].lint_models(
+                [model],
+                raise_on_error=False,
+            )
+            ls.publish_diagnostics(
+                params.text_document.uri,
+                SQLMeshLanguageServer._diagnostics_to_lsp_diagnostics(
```
izeigerman, line 62:
```suggestion
                self._diagnostics_to_lsp_diagnostics(
```

sqlmesh/lsp/main.py
```diff
@@ -110,6 +198,45 @@ def ensure_context_for_document(self, document: TextDocument) -> TextDocument:
...
+            else types.DiagnosticSeverity.Warning,
+        )
+
+    @staticmethod
+    def _diagnostics_to_lsp_diagnostics(
+        diagnostics: t.List[RuleViolationWithModelAndType],
+    ) -> t.List[types.Diagnostic]:
+        lsp_diagnostics: t.List[types.Diagnostic] = []
+        for diagnostic in diagnostics:
+            if diagnostic is None:
```
izeigerman, line 226: Why would it be None? The type signature doesn't reflect this

sqlmesh/lsp/main.py
```diff
@@ -110,6 +198,45 @@ def ensure_context_for_document(self, document: TextDocument) -> TextDocument:
...
+                continue
+            lsp_diagnostic = SQLMeshLanguageServer._diagnostic_to_lsp_diagnostic(diagnostic)
+            if lsp_diagnostic is not None:
+                lsp_diagnostics.append(lsp_diagnostic)
+        return lsp_diagnostics
+
+    @staticmethod
+    def _uri_to_path(uri: str) -> str:
```
izeigerman, line 234: Is this used anywhere?

sqlmesh/core/linter/rule.py
benfdking: To me, here you kind of need two classes: 
1. `RuleViolation` is the neat thing that a Rule emits
2. `RuleViolationWithModelAndType` is the type that comes from the rule where we have augmented it with severity and the rule that went in.

sqlmesh/lsp/main.py
benfdking: It's not pretty in this world, but for me, the reason to do this is that you accurately encode the states that a state can be: 
1. loaded with both Context and the dict
2. not loaded 

Rather than allowing the inconsistent states that sit in between.

PR #4153, opened 2025-04-16
Chore: Refine environment statements cli output

themisvaltinos:
This update revises the cli output of `before_all` `after_all` statements to 
- not include the python env differences on first plan
- use unidiff for it so that is in line with models instead of ndiff
- account for `--no-diff` flag and don't display the differences when its provided
- as part of this also decoupled `show_model_difference_summary`

Screenshot 2025-04-16 at 14 03 26

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -288,6 +288,7 @@ class Console(LinterConsole, StateExporterConsole, StateImporterConsole, Janitor
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
         no_diff: bool = True,
+        show_environment_statements: bool = True,
     ) -> None:
         """Displays a summary of differences for the given models."""
 
@@ -557,6 +558,7 @@ class NoopConsole(Console):
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
         no_diff: bool = True,
+        show_environment_statements: bool = True,
         ignored_snapshot_ids: t.Optional[t.Set[SnapshotId]] = None,
     ) -> None:
         pass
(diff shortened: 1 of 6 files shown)


sqlmesh/core/console.py
```diff
@@ -276,12 +276,13 @@ def stop_state_import(self, success: bool, input_file: Path) -> None:
...
 
     @abc.abstractmethod
-    def show_model_difference_summary(
+    def show_difference_summary(
         self,
         context_diff: ContextDiff,
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
         no_diff: bool = True,
+        show_environment_statements: bool = True,
```
izeigerman, line 285: Just double checking: are we going to have a separate method instead?

themisvaltinos: yes currently working on decoupling them

sqlmesh/core/context.py
```diff
@@ -1533,17 +1533,22 @@ def diff(self, environment: t.Optional[str] = None, detailed: bool = False) -> b
...
-            EnvironmentNamingInfo.from_environment_catalog_mapping(
-                self.config.environment_catalog_mapping,
-                name=environment,
-                suffix_target=self.config.environment_suffix_target,
-                normalize_name=context_diff.normalize_environment_name,
-            ),
-            self.default_catalog,
             no_diff=not detailed,
         )
+        if context_diff.has_changes:
```
georgesittas, line 1540: Why do we check this now? Why not for environment diffs?

georgesittas (approved):

sqlmesh/core/context.py
themisvaltinos: environment diffs does this internally and needs to display the appropriate messages accordingly. this is needed here since by decoupling them we don't want to display the model differences which weren't displayed before in this case since we were returning from the method instead before reaching that point

Issue #4079, opened 2025-04-03
Snowflake - Correctly handle cloning of transient tables for dev previews


erindru:
Given a model like so:

```
MODEL (
  name ANALYTICS_SILVER.EVENTS_PROJECTED,
  kind FULL,
  physical_properties (
    creatable_type = TRANSIENT
  ),
  cron '@daily',
  start '2021-07-08'
);

select * 
from ANALYTICS_BRONZE.RAW_DATALAKE_EVENTS;
```

If a plan to create a new environment results in a dev preview, SQLMesh tries to clone the table:
```
CREATE OR REPLACE TABLE "ANALYTICS_DB"."sqlmesh__ANALYTICS_SILVER"."ANALYTICS_SILVER__EVENTS_PROJECTED__3990895729__dev" CLONE "ANALYTICS_DB"."sqlmesh__ANALYTICS_SILVER"."ANALYTICS_SILVER__EVENTS_PROJECTED__2701799822"
```

This produces an error:
```
SQL compilation error: Transient object cannot be cloned to a permanent object.
```

The Snowflake adapter should check if the object is transient first before trying to perform a clone. If the object being cloned is transient, then the correct query is `CREATE OR REPLACE TRANSIENT TABLE`:

```
CREATE OR REPLACE TRANSIENT TABLE "ANALYTICS_DB"."sqlmesh__ANALYTICS_SILVER"."ANALYTICS_SILVER__EVENTS_PROJECTED__3990895729__dev" CLONE "ANALYTICS_DB"."sqlmesh__ANALYTICS_SILVER"."ANALYTICS_SILVER__EVENTS_PROJECTED__2701799822"
```

At a higher level, I wonder if we should actually be re-using the model `physical_properties` when creating derived tables like this

Slack ref: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1743715969481879

PR #4155, opened 2025-04-16
Fix(snowflake): Support cloning transient tables for dev previews

themisvaltinos:
This is an update for snowflake to support cloning transient tables for dev previews fixes: #4079 

When planning a new environment that results in the creation of a dev preview, creating a cloned table doesn't account for the physical properties. This ensures that if the source table is transient we account for it and clone using the correct `CREATE TRANSIENT TABLE` syntax.

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -469,3 +469,25 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin, ClusteredByMixi
                 f"Column comments for table '{table.alias_or_name}' not registered - this may be due to limited permissions.",
                 exc_info=True,
             )
+
+    def clone_table(
+        self,
+        target_table_name: TableName,
+        source_table_name: TableName,
+        replace: bool = False,
+        clone_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
+        **kwargs: t.Any,
+    ) -> None:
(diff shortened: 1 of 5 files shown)


sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -469,3 +469,25 @@ def _create_column_comments(
...
+    def clone_table(
+        self,
+        target_table_name: TableName,
+        source_table_name: TableName,
+        replace: bool = False,
+        clone_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        # The Snowflake adapter should use the transient property to clone transient tables
+        if physical_properties := kwargs.get("rendered_physical_properties"):
```
tobymao, line 482: i thought we weren't going to do this pattern anymore? this will fail on False and None {}

themisvaltinos: yes fair, in this case it’s safe since we ignore anything that’s not actual properties (`None`, `False`, ...) but happy to tweak it to avoid the pattern

erindru: Isnt this the exact place this pattern _should_ be used? It's basically saying "if rendered_physical_properties exists and has items"

I thought the "when not to use" related to if statements with a bunch of conditions, one being a walrus assignment in a place where `None` is a relevant value that should be handled

tobymao (approved):

PR #4157, opened 2025-04-16
fix: allow macro defs to references dates and other builtins

tobymao:


examples/sushi/models/waiter_revenue_by_day.sql
```diff
@@ -13,6 +13,8 @@ MODEL (
   grain (waiter_id, event_date)
 );
 
+@DEF(where_filter, o.event_date BETWEEN @start_date AND @end_date);
```
izeigerman, line 16: I suggest we add a proper test case in `test_model.py` instead

tobymao: added

izeigerman (approved):

PR #4158, opened 2025-04-16
fix: prevent past ttl values for environment and snapshot

eakmanrq:
Prior to this a user could enter a value like `1 week` and it would set the ttl in the past making all of their environments or snapshots immediately expire.

### DIFF
--- a/sqlmesh/core/config/root.py
+++ b/sqlmesh/core/config/root.py
@@ -41,6 +41,7 @@ from sqlmesh.core.config.ui import UIConfig
 from sqlmesh.core.loader import Loader, SqlMeshLoader
 from sqlmesh.core.notification_target import NotificationTarget
 from sqlmesh.core.user import User
+from sqlmesh.utils.date import to_timestamp, now, now_timestamp
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import field_validator, model_validator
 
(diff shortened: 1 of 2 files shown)


kelsin (approved):

sqlmesh/core/config/root.py
```diff
@@ -302,3 +303,20 @@ def dialect(self) -> t.Optional[str]:
...
+    if to_timestamp(v, relative_base=current_time) < to_timestamp(current_time):
+        raise ValueError(
+            f"TTL '{v}' is in the past. Please specify a relative time in the future. Ex: `in 1 week` instead of `1 week`."
+        )
+    return v
+
+
+if t.TYPE_CHECKING:
+    NoPastTTLString = str
+else:
```
erindru, line 319: Interesting pattern, I think its the first time i've seen this.

Is the purpose to alias the pydantic validator behind a more simple looking type, while not confusing MyPy?

erindru (approved):

sqlmesh/core/config/root.py
```diff
@@ -302,3 +303,20 @@ def dialect(self) -> t.Optional[str]:
     @property
     def fingerprint(self) -> str:
         return str(zlib.crc32(pickle.dumps(self.dict(exclude={"loader", "notification_targets"}))))
+
+
+def validate_no_past_ttl(v: str) -> str:
```
izeigerman, line 308: Can we stick to an existing pattern instead? I'm not sure I like the inconsistency

eakmanrq: Updated the other field validators: [`900b163` (#4158)](https://github.com/TobikoData/sqlmesh/pull/4158/commits/900b163dfb32e918c43c53407e5d659ba435a375)

izeigerman (approved):

PR #4167, opened 2025-04-17
refactor: further break apart console interface

benfdking:


sqlmesh/core/console.py
```diff
@@ -176,7 +184,45 @@ def stop_cleanup(self, success: bool = True) -> None:
...
+        context_diff: ContextDiff,
+        environment_naming_info: EnvironmentNamingInfo,
+        default_catalog: t.Optional[str],
+        no_diff: bool = True,
+    ) -> None:
+        """Displays a summary of differences for the given models."""
+
+
+class Console(
+    LinterConsole,
```
tobymao, line 218: in general multi inheritance is a bad pattern in python

izeigerman: These are all interfaces. The methods don't have implementation so it doesn't really matter

PR #4172, opened 2025-04-17
Chore!: bump sqlglot to v26.16.0

georgesittas:
@erindru could you double check that the Athena changes caused by this upgrade are legit? Based on this [comment](https://github.com/tobymao/sqlglot/pull/4962#discussion_r2036098959), I reckon that they should be good, just not quite sure.


georgesittas:
Bumped to v26.16.0 instead

benfdking (approved):

PR #4174, opened 2025-04-17
SQLMesh CLI Crash Course

sungchun12:
Get newcomers to learn the basics and some pro tips in 30 minutes. It's also designed as a fun copy/paste cheatsheet!

This is designed for newcomers to become "expert beginners" as fast as possible with sqlmesh. This is intended to cover 90% of the workflow, both in mental model and commands to run. 

The long tail 10% is a different scope.


docs/examples/sqlmesh_cli_crash_course.md
```diff
@@ -0,0 +1,1196 @@
...
+
+    ```bash
+    sqlmesh plan dev
+    ```
+
+    ```bash
+    sqlmesh plan <environment>
+    ```
+
+    If you want to move faster, you can add the `--auto-apply` flag to avoid the manual prompt.
```
afzaljasani, line 42: I would suggest reminding the user that --auto-apply is going to execute the plan. Maybe a disclaimer to only do this after they have seen what the full prompt details?

docs/examples/sqlmesh_cli_crash_course.md
```diff
@@ -0,0 +1,1196 @@
...
+You can make changes quickly and confidently through a simple command:
+
+- Plan the changes in your dev environment.
+- Apply the changes to your dev environment by entering `y` at the prompt.
+- Audit the changes (test data quality). This happens automatically when you apply the changes to dev.
+
+=== "SQLMesh"
+
+    ```bash
+    sqlmesh plan dev
```
afzaljasani, line 35: if someone runs sqlmesh plan and prod gets created then they will need to run sqlmesh plan dev --include-unmodified to create their first virtual data env. 

Otherwise they will run into this error :Error: Creating a new environment requires a change, but project files match the `prod` environment. Make a change or use the --include-unmodified flag to create a new environment without changes.

docs/examples/sqlmesh_cli_crash_course.md
sungchun12: updated

docs/examples/sqlmesh_cli_crash_course.md
sungchun12: updated

docs/examples/sqlmesh_cli_crash_course.md
```diff
@@ -0,0 +1,1198 @@
...
+
+## **Run on Production Schedule**
+
+SQLMesh schedules your transformation on a per-model basis in proper DAG order. This makes it easy to configure how often each step in your pipeline runs to backfill data without running when upstream models are late or failed. Rerunning from point of failure is also a default! Example below:
+
+`stg_transactions`(cron: `@hourly`) -> `fct_transcations`(cron: `@daily`). All times in UTC.
+
+1. `stg_transactions` runs hourly
+2. `fct_transcations` runs at 12am UTC if `stg_transactions` is fresh and updated since its most recent hour interval
+3. If `stg_transactions` failed from 11pm-11:59:59pm, it will prevent `fct_transcations` from running and put it in a `pending` state
```
treysp, line 918: The "pending" and "late" concepts are Tobiko Cloud only.

Should we make cloud- and OSS-specific sections here to delineate OSS scheduler features/behavior from cloud?

sungchun12: what happens in sqlmesh then?

sungchun12: updated language with anote

treysp (approved):

PR #4176, opened 2025-04-18
Fix: When initialising multiple connections pass concurrent tasks

themisvaltinos:
This update addresses an issue when the configured default gateway's concurrency (e.g. Athena with concurrent_tasks=4) differs from a model-specific gateway (e.g. BigQuery with concurrent_tasks=1) which is explicitly set in some of the models. In these cases, it leads to BigQuery not running in a multithreaded mode and failing to run concurrent jobs. 

The issue is that concurrent_tasks is passed to the snapshot evaluator, but the connection uses the gateway specific or default. This fix ensures that concurrent_tasks is passed to the connection as well to align all gateways for multithreaded use when applied.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2197,6 +2197,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         for gateway_name in self.config.gateways:
             if gateway_name != self.selected_gateway:
                 connection = self.config.get_connection(gateway_name)
+                connection.concurrent_tasks = self.concurrent_tasks
                 adapter = connection.create_engine_adapter()
                 self._engine_adapters[gateway_name] = adapter
         return self._engine_adapters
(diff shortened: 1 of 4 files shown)


sqlmesh/core/context.py
```diff
@@ -2197,6 +2197,7 @@ def engine_adapters(self) -> t.Dict[str, EngineAdapter]:
         for gateway_name in self.config.gateways:
             if gateway_name != self.selected_gateway:
                 connection = self.config.get_connection(gateway_name)
+                connection.concurrent_tasks = self.concurrent_tasks
```
izeigerman, line 2200: I don't like this implicit mutation. Can we instead add an optional `concurrent_tasks` argument to the `create_engine_adapter` method and use it if provided?

themisvaltinos: Right good point revised it to pass it to the method instead

izeigerman (approved):

PR #4178, opened 2025-04-18
Fix(motherduck): Attach multiple catalogs in MotherDuck

themisvaltinos:
This update addresses an issue in failing to attach multiple catalogs in MotherDuck by not raising an error internally when the path contains a name and by using `IF EXISTS` for attach. Also corrects minor error in docs.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -150,16 +150,12 @@ class DuckDBAttachOptions(BaseConfig):
         options_sql = f" ({', '.join(options)})" if options else ""
         alias_sql = ""
         # TODO: Add support for Postgres schema. Currently adding it blocks access to the information_schema
-        if self.type == "motherduck":
-            # MotherDuck does not support aliasing
-            md_db = self.path.replace("md:", "")
-            if md_db != alias.replace('"', ""):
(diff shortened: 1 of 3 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -150,16 +150,12 @@ def to_sql(self, alias: str) -> str:
         options_sql = f" ({', '.join(options)})" if options else ""
         alias_sql = ""
         # TODO: Add support for Postgres schema. Currently adding it blocks access to the information_schema
-        if self.type == "motherduck":
-            # MotherDuck does not support aliasing
-            md_db = self.path.replace("md:", "")
-            if md_db != alias.replace('"', ""):
-                raise ConfigError(
-                    f"MotherDuck does not support assigning an alias different from the database name {md_db}."
```
georgesittas, line 158: Do we no longer want this diagnostic?

georgesittas (approved):

sqlmesh/core/config/connection.py
georgesittas: Sounds good, thanks for clarifying 👍

eakmanrq (approved):

PR #4196, opened 2025-04-20
chore(vscode): not pacakge esbuild.js file

benfdking:


vscode/extension/.vscodeignore
```diff
@@ -13,3 +13,4 @@ vsc-extension-quickstart.md
 **/*.ts
 **/.vscode-test.*
 assets/logo.svg
+esbuild.js
```
themisvaltinos, line 16: along those lines could we also ignore `package-lock.json` isn't it for local installs? I'll defer to you though if its actually needed

themisvaltinos (approved):

vscode/extension/.vscodeignore
benfdking: image

^ it's already not included, that's where I saw the `esbuild.js` file.

PR #4222, opened 2025-04-22
Fix: Support on run start / on run end hooks of dbt packages

themisvaltinos:
This update improves the logic of supporting [dbt's on_run_start and on_run_end hooks](https://docs.getdbt.com/reference/project-configs/on-run-start-on-run-end) to also include `dbt_packages`.

To achieve this a list of environment statements is populated for each project which have each package's corresponding Jinja macro registry and statements. Also sourcing from the manifest's nodes directly rather than the runtime configuration (which doesn't contain this information) allows to get the hooks and associated macros from `dbt_packages` as well as the root project.

### DIFF
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -30,6 +30,7 @@ from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.jinja import (
     JinjaMacroRegistry,
     extract_macro_references_and_variables,
+    make_jinja_registry,
 )
 
 if sys.version_info >= (3, 12):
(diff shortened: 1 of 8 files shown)


sqlmesh/dbt/loader.py
```diff
@@ -238,59 +237,54 @@ def _load_requirements(self) -> t.Tuple[t.Dict[str, str], t.Set[str]]:
...
-                ],
-                after_all=[
-                    d.jinja_statement(stmt).sql(dialect=dialect) for stmt in on_run_end or []
-                ],
-                python_env=python_env,
-                jinja_macros=jinja_macros,
-            )
-        return None
+            context = project.context
+            for package_name, package in project.packages.items():
```
izeigerman, line 247: Can we confirm the order in which the project and package hooks are executed in dbt to make sure we match it

sqlmesh/dbt/loader.py
```diff
@@ -238,59 +237,54 @@ def _load_requirements(self) -> t.Tuple[t.Dict[str, str], t.Set[str]]:
...
+                context.set_and_render_variables(package.variables, package_name)
+                on_run_start: t.List[str] = []
+                on_run_end: t.List[str] = []
+                for hook in package.on_run_start.values():
+                    on_run_start.append(hook.sql)
+                for hook in package.on_run_end.values():
+                    on_run_end.append(hook.sql)
+
+                if statements := on_run_start + on_run_end:
+                    jinja_references, used_variables = extract_macro_references_and_variables(
```
izeigerman, line 257: This feels like something that could be a separate method

themisvaltinos: Revised to create the registry in a separate method. The extraction still needs to happen here to get the used variables

sqlmesh/dbt/loader.py
themisvaltinos: Yes the root project hooks are executed first and then the dbt packages hooks, so I added extra logic to guarantee they are executed first by adding them first in the list. Also for the `on_run_start`/`on_run_end` of each package to retain the order added to do the sorting based on the hooks `index` which dbt uses in the hook's node to retain the order

izeigerman (approved):

Issue #4090, opened 2025-04-07
Athena Engine iceberg bucket transforms not supported


darkcofy:
In athena when we want to optimise over high cardinality columns we use the athena bucket transforms
this allows us to sepcify a column and hash it to a set number of buckets as explained here

https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html

it is also suported by athena-dbt which we are currently trying to migrate from

izeigerman:
@darkcofy can you please share more info. What did you try in SQLMesh? How did it fail?

darkcofy:
Hey I tried setting them as an array in the partition by clause
```
Partition by ARRAY [bucket (event_name, bucket (farm_id, 5)]

```

As well as just a single one without an array 

Both threw errors in parsing

darkcofy:
> @darkcofy are you able to post your whole model file? That syntax looks incorrect. `partition_by` is meant to be provided in the `MODEL (.. )` block and not manually in the query


Hey here is my model block

```

model (
  name bronze_sqlmesh.entity_event,
   cron "@hourly",
   kind incremental_by_unique_key (
     unique_key (event_surrogate_key),
     auto_restatement_cron '@weekly'
   ),
   audits (
     unique_combination_of_columns(columns := (event_surrogate_key)),
     not_null(columns := (event_surrogate_key))
   ),
   partitioned_by day_partition,
   table_format iceberg,
   storage_format parquet
 );
```

erindru:
Ok, so assuming `day_partition` is an actual column name in the model, can you try:

```
model (
  name bronze_sqlmesh.entity_event,
  ...SNIP
   partitioned_by (day_partition, bucket(5, event_name), bucket(5, farm_id)),
 );
```

Note that the linked documentation defines the bucket syntax as `bucket(N, col)` and not `bucket(col, N)` like you originally specified

erindru:
Oh man, this is cursed.

Some operations require `CREATE TABLE` and some operations require `CREATE TABLE.. AS SELECT`. I'm *assuming* that your situation is generating a CTAS, if you were able to post the actual query that failed from the SQLMesh `logs/` folder, that would be helpful to confirm.

Anyway, on Athena, these operations hit two different engines under the hood.

- `CREATE TABLE` hits what we call the "Hive" engine
- `CREATE TABLE... AS SELECT` hits what we call the "Trino" engine

[...]

PR #4224, opened 2025-04-22
Fix!: Change how `partitioned_by` is parsed so that partition expressions with specialized AST nodes are captured

erindru:
This addresses issue #4090 

Prior to this PR, a partition expression in a `MODEL` block like:
```
MODEL (
 ...,
 partitioned_by (bucket(4, a), colb)
);
```

would be parsed into an `exp.Anonymous` node. SQLGlot gained support for `exp.PartitionedByBucket` (and `exp.PartitionedByTruncate`) in order to handle Iceberg transforms in a first class manner.

This is important because Athena inexplicably expects the arguments on the partition transforms to be in a different order depending on if the query is a `CREATE TABLE` or a `CTAS`.

This PR changes the `partitioned_by` property to be parsed with the SQLGlot `_parse_partitioned_by` parser instead of the fallback `_parse_bracket(_parse_field)` parser so that the correct AST nodes are created.

This has the effect of standardizing the parsing of this property.

This *change is breaking* because previously a single item was parsed as `exp.Paren` instead of `exp.Tuple` which meant it wasn't unpacked in the Pydantic field validator and ended up getting nested in the `partitioned_by` list.

Given the following model:
```
MODEL (
  name sqlmesh_audit.date_example,
  kind full,
  partitioned_by (date_id)
);
SELECT DATE('2025-02-01') as date_id
```

Before this change, `partitioned_by` looked like:
```
>>> model.partitioned_by
[Paren(
  this=Column(
    this=Identifier(this='date_id', quoted=True)))]
```

After this change, `partitioned_by` looks like:
```
>>> model.partitioned_by
[Column(
  this=Identifier(this='date_id', quoted=True))]
```

That is - the superfluous `exp.Paren` has been removed

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -134,7 +134,7 @@ jobs:
             - v1-nm-cache-{{ checksum "package-lock.json" }}
       - run:
           name: Install packages
-          command: npm ci
+          command: npm ci 
       - save_cache:
           key: v1-nm-cache-{{ checksum "package-lock.json" }}
           paths:
@@ -277,10 +277,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          #filters:
-          # branches:
-          #   only:
-          #     - main
+          filters:
+           branches:
+             only:
+               - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -134,7 +134,7 @@ jobs:
             - v1-nm-cache-{{ checksum "package-lock.json" }}
       - run:
           name: Install packages
-          command: npm ci 
+          command: npm ci
       - save_cache:
           key: v1-nm-cache-{{ checksum "package-lock.json" }}
           paths:
@@ -277,10 +277,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
+          # branches:
+          #   only:
+          #     - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/sqlmesh/migrations/v0081_update_partitioned_by.py
+++ b/sqlmesh/migrations/v0081_update_partitioned_by.py
@@ -1,0 +1,91 @@
+"""Remove superfluous exp.Paren references from partitioned_by"""
+
+import json
+
+import pandas as pd
+from sqlglot import exp
+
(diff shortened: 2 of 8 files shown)


tobymao (approved):

erindru:
In trying to write a migration for this, I stumbled across some extra complexity.

The Snapshot has the correct AST nodes when loaded from the project files.

However, in SQLMesh state, `partitioned_by` is just stored as a list of strings. A totally different code path is hit to turn this list of strings back into AST nodes and the AST nodes it returns are not correct (eg the `bucket` transform comes back as an `exp.Anonymous` node instead of an `exp.PartitionedByBucket` node).

To address this, i've raised [this PR](https://github.com/tobymao/sqlglot/pull/5006) against SQLGlot which will allow parsing the list of strings into an `exp.PartitionedByProperty` and thus trigger the creation of the correct AST nodes

.circleci/continue_config.yml
```diff
@@ -277,10 +277,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-           branches:
-             only:
-               - main
+          #filters:
```
erindru, line 280: todo: revert prior to merge

erindru:
@tobymao are you able to take another look? In particular at the migration. The scope of this PR increased since you first looked

sqlmesh/migrations/v0081_update_partitioned_by.py
```diff
@@ -0,0 +1,91 @@
...
+        ).from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+
+        if partitioned_by := parsed_snapshot["node"].get("partitioned_by"):
+            new_partitioned_by = []
+            for item in partitioned_by:
+                # rewrite '(foo)' to 'foo'
+                if item.startswith("(") and item.endswith(")"):
```
tobymao, line 53: why regex instead of parsing it?

tobymao: why don't we keep the same style? '(foo)' ?

tobymao (approved):

sqlmesh/migrations/v0081_update_partitioned_by.py
erindru: Isn't `startswith` a normal string match, not a regex?

Also, the goal was simply to strip redundant parenthesis so I thought it was much simpler to do this using string manipulation rather than trying to `parse_one()` each string (which requires a dialect), remove the `exp.Paren` nodes and then re-generate the strings

Issue #4055, opened 2025-03-30
Don't delete expired snapshots or environments records unless their associated data objects have been deleted


izeigerman:
We currently delete environment/snapshot records first, and only then do we attempt to clean up the views/tables associated with them.

This means that if a view or table deletion fails for any reason, the deletion process won't be reattempted later, as any information about its existence has already been removed from the state.

We need to change the order of operations and delete a record from the state only after the associated data object has been successfully deleted.

PR #4228, opened 2025-04-23
Fix: Enviroment & snapshot cleanup order

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/4055

This PR changes the order of deletions so that the actual views & physical tables are deleted **before** their state sync entries; This ensures that if the former fails, the state sync is not modified and a future `sqlmesh janitor` can pick up again the expired snapshots & environments on a later invocation.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -117,7 +117,7 @@ from sqlmesh.core.test import (
 from sqlmesh.core.user import User
 from sqlmesh.utils import UniqueKeyDict, Verbosity
 from sqlmesh.utils.dag import DAG
-from sqlmesh.utils.date import TimeLike, now_ds, to_timestamp, format_tz_datetime
+from sqlmesh.utils.date import TimeLike, now_ds, to_timestamp, format_tz_datetime, now_timestamp
 from sqlmesh.utils.errors import (
     CircuitBreakerError,
     ConfigError,
@@ -2350,11 +2350,14 @@ class GenericContext(BaseContext, t.Generic[C]):
         )
 
     def _run_janitor(self, ignore_ttl: bool = False) -> None:
-        # Clean up expired environments by removing their views and schemas
-        self._cleanup_environments()
+        current_ts = now_timestamp()
 
-        # Identify and delete expired snapshots
-        cleanup_targets = self.state_sync.delete_expired_snapshots(ignore_ttl=ignore_ttl)
+        # Clean up expired environments by removing their views and schemas
+        self._cleanup_environments(current_ts=current_ts)
+
+        cleanup_targets = self.state_sync.get_expired_snapshots(
+            ignore_ttl=ignore_ttl, current_ts=current_ts
+        )
 
         # Remove the expired snapshots tables
         self.snapshot_evaluator.cleanup(
@@ -2362,10 +2365,15 @@ class GenericContext(BaseContext, t.Generic[C]):
             on_complete=self.console.update_cleanup_progress,
         )
 
+        # Delete the expired snapshot records from the state sync
+        self.state_sync.delete_expired_snapshots(ignore_ttl=ignore_ttl, current_ts=current_ts)
+
         self.state_sync.compact_intervals()
 
-    def _cleanup_environments(self) -> None:
-        expired_environments = self.state_sync.delete_expired_environments()
+    def _cleanup_environments(self, current_ts: t.Optional[int] = None) -> None:
+        current_ts = current_ts or now_timestamp()
+
+        expired_environments = self.state_sync.get_expired_environments(current_ts=current_ts)
 
         cleanup_expired_views(
             default_adapter=self.engine_adapter,
@@ -2374,6 +2382,8 @@ class GenericContext(BaseContext, t.Generic[C]):
             warn_on_delete_failure=self.config.janitor.warn_on_delete_failure,
             console=self.console,
         )
+
+        self.state_sync.delete_expired_environments(current_ts=current_ts)
 
     def _try_connection(self, connection_name: str, validator: t.Callable[[], None]) -> None:
         connection_name = connection_name.capitalize()
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -281,6 +281,28 @@ class StateReader(abc.ABC):
             environment_names: An optional list of environment names to export. If not specified, all environments will be exported.
         """
 
+    @abc.abstractmethod
+    def get_expired_snapshots(
+        self, current_ts: int, ignore_ttl: bool = False
+    ) -> t.List[SnapshotTableCleanupTask]:
+        """Aggregates the id's of the expired snapshots and creates a list of table cleanup tasks.
+
+        Expired snapshots are snapshots that have exceeded their time-to-live
+        and are no longer in use within an environment.
+
+        Returns:
+           The list of table cleanup tasks.
+        """
+
+    @abc.abstractmethod
+    def get_expired_environments(self, current_ts: int) -> t.List[Environment]:
+        """Returns the expired environments.
+
+        Expired environments are environments that have exceeded their time-to-live value.
+        Returns:
+            The list of environments to remove, the filter to remove environments.
+        """
+
 
 class StateSync(StateReader, abc.ABC):
     """Abstract base class for snapshot and environment state management."""
@@ -309,7 +331,7 @@ class StateSync(StateReader, abc.ABC):
 
     @abc.abstractmethod
     def delete_expired_snapshots(
-        self, ignore_ttl: bool = False
+        self, ignore_ttl: bool = False, current_ts: t.Optional[int] = None
     ) -> t.List[SnapshotTableCleanupTask]:
         """Removes expired snapshots.
 
@@ -321,7 +343,7 @@ class StateSync(StateReader, abc.ABC):
                 all snapshots that are not referenced in any environment
 
         Returns:
-            The list of table cleanup tasks.
+            The list of snapshot table cleanup tasks.
         """
 
     @abc.abstractmethod
@@ -391,7 +413,9 @@ class StateSync(StateReader, abc.ABC):
         """
 
     @abc.abstractmethod
-    def delete_expired_environments(self) -> t.List[Environment]:
+    def delete_expired_environments(
+        self, current_ts: t.Optional[int] = None
+    ) -> t.List[Environment]:
         """Removes expired environments.
 
         Expired environments are environments that have exceeded their time-to-live value.
--- a/sqlmesh/core/state_sync/db/environment.py
+++ b/sqlmesh/core/state_sync/db/environment.py
@@ -162,43 +162,51 @@ class EnvironmentState:
             where=environment_filter,
         )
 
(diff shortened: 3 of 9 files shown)


sqlmesh/core/state_sync/db/environment.py
```diff
@@ -162,11 +162,12 @@ def finalize(self, environment: Environment) -> None:
             where=environment_filter,
         )
 
-    def delete_expired_environments(self) -> t.List[Environment]:
-        """Deletes expired environments.
+    def get_expired_environments(self) -> t.Tuple[t.List[Environment], exp.LTE]:
```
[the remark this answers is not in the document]
themisvaltinos: So line 2363 would take this cleanup task: https://github.com/TobikoData/sqlmesh/blob/14d63c21cf40006fd52a0d7ecd72f7ae80a301bd/sqlmesh/core/context.py#L2363

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -273,20 +273,35 @@ def unpause_snapshots(
     def invalidate_environment(self, name: str) -> None:
         self.environment_state.invalidate_environment(name)
 
+    def get_expired_snapshots(
+        self, ignore_ttl: bool = False
+    ) -> t.Tuple[t.Set[SnapshotId], t.List[SnapshotTableCleanupTask]]:
```
izeigerman, line 278: I think the public interface should be clean and only return what external caller cares about, which is the list of cleanup targets.

The only caller that cares about snapshot IDs is the facade itself. I suggest having a common private method which returns a tuple, and then use it in both public methods instead.

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -269,10 +270,16 @@ def _is_snapshot_used(snapshot: SharedVersionSnapshot) -> bool:
                         )
                     )
 
-        if expired_snapshot_ids:
-            self.delete_snapshots(expired_snapshot_ids)
         return expired_snapshot_ids, cleanup_targets
 
+    def delete_expired_snapshots(self, expired_snapshot_ids: t.Set[SnapshotId]) -> None:
```
izeigerman, line 275: Why do we even need this method? Why not call `delete_snapshots` directly?

sqlmesh/core/state_sync/db/facade.py
izeigerman: Actually, we don't even need a private method, since we have `self.snapshot_state.get_expired_snapshots(`

izeigerman: Ah, I see now that you might actually use the snapshot IDs in the context.

sqlmesh/core/context.py
```diff
@@ -2337,27 +2337,30 @@ def _run_janitor(self, ignore_ttl: bool = False) -> None:
...
+        _, cleanup_targets = self.state_sync.get_expired_snapshots(ignore_ttl=ignore_ttl)
 
         # Remove the expired snapshots tables
         self.snapshot_evaluator.cleanup(
             target_snapshots=cleanup_targets,
             on_complete=self.console.update_cleanup_progress,
         )
 
+        # Delete the expired snapshot records from the state sync
+        self.state_sync.delete_expired_snapshots(ignore_ttl=ignore_ttl)
```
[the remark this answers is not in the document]
izeigerman: Why not just expose `delete_snapshots` and call it instead?

sqlmesh/core/environment.py
```diff
@@ -228,6 +229,11 @@ class EnvironmentStatements(PydanticModel):
     jinja_macros: t.Optional[JinjaMacroRegistry] = None
 
 
+class EnvironmentCleanupTask(PydanticModel):
+    expired_environments: t.List[Environment]
+    filter_expr: exp.LTE
```
izeigerman, line 234: This is IMHO not a good interface and a significant abstraction leak. 

Separately, why environment names are not enough? Why do we need full environment objects?

sqlmesh/core/context.py
```diff
@@ -2337,27 +2337,30 @@ def _run_janitor(self, ignore_ttl: bool = False) -> None:
...
+
         self.state_sync.compact_intervals()
 
     def _cleanup_environments(self) -> None:
-        expired_environments = self.state_sync.delete_expired_environments()
+        cleanup_targets = self.state_sync.get_expired_environments()
```
izeigerman, line 2354: I think this should just return environment names that are then passed into `delete_environments`.

sqlmesh/core/state_sync/db/environment.py
```diff
@@ -181,24 +182,42 @@ def delete_expired_environments(self) -> t.List[Environment]:
...
-            for env in environments
+            for env in cleanup_targets.expired_environments
         ]:
             self.engine_adapter.delete_from(
                 self.environment_statements_table,
                 where=exp.or_(*expired_environments),
             )
 
-        return environments
+    def delete_expired_environments(self) -> t.List[Environment]:
```
izeigerman, line 212: Why do we even need this method? Shouldn't we just have `delete_environments` that takes a list of environment names?

sqlmesh/core/context.py
VaggelisD: Synced over this, we need the full `Enviroment` in `cleanup_expired_views` so will keep it as is for now.

sqlmesh/core/state_sync/db/snapshot.py
VaggelisD: Synced over this, `delete_expired_snapshots()` is used as a standalone function in `test_state_sync.py` so will leave it as is

sqlmesh/core/state_sync/db/facade.py
VaggelisD: I think you were right initially, `expired_snapshot_ids` is needed in `facade.py` but not in `context.py`, do we still want to break `get_expired_snapshots` apart in 2? 

Not sure if we revisited this during our sync after we went over the nuances

sqlmesh/core/context.py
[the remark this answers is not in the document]
VaggelisD: Similar discussion here https://github.com/TobikoData/sqlmesh/pull/4228/files/4329f8d9f9c53a8c71ebef936943bef9adba35ef#r2056482177

sqlmesh/core/state_sync/db/environment.py
VaggelisD: Synced over this too, similar to `delete_expired_snapshots()`, we use it as a standalone function in tests so we aligned on leaving it as is and cleaning it up in the future.

sqlmesh/core/state_sync/base.py
```diff
@@ -302,9 +301,23 @@ def delete_snapshots(self, snapshot_ids: t.Iterable[SnapshotIdLike]) -> None:
             snapshot_ids: A list of snapshot like objects to delete.
         """
 
+    @abc.abstractmethod
+    def get_expired_snapshots(
```
izeigerman, line 305: This should go into `StateReader` interface.

sqlmesh/core/state_sync/base.py
```diff
@@ -386,7 +399,27 @@ def finalize(self, environment: Environment) -> None:
         """
 
     @abc.abstractmethod
-    def delete_expired_environments(self) -> t.List[Environment]:
+    def get_expired_environments(self, current_ts: int) -> t.List[Environment]:
```
izeigerman, line 402: This too should go into `StateReader`

sqlmesh/core/state_sync/base.py
```diff
@@ -302,9 +301,23 @@ def delete_snapshots(self, snapshot_ids: t.Iterable[SnapshotIdLike]) -> None:
...
+
+        Returns:
+            The set of expired snapshot ids.
+            The list of table cleanup tasks.
+        """
+
     @abc.abstractmethod
     def delete_expired_snapshots(
-        self, ignore_ttl: bool = False
+        self, ignore_ttl: bool = False, current_ts: t.Optional[int] = None
```
izeigerman, line 320: Why is the current_ts optional here but not in `delete_environments`. Should be consistent I think

sqlmesh/core/state_sync/db/environment.py
```diff
@@ -162,30 +162,33 @@ def finalize(self, environment: Environment) -> None:
...
+                where=self._create_filter_expr(current_ts),
                 lock_for_update=True,
             ),
         )
-        environments = [self._environment_from_row(r) for r in rows]
+        expired_environments = [self._environment_from_row(r) for r in rows]
 
+        return expired_environments
+
+    def delete_environments(self, environments: t.List[Environment], current_ts: int) -> None:
```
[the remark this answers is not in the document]
izeigerman: I believe we concluded that we need to recompute the list of environments to delete to account for possible race conditions between when we computed the initial cleanup least and now.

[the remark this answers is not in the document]
izeigerman: So +1 to @themisvaltinos. We should get rid of `delete_environments` and just keep `delete_expired_environments`

sqlmesh/core/state_sync/base.py
VaggelisD: So currently we have:

- `delete_expired_snapshots()` & `delete_expired_environments()`: The ts is optional because these functions can be called on their own
- `get_expired_snapshots()` & `get_expired_environments()`: The ts here is required because I wanted to (1) constraint them on having a common timestamp (they "shoehorn" the same ts down to the delete functions) and (2) we don't have to do `current_ts = current_ts or now_timestamp()` at every call


What do you think? Should they all be optional?

tests/core/test_context.py
```diff
@@ -775,7 +776,13 @@ def test_janitor(sushi_context, mocker: MockerFixture) -> None:
     adapter_mock.dialect = "duckdb"
     state_sync_mock = mocker.MagicMock()
 
-    state_sync_mock.delete_expired_environments.return_value = [
+    now_ts = now_timestamp()
+    filter_expr = exp.LTE(
```
izeigerman, line 780: Dead code?

tests/core/test_context.py
```diff
@@ -775,7 +776,13 @@ def test_janitor(sushi_context, mocker: MockerFixture) -> None:
     adapter_mock.dialect = "duckdb"
     state_sync_mock = mocker.MagicMock()
 
-    state_sync_mock.delete_expired_environments.return_value = [
+    now_ts = now_timestamp()
```
izeigerman, line 779: this too

nickmuoh:
I believe this PR would fix the following reported issues in the slack channel:
* https://tobiko-data.slack.com/archives/C044BRE5W4S/p1741157298019859
* https://tobiko-data.slack.com/archives/C044BRE5W4S/p1742505134599459

izeigerman (approved):
One minor comment. lgtm otherwise. Thanks a lot for addressing comments!

tests/core/test_context.py
VaggelisD: Looks like I pasted that accidentally, `make style` didn't complain for some reason, thanks for catching that!

PR #4230, opened 2025-04-23
feat(vscode): add installation progress bar for sqlmesh enterprise

benfdking:
Adds progress notification when installing sqlmesh enterprise.

image

### DIFF
--- a/vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
+++ b/vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
@@ -1,14 +1,15 @@
 import path from "path"
-import { traceLog, traceVerbose } from "../common/log"
+import { traceInfo, traceLog, traceVerbose } from "../common/log"
 import { getInterpreterDetails } from "../common/python"
 import { Result, err, isErr, ok } from "../functional/result"
 import { getProjectRoot } from "../common/utilities"
-import { execFile } from "child_process"
-import { promisify } from "util"
 import { isPythonModuleInstalled } from "../python"
 import fs from "fs"
 import { ErrorType } from "../errors"
 import { isSignedIntoTobikoCloud } from "../../auth/auth"
+import { execAsync } from "../exec"
+import z from "zod"
+import { ProgressLocation, window } from "vscode"
 
 export type sqlmesh_exec = {
   workspacePath: string;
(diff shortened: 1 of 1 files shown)


vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
```diff
@@ -48,12 +49,114 @@ export const get_tcloud_bin = async (): Promise<Result<string, string>> => {
...
+    },
+    async (progress, token) => {
+      // Connect the cancellation token to our abort controller
+      token.onCancellationRequested(() => {
+        abortController.abort()
+        traceInfo("Sqlmesh enterprise installation cancelled")
+        window.showInformationMessage("Installation cancelled")
+      })
+      progress.report({ message: "Installing sqlmesh enterprise..." })
+      const installResult = await installSqlmeshEnterprise(abortController)
```
themisvaltinos, line 139: can this be renamed or is there a reason this shadows the outer scoped `installResult`?

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
```diff
@@ -48,12 +49,114 @@ export const get_tcloud_bin = async (): Promise<Result<string, string>> => {
...
+  if (isErr(tcloudBin)) {
+    return err(tcloudBin.error)
+  }
+  const called = await execAsync(tcloudBin.value, ["is_sqlmesh_installed"])
+  if (called.exitCode !== 0) {
+    return err(
+      `Failed to check if sqlmesh enterprise is installed: ${called.stderr}`
+    )
+  }
+  const parsed = isSqlmeshInstalledSchema.safeParse(JSON.parse(called.stdout))
```
themisvaltinos, line 75: does `called.stdout` need handling too so that json.parse doesn't throw an error or safeparse handles it?

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
```diff
@@ -88,6 +191,13 @@ export const sqlmesh_exec = async (): Promise<Result<sqlmesh_exec, ErrorType>> =
           type: "not_signed_in",
         })
       }
+      const ensured = await ensureSqlmeshEnterpriseInstalled()
+      if (isErr(ensured)) {
```
themisvaltinos, line 195: seeing a lot of this pattern with  if (isErr(..)) wondering if a helper could be useful. on the other hand it should be ok since you do have some different logic on the success paths

themisvaltinos (approved):

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
benfdking: Why would stdout need handling?

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
benfdking: Added lint rule

PR #4231, opened 2025-04-23
Fix: remove extraneous spacing between CLI annotation and runtime

treysp:
The CLI execution progress bar prints an annotation for each model after it is executed.

Currently, the annotation column width includes unnecessary space due to how it is calculated. This PR removes that space.

Column width calculations (`max` is over snapshots):
- Old: `max(len(annotation)) + max(len(audits))`
- Corrected: `max(len(annotation) + len(audits))`


izeigerman:
Can we somehow add a test with long model names and check the alignment?

PR #4238, opened 2025-04-24
feat: adds LinterContext

ravenac95:
Adds a LinterContext to LinterRules.

One of the things my team realized with the new linter rules (awesome, btw) but we also wanted to get the context of a model and it's relation to other models in some contexts of linting rules. 

For example, we have a case where anything downstream of a specific model we'd like it to include a specific tag or set of tags. This is to force the correct labeling of some models that are derived from the lineage of some other model. 

Notes:
* I figured it was bad form to pass the _full_ context into the Rule.
* In order to prevent a breaking change (that forces the Rule to include a `context`) object, I expose the context via the two new methods on the `Rule` base class. I'd prefer, honestly, to just pass in the `context` object but that would force all current users to change their implementations.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -312,6 +312,20 @@ class ExecutionContext(BaseContext):
         )
 
 
+class LinterContext:
+    """Limited context used only for linting"""
+
+    def __init__(self, model_retriever: t.Callable[[ModelOrSnapshot], t.Optional[Model]], dag: DAG):
+        self._dag = dag
+        self._model_retriever = model_retriever
+
+    def get_model(self, model_or_snapshot: ModelOrSnapshot) -> t.Optional[Model]:
+        return self._model_retriever(model_or_snapshot)
+
+    def get_dag(self) -> DAG:
+        return self._dag
+
+
 class GenericContext(BaseContext, t.Generic[C]):
     """Encapsulates a SQLMesh environment supplying convenient functions to perform various tasks.
 
@@ -2534,11 +2548,13 @@ class GenericContext(BaseContext, t.Generic[C]):
             list(self.get_model(model) for model in models) if models else self.models.values()
         )
         all_violations = []
+
+        linter_context = self.linter_context()
         for model in model_list:
             # Linter may be `None` if the context is not loaded yet
             if linter := self._linters.get(model.project):
                 lint_violation, violations = (
-                    linter.lint_model(model, console=self.console) or found_error
+                    linter.lint_model(linter_context, model, console=self.console) or found_error
                 )
                 if lint_violation:
                     found_error = True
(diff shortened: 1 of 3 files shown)


ravenac95:
I figured I'd add any tests after any review of the current design.

VaggelisD:
Hey @ravenac95, thank you for your kind words and for the contribution! Your use case makes sense, initially we had considered that exact scenario and passing `Context` to rules but decided to KISS, looks like it caught up already.

> I figured it was bad form to pass the full context into the Rule.

I get where you are coming from but I'm not sure if it's a considerable (or more accurately, worth the ROI) concern. I fear that moving forward with `LinterContext` will increase complexity by introducing the separate entity, will require work every time a new use case appears that requires new state from `Context` etc. I think just passing the full context will simplify things and make it truly future-proof, albeit putting some responsibility on the user to not modify it accidentally. 

<br />


> In order to prevent a breaking change (that forces the Rule to include a context) object, I expose the context via the two new methods on the Rule base class. I'd prefer, honestly, to just pass in the context object but that would force all current users to change their implementations.

Does that refer to passing `Context` as a kwarg in `check_model(...)`? I think passing `Context` in the `Rule` constructor is fine, or at the very least we can make it an optional (defaults to `None`) kwarg. 

Lastly, I think I'd drop the added `Rule` functions that simply call the underlying `Context` functions. Users can either use `self.context.<attr>` directly or we can consider redirecting the getters by overriding `Rule:: __getattr__` but that's more of a long shot.


<br />

Will be happy to get your input, will also wait for the other reviewers to chime in with their thoughts before we move on with the rest of the implementation.

ravenac95:
@VaggelisD Thanks for the quick response:

> I get where you are coming from but I'm not sure if it's a considerable (or more accurately, worth the ROI) concern. I fear that moving forward with LinterContext will increase complexity by introducing the separate entity, will require work every time a new use case appears that requires new state from Context etc. I think just passing the full context will simplify things and make it truly future-proof, albeit putting some responsibility on the user to not modify it accidentally.

Totally hear you, totally understand. I felt like the encapsulation here was theoretically better but practically more work so I totally get this. I was considering just putting the whole context here initially but opted against it. I'll make this change now.

> Does that refer to passing Context as a kwarg in check_model(...)? I think passing Context in the Rule constructor is fine, or at the very least we can make it an optional (defaults to None) kwarg.

Oh, I mostly meant that adding `kwargs` argument into `check_model` will static typing checks to break for all of the people who've created rules. Without static type checking, I think it's fine so perhaps it's not really a "breaking" change. 

> Lastly, I think I'd drop the added Rule functions that simply call the underlying Context functions. Users can either use self.context.<attr> directly or we can consider redirecting the getters by overriding Rule:: __getattr__ but that's more of a long shot.

Understood! I'll remove. Especially if the context is passed into `check_model` then this would become unnecessary.

tobymao:
thanks for the PR @ravenac95 but i think we're gonna go with this approach 

https://github.com/TobikoData/sqlmesh/pull/4247/files

PR #4243, opened 2025-04-24
Feat: Include stack trace in rendering exception handling

themisvaltinos:
Include the stack trace in the macro exception to help identify the error.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import traceback
 import typing as t
 from contextlib import contextmanager
 from functools import partial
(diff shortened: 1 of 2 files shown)


sqlmesh/core/renderer.py
```diff
@@ -211,12 +212,16 @@ def _resolve_table(table: str | exp.Table) -> str:
                     expressions = [e for e in parse(rendered_expression, read=self._dialect) if e]
 
                     if not expressions:
-                        raise ConfigError(f"Failed to parse an expression:\n{self._expression}")
+                        raise ConfigError(
+                            f"{traceback.format_exc()}"
```
izeigerman, line 216: Why do we include the stacktrace here? This is just parsing right?

izeigerman: If you're doing this for jinja's sake then I don't think it makes sense. As far as I remember jinja's stacktrace is pretty useless

PR #4249, opened 2025-04-25
Fix: 'duplicate definitions' python env bug due to module import order

georgesittas:
See the added test for [context](https://github.com/TobikoData/sqlmesh/pull/4249/files#diff-f3de4d159300cd3fc7e7073f004c2eac345751fe96c2d1754a0b76a453eb4098R9300-R9312) around this.

### DIFF
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -61,6 +61,16 @@ def _code_globals(code: types.CodeType) -> t.Dict[str, None]:
     return variables
 
 
+def _globals_match(obj1: t.Any, obj2: t.Any) -> bool:
+    return type(obj1) == type(obj2) and (
+        obj1 == obj2
+        or (
+            getattr(obj1, "__module__", None) == getattr(obj2, "__module__", None)
+            and getattr(obj1, "__name__", None) == getattr(obj2, "__name__", None)
+        )
+    )
+
+
 def func_globals(func: t.Callable) -> t.Dict[str, t.Any]:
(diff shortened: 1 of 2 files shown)


sqlmesh/utils/metaprogramming.py
```diff
@@ -61,6 +61,14 @@ def _code_globals(code: types.CodeType) -> t.Dict[str, None]:
     return variables
 
 
+def _globals_match(obj1: t.Any, obj2: t.Any) -> bool:
+    return obj1 == obj2 or (
```
tobymao, line 65: do you want == or is? we don't want False and 0 to match right

georgesittas: yeah, I guess we have this bug today because we used to compare the objects using `!=`:

```python
>>> False != 0
False
```

I'll fix this by checking the type first.

georgesittas: @tobymao I fixed this [here](https://github.com/TobikoData/sqlmesh/compare/d6a6d639265c3e0695ad6d45b7127107c436a959..96e18c9592de55b9f154433f27d219b9b3e4b7dd). There was actually a bug where we wouldn't detect clashing definitions for these globals, because we returned from `build_env`'s `walk` too early due to the `name in visited` condition. I also fixed that bug and added a test.

tobymao (approved):

sqlmesh/utils/metaprogramming.py
tobymao: looks good

PR #4254, opened 2025-04-25
Fix: Expand the parent's model query during query rendering if the parent is not categorized

izeigerman:
The uncategorized parent won’t have a physical table available, so if the user runs `evaluate` on a downstream child that relies on uncategorized upstream changes, the command will fail.

This update makes it so that uncategorized parents are inlined into the target model's query automatically during rendering.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -317,7 +317,9 @@ class BaseExpressionRenderer:
                 **table_mapping,
             }
             expand = set(expand) | {
-                name for name, snapshot in snapshots.items() if snapshot.is_embedded
+                name
+                for name, snapshot in snapshots.items()
+                if snapshot.is_embedded or not snapshot.categorized
             }
 
             if expand:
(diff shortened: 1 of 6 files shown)


sqlmesh/core/renderer.py
```diff
@@ -317,7 +317,9 @@ def _resolve_tables(
                 **table_mapping,
             }
             expand = set(expand) | {
-                name for name, snapshot in snapshots.items() if snapshot.is_embedded
+                name
+                for name, snapshot in snapshots.items()
+                if snapshot.is_embedded or not snapshot.categorized
             }
```
georgesittas, line 323: Does this also render uncategorized grandparents? I suppose the `render_query` in 342 handles this recursively?

georgesittas (approved):

tobymao (approved):

sqlmesh/core/renderer.py
izeigerman: Yes, because `expand` itself calls `render_query` which triggers the recursive sequence.

georgesittas (approved):

PR #4257, opened 2025-04-25
Fix!: exclude Semicolon expressions from model state

### LINKED ISSUE #4252: Comment at end of model script results in model query to be included in post statements
Having a comment at the end of a model script results in model query to be included in post statements in the snapshot metadata.

SQLMesh version: 0.176.0

## Reproducible example

Using the scaffolded example created by `sqlmesh init duckdb`.

Then adding a comment to the file `models/incremental_model.sql`: 

```sql
MODEL (
  name sqlmesh_example.incremental_model,
    time_column event_date
  ),
  start '2020-01-01',
  cron '@daily',
  grain (id, event_date)
);

SELECT
  id,
  item_id,
  event_date
WHERE
  event_date BETWEEN @start_date AND @end_date
;

--Just a comment
```

Then I did `sqlmesh plan`.

Inspecting the snapshots in the statestore reveils the following in the `snapshot` column (for "db"."sqlmesh_example"."incremental_model"):

```
...
    "pre_statements": [],
    "post_statements": [
      "SELECT\n  id,\n  item_id,\n  event_date\nFROM sqlmesh_example.seed_model\nWHERE\n  event_date BETWEEN @start_date AND @end_date\n;"
    ],
    "on_virtual_update": [],
    "query": "SELECT\n  id,\n  item_id,\n  event_date\nFROM sqlmesh_example.seed_model\nWHERE\n  event_date BETWEEN @start_date AND @end_date\n;",
    "source_type": "sql"
...
```

Without the comment the result is

```
    "pre_statements": [],
    "post_statements": [],
    "on_virtual_update": [],
    "query": "SELECT\n  id,\n  item_id,\n  event_date\nFROM sqlmesh_example.seed_model\nWHERE\n  event_date BETWEEN @start_date AND @end_date\n;",
    "source_type": "sql"
```

georgesittas:
Fixes #4252

I reproduced the issue locally by following the steps outlined in the above issue:

```python
>>> ctx.models['"db"."sqlmesh_example"."incremental_model"'].json()
'{"name":"sqlmesh_example.incremental_model","project":"","start":"2020-01-01","cron":"@daily","tags":[],"dialect":"duckdb","kind":{"name":"INCREMENTAL_BY_TIME_RANGE","on_destructive_change":"ERROR","dialect":"duckdb","forward_only":false,"disable_restatement":false,"time_column":{"column":"\\"event_date\\"","format":"%Y-%m-%d"},"partition_by_time_column":true},"partitioned_by":[],"clustered_by":[],"default_catalog":"db","audits":[],"grains":["(id, event_date)"],"references":[],"allow_partials":false,"signals":[],"enabled":true,"python_env":{},"jinja_macros":{"packages":{},"root_macros":{},"global_objs":{},"create_builtins_module":"sqlmesh.utils.jinja","top_level_packages":[]},"audit_definitions":{},"mapping_schema":{"\\"db\\"":{"\\"sqlmesh_example\\"":{"\\"seed_model\\"":{"id":"INT","item_id":"INT","event_date":"DATE"}}}},"extract_dependencies_from_query":true,"pre_statements":[],"post_statements":["SELECT\\n  id,\\n  item_id,\\n  event_date,\\nFROM\\n  sqlmesh_example.seed_model\\nWHERE\\n  event_date BETWEEN @start_date AND @end_date\\n;"],"on_virtual_update":[],"query":"SELECT\\n  id,\\n  item_id,\\n  event_date,\\nFROM\\n  sqlmesh_example.seed_model\\nWHERE\\n  event_date BETWEEN @start_date AND @end_date\\n;","source_type":"sql"}'
```

I observed that we hit [this line](https://github.com/TobikoData/sqlmesh/blob/c1ae64f6e13dee5ebf340ee2fee23e11f4acd01b/sqlmesh/core/dialect.py#L935) for the `Semicolon` expression and store the query's SQL in its `meta` dict. This explains the duplication of the model's query in the above serialized `post_statements` key.

My initial approach was to avoid setting `meta["sql"]` for `Semicolon` expressions, but then it occurred to me that we shouldn't be storing any `Semicolon` expressions in the model, because they can also affect the data/metadata hash.

I verified for the issue's examples that if I ran plan, removed the semicolon+comment, and then ran plan again then a diff was shown. This is because `Semicolon` is `gen`'d since it's in the `post_statements` list and we get back a `"SEMICOLON"` string in the `_data_hash_values`. I added a test for this.

I think we need a migration for this change. I'll take it to the finish line ~on Monday~ soon.

### DIFF
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -9332,6 +9332,7 @@ def test_python_env_references_are_unequal_but_point_to_same_definition(tmp_path
 
     db_path = str(tmp_path / "db.db")
     db_connection = DuckDBConnectionConfig(database=db_path)
+
     config = Config(
         gateways={"duckdb": GatewayConfig(connection=db_connection)},
         model_defaults=ModelDefaultsConfig(dialect="duckdb"),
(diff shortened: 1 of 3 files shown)


tests/core/test_model.py
```diff
@@ -9235,3 +9235,96 @@ def test_runtime_stage(evaluator):
...
+
+        --Just a comment
+        """
+    )
+
+    ctx = Context(paths=tmp_path, config=config)
+    model = ctx.get_model("sqlmesh_example.incremental_model_with_semicolon")
+
+    assert not model.pre_statements
+    assert not model.post_statements
```
georgesittas, line 9277: Without the change in `common.py`, both this assertion and the one in L9330 fail.

tobymao (approved):

PR #4258, opened 2025-04-25
Feat: return non-zero code when at least one audit fails in sqlmesh audit

### LINKED ISSUE #4099: retrun code from audit cmd is allways 0
Either the audit fail or not, the return code for the `sqlmesh audit` is allways 0.

It should be different of 0 in case of failure

georgesittas:
Fixes #4099


izeigerman (approved):
Thanks!

PR #4266, opened 2025-04-28
Fix: Avoid concurrent dialect patching in model testing

VaggelisD:
To [freeze time](https://sqlmesh.readthedocs.io/en/latest/concepts/tests/#freezing-time) through `execution_time` in unit tests we essentially patch the SQLGlot dialects globally to generate any `CURRENT_<date/time>()` reference as a cast at the specified time & type.

This works fine if tests are run sequentially but runs into race conditions in concurrent execution. This PR:

- Replaces the global patching for SQL models by instead transforming the corresponding AST nodes before generation

- Locks the global patching for Python models using the runner's `lock`; This is required because Python models can execute any SQL statement, so the patching is indeed necessary to cover that globally.

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -530,9 +530,7 @@ class ModelTest(unittest.TestCase):
         import time_machine
 
         lock_ctx: AbstractContextManager = (
-            self.CONCURRENT_RENDER_LOCK
-            if (self.concurrency and self._execution_time)
-            else nullcontext()
+            self.CONCURRENT_RENDER_LOCK if self.concurrency else nullcontext()
         )
         time_ctx: AbstractContextManager = nullcontext()
         dialect_patch_ctx: AbstractContextManager = nullcontext()
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import datetime
+import threading
 import typing as t
 import unittest
 from collections import Counter
-from contextlib import AbstractContextManager, nullcontext
+from contextlib import nullcontext, contextmanager, AbstractContextManager
 from itertools import chain
 from pathlib import Path
 from unittest.mock import patch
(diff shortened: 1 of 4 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -57,6 +58,7 @@ def __init__(
         path: Path | None = None,
         preserve_fixtures: bool = False,
         default_catalog: str | None = None,
+        lock: t.Optional[threading.Lock] = None,
```
izeigerman, line 61: I don't like the idea of passing concurrency primitives from outside. This is asking for trouble.

sqlmesh/core/test/runner.py
```diff
@@ -132,6 +132,7 @@ def _run_single_test(
             path=metadata.path,
             default_catalog=default_catalog,
             preserve_fixtures=preserve_fixtures,
+            lock=lock if num_workers > 1 else None,
```
izeigerman, line 135: And we're reusing a lock that had a completely different purpose originally. Double red flag

sqlmesh/core/test/definition.py
```diff
@@ -512,10 +519,34 @@ def _normalize_column_name(self, name: str) -> str:
...
+        if self._execution_time:
+            time_ctx = time_machine.travel(self._execution_time, tick=False)
+            dialect_patch_ctx = patch.dict(
+                self._test_adapter_dialect.generator_class.TRANSFORMS, self._transforms
+            )
+
+        with lock_ctx, time_ctx, dialect_patch_ctx:
+            yield
+
+    def _execute(self, query: exp.Query | str) -> pd.DataFrame:
```
izeigerman, line 547: should we make it just `str` to make sure someone doesn't pass an expression by accident in future?

izeigerman (approved):

sqlmesh/core/test/definition.py
VaggelisD: We might call `_execute` on other places such as when constructing the expected df, so I think allowing both is fine.

PR #4270, opened 2025-04-28
Fix: Add custom exception for macro evaluation errors

themisvaltinos:
Handle MacroEval errors similar to pythonmodel and signal evaluation errors that we want to catch.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -228,7 +228,9 @@ class BaseExpressionRenderer:
             try:
                 macro_evaluator.evaluate(definition)
             except Exception as ex:
-                raise_config_error(f"Failed to evaluate macro '{definition}'. {ex}", self._path)
+                raise_config_error(
+                    f"Failed to evaluate macro '{definition}'.\n\n{ex}\n", self._path
+                )
 
         resolved_expressions: t.List[t.Optional[exp.Expression]] = []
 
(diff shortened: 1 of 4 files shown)


sqlmesh/core/renderer.py
```diff
@@ -228,7 +228,9 @@ def _resolve_table(table: str | exp.Table) -> str:
             try:
                 macro_evaluator.evaluate(definition)
             except Exception as ex:
-                raise_config_error(f"Failed to evaluate macro '{definition}'. {ex}", self._path)
+                raise_config_error(
+                    f"Failed to evaluate macro '{definition}'.\n\n{ex}\n", self._path
```
izeigerman, line 232: Looks like too many empty lines. Can you share an example of how this is presented in CLI?

izeigerman (approved):
Looks good overall but I'm somewhat concerned about the number of empty lines in the CLI output

sqlmesh/core/renderer.py
themisvaltinos: yes they were meant to add a bit of spacing with an empty line between the error details and the message, so this is how it looked:

newline

happy to remove them if you'd prefer without the extra empty lines in between so that they are like this:

noemptyline

georgesittas: Imo the spacing looks great 👍

georgesittas: Imo the spacing looks great 👍

PR #4288, opened 2025-04-30
feat(vscode): better errors for missing sqlmesh_lsp

benfdking:
Handle the cases where 

- `sqlmesh_lsp` is not present 
- `pygls` and `lsprotocol` are not present, recommend installing `sqlmesh[lsp]`

Screenshot 2025-04-30 at 22 47 06

### DIFF
--- a/vscode/extension/src/commands/format.ts
+++ b/vscode/extension/src/commands/format.ts
@@ -3,22 +3,35 @@ import { execSync } from 'child_process'
 import { sqlmesh_exec } from '../utilities/sqlmesh/sqlmesh'
 import { err, isErr, ok, Result } from '../utilities/functional/result'
 import * as vscode from 'vscode'
-import { ErrorType, handleNotSginedInError } from '../utilities/errors'
+import {
+  ErrorType,
+  handleNotSginedInError,
+  handleSqlmeshLspNotFoundError,
+  handleSqlmeshLspDependenciesMissingError,
+} from '../utilities/errors'
 import { AuthenticationProviderTobikoCloud } from '../auth/auth'
 
 export const format =
-  (authProvider: AuthenticationProviderTobikoCloud) => async () => {
+  (authProvider: AuthenticationProviderTobikoCloud) =>
+  async (): Promise<void> => {
     traceLog('Calling format')
     const out = await internalFormat()
     if (isErr(out)) {
-      if (out.error.type === 'not_signed_in') {
-        await handleNotSginedInError(authProvider)
-        return
-      } else {
-        vscode.window.showErrorMessage(
-          `Project format failed: ${out.error.message}`,
-        )
-        return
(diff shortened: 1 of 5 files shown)


vscode/extension/src/commands/format.ts
```diff
@@ -3,22 +3,35 @@ import { execSync } from 'child_process'
...
-        return
-      } else {
-        vscode.window.showErrorMessage(
-          `Project format failed: ${out.error.message}`,
-        )
-        return
+      switch (out.error.type) {
+        case 'not_signed_in':
+          await handleNotSginedInError(authProvider)
+          return
```
mykalmax, line 23: Nit: `break` instead of return

vscode/extension/src/utilities/errors.ts
```diff
@@ -26,3 +36,40 @@ export const handleNotSginedInError = async (
...
+  if (error.is_tobiko_cloud) {
+    await window.showErrorMessage(
+      'LSP dependencies missing, make sure to include `lsp` in the `extras` section of your `tcloud.yaml` file.',
+    )
+  } else {
+    const install = await window.showErrorMessage(
+      'LSP dependencies missing, make sure to install `sqlmesh[lsp]`.',
+      'Install',
+    )
+    if (install === 'Install') {
```
mykalmax, line 66: should that `'Install'` be `const` ?

vscode/extension/src/commands/format.ts
```diff
@@ -3,22 +3,35 @@ import { execSync } from 'child_process'
...
-      if (out.error.type === 'not_signed_in') {
-        await handleNotSginedInError(authProvider)
-        return
-      } else {
-        vscode.window.showErrorMessage(
-          `Project format failed: ${out.error.message}`,
-        )
-        return
+      switch (out.error.type) {
+        case 'not_signed_in':
```
mykalmax, line 21: should we move `case`  options to be `enum`?

mykalmax (approved):

izeigerman (approved):

vscode/extension/src/commands/format.ts
benfdking: It's not the prettiest code, but in this case,e I explicitly want to return `rather` than `break`.

vscode/extension/src/utilities/errors.ts
benfdking: It could be, I think, for two instances right next to each other, I am slightly indifferent 🤷

vscode/extension/src/commands/format.ts
benfdking: I really think we don't, and my thinking behind it is that I want to be using discriminated unions and enums don't really neatly allow that. It's fully type safe and I have set linting up so that we have to check every case. To me enums or consts would just be indirection.

PR #4293, opened 2025-05-01
Fix of render definition method for virtual update statements

blecourt-private:
Casting virtual update expressions as VirtualUpdateStatement

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1319,12 +1319,14 @@ class SqlModel(_Model):
             result.extend(self.render_pre_statements())
             result.append(self.render_query() or self.query)
             result.extend(self.render_post_statements())
-            result.append(d.VirtualUpdateStatement(expressions=self.render_on_virtual_update()))
+            if self.render_on_virtual_update():
+                result.append(d.VirtualUpdateStatement(expressions=self.render_on_virtual_update()))
         else:
             result.extend(self.pre_statements)
             result.append(self.query)
             result.extend(self.post_statements)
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1319,12 +1319,14 @@ def render_definition(
             result.extend(self.render_pre_statements())
             result.append(self.render_query() or self.query)
             result.extend(self.render_post_statements())
-            result.extend(self.render_on_virtual_update())
+            if self.render_on_virtual_update():
```
tobymao, line 1322: we shouldn't call render twice, can you store the variable?

if virtual_update := self.render_on_virtual_update()

tests/core/test_model.py
```diff
@@ -1642,6 +1642,43 @@ def test_render_definition_partitioned_by():
...
+            f"""
+        MODEL (
+            name db.table,
+            kind FULL
+        );
+
+        select 1 as a;
+
+        ON_VIRTUAL_UPDATE_BEGIN;
+        GRANT SELECT ON VIEW @this_model TO ROLE role_name
```
themisvaltinos, line 1658: this is failing because it is missing a semicolon after role_name: `;`

sqlmesh/core/model/definition.py
themisvaltinos: since you assigned the variable you can use it now so have:
`result.append(d.VirtualUpdateStatement(expressions=virtual_update))` in line 1323

themisvaltinos (approved):

Issue #4294, opened 2025-05-01
Inconsistent behavior of imports inside macros


xardasos:
There is a strange inconsistent behavior of imports inside macros. These imports work if a given import is not repeated in other macros, but if there are 2 different macros trying to import the same enum, suddenly `sqlmesh plan` throws an error similar to the one below:
```
Error: Failed to load model definition at '/app/models/full_model.sql'.
cannot import name '_generate_next_value_' from 'enum' (/usr/local/lib/python3.12/enum.py)
```

I created a reproducer [project](https://github.com/xardasos/sqlmesh-imports-problem)

Commenting out the usage of EITHER `macro1` or `macro2` (both macros use the same import) in the [affected model](https://github.com/xardasos/sqlmesh-imports-problem/blob/main/sqlmesh-example/models/full_model.sql) makes the import error disappear.

Affected sqlmesh versions: >= 0.153.0, it seems that this problem doesn't appear in 0.151.2, perhaps it was introduced in this MR: https://github.com/TobikoData/sqlmesh/pull/3762

It looks like the root cause is that when we walk a given object for the first time (name_missing_from_env=true) then we don't walk recursively into external imports https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/metaprogramming.py#L337. However if the same object is being walked a second time (name_missing_from_env=false), then we do walk into the external imports (that's why we end up with `_generate_next_value_` in the serialized env)

PR #4295, opened 2025-05-01
Fix inconsistent imports behavior inside macros

xardasos:
Addresses issue: https://github.com/TobikoData/sqlmesh/issues/4294

### DIFF
--- a/sqlmesh/utils/metaprogramming.py
+++ b/sqlmesh/utils/metaprogramming.py
@@ -338,42 +338,42 @@ def build_env(
             ):
                 env[name] = (obj, is_metadata)
                 return
+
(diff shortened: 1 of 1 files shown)


georgesittas (approved):
> Hi @georgesittas, could you please take a look at this PR? I admit I haven't fully analyzed possible consequences of changing the walk logic like this so please be careful while reviewing :)

So my understanding is that we would previously walk objects, even though they were already in the environment, which seems wasteful in retrospect.

Additionally, falling through to the section after the branch seems questionable, because we would only end up overwriting the previous `env[name]` entry, right before returning from `walk`.

Given the above, this looks reasonable to me. Thanks for investigating and tackling this, great work!

PR #4298, opened 2025-05-01
Feat: Make this_env and views available in macros of before after all

themisvaltinos:
This update adds the `this_env` variable in the macros for before all after all statements as well as `views` which provides all the views of the virtual layer to be able to easily for example grant rights on the views:
```python
@macro()
def grant_select_privileges(evaluator):
    if evaluator.views:
        return [
            f"GRANT SELECT ON VIEW {view_name} /* sqlglot.meta replace=false */ TO ROLE admin_role;"
            for view_name in evaluator.views
        ]
```

By including the comment `/* sqlglot.meta replace=false */`, you ensure that the evaluator does not replace the view name with the physical table name.

### DIFF
--- a/docs/concepts/macros/macro_variables.md
+++ b/docs/concepts/macros/macro_variables.md
@@ -141,3 +141,4 @@ The following variables are also available in [`before_all` and `after_all` stat
 
 * @this_env - A string value containing the name of the current [environment](../environments.md).
 * @schemas - A list of the schema names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
+* @views - A list of the views names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
--- a/docs/concepts/macros/macro_variables.md
+++ b/docs/concepts/macros/macro_variables.md
@@ -141,4 +141,4 @@ The following variables are also available in [`before_all` and `after_all` stat
 
 * @this_env - A string value containing the name of the current [environment](../environments.md).
 * @schemas - A list of the schema names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
-* @views - A list of the views names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
+* @views - A list of the view names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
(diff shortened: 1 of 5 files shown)


docs/concepts/macros/macro_variables.md
```diff
@@ -140,4 +140,5 @@ SQLMesh provides additional predefined variables used to modify model behavior b
 The following variables are also available in [`before_all` and `after_all` statements](../../guides/configuration.md#before_all-and-after_all-statements), as well as in macros invoked within them.
 
 * @this_env - A string value containing the name of the current [environment](../environments.md).
-* @schemas - A list of the schema names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
\ No newline at end of file
+* @schemas - A list of the schema names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
+* @views - A list of the views names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
```
georgesittas, line 144:
```suggestion
* @views - A list of the view names of the [virtual layer](../../concepts/glossary.md#virtual-layer) of the current environment.
```

sqlmesh/core/macros.py
```diff
@@ -491,6 +491,26 @@ def gateway(self) -> t.Optional[str]:
...
 
+    @property
+    def snapshots(self) -> t.Dict[str, Snapshot]:
+        """Returns the snapshots if available."""
+        return self._snapshots
+
+    @property
+    def this_env(self) -> t.Optional[str]:
+        """Returns the name of the current environment in before after all."""
+        return self.locals.get("this_env")
```
georgesittas, line 502: Should this raise if the environment is not available? Same for the `schemas` and `views` properties.

sqlmesh/core/macros.py
```diff
@@ -491,6 +491,32 @@ def gateway(self) -> t.Optional[str]:
...
+    @property
+    def snapshots(self) -> t.Dict[str, Snapshot]:
+        """Returns the snapshots if available."""
+        return self._snapshots
+
+    @property
+    def this_env(self) -> str:
+        """Returns the name of the current environment in before after all."""
+        if "this_env" not in self.locals:
+            raise SQLMeshError("Environment name is not available in this context")
```
georgesittas, line 503: Should we instead explain how these are only available in the before/after all stages instead? This message is not particularly helpful for debugging.

georgesittas (approved):
One final suggestion, lgtm otherwise.

PR #4300, opened 2025-05-01
feat(vscode): add completion provider

benfdking:
- adds keywords from tokenizer
- adds models in project minus the model that the file constructs.

### DIFF
--- a/sqlmesh/lsp/completions.py
+++ b/sqlmesh/lsp/completions.py
@@ -1,0 +1,75 @@
+from functools import lru_cache
+from sqlglot import Dialect, Tokenizer
+from sqlmesh.lsp.custom import AllModelsResponse
+import typing as t
+from sqlmesh.lsp.context import LSPContext
+
+
+def get_sql_completions(context: t.Optional[LSPContext], file_uri: str) -> AllModelsResponse:
+    """
+    Return a list of completions for a given file.
+    """
+    return AllModelsResponse(
+        models=list(get_models(context, file_uri)),
+        keywords=list(get_keywords(context, file_uri)),
+    )
+
+
+def get_models(context: t.Optional[LSPContext], file_uri: t.Optional[str]) -> t.Set[str]:
+    """
+    Return a list of models for a given file.
+
+    If there is no context, return an empty list.
+    If there is a context, return a list of all models bar the ones the file itself defines.
+    """
+    if context is None:
+        return set()
+    all_models = set(model for models in context.map.values() for model in models)
+    if file_uri is not None:
+        models_file_refers_to = context.map[file_uri]
+        for model in models_file_refers_to:
+            all_models.discard(model)
+    return all_models
+
+
+def get_keywords(context: t.Optional[LSPContext], file_uri: t.Optional[str]) -> t.Set[str]:
+    """
+    Return a list of sql keywords for a given file.
+    If no context is provided, return ANSI SQL keywords.
+
+    If a context is provided but no file_uri is provided, returns the keywords
+    for the default dialect of the context.
+
+    If both a context and a file_uri are provided, returns the keywords
+    for the dialect of the model that the file belongs to.
+    """
+    if file_uri is not None and context is not None:
+        models = context.map[file_uri]
+        if models:
+            model = models[0]
+            if context.context.get_model(model) is not None:
+                return get_keywords_from_tokenizer(context.context.get_model(model).dialect)
+    if context is not None:
+        return get_keywords_from_tokenizer(context.context.default_dialect)
+    return get_keywords_from_tokenizer(None)
+
+
+@lru_cache()
+def get_keywords_from_tokenizer(dialect: t.Optional[str] = None) -> t.Set[str]:
+    """
+    Return a list of sql keywords for a given dialect. This is separate from
+    the direct use of Tokenizer.KEYWORDS.keys() because that returns a set of
+    keywords that are expanded, e.g. "ORDER BY" -> ["ORDER", "BY"].
+    """
+    tokenizer = Tokenizer
+    if dialect is not None:
+        try:
+            tokenizer = Dialect.get_or_raise(dialect).tokenizer_class
+        except Exception:
+            pass
+
+    expanded_keywords = set()
+    for keyword in tokenizer.KEYWORDS.keys():
+        parts = keyword.split(" ")
+        expanded_keywords.update(parts)
+    return expanded_keywords
--- a/sqlmesh/lsp/completions.py
+++ b/sqlmesh/lsp/completions.py
@@ -47,8 +47,10 @@ def get_keywords(context: t.Optional[LSPContext], file_uri: t.Optional[str]) -> 
         models = context.map[file_uri]
         if models:
             model = models[0]
-            if context.context.get_model(model) is not None:
-                return get_keywords_from_tokenizer(context.context.get_model(model).dialect)
+            model_from_context = context.context.get_model(model)
+            if model_from_context is not None:
+                if model_from_context.dialect:
+                    return get_keywords_from_tokenizer(model_from_context.dialect)
     if context is not None:
         return get_keywords_from_tokenizer(context.context.default_dialect)
     return get_keywords_from_tokenizer(None)
--- a/sqlmesh/lsp/custom.py
+++ b/sqlmesh/lsp/custom.py
@@ -1,0 +1,22 @@
+from lsprotocol import types
+import typing as t
+from sqlmesh.utils.pydantic import PydanticModel
+
+ALL_MODELS_FEATURE = "sqlmesh/all_models"
+
+
+class AllModelsRequest(PydanticModel):
+    """
+    Request to get all the models that are in the current project.
+    """
+
+    textDocument: types.TextDocumentIdentifier
+
+
+class AllModelsResponse(PydanticModel):
+    """
+    Response to get all the models that are in the current project.
+    """
+
+    models: t.List[str]
+    keywords: t.List[str]
--- a/sqlmesh/lsp/main.py
+++ b/sqlmesh/lsp/main.py
@@ -11,7 +11,9 @@ from pygls.server import LanguageServer
 from sqlmesh._version import __version__
 from sqlmesh.core.context import Context
 from sqlmesh.core.linter.definition import AnnotatedRuleViolation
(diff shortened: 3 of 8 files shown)


sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+    tables = list(model.query.find_all(exp.Table))
+    if len(tables) == 0:
+        return []
+    
+    references = []
+    for table in tables:
+        depends_on = model.depends_on
+
+        # Normalize the table reference
+        reference_name = table.this.this if table.db is None else f"{table.db}.{table.this.this}"
```
izeigerman, line 60: I think this doesn't work if the model name was quoted explicitly by the user. I think you want to do `table.sql(dialect=model.dialect)` instead.

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+        normalized_reference_name = normalize_model_name(reference_name, default_catalog=lint_context.context.default_catalog)
```
izeigerman, line 61: You need to pass `dialect=model.dialect` here.

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+    # Ensure the path is a sql model
+    if not document_uri.endswith(".sql"):
+        return []
+    
+    # Get the model 
+    models = lint_context.map[document_uri]
+    if models is None: 
+        return []
+    if len(models) == 0:
+        return []
```
VaggelisD, line 42: You can summarize both as `if not models`

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+    model_name = models[0]
+    model = lint_context.context.get_model(model_or_snapshot=model_name, raise_if_missing=False)
+    if model is None: 
+        return []
+    if not isinstance(model, SqlModel):
+        return []
```
VaggelisD, line 48: You can summarize both as `if not isinstance(model, SqlModel)`, if it is `None` it will also fail the check

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+    
+    # Find all possible references
+    tables = list(model.query.find_all(exp.Table))
+    if len(tables) == 0:
```
VaggelisD, line 52: [Nit] `if not tables` is more pythonic in general

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+        # Get the referenced model uri
+        referenced_model = lint_context.context.get_model(model_or_snapshot=normalized_reference_name, raise_if_missing=False)
+        if referenced_model is None:
+            continue
+        # Get the model uri
+        referenced_model_path = referenced_model._path
+        if referenced_model_path is None:
+            continue
```
VaggelisD, line 73: These can be `None` if there's no references, right? Also do note that we can have `exp.Table` objects that are not actual tables/models, for instance all of the createables in SQLGlot (e.g `CREATE INDEX foo`) wrap the identifiers in `exp.Table` objects.

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+        # Fully qualify the path in case
+        path = Path.resolve(Path(referenced_model_path))
+        referenced_model_path = f"file://{path}"
+        # Get the path to the file containing the reference
+        # file_path = document_uri.removeprefix("file://")
+        # read_file = open(file_path, "r").readlines() # Reading the file here is not needed for range calculation
```
VaggelisD, line 79: Dead code or remember to uncomment

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+        
+        # Extract metadata for positioning
+        this_meta = table.this.meta
+        this_start_line_0 = this_meta['line'] - 1  # Convert to 0-indexed
+        this_start_col_0 = this_meta['col'] - 1   # Convert to 0-indexed
+        
+        # End position: Based on the start of 'this' token + its length
+        end_char_0 = this_start_col_0 + len(table.this.this)
+        end_pos = Position(line=this_start_line_0, character=end_char_0)
```
VaggelisD, line 88: I'd create a helper that does `Position.from_identifier(...)` to abstract away the meta stuff if these become repeated

sqlmesh/lsp/reference.py
```diff
@@ -0,0 +1,114 @@
...
+            db_start_col_0 = db_meta['col'] - 1   # Convert to 0-indexed
+            start_pos = Position(line=db_start_line_0, character=db_start_col_0)
+        
+        # If there's a catalog qualifier, adjust the start position further
+        catalog = table.args.get('catalog')
+        if catalog is not None:
+            catalog_meta = catalog.meta
+            catalog_start_line_0 = catalog_meta['line'] - 1 # Convert to 0-indexed
+            catalog_start_col_0 = catalog_meta['col'] - 1  # Convert to 0-indexed
+            start_pos = Position(line=catalog_start_line_0, character=catalog_start_col_0)
```
VaggelisD, line 107: If you have an table of the form `catalog.db.table` then you'll go through the first branch (db) and then throw that away by visiting the second branch (catalog), since you don't need the db/catalog specific parts you can unify these both into:

```Python3
catalog_or_db = table.args.get('catalog') or table.args.get('db')

if catalog_or_db = 
  ....
  start_pos = Position(line=catalog_or_db_start_line_0, character=catalog_or_db_start_col_0)
```


```

File: sqlmesh/lsp/completions.py [hunk 10]
```diff
@@ -0,0 +1,58 @@
...
+@lru_cache(maxsize=1)
+def get_keywords_from_tokenizer() -> t.Set[str]:
+    keywords = list(Tokenizer.KEYWORDS.keys())
+    expanded_keywords = []
+
+    for keyword in keywords:
+        parts = keyword.split(" ")
+        expanded_keywords.extend(parts)
+
+    return set(expanded_keywords)
```
VaggelisD, line 58: I wonder whether both of these functions can be simplified into a module variable, e.g:

```Python3
from itertools import chain

EXPANDED_TOKENIZER_KEYWORDS = set(chain.from_iterable(keyword.split(" ") for keyword in Tokenizer.KEYWORDS.keys()))
```

sqlmesh/lsp/completions.py
```diff
@@ -0,0 +1,58 @@
...
+def get_models(context: t.Optional[LSPContext], file_uri: t.Optional[str]) -> t.Set[str]:
+    """
+    Return a list of models for a given file.
+
+    If there is no context, return an empty list.
+    If there is a context, return a list of all models bar the ones the file itself defines.
+    """
+    if context is None:
+        return set()
+    all_models = set(model for models in context.map.values() for model in models)
```
VaggelisD, line 27: Sorry for asking this again but when can we not have `LSPContext` and/or `file_uri`s?

VaggelisD (commented):
Broader question, not sure if I'm missing it in the code but is there context associated with these completion suggestions? 

For example, when the user has finished typing `FROM`, would they get the `models` list popping up or the entire universe of keywords?

sqlmesh/lsp/completions.py
izeigerman: +1 doesn't seem like it's used

sqlmesh/lsp/completions.py
```diff
@@ -0,0 +1,58 @@
...
+    Return a list of sql keywords for a given file.
+    If no context is provided, return ANSI SQL keywords.
+
+    If a context is provided but no file_uri is provided, returns the keywords
+    for the default dialect of the context.
+
+    If both a context and a file_uri are provided, returns the keywords
+    for the dialect of the model that the file belongs to.
+    """
+    return get_keywords_from_tokenizer()
```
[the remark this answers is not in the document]
izeigerman: I guess this is because we use `@lru_cache` ad we want to cache without arguments so that we only have a single instance vs an instance per `file_uri`

sqlmesh/lsp/completions.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@lru_cache(maxsize=1)
+def get_keywords_from_tokenizer() -> t.Set[str]:
+    keywords = list(Tokenizer.KEYWORDS.keys())
```
izeigerman, line 51: Aren't these dialect specific?

izeigerman: Shouldn't we use the default dialect from the context?

sqlmesh/lsp/custom.py
```diff
@@ -0,0 +1,22 @@
...
+
+ALL_MODELS_FEATURE = "sqlmesh/all_models"
+
+
+class AllModelsRequest(PydanticModel):
+    """
+    Request to get all the models that are in the current project.
+    """
+
+    textDocument: types.TextDocumentIdentifier
```
izeigerman, line 13: what's up with the naming style? :)

sqlmesh/lsp/completions.py
benfdking: You can have 
- No context when a project fails to load
- No file_uri when the file hasn't been saved yet

I haven't used it yet because the rest of the LSP relies very much on the project parsing at that moment in time, but I want to relax that restriction.

sqlmesh/lsp/custom.py
benfdking: This is the default naming style for lsp messages. 

Honestly for me it was just a bit of a coin flip that AI made a decision for when deciding on the 

In TS you do this camel, in Python you do snake, 🤷

sqlmesh/lsp/completions.py
benfdking: I wanted the function because I want to do this lazily because we don't know upfront the dialect.

sqlmesh/lsp/completions.py
benfdking: Yes sorry my bad.

sqlmesh/lsp/completions.py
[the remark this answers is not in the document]
benfdking: This is probably some slop that I didn't finish on my behalf...

PR #4301, opened 2025-05-02
Enable reload of modified models

blecourt-private:
Suggestion for loading models from definitions instead of files which would be helpful in [custom loaders](https://sqlmesh.readthedocs.io/en/stable/guides/customizing_sqlmesh/#modify-every-model).

Without reloading models after modifying them, custom loaders are more or less limited to using pure SQL in added or modified expressions. 

This MR showcases a way to load models from their rendered definitions. The PR currently only adds support for sql models but I imagine support for python models and seed models can be added in a similar fashion.

I suppose support for python models, seed models plus a bunch of tests need to be added to make this PR proper but before proceeding with this I would appreciate some feedback. It wouldn't make sense to put more work into this if I'm completely on the wrong track.

### DIFF
--- a/sqlmesh/core/loader.py
+++ b/sqlmesh/core/loader.py
@@ -601,7 +601,7 @@ class SqlMeshLoader(Loader):
                 out[model.fqn] = model
 
         return out
-        
+
     def load_materializations(self) -> None:
         with sys_path(self.config_path):
             self._load_materializations()


izeigerman:
"I've read the description and the PR a few times, but I'm still a bit unclear on what this is fixing exactly. Could you share steps to reproduce the issue—or even better, a test case that fails without this fix?

blecourt-private:
@sungchun12 this is the PR I mentioned in our meeting

izeigerman:
hey @blecourt-private ! Apologies about the delay. Thanks a lot for providing this context. I guess I don't fully follow how this update addresses the issue at hand. Perhaps you can add a test case which demonstrates this?

blecourt-private:
@izeigerman I'm planning to do so but I guess we're all busy 😉 . As soon as I have a few minutes to spare!

PR #4303, opened 2025-05-02
chore: make _path optional and more representative

benfdking:
At the moment, the path type is slightly confusing. Models can exist without a Path. Rather than mark the `_path` property as `None` we call `Path()` which is equivalent to calling `Path(".")` which is not particularly indicative it just returns your path to `cwd`. Optional is a much better representation.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -943,11 +943,8 @@ class GenericContext(BaseContext, t.Generic[C]):
     def config_for_node(self, node: str | Model | Audit) -> Config:
         if isinstance(node, str):
             path = self.get_snapshot(node, raise_if_missing=True).node._path
-            if path is None:
-                return self.config
-            config = self.config_for_path(path)
-            return config[0]
-        path = node._path
+        else:
+            path = node._path
         if path is None:
             return self.config
         return self.config_for_path(path)[0]  # type: ignore
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -942,10 +942,15 @@ class GenericContext(BaseContext, t.Generic[C]):
 
     def config_for_node(self, node: str | Model | Audit) -> Config:
         if isinstance(node, str):
-            return self.config_for_path(self.get_snapshot(node, raise_if_missing=True).node._path)[
-                0
-            ]  # type: ignore
-        return self.config_for_path(node._path)[0]  # type: ignore
+            path = self.get_snapshot(node, raise_if_missing=True).node._path
+            if path is None:
+                return self.config
+            config = self.config_for_path(path)
+            return config[0]
+        path = node._path
+        if path is None:
+            return self.config
+        return self.config_for_path(path)[0]  # type: ignore
 
     @property
     def models(self) -> MappingProxyType[str, Model]:
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1646,6 +1646,8 @@ class SeedModel(_Model):
     def seed_path(self) -> Path:
         seed_path = Path(self.kind.path)
         if not seed_path.is_absolute():
+            if self._path is None:
+                raise SQLMeshError(f"Seed model '{self.name}' has no path")
             return self._path.parent / seed_path
         return seed_path
 
@@ -2020,7 +2022,7 @@ def load_sql_based_model(
     expressions: t.List[exp.Expression],
     *,
     defaults: t.Optional[t.Dict[str, t.Any]] = None,
-    path: Path = Path(),
+    path: t.Optional[Path] = None,
     module_path: Path = Path(),
     time_column_format: str = c.DEFAULT_TIME_COLUMN_FORMAT,
     macros: t.Optional[MacroRegistry] = None,
@@ -2171,6 +2173,8 @@ Learn more at https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview
     # The name of the model will be inferred from its path relative to `models/`, if it's not explicitly specified
     name = meta_fields.pop("name", "")
     if not name and infer_names:
+        if path is None:
+            raise ValueError("Model must have a name", path)
         name = get_model_name(path)
 
     if not name:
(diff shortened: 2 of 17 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -2249,7 +2253,12 @@ def create_seed_model(
         seed_path = module_path.joinpath(*subdirs)
         seed_kind.path = str(seed_path)
     elif not seed_path.is_absolute():
-        seed_path = path / seed_path if path.is_dir() else path.parent / seed_path
+        if path is None:
+            seed_path = seed_path
```
georgesittas, line 2276: Typo?

sqlmesh/core/node.py
```diff
@@ -196,7 +196,7 @@ class _Node(PydanticModel):
     interval_unit_: t.Optional[IntervalUnit] = Field(alias="interval_unit", default=None)
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
-    _path: Path = Path()
+    _path: t.Optional[Path] = Path()
```
georgesittas: Is this still initialized to the cwd `Path()` on purpose? Or just an oversight in the refactor?

sqlmesh/lsp/reference.py
```diff
@@ -235,8 +238,10 @@ def get_model_definitions_for_a_path(
                 if referenced_model is None:
                     continue
                 referenced_model_path = referenced_model._path
+                if referenced_model_path is None:
+                    continue
                 # Check whether the path exists
-                if not referenced_model_path.is_file():
+                if referenced_model_path is None or not referenced_model_path.is_file():
```
georgesittas: The first term in this condition is redundant due to L241-242.

web/server/api/endpoints/models.py
```diff
@@ -121,11 +121,18 @@ def serialize_model(context: Context, model: Model, render_query: bool = False)
...
+            detail=f"Model {model.name} does not have a file path.",
+        )
+
     return models.Model(
         name=model.name,
         fqn=model.fqn,
-        path=str(model._path.absolute().relative_to(context.path).as_posix()),
-        full_path=str(model._path.absolute().as_posix()),
+        path=str(path.absolute().relative_to(context.path).as_posix()),
+        full_path=str(path.absolute().as_posix()),
```
benfdking: These should just be none

sqlmesh/core/model/definition.py
```diff
@@ -2171,6 +2173,8 @@ def load_sql_based_model(
     # The name of the model will be inferred from its path relative to `models/`, if it's not explicitly specified
     name = meta_fields.pop("name", "")
     if not name and infer_names:
+        if path is None:
+            raise ValueError("Model must have a name", path)
```
themisvaltinos: why is the path included, given that it's None. the string message should be enough

sqlmesh/dbt/converter/convert.py
```diff
@@ -207,6 +207,8 @@ def _convert_models(
 
         if model.kind.is_seed:
             # this will produce the original seed file, eg "items.csv"
+            if model._path is None:
+                raise ValueError(f"Unhandled model path: {model._path}")
```
themisvaltinos: again in these I'm confused about the inclusion of model._path in the error message when it's none

web/server/api/endpoints/models.py
```diff
@@ -121,11 +121,18 @@ def serialize_model(context: Context, model: Model, render_query: bool = False)
...
+            status_code=HTTP_404_NOT_FOUND,
+            detail=f"Model {model.name} does not have a file path.",
+        )
+
     return models.Model(
         name=model.name,
         fqn=model.fqn,
-        path=str(model._path.absolute().relative_to(context.path).as_posix()),
-        full_path=str(model._path.absolute().as_posix()),
+        path=str(path.absolute().relative_to(context.path).as_posix()) if path else None,
```
themisvaltinos, line 128: in line 125 you check for none and raise so here there is no need for a check as the path should exist

sqlmesh/core/context.py
```diff
@@ -942,10 +942,15 @@ def config_for_path(self, path: Path) -> t.Tuple[Config, Path]:
...
-        return self.config_for_path(node._path)[0]  # type: ignore
+            path = self.get_snapshot(node, raise_if_missing=True).node._path
+            if path is None:
+                return self.config
+            config = self.config_for_path(path)
+            return config[0]
+        path = node._path
+        if path is None:
+            return self.config
+        return self.config_for_path(path)[0]  # type: ignore
```
georgesittas: Nit: I'd suggest consolidating these two branches into one. If it's a string, get the node and let it fall through.

sqlmesh/dbt/converter/convert.py
```diff
@@ -219,12 +221,18 @@ def _convert_models(
             assert isinstance(model.kind, SeedKind)
             model.kind.path = str(Path("../seeds", seed_filename))
         else:
+            if model._path is None:
+                raise ValueError(f"Unhandled model path: {model._path}")
```
georgesittas: Should this also contain the model's name? This just interpolates `None` into the message.

sqlmesh/dbt/converter/convert.py
```diff
@@ -219,12 +221,18 @@ def _convert_models(
...
             if input_paths.models in model._path.parents:
                 model_filename = model._path.relative_to(input_paths.models)
             elif input_paths.snapshots in model._path.parents:
                 # /base/path/snapshots/foo.sql -> /output/path/models/dbt_snapshots/foo.sql
                 model_filename = "dbt_snapshots" / model._path.relative_to(input_paths.snapshots)
+            elif input_paths.tests in model._path.parents:
+                # /base/path/tests/foo.sql -> /output/path/audits/foo.sql
+                model_filename = "dbt_tests" / model._path.relative_to(input_paths.tests)
```
georgesittas: What triggered the need to add this branch?

sqlmesh/dbt/converter/convert.py
```diff
@@ -290,6 +298,8 @@ def _convert_standalone_audits(
 
         audit_definition_string = ";\n".join(stringified)
 
+        if audit._path is None:
+            continue
```
georgesittas, line 298: Should we warn? Seems like we'd otherwise drop the audit silently.

georgesittas (approved):
Seems good, just some minor comments from me.

PR #4306, opened 2025-05-02
Fix: use MarkdownConsole in non-interactive contexts

treysp:
Some terminals do not support unicode emoji like those used in the standard CLI output (e.g., running SQLMesh with Powershell.exe in Github Actions). Using the `MarkdownConsole` instead of standard `TerminalConsole` addresses this problem.

There is not a universal way to determine whether a given terminal supports Unicode, so we instead assume that unicode is not needed in any non-interactive context (typically a CI platform).

This PR:
- Automatically detects if we're in a non-interactive context and uses `MarkdownConsole` if so
  - Non-interactive contexts include CI (detected via env vars commonly set by CI systems) or anywhere stdout and stdin are not both tty
- Brings the MarkdownConsole output into alignment with recent CLI updates
- Allows the user to force terminal output of any runtime environment with the `SQLMESH_RUNTIME_ENVIRONMENT` env var
  - Accepted values: `terminal`, `databricks`, `google_colab`, `jupyter`, `debugger`, `non_interactive`, `ci`

Screenshot 2025-05-02 at 5 27 38 PM

### DIFF
--- a/sqlmesh/__init__.py
+++ b/sqlmesh/__init__.py
@@ -64,6 +64,19 @@ class RuntimeEnv(str, Enum):
 
         Unlike the rich implementation we try to split out by notebook type instead of treating it all as Jupyter.
         """
+        runtime_env_var = os.getenv("SQLMESH_RUNTIME_ENVIRONMENT")
+        if runtime_env_var:
+            runtime_env_var = runtime_env_var.lower().strip().replace(" ", "").replace("-", "_")
+            runtime_env_var = "non_interactive" if runtime_env_var == "ci" else runtime_env_var
+            runtime_env_var = "debugger" if "debug" in runtime_env_var else runtime_env_var
+            try:
+                return RuntimeEnv(runtime_env_var)
+            except ValueError:
+                valid_values = [f'"{member.value}"' for member in RuntimeEnv] + ['"ci"']
+                raise ValueError(
+                    f"Invalid SQLMESH_RUNTIME_ENVIRONMENT value: {runtime_env_var}. Must be one of {', '.join(valid_values)}."
+                )
+
         try:
             shell = get_ipython()  # type: ignore
             if os.getenv("DATABRICKS_RUNTIME_VERSION"):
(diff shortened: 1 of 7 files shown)


sqlmesh/core/console.py
```diff
@@ -1076,7 +1083,7 @@ def stop_promotion_progress(self, success: bool = True) -> None:
             self.promotion_progress.stop()
             self.promotion_progress = None
             if success:
-                self.log_success(f"\n{GREEN_CHECK_MARK} Virtual layer updated")
+                self.log_success(f"\n{self.CHECK_MARK} Virtual layer updated")
```
izeigerman, line 1086: Does it mean that non-interactive terminal will get a leading whitespace?

treysp: Yeah, that won't look good - will update so it's left-justified in non-interactive

sqlmesh/__init__.py
```diff
@@ -62,6 +64,19 @@ def get(cls) -> RuntimeEnv:
 
         Unlike the rich implementation we try to split out by notebook type instead of treating it all as Jupyter.
         """
+        runtime_env_var = os.getenv("SQLMESH_RUNTIME_ENVIRONMENT")
+        if runtime_env_var:
+            runtime_env_var = runtime_env_var.lower().strip().replace(" ", "").replace("-", "_")
```
izeigerman, line 69: what are these `replace()` calls and why do we need them?

sqlmesh/__init__.py
```diff
@@ -62,6 +64,19 @@ def get(cls) -> RuntimeEnv:
...
+            runtime_env_var = "non_interactive" if runtime_env_var == "ci" else runtime_env_var
+            runtime_env_var = "debugger" if "debug" in runtime_env_var else runtime_env_var
```
izeigerman, line 71: why do we have these special values?

sqlmesh/__init__.py
```diff
@@ -62,6 +64,19 @@ def get(cls) -> RuntimeEnv:
...
+            try:
+                return RuntimeEnv(runtime_env_var)
+            except ValueError:
+                valid_values = [f'"{member.value}"' for member in RuntimeEnv] + ['"ci"']
```
izeigerman, line 75: If `ci` is among supported values then we should just have it in the enum.

sqlmesh/__init__.py
```diff
@@ -75,6 +88,10 @@ def get(cls) -> RuntimeEnv:
 
         if debug_mode_enabled():
             return RuntimeEnv.DEBUGGER
+
+        if is_cicd_environment() or not is_interactive_environment():
```
izeigerman, line 92: Now that we have CI value should we assign it instead on `is_cicd_environment`? Do we need both values?

PR #4309, opened 2025-05-04
fix(vscode): support aliased references

benfdking:


sqlmesh/lsp/reference.py
```diff
@@ -55,7 +55,10 @@ def get_model_definitions_for_a_path(
         depends_on = model.depends_on
 
         # Normalize the table reference
-        reference_name = table.sql(dialect=model.dialect)
+        unaliased = table.copy()
+        if unaliased.args.get("alias") is not None:
+            unaliased.set("alias", None)
```
VaggelisD, line 60: There's all sorts of modifiers that can be attach to an `exp.Table`, such as joins, tablesamples etc. You'd need to drop all of them except `catalog`, `db` and `this` or leverage existing helpers such as `sqlmesh.core.dialect.find_tables` which should do that for you.

However, looking into the whole function again, I don't think you need to find (or clean up) the references on your own, isn't that provided by `model.depends_on`?

izeigerman: I guess the problem is that if there's an alias the `depends_on` check won't work correctly, though we still want to get a match. In all other cases you're right. `find_tables` could be a good suggestion.

benfdking: I've also realised ... it was failing with an exception and if say there are more examples where this can happen, I think it's better we wrap it in a try catch.

PR #4313, opened 2025-05-05
fix: pseudo-column type is BQ `TIMESTAMP` which maps to `TIMESTAMPTZ`

plaflamme:
[Documentation](https://cloud.google.com/bigquery/docs/querying-partitioned-tables#query_an_ingestion-time_partitioned_table) states that the `_PARTITIONTIME`'s pseudo-column type is `TIMESTAMP` (which is `TIMESTAMPTZ`)

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -289,7 +289,7 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 and bq_table.time_partitioning
                 and not bq_table.time_partitioning.field
             ):
-                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMPTZ")
+                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP", dialect="bigquery")
                 if bq_table.time_partitioning.type_ == "DAY":
                     columns["_PARTITIONDATE"] = exp.DataType.build("DATE")
 
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -289,7 +289,7 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 and bq_table.time_partitioning
                 and not bq_table.time_partitioning.field
             ):
-                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP")
+                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMPTZ")
                 if bq_table.time_partitioning.type_ == "DAY":
                     columns["_PARTITIONDATE"] = exp.DataType.build("DATE")
 


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -289,7 +289,7 @@ def create_mapping_schema(
                 and bq_table.time_partitioning
                 and not bq_table.time_partitioning.field
             ):
-                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP")
+                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMPTZ")
```
georgesittas, line 292:
```suggestion
                columns["_PARTITIONTIME"] = exp.DataType.build("TIMESTAMP", dialect="bigquery")
```

georgesittas (approved):
Thanks, this seems legit to me. I think we may need a migration for this one.

plaflamme:
> I think we may need a migration for this one.

In our case, I think this only showed up in `create_external_models`, so the migration would occur whenever someone runs that command? Not sure where this might show up elsewhere

georgesittas:
Sharing these conclusions, based on my investigation:

1. I did reproduce the behavior where `DATETIME` is written to the file instead of `TIMESTAMP`. The [latter is the correct](https://github.com/user-attachments/assets/a7d5133d-e779-4b10-b08b-db26b27e8fe1) type for the `_PARTITIONTIME` pseudo-column in BigQuery.
2. This change is only relevant to the `create_external_models` command, because it's [only reached when](https://github.com/TobikoData/sqlmesh/blob/3d368c871c5c4e6218a57f4a4131151ca85496bc/sqlmesh/core/engine_adapter/bigquery.py#L288) `include_pseudo_columns` is `True` and that only happens in the [schema loader](https://github.com/search?q=repo%3ATobikoData%2Fsqlmesh%20include_pseudo_columns%3DTrue&type=code). Specifically, the `columns` call only happens for **external tables**.
3. The BigQuery pseudo-columns are [excluded when expanding stars](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/qualify_columns.py#L667-L668), so in a typical `SELECT * FROM <ext. table>` model, the `_PARTITIONTIME` is not even included in the `columns_to_types`, so in this case the data type in the `external_models.yaml` file doesn't matter.
4. However… if one explicitly selects the column like in `SELECT id, _PARTITIONTIME AS foo FROM <ext. table>`, then these pseudo-columns _are_ included in `columns_to_types`, and the type is `Type.TIMESTAMP`, which generates `"DATETIME"` for BigQuery. This indeed [leads to a bug](https://github.com/user-attachments/assets/0558ca1d-09cb-4039-9f0e-b37ab38c49c7) where we create a table with `foo` as a `DATETIME` column, instead of `TIMESTAMP`.

Based on the above, this PR should only affect users _after_ they re-run `create_external_models` with this change in place, because it would change the type of the external model from `DATETIME` to `TIMESTAMP`, which will result in a breaking change at the edge of the DAG.

In that case, users can either apply a plan with the change (perhaps a forward-only one?), or move the previous `external_models.yaml` under the `external_models` [directory](https://sqlmesh.readthedocs.io/en/stable/concepts/models/external_models/#using-the-external_models-directory) to preserve the old behavior and avoid the diff altogether, since then the old type (`DATETIME`) will take precedence over the new one (`TIMESTAMP`).

cc @izeigerman for visibility.

PR #4314, opened 2025-05-05
feat: set query label session property in bq session

ncbkr:
we want to use the SQLMesh session properties to attach labels to the BigQuery jobs. Using these labels we want to enable ourselves to

- map BigQuery jobs to SQLMesh models
- map BigQuery jobs to owners/users of SQLMesh

In the model definition, this would like as follows

```
session_properties (
    query_label = [
      ('label1', 'value1'),
      ('label2', 'value2'),
      ('label3', 'value3')
    ]
  )
```

This is currently not possible in an easy way since all jobs are created by the same service account.

The screenshot below displays the information about a BigQuery job created by sqlmesh to backfill a model. In the last line you can see the label my_session_property_key: my_session_property_value. This comes from the session_properties of the sqlmesh model definition. According to the google [docs for BigQuery](https://cloud.google.com/bigquery/docs/adding-labels#adding-label-to-session), all jobs that are run within a session inherit the labels set at the beginning of the session using SET @@query_label = <my_labels>

image


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -182,7 +182,32 @@ def query_factory() -> Query:
     def _begin_session(self, properties: SessionProperties) -> None:
         from google.cloud.bigquery import QueryJobConfig
 
-        job = self.client.query("SELECT 1;", job_config=QueryJobConfig(create_session=True))
+        query_label = []
+        if "query_label" in properties and isinstance(properties["query_label"], exp.Array):
```
izeigerman, line 186: I think this can also be `exp.Tuple` not just `exp.Array`. Eg:
```
query_labels = ['label_a', 'label_b']
````
vs
```
query_labels = ('label_a', 'label_b')
```

File: sqlmesh/core/engine_adapter/bigquery.py [hunk 2]
```diff
@@ -182,7 +182,32 @@ def query_factory() -> Query:
...
+            for label_tuple in properties["query_label"].expressions:
+                if (
+                    isinstance(label_tuple, exp.Tuple)
+                    and len(label_tuple.expressions) == 2
+                    and all(isinstance(label, exp.Literal) for label in label_tuple.expressions)
```
izeigerman, line 191: what if it's an identifier and not a string literal eg: `query_labels = [ label_a ]` and not `query_labels = [ 'label_a' ]`. Might worth support both approaches.

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -182,7 +182,32 @@ def query_factory() -> Query:
...
+                ):
+                    query_label.append(
+                        (label_tuple.expressions[0].this, label_tuple.expressions[1].this)
+                    )
+                else:
+                    raise SQLMeshError(
+                        "Invalid query label format. Expected a tuple of two string literals."
```
izeigerman, line 198: I feel like this is something that should be captured at model load time and not at execution time. This seems too late.

izeigerman:
Thanks for your contribution! Can you please add some tests.

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: Yep, perhaps also `Paren` as an edge case.

ncbkr: Sorry for not including a usage example in the PR description. I imagined the model definition to look like this

```
session_properties (
    query_label = [
      ('key1', 'value1'),
      ('key2', 'value2'),
      ('key3', 'value3')
    ]
  )
```

I chose this format based on the google docs:

> a key-value pair for the label to assign to all future queries in the session. You can also add multiple key-value pairs, separated by a comma (for example, SET @@query_label = "key1:value1,key2:value2").

Also, this is consistent with the usage of `labels` in both, the virtual and phyiscal properties. 

This format is also why I explicitely check for an array.

What do you think about this format? Would you prefer another one?

e.g. choosing this format
```
query_labels = ['label_a', 'label_b'] | ('label_a', 'label_b')  # array or tuple
```

label_a and label_b would have to follow this pattern: `key:value`

ncbkr: @georgesittas would this be an edge case you had in mind?

```
query_labels = ('label_a')
```

So `query_label` is not a tuple or array but a single string `'key:vale'` wrapped in parentheses?

izeigerman: The format makes sense, but it might also be:
```
session_properties (
    query_label = (
      ('key1', 'value1'),
      ('key2', 'value2'),
      ('key3', 'value3')
    )
  )
  ```
  which would make a tuple of tuples rather than array of tuples.

File: sqlmesh/core/engine_adapter/bigquery.py (hunk 2 above)
Comment (georgesittas) [reply sqlmesh/core/engine_adapter/bigquery.py:191]: We could simply require strings; not sure what's the value of supporting identifiers here. Seems like an unnecessary increase in scope?

Comment (izeigerman) [reply sqlmesh/core/engine_adapter/bigquery.py:191]: Sounds good. I'm good with that

File: tests/core/test_model.py [hunk 4]
```diff
@@ -4248,6 +4250,64 @@ def test_model_session_properties(sushi_context):
...
+                    ('key1', 'value1')
+                )
+            )
+        );
+        SELECT a FROM tbl;
+        """,
+            default_dialect="bigquery",
+        )
+    )
+    assert model.session_properties == {
```
izeigerman, line 4305: Can we please also add a few test cases that fail the validation

sqlmesh/core/engine_adapter/bigquery.py
izeigerman: The properties are already validated here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/common.py#L425-L432. Can we please extend that instead?

sqlmesh/core/model/definition.py
```diff
@@ -2634,6 +2637,64 @@ def _validate_model_fields(klass: t.Type[_Model], provided_fields: t.Set[str], p
         raise_config_error(f"Invalid extra fields {extra_fields} in the model definition", path)
 
 
+def _validate_model_properties(meta_fields: dict[str, t.Any], path: Path) -> None:
```
izeigerman, line 2640: Instead of doing it here, can we have a custom Pydantic validator for `session_properties_` in `model/meta.py` and remove the field from the generic property validator here: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/common.py#L425-L432.

This way all model types will benefit from this and not just SQL models.

sqlmesh/core/model/definition.py
```diff
@@ -2634,6 +2637,64 @@ def _validate_model_fields(klass: t.Type[_Model], provided_fields: t.Set[str], p
...
+    # store for later validation of specific properties
+    model_properties: dict[str, t.Any] = {}
+
+    # validate that all properties kinds are key-value mappings
+    for kind in PROPERTIES:
```
izeigerman, line 2645: This is unnecessary. We already validate properties with this validator: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/common.py#L425-L432

tests/core/test_model.py
ncbkr: Good point. I added them

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -551,6 +552,48 @@ def test_begin_end_session(mocker: MockerFixture):
...
+                    exp.Tuple(
+                        expressions=[
+                            exp.Literal.string("key2"),
+                            exp.Literal.string("value2"),
+                        ]
+                    ),
+                ]
+            )
+        }
+    ):
```
georgesittas, line 575:
```suggestion
    with adapter.session({"query_label": parse_one("[('key1', 'value1'), ('key2', 'value2')]")}):
```

tests/core/test_model.py
```diff
@@ -4418,6 +4421,127 @@ def test_model_session_properties(sushi_context):
...
+                    ]
+                ),
+                exp.Tuple(
+                    expressions=[
+                        exp.Literal.string("key2"),
+                        exp.Literal.string("value2"),
+                    ]
+                ),
+            ]
+        )
```
georgesittas, line 4457: Ditto

tests/core/test_model.py
```diff
@@ -4418,6 +4421,127 @@ def test_model_session_properties(sushi_context):
...
+        SELECT a FROM tbl;
+        """,
+            default_dialect="bigquery",
+        )
+    )
+    assert model.session_properties == {
+        "query_label": exp.Paren(
+            this=exp.Tuple(expressions=[exp.Literal.string("key1"), exp.Literal.string("value1")])
+        )
+    }
```
georgesittas, line 4480: Ditto

georgesittas (approved):
Thanks @ncbkr! Left a few final comments as well.

sqlmesh/core/model/meta.py
```diff
@@ -310,6 +311,43 @@ def _refs_validator(cls, vs: t.Any, info: ValidationInfo) -> t.List[exp.Expressi
     def ignored_rules_validator(cls, vs: t.Any) -> t.Any:
         return LinterConfig._validate_rules(vs)
 
+    @field_validator("session_properties_", mode="before")
+    def session_properties_validator(cls, v: t.Any, info: ValidationInfo) -> t.Any:
+        # use the generic properties validator to parse the session properties
+        parsed_session_properties = parse_properties(type(cls), v, info)
```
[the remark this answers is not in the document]
ncbkr: @georgesittas  I wondered as well. When I just had `cls`, my linter complained. It said:

> an argument of type "Self@ModelMeta" is not compatible with parameter "cls" of type "Type[Unknown]" in the function "parse_properties"

so I used `type(cls)` to be compatible with the `Type` type hint

tests/core/engine_adapter/test_bigquery.py
ncbkr: oh yes. much nicer!

ncbkr:
Thank you as well for considering this change and providing feedback. This was very helpful. Happy to contribute to this great project

izeigerman (approved):

PR #4319, opened 2025-05-06
feat: store plan flags

toriwei:
identify and store flags a user sets manually

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1342,19 +1342,19 @@ class GenericContext(BaseContext, t.Generic[C]):
             "execution_time": execution_time,
             "create_from": create_from,
             "skip_tests": skip_tests,
-            "restate_models": list(restate_models) if restate_models else None,
+            "restate_models": list(restate_models) if restate_models is not None else None,
             "no_gaps": no_gaps,
             "skip_backfill": skip_backfill,
             "empty_backfill": empty_backfill,
             "forward_only": forward_only,
             "allow_destructive_models": list(allow_destructive_models)
-            if allow_destructive_models
+            if allow_destructive_models is not None
             else None,
             "no_auto_categorization": no_auto_categorization,
             "effective_from": effective_from,
             "include_unmodified": include_unmodified,
-            "select_models": list(select_models) if select_models else None,
-            "backfill_models": list(backfill_models) if backfill_models else None,
+            "select_models": list(select_models) if select_models is not None else None,
+            "backfill_models": list(backfill_models) if backfill_models is not None else None,
             "enable_preview": enable_preview,
             "run": run,
             "diff_rendered": diff_rendered,
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1161,11 +1161,11 @@ class GenericContext(BaseContext, t.Generic[C]):
         end: t.Optional[TimeLike] = None,
         execution_time: t.Optional[TimeLike] = None,
         create_from: t.Optional[str] = None,
-        skip_tests: bool = False,
+        skip_tests: t.Optional[bool] = None,
         restate_models: t.Optional[t.Iterable[str]] = None,
-        no_gaps: bool = False,
-        skip_backfill: bool = False,
-        empty_backfill: bool = False,
+        no_gaps: t.Optional[bool] = None,
+        skip_backfill: t.Optional[bool] = None,
+        empty_backfill: t.Optional[bool] = None,
         forward_only: t.Optional[bool] = None,
         allow_destructive_models: t.Optional[t.Collection[str]] = None,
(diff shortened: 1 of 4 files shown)


sqlmesh/core/context.py
```diff
@@ -1335,6 +1335,56 @@ def plan_builder(
...
+        user_defined_flags: t.Dict[
+            str,
+            t.Union[
+                t.Optional[TimeLike],
+                t.Optional[str],
+                t.Optional[bool],
+                t.Optional[t.Iterable[str]],
+                t.Optional[t.Collection[str]],
+            ],
+        ] = {k: v for k, v in kwargs.items() if v is not None}
```
crericha, line 1378: When we discussed, I thought we were going to put this logic in the plan builder class, not in the method. Can you please clarify with @izeigerman which he was referring too.

Also, I'm not so sure the union is useful over any. Especially if it has to be defined in multiple places.

sqlmesh/core/context.py
```diff
@@ -1335,6 +1335,56 @@ def plan_builder(
         Returns:
             The plan builder.
         """
+        kwargs: t.Dict[
+            str,
+            t.Union[
+                t.Optional[TimeLike],
```
izeigerman, line 1341: Can we have a type alias for this? Also I don't think we need both `Iterable` and `Collection`. Additionally, why are these optional? We should only include non-null fields.

sqlmesh/core/plan/builder.py
```diff
@@ -107,6 +107,18 @@ def __init__(
         ensure_finalized_snapshots: bool = False,
         interval_end_per_model: t.Optional[t.Dict[str, int]] = None,
         console: t.Optional[PlanBuilderConsole] = None,
+        user_defined_flags: t.Optional[
+            t.Dict[
+                str,
+                t.Union[
+                    t.Optional[TimeLike],
```
izeigerman, line 114: Please let's use a type alias + see my other comment wrt optional and iterable vs collection

sqlmesh/core/context.py
izeigerman: I'd also prefer to be concrete and just use `List[str]`.

sqlmesh/core/plan/definition.py
```diff
@@ -63,6 +63,19 @@ class Plan(PydanticModel, frozen=True):
     effective_from: t.Optional[TimeLike] = None
     execution_time: t.Optional[TimeLike] = None
 
+    user_defined_flags: t.Optional[
```
izeigerman, line 66: I believe `user_provided_*` is more accurate. We define the flags, users provide them

sqlmesh/core/context.py
toriwei: by that do you mean convert `Iterable` and `Collection` values to a list?

ex: plan_builder takes in `select_models: t.Optional[t.Collection[str]] = None,` so we coerce to a list in the kwargs dict: `"select_models": list(select_models) if select_models else None,`

toriwei: the kwargs values need to be optional, since user doesn't have to set those values. but you're correct, for the filtered flags dictionary, those are not optional. i will adjust

sqlmesh/core/context.py
```diff
@@ -1335,6 +1336,42 @@ def plan_builder(
...
+            "execution_time": execution_time,
+            "create_from": create_from,
+            "skip_tests": skip_tests,
+            "restate_models": list(restate_models) if restate_models else None,
+            "no_gaps": no_gaps,
+            "skip_backfill": skip_backfill,
+            "empty_backfill": empty_backfill,
+            "forward_only": forward_only,
+            "allow_destructive_models": list(allow_destructive_models)
+            if allow_destructive_models
```
izeigerman, line 1351:
```suggestion
            if allow_destructive_models is not None
```

sqlmesh/core/context.py
```diff
@@ -1335,6 +1336,42 @@ def plan_builder(
...
+            else None,
+            "no_auto_categorization": no_auto_categorization,
+            "effective_from": effective_from,
+            "include_unmodified": include_unmodified,
+            "select_models": list(select_models) if select_models else None,
```
izeigerman, line 1356:
```suggestion
            "select_models": list(select_models) if select_models is not None else None,
```

sqlmesh/core/context.py
```diff
@@ -1335,6 +1336,42 @@ def plan_builder(
...
+            "backfill_models": list(backfill_models) if backfill_models else None,
```
izeigerman, line 1357:
```suggestion
            "backfill_models": list(backfill_models) if backfill_models is not None else None,
```

izeigerman (approved):
please see the additional comments. LGTM once addressed

PR #4324, opened 2025-05-07
Fix columns example in tests doc

hliu1027:
For unit tests YAML, `<test_name>.inputs.<upstream_model>.columns` actually expects a dictionary. I tested that the current example syntax with `- ` fails, and the fixed version works.


georgesittas (approved):
Good catch, thanks for fixing!

PR #4325, opened 2025-05-07
Fix: Pop select model in tablediff to ensure backwards compatibility

themisvaltinos:
This fixes a bug in the `table_diff` command when used without the `select-model` to ensure backwards compatibility:
```bash
> sqlmesh table_diff dev:prod model
```
which would raise a keyword argument error as the select model would be passed in the kwargs. This fixes it by popping the select model from the kwargs beforehand regardless similarly to the handling in other commands.

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -908,13 +908,19 @@ def table_diff(
     """Show the diff between two tables or a selection of models when they are specified."""
     source, target = source_to_target.split(":")
     select_model = kwargs.pop("select_model", None)
-    select_models = {model} if model else select_model
-    obj.table_diff(
-        source=source,
-        target=target,
-        select_models=select_models,
-        **kwargs,
-    )
+
+    if model and select_model:
+        obj.console.log_error(
+            "The --select-model option cannot be used together with a model argument. Please choose one of them."
+        )
+    else:
+        select_models = {model} if model else select_model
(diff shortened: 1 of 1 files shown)


sqlmesh/cli/main.py
```diff
@@ -907,7 +907,8 @@ def table_diff(
 ) -> None:
     """Show the diff between two tables or a selection of models when they are specified."""
     source, target = source_to_target.split(":")
-    select_models = {model} if model else kwargs.pop("select_model", None)
+    select_model = kwargs.pop("select_model", None)
+    select_models = {model} if model else select_model
```
izeigerman, line 911: Shouldn't we fail if both were passed?

themisvaltinos: yes i was thinking of model taking precedence but will revise it to fail instead

izeigerman (approved):

sqlmesh/cli/main.py
```diff
@@ -907,13 +907,20 @@ def table_diff(
...
-    obj.table_diff(
-        source=source,
-        target=target,
-        select_models=select_models,
-        **kwargs,
-    )
+    select_model = kwargs.pop("select_model", None)
+
+    if model and select_model:
+        obj.console.log_error(
```
izeigerman, line 913: I suggest raising instead so that non-0 error code is returned.

PR #4327, opened 2025-05-07
refactor(vscode): move result into shared bus

benfdking:


vscode/extension/esbuild.js
```diff
@@ -15,6 +17,9 @@ async function main() {
     outfile: 'dist/extension.js',
     external: ['vscode'],
     logLevel: 'warning',
+    define: {
+      'process.env.NODE_ENV': IS_PRODUCTION ? '"production"' : '"development"',
```
mykalmax, line 21: why do we need double quoting  ?

mykalmax (approved):

vscode/extension/esbuild.js
benfdking: That's a good catch, I'll avoid it in the next on, as I can just undo this change for now.

Issue #3684, opened 2025-01-22
Add command to reset/remove entire project


treysp:
Delete all engine objects (state and warehouse), caches

serkef:
yes that would be very useful for people who do PoCs and are in the pre-prod phase. 
I've found myself deleting state db and wh more than once.

Happily-Coding:
It would also be great if that command could be run for a particular table and all its dependencies.

PR #4328, opened 2025-05-07
Feat: Add destroy command to remove entire project resources

themisvaltinos:
This update introduces the `sqlmesh destroy` command, which permanently deletes all engine-managed objects, including warehouse data, the associated state tables and cached SQLMesh assets, effectively resetting the entire project, fixes: #3684 

**Usage:** `sqlmesh destroy`

This might be useful for cleaning up development projects, proof-of-concept work or pre-production phases with temporary setups.

> ⚠️ This is destructive and irreversible, and it removes both development and the production environment, so it should only be used when a full removal is necessary.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -832,6 +832,19 @@ class GenericContext(BaseContext, t.Generic[C]):
 
         return success
 
+    @python_api_analytics
+    def destroy(self) -> bool:
+        success = False
+
+        if self.console.start_destroy():
+            try:
+                self._destroy()
+                success = True
+            finally:
+                self.console.stop_destroy(success=success)
+
+        return success
+
     @t.overload
     def get_model(
         self, model_or_snapshot: ModelOrSnapshot, raise_if_missing: Literal[True] = True
@@ -1554,16 +1567,19 @@ class GenericContext(BaseContext, t.Generic[C]):
         )
 
     @python_api_analytics
-    def invalidate_environment(self, name: str, sync: bool = False) -> None:
+    def invalidate_environment(
+        self, name: str, sync: bool = False, protect_prod: t.Optional[bool] = True
+    ) -> None:
         """Invalidates the target environment by setting its expiration timestamp to now.
 
         Args:
             name: The name of the environment to invalidate.
             sync: If True, the call blocks until the environment is deleted. Otherwise, the environment will
                 be deleted asynchronously by the janitor process.
+            protect_prod: If True, prevents invalidation of the production environment.
         """
         name = Environment.sanitize_name(name)
-        self.state_sync.invalidate_environment(name)
+        self.state_sync.invalidate_environment(name, protect_prod)
         if sync:
             self._cleanup_environments()
             self.console.log_success(f"Environment '{name}' deleted.")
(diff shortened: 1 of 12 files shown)


sqlmesh/core/context.py
```diff
@@ -1516,16 +1529,19 @@ def apply(
         )
 
     @python_api_analytics
-    def invalidate_environment(self, name: str, sync: bool = False) -> None:
+    def invalidate_environment(
+        self, name: str, sync: bool = False, protect_prod: t.Optional[bool] = True
```
izeigerman, line 1533: No, this shouldn't be in the public API. If you want to reuse the method I suggest having a shared private method instead.

sqlmesh/core/state_sync/db/environment.py
```diff
@@ -108,14 +108,15 @@ def update_environment_statements(
                 columns_to_types=self._environment_statements_columns_to_types,
             )
 
-    def invalidate_environment(self, name: str) -> None:
+    def invalidate_environment(self, name: str, protect_prod: t.Optional[bool] = True) -> None:
```
izeigerman, line 111: ditto: why the new argument is optional?

sqlmesh/core/context.py
themisvaltinos: yes good point actually only the invalidate method is used from it too so no need to call this

sqlmesh/core/state_sync/db/environment.py
themisvaltinos: right my bad no need for it

sqlmesh/core/context.py
```diff
@@ -1523,6 +1536,7 @@ def invalidate_environment(self, name: str, sync: bool = False) -> None:
             name: The name of the environment to invalidate.
             sync: If True, the call blocks until the environment is deleted. Otherwise, the environment will
                 be deleted asynchronously by the janitor process.
+            protect_prod: If True, prevents invalidation of the production environment.
```
georgesittas, line 1539: Outdated comment

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -313,18 +313,26 @@ def snapshots_exist(self, snapshot_ids: t.Iterable[SnapshotIdLike]) -> t.Set[Sna
...
             self.snapshot_state.auto_restatements_table,
             self.environment_state.environments_table,
+            self.environment_state.environment_statements_table,
             self.interval_state.intervals_table,
             self.plan_dags_table,
             self.version_state.versions_table,
         ):
             self.engine_adapter.drop_table(table)
+            if including_backup:
+                self.engine_adapter.drop_table(_backup_table_name(table))
```
georgesittas, line 329: What if a backup table doesn't exist? Have you tested this?

georgesittas (approved):
LGTM, just one question related to backup tables.

sqlmesh/core/context.py
themisvaltinos: good catch! thanks

sqlmesh/core/state_sync/db/facade.py
georgesittas: Cool thanks for double-checking, I forgot this was the default behavior.

izeigerman (approved):

PR #4333, opened 2025-05-07
Fix: Gracefully handle execution errors in before after all

themisvaltinos:
Catch and raise only the relevant error messages for `before all` and `after all` statements  (or dbt's `on_run_start` , `on_run_end` respectively) when an exception during their execution occurs

### DIFF
--- a/sqlmesh/core/environment.py
+++ b/sqlmesh/core/environment.py
@@ -262,24 +262,29 @@ def execute_environment_statements(
     end: t.Optional[TimeLike] = None,
     execution_time: t.Optional[TimeLike] = None,
 ) -> None:
-    rendered_expressions = [
-        expr
-        for statements in environment_statements
-        for expr in render_statements(
-            statements=getattr(statements, runtime_stage.value),
-            dialect=adapter.dialect,
-            default_catalog=default_catalog,
-            python_env=statements.python_env,
-            jinja_macros=statements.jinja_macros,
-            snapshots=snapshots,
-            start=start,
-            end=end,
-            execution_time=execution_time,
-            environment_naming_info=environment_naming_info,
-            runtime_stage=runtime_stage,
-            engine_adapter=adapter,
+    try:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/environment.py
```diff
@@ -282,4 +283,9 @@ def execute_environment_statements(
     if rendered_expressions:
         with adapter.transaction():
             for expr in rendered_expressions:
-                adapter.execute(expr)
+                try:
+                    adapter.execute(expr)
+                except Exception as e:
+                    raise SQLMeshError(
+                        f"An error occurred during execution of the following `{runtime_stage.value}` statement:\n\n{expr}\n\n{e}"
```
georgesittas, line 290: I forget: are we usually using backticks or single quotes for these error messages? Curious if this is consistent.

georgesittas (approved):

sqlmesh/core/environment.py
izeigerman: single quotes please

sqlmesh/core/environment.py
```diff
@@ -282,4 +283,9 @@ def execute_environment_statements(
     if rendered_expressions:
         with adapter.transaction():
             for expr in rendered_expressions:
-                adapter.execute(expr)
+                try:
+                    adapter.execute(expr)
```
izeigerman, line 287: Should we wrap the whole thing? Cause things can fail during rendering too

sqlmesh/core/environment.py
themisvaltinos: good catch thanks added single quotes instead

sqlmesh/core/environment.py
themisvaltinos: wrapped the rendering phase as well. just done it in separate try handlers to provide more specific error messages to be able to pinpoint better what went wrong and where

georgesittas (approved):

PR #4335, opened 2025-05-08
Enhances MSSQL connection with pyodbc support

mattiasthalen:
Adds support for `pyodbc` as an alternative driver for MSSQL connections, providing users with more flexibility and authentication options.

- Introduces a `driver` option to choose between `pymssql` and `pyodbc`.
- Implements connection string generation for `pyodbc` based on connection parameters.
- Includes validation for different authentication methods when using `pyodbc`.
- Updates dependencies to include `pyodbc`.

### DIFF
--- a/docs/integrations/engines/azuresql.md
+++ b/docs/integrations/engines/azuresql.md
@@ -31,6 +31,6 @@ pip install "sqlmesh[azuresql-odbc]"
 | `appname`         | The application name to use for the connection                   |    string    |    N     |
 | `conn_properties` | The list of connection properties                                | list[string] |    N     |
 | `autocommit`      | Is autocommit mode enabled. Default: false                       |     bool     |    N     |
-| `driver`         | The driver to use for the connection. Default: pymsql            |    string    |    N     |
+| `driver`         | The driver to use for the connection. Default: pymssql            |    string    |    N     |
 | `driver_name`     | The driver name to use for the connection. E.g., *ODBC Driver 18 for SQL Server* |    string    |    N     |
 | `odbc_properties` | The dict of ODBC connection properties. E.g., authentication: ActiveDirectoryServicePrincipal. See more [here](https://learn.microsoft.com/en-us/sql/connect/odbc/dsn-connection-string-attribute?view=sql-server-ver16). | dict |    N     |
--- a/docs/integrations/engines/azuresql.md
+++ b/docs/integrations/engines/azuresql.md
@@ -2,14 +2,17 @@
 
 [Azure SQL](https://azure.microsoft.com/en-us/products/azure-sql) is "a family of managed, secure, and intelligent products that use the SQL Server database engine in the Azure cloud."
 
-The Azure SQL adapter only supports authentication with a username and password. It does not support authentication with Microsoft Entra or Azure Active Directory.
-
 ## Local/Built-in Scheduler
 **Engine Adapter Type**: `azuresql`
 
 ### Installation
+#### User / Password Authentication:
 ```
 pip install "sqlmesh[azuresql]"
+```
+#### Microsoft Entra ID / Azure Active Directory Authentication:
+```
+pip install "sqlmesh[azuresql-odbc]"
 ```
 
 ### Connection options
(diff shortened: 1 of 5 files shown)


mattiasthalen:
@erindru hey, I just got around to do this after some onboarding at my new employer.

I also had some time to play around with the SQL Database object in fabric, and... I think just adding pyodbc allows us to connect to it like any other mssql. Dare I say it? It feels like it's really close to have some initial Fabric support! 👯 

I know, I know, it's not the lakehouse, nor the warehouse objects. But it's something!

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -220,11 +220,30 @@ def query_factory() -> Query:
...
-                conn.bulk_copy(temp_table.sql(dialect=self.dialect), rows)
+
+                if hasattr(conn, 'bulk_copy'):
+                    # Use bulk_copy if available
+                    rows: t.List[t.Tuple[t.Any, ...]] = list(
+                        df.replace({np.nan: None}).itertuples(index=False, name=None)  # type: ignore
+                    )
+                    conn.bulk_copy(temp_table.sql(dialect=self.dialect), rows)
+                else:
+                    # Fallback to executemany if bulk_copy is not available
```
erindru, line 232: Is there a reason why we can't fall back to the superclass implementation of `_df_to_source_queries` if `bulk_copy` isnt available?

pyproject.toml
```diff
@@ -37,7 +37,7 @@ classifiers = [
 
 [project.optional-dependencies]
 athena = ["PyAthena[Pandas]"]
-azuresql = ["pymssql"]
+azuresql = ["pymssql", "pyodbc"]
```
erindru, line 40: I'm not sure we should depend on both of these, users only need one or the other.

@izeigerman could this be a new optional dependency, eg:

```
[project.optional-dependencies]
azuresql = ["pymssql"]
azuresql-odbc = ["pyodbc"]
```

Or should we just say in the docs that you need to install `pyodbc` and set up the ODBC driver separately?

mattiasthalen:
@erindru I tried hitting the Warehouse endpoint, but there are several differences compared to MSSQL that doesn't work the same. E.g., information_schema is case sensitive.

Shouldn't be that hard to support it though. I'll probably have a look on that next.

Still refreshing to see that the SQL Database object was supported by just being able to login.

Ok, I was scratching my head on how to solve those failing tests 😅

Will fix asap and send for review 🎉

docs/integrations/engines/azuresql.md
```diff
@@ -2,29 +2,35 @@
...
 | `database`        | The target database                                              |    string    |    N     |
 | `charset`         | The character set used for the connection                        |    string    |    N     |
 | `timeout`         | The query timeout in seconds. Default: no timeout                |     int      |    N     |
 | `login_timeout`   | The timeout for connection and login in seconds. Default: 60     |     int      |    N     |
 | `appname`         | The application name to use for the connection                   |    string    |    N     |
 | `conn_properties` | The list of connection properties                                | list[string] |    N     |
-| `autocommit`      | Is autocommit mode enabled. Default: false                       |     bool     |    N     |
\ No newline at end of file
+| `autocommit`      | Is autocommit mode enabled. Default: false                       |     bool     |    N     |
+| `driver`         | The driver to use for the connection. Default: pymsql            |    string    |    N     |
```
treysp, line 34:
```suggestion
| `driver`         | The driver to use for the connection. Default: pymssql            |    string    |    N     |
```

docs/integrations/engines/mssql.md
```diff
@@ -24,3 +29,6 @@ pip install "sqlmesh[mssql]"
 | `appname`         | The application name to use for the connection               |    string    |    N     |
 | `conn_properties` | The list of connection properties                            | list[string] |    N     |
 | `autocommit`      | Is autocommit mode enabled. Default: false                   |     bool     |    N     |
+| `driver`         | The driver to use for the connection. Default: pymsql            |    string    |    N     |
```
treysp, line 32:
```suggestion
| `driver`         | The driver to use for the connection. Default: pymssql            |    string    |    N     |
```

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -220,11 +224,13 @@ def query_factory() -> Query:
                 columns_to_types_create = columns_to_types.copy()
                 self._convert_df_datetime(df, columns_to_types_create)
                 self.create_table(temp_table, columns_to_types_create)
+                conn = self._connection_pool.get()
```
erindru, line 227: Why did this need to change?

erindru:
Thanks, this is looking pretty good!

Just one more thing - there are no tests. Can you please add some tests to [test_connection_config](https://github.com/TobikoData/sqlmesh/blob/main/tests/core/test_connection_config.py) that show:

- The parameter validations you added are working
- The `_connection_kwargs_keys` are as expected for each variant (`pymssql` and `pyodbc`)
- `pyodbc.connect` gets invoked with the correct ODBC string

mattiasthalen:
Hey sorry for being slow to progress. It's been hectic at work, but I hopens have time this weekend and write the tests

erindru:
No problem, there is no expectation / time pressure on community contributions! Thanks for the update

Issue #4202, opened 2025-04-21
`@UNION` macro so that it takes in a condition that determines whether the right-hand side will be included in the union


michaelvanloo-ta:
Desired result:
- If condition is statically evaluatable and is false, then prune the query2
```
<query1>
@UNION(<condition>)
<query2>
```

Example:

```
MODEL (
  name cdc__production ,
  kind INCREMENTAL_BY_UNIQUE_KEY (
    unique_key ARRAY(tag_id, trend_time),
  ),
  start '2025-01-01'
);

  SELECT
    tag_id,
    trend_time,
    trend_status,
    trend_value,
    _dlt_load_id,
    _start_lsn,
    _operation
  FROM dbo_production_ct
  WHERE _dlt_load_id::DOUBLE::TIMESTAMP BETWEEN @start_ts AND @end_ts
  --Only union the snapshot table for the first run
  @UNION(@start_ts = '2025-01-01 00:00:00')
  SELECT 
    tag_id,
    trend_time,
    trend_status,
    trend_value,
    _dlt_load_id
    0::BINARY AS _start_lsn, /* mimic min lsn */
    2::INT AS _operation /* mimic snapshot as records inserted */
  FROM production_snapshot

```

PR #4337, opened 2025-05-08
Feat: Add condition in the union macro for conditional union of tables

themisvaltinos:
This update extends the `@UNION` macro  with a condition that is statically evaluatable and if it is false it short-circuits and it only keeps the first table, otherwise unions all tables, fixes: #4202 

Example:
```sql
MODEL (
  name sushi.table_4
);
  
@UNION(@get_date() > '1996-02-10', 'distinct', sushi.table_1, sushi.table_2, sushi.table_3)
```

the previous behaviour when the first argument can't be interpreted as a conditional True or False remains with the first argument being the union type and then the table names.

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -994,7 +994,7 @@ def union(
     """
 
     if not args:
-        raise SQLMeshError("At least one table is required for UNION.")
+        raise SQLMeshError("At least one table is required for the @UNION macro.")
 
     arg_idx = 0
     # Check for condition
@@ -1002,11 +1002,11 @@ def union(
     if isinstance(condition, bool):
         arg_idx += 1
         if arg_idx >= len(args):
-            raise SQLMeshError("Expected more arguments after condition.")
+            raise SQLMeshError("Expected more arguments after the condition of the `@UNION` macro.")
 
     # Check for union type
     type_ = exp.Literal.string("ALL")
-    if arg_idx < len(args) and isinstance(args[arg_idx], exp.Literal):
+    if isinstance(args[arg_idx], exp.Literal):
         type_ = args[arg_idx]  # type: ignore
         arg_idx += 1
     kind = type_.name.upper()
@@ -1014,13 +1014,9 @@ def union(
         raise SQLMeshError(f"Invalid type '{type_}'. Expected 'ALL' or 'DISTINCT'.")
 
     # Remaining args should be tables
-    tables = []
-    for table in args[arg_idx:]:
-        if isinstance(table, exp.Column):
-            table = exp.table_(table.this, db=table.args.get("table"), catalog=table.args.get("db"))
-        else:
-            table = exp.to_table(table, dialect=evaluator.dialect)  # type: ignore
-        tables.append(table)
+    tables = [
+        exp.to_table(e.sql(evaluator.dialect), dialect=evaluator.dialect) for e in args[arg_idx:]
+    ]
 
     columns = {
         column
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1008,6 +1008,53 @@ def union(
 
 
 @macro()
+def union_if(
+    evaluator: MacroEvaluator,
+    condition: exp.Expression,
+    type_: exp.Literal = exp.Literal.string("ALL"),
+    *tables: exp.Table,
+) -> exp.Query:
+    """Returns a UNION of the given tables if the condition is true, otherwise returns just the first table.
+       The behaviour remains of only choosing columns that have the same name and type.
+
+    Example:
+        >>> from sqlglot import parse_one
+        >>> from sqlglot.schema import MappingSchema
+        >>> from sqlmesh.core.macros import MacroEvaluator
+        >>> sql = "@UNION_IF(@start_ts = '2025-01-01 00:00:00', 'distinct', foo, bar)"
+        >>> MacroEvaluator(schema=MappingSchema({"foo": {"a": "int", "b": "string", "c": "string"}, "bar": {"c": "string", "a": "int", "b": "int"}})).transform(parse_one(sql)).sql()
+        'SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM foo UNION SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM bar'
+    """
+    kind = type_.name.upper()
+    if kind not in ("ALL", "DISTINCT"):
+        raise SQLMeshError(f"Invalid type '{type_}'. Expected 'ALL' or 'DISTINCT'.")
+
+    columns = {
+        column
+        for column, _ in reduce(
+            lambda a, b: a & b,  # type: ignore
+            (evaluator.columns_to_types(table).items() for table in tables),
+        )
+    }
+
+    projections = [
+        exp.cast(column, type_, dialect=evaluator.dialect).as_(column)
+        for column, type_ in evaluator.columns_to_types(tables[0]).items()
+        if column in columns
+    ]
+
+    result = evaluator.eval_expression(condition)
+    if not result:
+        # If condition is false, return just the first table with proper column casting
+        return exp.select(*projections).from_(tables[0])
+
+    return reduce(
+        lambda a, b: a.union(b, distinct=kind == "DISTINCT"),  # type: ignore
+        [exp.select(*projections).from_(t) for t in tables],
+    )
+
+
+@macro()
 def haversine_distance(
     _: MacroEvaluator,
     lat1: exp.Expression,
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1012,10 +1012,9 @@ def union_if(
     evaluator: MacroEvaluator,
     condition: exp.Expression,
     type_: exp.Literal = exp.Literal.string("ALL"),
-    *tables: exp.Table,
+    *tables: exp.Table | exp.Query,
 ) -> exp.Query:
-    """Returns a UNION of the given tables if the condition is true, otherwise returns just the first table.
-       The behaviour remains of only choosing columns that have the same name and type.
+    """Returns a UNION of the given tables or queries if the condition is true, otherwise returns just the first.
 
     Example:
         >>> from sqlglot import parse_one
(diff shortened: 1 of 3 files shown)


sqlmesh/core/macros.py
```diff
@@ -1007,6 +1007,53 @@ def union(
     )
 
 
+@macro()
+def union_if(
+    evaluator: MacroEvaluator,
+    condition: exp.Expression,
+    type_: exp.Literal = exp.Literal.string("ALL"),
+    *tables: exp.Table,
+) -> exp.Query:
```
georgesittas, line 1016: Can't we extend the existing `@UNION` macro somehow to make this work? Having a separate `_if` variant is not great.

Perhaps you could extend the argument list to make the `condition` argument [keyword-only](https://peps.python.org/pep-3102/). Alternatively, you could make it the first argument in the list, which is consistent with other macros, and try to migrate existing models or make it backwards-compatible by checking if the first argument is `'all'` or `'distinct'` to preserve the old behavior.

cc @tobymao

sqlmesh/core/macros.py
```diff
@@ -1007,6 +1007,53 @@ def union(
...
+            lambda a, b: a & b,  # type: ignore
+            (evaluator.columns_to_types(table).items() for table in tables),
+        )
+    }
+
+    projections = [
+        exp.cast(column, type_, dialect=evaluator.dialect).as_(column)
+        for column, type_ in evaluator.columns_to_types(tables[0]).items()
+        if column in columns
+    ]
```
VaggelisD, line 1044: Looks like these and the return statement are repeated in `union` and `union_if`, maybe we can extract them in a helper?

sqlmesh/core/macros.py
themisvaltinos: I was initially thinking of simply extending the existing @UNION macro as well, but ended up leaning against it for since you'd either need to compromise a bit on the parameter structure, as you mentioned, or require migration of existing models and for people to learn the new way. 

Also, looking back at the original issue, it seems there was a need to support both tables and select statements as inputs. That adds complexity, and having a dedicated variant might make that clearer and easier to manage in practice, even if it's not ideal from a naming standpoint. Though for the tables the union logic for tables could still be reused with a helper. But yes happy to go down whatever route sounds more sound.

tobymao: what does extending union look like?

tobymao: we could do the first one, in a backwards compatible way, and make it check if it's a string literal right?

tobymao: it only accepts all or distinct right now

themisvaltinos: ok thanks I will alter it this way

sqlmesh/core/macros.py
```diff
@@ -970,37 +970,70 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
 @macro()
 def union(
     evaluator: MacroEvaluator,
-    type_: exp.Literal = exp.Literal.string("ALL"),
-    *tables: exp.Table,
+    *args: exp.Expression,
```
georgesittas, line 973: I have a hunch that this change may not be correct. Type hints in macros have additional semantics, because the evaluator uses them to [coerce](https://github.com/TobikoData/sqlmesh/blob/5b2fa5b2fb94f5f6870de33649c340a4066651d9/sqlmesh/core/macros.py#L1375) the arguments into the specified type.

In this case, since the previous type hint was `exp.Table`, that means the evaluator would try to coerce the arguments into `Table` references:

```python
>>> from sqlglot.schema import MappingSchema
>>> sql = "@UNION('distinct', foo, bar)"
>>> MacroEvaluator(schema=MappingSchema({"foo": {"a": "int"}, "bar": {"a": "int"}})).transform(parse_one(sql))
> /.../sqlmesh/core/macros.py(991)union()
-> columns = {
(Pdb) p tables
(Table(
  this=Identifier(this=foo, quoted=False)), Table(
  this=Identifier(this=bar, quoted=False)))
```

Changing this to `Expression` means the arguments now won't be coerced into `Table`s and this may have unintended side-effects. For instance, it may result in producing a corrupted AST.

themisvaltinos: thanks @georgesittas I revised to coerce them with `to_table` and `table_` into tables based on their type, not sure if there is a better way to achieve this

also regarding the union type, just pointing that now this is optional so these are also allowed: 
- `@union(<condition>, table_1, table_2)`
- `@union(table_1, table_2)`
 and in that case uses the default `ALL`. is this ok or should I enforce the union type to be given which I believe was the previous behaviour?

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
+        >>> MacroEvaluator(schema=MappingSchema({"foo": {"a": "int", "b": "string", "c": "string"}, "bar": {"c": "string", "a": "int", "b": "int"}})).transform(parse_one(sql)).sql()
+        'SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM foo UNION SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM bar'
     """
+
+    if not args:
+        raise SQLMeshError("At least one table is required for UNION.")
+
+    arg_idx = 0
+    # Check for condition
+    condition = evaluator.eval_expression(args[arg_idx])
```
georgesittas, line 1001: I think if you do `@UNION(foo, bar)` then this `eval_expression` will fail, won't it?

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
+    if isinstance(condition, bool):
+        arg_idx += 1
+        if arg_idx >= len(args):
```
georgesittas, line 1004: [Nit] store `len(args)` at the beginning and reuse it.

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
         >>> sql = "@UNION('distinct', foo, bar)"
         >>> MacroEvaluator(schema=MappingSchema({"foo": {"a": "int", "b": "string", "c": "string"}, "bar": {"c": "string", "a": "int", "b": "int"}})).transform(parse_one(sql)).sql()
         'SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM foo UNION SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM bar'
+        >>> sql = "@UNION(True, 'distinct', foo, bar)"
+        >>> MacroEvaluator(schema=MappingSchema({"foo": {"a": "int", "b": "string", "c": "string"}, "bar": {"c": "string", "a": "int", "b": "int"}})).transform(parse_one(sql)).sql()
+        'SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM foo UNION SELECT CAST(a AS INT) AS a, CAST(c AS TEXT) AS c FROM bar'
     """
+
+    if not args:
+        raise SQLMeshError("At least one table is required for UNION.")
```
georgesittas, line 997:
```suggestion
        raise SQLMeshError("At least one table is required for the @UNION macro.")
```

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
+
+    arg_idx = 0
+    # Check for condition
+    condition = evaluator.eval_expression(args[arg_idx])
+    if isinstance(condition, bool):
+        arg_idx += 1
+        if arg_idx >= len(args):
+            raise SQLMeshError("Expected more arguments after condition.")
```
georgesittas, line 1005:
```suggestion
            raise SQLMeshError("Expected more arguments after the condition of the `@UNION` macro.")
```

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
+
+    # Check for union type
+    type_ = exp.Literal.string("ALL")
+    if arg_idx < len(args) and isinstance(args[arg_idx], exp.Literal):
```
georgesittas, line 1009: Don't we know by this point that `arg_idx` must be less than `len(args)`?

sqlmesh/core/macros.py
```diff
@@ -970,23 +970,58 @@ def safe_div(_: MacroEvaluator, numerator: exp.Expression, denominator: exp.Expr
...
         raise SQLMeshError(f"Invalid type '{type_}'. Expected 'ALL' or 'DISTINCT'.")
 
+    # Remaining args should be tables
+    tables = []
+    for table in args[arg_idx:]:
+        if isinstance(table, exp.Column):
+            table = exp.table_(table.this, db=table.args.get("table"), catalog=table.args.get("db"))
+        else:
+            table = exp.to_table(table, dialect=evaluator.dialect)  # type: ignore
+        tables.append(table)
```
georgesittas, line 1023: Should we do something like the following?

```python
tables = [
    exp.to_table(e.sql(evaluator.dialect), dialect=evaluator.dialect) for e in args[arg_idx:]
]
```

georgesittas (commented):
Remember to update docs on the updated API. Also let's test stuff like the `@UNION(foo, bar)` case I mentioned in a comment.

sqlmesh/core/macros.py
georgesittas: I think it's fine not enforcing the union type.

sqlmesh/core/macros.py
themisvaltinos: I have tried it in a project and in that case it is the table name so it won't be boolean. so in the condition will be `foo` unless you're talking of a different edge case. I added also a test for it in the tests

sqlmesh/core/macros.py
themisvaltinos: yes it's not needed as you point out in the other comment

sqlmesh/core/macros.py
themisvaltinos: yes it can't be

sqlmesh/core/macros.py
georgesittas: sounds good, I didn't test this properly so it was a false alarm, thanks for double-checking

georgesittas (approved):
Can you test whether this is breaking / needs a migration?

PR #4341, opened 2025-05-08
Feat!: Ensure metadata snapshots with modified audits are still audited 

### LINKED ISSUE #3970: Adding an audit is a metadata change and it doesn't run on plan application
In which case the audit's behavior is not validating when testing in dev. A problematic workflow may look like following:
 
1. The user adds a new audit to an existing model and references a non-existing projection in its query
2. SQLMesh detects a metadata change and applies the plan without evaluating anything
3. The user deploys to production where the audit fails during the next `sqlmesh run` execution

We need to come up with a way to verify added audits without evaluating associated models when applying the plan in a dev environment.

VaggelisD:
Fixes #3970

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -3256,10 +3256,12 @@ class DatabricksMagicConsole(CaptureTerminalConsole):
         num_audits_failed: int,
         audit_only: bool = False,
     ) -> None:
+        view_name, loaded_batches = self.evaluation_batch_progress[snapshot.snapshot_id]
+
         if audit_only:
+            print(f"Completed Auditing {view_name}")
             return
 
-        view_name, loaded_batches = self.evaluation_batch_progress[snapshot.snapshot_id]
         total_batches = self.evaluation_model_batch_sizes[snapshot]
 
         loaded_batches += 1
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -350,11 +350,14 @@ class Console(
         batched_intervals: t.Dict[Snapshot, Intervals],
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
+        audit_only: bool = False,
     ) -> None:
-        """Indicates that a new snapshot evaluation progress has begun."""
+        """Indicates that a new snapshot evaluation/auditing progress has begun."""
 
     @abc.abstractmethod
-    def start_snapshot_evaluation_progress(self, snapshot: Snapshot) -> None:
+    def start_snapshot_evaluation_progress(
+        self, snapshot: Snapshot, audit_only: bool = False
+    ) -> None:
         """Starts the snapshot evaluation progress."""
 
     @abc.abstractmethod
@@ -366,6 +369,7 @@ class Console(
         duration_ms: t.Optional[int],
         num_audits_passed: int,
         num_audits_failed: int,
+        audit_only: bool = False,
     ) -> None:
         """Updates the snapshot evaluation progress."""
 
@@ -507,10 +511,13 @@ class NoopConsole(Console):
         batched_intervals: t.Dict[Snapshot, Intervals],
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
+        audit_only: bool = False,
     ) -> None:
         pass
 
-    def start_snapshot_evaluation_progress(self, snapshot: Snapshot) -> None:
+    def start_snapshot_evaluation_progress(
+        self, snapshot: Snapshot, audit_only: bool = False
+    ) -> None:
         pass
 
     def update_snapshot_evaluation_progress(
@@ -521,6 +528,7 @@ class NoopConsole(Console):
         duration_ms: t.Optional[int],
         num_audits_passed: int,
         num_audits_failed: int,
+        audit_only: bool = False,
     ) -> None:
         pass
 
@@ -891,11 +899,12 @@ class TerminalConsole(Console):
         batched_intervals: t.Dict[Snapshot, Intervals],
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
+        audit_only: bool = False,
     ) -> None:
-        """Indicates that a new snapshot evaluation progress has begun."""
+        """Indicates that a new snapshot evaluation/auditing progress has begun."""
         if not self.evaluation_progress_live:
             self.evaluation_total_progress = make_progress_bar(
-                "Executing model batches", self.console
+                "Executing model batches" if not audit_only else "Auditing models", self.console
             )
 
             self.evaluation_model_progress = Progress(
@@ -916,8 +925,9 @@ class TerminalConsole(Console):
             batch_sizes = {
                 snapshot: len(intervals) for snapshot, intervals in batched_intervals.items()
             }
+            message = "Executing" if not audit_only else "Auditing"
             self.evaluation_total_task = self.evaluation_total_progress.add_task(
-                "Executing models...", total=sum(batch_sizes.values())
+                f"{message} models...", total=sum(batch_sizes.values())
             )
 
             # determine column widths
@@ -943,7 +953,9 @@ class TerminalConsole(Console):
             self.environment_naming_info = environment_naming_info
             self.default_catalog = default_catalog
 
-    def start_snapshot_evaluation_progress(self, snapshot: Snapshot) -> None:
+    def start_snapshot_evaluation_progress(
+        self, snapshot: Snapshot, audit_only: bool = False
+    ) -> None:
         if self.evaluation_model_progress and snapshot.name not in self.evaluation_model_tasks:
             display_name = snapshot.display_name(
                 self.environment_naming_info,
@@ -951,7 +963,7 @@ class TerminalConsole(Console):
                 dialect=self.dialect,
             )
             self.evaluation_model_tasks[snapshot.name] = self.evaluation_model_progress.add_task(
-                f"Evaluating {display_name}...",
+                f"{'Evaluating' if not audit_only else 'Auditing'} {display_name}...",
                 view_name=display_name,
                 total=self.evaluation_model_batch_sizes[snapshot],
             )
@@ -964,6 +976,7 @@ class TerminalConsole(Console):
         duration_ms: t.Optional[int],
         num_audits_passed: int,
         num_audits_failed: int,
+        audit_only: bool = False,
     ) -> None:
         """Update the snapshot evaluation progress."""
         if (
@@ -1003,7 +1016,7 @@ class TerminalConsole(Console):
                     self.evaluation_column_widths["duration"]
                 )
 
-                msg = f"{batch} {display_name}   {annotation}   {duration}".replace(
+                msg = f"{f'{batch} ' if not audit_only else ''}{display_name}   {annotation}   {duration}".replace(
                     self.AUDIT_PASS_MARK, self.GREEN_AUDIT_PASS_MARK
                 )
 
@@ -1015,7 +1028,10 @@ class TerminalConsole(Console):
 
             model_task_id = self.evaluation_model_tasks[snapshot.name]
             self.evaluation_model_progress.update(model_task_id, refresh=True, advance=1)
-            if self.evaluation_model_progress._tasks[model_task_id].completed >= total_batches:
+            if (
+                self.evaluation_model_progress._tasks[model_task_id].completed >= total_batches
+                or audit_only
+            ):
                 self.evaluation_model_progress.remove_task(model_task_id)
 
     def stop_evaluation_progress(self, success: bool = True) -> None:
@@ -3208,6 +3224,7 @@ class DatabricksMagicConsole(CaptureTerminalConsole):
         batched_intervals: t.Dict[Snapshot, Intervals],
         environment_naming_info: EnvironmentNamingInfo,
         default_catalog: t.Optional[str],
+        audit_only: bool = False,
     ) -> None:
         self.evaluation_model_batch_sizes = {
             snapshot: len(intervals) for snapshot, intervals in batched_intervals.items()
@@ -3215,7 +3232,9 @@ class DatabricksMagicConsole(CaptureTerminalConsole):
         self.evaluation_environment_naming_info = environment_naming_info
         self.default_catalog = default_catalog
 
-    def start_snapshot_evaluation_progress(self, snapshot: Snapshot) -> None:
+    def start_snapshot_evaluation_progress(
+        self, snapshot: Snapshot, audit_only: bool = False
+    ) -> None:
         if not self.evaluation_batch_progress.get(snapshot.snapshot_id):
             display_name = snapshot.display_name(
                 self.evaluation_environment_naming_info,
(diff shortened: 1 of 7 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1069,6 +1069,34 @@ def _data_hash_values(self) -> t.List[str]:
 
         return data  # type: ignore
 
+    def _audit_metadata(self) -> t.List[str]:
```
izeigerman, line 1072: `_audit_metadata_hash_values` to be consistent with `_data_hash_values`

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +551,72 @@ def _restatement_intervals_across_all_environments(
...
+        for to_be_audited_snapshot, previous_snapshot in to_be_audited_snapshots:
+            parent_snapshots = parent_snapshots_by_name(to_be_audited_snapshot, new_snapshots)
+
+            # The previous snapshot is the snapshot before the metadata change
+            # and contains the latest intervals that we should use for the new audit
+            for interval in previous_snapshot.intervals:
+                start, end = interval
+
+                try:
+                    scheduler._audit_snapshot(
```
izeigerman, line 598: I'm not sure I like this API. Can we somehow reuse the entire scheduler machinery as is by supporting `audit_only` mode in `run` API? With `audit_only` we'll bypass evaluate and interval recording but do everything else. We can also use the `restatements` argument to remove relevant intervals so that they can be reprocessed again.

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +551,72 @@ def _restatement_intervals_across_all_environments(
...
+                or not snapshot.is_model
+            ):
+                continue
+
+            previous_snapshot_id = snapshot.previous_version.snapshot_id(snapshot.name)
+            previous_snapshot = self.state_sync.get_snapshots([previous_snapshot_id])[
+                previous_snapshot_id
+            ]
+
+            new_audits = snapshot.model._audit_metadata()
```
izeigerman, line 575: I suggest having a proper top level method for this `audit_metadata_hash`

sqlmesh/core/plan/evaluator.py
VaggelisD: Yeah, that was my first attempt but felt like I'd be hacking `run()` to switch off all lines but the audit, so I went with my current approach. I can give it a shot in a new commit and we can compare the two. 

> We can also use the restatements argument to remove relevant intervals so that they can be reprocessed again.

Sorry if I'm not understanding, but won't this add perf regressions to this PR? Currently, we retrieve the intervals from the latest snapshot so we can run the audits **without** having to evaluate models, compute intervals, modify the state sync etc. 

Afaict, with restatements we can simply pass in `[model_start, today]` for each snapshot and get this for free, but in exchange we have to remove + recompute + restore the intervals all over again, only to get the same interval set back, right?

sqlmesh/core/plan/evaluator.py
VaggelisD: So you'd like a top-level method in the model definition that returns both the list and it's hash, right? 

```Python3
# core.model.definition

def audit_metadata_hash(self):
  audits = self._audit_metadata()
  return audits, hash_data(audits)
```

So then the highlighted section would be:

```Python3
_, previous_audits_hash = previous_snapshot.model.audit_metadata_hash()
new_audits, new_audits_hash = snapshot.model.audit_metadata_hash()

if (previous_audits_hash != new_audits_hash) and new_audits:
  ...
```

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+            [
+                s.previous_version.snapshot_id(s.name)
+                for s in metadata_snapshots
+                if s.previous_version
+            ]
+        ).values()
+
+        # Step 3: Compare the audit metadata hashes to determine if there was a change in the audits field
+        to_be_audited_snapshots = {}
+        for snapshot, previous_snapshot in zip(metadata_snapshots, previous_snapshots):
```
georgesittas, line 575: Based on L569, it seems like `previous_snapshots` may skip some snapshots in `metadata_snapshots`, i.e., if they don't have a `previous_version`. Hence, I wonder if pairing the two using `zip` here is correct. Could we end up comparing unrelated snapshots?

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+            new_audits, new_audits_hash = snapshot.model.audit_metadata_hash()
+            _, previous_audit_hash = previous_snapshot.model.audit_metadata_hash()
+
+            if previous_audit_hash != new_audits_hash and new_audits:
```
georgesittas, line 579: [Nit] Let's flip the condition so `new_audits` is first?

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+        previous_snapshots = self.state_sync.get_snapshots(
+            [
+                s.previous_version.snapshot_id(s.name)
+                for s in metadata_snapshots
+                if s.previous_version
+            ]
+        ).values()
+
+        # Step 3: Compare the audit metadata hashes to determine if there was a change in the audits field
+        to_be_audited_snapshots = {}
```
georgesittas, line 574: [Nit] consider renaming this to something like `audit_intervals_by_snapshot`.

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+            plan.start,
+            plan.end,
+            execution_time=plan.execution_time,
+            restatements=to_be_audited_snapshots,
+            end_bounded=plan.end_bounded,
+            interval_end_per_model=plan.interval_end_per_model,
+        )
+
+        if completion_status.is_failure:
+            raise PlanError("Plan application failed.")
```
georgesittas, line 601: What does failing actually look like? Do we have any signal about what caused the failure other than this message? Seems cryptic on its own.

sqlmesh/core/scheduler.py
```diff
@@ -149,8 +149,9 @@ def evaluate(
         execution_time: TimeLike,
         deployability_index: DeployabilityIndex,
         batch_index: int,
+        environment_naming_info: t.Optional[EnvironmentNamingInfo] = None,
         **kwargs: t.Any,
-    ) -> t.Tuple[t.List[AuditResult], t.List[AuditError]]:
+    ) -> t.List[AuditResult]:
```
georgesittas, line 154: Why did this change? Have we checked whether this affects other call sites of `evaluate`?

VaggelisD: The second list was returned to print the non-blocking audit failures outside of evaluate, but imo this seemed a bit leaking so I abstracted it all away in `_audit_snapshot`.

sqlmesh/core/plan/evaluator.py
VaggelisD: The actual logging is abstracted inside `_audit_snapshot` in the scheduler, and would look like:

```SQL

Models:
└── Metadata Updated:
    └── vaggelis.model1

---                                                                                                                                                                             
                                                                                                                                                                                
+++                                                                                                                                                                             
                                                                                                                                                                                
@@ -9,6 +9,9 @@                                                                                                                                                                 
                                                                                                                                                                                
     disable_restatement FALSE,                                                                                                                                                 
     on_destructive_change 'ERROR'                                                                                                                                              
   ),                                                                                                                                                                           
+  audits (UNIQUE_COMBINATION_OF_COLUMNS('columns' = (                                                                                                                          
+      event_date                                                                                                                                                               
+  ))),                                                                                                                                                                         
   allow_partials TRUE                                                                                                                                                          
 )                                                                                                                                                                              
 SELECT                                                                                                                                                                         

Metadata Updated: vaggelis.model1

Failed models

  "db"."vaggelis"."model1"

    'unique_combination_of_columns' audit error: 1 row failed
      
      Audit arguments
        columns := (event_date)
      
      Audit query
        SELECT ("event_date") AS "event_date" FROM (SELECT * FROM
        "db"."sqlmesh__vaggelis"."vaggelis__model1__440935667" AS "vaggelis__model1__440935667" WHERE
        "event_date" BETWEEN CAST('2025-05-01' AS DATE) AND CAST('2025-05-04' AS DATE)) AS "_q_0" WHERE TRUE
        GROUP BY ("event_date") HAVING COUNT(*) > 1
      
      
    ...
    raise PlanError("Plan application failed.")
sqlmesh.utils.errors.PlanError: Plan application failed.

```

The `PlanError` is used to halt the plan application.

sqlmesh/core/plan/evaluator.py
VaggelisD: This was mostly to satisfy mypy, it shouldn't be possible to **not** have previous versions if we're currently handling metadata snapshots, right? 

I can break that list comprehension apart and add asserts if it makes the code more clear

sqlmesh/core/model/definition.py
```diff
@@ -1069,6 +1069,38 @@ def _data_hash_values(self) -> t.List[str]:
...
+                        str(audit.skip),
+                        str(audit.blocking),
+                    ]
+                )
+
+        return metadata
+
+    def audit_metadata_hash(self) -> t.Tuple[t.List[str], str]:
+        hash_values = self._audit_metadata_hash_values()
+        return hash_values, hash_data(hash_values)
```
izeigerman, line 1102: Why does this return hash values?

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+                s.previous_version.snapshot_id(s.name)
+                for s in metadata_snapshots
+                if s.previous_version
+            ]
+        ).values()
+
+        # Step 3: Compare the audit metadata hashes to determine if there was a change in the audits field
+        to_be_audited_snapshots = {}
+        for snapshot, previous_snapshot in zip(metadata_snapshots, previous_snapshots):
+            new_audits, new_audits_hash = snapshot.model.audit_metadata_hash()
```
izeigerman, line 576: Is there a different way to check for audits?

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+            _, previous_audit_hash = previous_snapshot.model.audit_metadata_hash()
+
+            if previous_audit_hash != new_audits_hash and new_audits:
+                snapshot_start = min(i[0] for i in snapshot.intervals)
+                snapshot_end = max(i[1] for i in snapshot.intervals)
```
izeigerman, line 581: I thought we decided to use plan's start / end for this and not the whole snapshot's history.

sqlmesh/core/plan/evaluator.py
VaggelisD: We could also do `snapshot.model.audits` and not return the `hash_values` at all

sqlmesh/core/plan/evaluator.py
izeigerman: Make sure to use `snapshot.get_removal_interval` to compute the correct interval based on plan start / end. With `is_preview=True`.

VaggelisD: I remember bringing up the point that if we don't test the entirety of the historical data, then we might have a "false positive" audit. For example:
- Assume the model's intervals span `[01-01-2025, 01-05-2025]` (5 days)
- The plan start is set at `01-03-2025`
- We add a new `not_null(col)` audit. If we run it for `[01-03-2025, 01-05-2025]`, then we don't know if the audit holds for `[01-01-2025, 01-03-2025)` i.e if there are `NULL` records in that range.

Is that right?

sqlmesh/core/plan/evaluator.py
```diff
@@ -545,6 +547,59 @@ def _restatement_intervals_across_all_environments(
...
+
+        # Step 4: If there are any snapshots to be audited, we'll reuse the scheduler's
+        # internals to audit them by utilizing the restatement logic
+        scheduler = self.create_scheduler(snapshots.values())
+        completion_status = scheduler.audit(
+            plan.environment,
+            plan.start,
+            plan.end,
+            execution_time=plan.execution_time,
+            restatements=to_be_audited_snapshots,
```
izeigerman, line 595: I think `restatements` is not a valid concept in a context of an audit. Should we just remove this and then remove intervals inside `def audit` itself based on passed start and end?

sqlmesh/core/scheduler.py
```diff
@@ -426,11 +359,12 @@ def run_merged_intervals(
 
         batched_intervals = self.batch_intervals(merged_intervals, deployability_index)
 
-        self.console.start_evaluation_progress(
-            batched_intervals,
-            environment_naming_info,
-            self.default_catalog,
-        )
+        if not audit_only:
+            self.console.start_evaluation_progress(
```
izeigerman, line 363: Why don't we want the evaluation progress? I think we still do but with the caveat that it's audit only

VaggelisD: This has more moving parts than I wanted, e.g the `executing batches` and `evaluating models` progress bars. I can pushdown `audit_only` and further dissect the logic then.

tests/core/test_context.py
```diff
@@ -2071,3 +2072,42 @@ def test_audit():
     context.plan(no_prompts=True, auto_apply=True)
 
     assert context.audit(models=["dummy"], start="2020-01-01", end="2020-01-01") is True
+
+
+@use_terminal_console
+def test_audits_running_on_metadata_changes(tmp_path: pathlib.Path):
```
izeigerman, line 2078: Let's also add an integration test maybe?

VaggelisD: Is there a different scenario that you were thinking of?

sqlmesh/core/plan/evaluator.py
VaggelisD: Note, removing the intervals inside `audit` doesn't work because `_run_or_audit` will refresh them again. For this matter, I renamed `restatements` to `remove_intervals` on the latter, iirc this was one approach we had discussed to distinguish the actual restatements.

sqlmesh/core/plan/evaluator.py
VaggelisD: Synced on this, the plan start/end must be respected, it's the users responsibility if this case is hit ^

sqlmesh/core/model/definition.py
VaggelisD: Switched to returning only `hash_data(hash_values)`, we can also see if there are audits through `snapshot.model.audits` so returning `hash_values` is not needed.

sqlmesh/core/console.py
```diff
@@ -3183,7 +3202,11 @@ def update_snapshot_evaluation_progress(
         duration_ms: t.Optional[int],
         num_audits_passed: int,
         num_audits_failed: int,
+        audit_only: bool = False,
     ) -> None:
+        if audit_only:
+            return
```
izeigerman, line 3208: How do we update progress when it's audit only?

izeigerman: Or rather why don't we update it for databricks?

tests/core/test_context.py
izeigerman: Actually, this looks like an integration test already. Should we move it there?

izeigerman (approved):
One question, LGTM otherwise

sqlmesh/core/console.py
VaggelisD: Looks like I was missing a print there, added it. Thanks

PR #4345, opened 2025-05-08
fix(vscode): filter go to definitions by position

benfdking:
- before we were returning the references for a whole file
- now we filter by the position of the request

### DIFF
--- a/sqlmesh/lsp/main.py
+++ b/sqlmesh/lsp/main.py
@@ -14,7 +14,10 @@ from sqlmesh.core.linter.definition import AnnotatedRuleViolation
 from sqlmesh.lsp.completions import get_sql_completions
 from sqlmesh.lsp.context import LSPContext, ModelTarget
 from sqlmesh.lsp.custom import ALL_MODELS_FEATURE, AllModelsRequest, AllModelsResponse
-from sqlmesh.lsp.reference import get_model_definitions_for_a_path, filter_references_by_position
+from sqlmesh.lsp.reference import (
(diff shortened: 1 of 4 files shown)


sqlmesh/lsp/main.py
```diff
@@ -186,12 +186,18 @@ def goto_definition(
...
+                # Get all possible references
                 references = get_model_definitions_for_a_path(
                     self.lsp_context, params.text_document.uri
                 )
                 if not references:
                     return []
 
+                # Filter references by cursor position
+                filtered_references = filter_references_by_position(references, params.position)
+                if not filtered_references:
```
themisvaltinos, line 198: [nit]  if you don't need special handling for the empty case lines 198-199 are a bit redundant since you're already will be returning the empty list

sqlmesh/lsp/reference.py
```diff
@@ -14,6 +14,39 @@ class Reference(PydanticModel):
...
+    filtered_references = []
+
+    for reference in references:
+        # Check if position is within the reference range
+        range_start = reference.range.start
+        range_end = reference.range.end
+
+        # Position is within range if it's after or at start and before or at end
+        if (
+            range_start.line < position.line
```
themisvaltinos, line 39: references would never have invalid or reversed ranges (start after end) right

benfdking: nope they should be in correct order, that's an inheriant assumption I make in range.

Issue #4232, opened 2025-04-23
Python model with DataFrame and Columns, matched by name or order


wahmedswl:
Hi!

I am using SQLMesh with Python models. The columns order in the @model annotation and DataFrame is different and its throwing error

```
Failed models

  "postgres"."data_pipeline"."external_patient_statement_oibo"

    InvalidDatetimeFormat:
      invalid input syntax for type timestamp: "BR68282"

Error: Plan application failed.
```

If I have same columns and DataFrame order, it works. Does it map columns via Index or Name as I have same column name but different order. I am using SQLMesh with Postgres.

Thanks

erindru:
SQLMesh uses the `columns` that you define on the model to create the target table (and also figure out column level lineage). This is because these things happen at "creation" time (where we create empty tables to house the output of the model) vs "evaluation" time where we actually run the Python code of the model.

The expectation is that the Python model emits DataFrame's that match this column specification. SQLMesh will not reorder the DataFrame or adjust it in any way, it currently just mashes the returned DataFrame's into the table it created based on what you defined for `columns`.

This could perhaps be documented better but as you discovered, the two need to match and the duplication is necessary because SQLMesh cannot statically infer the columns of the DataFrame at load / creation time.

wahmedswl:
@erindru I guess you confused something. I don't have any problem with the columns duplication at the model definition and the dataframe. What i am seeing is, if the dataframe has different column order; it crashed.

Model and DataFrame; if have same columns it should work regardless of the columns order.

wahmedswl:
This is my model definition right now (stripped)

```
from datetime import datetime
import json
import typing as t
import xml.etree.ElementTree as ET
import numpy as np
import pandas as pd
from typing import Dict, Any

from sqlmesh import ExecutionContext, model
from sqlmesh.core.model import ModelKindName


@model(
    "data_pipeline.external_patient_statement_oibo",
    kind=dict(name=ModelKindName.FULL),
    columns={
        "account_number": "string",
        "statement_number": "string",
        "statement_date": "datetime",
        "remit_message": "string",
        "announcement": "string",
        "additional_message": "string",
        "patient_credit": "float",
        "total_patient_due": "float",
        "total_insurance_due": "float"
    },
    cron="@daily",
    grain="(account_number, statement_number)",
)
def execute(
    context: ExecutionContext,
    start: datetime,
    end: datetime,
    execution_time: datetime,
    **kwargs: t.Any,
):
[...]

Issue #3915, opened 2025-02-27
MSSQL: Order change in model columns definition crashed model plan


ebremst3dt:
Version used: 0.159.0
Engine: MSSQL

Example:
        
```
@model(
    columns={'chgdat': 'varchar(max)',
 'chgusr': 'varchar(max)',
 'compny': 'varchar(max)',
 'credat': 'varchar(max)',
 'creusr': 'varchar(max)',
 'digcod': 'varchar(max)',
 'dignam': 'varchar(max)',
 'migcod': 'varchar(max)',
 'sigcod': 'varchar(max)',
 'srtnam': 'varchar(max)',
 'srtnum': 'varchar(max)',
 'txtdsc': 'varchar(max)',
 'data_modified': 'date',
  'source_catalog': 'varchar(max)'
 
  }
```

In this model definition, the order is
1. data_modified
2. source_catalog

In my query it was:
1. source_catalog
2. data_modified

This results in MSSQLDatabaseException: (0, b'Unknown Error)

Having the same order makes the query run just fine.

I guess that the query informs how the dataframe is created, im just thinking sqlmesh should be able to pair the key values anyway, regardless of the order.

izeigerman:
@treysp can you please have a look

erindru:
This is fixed as of [v0.182.0](https://github.com/TobikoData/sqlmesh/releases/tag/v0.182.0)

PR #4348, opened 2025-05-09
Fix!: Allow python models to emit DataFrame's with a different column order 

erindru:
... than what is declared in `@model`. Addresses #3915 and #4232 

Prior to this PR, on many adapters the column order in the `DataFrame` emitted by a Python model had to match what was defined for `columns`.

This is because when we convert the `DataFrame` into a `VALUES(...)` list to insert it into the table, the values were in `DataFrame` order but the columns list was in `columns_to_types` order. If these didnt match, then we tried to insert the wrong value into the wrong column.

This PR adjusts the DataFrame to match `columns_to_types` before converting it to a `VALUES(...)` list.

**This PR fixes broken behaviour and thus may have unintended consequences**

Previously, if you had something like:
```
@model (
   "foo",
   "columns": { "a": "int", "b": "int" }
)
def execute(context: ExecutionContext, **kwargs: t.Any) -> pd.DataFrame:
   pd.DataFrame({"b": [1], "a": [2]})
```

Then this wouldnt fail because the column types match. So we would write this as `a=1` and `b=2` instead of what the user emitted in the DataFrame, `a=2` and `b=1`. I really hope nobody was relying on this

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -285,10 +285,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -285,10 +285,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          #filters:
-          #  branches:
-          #    only:
-          #      - main
+          filters:
+            branches:
+              only:
+                - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,9 +40,9 @@ athena = ["PyAthena[Pandas]"]
 azuresql = ["pymssql"]
 bigquery = [
     "google-cloud-bigquery[pandas]",
-    "google-cloud-bigquery-storage"
+    "google-cloud-bigquery-storage",
+    "bigframes>=1.32.0"
 ]
-bigframes = ["bigframes>=1.32.0"]
 clickhouse = ["clickhouse-connect"]
 databricks = ["databricks-sql-connector[pyarrow]"]
 dev = [
@@ -107,8 +107,7 @@ slack = ["slack_sdk"]
 snowflake = [
     "cryptography",
     "snowflake-connector-python[pandas,secure-local-storage]",
-    # as at 2024-08-05, snowflake-snowpark-python is only available up to Python 3.11
-    "snowflake-snowpark-python; python_version<'3.12'",
+    "snowflake-snowpark-python",
 ]
 trino = ["trino"]
 web = [
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -246,7 +246,12 @@ class EngineAdapter:
         assert isinstance(df, pd.DataFrame)
         num_rows = len(df.index)
         batch_size = sys.maxsize if batch_size == 0 else batch_size
+
+        # we need to ensure that the order of the columns in columns_to_types columns matches the order of the values
+        # they can differ if a user specifies columns() on a python model in a different order than what's in the DataFrame's emitted by that model
+        df = df[list(columns_to_types.keys())]
         values = list(df.itertuples(index=False, name=None))
+
         return [
             SourceQuery(
                 query_factory=partial(
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -249,7 +249,7 @@ class EngineAdapter:
 
         # we need to ensure that the order of the columns in columns_to_types columns matches the order of the values
         # they can differ if a user specifies columns() on a python model in a different order than what's in the DataFrame's emitted by that model
-        df = df[list(columns_to_types.keys())]
+        df = df[list(columns_to_types)]
         values = list(df.itertuples(index=False, name=None))
 
         return [
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -218,10 +218,13 @@ class MSSQLEngineAdapter(
             # as later calls.
             if not self.table_exists(temp_table):
                 columns_to_types_create = columns_to_types.copy()
-                self._convert_df_datetime(df, columns_to_types_create)
+                ordered_df = df[
+                    list(columns_to_types_create.keys())
+                ]  # reorder DataFrame so it matches columns_to_types
+                self._convert_df_datetime(ordered_df, columns_to_types_create)
                 self.create_table(temp_table, columns_to_types_create)
                 rows: t.List[t.Tuple[t.Any, ...]] = list(
-                    df.replace({np.nan: None}).itertuples(index=False, name=None)  # type: ignore
+                    ordered_df.replace({np.nan: None}).itertuples(index=False, name=None)  # type: ignore
                 )
                 conn = self._connection_pool.get()
                 conn.bulk_copy(temp_table.sql(dialect=self.dialect), rows)
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -219,7 +219,7 @@ class MSSQLEngineAdapter(
             if not self.table_exists(temp_table):
                 columns_to_types_create = columns_to_types.copy()
                 ordered_df = df[
-                    list(columns_to_types_create.keys())
+                    list(columns_to_types_create)
                 ]  # reorder DataFrame so it matches columns_to_types
                 self._convert_df_datetime(ordered_df, columns_to_types_create)
                 self.create_table(temp_table, columns_to_types_create)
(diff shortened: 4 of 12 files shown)


.circleci/continue_config.yml
```diff
@@ -285,10 +285,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru, line 288: TODO: revert prior to merge. I wanted to doublecheck the cloud databases as well because we sometimes override the DataFrame-handling functions in specific engine adapters

tests/core/engine_adapter/test_base.py
```diff
@@ -966,7 +966,7 @@ def test_merge_upsert(make_mocked_engine_adapter: t.Callable, assert_exp_eq):
 def test_merge_upsert_pandas(make_mocked_engine_adapter: t.Callable):
     adapter = make_mocked_engine_adapter(EngineAdapter)
 
-    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    df = pd.DataFrame({"id": [1, 2, 3], "ts": [4, 5, 6], "val": [1, 2, 3]})
```
erindru, line 969: @tobymao these tests were using DataFrame's that beared absolutely no resemblance to the `columns_to_types`, was that intended?

They appeared to rely on the fact that up until now column names were ignored and the values were just output in order. This particular one also had a missing column "val" which was present in `columns_to_types` but missing in the `DataFrame`. As far as I can tell, that should be an error

sqlmesh/core/engine_adapter/base.py
```diff
@@ -246,7 +246,12 @@ def _df_to_source_queries(
         assert isinstance(df, pd.DataFrame)
         num_rows = len(df.index)
         batch_size = sys.maxsize if batch_size == 0 else batch_size
+
+        # we need to ensure that the order of the columns in columns_to_types columns matches the order of the values
+        # they can differ if a user specifies columns() on a python model in a different order than what's in the DataFrame's emitted by that model
+        df = df[list(columns_to_types.keys())]
```
tobymao, line 252: you don't need keys

sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -218,10 +218,13 @@ def query_factory() -> Query:
             # as later calls.
             if not self.table_exists(temp_table):
                 columns_to_types_create = columns_to_types.copy()
-                self._convert_df_datetime(df, columns_to_types_create)
+                ordered_df = df[
+                    list(columns_to_types_create.keys())
```
tobymao, line 222: ditto

tests/core/engine_adapter/integration/__init__.py
```diff
@@ -509,6 +510,12 @@ def create_context(
         config.gateways = {self.gateway: config.gateways[self.gateway]}
 
         gateway_config = config.gateways[self.gateway]
+        if ephemeral_state_connection:
```
erindru, line 513: This allows the changes made [here](https://github.com/TobikoData/sqlmesh/commit/5385d00c71886073631e3f56f96992c9d553b45e) to still exercise using the warehouse as a state connection while allowing other tests that just want to run concurrently without clobbering each others state to do so

tests/core/engine_adapter/integration/test_integration_bigquery.py
```diff
@@ -433,3 +433,51 @@ def test_table_diff_table_name_matches_column_name(ctx: TestContext):
 
     assert row_diff.stats["join_count"] == 1
     assert row_diff.full_match_count == 1
+
+
+def test_bigframe_python_model_column_order(ctx: TestContext, tmp_path: Path):
```
erindru, line 438: This is more or less a duplicate of the same test in `test_integration`, just using BigFrame DataFrame's instead of Pandas.

Rather than adding `bigframe` and `snowpark` variants to the `test_type` fixture which would then generate 5 instances of every test per engine adapter when most tests only need 1 instance, i'm planning to refactor how this works in an upcoming PR

pyproject.toml
```diff
@@ -40,9 +40,9 @@ athena = ["PyAthena[Pandas]"]
 azuresql = ["pymssql"]
 bigquery = [
     "google-cloud-bigquery[pandas]",
-    "google-cloud-bigquery-storage"
+    "google-cloud-bigquery-storage",
+    "bigframes>=1.32.0"
```
erindru, line 44: I needed `bigframes` available in integration tests and decided to move it here for consistency with Snowflake where `snowpark` is installed as part of `pip install sqlmesh[snowflake]`

pyproject.toml
```diff
@@ -107,8 +107,7 @@ slack = ["slack_sdk"]
 snowflake = [
     "cryptography",
     "snowflake-connector-python[pandas,secure-local-storage]",
-    # as at 2024-08-05, snowflake-snowpark-python is only available up to Python 3.11
-    "snowflake-snowpark-python; python_version<'3.12'",
+    "snowflake-snowpark-python",
```
erindru, line 110: I removed the `python_version` constraint because I needed `snowpark` to be available for the new tests, our engine integration tests run on Python 3.12 and also the version limitation has been fixed upstream now

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -288,8 +288,25 @@ def _df_to_source_queries(
...
+                if temp_table.catalog
+                else None
+            )
+
             if is_snowpark_dataframe:
-                df.createOrReplaceTempView(temp_table.sql(dialect=self.dialect, identify=True))  # type: ignore
+                temp_table.set("catalog", database)
+                df_renamed = df.rename(
+                    {
+                        col: exp.to_identifier(col, quoted=True).sql(dialect=self.dialect)
```
tobymao, line 303: could you just do col.sql(dialect=self.dialect, identify=True)

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -279,10 +279,16 @@ def _ensure_pyspark_df(
     ) -> PySparkDataFrame:
         pyspark_df = self.try_get_pyspark_df(generic_df)
         if pyspark_df:
+            if columns_to_types:
+                # ensure Spark dataframe column order matches columns_to_types
+                pyspark_df = pyspark_df.select(*list(columns_to_types))
```
tobymao, line 284: don't think you need a list() call here

tobymao (approved):

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -2732,3 +2734,84 @@ def _use_warehouse_as_state_connection(gateway_name: str, config: Config):
 
     # will throw if one of the migrations produces an error, which can happen if we forget to take quoting or normalization into account
     sqlmesh_context.migrate()
+
+
+def test_python_model_column_order(ctx: TestContext, tmp_path_factory: pytest.TempPathFactory):
```
georgesittas, line 2739: FMI: when / why do we choose `tmp_path_factory` over just `tmp_path`?

georgesittas (approved):

sqlmesh/core/engine_adapter/snowflake.py
georgesittas: `col` is a string here, it doesn't look like this applies.

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -288,8 +288,25 @@ def _df_to_source_queries(
         is_snowpark_dataframe = snowpark and isinstance(df, snowpark.dataframe.DataFrame)
 
         def query_factory() -> Query:
+            # The catalog needs to be normalized before being passed to Snowflake's library functions because they
+            # just wrap whatever they are given in quotes without checking if its already quoted
```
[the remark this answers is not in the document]
georgesittas: Discussed in Slack– the reason is that `createOrReplaceTempView` _does_ normalize the column names, and also respects quoting. I assume this is true for both the dataframe receiving the method call and the method's argument table.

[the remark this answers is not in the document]
erindru: The catalog in particular is treated differently. Our engine adapter [rewrites it on the fly](https://github.com/TobikoData/sqlmesh/blob/223bc8eff0c68fc12e8112b579a1030d498332e7/sqlmesh/core/engine_adapter/snowflake.py#L465) to "fix" it to match what the user specified in the connection config, if it's been normalized based on the source dialect. 

Since we hand things off to the Snowflake library, we have to do it here too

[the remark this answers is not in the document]
georgesittas: Ah, I see, I had forgotten about that. Too much nuance, lol.

sqlmesh/core/engine_adapter/spark.py
erindru: Ah yes, you're right, originally the list wasn't unpacked with `*`.

tests/core/engine_adapter/integration/test_integration.py
erindru: According to the [docs](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-factory-fixture):

- `tmp_path_factory` - a session-level fixture
- `tmp_path` - a function level fixture

I was using `tmp_path_factory` because I was trying to shotgun debug a concurrency issue with tests clobbering each others files and I noticed other tests in this file were using `tmp_path_factory`.

Logically, `tmp_path` is the one to use unless youre trying to share a test directory amongst multiple tests. I've switched this test back to `tmp_path`, it will either work or remind me why I was using `tmp_path_factory`

PR #4351, opened 2025-05-09
feat: add on hover to models on lsp

benfdking:


georgesittas (approved):
I'd change this bit into a named closure and return that instead, lgtm otherwise:

```python
    return lambda r: (
        r.range.start.line < position.line
        or (r.range.start.line == position.line and r.range.start.character <= position.character)
    ) and (
        r.range.end.line > position.line
        or (r.range.end.line == position.line and r.range.end.character >= position.character)
    )
```

PR #854, opened 2023-05-16
Feat: display error tag on top level

mykalmax:
Attempts to handle different error cases, and display error message along with traceback to user for further investigation. Currently  errors  mostly are coming from: working with file system, running/applying plan, evaluating a model (or in general fetching data with plain sql), lineage. 

Create a descriptive payload for errors which are coming form API and display on UI in user friendly way.

Screenshot 2023-05-22 at 8 14 59 AM
Screenshot 2023-05-22 at 8 16 11 AM
Screenshot 2023-05-22 at 8 16 39 AM

PR #4352, opened 2025-05-09
feat(vscode): add custom api call method

benfdking:


sqlmesh/lsp/api.py
```diff
@@ -0,0 +1,70 @@
...
+
+    data: t.List[Model]
+
+    def __init__(self, **data: t.Any) -> None:
+        # Convert datetime objects to strings before passing to parent constructor
+        if "data" in data and isinstance(data["data"], list):
+            for model in data["data"]:
+                if hasattr(model, "details") and model.details:
+                    # Convert datetime fields to None or string to avoid serialization issues
+                    for field in ["stamp", "start", "cron_prev", "cron_next"]:
```
izeigerman, line 54: What are these fields? Only `start` here can actually be datetime

sqlmesh/lsp/api.py
```diff
@@ -0,0 +1,70 @@
...
+
+class ApiResponseGetModels(ApiResponse):
+    """
+    Response from the SQLMesh API for the get_models endpoint.
+    Specifies the data type more precisely as a list of models.
+    """
+
+    data: t.List[Model]
+
+    def __init__(self, **data: t.Any) -> None:
```
izeigerman, line 48: I'd suggest using a factory method instead. Or not inherit from `ApiResponse` and rely on pydantic validators instead.

sqlmesh/lsp/api.py
benfdking: https://github.com/TobikoData/sqlmesh/blob/1b073394b9c5c84f4be4173e9d50b99bd23f3cf5/web/server/models.py#L143

sqlmesh/lsp/api.py
benfdking: 👍

sqlmesh/lsp/api.py
georgesittas: It seems like these were [changed](https://github.com/TobikoData/sqlmesh/pull/854) quite some time ago. I can tell for certain that `stamp` is supposed to be a string ([docs](https://sqlmesh.readthedocs.io/en/stable/concepts/models/overview/?h=stamp#stamp)), so I wonder if the change was an oversight. The cron prev/next do [look](https://github.com/TobikoData/sqlmesh/blob/e773b59a626cc7607930f014f045c78f60add72f/sqlmesh/core/node.py#L103-L127) like datetime.

georgesittas (approved):

izeigerman (approved):

PR #4355, opened 2025-05-09
feat: add code href that links to the rule

benfdking:
- adds an href to the rule in the lsp which makes it possible to show
  a link to in the linting error that can take you to the code


sqlmesh/core/linter/rule.py
```diff
@@ -40,6 +40,34 @@ def violation(self, violation_msg: t.Optional[str] = None) -> RuleViolation:
         """Create a RuleViolation instance for this rule"""
         return RuleViolation(rule=self, violation_msg=violation_msg or self.summary)
 
+    def get_definition_location(self) -> t.Dict[str, t.Any]:
```
benfdking, line 43: This should return a proper type

VaggelisD (approved):

PR #4406, opened 2025-05-14
Chore: Enable tests for native DataFrame implementations

erindru:
In implementing #4348 I noticed a gap in our engine adapter integration tests.

Currently:
 - For each engine adapter integration test, we parameterize it by engine (so it gets called at least once per engine)
 - Many engine adapter methods take a `QueryOrDF`, that is - a SQL Query or some kind of DataFrame. So we also parameterize each test by `query` and `df`
 - However, `df` really only refers to a Pandas DataFrame. At the time the tests were originally written, there was only one other DataFrame implementation - `pyspark`.
 - So we also parameterize each test by `pyspark` as well even though this is only relevant for two engines - Spark and Databricks
 - The net result is, we have parameterized each test 3 times _per engine adapter_ - once for `query`, `df` and `pyspark`. In many cases this isnt needed, which leads to test logic that looks like:
    ```
    if ctx.test_type != "query":
         pytest.skip("Test only needs to run for query")
    ```
 - So we generate a lot of tests that just exist to get skipped, while also forgetting about the other native DataFrame types: `bigframe` (BigQuery) and `snowpark` (Snowflake) which currently have zero test coverage
 - But if we add these to the existing `test_type` fixture to make it parametrized over (`query`, `df`, `pyspark`, `bigframe`, `snowpark`) then we would get 5 versions of each test _per engine adapter_ - just to skip most of them because `bigframe` is only relevant for BigQuery and `snowpark` is only relevant for Snowflake

So I decided to do some refactoring.

This PR changes things to:
 - Refactor how the engines themselves are defined to prevent duplication
   - For example, rather than defining `trino` 4 times, once per supported catalog - we define it once and generate a definition per supported catalog
   - If an engine has a native DataFrame type, it gets recorded on the engine definition so we only generate a test parameter for that engine rather than for every engine
 - This allowed the creation of 3 `TestContext` fixtures for a test to use depending on what it needs.
   - If a test only cares about SQL queries, it can depend on the `ctx` fixture which will only invoke the test once per engine adapter
   - If a test only cares about DataFrame's, it can depend on the `ctx_df` fixture which will invoke the test once for each DataFrame implementation supported by that engine
   - If a test cares about both SQL Queries and DataFrame's, it can depend on the `ctx_query_and_df` fixture which will call the test the correct amount of times for each engine - once for `query` and once for each DataFrame implementation supported by that engine

This has allowed the removal of the logic that was skipping tests based on `test_type` while also calling our parametrized tests for the native DataFrame implementations which were previously being skipped.

This uncovered a bunch of bugs / inconsistencies in how we were handling native DataFrame's which I have also fixed in this PR

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -297,10 +297,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          #filters:
-          #  branches:
-          #    only:
-          #      - main
+          filters:
+            branches:
+              only:
+                - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -281,7 +281,6 @@ workflows:
                 - trino
                 - spark
                 - clickhouse
-                - clickhouse-cluster
                 - risingwave
       - engine_tests_cloud:
           name: cloud_engine_<< matrix.engine >>
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -297,10 +297,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
--- a/Makefile
+++ b/Makefile
@@ -126,54 +126,51 @@ engine-%-down:
 ##################
 
 clickhouse-test: engine-clickhouse-up
-	pytest -n auto -x -m "clickhouse" --retries 3 --junitxml=test-results/junit-clickhouse.xml
-
-clickhouse-cluster-test: engine-clickhouse-up
-	pytest -n auto -x -m "clickhouse_cluster" --retries 3 --junitxml=test-results/junit-clickhouse-cluster.xml
+	pytest -n auto -m "clickhouse" --retries 3 --junitxml=test-results/junit-clickhouse.xml
 
 duckdb-test: engine-duckdb-install
-	pytest -n auto -x -m "duckdb" --retries 3 --junitxml=test-results/junit-duckdb.xml
+	pytest -n auto -m "duckdb" --retries 3 --junitxml=test-results/junit-duckdb.xml
 
 mssql-test: engine-mssql-up
-	pytest -n auto -x -m "mssql" --retries 3 --junitxml=test-results/junit-mssql.xml
+	pytest -n auto -m "mssql" --retries 3 --junitxml=test-results/junit-mssql.xml
 
 mysql-test: engine-mysql-up
-	pytest -n auto -x -m "mysql" --retries 3 --junitxml=test-results/junit-mysql.xml
+	pytest -n auto -m "mysql" --retries 3 --junitxml=test-results/junit-mysql.xml
 
 postgres-test: engine-postgres-up
-	pytest -n auto -x -m "postgres" --retries 3 --junitxml=test-results/junit-postgres.xml
+	pytest -n auto -m "postgres" --retries 3 --junitxml=test-results/junit-postgres.xml
 
 spark-test: engine-spark-up
-	pytest -n auto -x -m "spark or pyspark" --retries 3 --junitxml=test-results/junit-spark.xml
+	pytest -n auto -m "spark" --retries 3 --junitxml=test-results/junit-spark.xml
 
 trino-test: engine-trino-up
-	pytest -n auto -x -m "trino or trino_iceberg or trino_delta or trino_nessie" --retries 3 --junitxml=test-results/junit-trino.xml
+	pytest -n auto -m "trino" --retries 3 --junitxml=test-results/junit-trino.xml
 
 risingwave-test: engine-risingwave-up
-	pytest -n auto -x -m "risingwave" --retries 3 --junitxml=test-results/junit-risingwave.xml
+	pytest -n auto -m "risingwave" --retries 3 --junitxml=test-results/junit-risingwave.xml
 
 #################
 # Cloud Engines #
 #################
 
 snowflake-test: guard-SNOWFLAKE_ACCOUNT guard-SNOWFLAKE_WAREHOUSE guard-SNOWFLAKE_DATABASE guard-SNOWFLAKE_USER guard-SNOWFLAKE_PASSWORD engine-snowflake-install
-	pytest -n auto -x -m "snowflake" --retries 3 --junitxml=test-results/junit-snowflake.xml
+	pytest -n auto -m "snowflake" --retries 3 --junitxml=test-results/junit-snowflake.xml
 
 bigquery-test: guard-BIGQUERY_KEYFILE engine-bigquery-install
-	pytest -n auto -x -m "bigquery" --retries 3 --junitxml=test-results/junit-bigquery.xml
+	pytest -n auto -m "bigquery" --retries 3 --junitxml=test-results/junit-bigquery.xml
 
 databricks-test: guard-DATABRICKS_CATALOG guard-DATABRICKS_SERVER_HOSTNAME guard-DATABRICKS_HTTP_PATH guard-DATABRICKS_ACCESS_TOKEN guard-DATABRICKS_CONNECT_VERSION engine-databricks-install
 	pip install 'databricks-connect==${DATABRICKS_CONNECT_VERSION}'
-	pytest -n auto -x -m "databricks" --retries 3 --junitxml=test-results/junit-databricks.xml
+	pytest -n auto -m "databricks" --retries 3 --junitxml=test-results/junit-databricks.xml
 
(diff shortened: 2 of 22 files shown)


Makefile
```diff
@@ -126,54 +126,51 @@ engine-%-down:
...
 clickhouse-test: engine-clickhouse-up
-	pytest -n auto -x -m "clickhouse" --retries 3 --junitxml=test-results/junit-clickhouse.xml
-
-clickhouse-cluster-test: engine-clickhouse-up
-	pytest -n auto -x -m "clickhouse_cluster" --retries 3 --junitxml=test-results/junit-clickhouse-cluster.xml
+	pytest -n auto -m "clickhouse" --retries 3 --junitxml=test-results/junit-clickhouse.xml
 
 duckdb-test: engine-duckdb-install
-	pytest -n auto -x -m "duckdb" --retries 3 --junitxml=test-results/junit-duckdb.xml
+	pytest -n auto -m "duckdb" --retries 3 --junitxml=test-results/junit-duckdb.xml
```
erindru, line 132: I removed `-x` ("stop after first failure") because I notice it can interfere with `pytest-retry`

Makefile
```diff
@@ -126,54 +126,51 @@ engine-%-down:
...
 postgres-test: engine-postgres-up
-	pytest -n auto -x -m "postgres" --retries 3 --junitxml=test-results/junit-postgres.xml
+	pytest -n auto -m "postgres" --retries 3 --junitxml=test-results/junit-postgres.xml
 
 spark-test: engine-spark-up
-	pytest -n auto -x -m "spark or pyspark" --retries 3 --junitxml=test-results/junit-spark.xml
+	pytest -n auto -m "spark" --retries 3 --junitxml=test-results/junit-spark.xml
 
 trino-test: engine-trino-up
-	pytest -n auto -x -m "trino or trino_iceberg or trino_delta or trino_nessie" --retries 3 --junitxml=test-results/junit-trino.xml
```
erindru, line 150: We always run the trino tests as a single unit so I have consolidated this

Makefile
```diff
@@ -126,54 +126,51 @@ engine-%-down:
 ##################
 
 clickhouse-test: engine-clickhouse-up
-	pytest -n auto -x -m "clickhouse" --retries 3 --junitxml=test-results/junit-clickhouse.xml
```
erindru, line 129: Both `clickhouse` and `clickhouse-cluster` are now marked as `clickhouse` so they only need a single consolidated entry (similar to `trino`).

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2002,6 +2004,19 @@ def _fetch_native_df(
             self.execute(query, quote_identifiers=quote_identifiers)
             return self.cursor.fetchdf()
 
+    def _native_df_to_pandas_df(
```
erindru, line 2007: I only did this to support the interface for `create_view()` which takes a `QueryOrDF` and noticed that the PySpark implementation was just calling `toPandas()` on the PySpark DataFrame.

However, i'm wondering if we rip this out and make `create_view()` only take a `Query`. The reason is that we [stopped supporting views in Python models](https://github.com/TobikoData/sqlmesh/pull/2908) so I dont see where in the codebase a DataFrame can still be legitimately shoved into a view

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -1107,6 +1108,39 @@ def _normalize_decimal_value(self, col: exp.Expression, precision: int) -> exp.E
...
+
+    def _columns_to_types(
+        self, query_or_df: QueryOrDF, columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+    ) -> t.Optional[t.Dict[str, exp.DataType]]:
+        if (
+            not columns_to_types
+            and bigframes
+            and isinstance(query_or_df, bigframes.dataframe.DataFrame)
+        ):
+            # using dry_run=True attempts to prevent the DataFrame from being materialized just to read the column types from it
```
erindru, line 1129: In a real scenario with a Python model, I believe `columns_to_types` is a required field and is read off the model.

However, a bunch of existing tests deliberately didn't supply `columns_to_types` to EngineAdapter methods so that `DataFrame` type inference could be tested, so I added a basic implementation for the native dataframes (`bigframe` and `snowpark` since `pyspark` already had a comprehensive implementation)

tests/core/engine_adapter/integration/__init__.py
```diff
@@ -30,6 +33,120 @@
...
+                    # ref: https://duckdb.org/docs/connect/concurrency.html#writing-to-duckdb-from-multiple-processes
+                    pytest.mark.xdist_group("engine_integration_duckdb"),
+                ]
+            )
+        return marks
+
+
+ENGINES = [
+    # Docker engines that can be locally tested
+    IntegrationTestEngine("duckdb"),
```
erindru, line 70: This is the new way to define engines and all their variants. This allows `generate_pytest_params()` below to generate the correct `pytest.param` objects for each engine so we dont for example end up with a `pyspark` test variant for `duckdb` like we have now

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -82,185 +82,7 @@ def dev_table_name_for(self, snapshot: Snapshot) -> str:
         return self.table_name_for(snapshot, is_deployable=False)
 
 
-@pytest.fixture(params=["df", "query", "pyspark"])
-def test_type(request):
-    return request.param
-
-
-@pytest.fixture(
```
erindru, line 90: All of these are now removed in favour of being dynamically generated from the `IntegrationTestEngine` objects

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -2736,34 +2521,55 @@ def _use_warehouse_as_state_connection(gateway_name: str, config: Config):
     sqlmesh_context.migrate()
 
 
-def test_python_model_column_order(ctx: TestContext, tmp_path: pathlib.Path):
-    if ctx.test_type == "pyspark" and ctx.dialect in ("spark", "databricks"):
-        # dont skip
-        pass
-    elif ctx.test_type != "df":
-        pytest.skip("python model column order test only needs to be run once per db")
+def test_python_model_column_order(ctx_df: TestContext, tmp_path: pathlib.Path):
```
erindru, line 2524: After the changes in this PR, I was able to consolidate this test to run parameterized for every DataFrame implementation for every engine.

Previously, BigQuery and Snowflake had a copy+pasted version in `test_integration_<engine>.py`

tests/core/engine_adapter/integration/test_integration_clickhouse.py
```diff
@@ -1,46 +1,36 @@
...
+from tests.core.engine_adapter.integration import (
+    TestContext,
+    generate_pytest_params,
+    ENGINES_BY_NAME,
+    IntegrationTestEngine,
+)
+
 
 @pytest.fixture(
-    params=[
```
erindru, line 10: Previously, these were all duplicated from `test_integration.py` and needed to be kept in sync.

Now, we can just generate them from the definitions

sqlmesh/core/engine_adapter/base.py
erindru: Oh, it [looks like](https://github.com/TobikoData/sqlmesh/blob/1837bb649e1251ea3fdafffddf1026c86d6e20d2/sqlmesh/core/test/definition.py#L193) we still need to shove DataFrame's into views for Model tests

.circleci/continue_config.yml
```diff
@@ -298,10 +297,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru, line 300: TODO: revert prior to merge

georgesittas (approved):
Great work, nice!

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2002,6 +2004,19 @@ def _fetch_native_df(
...
+        self,
+        query_or_df: QueryOrDF,
+    ) -> t.Union[Query, pd.DataFrame]:
+        """
+        Take a "native" DataFrame (eg Pyspark, Bigframe, Snowpark etc) and convert it to Pandas
+        """
+        if isinstance(query_or_df, (exp.Query, exp.DerivedTable, pd.DataFrame)):
+            return query_or_df
```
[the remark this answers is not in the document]
georgesittas: To be clear, I'm not suggesting that we change anything related to this as part of the PR. Just meant to point it out since I stumbled upon it again.

> Would it be possible that parsing a weird and wacky SQL query could produce an exp.DerivedTable node and we just dont have a test that covers this?

The only subclasses of `DerivedTable` are `UDTF`, `CTE` and `Subquery`. The last one is a `Query`, so it should be handled by the instance check. The previous two shouldn't ever be created on their own and passed around as "query or df", because they lack some characteristics that `Query` instances have.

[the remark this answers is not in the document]
erindru: Ah, I see. Thanks for the background.

Yeah it would definitely help simplify `isinstance` checks because annoyingly you cant use the `Query` or `DF` typehints for `isinstance` checks

PR #4409, opened 2025-05-14
chore(vscode): clean up format and improve logs

benfdking:


vscode/extension/src/utilities/exec.ts
```diff
@@ -12,14 +12,26 @@ export interface CancellableExecOptions extends ExecOptions {
   signal?: AbortSignal
 }
 
-export function execAsync(
+export async function execAsync(
```
themisvaltinos, line 15: now that this is async, does the way AbortSignal is handled in execAsyncCore with child.kill() risk leaving behind hanging processes or it is always graceful? (I'm not fully familiar with this kind of handling, just asking to understand)

themisvaltinos (approved):

vscode/extension/src/utilities/exec.ts
benfdking: I reformated to make it slightly safer 🙏

PR #4417, opened 2025-05-15
fix(vscode): make nodes no longer transparent

benfdking:


georgesittas (approved):

themisvaltinos (approved):
nice. also react-flow__attribution is missing a dot

PR #4423, opened 2025-05-15
Chore: do not warn if non-literal is used in then/else_value of PIVOT macro

### LINKED ISSUE #4421: Coercion of expression '<column_name>' to type '<class 'sqlglot.expressions.Literal'>' failed. Using non coerced expression at '<file_name>'
Not sure whether this is a bug or a faulty implementation on my end.

```sql
@DEF(overage_types, [
  'users',
  'messages'
  ]);

SELECT
  ...,
  @PIVOT(overage_type, @overage_types, then_value := usage, suffix := '_usage', quote := FALSE),
  ...
```
Yields
```
  SUM(CASE WHEN `_stg`.`overage_type` = 'users' THEN `_stg`.`usage` ELSE 0 END) AS `users_usage`,                                                                  
  SUM(CASE WHEN `_stg`.`overage_type` = 'messages' THEN `_stg`.`usage` ELSE 0 END) AS `messages_usage`  
```
Which is expected, but also shows this error in the CLI:
```
Coercion of expression 'usage' to type '<class 'sqlglot.expressions.Literal'>' failed. Using non coerced expression at 
'my_model.sql'
```
Things I've tried:
* `@PIVOT(overage_type, @overage_types, then_value := _stg.usage, suffix := '_usage', quote := FALSE),`
* `@PIVOT(overage_type, @overage_types, then_value := usage, suffix := _usage, quote := TRUE),`. This doesn't work because of an issue with suffixes.
* I checked the class [Literal](https://github.com/tobymao/sqlglot/blob/main/sqlglot/expressions.py#L2460) but I don't expect it to be an issue on sqlglot side.

I expect the issue is with the fact quoting is set to FALSE and the `suffix` ([ref](https://github.com/TobikoData/sqlmesh/blob/d0129bbda6099d5a753bf37fcdbeeee55745459e/sqlmesh/core/macros.py#L1096)) is expected to be of type string. But when I pass it on as string, it's incorrectly projecting the aliases.

### Context
* Engine: BigQuery
* SQLMesh v0.180.0

georgesittas:
Fixes #4421

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1094,8 +1094,8 @@ def pivot(
     cmp: exp.Literal = exp.Literal.string("="),
     prefix: exp.Literal = exp.Literal.string(""),
     suffix: exp.Literal = exp.Literal.string(""),
-    then_value: exp.Literal = exp.Literal.number(1),
-    else_value: exp.Literal = exp.Literal.number(0),
+    then_value: t.Union[exp.Literal, exp.Expression] = exp.Literal.number(1),
+    else_value: t.Union[exp.Literal, exp.Expression] = exp.Literal.number(0),
     quote: exp.Boolean = exp.true(),
     distinct: exp.Boolean = exp.false(),
 ) -> t.List[exp.Expression]:
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1087,17 +1087,17 @@ def haversine_distance(
 @macro()
 def pivot(
     evaluator: MacroEvaluator,
-    column: exp.Column,
-    values: t.Union[exp.Array, exp.Tuple],
-    alias: exp.Boolean = exp.true(),
-    agg: exp.Literal = exp.Literal.string("SUM"),
-    cmp: exp.Literal = exp.Literal.string("="),
-    prefix: exp.Literal = exp.Literal.string(""),
-    suffix: exp.Literal = exp.Literal.string(""),
-    then_value: t.Union[exp.Literal, exp.Expression] = exp.Literal.number(1),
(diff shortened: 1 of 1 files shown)


sqlmesh/core/macros.py
```diff
@@ -1094,8 +1094,8 @@ def pivot(
     cmp: exp.Literal = exp.Literal.string("="),
     prefix: exp.Literal = exp.Literal.string(""),
     suffix: exp.Literal = exp.Literal.string(""),
-    then_value: exp.Literal = exp.Literal.number(1),
-    else_value: exp.Literal = exp.Literal.number(0),
+    then_value: t.Union[exp.Literal, exp.Expression] = exp.Literal.number(1),
+    else_value: t.Union[exp.Literal, exp.Expression] = exp.Literal.number(0),
```
izeigerman, line 1098: What's the point of a union here if Expression is a superclass of Literal?

sqlmesh/core/macros.py
```diff
@@ -1087,17 +1087,17 @@ def haversine_distance(
...
+    values: t.List[SQL],
+    alias: bool = True,
+    agg: SQL = SQL("SUM"),
+    cmp: SQL = SQL("="),
+    prefix: SQL = SQL(""),
+    suffix: SQL = SQL(""),
+    then_value: SQL = SQL("1"),
+    else_value: SQL = SQL("0"),
+    quote: bool = True,
+    distinct: bool = False,
```
georgesittas, line 1100: Since we only use these arguments to construct a string to parse, I figured that coercing the corresponding expressions into their `sql()` representations should suffice.

cc @tobymao @z3z1ma

PR #4427, opened 2025-05-15
Feat: more cores for loading

themisvaltinos:
This update adds a `ProcessPoolExecutor` for parallel loading of a project's models. It also adds a mock executor for single-process scenarios, such as when the system doesn’t support fork. Also, it refactors `optimized_query_cache_pool` to use this unified execution logic, eliminating the need for separate sequential and parallel code paths.


Makefile
```diff
@@ -64,7 +64,7 @@ engine-up: engine-clickhouse-up engine-mssql-up engine-mysql-up engine-postgres-
 engine-down: engine-clickhouse-down engine-mssql-down engine-mysql-down engine-postgres-down engine-spark-down engine-trino-down
 
 fast-test:
-	pytest -n auto -m "fast and not cicdonly" && pytest -m "isolated"
+	pytest -n auto -m "fast and not cicdonly" && pytest -m "isolated" && pytest -m "isolated2"
```
izeigerman, line 67: This label name is not very descriptive. Can we be more precise

sqlmesh/core/loader.py
```diff
@@ -473,48 +517,62 @@ def _load_sql_models(
...
+        for path in paths.copy():
+            cached_models = cache.get(path)
+            if cached_models:
+                paths.remove(path)
+                for model in cached_models:
+                    if model.enabled:
+                        models[model.fqn] = model
+
+        if paths:
+            defaults = dict(
```
izeigerman, line 531: This overlaps with the `defaults` that is in the dict itself. Can we have a more descriptive name for this?

sqlmesh/core/loader.py
```diff
@@ -225,37 +283,40 @@ def _load_external_models(
...
+                                **{
+                                    "dialect": self.config.model_defaults.dialect,
+                                    "default_catalog": self.context.default_catalog,
+                                    **row,
+                                },
+                            )
+                            for row in YAML().load(file.read())
+                        ]
+
+                    cache.put(external_models, path)
```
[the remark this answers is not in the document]
tobymao: i'm not sure. this is the simplest because the loading all happens in separate processes and must all be complete separate, we only communicate through the file system

sqlmesh/core/loader.py
```diff
@@ -69,21 +68,80 @@ class LoadedProject:
...
+    cache: t.Optional[CacheBase] = None,
+) -> None:
+    global _defaults, _cache, _config, _selected_gateway
+    _defaults = defaults
+    _cache = cache
+    _config = config
+    _selected_gateway = selected_gateway
+
+
+def load_sql_models(path: Path) -> t.Tuple[Path, list[Model]]:
```
izeigerman, line 98: I don't see the 1st item of the return value being used anywhere. Is it used?

sqlmesh/core/loader.py
```diff
@@ -473,48 +517,62 @@ def _load_sql_models(
...
+                                models[model.fqn] = model
+                    except Exception as ex:
+                        errors.append(self._failed_to_load_model_error(path, str(ex)))
 
-                if isinstance(model, SeedModel):
-                    seed_path = model.seed_path
-                    self._track_file(seed_path)
+            if errors:
+                error_string = "\n".join(errors)
+                raise ConfigError(f"Failed to load models\n\n{error_string}")
```
izeigerman, line 575: [Nit] I believe `Failed to load models\n\n` is redundant

sqlmesh/core/loader.py
```diff
@@ -69,21 +68,80 @@ class LoadedProject:
...
+    assert _config
+
+    gateway_name = gateway_name or _selected_gateway
+
+    try:
+        gateway = _config.get_gateway(gateway_name)
+    except ConfigError:
+        from sqlmesh.core.console import get_console
+
+        get_console().log_warning(
```
izeigerman, line 119: Does this respect `--ignore-warnings` when this runs in a worker process?

sqlmesh/core/loader.py
```diff
@@ -69,21 +68,80 @@ class LoadedProject:
...
     @abc.abstractmethod
-    def get_or_load_models(
-        self, target_path: Path, loader: t.Callable[[], t.List[Model]]
-    ) -> t.List[Model]:
-        """Get or load all models from cache."""
+    def put(self, models: t.List[Model], path: Path) -> bool:
+        pass
+
+    @abc.abstractmethod
+    def get(self, path: Path) -> t.List[Model]:
```
izeigerman, line 75: This interface is only needed in one place but impacts all other places. Can we keep both interfaces and revert unrelated changes?

sqlmesh/core/loader.py
themisvaltinos: Renamed to `model_loading_defaults` for clarity, couldn't think of a better alternative

sqlmesh/core/loader.py
themisvaltinos: no this was leftover from before I added the futures_to_paths dictionary. The path is already accessible so it’s no longer needed. I removed it and updated _load_sql_models to use the simplified return value—loaded = future.result() instead of unpacking with _.

sqlmesh/core/loader.py
themisvaltinos: removed it

sqlmesh/core/loader.py
themisvaltinos: good catch, I tested it and your suspicion is correct it wasn't working because the console wasn’t shared across processes. I added a console parameter to _init_model_defaults to pass self._console to worker processes when creating the process pool, so that each worker used set_console(console) to match the parent’s console and preserves  ignore_warnings along with all console settings

sqlmesh/core/loader.py
themisvaltinos: reverted to use in all other places except for sqlmodels the previous interface

sqlmesh/utils/process.py
```diff
@@ -0,0 +1,64 @@
...
+        """Execute the function synchronously and return a Future with the result."""
+        future = Future()
+        try:
+            result = fn(*args, **kwargs)
+            future.set_result(result)
+        except Exception as e:
+            future.set_exception(e)
+        return future
+
+    def map(self, fn, *iterables, timeout=None, chunksize=1):
```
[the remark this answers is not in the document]
themisvaltinos: since shutdown was called in `__exit__` and since the synchronous executor runs everything in the main process, there are no resources to release or cleanup and futures are complete from the start this is why I didn't add it. but I revise to have it to keep the api similar to python’s ProcessPoolExecutor

sqlmesh/core/model/definition.py
```diff
@@ -812,8 +812,10 @@ def convert_to_time_column(
         return exp.convert(time)
 
     def set_mapping_schema(self, schema: t.Dict) -> None:
+        # Make a shallow copy to avoid modifying the original in case they're the same
```
izeigerman, line 815: Even with the comment I don't really understand this maneuver. Does this mean that we're passing the same exact reference here? This is problematic then and we should return a copy upstream whenever we obtain the reference.

sqlmesh/core/model/schema.py
```diff
@@ -102,17 +81,25 @@ def process_models(completed_model: t.Optional[Model] = None) -> None:
                     )
                 )
 
+    errors: t.List[str] = []
```
izeigerman, line 84: Should this be a list if we exit the loop if at least one error was encountered? Should this be `t.Optional`?

sqlmesh/core/loader.py
```diff
@@ -73,17 +73,90 @@ def get_or_load_models(
         self, target_path: Path, loader: t.Callable[[], t.List[Model]]
     ) -> t.List[Model]:
         """Get or load all models from cache."""
+        pass
+
+    @abc.abstractmethod
+    def put(self, models: t.List[Model], path: Path) -> bool:
```
izeigerman, line 79: Can we please add some docstrings. Eg. I'm not sure what does the boolean return value suppose to represent.

sqlmesh/core/loader.py
```diff
@@ -73,17 +73,90 @@ def get_or_load_models(
...
+
+def load_sql_models(path: Path) -> t.List[Model]:
+    assert _defaults
+    assert _cache
+
+    with open(path, "r", encoding="utf-8") as file:
+        expressions = parse(file.read(), default_dialect=_defaults["dialect"])
+    models = load_sql_based_models(expressions, path=Path(path).absolute(), **_defaults)
+
+    return [] if _cache.put(models, path) else models
```
izeigerman, line 121: Don't we always expect `True` here provided that we load SQL models only?

izeigerman (commented):
This looks much cleaner, thank you!

sqlmesh/core/loader.py
themisvaltinos: yes I see what you're saying we cache the models so subsequently we could retrieve them from the cache. I had the same thought, but the issue arises with **seed models** since we don’t cache these so they will be missing. That’s why the `put` call can return `False` in those cases here

sqlmesh/core/model/schema.py
themisvaltinos: so my thinking was less about having the final exception include more errors for information and more about in flight futures and I don't know if there is a way to just stop the for loop which uses `as_completed`. firstly we cannot simply raise and stop in the first exception since it is inside a with context so it will be swallowed. also even though the with ensures that when the block exits, it waits for the running tasks to complete and shuts down the executor, but multiple futures can finish around the same time and my concern is that it is possible multiple exceptions could be added to errors simultaneously

sqlmesh/core/model/definition.py
themisvaltinos: yes the issue is since `set_mapping_schema` clears and replaces the schema entirely (which is necessary when a model might be processed multiple times and then has additional information). however if the dictionary is the same it would clear the old one which happens to be the identical to the schema that was passed in so it then updates with an empty dictionary. I revised the logic ustream instead of doing the copy here to not call set_mapping_schema when the two dictionaries are identical, I don't know if this is better

sqlmesh/core/model/schema.py
izeigerman: >  firstly we cannot simply raise and stop in the first exception since it is inside a with context so it will be swallowed.

why would it be swallowed? 

> but multiple futures can finish around the same time and my concern is that it is possible multiple exceptions could be added to errors simultaneously

you iterate futures in order. Even if they finish simultaneously the loop code will be executed sequentially. Am I missing something?

sqlmesh/core/loader.py
izeigerman: ah, makes sense. I forgot that seed models also go through this path

sqlmesh/core/loader.py
```diff
@@ -473,48 +549,62 @@ def _load_sql_models(
...
             self._track_file(path)
+            paths.add(path)
+
+        for path in paths.copy():
+            cached_models = cache.get(path)
+            if cached_models:
+                paths.remove(path)
+                for model in cached_models:
+                    if model.enabled:
+                        models[model.fqn] = model
```
VaggelisD, line 560: We can pull this on the for-loop above to avoid copying the list

sqlmesh/core/loader.py
```diff
@@ -473,48 +549,62 @@ def _load_sql_models(
...
+
+        if paths:
```
VaggelisD, line 562: We can probably spare an indent level by reversing the order here

```
if not paths:
  return models
```

sqlmesh/core/loader.py
```diff
@@ -473,48 +549,62 @@ def _load_sql_models(
...
-                    self._raise_failed_to_load_model_error(
-                        path, f"Duplicate SQL model name: '{model.name}'."
-                    )
-                if model.enabled:
-                    models[model.fqn] = model
+            errors: t.List[str] = []
+            with create_process_pool_executor(
+                initializer=_init_model_defaults,
+                initargs=(self.config, gateway, model_loading_defaults, cache, self._console),
+                max_workers=c.MAX_FORK_WORKERS,
```
VaggelisD, line 585: ditto about `max_workers`, it's the default value

sqlmesh/core/loader.py
```diff
@@ -473,48 +549,62 @@ def _load_sql_models(
...
+            ) as pool:
+                futures_to_paths = {pool.submit(load_sql_models, path): path for path in paths}
+                for future in concurrent.futures.as_completed(futures_to_paths):
+                    path = futures_to_paths[future]
+                    try:
+                        loaded = future.result()
+                        for model in loaded or cache.get(path):
```
VaggelisD, line 592: Why do we have to repeat `cache.get(path)` here, weren't those processed earlier?

sqlmesh/core/model/schema.py
```diff
@@ -102,17 +81,26 @@ def process_models(completed_model: t.Optional[Model] = None) -> None:
...
-                fqn, entry_name, data_hash, metadata_hash, mapping_schema = future.result()
-                model = models[fqn]
-                model._data_hash = data_hash
-                model._metadata_hash = metadata_hash
-                model.set_mapping_schema(mapping_schema)
-                optimized_query_cache.with_optimized_query(model, entry_name)
-                _update_schema_with_model(schema, model)
-                process_models(completed_model=model)
+                try:
+                    futures.remove(future)
```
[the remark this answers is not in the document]
themisvaltinos: this is how it always was and I couldn't think of a better way to handle it safely

sqlmesh/core/model/schema.py
```diff
@@ -102,17 +81,26 @@ def process_models(completed_model: t.Optional[Model] = None) -> None:
                     )
                 )
 
+    errors: t.List[str] = []
     with optimized_query_cache_pool(optimized_query_cache) as executor:
         process_models()
 
-        while futures:
+        while futures and not errors:
```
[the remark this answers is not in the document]
themisvaltinos: this was revised

sqlmesh/core/model/cache.py
```diff
@@ -66,9 +62,22 @@ def get_or_load(
                 model.full_depends_on
 
             self._file_cache.put(name, entry_id, value=models)
-
         return models
 
+    def put(self, models: t.List[Model], name: str, entry_id: str = "") -> bool:
+        if isinstance(models, list) and isinstance(seq_get(models, 0), (SqlModel, ExternalModel)):
```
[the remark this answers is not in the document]
themisvaltinos: no good point revised it

sqlmesh/core/loader.py
```diff
@@ -473,48 +549,62 @@ def _load_sql_models(
...
-                if model.fqn in models:
-                    self._raise_failed_to_load_model_error(
-                        path, f"Duplicate SQL model name: '{model.name}'."
-                    )
-                if model.enabled:
-                    models[model.fqn] = model
+            errors: t.List[str] = []
+            with create_process_pool_executor(
+                initializer=_init_model_defaults,
+                initargs=(self.config, gateway, model_loading_defaults, cache, self._console),
```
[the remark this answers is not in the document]
themisvaltinos: changed it to only pass what is needed from the config

sqlmesh/core/model/schema.py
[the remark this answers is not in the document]
VaggelisD: Does the remove actually serve a purpose here, do we need it?

sqlmesh/core/loader.py
themisvaltinos: default arguments are evaluated once at function definition time, so `c.MAX_FORK_WORKERS` gets fixed when the module is imported. If tests later change or mock this value, the function’s default won’t reflect it, potentially causing it to use the wrong executor. Explicitly passing `max_workers=c.MAX_FORK_WORKERS` ensures the correct, current value is used, avoiding issues like tests hanging. But I removed the default value from the function instead.

sqlmesh/core/loader.py
themisvaltinos: yes revised it

sqlmesh/core/model/cache.py
```diff
@@ -149,11 +158,8 @@ def _entry_name(model: SqlModel) -> str:
         return f"{model.name}_{crc32(hash_data)}"
 
 
-def optimized_query_cache_pool(optimized_query_cache: OptimizedQueryCache) -> ProcessPoolExecutor:
-    # fork doesnt work on Windows. ref: https://docs.python.org/3/library/multiprocessing.html#multiprocessing-start-methods
-    context_type = "spawn" if IS_WINDOWS else "fork"
-    return ProcessPoolExecutor(
```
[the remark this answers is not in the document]
themisvaltinos: yes that's on purpose because spawn will fail pickling

sqlmesh/core/model/schema.py
themisvaltinos: thanks for pointing this out it was my mistake looking into it deeper the way I was shutting down and handling exit which didn't return False so it suppressed the exception. I revised this so that now we can exit when one error was encountered

sqlmesh/core/model/schema.py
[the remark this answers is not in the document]
themisvaltinos: yes look at line 99 process_models is called again inside the loop

izeigerman (approved):

Issue #4397, opened 2025-05-13
Bug: Snowpark `SCD Type 2 By Time` models with Snowflake Engine concurrency error


zunedz:
I have Snowpark `SCD Type 2 By Time` models that runs in Snowflake. However, if I specify `concurrent_tasks` to be more than 1, I will have errors whenever I run `sqlmesh run`. Below is the error message I received when I set `concurrent_tasks` to 4: 

```py
Failed models

  "<DATABASE>"."<SCHEMA>"."MODEL_1"

    ProgrammingError:
      002003 (42S02): 01bc52f0-3201-9bfe-0007-878208226142: SQL compilation error:
      Object '<DATABASE>.SQLMESH__<SCHEMA>.__TEMP_<SCHEMA>__MODEL_1__2639487389_RNI5LATL' does
not exist or not authorized.

  "<DATABASE>"."<SCHEMA>"."MODEL_2"

    ProgrammingError:
      002003 (42S02): 01bc52f0-3201-9bfe-0007-87820822614a: SQL compilation error:
      Object '<DATABASE>.SQLMESH__<SCHEMA>.__TEMP_<SCHEMA>__MODEL_2__1485328045_ZEF1CCB4' does not exist or
not authorized.

  "<DATABASE>"."<SCHEMA>"."MODEL_3"

    ProgrammingError:
      002003 (42S02): 01bc52f0-3201-9b3c-0007-878208217e9e: SQL compilation error:
      Object '<DATABASE>.SQLMESH__<SCHEMA>.__TEMP_<SCHEMA>__MODEL_3"__217620242_H719ZQ6R' does not
exist or not authorized.
```

izeigerman:
Hey @zunedz we'd need the logs in order to investigate the issue. Anything you can provide?

erindru:
I think I just independently discovered this bug in #4406 .

I believe it is to do with the Snowpark session being non threadsafe. What I observed (at least in our test suite) is that the "catalog" and "schema" of the Snowpark session appear to be reset when it's accessed from another thread, which cause the `SNOWPARK_TEMP_TABLE_*` tables to get created in the wrong place. When we create our view over the top, it cant be.

This shows as a `Object does not exist or not authorized.` error in the Python client but if you look server side in the Snowflake web interface it is probably prefixed with
```
SQL compilation error:
Failure during expansion of view( '__temp_TEST_TABLE_l05oztp5': SQL compilation error:
<original error you see in SQLMesh>
```

[...]

PR #4431, opened 2025-05-16
Fix(snowflake): Allow models that utilize Snowpark to execute concurrently

erindru:
Addresses #4397 

Prior to this PR, a Snowpark session would get shared across threads.

This lead to strange errors like `Object does not exist or not authorized` as threads would all try to start a transaction, upload a DataFrame and return a pointer to a table which eventually lead to an error as they clobbered each others state.

This PR makes each Snowpark session thread-local and adds a test to ensure concurrent execution works

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -328,15 +328,15 @@ workflows:
             parameters:
               engine:
                 - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-                - athena
-          filters:
-            branches:
-              only:
-                - main
+                #- databricks
+                #- redshift
+                #- bigquery
+                #- clickhouse-cloud
+                #- athena
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
       - trigger_private_tests:
           requires:
             - style_and_cicd_tests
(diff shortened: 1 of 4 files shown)


.circleci/continue_config.yml
```diff
@@ -292,15 +292,15 @@ workflows:
...
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-                - athena
-          filters:
-            branches:
-              only:
-                - main
+                #- databricks
```
erindru, line 295: TODO: revert prior to merge

georgesittas (approved):

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -69,6 +70,10 @@ class SnowflakeEngineAdapter(GetCurrentCatalogFromFunctionMixin, ClusteredByMixi
     )
     MANAGED_TABLE_KIND = "DYNAMIC TABLE"
 
+    def __init__(self, *args: t.Any, **kwargs: t.Any):
+        super().__init__(*args, **kwargs)
+        self._snowpark_threadlocal = threading.local()
```
izeigerman, line 75: We have `set_attribute` / `get_attribute` in `connection_pool` specifically for this purpose. Can we use that instead?

erindru: Oh, thanks, I didn't notice these. I've refactored to use them

izeigerman (approved):

PR #4440, opened 2025-05-16
fix(vscode): improve handling of uris

benfdking:


mykalmax (approved):

themisvaltinos (approved):
should we do the same in other parts as well like in `completion.ts` that is using still the uri.fsPath instead of  parsing uris with the vscode-uri library

PR #4441, opened 2025-05-16
Add iceberg support to table_diff

### LINKED ISSUE #4208: `table_diff` on Athena fails for Iceberg tables
Based on issue a user reported in Slack: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1744125717966919?thread_ts=1742833178.418699&cid=C044BRE5W4S

The table_diff [creates a temporary table](https://github.com/TobikoData/sqlmesh/blob/9952bbba55b53c5f8dff06fa8d2cc82636406e92/sqlmesh/core/table_diff.py#L425) in order to hold the diff results.

In the case of Athena, this temporary table will be created as Hive by default. However, if the source table is Iceberg, the temporary table also  needs to be created as Iceberg because some data types are not compatible between the two.

Timestamp fields are a good example:
```
pyathena.error.OperationalError: NOT_SUPPORTED: Unsupported Hive type: timestamp(6) with time zone. You may need to manually clean the data at location 's3://athena/sqlmesh/tables/1543f306-59e7-4310-b2b0-f75e0558761a' before retrying. Athena will not delete data in your account.
```

williebsweet:
Resolves #4208

PR contains the following:
- Added a check of both source and target table formats 
- Added an error, should someone try to diff an Iceberg table and a Hive table
- Set the table format for a Iceberg/Iceberg diff to Iceberg
- Tests for various combinations of source/target/temp table formats


sqlmesh/core/engine_adapter/base.py
```diff
@@ -2203,6 +2203,19 @@ def temp_table(
             query_or_df, columns_to_types=columns_to_types, target_table=name
         )
 
+        if hasattr(self, "_query_table_type_or_raise") and hasattr(self, "source_table"):
```
izeigerman, line 2206: Should this logic be in `athena.py` where these attributes are available?

izeigerman:
@williebsweet thanks for the submission! Some tests will be helpful

sqlmesh/core/engine_adapter/base.py
erindru: I... dont actually see how this works. Where does `self.source_table` come from? This isn't an engine adapter property.

table_diff [calls](https://github.com/TobikoData/sqlmesh/blob/529e52fa9d890dae3e56bb958e49f5788676e5dc/sqlmesh/core/table_diff.py#L434) `EngineAdapter.temp_table(...)` and passes the query that it's trying to store the results of, which looks something like:
```
with source as (select ... from source_table),
target as (select ... from target_table),
joined as (select ... from source full outer join target)
select * from joined
```

I actually think it's table_diff that should be supplying the table format, not the engine adapter introspecting the query to try and look up all the tables referenced in the query to see if any of them are Iceberg.

williebsweet: @erindru Something more like [this previous commit](https://github.com/TobikoData/sqlmesh/pull/4441/commits/d45c73e487a5fc6c84f1cf7d6ea02b37822ead9a)?

erindru: Yes, exactly. With perhaps a comment to show why there is an Athena-specific check and the addition of some tests

williebsweet:
@erindru I moved the logic back into `table_diff`. I did the following:
- Added a check of both source and target table formats 
- Added an error, should someone try to diff an Iceberg table and a Hive table
- Set the table format for a Iceberg/Iceberg diff to Iceberg

@izeigerman I also added tests for the scenarios above.

williebsweet:
Forgot to mention - I also tested this with my company's SQLMesh project (installing the `--editable` package) and I was able to run a diff. 🎆

sqlmesh/core/table_diff.py
```diff
@@ -431,7 +433,26 @@ def name(e: exp.Expression) -> str:
...
+                # the formats be the same for the source, target, and temp tables.
+                source_table_type = self.adapter._query_table_type(self.source_table)
+                target_table_type = self.adapter._query_table_type(self.target_table)
+
+                if source_table_type == "iceberg" and target_table_type == "iceberg":
+                    temp_table_kwargs["table_format"] = "iceberg"
+                # Sets the temp table's format to Iceberg.
+                # If neither source nor target table is Iceberg, it defaults to Hive (Athena's default).
+                elif source_table_type == "iceberg" or target_table_type == "iceberg":
+                    raise SQLMeshError(
```
erindru, line 448: I _think_ this is right. Table diff relies on copying a sample of data from each table into a single table, which on most engines is fine because they have a consistent table format.

I know that timestamp types in particular are handled differently between Athena/Iceberg and Athena/Hive so I dont think we can say "if Iceberg is detected - make the diff table Iceberg" because copying Hive data into it verbatim can still cause an error

erindru:
Thanks @williebsweet ! This looks much better

erindru (approved):

PR #4445, opened 2025-05-17
Fix: partitioned_by changes are breaking

treysp:
Currently, changes to partitioned_by in forward-only models are not classified as breaking. Because you cannot repartition a table via `ALTER` statements, this means the only way to repartition a forward-only model is to manually drop the physical table.

This PR makes partitioned_by changes always breaking.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1447,6 +1447,9 @@ class SqlModel(_Model):
         if self.lookback != previous.lookback:
             return None
 
+        if self.partitioned_by != previous.partitioned_by:
+            return None
+
         try:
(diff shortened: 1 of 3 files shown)


erindru:
I think it's more nuanced than this and we need to delegate to the engine adapter.

For example, Iceberg [specifically supports partition evolution](https://iceberg.apache.org/docs/latest/evolution/) and it [is implemented ](https://trino.io/docs/current/connector/iceberg.html#alter-table-set-properties) in Trino. In that case, the partitioning on an Iceberg table can be changed in a forward_only fashion without a full rebuild.

Not sure where the best place to put the check is though

treysp:
Decision: warn and do this change via table swap on backend (rather than breaking)

PR #4450, opened 2025-05-19
feat(vscode): add end point for column lineage

benfdking:


sqlmesh/lsp/main.py
```diff
@@ -92,18 +94,37 @@ def all_models(ls: LanguageServer, params: AllModelsRequest) -> AllModelsRespons
...
+            if (
+                request.url.startswith("/api/lineage")
+                and request.url.count("/") == 3
+                and request.method == "GET"
+            ):
                 name = request.url.split("/")[-1]
                 lineage = model_lineage(name, self.lsp_context.context)
                 non_set_lineage = {k: v for k, v in lineage.items() if v is not None}
                 return ApiResponseGetLineage(data=non_set_lineage)
+            if (
```
themisvaltinos, line 113: I notice a lot of reusability of same comparisons as above conditional for the method etc. maybe refactor also to be a bit more readable of what is happening with some comments and reuse what you can and handle slashes with something like urlparse (except if it doesn't work on windows ):
```python
    parsed_url = urlparse(request.url)
    path_parts = parsed_url.path.strip("/").split("/")

    if request.method == "GET":
        if path_parts == ["api", "models"]:
            return ApiResponseGetModels(data=get_models(self.lsp_context.context))

        if path_parts[:2] == ["api", "lineage"]:
            if len(path_parts) == 3:
                # /api/lineage/{model}
                model_name = unquote(path_parts[2])
                lineage = model_lineage(model_name, self.lsp_context.context)
                non_set_lineage = {k: v for k, v in lineage.items() if v is not None}
                return ApiResponseGetLineage(data=non_set_lineage)

            elif len(path_parts) == 4:
                # /api/lineage/{model}/{column}
                model_name = unquote(path_parts[2])
                column = unquote(path_parts[3])
                logging.info(f"Column lineage request: {model_name} {column}")
                column_lineage_response = column_lineage(
                    model_name, column, False, self.lsp_context.context
                )
                return ApiResponseGetColumnLineage(data=column_lineage_response)
```

themisvaltinos (approved):

sqlmesh/lsp/main.py
benfdking: 🙏

PR #4459, opened 2025-05-19
Feat(duckdb): Add support for authentication using secrets

themisvaltinos:
This PR adds support for [DuckDB's Secrets Manager](https://duckdb.org/docs/stable/configuration/secrets_manager.html) in SQLMesh for cloud storage authentication. 

This is supported in DuckDB v0.10.0 + and provides a more secure way to manage authentication credentials for external services like S3 versus the [legacy authentication method](https://duckdb.org/docs/stable/extensions/httpfs/s3api_legacy_authentication.html) that had reliability issues and didn't work with new duckdb features [like attaching catalogs](https://duckdb.org/docs/stable/guides/network_cloud_storage/duckdb_over_https_or_s3.html).

These now can be set in the `secrets` and will create the corresponding secret. For example for S3:

  ```yaml
  gateways:
    duckdb:
      connection:
        type: duckdb
        catalogs:
          local: local.db
          remote: "s3://bucket/data/remote.duckdb"
        extensions:
          - name: httpfs
        secrets:
          - type: s3
            region: "YOUR_AWS_REGION"
            key_id: "YOUR_AWS_ACCESS_KEY"
            secret: "YOUR_AWS_SECRET_KEY"
  ```
  
After configuring the secrets, the s3 paths or tables can directly be referenced in SQL queries without additional authentication steps.


sqlmesh/core/config/connection.py
```diff
@@ -283,6 +286,22 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
                 except Exception as e:
                     raise ConfigError(f"Failed to set connector config {field} to {setting}: {e}")
 
+            if self.secrets_config and version.parse(duckdb.__version__) >= version.parse("0.10.0"):
```
izeigerman, line 289: Should we let them know that `secrets_config` is not used if the version doesn't match the criteria?

sqlmesh/core/config/connection.py
```diff
@@ -201,6 +202,7 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
         catalogs: Key is the name of the catalog and value is the path.
         extensions: A list of autoloadable extensions to load.
         connector_config: A dictionary of configuration to pass into the duckdb connector.
+        secrets_config: A list of dictionaries used to generate DuckDB secrets for authenticating with external services (e.g. S3).
```
izeigerman, line 205: `_config` seems redundant since this is defined in the config already. Can this be just `secrets`?

sqlmesh/core/config/connection.py
themisvaltinos: yes good point I'll add an appropriate message that this is supported in duckdb versions after 0.10.0

georgesittas (approved):

izeigerman (approved):

PR #4467, opened 2025-05-19
Chore: Get fast-test working on Windows

erindru:
This extends the Windows tests to include `make fast-test` instead of just the LSP tests.

There is still more work to do in order to get `make cicd-test` fully working on Windows so that can be addressed in followup PR's as bandwidth allows

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -105,24 +105,29 @@ jobs:
       size: large
     steps:
       - halt_unless_core
+      - run:
+          name: Enable symlinks in git config
+          command: git config --global core.symlinks true
       - checkout
       - run:
-          name: Install Python 3.9
+          name: Install System Dependencies
           command: |
-            choco install python --version=3.9 -y
+            choco install make which -y
             refreshenv
-      - run: 
-          name: Install make
-          command: choco install make -y
       - run:
           name: Install SQLMesh dev dependencies
-          command: make install-dev
-      - run:
-          name: Fix Git URL override
-          command: git config --global --unset url."ssh://git@github.com".insteadOf      
+          command: |
+            python -m venv venv
+            . ./venv/Scripts/activate
+            python.exe -m pip install --upgrade pip
+            make install-dev
       - run:
           name: Run fast unit tests
-          command: make fast-test
+          command: |
+            . ./venv/Scripts/activate
+            which python
+            python --version
+            make fast-test
       - store_test_results:
           path: test-results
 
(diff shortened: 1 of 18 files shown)


.circleci/continue_config.yml
```diff
@@ -99,35 +99,35 @@ jobs:
...
+      - run:
+          name: Enable symlinks in git config
+          command: git config --global core.symlinks true
       - checkout
       - run:
-          name: Install Python 3.9
+          name: Install System Dependencies
           command: |
-            choco install python --version=3.9 -y
+            choco install make which -y
```
erindru, line 115: Installing python 3.9 [wasnt actually working](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/20865/workflows/39ea5db0-1b42-4ed8-8865-70b9b327fb19/jobs/209337) and the tests were being run with Python 3.12.

I was unable to actually install Python 3.9 because even with `--allow-downgrade`, Chocolatey would exit with an error. So i've left everything on Python 3.12 which does match our engine adapter tests

.circleci/continue_config.yml
```diff
@@ -99,35 +99,35 @@ jobs:
...
     executor:
       name: windows/default
       size: large
-    environment:
-      PYTEST_XDIST_AUTO_NUM_WORKERS: 4
     steps:
       - halt_unless_core
+      - run:
+          name: Enable symlinks in git config
+          command: git config --global core.symlinks true
```
erindru, line 110: Our DBT adapter tests rely on symlinks for `dbt_packages` which Windows has particularly bad support for. 

Thankfully it looks like the CirceCI runners have "Developer Mode" enabled which turns on the feature at the OS level, so it's just a matter of configuring `git` to create them on checkout

examples/sushi/linter/__init__.py
```diff
@@ -0,0 +1 @@
+# this makes "linter" a package so "linter.user" is a valid module for importlib.import_module() to load
```
benfdking, line 1: If this is the case, we should document it in our docs.

benfdking (approved):
Just document `__init__.py` if that's the case.

examples/sushi/linter/__init__.py
erindru: I noticed afterwards that this test only fails due to lack of `__init__.py` when running in the VSCode test runner. I think it's because the VSCode test runner already loads a package called `linter`

I later realised that the test passes fine from the CLI

PR #4484, opened 2025-05-20
fix!: check for base branch when deploying to prod

eakmanrq:
Currently if a user branches off a non-prod branch and then has a required approver approve, the bot will deploy and merge the branch like it is going to production. This PR updates the behavior so it checks if the base branch is prod and if not then it will not do the prod deploy. 

It is technically breaking since if a user currently has a prod branch name that isn't `main` or `master` and uses required approvers then it will stop deploying to prod. I think most likely this won't actually break for anyone though.

Looking for feedback:
* The config property is `prod_branch_names` but really there can be only a single prod branch. It takes a list though since I want to cover both `main` and `master`. Could just pick `main` or another option is to accept a `str` only too and just document a single value. A validator will convert it to a list and if not provided then it will still be `main` and `master`. Could also just leave it as is.


kelsin (approved):

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -444,6 +444,10 @@ def modified_snapshots(self) -> t.Dict[SnapshotId, t.Union[Snapshot, SnapshotTab
     def removed_snapshots(self) -> t.Set[SnapshotId]:
         return set(self.prod_plan_with_gaps.context_diff.removed_snapshots)
 
+    @property
+    def is_branched_from_prod(self) -> bool:
```
izeigerman, line 448: A bit misleading name. It can be branched from anything, the target what matters here. So isn't this more of `targets_prod_branch`?

PR #4492, opened 2025-05-21
Fix: propagate correct snapshot dict to renderer for virtual statements

georgesittas:
The issue is that `snapshots` in the `SnapshotEvaluator._promote_snapshot` method was a `SnapshotId -> Snapshot` dict instead of a `str -> Snapshot`, which is expected by the rendering methods. Since this dict flows into the macro evaluator, `columns_to_types` lookup using string keys would always fail and we'd fail.

I also updated the `_statement_renderer` and `_create_renderer` to pass the `mapping_schema` for consistency, since we do this for the query renderer as well. This is unrelated to the above issue, so happy to revert if we're unsure if it's safe.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -978,7 +978,6 @@ class SnapshotEvaluator:
                 end=end,
                 execution_time=execution_time,
                 engine_adapter=adapter,
-                snapshots=snapshots,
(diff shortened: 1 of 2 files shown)


izeigerman (approved):
Looks good, except I don't understand the motivation behind the mapping schema change. I suggest to revert that and stick to fixing the issue at hand

PR #4494, opened 2025-05-21
Fix: Normalize before applying quotes in get_relation and list_relations methods of the runtime dbt adapter

izeigerman:


tobymao (approved):

sqlmesh/dbt/adapter.py
```diff
@@ -262,10 +259,15 @@ def load_relation(self, relation: BaseRelation) -> t.Optional[BaseRelation]:
         return self._table_to_relation(mapped_table)
 
     def list_relations(self, database: t.Optional[str], schema: str) -> t.List[BaseRelation]:
-        reference_relation = self.relation_type.create(
-            database=database, schema=schema, quote_policy=self.quote_policy
+        target_schema = exp.Table(
+            this=None,
+            db=exp.to_identifier(schema),
+            catalog=exp.to_identifier(database) if database else None,
         )
```
georgesittas, line 266: Nit: I think we have a `schema_` helper in `dialect.py` which does this.

georgesittas (approved):

sqlmesh/dbt/adapter.py
izeigerman: perfect, thanks

PR #4495, opened 2025-05-22
Feat(experimental): DBT project conversion

erindru:
This PR contains an initial implementation of a command that can take a DBT project, read it into memory and write the result out as a native SQLMesh project.

To invoke it, use:
```
$ sqlmesh dbt convert -i <input_path> -o <output_path>
```

The way this works is that the project is first loaded into memory using our existing `DbtLoader`.
 - This means that the existing SQLMesh mappings from DBT model types -> SQLMesh model types are utilized
 - It also means that our existing DBT shims in the Jinja context can be utilized

The resulting models and macros are extracted from the context and their Jinja is parsed into an AST (using the Jinja library). A bunch of AST transforms are run to replace DBT-isms with SQLMesh native concepts as much as possible:
- `{{ ref() }}` and `{{ source() }}` calls are replaced with the actual model names they reference (where possible)
- `{% is_incremental() %}` blocks are removed
- ...etc

Jinja has no way of turning a Jinja AST back into a Jinja template string (since its goal is to generate Python code, not more Jinja) so I wrote a `JinjaGenerator` class to go from AST back to `str`.

If, after applying all the Jinja AST transforms, there is still some Jinja left in the result then it is written to the SQLMesh model file as a Jinja model surrounded with `JINJA_QUERY_BEGIN; JINJA_END;` blocks.

If there is no Jinja left after applying the transforms, it is written directly as a native SQL model with no Jinja wrapping.

Macros that call into DBT packages are also handled. The dependency tree is migrated and put in the target folder under `macros/__dbt_packages__` so that the macro hierarchy is still available when the macros are called on the SQLMesh side.

A migrated project isnt _truly_ native until all the dbt-isms have been removed. If the native loader detects it is loading a migrated project, it injects DBT shims into the Jinja context to make the migrated macros still work. The bulk of the DBT shim code is re-used from our existing DBT loader.

Known limitations:
 - The source DBT project must be loadable by the SQLMesh DBT loader. It doesnt need to be _runnable_, just loadable.
 - Currently only works for BigQuery and DuckDB as these were the initial focus. Other DB types can be added by implementing `from_sqlmesh()` in the relevant dbt `TargetConfig` class
 - Jinja handles whitespace stripping at parse time, so the AST has no idea if eg a `{%-` or `{%` block was used. This can lead to less than ideal formatting once the AST transforms are run and the resulting AST is turned back into a Jinja string.
 - Only the jinja constructs that were present in the test projects are handled by the Jinja generator, some more esoteric AST nodes are currently not handled
 - Audits are not generalized by our DBT loader so each model currently gets its own version as an inline audit
 - `{{ source() }}` calls on the DBT side that used dynamic inputs and were aliased in the DBT config are not correctly migrated
 - pre/post hooks / statements are not currently handled


sqlmesh/cli/main.py
```diff
@@ -1203,3 +1204,48 @@ def state_import(obj: Context, input_file: Path, replace: bool, no_confirm: bool
...
+    "--external-models-file/--no-external-models-file",
+    is_flag=True,
+    default=True,
+    help="Generate external_models.yaml (requires connectivity to the data warehouse)",
+)
+@click.option("--no-prompts", is_flag=True, help="Disable interactive prompts", default=False)
+@click.pass_obj
+@error_handler
+@cli_analytics
+def dbt_convert(
```
izeigerman, line 1240: Should we instead extend the `init` command like we do for dlt generation?

sqlmesh/core/model/definition.py
```diff
@@ -2448,13 +2451,42 @@ def _create_model(
...
         )
     else:
         jinja_macros = JinjaMacroRegistry()
 
-    for jinja_macro in jinja_macros.root_macros.values():
-        used_variables.update(extract_macro_references_and_variables(jinja_macro.definition)[1])
+    # extract {{ var() }} references used in all jinja macro dependencies to check for any variables specific
+    # to a migrated DBT package and resolve them accordingly
+    # vars are added into __sqlmesh_vars__ in the Python env so that the native SQLMesh var() function can resolve them
+    if migrated_dbt_project_name:
```
izeigerman, line 2464: Can this be encapsulated into its own function?

sqlmesh/core/model/kind.py
```diff
@@ -491,6 +491,18 @@ def _merge_filter_validator(
 
         return v.transform(d.replace_merge_table_aliases)
 
+    @field_validator("batch_concurrency", mode="before")
```
izeigerman, line 494: Why is this needed? There's already a validator for this field

sqlmesh/dbt/converter/console.py
```diff
@@ -0,0 +1,106 @@
...
+        "[progress.percentage]{task.percentage:>3.1f}%",
+        "•",
+        MofNCompleteColumn(),
+        "•",
+        TimeElapsedColumn(),
+        console=console,
+    )
+
+
+class DbtConversionConsole(TerminalConsole):
```
izeigerman, line 30: Does this need to inherit `TerminalConsole`?

sqlmesh/core/loader.py
```diff
@@ -384,15 +389,42 @@ def _raise_failed_to_load_model_error(self, path: Path, error: t.Union[str, Exce
 class SqlMeshLoader(Loader):
     """Loads macros and models for a context using the SQLMesh file formats"""
 
+    @property
+    def is_migrated_dbt_project(self) -> bool:
+        return self.migrated_dbt_project_name is not None
+
+    @property
+    def migrated_dbt_project_name(self) -> t.Optional[str]:
+        return self.config.variables.get(c.MIGRATED_DBT_PROJECT_NAME)
```
izeigerman: Since we know from the config that this a migrated project, can we put all of the logic below into a separate loader that inherits from `SqlMeshLoader`?

sqlmesh/core/loader.py
```diff
@@ -417,16 +449,51 @@ def _load_scripts(self) -> t.Tuple[MacroRegistry, JinjaMacroRegistry]:
...
 
         self._macros_max_mtime = macros_max_mtime
 
         macros = macro.get_registry()
         macro.set_registry(standard_macros)
 
+        if self.is_migrated_dbt_project:
+            from sqlmesh.dbt.target import TARGET_TYPE_TO_CONFIG_CLASS
+
+            connection_config = self.context._connection_config
```
izeigerman: Why not use the equivalent public method?

sqlmesh/core/loader.py
```diff
@@ -417,16 +449,51 @@ def _load_scripts(self) -> t.Tuple[MacroRegistry, JinjaMacroRegistry]:
...
+                        {
+                            "target": dbt_config_type.from_sqlmesh(
+                                self.context._connection_config,
+                                name=self.config.default_gateway_name,
+                            ).attribute_dict()
+                        }
+                    )
+                except NotImplementedError:
+                    raise ConfigError(
+                        f"No DBT 'Target Type' mapping for connection type: {connection_config.type_}"
```
izeigerman:
```suggestion
                        f"Unsupported dbt target type: {connection_config.type_}"
```

sqlmesh/core/model/definition.py
```diff
@@ -2448,13 +2451,42 @@ def _create_model(
...
+        # note: JinjaMacroRegistry is trimmed here so "all_macros" should be just be all the macros used by this model
+        for _, _, jinja_macro in jinja_macros.all_macros:
+            _, extracted_variable_names = extract_macro_references_and_variables(
```
izeigerman: Previously we were doing this unconditionally, but now it appears that we only extract if this is a migrated dbt project? Why is that?

sqlmesh/dbt/converter/common.py
```diff
@@ -0,0 +1,40 @@
...
+SQLGlotTransform = t.Callable[[exp.Expression], t.Optional[exp.Expression]]
+
+
+def _sqlmesh_predefined_macro_variables() -> t.Set[str]:
+    def _gen() -> t.Iterable[str]:
+        for suffix in ("dt", "date", "ds", "ts", "tstz", "hour", "epoch", "millis"):
+            for prefix in ("start", "end", "execution"):
+                yield f"{prefix}_{suffix}"
+
+        for item in ("runtime_stage", "gateway", "this_model", "this_env", "model_kind_name"):
```
izeigerman, line 22: There's no way we'll remember to update this once we add more macro variables

sqlmesh/core/loader.py
```diff
@@ -417,16 +449,51 @@ def _load_scripts(self) -> t.Tuple[MacroRegistry, JinjaMacroRegistry]:
...
             with open(path, "r", encoding="utf-8") as file:
-                jinja_macros.add_macros(
-                    extractor.extract(file.read(), dialect=self.config.model_defaults.dialect)
-                )
+                try:
+                    package: t.Optional[str] = None
+                    if self.is_migrated_dbt_project:
+                        if path.is_relative_to(migrated_dbt_package_base_path):
+                            package = str(
+                                path.relative_to(migrated_dbt_package_base_path).parents[0]
```
izeigerman: Don't we have `infer_dbt_package_from_path` for this?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+
+    _fix_meta_field_converters()
+
+    console.log_message(
+        f"Converting project at '{src}' to '{dest}' (external models: {output_external_models})"
+    )
+
+    ctx, dbt_project = _load_project(src)
+    dbt_load_context = dbt_project.context
+    dbt_project.load
```
izeigerman: What does this do?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+        # External models
+        if len(ctx.configs) == 1:
+            # The SQLMesh context is currently tied to the source DBT project location
+            # Update the config paths to be in the target SQLMesh project location so that create_external_models() writes the external models file in the right place
+            ctx.configs = {dest: config for _, config in ctx.configs.items()}
+            console.log_message("Creating external models file")
+            ctx.create_external_models()
+        else:
+            console.log_warning(
+                "More than 1 config detected; not sure how to handle this. External models file not created"
```
izeigerman: Not sure this is a helpful message. What can user do to mitigate this?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+
+    # Generate SQLMesh config
+    # TODO: read all profiles from config and convert to gateways instead of just the current profile?
+    console.log_message("Writing SQLMesh config")
+    new_config = _generate_sqlmesh_config(ctx, dbt_project, dbt_load_context)
+    (dest / "config.yml").write_text(yaml.dump(new_config))
+
+    if output_external_models:
+        # External models
+        if len(ctx.configs) == 1:
```
izeigerman: How can it be more than 1?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+
+    console.complete_macros_conversion()
+
+    # Generate SQLMesh config
+    # TODO: read all profiles from config and convert to gateways instead of just the current profile?
+    console.log_message("Writing SQLMesh config")
+    new_config = _generate_sqlmesh_config(ctx, dbt_project, dbt_load_context)
+    (dest / "config.yml").write_text(yaml.dump(new_config))
+
+    if output_external_models:
```
izeigerman: Do we even need to do this as part of the conversion? Can't this just be done as a follow-up command?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+        ):  # note: should be no PythonModel's in a DBT project
+            model_to_render = model.model_copy(
+                update=dict(depends_on_=filtered_depends_on(ctx, model))
+            )
+
+            if isinstance(model, SqlModel):
+                test_rendered = _render(model.query)
+                if is_self_referencing(model, test_rendered):
+                    report.self_referencing_models.append((model_output_path, model))
+                    # if it's self referencing, we need to output the columns or we get an error like: Error: Self-referencing models require inferrable column types.
```
izeigerman: Can we still convert them as INCREMENTAL_UNMANAGED and keep the `is_incremental()` macro?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+
+    console.log_message("All done")
+
+
+def _fix_meta_field_converters() -> None:
+    from sqlmesh.core.model.definition import META_FIELD_CONVERTER
+
+    orig_depends_on = META_FIELD_CONVERTER["depends_on_"]
+
+    # The 'depends_on' key always gets rendered but the value is missing if its an empty list
```
izeigerman: I'm confused why we need this. And even if we do why can't we update the original converter

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+                    output_columns = model.columns_to_types or {
+                        "COLUMN_TYPES_CANNOT_BE_STATICALLY_ANALYZED": exp.DataType.build("unknown")
+                    }
+                    model_to_render = model_to_render.model_copy(
+                        update=dict(columns_to_types_=output_columns)
+                    )
+
+        rendered_queries = [
+            _render(q)
+            for q in model_to_render.render_definition(render_query=False, include_python=False)
```
izeigerman, line 260: I'm pretty sure this will render all default values too. Can we reduce clutter by omitting those?

sqlmesh/dbt/converter/convert.py
```diff
@@ -0,0 +1,502 @@
...
+        # however, we still need to keep it for models where the entire query is a macro call, otherwise the SQLMesh native loader will throw:
+        # - ConfigError: Dependencies must be provided explicitly for models that can be rendered only at runtime
+        model_to_render = model.model_copy(
+            update=dict(depends_on_=None if len(model.depends_on) > 0 else set())
+        )
+        if isinstance(
+            model, (SqlModel, SeedModel)
+        ):  # note: should be no PythonModel's in a DBT project
+            model_to_render = model.model_copy(
+                update=dict(depends_on_=filtered_depends_on(ctx, model))
```
izeigerman: Isn't this something we can get from `DbtContext.models.dependencies.refs`?

sqlmesh/dbt/converter/jinja_builtins.py
```diff
@@ -0,0 +1,107 @@
...
+def migrated_ref(
+    dbt_api: Api,
+    database: t.Optional[str] = None,
+    schema: t.Optional[str] = None,
+    identifier: t.Optional[str] = None,
+    version: t.Optional[int] = None,
+    sqlmesh_model_name: t.Optional[str] = None,
+) -> BaseRelation:
+    if version:
+        raise ValueError("Handling version isnt implemented")
```
izeigerman:
```suggestion
        raise ValueError("dbt model versions are not supported in converted projects.")
```

sqlmesh/dbt/converter/jinja_builtins.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def source(dbt_api: Api, source_name: str, table_name: str) -> BaseRelation:
+        # some source() calls cant be converted to __migrated_source() calls because they contain dynamic parameters
+        # this is a fallback and will be wrong in some situations because `sources` in DBT can be aliased in config
+        # TODO: maybe we migrate sources into the SQLMesh variables so we can look them up here?
+        return dbt_api.Relation.create(database=source_name, identifier=table_name)
+
+    def ref(dbt_api: Api, ref_name: str, package: t.Optional[str] = None) -> BaseRelation:
+        # some ref() calls cant be converted to __migrated_ref() calls because they contain dynamic parameters
+        raise NotImplementedError(f"Not sure how to resolve ref: {ref_name}")
```
izeigerman: We need a more actionable error here

sqlmesh/dbt/converter/sqlglot_transforms.py
```diff
@@ -0,0 +1,26 @@
+from sqlglot import exp
+import sqlmesh.core.dialect as d
+from sqlmesh.dbt.converter.common import SQLMESH_PREDEFINED_MACRO_VARIABLES, SQLGlotTransform
+
+
+def unwrap_macros_in_string_literals() -> SQLGlotTransform:
```
izeigerman: Not sure we need a whole separate module for this. Can't this be a part of `jinja.py`?

sqlmesh/cli/main.py
erindru: We could but i'm not so sure.

Right now imo the `init` command is complicated by dlt generation with a bunch of `--dlt-*` specific parameters that are defined on the command but meaningless unless `--template` is set to `dlt`.

I see the dbt conversion process also needing a bunch of parameters to control how the conversion is performed. It's also not really a one-off `init`, I expect it would probably be run a few times with different parameters (or model changes) until the output looks right.

sqlmesh/core/model/kind.py
erindru: Oh, are you able to point me to the existing validator? I'm having trouble locating it

I added this because if you called `model.render_definition()` and it wrote something like:
```
        MODEL (
            name db.table,
                unique_key a,
                batch_concurrency 1
            )
        );
```

And then you tried to read it back, you get a Pydantic error:
```
E           pydantic_core._pydantic_core.ValidationError: 1 validation error for IncrementalByUniqueKeyKind
E           batch_concurrency
E             Input should be 1 [type=literal_error, input_value=Literal(this=1, is_string=False), input_type=Literal]
```

I added a test in `test_model.py` to show this behaviour. It fails because it wants an `int` but it gets given an `exp.Literal` instead

sqlmesh/dbt/converter/common.py
erindru: Yeah I added this when I was in "hack whatever it takes to get the thing to work" mode.

Is there somewhere that can be referenced to get a definitive list?

sqlmesh/dbt/converter/console.py
erindru: I was using it for things like `_print` and `_confirm`. How would you prefer this was structured?

sqlmesh/dbt/converter/convert.py
erindru: I suppose it could, I just wanted the conversion to be "complete" in the sense it would produce a native project with everything in place.

Shall I remove it to simplify things?

sqlmesh/dbt/converter/convert.py
erindru: It's more to trigger a bug report. I didn't have an example of how someone could get into this situation but the data structure makes it possible.

If someone did hit this then it would be helpful to understand how so we can add a test

sqlmesh/dbt/converter/convert.py
erindru: We need this because if we don't have it, the following can occur when calling `model.render_definition()`:

```
MODEL (
   ...,
   depends_on ,
   ...
)
```

Which is a syntax error when you try to read it back.

Initially I was trying to wrap things rather than change them. I'll update the original converter instead

sqlmesh/dbt/converter/convert.py
erindru: Are you referring to the `include_defaults` parameter? It defaults to `False`

sqlmesh/dbt/converter/jinja.py
```diff
@@ -0,0 +1,560 @@
...
+
+
+def lpad_windowed(iterable: t.Iterable[j.Node]) -> t.Iterator[t.Tuple[t.Optional[j.Node], j.Node]]:
+    for prev, curr in windowed(chain([None], iterable), 2):
+        if curr is None:
+            raise ValueError("Current item cannot be None")
+        yield prev, curr
+
+
+class JinjaGenerator:
```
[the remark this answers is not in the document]
erindru: I did think about that but having the class made it easy to do the following:

```
generator_fn_name = f"_generate_{node_type.__name__.lower()}"

if generator_fn := getattr(self, generator_fn_name, None):
    generator_fn(...)
```

i.e it gives a handle `self` to call `getattr` on and limits the functions to just what are available on this class

sqlmesh/dbt/converter/sqlglot_transforms.py
erindru: It can be, I originally had a bunch more SQLGlot transforms but they turned out to be more fragile than I hoped so I removed them all except this one

sqlmesh/dbt/converter/convert.py
erindru: `ctx.configs` is defined as `t.Dict[Path, Config]` so based on that there can be any number.

However having said that I don't think this check matters, i'll remove it

sqlmesh/core/model/definition.py
erindru: It's an oversight although surprisingly no tests failed. I'll revert it

sqlmesh/core/loader.py
erindru: We sure do, and it was written after this. I'll update it

sqlmesh/dbt/converter/convert.py
erindru: Absolutely nothing, i'll remove it

sqlmesh/dbt/converter/convert.py
erindru: Some are legitimately self referencing, `is_self_referencing` also returns `True` if `model.depends_on_self`.

For the others that only become self referencing after removing `is_incremental()`, I guess we can test for that and not apply the transform which will result in them getting output with the `is_incremental()` checks intact

sqlmesh/dbt/converter/convert.py
erindru: Not really.

The problem here is, in order to work with the native loader, `model.depends_on` needs to point to other **models** in the project. Not FQN strings or DBT ref names.

Given the `sushi_dbt` example project, in a model loaded by the `DbtLoader`, `model.depends_on` looks something like:

```
>>> ctx.models['"memory"."sushi"."customers"'].depends_on
{'"memory"."sushi_raw"."orders"'}
```

That is, it is a set of FQN strings, not model names

If you call `model.render_definition()` with that, you get:
```
MODEL (
name sushi.customers,
...
depends_on ("memory"."sushi_raw"."orders"),
...
)
```

This cant be loaded by the native loader, it just throws an error `ConfigError: Failed to load model definition: 'Dot' object has no attribute 'catalog'`.

The items in this list need to be model names, not model FQN's.

`DbtContext.models.dependencies.refs` doesn't return model names, it returns ref names from the flat DBT global namespace. For example:
```
>>> dbt_load_context.models["customers"].dependencies.refs
{'orders'}
```

`"orders"` isnt a valid value for `depends_on`, it needs to be `sushi_raw.orders`. Why? Because the model name is `sushi_raw.orders` and not `orders`:

```
>>> ctx.models['"memory"."sushi_raw"."orders"'].name
'sushi_raw.orders'
```

So what `filtered_depends_on()` does is resolve the FQN strings in `model.depends_on` to actual model names so the result can be loaded by the native loader.

Granted, its implementation has changed from the original version, so i'll rename it to `resolve_fqns_to_model_names()` to better communicate what it's trying to do

sqlmesh/core/model/definition.py
erindru: I've moved the logic to `_extract_migrated_dbt_variable_references`

sqlmesh/core/loader.py
erindru: I've moved the logic to a `MigratedDbtProjectLoader`.

I couldn't see a good way to hook into the `_load_scripts` lifecycle of the parent `SqlMeshLoader` to change the builtins module and infer / set the dbt package name for macros under `__dbt_packages__/`, so I ended up duplicating a bunch of the existing code.

Not super happy with this but modifying the parent `SqlMeshLoader` to take callbacks to set these things would be just as invasive as the original changes

sqlmesh/core/loader.py
erindru: I've updated this

sqlmesh/core/model/definition.py
erindru: I've added it to the `else` condition

sqlmesh/dbt/converter/convert.py
erindru: I've removed the external model creation entirely

sqlmesh/dbt/converter/convert.py
erindru: I ended up removing it

sqlmesh/dbt/converter/convert.py
erindru: I've updated this to test-load the converted model.

If it throws the  `Self-referencing models require inferrable column types` error then I re-run the conversion but excluding the transforms that remove `is_incremental()`.

This does work although on the 1000+ model test project it does have the downside of `context.upsert_model()`  busts the fingerprint cache so it spends a bunch of extra time re-rendering queries of the dependency tree

sqlmesh/dbt/converter/convert.py
erindru: I've updated the original converter

sqlmesh/dbt/converter/jinja_builtins.py
erindru: I've updated this to `Unable to resolve ref: {ref_name}. Please replace it with an actual model name or use a SQLMesh macro to generate dynamic model name.`

sqlmesh/dbt/converter/sqlglot_transforms.py
erindru: I've moved this and removed the `sqlglot_transforms` module

sqlmesh/core/model/kind.py
izeigerman: The validator is defined through the type annotation: 
https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/pydantic.py#L342C5-L342C23 which in turn calls https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/pydantic.py#L204

sqlmesh/dbt/converter/console.py
izeigerman: Those are just one-liners, and I’m not sure they’re worth coupling over

sqlmesh/dbt/converter/jinja.py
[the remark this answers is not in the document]
izeigerman: Just curious: can't you replace `self` with `__module__`? No objections otherwise

sqlmesh/dbt/converter/common.py
izeigerman: Good question. I don't think we have any centralized place atm.

sqlmesh/dbt/converter/convert.py
izeigerman: I see what you're saying, but I think this can still be done through the manifest helper. Eg.
1. Get the ref from `DbtContext.models.dependencies.refs`
2. Get the model config from the manifest using the ref: `DbtContext.manifest.models()[<extracted_ref]`
3. Use the returned `ModelConfig` to get the `canonical_name` of the model. Which should be `<schema>.<name>` and not FQN.

sqlmesh/core/model/kind.py
erindru: Ah, that's true for other model kinds but not `IncrementalByUniqueKeyKind` where it was defined as:

```
batch_concurrency: t.Literal[1] = 1
```

which doesnt invoke the validator.

However it looks like changing it to the following does the trick:

```
batch_concurrency: t.Annotated[t.Literal[1], BeforeValidator(positive_int_validator)] = 1
```

sqlmesh/dbt/converter/jinja.py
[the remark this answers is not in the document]
erindru: Probably but it does mean that it's easier to accidentally call something unrelated that is defined in the module but not part of the generator.

I guess i'm just thinking in terms of self-containment. I could probably add a `generate()` top level module function and make the `JinjaGenerator` class private to make this API more ergonomic, but I do see it evolving a bit as this feature evolves

sqlmesh/dbt/converter/console.py
erindru: No problem, i've made this class standalone now

sqlmesh/dbt/converter/convert.py
erindru: > Get the ref from DbtContext.models.dependencies.refs

^ how does this part work? `DbtContext.models` returns DBT model names, not SQLMesh model names.

The entire conversion process is currently based around the loaded SQLMesh `Context`, not the `DbtContext`. As part of building the SQLMesh `Context`, the DBT `ModelConfig` objects have had `to_sqlmesh()` called on them which turns them into SQLMesh `Model` objects.

In the `sushi_dbt` project, this for example takes the DBT model `customers` and converts it to the SQLMesh model `default.customers`.

When the conversion process iterates through SQLMesh models in the SQLMesh context and converts them, let's say it just hit `default.customers`. The environment looks something like:

```
>>> model.name
'default.customers'

>>> model.depends_on
{'"sushi_dbt"."default_raw"."orders"'}

>>> dbt_context.models.keys()
dict_keys(['customer_revenue_by_day', 'customers', 'top_waiters', 'waiter_revenue_by_day_v1', 'waiter_revenue_by_day_v2', 'waiter_as_customer_by_day', 'waiters', 'test_nested_ref'])

>>> sqlmesh_context.models.keys()
dict_keys([..SNIP, '"sushi_dbt"."default_raw"."orders"', ...SNIP])
```

I couldn't find a reverse mapping to take the `model.name` of `sushi.customers` and map it back to the DBT ref `customers` so it can be looked up in `dbt_context.models`.

But even if one existed, wouldn't this just be repeating work the DBT loader already did anyway?

I suppose I could try poking into `model.jinja_macros.global_objs` to try and figure something out but I don't see how that is simpler or better than the current approach of just looking up the fqn strings in `sqlmesh_context.models` which is already indexed by the fqn strings?

PR #4496, opened 2025-05-22
Chore: Test retry flakiness and migrate pytest config to pyproject.toml

erindru:
This was motivated by the following [test failure on main](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/20958/workflows/008e93a4-5902-4254-b558-27b66a9af61d/jobs/211276):

```
ValueError: ctx did not yield a value
```

This was a regression I introduced in #4406 

The tl;dr is that `pytest-retry` [deliberately wont retry](https://github.com/str0zzapreti/pytest-retry/issues/33#issuecomment-1962798422) failures in setting up test fixtures. Our `TestContext` fixture can fail when engines like Trino decide to throw an exception on a `CREATE SCHEMA` statement instead of actually creating the schema.

Not yielding the fixture at all or yielding `None` still results in an error in the fixture setup phase which means the test wont be retried.

So i've updated it to capture / log the exception and yield the `TestContext` to the test anyway. In doing so, I am anticipating an error in the test itself instead of the fixture, which will trigger `pytest-retry` to rerun the individual test instead of failing the entire test run like what happens now.

I also wanted to add `retry_delay=10` as a default to all the integration tests that were using `--retries`, which motivated moving the pytest config from `pytest.ini` to `pyproject.toml` to make it consistent with our other tooling config


.circleci/continue_config.yml
```diff
@@ -333,10 +333,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru, line 336: TODO: revert prior to merge

tests/core/engine_adapter/integration/conftest.py
```diff
@@ -106,25 +106,23 @@ def _create(
...
 
-        skip = False
         try:
             ctx.init()
-        except Exception:
-            # We need to catch this exception because if there is an error during setup, pytest-retry aborts immediately
-            # instead of retrying
+        except:
+            # pytest-retry doesnt work if there are errors in fixture setup (ref: https://github.com/str0zzapreti/pytest-retry/issues/33 )
+            # we can do is log the exception and return a partially-initialized context to the test, which should
```
georgesittas, line 113:
```suggestion
            # what we can do is log the exception and return a partially-initialized context to the test, which should
```

georgesittas (approved):

PR #4498, opened 2025-05-22
Fix: prompt if there are any uncategorized snapshots in a plan

georgesittas:
See the added test for reference. The issue that we saw in the wild looked something like this:

```
-  event_date
+  event_date,
+  1 AS z
 WHERE
   event_date BETWEEN @start_date AND @end_date
Directly Modified: sqlmesh_example.incremental_model
└── Indirectly Modified Children:
Models needing backfill:
└── sqlmesh_example.incremental_model: [2020-01-01 - 2025-05-20]
Error: Can't apply a plan with uncategorized changes.
```

This happened after running `sqlmesh plan` with a config that switches off auto-categorization:

```python
PlanConfig(
    auto_categorize_changes=CategorizerConfig(
        external=AutoCategorizationMode.OFF,
        python=AutoCategorizationMode.OFF,
        sql=AutoCategorizationMode.OFF,
        seed=AutoCategorizationMode.OFF,
    ),
)
```

One workaround for this^ issue today is to set `no_prompts=False` in the plan config, but knowing this or having to reach out for it is not as good of an experience, which is what motivated this PR.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1269,8 +1269,11 @@ class GenericContext(BaseContext, t.Generic[C]):
             skip_linter=skip_linter,
         )
 
-        if no_auto_categorization:
+        plan = plan_builder.build()
+
+        if no_auto_categorization or plan.uncategorized:
             # Prompts are required if the auto categorization is disabled
+            # or if there are any uncategorized snapshots in the plan
             no_prompts = False
 
         self.console.plan(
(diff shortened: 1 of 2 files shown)


sqlmesh/core/context.py
```diff
@@ -1269,8 +1269,11 @@ def plan(
             skip_linter=skip_linter,
         )
 
-        if no_auto_categorization:
+        plan = plan_builder.build()
+
+        if no_auto_categorization or plan.uncategorized:
```
georgesittas, line 1274: FYI: I intentionally did not remove the following check in `apply` as a defensive mechanism: https://github.com/TobikoData/sqlmesh/blob/ec92d47868f03f1b3d29311d7bddbb89c68f89c5/sqlmesh/core/context.py#L1545-L1546

izeigerman (approved):

PR #4506, opened 2025-05-22
Chore: add pyproject.toml in examples/custom_materializations

georgesittas:
Quick fix to silence the warning shown below:

```bash
$ make install-dev
...
Building wheels for collected packages: custom_materializations, sqlmesh
  DEPRECATION: Building 'custom_materializations' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of 'custom_materializations'. Discussion can be found at https://github.com/pypa/pip/issues/6334
```

References:
- https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-pyproject-toml-mandatory
- https://github.com/TobikoData/sqlmesh/blob/a1c48404b8ef476e72e95deab0d8d3b7bbaadbcc/pyproject.toml#L144-L146

### DIFF
--- a/examples/custom_materializations/pyproject.toml
+++ b/examples/custom_materializations/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 61.0", "setuptools_scm"]
+build-backend = "setuptools.build_meta"


erindru:
Shouldnt we remove `setup.py` altogether and set the entrypoint in `pyproject.toml` as described here: https://sqlmesh.readthedocs.io/en/stable/guides/custom_materializations/#python-packaging ?

I dont think we are using any feature that requires `setup.py` to be present

georgesittas:
Ah, I missed that page. Yeah, this makes sense. I'll open a followup PR shortly.

georgesittas:
@erindru https://github.com/TobikoData/sqlmesh/pull/4514

PR #4509, opened 2025-05-22
Chore!: configure logging before loading config to log errors properly

georgesittas:
This [commit](https://github.com/TobikoData/sqlmesh/commit/cc3e4931ed4ef361f90e340b4bc18793009935c5) introduced new error handling logic for connection configs: if sqlmesh fails to import an engine driver, then it suggests running `pip install` as a possible solution to the issue.

However, problems can arise at import time even if a driver _is_ installed. For example, `pymssql` is known to have [compatibility issues](https://github.com/pymssql/pymssql/issues/769) with M-series Mac systems. This leads to confusing error messages as shown below:

```
((.venv) ) ➜  sqlmesh info
Error: Failed to import the 'mssql' engine library. Please run `pip install "sqlmesh[mssql]"`.
((.venv) ) ➜  playground pip show pymssql  # The above message is misleading, pymssql is already installed
Name: pymssql
Version: 2.3.4
...
((.venv) ) ➜  SQLMESH_DEBUG=1 sqlmesh info  # This includes the actual error in the stack trace
...
  File "src/pymssql/_pymssql.pyx", line 1, in init pymssql._pymssql
ImportError: dlopen(.../sqlmesh/.venv/lib/python3.12/site-packages/pymssql/_mssql.cpython-312-darwin.so, 0x0002): symbol not found in flat namespace '_bcp_batch'
```

This PR addresses the issue by ensuring that the actual import error is raised when in debug mode, or logged otherwise.

One obstacle I faced was that `configure_logging` currently [depends on](https://github.com/TobikoData/sqlmesh/blob/ac25787cf4cfffb8fb3654392c6c321ef07a7941/sqlmesh/cli/main.py#L103-L112) `load_configs`, because the latter reads the `log_limit`, which is needed for purging excess log files.

I worked around this issue by decoupling the excess log file purging logic from `configure_logging`, so that we can run the latter before loading the configs, allowing the errors to be properly logged in a file instead of being printed in stdout.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -26,7 +26,7 @@ from sqlmesh.core.config.common import (
 )
 from sqlmesh.core.engine_adapter.shared import CatalogSupport
 from sqlmesh.core.engine_adapter import EngineAdapter
-from sqlmesh.utils import str_to_bool
+from sqlmesh.utils import debug_mode_enabled, str_to_bool
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.pydantic import (
     ValidationInfo,
@@ -66,7 +66,11 @@ def _get_engine_import_validator(
             return data
         try:
             importlib.import_module(import_name)
-        except ImportError:
+        except ImportError as e:
+            if debug_mode_enabled():
+                raise
+
+            logger.error(str(e))
             raise ConfigError(
                 f"Failed to import the '{engine_type}' engine library. Please run `pip install \"sqlmesh[{extra_name}]\"`."
             )
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -66,13 +66,17 @@ def _get_engine_import_validator(
             return data
         try:
             importlib.import_module(import_name)
-        except ImportError as e:
+        except ImportError:
             if debug_mode_enabled():
                 raise
 
-            logger.error(str(e))
+            logger.exception("Failed to import the engine library")
+
             raise ConfigError(
-                f"Failed to import the '{engine_type}' engine library. Please run `pip install \"sqlmesh[{extra_name}]\"`."
+                f"Failed to import the '{engine_type}' engine library. This may be due to a missing "
+                "or incompatible installation. Please ensure the required dependency is installed by "
+                f'running: pip install "sqlmesh[{extra_name}]". For more details, check the logs '
+                "in the 'logs/' folder, or rerun the command with the '--debug' flag."
             )
 
         return data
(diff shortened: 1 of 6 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -66,7 +66,11 @@ def validate(cls: t.Any, data: t.Any) -> t.Any:
...
         try:
             importlib.import_module(import_name)
-        except ImportError:
+        except ImportError as e:
+            if debug_mode_enabled():
+                raise
+
+            logger.error(str(e))
             raise ConfigError(
                 f"Failed to import the '{engine_type}' engine library. Please run `pip install \"sqlmesh[{extra_name}]\"`."
```
georgesittas, line 75: I wonder if we should rephrase this, now that we've seen a `pip install` may not always be the issue. E.g.:

> This may be due to a missing or incompatible installation. Please ensure that the required dependency is installed by running `pip install \"sqlmesh[{extra_name}]\"`. For more details, check the generated log file under the 'logs/' directory or retry using the '--debug' option."

sqlmesh/core/config/connection.py
```diff
@@ -66,7 +66,11 @@ def validate(cls: t.Any, data: t.Any) -> t.Any:
             return data
         try:
             importlib.import_module(import_name)
-        except ImportError:
+        except ImportError as e:
+            if debug_mode_enabled():
+                raise
+
+            logger.error(str(e))
```
izeigerman, line 73: This swallows the original stacktrace. Maybe
```
logger.exception("Failed to import the engine library")
```
`exception` will pick up the correct exception from the context, no need to explicitly pass `e`.

sqlmesh/core/config/connection.py
izeigerman: > This may be due to a missing or incompatible installation. Please ensure the required dependency is installed by running: `pip install "sqlmesh[{extra_name}]"`. For more details, check the logs in the 'logs/' folder, or rerun the command with the '--debug' flag.

yes, good idea

izeigerman (approved):
LGTM once the comments are addressed

sqlmesh/core/config/connection.py
georgesittas: Perfect, TIL!

sqlmesh/core/config/connection.py
```diff
@@ -67,8 +67,16 @@ def validate(cls: t.Any, data: t.Any) -> t.Any:
...
+            if debug_mode_enabled():
+                raise
+
+            logger.exception("Failed to import the engine library")
+
             raise ConfigError(
-                f"Failed to import the '{engine_type}' engine library. Please run `pip install \"sqlmesh[{extra_name}]\"`."
+                f"Failed to import the '{engine_type}' engine library. This may be due to a missing "
+                "or incompatible installation. Please ensure the required dependency is installed by "
+                f'running: pip install "sqlmesh[{extra_name}]". For more details, check the logs '
```
izeigerman, line 78:
```suggestion
                f'running: `pip install "sqlmesh[{extra_name}]"`. For more details, check the logs '
```

PR #4510, opened 2025-05-22
Feat: expose model fqn in the macro evaluator

georgesittas:
There's no easy way to access the unresolved name of a model within a macro that it calls– `this_model` is usually the resolved version, corresponding to either a physical table, or a view if we're in the promoting stage.

This PR adds a new property in the macro evaluator that can be used to access the model's fully-qualified name. One example where the unresolved name is helpful can be seen in the added test.

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -173,6 +173,7 @@ class MacroEvaluator:
         default_catalog: t.Optional[str] = None,
         path: Path = Path(),
         environment_naming_info: t.Optional[EnvironmentNamingInfo] = None,
+        model_fqn: t.Optional[str] = None,
     ):
         self.dialect = dialect
         self.generator = MacroDialect().generator()
@@ -198,6 +199,7 @@ class MacroEvaluator:
         self._snapshots = snapshots if snapshots is not None else {}
         self._path = path
         self._environment_naming_info = environment_naming_info
+        self._model_fqn = model_fqn
 
         prepare_env(self.python_env, self.env)
         for k, v in self.python_env.items():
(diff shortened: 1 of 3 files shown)


tobymao (approved):
documentation?

georgesittas:
Yep, I'll update it.

sqlmesh/core/macros.py
```diff
@@ -476,6 +478,12 @@ def this_model(self) -> str:
             raise SQLMeshError("Model name is not available in the macro evaluator.")
         return this_model.sql(dialect=self.dialect, identify=True, comments=False)
 
+    @property
+    def this_model_fqn(self) -> str:
```
izeigerman, line 482: Do we need the same API in the `ExecutionContext`? Also what's wrong with `evaluator.this_model.fqn`?

izeigerman: Ah shit, `this_model` is not a model.

izeigerman: Shouldn't we expose `@this_model_fqn` then?

georgesittas: > Do we need the same API in the ExecutionContext?

I can look into that as well.

> Also what's wrong with evaluator.this_model.fqn?
> Ah shoot, this_model is not a model.
> Shouldn't we expose @this_model_fqn then?

So we actually have `locals["this_model"]` which is a `Table` reference. This is what `evaluator.this_model` is based on. The issue however is that it's [already resolved](https://github.com/TobikoData/sqlmesh/blob/73e2ef4a8d811461b8777eedec96101a710b9790/sqlmesh/core/renderer.py#L135-L140).

izeigerman: Right. My question is if we expose this through `evaluator` do we want to expose it directly in SQL too

georgesittas: Ah, yeah I was wondering the same, but won't we end up resolving the unresolved version after the macros are rendered? Or am I missing something? In which case, it'd be meaningless to expose that variable in `locals`.

izeigerman: I see

izeigerman (approved):

PR #4519, opened 2025-05-23
Feat: Allow supplying complex connection config fields as JSON strings

erindru:
Prior to this PR, there was no clear way to use an environment variable to set a "complex" field on a connection, such as the `scopes` property on a BigQuery connection which takes a list of strings, or the `keyfile_json` property which takes a dict.

Attempting to do so would just throw a Pydantic error:
```sh
$ SQLMESH__GATEWAYS__BIGQUERY__CONNECTION__SCOPES='["http://..."]' sqlmesh info
Error: Invalid 'bigquery' connection config:
  Invalid field 'scopes':
    Input should be a valid list

Verify your config.yaml and environment variables.
```

This PR adds a generic `mode=before` validator onto the `ConnectionConfig` base class to check if any "complex type" fields are being set as a string. If they are, and the string contains JSON, it parses the JSON string to a Python value before continuing.

This means that all the connection config subclasses transparently gain the ability to handle JSON values without having to manually add validators to existing fields or remember to add validators to new fields

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -13,6 +13,7 @@ from functools import partial
 
 import pydantic
 from pydantic import Field
+from pydantic_core import from_json
 from packaging import version
 from sqlglot import exp
 from sqlglot.helper import subclasses
@@ -33,6 +34,7 @@ from sqlmesh.utils.pydantic import (
     field_validator,
     model_validator,
     validation_error_message,
+    get_concrete_types_from_typehint,
 )
 from sqlmesh.utils.aws import validate_s3_uri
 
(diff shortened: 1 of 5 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -177,6 +179,41 @@ def get_catalog(self) -> t.Optional[str]:
...
+        of turning this string into the corresponding Python list or dict.
+
+        Rather than doing this piecemeal on every config subclass, this provides a generic implementatation
+        to identify fields that may be be supplied as JSON strings and handle them transparently
+        """
+        if data and isinstance(data, dict):
+            for maybe_json_field_name in cls._get_list_and_dict_field_names():
+                if (value := data.get(maybe_json_field_name)) and isinstance(value, str):
+                    # crude JSON check as we dont want to try and parse every string we get
+                    if value.startswith("{") or value.startswith("["):
```
izeigerman, line 199: Should we trim the string to remove any whitespace characters?

izeigerman (approved):

sqlmesh/core/config/connection.py
erindru: I was about to say no and err on the side of being strict, but then I saw [this](https://stackoverflow.com/questions/4150621/are-whitespace-characters-insignificant-in-json) and who am I to argue with RFC 4627. I've adjusted the logic

PR #4525, opened 2025-05-23
Chore: Trey edits to macros concepts doc this_model section

treysp:


docs/concepts/macros/sqlmesh_macros.md
```diff
@@ -1681,32 +1681,70 @@ def some_macro(evaluator):
     ...
 ```
 
-#### Accessing the invoking model's resolved name
+#### Accessing model, physical table, and virtual layer view names
 
-The physical table name of the invoking model can be accessed within a Python macro function through its evaluation context's `this_model` property.
+All SQLMesh models have a name in their `MODEL` specification. We refer to that as the model's "unqualified" and "unresolved" name because it may not correspond to any specific object in the SQL engine.
```
georgesittas, line 1686: I'd remove the "unqualified" bit because a model's name can be qualified as well.

georgesittas (approved):
Way cleaner, thanks a lot Trey!

Issue #4472, opened 2025-05-20
mssql: FULL materialization strategy is extremely slow using MERGE INTO existing table ... ON 1 = 0


martinburch:
Following SQLMesh's initial (fast) `plan` of a `FULL` model on the `mssql` adapter, subsequent SQLMesh `run`s, which rebuild the model completely, are very slow (hours or more).

This is caused by using the pattern MERGE INTO ... ON (1 = 0) existing table, like this
```
MERGE INTO [schema].[schema__table__1460139764] AS [__MERGE_TARGET__]
USING ( ... ) AS [__MERGE_SOURCE__]
ON (1 = 0)
WHEN NOT MATCHED BY SOURCE THEN DELETE
WHEN NOT MATCHED THEN INSERT ( ... ) VALUES ( ... );
```
I can see this query is `suspended` in the MS SQL query engine with 100% CPU use for hours.

The materialisation strategy for `FULL` `mssql` models is not specified in the docs, but no other database uses `MERGE` for `FULL`.
https://sqlmesh.readthedocs.io/en/stable/concepts/models/model_kinds/?h=merge#materialization-strategy_2
I suggest using the `Postgres` method for `mssql`: `DROP TABLE, CREATE TABLE, INSERT`

> _How do I know the MERGE pattern is the problem?_
> 
> When I terminate the stuck `run` process and then
> ```
> TRUNCATE TABLE [schema].[schema__table__1460139764];
> ```
> and re-run `sqlmesh run` then the `run` is as fast as the initial `plan`. So SQLMesh is still doing the `MERGE INTO` except now it's comparing `0` records.
> 
> The problem is one of squares: if I have a table of 2 million records, and I'm doing a full rebuild, I have 2 million new records which must be compared, and that's 4 trillion comparisons. Even if each comparison only took 1 millisecond, the comparisons would take over 100 years to complete!

martinburch:
I've tried implementing the TRUNCATE TABLE hack in the model's pre-statement but SQLMesh is not happy, `Could not render or parse jinja` (_yes, the jinja itself is valid_ see http://jinja.quantprogramming.com) for something like...

```
JINJA_STATEMENT_BEGIN;
  {% if runtime_stage == 'creating' %}
    IF OBJECT_ID('{{ this_model }}') IS NOT NULL TRUNCATE TABLE {{ this_model }};
  {% endif %}
JINJA_END;
```

[...]

PR #4531, opened 2025-05-26
Fix(mssql): Use truncate+insert for FULL models instead of merge

erindru:
Addresses issue #4472

This PR:
 - Overrides `delete_from()` to use `TRUNCATE` if there is no WHERE clause. This works because [truncate is works within a transaction](https://learn.microsoft.com/en-us/sql/t-sql/statements/truncate-table-transact-sql?view=sql-server-ver15#remarks) and is more efficient than using `DELETE`
 - Uses the `DELETE+INSERT` strategy instead of `MERGE` if `_insert_overwrite_by_condition` is called without a condition. `_insert_overwrite_by_condition` is used by `replace_table()` which is how the `FULL` materialization strategy is implemented.

The net result is that full data replacements in existing tables become a lot more efficient

### DIFF
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -10,7 +10,11 @@ from pandas.api.types import is_datetime64_any_dtype  # type: ignore
 from sqlglot import exp
 
 from sqlmesh.core.dialect import to_schema
-from sqlmesh.core.engine_adapter.base import EngineAdapterWithIndexSupport
+from sqlmesh.core.engine_adapter.base import (
+    EngineAdapterWithIndexSupport,
+    EngineAdapter,
+    InsertOverwriteStrategy,
+)
 from sqlmesh.core.engine_adapter.mixins import (
     GetCurrentCatalogFromFunctionMixin,
     InsertOverwriteWithMergeMixin,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -281,3 +285,45 @@ def _rename_table(
...
+        source_queries: t.List[SourceQuery],
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        where: t.Optional[exp.Condition] = None,
+        insert_overwrite_strategy_override: t.Optional[InsertOverwriteStrategy] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        if not where or where == exp.true():
+            # this is a full table replacement, call the base strategy to do DELETE+INSERT
+            # which will result in TRUNCATE+INSERT due to how we have overridden self.delete_from()
+            return EngineAdapter._insert_overwrite_by_condition(
```
erindru, line 301: `super(EngineAdapter, self)` threw an exception and `super(EngineAdapterWithIndexSupport, self)` returned the `_insert_overwrite_by_condition` method from `InsertOverwriteWithMergeMixin` instead of from `EngineAdapter`.

So I worked around this by making a direct call to `EngineAdapter`

georgesittas (approved):

Issue #4493, opened 2025-05-21
Loading snapshots should gracefully handle import errors


tobymao:
If a snapshot in state references a 3rd party library that is unavailable on a local system, SQLMesh fails to load. We should handle this gracefully.

A user created a macro library and referenced those functions in a model. They decided they didn't want to use that library anymore so deleted it resulting in a crash.

```
Traceback (most recent call last):
  File "/Users/gesus/.pyenv/versions/3.9.17/lib/python3.9/concurrent/futures/process.py", line 246, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/Users/gesus/.pyenv/versions/3.9.17/lib/python3.9/concurrent/futures/process.py", line 205, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/Users/gesus/.pyenv/versions/3.9.17/lib/python3.9/concurrent/futures/process.py", line 205, in <listcomp>
    return [fn(*args) for args in chunk]
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/model/cache.py", line 175, in load_optimized_query
    entry_name = _optimized_query_cache.put(model)
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/model/cache.py", line 130, in put
    self._put(name, model)
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/model/cache.py", line 134, in _put
    optimized_query = model.render_query()
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/model/definition.py", line 1297, in render_query
    query = self._query_renderer.render(
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/renderer.py", line 513, in render
    expressions = super()._render(
  File "/Users/gesus/Repos/Adonis/core/adonis-core/.venv/lib/python3.9/site-packages/sqlmesh/core/renderer.py", line 155, in _render
    **{**render_kwargs, **prepare_env(self._python_env), **variables},
[...]

PR #4532, opened 2025-05-26
Fix: Catch errors caching optimized model queries so that SQLMesh can still load

erindru:
Addresses #4493 

I was able to produce an `ModuleNotFoundError` on `sqlmesh plan` by:
 - Creating a module in `site-packages` that exposes a macro
 - Importing that macro in `macros/__init__.py`
 - Referencing that macro in a model
 - `sqlmesh plan` - applied successfully
 - Deleting the module from `site-packages`
 - Updating `macros/__init__.py` to include the macro definition inline instead of importing it
 - **clearing the cache**
 - `sqlmesh plan` - `ModuleNotFoundError: no module named <site-packages module>`

When debugging the `ModuleNotFoundError`, I noticed that it was triggered when putting values into the `OptimizedQueryCache` by the `model.render_query()` call that produces the value to be cached.

I also noticed that in `SnapshotCache`, we catch rendering errors, [log the exception](https://github.com/TobikoData/sqlmesh/blob/b1f9960d5de89a28e70f0081c00d8634f4fdfab0/sqlmesh/core/snapshot/cache.py#L100) and move on.

So I applied the same logic to `OptimizedQueryCache` which allowed me to load my project and apply a new plan with the updated import.

### DIFF
--- a/sqlmesh/core/model/cache.py
+++ b/sqlmesh/core/model/cache.py
@@ -141,12 +141,18 @@ class OptimizedQueryCache:
         return name
 
     def _put(self, name: str, model: SqlModel) -> None:
-        optimized_query = model.render_query()
-        new_entry = OptimizedQueryCacheEntry(
-            optimized_rendered_query=optimized_query,
-            renderer_violations=model.violated_rules_for_query,
-        )
-        self._file_cache.put(name, value=new_entry)
+        try:
+            optimized_query = model.render_query()
+
+            new_entry = OptimizedQueryCacheEntry(
+                optimized_rendered_query=optimized_query,
+                renderer_violations=model.violated_rules_for_query,
+            )
+            self._file_cache.put(name, value=new_entry)
+        except:
+            # this can happen if the model query references some python library or function that was available
+            # at the time the model was created but has since been removed locally
+            logger.exception(f"Failed to cache optimized query for model '{model.name}'")
 
     @staticmethod
     def _entry_name(model: SqlModel) -> str:
--- a/sqlmesh/core/model/cache.py
+++ b/sqlmesh/core/model/cache.py
@@ -141,18 +141,13 @@ class OptimizedQueryCache:
         return name
 
     def _put(self, name: str, model: SqlModel) -> None:
-        try:
-            optimized_query = model.render_query()
+        optimized_query = model.render_query()
 
-            new_entry = OptimizedQueryCacheEntry(
-                optimized_rendered_query=optimized_query,
-                renderer_violations=model.violated_rules_for_query,
-            )
-            self._file_cache.put(name, value=new_entry)
-        except:
-            # this can happen if the model query references some python library or function that was available
-            # at the time the model was created but has since been removed locally
-            logger.exception(f"Failed to cache optimized query for model '{model.name}'")
+        new_entry = OptimizedQueryCacheEntry(
+            optimized_rendered_query=optimized_query,
+            renderer_violations=model.violated_rules_for_query,
+        )
+        self._file_cache.put(name, value=new_entry)
 
     @staticmethod
     def _entry_name(model: SqlModel) -> str:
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/cache.py
```diff
@@ -132,12 +132,18 @@ def put(self, model: Model) -> t.Optional[str]:
         return name
 
     def _put(self, name: str, model: SqlModel) -> None:
-        optimized_query = model.render_query()
-        new_entry = OptimizedQueryCacheEntry(
-            optimized_rendered_query=optimized_query,
-            renderer_violations=model.violated_rules_for_query,
-        )
-        self._file_cache.put(name, value=new_entry)
+        try:
```
tobymao, line 135: is there ever a case where we do this but we actually do want to error out? i'm afraid this is a bit too lenient, should we instead put this in the snapshot loader?

georgesittas:
Did you get to the bottom of why it failed to render the query? Why did it have the old `python_env` and tried to import the non-existent package?

erindru:
> Why does it fail when we try to put in the optimized query file cache?

We call `model.render_query()` as part of the `_put` operation; thats where the error gets thrown as it tries to import libraries that no longer exist in the current python environment.

The model itself comes from state, I believe its the result of loading snapshots from state as part of the the plan. I traced through the execution of `sqlmesh plan` on my project that produces this error:

1. PlanBuilder creates a `ContextDiff`
2. `ContextDiff` calls `state_reader.get_snapshots()` so it has something to diff against
3. Most of the snapshots are fetched from the local cache, except for my modified model which is [tagged as "missing" and fetched from state](https://github.com/TobikoData/sqlmesh/blob/01f7af80949f56a149dee6c240f05165faa87dc5/sqlmesh/core/state_sync/cache.py#L67)
4. Fetching from state [calls `get_or_load` on the `SnapshotCache`](https://github.com/TobikoData/sqlmesh/blob/01f7af80949f56a149dee6c240f05165faa87dc5/sqlmesh/core/state_sync/db/snapshot.py#L484)
5. `get_or_load` is a cache miss because I nuked the cache prior, so:
6. As part of loading a snapshot from state into the `SnapshotCache`, it also prepopulates the optimized query cache based on `snapshot.model` - [calling `load_optimized_query` for every snapshot](https://github.com/TobikoData/sqlmesh/blob/4dbff925b6dba8ef6e8b12915e27915eae7a5e4b/sqlmesh/core/snapshot/cache.py#L60) to warm the cache
7. `load_optimized_query` [calls `OptimizedQueryCache.put()`](https://github.com/TobikoData/sqlmesh/blob/4dbff925b6dba8ef6e8b12915e27915eae7a5e4b/sqlmesh/core/model/cache.py#L184)
8. This is what [results in the `model.render_query()` call](https://github.com/TobikoData/sqlmesh/blob/4dbff925b6dba8ef6e8b12915e27915eae7a5e4b/sqlmesh/core/model/cache.py#L144) that throws the exception

Rendering the model in the local environment that doesnt have the python library available throws an exception because when we [instantiate the MacroEvaluator](https://github.com/TobikoData/sqlmesh/blob/a69f1a4d8ceb780f27c8ba898fc1d50617e889fe/sqlmesh/core/renderer.py#L152), we call `prepare_env` which [`eval()`'s all the `Executable` payloads against the current Python interpreter](https://github.com/TobikoData/sqlmesh/blob/133f1d3ed948ea299d3e4c47e504b224dc031a17/sqlmesh/utils/metaprogramming.py#L532).

In the case of my example, it runs `eval("from erin_lib.macros import do_something")` where `erin_lib.macros` used to exist but I deleted it. This causes the `eval` to fail, which throws the error.

georgesittas:
Nice, it all makes sense. Appreciate the detailed write up, Erin!

georgesittas (approved):

sqlmesh/core/model/cache.py
erindru: I switched the try/catch to be inside `load_optimized_query` instead, which is the function called by each process in the process pool to prepopulate the cache

PR #4534, opened 2025-05-26
feat: File Name Equals Model Name Linter

tjirab:
A linter to assert whether the file name (e.g. mymodel in foo/bar/mymodel.sql) matches the model name (e.g. mymodel in myschema.mymodel).


izeigerman:
Hey @tjirab ! Thanks for your contribution! Can you please run `make style` and add a test for this

izeigerman:
Also can you please share a bit more motivation behind this rule and why the following config cannot be used instead?
```
model_naming:
    infer_names: True
```
?

PR #4552, opened 2025-05-27
Fix: improve error handling for invalid environment connections

### LINKED ISSUE #4550: Error: Missing connection type.
If you define an environment variable relating to a gateway that does not exist in your config, it will cause sqlplan to not run anymore, with no further information.

So if you take your simple sushi example code: https://github.com/TobikoData/sqlmesh-examples/tree/main/001_sushi/1_simple

Then set an environment variable for a random gateway:

```
SQLMESH__GATEWAYS__ABCDEF__CONNECTION__PASSWORD = "...."
```

Every sqlmesh command fails:

```
sqlmesh plan
Error: Missing connection type.
```

This is also really difficult to troubleshoot because running sqlmesh --debug does not give you any extra information about what's going on. Who would think having an extra environment variable would break the entire program?

This may be related to https://github.com/TobikoData/sqlmesh/issues/4106

georgesittas:
I was debating whether it's worth making this a bit more complicated to capture the gateway name that is related to the failure, but for now I opted for the simpler approach. Happy to pivot. FYI @izeigerman, since you worked on this section recently.

Fixes #4550


sqlmesh/core/config/connection.py
```diff
@@ -1974,13 +1974,18 @@ def _connection_config_validator(
...
     try:
         return parse_connection_config(v)
     except pydantic.ValidationError as e:
         raise ConfigError(
             validation_error_message(e, f"Invalid '{v['type']}' connection config:")
-            + "\n\nVerify your config.yaml and environment variables."
+            + check_config_and_vars_msg
         )
+    except ConfigError as e:
+        raise ConfigError(f"Invalid connection config: {e}" + check_config_and_vars_msg)
```
izeigerman, line 1988: Why not include the connection type here. Also use `validation_error_message` instead which will ensure the correct formatting of the underlying exeception.

izeigerman (approved):

PR #4566, opened 2025-05-27
feat(vscode): gtd for macros

benfdking:
- [x] Test with builtins


examples/multi/repo_2/config.yaml
```diff
@@ -10,7 +10,7 @@ gateways:
     connection:
       type: duckdb
 
-default_gateway: local
+default_gateway: memory
```
izeigerman, line 13: Why did this change?

PR #4567, opened 2025-05-28
Fix: Dependency handling when converting on-run-start / on-run-end hooks in dbt projects

izeigerman:
Previously, the conversion process didn't handle refs and sources in `on-run-start` / `on-run-end` hooks. This fix addresses the issue by sourcing ref and source dependencies directly from the manifest, in addition to extracting them by statically analyzing the hook's Jinja code.

### DIFF
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -28,7 +28,6 @@ from sqlmesh.utils import UniqueKeyDict
 from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.jinja import (
     JinjaMacroRegistry,
-    extract_macro_references_and_variables,
     make_jinja_registry,
 )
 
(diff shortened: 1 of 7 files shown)


themisvaltinos (approved):
the change looks good, but I noticed an issue in the `_load_environment_statements` which is my mistake that `environment_statements` is overwritten on each iteration of the loop so it doesn’t accumulate `EnvironmentStatements` across multiple projects. I believe to fix this we should replace the assignment with:

```python
environment_statements.extend(
    statements
    for _, statements in sorted(
        hooks_by_package_name.items(),
        key=lambda item: 0 if item[0] == context.project_name else 1,
    )
)
```
and properly indent it. But if not addressed here, I can handle this in a subsequent PR

izeigerman:
@themisvaltinos good catch. I've addressed your comment but keep in mind the following:
1. It appears that the dbt loader only supports a single dbt project today
2. Multiple dbt projects can have overlapping packages in which case we don't want the hooks to be duplicated. So I made `hooks_by_package_name` shared across projects instead.

Issue #4497, opened 2025-05-22
mono repo -> multi repos throwing Error: Duplicate key ...in UniqueKeyDict<models>


lafirm:
We initially used a monorepo setup but recently decided to switch to a multi-repo structure. To support this, we introduced the project variable in config.yaml as described in the [SQLMesh multi-repo guide](https://sqlmesh.readthedocs.io/en/latest/guides/multi_repo/). However, adding the project variable to our existing repository is causing the following error:

> Error: Duplicate key '"iceberg"."schema"."table_1"' found in UniqueKeyDict<models>. Call dict.update(...) if this is intentional.

Is there an easy or recommended way to migrate from a monorepo to a multi-repo setup? Thank you!

tobymao:
can you provide a fully reproducible minimal example?

lafirm:
> can you provide a fully reproducible minimal example?

sure, the below steps will produce the error.

`sqlmesh init duckdb`

`sqlmesh plan`

and apply the plan. Once done, introduce the project variable in the config.yaml like this,

```
project: demo

gateways:
  duckdb:
    connection:
      # For more information on configuring the connection to your execution engine, visit:
      # https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#connections
      # https://sqlmesh.readthedocs.io/en/stable/integrations/engines/duckdb/#connection-options
      type: duckdb
      database: db.db
      # concurrent_tasks: 1
      # register_comments: True
      # pre_ping: False
      # pretty_sql: False
      # catalogs: 
      # extensions: 
      # connector_config: 
      # token: 


default_gateway: duckdb

model_defaults:
  dialect: duckdb
  start: 2025-05-21
```


then again run

`sqlmesh plan`

which will throw the below error,

> Error: Duplicate key '"db"."sqlmesh_example"."incremental_model"' found in UniqueKeyDict<models>. Call dict.update(...) if this is intentional.

Note:
Model name in the error output is different every time and the error occurs even with -p arg in the command, refer below for the sequence of tries and errors in th cli

> .venv(bot)@K6W46PHGL7 sqlmesh-repo % sqlmesh plan       
> Error: Duplicate key '"db"."sqlmesh_example"."incremental_model"' found in UniqueKeyDict<models>. Call dict.update(...) if this is intentional.
> .venv(bot)@K6W46PHGL7 sqlmesh-repo % sqlmesh plan       
[...]

PR #4569, opened 2025-05-28
Fix: handle duplicate key error when we add a project var or change the project name

lafirm:
Should fix https://github.com/TobikoData/sqlmesh/issues/4497

I found a bug while attempting to convert a monorepo to a multi-repo setup by adding a project variable in the config YAML. Upon investigation, I discovered that the same error also occurs when attempting to rename a project.

The root cause appears to be a line of code that tries to [update the _models variable](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/context.py#L633), which is a UniqueKeyDict that already contains the project models. It raises the Duplicatekey error when the models are re-added.

To address this, I’ve submitted a simple and humble PR that assumes we can safely skip adding models if they already exist. I’d appreciate it if you could review the changes and share your thoughts.

_

> Additionally, I have a request: would it be possible to release a bugfix version 0.174.3 with this patch? We’re currently unable to upgrade sqlmesh beyond 0.174.* for the time being, but we do need this fix.

_

Thanks for your time and support!

### DIFF
--- a/tests/core/integration/test_multi_repo.py
+++ b/tests/core/integration/test_multi_repo.py
@@ -421,6 +421,111 @@ def test_multi_hybrid(mocker):
     validate_apply_basics(context, c.PROD, plan.snapshots.values())
 
 
+def test_multi_repo_no_project_to_project(copy_to_temp_path):
+    paths = copy_to_temp_path("examples/multi")
+    repo_1_path = f"{paths[0]}/repo_1"
+    repo_1_config_path = f"{repo_1_path}/config.yaml"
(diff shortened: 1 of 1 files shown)


izeigerman:
@lafirm thanks a lot for your contribution!

I'm a little concerned that this fix may mask a real issue. The idea there is that models sourced from the `prod` environment should **not** overlap with models loaded locally. So we should, under no circumstance, override the local version of the model with the one sourced from `prod`. Otherwise, user changes could get lost.

@georgesittas @tobymao wdyt? Is there a better approach to fixing this?

PR #4575, opened 2025-05-29
Feat: make automatic python dependency inference opt-out

georgesittas:


sqlmesh/core/context.py
```diff
@@ -384,6 +384,7 @@ def __init__(
         self.auto_categorize_changes = self.config.plan.auto_categorize_changes
         self.selected_gateway = gateway or self.config.default_gateway_name
         self.gateway_managed_virtual_layer = self.config.gateway_managed_virtual_layer
+        self.infer_python_package_requirements = self.config.infer_python_package_requirements
```
izeigerman, line 387: Is there a point in having this top-level value?

sqlmesh/core/config/root.py
```diff
@@ -146,6 +147,7 @@ class Config(BaseConfig):
         default=EnvironmentSuffixTarget.default
     )
     gateway_managed_virtual_layer: bool = False
+    infer_python_package_requirements: bool = True
```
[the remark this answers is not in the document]
georgesittas: Will rename.

sqlmesh/core/context.py
izeigerman: Yeah. I missing the point of having this value

georgesittas: I did it to be consistent with `gateway_managed_virtual_layer`, since it's the same pattern. Would you like me to change both of them?

georgesittas: Done

georgesittas:
This is ready for another review, thanks for the suggestions!

sqlmesh/core/context_diff.py
```diff
@@ -475,29 +483,41 @@ def _build_requirements(
...
-                                    and dist not in IGNORED_PACKAGES
-                                    and dist not in excluded_requirements
-                                ):
-                                    requirements[dist] = metadata.version(dist)
-                    except metadata.PackageNotFoundError:
-                        from sqlmesh.core.console import get_console
-
-                        get_console().log_warning(f"Failed to find package for {lib}.")
+        if not snapshot.is_model:
+            continue
```
georgesittas, line 499: I went ahead and de-nested this section a bit as well. Didn't change anything besides negating conditions and short-circuiting with `continue`.

PR #4578, opened 2025-05-29
Refactor: Move engine adapter instantiation after loading on-demand

themisvaltinos:
This update refactors context to defer engine adapter creation until after the context is loaded creating it in the snapshot evaluation instantiation when needed. It also adapts the default_catalog_per_gateway mapping to be sourced from the scheduler config instead.

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -55,6 +55,14 @@ class SchedulerConfig(abc.ABC):
         """
 
     @abc.abstractmethod
+    def get_default_catalog_per_gateway(self, context: GenericContext) -> t.Dict[str, str]:
+        """Returns the default catalog for each gateway.
+
+        Args:
+            context: The SQLMesh Context.
+        """
+
+    @abc.abstractmethod
     def state_sync_fingerprint(self, context: GenericContext) -> str:
         """Returns the fingerprint of the State Sync configuration.
 
@@ -138,6 +146,13 @@ class BuiltInSchedulerConfig(_EngineAdapterStateSyncSchedulerConfig, BaseConfig)
 
     def get_default_catalog(self, context: GenericContext) -> t.Optional[str]:
         return context.engine_adapter.default_catalog
+
+    def get_default_catalog_per_gateway(self, context: GenericContext) -> t.Dict[str, str]:
+        return {
+            name: adapter.default_catalog
+            for name, adapter in context.engine_adapters.items()
+            if adapter.default_catalog
+        }
 
 
 SCHEDULER_CONFIG_TO_TYPE = {
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -62,7 +62,9 @@ from sqlmesh.core.config import (
     Config,
     load_configs,
 )
+from sqlmesh.core.config.connection import ConnectionConfig
 from sqlmesh.core.config.loader import C
+from sqlmesh.core.config.root import RegexKeyDict
 from sqlmesh.core.console import get_console
 from sqlmesh.core.context_diff import ContextDiff
 from sqlmesh.core.dialect import (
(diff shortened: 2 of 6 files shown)


sqlmesh/core/config/scheduler.py
```diff
@@ -139,6 +147,13 @@ def create_plan_evaluator(self, context: GenericContext) -> PlanEvaluator:
     def get_default_catalog(self, context: GenericContext) -> t.Optional[str]:
```
izeigerman, line 147: Can we remove this now?

sqlmesh/core/config/scheduler.py
```diff
@@ -139,6 +147,13 @@ def create_plan_evaluator(self, context: GenericContext) -> PlanEvaluator:
...
         return context.engine_adapter.default_catalog
 
+    def get_default_catalog_per_gateway(self, context: GenericContext) -> t.Dict[str, str]:
+        return {
+            name: adapter.default_catalog
+            for name, adapter in context.engine_adapters.items()
+            if adapter.default_catalog
```
izeigerman, line 154: This call cause interaction with the database we don't want to make it twice here.

sqlmesh/core/context.py
```diff
@@ -435,14 +429,6 @@ def __init__(
         self.users = list({user.username: user for user in self.users}.values())
         self._register_notification_targets()
 
-        if (
-            self.config.environment_catalog_mapping
-            and not self.engine_adapter.catalog_support.is_multi_catalog_supported
-        ):
-            raise SQLMeshError(
```
izeigerman, line 442: Can we keep this but only check if the engine adapter is available?

sqlmesh/core/context.py
```diff
@@ -2508,23 +2496,47 @@ def _model_tables(self) -> t.Dict[str, str]:
...
         """Returns the default catalogs for each engine adapter."""
-        if self._default_catalog_per_gateway is None:
-            self._default_catalog_per_gateway = {
-                name: adapter.default_catalog
-                for name, adapter in self.engine_adapters.items()
-                if adapter.default_catalog
-            }
-        return self._default_catalog_per_gateway
+        if self.gateway_managed_virtual_layer:
+            return self._scheduler.get_default_catalog_per_gateway(self)
```
izeigerman, line 2511: Why do we not cache this any more?

sqlmesh/core/context.py
```diff
@@ -2508,23 +2496,47 @@ def _model_tables(self) -> t.Dict[str, str]:
...
     def default_catalog_per_gateway(self) -> t.Dict[str, str]:
         """Returns the default catalogs for each engine adapter."""
-        if self._default_catalog_per_gateway is None:
-            self._default_catalog_per_gateway = {
-                name: adapter.default_catalog
-                for name, adapter in self.engine_adapters.items()
-                if adapter.default_catalog
-            }
-        return self._default_catalog_per_gateway
+        if self.gateway_managed_virtual_layer:
```
izeigerman, line 2510: Why do we only do this for `gateway_managed_virtual_layer`?

sqlmesh/core/context.py
```diff
@@ -2508,23 +2496,47 @@ def _model_tables(self) -> t.Dict[str, str]:
...
+
+    @cached_property
+    def concurrent_tasks(self) -> int:
+        if self._concurrent_tasks is None:
+            self._concurrent_tasks = self._connection_config.concurrent_tasks
+        return self._concurrent_tasks
+
+    @cached_property
+    def _connection_config(self) -> ConnectionConfig:
+        return self.config.get_connection(self.gateway)
```
izeigerman, line 2522: Shouldn't this be `self.selected_gateway`?

sqlmesh/core/context.py
themisvaltinos: I was thinking of the scenario without multiple virtual layers that the catalog should be shared but good point revise to not do it conditionally

sqlmesh/core/context.py
themisvaltinos: changed it to cache it

sqlmesh/core/config/scheduler.py
themisvaltinos: yes good point changed it to be called once

PR #4585, opened 2025-05-29
Chore: Improve plan error handling when --select-model is used and the remote snapshot can't be rendered

erindru:
This relates to #4493 

When `--select-model` is used with `sqlmesh plan`, the following occurs:
 - We identify which models are selected and use the local versions of those
 - For all other models, we fetch the latest version from state
 - We also identify dependencies by calling `model.depends_on` and fetch them either locally or from state depending on if they were part of `--select-model`
 - the ultimate goal of replacing local versions with remote versions for unselected models is to trick `ContextDiff` into only showing changes for the selected models

The problem is that `model.depends_on` has side effects. It causes a `render_query()` call so it can pull out all the table references, and this can fail if there is a macro reference that contains a python import that *used* to be valid but is no longer valid in the local environment.

At this point, the user has two options:
 - Fix the local environment to put the missing library back so the python `import` succeeds
 - Fix the affected model locally and include it in the `--select-model` expression so the local version is used

This PR updates the current error from:
Screenshot From 2025-05-30 08-44-53

to:
Screenshot From 2025-05-30 08-46-50

(and adds an integration test to verify the behaviour)

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1450,12 +1450,18 @@ class GenericContext(BaseContext, t.Generic[C]):
 
         models_override: t.Optional[UniqueKeyDict[str, Model]] = None
         if select_models:
-            models_override = model_selector.select_models(
-                select_models,
-                environment,
-                fallback_env_name=create_from or c.PROD,
-                ensure_finalized_snapshots=self.config.plan.use_finalized_state,
-            )
+            try:
+                models_override = model_selector.select_models(
+                    select_models,
+                    environment,
+                    fallback_env_name=create_from or c.PROD,
+                    ensure_finalized_snapshots=self.config.plan.use_finalized_state,
+                )
+            except SQLMeshError as e:
+                logger.exception(e)  # ensure the full stack trace is logged
+                raise PlanError(
+                    f"{e}\nCheck the SQLMesh log file for the full stack trace.\nIf the model has been fixed locally, please ensure that the --select-model expression includes it."
+                )
             if not backfill_models:
                 # Only backfill selected models unless explicitly specified.
                 backfill_models = model_selector.expand_model_selections(select_models)
--- a/sqlmesh/core/selector.py
+++ b/sqlmesh/core/selector.py
@@ -16,6 +16,7 @@ from sqlmesh.core.model import update_model_schemas
 from sqlmesh.utils import UniqueKeyDict
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.git import GitClient
+from sqlmesh.utils.errors import SQLMeshError
 
 
 if t.TYPE_CHECKING:
@@ -111,7 +112,16 @@ class Selector:
         def get_model(fqn: str) -> t.Optional[Model]:
             if fqn not in all_selected_models and fqn in env_models:
                 # Unselected modified or added model.
-                return env_models[fqn]
+                model_from_env = env_models[fqn]
+                try:
+                    # this triggers a render_query() which can throw an exception
+                    model_from_env.depends_on
+                    return model_from_env
+                except Exception as e:
+                    raise SQLMeshError(
+                        f"Model '{model_from_env.name}' sourced from state cannot be rendered "
+                        f"in the local environment due to:\n> {str(e)}"
+                    ) from e
             if fqn in all_selected_models and fqn in self._models:
                 # Selected modified or removed model.
                 return self._models[fqn]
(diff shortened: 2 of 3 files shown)


sqlmesh/core/selector.py
```diff
@@ -111,7 +112,16 @@ def select_models(
...
             if fqn not in all_selected_models and fqn in env_models:
                 # Unselected modified or added model.
-                return env_models[fqn]
+                model_from_env = env_models[fqn]
+                try:
+                    # this triggers a render_query() which can throw an exception
+                    model_from_env.depends_on
+                    return model_from_env
+                except Exception as e:
+                    raise SQLMeshError(
```
erindru, line 121: This is done in two stages (here, and in `plan_builder()`) because I wanted to capture the model name but not tie this call site to a `PlanError` since the `Selector` is not just used for plans.

So a `SQLMeshError` is raised here and then wrapped in a `PlanError` by `plan_builder()` which also adds some extra plan-specific text to the error message

sqlmesh/core/context.py
```diff
@@ -1450,12 +1450,18 @@ def plan_builder(
...
+            try:
+                models_override = model_selector.select_models(
+                    select_models,
+                    environment,
+                    fallback_env_name=create_from or c.PROD,
+                    ensure_finalized_snapshots=self.config.plan.use_finalized_state,
+                )
+            except SQLMeshError as e:
+                logger.exception(e)  # ensure the full stack trace is logged
+                raise PlanError(
```
izeigerman, line 1462: Selectors are also used for runs. Not sure `PlanError` is appropriate

sqlmesh/core/context.py
```diff
@@ -1450,12 +1450,18 @@ def plan_builder(
...
+                    f"{e}\nCheck the SQLMesh log file for the full stack trace.\nIf the model has been fixed locally, please ensure that the --select-model expression includes it."
```
izeigerman, line 1463: Are we sure the only exceptions thrown by this method are caused by the user? 

> If the model has been fixed locally

Ideally we should indicate which model caused this.

sqlmesh/core/context.py
izeigerman: nvm, this is `plan_builder`. sorry

sqlmesh/core/selector.py
```diff
@@ -111,7 +112,16 @@ def select_models(
...
+                        f"Model '{model_from_env.name}' sourced from state cannot be rendered "
+                        f"in the local environment due to:\n> {str(e)}"
```
izeigerman, line 123: What's `>` for?

sqlmesh/core/context.py
izeigerman: k, found the name, thanks!

sqlmesh/core/selector.py
```diff
@@ -111,7 +112,16 @@ def select_models(
         def get_model(fqn: str) -> t.Optional[Model]:
             if fqn not in all_selected_models and fqn in env_models:
                 # Unselected modified or added model.
-                return env_models[fqn]
+                model_from_env = env_models[fqn]
+                try:
+                    # this triggers a render_query() which can throw an exception
+                    model_from_env.depends_on
```
izeigerman, line 118: What if the snapshot was cached? I believe in that case `render_query` will just return the cached value. And same goes for `depends_on`.

izeigerman: I guess it will just fail downstream then

izeigerman: Also why even bother with `depends_on`. Why not go to the source and call `render_query()` directly?

erindru: > What if the snapshot was cached? 

There is a try/catch around the cache already which means that problematic snapshots don't get cached 

> Also why even bother with depends_on

The code afterwards uses `depends_on`, so if we ever changed the implementation of `depends_on` to not render the query every time then wouldn't we be be stuck with an extra unnecessary `render_query()` call here?

sqlmesh/core/selector.py
erindru: It just creates a visual cue that this is the error to focus on (see screenshot). I can remove it but then the output will look like:

```
Error: model 'blah' sourced from state cannot be rendered in the local environment due to:
No module named 'erin_lib.macros'
Check the SQLMesh log file for the full stack trace.
If the model has been fixed locally, please ensure that the --select-model expression includes it.
```

i.e there is nothing to draw attention to the actual error that got captured

tests/cli/test_integration_cli.py
```diff
@@ -16,6 +17,10 @@ def __call__(
     ) -> subprocess.CompletedProcess: ...
 
 
+class CreateSitePackageType(t.Protocol):
```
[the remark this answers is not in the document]
erindru: :) I got sick of typing out type signatures like `t.Iterator[t.Callable[[str], t.Tuple[str, Path]]` in multiple places

PR #4595, opened 2025-05-30
Fix: Compare columns by using the schema differ based on snapshots dialect

themisvaltinos:
This update retrieves the schema differ when comparing columns based on the snapshot's dialect, instead of using a generic differ when comparing columns. This approach suggested by @georgesittas avoids the need to initialise an engine adapter by fetching the schema differ directly from the class using the dialect


georgesittas (approved):
Looks great :)

PR #4599, opened 2025-05-30
Chore: update docs on `@this_model`

georgesittas:
I took a stab at improving the docs around `this_model` and also left a note where I tried to explain the difference between the `@{variable}` and `@variable` macro syntax.

Not sure if this is super clear, so please share any feedback you have.

### DIFF
--- a/docs/concepts/macros/macro_variables.md
+++ b/docs/concepts/macros/macro_variables.md
@@ -130,10 +130,18 @@ SQLMesh provides additional predefined variables used to modify model behavior b
     * 'auditing' - The audit is being run.
     * 'testing' - The model query logic is being evaluated in the context of a unit test.
 * @gateway - A string value containing the name of the current [gateway](../../guides/connections.md).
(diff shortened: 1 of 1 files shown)


erindru:
Ah, interesting. Since this wasnt documented, I assumed that the `@{this_model}` syntax was for resolving ambiguities in how the macro was called.

Consider the following post statement:
```
@if(
   @runtime_stage = 'creating',
   create clustered index @resolve_template('PK_@{table_name}') on @this_model (item_id)
);
```

This errors with:
```
Macro 'this_model' does not exist.
 at '/home/erin/sqlmesh-projects/mssql/models/full_model.sql'
```

Why? Because it treats this:
```
@this_model (item_id)
```

As passing parameters to the `@this_model` macro, which doesnt take any parameters.

Changing it to this:
```
@if(
   @runtime_stage = 'creating',
   create clustered index @resolve_template('PK_@{table_name}') on @{this_model} (item_id)
);
```

Indicates that `(item_id)` is standalone and not part of the invocation of the `@this_model` macro.

This allows the statemet to render, but unfortunately leads to double identifier escaping

PR #4605, opened 2025-06-01
refactor: make it possible to get a models macroevaluator

benfdking:


benfdking:
> Missing description. What's the motivation for this?

Sorry, this is mainly so one can have access to the macroevaluator.locals outside a render. This is go to definition and hover on definition wok.

izeigerman:
If you plan on calling it externally, should it not be private any more?

sqlmesh/core/renderer.py
```diff
@@ -188,6 +238,22 @@ def _resolve_table(table: str | exp.Table) -> str:
         }
 
         variables = kwargs.pop("variables", {})
+
+        if this_model:
+            render_kwargs["this_model"] = this_model
+
+        macro_evaluator = self._create_macro_evaluator(
```
georgesittas, line 251: I discussed with Ben about this, but should we maintain a `_macro_evaluator` instance attribute that gets overwritten on each `_render`? So you can then expose a getter property for it and keep the previous logic.

sqlmesh/core/renderer.py
```diff
@@ -68,10 +68,97 @@ def __init__(
...
 
     def update_schema(self, schema: t.Dict[str, t.Any]) -> None:
         self.schema = d.normalize_mapping_schema(schema, dialect=self._dialect)
 
+    @property
+    def macro_evaluator(self) -> t.Optional[MacroEvaluator]:
+        """Returns the cached macro evaluator from the last render operation."""
+        return self._macro_evaluator
+
+    def _create_macro_evaluator(
```
georgesittas, line 81: Do we still want this?

PR #4607, opened 2025-06-01
chore(lsp): move autocomplete to the lsp

benfdking:


sqlmesh/lsp/main.py
```diff
@@ -456,6 +457,47 @@ def workspace_diagnostic(
...
+                    )
+
+                return types.CompletionList(
+                    is_incomplete=False,
+                    items=completion_items,
+                )
+
+            except Exception as e:
+                get_sql_completions(None, URI(params.text_document.uri))
+                return None
```
georgesittas, line 499: L498 appears to be dead code, I assume you meant to return that instead of `None`? Also, should we push this logic to `get_autocomplete`, i.e. have it work as if it received `None` if it fails for a given dialect?

Issue #4367, opened 2025-05-12
Introduce an option to always compare against production when planning


izeigerman:
Currently, the plan's behavior varies depending on whether the target dev environment already exists:
* If the target dev environment doesn't exist, the plan compares local models against the current state of the production environment, effectively creating a copy of production with the planned changes
* If the target dev environment already exists, the local models are compared against the current state of this target dev environment

This difference in behavior is pretty nuanced and often confuses users. Additionally, the plan's behavior can be considered inconsistent, as changes may be categorized differently depending on whether the target environment already exists.

At the same time, it appears that the user expects all changes to be made in relation to production regardless of the current state of the dev environment. Additionally, this ensures a certain level of consistency and predictability in change categorization.

The downside of always comparing to production is that incremental changes may lead to additional backfills, since intermediate changes are lost. For example, when a breaking change to model A is followed by a non-breaking change to the same model, downstream models will be backfilled only once when comparing against the target dev environment - but twice when comparing to production. This is because, in the latter case, the change will be categorized as breaking both times.

Therefore, we can introduce this behavior as an opt-in feature, making it configurable as part of the plan configuration. The tradeoff described above should be thoroughly documented.

Note that if the production environment doesn't exist yet, SQLMesh should still attempt to compare against the target environment.

ig-raiderler:
This is a well-structured codebase. I appreciate the clean separation of concerns.

ig-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #4615, opened 2025-06-02
Feat: Add plan option to always compare against prod

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/4367

This PR introduces the plan config option `always_recreate_environment`; Upon enabling it, local state will always be compared against the prod (by recreating the target env), while the changes will still be committed to the target environment.


VaggelisD:
> needs documentation

Will follow up on that shortly, wanted to align on the implementation and direction before diving into it

sqlmesh/core/context.py
```diff
@@ -2593,6 +2593,25 @@ def _snapshots(
 
         return {name: stored_snapshots.get(s.snapshot_id, s) for name, s in snapshots.items()}
 
+    def _get_target_environment(self, environment: t.Optional[str] = None) -> t.Tuple[str, str]:
```
izeigerman: Doesn't this belong to `context_diff` instead?

sqlmesh/core/context_diff.py
```diff
@@ -130,6 +133,13 @@ def create(
         environment = environment.lower()
         env = state_reader.get_environment(environment)
 
+        initial_environment = initial_environment or environment
```
izeigerman: This seems backwards to me. We should just pass `compare_against_prod` value into `ContextDiff` and let it figure out what to compare against. The "initial environment" concept doesn't make much sense to me. `ContextDiff` should always be associated with one target environment.

sqlmesh/core/context_diff.py
```diff
@@ -88,6 +88,8 @@ class ContextDiff(PydanticModel):
     """Environment statements."""
     diff_rendered: bool = False
     """Whether the diff should compare raw vs rendered models"""
+    initial_environment: str = ""
```
izeigerman: I don't think we need any additional properties for `ContextDiff`.

sqlmesh/core/context_diff.py
VaggelisD: So we have 2 concepts at play:
1. We must maintain which environment to promote the changes to (the initial target environment)
2. We must choose which is the appropriate `ContextDiff` during runtime depending on the flag

So, my main issue here was that afaict `ContextDiff` is also responsible for propagating the to-be-promoted environment [to plan evaluator](https://github.com/TobikoData/sqlmesh/pull/4615/files#diff-55f918f24b35e078fccfe4f341f7dd31486e851661658cfcf557cf3a04d0b74f) (i.e bullet n1), so we must somehow differentiate bullet n1 and bullet n2.

I do agree that not needing the `initial_environment` here and being able to reverse-engineer (1) during plan evaluation would be the simplest solution here.

sqlmesh/core/context.py
VaggelisD: Yeah can move it there, was wondering if it's better to keep it around `Context` for easier config access or move it to `ContextDiff` with the caveat that it'll need the flag propagated to a couple functions

sqlmesh/core/context_diff.py
izeigerman: > We must maintain which environment to promote the changes to (the initial target environment)

This makes sense.

> We must choose which is the appropriate ContextDiff during runtime depending on the flag

This should only matter at `ContextDiff` **construction** time, no? Once constructed, only n1 is relevant.

VaggelisD: > This should only matter at ContextDiff construction time, no? Once constructed, only n1 is relevant.

Yeah I can see the point, that was my first approach as well but my thinking was that this would leave the `ContextDiff` object at an incorrect state; This is because `ContextDiff::environment`'s purpose is _"the environment to diff against"_, and this will have the wrong value past construction time.

For instance, this attribute is also used at `Console::show_environment_difference_summary()`, so the CLI would print `Differences from the <initial_env>` instead of `Differences from prod` and so on.

Imo, the important change of this PR is that before, the env to diff against was tightly coupled to the one which the changes will be promoted to, whereas now these are decoupled and sort of distinction would probably save us from some env related bugs.

What do you think?

sqlmesh/core/context_diff.py
```diff
@@ -479,6 +499,23 @@ def text_diff(self, name: str) -> str:
...
+def _get_target_environment(
+    environment: str, state_reader: StateReader, always_compare_against_prod: bool = False
+) -> str:
+    if always_compare_against_prod:
+        prod = state_reader.get_environment(c.PROD)
+        if prod:
+            logger.warning(
+                f"Comparing against production environment instead of {environment}. Note that this may lead to "
+                "additional backfills as accumulated changes are still pushed to the target environment."
+            )
```
georgesittas: I worry this is going to produce noise and not be helpful for users who intentionally want to compare against prod. Was the intention to maybe check whether `environment != c.PROD` and only then warn?

tests/core/test_integration.py
```diff
@@ -6208,3 +6209,107 @@ def test_render_path_instead_of_model(tmp_path: Path):
...
+    assert plan.context_diff.environment == "prod"
+
+    assert (
+        next(iter(plan.context_diff.snapshots.values())).change_category
+        == SnapshotChangeCategory.BREAKING
+    )
+
+    output = plan_with_output(ctx, "dev")
+
+    assert "Differences from the `prod` environment" in output.stdout
```
georgesittas, line 6334: Is `plan_with_output` producing a fresh `output.stdout` stream every time it runs? Just double-checking whether this assertion is meaningful and not trivially true due to L6280.

sqlmesh/core/context_diff.py
VaggelisD: Can remove it, was mostly a reminder in case folks don't remember they've set/unset that flag but it should be somewhat obvious judging by the `Differences from <env>` message in CLI.

tests/core/test_integration.py
VaggelisD: Yep each `output` contains the output of that plan only, thanks for the heads up 👍

sqlmesh/core/context_diff.py
VaggelisD: Similar discussion here https://github.com/TobikoData/sqlmesh/pull/4615#discussion_r2123114828

sqlmesh/core/context_diff.py
georgesittas: Yeah, I think reiterating this with a warning is not good UX. Either include a small indication using the console, i.e., that the flag is set (so interpreting the above message is straightforward), or just leave it as is.

Experience has shown that non-actionable warnings do more harm than good. Folks justifiably ignore them.

sqlmesh/core/context_diff.py
izeigerman: > For instance, this attribute is also used at Console::show_environment_difference_summary(), so the CLI would print Differences from the <initial_env> instead of Differences from prod and so on.


ok, I buy this argument. I'm aligned that `ContextDiff` should represent what environment it compares against, not the one targeted by the plan.

At the same time, I don't want `ContextDiff` to be associated with more than one environment. I think this creates unnecessary confusion. Can we decouple the environment we pass into `ContextDiff` from the environment we pass into `PlanBuilder`? I.e. get rid of `name=self._context_diff.initial_environment,` in `PlanBuilder` and pass the target environment explicitly in the constructor.

This way, we can get rid of `initial_environment`, only keep `environment` in ContextDiff which represents the environment being compared against.

The reason I suggested moving the environment selection into `ContextDiff.create` is because part of the environment selection logic already resides [there] (https://github.com/TobikoData/sqlmesh/blob/6e08b44b194715c1eb867505da75fa6805ea9a7e/sqlmesh/core/context_diff.py#L150-L159) related to handling the `create_from` argument. I don't mind doing this outside `ContextDiff` but I do want to have the logic that selects the environment to compare against in one place.

VaggelisD: Sounds good, I deduplicated the two types of environments (plan building and console logging). 

`ContextDiff` will always have the diff environment while the target environment is propagated to `Plan::environment_naming_info`

sqlmesh/core/context_diff.py
```diff
@@ -127,24 +129,35 @@ def create(
         Returns:
             The ContextDiff object.
         """
-        environment = environment.lower()
-        env = state_reader.get_environment(environment)
+        initial_environment = environment.lower()
+        initial_env = state_reader.get_environment(initial_environment)
```
georgesittas: Nit: rename `initial_environment` to `initial_environment_name` to make it clearer it's a string, compared to `initial_env`, which is an `Environment`.

sqlmesh/core/context_diff.py
```diff
@@ -127,24 +129,35 @@ def create(
...
+                s.snapshot_id for s in initial_env.promoted_snapshots
+            }
+
+        # Find the proper environment to diff against, this might be different than the initial (i.e user provided) environment
+        # e.g it will default to prod if the plan option `always_compare_against_prod` is set.
+        environment = _get_diff_environment(environment, state_reader, always_compare_against_prod)
+        env = (
+            initial_env
+            if (initial_environment == environment)
+            else state_reader.get_environment(environment)
```
izeigerman: This is a bit weird piece of logic. We've already fetched `prod` in `_get_diff_environment` if `always_compare_against_prod` do we fetch it again here?  If feels like in the worst case scenario we'll be fetching 3 environments. It should be at most 1. 

Shouldn't we just extend the condition above?
```
environment = environment.lower()
env = state_reader.get_environment(environment)
if always_compare_against_prod:
    prod_environment = state_reader.get_environment(c.PROD)
    if prod_environment:
        environment = prod_environment
    else:
        # Print warning that prod doesn't exist
if environment is None or environment.expired:
     # Existing logic follows
```

docs/guides/configuration.md
```diff
@@ -381,6 +381,115 @@ Example showing default values:
     )
     ```
 
+
+### Always comparing against production
```
izeigerman, line 385: Let's also update the Configuration reference doc

sqlmesh/core/context_diff.py
izeigerman: Actually, I think we're overcomplicating this. We can make it easier. Instead of doing `always_compare_against_prod`, we can just have `always_create_from` flag. And so the **only** change we'd need in this module is:
```
if env is None or env.expired or always_create_from:
```
I don't think we need to change anything else here.

izeigerman: This also means, that we don't need to make any changes to the console, because we'd always treat the environment as fresh.

I'd also change the name of the configuration to be something like `always_init_from_prod`.

VaggelisD: I went ahead and renamed it to `always_init_from_prod` but I think there's a slight confusion bound to happen; If the user actually uses `create_from` in the plan, then we will actually use _that_ environment instead of prod. 

So maybe `always_create_from` is more fitting, or instead `always_init_from_prod` should also  _force_ `ContextDiff::create_from = c.PROD` (?)

sqlmesh/core/plan/builder.py
```diff
@@ -156,7 +157,7 @@ def __init__(
         self.override_end = end is not None
         self.environment_naming_info = EnvironmentNamingInfo.from_environment_catalog_mapping(
             environment_catalog_mapping or {},
-            name=self._context_diff.environment,
+            name=environment,
```
izeigerman: so can we go back to using `context_diff` here again?

sqlmesh/core/console.py
```diff
@@ -2904,6 +2905,7 @@ def show_environment_difference_summary(
         Args:
             context_diff: The context diff to use to print the summary.
             no_diff: Hide the actual environment statements differences.
+            environment: The initial target environment
```
izeigerman: we need to remove this leftover

sqlmesh/core/console.py
```diff
@@ -1532,6 +1532,7 @@ def show_environment_difference_summary(
         Args:
             context_diff: The context diff to use to print the summary
             no_diff: Hide the actual environment statement differences.
+            environment: The initial target environment
```
izeigerman: ditto

sqlmesh/core/context.py
```diff
@@ -2607,6 +2609,7 @@ def _context_diff(
         force_no_diff: bool = False,
         ensure_finalized_snapshots: bool = False,
         diff_rendered: bool = False,
+        always_init_from_prod: bool = False,
```
izeigerman: why do we need this argument? can't we access the config in this method directly?

VaggelisD: Because `_context_diff` can be used elsewhere such as `sqlmesh diff <env>` and we _don't_ want to propagate the flag there imo, since it's a plan specific option. For instance:

- Diff behavior if we always propagate the config flag:
```bash
❯ sqlmesh diff prod

New environment `prod` will be created from `prod`

Models:
└── Directly Modified:
    └── sqlmesh_example.test_model                                                                                                                                                              
        ---                                                                                                                                                                                     
                                                                                                                                                                                                
        +++                                                                                                                                                                                     
                                                                                                                                                                                                
        @@ -6,4 +6,4 @@                                                                                                                                                                         
                                                                                                                                                                                                
           kind FULL                                                                                                                                                                            
         )                                                                                                                                                                                      
         SELECT                                                                                                                                                                                 
        -  10 AS col                                                                                                                                                                            
        +  11 AS col    
           
                                                                                                                                                             
❯ sqlmesh diff dev

New environment `dev` will be created from `prod`

Models:
└── Directly Modified:
    └── sqlmesh_example__dev.test_model                                                                                                                                                         
        ---                                                                                                                                                                                     
                                                                                                                                                                                                
        +++                                                                                                                                                                                     
                                                                                                                                                                                                
        @@ -6,4 +6,4 @@                                                                                                                                                                         
                                                                                                                                                                                                
           kind FULL                                                                                                                                                                            
         )                                                                                                                                                                                      
         SELECT                                                                                                                                                                                 
        -  10 AS col                                                                                                                                                                            
        +  11 AS col           
```

<br />

- Diff behavior in this PR:

```bash
❯ sqlmesh diff prod

Models:
└── Directly Modified:
    └── sqlmesh_example.test_model                                                                                                                                                              
        ---                                                                                                                                                                                     
                                                                                                                                                                                                
        +++                                                                                                                                                                                     
                                                                                                                                                                                                
        @@ -6,4 +6,4 @@                                                                                                                                                                         
                                                                                                                                                                                                
           kind FULL                                                                                                                                                                            
         )                                                                                                                                                                                      
         SELECT                                                                                                                                                                                 
        -  10 AS col                                                                                                                                                                            
        +  11 AS col         

                                                                                                                                                                 
❯ sqlmesh diff dev

No changes to plan: project files match the `dev` environment
```

izeigerman (approved):

docs/guides/configuration.md
```diff
@@ -381,6 +381,117 @@ Example showing default values:
...
+    from sqlmesh.core.config import (
+        Config,
+        ModelDefaultsConfig,
+        PlanConfig,
+    )
+
+    config = Config(
+        model_defaults=ModelDefaultsConfig(dialect=<dialect>),
+        plan=PlanConfig(
+            always_compare_against_prod=True,
```
themisvaltinos, line 412: the flag is different in the example: always_recreate_environment

PR #4631, opened 2025-06-02
feat: remove module pandas import to improve import speed

eakmanrq:
Pandas takes a while to import and currently if you import nearly anything from SQLMesh, including things like connection config, you are forced to import pandas. This is due to the fact that we import pandas at the module level (top of the file). 

This PR changes all pandas imports to be at the local import scope (within the function that is using it). It also uses tidy to ban module level imports of pandas going forward. 

Exceptions:
* Example projects
* Tests

More can be done to improve this. For example there are sometimes where we could import even more locally in the function (within a specific if block).  Also a lot of the time we import pandas just to see if something is a pandas dataframe. We could replace this with a duck-typing approach to avoid import pandas when the object isn't actually a pandas dataframe. This is a good first step though.


sqlmesh/core/table_diff.py
```diff
@@ -75,6 +76,21 @@ class RowDiff(PydanticModel, frozen=True):
...
+            cls._resolve_types()
+        return super().__new__(cls)
+
+    @classmethod
+    def _resolve_types(cls) -> None:
+        """Pandas is imported by type checking so we need to resolve the types with the real import before instantiating."""
+        import pandas as pd  # noqa
+
+        cls.model_rebuild()
+        cls._types_resolved = True
```
eakmanrq: This required some custom logic. The issue is that pandas is imported for type checking purposes but then when Pydantic instantiates the object it doesn't actually have pandas available and errors. So what I did was overrode new and had it rebuild the model on the first instantiation. Confirmed in tests that the issue would repro before and doesn't anymore after this change.

sqlmesh/utils/pandas.py
```diff
@@ -1,42 +1,50 @@
...
+    }
+    try:
+        import pyarrow  # type: ignore  # noqa
+
+        # Only add this if pyarrow is installed
+        mappings[pd.StringDtype("pyarrow")] = exp.DataType.build("text")
+    except ImportError:
+        pass
+
+    return mappings
```
eakmanrq, line 47: Move this from a global variable to defined inside this function and then put an lru cache on it. This was only accessed within this file.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -257,7 +262,7 @@ def _df_to_source_queries(
             SourceQuery(
                 query_factory=partial(
                     self._values_to_sql,
-                    values=values,
+                    values=values,  # type: ignore
```
georgesittas, line 265: Why did we need this `# type: ignore`?

sqlmesh/core/table_diff.py
```diff
@@ -75,6 +76,21 @@ class RowDiff(PydanticModel, frozen=True):
...
+
+    def __new__(cls, *args: t.Any, **kwargs: t.Any) -> RowDiff:
+        if not cls._types_resolved:
+            cls._resolve_types()
+        return super().__new__(cls)
+
+    @classmethod
+    def _resolve_types(cls) -> None:
+        """Pandas is imported by type checking so we need to resolve the types with the real import before instantiating."""
+        import pandas as pd  # noqa
```
georgesittas, line 89: Why do need to `# noqa` here?

georgesittas (approved):
LGTM, a couple of questions.

sqlmesh/core/table_diff.py
eakmanrq: Its not actually referenced so this is to prevent check failures saying there is an unused import. It is implicitly used when pydantic rebuilds.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Yeah this looked like a strange type check error. I didn't see how it was related to the change so I just ignored it. The logic here didn't change with the PR.

sqlmesh/core/table_diff.py
```diff
@@ -75,6 +76,21 @@ class RowDiff(PydanticModel, frozen=True):
...
+    _types_resolved: t.ClassVar[bool] = False
+
+    def __new__(cls, *args: t.Any, **kwargs: t.Any) -> RowDiff:
+        if not cls._types_resolved:
+            cls._resolve_types()
+        return super().__new__(cls)
+
+    @classmethod
+    def _resolve_types(cls) -> None:
+        """Pandas is imported by type checking so we need to resolve the types with the real import before instantiating."""
```
izeigerman:
```suggestion
        # Pandas is imported by type checking so we need to resolve the types with the real import before instantiating.
```
This is not really a docstring to this method

tests/fixtures/migrations/snapshots.json
```diff
@@ -1 +1 @@
...
...], \"audits\": [], \"expressions\": [], \"jinja_macros\": {\"packages\": {}, \"root_macros\": {}, \"global_objs\": {}}, \"seed\": {\"content\": \"id,name\\n0,Toby\\n1,Tyson\\n2,Ryan\\n3,George\\n4,Chris\\n5,Max\\n6,Vincent\\n7,Iaroslav\\n8,Emma\\n9,Maia\\n10,Jim\\n\"}, \"source_type\": \"seed\"}, \"parents\": [], \"audits\": [], \"intervals\": [[1672531200000, 1673136000000]], \"dev_intervals\": [], \"created_ts\": 1680814464932, \"updated_ts\": 1680814464932, \"ttl\": \"in 1 week\", \"previous_versions\": [{\"fingerprint\": {\"data_hash\": \"1876476880\", \"metadata_hash\": \"570478986\", \"parent_data_hash\": \"0\", \"parent_metadata_hash\": \"0\"}, \"version\": \"2505706914\"}], \"indirect_versions\": {\"sushi.waiter_as_customer_by_day\": [{\"fingerprint\": {\"data_hash\": \"486172035\", \"metadata_hash\": \"1992853678\", \"parent_data_hash\": \"2154574190\", \"parent_metadata_hash\": \"1349779748\"}, \"version\": \"1267397572\"}, {\"fingerprint\": {\"data_hash\": \"486172035\", \"metadata_hash\": \"1992853678\", \"parent_data_hash\": \"2824767713\", \"parent_metadata_hash\": \"1349779748\"}, \"version\": \"3668757715\"}]}, \"version\": \"1204702829\", \"change_category\": 1}"}}
```
izeigerman: Why did this change?

eakmanrq: thanks for catching. It was a bad find/replace but it just added a comment so it didn't affect the result. Fixing it.

Issue #4568, opened 2025-05-28
Improve INCREMENTAL_BY_UNIQUE_KEY merge


codykonior:
For MSSQL/T-SQL the MERGE statement generated for a INCREMENTAL_BY_UNIQUE_KEY model could use improvement.

It typically does:

```
MERGE INTO ... [__MERGE_TARGET__] 
USING (...)  AS [__MERGE_SOURCE__] 
ON Target Unique Key Columns = Source Unique Key Columns
WHEN MATCHED 
THEN UPDATE SET Each Target Column = Each Source Column
```

This is bad for MSSQL, updating columns to their existing values isn't a no-operation. It overwrites every single column and is all logged. So we'd like to prevent any of that where not necessary.

For this engine it's much better to do:

```
MERGE INTO ... [__MERGE_TARGET__] 
USING (...)  AS [__MERGE_SOURCE__] 
ON Target Unique Key Columns = Source Unique Key Columns
WHEN MATCHED AND EXISTS (SELECT Each Target Column EXCEPT SELECT Each Source Column) (except for the unique columns)
THEN UPDATE SET Each Target Column = Each Source Column (except for the unique columns)
```

The first part of this change with the EXISTS prevents any update happening if all of the columns already match, which is an extremely common case no matter what you're doing. Also to note the syntax for this (instead of comparing each column individually like WHEN MATCHED AND A1 = A2 AND B1 = B2 etc), that invites NULL handling issues, and this syntax neatly sidesteps all of that because for MSSQL the SELECT/EXCEPT/SELECT honours NULL comparisons. It is well accepted as the right way to do things; other engines it may not work though, I don't know.

The second part of that change is being a bit more careful with the UPDATE SET columns to not update the unique columns we just compared as part of the match. The reason for this is, for normal clustered table, the original syntax in MSSQL causes a deferred update where it internally deletes and inserts the row. When these aren't specifically set, it can update the row in place (assuming all the data types fit etc). I'm not 100% sure this occurs on the heap tables sqlmesh is using, but it should be avoided anyway.

Together these will result in much reduced CPU and IO.

For the user-presented implementation I'd like to suggest allowing a keyword on a model like:

```
MODEL (
    name [xxx].[yyy],
    kind INCREMENTAL_BY_UNIQUE_KEY (
        unique_key [My Id],
        match_exists_except [Some Column I Want To Exclude]
    ),
    cron '*/5 * * * *'
);
```

[...]

PR #4632, opened 2025-06-03
MSSQL Merge improvements #4568

codykonior:
I'm hoping to add a test for this, but this is the code I've been running for the past week.


codykonior:
For some reason it included all the other commits from others, I'll try again with a fresh branch

PR #4634, opened 2025-06-03
faet(vscode): support multiple workspaces

benfdking:
[ci skip]


sqlmesh/lsp/main.py
```diff
@@ -506,6 +529,23 @@ def _ensure_context_for_document(
                         )
             path = path.parent
 
+        # If still no context found, try the workspace folders
+        if not loaded:
+            for workspace_folder in self.workspace_folders:
```
themisvaltinos, line 534: nit maybe refactor this since its identically used  in both places

PR #4600, opened 2025-05-31
Add data path and encrypted to duckdb catalog

mattiasthalen:
This pull request introduces support for the `ducklake` catalog type in DuckDB, including the ability to specify a `DATA_PATH` and activate encryption of the parquet files. The changes span documentation updates, configuration enhancements, SQL generation modifications, and new test cases to ensure functionality.

### Documentation Updates:
* Added a new section in `docs/integrations/engines/duckdb.md` demonstrating how to configure the `ducklake` catalog type in both YAML and Python formats, including the use of `DATA_PATH` & `ENCRYPTED`.

### Configuration Enhancements:
* Updated the `DuckDBAttachOptions` class in `sqlmesh/core/config/connection.py` to include a new `data_path` attribute for specifying the `DATA_PATH` option in `ducklake` catalogs. In addition to that, this also adds the `ENCRYPTED` option to activate the parquet encryption.

### SQL Generation Modifications:
* Enhanced the `to_sql` method in `DuckDBAttachOptions` to include the `DATA_PATH` option when the catalog type is `ducklake` and a `data_path` is provided. And also the `ENCRYPTED` flag, if set to True.

### Testing Additions:
* Added a new test case in `tests/core/test_connection_config.py` to validate the configuration and SQL generation for `ducklake` catalogs with a `DATA_PATH` and `ENCRYPTED`.

Closes #4590

PR #4635, opened 2025-06-03
Add ducklake `DATA_INLINING_ROW_LIMIT` attach option

mrjsj:
This PR adds support for configuring the `DATA_INLINING_ROW_LIMIT` (https://ducklake.select/docs/stable/duckdb/advanced_features/data_inlining
) parameter in DuckLake catalogs. This allows users to control the number of rows that can be inlined in a DuckLake catalog.

Changes:
- Added `data_inlining_row_limit` configuration option to `DuckDBAttachOptions`
- Updated documentation with example configuration
- Added tests

#4600 already did ground work for this change.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -224,6 +224,7 @@ class DuckDBAttachOptions(BaseConfig):
     # DuckLake specific options
     data_path: t.Optional[str] = None
     encrypted: bool = False
+    data_inlining_row_limit: t.Optional[int] = None
 
     def to_sql(self, alias: str) -> str:
         options = []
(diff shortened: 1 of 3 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -240,6 +241,8 @@ def to_sql(self, alias: str) -> str:
                 options.append(f"DATA_PATH '{self.data_path}'")
             if self.encrypted:
                 options.append("ENCRYPTED")
+            if self.data_inlining_row_limit:
+                options.append(f"DATA_INLINING_ROW_LIMIT {self.data_inlining_row_limit}")
```
georgesittas: Should this check be `if self.data_inlining_row_limit is not None: ...` instead? What if the limit is 0?

georgesittas (approved):
Thanks for the contribution! Could you please sign the CLA?

sqlmesh/core/config/connection.py
mrjsj: great catch

mrjsj: i also updated this for `data_path` as `attach 'ducklake:db.ducklake' (DATA_PATH '');` is also valid.

georgesittas (approved):

PR #4636, opened 2025-06-03
feat: render model without active editor

benfdking:


sqlmesh/lsp/main.py
```diff
@@ -117,6 +120,17 @@ def render_model(ls: LanguageServer, params: RenderModelRequest) -> RenderModelR
...
 
+        @self.server.feature(ALL_MODELS_FOR_RENDER_FEATURE)
+        def all_models_for_render(
+            ls: LanguageServer, params: AllModelsForRenderRequest
+        ) -> AllModelsForRenderResponse:
+            if self.lsp_context is None:
+                current_path = Path.cwd()
+                self._ensure_context_in_folder(current_path)
+            if self.lsp_context is None:
+                raise RuntimeError("No context found")
```
VaggelisD, line 136: The branch here is repeated, is that meant to be? If yes, I'd assume `self._ensure_context_in_folder` would encapsulate the assertion inside it

georgesittas (approved):

sqlmesh/lsp/context.py
```diff
@@ -148,3 +148,29 @@ def lint_model(self, uri: URI) -> t.List[AnnotatedRuleViolation]:
...
+        ] + [
+            ModelForRendering(
+                name=audit.name,
+                fqn=audit.fqn,
+                description=audit.description,
+                uri=URI.from_path(audit._path).value,
+            )
+            for audit in self.context.standalone_audits.values()
+            if audit._path is not None
+        ]
```
VaggelisD, line 176: You can simplify that as:

```Python
return [ModelForRendering(...) for node in [*self.context.models.values(), self.context.standalone_audits.values()] if isinstance(node, (SqlModel, StandaloneAudit) and node._path]  ```

```

sqlmesh/lsp/custom.py
```diff
@@ -46,3 +46,29 @@ class RenderModelResponse(PydanticModel):
     """
 
     models: t.List[RenderModelEntry]
+
+
+ALL_MODELS_FOR_RENDER_FEATURE = "sqlmesh/all_models_for_render"
+
+
+class ModelForRendering(PydanticModel):
```
VaggelisD, line 54: I think we should rename those classes as nodes instead of models, since we can also render audits etc.

sqlmesh/lsp/context.py
benfdking: slighlty more complicated with the ifs

sqlmesh/lsp/custom.py
benfdking: I don't have a good name though because it's not 'all nodes' so don't want to confuse it. We should rename it at some point though.

sqlmesh/lsp/main.py
benfdking: Yep, you could refactor to return it and then not error if it existed, just exists in a few places at the moment so will make separate PR where we refactor this all

PR #4637, opened 2025-06-03
feat(vscode): rerender on save model

benfdking:
Clipboard-20250603-123556-813

### DIFF
--- a/vscode/extension/src/commands/renderModel.ts
+++ b/vscode/extension/src/commands/renderModel.ts
@@ -4,6 +4,51 @@ import { isErr } from '@bus/result'
 import { RenderModelEntry } from '../lsp/custom'
 import { RenderedModelProvider } from '../providers/renderedModelProvider'
 
+export async function reRenderModelForSourceFile(
+  sourceUri: string,
+  lspClient: LSPClient,
+  renderedModelProvider: RenderedModelProvider,
+): Promise<void> {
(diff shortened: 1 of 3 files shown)


vscode/extension/src/commands/renderModel.ts
```diff
@@ -4,6 +4,39 @@ import { isErr } from '@bus/result'
...
+  }
+
+  // Check if we got any models
+  if (!result.value.models || result.value.models.length === 0) {
+    return
+  }
+
+  // For auto-rerender, we'll use the first model found
+  // In the future, we could store which specific model was originally rendered
+  const selectedModel = result.value.models[0]
```
georgesittas: What could the other models be here (I mean `models[1:]`)?

georgesittas (approved):

vscode/extension/src/commands/renderModel.ts
benfdking: Oh this is bad ... I need to fix this. Think blueprints

georgesittas: Does this mean that only the first blueprint would get re-rendered, for example?

benfdking: Yep this is a mistake ... I should add a test for it.

benfdking: Didn't add a test but fixed this.

PR #4638, opened 2025-06-03
feat(lsp): add keywords in query to autocompete

benfdking:


sqlmesh/lsp/completions.py
```diff
@@ -97,3 +108,62 @@ def get_keywords_from_tokenizer(dialect: t.Optional[str] = None) -> t.Set[str]:
...
+    """
+    Extract identifiers from SQL content using the tokenizer.
+    Only extracts identifiers (variable names, table names, column names, etc.)
+    that are not SQL keywords.
+    """
+    if not content:
+        return set()
+
+    tokenizer_class = Tokenizer
+    if dialect is not None:
```
georgesittas: You don't have to check if `dialect is not None `here, `get_or_raise` returns the base tokenizer on `None`.

sqlmesh/lsp/completions.py
```diff
@@ -97,3 +108,62 @@ def get_keywords_from_tokenizer(dialect: t.Optional[str] = None) -> t.Set[str]:
...
+    """
+    if file_uri is not None and context is not None and file_uri.to_path() in context.map:
+        file_info = context.map[file_uri.to_path()]
+
+        # Handle ModelInfo objects
+        if isinstance(file_info, ModelTarget) and file_info.names:
+            model_name = file_info.names[0]
+            model_from_context = context.context.get_model(model_name)
+            if model_from_context is not None and model_from_context.dialect:
+                return model_from_context.dialect
```
georgesittas: No need to check if the `dialect` is truthy, it's always expected to be. Ditto for the audits below.

sqlmesh/lsp/completions.py
```diff
@@ -97,3 +108,62 @@ def get_keywords_from_tokenizer(dialect: t.Optional[str] = None) -> t.Set[str]:
...
+        except Exception:
+            pass
+
+    keywords = set()
+    try:
+        tokenizer = tokenizer_class()
+        tokens = tokenizer.tokenize(content)
+        for token in tokens:
+            # Extract identifiers (variables, table names, column names, etc.)
+            if token.token_type == TokenType.VAR and token.text:
```
georgesittas: What about `IDENTIFIER` (e.g. `"foo"`) and `STRING` (e.g. `'foo'`) tokens? Should we factor the set of interesting tokens out into a global variable and do a look up `token.token_type` in it instead?

sqlmesh/lsp/completions.py
```diff
@@ -97,3 +108,62 @@ def get_keywords_from_tokenizer(dialect: t.Optional[str] = None) -> t.Set[str]:
...
+                # Skip if it's a SQL keyword
+                if token.text.upper() not in tokenizer_class.KEYWORDS:
+                    keywords.add(token.text)
```
georgesittas: Why do we need this?

benfdking: Because I don't want to add in the keywords twice.

sqlmesh/lsp/completions.py
benfdking: I made it more generic, I think we just want to check waht's not matching if both upper.

georgesittas (approved):

PR #4639, opened 2025-06-03
Feat: add support for Trino's authorization session property

georgesittas:
This adds first-class support for Trino's `SET SESSION AUTHORIZATION` ([docs](https://trino.io/docs/current/sql/set-session-authorization.html)), using the existing `session_properties` mapping. One can now do the following:

```sql
MODEL (
    name <model_name>,
    ...,
    session_properties (
        authorization = 'foo'
    )
);
```

which will result in the following statement:

```sql
SET SESSION AUTHORIZATION 'foo';
```

to be run **before** anything relevant to this model is executed, and then the following statement:

```sql
RESET SESSION AUTHORIZATION;
```

will run **after** all statements relevant to this model have been executed.

### DIFF
--- a/sqlmesh/core/engine_adapter/trino.py
+++ b/sqlmesh/core/engine_adapter/trino.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import pandas as pd
 import re
 import typing as t
 from functools import lru_cache
@@ -27,7 +26,6 @@ from sqlmesh.core.engine_adapter.shared import (
     SourceQuery,
     set_catalog,
 )
-from sqlmesh.utils.errors import SQLMeshError
 from sqlmesh.core.schema_diff import SchemaDiffer
 from sqlmesh.utils.date import TimeLike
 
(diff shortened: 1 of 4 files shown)


sqlmesh/core/engine_adapter/trino.py
```diff
@@ -88,6 +91,26 @@ def get_catalog_type(self, catalog: t.Optional[str]) -> str:
...
+    def session(self, properties: SessionProperties) -> t.Iterator[None]:
+        authorization = properties.get("authorization")
+        if not authorization:
+            yield
+            return
+
+        if isinstance(authorization, str):
+            authorization = exp.Literal.string(authorization)
+        if not (isinstance(authorization, exp.Expression) and authorization.is_string):
+            raise SQLMeshError(f"Invalid authorization: '{authorization}'")
```
izeigerman: Can this be validated at the model creation time?

georgesittas: Done: https://github.com/TobikoData/sqlmesh/pull/4639/commits/8ed4d0ff0ce84132162006e45d2c96a99d0e8273

PR #4641, opened 2025-06-03
feat: add inlay type hints for sqlmesh sql models

kelsin:


sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,113 @@
...
+                col = meta.get("col")
+
+                name = select_exp.alias_or_name
+                if name not in columns_to_types:
+                    continue
+
+                if line < start_line or line > end_line:
+                    continue
+
+                type_label = str(columns_to_types.get(name))
```
benfdking: @georgesittas mentioned the `UNKNOWN` that I think we want to avoid showing. Don't know how to show it in a test.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,113 @@
...
+) -> t.List[types.InlayHint]:
+    hints: t.List[types.InlayHint] = []
+    try:
+        query = normalize_identifiers(query.copy(), dialect=dialect)
+        root = build_scope(query)
+
+        if not root:
+            return []
+
+        for select in find_all_in_scope(root.expression, exp.Select):
```
benfdking: I don't think we want to go through every scope, only the outmost one because I think that's the only one we are confident of at the moment of the types?

I may be misunderstanding this though, so would check with @georgesittas

sqlmesh/lsp/main.py
```diff
@@ -324,6 +323,27 @@ def hover(ls: LanguageServer, params: types.HoverParams) -> t.Optional[types.Hov
...
 
+        @self.server.feature(types.TEXT_DOCUMENT_INLAY_HINT)
+        def inlay_hint(
+            ls: LanguageServer, params: types.InlayHintParams
+        ) -> t.List[types.InlayHint]:
+            """Implement type hints for sql columns as inlay hints"""
+            try:
+                uri = URI(params.text_document.uri)
+                self._ensure_context_for_document(uri)
+                document = ls.workspace.get_text_document(params.text_document.uri)
```
benfdking: I can't seem to understand why you are reading the whole the document here. Just a thing that matters as this is a full file read. Feels like an unused variable, which I wish our linter would catch.

benfdking (commented):
A few little things but nice work! 🚀

tests/lsp/test_hints.py
```diff
@@ -0,0 +1,62 @@
...
+from sqlmesh.core.context import Context
+from sqlmesh.lsp.context import LSPContext, ModelTarget
+from sqlmesh.lsp.hints import get_hints
+from sqlmesh.lsp.uri import URI
+
+
+@pytest.mark.fast
+def test_hints() -> None:
+    context = Context(paths=["examples/sushi"])
+    lsp_context = LSPContext(context)
```
benfdking, line 16: I have been adding a small integration test for each feature as well. It's minor but is a nice confirmation that it's correct. An easy way to test for doing the showing double you talked about.

sqlmesh/lsp/hints.py
kelsin: Yeah this should just be going over every Select section of the root scope which is only outermost.

sqlmesh/lsp/main.py
kelsin: It was referenced in the error. Fixing.

tests/lsp/test_hints.py
kelsin: Where are these located?

sqlmesh/lsp/hints.py
kelsin: Added.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+
+                if line < start_line or line > end_line:
+                    continue
+
+                data_type = columns_to_types.get(name)
+
+                if not data_type or data_type.is_type(exp.DataType.Type.UNKNOWN):
+                    continue
+
+                type_label = str(data_type)
```
VaggelisD: Generating expressions with `str(...)` is equivalent to `expr.sql()` i.e using the base generator, we should probably be doing `expr.sql(dialect)` to match the dialect semantics. 

For instance, `Type.DATETIME` will be generated as `"DATETIME"` in the base generator but in some dialects it might be mapped to `"TIMESTAMP"`.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+                    meta = select_exp.parts[-1]._meta
+                else:
+                    continue
+
+                line = meta.get("line") - 1  # Lines from sqlglot are 1 based
+                col = meta.get("col")
+
+                name = select_exp.alias_or_name
+                if name not in columns_to_types:
+                    continue
```
VaggelisD: This looks like it can be removed,  it's mostly repeating the key check with the branch that's following it:

```Python3
                data_type = columns_to_types.get(name)

                if not data_type or data_type.is_type(exp.DataType.Type.UNKNOWN):
                    continue
```

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+    if path.suffix != ".sql":
+        return []
+
+    if path not in lsp_context.map:
+        return []
+
+    file_info = lsp_context.map[path]
+
+    # Process based on whether it's a model or standalone audit
+    if isinstance(file_info, ModelTarget):
```
VaggelisD: [Nit] we can maybe flip the condition here to save on an indentation level? e.g

```
if not isinstance(file_info, ModelTarget):
  return []

model = ...
```

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns:
+        A list of hints to apply to the document
+    """
+    path = document_uri.to_path()
+    if path.suffix != ".sql":
+        return []
+
+    if path not in lsp_context.map:
+        return []
```
VaggelisD, line 38: [Future] I think we do similar processing elsewhere i.e collecting the model targets, differentiating them from standalone audits etc. 

It would help a lot if we slowly started abstracting some of that repeating logic, the optionality of all these objects makes for a somewhat verbose if-else chain

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+    file_info = lsp_context.map[path]
+
+    # Process based on whether it's a model or standalone audit
+    if isinstance(file_info, ModelTarget):
+        # It's a model
+        model = lsp_context.context.get_model(
+            model_or_snapshot=file_info.names[0], raise_if_missing=False
+        )
+        if model is None or not isinstance(model, SqlModel):
+            return []
```
VaggelisD: [Nit] These checks can be merged into one i.e `if not isinstance(model, SqlModel)` will also take care of `None`

VaggelisD (commented):
Left some comments to consider before the big guns (George), this is an awesome feature 💯

vscode/extension/tests/hints.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    await window.waitForSelector('text=grain')
+    await window.waitForSelector('text=Loaded SQLMesh Context')
+
+    // Wait a moment for hints to appear
+    await window.waitForTimeout(500)
+
+    // Check if the hints are visible
+    expect(
+      await window.locator('text=customer_id::INT').count(),
+    ).toBe(1)
```
VaggelisD: Can we also add a couple more parameterized & complex types? The former would be `VARCHAR(100)` while the latter would be something like `STRUCT<INT, STRUCT<TEXT, ARRAY<INT>>>` (we've seen such monstrosities before in dialects like BQ)

VaggelisD: Too complex types might have to be truncated (?)

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+                type_label = str(data_type)
+                hints.append(
+                    types.InlayHint(
+                        label=f"::{type_label}",
+                        kind=types.InlayHintKind.Type,
+                        padding_left=False,
+                        padding_right=True,
+                        position=types.Position(line=line, character=col),
+                    )
+                )
```
benfdking: image

I think we probably want to fail silently.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return hints
```
benfdking: Do we want to add it twice? I think if you specify it in the non-alias, we can skip it for the alias?

image

benfdking (commented):
Again lovely work!

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+def _get_type_hints_for_model_from_query(
+    query: Expression,
+    dialect: str,
+    columns_to_types: t.Dict[str, exp.DataType],
+    start_line: int,
+    end_line: int,
+) -> t.List[types.InlayHint]:
+    hints: t.List[types.InlayHint] = []
+    try:
+        query = normalize_identifiers(query.copy(), dialect=dialect)
```
georgesittas, line 124: I wonder if we should start extracting these normalized queries out into the `LspContext`, possibly caching them while at it. It's easy to forget to do this, and also not straightforward to understand why this is needed unless you're familiar with normalization specifics.

sqlmesh/lsp/hints.py
georgesittas: Yeah this would only pick up the `Select` instances in the outer scope. I was initially skeptical of the loop, but it may actually be necessary for things like `UNION` model queries.

Can we test if the inlay is present in all of the `UNION` branches?

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+        root = build_scope(query)
+
+        if not root:
+            return []
+
+        for select in find_all_in_scope(root.expression, exp.Select):
+            for select_exp in select.expressions:
+                if not select_exp:
+                    continue
```
georgesittas: I don't think this can ever be falsey.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+
+                if isinstance(select_exp, exp.Alias):
+                    meta = select_exp.args["alias"]._meta
+                elif isinstance(select_exp, exp.Column):
+                    meta = select_exp.parts[-1]._meta
+                else:
+                    continue
+
+                line = meta.get("line") - 1  # Lines from sqlglot are 1 based
+                col = meta.get("col")
```
georgesittas: Let's ensure `meta` contains the line & column info & use `[...]` lookup syntax, before proceeding with this logic. Otherwise you may end up getting a `TypeError` (`None - 1`).

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,118 @@
...
+
+                type_label = str(data_type)
+                hints.append(
+                    types.InlayHint(
+                        label=f"::{type_label}",
+                        kind=types.InlayHintKind.Type,
+                        padding_left=False,
+                        padding_right=True,
+                        position=types.Position(line=line, character=col),
+                    )
```
georgesittas: I wonder if we want to include the `::` in the inlay. This could make it awkward to deal with things that aren't identifiers, e.g., `CAST` projections.

```
MODEL (name test);

SELECT CAST(x AS INT);
```

^ This renders into the following, so SQLMesh infers the column's name to be the `CAST` argument:

```
SELECT
  CAST("x" AS INT) AS "x"
```

I wonder if always using an inlay like `<type>` at the end of projection's line is a better format for this. Thoughts?

sqlmesh/lsp/hints.py
georgesittas: Discussed offline – we'll address this at a later stage when we start refactoring.

sqlmesh/lsp/hints.py
kelsin: Excellent, fixing!

sqlmesh/lsp/hints.py
kelsin: Good catch, I added this when I was debugging my stupid `,` vs `.` problem lol

kelsin:
Most easy comments addressed. Some discussions left of course!

sqlmesh/lsp/hints.py
kelsin: This only finds the LAST segment of the union. Raising discussion in slack: https://tobikodata.slack.com/archives/C040V7ZP9NH/p1749056892168249

sqlmesh/lsp/hints.py
kelsin: We'd have to dig inside each select and determine if the casting they are doing is top level and if it is then don't show. It's a lot of logic to handle everything that can be there. We can do that if you want, just seemed like a lot of work for the first pass. We could also JUST handle the simple case like this when the alias is only for a cast, and if so don't put a type in.

vscode/extension/tests/hints.spec.ts
kelsin: Added these to a new unit test. If we want to truncate that's def a discussion.

sqlmesh/lsp/hints.py
kelsin: We'd have to decide how to deal with multiple on the same line (if they aren't on different lines)

sqlmesh/lsp/hints.py
kelsin: Added a new unit test for this.

sqlmesh/lsp/hints.py
benfdking: I am not sure we want them at the end of the line. I just think for explicit casts where the type is obvious we don't show the inline.

vscode/extension/tests/hints.spec.ts
benfdking: I think let's leave for the time being. If we need to truncate later that's a future iteration.

sqlmesh/lsp/hints.py
benfdking: I would start with the very very simple, if 
- direct cast -> column_a::INT don't readd
- for alias -> column_a::INT as column_b don't read 

^ To me, this will cover 99% of the cases where I think this is too verbose.  I think max do one layer.

sqlmesh/lsp/hints.py
benfdking: I think this abstraction can wait.

sqlmesh/lsp/hints.py
kelsin: Cool, the first case is already handled. We only deal with `exp.Column` or `exp.Alias` items. For the Alias case I'll add a test and make sure this happens!

kelsin: Both cases now handled (and in tests)

kelsin:
@benfdking @georgesittas ok I want a review on how I'm doing the selection of things. I don't fully know sqlglot. You can see in the tests what type of queries. Would love more tests that don't work.

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,157 @@
...
+        if isinstance(select_exp, exp.Alias):
+            if isinstance(select_exp.this, exp.Cast):
+                continue
+
+            meta = select_exp.args["alias"]._meta
+
+        elif isinstance(select_exp, exp.Column):
+            meta = select_exp.parts[-1]._meta
+        else:
+            continue
```
georgesittas:
```suggestion
        if isinstance(select_exp, exp.Alias):
            meta = select_exp.args["alias"].meta
        elif isinstance(select_exp, exp.Column):
            meta = select_exp.parts[-1].meta
        else:
            continue
```

sqlmesh/lsp/hints.py
```diff
@@ -0,0 +1,157 @@
...
+    try:
+        query = normalize_identifiers(query.copy(), dialect=dialect)
+        root = build_scope(query)
+
+        if not root:
+            return []
+
+        return _get_type_hints_for_expression(
+            root.expression, dialect, columns_to_types, start_line, end_line
+        )
```
georgesittas: I don't think we need to build a scope here. I'd get rid of `_get_type_hints_for_expression` and do something like:

```python
return [
    _get_type_hints_for_select(select, ...)
    for q in query.walk(prune=lambda n: not isinstance(n, exp.SetOperation))
    if isinstance(select := q.unnest(), exp.Select)
]
```

georgesittas (commented):
A couple of comments from me @kelsin, LGTM otherwise.

sqlmesh/lsp/hints.py
kelsin: This doesn't appear to be the same logic. The inside if checks if the thing being aliased (`select_exp.this` is a Cast)

georgesittas (approved):
Awesome work 👍

PR #4642, opened 2025-06-03
Feat: Introduce plan explain mode

izeigerman:
This update introduces the new `--explain` flag to the `sqlmesh plan` command which prints out a detailed explanation of the plan instead of evaluating it.

Each model represented in a format "\<view name in the target environment> -> \<the physical table / view that will be used for this model version>".

Presently, this command is only supported in the terminal. Notebook support will be tackled separately. 

Below are a few example explanations printed for different typical scenarios.

Forward-only change in dev:
Screenshot 2025-06-03 at 1 24 59 PM

Forward-only change in prod:
Screenshot 2025-06-03 at 1 25 28 PM

Breaking change in dev:
Screenshot 2025-06-03 at 1 25 47 PM

Breaking change in production after previously backfilling it in dev:
Screenshot 2025-06-03 at 1 30 42 PM

Model restatement without changes:
Screenshot 2025-06-03 at 1 33 47 PM

### DIFF
--- a/sqlmesh/core/plan/explainer.py
+++ b/sqlmesh/core/plan/explainer.py
@@ -1,0 +1,486 @@
+import abc
+import typing as t
+
+from dataclasses import dataclass
+from rich.console import Console as RichConsole
+from rich.tree import Tree
+from sqlglot.dialects.dialect import DialectType
+from sqlmesh.core.console import Console, TerminalConsole, get_console
+from sqlmesh.core.environment import EnvironmentNamingInfo
+from sqlmesh.core.plan.definition import EvaluatablePlan, SnapshotIntervals
+from sqlmesh.core.plan.evaluator import (
+    PlanEvaluator,
+    get_audit_only_snapshots,
+    get_snapshots_to_create,
+)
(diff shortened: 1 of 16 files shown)


sqlmesh/core/plan/explainer.py
```diff
@@ -0,0 +1,281 @@
...
+        explainer_console = _get_explainer_console(
+            self.console, plan.environment, self.default_catalog
+        )
+        explainer_console.explain(plan_stages)
+
+
+class ExplainerConsole(abc.ABC):
+    @abc.abstractmethod
+    def explain(self, stages: t.List[stages.PlanStage]) -> None:
+        pass
```
georgesittas, line 52: Is the vision to gradually move the abstract console classes in `console.py` to each respective module? Curious to understand because this seems like a new pattern.

izeigerman: > Is the vision to gradually move the abstract console classes in console.py to each respective module?

I'm less opinionated about the module structure, but otherwise - yes. We do want to breakdown the giant console class into smaller, domain-specific entities.

PR #4643, opened 2025-06-03
Fix: Support coercion of SQL literal expressions into literal types for macro arguments

izeigerman:


tests/core/test_macros.py
```diff
@@ -1087,3 +1093,33 @@ def test_macro_with_spaces():
...
+    assert macro_evaluator.transform(expression).sql() == "'test_literal_a'"
+
+    expression = d.parse_one("@TEST_LITERAL_TYPE('test_literal_b')")
+    assert macro_evaluator.transform(expression).sql() == "'test_literal_b'"
+
+    expression = d.parse_one("@TEST_LITERAL_TYPE(1)")
+    assert macro_evaluator.transform(expression).sql() == "'1'"
+
+    expression = d.parse_one("@TEST_LITERAL_TYPE(True)")
+    assert macro_evaluator.transform(expression).sql() == "'True'"
```
georgesittas, line 1109: Are these two correct? Why do we coerce them both into strings instead of python `int` and `bool`, respectively?

izeigerman: This is the implementation detail of the macro that we're testing:
```
    @macro()
    def test_literal_type(evaluator, a: t.Literal["test_literal_a", "test_literal_b", 1, True]):
        if isinstance(a, exp.Expression):
            raise SQLMeshError("Coercion failed")
        return f"'{a}'"
```

georgesittas: Ah, whoops, thanks for pointing out.

georgesittas (approved):

PR #4646, opened 2025-06-04
fix: cicd bot correctly surface plan errors

eakmanrq:
Prior to this PR, plan errors would not be surfaced back to the user and would give a confusing `action_required` check title. Now the title communicates a plan error and then the plan error is shown in the summary.


erindru:
Oh, interesting. Have we [raised similar PR's](https://github.com/TobikoData/sqlmesh/pull/4648) at the same time?

PR #4652, opened 2025-06-04
Feat(lsp): Add support for go to and find all References for CTEs

themisvaltinos:
This update adds the "Find All References" and "Go to References" functionality for Common Table Expressions (CTEs) when triggered from its definition for the latter or any usage location for both

cte_references

### DIFF
--- a/sqlmesh/lsp/main.py
+++ b/sqlmesh/lsp/main.py
@@ -112,13 +112,22 @@ class SQLMeshLanguageServer:
         @self.server.feature(ALL_MODELS_FEATURE)
         def all_models(ls: LanguageServer, params: AllModelsRequest) -> AllModelsResponse:
             uri = URI(params.textDocument.uri)
+
+            # Get the document content
+            content = None
+            try:
+                document = ls.workspace.get_text_document(params.textDocument.uri)
+                content = document.source
+            except Exception:
+                pass
+
             try:
                 context = self._context_get_or_load(uri)
-                return context.get_autocomplete(uri)
+                return context.get_autocomplete(uri, content)
             except Exception as e:
                 from sqlmesh.lsp.completions import get_sql_completions
 
-                return get_sql_completions(None, URI(params.textDocument.uri))
+                return get_sql_completions(None, URI(params.textDocument.uri), content)
 
         @self.server.feature(RENDER_MODEL_FEATURE)
         def render_model(ls: LanguageServer, params: RenderModelRequest) -> RenderModelResponse:
@@ -491,8 +500,16 @@ class SQLMeshLanguageServer:
                 uri = URI(params.text_document.uri)
                 context = self._context_get_or_load(uri)
 
+                # Get the document content
+                content = None
+                try:
+                    document = ls.workspace.get_text_document(params.text_document.uri)
+                    content = document.source
+                except Exception:
+                    pass
+
                 # Get completions using the existing completions module
-                completion_response = context.get_autocomplete(uri)
+                completion_response = context.get_autocomplete(uri, content)
 
                 completion_items = []
                 # Add model completions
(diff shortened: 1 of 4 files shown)


sqlmesh/lsp/main.py
```diff
@@ -366,6 +364,35 @@ def goto_definition(
...
+
+                # Convert references to Location objects
+                locations = []
+                for ref in cte_references:
+                    locations.append(
+                        types.Location(
+                            uri=ref.uri,
+                            range=ref.range,
+                        )
+                    )
```
georgesittas:
```suggestion
                locations = [types.Location(uri=ref.uri, range=ref.range) for ref in cte_references]
```

sqlmesh/lsp/reference.py
```diff
@@ -478,3 +469,78 @@ def get_built_in_macro_reference(macro_name: str, macro_range: Range) -> t.Optio
...
+    Returns:
+        A list of references to the CTE (including its definition and all usages)
+    """
+    references = get_model_definitions_for_a_path(lint_context, document_uri)
+
+    # Filter for CTE references (those with target_range set and same URI)
+    # TODO: Consider extending Reference class to explicitly indicate reference type instead
+    cte_references = [
+        ref for ref in references if ref.target_range is not None and ref.uri == document_uri.value
+    ]
```
georgesittas, line 496: Using `target_range` as a proxy for determining whether a reference corresponds to a CTE is confusing, imo. I think we should make this more explicit and document why its existence means the ref is a CTE.

sqlmesh/lsp/reference.py
```diff
@@ -478,3 +469,78 @@ def get_built_in_macro_reference(macro_name: str, macro_range: Range) -> t.Optio
...
+        return []
+
+    target_cte_definition_range = None
+    for ref in cte_references:
+        # Check if cursor is on a CTE usage
+        if _position_within_range(position, ref.range):
+            target_cte_definition_range = ref.target_range
+            break
+        # Check if cursor is on the CTE definition
+        elif ref.target_range and _position_within_range(position, ref.target_range):
```
georgesittas, line 508: I guess the `target_range` check here is only used to satisfy mypy, right?

themisvaltinos: yes exacrly

sqlmesh/lsp/reference.py
themisvaltinos: yes this is why I left the comment  since I also figured it out by the comment in line 204, so I can extend the Reference with ref type if @benfdking you agree as well, or some other way, or document it

benfdking: Yep, I think we'll want a discriminated union of some sort to define the types of these that we support.

tests/lsp/test_reference_cte_find_all.py
```diff
@@ -0,0 +1,140 @@
...
+from tests.lsp.test_reference_cte import find_ranges_from_regex
+
+
+def test_cte_find_all_references():
+    context = Context(paths=["examples/sushi"])
+    lsp_context = LSPContext(context)
+
+    sushi_customers_path = next(
+        path
+        for path, info in lsp_context.map.items()
```
benfdking, line 15: I think we should also have an e2e test for this under `vscode/extension/tests` in `vscode/extension` there is a guide on how to set this up.

benfdking (commented):
I think we should definitely add an integration test.

tests/lsp/test_reference_cte_find_all.py
themisvaltinos: thanks added e2e tests

sqlmesh/lsp/main.py
themisvaltinos: thanks revised it

PR #4661, opened 2025-06-04
Feat: add typo suggestions to MODEL block field errors

treysp:
Before
```
  Invalid extra fields {'descriptions', 'dialects'} in the model definition
```

```
  Invalid extra fields {'batch_sizes'} in the model kind 'INCREMENTAL_BY_TIME_RANGE'
```

After
```
  Invalid field names present in the MODEL block: 'dialects', 'descriptions'
  
    - dialects: Did you mean 'dialect'?
    - descriptions: Did you mean 'description'?
```

```
  Invalid field name present in the MODEL block 'kind INCREMENTAL_BY_TIME_RANGE(' field: 'batch_sizes'. Did you mean 'batch_size'?
```


sqlmesh/core/model/kind.py
```diff
@@ -1015,7 +1015,9 @@ def create_model_kind(v: t.Any, dialect: str, defaults: t.Dict[str, t.Any]) -> M
                 actual_kind_type, _ = custom_materialization
                 return actual_kind_type(**props)
 
-        validate_extra_and_required_fields(kind_type, set(props), f"model kind '{name}'")
+        validate_extra_and_required_fields(
+            kind_type, set(props), f"MODEL block 'kind {name}(' field"
```
izeigerman: the open parenthesis looks somewhat of weird

izeigerman (approved):

sqlmesh/core/model/kind.py
treysp: removed

sqlmesh/core/model/common.py
```diff
@@ -267,13 +268,33 @@ def validate_extra_and_required_fields(
...
     if extra_fields:
-        raise_config_error(f"Invalid extra fields {extra_fields} in the {entity_name}")
+        extra_field_names = "'" + "', '".join(extra_fields) + "'"
+
+        all_fields = klass.all_fields()
+        close_matches = {}
+        for field in extra_fields:
+            matches = get_close_matches(field, all_fields, n=1)
+            if matches:
+                close_matches[field] = matches[0]
```
benfdking, line 285: not important: This looks great to me and is a lovely bit of kit!

@izeigerman this is an example of the thing I talked about on Tuesday when I was saying building with sympathy for the "LSP"

To me, this is a perfect example of something that could be a diagnostic, e.g., if this were a structured error, we could
1. Highlight it in line, e.g., under the misspelled word
2. Suggest a code action that could correct the mistyped word to the 'close_match' 

Actually even more than just for the LSP if you printed it correctly, you could print a link to the file with the right code line that IDE's would pick up as a link to the right file and place.

benfdking (approved):

georgesittas (approved):

sqlmesh/core/model/common.py
izeigerman: I hear you. Though, I don't believe all the relevant information is available here—for example, the position of the field in the code. So Trey wouldn't be able to express this sympathy in the PR even if he wanted to, since he doesn't have access to the original source, only the Pydantic model object.

PR #4663, opened 2025-06-05
feat: add position details to linter

benfdking:
- adds ability for linter to return range to specify where the error
  gets returned
- maps that correctly over for the lsp
- for the builtin selectstar rule, add the range

### DIFF
--- a/sqlmesh/core/linter/definition.py
+++ b/sqlmesh/core/linter/definition.py
@@ -8,7 +8,7 @@ import operator as op
 from collections.abc import Iterator, Iterable, Set, Mapping, Callable
 from functools import reduce
 from sqlmesh.core.model import Model
-from sqlmesh.core.linter.rule import Rule, RuleViolation
+from sqlmesh.core.linter.rule import Rule, RuleViolation, Range
 from sqlmesh.core.console import LinterConsole, get_console
 
 if t.TYPE_CHECKING:
@@ -74,6 +74,7 @@ class Linter:
                 violation_msg=violation.violation_msg,
                 model=model,
                 violation_type="error",
+                violation_range=violation.violation_range,
             )
             for violation in error_violations
         ] + [
@@ -82,6 +83,7 @@ class Linter:
                 violation_msg=violation.violation_msg,
                 model=model,
                 violation_type="warning",
+                violation_range=violation.violation_range,
             )
             for violation in warn_violations
         ]
@@ -149,7 +151,8 @@ class AnnotatedRuleViolation(RuleViolation):
         violation_msg: str,
         model: Model,
         violation_type: t.Literal["error", "warning"],
+        violation_range: t.Optional[Range] = None,
     ) -> None:
-        super().__init__(rule, violation_msg)
+        super().__init__(rule, violation_msg, violation_range)
         self.model = model
         self.violation_type = violation_type
--- a/sqlmesh/core/linter/helpers.py
+++ b/sqlmesh/core/linter/helpers.py
@@ -1,0 +1,115 @@
+from pathlib import Path
+
+from sqlmesh.core.linter.rule import Position, Range
+from sqlmesh.utils.pydantic import PydanticModel
+import typing as t
+
+
+class TokenPositionDetails(PydanticModel):
+    """
+    Details about a token's position in the source code in the structure provided by SQLGlot.
+
+    Attributes:
+        line (int): The line that the token ends on.
+        col (int): The column that the token ends on.
+        start (int): The start index of the token.
+        end (int): The ending index of the token.
+    """
+
+    line: int
+    col: int
+    start: int
+    end: int
+
+    @staticmethod
+    def from_meta(meta: t.Dict[str, int]) -> "TokenPositionDetails":
+        return TokenPositionDetails(
+            line=meta["line"],
+            col=meta["col"],
+            start=meta["start"],
+            end=meta["end"],
+        )
+
+    def to_range(self, read_file: t.Optional[t.List[str]]) -> Range:
+        """
+        Convert a TokenPositionDetails object to a Range object.
+
+        In the circumstances where the token's start and end positions are the same,
+        there is no need for a read_file parameter, as the range can be derived from the token's
+        line and column. This is an optimization to avoid unnecessary file reads and should
+        only be used when the token represents a single character or position in the file.
+
+        If the token's start and end positions are different, the read_file parameter is required.
+
+        :param read_file: List of lines from the file. Optional
+        :return: A Range object representing the token's position
+        """
+        if self.start == self.end:
+            # If the start and end positions are the same, we can create a range directly
+            return Range(
+                start=Position(line=self.line - 1, character=self.col - 1),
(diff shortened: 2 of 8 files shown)


sqlmesh/core/linter/definition.py
```diff
@@ -149,7 +151,8 @@ def __init__(
         violation_msg: str,
         model: Model,
         violation_type: t.Literal["error", "warning"],
+        violation_range: t.Optional["Range"] = None,
```
georgesittas: Nit: do we need to use a forward ref here?
```suggestion
        violation_range: t.Optional[Range] = None,
```

sqlmesh/core/linter/helpers.py
```diff
@@ -0,0 +1,115 @@
...
+    Read the file and return the content within the specified range.
+
+    Args:
+        file: Path to the file to read
+        text_range: The range of text to extract
+
+    Returns:
+        The content within the specified range
+    """
+    with file.open("r") as f:
```
```suggestion
    with file.open("r", encoding="utf-8") as f:
```

tests/core/linter/test_linter.py
georgesittas, line 1: Empty file.

tests/lsp/test_diagnostics.py
```diff
@@ -0,0 +1,65 @@
...
+from sqlmesh.lsp.context import LSPContext
+from sqlmesh.lsp.uri import URI
+
+
+def test_diagnostic_on_sushi() -> None:
+    # Copy sushi example to a temporary directory
+    sushi_original_path = Path(__file__).parent.parent.parent / "examples" / "sushi"
+
+    tempfolder = tempfile.TemporaryDirectory()
+    temp_path = Path(tempfolder.name)
```
georgesittas: FYI, there is a `tmp_path` pytest fixture you can use here. May help cleaning up the testing setup by avoiding the `tempfile` and `copytree` gymnastics.

sqlmesh/core/linter/rule.py
```diff
@@ -22,6 +22,20 @@ class RuleLocation(PydanticModel):
     start_line: t.Optional[int] = None
 
 
+class Position(PydanticModel):
+    """The position of a rule violation in a file, the position follows the LSP standard."""
```
georgesittas: Why do these wrapper classes inherit from `PydanticModel` instead of making them data classes? I think the latter is more lightweight, so may be worth exploring if we don't care about serializing these objects.

sqlmesh/core/linter/rule.py
```diff
@@ -22,6 +22,20 @@ class RuleLocation(PydanticModel):
...
+
+    line: int
+    character: int
+
+
+class Range(PydanticModel):
+    """The range of a rule violation in a file. The range follows the LSP standard."""
+
+    start: Position
+    end: Position
```
georgesittas: Should we document that this represents a `[start, end)` range?

sqlmesh/core/linter/rule.py
```diff
@@ -22,6 +22,20 @@ class RuleLocation(PydanticModel):
     start_line: t.Optional[int] = None
 
 
+class Position(PydanticModel):
+    """The position of a rule violation in a file, the position follows the LSP standard."""
+
+    line: int
+    character: int
```
georgesittas, line 31: Does "character" mean column offset in the line? Why not use `column` or something similar so that the naming is consistent?

I also suggest we start documenting some details, like whether it's 1-based indexing and stuff, so reading the docstring makes it obvious to understand what's going on.

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -4,21 +4,39 @@
...
+        query = model.query
+        if query.is_star:
+            violation_range = self._get_range(model)
+            return self.violation(violation_range=violation_range)
+        return None
+
+    def _get_range(self, model: SqlModel) -> t.Optional[Range]:
+        """Get the range of the violation if available."""
+        try:
+            if len(model.query.expressions) == 1 and isinstance(model.query.expressions[0], Star):
```
georgesittas, line 31: Should this instead call `find(Star)` on the model's query? Note that SQL like `t.*` won't be picked up by this check because the projection is a `Column` that contains a `Star` child in `this`.

sqlmesh/core/linter/rule.py
benfdking: To me this is really what I was hoping 

`The range follows the LSP standard.`

^ I don't want to reinvent the wheel, I just want them to follow the LSP standard. So if you are confused, just go there.

sqlmesh/core/linter/rule.py
benfdking: idem, just want to specify LSP standard

sqlmesh/core/linter/rule.py
benfdking: 🤷  I think serializing could quickly become useful but as you say easier to add after.

sqlmesh/core/linter/rules/builtin.py
benfdking: I think this can be improved, but I'm hesitant to do so right away because it adds complexity. For me, I'm just in the simplest case of returning a range, which I think we can expand on over time. 

There's also a lot of other changes so rather do this later.

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -4,21 +4,39 @@
...
-    """Query should not contain SELECT * on its outer most projections, even if it can be expanded."""
+    """Query should not contain SELECT * on its outermost projections, even if it can be expanded."""
 
     def check_model(self, model: Model) -> t.Optional[RuleViolation]:
+        # Only applies to SQL models, as other model types do not have a query.
         if not isinstance(model, SqlModel):
             return None
-
-        return self.violation() if model.query.is_star else None
+        query = model.query
```
VaggelisD: we don't really use `query` anywhere, it can be inlined back into the branch

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -4,21 +4,39 @@
...
+            violation_range = self._get_range(model)
+            return self.violation(violation_range=violation_range)
+        return None
+
+    def _get_range(self, model: SqlModel) -> t.Optional[Range]:
+        """Get the range of the violation if available."""
+        try:
+            if len(model.query.expressions) == 1 and isinstance(model.query.expressions[0], Star):
+                position = TokenPositionDetails.from_meta(model.query.expressions[0].meta)
+                if position.start != position.end:
```
VaggelisD: This branch doesn't seem needed, if `start` and `end` aren't provided then `to_range` will raise

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -4,21 +4,39 @@
...
+                    return None
+                return position.to_range(None)
```
VaggelisD: This `return` chain seems a bit verbose, can we simplify somehow like:

```
range = None

try: 
  ...
  range = position_to_range()
except:
  pass

return range
```

PR #4667, opened 2025-06-05
Add macros to LSP completion

benfdking:
## Summary
- expose macros in `AllModelsResponse`
- include macros in Python/TypeScript completion handling
- test that macros show in completions

## Testing
- `pytest -q` *(fails: ModuleNotFoundError: No module named 'pymysql')*

------
https://chatgpt.com/codex/tasks/task_e_6841b7a645948330889be83831471693

### DIFF
--- a/sqlmesh/lsp/completions.py
+++ b/sqlmesh/lsp/completions.py
@@ -1,6 +1,7 @@
 from functools import lru_cache
 from sqlglot import Dialect, Tokenizer
 from sqlmesh.lsp.custom import AllModelsResponse
+from sqlmesh import macro
 import typing as t
 from sqlmesh.lsp.context import AuditTarget, LSPContext, ModelTarget
 from sqlmesh.lsp.uri import URI
@@ -26,6 +27,7 @@ def get_sql_completions(
     return AllModelsResponse(
         models=list(get_models(context, file_uri)),
         keywords=all_keywords,
+        macros=list(get_macros(context)),
     )
 
 
@@ -54,6 +56,11 @@ def get_models(context: t.Optional[LSPContext], file_uri: t.Optional[URI]) -> t.
                 all_models.discard(model)
 
     return all_models
+
+
+def get_macros(context: t.Optional[LSPContext]) -> t.Set[str]:
+    """Return a list of all macros available in the context."""
+    return set(macro.get_registry().keys())
 
 
 def get_keywords(context: t.Optional[LSPContext], file_uri: t.Optional[URI]) -> t.Set[str]:
@@ -159,3 +166,4 @@ def extract_keywords_from_content(content: str, dialect: t.Optional[str] = None)
         pass
 
     return keywords
+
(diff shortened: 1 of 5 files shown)


sqlmesh/lsp/completions.py
```diff
@@ -56,6 +58,17 @@ def get_models(context: t.Optional[LSPContext], file_uri: t.Optional[URI]) -> t.
...
 
+def get_macros(context: t.Optional[LSPContext]) -> t.Set[str]:
+    """Return a set of all macros with the ``@`` prefix."""
+    names = set(macro.get_registry().keys())
+    if context is not None:
+        try:
+            names.update(context.context._macros.keys())
+        except Exception:
+            pass
+    return {f"@{name}" for name in names}
```
georgesittas: This only returns the builtin macros IIRC, not all of them. Nit: both `keys()` calls can be removed.

sqlmesh/lsp/completions.py
```diff
@@ -56,6 +60,17 @@ def get_models(context: t.Optional[LSPContext], file_uri: t.Optional[URI]) -> t.
     return all_models
 
 
+def get_macros(context: t.Optional[LSPContext], file_uri: t.Optional[URI]) -> t.Set[str]:
+    """Return a set of all macros with the ``@`` prefix."""
+    names = set(macro.get_registry())
```
themisvaltinos, line 65: I don't think this is necessary since these are already contained in context.context._macros. The only case where it might make sense is if context can ever be None, but can that actually happen here? If so, then I suppose this is justified

sqlmesh/lsp/main.py
```diff
@@ -595,6 +595,16 @@ def completion(
                             detail="SQLMesh Model",
                         )
                     )
+                # Add macro completions
+                for macro_name in completion_response.macros:
+                    completion_items.append(
+                        types.CompletionItem(
+                            label=f"@{macro_name}",  # shown in UI
+                            insert_text=f"{macro_name}",  # text that will be inserted
```
themisvaltinos: I might be reading it wrong but if the user types `@eac` and selects `@each` wouldn't this insert `each` instead of `@each` since the insert text doesn't contain the `@` symbol?

sqlmesh/lsp/completions.py
benfdking: That's the goal. Why do you say that can never happen there?

sqlmesh/lsp/main.py
benfdking: I've adjusted this a little.

sqlmesh/lsp/completions.py
themisvaltinos: ok right I guess it happens because someone might start typing without the context having loaded yet

sqlmesh/lsp/main.py
themisvaltinos: looks good thanks 👍

sqlmesh/lsp/completions.py
benfdking: Yep, that's the plan!

PR #4668, opened 2025-06-05
Feat: Refactor model test output [skip ci]

VaggelisD:
**Before this PR**:

- Non-verbose:
before_vv

- Verbose
before

<br />
<br />

**After this PR**
- Non-verbose
after

- Verbose
after_vv

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -463,7 +463,7 @@ class Console(
 
     @abc.abstractmethod
     def log_test_results(
-        self, result: unittest.result.TestResult, output: t.Optional[str], target_dialect: str
+        self, result: ModelTextTestResult, output: t.Optional[str], target_dialect: str
     ) -> None:
         """Display the test result and output.
 
@@ -498,7 +498,9 @@ class Console(
         """Stop loading for the given id."""
 
     @abc.abstractmethod
-    def log_unit_test_results(self, result: ModelTextTestResult, test_duration: float) -> None:
+    def log_unit_test_results(
+        self, result: ModelTextTestResult, test_duration: t.Optional[float] = None
+    ) -> None:
         """Print the unit test results."""
 
 
@@ -674,7 +676,7 @@ class NoopConsole(Console):
             plan_builder.apply()
 
     def log_test_results(
-        self, result: unittest.result.TestResult, output: t.Optional[str], target_dialect: str
+        self, result: ModelTextTestResult, output: t.Optional[str], target_dialect: str
     ) -> None:
         pass
 
@@ -782,7 +784,9 @@ class NoopConsole(Console):
     def stop_destroy(self, success: bool = True) -> None:
         pass
 
-    def log_unit_test_results(self, result: ModelTextTestResult, test_duration: float) -> None:
+    def log_unit_test_results(
+        self, result: ModelTextTestResult, test_duration: t.Optional[float] = None
+    ) -> None:
         pass
 
 
@@ -1961,9 +1965,13 @@ class TerminalConsole(Console):
             plan_builder.apply()
 
     def log_test_results(
-        self, result: unittest.result.TestResult, output: t.Optional[str], target_dialect: str
+        self, result: ModelTextTestResult, output: t.Optional[str], target_dialect: str
     ) -> None:
         divider_length = 70
+
+        self.log_unit_test_results(result)
+
+        self._print("\n")
         if result.wasSuccessful():
             self._print("=" * divider_length)
             self._print(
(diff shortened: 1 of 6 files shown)


sqlmesh/core/console.py
```diff
@@ -496,6 +497,10 @@ def loading_start(self, message: t.Optional[str] = None) -> uuid.UUID:
     def loading_stop(self, id: uuid.UUID) -> None:
         """Stop loading for the given id."""
 
+    @abc.abstractmethod
+    def log_unit_test_results(self, result: ModelTextTestResult, test_duration: float) -> None:
+        """Print the unit test results."""
+
```
georgesittas: Let's factor this out into a `UnitTestingConsole`.

sqlmesh/core/test/definition.py
```diff
@@ -10,6 +10,11 @@
 from pathlib import Path
 from unittest.mock import patch
 
+from rich.table import Table
+from rich.align import Align
+
+import numpy as np
+import pandas as pd
```
georgesittas: We don't allow top-level pandas/numpy imports anymore.

sqlmesh/core/test/definition.py
```diff
@@ -60,6 +66,8 @@ def __init__(
         preserve_fixtures: bool = False,
         default_catalog: str | None = None,
         concurrency: bool = False,
+        verbosity: Verbosity = Verbosity.DEFAULT,
+        rich_output: bool = True,
```
georgesittas: Why do we need this option and not just default to rich output all the time?

sqlmesh/core/test/definition.py
```diff
@@ -294,10 +305,29 @@ def _to_hashable(x: t.Any) -> t.Any:
...
+
+                    levels = t.cast(MultiIndex, diff.columns).levels[0]
+                    for col in levels:
+                        col_diff = diff[col]
+                        if not col_diff.empty:
+                            table = df_to_table(
+                                f"[bold red]Column '{col}' mismatch[/bold red]", col_diff
+                            )
+                            args.append(table)
+
```
georgesittas, line 329: What does this do?

georgesittas (commented):
I like this, curious to see what others have to say. Is it possible to point out the exact differences per different rows reported, similar to `git diff`?

sqlmesh/core/test/definition.py
VaggelisD: Yep rebasing issue, will remove

sqlmesh/core/test/definition.py
VaggelisD: My issue with that had to do with `test_test.py`, we check the payload itself in the exception (`_check_successful_or_raise`) instead of the output. That used to be a `str` but is now a tuple of `str | Rich.Table` so the comparison fails.

By allowing the old style we can maintain the tests as they are e.g `test_row_order`

sqlmesh/core/test/definition.py
VaggelisD: This is for the verbose mode. Essentially:
- Non verbose prints the entire `diff` dataframe, which has the columns `("col_1: Expected", "col_1: Actual", ..., "col_n: Expected", "col_n: Actual")`
- Verbose splits the `diff` apart in tables that contain the `Actual` vs `Expected` pairs for each `col_i`

sqlmesh/core/test/definition.py
georgesittas: Doesn't seem like a good argument to keep it. We should instead update the tests, imo, unless there's a compelling reason to expose a non-rich formatting option.

How does this look in formats like CI/CD bot, magics, etc?

tests/core/test_test.py
```diff
@@ -924,10 +925,10 @@ def test_missing_column_failure(sushi_context: Context, full_model_without_ctes:
             context=sushi_context,
         ).run(),
         expected_msg=(
-            "AssertionError: Data mismatch (exp: expected, act: actual)\n\n"
-            "  value        ds    \n"
-            "    exp act   exp act\n"
-            "0  None   2  None   3\n"
+            "AssertionError: Data mismatch\n\n"
```
themisvaltinos: now that you use rich check `capture_console_output` and `strip_ansi_codes` methods in `test_table_diff.py` so that the assertions strip the output of rich styling so that they can work in the tests. maybe if you end up using these they can be moved to some utils folder so that they can be shared

VaggelisD:
Still trying to figure out the smaller output details so will close this PR and reopen once those are cleaner (plus comments are addressed)

PR #4671, opened 2025-06-05
Feat(table_diff): Add option for case insensitive schema comparisons

erindru:


sqlmesh/core/table_diff.py
```diff
@@ -36,29 +36,78 @@ class SchemaDiff(PydanticModel, frozen=True):
     source_alias: t.Optional[str] = None
     target_alias: t.Optional[str] = None
     model_name: t.Optional[str] = None
+    ignore_case: bool = False
+
+    @property
+    def _normalized_source_schema(self) -> t.Dict[str, exp.DataType]:
```
izeigerman: I believe "normalized" means something else in other part of the code base. Should not overload the term?

sqlmesh/cli/main.py
```diff
@@ -938,6 +938,11 @@ def create_external_models(obj: Context, **kwargs: t.Any) -> None:
     multiple=True,
     help="Specify one or more models to data diff. Use wildcards to diff multiple models. Ex: '*' (all models with applied plan diffs), 'demo.model+' (this and downstream models), 'git:feature_branch' (models with direct modifications in this branch only)",
 )
+@click.option(
+    "--schema-diff-ignore-case",
```
izeigerman, line 942: Is this relevant for magics?

sqlmesh/core/table_diff.py
erindru: I originally called this `_lowercased_source_schema` but that was misleading because it is only lowercased if `ignore_case` is set to `True`.

Maybe `_comparable_source_schema`?

sqlmesh/cli/main.py
erindru: I guess so, I keep forgetting we have a duplicate copy of our CLI arguments and help text to support Notebook magics.

I'll add it

PR #4680, opened 2025-06-06
Feat(lsp): Add support for find and go to references for Model usages

themisvaltinos:
A continuation of #4652  this update adds the "Find All References" and "Go to References" functionality for Models in the VSCode extension. This supports cross-file navigation between model definitions and their usages throughout the project, including model references in audits.

modelrefs


sqlmesh/lsp/main.py
```diff
@@ -395,8 +395,8 @@ def goto_definition(
                 references = get_references(self.lsp_context, uri, params.position)
                 location_links = []
                 for reference in references:
-                    # Use target_range if available (for CTEs), otherwise default to start of file
-                    if reference.target_range:
+                    # Use target_range if available (CTEs, Macros), otherwise default to start of file
+                    if not isinstance(reference, LSPModelReference) and reference.target_range:
```
benfdking: nit: Now that you know it's that type do you need to check the range

sqlmesh/lsp/reference.py
```diff
@@ -23,21 +24,37 @@
...
-        uri: The uri of the referenced model
-        markdown_description: The markdown description of the referenced model
-        target_range: The range of the definition for go-to-definition (optional, used for CTEs)
-    """
+class LSPBaseReference(PydanticModel):
+    """Base class for all LSP reference types."""
 
     range: Range
     uri: str
     markdown_description: t.Optional[str] = None
```
benfdking: Why have a base reference if the description is optional? Would it not be worth bringing that up and getting rid of this base class?

benfdking (commented):
Small things

sqlmesh/lsp/main.py
themisvaltinos: no good point removed it

sqlmesh/lsp/reference.py
themisvaltinos: revised it to not have the base reference class and moved them to the corresponding classes

benfdking (approved):

PR #4686, opened 2025-06-06
Enhances MSSQL connection with pyodbc support v2

mattiasthalen:
Adds support for `pyodbc` as an alternative driver for MSSQL connections, providing users with more flexibility and authentication options.

- Introduces a `driver` option to choose between `pymssql` and `pyodbc`.
- Implements connection string generation for `pyodbc` based on connection parameters.
- Includes validation for different authentication methods when using `pyodbc`.
- Updates dependencies to include `pyodbc`.

>**NOTE** Had to create a new PR, the merge conflicts was gnarly, easier to just restart.
> Replaces PR #4335

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -1444,40 +1444,26 @@ class MSSQLConnectionConfig(ConnectionConfig):
         if not isinstance(data, dict):
             return data
 
-        check_import = str_to_bool(str(data.pop("check_import", True)))
-        if not check_import:
-            return data
-
         driver = data.get("driver", "pymssql")
 
-        try:
-            if driver == "pymssql":
-                importlib.import_module("pymssql")
-            elif driver == "pyodbc":
-                importlib.import_module("pyodbc")
-            else:
-                raise ValueError(f"Unsupported driver: {driver}")
-        except ImportError:
-            if debug_mode_enabled():
-                raise
+        # Define the mapping of driver to import module and extra name
+        driver_configs = {"pymssql": ("pymssql", "mssql"), "pyodbc": ("pyodbc", "mssql-odbc")}
 
-            logger.exception("Failed to import the MSSQL engine library")
+        if driver not in driver_configs:
+            raise ValueError(f"Unsupported driver: {driver}")
 
-            if driver == "pymssql":
-                extra_name = "mssql"
-            elif driver == "pyodbc":
-                extra_name = "mssql-odbc"
-            else:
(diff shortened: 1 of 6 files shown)


erindru:
Nice work!

sqlmesh/core/config/connection.py
```diff
@@ -1422,17 +1422,66 @@ class MSSQLConnectionConfig(ConnectionConfig):
...
     type_: t.Literal["mssql"] = Field(alias="type", default="mssql")
 
-    _engine_import_validator = _get_engine_import_validator("pymssql", "mssql")
+    @model_validator(mode="before")
+    @classmethod
+    def _mssql_engine_import_validator(cls, data: t.Any) -> t.Any:
+        if not isinstance(data, dict):
+            return data
+
+        check_import = str_to_bool(str(data.pop("check_import", True)))
```
erindru: There is a bit of code duplication here but I guess it's because `_get_engine_import_validator` assumes one library / extra per engine adapter and we don't know which one to check until the value for "driver" has been read.

Are we able to conditionally delegate to either `_get_engine_import_validator("pymssql", "mssql")` or `_get_engine_import_validator("pyodbc", "mssql-odbc")` depending on what `driver` is set to?

sqlmesh/core/config/connection.py
```diff
@@ -1422,17 +1422,52 @@ class MSSQLConnectionConfig(ConnectionConfig):
...
+
+        import_module, extra_name = driver_configs[driver]
+
+        # Conditionally delegate to the existing _get_engine_import_validator
+        # Create a validator for the specific driver and call its inner function
+        validator_func = _get_engine_import_validator(import_module, driver, extra_name)
+
+        # Extract the inner validate function from the decorated validator
+        # The validator_func has a __wrapped__ attribute that contains the original function
+        inner_validate = getattr(validator_func, "__wrapped__", validator_func)
```
erindru: It looks like this doesn't work for Python 3.9. I think what we need to is make the `@model_validator` decorator in `_get_engine_import_validator` conditional.

Something like:
 - change the `_get_engine_import_validator` signature to take an extra parameter, `decorate: bool = True`
 - remove the `@model_validator` annotation on the inner method
 - ```py
   return model_validator(mode="before")(validate) if decorate else validate
   ```

If we did this, we can use it here like:

```py
return _get_engine_import_validator(import_module, driver, extra_name, decorate=False)(cls, data)
```

erindru (approved):
Thanks for bearing with me!

PR #4692, opened 2025-06-09
Feat(lsp): Add find all and go to references support for Macros

themisvaltinos:
Building on #4652  and #4680, this update adds "Find All References" and "Go to References" support for macros in the SQLMesh VSCode extension.

macrorefs

### DIFF
--- a/sqlmesh/lsp/reference.py
+++ b/sqlmesh/lsp/reference.py
@@ -587,13 +587,88 @@ def get_cte_references(
     return matching_references
 
 
+def get_macro_find_all_references(
+    lint_context: LSPContext, document_uri: URI, position: Position
+) -> t.List[LSPMacroReference]:
+    """
+    Get all references to a macro at a specific position in a document.
+
+    This function finds all usages of a macro across the entire project.
+
+    Args:
+        lint_context: The LSP context
+        document_uri: The URI of the document
+        position: The position to check for macro references
+
+    Returns:
+        A list of references to the macro across all files
+    """
+    # First, get the macro references in the current file to identify the target macro
+    current_file_references = [
+        ref
+        for ref in get_macro_definitions_for_a_path(lint_context, document_uri)
+        if isinstance(ref, LSPMacroReference)
+    ]
+
+    # Find the macro reference at the cursor position
+    target_macro_uri: t.Optional[str] = None
+    target_macro_target_range: t.Optional[Range] = None
+
+    for ref in current_file_references:
+        if _position_within_range(position, ref.range):
+            target_macro_uri = ref.uri
+            target_macro_target_range = ref.target_range
+            break
+
+    if target_macro_uri is None:
+        return []
+
+    # Start with the macro definition
+    all_references: t.List[LSPMacroReference] = [
+        LSPMacroReference(
+            uri=target_macro_uri,
+            range=target_macro_target_range,
+            target_range=target_macro_target_range,
+            markdown_description=None,
+        )
+    ]
+
+    # Search through all SQL and audit files in the project
+    for path, target in lint_context.map.items():
+        if not isinstance(target, (ModelTarget, AuditTarget)):
+            continue
+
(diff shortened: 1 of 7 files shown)


sqlmesh/lsp/reference.py
```diff
@@ -587,13 +587,88 @@ def get_cte_references(
     return matching_references
 
 
+def get_macro_find_all_references(
+    lint_context: LSPContext, document_uri: URI, position: Position
```
benfdking: name

```suggestion
    lsp_context: LSPContext, document_uri: URI, position: Position
```

sqlmesh/lsp/reference.py
```diff
@@ -587,13 +587,88 @@ def get_cte_references(
...
+    target_macro_target_range: t.Optional[Range] = None
+
+    for ref in current_file_references:
+        if _position_within_range(position, ref.range):
+            target_macro_uri = ref.uri
+            target_macro_target_range = ref.target_range
+            break
+
+    if target_macro_uri is None:
+        return []
```
benfdking: nit: I feel like this could be slightly cleaner with a filter and then you could build a pipe of filter with the above filter where you are filtering out for `LSPMacroReferences`

sqlmesh/lsp/reference.py
```diff
@@ -587,13 +587,88 @@ def get_cte_references(
...
+            range=target_macro_target_range,
+            target_range=target_macro_target_range,
+            markdown_description=None,
+        )
+    ]
+
+    # Search through all SQL and audit files in the project
+    for path, target in lint_context.map.items():
+        if not isinstance(target, (ModelTarget, AuditTarget)):
+            continue
```
benfdking: nit: aren't they the only two types at the moment? might be able to skip but I am guessing this might have to do with typing.

tests/lsp/test_reference_macro_find_all.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Verify the macro definition is included
+    definition_refs = [ref for ref in all_references if "utils.py" in ref.uri]
+    assert len(definition_refs) >= 1, "Should include the macro definition in utils.py"
+
+    # Verify the usage in top_waiters is included
+    usage_refs = [ref for ref in all_references if "top_waiters" in ref.uri]
+    assert len(usage_refs) >= 1, "Should include the usage in top_waiters.sql"
+
+    # breakpoint()
```
benfdking: nit: 

```suggestion
```

tests/lsp/test_reference_macro_find_all.py
```diff
@@ -0,0 +1,189 @@
...
+        # Usage in customers.sql
+        {
+            "uri_file": "customers.sql",
+            "range": ((36, 7), (36, 14)),
+        },
+        # Usage in top_waiters.sql
+        {
+            "uri_file": "top_waiters.sql",
+            "range": ((12, 5), (12, 12)),
+        },
```
benfdking: nit: To make these less brittle, I often use `read_range_from_file` to test for the output I expect rather than hard-coding

sqlmesh/lsp/reference.py
themisvaltinos: excellent point, refactoring it now

sqlmesh/lsp/reference.py
themisvaltinos: yes removed this check from both places

tests/lsp/test_reference_macro_find_all.py
themisvaltinos: revised it to use read_range_from_file instead

sqlmesh/lsp/reference.py
themisvaltinos: refactored it to use the filtering logic in one go since it's also more efficient since it stops in the first hit. also changed the model function as well to use similar logic

PR #4695, opened 2025-06-09
Chore!: bump sqlglot to v26.26.0

georgesittas:


tests/core/test_dialect.py
```diff
@@ -643,9 +643,6 @@ def test_model_normalization_quote_flexibility():
         normalize_model_name('"catalog"."db"."table"', default_catalog=None, dialect="spark")
         == '"catalog"."db"."table"'
     )
-    # It doesn't work the other way which is what we currently expect
-    with pytest.raises(ParseError):
-        normalize_model_name("`catalog`.`db`.`table`", default_catalog=None, dialect=None)
```
georgesittas, line 648: This change is due to this [commit](https://github.com/tobymao/sqlglot/commit/26077a47d9db750f44ab1baf9a434596b5bb613b). The `normalize_model_name` helper [relies](https://github.com/TobikoData/sqlmesh/blob/f87d3f0a6dad639584a4b70540ac19010c0ca92c/sqlmesh/core/dialect.py#L1189) on `to_table` under the hood, which is now more lenient and doesn't fail despite the presence of backticks & absence of a dialect like `"bigquery"`.

Basically, the behavior introduced by that sqlglot change is that this table name now literally includes the backticks in the table parts.

erindru (approved):

PR #4696, opened 2025-06-09
Fix: normalize datetime values to strings in unit tests

georgesittas:
Fixes the issue discussed in this Slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1749046530513469

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -258,12 +258,21 @@ class ModelTest(unittest.TestCase):
         actual = actual.replace({np.nan: None})
         expected = expected.replace({np.nan: None})
 
+        # We define this here to avoid a top-level import of numpy and pandas
+        DATETIME_TYPES = (
+            datetime.datetime,
+            datetime.date,
+            datetime.time,
+            np.datetime64,
+            pd.Timestamp,
+        )
+
         def _to_hashable(x: t.Any) -> t.Any:
             if isinstance(x, (list, np.ndarray)):
                 return tuple(_to_hashable(v) for v in x)
             if isinstance(x, dict):
                 return tuple((k, _to_hashable(v)) for k, v in x.items())
-            return str(x) if not isinstance(x, t.Hashable) else x
+            return str(x) if isinstance(x, DATETIME_TYPES) or not isinstance(x, t.Hashable) else x
 
         actual = actual.apply(lambda col: col.map(_to_hashable))
         expected = expected.apply(lambda col: col.map(_to_hashable))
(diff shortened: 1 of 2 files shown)


sqlmesh/core/test/definition.py
```diff
@@ -258,12 +258,21 @@ def assert_equal(
...
+            pd.Timestamp,
+        )
+
         def _to_hashable(x: t.Any) -> t.Any:
             if isinstance(x, (list, np.ndarray)):
                 return tuple(_to_hashable(v) for v in x)
             if isinstance(x, dict):
                 return tuple((k, _to_hashable(v)) for k, v in x.items())
-            return str(x) if not isinstance(x, t.Hashable) else x
+            return str(x) if isinstance(x, DATETIME_TYPES) or not isinstance(x, t.Hashable) else x
```
VaggelisD, line 275: Do we have to pass all values through `_to_hashable` (i.e reduce them down to primitive python types) for dataframe comparison to work properly?

tests/core/test_test.py
```diff
@@ -2609,3 +2609,34 @@ def test_model_test_text_result_reporting_no_traceback(
...
+      inputs:
+        temp_model_with_timestamp:
+          rows:
+            - id: "id1"
+              timestamp_col: "2024-01-02T15:00:00"
+      outputs:
+        query:
+          rows:
+            - id: id1
+              agg_timestamp_col: ["2024-01-02T15:00:00.000000"]
```
VaggelisD, line 2635: Would the same output comparison work properly if we compared an equivalent datetime value from a different string? E.g `2024-01-02T15:00:00.00" or one based on a different timezone that results in the same time?

Ideally we'd normalize all datetime types into a common representation at a common timezone (?)

VaggelisD (approved):
Don't really have a concrete comment, unless I'm downplaying the power of `str(x)`, my hunch is this cause false-negative comparisons again but I don't have a better solution at hand.

sqlmesh/core/test/definition.py
georgesittas: Yep, exactly.

tests/core/test_test.py
georgesittas: I think so, there's a test for this: https://github.com/TobikoData/sqlmesh/blob/f87d3f0a6dad639584a4b70540ac19010c0ca92c/tests/core/test_test.py#L1220

VaggelisD: This looks like it's on the same timezone from a quick look, I was referring to a scenario such as :

```
inputs:
   ...
   time_col: "2020-01-01 05:00:00+05"

outputs:
   ...
   time_col: "2020-01-01 00:00:00+00"
```

this test should be correct, right? But if it's a string comparison (or at best, normalized to a certain precision with zeroes) then it'd fail

georgesittas: I don't think that we encourage specifying time zones in the timestamp values (in the sense that it's unsupported). 

The `+zone` syntax is only mentioned in relation to the `execution_time` variable in the docs, and I think that was intentional at the time.

izeigerman (approved):

PR #4697, opened 2025-06-09
fix: dockerfile for UI

mykalmax:


web/Dockerfile.app
```diff
@@ -4,19 +4,19 @@ WORKDIR /app
...
 
-# Copy the rest of the application
-COPY web/client/ /app/web/client/
+# Copy source files
+COPY web/client/ ./web/client/
+COPY vscode/extension/ ./vscode/extension/
+COPY vscode/bus/ ./vscode/bus/
 
 # Install dependencies
 RUN pnpm install --frozen-lockfile
```
benfdking: nit: To make the docker cache useable, I would install before `COPY`-ing over source files.

PR #4700, opened 2025-06-10
chore(docs): add tobiko logo to tobiko cloud nav tab

caiters:
Added logo to the Tobiko Cloud item in the navigation (targeting the last item in the navigation):

image

- attempted to target by href, but the url for the page changes when the page is active
- another method is doing a theme override, but the nav link file is complex and we would miss out on updates to original theme. i don't think it's worth it
- added a comment to the tobiko cloud section that notes the css has to be updated if another item is added after it in the navigation, but I think it's likely this will remain the last item at least for the near future.


treysp:
Thank you!

treysp (approved):

PR #4702, opened 2025-06-10
Fix: Make plan dates relative to `--execution-time`

erindru:
This addresses some issues with setting `--execution-time` on a plan, which is intended to run a plan "as at" a certain time.

Currently, if you specify a relative start date like `--start '2 days ago'` and then you specify a fixed `--execution-time`, eg `--execution-time '2020-01-05'`, SQLMesh will make the start date relative to the current time and not the execution time.

This gets confusing because it makes the plan start / end and any relative dates show no relation to the execution time.

This is also causes problems when `--execution-time` is set to a value older than the latest prod interval for a model, eg:
```
$ sqlmesh plan dev --execution-time 2020-01-05 --start '2 days ago'
...SNIP
Error: Start date / time (2 days ago) can't be greater than end date / time (1749168000000)
```

This PR tightens up the behavior, adds new checks at the plan building stage and adds tests to verify the relative date behaviour.

After this PR, the above command works as intended by making the plan end time not able to exceed the execution time

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -41,6 +41,7 @@ from sqlmesh.utils.date import (
     to_datetime,
     yesterday_ds,
     to_timestamp,
+    time_like_to_str,
 )
 from sqlmesh.utils.errors import NoChangesPlanError, PlanError
 
@@ -55,6 +56,7 @@ class PlanBuilder:
         start: The start time to backfill data.
         end: The end time to backfill data.
         execution_time: The date/time time reference to use for execution time. Defaults to now.
+            If :start or :end are relative time expressions, they are interpreted as relative to the :execution_time
         apply: The callback to apply the plan.
         restate_models: A list of models for which the data should be restated for the time range
             specified in this plan. Note: models defined outside SQLMesh (external) won't be a part
@@ -172,6 +174,22 @@ class PlanBuilder:
         """Indicates whether this plan allows to set the start and end dates."""
         return self._is_dev or bool(self._restate_models)
 
+    @property
+    def start(self) -> t.Optional[TimeLike]:
+        if self._start and self._execution_time:
+            return to_datetime(self._start, relative_base=to_datetime(self._execution_time))
+        return self._start
+
+    @property
+    def end(self) -> t.Optional[TimeLike]:
+        if self._end and self._execution_time:
+            return to_datetime(self._end, relative_base=to_datetime(self._execution_time))
+        return self._end
+
+    @property
+    def execution_time(self) -> TimeLike:
+        return self._execution_time or now()
+
     def set_start(self, new_start: TimeLike) -> PlanBuilder:
         self._start = new_start
         self.override_start = True
@@ -256,7 +274,7 @@ class PlanBuilder:
         )
 
         restatements = self._build_restatements(
-            dag, earliest_interval_start(self._context_diff.snapshots.values())
+            dag, earliest_interval_start(self._context_diff.snapshots.values(), self.execution_time)
         )
         models_to_backfill = self._build_models_to_backfill(dag, restatements)
 
@@ -269,8 +287,8 @@ class PlanBuilder:
         plan = Plan(
             context_diff=self._context_diff,
             plan_id=self._plan_id,
-            provided_start=self._start,
-            provided_end=self._end,
+            provided_start=self.start,
+            provided_end=self.end,
             is_dev=self._is_dev,
             skip_backfill=self._skip_backfill,
             empty_backfill=self._empty_backfill,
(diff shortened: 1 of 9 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -172,6 +174,22 @@ def is_start_and_end_allowed(self) -> bool:
...
+
+    @property
+    def end(self) -> t.Optional[TimeLike]:
+        if self._end and self._execution_time:
+            return to_datetime(self._end, relative_base=to_datetime(self._execution_time))
+        return self._end
+
+    @property
+    def execution_time(self) -> TimeLike:
+        return self._execution_time or now()
```
izeigerman, line 203: Should we fall back to `now()` in constructor? So that we have it return the same value throughout the plan building process?

sqlmesh/core/plan/builder.py
```diff
@@ -289,7 +307,7 @@ def build(self) -> Plan:
             selected_models_to_backfill=self._backfill_models,
             models_to_backfill=models_to_backfill,
             effective_from=self._effective_from,
-            execution_time=self._execution_time,
+            execution_time=self.execution_time,
```
izeigerman: I wonder whether we should keep `None` if the `execution_time` wasn't specified explicitly. This way the `execution_time` will be determined at runtime, which is something I believe we want.

sqlmesh/core/plan/definition.py
```diff
@@ -80,7 +81,11 @@ def start(self) -> TimeLike:
 
     @cached_property
     def end(self) -> TimeLike:
-        return self.provided_end or now()
+        return self.provided_end or self.execution_time
+
+    @property
```
izeigerman: Should this be cached too?

sqlmesh/core/plan/builder.py
erindru: Yep, that seems to work. I've updated it

sqlmesh/core/plan/builder.py
erindru: It's been defaulted to `now()` in the PlanBuilder constructor, right?

In general, allowing it to be `None` is a bit annoying because every place that needs to use the value has to implement fallback behaviour if the value is not specified.

If the fallback is always "default to now()" then isn't it better to make this explicit up front?

sqlmesh/core/plan/definition.py
erindru: Yeah, I was wondering why the other one was cached and then I realised it was to keep a stable value since multiple calls to `now()` can result in different timestamps.

I've updated it

sqlmesh/core/plan/builder.py
erindru: I spoke to soon, it breaks a bunch of tests. Investigating...

sqlmesh/core/plan/builder.py
izeigerman: > It's been defaulted to now() in the PlanBuilder constructor, right?

sorry, the 2 suggestions were made independently. As far as I can tell we don't fall back to `now()` in constructor in main.

> In general, allowing it to be None is a bit annoying because every place that needs to use the value has to implement fallback behaviour if the value is not specified.

I think it's fine to have it as `@cached_property` in the builder and elsewhere. But I'd prefer the value to be correctly set at runtime since there can be a non-insignificant delay between plan construction and application.

erindru: Ok, so if I understand correctly, if it's sent to the `PlanBuilder` as `None` then that should still be `None` in the resulting `Plan` as well?

i.e as part of building the plan, don't default it to `now()`?

izeigerman: That's exactly what I mean, yes

sqlmesh/core/plan/builder.py
erindru: I've reverted defaulting to `now()` in the constructor and left a comment for next time, based on the conversation below

sqlmesh/core/plan/builder.py
erindru: Understood. I've removed the default in the constructor and use it like `self._execution_time or now()` when it is referenced in the `PlanBuilder` in places that might be dealing with relative time strings.

I stopped short of adding `@cached_property` for it because I thought it might be confusing having `self.execution_time` used within the PlanBuilder but `self._execution_time` being used when constructing the `Plan`

izeigerman: > I stopped short of adding @cached_property for it because I thought it might be confusing having self.execution_time used within the PlanBuilder but self._execution_time being used when constructing the Plan

Isn't this problematic too? Now we get inconsistent `execution_time` throughout the plan building process

erindru: Ok, i've reverted this back to my original design except with `@cached_property` instead of `@property` and added a comment explaining what's going on

PR #4703, opened 2025-06-10
Fix: skip project loading for clean, destroy and janitor commands

georgesittas:
`sqlmesh clean` failed, while debugging a dbt project in the wild, due to a model that wasn't able to be loaded. Verified that manually running `rm -r .cache` and re-loading the project fixed the issue.

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -23,12 +23,15 @@ from pathlib import Path
 logger = logging.getLogger(__name__)
 
 SKIP_LOAD_COMMANDS = (
+    "clean",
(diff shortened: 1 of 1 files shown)


sqlmesh/cli/main.py
```diff
@@ -23,12 +23,14 @@
 logger = logging.getLogger(__name__)
 
 SKIP_LOAD_COMMANDS = (
+    "clean",
```
themisvaltinos, line 26: we should I suppose include the janitor as well

georgesittas: done

PR #4708, opened 2025-06-10
feat(lsp): custom methods always return errors

benfdking:


sqlmesh/lsp/main.py
```diff
@@ -105,8 +113,128 @@ def _create_lsp_context(self, paths: t.List[Path]) -> t.Optional[LSPContext]:
...
+                self._ensure_context_in_folder(current_path)
+            if self.lsp_context is None:
+                raise RuntimeError("No context found")
+
+            # Call the format method on the context
+            self.lsp_context.context.format()
+            return FormatProjectResponse()
+        except Exception as e:
+            ls.log_trace(f"Error formatting project: {e}")
+            return FormatProjectResponse()
```
themisvaltinos, line 167: why do we return response on error instead of raising?

sqlmesh/lsp/api.py
```diff
@@ -28,7 +31,11 @@ class ApiRequest(PydanticModel):
     body: t.Optional[t.Dict[str, t.Any]] = None
 
 
-class ApiResponseGetModels(PydanticModel):
+class BaseAPIResponse(CustomMethodResponseBaseClass):
+    error: t.Optional[str] = None
```
themisvaltinos, line 35: why is error needed here and why is response_error not enough? I wen thought the code but can't understand its purpose of inheriting from CustomMethodResponseBaseClass

themisvaltinos (approved):

PR #4715, opened 2025-06-11
Chore!: Refactor `sqlmesh test` output to use rich

VaggelisD:
**Before this PR**:

- Default:
before

<br />

- Verbose
before_vv

<br />
<br />

**After this PR**
- Default
after

- Verbose (columns are split pairwise)
after_vv

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -7,6 +7,7 @@ import unittest
 import uuid
 import logging
 import textwrap
+from itertools import zip_longest
 from pathlib import Path
 from hyperscript import h
 from rich.console import Console as RichConsole
@@ -26,6 +27,7 @@ from rich.table import Table
 from rich.tree import Tree
 from sqlglot import exp
 
+from sqlmesh.core.test.result import ModelTextTestResult
 from sqlmesh.core.environment import EnvironmentNamingInfo, EnvironmentSummary
 from sqlmesh.core.linter.rule import RuleViolation
 from sqlmesh.core.model import Model
@@ -46,6 +48,7 @@ from sqlmesh.utils.errors import (
     NodeAuditsErrors,
     format_destructive_change_msg,
 )
+from sqlmesh.utils.rich import strip_ansi_codes
 
 if t.TYPE_CHECKING:
     import ipywidgets as widgets
@@ -316,6 +319,12 @@ class PlanBuilderConsole(BaseConsole, abc.ABC):
         """Display a destructive change error or warning to the user."""
 
 
+class UnitTestConsole(abc.ABC):
+    @abc.abstractmethod
+    def log_test_results(self, result: ModelTextTestResult, target_dialect: str) -> None:
+        """Display the test result and output."""
+
+
 class Console(
     PlanBuilderConsole,
     LinterConsole,
@@ -327,6 +336,7 @@ class Console(
     DifferenceConsole,
     TableDiffConsole,
     BaseConsole,
+    UnitTestConsole,
     abc.ABC,
 ):
     """Abstract base class for defining classes used for displaying information to the user and also interact
@@ -461,9 +471,7 @@ class Console(
         """
 
     @abc.abstractmethod
-    def log_test_results(
-        self, result: unittest.result.TestResult, output: t.Optional[str], target_dialect: str
-    ) -> None:
+    def log_test_results(self, result: ModelTextTestResult, target_dialect: str) -> None:
         """Display the test result and output.
 
         Args:
@@ -495,6 +503,10 @@ class Console(
     @abc.abstractmethod
     def loading_stop(self, id: uuid.UUID) -> None:
         """Stop loading for the given id."""
+
+    @abc.abstractmethod
+    def log_unit_test_results(self, result: ModelTextTestResult) -> None:
+        """Print the unit test results."""
 
 
 class NoopConsole(Console):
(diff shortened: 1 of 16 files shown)


sqlmesh/core/console.py
```diff
@@ -461,9 +471,7 @@ def plan(
         """
 
     @abc.abstractmethod
-    def log_test_results(
-        self, result: unittest.result.TestResult, output: t.Optional[str], target_dialect: str
-    ) -> None:
+    def log_test_results(self, result: ModelTextTestResult, target_dialect: str) -> None:
         """Display the test result and output.
```
georgesittas: This is unnecessary now with the interface being inherited.

sqlmesh/core/console.py
```diff
@@ -496,6 +504,10 @@ def loading_start(self, message: t.Optional[str] = None) -> uuid.UUID:
     def loading_stop(self, id: uuid.UUID) -> None:
         """Stop loading for the given id."""
 
+    @abc.abstractmethod
+    def log_unit_test_results(self, result: ModelTextTestResult) -> None:
+        """Print the unit test results."""
+
```
georgesittas: Why do we have two different methods for this?

sqlmesh/core/context.py
```diff
@@ -2044,6 +2043,7 @@ def test(
         verbosity: Verbosity = Verbosity.DEFAULT,
         preserve_fixtures: bool = False,
         stream: t.Optional[t.TextIO] = None,
+        log_results: bool = True,
```
georgesittas: Not sure about adding this to the public context method. Can you elaborate on why this was necessary?

sqlmesh/core/test/result.py
```diff
@@ -41,6 +44,12 @@ def addSubTest(
 
         super().addSubTest(test, subtest, err)
 
+    def _print_char(self, char: str) -> None:
+        from sqlmesh.core.console import TerminalConsole
+
+        if self.console and isinstance(self.console, TerminalConsole):
+            self.console._print(char, end="")
```
georgesittas:
```suggestion
        if isinstance(self.console, TerminalConsole):
            self.console._print(char, end="")
```

sqlmesh/core/test/result.py
```diff
@@ -51,7 +60,18 @@ def addFailure(self, test: unittest.TestCase, err: ErrorType) -> None:
...
         exctype, value, _ = err
+
+        if value and value.args:
+            exception_msg, rich_tables = value.args[:1], value.args[1:]
+            value.args = exception_msg
+
+            if rich_tables:
+                self.failure_tables.append(rich_tables)
+
+        self._print_char("F")
```
georgesittas, line 71: Can tests be skipped? Do we mimic unittest's output in that case?

sqlmesh/core/console.py
VaggelisD: A couple points:
- I wanted to separate the actual `ModelTestTextResult` printing from the (existing) `log_test_results`; The former prints the unittest relevant parts while the latter also adds more information such as the test summary
- On certain consoles like the NotebookConsole, we want to serialize the same report into a `str` and push it down to a component / send it over the network, so the distinction helps avoid recursion

sqlmesh/core/test/result.py
VaggelisD: I don't think we allow skipping, I can only think of using ignore patterns

sqlmesh/core/context.py
VaggelisD: I think this is because CI/CD calls `run_tests` but also generates it's own report in `update_test_check`, so we'd end up printing this twice in the logs, will verify this is the case again

sqlmesh/core/console.py
izeigerman: sorry, I still don't follow. Having 2 methods looks somewhat confusing to me

izeigerman:
Shouldn't a single `-v` be enough for the pairwise mode? Actually, I wonder whether we should just automatically switch to the pairwise mode if the number of columns exceeds certain threshold. Ideally, this should depend on the current width of the screen. Why even output something we know is probably unreadable?

sqlmesh/core/console.py
VaggelisD: Sorry, the way I have it is indeed confusing, but basically:
- `log_unit_test_results` is essentially a helper method; It only has to be defined once, and mostly outputs the unittest framework and the rich tables as text (`TerminalConsole`)
- `log_test_results` is the actual console method and reuses  `log_unit_test_results` differently depending on the console hierarchy, e.g for `NotebookConsole` it redirects that output into a display box. 

On my latest commit I'll make `log_unit_test_results` a private method and also remove it from being an `abc` method of `Console`, hopefully that'll be a bit better.

sqlmesh/core/console.py
```diff
@@ -2492,6 +2499,63 @@ def show_linter_violations(
         else:
             self.log_warning(msg)
 
+    def _log_unit_test_results(self, result: ModelTextTestResult) -> None:
```
izeigerman: Ok, this makes more sense. The only nitpick I have is that the `_log_unit_test_results` name is not that distinct from `_log_test_results`. Is there any way we can be more specific? Like `_log_test_details` or `_log_test_info` or `_log_test_summary`

izeigerman:
> That is a good idea, but do we think this might cause too much output and render the default version useless? Maybe a different idea would be to break the initial table apart every N columns, e.g:

I think I like that

sqlmesh/core/console.py
VaggelisD: Done

georgesittas (approved):
LGTM, though, I'd let Iaroslav and Sung take another look at this. Looks nice 👍

PR #4718, opened 2025-06-11
Feat(lsp): Add rename functionality for CTEs in vscode

themisvaltinos:
This update adds CTE renaming (F2) to the SQLMesh LSP server, letting users rename CTEs and their references with proper scope handling. Still to do:
- [x] Handle CTE with specific column reference or star [ Tackled here: #4763  ]
- [x] Add end to end test

### DIFF
--- a/vscode/extension/tests/rename_cte.spec.ts
+++ b/vscode/extension/tests/rename_cte.spec.ts
@@ -1,0 +1,187 @@
+import { test, expect } from '@playwright/test'
+import path from 'path'
+import fs from 'fs-extra'
+import os from 'os'
+import { startVSCode, SUSHI_SOURCE_PATH } from './utils'
+
+// Keyboard shortcuts
+const RENAME_KEY = 'F2'
+const FIND_ALL_REFERENCES_KEY =
+  process.platform === 'darwin' ? 'Alt+Shift+F12' : 'Ctrl+Shift+F12'
+
+test.describe('CTE Rename', () => {
+  let tempDir: string
+  let window: any
+  let close: () => Promise<void>
+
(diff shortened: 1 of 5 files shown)


benfdking:
```
WITH A AS (
  SELECT 1 AS B
)

SELECT A.B AS C FROM A WHERE A.B > 2
```

^ Do the two As in `A.B` get renamed.

themisvaltinos:
> ^ Do the two As in `A.B` get renamed.

thanks after #4763 now yes they do as they're picked up as references. added tests for it and end to end for rename functionality in general

vscode/extension/tests/rename_cte.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+// Keyboard shortcuts
+const RENAME_KEY = 'F2'
+const FIND_ALL_REFERENCES_KEY =
+  process.platform === 'darwin' ? 'Alt+Shift+F12' : 'Ctrl+Shift+F12'
+
+test.describe('CTE Rename', () => {
+  let tempDir: string
+  let window: any
+  let close: () => Promise<void>
```
benfdking: Although this is slightly cleaner on the surface, it has a nasty effect in implicitly tying tests together. 

Happy ho have functions with shared code, but I really don't think we should be tieing tests together. It means you can have side-effects and when it comes to running things in parallel we'll have to relook at it.

themisvaltinos: changed so that tests create their own isolated vscode instance and clean up, let me know if this is what you had in mind @benfdking

PR #4723, opened 2025-06-11
Fix: Import validation for the snowflake connection config

izeigerman:
I've noticed that the top-level `snowflake` import is sometimes not enough to accurately conclude whether the required library has been installed or not.

Additionally, rearranged the model validators so that the import validator always runs first.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -540,7 +540,6 @@ class SnowflakeConnectionConfig(ConnectionConfig):
     type_: t.Literal["snowflake"] = Field(alias="type", default="snowflake")
 
     _concurrent_tasks_validator = concurrent_tasks_validator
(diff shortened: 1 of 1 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -567,6 +566,10 @@ def _validate_authenticator(cls, data: t.Any) -> t.Any:
 
         return data
 
+    _engine_import_validator = _get_engine_import_validator(
+        "snowflake.connector.network", "snowflake"
+    )
```
georgesittas, line 571: FMI, how does moving this affect the execution / validation order? Asking based on this bit in the description:

> Additionally, rearranged the model validators so that the import validator always runs first.

georgesittas (approved):

sqlmesh/core/config/connection.py
izeigerman: We want this to run before `_validate_authenticator` which uses the imports

PR #4726, opened 2025-06-11
Feat: interactive sqlmesh init

treysp:
This PR adds an interactive `sqlmesh init` that allows users to choose a template, engine, and CLI mode from a list of options. A few other init changes are included as well.

Changes: 
- `init` now takes an engine name instead of a SQL dialect
- If a dbt template is specified, we check for the existence of a `dbt_project.yml` file and error if it's not present
- Linting rules `ambiguousorinvalidcolumn` and `invalidselectstarexpansion` automatically added to config
- New configuration macro function `user()` for injecting OS username into strings - must be called within curly braces `{{ user() }}`

The interactive init flow occurs when `sqlmesh init` is not passed an engine name (unless the `dbt` template option was passed and we infer the engine from the dbt project config).

The interactive init flow includes a new "flow" CLI mode, implemented via plan config args `no_diff`, `no_prompts`, and `auto_apply`.

Interactive input
Screenshot 2025-06-11 at 5 52 11 PM

Output (coloring from terminal, not SQLMesh)
Screenshot 2025-06-11 at 5 52 49 PM

Additional config.yaml content for FLOW CLI mode
```
# FLOW: Minimal prompts, automatic changes, summary output
# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#plan

plan:
  no_diff: true             # Hide detailed text differences for changed models
  no_prompts: true          # No interactive prompts
  auto_apply: true          # Apply changes automatically

# --- Optional: Set a default target environment ---
# This is intended for local development to prevent users from accidentally applying plans to the prod environment.
# It is a development only config and should NOT be committed to your git repo.
# https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#default-target-environment

# Uncomment the following line to use a default target environment derived from the logged in user's name.
# default_target_environment: dev_{{ user() }}

# Example usage:
# sqlmesh plan            # Automatically resolves to: sqlmesh plan dev_yourname
# sqlmesh plan prod       # Specify `prod` to apply changes to production
```


sqlmesh/cli/main.py
```diff
@@ -159,24 +163,80 @@ def cli(
...
         template=project_template,
+        engine_type=engine_type,
+        cli_mode=cli_mode or InitCliMode.DEFAULT,
         pipeline=dlt_pipeline,
         dlt_path=dlt_path,
     )
 
+    engine_install_text = ""
+    if engine_type and engine_type not in ("duckdb", "motherduck", "spark"):
+        engine_install_text = f'• Run command in CLI to install your SQL engine\'s Python dependencies: pip install "sqlmesh\\[{engine_type.replace("_", "")}]"\n'
```
georgesittas: Is it expected that there's no target for Spark? Won't folks need `pyspark` installed?

sqlmesh/cli/main.py
```diff
@@ -1219,3 +1279,115 @@ def state_import(obj: Context, input_file: Path, replace: bool, no_confirm: bool
     """Import a state export file back into the state database"""
     confirm = not no_confirm
     obj.import_state(input_file=input_file, clear=replace, confirm=confirm)
+
+
+def _interactive_init(
```
georgesittas: Nit: I feel like these should go to `example_project.py`, so that we only have CLI command entrypoints in this file.

sqlmesh/cli/main.py
treysp: Good catch - there's no sqlmesh extra, but pyspark isn't built in so they'd still need to install.

sqlmesh/cli/example_project.py
```diff
@@ -97,7 +120,28 @@ def _gen_config(
...
-    return default_configs[template]
+
+    simple_cli_mode = """
+# --- SIMPLE CLI MODE ---
+# Minimal prompts, automatic changes, summary output
+# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#plan
+
+plan:
+  no_diff: true             # Hide detailed text differences for changed models
+  use_finalized_state: true # Compare only against finalized snapshots
```
izeigerman: This shouldn't be here

sqlmesh/cli/example_project.py
```diff
@@ -2,30 +2,40 @@
 from enum import Enum
```
izeigerman: I think this module can be renamed to something more accurate. Like `project_init` or `init_wizard`

sqlmesh/cli/example_project.py
```diff
@@ -97,7 +120,28 @@ def _gen_config(
...
+  no_diff: true             # Hide detailed text differences for changed models
+  use_finalized_state: true # Compare only against finalized snapshots
+  no_prompts: true          # No interactive prompts
+  auto_apply: true          # Apply changes automatically
+
+# --- Optional: Set a default target environment ---
+# default_target_environment: dev_{{ env_var('USER', 'your_name') }}
+
+# Example usage:
+# export USER=your_name
```
izeigerman: Why do you need this export?

sqlmesh/cli/example_project.py
```diff
@@ -97,7 +120,28 @@ def _gen_config(
...
+# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/#plan
+
+plan:
+  no_diff: true             # Hide detailed text differences for changed models
+  use_finalized_state: true # Compare only against finalized snapshots
+  no_prompts: true          # No interactive prompts
+  auto_apply: true          # Apply changes automatically
+
+# --- Optional: Set a default target environment ---
+# default_target_environment: dev_{{ env_var('USER', 'your_name') }}
```
izeigerman: Should we just set this?

sqlmesh/cli/example_project.py
```diff
@@ -97,7 +120,28 @@ def _gen_config(
 
     default_configs[ProjectTemplate.EMPTY] = default_configs[ProjectTemplate.DEFAULT]
     default_configs[ProjectTemplate.DLT] = default_configs[ProjectTemplate.DEFAULT]
-    return default_configs[template]
+
+    simple_cli_mode = """
+# --- SIMPLE CLI MODE ---
```
izeigerman: Why do we need this header?

sqlmesh/core/config/connection.py
```diff
@@ -2067,6 +2116,9 @@ class RisingwaveConnectionConfig(ConnectionConfig):
     pre_ping: bool = True
 
     type_: t.Literal["risingwave"] = Field(alias="type", default="risingwave")
+    dialect: t.ClassVar[t.Literal["risingwave"]] = "risingwave"
+    display_name: t.ClassVar[t.Literal["RisingWave"]] = "RisingWave"
+    display_order: t.ClassVar[t.Literal[16]] = 16
```
izeigerman: [Nit] I suggest to make class variables all caps to follow the convention for class constants

sungchun12:
Feels great to use!

sqlmesh/cli/example_project.py
sungchun12: ah fair point, we could prompt the user to enter their name, but that may scare people off thinking we track user data. I'm open to something like the below.

```shell
For simple mode, enter your name to automatically apply changes to your personal dev environment. We do not track this information. 
```

From there, it can leave the example usage commented out so people have quick guidance.

```yaml
# --- Optional: Set a default target environment ---
default_target_environment: dev_{{ env_var('USER', 'user123') }}

# Example usage:
# sqlmesh plan                # Resolves to: sqlmesh plan dev_user123
# sqlmesh plan prod       # To apply changes to production
```

However, I am hesitant to push for this because this may have cascading effects on the docs, the github cicd bot defaults, and how people build a mental model for applying changes to prod: "When do I need to list out prod vs. not with `sqlmesh plan` given some users use default for simple mode?" 

What do you think? Am I overcomplicating the impact here?

cc @treysp as this is different from what I said in the internal slack thread.

sqlmesh/cli/example_project.py
sungchun12: see above

sqlmesh/cli/example_project.py
treysp: I think setting this for people will end up introducing confusion, so we should either leave it commented out or omit entirely. 

Maybe we could include a link to a page with other config changes they might want in the final CLI output? That way they're aware that other options exist but we're not pushing them?

If we want to leave it commented out in the config file, we should change the `export` line to an abbreviated explanation of how the env vars work and the implications of turning this on.

sungchun12: @treysp Alright, we have a shared hesitation. Let's do a mix of both your recommendations. Something like the below.

```yaml
# --- Optional: Set a default target environment ---
# # explain how this config should NOT be git committed as they may mess up the github bot and how others interact with the project when they first run it so they don't have to finagle changing this all the time if they're using DEFAULT mode for example.
# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/?h=config#environments
# default_target_environment: dev_{{ env_var('USER', 'your_name') }} 

# Example usage:
# export USER=your_name # explain how environment variables work, how to set them permanently vs. temporarily 
# sqlmesh plan            # Resolves to: sqlmesh plan dev_your_name
# sqlmesh plan prod       # To apply changes to production
```

izeigerman: Folks, in mac and linux this environment variable is set by the system. You don't need to set it manually. I'm sure something similar is also available in windows. This is why I asked the question.

izeigerman: See https://unix.stackexchange.com/questions/76354/who-sets-user-and-username-environment-variables

It appears that windows has `USERNAME` available but this needs to be verified.

Alternatively, we can add our own Jinja macro like `{{ user() }}` which returns `os.getlogin()`. And then it becomes:
```
dev_{{ user() }}
```

Up to you, but the env variable approach seems easier.

sungchun12: > Folks, in mac and linux this environment variable is set by the system. You don't need to set it manually. I'm sure something similar is also available in windows. This is why I asked the question.

I realized there was unintentional conflation with `USER` being the default env var in the yaml config. I forgot it's a reserved system env var which is better for us as I reflected on it. With that in mind, I recommend something like the below.

```yaml
# --- Optional: Set a default target environment ---
# explain how this config should NOT be git committed as they may mess up the github bot and how others interact with the project when they first run it so they don't have to finagle changing this all the time if they're using DEFAULT mode for example.
# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/?h=config#environments
# default_target_environment: dev_{{ env_var('USER') }} # this will inherit from your system environment variable with the same name
# for windows:
# default_target_environment: dev_{{ env_var('USERNAME') }} # this will inherit from your system environment variable with the same name


# Example usage:
# export USER=your_name # you can override the above default manually
# for windows:
# set USERNAME=your_name # you can override the above default manually
# sqlmesh plan            # Resolves to: sqlmesh plan dev_your_name
# sqlmesh plan prod       # To apply changes to production
```

treysp: Ahh, thank you, i was confused

Since this feature is complex enough to need explanation, should we just put all of this in the docs and link to it in the `init` output (rather than putting it here in the config file)?

sungchun12: This config is optimized for the `getting started` problem. I recommend both. Here's a cleaned up version. I want this to be in the user's face immediately because concerns of accidentally writing to prod comes up in every conversation I have with a user. We shouldn't give people homework to answer this basic question in FLOW mode. In the docs, we can explain that people can customize this with another env var to NOT use the system env var or if they want to do something like the below.

`default_target_environment: dev_{{ env_var('DEV_USER','prod') }}`

```yaml
# --- Optional: Set a default target environment ---
# This is intended for local development to prevent users from accidentally applying plans to the prod environment
# It is a development only config and should NOT be committed in your git repo
# https://sqlmesh.readthedocs.io/en/stable/reference/configuration/?h=config#environments

# Uncomment the following line to use a default target environment tied to your system environment variable USER/USERNAME automatically
# default_target_environment: dev_{{ user() }}

# Example usage:
# sqlmesh plan            # Resolves to: sqlmesh plan dev_yourname
# sqlmesh plan prod       # To apply changes to production
```

sqlmesh/cli/project_init.py
```diff
@@ -0,0 +1,504 @@
...
+
+config = sqlmesh_config(Path(__file__).parent)
+""",
+    }
+
+    default_configs[ProjectTemplate.EMPTY] = default_configs[ProjectTemplate.DEFAULT]
+    default_configs[ProjectTemplate.DLT] = default_configs[ProjectTemplate.DEFAULT]
+
+    flow_cli_mode = """
+# Minimal prompts, automatic changes, summary output
```
sungchun12:
```suggestion
# FLOW: Minimal prompts, automatic changes, summary output
```

sungchun12: Let's make it clear what choosing flow meant during `init`

docs/guides/configuration.md
```diff
@@ -151,6 +151,53 @@ The examples specify a Snowflake connection whose password is stored in an envir
...
 
+#### Default target environment
+
+The SQLMesh `plan` command acts on the `prod` environment by default (i.e., `sqlmesh plan` is equivalent to `sqlmesh plan prod`).
+
+In some organizations, users never run plans directly against `prod` - they do all SQLMesh work in a development environment unique to them. In a standard SQLMesh configuration, this means they need to include their development environment name every time they issue the `plan` command (e.g., `sqlmesh plan dev_tony`).
+
+If your organization works like this, it may be convenient to change the `plan` command's default environment from `prod` to each user's development environment. That way people can issue `sqlmesh plan` without typing the environment name every time.
+
+The SQLMesh configuration `user()` function returns the name of the user currently logged in and running SQLMesh. Call it inside Jinja curly braces to use default target environments based on the user name. The curly braces allow you to combine the user name with a prefix or suffix.
```
sungchun12: Explain how we retrieve the logged in user name. "We check your system environment variable `USER` for MacOS/Linux and `USERNAME` for Windows." Something like that.

sqlmesh/cli/project_init.py
```diff
@@ -0,0 +1,508 @@
...
+# sqlmesh plan prod       # Specify `prod` to apply changes to production
+"""
+
+    return default_configs[template] + (flow_cli_mode if cli_mode == InitCliMode.FLOW else "")
+
+
+@dataclass
+class ExampleObjects:
+    schema_name: str
+    full_model_name: str
```
izeigerman: I'd suggest to make it more generic without coupling this with a specific project structure:
```
models: t.Dict[str, str]
seeds: t.Dict[str, str]
...
```
with key being the name and the value being definition. For seeds the value could be tuple

PR #4729, opened 2025-06-12
refactor: expose macro evaluator in renderer

benfdking:
See https://github.com/TobikoData/sqlmesh/pull/4605 for context

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -68,6 +68,13 @@ class BaseExpressionRenderer:
         self._cache: t.List[t.Optional[exp.Expression]] = []
         self._model_fqn = model_fqn
         self._optimize_query_flag = optimize_query is not False
+        # Cache of the macro evaluator to be able to reuse it outside the render method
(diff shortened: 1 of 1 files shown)


sqlmesh/core/renderer.py
```diff
@@ -68,6 +68,13 @@ def __init__(
         self._cache: t.List[t.Optional[exp.Expression]] = []
         self._model_fqn = model_fqn
         self._optimize_query_flag = optimize_query is not False
+        # Cache of the macro evaluator to be able to reuse it outside the render method
+        self._macro_evaluator: t.Optional[MacroEvaluator] = None
+
+    @property
+    def macro_evaluator(self) -> t.Optional[MacroEvaluator]:
+        """Returns the cached macro evaluator from the last render operation."""
+        return self._macro_evaluator
```
izeigerman, line 77: This doesn't feel like a robust way of doing this. Can we instead have a method that constructs the MacroEvaluator instance? And then call it externally AND in `_render`?

PR #4732, opened 2025-06-13
Chore: Update the plan evaluator to use stages

izeigerman:
This update unifies plan evaluation and explanation by utilizing a common intermediate step: building the list of plan stages first.

The plan stages now contain a fully precomputed set of evaluation steps, removing dynamic behavior from the evaluation process. As a result, the following evaluation details are known ahead of time, before evaluation begins, and without producing any side effects:
- A set of snapshots for which physical tables will be checked and created if missing
- All missing intervals that will be backfilled
- A set of snapshots for which views will be created / deleted in the virtual layer
- A set of snapshots for which the schema will be migrated
- The order of operations

The only remaining dynamic step is the creation of physical tables, as it requires querying the data warehouse to determine which tables are missing. We want to avoid touch the data warehouse during the stage construction process.

Among the many benefits of the new approach is that plan evaluations and explanations are now always in sync, as they rely on the same underlying logic for deriving plan stages.

### DIFF
--- a/sqlmesh/core/state_sync/base.py
+++ b/sqlmesh/core/state_sync/base.py
@@ -109,6 +109,17 @@ class StateReader(abc.ABC):
         """
 
     @abc.abstractmethod
+    def refresh_snapshot_intervals(self, snapshots: t.Collection[Snapshot]) -> t.List[Snapshot]:
+        """Updates given snapshots with latest intervals from the state.
+
+        Args:
+            snapshots: The snapshots to refresh.
+
+        Returns:
+            The updated snapshots.
+        """
+
+    @abc.abstractmethod
     def nodes_exist(self, names: t.Iterable[str], exclude_external: bool = False) -> t.Set[str]:
         """Returns the node names that exist in the state sync.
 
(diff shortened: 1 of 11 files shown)


sqlmesh/core/state_sync/base.py
```diff
@@ -108,6 +108,17 @@ def snapshots_exist(self, snapshot_ids: t.Iterable[SnapshotIdLike]) -> t.Set[Sna
             A set of all the existing snapshot ids.
         """
 
+    @abc.abstractmethod
+    def refresh_snapshot_intervals(self, snapshots: t.Collection[Snapshot]) -> t.List[Snapshot]:
```
erindru, line 112: Haha nice, finally fixed this :)

tests/core/test_plan_evaluator.py
```diff
@@ -65,46 +64,18 @@ def test_builtin_evaluator_push(sushi_context: Context, make_snapshot):
...
         == 2
     )
     assert sushi_context.engine_adapter.table_exists(new_model_snapshot.table_name())
     assert sushi_context.engine_adapter.table_exists(new_view_model_snapshot.table_name())
-
-
-@pytest.mark.parametrize(
-    "change_category", [SnapshotChangeCategory.BREAKING, SnapshotChangeCategory.FORWARD_ONLY]
-)
-def test_update_intervals_for_new_snapshots(
```
erindru, line 81: Was this not needed anymore? I assume its related to the removal of this call:

```
snapshot.dev_intervals = snapshot.intervals.copy()
```

in `_update_intervals_for_new_snapshots`

erindru (approved):
Nice, this makes implicit behaviour much more explicit / easier to follow

tests/core/test_plan_evaluator.py
izeigerman: Yeah, that method is gone.

PR #4739, opened 2025-06-14
chore(vscode): better handling no sqlmesh

benfdking:


vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
```diff
@@ -419,3 +437,13 @@ export const sqlmeshLspExec = async (): Promise<
     })
   }
 }
+
+async function doesExecutableExist(executable: string): Promise<boolean> {
```
themisvaltinos, line 441:
```suggestion
async function doesExecutableExist(executable: string): Promise<boolean> {
  const command = process.platform === 'win32' ? 'where.exe' : 'which'
  try {
    const result = await execAsync(command, [executable])
    return result.exitCode === 0
  } catch {
    return false
  }
}
```

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
```diff
@@ -287,6 +287,12 @@ export const sqlmeshExec = async (): Promise<
       args: [],
     })
   } else {
+    const exists = await doesExecutableExist(sqlmesh)
+    if (!exists) {
+      return err({
```
themisvaltinos, line 292: maybe add tracing before return (and in sqlmesh lsp not found similarly) for logging the way that you have in other places: 

```ts
traceInfo('SQLMesh not found.')
```

themisvaltinos (approved):

vscode/extension/src/utilities/sqlmesh/sqlmesh.ts
benfdking: There's one in the error handlers itself, but I do think I want to add one to the checker.

PR #4740, opened 2025-06-15
chore: dispose correctly in lineage panel

benfdking:


vscode/extension/src/webviews/lineagePanel.ts
```diff
@@ -58,12 +63,12 @@ export class LineagePanel implements WebviewViewProvider, Disposable {
...
       // Allow scripts in the webview
       enableScripts: true,
-      localResourceRoots: [this._extensionUri],
+      localResourceRoots: [this.extensionUri],
     }
 
     // Set content options for external URL access
     // Set up message listener for events from the iframe
-    webviewView.webview.onDidReceiveMessage(
+    const disposeable = webviewView.webview.onDidReceiveMessage(
```
themisvaltinos: since there are multiple async operations inside the switches, thinking if something goes wrong in one of the await calls if it fails silently or unpredictably and if it makes sense to wrap these in a try-catch block

vscode/extension/src/webviews/lineagePanel.ts
```diff
@@ -16,37 +16,42 @@ export class LineagePanel implements WebviewViewProvider, Disposable {
 
   private panel: WebviewView | undefined
   private lsp: LSPClient
-  private _extensionUri: Uri
+  private readonly extensionUri: Uri
+
+  private readonly disposeables: Disposable[] = []
```
themisvaltinos: typo: `disposables`

Issue #3251, opened 2024-10-14
Support environment specific catalogs


mpcarter:
Add a new option for `environment_suffix_target: catalog`. 

The desired behavior is that non-prod virtual environments would dynamically create new catalogs/databases. For example, `sqlmesh plan dev` results in:

```
mydb__dev.sushi.customers
```

The rationale is that this makes development with downstream consumers easier (e.g. reporting tools). Many tools will reference objects by `schema.object`. So, currently, if I want to test how a report or dataset behaves before promoting a model change to prod, I need to temporarily change the reference name to the schema or object inside the tool. If only the catalog/database name changed, then it is just a matter of temporarily updating the connection (which can usually be an environment variable) to use a different catalog.

Database engine limitations would mean only a subset of them could support this option. Primarily Snowflake, but also MotherDuck, Redshift, MSSQL, possibly Databricks.

grantith:
Any thoughts on whether this is reasonably incorporated? I'm hankering to deploy sqlmesh as our long-term dbt alternative but without dynamic catalogs instead of schemas it makes the suite of tools and automations relying on our models unwieldy to configure.

If anyone familiar with the code-base thinks this would be remotely accessible to a first time contributor I'm happy to start digging in and see what I can come up with.

dcohen24:
Given that :

**environment_catalog_mapping** 
A mapping from regular expressions to catalog names. 
The catalog name is used to determine the target catalog for a given environment. | dict[string, string] | N


There is already a function to disable environment specific schemas (but it flips the env to change the table name)
https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#physical-table-schemas

isnt what we are asking effectively:
To add to the   environment_suffix_target config a `none` option

so we have `table`,`schema` (and add `none`)
none would in theory only be valid if environment_catalog_mapping is set?

sungchun12:
> Given that :
> 
> **environment_catalog_mapping** A mapping from regular expressions to catalog names. The catalog name is used to determine the target catalog for a given environment. | dict[string, string] | N
> 
> There is already a function to disable environment specific schemas (but it flips the env to change the table name) https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#physical-table-schemas
> 
> isnt what we are asking effectively: To add to the environment_suffix_target config a `none` option
> 
> so we have `table`,`schema` (and add `none`) none would in theory only be valid if environment_catalog_mapping is set?

[...]

PR #4742, opened 2025-06-16
Feat: Allow virtual environments to be given dedicated catalogs

erindru:
Addresses #3251 

Up until now, there has been no way to say to SQLMesh "Create a virtual environment with identical schema and view naming to `prod`, just under a different catalog".

The closest thing was `environment_catalog_mapping` which technically did allow virtual environment views to go into a different catalog but it did nothing to rename the schemas.

This meant if you had something like:

```
environment_catalog_mapping:
   '^prod$': 'prod',
  '^dev$': 'dev'
```

And ran `sqlmesh plan dev`, the schemas would still have the `__dev` suffix, eg `dev.example_schema__dev.example_table` - even though they are created under the `dev` catalog.

This behaviour meant that it is not trivial to point a downstream report written against the prod environment at your dev environment because the objects are still named differently.

This PR extends the existing `environment_suffix_target` option to give it another value - `catalog`. After this PR, setting some config like:
```
environment_suffix_target: catalog
```

And running `sqlmesh plan dev` will cause the virtual layer to be created under eg `dev.example_schema.example_table` instead of `dev.example_schema__dev.example_table`.

In this initial implementation, SQLMesh will not create or drop catalogs for you. Catalogs are a lot less trivial to create than schemas because they tend to have extra options like "where to store the data files", "what tablespace to use" etc which SQLMesh doesnt currently have a good way to specify.

So the assumption is that any environment-specific catalogs have already been created in the target database and SQLMesh is just utilizing them. On the plus side, it means the SQLMesh user does not need any extra permissions to create catalogs


sungchun12:
@dcohen24 we're working on this as promised!

dcohen24:
Great.  Will need your help to think about this one :
" SQLMesh will not attempt to create catalogs on demand or drop them as part of janitor cleanup. Using `environment_suffix_target "

This will work great for staging (and dev)... need to think about how we might handle individual user / feature branches (that wouldnt have a pre-built catalog)

sqlmesh/core/config/root.py
```diff
@@ -219,6 +219,23 @@ def _normalize_and_validate_fields(cls, data: t.Any) -> t.Any:
...
+            if (
+                environment_suffix_target.lower() == EnvironmentSuffixTarget.CATALOG.lower()
+                and "environment_catalog_mapping" not in data
+            ):
+                # set the default environment_catalog_mapping for when environment_suffix_target=catalog
+                # but no explicit mapping of environments to catalogs is set by the user
+                data["environment_catalog_mapping"] = {
+                    # no override for prod, use the default catalog on the connection
+                    # note that we cant pass None here without failing the Pydantic validator, so we pass a string constant that can be interpreted as None later
+                    f"^{c.PROD}$": c.SQLMESH_NONE,
```
izeigerman: I don't believe `environment_catalog_mapping` applies to the prod environment. Isn't this redundant?

sqlmesh/core/config/root.py
```diff
@@ -219,6 +219,23 @@ def _normalize_and_validate_fields(cls, data: t.Any) -> t.Any:
...
             }
 
+        if (environment_suffix_target := data.get("environment_suffix_target")) and isinstance(
+            environment_suffix_target, str
+        ):
+            if (
+                environment_suffix_target.lower() == EnvironmentSuffixTarget.CATALOG.lower()
+                and "environment_catalog_mapping" not in data
+            ):
+                # set the default environment_catalog_mapping for when environment_suffix_target=catalog
```
izeigerman: Shouldn't we fail if `environment_catalog_mapping` is provided together with the `catalog` suffix target?  IMHO the 2 seem to be mutually exclusive

sqlmesh/core/environment.py
```diff
@@ -89,8 +93,13 @@ def from_environment_catalog_mapping(
         construction_kwargs = dict(name=name, **kwargs)
         for re_pattern, catalog_name in environment_catalog_mapping.items():
             if re.match(re_pattern, name):
+                if catalog_name == c.SQLMESH_NONE:
```
izeigerman: I'm not sure this is needed

sqlmesh/core/config/root.py
```diff
@@ -219,6 +219,23 @@ def _normalize_and_validate_fields(cls, data: t.Any) -> t.Any:
...
+                # but no explicit mapping of environments to catalogs is set by the user
+                data["environment_catalog_mapping"] = {
```
izeigerman: Overriding this value in the config itself seems a bit implicit. Wouldn't it be better to make this logic explicit in the `Context.environment_catalog_mapping`

docs/guides/configuration.md
```diff
@@ -271,20 +273,68 @@ Config example:
...
+- For the `prod` environment, the default catalog as configured in the gateway will be used
+- For any other environment, eg `dev`, the environment name will be used as the catalog
+
+Therefore, a model named `my_schema.users` in an environment called `dev` will have its view created as `dev.my_schema.users`.
+
+If you would like more control over the catalog name to use for a given environment, you may optionally configure [environment_catalog_mapping](#environment-view-catalogs) as described below.
+
+!!! warning "Caveats"
+    - Using `environment_suffix_target: catalog` only works on engines that support querying across different catalogs. If your engine does not support cross-catalog queries then you will need to use `environment_suffix_target: schema` or `environment_suffix_target: table` instead.
+    - SQLMesh will not attempt to create catalogs on demand or drop them as part of janitor cleanup. Using `environment_suffix_target: catalog` assumes the catalogs already exist in the target database and are being managed outside of SQLMesh.
```
dcohen24: So: SQLMesh will not do this.  Is there a fallback?  what will happen when a feature branch/catalog is attempted to be spun up?

sqlmesh/core/config/root.py
erindru: It does apply. The documentation mentions it explicitly and the existing tests in `test_environment.test_from_environment_catalog_mapping` confirm it

sqlmesh/core/config/root.py
erindru: They aren't, they work together to give users control over the catalog names.

By default, if you dont configure an `environment_catalog_mapping`, you get the default mapping of "catalog is named after the environment 1:1".

But if you want more control over how the catalog names are generated, you can specify a custom `environment_catalog_mapping`.

Therefore, `environment_suffix_target: catalog` is implemented _in terms_ of `environment_catalog_mapping`

sqlmesh/core/environment.py
erindru: It's needed because prod always uses the default catalog, not a catalog called `prod`.

However, by the time we are working out what to call something within `QualifiedViewName`, the `catalog` property has already been normalized and set to the default catalog.

Since we dont want to override it, we make `catalog_name_override` = `None` for just the `prod` environment, and only for our default mapping so if the user specifies a custom mapping for `prod` it can still be overridden

sqlmesh/core/config/root.py
erindru: Oh, yes, I missed that was a property. I will move this

docs/guides/configuration.md
erindru: The difficulty right now is that `CREATE CATALOG` / `CREATE DATABASE` generally comes with a bunch of options to customize the catalog and SQLMesh does not have a good way of specifying them at the moment

However, we might be able to support a basic case with engines like Snowflake that don't make this _as_ difficult, let me revisit this

dcohen24: that would be awesome [selfishly...we will be snowflake]... I suppose it also be like a custom materialization /abstract class.  Push it on user to do build up /teardown

sqlmesh/core/environment.py
erindru: I've updated this so you can specify a Python `None` which means "no mapping, fall back to the default" - rather than a string constant meaning the same thing.

It did require type changes in a few places but I think the overall result is less hacky

sqlmesh/core/config/root.py
erindru: I've moved this logic

docs/guides/configuration.md
erindru: I've created an initial implementation for Snowflake where SQLMesh will run `CREATE DATABASE IF NOT EXISTS <env_name>` to create a catalog and `DROP DATABASE IF EXISTS <env_name>` to clean up when the env expires.

Doing this automatically makes me slightly nervous because SQLMesh cannot distinguish between catalogs it created and catalogs that others created. So someone could map a SQLMesh virtual environment to an existing catalog with other data in it and the Janitor will happily drop that catalog when the SQLMesh virtual environment expires, which will also drop the other data.

Do you see that being a problem in your use-case?

sqlmesh/core/environment.py
```diff
@@ -82,17 +89,22 @@ def sanitize_names(cls, values: t.Iterable[str]) -> t.Set[str]:
...
         name: str = c.PROD,
         **kwargs: t.Any,
     ) -> T:
         construction_kwargs = dict(name=name, **kwargs)
-        for re_pattern, catalog_name in environment_catalog_mapping.items():
+
+        for re_pattern, catalog_name_override in environment_catalog_mapping.items():
             if re.match(re_pattern, name):
+                if catalog_name_override and re_pattern.groups:
+                    catalog_name_override = re.sub(re_pattern, catalog_name_override, name, count=1)
```
erindru: I switched this to use regex capture groups instead of an `@{environment_name}` string substitution for maximum flexibility

docs/guides/configuration.md
erindru: Actually, thinking about this more, i've adjusted SQLMesh to set `COMMENT = 'sqlmesh_managed'` on the databases it creates.

It will then only drop databases with this comment set. That should help prevent accidents

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1217,6 +1218,18 @@ def drop_view(
             **kwargs,
         )
 
+    def create_catalog(self, catalog_name: str | exp.Identifier) -> None:
+        return self._create_catalog(exp.parse_identifier(catalog_name, dialect=self.dialect))
+
+    def _create_catalog(self, catalog_name: exp.Identifier) -> None:
+        raise NotImplementedError()
```
izeigerman: We need a more reasonable error here indicating what happened and for which engine

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1217,6 +1218,18 @@ def drop_view(
...
+
+    def drop_catalog(self, catalog_name: str | exp.Identifier) -> None:
+        return self._drop_catalog(exp.parse_identifier(catalog_name, dialect=self.dialect))
+
+    def _drop_catalog(self, catalog_name: exp.Identifier) -> None:
+        raise NotImplementedError()
```
izeigerman: ditto

sqlmesh/core/context.py
```diff
@@ -2575,21 +2576,48 @@ def test_connection_config(self) -> ConnectionConfig:
...
+            # - environment_suffix_target=catalog
+            # - no explicit mapping of environments to catalogs is set by the user
+            # this allows the feature to work without making the user configure a default mapping explicitly
+
+            default_catalog = self.default_catalog
+
+            return compile_regex_mapping(
+                {
+                    # no override for prod, use the default catalog
+                    f"^{c.PROD}$": None,
```
izeigerman: would it be easy to define this in yaml?

sqlmesh/core/config/root.py
erindru: After internal discussion, we don't want to give users the ability to override how the catalog names are generated because it's too easy to get subtly wrong.

Based on that, these two settings are indeed mutually exclusive, i've adjusted the implementation to throw a `ConfigError` if both are specified

sqlmesh/core/environment.py
erindru: This has now been removed entirely since this PR no longer touches `environment_catalog_mapping`

sqlmesh/core/config/root.py
erindru: This has now been removed entirely since this PR no longer touches `environment_catalog_mapping`

sqlmesh/core/environment.py
erindru: This has now been removed entirely since this PR no longer touches environment_catalog_mapping

sqlmesh/core/engine_adapter/base.py
erindru: This codepath doesnt actually get hit because calls to `create_catalog` and `drop_catalog` are gated behind a check on `adapter.SUPPORTS_CREATE_DROP_CATALOG` and it's implemented for the two adapters where that returns `True` (duckdb and snowflake)

But i've updated the message anyway

sqlmesh/core/context.py
erindru: It's a moot point now since I rolled back capture groups and the rest of the changes to `environment_catalog_mapping`

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1217,6 +1218,22 @@ def drop_view(
...
+    def _create_catalog(self, catalog_name: exp.Identifier) -> None:
+        raise NotImplementedError(
+            f"Unable to create catalog '{catalog_name.sql(dialect=self.dialect)}' as automatic catalog management is not implemented in the {self.dialect} engine."
+        )
+
+    def drop_catalog(self, catalog_name: str | exp.Identifier) -> None:
+        return self._drop_catalog(exp.parse_identifier(catalog_name, dialect=self.dialect))
+
+    def _drop_catalog(self, catalog_name: exp.Identifier) -> None:
+        raise NotImplementedError(
```
izeigerman: I think this should be a SQLMeshError in order to bubble up to the user correctly.

izeigerman (approved):

PR #4751, opened 2025-06-16
feat: Add support for Microsoft Fabric Warehouse

fresioAS:
# Add Microsoft Fabric Engine Support

## Overview

This PR adds support for Microsoft Fabric as a new execution engine in SQLMesh. Users can now connect to and execute queries on Microsoft Fabric Data Warehouses.

## Changes

*   **Documentation:**
    *   Added `docs/integrations/engines/fabric.md` with Fabric connection options, installation, and configuration instructions.
    *   Listed Fabric in `docs/guides/configuration.md` and `docs/integrations/overview.md`.
    *   Updated `mkdocs.yml` to include the new Fabric documentation page.

*   **Core Implementation:**
    *   Added `FabricConnectionConfig`, inheriting from `MSSQLConnectionConfig`, with Fabric-specific defaults and validation.
    *   Registered the new Fabric engine adapter (`FabricAdapter`) in the registry.
    *   Added `sqlmesh/core/engine_adapter/fabric.py` with Fabric-specific logic, including the use of `DELETE`/`INSERT` for overwrite operations.

*   **Testing:**
    *   Added `tests/core/engine_adapter/test_fabric.py` for adapter logic, table checks, insert/overwrite, and replace query tests.
    *   Updated `tests/core/test_connection_config.py` for config validation and ODBC connection string generation.

*   **Configuration:**
    *   Updated `pyproject.toml` to add a `fabric` test marker.
    *   Registered Fabric in all relevant config and adapter modules.


fresioAS:
Thanks. I will investigate later - perhaps we need `schema = self._get_schema_name(table_name)` in `create_schema()` function?
Feel free to contribute if you find a fix. I have not battle tested this outside of SQLMesh init, just wanted to get the code out there for draft first!

mattiasthalen:
@georgesittas, would you say that most of these changes would be more suitable in sqlglot, a fabric-tsql dialect, if you will.

Seeing as there are more differences between tsql and the version in fabric.

mattiasthalen:
@erindru, I don't think there should be any separation between warehouse and Lakehouse. Both use the same type of sql endpoint, the "fabric flavored t-sql".

The only difference I can think of is wether or not the Lakehouse supports schemas. As of now, you get the option to activate schemas when creating a Lakehouse. And that comes with its own issues, e.g., a weaker API.

This might merit a parameter to tell if the catalog/database is a Lakehouse with/without schema, or a warehouse. But I agree that a different engine is overkill.

With that said, the current code in this PR can actually query a Lakehouse. The host/endpoint used is the same for LH & WH, and you specify which one by the catalog/database.

Same thing happens with the sql database object, they share host/endpoint and you select the object by setting the database.

erindru:
In that case, a coherent `fabric` implementation that works transparently across both would be even better.

MS has probably improved this since I last looked, but isn't Lakehouse based on Spark SQL and Warehouse based on the "Polaris" flavour of TSQL?

mattiasthalen:
Well, yeah. Spark SQL is used in a Lakehouse to create tables. But you can use the SQL endpoint to query it, and I think you can create views with it. The warehouse can use both tsql and spark.

fresioAS:
The latest commit including the dialect found with this sqlglot [fork](https://github.com/fresioAS/sqlglot/tree/fabric_dialect) allows me to reference lakehouse external data. Now there is most likely some overlaps between the engine and the dialect now, and also there is a good amount of generated code that is probably irrelevant. 

Try it out @mattiasthalen and check if we get a bit closer towards the goal

mattiasthalen:
You're fast! I haven't even fired up a codespace for sqlglot yet. 

Did a quick test, but all I got was that there is no fabric dialect. Not sure if the error comes from sqlmesh or sqlglot. ☺️

mattiasthalen:
@georgesittas / @erindru, what's needed for the config to be available for Python configs?

sqlmesh/core/config/connection.py
```diff
@@ -43,7 +43,14 @@
...
 logger = logging.getLogger(__name__)
 
-RECOMMENDED_STATE_SYNC_ENGINES = {"postgres", "gcp_postgres", "mysql", "mssql", "azuresql"}
+RECOMMENDED_STATE_SYNC_ENGINES = {
+    "postgres",
+    "gcp_postgres",
+    "mysql",
+    "mssql",
+    "azuresql",
+    "fabric",
```
mattiasthalen: I don't think fabric should be a recommended state engine. what you would want to do is setup a SQL Database object and connect to it via the mssql engine.

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -26,334 +17,35 @@ class FabricAdapter(MSSQLEngineAdapter):
...
-        if not self.table_exists(table_name):
-            self.execute(create_table_sql)
-
-        if table_description and self.comments_enabled:
-            qualified_table_for_comment = self._fully_qualify(table_name)
-            self._create_table_comment(qualified_table_for_comment, table_description)
-            if column_descriptions and self.comments_enabled:
-                self._create_column_comments(qualified_table_for_comment, column_descriptions)
-
     def table_exists(self, table_name: TableName) -> bool:
```
mattiasthalen: Is `table_exists()` needed now that sqlglot uses the upper case information schema in tsql (inherited by fabric)?

fresioAS: The mssql adapter itself is querying in lowercase for `table_exists()` - the alternative would be to uppercase this part in the mssql adapter?

sqlmesh/core/config/connection.py
fresioAS: No strong opinion on this, but I don't see why it could not work as a state engine?

mattiasthalen: I think that list is for the state engines they recommend, that doesn't mean it's not supported if it's not listed there. :) `FORBIDDEN_STATE_SYNC_ENGINES` on the other hand, that would prohibit it.

E.g., DuckDB works for state, but it's not a recommended state engine. I think they only recommend OLTP dbs.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Ah, true that.

@georgesittas, should we do the same as in sqlglot and update the mssql to use upper case?

mattiasthalen:
> > I got it all working here: https://github.com/mattiasthalen/sqlmesh/tree/fabric Basically the same as yours, but I moved the information_schema in upper case to MSSQL.
> > It requires the SQLGlot from main, hopefully that will be released soon.
> 
> Makes the codebase simpler - probably a standalone PR to change the mssql adapter

Submitted #4795

sqlmesh/core/engine_adapter/fabric.py
georgesittas: Seems reasonable 👍

docs/integrations/engines/fabric.md
mattiasthalen, line 1: Might want to add a section on using the SQL Database object and MSSQL as the preferred state engine in Fabric. That would give us OLTP for state.

mattiasthalen (approved):
Looks great!
I have a pending [PR in SQLGlot](https://github.com/tobymao/sqlglot/pull/5280) that will sort the datetime2 errors

docs/integrations/engines/fabric.md
fresioAS: added a note similar to what was already present in Trino/Spark docs

georgesittas:
Hey folks 👋 

Just a heads up that I've released the latest Fabric fixes / additions in SQLGlot and there's a [PR](https://github.com/TobikoData/sqlmesh/pull/4831) up here to bump its version. Keep an eye out for when it's merged so you can rebase and run the CI with the latest state of the other repo.

sqlmesh/core/config/connection.py
```diff
@@ -1635,6 +1641,34 @@ def _extra_engine_config(self) -> t.Dict[str, t.Any]:
...
+    """
+    Fabric Connection Configuration.
+    Inherits most settings from MSSQLConnectionConfig and sets the type to 'fabric'.
+    It is recommended to use the 'pyodbc' driver for Fabric.
+    """
+
+    type_: t.Literal["fabric"] = Field(alias="type", default="fabric")  # type: ignore
+    DIALECT: t.ClassVar[t.Literal["fabric"]] = "fabric"  # type: ignore
+    DISPLAY_NAME: t.ClassVar[t.Literal["Fabric"]] = "Fabric"  # type: ignore
+    DISPLAY_ORDER: t.ClassVar[t.Literal[17]] = 17  # type: ignore
```
georgesittas, line 1703: Just FMI: why are these `# type: ignore` comments needed?

fresioAS: It did not pass the check since this subclass intentionally redefines the attributes from its parent - MSSQLConnectionConfig. Let me know if you have a better way of redefining this.

mattiasthalen:
Found a pesky bug, it tries to execute `MERGE` when running `sqlmesh run`:
```bash
"das"."raw"."raw__northwind__territories"

ProgrammingError:
      ('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Incorrect syntax
near 'MERGE'. (102) (SQLExecDirectW)")
```

The model config is:
```python
@model(
    "das.raw.@{name}",
    is_sql=True,
    kind=dict(
        name=ModelKindName.INCREMENTAL_BY_UNIQUE_KEY,
        unique_key="_record__hash",
        disable_restatement=True
    ),
    blueprints=generate_blueprints("northwind"),
)
```

```log
2025-07-01 11:57:15,717 - MainThread - sqlmesh.core.state_sync.db.interval - INFO - Compacting 0 intervals for 0 snapshots (interval.py:129)
2025-07-01 11:57:15,814 - MainThread - sqlmesh.core.state_sync.db.interval - INFO - Compacting 0 intervals for 0 snapshots (interval.py:129)
2025-07-01 11:57:18,414 - ThreadPoolExecutor-1_0 - sqlmesh.core.snapshot.evaluator - INFO - Evaluating snapshot SnapshotId<"das"."raw"."raw__northwind__territories": 3192609146> (evaluator.py:636)
2025-07-01 11:57:18,415 - ThreadPoolExecutor-1_0 - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: SELECT DB_NAME(); (base.py:2201)
2025-07-01 11:57:19,571 - ThreadPoolExecutor-1_0 - sqlmesh.core.snapshot.evaluator - INFO - Inserting batch (2025-06-27 13:00:00, 2025-07-01 11:55:00) into das.sqlmesh__raw.raw__raw__northwind__territories__737428378' (evaluator.py:690)
2025-07-01 11:57:19,571 - ThreadPoolExecutor-1_0 - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: SELECT DB_NAME(); (base.py:2201)
2025-07-01 11:57:19,608 - ThreadPoolExecutor-1_0 - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: MERGE INTO [sqlmesh__raw].[raw__raw__northwind__territories__737428378] AS [__MERGE_TARGET__] USING (SELECT CAST([territory_id] AS VARCHAR(MAX)) AS [territory_id], CAST([territory_description] AS VARCHAR(MAX)) AS [territory_description], CAST([region_id] AS BIGINT) AS [region_id], CAST([_dlt_load_id] AS VARCHAR(MAX)) AS [_dlt_load_id], CAST([_dlt_id] AS VARCHAR(MAX)) AS [_dlt_id], HASHBYTES('SHA2_256', CONCAT(COALESCE(CAST([territory_id] AS VARCHAR(MAX)), '_sqlmesh_surrogate_key_null_'), '|', COALESCE(CAST([territory_description] AS VARCHAR(MAX)), '_sqlmesh_surrogate_key_null_'), '|', COALESCE(CAST([region_id] AS VARCHAR(MAX)), '_sqlmesh_surrogate_key_null_'))) AS [_record__hash], CAST(DATEADD([MICROSECOND], CAST(ROUND(CAST([_dlt_load_id] AS FLOAT) * 1e6, 0) AS BIGINT), CAST('1970-01-01' AS DATETIME2(6))) AS DATETIME2(6)) AS [_record__loaded_at] FROM [landing_zone].[northwind].[raw__northwind__territories] AS [raw__northwind__territories]) AS [__MERGE_SOURCE__] ON [__MERGE_TARGET__].[_record__hash] = [__MERGE_SOURCE__].[_record__hash] WHEN MATCHED AND EXISTS(SELECT [__MERGE_TARGET__].[territory_id], [__MERGE_TARGET__].[territory_description], [__MERGE_TARGET__].[region_id], [__MERGE_TARGET__].[_dlt_load_id], [__MERGE_TARGET__].[_dlt_id], [__MERGE_TARGET__].[_record__loaded_at] EXCEPT SELECT [__MERGE_SOURCE__].[territory_id], [__MERGE_SOURCE__].[territory_description], [__MERGE_SOURCE__].[region_id], [__MERGE_SOURCE__].[_dlt_load_id], [__MERGE_SOURCE__].[_dlt_id], [__MERGE_SOURCE__].[_record__loaded_at]) THEN UPDATE SET [__MERGE_TARGET__].[territory_id] = [__MERGE_SOURCE__].[territory_id], [__MERGE_TARGET__].[territory_description] = [__MERGE_SOURCE__].[territory_description], [__MERGE_TARGET__].[region_id] = [__MERGE_SOURCE__].[region_id], [__MERGE_TARGET__].[_dlt_load_id] = [__MERGE_SOURCE__].[_dlt_load_id], [__MERGE_TARGET__].[_dlt_id] = [__MERGE_SOURCE__].[_dlt_id], [__MERGE_TARGET__].[_record__loaded_at] = [__MERGE_SOURCE__].[_record__loaded_at] WHEN NOT MATCHED THEN INSERT ([territory_id], [territory_description], [region_id], [_dlt_load_id], [_dlt_id], [_record__hash], [_record__loaded_at]) VALUES ([__MERGE_SOURCE__].[territory_id], [__MERGE_SOURCE__].[territory_description], [__MERGE_SOURCE__].[region_id], [__MERGE_SOURCE__].[_dlt_load_id], [__MERGE_SOURCE__].[_dlt_id], [__MERGE_SOURCE__].[_record__hash], [__MERGE_SOURCE__].[_record__loaded_at]); (base.py:2201)
2025-07-01 11:57:19,644 - MainThread - sqlmesh.core.scheduler - INFO - Execution failed for node ('"das"."raw"."raw__northwind__territories"', ((1751029200000, 1751370900000), 0)) (scheduler.py:460)
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/utils/concurrency.py", line 69, in _process_node
    self.fn(node)
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/scheduler.py", line 419, in evaluate_node
    audit_results = self.evaluate(
                    ^^^^^^^^^^^^^^
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/scheduler.py", line 172, in evaluate
    wap_id = self.snapshot_evaluator.evaluate(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 159, in evaluate
    result = self._evaluate_snapshot(
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 788, in _evaluate_snapshot
    apply(query_or_df, index)
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 696, in apply
    evaluation_strategy.insert(
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 1606, in insert
    self.adapter.merge(
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/shared.py", line 349, in internal_wrapper
    resp = func(*list_args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/mssql.py", line 267, in merge
    self._merge(
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 1440, in _merge
    self.execute(exp.Merge(this=this, using=using, on=on, whens=whens))
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 2183, in execute
    self._execute(sql, **kwargs)
  File "/workspaces/fabric/.venv/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 2204, in _execute
pyodbc.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Incorrect syntax near 'MERGE'. (102) (SQLExecDirectW)")

The above exception was the direct cause of the following exception:

sqlmesh.utils.concurrency.NodeExecutionFailedError: Execution failed for node ('"das"."raw"."raw__northwind__territories"', ((1751029200000, 1751370900000), 0))
2025-07-01 11:57:19,650 - MainThread - root - INFO - Shutting down the event dispatcher (dispatcher.py:159)
```

mattiasthalen:
@georgesittas / @erindru, from my understanding, the merge error I got is due to the unique key incremental strategy, I thought this would be handled be fabric setting the strategy to delete insert. What is needed to fix that, override the merge function in mssql?

erindru:
The `INCREMENTAL_BY_UNIQUE_KEY` kind calls `merge()` [explicitly](https://github.com/TobikoData/sqlmesh/blob/649fdf003b3e9c92f784a64269374e4ae020270d/sqlmesh/core/snapshot/evaluator.py#L1606). Check out `LogicalMergeMixin` which is a way of emulating merge on engines that dont support merge.

`InsertOverwriteStrategy` applies to the other incremental model types that are based on a time column or partition key

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,46 @@
...
+from sqlglot import exp
+from sqlmesh.core.engine_adapter.mssql import MSSQLEngineAdapter
+from sqlmesh.core.engine_adapter.shared import InsertOverwriteStrategy, SourceQuery
+from sqlmesh.core.engine_adapter.base import EngineAdapter
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core._typing import TableName
+
+
+class FabricAdapter(MSSQLEngineAdapter):
```
mattiasthalen: I added this to my own fork, and it solved the merge issue with incremental_by_unique_key:
```suggestion
from sqlmesh.core.engine_adapter.mixins import LogicalMergeMixin

class FabricAdapter(LogicalMergeMixin, MSSQLEngineAdapter):
```

mattiasthalen (approved):

georgesittas:
The work so far looks reasonable to me, but I think we need to have integration tests to consider this ready. At least the sushi project should be able to run end-to-end without issues.

@treysp @erindru since you folks have already implemented a few adapters, I'll let you take a look as well and verify the above claim re: testing.

mattiasthalen:
@georgesittas, wouldn't that need access to a fabric capacity? Trying to wrap my head around it ☺️

erindru:
@mattiasthalen yes, it would. In short, a new adapter can't be merged until it passes our integration test suite.

We can assist with this however. I'm going to do some work to enable integration tests for Fabric, with the intention that you set some environment variables to point to a Fabric endpoint you control (the same way the rest of our integration tests work).

This should enable you to run the tests locally and address any gaps. I assume you have access to a Fabric capacity since you've been testing it so far?

mattiasthalen:
@erindru yes ☺️ I have a trial capacity at work

sqlmesh/core/config/connection.py
```diff
@@ -1635,6 +1641,34 @@ def _extra_engine_config(self) -> t.Dict[str, t.Any]:
...
+    driver: t.Literal["pyodbc"] = "pyodbc"
```
erindru, line 1704: This driver literal doesn't seem to work. If I don't manually specify it in my Fabric connection config:
```
gateways:
  fabric:
    connection:
      type: fabric
      #driver: pyodbc
```

Then the `pymssql` check still happens and SQLMesh aborts with:
```
Error: Failed to import the 'pymssql' engine library.
```

mattiasthalen:
Got the test suite running. Just need to find the mental capacity to decode what's wrong 😅

mattiasthalen:
@erindru using sqlglot@main, for `test_to_time_column` there's an issue with the data type returned to ODBC:
```bash
pyodbc.ProgrammingError: ('ODBC SQL type -155 is not yet supported.  column-index=0  type=-155', 'HY106')
```

The generated SQL works fine in Fabric:
```sql
SELECT CAST('2020-01-01 00:00:00+00:00' AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC' AS the_col;
```

So my guess is that we need to override the `_fetch_native_df` method for Fabric, or should we change SQLGlot to cast it to `datetime2`?
```sql
SELECT CAST(CAST('2020-01-01 00:00:00+00:00' AS DATETIMEOFFSET(6)) AT TIME ZONE 'UTC' AS DATETIME2(6)) AS the_col;
```

fresioAS:
Glad to see you are on top of this @mattiasthalen 
I’m on holiday, so no contribution from me until 16th of July

mattiasthalen:
> Glad to see you are on top of this @mattiasthalen I’m on holiday, so no contribution from me until 16th of July

Thanks for the heads up, you had me worried that you'd given up 😅

mattiasthalen:
@erindru, good catch. so something like this #4930 ?

georgesittas:
> Going in circles :) Fabric only supports DATETIMEOFFSET when used with AT TIME ZONE, and this is handled by SQLGlot, but that doesn't solve the create table statement.

Is it possible to do something like `[value] DATETIMEOFFSET(6) AT TIME ZONE 'UTC'`, or is that syntax not supported? Btw, this is the scenario I was thinking of when I made that comment about `datatype_sql` vs `cast_sql`, namely: DDLs :)

But, more importantly, where does the `TIMESTAMPTZ` come from in this test? We have some engine-specific logic in it, I wonder if we need another branch to handle Fabric specifically?

mattiasthalen:
> If `test_to_time_column` is working, then that's the major one for SQLMesh models.
> 
> `test_value_normalization` is used for table diff. The goal is to make eg all `TIMESTAMPTZ` columns result in the same _string_ value across all our engines so that the results are comparable.
> 
> You could override the value normalization logic for Fabric to add extra casts / do whatever is needed to achieve the correct string values by overriding the `_normalize_timestamp_value()` function on `RowDiffMixin` (which is included by `MSSQLEngineAdapter` and thus inherited by `FabricEngineAdapter`)

Ok, but at the same time, since Fabric doesn't allow `DATETIMEOFFSET` in the DDL, wouldn't it be better to have SQLGlot map `TIMESTAMPTZ` to `DATETIME2`, with the "cast datetimeoffset at time zone as datetime2" logic? 

>**[datetimeoffset-transact-sql](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql):**
>In Microsoft Fabric, currently you cannot create columns with the datetimeoffset data type, but you can use datetimeoffset for converting data with the [AT TIME ZONE (Transact-SQL)](https://learn.microsoft.com/en-us/sql/t-sql/queries/at-time-zone-transact-sql?view=fabric) function

mattiasthalen:
In practical terms, that's an issue anyway with Fabric, you have to strip TZ info. So I assume best practice here is to store the TZ info next to the timestamp, but that's out of scope IMHO

sqlmesh/core/config/connection.py
```diff
@@ -1635,6 +1641,34 @@ def _extra_engine_config(self) -> t.Dict[str, t.Any]:
...
+    def _engine_adapter(self) -> t.Type[EngineAdapter]:
+        from sqlmesh.core.engine_adapter.fabric import FabricAdapter
+
+        return FabricAdapter
+
+    @property
+    def _extra_engine_config(self) -> t.Dict[str, t.Any]:
+        return {
+            "database": self.database,
+            "catalog_support": CatalogSupport.REQUIRES_SET_CATALOG,
```
mattiasthalen, line 1734: Fabric doesn't allow creating/dropping catalogs, so this must be set to single catalog.
```suggestion
            "catalog_support": CatalogSupport.SINGLE_CATALOG_ONLY,
```

mattiasthalen (changes requested):
Catalog support must be updated

erindru:
> The only way to create catalogs/dbs are via API, not SQL.

So Fabric does support this, just not via SQL? Would it be worth making an API call in the engine adapter instead to implement these operations? We do this already for BigQuery

mattiasthalen:
> > The only way to create catalogs/dbs are via API, not SQL.
> 
> 
> 
> So Fabric does support this, just not via SQL? Would it be worth making an API call in the engine adapter instead to implement these operations? We do this already for BigQuery

Yeah: https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/items/create-warehouse?tabs=HTTP

But could we introduce this in a future PR 😅

erindru:
Fair :) let's get the base implementation working first.

I'm not sure what effect changing an adapter from `SINGLE_CATALOG_ONLY` to `REQUIRES_SET_CATALOG` or `FULL_SUPPORT` will have on existing data stored in state though

sqlmesh/core/config/connection.py
georgesittas: @mattiasthalen has this been addressed?

georgesittas (commented):
Looks good, I'm also back and reviewing the pending SQLGlot PRs for this. Nice work folks!

sqlmesh/core/config/connection.py
mattiasthalen: Not in this PR, but it's there in my branch. https://github.com/TobikoData/sqlmesh/pull/4950/commits/4412fc9a6c194dc49ffb92c746d4db301bad1463

Not sure if that's the way to do it ☺️

fresioAS: i will update this PR later today

mattiasthalen: Nice to see you back @fresioAS! ☺️

There's an issue in sqlglot's tsql dialect. The ddl for varchar without precision creates a varchar(1), that affects one of the integration tests.

I'm just about to leave for vacation and won't have time to handle it. But I'll be back on Sunday.

fresioAS: @mattiasthalen Ok. will take a look this week. Summer time.  I'm off on vacation again on Monday 😄

fresioAS:
Latest code changes added!

Engine tests failing on pyodbc - not sure how to address this?

mattiasthalen:
@fresioAS I'm on my phone so I have a hard time seeing the details. Is it ok if you show me what test and the error?

sqlmesh/core/config/connection.py
```diff
@@ -1506,7 +1512,14 @@ def _mssql_engine_import_validator(cls, data: t.Any) -> t.Any:
...
+        default_driver = "pymssql"
+        if hasattr(cls, "model_fields") and "driver" in cls.model_fields:
+            field_info = cls.model_fields["driver"]
+            if hasattr(field_info, "default") and field_info.default is not None:
+                default_driver = field_info.default
+
+        driver = data.get("driver", default_driver)
 
         # Define the mapping of driver to import module and extra name
         driver_configs = {"pymssql": ("pymssql", "mssql"), "pyodbc": ("pyodbc", "mssql-odbc")}
```
mattiasthalen: This might mean that for fabric, you have to install the `mssql-odbc` extra, since fabric is never an entry in the dictionary.

tests/core/engine_adapter/integration/config.yaml
```diff
@@ -186,5 +186,18 @@ gateways:
...
+      driver: pyodbc
+      host: {{ env_var("FABRIC_HOST") }}
+      user: {{ env_var("FABRIC_CLIENT_ID") }}
+      password: {{ env_var("FABRIC_CLIENT_SECRET") }}    
+      database: {{ env_var("FABRIC_DATABASE") }}
+      odbc_properties:
+        Authentication: ActiveDirectoryServicePrincipal        
+    state_connection:
+      type: duckdb
+
```
mattiasthalen: @erindru / @georgesittas the errors we are seeing on other dialects, e.g., duckdb, goes away if this is commented out. The funny thing is that I do not get this errors in my codespace, even with a fresh one.

I tried repicating it by unsetting the env vars, but still didn't get the errors.

So I'm lost here on what to do.

erindru: Hmm, `pyodbc` is part of the `[dev]` extra so I would expect it to be available.

I wonder if it's something else, eg `odbc` not being available at the system level. But since we swallow the ImportError we lose context.

In the CircleCI config, I can see we install `unixodbc-dev` for `style_and_cicd_tests` but not the engine tests, maybe that's it

fresioAS: this worked - I also had to add this code to make circleci process properly
`      - run:
          name: Fetch full git history and tags
          command: git fetch --unshallow --tags || true`

erindru: ^ can you please explain more why that change is necessary? Nothing existing needs it, including @mattiasthalen 's branch that runs just fine without it

mattiasthalen (approved):
Looks good, we're getting close! 🎉

.circleci/continue_config.yml
```diff
@@ -59,6 +59,9 @@ jobs:
     steps:
       - halt_unless_core
       - checkout
+      - run:
+          name: Fetch full git history and tags
+          command: git fetch --unshallow --tags || true
```
erindru: Are you able to explain why these changes are necessary? No other engine needed this, including Fabric in [this pr](https://github.com/TobikoData/sqlmesh/pull/4950)

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+from sqlmesh.utils.errors import SQLMeshError
+
+if t.TYPE_CHECKING:
+    from sqlmesh.core._typing import TableName, SchemaName
+
+
+from sqlmesh.core.engine_adapter.mixins import LogicalMergeMixin
+
+logger = logging.getLogger(__name__)
+requests = optional_import("requests")
```
erindru: `requests` is part of the [direct SQLMesh dependencies](https://github.com/TobikoData/sqlmesh/blob/a52a2189bf1d6c5fdcb8864a16efe5c42c681f4d/pyproject.toml#L23) so shouldn't need `optional_import()`

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+
+
+class FabricAdapter(LogicalMergeMixin, MSSQLEngineAdapter):
```
erindru: Nit: This should be called `FabricEngineAdapter` to match our existing naming convention

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+        self,
+        schema_name: SchemaName,
+        ignore_if_exists: bool = True,
+        **kwargs: t.Any,
+    ) -> None:
+        """
+        Override create_schema to handle catalog-qualified schema names.
+        Fabric doesn't support 'CREATE SCHEMA [catalog].[schema]' syntax.
+        """
+        logger.debug(f"create_schema called with: {schema_name} (type: {type(schema_name)})")
```
erindru: nit: vestigal debug log?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        view_properties: t.Optional[t.Dict[str, exp.Expression]] = None,
+        **create_kwargs: t.Any,
+    ) -> None:
+        """
+        Override create_view to handle catalog-qualified view names and ensure schema exists.
+        Fabric doesn't support 'CREATE VIEW [catalog].[schema].[view]' syntax.
+        """
+        logger.debug(f"create_view called with: {view_name} (type: {type(view_name)})")
```
erindru: nit: vestigal debug log?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+            catalog_name = table.catalog
+            schema_name = table.db or ""
+            view_only = table.name
+
+            logger.debug(
+                f"Detected catalog.schema.view format: catalog='{catalog_name}', schema='{schema_name}', view='{view_only}'"
+            )
+
+            # Switch to the catalog first
+            self.set_current_catalog(catalog_name)
```
erindru: This logic to switch catalog based on the incoming table name seems to be repeated in a few places, could it be consolidated?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+        self._original_connection_factory = self._connection_pool._connection_factory  # type: ignore
+        # Replace the connection factory with our custom one
+        self._connection_pool._connection_factory = self._create_fabric_connection  # type: ignore
+
+    def _create_fabric_connection(self) -> t.Any:
+        """Custom connection factory that uses the target catalog if set."""
+        # If we have a target catalog, we need to modify the connection parameters
+        if self._target_catalog:
+            # The original factory was created with partial(), so we need to extract and modify the kwargs
+            if hasattr(self._original_connection_factory, "keywords"):
```
erindru: This.. looks brittle and hacky. The connection factory exposed in `FabricConnectionConfig` should be able to handle this.

However, it looks like our current connection pool implementation doesnt allow to manipulate the connection behaviour at runtime / pass extra runtime args when creating connections, i'll check internally as to what to do here.

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,673 @@
...
+
+        # If already using the requested catalog, do nothing
+        if current_catalog and current_catalog == catalog_name:
+            logger.debug(f"Already using catalog '{catalog_name}', no action needed")
+            return
+
+        logger.info(f"Switching from catalog '{current_catalog}' to '{catalog_name}'")
+
+        # Set the target catalog for our custom connection factory
+        self._target_catalog = catalog_name
```
erindru: Will storing this as a class property cause issues when multiple threads are running concurrently and potentially operating on different catalogs?

The connection pool has `get_attribute()` / `set_attribute()` for threadlocal storage, `_target_catalog` could be wrapped in an `@property` that uses this threadlocal storage instead of being set directly on the EngineAdapter class instance

mattiasthalen: I'll give it a go and report back ☺️

sqlmesh/core/engine_adapter/fabric.py
erindru: Ok, so, theoretically this should be able to work by wrapping the connection factory as it passes through `__init__`.

Something like:

```py
def __init__(
    self,
    connection_factory_or_pool: t.Union[t.Callable[[], t.Any], ConnectionPool],
    *args,
    **kwargs
):
    if not isinstance(connection_factory_or_pool, ConnectionPool):
      connection_factory_or_pool = lambda: connection_factory_or_pool(database=self._target_catalog)
    
    super().__init__(connection_factory_or_pool, *args, **kwargs)
```

In order to ensure that a new connection is created when `_target_catalog` is changed, you'd probably want to turn it into a property, and then in the setter close the current connection if there has been a change so the next query opens a new connection with the new target catalog

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Fixed in my branch, https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse. Waiting to have it pulled into this PR.

mattiasthalen:
@fresioAS, Ive addressed all the comments above, where I've referenced my copy of your branch. Could you merge that in? https://github.com/mattiasthalen/sqlmesh/tree/add_fabric_warehouse

.circleci/continue_config.yml
fresioAS: I added it because I needed this part to run in CircleCI
`      - run:
          name: Install ODBC
          command: sudo apt-get install unixodbc-dev`
As far as my understanding goes, now that the full depth repo has been fetched it can be removed again and someone else can sort this out if they need to do something similar!

.circleci/continue_config.yml
```diff
@@ -217,6 +217,9 @@ jobs:
       - run:
           name: Install OS-level dependencies
           command: ./.circleci/install-prerequisites.sh "<< parameters.engine >>"
+      - run:
+          name: Install ODBC
+          command: sudo apt-get install unixodbc-dev
```
erindru: This is already included in the `install-prerequisites.sh` script above (but I appreciate that was not the case when this PR was first opened, @mattiasthalen added it)

sqlmesh/core/config/connection.py
```diff
@@ -1690,6 +1696,77 @@ def _extra_engine_config(self) -> t.Dict[str, t.Any]:
...
+    autocommit: t.Optional[bool] = True
+
+    @property
+    def _engine_adapter(self) -> t.Type[EngineAdapter]:
+        from sqlmesh.core.engine_adapter.fabric import FabricEngineAdapter
+
+        return FabricEngineAdapter
+
+    @property
+    def _connection_factory(self) -> t.Callable:
```
erindru, line 1716: This seems quite verbose. Could it be reduced to something like:

```python
    @property
    def _connection_factory(self) -> t.Callable:
        base_factory = super()._connection_factory

        def create_fabric_connection(
            target_catalog: t.Optional[str], *args: t.Any, **kwargs: t.Any
        ) -> t.Callable:
            kwargs["database"] = target_catalog or self.database
            return base_factory(*args, **kwargs)

        return create_fabric_connection
```

Also, you dont really need to exhaustively document the _what_ of these wrapper functions, a simple comment explaining the _why_ is sufficient.

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+    SUPPORTS_INDEXES = False
+    SUPPORTS_TRANSACTIONS = False
+    SUPPORTS_CREATE_DROP_CATALOG = True
+    INSERT_OVERWRITE_STRATEGY = InsertOverwriteStrategy.DELETE_INSERT
+
+    def __init__(
+        self, connection_factory_or_pool: t.Union[t.Callable, t.Any], *args: t.Any, **kwargs: t.Any
+    ) -> None:
+        # Handle the connection factory wrapping before calling super().__init__
+        if not hasattr(connection_factory_or_pool, "get"):  # It's a connection factory, not a pool
```
erindru: This also seems quite verbose. In conjunction with the change in the connection config, could it be reduced to something like:

```python
   from sqlmesh.utils.connection_pool import ConnectionPool

    def __init__(
        self, connection_factory_or_pool: t.Union[t.Callable, t.Any], *args: t.Any, **kwargs: t.Any
    ) -> None:
        if not isinstance(connection_factory_or_pool, ConnectionPool):
            original_connection_factory = connection_factory_or_pool
            connection_factory_or_pool = lambda *args, **kwargs: original_connection_factory(
                target_catalog=self._target_catalog, *args, **kwargs
            )

        super().__init__(connection_factory_or_pool, *args, **kwargs)
```

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+        client_secret = self._extra_config.get("password")
+
+        if not all([tenant_id, client_id, client_secret]):
+            raise SQLMeshError(
+                "Service Principal authentication requires tenant_id, client_id, and client_secret "
+                "in the Fabric connection configuration"
+            )
+
+        if not requests:
+            raise SQLMeshError("requests library is required for Fabric authentication")
```
erindru: This check is no longer needed right? Since requests is now imported directly?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+
+    @retry(
+        wait=wait_exponential(multiplier=1, min=1, max=30),
+        stop=stop_after_attempt(60),
+        retry=retry_if_result(lambda result: result not in ["Succeeded", "Failed"]),
+    )
+    def _check_operation_status(self, location_url: str, operation_name: str) -> str:
+        """Check the operation status and return the status string."""
+        if not requests:
+            raise SQLMeshError("requests library is required for Fabric catalog operations")
```
erindru: This check is no longer needed right? Since requests is now imported directly?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+        """Get authentication headers for Fabric REST API calls."""
+        access_token = self._get_access_token()
+        return {"Authorization": f"Bearer {access_token}", "Content-Type": "application/json"}
+
+    def _make_fabric_api_request(
+        self, method: str, endpoint: str, data: t.Optional[t.Dict[str, t.Any]] = None
+    ) -> t.Dict[str, t.Any]:
+        """Make a request to the Fabric REST API."""
+        if not requests:
+            raise SQLMeshError("requests library is required for Fabric catalog operations")
```
erindru: This check is no longer needed right? Since requests is now imported directly?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+                        "message", str(error_response)
+                    )
+            except (ValueError, AttributeError):
+                error_details = response.text if hasattr(response, "text") else str(e)
+
+            raise SQLMeshError(f"Fabric API HTTP error ({response.status_code}): {error_details}")
+        except requests.exceptions.RequestException as e:
+            raise SQLMeshError(f"Fabric API request failed: {e}")
+
+    def _make_fabric_api_request_with_location(
```
erindru: This whole method looks very much like a duplicate of `_make_fabric_api_request` with a mild tweak copy a response header into the body payload before returning it.

Is there a way to consolidate the main logic so it's not duplicated?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+        warehouse_name = catalog_name.sql(dialect=self.dialect, identify=False)
+
+        logger.info(f"Deleting Fabric warehouse: {warehouse_name}")
+
+        try:
+            # Get the warehouse ID by listing warehouses
+            warehouses = self._make_fabric_api_request("GET", "warehouses")
+            warehouse_id = None
+
+            for warehouse in warehouses.get("value", []):
```
erindru: a more Pythonic way of doing this might be something like:

```python
            warehouse_id = next(
                (
                    warehouse.get("id")
                    for warehouse in warehouses.get("value", [])
                    if warehouse.get("displayName") == warehouse_name
                ),
                None,
            )
```

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+        exists: bool = True,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """
+        Override create_table to ensure schema exists before creating tables.
+        """
+        # Ensure the schema exists before creating the table
+        self._ensure_schema_exists(table_name)
```
erindru: Dont you already do this in `_create_table` above? If you remove this `create_table` method override entirely, wont the parent class call `self._create_table` for you?

sqlmesh/core/engine_adapter/fabric.py
```diff
@@ -0,0 +1,656 @@
...
+        exists: bool = True,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """
+        Override ctas to ensure schema exists before creating tables.
+        """
+        # Ensure the schema exists before creating the table
+        self._ensure_schema_exists(table_name)
```
erindru: Same for this, if you dont override `ctas()`, won't `self._create_table` still get called by the parent class and do the schema exists check?

sqlmesh/core/engine_adapter/fabric.py
erindru: I still think it's too complex, i've added a comment with a suggestion to simplify it. Let me know what you think

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Correct, cleaned up.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Correct, cleaned up.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Correct, cleaned up.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed, waiting for it to be pulled.

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed

sqlmesh/core/config/connection.py
mattiasthalen: Adressed

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed

sqlmesh/core/engine_adapter/fabric.py
mattiasthalen: Adressed

.circleci/continue_config.yml
fresioAS: removed

fresioAS:
> To be perfectly frank, I think theyre AI slop which we can't accept into the codebase. You were doing well until [776b840](https://github.com/TobikoData/sqlmesh/pull/4751/commits/776b8409ffa6e477403f4f66ca6cbb4d20001bf0) which went totally off the rails, and the followup commit didn't really improve much.

I suggest we roll back and take it from there - @mattiasthalen can you fix a new merge?

mattiasthalen:
> > you think they are ok, we could keep it
> 
> 
> 
> To be perfectly frank, I think theyre AI slop which we can't accept into the codebase. You were doing well until [776b840](https://github.com/TobikoData/sqlmesh/pull/4751/commits/776b8409ffa6e477403f4f66ca6cbb4d20001bf0) which went totally off the rails, and the followup commit didn't really improve much.
> 
> 
> 
> In general we are happy to accept clean, concise code that follows python idioms, has a minimum required set of changes to achieve the stated functionality and overall meets a similar level of quality to the existing code in the codebase.
> 
> 
> 
> Not whatever this is:
> 
> ```
> 
>  Global caches for performance optimization
> 
> _signature_inspection_cache: t.Dict[
> 
>     int, bool
> 
> ] = {}  # Cache for connection factory signature inspection
> 
> _signature_cache_lock = threading.RLock()  # Thread-safe access to signature cache
> 
> _warehouse_list_cache: t.Dict[
> 
>     str, t.Tuple[t.Dict[str, t.Any], float]
> 
> ] = {}  # Cache for warehouse listings
> 
> _warehouse_cache_lock = threading.RLock()  # Thread-safe access to warehouse cache
> 
> ```
> 
> 
> 
> Don't get me wrong - we would love to include Fabric support in SQLMesh and appreciate the community effort.
> 
> 
> 
> But remember, the code still needs to be maintained by our team after its merged, which means the code needs to be in a maintainable state _before_ we can consider merging it.
> 
> 

I respect that ☺️

fresioAS:
I'll test the latest PR from @erindru on my live project tomorrow - will report back!

erindru:
Nice!

> I ran into an issue with the janitor failing on clean up of a catalog I had previously deleted 

Based on your stack trace, it looks like:
 - The janitor wanted to drop an expired view. There was a reference in state and it had no idea you'd already performed the drop manually
 - In Fabric one cannot simply drop a view without first activating the correct catalog
 - Since you had already manually deleted the catalog, the switch failed
 
 This is more an issue with how the Janitor reconciles state vs reality (and already happens in various forms on other databases) so fixing it is outside the scope of this PR

> I'm a bit unsure how it initially sets my_lakehouse as current catalog, perhaps because of external_models?

Yeah I noticed this as well. If you specify an invalid database on the connection, it just "helpfully" picks one to use anyway and doesn't throw an error. For me it was the first one in the list, does that line up with what you observed?

fresioAS:
> Yeah I noticed this as well. If you specify an invalid database on the connection, it just "helpfully" picks one to use anyway and doesn't throw an error. For me it was the first one in the list that you get in the Fabric UI, does that line up with what you observed?

I'm not sure if that is the case for me - I have multiple Lakehouses/Warehouses in the GUI above the "default" database.

But it got me thinking. I pointed my test project directly to the Lakehouse - and this actually worked (by hitting the SQL analytics endpoint). The views were created on top of the Lakehouse!

Writing to tables is not allowed in Fabric through the SQL endpoint though - so any materialized model will fail by connecting directly to a Lakehouse from this PR.

mattiasthalen (approved):

PR #4752, opened 2025-06-17
Minor fix for Windows web UI interface

codetalker-ai:
On Windows OS running UI will render Source Code to be "File not found" since relative path on Windows is two backslashes `\\`, while SQLMesh uses POSIX format.
Aligned relative path to be POSIX format as well.


georgesittas (approved):
Thanks!

codykonior-hsi:
Wow, this finally makes the sqlmesh ui work for me on Windows. It never worked before, it would always say it couldn't find any file I clicked on.

I hope it gets merged in, I'll put it in my private patches for now.

StuffbyYuki:
@codetalker-ai can you rebass your branch and push it again? It looks like there is some conflict. Once it passes tests we can merge it

StuffbyYuki:
@codetalker-ai Just following up on my last comment

PR #4753, opened 2025-06-17
Feat: add formatter options to render command

georgesittas:


benfdking:
My wonder here is whether we should try to standardise it into a struct that is much easier to reuse that the flags pass in. For me if you want to format it in a certain way, you probably want it in your config.

sqlmesh/magics.py
```diff
@@ -579,23 +628,43 @@ def evaluate(self, context: Context, line: str) -> None:
...
+            start=render_opts.pop("start", None),
+            end=render_opts.pop("end", None),
+            execution_time=render_opts.pop("execution_time", None),
+            expand=render_opts.pop("expand", False),
+        )
+
+        no_format = render_opts.pop("no_format", False)
+
+        format_options = {}
+        if render_opts.get("no_rewrite_casts"):
```
erindru: This feels a bit awkward, is it turning `--no-rewrite-casts=True` into `--rewrite-casts=False`?

The logic also appears to be duplicated here and in `cli.main.render()`

Would the `--rewrite-casts/--no-rewrite-casts` [syntax](https://click.palletsprojects.com/en/stable/options/#boolean) of `@click.option` describe this better?

iirc that will give you a single boolean `rewrite_casts` that is either `True` if `--rewrite-casts` is set or `False` if `--no-rewrite-casts` is set

georgesittas: TIL, nice! I'll try it out.

georgesittas: Refactored the PR to use this feature, thanks. The magics `argument` function does not support it, so I left it as is. Additionally, I excluded this argument from `render` for simplicity, since the cast rewriting is implemented within the context's `format` method.

sqlmesh/cli/main.py
```diff
@@ -213,7 +215,17 @@ def render(
         expand=expand,
     )
 
-    sql = rendered.sql(pretty=True, dialect=ctx.obj.config.dialect if dialect is None else dialect)
+    format_config = ctx.obj.config_for_node(model).format
+    format_kwargs = {
```
themisvaltinos, line 219: nit: for clarity maybe this could be renamed from `format_kwargs` that are the parameters name to something else

sqlmesh/cli/main.py
```diff
@@ -213,7 +215,17 @@ def render(
...
+        **format_config.generator_options,
+        **{k: v for k, v in format_kwargs.items() if v is not None},
+    }
+
+    sql = rendered.sql(
+        pretty=True,
+        dialect=ctx.obj.config.dialect if dialect is None else dialect,
```
themisvaltinos, line 226: nit: I know this was how it was originally here before too but could it be `dialect or ctx.obj.config.dialect` or we have to explicitly check for none

themisvaltinos (approved):
Looks good to me!

sqlmesh/cli/main.py
georgesittas: I think checking for `None` is fine here. Theoretically speaking, if you want to you can specify `""` as the dialect, which corresponds to the base SQLGlot dialect.

sqlmesh/cli/main.py
georgesittas: It's the same mapping essentially, just filtered out the `None`s, hence why I used the same name.

PR #4759, opened 2025-06-18
refactor(lsp): make lsp more predictable and stable

benfdking:


sqlmesh/lsp/main.py
```diff
@@ -679,79 +615,94 @@ def _get_diagnostics_for_uri(self, uri: URI) -> t.Tuple[t.List[types.Diagnostic]
...
     def _ensure_context_for_document(
         self,
-        document_uri: URI,
+        document_uri: t.Optional[URI] = None,
     ) -> None:
         """
         Ensure that a context exists for the given document if applicable by searching
         for a config.py or config.yml file in the parent directories.
         """
+        if document_uri is not None:
```
themisvaltinos, line 633: looking at the `_ensure_context_in_folder` implementation I'd simplify this as :  
```suggestion
        if document_uri is not None:
            document_path = document_uri.to_path()
            if document_path.is_file():
                self._ensure_context_in_folder(document_path.parent)
                return

        self._ensure_context_in_folder()
```

sqlmesh/lsp/main.py
```diff
@@ -297,22 +256,16 @@ def did_open(ls: LanguageServer, params: types.DidOpenTextDocumentParams) -> Non
...
-
-            # Reload the entire context and create a new LSPContext
-            if self.lsp_context is not None:
-                if self._create_lsp_context(list(self.lsp_context.context.configs)):
-                    return
-
-            context = self._context_get_or_load(uri)
-            models = context.map[uri.to_path()]
-            if models is None or not isinstance(models, ModelTarget):
+            if self.lsp_context is None:
```
themisvaltinos, line 259: would it make sense to use _context_get_or_load also here instead?
```
context = self._context_get_or_load(uri)
context.context.load()
self.lsp_context = LSPContext(context.context)
```

benfdking: To me, this is slightly different because I think if you save and there is no context, you don't even want to try to make a context. It is slightly different because it is 'did_save' and so it should only really be used to update the context. Curious though what you think.

sqlmesh/lsp/main.py
```diff
@@ -679,79 +615,94 @@ def _get_diagnostics_for_uri(self, uri: URI) -> t.Tuple[t.List[types.Diagnostic]
...
     ) -> None:
         """
         Ensure that a context exists for the given document if applicable by searching
         for a config.py or config.yml file in the parent directories.
         """
+        if document_uri is not None:
+            document_path = document_uri.to_path()
+            if document_path.is_file() and document_path.suffix not in (".sql", ".py"):
+                document_folder = document_path.parent
+                if document_folder.is_dir():
```
benfdking: I think this is a mistake as well

sqlmesh/lsp/main.py
themisvaltinos: no you're right. the flow wasn’t fully clear to me of the loading happening regardless when opening a project. so the scenario I was thinking was where vscode starts with files already open from the previous session. the user edits a file and saves but since the file was already open, no did_open is sent, then did_save fires while lsp_context is None

sqlmesh/lsp/main.py
benfdking: For some reason this doesn't pass my relative tests.

PR #4760, opened 2025-06-18
Fix: Reporting the absence of physical layer updates when explaining a plan

izeigerman:
The explainer was incorrectly printing
```
Validate SQL and create physical layer tables and views if they do not exist
```
with an empty list of models if there were no physical updates to perform. This fix makes it so that 
```
SKIP: No physical layer updates to perform
```
is printed instead.

### DIFF
--- a/sqlmesh/core/plan/explainer.py
+++ b/sqlmesh/core/plan/explainer.py
@@ -90,16 +90,16 @@ class RichExplainerConsole(ExplainerConsole):
         return Tree("[bold]Execute after all statements[/bold]")
 
     def visit_physical_layer_update_stage(self, stage: stages.PhysicalLayerUpdateStage) -> Tree:
-        if not stage.snapshots:
+        snapshots = [
+            s for s in stage.snapshots if s.snapshot_id in stage.snapshots_with_missing_intervals
+        ]
+        if not snapshots:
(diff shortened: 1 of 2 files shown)


themisvaltinos (approved):

sqlmesh/core/plan/explainer.py
```diff
@@ -90,7 +90,7 @@ def visit_after_all_stage(self, stage: stages.AfterAllStage) -> Tree:
         return Tree("[bold]Execute after all statements[/bold]")
 
     def visit_physical_layer_update_stage(self, stage: stages.PhysicalLayerUpdateStage) -> Tree:
-        if not stage.snapshots:
+        if not stage.snapshots_with_missing_intervals:
             return Tree("[bold]SKIP: No physical layer updates to perform[/bold]")
```
benfdking: I understand this as purely basic logic and I'm missing the context. So correct me if I am wrong:

We should show this message when no `snapshots` are in `snapshots_with_missing_intervals`. Is there a case where you still have `snapshots_with_missing_intervals` but none of the `snapshots` are in them?

izeigerman: This is a great catch, thank you

PR #4763, opened 2025-06-19
Fix(lsp): Extend support for table references with columns

themisvaltinos:
This adds support to detect table references in column expressions for models and ctes (e.g. **sushi.orders**.customer_id). (needed for: https://github.com/TobikoData/sqlmesh/pull/4718)

### DIFF
--- a/sqlmesh/lsp/reference.py
+++ b/sqlmesh/lsp/reference.py
@@ -208,6 +208,17 @@ def get_model_definitions_for_a_path(
                                     target_range=target_range,
                                 )
                             )
+
+                            column_references = _process_column_references(
+                                scope=scope,
+                                reference_name=table.name,
+                                read_file=read_file,
+                                referenced_model_uri=document_uri,
+                                description="",
+                                reference_type="cte",
+                                cte_target_range=target_range,
+                            )
+                            references.extend(column_references)
                     continue
 
                 # For non-CTE tables, process as before (external model references)
@@ -276,6 +287,19 @@ def get_model_definitions_for_a_path(
                             target_range=yaml_target_range,
                         )
                     )
+
+                    column_references = _process_column_references(
+                        scope=scope,
+                        reference_name=normalized_reference_name,
+                        read_file=read_file,
+                        referenced_model_uri=referenced_model_uri,
+                        description=description,
+                        yaml_target_range=yaml_target_range,
+                        reference_type="external_model",
+                        default_catalog=lint_context.context.default_catalog,
+                        dialect=dialect,
+                    )
+                    references.extend(column_references)
                 else:
                     references.append(
                         LSPModelReference(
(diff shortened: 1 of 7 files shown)


sqlmesh/lsp/reference.py
```diff
@@ -288,6 +312,18 @@ def get_model_definitions_for_a_path(
...
+                        reference_name=normalized_reference_name,
+                        read_file=read_file,
+                        referenced_model_uri=referenced_model_uri,
+                        description=description,
+                        reference_type="model",
+                        default_catalog=lint_context.context.default_catalog,
+                        dialect=dialect,
+                    )
+                    references.extend(column_references)
+    # breakpoint()
```
benfdking:
```suggestion
```

sqlmesh/lsp/reference.py
```diff
@@ -735,6 +771,104 @@ def _position_within_range(position: Position, range: Range) -> bool:
...
+
+    Args:
+        column: The column expression
+        read_file: The file content as list of lines
+
+    Returns:
+        The Range covering the table reference in the column
+    """
+
+    table_parts = column.parts[:-1] if len(column.parts) > 1 else [column.parts[0]]
```
georgesittas: Not sure about the `else` branch here. If there's only one part in a `Column` instance, isn't that the column name and not a table?

sqlmesh/lsp/reference.py
```diff
@@ -735,6 +771,104 @@ def _position_within_range(position: Position, range: Range) -> bool:
...
+                        )
+                    )
+            else:
+                table_parts = [part.name for part in column.parts[:-1]]
+                table_ref = ".".join(table_parts)
+                normalized_reference_name = normalize_model_name(
+                    table_ref,
+                    default_catalog=default_catalog,
+                    dialect=dialect,
+                )
```
georgesittas: This section seems a little sketchy: `part.name` gets rid of the quotes, so `normalize_model_name` will normalize even case-sensitive (quoted) identifiers. Has this been tested / taken into account?

sqlmesh/lsp/reference.py
themisvaltinos: actually yes there's no need for it altogether since if we're here it would have more than one part so `table_parts = column.parts[:-1]` will be enough

sqlmesh/lsp/reference.py
themisvaltinos: yes the `reference_name` it compares it too has gone through the same normalisation in `get_model_definitions_for_a_path` this is why, but I will test further with more edge cases to ensure

themisvaltinos: revised it to use .sql(0 instead which is used in get_model_definitions_for_a_path as well and added a test for it : `test_quoted_uppercase_table_and_column_references`

PR #4765, opened 2025-06-19
ci: create release of vscode extension

benfdking:


.github/workflows/release_extension.yaml
```diff
@@ -0,0 +1,51 @@
+name: Release VSCode Extension
+on:
+  push:
+    tags:
+      - 'vscode@v*'
```
izeigerman, line 5: You need to either change the prefix here or in the CircleCI config: https://github.com/TobikoData/sqlmesh/blob/main/.circleci/config.yml#L17. Since both tags happen to start with `v`

PR #4766, opened 2025-06-19
fix(vscode): allow uv pip executable called through tcloud

benfdking:
- sets up environment where tcloud is called so can run `uv`

* [x] manual test
* [x] e2e tests


benfdking:
https://github.com/TobikoData/sqlmesh-enterprise/issues/2894

izeigerman (approved):

PR #4586, opened 2025-05-29
Fix: dbt-core 1.9.5 freshness error

treysp:
With 1.9.5, dbt-core added a `freshness` parameter to model and source configs (in addition to the existing top-level properties).

We pass all contents of both a source's top-level keys and config into our `SourceConfig` constructor, so we get a duplicate `freshness` key error.

This PR uses the dbt.parser.sources `merge_freshness` function to merge the two freshnesses before passing to `SourceConfig`.

closes #4574

PR #4778, opened 2025-06-21
feat(duckdb): add file systems configuration option

mrjsj:
Adds configuration option for specifying a fsspec filesystem.

Currently implemented to enable write-support for DuckDB on Azure Storage Accounts, however, it is maybe useful for other filesystems too.


mrjsj:
Last commit (https://github.com/TobikoData/sqlmesh/pull/4778/commits/c47582d9ccefff778183850d918ffef6d9e8199b) is a fix the the recent addition of `merge_freshness` in dbt 1.9.5 (see #4586).
However, CI installs dbt >= 1.10.0 in which the function is renamed `merge_source_freshness`, so mypy errors.

I'm not sure this method is acceptable.

docs/integrations/engines/duckdb.md
```diff
@@ -18,6 +18,7 @@
...
 | `connector_config` | Configuration to pass into the duckdb connector.                                                                                                                                                                                                |  dict  |    N     |
 | `secrets`   | Configuration for authenticating external sources (e.g., S3) using DuckDB secrets.                                                                                                                                                                                                |  dict  |    N     |
+| `file_systems`   | Configuration for registering `fsspec` filesystems to the DuckDB connection. secrets.                                                                                                                                                                                                |  dict  |    N     |
```
erindru: Registering filesystems to the duckdb connection secrets?

docs/integrations/engines/duckdb.md
```diff
@@ -256,4 +257,37 @@ After configuring the secrets, you can directly reference S3 paths in your catal
 
 Refer to the official DuckDB documentation for the full list of [supported S3 secret parameters](https://duckdb.org/docs/stable/extensions/httpfs/s3api.html#overview-of-s3-secret-parameters) and for more information on the [Secrets Manager configuration](https://duckdb.org/docs/configuration/secrets_manager.html).
 
-> Note: Loading credentials at runtime using `load_aws_credentials()` or similar deprecated functions may fail when using SQLMesh.
\ No newline at end of file
+> Note: Loading credentials at runtime using `load_aws_credentials()` or similar deprecated functions may fail when using SQLMesh.
+
+##### File system configuration example for Microsoft Onelake
+
+The `file_systems` accepts a list of file systems to register in the DuckDB connection. This is especially useful for Azure Storage Accounts, as it adds write support for DuckDB which is not natively supported by DuckDB (yet).
```
erindru: Nit: would it make more sense to call this property `filesystems` and not `file_systems`?

Given that the [syntax to register it](https://duckdb.org/docs/stable/guides/python/filesystems.html#example) is `register_filesystem(filesystem('foo'))`, not `register_file_system(file_system('foo'))` and the docs refer to `filesystems` and not `file_systems`

sqlmesh/dbt/target.py
```diff
@@ -147,6 +148,7 @@ class DuckDbConfig(TargetConfig):
     extensions: t.Optional[t.List[str]] = None
     settings: t.Optional[t.Dict[str, t.Any]] = None
     secrets: t.Optional[t.List[t.Dict[str, t.Any]]] = None
+    file_systems: t.Optional[t.List[t.Dict[str, t.Any]]] = None
```
erindru: Did you test this? As far as I can tell, the dbt-duckdb adapter [also calls this property](https://github.com/duckdb/dbt-duckdb?tab=readme-ov-file#duckdb-extensions-settings-and-filesystems) `filesystems` and not `file_systems`

erindru:
> I'm not sure this method is acceptable

I think we have used this method elsewhere in the codebase since we try to support a bunch of DBT versions

docs/integrations/engines/duckdb.md
mrjsj: Oops. Copy/paste gone wrong

docs/integrations/engines/duckdb.md
mrjsj: You're right. It my head it's `file_systems`, but should just align with the syntax. I'll change it to `filesystems`

mrjsj:
> > I'm not sure this method is acceptable
> 
> I think we have used this method elsewhere in the codebase since we try to support a bunch of DBT versions
> 
> EDIT: I've raised a PR to fix the unrelated DBT issue [here](https://github.com/TobikoData/sqlmesh/pull/4781)

Good - I'll remove my fix and rebase.

sqlmesh/dbt/target.py
mrjsj: Honerstly, no, I naïvely copied the `secrets` implementation and didn't give a thought. I'll rename `file_systems` to `filesystems` cf. previous comment.

erindru:
Oh, I actually liked your [original method](https://github.com/TobikoData/sqlmesh/commit/cd8f5e0696c772bffa16cd707dfdcefaa04aa693#diff-de5de8bb701edca7af9847e114df48bfceefda5f991eaa196a2dfe765a5af9a3R281-R288) of separating the two instead of taking the dbt approach of jumbling them altogether.

There are two parts to implementing this:
 - Implementing on the the SQLMesh native side in `DuckDbConnectionConfig`
 - Updating the DBT loader to map however DBT represents this back to the SQLMesh equivalent

The two structures don't need to be identical.

@izeigerman do you have an opinion on this? I guess copying what DBT does 1:1 makes it easier to map, and also we kind of take this approach on our `ConnectionConfig` already with the `type_` field

mrjsj:
> Oh, I actually liked your [original method](https://github.com/TobikoData/sqlmesh/commit/cd8f5e0696c772bffa16cd707dfdcefaa04aa693#diff-de5de8bb701edca7af9847e114df48bfceefda5f991eaa196a2dfe765a5af9a3R281-R288) of separating the two instead of taking the dbt approach of jumbling them altogether.
> 
> There are two parts to implementing this:
> 
>     * Implementing on the the SQLMesh native side in `BaseDuckDBConnectionConfig`
> 
>     * Updating the DBT loader to map however DBT represents this back to the SQLMesh equivalent
> 
> 
> The two structures don't need to be identical.
> 
> @izeigerman do you have an opinion on this? I guess copying what DBT does 1:1 makes it easier to map, and also we kind of take this approach on our `ConnectionConfig` already with the `type_` field

I did, too, but as you mentioned, I figured it would be easier to map.

However, I'll change it to what you think is best.

erindru (approved):
I guess the structure doesn't matter too much in the grand scheme of things so keeping it compatible with dbt helps with interoperability.

Thanks for adding this!

Issue #4069, opened 2025-04-01
Add relative start to plan and add support for it to CI/CD bot


eakmanrq:
Currently "start" for a plan is shared across all models while "end" is model-specific. In order to complete this issue, start must also become model-specific like end. It might be best to isolate that change in PR and validate it's behavior before adding "relative start" described below.

"relative start" is a relative time value ("1 week ago", "2 months ago", ect.) that a user can provide instead of start. What it means it the following:
* Take your model-specific end, and subtract the relative amount from it
* If the computed start/end range does not cover a complete interval, then move start further back until 1 interval is loaded

Once done, all models will have between 1 and N cron intervals loaded where N is technically the number of intervals most granular model can have for the relative time provided. A realistic example, if I have a mixture of daily and monthly models, then the range for "1 week ago" would be between 1 and 7 with 1 being the monthly models and 7 being the daily models.

Note that this is a new argument so existing "start" is untouched and the behavior should be the same. We should consider validating that what is provided to plan for "relative_start" is truly a relative value and not an absolute.

Once this is added to plan, then the CI/CD bot needs updated to add a new config which is `default_pr_relative_start`. The value provided for this would be used for the plan's `relative_start`. Add a deprecation notice for `default_pr_start` and tell users to use `default_pr_relative_start` instead. We are deprecating here since this is a change in functionality.

erindru:
This can be achieved as of [v0.200.0](https://github.com/TobikoData/sqlmesh/releases/tag/v0.200.0) with the following config:

```
cicd_bot:
  default_pr_start: '1 week ago'
  pr_min_intervals: 1
```

PR #4780, opened 2025-06-22
Feat: Allow specifying a minimum number of intervals to include for dev plans with a relative start date

erindru:
This addresses part of issue #4069 , albeit in a slightly different way to what is described in the ticket.

This PR adds a new plan option, `--relative-start-min-intervals`, intended to be used like so:

```
sqlmesh plan dev --start '2 weeks ago' --relative-start-min-intervals 1
```

What this option does is ensure that all models in the new environment have at least 1 interval backfilled, even if their interval unit is larger than the relative time period specified for `--start` by the user.

The immediate use-case is for PR environments created by the CI/CD bot which would allow you to say things like:
 - "always create PR envs with 2 weeks worth of data in them" and still have this include monthly models
 - "always create PR envs with 1 days worth of data in them" and still have this include weekly and monthly models

Right now these are excluded which can result in downstream daily models missing data in PR envs.

This option can currently only be used when `--start` is specified as a relative date or an error will be thrown which is why I called it `--relative-start-min-intervals`.

This limit is arbitrary though, technically it should work for absolute time ranges as well. In that case, a better name that is also less of a mouthful would be `--min-intervals`


sqlmesh/cli/main.py
```diff
@@ -511,6 +511,11 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     help="Explain the plan instead of applying it.",
     default=None,
 )
+@click.option(
+    "--min-intervals",
+    default=0,
+    help="In new environments created against a specific time range, ensure that models contain at least this many intervals",
```
izeigerman: I don't think this only impacts new environments, does it?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1783,26 +1783,40 @@ def missing_intervals(
     interval_end_per_model: t.Optional[t.Dict[str, int]] = None,
     ignore_cron: bool = False,
     end_bounded: bool = False,
+    min_intervals: t.Optional[int] = None,
```
izeigerman: I don't know if we should do this as part of `missing_intervals`. What I'd do instead is adjust `start` upstream in a different function rather than make this implementation even more complex.

sqlmesh/core/snapshot/definition.py
```diff
@@ -1826,17 +1840,35 @@ def missing_intervals(
...
+                max(missing_interval_start_date, inferred_start_date),
+                missing_interval_end_date,
+                execution_time=execution_time,
+                deployability_index=deployability_index,
+                ignore_cron=ignore_cron,
+                end_bounded=end_bounded,
+            )
+
+        while True:
+            intervals = _get_missing_intervals()
```
izeigerman: So for snapshots without any missing intervals we'll do this operation: `(provided_start_ms - model_start_ms) / interval_unit_ms ` times? This seems really bad

izeigerman: I think we should just adjust start like we do for previews

sqlmesh/cli/main.py
erindru: Yes, that was meant to say dev environments (there was a check that threw an error if you specified this on prod).

I've updated the text and removed the dev environment check because specifying this on `prod` is harmless in the sense that it doesnt do anything because `prod` doesnt support `--start` and `--end` so already considers the full time range

sqlmesh/core/snapshot/definition.py
erindru: We can't do this upstream in the plan builder. Here's why:

Let's say I have Model A (monthly) and Model B (daily). The dag is `A -> B`.

If I run `sqlmesh plan dev --start '1 week ago'` right now, I get 7 days of B and none of A.

If I run `sqlmesh plan dev --start '1 week ago' --min-intervals 1`, in order to satisfy `--min-intervals` at the plan building stage I would have to wind the start back by 1 month to include the monthly model A.

The problem with winding back the plan start date is, I now get ~30 missing intervals of B and 1 of A, where I actually want 7 missing intervals of B and 1 of A.

So the only place where we can adjust the start date on a per-model basis is inside `missing_intervals()`. Everywhere else it's a global value that affects all models

sqlmesh/core/snapshot/definition.py
erindru: Yeah this implementation was pretty inefficient, i've updated it to adjust the start date that gets passed to `Snapshot.missing_intervals` instead.

erindru: This logic to work out the per-model start dates has now been moved to `Context._calculate_start_override_per_model()`

sqlmesh/core/context.py
```diff
@@ -2864,6 +2885,58 @@ def _get_plan_default_start_end(
...
+        plan_end_dt = to_datetime(
+            plan_end or plan_execution_time_dt, relative_base=plan_execution_time_dt
+        )
+
+        for model_fqn in backfill_model_fqns:
+            snapshot = snapshots_by_model_fqn.get(model_fqn)
+            if not snapshot:
+                continue
+
+            starting_point = plan_end_dt
```
izeigerman: Shouldn't we use `interval_end_per_model` here instead of a global end?

sqlmesh/core/context.py
```diff
@@ -1553,6 +1573,7 @@ def plan_builder(
             ),
             end_bounded=not run,
             ensure_finalized_snapshots=self.config.plan.use_finalized_state,
+            start_override_per_model=start_override_per_model,
```
izeigerman, line 1584: Should the name be consistent with `interval_end_per_model`? I don't care which one is it, but I feel like they represent similar thing and should be named similarly.

sqlmesh/core/snapshot/definition.py
```diff
@@ -1806,13 +1809,16 @@ def missing_intervals(
...
     interval_end_per_model = interval_end_per_model or {}
     deployability_index = deployability_index or DeployabilityIndex.all_deployable()
 
     for snapshot in snapshots.values():
         if not snapshot.evaluatable:
             continue
-        snapshot_start_date = start_dt
+
+        snapshot_start_override = start_override_per_model.get(snapshot.name, None)
+        snapshot_start_date = snapshot_start_override or start_dt
```
izeigerman:
```
snapshot_start_date = start_override_per_model.get(snapshot.name, start_dt)
```
?

sqlmesh/core/snapshot/definition.py
```diff
@@ -1891,7 +1897,7 @@ def compute_missing_intervals(
     Returns:
         A list of all timestamps in this range.
     """
-    if start_ts == end_ts:
+    if start_ts >= end_ts:
```
izeigerman: Why was this changed?

sqlmesh/core/snapshot/definition.py
erindru: This has been shifted upstream by modifying the plan builder to create / populate a new property

sqlmesh/core/snapshot/definition.py
erindru: The original implementation would sometimes return intervals where the start date was later than the end date once relative timestamps had been resolved (which is error).

I reverted this change and the tests I added in this PR are still passing so I guess it is no longer necessary

sqlmesh/core/snapshot/definition.py
erindru: Good point, updated

sqlmesh/core/context.py
erindru: I've renamed `interval_end_per_model` to `end_override_per_model` on the Plan to match

sqlmesh/core/context.py
erindru: Since it's an explicit goal not to backfill anything beyond what exists in prod, yes.

I've updated this

sqlmesh/core/context.py
```diff
@@ -2887,19 +2909,95 @@ def _get_plan_default_start_end(
...
+        plan_start_dt = to_datetime(plan_start, relative_base=plan_execution_time_dt)
+        plan_end_dt = to_datetime(
+            plan_end or plan_execution_time_dt, relative_base=plan_execution_time_dt
+        )
+
+        # we need to take the DAG into account so that parent models can be expanded to cover at least as much as their children
+        # for example, A(hourly) <- B(daily)
+        # if min_intervals=1, A would have 1 hour and B would have 1 day
+        # but B depends on A so in order for B to have 1 valid day, A needs to be expanded to 24 hours
+        backfill_dag = self.dag.prune(*backfill_model_fqns)
```
izeigerman: Are you sure we can reuse this DAG? Wouldn't the loaded DAG be different if they use a selector?

sqlmesh/core/context.py
```diff
@@ -2887,19 +2909,95 @@ def _get_plan_default_start_end(
...
+
+        # So that our min start is per tree branch and not global to the entire DAG
+        reversed_subdags = [
+            backfill_dag.subdag(root_node, *backfill_dag.downstream(root_node)).reversed
+            for root_node in backfill_dag.roots
+        ]
+
+        # start from the leaf nodes and work back towards the root because the min_start at the root node is determined by the calculated starts in the leaf nodes
+        for subdag in reversed_subdags:
```
izeigerman: Wouldn't this contain overlapping subdags? Why can't we reverse the whole DAG and just traverse it in one go? So something like:
```
reversed_dag = dag.reversed
for model_fqn in reversed_dag:
    snapshot = snapshots_by_model_fqn[model_fqn]
    
    # Get the minimum start from all immediate children of this snapshot
    min_child_start = min([
        start_overrides.get(fqn, sys.max) 
        for fqn in reversed_dag.get(model_fqn, set())
    ])
    
    # Proceed with computing the start for this snapshot and taking a min of computed start and min_child_start
    
    ```

File: sqlmesh/core/context.py (hunk 8 above)
Comment (erindru) [reply sqlmesh/core/context.py]: I thought it was fine due to the pruning, but i've updated the code to construct a new DAG

File: sqlmesh/core/context.py (hunk 9 above)
Comment (erindru) [reply sqlmesh/core/context.py]: Yep, the key difference here is only checking immediate children which I missed when scanning the DAG API originally (I thought it had to be all downstream nodes).

I've adjusted it as per your suggestion and also set an override whether it's needed or not so there is always a value for each node in the `start_overrides` dict

File: sqlmesh/core/context.py [hunk 10]
```diff
@@ -2850,15 +2872,15 @@ def _plan_preview_enabled(self) -> bool:
...
+        max_interval_end_per_model: t.Dict[str, datetime],
         backfill_models: t.Optional[t.Set[str]],
         modified_model_names: t.Set[str],
         execution_time: t.Optional[TimeLike] = None,
     ) -> t.Tuple[t.Optional[int], t.Optional[int]]:
         if not max_interval_end_per_model:
             return None, None
 
-        default_end = max(max_interval_end_per_model.values())
+        default_end = to_timestamp(max(max_interval_end_per_model.values()))
```
izeigerman, line 2883: Just curious, why did we change this to datetime only to convert back to timestamp later?

izeigerman (approved):

sqlmesh/core/context.py
erindru: Rome wasnt built in a day and the rest of the code in that method was ints.

One day I hope we will use proper types internally and push `TimeLike` and co back to the edges / user input handling only

PR #4786, opened 2025-06-23
fix(vscode): passing any lineage error through

benfdking:


vscode/extension/src/webviews/lineagePanel.ts
```diff
@@ -97,12 +98,40 @@ export class LineagePanel implements WebviewViewProvider, Disposable {
...
-                    result: response,
-                  },
+                let responseCallback: CallbackEvent
+                if (isErr(response)) {
+                  let errorMessage: string
+                  switch (response.error.type) {
+                    case 'generic':
+                      errorMessage = response.error.message
+                      break
+                    case 'invalid_state':
```
themisvaltinos, line 108: would some extra info make sense?
```
    case 'generic':
      return error.message
    case 'invalid_state':
      return `Invalid state: ${error.message}`
    case 'sqlmesh_outdated':
      return `SQLMesh version issue: ${error.message}`
    default:
      return 'Unknown error occurred'
```

vscode/extension/src/lsp/lsp.ts
```diff
@@ -256,6 +256,9 @@ export class LSPClient implements Disposable {
 
     try {
       const result = await this.client.sendRequest<Response>(method, request)
+      if (result.response_error) {
```
themisvaltinos, line 259: `if (result && 'response_error' in result && result.response_error) ` in case it doesn't exist?

themisvaltinos (approved):

vscode/extension/src/lsp/custom.ts
```diff
@@ -55,9 +54,11 @@ export interface AbstractAPICallRequest {
 export interface AbstractAPICall {
   method: 'sqlmesh/api'
   request: AbstractAPICallRequest
-  response: object
+  response: AbstractAPICallResponse
 }
 
+type AbstractAPICallResponse = object & BaseResponse
```
themisvaltinos, line 60: maybe this interface as well?
```
interface AbstractAPICallResponse extends BaseResponse {
  [key: string]: any
}
```

vscode/extension/src/lsp/lsp.ts
benfdking: You shouldn't need that as it is typed. If we don't trust the typing, rather than do this we could use zod to validate the response first.

vscode/extension/src/lsp/custom.ts
benfdking: │   60:1  error  A record is preferred over an index signature  @typescript-eslint/consistent-indexed-object-style
│ ✖ 1 problem (1 error, 0 warnings)

lint error so didn't do.

vscode/extension/src/webviews/lineagePanel.ts
benfdking: 👍

PR #4787, opened 2025-06-23
fix(lsp): return context error in case it failed

benfdking:
- adds the context state to the lsp
- adds only partially working tests to extension where
  assertions need to be readded in the future

### DIFF
--- a/sqlmesh/lsp/main.py
+++ b/sqlmesh/lsp/main.py
@@ -56,6 +56,33 @@ from sqlmesh.lsp.rename import prepare_rename, rename_symbol, get_document_highl
 from sqlmesh.lsp.uri import URI
 from web.server.api.endpoints.lineage import column_lineage, model_lineage
 from web.server.api.endpoints.models import get_models
+from typing import Union
+from dataclasses import dataclass
+
+
+@dataclass
+class NoContext:
+    """State when no context has been attempted to load."""
+
+    pass
+
+
+@dataclass
+class ContextLoaded:
+    """State when context has been successfully loaded."""
+
+    lsp_context: LSPContext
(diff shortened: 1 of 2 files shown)


sqlmesh/lsp/main.py
```diff
@@ -56,6 +56,33 @@
...
+
+@dataclass
+class ContextFailed:
+    """State when context failed to load with an error message."""
+
+    error_message: str
+    context: t.Optional[Context] = None
+
+
+ContextState = Union[NoContext, ContextLoaded, ContextFailed]
```
themisvaltinos, line 85: im wondering whether this is the best approach since now since only `ContextLoaded` has the lsp_context we would need to have either type guards or assertions in quite a lot of places

sqlmesh/lsp/main.py
```diff
@@ -670,11 +723,13 @@ def _get_diagnostics_for_uri(self, uri: URI) -> t.Tuple[t.List[types.Diagnostic]
...
+        if isinstance(self.context_state, ContextFailed):
+            raise RuntimeError(self.context_state.error_message)
+        if isinstance(self.context_state, NoContext):
             self._ensure_context_for_document(document_uri)
-        if self.lsp_context is None:
-            raise RuntimeError("No context found able to get or load")
-        return self.lsp_context
+        if not isinstance(self.context_state, ContextLoaded):
+            raise RuntimeError("Context is not loaded")
+        return self.context_state.lsp_context
```
themisvaltinos, line 732:
```suggestion
        assert isinstance(self.context_state, ContextLoaded)
        return self.context_state.lsp_context
```

PR #4789, opened 2025-06-23
feat(vscode): print env into terminal

benfdking:


vscode/extension/src/commands/printEnvironment.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    if (isErr(envResult)) {
+      await vscode.window.showErrorMessage(envResult.error)
+      return
+    }
+
+    const env = envResult.value
+
+    // Create a new terminal with the SQLMesh environment
+    const terminal = vscode.window.createTerminal({
```
themisvaltinos, line 18: maybe possible future improvement instead of always creating a new terminal to see if one exists already

themisvaltinos (approved):

vscode/extension/src/commands/printEnvironment.ts
benfdking: My hesitation is that the whole point of this is to show what the lsp runs in. If you reuse, you don't quite get that.

PR #4790, opened 2025-06-23
Fix: simplify complex expressions when validating boolean pydantic fields

georgesittas:
Addresses the bug reported in this Slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1750228449327769.

T-SQL generates `Boolean` values like `TRUE` into `(1 = 1)`, which causes issues when it is used in the context of a `MODEL` property, such as `enabled`.

For example, if a macro is used and the meta block is rendered, the resulting `Model` expression is generated and re-parsed in the macro evaluator, resulting in a new `Paren(Eq(Literal(1), Literal(1))` expression for the `enabled` field, which resulted in a falsey value with the old logic.

### DIFF
--- a/sqlmesh/core/model/common.py
+++ b/sqlmesh/core/model/common.py
@@ -336,10 +336,18 @@ def parse_expression(
 
 
 def parse_bool(v: t.Any) -> bool:
-    if isinstance(v, exp.Boolean):
-        return v.this
     if isinstance(v, exp.Expression):
+        if not isinstance(v, exp.Boolean):
+            from sqlglot.optimizer.simplify import simplify
+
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/common.py
```diff
@@ -336,10 +336,18 @@ def parse_expression(
 
 
 def parse_bool(v: t.Any) -> bool:
-    if isinstance(v, exp.Boolean):
-        return v.this
     if isinstance(v, exp.Expression):
```
georgesittas, line 339: I've verified by putting an `exit(1)` within this branch that it wasn't reached by any of our tests, prior to this PR. The only dialect I'm aware of being impacted by this change is T-SQL.

tobymao (approved):

PR #4799, opened 2025-06-23
Fix(cicd_bot): Actually inherit project level auto categorization config as per docs

erindru:
The [docs](https://sqlmesh.readthedocs.io/en/stable/integrations/github/#configuration-properties) for the CICD bot say:

> Auto categorization behavior to use for the bot. **If not provided then the project-wide categorization behavior is used**.

However, this is currently not true. The CICD bot sets `CategorizerConfig.all_off()` if it's not explicitly defined which disables auto change categorization entirely by default.

This is a high friction default because it means that a `sqlmesh plan` that works fine locally throws an error about uncategorized snapshots when run on the CICD bot.

This PR updates the bot config to actually inherit the project-level auto categorization settings if no specific settings are provided, ensuring consistency by default between a local `sqlmesh plan` and a plan run by the bot.


sqlmesh/core/config/root.py
```diff
@@ -256,6 +256,10 @@ def _normalize_identifiers(key: str) -> None:
         if self.physical_schema_mapping:
             _normalize_identifiers("physical_schema_mapping")
 
+        if self.cicd_bot and not self.cicd_bot.auto_categorize_changes_:
```
izeigerman, line 259: I believe the name of the root validator no longer reflects its contents

izeigerman:
I believe this was an intentional choice. @eakmanrq do you happen to have additional context?

erindru:
Based on internal conversation, i'll raise a new PR that adjusts the CI/CD bot defaults to align with the CLI

erindru:
New PR: https://github.com/TobikoData/sqlmesh/pull/4900

PR #4800, opened 2025-06-24
feat: add fixes to linting errors

benfdking:
- add notifications when to refresh inlays and diagnostics
- add feature to core
- refactor the lsp a little bit

### DIFF
--- a/sqlmesh/core/linter/rule.py
+++ b/sqlmesh/core/linter/rule.py
@@ -39,6 +39,22 @@ class Range:
     end: Position
 
 
+@dataclass(frozen=True)
+class TextEdit:
+    """A text edit to apply to a file."""
+
+    range: Range
+    new_text: str
+
+
+@dataclass(frozen=True)
+class Fix:
+    """A fix that can be applied to resolve a rule violation."""
+
+    title: str
+    edits: t.List[TextEdit]
+
+
 class _Rule(abc.ABCMeta):
     def __new__(cls: Type[_Rule], clsname: str, bases: t.Tuple, attrs: t.Dict) -> _Rule:
         attrs["name"] = clsname.lower()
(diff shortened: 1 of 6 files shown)


sqlmesh/lsp/context.py
```diff
@@ -133,8 +146,8 @@ def lint_model(self, uri: URI) -> t.List[AnnotatedRuleViolation]:
         path = uri.to_path()
 
         # Check cache first
-        if path in self._lint_cache:
-            return self._lint_cache[path]
+        # if path in self._lint_cache:
```
benfdking: uncomment

sqlmesh/core/linter/rule.py
```diff
@@ -39,6 +39,22 @@ class Range:
     end: Position
 
 
+@dataclass(frozen=True)
+class TextEdit:
```
themisvaltinos, line 43: curious why did you create this class instead of using the lspprotocol types' one

benfdking: We don't require the lspprotocol types in 'core' yet and so that's why I always map over. I do think at some point it might be just worth using the types from the library.

themisvaltinos: ok I see in that case maybe we/i should change in rename.py as well which I used the lsp one

sqlmesh/lsp/context.py
```diff
@@ -150,6 +163,86 @@ def lint_model(self, uri: URI) -> t.List[AnnotatedRuleViolation]:
...
+                            lsp_diagnostic.range.start.line,
+                            lsp_diagnostic.range.start.character,
+                            lsp_diagnostic.range.end.line,
+                            lsp_diagnostic.range.end.character,
+                        ),
+                    )
+                    violation_map[key] = violation
+
+        # Get diagnostics in the requested range
+        diagnostics = params.context.diagnostics if params.context else []
```
themisvaltinos, line 196: should we move this up and do:
```
diagnostics = params.context.diagnostics if params.context else []
if not diagnostics:
    return None
```
if I don't read this incorrectly when we don't have diagnostics there is no need to do the violation mapping or anything else

sqlmesh/lsp/context.py
```diff
@@ -186,3 +279,68 @@ def get_completions(
...
+    @staticmethod
+    def diagnostic_to_lsp_diagnostic(
+        diagnostic: AnnotatedRuleViolation,
+    ) -> t.Optional[types.Diagnostic]:
+        if diagnostic.model._path is None:
+            return None
+        if not diagnostic.violation_range:
+            with open(diagnostic.model._path, "r", encoding="utf-8") as file:
+                lines = file.readlines()
+            range = types.Range(
```
themisvaltinos: nit: rename to `diagnostic_range` to avoid shadowing python's range

themisvaltinos (approved):
nice! lgtm some few minor nits

sqlmesh/lsp/context.py
benfdking: 🤷 , this achieves something very similar if the for loop is empty?

PR #4804, opened 2025-06-25
Fix: Push filter down to query for `get_expired_environments`

VaggelisD:


sqlmesh/core/state_sync/db/environment.py
```diff
@@ -165,20 +165,30 @@ def finalize(self, environment: Environment) -> None:
             where=environment_filter,
         )
 
+    def _fetch_environment_summaries(
```
izeigerman: [Nit] Let's make sure that the private methods follow the public ones

izeigerman (approved):

PR #4812, opened 2025-06-25
Feat!: improve signal CLI UX

georgesittas:
Current state of the CLI UX for signals on this branch:

image

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -1144,10 +1144,13 @@ class TerminalConsole(Console):
 
         # Color coding to help detect partial interval ranges quickly
         if ready_intervals == check_intervals:
+            msg = "All ready"
             color = "green"
         elif ready_intervals:
+            msg = "Some ready"
             color = "yellow"
         else:
+            msg = "None ready"
             color = "red"
 
         if self.verbosity < Verbosity.VERY_VERBOSE:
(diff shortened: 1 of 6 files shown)


sqlmesh/core/console.py
```diff
@@ -3810,22 +3938,35 @@ def _format_audits_errors(error: NodeAuditsErrors) -> str:
     return "  " + "\n".join(error_messages)
 
 
-def _format_evaluation_model_interval(snapshot: Snapshot, interval: Interval) -> str:
+def _format_interval(snapshot: Snapshot, interval: Interval, prefix: str = "") -> str:
```
izeigerman: I don't think we should try and reuse this. We should always format intervals. Let's not couple the two

sqlmesh/core/console.py
```diff
@@ -1050,6 +1109,86 @@ def stop_evaluation_progress(self, success: bool = True) -> None:
...
+            num_ready_intervals = len(formatted_ready_intervals)
+            if num_ready_intervals > 3:
+                formatted_ready_intervals = formatted_ready_intervals[:3]
+                formatted_ready_intervals.append(f"... and {num_ready_intervals - 3} more")
+
+            check = ", ".join(formatted_check_intervals)
+            tree.add(f"check: {check}")
+
+            ready = ", ".join(formatted_ready_intervals)
+            tree.add(f"[{color}]ready: {ready}[/{color}]")
```
themisvaltinos: Currently, the legend is always `ready`  and its actual status is conveyed through colour. Personally, along with the corresponding colours I’d prefer more explicit messaging for the three cases as well for example:
- All intervals ready
- Some intervals unready
- No intervals ready

and then list which intervals are ready below. It’s a minor preference, but relying solely on colour feels a bit implicit, especially from an accessibility perspective and distinguishing green and yellow in some terminals

sqlmesh/core/console.py
georgesittas: I ended up reusing the interval formatting logic for both, but decoupled it from the additional checks related to the model's kind. You can see this diff here: https://github.com/TobikoData/sqlmesh/pull/4812/commits/05c5b07ed0676b3505a83621a2053e4aa437cf36.

sqlmesh/core/console.py
georgesittas: I'd be curious to hear @sungchun12's take on this. I hear you, esp. the bit regarding accessibility. I think trying to improve the messaging may be the better thing here.

sqlmesh/core/console.py
```diff
@@ -1050,6 +1112,86 @@ def stop_evaluation_progress(self, success: bool = True) -> None:
...
+        snapshot: Snapshot,
+        signal_name: str,
+        signal_idx: int,
+        total_signals: int,
+        ready_intervals: Intervals,
+        check_intervals: Intervals,
+        duration: float,
+    ) -> None:
+        """Updates the signal checking progress."""
+        tree = Tree(f"[{signal_idx + 1}/{total_signals}] {signal_name} {duration:.2f}s")
```
georgesittas, line 1135: I'm leaving the right-justification of the duration for a followup PR (if we think it's worth it), as it will require some width calculations similar to what we do in the evaluation stage.

sqlmesh/core/console.py
georgesittas: Updated:

Non-verbose mode:
Screenshot 2025-07-01 at 7 06 24 PM

Very-verbose mode:
Screenshot 2025-07-01 at 7 06 35 PM

sungchun12: Aligned with the new images

PR #4814, opened 2025-06-25
Feat: Add support for Restatements on SCD Type 2 models

### LINKED ISSUE #3642: Support Restatements on SCD Type 2 and Use Them When Creating Preview Clone
Currently SCD Type 2 supports only restating the model from the beginning of time. One could also restate a model at any point in history too but also end with the latest date. This how how it would work:

Lets say I am restating Jan 10th
* Delete from the table where all valid_from >= Jan 10th
* Update the table where all valid_to >= Jan 10th

This would reset the table to look like it was last run before the given start date. 

Ideally, once this is implemented, when we create a preview clone of a model we would then configure the clone to restate this latest model with the new logic. 

Suggestion for resolving that is to update the SCD Type 2 logic to always go through this delete/update logic on each insert. That way in most cases where we are just processing the newest interval it would be no-op (does require a scan though) and when we have to restate it would delete/update as expected.

themisvaltinos:
This update aims to enable restatements from a specific point for SCD Type 2 (previously only supported full restatements), closes #3642 

The logic is updated for the latest and static parts so that it enables restatements when a custom start date is used, while for the end date defaults to the latest interval.


sqlmesh/core/snapshot/definition.py
```diff
@@ -795,6 +795,21 @@ def get_removal_interval(
...
+        if not is_preview and self.is_scd_type_2 and self.intervals:
+            requested_start, requested_end = removal_interval
+            latest_end = max(interval[1] for interval in self.intervals)
+            if requested_end != latest_end:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
+                    f"SCD Type 2 model '{self.model.name}' does not support end date in restatements.\n"
+                    f"Requested end date [{to_ts(requested_end)}] doesn't match latest interval end date [{to_ts(latest_end)}].\n"
+                    f"You can set start date but end date must be the latest interval so ."
```
erindru: > end date must be the latest interval so .

so... ?

themisvaltinos: Good catch, I meant to delete the whole line there since it felt repetitive

sqlmesh/core/snapshot/definition.py
```diff
@@ -797,6 +797,20 @@ def get_removal_interval(
 
             removal_interval = expanded_removal_interval
 
+        # SCD Type 2 validation that end date is the latest interval if it was provided
+        if not is_preview and self.is_scd_type_2 and self.intervals:
+            requested_start, requested_end = removal_interval
+            latest_end = max(interval[1] for interval in self.intervals)
```
izeigerman: Intervals are sorted already, so you can just do `self.intervals[-1][1]`

sqlmesh/core/snapshot/definition.py
```diff
@@ -797,6 +797,20 @@ def get_removal_interval(
...
+            if requested_end != latest_end:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
+                    f"SCD Type 2 model '{self.model.name}' does not support end date in restatements.\n"
+                    f"Requested end date [{to_ts(requested_end)}] doesn't match latest interval end date [{to_ts(latest_end)}].\n"
```
izeigerman: It warns about the problem but it doesn't say anything about what will happen. I suggest following the pattern of a similar warning above:

> Expanding the requested restatement intervals from...

sqlmesh/core/snapshot/definition.py
```diff
@@ -797,6 +797,20 @@ def get_removal_interval(
 
             removal_interval = expanded_removal_interval
 
+        # SCD Type 2 validation that end date is the latest interval if it was provided
+        if not is_preview and self.is_scd_type_2 and self.intervals:
+            requested_start, requested_end = removal_interval
+            latest_end = max(interval[1] for interval in self.intervals)
+            if requested_end != latest_end:
```
izeigerman: What about `request_end` being greater than `latest_end`? Isn't that fine?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1747,12 +1754,44 @@ def remove_managed_columns(
                 # Historical Records that Do Not Change
                 .with_(
                     "static",
-                    existing_rows_query.where(valid_to_col.is_(exp.Null()).not_()),
+                    existing_rows_query.where(valid_to_col.is_(exp.Null()).not_())
+                    if truncate
+                    else existing_rows_query.where(
+                        exp.and_(
+                            valid_from_col <= cleanup_ts,
```
eakmanrq: Is this needed? The other condition is that `valid_to_col < cleanup_ts` which means `valid_from_col` would also be `<= cleanupo_ts` right?

sqlmesh/core/snapshot/definition.py
themisvaltinos: thank you! changed it forgot they were sorted

sqlmesh/core/snapshot/definition.py
themisvaltinos: added a similar message to say it will be set to the latest interval end

sqlmesh/core/engine_adapter/base.py
themisvaltinos: yes it is indeed redundant, removed this and kept only the valid_to condition

izeigerman (approved):

PR #4819, opened 2025-06-25
Fix: Prevent formatting=false throwing an error

erindru:
Prior to this, if you tried to use the `formatting` property to disable formatting on your model, the following would occur:

```
$ sqlmesh plan

Models:
└── Directly Modified:
Error: Name needs to be a string or an Identifier, got: <class 'bool'>
```

This is because when `model.render_definition()` was called, there was no entry in `META_FIELD_CONVERTER` for the `formatting` property, so it tried to use a default that was expecting a SQLGlot expression or string.

This PR adds the entry to `META_FIELD_CONVERTER` and checks that `render_definition()` now works

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2886,6 +2886,7 @@ META_FIELD_CONVERTER: t.Dict[str, t.Callable] = {
             for name, args in values
         )
     ),
+    "formatting": lambda value: str(value),
 }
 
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -2886,6 +2886,7 @@ def render_expression(
             for name, args in values
         )
     ),
+    "formatting": lambda value: str(value),
```
georgesittas: 🙃 
```suggestion
    "formatting": str,
```

georgesittas (approved):

sqlmesh/core/model/definition.py
erindru: Man, that's a galaxy brain move. It never occurred to me to use `str` like that

PR #4820, opened 2025-06-25
fix: set default for append-newline flag to None

kelsin:


sqlmesh/cli/main.py
```diff
@@ -355,6 +356,7 @@ def evaluate(
     "--append-newline",
     is_flag=True,
     help="Include a newline at the end of each file.",
+    default=None,
```
kelsin, line 359: This is the only change that wasn't caused by precommit ruff formatting.

izeigerman (approved):

PR #4832, opened 2025-06-26
Feat!: Tag queries with their plan id

VaggelisD:
This PR enhances the `SnapshotEvaluator` such that it appends a comment containing the `plan_id` at each query generated under it.

This can help associate historical queries with a plan after the fact and paves the way for tagging queries with labels/tags in engines like BigQuery and Snowflake.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -116,7 +116,7 @@ from sqlmesh.core.test import (
     run_tests,
 )
 from sqlmesh.core.user import User
-from sqlmesh.utils import UniqueKeyDict, Verbosity
+from sqlmesh.utils import UniqueKeyDict, Verbosity, CorrelationId
 from sqlmesh.utils.concurrency import concurrent_apply_to_values
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.date import (
@@ -418,7 +418,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             self.config.get_state_connection(self.gateway) or self.connection_config
         )
 
-        self._snapshot_evaluator: t.Optional[SnapshotEvaluator] = None
+        self._snapshot_evaluators: t.Dict[t.Optional[CorrelationId], SnapshotEvaluator] = {}
 
         self.console = get_console()
         setattr(self.console, "dialect", self.config.dialect)
@@ -446,18 +446,22 @@ class GenericContext(BaseContext, t.Generic[C]):
             self._engine_adapter = self.connection_config.create_engine_adapter()
         return self._engine_adapter
 
-    @property
-    def snapshot_evaluator(self) -> SnapshotEvaluator:
-        if not self._snapshot_evaluator:
-            self._snapshot_evaluator = SnapshotEvaluator(
+    def snapshot_evaluator(
+        self, correlation_id: t.Optional[CorrelationId] = None
+    ) -> SnapshotEvaluator:
+        # Cache snapshot evaluators by correlation_id to avoid old correlation_ids being attached to future Context operations
+        if correlation_id not in self._snapshot_evaluators:
+            self._snapshot_evaluators[correlation_id] = SnapshotEvaluator(
                 {
-                    gateway: adapter.with_log_level(logging.INFO)
+                    gateway: adapter.with_settings(
+                        log_level=logging.INFO, correlation_id=correlation_id
+                    )
                     for gateway, adapter in self.engine_adapters.items()
                 },
                 ddl_concurrent_tasks=self.concurrent_tasks,
                 selected_gateway=self.selected_gateway,
             )
-        return self._snapshot_evaluator
+        return self._snapshot_evaluators[correlation_id]
 
     def execution_context(
         self,
@@ -538,7 +542,9 @@ class GenericContext(BaseContext, t.Generic[C]):
 
         return self.create_scheduler(snapshots)
 
-    def create_scheduler(self, snapshots: t.Iterable[Snapshot]) -> Scheduler:
+    def create_scheduler(
+        self, snapshots: t.Iterable[Snapshot], correlation_id: t.Optional[CorrelationId] = None
+    ) -> Scheduler:
         """Creates the built-in scheduler.
 
         Args:
@@ -549,7 +555,7 @@ class GenericContext(BaseContext, t.Generic[C]):
         """
         return Scheduler(
             snapshots,
-            self.snapshot_evaluator,
+            self.snapshot_evaluator(correlation_id),
             self.state_sync,
             default_catalog=self.default_catalog,
             max_workers=self.concurrent_tasks,
@@ -714,7 +720,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             NotificationEvent.RUN_START, environment=environment
         )
         analytics_run_id = analytics.collector.on_run_start(
-            engine_type=self.snapshot_evaluator.adapter.dialect,
+            engine_type=self.snapshot_evaluator().adapter.dialect,
             state_sync_type=self.state_sync.state_type(),
         )
         self._load_materializations()
@@ -1076,7 +1082,7 @@ class GenericContext(BaseContext, t.Generic[C]):
             and not parent_snapshot.categorized
         ]
 
-        df = self.snapshot_evaluator.evaluate_and_fetch(
+        df = self.snapshot_evaluator().evaluate_and_fetch(
             snapshot,
             start=start,
             end=end,
@@ -1588,7 +1594,12 @@ class GenericContext(BaseContext, t.Generic[C]):
                 default_catalog=self.default_catalog,
                 console=self.console,
             )
-            explainer.evaluate(plan.to_evaluatable())
+            explainer.evaluate(
+                plan.to_evaluatable(),
+                snapshot_evaluator=self.snapshot_evaluator(
+                    correlation_id=CorrelationId.from_plan_id(plan.plan_id)
+                ),
+            )
             return
 
         self.notification_target_manager.notify(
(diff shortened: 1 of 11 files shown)


sqlmesh/core/context.py
```diff
@@ -446,12 +446,14 @@ def engine_adapter(self) -> EngineAdapter:
             self._engine_adapter = self.connection_config.create_engine_adapter()
         return self._engine_adapter
 
-    @property
-    def snapshot_evaluator(self) -> SnapshotEvaluator:
-        if not self._snapshot_evaluator:
+    def snapshot_evaluator(self, job_id: t.Optional[str] = None) -> SnapshotEvaluator:
+        # Cache snapshot evaluators by job_id to avoid old job_ids being attached to future Context operations
```
erindru: Would `@cache` or `@lru_cache` work for this? iirc it [takes the method arguments](https://docs.python.org/3/library/functools.html#functools.lru_cache) into account

sqlmesh/core/plan/explainer.py
```diff
@@ -37,7 +38,10 @@ def __init__(
         self.console = console or get_console()
 
     def evaluate(
-        self, plan: EvaluatablePlan, circuit_breaker: t.Optional[t.Callable[[], bool]] = None
+        self,
+        plan: EvaluatablePlan,
+        snapshot_evaluator: SnapshotEvaluator,
```
erindru, line 43: where is the added `snapshot_evaluator` argument used here?

tests/conftest.py
```diff
@@ -266,10 +266,10 @@ def duck_conn() -> duckdb.DuckDBPyConnection:
 def push_plan(context: Context, plan: Plan) -> None:
     plan_evaluator = BuiltInPlanEvaluator(
         context.state_sync,
-        context.snapshot_evaluator,
         context.create_scheduler,
         context.default_catalog,
     )
+    plan_evaluator.snapshot_evaluator = context.snapshot_evaluator(job_id=plan.plan_id)
```
erindru: nit: is there a reason you set the property afterwards vs just passing `context.snapshot_evaluator(job_id=plan.plan_id)` to the `BuiltInPlanEvaluator` constructor?

tests/core/test_integration.py
```diff
@@ -1137,7 +1137,7 @@ def test_non_breaking_change_after_forward_only_in_dev(
     init_and_plan_context: t.Callable, has_view_binding: bool
 ):
     context, plan = init_and_plan_context("examples/sushi")
-    context.snapshot_evaluator.adapter.HAS_VIEW_BINDING = has_view_binding
+    context.snapshot_evaluator().adapter.HAS_VIEW_BINDING = has_view_binding
```
erindru, line 1141: Since there is a bunch of these places that don't care about the `job_id`, it may be worth re-adding the `context.snapshot_evaluator` property and just have it return `self.snapshot_evaluator(job_id=None)`.

I'm not 100% on the typing overload syntax to know if that's actually possible though

sqlmesh/core/plan/explainer.py
izeigerman: It's a part of ABC

tests/conftest.py
izeigerman: That was my suggestion. The interface of the PlanEvaluator assumes that the same instance can be reused for multiple plans. Setting `snapshot_evaluator` in the constructor fixes the job ID across all plans. 

So we should either pass both the plan and the snapshot evaluator in constructor or pass both of them in the `evaluate` method. I suggested the latter.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2175,6 +2178,9 @@ def execute(
                 else:
                     sql = t.cast(str, e)
 
+                if self._job_id:
+                    sql = f"/* sqlmesh_ref: {self._job_id} */ {sql}"
```
izeigerman: Instead of obscure `sqlmesh_ref`, can we call it "Correlation ID"? Ideally we should be able to tell whether it's a plan or a run. Can `job_id` be a dataclass instead of just a string and include a job type? Maybe let's call it `CorrelationId` since "job" is a very overloaded term.

tests/core/test_integration.py
izeigerman: Can you please clarify what exactly will we be solving by doing this?

erindru: Mostly just ergonomics, accessing a property vs a method call. But I guess it doesnt really matter in the grand scheme of things

sqlmesh/core/context.py
VaggelisD: You're right, but I just thought of an oversight with both functools and my current implementation: When swapping the `_snapshot_evaluator`, the previous one should be closed through `SnapshotEvaluator::close()` in order to release the adapters.

I guess the functools abstraction hinders this in general since we can't access the objects (?), but I can adapt my current implementation for that matter.

tests/core/test_integration.py
VaggelisD: Yeah I had the same idea as well, I think it can be done and would revert a lot of changes in this PR but it's considered an anti-pattern afaict so I'm split. Happy to consider either direction

sqlmesh/core/engine_adapter/base.py
VaggelisD: Implemented in my latest commit

sqlmesh/core/context.py
```diff
@@ -446,18 +446,18 @@ def engine_adapter(self) -> EngineAdapter:
...
-    def snapshot_evaluator(self) -> SnapshotEvaluator:
-        if not self._snapshot_evaluator:
-            self._snapshot_evaluator = SnapshotEvaluator(
+    def snapshot_evaluator(self, job_id: t.Optional[CorrelationId] = None) -> SnapshotEvaluator:
+        # Cache snapshot evaluators by job_id to avoid old job_ids being attached to future Context operations
+        if job_id not in self._snapshot_evaluators:
+            self._snapshot_evaluators[job_id] = SnapshotEvaluator(
                 {
-                    gateway: adapter.with_log_level(logging.INFO)
+                    gateway: adapter.with_settings(level=logging.INFO, job_id=job_id)
```
georgesittas: Nit: I'd rename `level` to `log_level` which is more explicit.

sqlmesh/core/plan/evaluator.py
```diff
@@ -60,20 +63,20 @@ def evaluate(
...
+            snapshot_evaluator: The snapshot evaluator to use.
+            circuit_breaker: The circuit breaker to use.
         """
 
 
 class BuiltInPlanEvaluator(PlanEvaluator):
     def __init__(
         self,
         state_sync: StateSync,
-        snapshot_evaluator: SnapshotEvaluator,
```
georgesittas, line 70: Shouldn't we mark the PR as breaking?

sqlmesh/utils/__init__.py
```diff
@@ -382,3 +386,23 @@ def to_snake_case(name: str) -> str:
     return "".join(
         f"_{c.lower()}" if c.isupper() and idx != 0 else c.lower() for idx, c in enumerate(name)
     )
+
+
+class JobType(Enum):
+    PLAN = "PLAN"
+    RUN = "RUN"
```
georgesittas: Nit: it appears that we inject comments like `/* PLAN: <id> */`. Should this be `SQLMesh plan: <id>` instead? Just to namespace the comment properly and avoid putting the generic "PLAN" keyword which may show elsewhere. Just makes it slightly less likely to avoid false positives when searching the db query history.

georgesittas (commented):
Nice work, looks good. Is the next step to manually intercept the adapter instantiations elsewhere (e.g., unit tests), to ensure the IDs are properly generated in their `execute` as well?

sqlmesh/utils/__init__.py
VaggelisD: Good idea 👍

sqlmesh/core/context.py
izeigerman: > hen swapping the _snapshot_evaluator, the previous one should be closed through SnapshotEvaluator::close() in order to release the adapters.

If we need to close this resource than I'd suggest to not cache this at all and make it a context manager instead.

sqlmesh/core/context.py
```diff
@@ -2334,8 +2339,8 @@ def print_environment_names(self) -> None:
 
     def close(self) -> None:
         """Releases all resources allocated by this context."""
-        if self._snapshot_evaluator:
-            self._snapshot_evaluator.close()
+        for evaluator in self._snapshot_evaluators.values():
+            evaluator.close()
```
izeigerman, line 2350: If we close here, we should probably clear the dict

sqlmesh/utils/__init__.py
```diff
@@ -382,3 +386,25 @@ def to_snake_case(name: str) -> str:
     return "".join(
         f"_{c.lower()}" if c.isupper() and idx != 0 else c.lower() for idx, c in enumerate(name)
     )
+
+
+class JobType(Enum):
+    PLAN = "SQLMESH_PLAN"
```
izeigerman, line 389: I think `SQLMESH` prefix is redundant

sqlmesh/core/context.py
VaggelisD: Ended up cleaning up the resources in `Context::close` as was done before, can make it into a context manager if need be

sqlmesh/utils/__init__.py
```diff
@@ -382,3 +386,23 @@ def to_snake_case(name: str) -> str:
...
+    """ID that is added to each query in order to identify the job that created it."""
+
+    job_type: JobType
+    job_id: str
+
+    def __str__(self) -> str:
+        return f"{self.job_type.value}: {self.job_id}"
+
+    @classmethod
+    def from_plan(cls, plan: Plan) -> CorrelationId:
```
izeigerman: [Nit] Passing the whole plan here seems a little excessive, provided you only need a plan ID. So `from_plan_id` would've been more appropriate.

georgesittas (approved):

PR #4836, opened 2025-06-27
Fix: bug in model creation related to metadata-only statements

georgesittas:
The `statements` list can contain pairs of expressions & booleans, which we didn't take into account in the line I changed.


sqlmesh/core/model/definition.py
```diff
@@ -2453,7 +2453,7 @@ def _create_model(
             statements.extend(property_values.expressions)
 
     jinja_macro_references, used_variables = extract_macro_references_and_variables(
-        *(gen(e) for e in statements)
+        *(gen(e if isinstance(e, exp.Expression) else e[0]) for e in statements)
```
izeigerman, line 2456: Are we sure this is a non-breaking change?

georgesittas: It should be fine, I think. These `gen`erated strings are used for extracting macro & variable references. If I'm not mistaken, we'd previously produce `"(True, <expression SQL>)"` due to reaching [this](https://github.com/tobymao/sqlglot/blob/main/sqlglot/optimizer/simplify.py#L1371-L1373) section. After this PR we'll only search `<expression SQL>`.

PR #4838, opened 2025-06-27
Feat: add dtntz date macro variant

georgesittas:
Addresses the issue discussed in this Slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1749739413928599.

The `_dt` date macro variables always inject a time zone in the resulting `datetime` object, even if it's `UTC`. This is problematic in certain cases, e.g., when these values are compared to a zone-unaware timestamps. For instance, Redshift isn't clever enough when comparing zone-aware and unaware timestamps, leading to full scans instead of pruning.

The only alternative that exists today to work around this issue is to have string versions of these date values and cast them, which isn't a great experience. This PR covers this feature gap.


erindru (approved):
We might want to document this on the Redshift engine adapter page for people to consider if they're having trouble with `_dt`

Issue #4405, opened 2025-05-13
`.env` file to load in environment variables easily for gateway configs


sungchun12:
Make it easy to configure gateway configs with environment variables with a `.env` file to prevent the tedious exercise of doing `export PASSWORD=jaiowjfe` each time a user opens a new shell session or doesn't want to hard code in a bash config. 

Can use this: https://github.com/theskumar/python-dotenv

Internal slack thread: https://tobikodata.slack.com/archives/C08PZF4PL21/p1747171622029019

mattiasthalen:
That's one of the main reasons I do my config in Python, so that I can just throw in `load_dotenv()`.

Is this something that would be added where the config is loaded, or even closer to the entrypoint?

sungchun12:
> That's one of the main reasons I do my config in Python, so that I can just throw in `load_dotenv()`.
> 
> Is this something that would be added where the config is loaded, or even closer to the entrypoint?

Yes, in the same project root directory of the config: `config.yaml`, `config.py`

PR #4840, opened 2025-06-27
Feat: Add support for dot env file to load variables from

themisvaltinos:
This update allows environment variables being loaded from a `.env` file in the project directory, fixes: #4405 

Example `.env` file:

```bash
SNOWFLAKE_PW=my_secret_password
S3_BUCKET=s3://my-data-bucket/warehouse

# Override SQLMesh config values
SQLMESH__DEFAULT_GATEWAY=production
SQLMESH__MODEL_DEFAULTS__DIALECT=snowflake
```

### DIFF
--- a/sqlmesh/core/config/loader.py
+++ b/sqlmesh/core/config/loader.py
@@ -6,6 +6,7 @@ import typing as t
 from pathlib import Path
 
 from pydantic import ValidationError
+from dotenv import load_dotenv
 from sqlglot.helper import ensure_list
 
 from sqlmesh.core import constants as c
@@ -34,6 +35,9 @@ def load_configs(
         for path in ensure_list(paths)
         for p in (glob.glob(str(path)) or [str(path)])
     ]
+
+    for path in absolute_paths:
+        load_dotenv(dotenv_path=path / ".env", override=True)
 
     if not isinstance(config, str):
         if type(config) != config_type:
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -98,7 +98,32 @@ All software runs within a system environment that stores information as "enviro
 
 SQLMesh can access environment variables during configuration, which enables approaches like storing passwords/secrets outside the configuration file and changing configuration parameters dynamically based on which user is running SQLMesh.
 
-You can use environment variables in two ways: specifying them in the configuration file or creating properly named variables to override configuration file values.
+You can specify environment variables in the configuration file or by storing them in a `.env` file.
+
+### .env files
+
+SQLMesh automatically loads environment variables from a `.env` file in your project directory. This provides a convenient way to manage environment variables without having to set them in your shell.
+
+Create a `.env` file in your project root with key-value pairs:
+
+```bash
+# .env file
+SNOWFLAKE_PW=my_secret_password
+S3_BUCKET=s3://my-data-bucket/warehouse
+DATABASE_URL=postgresql://user:pass@localhost/db
+
+# Override specific SQLMesh configuration values
+SQLMESH__DEFAULT_GATEWAY=production
+SQLMESH__MODEL_DEFAULTS__DIALECT=snowflake
+```
+
+See the [overrides](#overrides) section for a detailed explanation of how these are defined.
+
+The rest of the `.env` file variables can be used in your configuration files with `{{ env_var('VARIABLE_NAME') }}` syntax in YAML or accessed via `os.environ['VARIABLE_NAME']` in Python.
+
+**Important considerations:**
+- Add `.env` to your `.gitignore` file to avoid committing sensitive information
+- SQLMesh will only load the `.env` file if it exists in the project directory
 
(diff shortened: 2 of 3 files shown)


sqlmesh/core/config/loader.py
```diff
@@ -6,6 +6,7 @@
 from pathlib import Path
 
 from pydantic import ValidationError
+from dotenv import load_dotenv
```
georgesittas, line 9: This is a new dependency that needs to be included in `pyproject.toml`, right?

sqlmesh/core/config/loader.py
```diff
@@ -35,6 +36,9 @@ def load_configs(
         for p in (glob.glob(str(path)) or [str(path)])
     ]
 
+    for path in absolute_paths:
+        load_dotenv(dotenv_path=path / ".env", override=True)
+
```
georgesittas: We can make this feature more generic by supporting arbitrary user-supplied "env" files, not necessarily constraining their name to `.env`, or treating them as special SQLMesh project files. That way one can configure where the file resides, e.g., through the CLI, and we can load it regardless of its location.

For example, I'd like to have such a file in my home directory and just reference it, instead of copying over all necessary variables to the repo and having to put it in `.gitignore` to avoid leaking them.

@izeigerman @sungchun12 what do you think?

sungchun12: Agreed and I still want a clean default where it detects .env in the root directory and uses that first. If people want to customize, they can do what you suggested.

sqlmesh/core/config/loader.py
themisvaltinos: it is already pulled in as a dependency of `uvicorn` and installed but I'll add it explicitly in the .toml as well for good measure

sqlmesh/core/config/loader.py
themisvaltinos: yeah makes sense I'll add the option to allow a custom path to the directory and name

sqlmesh/core/config/loader.py
izeigerman: Yes, let's not rely on transitivity

sqlmesh/core/config/loader.py
izeigerman: How is the customization going to look like? An environment variable with a path?

themisvaltinos: I just pushed what I had in mind @izeigerman to enable this, I don't know if there are possible better alternatives

I was thinking setting it in the config is not possible due to the chicken-and-egg problem so the easier thing would be to be set once like: `export SQLMESH_DOTENV_PATH=/path/to/custom/.custom_env`

sungchun12: I can see users wanting to set a flag like this ad hoc:

I theorize this will be preferred for testing different env vars (ex: testing permissions between user roles in applying plans. Users can do this in addition to: `export SQLMESH_DOTENV_PATH=/path/to/custom/.custom_env` like you have above.
`sqlmesh plan --env-vars /path/to/custom/.custom_env`

AND I can see users wanting this:

I theorize this will be preferred for the custom paths. Edit: However, I realize this maps directly to your chicken and egg problem. If `.env` doesn't exist in the project root directory, can this config be the default? If `.env` does exist, error out and tell the user to either remove the `.env` file or remove the `env_vars_path` config.
```yaml
# config.yaml
env_vars_path:  /path/to/custom/.custom_env
```

izeigerman:
```
# config.yaml
env_vars_path:  /path/to/custom/.custom_env
```

I don't think this works since we're in a process of loading a config here, which creates a chicken and egg problem.
```
sqlmesh --env-vars ... plan
```
approach may work.

sqlmesh/magics.py
```diff
@@ -166,6 +167,9 @@ def _shell(self) -> t.Any:
     @argument("--ignore-warnings", action="store_true", help="Ignore warnings.")
     @argument("--debug", action="store_true", help="Enable debug mode.")
     @argument("--log-file-dir", type=str, help="The directory to write the log file to.")
+    @argument(
+        "--dotenv", type=str, help="Path to a custom .env file to load environment variables from."
```
izeigerman, line 171: @sungchun12 you good with this name?

izeigerman (approved):

sqlmesh/magics.py
sungchun12: This works

pyproject.toml
```diff
@@ -79,6 +79,7 @@ dev = [
     "PyAthena[Pandas]",
     "PyGithub>=2.6.0",
     "pyperf",
+    "python-dotenv",
```
georgesittas, line 82: @themisvaltinos we need to move this to `dependencies`– this isn't a dev dependency, because the `.env` functionality requires that this package is installed on the user's system.

PR #4844, opened 2025-06-27
Fix: Make start and end macros produce different hashes

### LINKED ISSUE #3500: Date Macros result in same Data Hash
The data hash for

select @start_ds is the same as select @end_ds because they both evaluate to '1970-01-0',

ideally we need to disambiguate this so that users using these macros in their queries get a diff.

MikeWallis42:
Fixes #3500

Prior to this, as they have the same default value, if only 1 is used and then changed that would not result in a new data_hash and therefor would not be considered a breaking change when planning.

I'm fully aware that this may need a migration script, please let me know if so as I might need some assistance.

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 from contextlib import contextmanager
+from datetime import timedelta
 from functools import partial
 from pathlib import Path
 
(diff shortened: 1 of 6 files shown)


tests/core/test_config.py
```diff
@@ -990,6 +990,7 @@ def test_environment_statements_config(tmp_path):
 def test_pydantic_import_error() -> None:
     class TestConfig(DuckDBConnectionConfig):
         pass
+    TestConfig.model_rebuild()
```
MikeWallis42, line 993: I tried to fix a local issue here but didn't completely fix it.

sqlmesh/core/renderer.py
```diff
@@ -173,7 +174,7 @@ def _resolve_table(table: str | exp.Table) -> str:
         )
 
         start_time, end_time = (
-            make_inclusive(start or c.EPOCH, end or c.EPOCH, self._dialect)
+            make_inclusive(start or c.EPOCH, end or c.EPOCH + timedelta(days=1), self._dialect)
```
izeigerman, line 177: I don't think this is the only place where we use epoch as a placeholder. I wonder whether we should instead have 3 constants:
```
START_PLACEHOLDER
END_PLACEHOLDER
EXECUTION_TIME_PLACEHOLDER
```
and use these instead throughout.

StuffbyYuki:
@MikeWallis42 Wanted to follow up on this one. Or we can have somebody else pick this one up

PR #4847, opened 2025-06-30
Fix: Normalize when_matched and merge_filter expressions to the source dialect

### LINKED ISSUE #4843: The "when_matched" expression of the INCREMENTAL_BY_UNIQUE key kind is not normalized
The identifiers within the `when_matched` expressions get quoted but normalized. So for example if `when_matched` is set to something like:
```
    when_matched (
      WHEN MATCHED THEN UPDATE SET
        __MERGE_TARGET__.key_a = __MERGE_SOURCE__.key_a,
        __MERGE_TARGET__.key_b = __MERGE_SOURCE__.key_b,
    ),
```
This will fail in Snowflake because the final query will reference `"key_a"` and `"key_b"` - quoted and lowercased.

erindru:
Fixes #4843 

Currently, neither the `when_matched` or `merge_filter` user-supplied expressions are normalized. This can lead to issues when queries are executed, because we quote by default.

So an expression in a Model definition like:
```
  when_matched (
      WHEN MATCHED THEN UPDATE SET target.key_a = source.key_a,
    ),
```

Would get executed as:
```
... WHEN MATCHED THEN UPDATE SET "__MERGE_TARGET__"."key_a" = "__MERGE_SOURCE__"."key_a" ...
```

which is incorrect for databases like Snowflake as unquoted identifiers should be normalized to uppercase like so:

```
... WHEN MATCHED THEN UPDATE SET "__MERGE_TARGET__"."KEY_A" = "__MERGE_SOURCE__"."KEY_A" ...
```

This PR normalizes the user input for the `when_matched` and `merge_filter` properties of `IncrementalByUniqueKeyKind` as well as normalizing the `MERGE_SOURCE_ALIAS` and `MERGE_TARGET_ALIAS` constants so they are consistent with the user query snippet


tests/core/test_model.py
```diff
@@ -5400,9 +5400,9 @@ def test_when_matched():
   kind INCREMENTAL_BY_UNIQUE_KEY (
     unique_key ("purchase_order_id"),
     when_matched (
-      WHEN MATCHED AND __MERGE_SOURCE__._operation = 1 THEN DELETE
-      WHEN MATCHED AND __MERGE_SOURCE__._operation <> 1 THEN UPDATE SET
-        __MERGE_TARGET__.purchase_order_id = 1
+      WHEN MATCHED AND __merge_source__._operation = 1 THEN DELETE
```
erindru: Our tests tend to use dialects where a normalized identifier is lowercased, so fixing this issue required a bunch of edits to test SQL

sqlmesh/core/model/kind.py
```diff
@@ -468,15 +468,20 @@ def _when_matched_validator(
...
             if v.startswith("("):
                 v = v[1:-1]
 
-            return t.cast(exp.Whens, d.parse_one(v, into=exp.Whens, dialect=get_dialect(info.data)))
+            v = t.cast(exp.Whens, d.parse_one(v, into=exp.Whens, dialect=dialect))
+        else:
+            v = t.cast(exp.Whens, v.transform(d.replace_merge_table_aliases, dialect=dialect))
 
-        return t.cast(exp.Whens, v.transform(d.replace_merge_table_aliases))
+        return normalize_identifiers(v, dialect=dialect)
```
izeigerman: If we do it here, then wouldn't this require a migration? Since this changes how the expressions are stored in the state. Don't we need a migration script and to make this a breaking change?

sqlmesh/core/model/kind.py
```diff
@@ -486,11 +491,16 @@ def _merge_filter_validator(
...
             return v
+
+        dialect = get_dialect(info.data)
+
         if isinstance(v, str):
             v = v.strip()
-            return d.parse_one(v, dialect=get_dialect(info.data))
+            v = d.parse_one(v, dialect=dialect)
+        else:
+            v = v.transform(d.replace_merge_table_aliases, dialect=dialect)
```
izeigerman, line 501: why was this added?

sqlmesh/core/model/kind.py
izeigerman: Also note, that in other places (eg. partitioned_by, time column, etc) we quote identifiers in addition to normalizing them.

erindru: I've updated this to normalize, quote and attach `.meta["dialect"]` to the returned expressions so that when they get serialized to JSON for state, they are serialized according to the correct dialect.

I re-used the `_get_field` private function from `sqlmesh.utils.pydantic` which did all of this already but added a public function to expose it

tests/core/test_model.py
```diff
@@ -5366,13 +5366,13 @@ def test_when_matched():
...
-    expected_when_matched = "(WHEN MATCHED THEN UPDATE SET __MERGE_TARGET__.salary = COALESCE(__MERGE_SOURCE__.salary, __MERGE_TARGET__.salary))"
+    expected_when_matched = "(WHEN MATCHED THEN UPDATE SET `__merge_target__`.`salary` = COALESCE(`__merge_source__`.`salary`, `__merge_target__`.`salary`))"
 
     model = load_sql_based_model(expressions, dialect="hive")
-    assert model.kind.when_matched.sql() == expected_when_matched
+    assert model.kind.when_matched.sql(dialect="hive") == expected_when_matched
 
     model = SqlModel.parse_raw(model.json())
-    assert model.kind.when_matched.sql() == expected_when_matched
+    assert model.kind.when_matched.sql(dialect="hive") == expected_when_matched
```
erindru, line 5375: Dialects started becoming very relevant for `.sql()` calls in many tests now that we quote identifiers

tests/core/test_model.py
```diff
@@ -7894,7 +7953,7 @@ def test_merge_filter():
                 source.ds > (SELECT MAX(ds) FROM db.test) AND
                 source.ds > @start_ds AND
                 source._operation <> 1 AND
-                target.start_date > dateadd(day, -7, current_date)
+                target.start_date > date_add(current_date, interval 7 day)
```
erindru, line 7956: `dateadd(day, -7, current_date)` isnt actually a duckdb function. Due to this, `day` was getting treated as a column `"day"` which is incorrect.

So I changed this to the [correct syntax](https://duckdb.org/docs/stable/sql/functions/date.html#date_adddate-interval)

sqlmesh/core/model/kind.py
erindru: I also added an empty migration

Issue #4307, opened 2025-05-02
Make cache directory location configurable


treysp:
Currently, the `.cache` directory is always colocated with a project's config file, but that has negative side effects in some contexts.

PR #4869, opened 2025-07-01
Feat: Add support for configurable cache directory

themisvaltinos:
By default SQLMesh stores its cache in a `.cache` folder inside the project directory. This introduces a configuration option to specify a custom cache location, either as a relative path or an absolute path. Fixes: #4307 

```yaml
# Relative path to project directory
cache_dir: .my_custom_cache

# Absolute path
cache_dir: /tmp/sqlmesh_cache
```

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -105,7 +105,7 @@ class _EngineAdapterStateSyncSchedulerConfig(SchedulerConfig):
 
         schema = context.config.get_state_schema(context.gateway)
         return EngineAdapterStateSync(
-            engine_adapter, schema=schema, context_path=context.cache_dir, console=context.console
+            engine_adapter, schema=schema, cache_dir=context.cache_dir, console=context.console
         )
 
     def state_sync_fingerprint(self, context: GenericContext) -> str:
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -105,7 +105,7 @@ class _EngineAdapterStateSyncSchedulerConfig(SchedulerConfig):
 
         schema = context.config.get_state_schema(context.gateway)
         return EngineAdapterStateSync(
-            engine_adapter, schema=schema, context_path=context.path, console=context.console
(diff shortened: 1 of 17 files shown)


tests/core/test_context.py
```diff
@@ -652,6 +652,96 @@ def test_clear_caches(tmp_path: pathlib.Path):
...
+        f"""test_config = Config(
+    gateways={{"in_memory": GatewayConfig(connection=DuckDBConnectionConfig())}},
+    default_gateway="in_memory",
+    plan=PlanConfig(
+        auto_categorize_changes=CategorizerConfig(
+            sql=AutoCategorizationMode.SEMI, python=AutoCategorizationMode.OFF
+        )
+    ),
+    model_defaults=model_defaults,
+    cache_dir="{custom_cache_dir}",
```
erindru: I think injecting the path as a quoted string here is breaking the tests on Windows, because it ends up being `cache_dir="C:\some\path"` and the \ is treated as an escape

themisvaltinos: thanks you’re right updated the test to use as posix to have forward slashes in all platforms

sqlmesh/core/config/scheduler.py
```diff
@@ -105,7 +105,7 @@ def create_state_sync(self, context: GenericContext) -> StateSync:
 
         schema = context.config.get_state_schema(context.gateway)
         return EngineAdapterStateSync(
-            engine_adapter, schema=schema, context_path=context.path, console=context.console
+            engine_adapter, schema=schema, context_path=context.cache_dir, console=context.console
```
erindru: Should the `context_path` parameter be renamed to `cache_dir`?

erindru (commented):
Looks good, this will definitely help in Notebook environments (Fabric in particular) which limit the number of files in the persistent storage.

I notice that `cache_dir` vs `cache_path` is used a bit inconsistently though? Should it be one or the other?

sqlmesh/core/config/scheduler.py
themisvaltinos: yes good point renamed to use the `cache_dir` name here and everywhere else that I was using cache_path for consistency

Issue #4120, opened 2025-04-10
CICD bot should show all details of the `prod` plan in the preview


plaflamme:
Currently, the `Prod Plan Preview` step summary in GitHub shows the diff against prod and lists the indirectly affected models. On the other hand, the `PR Environment Synced` step summary shows the categorization of directly modified models, but only for the change to the virtual environment.

It's currently not possible to know what the categorization for indirectly modified models are for the eventual `prod` plan. That is, we can only see what models are affected, but we cannot (clearly) see that they are indirectly modified non-breaking, breaking or forward-only. Furthermore, the "backfill" list doesn't show the dates that need to be backfilled or whether a model is a view and its view will be recreated (like the cli output shows).

This leads to the situation where it's hard to say for sure what will **actually** happen when a PR is `/deploy`ed. Ideally, the `Prod Plan Preview` would show everything explicitly, that is, list **all** affected model, the categorization of the change for each of them and the list of all backfilled models and the corresponding dates (or that a view will be re-created). Without this, it's difficult to know exactly what will be deployed since there might be many modifications to the virtual environment that the latest `PR Environment Synced` doesn't show.

erindru:
Hey @plaflamme are the screenshots in [my PR](https://github.com/TobikoData/sqlmesh/pull/4872) more like what you were thinking?

plaflamme:
@erindru yes, that's great, thanks!

erindru:
These changes have been released in [v0.200.0](https://github.com/TobikoData/sqlmesh/releases/tag/v0.200.0)

PR #4872, opened 2025-07-02
Feat(cicd_bot): Improve output of 'PR Environment Synced' and 'Prod Plan Preview' steps

erindru:
Addresses #4120 

# Prod Plan Preview
I've updated this step to more closely align with the CLI by showing the change category as well as any indirectly modified children. I've also added an intro text explaining what it's trying to show.

## Before
Prod Plan Preview - Before

## After
Prod Plan Preview - After

# PR Environment Synced
I've updated this step to switch from a tabular view to a list view so that it can more selectively show certain bits of info and more closely align with the CLI. It now shows:
 - A warning about what's causing the PR environment to not contain the full set of data (and thus why a prod deploy wouldnt be a virtual update)
 - More information about the work being done (eg `[full refresh]` / `[recreate view]`)
 - For incremental models, more information about intervals. We show intervals that were loaded as part of the PR environment but also intervals that were *not* loaded, to show that there is a gap that will result in work being done on deploy (as opposed to a pure virtual update)

## Before
PR Environment Synced - Before

## After
PR Environment Synced - After

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -9,11 +9,10 @@ import re
 import traceback
 import typing as t
 from enum import Enum
-from typing import List
 from pathlib import Path
+from dataclasses import dataclass
 
 import requests
-from hyperscript import Element, h
 from sqlglot.helper import seq_get
 
 from sqlmesh.core import constants as c
@@ -21,13 +20,13 @@ from sqlmesh.core.console import SNAPSHOT_CHANGE_CATEGORY_STR, get_console, Mark
 from sqlmesh.core.context import Context
 from sqlmesh.core.test.result import ModelTextTestResult
 from sqlmesh.core.environment import Environment
-from sqlmesh.core.plan import Plan, PlanBuilder
+from sqlmesh.core.plan import Plan, PlanBuilder, SnapshotIntervals
+from sqlmesh.core.plan.definition import UserProvidedFlags
 from sqlmesh.core.snapshot.definition import (
     Snapshot,
     SnapshotChangeCategory,
     SnapshotId,
     SnapshotTableInfo,
-    format_intervals,
 )
 from sqlglot.errors import SqlglotError
 from sqlmesh.core.user import User
@@ -416,6 +415,14 @@ class GithubController:
             return None
 
     @property
+    def pr_plan_flags(self) -> t.Optional[t.Dict[str, UserProvidedFlags]]:
+        if pr_plan := self.pr_plan_or_none:
+            return pr_plan.user_provided_flags
+        if pr_plan_builder := self._pr_plan_builder:
+            return pr_plan_builder._user_provided_flags
+        return None
+
+    @property
     def prod_plan(self) -> Plan:
         if not self._prod_plan_builder:
             self._prod_plan_builder = self._context.plan_builder(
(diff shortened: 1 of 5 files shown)


sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -1115,3 +1107,232 @@ def _chunk_up_api_message(self, message: str) -> t.List[str]:
...
+
+    def _generate_plan_flags_section(
+        self, user_provided_flags: t.Dict[str, UserProvidedFlags]
+    ) -> str:
+        # collapsed section syntax:
+        # https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections#creating-a-collapsed-section
+        section = "<details>\n\n<summary>Plan flags</summary>\n\n"
+        for flag_name, flag_value in user_provided_flags.items():
+            section += f"- `{flag_name}` = `{flag_value}`\n"
+        section += "\n</summary>"
```
themisvaltinos: if I read the html correctly the section here should close with `</details>` instead. since the `</summary>` has already being closed in line 1122

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -1115,3 +1107,232 @@ def _chunk_up_api_message(self, message: str) -> t.List[str]:
...
+    plan: Plan
+
+    @property
+    def snapshot(self) -> Snapshot:
+        if self.is_removed:
+            raise ValueError("Removed snapshots only have SnapshotTableInfo available")
+        return self.plan.snapshots[self.snapshot_id]
+
+    @property
+    def snapshot_table_info(self) -> SnapshotTableInfo:
```
themisvaltinos, line 1208: thinking if it'd make sense to use cached_property instead for properties like this that are frequently accessed

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -1115,3 +1107,232 @@ def _chunk_up_api_message(self, message: str) -> t.List[str]:
...
+            # in-depth interval info is only relevant for incremental models
+            item += f"  **Dates loaded in PR:** [{self.loaded_intervals_rendered}]\n"
+            if self.missing_intervals:
+                item += f"  **Dates *not* loaded in PR:** [{self.missing_intervals_formatted}]\n"
+
+        return item
+
+    def _format_intervals(self, intervals: SnapshotIntervals) -> str:
+        preview_modifier = " (**preview**)" if self.is_dev_preview else ""
+        return f"{intervals.format_intervals(self.snapshot.node.interval_unit)}{preview_modifier}"
```
themisvaltinos, line 1339: here intervals.format_intervals() would never return None right? (if it did this would result in: `None<previewmodifier>` but im guessing from the control flow there will always be intervals here)

erindru: It returns `str` as opposed to `t.Optional[str]`, so if an empty list of intervals was passed then it could theoretically return an empty string.

However as you noticed it's only ever called when we have intervals to display

sqlmesh/integrations/github/cicd/controller.py
erindru: Yeah I always wonder if this is a micro-optimization but I guess this property is called several times during the rendering of `SnapshotSummaryRecord` so it wouldn't hurt

sqlmesh/integrations/github/cicd/controller.py
erindru: Well spotted!

sungchun12:
what does the plan flags toggle do when you click it? screenshot?

sungchun12:
Overall looks great!

erindru:
The plan flags toggle is mostly there for debugging / reasoning about the bot output when people ask questions like "why didn't it show me X" or "why didn't it throw a linter error". It looks like this when expanded:

Screenshot From 2025-07-03 14-38-32

PR #4895, opened 2025-07-03
Chore: Reintroduce tagging queries with correlation ID

VaggelisD:


sqlmesh/core/engine_adapter/base.py
```diff
@@ -147,15 +147,15 @@ def __init__(
...
 
-    def with_settings(self, log_level: int, **kwargs: t.Any) -> EngineAdapter:
+    def with_settings(self, **kwargs: t.Any) -> EngineAdapter:
         adapter = self.__class__(
             self._connection_pool,
             dialect=self.dialect,
             sql_gen_kwargs=self._sql_gen_kwargs,
             default_catalog=self._default_catalog,
-            execute_log_level=log_level,
+            execute_log_level=kwargs.pop("log_level", self._execute_log_level),
```
VaggelisD: `log_level` is not really a required arg anymore but I kept the naming. Do we want to remove it in favor of `execute_log_level`?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -147,15 +147,15 @@ def __init__(
...
             register_comments=self._register_comments,
-            null_connection=True,
+            null_connection=self._extra_config.pop("null_connection", True),
```
VaggelisD: After the first call to`adapter.with_settings`, `null_collection` will end up in `_extra_configs`. So, if we call `adapter.with_settings` from the same adapter again we'd set `null_collection` twice which ends up in an error.

We can also solve that by dragging all the kwargs outside of the constructor call

tests/core/test_table_diff.py
```diff
@@ -337,9 +337,9 @@ def test_generated_sql(sushi_context_fixed_date: Context, mocker: MockerFixture)
 
     # make with_settings() return the current instance of engine_adapter so we can still spy on _execute
     mocker.patch.object(
-        engine_adapter, "with_settings", new_callable=lambda: lambda _: engine_adapter
+        engine_adapter, "with_settings", new_callable=lambda: lambda **kwargs: engine_adapter
     )
-    assert engine_adapter.with_settings(1) == engine_adapter
+    assert engine_adapter.with_settings() == engine_adapter
```
VaggelisD, line 342: This was a dummy `log_level` value which is not required anymore

sqlmesh/core/engine_adapter/base.py
izeigerman: Why did we touch this at all? Why not just keep `True` like it used to be? I don't think we need this change

sqlmesh/core/engine_adapter/base.py
izeigerman: Yes, let's use `execute_log_level`

sqlmesh/core/plan/evaluator.py
```diff
@@ -88,6 +89,9 @@ def evaluate(
         circuit_breaker: t.Optional[t.Callable[[], bool]] = None,
     ) -> None:
         self._circuit_breaker = circuit_breaker
+        self.snapshot_evaluator = self.snapshot_evaluator.set_correlation_id(
```
izeigerman, line 92: I don't see us closing this snapshot evaluator in the `finally` block of the plan evaluator

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1186,6 +1187,18 @@ def _execute_create(
         )
         adapter.execute(snapshot.model.render_post_statements(**create_render_kwargs))
 
+    def set_correlation_id(self, correlation_id: CorrelationId) -> SnapshotEvaluator:
+        return SnapshotEvaluator(
+            {
+                gateway: adapter.with_settings(
+                    log_level=adapter._execute_log_level, correlation_id=correlation_id
```
izeigerman: Why set `log_level`? Shouldn't `with_settings` not override if the argument is not passed?

sqlmesh/core/plan/evaluator.py
VaggelisD: Yeah, that ended up causing problems due to the shared connections, had many failing tests. I imagine the scenario is the following:
- We load a `Context` which initializes all the adapters and `Context::snapshot_evaluator`
- We run `sqlmesh plan` which creates a new `SnapshotEvaluator`, but since it reuses the previous connection pools, all the previous conns are dropped at `finally` 
- If we attempt to reuse `Context`, we reestablish the connections but all the previous results are dropped if e.g we are using DuckDB

sqlmesh/core/engine_adapter/base.py
VaggelisD: If we do:

```Python3

[1] adapter = DuckDbEngineAdapter()

[2] new_adapter = adapter.with_settings(...)

[3] newer_adapter = new_adapter.with_settings(...)
```

At [3] we will get the error:

```Python3
TypeError: sqlmesh.core.engine_adapter.duckdb.DuckDBEngineAdapter() got multiple values for keyword argument 'null_connection'
```

This is because `null_collection` is stored as `_extra_config` in [2] but at [3] it's passed twice, once as hardcoded `True` and another time from `**new_adapter._extra_config`

izeigerman: Ok, sounds good. But then we should not do `self._extra_config.pop` because it mutates the adapter that we're copying.

sqlmesh/core/plan/evaluator.py
izeigerman: That's fine. This means you need to call `.recycle` instead of `close`.

VaggelisD: TIL, applied 👍

PR #4899, opened 2025-07-03
Fix: establish transaction and session before snapshot promotion/demotion

treysp:


erindru (commented):
I was going to comment saying that this needed a test, but I think that updating the failing tests (which are failing because the new calls are unexpected) to now expect the calls should cover this

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -988,40 +992,59 @@ def _promote_snapshot(
...
+                if environment_naming_info.gateway_managed
+                else self.adapter
+            )
+            view_name = snapshot.qualified_view_name.for_environment(
+                environment_naming_info, dialect=adapter.dialect
+            )
+            with (
+                adapter.transaction(),
+                adapter.session(
+                    snapshot.model.render_session_properties(
```
erindru: It looks like these are slightly different to the `render_kwargs` that `_promote_snapshot()` passes to `model.render_session_properties`.

For example, we pass `snapshots` here in `_demote_snapshot()` but not in `_promote_snapshot()`. And `_promote_snapshot()` gets a `table_mapping` property which is absent here.

Do you think these should be consistent so macros defined in `session_properties` can rely on the same things being present at both `RuntimeStage.PROMOTING` and `RuntimeStage.DEMOTING`?

themisvaltinos: in promote the `table_mapping` is mainly to be used for the `on_virtual_update` statements which are not relevant in demotion, but I do agree maybe they should be consistent across both, since we don't know what dynamic macros there might be in `session_properties` potentially

treysp: I was passing `snapshots` based on the `_migrate_snapshot`, but on further inspection I was both doing it wrong and don't think it's necessary.

I removed `snapshots` and added `table_mapping` to demote, so now promote/demote pass the same render args.

sqlmesh/core/plan/evaluator.py
```diff
@@ -382,12 +384,24 @@ def _promote_snapshots(
...
+        snapshots: t.Dict[SnapshotId, Snapshot],
+        deployability_index: t.Optional[DeployabilityIndex] = None,
         on_complete: t.Optional[t.Callable[[SnapshotInfoLike], None]] = None,
     ) -> None:
         self.snapshot_evaluator.demote(
-            target_snapshots, environment_naming_info, on_complete=on_complete
+            target_snapshots,
+            environment_naming_info,
+            snapshots=snapshots,
+            table_mapping=to_view_mapping(
```
izeigerman, line 396: I believe in other places we only pass parent snapshots and not all of them. It'd be annoying to recreate mapping for ALL snapshots every time we demote. This is potentially N^2 when demoting many of them.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -989,40 +999,60 @@ def _promote_snapshot(
...
-        adapter = (
-            self.get_adapter(snapshot.model_gateway)
-            if environment_naming_info.gateway_managed
-            else self.adapter
-        )
-        view_name = snapshot.qualified_view_name.for_environment(
-            environment_naming_info, dialect=adapter.dialect
-        )
-        _evaluation_strategy(snapshot, adapter).demote(view_name)
+        if snapshot.is_model:
```
izeigerman: Can we just do `if not snapshot.is_model: return`? No point in nesting the entire body of the function.

sqlmesh/engines/commands.py
```diff
@@ -49,8 +48,10 @@ class PromoteCommandPayload(PydanticModel):
 
 
 class DemoteCommandPayload(PydanticModel):
-    snapshots: t.List[SnapshotTableInfo]
+    target_snapshots: t.List[Snapshot]
```
izeigerman: Why this changed? This entire module is a leftover from the Airflow integration and needs to be removed. Can you please create a follow-up PR doing that but revert this change in the meantime.

izeigerman: Same goes for the `targets.py` module.

izeigerman: https://github.com/TobikoData/sqlmesh/pull/4924

PR #4900, opened 2025-07-04
Chore!: Update ci/cd bot default behaviour to match CLI behaviour

erindru:
This PR brings the CI/CD bot to partial parity with the CLI in terms of default behaviour, with a path to full parity in future.

Prior to this, by default the CI/CD bot would:
 - Throw an error if the plan had any uncategorized changes, with the assumption that the user had already categorized the changes in their own local environment prior to raising the PR
 - Not backfill any data. This effectively delays the backfill until prod deployment which can make the prod deployment take a long time, rather than doing the backfill work in the PR environment and making the prod deployment a simple virtual update
 
This is in contrast to the CLI, which, when running a command like `sqlmesh plan pr_env` results in a much more streamlined experience as it:
 - Automatically categorizes changes as breaking / non breaking
 - Automatically backfills data into the changed models so that deploying to `prod` can be a virtual update

This PR takes one step towards treating the bot like an automated extension of the CLI.

Internal analytics showed that 97% of users were already overriding `auto_categorize_changes` from the default, so this PR updates the default to be inline with what the documentation was suggesting and inherit the project behaviour (auto-categorize all changes).

However, internal analytics also showed that close to 33% of users were either leaving `skip_pr_backfill` at the default of `true` or not setting it at all.

Rather than break these workflows straight away, if `skip_pr_backfill` is not explicitly set, SQLMesh will produce a warning like so:

```
[WARNING] `skip_pr_backfill` is unset, defaulting it to `true` (no data will be backfilled).
Future versions of SQLMesh will default to `skip_pr_backfill: false` to align with the CLI default behaviour.
If you would like to preserve the current behaviour and remove this warning, please explicitly set `skip_pr_backfill: true` in the bot config.

For more information on configuring the bot, see: https://sqlmesh.readthedocs.io/en/stable/integrations/github/
```

To remove the warning, explicitly set `skip_pr_backfill` to the desired value.

In a future release, we will update the default value to match the CLI after everyone has had a chance to update their config.


sqlmesh/core/config/root.py
```diff
@@ -260,6 +260,18 @@ def _normalize_identifiers(key: str) -> None:
 
         return self
 
+    @model_validator(mode="after")
+    def _inherit_project_config_in_cicd_bot(self) -> Self:
+        if self.cicd_bot:
+            # inherit the project-level settings into the CICD bot if they have not been explicitly overridden
+            if self.cicd_bot.auto_categorize_changes_ is None:
```
erindru, line 267: The documentation claimed that these properties defaulted to the project's config value, however that wasn't working.

This Pydantic validator ensures that the project config is copied down to the bot config if the bot config has not set these fields already.

It needs to happen here because the bot config does not have a reference to the project config to look up these values afterwards

georgesittas (approved):

PR #4903, opened 2025-07-04
Fix: Ensure correct datatypes are fetched for RisingWave dialect

MisterWheatley:
Prior to this PR, the RisingWave dialect used the `BasePostgressAdapter` implementation of `def columns(..)` to fetch column names and data-types.

However the PG catalog tables in RisingWave doesn't support e.g. `struct` types instead returning `???` as datatype. This PR updates the RisingWave adapter to use the [RisingWave catalogs](https://docs.risingwave.com/sql/system-catalogs/rw-catalog) instead to get correct datatypes.

Edit: Also added a integration test file for RisingWave upon request of maintainers for the columns function.


sqlmesh/core/engine_adapter/risingwave.py
```diff
@@ -32,5 +33,39 @@ class RisingwaveEngineAdapter(PostgresEngineAdapter):
...
+        sql = (
+            exp.select("rw_columns.name AS column_name", "rw_columns.data_type AS data_type")
+            .from_("rw_catalog.rw_columns")
+            .join("rw_catalog.rw_relations", on="rw_relations.id=rw_columns.relation_id")
+            .join("rw_catalog.rw_schemas", on="rw_schemas.id=rw_relations.schema_id")
+            .where(
+                exp.and_(
+                    exp.column("name", db="rw_relations").eq(table.alias_or_name),
+                    exp.column("name", db="rw_columns").neq("_row_id"),
+                    exp.column("name", db="rw_columns").neq("_rw_timestamp"),
```
georgesittas: why are we setting `db` here instead of `table`?

sqlmesh/core/engine_adapter/risingwave.py
```diff
@@ -32,5 +33,39 @@ class RisingwaveEngineAdapter(PostgresEngineAdapter):
...
+                )
+            )
+        )
+
+        if table.args.get("db"):
+            sql = sql.where(
+                exp.column("name", db="rw_schemas", quoted=False).eq(table.args["db"].name)
```
georgesittas: Ditto.

tests/core/engine_adapter/test_risingwave.py
```diff
@@ -15,6 +15,43 @@ def adapter(make_mocked_engine_adapter):
     return adapter
 
 
+def test_columns(adapter: t.Callable):
```
georgesittas, line 18: Let's include an integration test for this please, containing the problematic case you described in the description. You can create a `test_integration_risingwave`, similar to other integration testing modules.

sqlmesh/core/engine_adapter/risingwave.py
MisterWheatley: The above is directly copied from https://github.com/TobikoData/sqlmesh/blob/61455f2ac06866ead19aefb8759751dbad0f38ce/sqlmesh/core/engine_adapter/base_postgres.py#L50

I was following the example set in the `BasePostgresAdapter` by using `db`

tests/core/engine_adapter/test_risingwave.py
MisterWheatley: I'll try and get started on this next week.

georgesittas: Thanks!

sqlmesh/core/engine_adapter/risingwave.py
georgesittas: Notice how we only access it for `table`, though. For columns, i.e. objects constructed via `column`, a reference like `a.b` produces a `b` for `this` and `a` for `table`. The `db` part would come _before_ `a`, and similarly the `catalog` part would come before the `db` part.

MisterWheatley: Thank you for the clarification. I will update next week to use `table` over `db`. 

I guess I was just blindly following the existing code, but did consider whether it was due to some interaction with Postgres as a single catalog engine. It seemed to work so I didn't question it too deeply.

sqlmesh/core/engine_adapter/risingwave.py
MisterWheatley: Fixed in https://github.com/TobikoData/sqlmesh/pull/4903/commits/9c4a9caf2af29dda1d8d9e3ae9ae857f13e545d9

sqlmesh/core/engine_adapter/risingwave.py
MisterWheatley: Fixed in https://github.com/TobikoData/sqlmesh/pull/4903/commits/9c4a9caf2af29dda1d8d9e3ae9ae857f13e545d9

tests/core/engine_adapter/test_risingwave.py
MisterWheatley: Added my attempt at a integration test in https://github.com/TobikoData/sqlmesh/pull/4903/commits/c19a86f015c44ebff3cc52c9fa15c9d1f1694723

It runs locally on my dev setup when running `pytest tests/core/engine_adapter/integration/test_integration_risingwave.py` but let me know if you have feedback on the test.

I took some inspiration from the other adapter integration tests, but I am not that familiar with pytest (yet), so please do let me know if something needs to be changed so that I can learn.

sqlmesh/core/engine_adapter/risingwave.py
```diff
@@ -32,5 +33,39 @@ class RisingwaveEngineAdapter(PostgresEngineAdapter):
...
+            .where(
+                exp.and_(
+                    exp.column("name", table="rw_relations").eq(table.alias_or_name),
+                    exp.column("name", table="rw_columns").neq("_row_id"),
+                    exp.column("name", table="rw_columns").neq("_rw_timestamp"),
+                )
+            )
+        )
+
+        if table.args.get("db"):
```
georgesittas:
```suggestion
        if table.db:
```

sqlmesh/core/engine_adapter/risingwave.py
```diff
@@ -32,5 +33,39 @@ class RisingwaveEngineAdapter(PostgresEngineAdapter):
...
+            sql = sql.where(
+                exp.column("name", table="rw_schemas", quoted=False).eq(table.args["db"].name)
+            )
```
georgesittas:
```suggestion
            sql = sql.where(exp.column("name", table="rw_schemas").eq(table.db))
```

tests/core/engine_adapter/integration/test_integration_risingwave.py
```diff
@@ -0,0 +1,87 @@
...
+        f"NULL::{data_type.sql(dialect='risingwave')} AS {col_name}"
+        for col_name, data_type in risingwave_columns_with_datatypes.items()
+    ]
+    query: exp.Query = exp.maybe_parse(
+        f"""
+        SELECT
+          {",".join(query_cols)}
+        """,
+        dialect="risingwave",
+    )
```
georgesittas:
```suggestion
    query = exp.select(
        *[
            exp.cast(exp.null(), dtype).as_(name)
            for name, dtype in risingwave_columns_with_datatypes.items()
        ]
    )
```

georgesittas (approved):
LGTM– thanks for addressing comments! A few more minor ones from me and the PR should be good to merge.

sqlmesh/core/engine_adapter/risingwave.py
```diff
@@ -32,5 +33,37 @@ class RisingwaveEngineAdapter(PostgresEngineAdapter):
...
+
+        if table.db:
+            sql = sql.where(exp.column("name", table="rw_schemas").eq(table.db))
+
+        self.execute(sql)
+        resp = self.cursor.fetchall()
+        if not resp:
+            raise SQLMeshError(f"Could not get columns for table {table_name}. Table not found.")
+        return {
+            column_name: exp.DataType.build(data_type, dialect=self.dialect, udt=True)
```
erindru, line 64: I had to check what `udt=True` does but it looks like we still try to parse into a known datatype and only use a user-defined type as a fallback, so this LGTM :+1:

erindru (approved):

PR #4905, opened 2025-07-04
Fix: parse runtime-rendered fields, extract python env deps from merge_filter

georgesittas:
While iterating on a separate task locally, I observed that there are a couple of gaps in our "runtime-rendered" property handling logic:

- We don't parse Python strings corresponding to said properties. This means we don't analyze them to extract referenced variables, which can lead to issues at runtime (e.g., cadence runs), due to these variables not being present in `python_env`, and hence failing to render.
- We don't analyze `merge_filter` for variables. This is true for both Python and SQL models. Again, this means that if a variable is referenced in `merge_filter`, it may not be present in the `python_env` of the model and hence we may fail to render it at runtime.

This PR addresses both of the issues above and expands our testing coverage to ensure we properly detect the referenced variables in runtime-rendered properties.

I'm not sure yet if this needs a migration script, because if someone references variables in said properties today, then cadence runs would've failed for them anyway if the variables weren't present in the `python_env`. I'd like some eyes on this to double-check my reasoning.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -77,8 +77,6 @@ PROPERTIES = {"physical_properties", "session_properties", "virtual_properties"}
 RUNTIME_RENDERED_MODEL_FIELDS = {
     "audits",
     "signals",
-    "description",
-    "cron",
     "merge_filter",
 } | PROPERTIES
 
@@ -2469,6 +2467,9 @@ def _create_model(
         if isinstance(property_values, exp.Tuple):
             statements.extend(property_values.expressions)
 
+    if isinstance(getattr(kwargs.get("kind"), "merge_filter", None), exp.Expression):
+        statements.append(kwargs["kind"].merge_filter)
+
     jinja_macro_references, used_variables = extract_macro_references_and_variables(
         *(gen(e if isinstance(e, exp.Expression) else e[0]) for e in statements)
     )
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -77,8 +77,6 @@
 RUNTIME_RENDERED_MODEL_FIELDS = {
     "audits",
     "signals",
-    "description",
-    "cron",
```
georgesittas, line 81: These were added some time ago to avoid parsing & rendering things like `cron="@daily"`, or `description="... @bar ..."`, where `@bar` is just a literal substring of the description, not a variable reference.

I'm not sure if this is the right approach, because we may have more fields with "string literals", where we don't care about rendering variables. I decided to simply preserve what we do today to avoid increasing the scope, but this is probably worth discussing & addressing in a followup PR.

themisvaltinos (approved):
looks good to me, the only concern is regarding the need of a migration script, but I think your theory is correct, since if someone had macro vars in the `merge_filter` before they would have failed to be rendered

sqlmesh/core/model/definition.py
```diff
@@ -2749,23 +2750,37 @@ def render_field_value(value: t.Any) -> t.Any:
 
         return value
 
+    def parse_strings_with_macro_refs(value: t.Any) -> t.Any:
```
izeigerman: I feel like this function needs to be in utils somewhere

sqlmesh/core/model/definition.py
```diff
@@ -2749,23 +2750,37 @@ def render_field_value(value: t.Any) -> t.Any:
...
+            value = [parse_strings_with_macro_refs(v) for v in value]
+
+        return value
+
     for field_name, field_info in ModelMeta.all_field_infos().items():
         field = field_info.alias or field_name
+        field_value = fields.get(field)
 
-        if field in RUNTIME_RENDERED_MODEL_FIELDS:
+        if field in ("cron", "description") or field_value is None:
```
izeigerman: We won't find this next time we introduce a similar field. Should we have a new named constant like `UNRENDERABLE_MODEL_FIELDS`

sqlmesh/core/model/definition.py
```diff
@@ -2749,23 +2750,37 @@ def render_field_value(value: t.Any) -> t.Any:
...
             continue
 
-        field_value = fields.get(field)
-        if field_value is None:
+        if field in RUNTIME_RENDERED_MODEL_FIELDS:
+            fields[field] = parse_strings_with_macro_refs(field_value)
```
izeigerman: I'm somewhat confused by how these runtime-rendered fields (signals, audits, etc) might end up being strings? Aren't they always expressions?

georgesittas: The issue appears with python models, where you could encode values as strings that are later parsed during validation.

sqlmesh/core/model/definition.py
georgesittas: I'm considering moving it under `models/common.py` since we have some similar converters / validators there + given we don't have a relevant file under `utils/`

sqlmesh/core/model/definition.py
georgesittas: Amended test and got rid of `description`, so just comparing to `cron` here now. Left a comment to explain why.

sqlmesh/core/model/definition.py
georgesittas: Done, now lives under `common.py`

tests/core/test_model.py
```diff
@@ -6436,7 +6436,7 @@ def end_date_macro(evaluator: MacroEvaluator, var: bool):
         owner="@IF(@gateway = 'dev', @{dev_owner}, @{prod_owner})",
         stamp="@{stamp}",
         tags=["@{tag1}", "@{tag2}"],
-        description="Model desc @{test_}",
+        description="'Model desc @{test_}'",
```
izeigerman, line 6439: This probably makes it a breaking change

PR #4908, opened 2025-07-04
Fix!: Always recreate materialized view

VaggelisD:
Materialized views in engines like BigQuery and Snowflake are invalidated if the underlying table(s) are replaced. Toy repro scenario:

```SQL
1. CREATE TABLE tbl AS (SELECT 1 AS col);

2. CREATE MATERIALIZED VIEW mview AS (SELECT * FROM tbl);

3. CREATE OR REPLACE TABLE tbl AS (SELECT 2 AS col);

4. SELECT * FROM mview;
```

<br />

This flow can be mirrored in the following toy SQLMesh project:

```SQL
# Full model
MODEL (name example_schema.t1, kind FULL); SELECT 1 AS col;

# Materialized view
MODEL (
  name example_schema.t2,
  kind VIEW (
    materialized true
  )
);

SELECT * FROM example_schema.t1;

```

<br />

Running the very first plan will yield the following error in either engine, e.g:
```
google.api_core.exceptions.BadRequest: 400 Materialized view <t2> references table <t1> which was deleted and recreated. The view must be deleted and recreated as well
```

<br />

This is because, during the plan:
1. The physical table for `t1` is created as a dry run
2. The materalized view for `t2` is created as a dry run
3. The physical table for `t1` is recreated/replaced as part of the `FULL` insertion strategy
4. The materialized view exists, and based on the pre-existing `ViewStrategy` it is not recreated

This PR alters (4) such that any materialized view with upstream dependencies is recreated from scratch in order to restore correctness.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -1876,6 +1876,9 @@ class ViewStrategy(PromotableStrategy):
         )
         snapshot = kwargs["snapshot"]
         snapshots = kwargs["snapshots"]
+
+        # We must replace a materialized view if its query has changed or if it depends on a table.
+        # The latter is because if the underlying table is replaced, the materialized view is invalidated in some of the engines.
         if (
             (
                 isinstance(query_or_df, exp.Expression)
@@ -1887,6 +1890,7 @@ class ViewStrategy(PromotableStrategy):
                     engine_adapter=self.adapter,
                 )
                 == query_or_df
+                and not model.depends_on
             )
             or self.adapter.HAS_VIEW_BINDING
         ) and self.adapter.table_exists(table_name):
--- a/tests/core/engine_adapter/integration/test_integration_bigquery.py
+++ b/tests/core/engine_adapter/integration/test_integration_bigquery.py
@@ -400,3 +400,50 @@ def test_table_diff_table_name_matches_column_name(ctx: TestContext):
 
     assert row_diff.stats["join_count"] == 1
     assert row_diff.full_match_count == 1
+
+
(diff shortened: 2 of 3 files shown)


georgesittas (approved):

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1887,6 +1890,7 @@ def insert(
                     engine_adapter=self.adapter,
                 )
                 == query_or_df
+                and not model.depends_on
```
izeigerman: I wonder what's the point of this check and whether we should just remove this whole condition and re-create the materialized view always unconditionally. When does the model have no upstream dependencies (and it's not a seed)?

VaggelisD: Well, this whole condition ended up being a micro optimization i.e if the MV was a pure `SELECT`, but that would not make much sense I guess. Will completely remove it

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1876,20 +1876,12 @@ def insert(
...
-                and model.render_query(
-                    snapshots=snapshots,
-                    deployability_index=deployability_index,
-                    engine_adapter=self.adapter,
-                )
-                == query_or_df
-            )
-            or self.adapter.HAS_VIEW_BINDING
-        ) and self.adapter.table_exists(table_name):
+            not snapshot.is_materialized_view
```
izeigerman, line 1881: Previously we were skipping *only* for materialized views and never for regular ones. This changes it. I think we should remove the whole skipping logic up to and including line 1886.

izeigerman: Never mind, I was wrong

izeigerman (approved):

PR #4912, opened 2025-07-04
feat: add named secret support DuckDB

eakmanrq:
Prior to this PR, users were forced to use the default secret name when creating secrets. Now they can still do that or create secrets with a specific name. This is required to support a use case where users want to define multiple secrets for a single type (ex: multiple secrets for S3) since before they would get an error saying the default name was already defined.

### DIFF
--- a/docs/integrations/engines/duckdb.md
+++ b/docs/integrations/engines/duckdb.md
@@ -10,15 +10,15 @@
 
 ### Connection options
 
-| Option             | Description                                                                                                                                                                                                                                     |  Type  | Required |
-|--------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:------:|:--------:|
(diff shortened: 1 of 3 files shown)


docs/integrations/engines/duckdb.md
```diff
@@ -273,9 +351,9 @@ The `filesystems` accepts a list of file systems to register in the DuckDB conne
           type: duckdb
           catalogs:
             ducklake:
-            type: ducklake
-            path: myducklakecatalog.duckdb
-            data_path: abfs://MyFabricWorkspace/MyFabricLakehouse.Lakehouse/Files/DuckLake.Files
+              type: ducklake
+              path: myducklakecatalog.duckdb
+              data_path: abfs://MyFabricWorkspace/MyFabricLakehouse.Lakehouse/Files/DuckLake.Files
```
eakmanrq, line 356: This is a typo I found while reviewing the docs after the change in this PR

PR #4915, opened 2025-07-07
Fix(cicd_bot): Don't truncate backfill model list

erindru:
Related to #4120 

Currently, if there are lots of models to backfill, we "helpfully" truncate the list in the `Prod Plan Preview` step:
01 - Prod Plan Preview -  Before, truncated

However, this is undesirable in the CI/CD bot because there is no equivalent of the `-vv` option to set the output verbosity. This means anyone reviewing the bot output has no way of actually knowing the full extent of the models to be backfilled.

This PR updates the CI/CD bot to always show the full list of models affected:
01 - Prod Plan Preview -  After

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -486,7 +486,7 @@ class GithubController:
 
     def get_plan_summary(self, plan: Plan) -> str:
         # use Verbosity.VERY_VERBOSE to prevent the list of models from being truncated
-        # this is particularly importanmt for the "Models needing backfill" list because
+        # this is particularly important for the "Models needing backfill" list because
         # there is no easy way to tell this otherwise
         orig_verbosity = self._console.verbosity
         self._console.verbosity = Verbosity.VERY_VERBOSE
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -485,6 +485,12 @@ class GithubController:
                 print(f"{key}={value}", file=fh)
 
     def get_plan_summary(self, plan: Plan) -> str:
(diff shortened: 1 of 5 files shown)


tests/integrations/github/cicd/test_integration.py
```diff
@@ -298,7 +298,7 @@ def test_merge_pr_has_non_breaking_change(
     assert GithubCheckConclusion(prod_plan_preview_checks_runs[2]["conclusion"]).is_success
 
     expected_prod_plan_directly_modified_summary = """**Directly Modified:**
-* `sushi.waiter_revenue_by_day` (Non-breaking)
+* `memory.sushi.waiter_revenue_by_day` (Non-breaking)
```
erindru, line 301: This is a side effect of setting `Verbosity.VERY_VERBOSE`.

However, it does have the effect of making the output of this step more closely align with the `PR Environment Synced` step which is using fully qualified names already

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -485,6 +485,12 @@ def _append_output(cls, key: str, value: str) -> None:
                 print(f"{key}={value}", file=fh)
 
     def get_plan_summary(self, plan: Plan) -> str:
+        # use Verbosity.VERY_VERBOSE to prevent the list of models from being truncated
+        # this is particularly importanmt for the "Models needing backfill" list because
```
georgesittas:
```suggestion
        # this is particularly important for the "Models needing backfill" list because
```

georgesittas (approved):

sqlmesh/integrations/github/cicd/controller.py
erindru: This is how you know I wrote the code and not an AI agent

georgesittas: Haha fair enough, I respect that 😄

PR #4925, opened 2025-07-07
fix!: make repr deterministic for fingerprinting

eakmanrq:
Prior to this PR, models could show changes to `__sqlmesh__vars__` even though there is not an actual change to the model. Looking at the code and it appears that the dictionary is not deterministic so this PR aims to address that shortcoming.

### DIFF
--- a/sqlmesh/migrations/v0085_deterministic_repr.py
+++ b/sqlmesh/migrations/v0085_deterministic_repr.py
@@ -3,13 +3,21 @@ When serializing some objects, like `__sqlmesh__vars__`, the order of keys in th
 and therefore this migration applies deterministic sorting to the keys of the dictionary.
 """
 
-import ast
 import json
 import typing as t
+from dataclasses import dataclass
 
 from sqlglot import exp
 
 from sqlmesh.utils.migration import index_text_type, blob_text_type
+
+
+# Make sure `SqlValue` is defined so it can be used by `eval` call in the migration
+@dataclass
+class SqlValue:
+    """A SQL string representing a generated SQLGlot AST."""
+
+    sql: str
 
 
 def _deterministic_repr(obj: t.Any) -> str:
@@ -27,7 +35,10 @@ def _deterministic_repr(obj: t.Any) -> str:
             return type(o)(normalized)
         return o
 
-    return repr(_normalize_for_repr(obj))
+    try:
+        return repr(_normalize_for_repr(obj))
+    except Exception:
+        return repr(obj)
 
 
 def migrate(state_sync, **kwargs):  # type: ignore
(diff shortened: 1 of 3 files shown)


sqlmesh/utils/metaprogramming.py
```diff
@@ -633,6 +635,40 @@ def print_exception(
...
+    Returns:
+        A deterministic string representation of the object.
+    """
+
+    def _normalize_for_repr(o: t.Any) -> t.Any:
+        if isinstance(o, dict):
+            # Create a new dict with sorted keys, handling mixed types by using string representation for sorting
+            try:
+                sorted_items = sorted(o.items())
+            except TypeError:
```
georgesittas: Should we catch the base `Exception` class here instead? I'm worried `TypeError` may be too specific. To me, we should ideally try to normalize and if it fails for whatever reason just fall back to the previous behavior.

eakmanrq: Yeah was talking to Iaroslav about this too and it seems like we can just always do the fallback behavior

sqlmesh/migrations/v0085_deterministic_repr.py
```diff
@@ -0,0 +1,140 @@
...
+        parsed_snapshot = json.loads(snapshot)
+        python_env = parsed_snapshot["node"].get("python_env")
+
+        if python_env:
+            for key, executable in python_env.items():
+                if isinstance(executable, dict) and executable.get("kind") == "value":
+                    old_payload = executable["payload"]
+                    try:
+                        # Try to parse the old payload and re-serialize it deterministically
+                        parsed_value = ast.literal_eval(old_payload)
```
georgesittas: This doesn't look right– we should mimic the `value` evaluation logic and use `eval` instead ([source](https://github.com/TobikoData/sqlmesh/blob/c9c1429335f055d6adf10d60f9c60557d963f407/sqlmesh/utils/metaprogramming.py#L531-L532)). You will also need to define `SqlValue` ([source](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/utils/metaprogramming.py#L382-L386)) in the migration script, so that it's present in `eval`'s `globals` context. If we do this, I believe we should be able to get rid of the complicated try-except logic.

sqlmesh/migrations/v0085_deterministic_repr.py
```diff
@@ -0,0 +1,140 @@
...
+            "unrestorable",
+        ).from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        python_env = parsed_snapshot["node"].get("python_env")
+
+        if python_env:
+            for key, executable in python_env.items():
+                if isinstance(executable, dict) and executable.get("kind") == "value":
```
georgesittas, line 85: What about lists and tuples that contain nested dicts?

sqlmesh/utils/metaprogramming.py
```diff
@@ -633,6 +635,35 @@ def print_exception(
...
+        if isinstance(o, dict):
+            sorted_items = sorted(o.items(), key=lambda x: str(x[0]))
+            return {k: _normalize_for_repr(v) for k, v in sorted_items}
+        if isinstance(o, (list, tuple)):
+            # Recursively normalize nested structures
+            normalized = [_normalize_for_repr(item) for item in o]
+            return type(o)(normalized)
+        return o
+
+    return repr(_normalize_for_repr(obj))
```
georgesittas: Already discussed in Slack, but sharing here for vis. – I suggest wrapping this whole thing in a try-except and ensuring we have a safe fall back to the previous logic if normalization fails for whatever reason.

sqlmesh/migrations/v0085_deterministic_repr.py
eakmanrq: The `_deterministic_repr` is just being applied to executable value kinds: https://github.com/TobikoData/sqlmesh/blob/67b935624fbac7c92f0aead11b92967c0283cf96/sqlmesh/utils/metaprogramming.py#L426-L430

That is why this if condition is so restrictive. Nested lists/tuples are handled in the `_deterministic_repr` implementation. Does that address your concern?

sqlmesh/migrations/v0085_deterministic_repr.py
eakmanrq: Thanks for the suggestion. Commit: [`b6d5b68` (#4925)](https://github.com/TobikoData/sqlmesh/pull/4925/commits/b6d5b68cd4119b968a6369ba078bdf36870b592a)

sqlmesh/migrations/v0085_deterministic_repr.py
georgesittas: Ah, I see, so you're saying that the "top-level" value is always a dict because this only concerns variables?

I was thinking of an executable constructed like `Executable.value([1, 2, 3])`, but if this doesn't represent an actual payload we may encounter then feel free to ignore.

PR #4930, opened 2025-07-07
fix!: Add DATETIMEOFFSET handling in pyodbc for MSSQL

mattiasthalen:
This pull request introduces enhancements to MSSQL connection handling, particularly for the `pyodbc` driver, and updates dependency specifications for `pyodbc` across the project. The most significant changes include adding support for MSSQL's `DATETIMEOFFSET` type, improving test coverage for this feature, and ensuring the `pyodbc` version is explicitly specified to avoid compatibility issues.

### MSSQL Connection Enhancements:
* [`sqlmesh/core/config/connection.py`](diffhunk://#diff-95f5a476d8642ba1937252ee40000b28f73c881331613e66abd63dac92514416L1627-R1652): Added support for MSSQL's `DATETIMEOFFSET` type by implementing a custom output converter at the connection level for `pyodbc`. The converter handles binary data for the `DATETIMEOFFSET` SQL type (-155) and converts it into Python's `datetime` objects with timezone information.

### Test Coverage Improvements:
* [`tests/core/test_connection_config.py`](diffhunk://#diff-417bffd57708967673f0a4aeba0ab09672d30d5221fcb14a11f7bebc72fbef9fR1524-R1662): Added tests to validate the `DATETIMEOFFSET` handling for `pyodbc`. These include scenarios for positive and negative timezone offsets, ensuring proper conversion of binary data to `datetime` objects. Additionally, confirmed that `_cursor_init` is no longer required for `pymssql` due to the connection-level setup.

### Dependency Updates:
* [`pyproject.toml`](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L43-R43): Updated the `pyodbc` dependency to require version `>=5.0.0` across multiple optional dependency groups (`azuresql-odbc`, `mssql-odbc`, and `dev`) to ensure compatibility with the new features. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L43-R43) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L81-R81) [[3]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L111-R111)

### DIFF
--- a/tests/core/test_connection_config.py
+++ b/tests/core/test_connection_config.py
@@ -1683,16 +1683,3 @@ def test_mssql_pyodbc_connection_negative_timezone_offset():
         expected_dt = datetime(2023, 1, 1, 12, 0, 0, 0, timezone(timedelta(hours=-8, minutes=0)))
         assert result == expected_dt
         assert result.tzinfo == timezone(timedelta(hours=-8))
-
-
-def test_mssql_no_cursor_init_for_pymssql():
-    """Test that _cursor_init is not needed for pymssql driver."""
-    config = MSSQLConnectionConfig(
-        host="localhost",
-        driver="pymssql",
-        check_import=False,
-    )
-
-    # Since we moved output converter setup to connection level,
-    # there's no cursor init needed for any driver
-    assert not hasattr(config, "_cursor_init") or config._cursor_init is None


sqlmesh/core/config/connection.py
```diff
@@ -1520,6 +1520,41 @@ def _mssql_engine_import_validator(cls, data: t.Any) -> t.Any:
...
+    def _cursor_init(self) -> t.Optional[t.Callable[[t.Any], None]]:
+        """Initialize the cursor with output converters for MSSQL-specific data types."""
+        # Only apply pyodbc-specific cursor initialization when using pyodbc driver
+        if self.driver != "pyodbc":
+            return None
+
+        def init(cursor: t.Any) -> None:
+            # Get the connection from the cursor and set the output converter
+            conn = cursor.connection
+            if hasattr(conn, "add_output_converter"):
```
erindru: Since these are meant to be attached to connections and not cursors, wouldnt line 1662 be the place to do it?

```
return pyodbc.connect(conn_str, autocommit=kwargs.get("autocommit", False))
```
vs
```
conn = pyodbc.connect(conn_str, autocommit=kwargs.get("autocommit", False))
conn.add_output_converter(...)
return conn
```

georgesittas (approved):
This looks reasonable to me, @mattiasthalen.

Just double-checking, but this is backwards-compatible, yes? In the sense that `DATETIMEOFFSET` would previously lead to an error, if pyodbc was used?

mattiasthalen:
Yes, without it PYODBC would throw an error about data type -155 ☺️

tests/core/test_connection_config.py
```diff
@@ -1521,3 +1521,168 @@ def test_mssql_pymssql_connection_factory():
...
+
+        result = converter_func(binary_data)
+
+        expected_dt = datetime(2023, 1, 1, 12, 0, 0, 0, timezone(timedelta(hours=-8, minutes=0)))
+        assert result == expected_dt
+        assert result.tzinfo == timezone(timedelta(hours=-8))
+
+
+def test_mssql_pyodbc_connection_no_add_output_converter():
+    """Test that connection gracefully handles pyodbc without add_output_converter."""
```
erindru: FMI: when does the `Connection` object returned by `pyodbc` not have an `add_output_converter` method on it?

Otherwise, LGTM :+1:

mattiasthalen: Not available in pyodbc < 4.0.22. But perhaps it's better to pin the version? ☺️ and besides, 4.0.21 is  almost 8 years old, it's on 5.2.0 now 😅

erindru: In that case, I think its reasonable to require `pyodbc >= 5.0.0` (released october 2023) in our dependencies and then we can remove all the `hasattr` checks and also not have to worry about any breaking changes from 4.x to 5.x

mattiasthalen: I will make it happen... tomorrow.. today? It's 02:00 here, need some sleep ☺️

erindru: Sure, no rush! Thanks for bearing with us!

georgesittas (approved):
I'll mark this as breaking due to the version constraint.

tests/core/test_connection_config.py
```diff
@@ -1521,3 +1521,142 @@ def test_mssql_pymssql_connection_factory():
...
+        )
+
+        result = converter_func(binary_data)
+
+        expected_dt = datetime(2023, 1, 1, 12, 0, 0, 0, timezone(timedelta(hours=-8, minutes=0)))
+        assert result == expected_dt
+        assert result.tzinfo == timezone(timedelta(hours=-8))
+
+
+def test_mssql_no_cursor_init_for_pymssql():
```
erindru: I think this test is kind of redundant now and can be removed?

mattiasthalen: Sorry, missed that one

erindru (approved):
Great work, thanks!

This ended up in a much simpler place than where we started

PR #4933, opened 2025-07-08
feat: ability to find a model block

benfdking:
- added the use of it in a builtin


sqlmesh/core/linter/helpers.py
```diff
@@ -113,3 +114,40 @@ def read_range_from_file(file: Path, text_range: Range) -> str:
...
+    dialect: str,
+) -> t.Optional[Range]:
+    """
+    Get the range of the model block in an SQL file.
+    """
+    tokens = tokenize(sql, dialect=dialect)
+
+    # Find start of the model block
+    start = next(
+        (t for t in tokens if t.token_type is TokenType.VAR and t.text == "MODEL"),
```
```suggestion
        (t for t in tokens if t.token_type is TokenType.VAR and t.text.upper() == "MODEL", None),
```

sqlmesh/core/linter/helpers.py
```diff
@@ -113,3 +114,40 @@ def read_range_from_file(file: Path, text_range: Range) -> str:
...
+    )
+    end = next(t for t in tokens if t.token_type is TokenType.SEMICOLON)
```
```suggestion
    end = next(t for t in tokens if t.token_type is TokenType.SEMICOLON, None)
```

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -93,7 +93,25 @@ class NoMissingAudits(Rule):
     """Model `audits` must be configured to test data quality."""
 
     def check_model(self, model: Model) -> t.Optional[RuleViolation]:
-        return self.violation() if not model.audits and not model.kind.is_symbolic else None
+        if model.audits or model.kind.is_symbolic:
+            return None
+        if (
+            not isinstance(model, SqlModel)
```
georgesittas: Do we need this? What about seed models? Perhaps the suffix check is fine.

georgesittas (approved):

PR #4935, opened 2025-07-08
Fix: Do not attach `correlation_id` for Athena queries

VaggelisD:
Athena does not seem to deal with comments well, e.g:

```Python3
>>> self._execute('/* test */ DESCRIBE foo')
pyathena.error.OperationalError: FAILED: ParseException line 1:0 cannot recognize input near '/' '*' 'test'
```

This can cause `base.py::table_exists()` to fail silently (due to the try/catch) and thus not drop tables when needed.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -336,10 +336,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          # filters:
-          #   branches:
-          #     only:
-          #       - main
+          filters:
+            branches:
+              only:
+                - main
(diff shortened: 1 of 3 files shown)


.circleci/continue_config.yml
```diff
@@ -336,10 +336,10 @@ workflows:
...
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
+          #   branches:
+          #     only:
+          #       - main
```
VaggelisD: TODO: Reinstate before merging

PR #4936, opened 2025-07-08
Fix!: mark vars referenced in metadata macros as metadata

georgesittas:
Macro variable references are always treated as non-metadata today. This means that if, for example, a variable is referenced within a metadata-only macro, changing its value will result in a breaking change, which is inconsistent.

This PR alters this behavior, similar to the [macro metadata-only status propagation](https://github.com/TobikoData/sqlmesh/commit/fc7fe1bd0f42aa6aa6b4ca880075d18e21c591da):

- Variables referenced within metadata-only macro _definitions_ can be treated as metadata-obly
- Variables referenced in metadata-only macro _calls_ can be treated as metadata-only
- Variables referenced within metadata properties can be treated as metadata-only

I intentionally say "can" instead of "will" above, because we need to factor in all references of a variable to decide whether it's a metadata-only reference. The rules implemented here are similar to those we apply for macros: a non-metadata occurrence overrules all metadata occurrences.

Additionally, this PR introduces trimming for blueprint variables. Certain blueprint variables, e.g. used in model names, aren't required after loading, while others are because they may be referenced in the model's statements or in "runtime-rendered" properties (e.g., `merge_filter`).

The former category can be omitted from the model's `python_env`, thus reducing its snapshot's size, as long as a variable is only referenced in the meta block and in fields that are static after loading the model.

Both of these changes are quite breaking, so I'm planning to implement a migration script to at least warn about this. I'm also planning to increase the testing coverage.


sqlmesh/core/model/common.py
```diff
@@ -126,43 +148,107 @@ def make_python_env(
...
 
 
+def _extract_macro_func_variable_references(macro_func: exp.Expression) -> t.Set[str]:
+    references = set()
+
+    # Don't descend into nested MacroFunc nodes besides @VAR() and @BLUEPRINT_VAR(), because
+    # they will be handled in a separate call of _extract_macro_func_variable_references.
+    def _prune_nested_macro_func(expression: exp.Expression) -> bool:
+        return (
+            type(n) is d.MacroFunc
```
themisvaltinos: shouldn't here these variables all be `expression` instead of `n` in `_prune_nested_macro_func `?

sqlmesh/core/model/common.py
```diff
@@ -126,43 +148,107 @@ def make_python_env(
...
+        )
 
     if blueprint_variables:
+        metadata_blueprint_variables = {
+            k: SqlValue(sql=v.sql(dialect=dialect)) if isinstance(v, exp.Expression) else v
+            for k, v in blueprint_variables.items()
+            if k in metadata_used_variables
+        }
         blueprint_variables = {
             k: SqlValue(sql=v.sql(dialect=dialect)) if isinstance(v, exp.Expression) else v
```
themisvaltinos: I imagine in a subsequent pr we'll turn these keys to lower()

sqlmesh/core/model/common.py
```diff
@@ -126,43 +148,107 @@ def make_python_env(
...
-    used_variables = (used_variables or set()) | python_used_variables
+    for var_name, is_metadata in python_used_variable_referenced_in_metadata_expression.items():
+        used_variable_referenced_in_metadata_expression[var_name] = (
+            used_variable_referenced_in_metadata_expression.get(var_name, True) and is_metadata
+        )
+
+    metadata_used_variables = set()
+    for used_var, macro_names in (macro_funcs_by_used_var or {}).items():
+        if used_variable_referenced_in_metadata_expression.get(used_var) or all(
+            name in python_env and python_env[name].is_metadata for name in macro_names
```
themisvaltinos: is maybe this needed for is_metadata: `name in python_env and getattr(python_env.get(name), 'is_metadata', False)` to handle none or `all` handles nones just fine?

sqlmesh/core/model/common.py
```diff
@@ -77,16 +82,27 @@ def make_python_env(
...
                         )
-                    used_variables.add(args[0].this.lower())
+
+                    var_name = args[0].this.lower()
+                    used_variable_referenced_in_metadata_expression[var_name] = (
+                        used_variable_referenced_in_metadata_expression.get(var_name, True)
+                        and bool(is_metadata)
+                    )
+                else:
+                    for var_ref in _extract_macro_func_variable_references(macro_func_or_var):
```
themisvaltinos: I wonder if the complexity because of this nested walk with `_extract_macro_func_variable_references` inside the loop which uses `find_all` is too much and can be reduced

sqlmesh/core/model/common.py
```diff
@@ -126,43 +148,107 @@ def make_python_env(
...
         blueprint_variables=blueprint_variables,
     )
-    used_variables = (used_variables or set()) | python_used_variables
+    for var_name, is_metadata in python_used_variable_referenced_in_metadata_expression.items():
+        used_variable_referenced_in_metadata_expression[var_name] = (
+            used_variable_referenced_in_metadata_expression.get(var_name, True) and is_metadata
+        )
+
+    metadata_used_variables = set()
+    for used_var, macro_names in (macro_funcs_by_used_var or {}).items():
```
themisvaltinos: this might need a small comment over it because it took a while to figure out what was happening, also a question because I'm still not entirely certain if I read this correctly:  this `used_variable_referenced_in_metadata_expression.get(used_var)` is False when it's non-metadata and it won't be added to `metadata_used_variables`  _unless_ all macros using it are metadata-only?

sqlmesh/core/model/common.py
georgesittas: Great catch, thank you! This was originally a lambda that used `n` for the argument name, hence how it slipped through the cracks. Interestingly enough, using `n` works too, because it binds to the local `n` loop variable.

sqlmesh/core/model/common.py
georgesittas: Yes, I think this can be done in a followup PR. I'd like to decouple it from this one, as it's already quite big.

sqlmesh/core/model/common.py
georgesittas: `python_env[name]` shouldn't ever result in `None`, see `python_env`'s type hint.

sqlmesh/core/model/common.py
georgesittas: The nested walk does not descend into nested macro function calls– this is what `prune` is for.

sqlmesh/core/model/common.py
georgesittas: `used_variable_referenced_in_metadata_expression[used_var]` should be false whenever this variable is referenced in non-metadata contexts, such as macro calls appearing in the model's query, etc.

If the variable is only referenced in metadata contexts, like in the virtual statement section, under the `audits` property, etc, then it should be true.

If it's false, then the subsequent `all` checks whether the value of the variable only "flows" into metadata macros. So, even if you have a pre-statement like `@foo(@v)`, if `foo` is metadata-only, then the idea is that `v` can also be metadata-only since it's value doesn't really affect the data hash.

georgesittas: Left a comment

sqlmesh/core/model/common.py
themisvaltinos: ooh I see

sqlmesh/core/model/common.py
themisvaltinos: ok I see thank you for explaining

sqlmesh/core/model/common.py
```diff
@@ -77,16 +82,27 @@ def make_python_env(
...
                         raise_config_error(
                             f"The variable name must be a string literal, '{args[0].sql()}' was given instead",
                             path,
                         )
-                    used_variables.add(args[0].this.lower())
+
+                    var_name = args[0].this.lower()
+                    used_variable_referenced_in_metadata_expression[var_name] = (
+                        used_variable_referenced_in_metadata_expression.get(var_name, True)
+                        and bool(is_metadata)
```
izeigerman: Shouldn't `is_metadata` already be a bool? Why do we keep converting it everywhere?

sqlmesh/core/model/common.py
```diff
@@ -126,43 +148,110 @@ def make_python_env(
...
 ) -> t.Dict[str, Executable]:
-    _, python_used_variables = parse_dependencies(
+    _, python_used_variable_referenced_in_metadata_expression = parse_dependencies(
         python_env,
         None,
         strict_resolution=strict_resolution,
         variables=variables,
         blueprint_variables=blueprint_variables,
     )
-    used_variables = (used_variables or set()) | python_used_variables
```
izeigerman, line 152: One thing I'm confused about is why do we no longer pass variables used in **non**-metadata expressions.

izeigerman: ok, I'm misinterpreting. Looks like this dict contains all used variables and the value determines whether it's metadata? Is the argument name misleading then? Should it remain `used_variables`?

sqlmesh/core/model/common.py
```diff
@@ -205,12 +295,16 @@ def blueprint_var(var_name: str, default: t.Optional[t.Any] = None) -> t.Optiona
     local_env = dict.fromkeys(("context", "evaluator"), VariableResolutionContext)
 
     depends_on = set()
-    used_variables = set()
+    used_variable_referenced_in_metadata_expression: t.Dict[str, bool] = {}
```
izeigerman: ditto about the name

izeigerman (approved):

sqlmesh/core/model/common.py
georgesittas: > ok, I'm misinterpreting. Looks like this dict contains all used variables and the value determines whether it's metadata?

Correct. I can see how the name may be confusing. The rationale was to reflect in the name that this is a dict that contains additional info re: metadata or not.

sqlmesh/core/model/common.py
georgesittas: I'm testing it out locally, this refactor may be possible.

georgesittas: Follow-up commit for reference: https://github.com/TobikoData/sqlmesh/pull/4936/commits/16d626044389a3911ecae5807d9d47a0b768981d. I didn't find a reason besides wanting to store `None` over `False` in the `Executable` pydantic model, so I refactored it by pushing this logic down to the relevant instantiations of `Executable`.

sqlmesh/core/model/common.py
georgesittas: Reinstated `used_variables`: https://github.com/TobikoData/sqlmesh/pull/4936/commits/88ff730e56e04b4b1aad9c40401b21477cdaaf19

georgesittas:
@izeigerman thank you for the review– addressed all comments. Planning to merge once CI's green.

georgesittas:
@izeigerman since my last comment in this PR, I discovered a bug and refactored my approach to patch it– can you please take another look when you find some bandwidth?

## The bug

Given the following macros:

```python
from sqlmesh import macro

@macro()
def macro1(evaluator):
    return evaluator.var("foo")

@macro(metadata_only=True)
def macro2(evaluator, var_value):
    return 1
```

and this model:

```sql
MODEL (
  name test_model,
  kind FULL,
);

SELECT
  @macro1() AS col,
  @macro2(@foo) AS col2
```

I verified that `@foo` was being marked as metadata-only, despite it being referenced within `macro1`, which is not metadata-only:

```python
>>> from sqlmesh import Context
>>> ctx = Context()
>>> ctx.models['"test_model"'].python_env
{'macro1': Executable<payload: def macro1(evaluator):
    return evaluator.var('foo'), name: macro1, path: macros/test.py>, 'macro2': Executable<payload: def macro2(evaluator, var_value):
    return 1, name: macro2, path: macros/test.py, is_metadata: True>, '__sqlmesh__vars__metadata__': Executable<payload: {'foo': "'id'"}, kind: value, is_metadata: True>}
```

## Root cause

The problematic logic was located [here](https://github.com/TobikoData/sqlmesh/pull/4936/commits/a8ee3a67ca20e9e81e194c5699f3a0564dfd8cba#diff-2851f45472a26ea4aa1a353ea73526ad92d304ad57688ee3f313dadb821f60c1L211-R210). The value of `used_variables` was `{'foo': False}` for the above project, while the value of `macro_funcs_by_used_var` was `defaultdict(<class 'set'>, {'foo': {'macro2'}})`.

This meant that `used_variables.get(used_var)` would yield false and so we checked the `all(...)` condition that only looked at `macro2`, which is indeed metadata-only but that is not enough to mark `foo` as metadata-only.

## Fix

I addressed the above issue in this [commit](https://github.com/TobikoData/sqlmesh/pull/4936/commits/142dbc243c410de6ee4c466a3b56a28201a468f6). Below is my analysis that led to this refactor:

### Terminology

- *Metadata status*: the value of `is_metadata` for an `Executable` instance
- *Metadata-only*: when a variable or macro `Executable` has `is_metadata` set to `True`
- *Metadata expression*: expressions such as `audits (...)`, or virtual statement blocks that only affect the metadata hash
- *Direct variable reference*: a variable that is referenced directly in SQL code
- *Indirect variable reference*: a variable that is referenced indirectly in SQL code, e.g., an invoked macro references it internally

### Observations

1. We can always infer the metadata status of an indirect variable reference; it inherits it from the macro that references it.

2. We can't determine the metadata status of direct variable references that are under some macro function call until later. The reason is that the macro definition walking happens after we analyze the model's ASTs, so only after that is done we can have the full picture w.r.t. the metadata status of said macro call, which affects the metadata status of the variable itself. For example, a variable reference that is used as an argument of a metadata-only macro is itself metadata-only. The only exception to this rule is when the variable appears in a metadata expression, since everything under it is treated as metadata-only.

3. If there is at least one reference of a variable outside of a metadata expression and not under any macro function call, then that variable is guaranteed to be non-metadata. Otherwise, every reference of said variable may appear in either a metadata expression, or under a macro function call. If there is at least one reference under a macro function call, we must defer the metadata status inference for that variable until after we walk the macro objects and know their metadata statuses. Then, the status of the metadata is all of these statuses combined through conjunction.

PR #4938, opened 2025-07-09
Chore(cicd_bot): Show warnings in the check output

erindru:
Prior to this PR, warnings were logged to the console only. This made them vastly less noticeable/actionable because you'd only ever see them if you were trying to troubleshoot an issue.

PR #4900 introduces a warning that we want people to take notice of / action.
So, this PR:
 - Refactors how the `PR Environment Summary` Markdown is generated as it was getting quite large and unwieldy
 - Update `MarkdownConsole` errors / warnings captured output to produce [alert blocks](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) instead of code blocks
 - Utilize these when writing the check output

Before (warnings not shown in check output):
Screenshot From 2025-07-09 17-48-05

After (warnings shows in check output so they can be actioned):
Screenshot From 2025-07-09 17-43-19


sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -522,13 +522,118 @@ def get_plan_summary(self, plan: Plan) -> str:
             if not difference_summary and not missing_dates:
                 return f"No changes to apply.{plan_flags_section}"
 
-            return f"{difference_summary}\n{missing_dates}{plan_flags_section}"
+            warnings_block = self._console.captured_warnings
```
themisvaltinos: fmi what is the reason that here you don't consume so the error and warnings remain but in `get_pr_environment_summary` you use `consume_captured_warnings` instead

sqlmesh/core/console.py
```diff
@@ -3084,9 +3097,8 @@ def log_skipped_models(self, snapshot_names: t.Set[str]) -> None:
             super().log_skipped_models(snapshot_names)
 
     def log_failed_models(self, errors: t.List[NodeExecutionFailedError]) -> None:
-        if errors:
-            self._errors.append("\n".join(str(ex) for ex in errors))
-            super().log_failed_models(errors)
+        self._errors.extend([str(ex) for ex in errors])
```
themisvaltinos: should this have similar logic check like in log_error ` if message not in self._errors:` or the errors can't have duplicates here

sqlmesh/core/console.py
```diff
@@ -3431,18 +3443,25 @@ def show_linter_violations(
...
-                "\n"  # so that the closing ``` ends up on a newline which is important for GitHub
-            )
+    @property
+    def captured_warnings(self) -> str:
+        return self._render_alert_block("WARNING", self._warnings)
 
-        self._print(f"```\n\\[WARNING] {short_message}```\n\n")
+    @property
+    def captured_errors(self) -> str:
+        return self._render_alert_block("ERROR", self._errors)
```
themisvaltinos: how does this look in this `ERROR` case? is it like the caution one here: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

themisvaltinos (approved):
lgtm the alert block is really nice there! one question I have is about the error alert and how it looks because I can't find examples of it on GitHub

eakmanrq:
The warning looks great! Some things to think about (sorry if this is addressed in PR I just quickly reviewed it)
* There are some warning that will keep showing up and users will want to ignore. The example from the screenshot is good since they might have accepted they are using a non-prod state db. Do you think they would be annoyed seeing this at the top every time? Should it potentially be collapsable and start collapsed so they can opt-in to seeing the information? The plan information on the PR comment does something like this currently
* Checks have a max length and I wonder what users would prefer between truncating the normal check information and truncating warning/errors. Have you thought about that? 
* Are you aware of any warning that could get spammy that we send out in SQLMesh? I haven't worked on console since it added this warning logic so I'm not familiar with it but checking if we are concerned about noisy warnings spamming the top of the check

sqlmesh/core/console.py
erindru: :man_facepalming: I should actually read the documentation I linked.

Yes, this should be `CAUTION` and not `ERROR` because there is no `ERROR`

sqlmesh/integrations/github/cicd/controller.py
erindru: Good question.

I need to revisit this, upon further reflection it's essentially a no-op and because the warnings already get consumed in `get_pr_environment_summary` which means `self._console.captured_warnings` always returns `''`

sqlmesh/core/console.py
erindru: Yeah I was torn between preserving the existing logic or adjusting it.

I should just make this consistent with `_warnings`

erindru:
Oh, interesting, it looks like collapsible sections can go _inside_ alert blocks:
> [!WARNING]
> <details>
> <summary>The PR plan produced the following warnings:</summary>
> foo
> </details>

Would that be preferable, coupled with truncating the warning text if it goes over a certain threshold?

sqlmesh/core/console.py
```diff
@@ -3110,6 +3122,11 @@ class MarkdownConsole(CaptureTerminalConsole):
     AUDIT_PADDING = 7
 
     def __init__(self, **kwargs: t.Any) -> None:
+        self.alert_block_max_content_length = int(kwargs.pop("alert_block_max_content_length", 500))
+        self.alert_block_collapsible_threshold = int(
+            kwargs.pop("alert_block_collapsible_threshold", 200)
+        )
```
erindru, line 3128: I've hardcoded these thresholds for now but it would be a low lift to make them configurable if required.

`alert_block_max_content_length` - maximum amount of text to put inside a `[!WARNING]` block
`alert_block_collapsible_threshold` - if the text is longer than this, it isn't truncated (unless it exceeds `alert_block_max_content_length`), but it's wrapped inside a collapsible section within the `[!WARNING]` block

sqlmesh/integrations/github/cicd/controller.py
erindru: I've updated these to the `consume()` equivalents. Most of the time, warnings in particular will be already consumed by the `PR Environment Synced` step, this would just capture any extra ones raised between the steps

sqlmesh/core/console.py
erindru: I've added in deduplication step. However, I need to do more testing because im not actually sure if this codepath is ever hit.

I've noticed another issue with how failures are reported when a model throws an exception but i'll raise a new PR for that

sqlmesh/core/console.py
erindru: I've updated this

sqlmesh/core/console.py
themisvaltinos: yes I feel the collapsible section within the block is much better and since it is collapsible I don't see the need to truncate and have the please check the console message anymore

sqlmesh/core/console.py
themisvaltinos: great and thanks for providing a example of how it looks

sqlmesh/integrations/github/cicd/controller.py
themisvaltinos: yes i see, it is indeed a bit hard to follow the steps and see what has been consumed

PR #4940, opened 2025-07-09
feat(lsp): return load time errors as diagnostics

benfdking:
- adds path in places to make this as possible with
- refreshes well on typing

image

### DIFF
--- a/sqlmesh/core/model/common.py
+++ b/sqlmesh/core/model/common.py
@@ -41,7 +41,7 @@ def make_python_env(
     macros: MacroRegistry,
     variables: t.Optional[t.Dict[str, t.Any]] = None,
     used_variables: t.Optional[t.Set[str]] = None,
-    path: t.Optional[str | Path] = None,
+    path: t.Optional[Path] = None,
     python_env: t.Optional[t.Dict[str, Executable]] = None,
     strict_resolution: bool = True,
     blueprint_variables: t.Optional[t.Dict[str, t.Any]] = None,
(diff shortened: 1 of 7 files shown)


sqlmesh/core/model/common.py
```diff
@@ -41,7 +41,7 @@ def make_python_env(
     macros: MacroRegistry,
     variables: t.Optional[t.Dict[str, t.Any]] = None,
     used_variables: t.Optional[t.Set[str]] = None,
-    path: t.Optional[str | Path] = None,
+    path: t.Optional[Path] = None,
```
themisvaltinos, line 44: why did you change this here? it seems a bit dangerous

sqlmesh/lsp/errors.py
```diff
@@ -0,0 +1,51 @@
+from lsprotocol.types import Diagnostic, DiagnosticSeverity, Range, Position
+
+from sqlmesh.lsp.uri import URI
+from sqlmesh.utils.errors import (
+    ConfigError,
+)
+import typing as t
+
+type ContextFailedError = t.Union[str, ConfigError, Exception]
```
themisvaltinos:
```suggestion 
ContextFailedError = t.Union[str, ConfigError, Exception]
```
lose the type

sqlmesh/core/model/common.py
benfdking: It makes downstream much simpler, and nothing seems to be passing in a string?

PR #4941, opened 2025-07-09
fix: add `'ducklake` prefix if not present

eakmanrq:
Follow up to https://github.com/TobikoData/sqlmesh/pull/4913

Documentation consistently shows including `'ducklake` in the path: https://ducklake.select/docs/stable/duckdb/usage/choosing_a_catalog_database

It seems like if you exclude it from the path but included the type then it would also work. If you include both though you would get an error. Therefore since the last PR excluded it from TYPE we had some users not include it in the path and had an error.

This PR ensures the `'ducklake` prefix is always there if the type of "ducklake" is provided.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -238,7 +238,10 @@ class DuckDBAttachOptions(BaseConfig):
             options.append("READ_ONLY")
 
         # DuckLake specific options
+        path = self.path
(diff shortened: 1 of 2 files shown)


themisvaltinos (approved):
should we maybe also update the `docs/integrations/engines/duckdb.md` docs to have in the ducklake examples the prefix

tobymao (approved):

PR #4951, opened 2025-07-10
Chore: Update and clarify blueprints doc example macro syntax

themisvaltinos:
Add string in the blueprints example in the docs and clarify its usage with the regular and curly braces macro syntax.


eakmanrq (approved):
Thanks!

PR #4953, opened 2025-07-11
Feat(cicd_bot): Allow forward-only plans based on branch suffix

erindru:
Prior to this, there was no way to trigger the CICD bot to perform a `--forward-only` plan.

This PR:
 - Adds a new bot property, `forward_only_branch_suffix` which defaults to `-forward-only`. This means that any PR's where the git branch ends in `-forward-only`, eg `erin/update-sales-model-forward-only` will have the `--forward-only` flag added to the PR environment and prod deployment plans
 - Adds a tip to the comment the bot creates against the PR on how to restate historical data
 
The `PR Environment Synced` step looks like (notice the `(Forward Only)` indicators and `forward_only=True` in the plan flags):
Screenshot From 2025-07-11 12-21-59

The bot comment on the PR after prod deployment looks like:
Screenshot From 2025-07-11 13-26-32


plaflamme:
FWIW: a more general solution would be to allow more sophisticated commands. For example, we have customized our CI to allow things like:

```
/categorize non-breaking
```

Which categorizes all changes as non-breaking. Similarly for `breaking` and `forward-only`. Ideally, we'd allow specifying this on a per-model basis, perhaps using some kind of form or just something like `/catgorize non-breaking model-a, model-b`, but that seems rather tedious... Perhaps the bot could submit a comment as a form with tickboxes and then the user can do `/plan` to trigger a new plan with the categorization taken from the form, e.g.:

### model-a
- [ ] breaking
- [ ] non-breaking
- [ ] forward-only

In practice though, we've found that applying the same categorization on all changes typically works fine...

erindru:
Thanks @plaflamme , we did discuss a more generic solution for specifying plan flags internally but ultimately decided to not go down that path just yet.

Regarding categorization in particular, we noticed that most people are using auto categorization so we currently have no plans to enable manual categorization via the bot.

sqlmesh/integrations/github/cicd/controller.py
```diff
@@ -485,6 +495,26 @@ def _append_output(cls, key: str, value: str) -> None:
...
+            snapshot = plan.snapshots[snapshot_id]
+            if snapshot.is_incremental:
+                example_model_name = snapshot.node.name
+                break
+
+        return (
+            "> [!TIP]\n"
+            "> In order to see this forward-only plan retroactively apply to historical intervals on the production model, run the below for date ranges in scope:\n"
+            "> \n"
+            f"> `$ sqlmesh plan --restate-model {example_model_name} --start YYYY-MM-dd --end YYYY-MM-DD`\n"
```
sungchun12:
```suggestion
            f"> `$ sqlmesh plan --restate-model {example_model_name} --start YYYY-MM-DD --end YYYY-MM-DD`\n"
```

quick nit

tests/integrations/github/cicd/test_github_commands.py
```diff
@@ -1366,3 +1369,84 @@ def test_comment_command_deploy_prod_no_deploy_detected_yet(
...
+    assert prod_checks_runs[1]["output"]["title"] == "Deployed to Prod"
+    assert GithubCheckConclusion(prod_checks_runs[1]["conclusion"]).is_success
+
+    # PR comment should be updated with forward-only tip
+    assert len(created_comments) == 1
+    assert (
+        """> [!TIP]
+> In order to see this forward-only plan retroactively apply to historical intervals on the production model, run the below for date ranges in scope:
+> 
+> `$ sqlmesh plan --restate-model sushi.customer_revenue_by_day --start YYYY-MM-dd --end YYYY-MM-DD`"""
```
sungchun12:
```suggestion
> `$ sqlmesh plan --restate-model sushi.customer_revenue_by_day --start YYYY-MM-DD --end YYYY-MM-DD`"""
```

same nit

sqlmesh/integrations/github/cicd/config.py
```diff
@@ -33,6 +33,9 @@ class GithubCICDBotConfig(BaseConfig):
     pr_environment_name: t.Optional[str] = None
     pr_min_intervals: t.Optional[int] = None
     prod_branch_names_: t.Optional[str] = Field(default=None, alias="prod_branch_name")
+    forward_only_branch_suffix_: t.Optional[str] = Field(
+        default=None, alias="foward_only_branch_suffix"
```
themisvaltinos: there is a typo in the alias  `foward` instead of `forward` in case this affects the config

erindru: Ah, thanks for spotting. The real question is why didn't a test fail :thinking:

erindru: Oh, test contains same typo :facepalm:

PR #4964, opened 2025-07-14
Chore(cicd_bot): Tidy up failure output

erindru:
Prior to this, if there was an error creating a PR environment, the output was a bit untidy:

- `Plan application failed` message was duplicated
- Warnings were duplicated
- Skipped models / failed models lists had inconsistent formatting
- Failed models code block formatting was off

This PR addresses these issues.

# Before
Screenshot From 2025-07-14 14-15-12

# After
Screenshot From 2025-07-14 14-18-45


sqlmesh/core/console.py
```diff
@@ -3127,6 +3141,11 @@ def __init__(self, **kwargs: t.Any) -> None:
             kwargs.pop("alert_block_collapsible_threshold", 200)
         )
 
+        # capture_only = True: capture but dont print to console
+        # capture_only = False: capture and also print to console
+        self.warning_capture_only = kwargs.pop("warning_capture_only", False)
```
erindru, line 3146: I can't default these to `true` because the `MarkdownConsole` is used as the default console in CircleCI and a bunch of existing tests depend on these getting printed to console.

It's just the bot that wants to capture them and not print to console so it can give them special status in the Check output

themisvaltinos (approved):

Issue #4960, opened 2025-07-12
`sqlmesh destroy` deletes external resources without warning - caused production data loss


asingamaneni:
<h3>Problem Description</h3>
<p>The <code>sqlmesh destroy</code> command currently deletes <strong>ALL resources in target schemas/datasets</strong>, including tables and schemas created and managed by external tools like Terraform. This behavior caused catastrophic data loss in my production environment and represents a critical safety gap.</p>
<h3>What Happened</h3>
<ol>
<li>I had a BigQuery dataset <code>ingest</code> created and managed by Terraform</li>
<li>This dataset contained multiple tables:
<ul>
<li>Some loaded via external ETL processes</li>
<li>Some created by SQLMesh (my C tables)</li>
</ul>
</li>
<li>When I ran <code>sqlmesh destroy</code>, it:
<ul>
<li>❌ Asked for confirmation but <strong>didn't show what would be destroyed</strong></li>
<li>❌ Deleted ALL tables in the <code>ingest</code> dataset (including external ones)</li>
<li>❌ Dropped the entire <code>ingest</code> schema/dataset</li>
<li>❌ No way to recover the external data</li>
</ul>
</li>
</ol>
<p><strong>This was absolutely terrifying and traumatic to deal with.</strong></p>
<h3>Expected Behavior</h3>
<p>I expected SQLMesh to behave like Terraform - only managing resources <strong>it created</strong>. The destroy command should:</p>
<ul>
<li>✅ Only delete SQLMesh-managed resources</li>
<li>✅ Respect boundaries with external tools</li>
<li>✅ <strong>Show exactly what will be destroyed before asking for confirmation</strong></li>
<li>✅ Offer dry-run capability</li>
</ul>
<h3>Current Documentation Gap</h3>
<p>The <a href="https://sqlmesh.readthedocs.io/en/stable/reference/cli/">CLI documentation</a> only states:</p>
<blockquote>
<p>"Removes all project resources, including warehouse objects, state tables, the SQLMesh cache and any build artifacts."</p>
</blockquote>
<p>This is dangerously vague and doesn't warn users about external resource deletion.</p>
<h3>Comparison with Industry Standards</h3>

Tool | Behavior | Safety Features
-- | -- | --
Terraform | Only destroys managed resources | ✅ Shows destruction plan, confirmation prompts, prevent_destroy
dbt | Does NOT auto-delete any resources | ✅ Manual cleanup only
[...]

PR #4965, opened 2025-07-14
Chore: Show resources to be deleted and enhance warnings in destroy

themisvaltinos:
This enhances the warning of the sqlmesh destroy command and lists the resources to be deleted before re-prompting the user to further confirm, fixes: #4960 

destroy


tobymao:
we don't need two commands, just one is enough

sqlmesh/core/console.py
```diff
@@ -1277,14 +1277,20 @@ def stop_cleanup(self, success: bool = False) -> None:
             self.log_error("Cleanup failed!")
 
     def start_destroy(self) -> bool:
-        self.log_warning(
+        self.log_error(
             (
-                "This will permanently delete all engine-managed objects, state tables and SQLMesh cache.\n"
+                "!!!  EXTREME CAUTION: DESTRUCTIVE OPERATION !!!\n\n"
+                "The 'destroy' command will PERMANENTLY DELETE:\n"
```
tobymao: it's not necessary permament if you have other mechanisms to recover, so i wouldn't say that

asingamaneni:
> @themisvaltinos Thank you for putting this together and very helpful. It would be good to be able to show what objects will be deleted before the confirmation, so that the user can see the list before they confirm with "Y".

Ah, I see in the screenshot you have posted. The user needs to confirm twice - please ignore my above comment!

tests/integrations/jupyter/test_magics.py
```diff
@@ -898,19 +898,18 @@ def test_destroy(
...
+        "Snapshot tables to be deleted:",
+        "This action will DELETE ALL the above resources managed by SQLMesh AND\npotentially external resources created by other tools in these schemas.",
+        "Are you ABSOLUTELY SURE you want to proceed with deletion? [y/n]:",
         "Environment 'prod' invalidated.",
         "Deleted object memory.sushi",
         'Deleted object "memory"."raw"."model1"',
-        'Deleted object "memory"."raw"."model1"',
-        'Deleted object "memory"."raw"."model2"',
         'Deleted object "memory"."raw"."model2"',
-        'Deleted object "memory"."raw"."demographics"',
```
erindru, line 910: Curious why `memory.raw.demographics` no longer shows up

erindru (approved):

tests/integrations/jupyter/test_magics.py
themisvaltinos: good point it still shows up this simply checks parts of the output but I added it back as well in the assertions

Issue #4952, opened 2025-07-10
Multi-repo environment statements and dependencies change depending on which repo planned


crericha:
If multiple repos are planned at the same time, all their environment statements and dependencies are included. Then if later, only one environment is planned, the environment statements and dependencies for all other environments are removed. Its not clear if this is a CLI only issue or more than that.

To reproduce, use the multi example project with the following steps:
1. sqlmesh -p repo_1 -p repo_2 plan
2. Modify metadata in repo_2/models/d.sql
3. sqlmesh -p repo_2 plan

Output will look similar to the following (this is cloud output, so it may vary a bit):

(python-default/3.9.21) ~/code/tobiko/projects/multi_repo % sqlmesh -p repo_2 plan 

Differences from the `prod` environment:

```
Environment statements:

before_all:                                                                                                                                                                     
  - CREATE TABLE IF NOT EXISTS before_1 AS select @one()                                                                                                                        
    CREATE TABLE IF NOT EXISTS before_2 AS select @two()                                                                                                                        
                                                                                                                                                                                
after_all:                                                                                                                                                                      
  - CREATE TABLE IF NOT EXISTS after_1 AS select @dup()                                                                                                                         
    CREATE TABLE IF NOT EXISTS after_2 AS select @dup()                                                                                                                         
                                                                                                                                                                                
dependencies:                                                                                                                                                                   
@@ -1,9 +1,3 @@                                                                                                                                                                 
                                                                                                                                                                                
-# macros/__init__.py                                                                                                                                                           
-def dup(context):                                                                                                                                                              
-    return "'repo_1'"                                                                                                                                                          
-# macros/__init__.py                                                                                                                                                           
-def one(context):                                                                                                                                                              
-    return 1                                                                                                                                                                   
 # macros/__init__.py                                                                                                                                                           
 def dup(context):                                                                                                                                                              
     return "'repo_2'"                                                                                                                                                          
                                                                                                                                                                                
Models:
└── Metadata Updated:
    └── silver.d

---                                                                                                                                                                             
                                                                                                                                                                                
+++                                                                                                                                                                             
                                                                                                                                                                                
@@ -1,6 +1,7 @@                                                                                                                                                                 
                                                                                                                                                                                
 MODEL (                                                                                                                                                                        
   name silver.d,                                                                                                                                                               
   project repo_2,                                                                                                                                                              
+  tags silver,                                                                                                                                                                 
   dialect duckdb                                                                                                                                                               
[...]

PR #4966, opened 2025-07-14
Fix!: Add project in environment statements for consistent multi-repo plans

themisvaltinos:
This update adds the `project` information if it's available in a multi-repo setup in the environment statements so that plans triggered from one repo retain the other repos statements from state (mirroring the logic we have for models of a multi repo setup), fixes: #4952

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -622,6 +622,15 @@ class GenericContext(BaseContext, t.Generic[C]):
                 BUILTIN_RULES.union(project.user_rules), config.linter
             )
 
+        # Load environment statements from state for projects not in current load
+        if any(self._projects):
+            prod = self.state_reader.get_environment(c.PROD)
+            if prod:
+                existing_statements = self.state_reader.get_environment_statements(c.PROD)
+                for stmt in existing_statements:
+                    if stmt.project and stmt.project not in self._projects:
+                        self._environment_statements.append(stmt)
+
         uncached = set()
 
         if any(self._projects):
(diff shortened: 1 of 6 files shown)


sqlmesh/core/context.py
```diff
@@ -622,6 +622,15 @@ def load(self, update_schemas: bool = True) -> GenericContext[C]:
                 BUILTIN_RULES.union(project.user_rules), config.linter
             )
 
+        # Load environment statements from state for projects not in current load
+        if any(self._projects):
```
tobymao, line 626: this is kind of a side note, but how do you delete a project?

tobymao: is a migration necessary for this?

tobymao (approved):

sqlmesh/core/context.py
themisvaltinos: > this is kind of a side note, but how do you delete a project?

good question, I was wondering the same. so I believe the only way currently is to remove from the project we want to remove the local models, macros, audits, before all/after all hooks and then run plan _including_ this project in the command

themisvaltinos: > is a migration necessary for this?

i was thinking not since with the current logic multi-repo setups using environment statements wouldn’t work correctly as when someone ran a plan with all projects and then again with individual ones as their statements would be deleted as described in the issue. and since the state doesn’t have the information to tell us which project the hooks belong to, we’d need to reload each project from scratch to determine where they belong so without running plan we couldn't load this information in the state's statements

tobymao: let's add a to do to delete projects i guess

PR #4975, opened 2025-07-16
Feat: correctly handle the generation of VALUES expressions using macros

georgesittas:
This PR aims to address a feature gap. This is the behavior in main today:

```python
>>>
>>> MacroEvaluator().transform(parse_one("SELECT * FROM (VALUES @EACH([1, 2, 3], v -> (v)) ) as v")).sql()
'SELECT * FROM (VALUES ((1), (2), (3))) AS v'
>>> MacroEvaluator().transform(parse_one("SELECT * FROM (VALUES (@EACH([1, 2, 3], v -> (v))) ) as v")).sql()
'SELECT * FROM (VALUES ((1), (2), (3))) AS v'
```

Notice how both inputs result in the same output, which produces a single row with three columns. There is no obvious way to produce three _columns_, using `@EACH`, due to how the `VALUES` parsing works today.

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -419,6 +419,20 @@ def _parse_limit(
     return macro
 
 
+def _parse_value(self: Parser, values: bool = True) -> t.Optional[exp.Expression]:
+    wrapped = self._match(TokenType.L_PAREN, advance=False)
+
+    # The base _parse_value method always constructs a Tuple instance. This is problematic when
+    # generating values with a macro function, because it's impossible to tell whether the user's
+    # intention was to construct a row or a column with the VALUES expression. To avoid this, we
+    # amend the AST such that the Tuple is replaced by the macro function call itself.
+    expr = self.__parse_value()  # type: ignore
+    if expr and not wrapped and isinstance(seq_get(expr.expressions, 0), MacroFunc):
+        return expr.expressions[0]
+
+    return expr
+
+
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -419,6 +419,20 @@ def _parse_limit(
...
 
+def _parse_value(self: Parser, values: bool = True) -> t.Optional[exp.Expression]:
+    wrapped = self._match(TokenType.L_PAREN, advance=False)
+
+    # The base _parse_value method always constructs a Tuple instance. This is problematic when
+    # generating values with a macro function, because it's impossible to tell whether the user's
+    # intention was to construct a row or a column with the VALUES expression. To avoid this, we
+    # amend the AST such that the Tuple is replaced by the macro function call itself.
+    expr = self.__parse_value()  # type: ignore
+    if expr and not wrapped and isinstance(seq_get(expr.expressions, 0), MacroFunc):
```
themisvaltinos, line 430: lgtm. I imagine this preserves any other functionality and in this first position it should be `each` always this macro function?

themisvaltinos (approved):

sqlmesh/core/dialect.py
georgesittas: It doesn't have to be `@EACH`– other macro functions could also return lists. In that case, we should interpret the items in the list as either columns or rows depending on the presence of the parentheses after `VALUES`.

PR #4976, opened 2025-07-16
Fix: Don't use SCD type 2 restatement logic in regular runs

themisvaltinos:
This update introduces an `is_restatement` flag and logic in `run_merged_intervals` to distinguish between restatements (which should clean up historical data from the restatement start date onwards) and regular runs (which should only append new changes). currently SCD Type 2 models with offset  cron greater than the interval start incorrectly used restatement cleanup logic during regular scheduled runs, which can be demonstrated from the added `test_scd_type_2_regular_run_with_offset` test.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -234,6 +234,12 @@ class BuiltInPlanEvaluator(PlanEvaluator):
             return
 
         scheduler = self.create_scheduler(stage.all_snapshots.values(), self.snapshot_evaluator)
+        # Convert model name restatements to snapshot ID restatements
+        restatements_by_snapshot_id = {
+            stage.all_snapshots[name].snapshot_id: interval
+            for name, interval in plan.restatements.items()
+            if name in stage.all_snapshots
+        }
         errors, _ = scheduler.run_merged_intervals(
             merged_intervals=stage.snapshot_to_intervals,
             deployability_index=stage.deployability_index,
@@ -242,6 +248,7 @@ class BuiltInPlanEvaluator(PlanEvaluator):
             circuit_breaker=self._circuit_breaker,
             start=plan.start,
             end=plan.end,
+            restatements=restatements_by_snapshot_id,
         )
         if errors:
             raise PlanError("Plan application failed.")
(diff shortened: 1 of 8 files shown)


sqlmesh/core/plan/evaluator.py
```diff
@@ -234,6 +234,12 @@ def visit_backfill_stage(self, stage: stages.BackfillStage, plan: EvaluatablePla
             return
 
         scheduler = self.create_scheduler(stage.all_snapshots.values(), self.snapshot_evaluator)
+        # Convert model name restatements to snapshot ID restatements
+        restatements_by_snapshot_id = {
+            stage.all_snapshots[name].snapshot_id: interval
+            for name, interval in plan.restatements.items()
+            if name in stage.all_snapshots
```
eakmanrq: Why is this if needed? Is there a reason to believe a name could be in restatements but not in snapshots?

sqlmesh/core/scheduler.py
```diff
@@ -447,6 +450,12 @@ def evaluate_node(node: SchedulingUnit) -> None:
                         execution_time=execution_time,
                     )
                 else:
+                    # Determine if this snapshot and interval is a restatement (for SCD type 2)
+                    is_restatement = (
+                        restatements is not None
+                        and snapshot.snapshot_id in restatements
+                        and start >= restatements[snapshot.snapshot_id][0]
```
eakmanrq: Can you explain why this start check is needed?

sqlmesh/core/plan/evaluator.py
themisvaltinos: no I was simply being cautious since I thought this `stage.all_snapshots[name].snaphot_id` would raise a key error in an edge case. but following the control flow when we're here it should be present

sqlmesh/core/scheduler.py
themisvaltinos: I thought a partial restatement shouldn’t happen in an interval that the interval’s start is earlier than the start date of the restatement, but that’s not actually the case since the logic can handle it so i removed it

eakmanrq (approved):

PR #4977, opened 2025-07-16
fix: gather `engine_adapters` from all projects

etonlels:
Trying to use `sqlmesh render` with multiple projects can fail with a [`Gateway '{gateway}' not found in the available engine adapters.` error](https://github.com/TobikoData/sqlmesh/blob/9b26320f3425348f3fa008c3edcfed3f779af11d/sqlmesh/core/context.py#L2706) because the `engine_adapters` attribute on a `Context` only looks at `self.config` instead of `self.configs`. This PR modifies the definition of `engine_adapters` to look at all projects to avoid this issue.


etonlels:
> can you also add a test for this?

Certainly. Thank you for the starting point 🙏

themisvaltinos (approved):

PR #4981, opened 2025-07-17
Fix(duckdb): Only SET connector_config values on cursor init if they are different

erindru:
Prior to this, a duckdb connection config like:

```
type: duckdb
connector_config:
  temp_directory: /tmp/duckdb/
  memory_limit: '1GB'
```

would result in a `SET` call for each item in `connector_config` every time a new cursor was initialized.

This isnt a problem for many properties, but ones like `temp_directory` have side effects. If DuckDB has had to spill data to disk (and created files in `temp_directory`), if you try to `SET temp_directory='...'`, even to the same value, it will throw an error:
```
duckdb.duckdb.NotImplementedException: Not implemented Error: Cannot switch temporary directory after the current one has been used
```

This PR changes the cursor init to read the existing values and only make `SET` calls if the configured values differ from what is already set


sqlmesh/core/config/connection.py
```diff
@@ -348,11 +348,28 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
+            if self.connector_config:
+                option_names = list(self.connector_config)
+                in_part = ",".join(["?" for _ in range(len(option_names))])
+
+                cursor.execute(
+                    f"SELECT name, value FROM duckdb_settings() WHERE name IN ({in_part})",
+                    option_names,
+                )
+
+                existing_values = {r[0]: r[1] for r in cursor.fetchall()}
```
themisvaltinos:
```suggestion
                existing_values = {field: setting for field, setting in cursor.fetchall()}
```
nit: maybe something more explicit like this so that it's more clear when going through the code

sqlmesh/core/config/connection.py
```diff
@@ -348,11 +348,28 @@ def init(cursor: duckdb.DuckDBPyConnection) -> None:
...
                     raise ConfigError(f"Failed to load extension {extension['name']}: {e}")
 
-            for field, setting in self.connector_config.items():
-                try:
-                    cursor.execute(f"SET {field} = '{setting}'")
-                except Exception as e:
-                    raise ConfigError(f"Failed to set connector config {field} to {setting}: {e}")
+            if self.connector_config:
+                option_names = list(self.connector_config)
+                in_part = ",".join(["?" for _ in range(len(option_names))])
```
themisvaltinos:
```suggestion
                in_part = ",".join("?" for _ in option_names)
```
nit: this should achieve the same

themisvaltinos (approved):

sqlmesh/core/config/connection.py
erindru: thanks, it didnt even occur to me to unpack the tuple like that. will update

PR #4982, opened 2025-07-17
Feat: Allow some control of table naming at the physical layer

### LINKED ISSUE #4403: Ability to create physical layer tables using a hash function
## Description

Currently, some database engines, such as Postgres, have a limit on the length of object names. As a result, sqlmesh users are either heavily restricted in naming conventions or experience conflicts between production and development environments when creating tables.

**Example model name: `analytics_intermediate.int_booking_to_lesson_transitions_logs__extracted`**

In the logs, we see the full physical table name as:

 ```
sqlmesh__analytics_intermediate.analytics_intermediate__int_booking_to_lesson_transitions_logs__extracted__4174565263
``` 

but in Postgres it is created as: 

```
sqlmesh__analytics_intermediate.analytics_intermediate__int_booking_to_lesson_transitions_logs_
``` 

which effectively causes every sqlmesh plan command to work with the same snapshot version.

This leads to several issues:

- Data loss when running plan from any environment (prod/dev), as the same snapshot is reused.

- Schema changes require table deletion — since the table is created using `CREATE TABLE IF NOT EXISTS`, it keeps the existing schema unless dropped. Meanwhile, new inserts use the updated schema, leading to Postgres errors at that stage.

## Potential Solution

A possible solution is to allow physical table names to be generated using a hash function. For example, if we applied the md5 hash function to the snapshot name, the resulting physical table path could look like:

```
"<physical_schema>"."e3b5abfa001550e395b664bd0abde058"
```

This approach would:

- Let users control the length of physical table names.

- Eliminate the issues described above.

### PR DESCRIPTION
**Pull request by erindru**
Fixes #4403 

Up until now, there has been no control over how SQLMesh generates table names for the physical layer. This presented some issues on engines like Postgres that by default have very short identifier length limits that get triggered unexpectedly because SQLMesh generates table names that contain some extra metadata alongside the model name.

This PR introduces a project-level property called `physical_table_naming_convention` that has the following values:

- `schema_and_table` - include both the schema name and table name from the model in the snapshot table name (current behaviour; default)
- `table_only` - include the table name only from the model in the snapshot table name with the assumption that the table is being created within a physical schema that already contains the schema name from the model
- `hash_md5` - produce snapshot table names based on a MD5 hash of what we would have called the table if we were using `schema_and_table`. This leads to predictable identifier lengths at the expense of readability

### CODE REVIEW

File: docs/guides/configuration.md [hunk 1]
```diff
@@ -436,6 +440,89 @@ Given the example of a model called `my_schema.users` with a default catalog of
...
+    ```
+
+This will cause SQLMesh to omit the model schema from the table name and generate physical names that look like (using the above example):
+```
+# <catalog>.sqlmesh__<schema>.<table>__<fingerprint>
+
+warehouse.sqlmesh__finance_mart.transaction_events_over_threshold__<fingerprint>
+```
+
+Notice that the model schema name is no longer part of the physical table name. This allows for slightly longer model names on engines with low identifier length limits, which may be useful for your project.
```
georgesittas, line 493: Reading the docs right now doesn't make it obvious why the schema is included in the physical table's name by default. Should we include an example to explain the rationale behind that choice?

georgesittas:
Did a quick first pass– the implementation seems reasonable.

docs/guides/configuration.md
erindru: To be honest, I don't know what the rationale behind the original choice was.

The only thing I can think of is to allow someone to set:

```
physical_schema_mapping:
  '.*': some_schema
```

Which would map every model to the same physical schema regardless of the model schema. In this situation, including the model schema within the model's table name would be helpful to disambiguate `foo.model_a` and `bar.model_a` if they would both be written as `some_schema.model_a` at the physical layer

erindru: So I believe the default exists because it was the lowest common denominator that worked across any rendition  of `physical_schema_mapping`.

I've updated the docs

docs/reference/configuration.md
```diff
@@ -36,8 +36,9 @@ Configuration options for SQLMesh environment creation and promotion.
...
-| `gateway_managed_virtual_layer`   | Whether SQLMesh views of the virtual layer will be created by the default gateway or model specified gateways - [additional details](../guides/multi_engine.md#gateway-managed-virtual-layer). (Default: False)                                                                                                                 | boolean               | N        |
-| `infer_python_dependencies`   | Whether SQLMesh will statically analyze Python code to automatically infer Python package requirements. (Default: True)                                                                                                                 | boolean               | N        |
+| `physical_table_naming_convention`| Sets which parts of the model name are included in the physical table names. Options are `schema_and_table` or `table_only` - [additional details](../guides/configuration.md#physical-table-naming-convention). (Default: `schema_and_table`)                                                 | string               | N        |
```
izeigerman: Note: The section you're adding this to is called `Environment`. I don't think this is the right place for it, or we should rename the section.

sqlmesh/core/config/root.py
```diff
@@ -147,6 +148,7 @@ class Config(BaseConfig):
     environment_suffix_target: EnvironmentSuffixTarget = Field(
         default=EnvironmentSuffixTarget.default
     )
+    physical_table_naming_convention: t.Optional[TableNamingConvention] = None
```
izeigerman: Why not use `schema_and_table` as a default?

sqlmesh/core/snapshot/definition.py
```diff
@@ -333,6 +334,7 @@ class SnapshotInfoMixin(ModelKindMixin):
     # This can be removed from this model once Pydantic 1 support is dropped (must remain in `Snapshot` though)
     base_table_name_override: t.Optional[str]
     dev_table_suffix: str
+    table_naming_convention: t.Optional[TableNamingConvention] = None
```
izeigerman: Ditto: I believe we have an appropriate default for this

sqlmesh/core/snapshot/definition.py
```diff
@@ -605,6 +609,9 @@ class Snapshot(PydanticModel, SnapshotInfoMixin):
     base_table_name_override: t.Optional[str] = None
     next_auto_restatement_ts: t.Optional[int] = None
     dev_table_suffix: str = "dev"
+    table_naming_convention_: t.Optional[TableNamingConvention] = Field(
```
izeigerman: When reusing the physical version (eg. FORWARD_ONLY, METADATA) we should inherit the convention of the previous snapshot just like we do the physical schema: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/definition.py#L1015

sqlmesh/core/snapshot/definition.py
```diff
@@ -1568,14 +1579,41 @@ def table_name(
...
+        # just take a MD5 hash of what we would have generated anyway using SCHEMA_AND_TABLE
+        value_to_hash = table_name(
+            physical_schema=physical_schema,
+            name=name,
+            version=version,
+            catalog=catalog,
+            suffix=suffix,
+            naming_convention=TableNamingConvention.SCHEMA_AND_TABLE,
+        )
+        full_name = f"{c.SQLMESH}_md5__{md5(value_to_hash)}"
```
izeigerman, line 1598: `_md5` is redundant here imho

sqlmesh/core/snapshot/definition.py
erindru: The nice thing about using `None` is that it makes the serialized JSON payloads smaller. But I dont have a strong opinion on this, so i've changed these to default to `TableNamingConvention.default`

sqlmesh/core/snapshot/definition.py
erindru: Ahh, thanks. I had a suspicion something like this was lurking.

I've added a test to show that categorizing a snapshot as `FORWARD_ONLY` inherits the `table_naming_convention` from the `previous_version`

docs/reference/configuration.md
erindru: I put it here because the other `physical_*` properties were here, but you're right this could use a refactor. I've:

- renamed this section to `Environments (Virtual Layer)`
- moved the project properties like `log_limit` to the `Projects` section
- moved the model properties like `time_column_format` to the `Models` section
- added a section called `Database (Physical Layer)` to house the `physical_*` properties

izeigerman (approved):

PR #4983, opened 2025-07-17
Feat: Add signal listener and methods in scheduler

themisvaltinos:
This adds the provisionally called `SignalListener` class with methods which are called when signals are registered (`on_signal_register`), when a signal starts (`on_signal_start`) and when it finishes with success or failure (`on_signal_end`).

### DIFF
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -49,7 +49,6 @@ from sqlmesh.utils.errors import (
     SQLMeshError,
     SignalEvalError,
 )
-from sqlmesh.utils.pydantic import serialize_expressions
 
 if t.TYPE_CHECKING:
     from sqlmesh.core.context import ExecutionContext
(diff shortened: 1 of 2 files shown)


sqlmesh/utils/pydantic.py
```diff
@@ -60,6 +60,16 @@ def _expression_encoder(e: exp.Expression) -> str:
     return e.meta.get("sql") or e.sql(dialect=e.meta.get("dialect"))
 
 
+def serialize_expressions(kwargs: t.Dict[str, t.Optional[exp.Expression]]) -> t.Dict[str, str]:
+    serialized_kwargs: t.Dict[str, str] = {}
+    for key, value in kwargs.items():
+        if isinstance(value, exp.Expression):
+            serialized_kwargs[key] = _expression_encoder(value)
+        else:
+            serialized_kwargs[key] = str(value)
```
georgesittas: Checking whether `value` is an `Expression` seems redundant, based on `kwargs`' type hint.

georgesittas (approved):

sqlmesh/core/scheduler.py
```diff
@@ -60,6 +61,37 @@
 SchedulingUnit = t.Tuple[str, t.Tuple[Interval, int]]
 
 
+class SignalListener:
+    def on_signal_register(
+        self,
+        snapshot_id: SnapshotId,
+        signals: t.Dict[str, t.Dict[str, str]],
```
izeigerman: Passing serialized arguments means that we design this interface for one use case and one use case only. I think it'd be more extendable if we just pass a `Snapshot` instance instead and let the implementation figure out what to do with it.

sqlmesh/core/scheduler.py
```diff
@@ -60,6 +61,37 @@
...
+    ) -> None:
+        pass
+
+    def on_signal_start(
+        self,
+        snapshot_id: SnapshotId,
+        signal_name: str,
+        signal_index: int,
+        intervals: Intervals,
+        signal_kwargs: t.Dict[str, str],
```
izeigerman: Ditto: I'd rather just pass a snapshot

sqlmesh/core/scheduler.py
izeigerman: Btw, shouldn't the registration be performed once for ALL snapshots with signals? And not the individual one?

themisvaltinos: > Btw, shouldn't the registration be performed once for ALL snapshots with signals? And not the individual one?

yes will revise to do it all snapshots at once and since passing snapshot instead it'd be easier

sqlmesh/core/scheduler.py
themisvaltinos: revised to pass snapshot here instead. I'm not sure if you meant to also not pass the signal related args?

PR #4984, opened 2025-07-17
fix!: correctly identify keys to make deterministic

eakmanrq:
Fixing a mistake in this PR: https://github.com/TobikoData/sqlmesh/pull/4925

The issue is that the user could have defined a variable in the global namespace and this PR would sort the dictionary which would cause an issue for the user if they rely on the dictionary order. Therefore this PR makes the sorting logic focused just on the known dictionaries that we control: sqlmesh vars and sqlmesh blueprint vars. It also changes the sorting to only sort on the root level dict instead of doing a recursive sort. This is because a user could define a dict as a value in variable/blueprint variables and we don't want to change that. 

The previous 85 migration is updated to do the correct thing. 86 migration was added to try to detect if the user may have been impacted by this bug and if so it will log a warning.

### DIFF
--- a/sqlmesh/migrations/v0085_deterministic_repr.py
+++ b/sqlmesh/migrations/v0085_deterministic_repr.py
@@ -12,6 +12,9 @@ from sqlglot import exp
 from sqlmesh.utils.migration import index_text_type, blob_text_type
 
 
+KEYS_TO_MAKE_DETERMINISTIC = ["__sqlmesh__vars__", "__sqlmesh__blueprint__vars__"]
+
+
 # Make sure `SqlValue` is defined so it can be used by `eval` call in the migration
 @dataclass
 class SqlValue:
@@ -82,6 +85,8 @@ def migrate(state_sync, **kwargs):  # type: ignore
 
         if python_env:
             for key, executable in python_env.items():
+                if key not in KEYS_TO_MAKE_DETERMINISTIC:
+                    continue
                 if isinstance(executable, dict) and executable.get("kind") == "value":
                     old_payload = executable["payload"]
                     try:
--- a/sqlmesh/migrations/v0085_deterministic_repr.py
+++ b/sqlmesh/migrations/v0085_deterministic_repr.py
@@ -4,12 +4,16 @@ and therefore this migration applies deterministic sorting to the keys of the di
 """
 
 import json
+import logging
 import typing as t
 from dataclasses import dataclass
 
 from sqlglot import exp
 
 from sqlmesh.utils.migration import index_text_type, blob_text_type
+
+
+logger = logging.getLogger(__name__)
 
 
 KEYS_TO_MAKE_DETERMINISTIC = ["__sqlmesh__vars__", "__sqlmesh__blueprint__vars__"]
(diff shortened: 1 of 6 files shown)


sqlmesh/migrations/v0085_deterministic_repr.py
```diff
@@ -20,25 +23,13 @@ class SqlValue:
...
-        return o
-
+def non_recursive_dict_sort(obj: t.Any) -> str:
     try:
-        return repr(_normalize_for_repr(obj))
+        if isinstance(obj, dict):
+            obj = dict(sorted(obj.items(), key=lambda x: str(x[0])))
     except Exception:
-        return repr(obj)
+        pass
```
crericha: Why are we silently ignoring an exception?

sqlmesh/migrations/v0085_deterministic_repr.py
```diff
@@ -82,12 +73,14 @@ def migrate(state_sync, **kwargs):  # type: ignore
 
         if python_env:
             for key, executable in python_env.items():
+                if key not in KEYS_TO_MAKE_DETERMINISTIC:
+                    continue
                 if isinstance(executable, dict) and executable.get("kind") == "value":
                     old_payload = executable["payload"]
```
crericha, line 83: No deep copy needed here?

eakmanrq: No because we eval the old payload which converts it from string to dict.

sqlmesh/migrations/v0085_deterministic_repr.py
eakmanrq: An exception would seem to indicate that the dictionary is not sortable. So rather than raising we just don't sort. Will update to log a warning.

sqlmesh/migrations/v0086_check_deterministic_bug.py
```diff
@@ -0,0 +1,79 @@
...
+                    and isinstance(executable, dict)
+                    and executable.get("kind") == "value"
+                ):
+                    try:
+                        parsed_value = eval(executable["payload"])
+                        if isinstance(parsed_value, dict):
+                            get_console().log_warning(warning)
+                            return
+                    except Exception:
+                        pass
```
crericha: No warning log if an exception?

eakmanrq: Same as above will log

sqlmesh/utils/metaprogramming.py
```diff
@@ -635,36 +636,13 @@ def print_exception(
...
-        if isinstance(o, dict):
-            sorted_items = sorted(o.items(), key=lambda x: str(x[0]))
-            return {k: _normalize_for_repr(v) for k, v in sorted_items}
-        if isinstance(o, (list, tuple)):
-            # Recursively normalize nested structures
-            normalized = [_normalize_for_repr(item) for item in o]
-            return type(o)(normalized)
-        return o
-
+def non_recursive_dict_sort(obj: t.Any) -> str:
```
crericha: Is non_recursive important for the name or only the implementation? Wondering if _dict_sort is a good enough name?

sqlmesh/utils/metaprogramming.py
```diff
@@ -635,36 +636,13 @@ def print_exception(
...
     try:
-        return repr(_normalize_for_repr(obj))
+        if isinstance(obj, dict):
+            obj = dict(sorted(obj.items(), key=lambda x: str(x[0])))
     except Exception:
-        return repr(obj)
+        pass
```
crericha: Return nothing on exception?

crericha:
Will this mean user dicts are still non-deterministic from plan to plan?

eakmanrq:
> Will this mean user dicts are still non-deterministic from plan to plan?

Yes I have basically changed the implementation to be as conservative as possible. If we see an issue with this in the future we can expand to include them.

sqlmesh/utils/metaprogramming.py
erindru: I think non_recursive is helpful because it communicates that only the first layer is considered and nested objects are ignored

eakmanrq: Already updated it... I didn't feel strongly either way.

PR #4993, opened 2025-07-21
Chore: Add cron in the fields to render for python models

themisvaltinos:
This update adds support for resolving variables in cron for python models. The updated existing tests show the use for macro variables and the newly added tests for blueprint python models.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2786,10 +2786,6 @@ def render_meta_fields(
         ) or field_value is None:
             continue
 
-        # If it contains a macro reference we need to render it similar to sql models
-        if field == "cron" and isinstance(field_value, str):
-            field_value = exp.to_identifier(field_value)
-
         if field in RUNTIME_RENDERED_MODEL_FIELDS:
             fields[field] = parse_strings_with_macro_refs(field_value, dialect)
             continue
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -74,8 +74,6 @@ if t.TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-UNRENDERABLE_MODEL_FIELDS = {"cron", "description"}
-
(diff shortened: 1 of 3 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -84,6 +84,16 @@
...
 
+CRON_ALIASES = {
+    "@midnight",
+    "@hourly",
+    "@daily",
+    "@weekly",
+    "@monthly",
+    "@yearly",
+    "@annually",
+}
```
georgesittas: What about `CRON_SHORTCUTS`?

sqlmesh/core/model/definition.py
```diff
@@ -2771,9 +2781,15 @@ def render_field_value(value: t.Any) -> t.Any:
...
         # We don't want to parse python model cron="@..." kwargs (e.g. @daily) into MacroVar
-        if field == "cron" or field_value is None:
+        if (
+            field == "cron" and isinstance(field_value, str) and field_value.lower() in CRON_ALIASES
+        ) or field_value is None:
             continue
 
+        # If it contains a macro reference we need to render it similar to sql models
+        if field == "cron" and isinstance(field_value, str):
+            field_value = exp.to_identifier(field_value)
```
georgesittas: Is this needed? Won't the control flow reach L2820, where `field_value` will be parsed & rendered?

themisvaltinos: without it a cron like: `"*/@{min} * * * *"` would be parsed as a multiplication expression because of the *

georgesittas: I see, but I'm still not sure whether this is a pattern we should support necessarily.

We don't parse these cron expressions for SQL models unless they're string literals, so allowing them for python and resorting to hacky solutions like using `to_identifier` to work around this limitation feels weird to me.

I wonder if we should just document that one needs to do `"@'...@{var}...'"` to work around the issue.

themisvaltinos: yes I see what you're saying. updated the docs with clarification over this and removed the use of to_identifier, let me know if this looks better

PR #5001, opened 2025-07-23
Feat!: introduce migration pre-checks

georgesittas:
Some SQLMesh versions introduce breaking changes that cannot be fully handled by simply manipulating state. These upgrades typically require users to update their project definition and apply a plan to complete the migration process. Without doing so, cadence runs may start failing after the upgrade, and applying a plan becomes necessary to restore normal operation.

The problem is that we currently have no way to inform the user that this additional step is required. The best we can do is issue a warning during the migration process ([example](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/migrations/v0078_warn_if_non_migratable_python_env.py#L94)), which is already too late. Users don’t get a chance to prepare for such a breaking upgrade and are forced to to fix their project ASAP or resort to rolling back their state.

This PR aims to address the above problem by introducing a new concept: pre-checks. These are scripts that correspond to a particular migration script and contain a single function `pre_check`.

The `pre_check` function's only purpose is to inspect the current state and determine whether the project will require a plan application after the migration is complete. It returns a list of messages that will be shown to the user, explaining any necessary changes on their end and outlining the steps they need to take.

It’s the responsibility of the developer introducing the breaking change to implement the `pre_check` function, in addition to the `migrate` function, when applicable.


sqlmesh/core/state_sync/base.py
```diff
@@ -64,6 +64,10 @@ def _schema_version_validator(cls, v: t.Any) -> int:
     importlib.import_module(f"sqlmesh.migrations.{migration}")
     for migration in sorted(info.name for info in pkgutil.iter_modules(migrations.__path__))
 ]
+PRE_CHECKS = {
+    pre_check: importlib.import_module(f"sqlmesh.pre_checks.{pre_check}")
```
izeigerman: Can pre-checks share the module with the migration function? I'm not sure I want to have them separated

georgesittas: Discussed offline– will be moving forward with this ^ approach instead.

sqlmesh/core/console.py
```diff
@@ -1472,6 +1490,33 @@ def stop_env_migration_progress(self, success: bool = True) -> None:
...
                 self.log_success("Environments migrated successfully")
 
+    def log_pre_check_warnings(
+        self,
+        pre_check_warnings: t.List[t.Tuple[str, t.List[str]]],
+        pre_check_only: bool,
+    ) -> bool:
+        if pre_check_warnings:
+            for pre_check, warnings in pre_check_warnings:
+                tree = Tree(f"[bold]Pre-migration warnings for {pre_check}[/bold]")
```
izeigerman: Not sure whether there's any point in showing the internal module name here

georgesittas: This is the name of the migration script's stem, like `vXXXX_foo_bar`. I mostly included it as a breadcrumb for the devs to understand where the warning came from, if needed to debug / explain a situation. It's kinda hypothetical though, so I don't mind removing it if you think it's noisy.

izeigerman: We should use logs for breadcrumbs. The CLI output is reserved for the user

georgesittas: I see, fair enough. I'll clean this up, thanks.

georgesittas: Done: https://github.com/TobikoData/sqlmesh/pull/5001/commits/e08d38eb501a21aeee7e3d6a97bc08d83e4ae2f7

georgesittas:
This has been deprioritized, so I'm closing it for now to reduce clutter in the PR list. Will keep the branch intact.

PR #5009, opened 2025-07-24
feat!: linting rule no unregistered external models

benfdking:


sqlmesh/core/linter/rules/builtin.py
```diff
@@ -110,4 +110,33 @@ def check_model(self, model: Model) -> t.Optional[RuleViolation]:
...
+                not_registered_external_models.add(depends_on_model)
+
+        if not not_registered_external_models:
+            return None
+
+        return RuleViolation(
+            rule=self,
+            violation_msg=f"Model '{model.name}' depends on unregistered external models: "
+            f"{', '.join(m for m in not_registered_external_models)}. "
+            "Please register them in the external_models.yaml file.",
```
georgesittas: If the `external_models.yaml` file is missing and the user is unaware of it, then this may not be very helpful. Perhaps worth mentioning the `sqlmesh create_external_models` command here.

tests/core/linter/test_builtin.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_no_missing_external_models(tmp_path, copy_to_temp_path) -> None:
+    """
+    Tests that the linter correctly identifies unregistered external model dependencies.
+
+    This test removes the `external_models.yaml` file from the sushi example project,
+    enables the linter, and verifies that the linter raises a violation for a model
+    that depends on unregistered external models.
+    """
+    sushi_paths = copy_to_temp_path("examples/sushi")
```
VaggelisD, line 14: Do we even need the sushi project here? I think we could get the same violation but simply creating an empty project with a model such as `SELECT * FROM external_model`. 

This way we don't have to:
- Delete the `external_models.yaml` file, it doesn't exist in the first place
- We don't have to copy the whole sushi dir
- We don't have to string replace the config, we can create one from scratch

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -110,4 +110,31 @@ def check_model(self, model: Model) -> t.Optional[RuleViolation]:
             return self.violation()
 
 
+class NoMissingExternalModels(Rule):
+    """All external models must be registered in the external_models.yaml file"""
+
+    def check_model(self, model: Model) -> t.Optional[RuleViolation]:
+        # Ignore external models themselves, because either they are registered,
```
VaggelisD, line 117: [Nit] This can be inlined in the for loop

izeigerman:
The rule looks good, but I agree with @georgesittas. Can we introduce a list of rules that are enabled by default to avoid making this a breaking change.

tests/core/linter/test_builtin.py
benfdking: I am curious to understand the objection to this test. 

I don't particularly understand the rationale of the things you mention we avoid. While they are slightly inconvenient, the test is 50 lines, but because it's testing a fully fledged project, it gives me a lot of confidence. The only argument I can get is slow tests, but if we're optimising for that, we should be using other tools. 

Another way, I kind of read this is maybe about a philosophical difference in testing: The test you describe is the most isolated, unit-y test you can build, it's as minimal a test as possible to prove the thing works whereas mine feels like I start with the integration tests and test the function way more thoroughly because I exercise through a lot more scenarios. A minimal unit test wouldn't reveal, for example, any false positive, but running through sushi, which exercises a lot of variations, just might. 

Someone once described this testing to me as "happy path" testing. You build a robust happy e2e test (e.g. our sushi project) and then you exercise paths around it by altering the inputs.

VaggelisD: It's not an outright objection but it also doesn't feel like the right tool for this. My reasoning being:
- There's a lot more noise compared to a test that'd simply use a dummy model and a `Config` object. This makes understanding & updating the test harder and some parts of it may be considered "hardcoded" such as the YAML overwrite
- There's obviously some performance overhead, and although we are not optimising for that per se, it shouldn't be neglected
- Imo you don't "win" anything in terms of integration testing by using `sushi` instead of a simpler project for this case

It makes sense to use the pre-existing projects such as `sushi` or `multi` when it comes to testing the entirety of the linter end to end, but otherwise this feels like it uses more real estate than is required for such a test case

georgesittas:
Getting this in to include with v0.203.0, can revisit the test if needed.

PR #5013, opened 2025-07-24
feat(vscode): adding ability to update columns

benfdking:


sqlmesh/lsp/commands.py
```diff
@@ -0,0 +1,7 @@
+from sqlmesh.utils.pydantic import PydanticModel
+
+EXTERNAL_MODEL_UPDATE_COLUMNS = "sqlmesh.external_model_update_columns"
+
+
+class ExternalModelUpdateColumnsRequest(PydanticModel):
+    model_name: str
```
benfdking: Unnecessary

sqlmesh/lsp/context.py
```diff
@@ -228,8 +236,42 @@ def get_code_actions(
...
+                ],
+            )
+            code_action = types.CodeLens(
+                range=range,
+                command=command,
+            )
+            code_actions.append(code_action)
 
-        return code_actions if code_actions else None
+        return code_actions if code_actions else []
```
benfdking: Rename variable

sqlmesh/lsp/context.py
```diff
@@ -228,8 +236,42 @@ def get_code_actions(
                         edit=types.WorkspaceEdit(changes={params.text_document.uri: text_edits}),
                     )
                     code_actions.append(code_action)
+        return code_actions
```
benfdking: Revert this

vscode/extension/package.json
```diff
@@ -134,6 +134,7 @@
     "package": "rm -rf ./src_react && mkdir -p ./src_react && cd ../react && pnpm run build && cd ../extension && cp -r ../react/dist/* ./src_react && pnpm run check-types && node esbuild.js --production"
   },
   "dependencies": {
+    "@duckdb/node-api": "1.3.2-alpha.25",
```
benfdking, line 137: Make it non-alpha

sqlmesh/lsp/main.py
```diff
@@ -368,6 +369,47 @@ def function_call(ls: LanguageServer, params: t.Any) -> t.Dict[str, t.Any]:
 
             self.server.feature(name)(create_function_call(method))
 
+        # Check if command is already registered to prevent duplicate registration errors
+        if EXTERNAL_MODEL_UPDATE_COLUMNS not in getattr(self.server, "_commands", {}):
```
benfdking: Actually can we revert this complexity now

sqlmesh/lsp/context.py
```diff
@@ -298,6 +303,42 @@ def get_code_actions(
...
+                arguments=[
+                    name,
+                ],
+            )
+            code_lenses.append(
+                types.CodeLens(
+                    range=to_lsp_range(range),
+                    command=command,
+                )
+            )
```
benfdking: this could be collapsed

sqlmesh/lsp/context.py
```diff
@@ -399,3 +434,72 @@ def diagnostic_to_lsp_diagnostic(
...
+        existing_model_columns = existing_model.get("columns")
+
+        # Get the adapter and fetch columns
+        adapter = self.context.engine_adapter
+        # Get columns for the model
+        new_columns = get_columns(
+            adapter=adapter,
+            dialect=self.context.config.model_defaults.dialect,
+            table=model_name,
+            strict=True,
```
georgesittas, line 466: Can `strict` here cause an abrupt exit due to possible exceptions? Is that gracefully handled elsewhere in the LSP?

sqlmesh/lsp/context.py
```diff
@@ -399,3 +434,72 @@ def diagnostic_to_lsp_diagnostic(
...
+        )
+        # Compare existing columns and matching types and if they are the same, do not update
+        if existing_model_columns is not None:
+            if existing_model_columns == new_columns:
+                return False
+
+        # Model index to update
+        model_index = next(
+            (i for i, model in enumerate(models) if model.get("name") == model_name), None
+        )
```
georgesittas, line 476: Can you double-check whether this condition is correct? I recall that some entries in the schema file can contain a `gateway`. For this reason, I'm unsure whether the model name can be used as a unique key here.

georgesittas (approved):
Looks ok, just a couple of comments.

sqlmesh/lsp/context.py
benfdking: Yep every method in the LSP is wrapped by a try catch block at the moment including this one. I just checked.

sqlmesh/lsp/context.py
benfdking: You are right but in the context of this it will always match as it came from it. That said I have clarified that in the commments..

PR #5014, opened 2025-07-24
Fix!: Dont normalize aliases in merge and when matched

themisvaltinos:
This pr : https://github.com/TobikoData/sqlmesh/pull/4847 along with the columns added normalisation for MERGE_SOURCE_ALIAS and MERGE_TARGET_ALIAS constants that since are capitalised in the merge statement are mismatched and lead to error when executing the merge for dialects like Postgres so we'd get:
 ```sql
 MERGE INTO table AS "__MERGE_TARGET__" USING (
SELECT ... AS "__MERGE_SOURCE__" ON 
WHEN MATCHED THEN UPDATE SET "id" = "__merge_source__"."id"  ... ) # this will be set to lowercase due to normalisation
 ```
 instead of:
  ```sql
 MERGE INTO table AS "__MERGE_TARGET__" USING (
SELECT ... AS "__MERGE_SOURCE__" ON 
WHEN MATCHED THEN UPDATE SET "id" = "__MERGE_SOURCE__"."id"  ... )
 ```
This pr leaves the normalisation logic, but reverts to the previous behaviour for the aliases.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -303,10 +303,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          # filters:
-          #   branches:
-          #     only:
-          #       - main
+          filters:
+            branches:
+              only:
+                - main
       - ui_style
       - ui_test
       - vscode_test
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -2,9 +2,9 @@ on:
   push:
     branches:
       - main
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    branches:
+      - main
 concurrency:
   group: 'pr-${{ github.event.pull_request.number }}'
   cancel-in-progress: true


themisvaltinos:
> Does this require a migration?

if I’m not missing anything im not sure it is needed, because the merge into would fail unless the engine is one with uppercase like Snowflake and in that case nothing would change in state regarding the aliases. my thinking is if a when matched was in state and the dialect was case-insensitive but not uppercase (like Postgres) it simply wouldn't work as the alias would be lowercase causing `sqlmesh run` to fail

sqlmesh/core/dialect.py
```diff
@@ -1412,18 +1412,10 @@ def replace_merge_table_aliases(
...
-    )
-
-    normalized_merge_target_alias = quote_identifiers(
-        normalize_identifiers(exp.to_identifier(MERGE_TARGET_ALIAS), dialect), dialect=dialect
-    )
-
     if isinstance(expression, exp.Column) and (first_part := expression.parts[0]):
         if first_part.this.lower() in ("target", "dbt_internal_dest", "__merge_target__"):
-            first_part.replace(normalized_merge_target_alias)
+            first_part.replace(exp.to_identifier(MERGE_TARGET_ALIAS, quoted=True))
```
erindru, line 1426: It took me some time to grok this but I see why it works. `quoted=True` is the key part to prevent it being un-done later.

This essentially ensures that everything is uppercase regardless of the engine's normalization strategy, which lines up with what `EngineAdapter._merge` uses.

`EngineAdapter._merge` produces an _unquoted_ uppercase value, which gets quoted by default during `EngineAdapter.execute`, which is why it needs to either be uppercase here or normalized in `EngineAdapter._merge`

.circleci/continue_config.yml
```diff
@@ -303,10 +303,10 @@ workflows:
                 - bigquery
                 - clickhouse-cloud
                 - athena
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
```
themisvaltinos: Todo: to reinstate prior to merge

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -1897,6 +1897,125 @@ def _mutate_config(current_gateway_name: str, config: Config):
         ctx.cleanup(context)
 
 
+def test_incremental_by_unique_key_model_when_matched(ctx: TestContext):
+    if not ctx.supports_merge:
+        pytest.skip(f"{ctx.dialect} on {ctx.gateway} doesnt support merge")
+
+    # DuckDB and some other engines use logical_merge which doesn't support when_matched
+    if ctx.dialect not in ["bigquery", "databricks", "postgres", "snowflake", "spark"]:
```
erindru: nit: would a more robust test check if `ctx.engine_adapter` is an instance of `LogicalMergeMixin` instead?

sqlmesh/core/model/kind.py
```diff
@@ -478,10 +478,9 @@ def _when_matched_validator(
                 v = v[1:-1]
 
             v = t.cast(exp.Whens, d.parse_one(v, into=exp.Whens, dialect=dialect))
-        else:
-            v = t.cast(exp.Whens, v.transform(d.replace_merge_table_aliases, dialect=dialect))
 
-        return validate_expression(v, dialect=dialect)
+        v = validate_expression(v, dialect=dialect)
+        return t.cast(exp.Whens, v.transform(d.replace_merge_table_aliases, dialect=dialect))
```
erindru, line 483: Previously we only ran the `replace_merge_table_aliases` transform when loading from disk. When reading back from state, we assumed it had already been applied so didn't apply it again.

However, applying it regardless like you do here will transparently "fix" what was in state, right? So that when people upgrade SQLMesh, everything should still match

tests/core/engine_adapter/integration/test_integration.py
themisvaltinos: yes good point and much cleaner this way revised it

sqlmesh/core/model/kind.py
themisvaltinos: yes that was my thinking behind it to foolproof it

PR #5033, opened 2025-07-27
Feat: Add Doris adapter

xinge-ji:
Engine Adapter Implementation
Doris Adapter: Implemented `DorisEngineAdapter` to support Doris-specific SQL behavior (`sqlmesh/core/engine_adapter/doris.py`)

Connection Configuration
Doris Connection Config: Added a new `DorisConnectionConfig`, including support for basic authentication and configurable HTTP port (`sqlmesh/core/config/connection.py`)

Documentation Updates
Doris Guide: Added a detailed guide for Doris integration, covering setup, model support, connection configuration, and limitations (`docs/integrations/engines/doris.md`)


erindru:
Hey @xinge-ji , let us know when you'd like us to review.

We typically ignore a PR still in 'Draft' state, and also PR's with failing tests and merge conflicts with main (with the assumption that the requester is still working on the PR)

tests/core/engine_adapter/integration/test_integration_doris.py
```diff
@@ -0,0 +1,35 @@
...
+
+from sqlmesh.core.engine_adapter import DorisEngineAdapter
+from tests.core.engine_adapter.integration import TestContext
+from tests.core.engine_adapter.integration import (
+    generate_pytest_params,
+    ENGINES_BY_NAME,
+    IntegrationTestEngine,
+)
+
+pytestmark = [pytest.mark.doris, pytest.mark.engine, pytest.mark.slow]
```
erindru: I this `pytestmark` is why the "style_and_cicd_tests" task is failing - because it's causing the tests in this file to get included when they shouldnt be, they should only be included in the "engine_doris" task.

If you remove it, "style_and_cicd_tests" should stop failing on this test. See the other `test_integration_<engine>.py` files for an example

erindru:
> Would it be possible to add it in the ci environment?

You can adjust the ci environment by editing `.circleci/continue_config.yml` but I think in this case it's caused by the doris-specific tests getting picked up by the general test task

tests/core/engine_adapter/integration/test_integration_doris.py
xinge-ji: > I this `pytestmark` is why the "style_and_cicd_tests" task is failing - because it's causing the tests in this file to get included when they shouldnt be, they should only be included in the "engine_doris" task.
> 
> If you remove it, "style_and_cicd_tests" should stop failing on this test. See the other `test_integration_<engine>.py` files for an example

Thank you! This is now ready for review.

docs/integrations/engines/doris.md
```diff
@@ -0,0 +1,397 @@
...
+      buckets = 16
+    ),
+    enable_unique_key_merge_on_write = 'true'
+  )
+);
+```
+
+## Table Properties
+
+The Doris adapter supports a comprehensive set of table properties that can be configured in the `physical_properties` section of your model. These properties are processed by the `_build_table_properties_exp` method.
```
erindru: `_build_table_properties_exp` is an internal method of `EngineAdapter` and doesnt really belong in user-facing docs

docs/integrations/engines/doris.md
```diff
@@ -0,0 +1,397 @@
...
+Doris supports range partitioning to improve query performance. SQLMesh automatically translates partitioning into Doris's `PARTITION BY RANGE` syntax.
+
+**Custom Partition Expression:**
+```sql
+MODEL (
+  name my_partitioned_model,
+  kind INCREMENTAL_BY_TIME_RANGE(time_column (event_date, '%Y-%m-%d')),
+  partitioned_by event_date
+  physical_properties (
+    partitioned_by_expr = 'FROM ("2000-11-14") TO ("2099-11-14") INTERVAL 2 YEAR',
```
erindru: SQLMesh does support expressions in the existing `partitioned_by` property, eg:

`partitioned_by date_trunc(event_timestamp)`

or eg for Iceberg transforms on Athena/Trino:

`partitioned_by (day(cola), truncate(colb, 8))`

Is there a reason why we couldn't make this more ergonomic for Doris? eg something like:

```
MODEL (
  name my_partitioned_model,
  ...
  partitioned_by RANGE(event_date), --or LIST(event_date) for list partitioning
  physical_properties (
     partitions = (
         `default` FROM ("2000-11-14") TO ("2099-11-14") INTERVAL 2 YEAR),
         `p2025` VALUES [("2025-01-01"), ("2026-01-01"))',
         `other` VALUES LESS THAN MAXVALUE
      )
  )
```
to define the partition type (`RANGE`, `LIST` - etc) in `partitioned_by` and then the partitions themselves in `physical_properties`.

I appreciate this might require some tweaks to the parser to achieve, let me know what you think

sqlmesh/core/config/connection.py
```diff
@@ -50,6 +50,8 @@
     "trino",
     # Nullable types are problematic
     "clickhouse",
+    # Do not support table name starts with "_"
```
erindru, line 60: Wow, that's quite a random limitation :)

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+            )
+            .from_(exp.table_("tables", db="information_schema"))
+            .where(exp.column("table_schema").eq(to_schema(schema_name).db))
+        )
+        if object_names:
+            query = query.where(exp.column("table_name").isin(*object_names))
+        df = self.fetchdf(query)
+
+        result = []
+        for row in df.itertuples(index=False, name=None):
```
erindru: `self.fetchall()` might be better here because it already returns a list of tuples. Something like:

```
for schema_name, table_name, table_type in self.fetchall(query):
   ... rest of logic ...
```

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+                        "source_table",
+                    }:
+                        properties[k] = v
+
+                if properties:
+                    props = []
+                    for k, v in properties.items():
+                        v_value = v.this if isinstance(v, exp.Literal) else str(v)
+                        props.append(f"'{k}'='{v_value}'")
+                    doris_clauses.append(f"PROPERTIES ({', '.join(props)})")
```
erindru: This is getting into the realm of raw string concatenation which we generally try to avoid.

SQLGlot has `exp.Properties` for this, eg:

```
>>> exp.Properties(expressions=[
   exp.Property(this="foo", value=exp.Literal.string("bar")),
   exp.Property(this="baz", value=exp.Literal.string("bing")),
]).sql(dialect="doris")

"PROPERTIES ('foo'='bar', 'baz'='bing')"
```

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+
+        # Generate the query SQL
+        with source_queries[0] as query:
+            query_sql = query.sql(dialect=self.dialect, identify=True)
+
+            # Build the complete SQL following Doris grammar order
+            view_name_sql = exp.to_table(view_name).sql(dialect=self.dialect, identify=True)
+
+            # Start with CREATE MATERIALIZED VIEW
+            clauses = [f"CREATE MATERIALIZED VIEW {view_name_sql}"]
```
erindru: Is there a reason why you cant use AST nodes for this? To see what node types you could use, you can leverage the SQLGlot parser:

```
>>> parse_one("create materialized view foo as select 1", dialect="doris")
Create(
  this=Table(
    this=Identifier(this=foo, quoted=False)),
    ...etc
```

Generating AST nodes as strings in the target dialect and then using string formatting to put them back inside another string (like you do here with `view_name`) should be avoided if possible

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+
+    def _rename_table(
+        self,
+        old_table_name: TableName,
+        new_table_name: TableName,
+    ) -> None:
+        old_table_sql = exp.to_table(old_table_name).sql(dialect=self.dialect, identify=True)
+        new_table_sql = exp.to_table(new_table_name).sql(dialect=self.dialect, identify=True)
+
+        self.execute(f"ALTER TABLE {old_table_sql} RENAME {new_table_sql}")
```
erindru: Did the superclass implementation of this not work? It does:

```
>>> exp.rename_table("table_a", "table_b").sql(dialect="doris")
'ALTER TABLE table_a RENAME TO table_b'
```

If `RENAME TO` is incorrect syntax, this needs to be fixed in upstream SQLGlot rather than worked around here

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+                    )
+                    add_partition = False
+            if add_partition:
+                partitioned_by_expr = table_properties_copy.pop("partitioned_by_expr", None)
+
+                # If partitioned_by is provided but partitioned_by_expr is not, add dynamic partition properties
+                if partitioned_by and not partitioned_by_expr:
+                    # Define the required dynamic partition properties
+                    dynamic_partition_props = {
+                        "dynamic_partition.enable": "true",
```
erindru: What are these default values? They seem kind of arbitrary, is this a common idiom when using Doris?

sqlmesh/core/engine_adapter/doris.py
```diff
@@ -0,0 +1,1048 @@
...
+                v = exp.Literal.string(str(v))
+            generic_properties.append(exp.Property(this=exp.Literal.string(str(k)), value=v))
+        if generic_properties:
+            properties.extend(generic_properties)
+
+        if properties:
+            return exp.Properties(expressions=properties)
+        return None
+
+    def _get_temp_table(
```
erindru: I initially thought that this was a direct copy of the superclass method, and it mostly is - except I can see you've removed the leading underscores from `__temp_` because Doris doesnt support leading underscores.

Rather than copy+paste the entire superclass method, perhaps a new method could be added to `EngineAdapter` called something like `_get_temp_table_name()` that returns the usual table name string by default but you can override for Doris to return one without leading underscores?

sqlmesh/core/model/meta.py
```diff
@@ -464,9 +464,23 @@ def batch_concurrency(self) -> t.Optional[int]:
     @cached_property
     def physical_properties(self) -> t.Dict[str, exp.Expression]:
         """A dictionary of properties that will be applied to the physical layer. It replaces table_properties which is deprecated."""
+        properties = {}
         if self.physical_properties_:
-            return {e.this.name: e.expression for e in self.physical_properties_.expressions}
-        return {}
+            properties = {e.this.name: e.expression for e in self.physical_properties_.expressions}
+
+        # For INCREMENTAL_BY_UNIQUE_KEY models for doris dialect, add the unique_key to physical_properties
```
erindru: This doesn't seem right. `IncrementalByUniqueKeyStrategy` passes the key as a kwarg called `unique_key` when it calls `engine_adapter.merge(...)`.

At this point, within the engine adapter, could you inject it into the table properties?

I notice that it [doesnt seem to do it](https://github.com/TobikoData/sqlmesh/blob/8fbf26c3b52c98285254e61e1457f9e65938d99b/sqlmesh/core/snapshot/evaluator.py#L1647) on the first insert, so if it needs to be known at creation time, I think the thing to do is modify the `_replace_query_for_model` call to pass `model.unique_key` in the [replace_query()](https://github.com/TobikoData/sqlmesh/blob/8fbf26c3b52c98285254e61e1457f9e65938d99b/sqlmesh/core/snapshot/evaluator.py#L1399) call and then you can access it in the Doris engine adapter

tests/core/engine_adapter/integration/__init__.py
```diff
@@ -493,6 +494,18 @@ def get_table_comment(
             """
         elif self.dialect == "clickhouse":
             query = f"SELECT name, comment FROM system.tables WHERE database = '{schema_name}' AND name = '{table_name}'"
+        elif self.dialect == "doris":
+            # Doris uses MySQL-compatible information_schema
```
erindru: If that's true, should doris be added to the `elif self.dialect in ["mysql", "snowflake"]` branch instead?

tests/core/engine_adapter/integration/__init__.py
```diff
@@ -588,6 +601,18 @@ def get_column_comments(
         elif self.dialect in ["spark", "databricks", "clickhouse"]:
             query = f"DESCRIBE TABLE {schema_name}.{table_name}"
             comment_index = 2 if self.dialect in ["spark", "databricks"] else 4
+        elif self.dialect == "doris":
+            # Doris uses MySQL-compatible information_schema
```
erindru: Same for this, can it be added to the MySQL branch?

tests/core/engine_adapter/integration/docker/compose.doris.yaml
```diff
@@ -0,0 +1,80 @@
+services:
+  fe-01:
+    image: apache/doris:fe-2.1.10
+    container_name: doris-fe-01
+    hostname: fe-01
+    environment:
+      - FE_SERVERS=fe1:172.20.80.2:9030
```
erindru, line 7: FMI: why do we need to hardcode IP addresses like this? Does Doris not support DNS lookup?

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -336,13 +350,26 @@ def test_materialized_view(ctx_query_and_df: TestContext):
...
+        ctx.compare_with_current(view, input_data)
     # Make sure that dropping a materialized view also works
-    ctx.engine_adapter.drop_view(view, materialized=True)
-    results = ctx.get_metadata_results()
-    assert len(results.materialized_views) == 0
+    if ctx.engine_adapter.dialect == "doris":
+        # Wait for the materialized view to be created
+        import time
+
+        time.sleep(1)
```
erindru: `time.sleep()`'s like this tend to lead to brittle tests. Are you able to use the `tenacity` module used elsewhere in the codebase instead? It can retry on a cadence and give up after a certain limit has passed

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -555,6 +582,35 @@ def test_insert_overwrite_by_time_partition(ctx_query_and_df: TestContext):
...
+    if ctx.dialect == "doris":
+        ds_type = "date"
+
+    # Get current year and create dates for testing. Doris cannot have more than 500 history partitions.
+    current_year = datetime.now().year
+    current_date = datetime(current_year, 1, 1)
+
+    # Convert dates based on ds_type
+    if ctx.dialect == "doris":
+        # For Doris with DATE type, use pandas date objects
```
erindru: Why is this necessary vs what this test was doing before?

erindru (changes requested):
Thanks for this very complete PR @xinge-ji , nice work!

I've completed a first pass reviewing it, let me know if anything seems off. 

I've never used Doris before so i'm approaching it from the perspective of how the other engine adapters tend to work and the general concepts established within the SQLMesh codebase

docs/integrations/engines/doris.md
xinge-ji: I've removed the reference to `_build_table_properties_exp` from the doc.

docs/integrations/engines/doris.md
xinge-ji: I have adjusted the parser to support partitioned_by `RANGE(event_date)` / `LIST(values)` syntax and defined the partition ranges/bounds in `physical_properties` as `partitions` as you suggested.

sqlmesh/core/engine_adapter/doris.py
xinge-ji: Thanks for the suggestion, I have updated the code.

sqlmesh/core/engine_adapter/doris.py
xinge-ji: I've refactored the implementation to avoid raw string concatenation

sqlmesh/core/engine_adapter/doris.py
xinge-ji: This is a leftover from a previous workaround for a SQLGlot limitation. That limitation has been fixed and the code refactored

sqlmesh/core/engine_adapter/doris.py
xinge-ji: I have fixed this in SQLGlot and removed this

sqlmesh/core/engine_adapter/doris.py
xinge-ji: These default values are indeed somewhat arbitrary. They were added primarily to satisfy the `test_integration` requirements for partition handling in the test environment.

sqlmesh/core/engine_adapter/doris.py
xinge-ji: I noticed the existing `_get_temp_table_name()` method in the base EngineAdapter after your suggestion, and I've now overridden it in the Doris adapter to handle the underscore restriction

sqlmesh/core/model/meta.py
xinge-ji: I've modified the implementation to pass `unique_key` through `_replace_query_for_model()` to `replace_query()`

tests/core/engine_adapter/integration/__init__.py
xinge-ji: Good catch, I have moved it to the `mysql` branch

tests/core/engine_adapter/integration/__init__.py
xinge-ji: I have moved it to the mysql branch

tests/core/engine_adapter/integration/docker/compose.doris.yaml
xinge-ji: The official [Docker example](https://github.com/apache/doris/blob/cd14249caab148b7e03a14adff36a43bdbcf2be5/docker/runtime/docker-compose-demo/build-cluster/docker-compose/3fe_3be/docker-compose.yaml) uses hardcoded IPs. When I was setting up the development environment, I attempted to replace the IPs with localhost for simplicity, but the internal container networking did not resolve it correctly, and the setup failed.

tests/core/engine_adapter/integration/test_integration.py
xinge-ji: Sure, I switched to `tenacity`

tests/core/engine_adapter/integration/test_integration.py
xinge-ji: The old test used hardcoded dates from `2022-01-01`. In a Doris table with dynamic partitioning enabled, there's a default limit (`max_dynamic_partition_num`) of 500 partitions. If the date in the test is too far in the past (like 2022), it can fall outside the range of partitions that Doris automatically manages, causing the insert to fail.

sqlmesh/core/engine_adapter/doris.py
xinge-ji: I updated the date definitions in `test_insert_overwrite_by_time_partition` and `test_insert_overwrite_by_time_partition_source_columns` (from `test_integration.py`) and removed these default values.

StuffbyYuki:
@erindru hey man, not sure how active you're on this repo but it'd be awesome if we could get your another review on this!

StuffbyYuki:
@xinge-ji Can you take a look at the conflicts that might need to be resolved?

PR #5037, opened 2025-07-28
Fix: Inconsistent behaviour when a macro returns a list containing a single array vs multiple arrays

### LINKED ISSUE #4536: `each` passed to `@reduce` has inconsistent behaviour
Consider the following:

```python
>>> me=MacroEvaluator()

# @reduce over an array of one element, returns the first element
>>> me.transform(parse_one("@reduce([[1]], (x,y) -> x+y)"))
Array(
  expressions=[
    Literal(this=1, is_string=False)])

# @each wraps items in an array, returns an Array of array
>>> me.transform(parse_one("@each([1], a -> [@a])"))
[Array(
  expressions=[
    Literal(this=1, is_string=False)])]

# combining these two, we would expect @reduce to still return the first element, but we get the literal 1 instead:
>>> me.transform(parse_one("@reduce(@each([1], a -> [a]), (x,y) -> x+y)"))
Literal(this=1, is_string=False)

```

erindru:
Closes #4536 

Prior to this, the following behavior occurred:

```python
>>> me = MacroEvaluator()

>>> me.transform(parse_one("select @each([1], a -> [@a])")).sql()
'SELECT ARRAY(1)' 

>>> me.transform(parse_one("select @reduce([[1]], (x,y) -> x + y)")).sql()
'SELECT ARRAY(1)' 

>>> me.transform(parse_one("select @reduce(@each([1, 2], a -> [@a]), (x,y) -> x + y)")).sql() 
'SELECT ARRAY(1) + ARRAY(2)'

>>> me.transform(parse_one("select @reduce(@each([1], a -> [@a]), (x,y) -> x + y)")).sql() 
'SELECT 1'
```

That last item is unexpected, it should be `SELECT ARRAY(1)` because `@each` returns 1 item of type array.

Unfortunately, `_norm_var_arg_lambda()` [assumes this item is varargs](https://github.com/TobikoData/sqlmesh/blob/9b04f45d77be5e1260fce6ae9edbe3bf891840eb/sqlmesh/core/macros.py#L627) and unpacks it, rather than passing it as-is.

Sometimes, we do actually want it to be unpacked though. If we didnt, `select @each([1, 2], a -> [@a])` would result in `SELECT ARRAY(ARRAY(1), ARRAY(2))` instead of `SELECT ARRAY(1), ARRAY(2)`.

So this PR wraps the output of a macro function in an extra `exp.Array` to prevent `_norm_var_arg_lambda()` from treating it as varargs, but _only if it's being used as input to another function_, which preserves the semantics of still unpacking it for a `SELECT` column list.


erindru:
I guess every bug fix is breaking if someone was depending on the buggy behaviour

PR #5039, opened 2025-07-28
refactor(lsp): move references to internal types

benfdking:
[ci skip]


sqlmesh/core/linter/rule.py
```diff
@@ -30,6 +30,19 @@ class Position:
     line: int
     character: int
 
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, Position):
+            return NotImplemented
+        return (self.line, self.character) == (o.line, o.character)
+
+    def __gt__(self, o: "Position") -> bool:
```
georgesittas: Why is this `o: "Position"` but for `__eq__` we have `o: object`?

benfdking: I have just simplified it and used the default 🙏 It was a bit arbitrary and copied from lspprotocal

PR #5040, opened 2025-07-28
Chore: do not fail if sqlglot optimization rules raise an exception

georgesittas:
Sometimes, rendering may fail without raising a `SqlglotError`, e.g., if there is a bug and one of `qualify`, `annotate_types` or `simplify` fail. I believe logging those errors is helpful, but failing to load altogether seems too abrupt. This PR intercepts those exceptions and logs a warning instead of re-raising them.


georgesittas:
Discussed offline– closing this because users have the option of setting `optimize_query false` and bypassing uncaught errors.

PR #5045, opened 2025-07-28
Fix!: normalize blueprint variables

georgesittas:
The following model fails to load today, because blueprint variables are not normalized:

```sql
MODEL (
  name test_schema.test_model,
  kind FULL,
  blueprints (
    (Test_Var := 1)
  )
);

SELECT
  @test_var AS col
```

This PR aims to address this issue. I also added a migration script that normalizes all existing blueprint
variable keys in the `python_env`, unless there are keys that collide when normalized. In that case, we
raise an `AssertionError` to avoid messing up the state. This allows the user to rename their blueprint
variables in order to avoid this issue and then re-run the migration.

### DIFF
--- a/sqlmesh/migrations/v0087_normalize_blueprint_variables.py
+++ b/sqlmesh/migrations/v0087_normalize_blueprint_variables.py
@@ -1,0 +1,134 @@
+"""
+Normalizes blueprint variables, so Customer_Field is stored as customer_field in the `python_env`:
+
+MODEL (
+  ...
+  blueprints (
+    Customer_Field := 1
+  )
+);
+
+SELECT
+  @customer_field AS col
+"""
+
+import json
+import logging
+from dataclasses import dataclass
+
+from sqlglot import exp
+from sqlmesh.utils.migration import index_text_type, blob_text_type
+
+
+logger = logging.getLogger(__name__)
+
+
+SQLMESH_BLUEPRINT_VARS = "__sqlmesh__blueprint__vars__"
+
+
+# Make sure `SqlValue` is defined so it can be used by `eval` call in the migration
+@dataclass
+class SqlValue:
+    """A SQL string representing a generated SQLGlot AST."""
+
+    sql: str
+
+
+def migrate(state_sync, **kwargs):  # type: ignore
+    import pandas as pd
+
+    engine_adapter = state_sync.engine_adapter
+    schema = state_sync.schema
+    snapshots_table = "_snapshots"
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    migration_needed = False
+    new_snapshots = []
+
+    for (
+        name,
+        identifier,
+        version,
+        snapshot,
+        kind_name,
+        updated_ts,
+        unpaused_ts,
+        ttl_ms,
+        unrestorable,
+    ) in engine_adapter.fetchall(
+        exp.select(
+            "name",
+            "identifier",
+            "version",
+            "snapshot",
+            "kind_name",
+            "updated_ts",
+            "unpaused_ts",
+            "ttl_ms",
+            "unrestorable",
+        ).from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        node = parsed_snapshot["node"]
+        python_env = node.get("python_env") or {}
+
+        # Intentionally checking for falsey value here, since that accounts for empty dicts and None
+        if blueprint_vars_executable := python_env.get(SQLMESH_BLUEPRINT_VARS):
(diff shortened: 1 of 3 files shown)


sqlmesh/migrations/v0087_normalize_blueprint_variables.py
```diff
@@ -0,0 +1,134 @@
...
+                        "SQLMesh version to the previously-installed one, rename either of these variables, "
+                        "apply the corresponding plan and try again."
+                    )
+
+                    del blueprint_vars[var]
+                    blueprint_vars[lowercase_var] = value
+                    migration_needed = True
+
+            if migration_needed:
+                blueprint_vars_executable["payload"] = repr(blueprint_vars)
```
izeigerman, line 102: `migration_needed` seems to be shared across all snapshots. Does this mean that it's enough for at least one snapshot to have blueprint variables for us to rewrite all of them?

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Intentionally checking for falsey value here, since that accounts for empty dicts and None
+        if blueprint_vars_executable := python_env.get(SQLMESH_BLUEPRINT_VARS):
+            blueprint_vars = eval(blueprint_vars_executable["payload"])
+
+            for var, value in dict(blueprint_vars).items():
+                lowercase_var = var.lower()
+                if var != lowercase_var:
+                    # Ensures that we crash instead of overwriting snapshot payloads incorrectly
+                    assert lowercase_var not in blueprint_vars, (
```
izeigerman: I don't think this suggestion helps, because the snapshot will still remain in the state despite them making changes locally and applying them. I don't think we're allowed to fail.

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
```diff
@@ -0,0 +1,134 @@
...
+        ).from_(snapshots_table),
+        quote_identifiers=True,
+    ):
+        parsed_snapshot = json.loads(snapshot)
+        node = parsed_snapshot["node"]
+        python_env = node.get("python_env") or {}
+
+        # Intentionally checking for falsey value here, since that accounts for empty dicts and None
+        if blueprint_vars_executable := python_env.get(SQLMESH_BLUEPRINT_VARS):
+            blueprint_vars = eval(blueprint_vars_executable["payload"])
```
izeigerman, line 81: should this be `literal_eval`?

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
georgesittas: Yeah, that conclusion is correct. To be honest, I didn't give this much thought. We generally don't have this pattern where we check if a migration is needed, and in a couple other places where we _do_ have it, the same logic applies (see v0085 and v0081).

The reason I avoided trying to optimize this further and do finer-grained updates was (1) the fact that it's not an established pattern and (2) I didn't consider it bad enough to increase the risk of running this script.

Interested to see your thoughts around these.

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
georgesittas: You're probably right, I missed the fact that even if they applied the change to prod, they could have the same problematic snapshot in other environments that would need to all be amended.

Any recommendations around this? Should we warn and treat it as a super edge case, not taking any measures? Or should I just update the error message to let the user know that they'll need to run forward only plans for all relevant environments to amend the underlying snapshots and avoid clobbering the ones that are actively used?

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
georgesittas: We used to rely on `literal_eval` but switched to plain `eval` at some point: https://github.com/TobikoData/sqlmesh/blob/9238534b3273d040c878cd162853063d1ec605e5/sqlmesh/utils/metaprogramming.py#L538-L539

Just following that logic here.

izeigerman: ah, yes, because the value can now be a special type

sqlmesh/migrations/v0087_normalize_blueprint_variables.py
georgesittas: Updated to warn instead of raise or overwrite the state: https://github.com/TobikoData/sqlmesh/pull/5045/commits/31216b42d45532faf4a508dd29ca9bbeb413dbb9. The user should now see that warning + a diff on the first plan if they had blueprints that resolved to the same lowercase name in a model.

PR #5019, opened 2025-07-25
feat(vscode): add test recognition to vscode

benfdking:
Clipboard-20250728-153210-711

PR #5049, opened 2025-07-28
feat: location in nomissingexternal models linting error

benfdking:
Depends on 

* [x] https://github.com/TobikoData/sqlmesh/pull/5019

- Adds the range to missing external models 
- Fixes a bug where non registered external models where not highlighting full thing for example in `raw.demographics` was only highlighting `demographics`


sqlmesh/utils/lineage.py
```diff
@@ -0,0 +1,404 @@
...
+    range: Range
+    target_range: t.Optional[Range] = None
+    path: t.Optional[Path] = None
+    """The path of the external model, typically a YAML file, it is optional because
+    external models can be unregistered and so the path is not available."""
+
+    markdown_description: t.Optional[str] = None
+
+
+class LSPCteReference(PydanticModel):
```
izeigerman: Provided that this is now in the common utils package, I wonder whether we should remove the `LSP` prefix from type names.

sqlmesh/utils/lineage.py
```diff
@@ -0,0 +1,404 @@
...
+    references: t.List[Reference] = []
+
+    # Build a scope tree to properly handle nested CTEs
+    try:
+        query = normalize_identifiers(query.copy(), dialect=dialect)
+        root_scope = build_scope(query)
+    except Exception:
+        root_scope = None
+
+    if root_scope:
```
izeigerman: Should we return earlier instead of nesting this whole big block of code?

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -129,10 +137,74 @@ def check_model(self, model: Model) -> t.Optional[RuleViolation]:
...
+            normalize_model_name(
+                table=read_range_from_string(read_file, ref.range),
+                default_catalog=model.default_catalog,
+                dialect=model.dialect,
+            ): ref
+            for ref in references
+            if isinstance(ref, LSPExternalModelReference) and ref.path is None
+        }
+
+        # Ensure that depends_on are a subset of the external references. If not, return generic violation.
```
izeigerman: The condition below doesn't check for a subset. It checks for equality. Which one is not accurate: the code or the comment?

izeigerman (approved):

PR #5063, opened 2025-07-29
Chore: Pin dateparser version

erindru:
`dateparser==1.2.2` seems to cause some [date-related tests](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/22855/workflows/05995099-1db7-41c9-a8b2-340e9079efe2/jobs/250402) to fail

This pins it to "anything compatible with 1.2.1" which fixes things for me locally at least


tobymao:
what's the issue?

erindru:
Still working on pinning down exactly what causes it, I just wanted to provide an option to move forward in the short term.

Basically, [this function](https://github.com/TobikoData/sqlmesh/blob/5ccd57eb91d14cb708467d6c6603ae8472bb8b68/sqlmesh/utils/date.py#L190) started returning `None` in some situations again.

It appears to be nuanced, im trying to track down the root cause which may be in a transitive dependency.

- Fresh venv, dateparser==1.2.2 resolved - tests fail
- Fresh venv, dateparser==1.2.1 resolved - tests pass
- My existing development venv, manually `pip install --upgrade dateparser==1.2.2` - tests pass for some reason

erindru:
Merging this to unblock the pipeline in the short term, i'll raise another PR when I figure out what the root cause is

PR #5072, opened 2025-07-30
Use PandasCursor for Athena dataframes in fetchdf magic

williebsweet:
Only applies to the `%%fetchdf` magic command (although it could be expanded) - replacing the generic pandas read_sql_query() with PandasCursor for improved I/O performance.

Anecdotally, I had queries that were taking 30+ min to execute that are now taking ~2 min.

### DIFF
--- a/sqlmesh/magics.py
+++ b/sqlmesh/magics.py
@@ -611,7 +611,13 @@ class SQLMeshMagics(Magics):
     def fetchdf(self, context: Context, line: str, sql: str) -> None:
         """Fetches a dataframe from sql, optionally storing it in a variable."""
         args = parse_argstring(self.fetchdf, line)
-        df = context.fetchdf(sql)
+        
+        # Check if we're using Athena and use PandasCursor directly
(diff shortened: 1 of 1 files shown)


erindru:
Nope, uv has never been the default, it just happens to be what that user was using.

Generally what we do is something like:

```
$ python -m venv sqlmesh-env
$ . ./sqlmesh-env/bin/activate
(sqlmesh-env) $ make install-dev 
(sqlmesh-env) $ make install-pre-commit #only have to do this once
# make some changes
(sqlmesh-env) $ git commit ....
```

The pre-commit hooks should run on `git commit` and abort the commit if they fail. You should also be able to run `make style` for something similar. The venv that you ran `make install-dev` needs to be active for all of these.

Is there a specific error you're getting?

williebsweet:
@erindru I was able to resolve my issue by not using `uv`. 

But weirdly, `make style` and `make py-style` pass locally, but the latter failed in the last CI run.

PR #5076, opened 2025-07-30
Fix: Reinstate dateparser==1.2.2 with a workaround for the upstream issue

erindru:
`dateparser==1.2.2` introduced a [regression](https://github.com/scrapinghub/dateparser/issues/1282) that essentially causes the following to fail:

```
>>> import dateparser
>>> dateparser.parse("1 minute ago")
# (returns None instead of a datetime)
```

This breaks a bunch of tests in SQLMesh.

However, `dateparser==1.2.2` *also* introduced a cache that shaves several seconds off SQLMesh startup time which can add up over many invocations of SQLMesh.

So this PR adds a workaround for the upstream regression until it can be resolved upstream. The workaround is simply deleting the `dateparser_tz_cache.pkl` pickle file that is distributed with `dateparser==1.2.2` and allowing it to be regenerated on first load.

However, we need to be careful just to delete the one that was distributed from upstream and not one that has been generated locally, since it's the source of the startup time performance improvement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dependencies = [
     "click",
     "croniter",
     "duckdb>=0.10.0,!=0.10.3",
-    "dateparser<=1.2.1",
+    "dateparser>=1.2.2",
     "hyperscript>=0.1.0",
     "importlib-metadata; python_version<'3.12'",
     "ipywidgets",
--- a/sqlmesh/__init__.py
+++ b/sqlmesh/__init__.py
@@ -5,6 +5,13 @@
 
 from __future__ import annotations
 
+# Work around dateparser upstream issue: https://github.com/scrapinghub/dateparser/issues/1282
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
     "click",
     "croniter",
     "duckdb>=0.10.0,!=0.10.3",
-    "dateparser<=1.2.1",
+    "dateparser>=1.2.2",
```
erindru, line 14: This pin forces SQLMesh to use the version of dateparser that contains the performance enhancements (which dont exist in any other version), but also the parsing regression that this PR provides a workaround for

erindru:
Closing [in favour](https://tobikodata.slack.com/archives/C04DTU8RF4J/p1753918736923819?thread_ts=1753912945.533369&cid=C04DTU8RF4J) of keeping the `<= 1.2.1` pin and accepting the performance penalty until dateparser is fixed upstream.

PR #5079, opened 2025-07-31
feat: helper to find key in model block

benfdking:
- to be used by linter in the future

### DIFF
--- a/sqlmesh/core/linter/helpers.py
+++ b/sqlmesh/core/linter/helpers.py
@@ -158,3 +158,55 @@ def get_range_of_model_block(
     return Range(
         start=start_position.to_range(splitlines).start, end=end_position.to_range(splitlines).end
     )
+
+
+def get_range_of_a_key_in_model_block(
+    sql: str,
+    dialect: str,
+    key: str,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/linter/helpers.py
```diff
@@ -158,3 +158,55 @@ def get_range_of_model_block(
...
+    tokens_of_interest = tokens[start_index + 1 : end_index]
+    # Find the key token
+    key_token = next(
+        (
+            t
+            for t in tokens_of_interest
+            if t.token_type is TokenType.VAR and t.text.upper() == key.upper()
+        ),
+        None,
+    )
```
georgesittas, line 202: Just a heads up that this looks good as a best-effort approach, just not sure if it _guarantees_ you'll match the exact token. Think that the token corresponding to the property's key could appear elsewhere in the `MODEL` block

 Probably not worth trying to further improve it, but a simple heuristic if you notice any issues would be to check if the previous token is a comma, since properties are comma-separated.

georgesittas (approved):

Issue #5067, opened 2025-07-30
Cte Tests Not Working


josh-reynolds1999:
I am attempting to follow the instructions laid out here: https://sqlmesh.readthedocs.io/en/latest/concepts/tests/#testing-ctes relating to adding cte tests. I have followed the steps to update the model to include a cte and updated the test to include the cte section. However, if I leave the final output for the query the same or even remove it but I update the expected rows for the cte the test passes. I would have expected the test to fail if the expected rows for the cte are incorrect. There is no configuration I can set up for the test that causes it to fail for issues with the cte expected data.

My model looks like this:

```
MODEL (
  name sqlmesh_example.full_model,
  kind FULL,
  cron '@daily',
  grain item_id,
  audits (assert_positive_order_ids),
);


WITH filtered_orders_cte AS (
  SELECT
    id,
    item_id
  FROM
    sqlmesh_example.incremental_model
  WHERE
    item_id = 1
)
SELECT
  item_id,
  COUNT(DISTINCT id) AS num_orders,
FROM
  filtered_orders_cte
GROUP BY item_id
```

I have my test file as follows which I believe should fail as the cte should only inlcude the records with item_id = 1 whereas I have updated it to select the other record meaning the expected is not the same as the actual:

```
test_example_full_model:
  model: sqlmesh_example.full_model
  inputs:
    sqlmesh_example.incremental_model:
        rows:
        - id: 1
          item_id: 1
        - id: 2
          item_id: 1
        - id: 3
          item_id: 2
  outputs:
    ctes:
      filtered_orders_cte:
        rows:
          - id: 3
            item_id: 2
    query:
      rows:
      - item_id: 1
        num_orders: 2
```

However after running `sqlmesh test` it shows that it successfully ran the test:

<img width="484" height="76" alt="Image" src="https://github.com/user-attachments/assets/801d5c6c-a3af-466f-adb7-3cfa10b36cea" />

PR #5081, opened 2025-07-31
Fix: Unit test CTE failures not being captured

VaggelisD:
Fixes https://github.com/TobikoData/sqlmesh/issues/5067

In `unittest` internals `TestCase::addSubTest` appends the result of that subtest directly to `self.errors` or `self.failures`:

```Python3
    def addSubTest(self, test, subtest, err):
        """Called at the end of a subtest.
        'err' is None if the subtest ended successfully, otherwise it's a
        tuple of values as returned by sys.exc_info().
        """
        # By default, we don't do anything with successful subtests, but
        # more sophisticated test results might want to record them.
        if err is not None:
            if getattr(self, 'failfast', False):
                self.stop()
            if issubclass(err[0], test.failureException):
                errors = self.failures
            else:
                errors = self.errors
            errors.append((subtest, self._exc_info_to_string(err, test)))
            self._mirrorOutput = True

```

However, our custom classes expected that this'd happen through `addError(...)` or `addFailure(...)` so that we can intercept these exceptions (since they get turned into strings before being stored in their parent classes).

To solve this regression, `ModelTestTextResult::addSubTest` now follows the parent implementation and instead calls the respective APIs which do fill our own `original_failures` & `original_error` structures.

The output for that looks like:

```bash
❯ sqlmesh test
F

FAIL: test_example_full_model (/Users/vaggelisd/Desktop/tobiko/test_dir/tests/test.yaml)
         Data mismatch (CTE "filtered_orders_cte")          
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
┃   Row    ┃      id: Expected       ┃     id: Actual      ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━┩
│    0     │            5            │          1          │
└──────────┴─────────────────────────┴─────────────────────┘

Test Failure Summary
Ran 1 tests against duckdb in 0.09 seconds. 

Failed tests (1):
 •<path>::test_example_full_model

``` 

Docs
-------
[unittest.TestResult](https://github.com/python/cpython/blob/main/Lib/unittest/result.py) | [unittest.TestCase](https://github.com/python/cpython/blob/9d3b53c47fab9ebf1f40d6f21b7d1ad391c14cd7/Lib/unittest/case.py) | [unittest.runner](https://github.com/python/cpython/blob/main/Lib/unittest/runner.py)

### DIFF
--- a/tests/integrations/github/cicd/test_github_commands.py
+++ b/tests/integrations/github/cicd/test_github_commands.py
@@ -479,7 +479,7 @@ def test_run_all_test_failed(
     assert (
         """sqlmesh.utils.errors.TestError: some error""" in test_checks_runs[2]["output"]["summary"]
     )
-    assert """Failed tests (1):""" in test_checks_runs[2]["output"]["summary"]
+    assert """Failed tests""" in test_checks_runs[2]["output"]["summary"]
 
     assert "SQLMesh - Prod Plan Preview" in controller._check_run_mapping
     prod_plan_preview_checks_runs = controller._check_run_mapping[


sqlmesh/core/test/definition.py
```diff
@@ -317,6 +317,13 @@ def _to_hashable(x: t.Any) -> t.Any:
             #
             # This is a bit of a hack, but it's a way to get the best of both worlds.
             args: t.List[t.Any] = []
+
+            failed_subtest = ""
+
+            if subtest := getattr(self, "_subtest", None):
+                if cte := subtest.params.get("cte"):
+                    failed_subtest = f" (CTE {cte})"
```
VaggelisD, line 325: This is to differentiate the CTE mismatch from the output one, e.g if both subtests fail in a single test that'd be the output:


```bash
FF

FAIL: test_example_full_model (/Users/vaggelisd/Desktop/tobiko/test_dir/tests/test.yaml)
         Data mismatch (CTE "filtered_orders_cte")          
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
┃   Row    ┃      id: Expected       ┃     id: Actual      ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━┩
│    0     │            5            │          1          │
└──────────┴─────────────────────────┴─────────────────────┘

FAIL: test_example_full_model (/Users/vaggelisd/Desktop/tobiko/test_dir/tests/test.yaml)
                       Data mismatch                        
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
┃   Row    ┃      id: Expected       ┃     id: Actual      ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━┩
│    0     │            2            │          1          │
└──────────┴─────────────────────────┴─────────────────────┘

Test Failure Summary
Ran 1 tests against duckdb in 0.14 seconds. 

Failed tests (1):
 • /Users/vaggelisd/Desktop/tobiko/test_dir/tests/test.yaml::test_example_full_model
```

sqlmesh/core/test/result.py
```diff
@@ -117,4 +122,14 @@ def merge(self, other: ModelTextTestResult) -> None:
...
+
+    def get_fail_and_error_tests(self) -> t.List[ModelTest]:
+        # If tests contain failed subtests (e.g testing CTE outputs) we don't want
+        # to report it as different test failures
+        test_name_to_test = {
+            test.test_name: test
+            for test, _ in self.failures + self.errors
+            if isinstance(test, ModelTest)
+        }
+        return list(test_name_to_test.values())
```
VaggelisD, line 135: Simple abstraction to simplify `console.py`

tests/core/test_test.py
```diff
@@ -2407,8 +2415,8 @@ def test_test_output(tmp_path: Path) -> None:
 
     # Case 3: Assert that concurrent execution is working properly
     for i in range(50):
-        copyfile(original_test_file, tmp_path / "tests" / f"test_success_{i}.yaml")
-        copyfile(new_test_file, tmp_path / "tests" / f"test_failure_{i}.yaml")
+        copy_test_file(original_test_file, tmp_path / "tests" / f"test_success_{i}.yaml", i)
+        copy_test_file(new_test_file, tmp_path / "tests" / f"test_failure_{i}.yaml", i)
```
VaggelisD, line 2419: Previously we copied these files directly but unit tests don't work well if the name is duplicated, I'll follow up with a UX fix

sqlmesh/core/console.py
```diff
@@ -2721,28 +2720,15 @@ def _log_test_details(
...
-        failures = result.failures
-        skipped = result.skipped
-
-        infos = []
-        if failures:
-            infos.append(f"failures={len(failures)}")
-        if errors:
-            infos.append(f"errors={len(errors)}")
-        if skipped:
-            infos.append(f"skipped={skipped}")
```
VaggelisD, line 2739: dead code, `infos` is no longer used

sqlmesh/core/test/definition.py
```diff
@@ -317,6 +317,13 @@ def _to_hashable(x: t.Any) -> t.Any:
             #
             # This is a bit of a hack, but it's a way to get the best of both worlds.
             args: t.List[t.Any] = []
+
+            failed_subtest = ""
+
+            if subtest := getattr(self, "_subtest", None):
```
izeigerman, line 323: I wonder if it even makes sense at this point to rely on subtests rather than write our own context manager which sets the current CTE and then resets it.

izeigerman: Are there any other benefits that subtests offer us here?

VaggelisD: I had the same thought as I was working on this, I think it's merit is that it triggers the callback `addSubTest` when it finishes, otherwise we'd have to schedule that ourselves.

Not really a strong argument for it, but I'm not sure if there's a strong argument against it either. Happy to reconsider if you had a better thought with the custom CM

PR #5085, opened 2025-07-31
Feat: handle additional pseudocolumn types in bigquery when resolving…

z3z1ma:
… external table schemas.


z3z1ma:
my brain sql'ed when it shouldve python'ed - let me amend

PR #5087, opened 2025-07-31
Feat!: Dev-only VDE mode

izeigerman:
This PR adds support for the `dev_only` Virtual Data Environment (VDE) mode. 

In this mode VDE is applied only in development environments, while in production, model tables and views are updated directly, bypassing the virtual layer. This also means that physical tables in production will be created using the original, unversioned model names. Users will continue to benefit from VDE and data reuse across development environments.

The new mode can be enabled in the project configuration:
```yaml
virtual_environment_mode: 'dev_only'
```

Switching the mode for an existing project will result in a complete rebuild of all models in the project.


sqlmesh/core/snapshot/definition.py
```diff
@@ -863,16 +882,19 @@ def merge_intervals(self, other: t.Union[Snapshot, SnapshotIntervals]) -> None:
...
-                self.add_interval(start, end)
+        if self.is_no_rebuild or self.virtual_environment_mode.is_full or not self.is_paused:
+            # If the virtual environment mode is not full we can only merge prod intervals if this snapshot
+            # is currently promoted in production or if it's forward-only / metadata / indirect non-breaking.
+            # Otherwise, we want to ignore any existing intervals and backfill this snapshot from scratch.
+            effective_from_ts = self.normalized_effective_from_ts or 0
+            apply_effective_from = effective_from_ts > 0 and self.identifier != other.identifier
+            for start, end in other.intervals:
+                # If the effective_from is set, then intervals that come after it must come from
+                # the current snapshost.
```
erindru: (historical) nit: snapshost

sqlmesh/core/snapshot/definition.py
```diff
@@ -1013,15 +1035,18 @@ def categorize_as(self, category: SnapshotChangeCategory) -> None:
             category: The change category to assign to this snapshot.
         """
         self.dev_version_ = self.fingerprint.to_version()
-        reuse_previous_version = category in (
+        is_no_rebuild = category in (
```
erindru: Is this logic duplicated here and in the `self.is_no_rebuild()` property?

As a sidenote - thanks for renaming this and `_should_force_breaking_change`, the new names are much clearer as to the intent compared to the original names

docs/guides/configuration.md
```diff
@@ -538,6 +538,44 @@ sqlmesh_md5__d3b07384d113edec49eaa6238ad5ff00__dev
 
 This has a downside that now it's much more difficult to determine which table corresponds to which model by just looking at the database with a SQL client. However, the table names have a predictable length so there are no longer any surprises with identfiers exceeding the max length at the physical layer.
 
+#### Virtual Data Environment modes
```
sungchun12:
```suggestion
#### Virtual Data Environment Modes
```

docs/guides/configuration.md
```diff
@@ -538,6 +538,44 @@ sqlmesh_md5__d3b07384d113edec49eaa6238ad5ff00__dev
...
+
+By default, Virtual Data Environments (VDE) are applied across both development and production environments. This allows SQLMesh to reuse physical tables when appropriate, even when promoting from development to production.
+
+However, users may sometimes prefer their production environment to be non-virtual. The non-exhaustive list of reasons may include:
```
sungchun12:
```suggestion
However, users may prefer their production environment to be non-virtual. The non-exhaustive list of reasons may include:
```

sqlmesh/core/snapshot/definition.py
izeigerman: > Is this logic duplicated here and in the self.is_no_rebuild() property?

yes, but only because `_should_force_breaking_change` relies on `self.change_category` which hasn't been set yet at this point.

erindru: Oh, I was more thinking along the lines of adding a constant something like `NON_REBUILDABLE_CHANGE_CATEGORIES={SnapshotChangeCategory.FORWARD_ONLY, ...}` and then referencing it in both places rather than maintaining two identical lists

sqlmesh/core/config/common.py
```diff
@@ -49,6 +49,35 @@ def __repr__(self) -> str:
...
 
 
+class VirtualEnvironmentMode(str, Enum):
+    """Mode for virtual environment behavior.
+
+    FULL: Use full virtual environment functionality with versioned table names and virtual layer updates.
+    DEV_ONLY: Bypass virtual environments in production, using original unversioned model names.
+    """
+
+    FULL = "full"
```
eakmanrq, line 59: Minor: Is the term "FULL" being used because of "FULL" in auto-categorization? "ALL" feels better to me but I'm wondering if there is some intention behind using "FULL" that I am missing.

sqlmesh/core/config/root.py
```diff
@@ -260,6 +262,11 @@ def _normalize_identifiers(key: str) -> None:
                 "Please specify one or the other"
             )
 
+        if self.plan.use_finalized_state and not self.virtual_environment_mode.is_full:
+            raise ConfigError(
+                "Using the finalized state is only supported when `virtual_environment_mode` is set to `full`."
+            )
```
eakmanrq, line 268: Is this because a "partial" state doesn't make sense if you aren't using VDE in prod? If so, what would be the negative in allowing this? Just doesn't have much meaning?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2483,6 +2515,35 @@ def _truncate_table(self, table_name: TableName) -> None:
...
+        self, target_name: TableName, expected_type: DataObjectType
+    ) -> bool:
+        """Drops a data object if it exists and is not of the expected type.
+
+        Args:
+            target_name: The name of the data object to check.
+            expected_type: The expected type of the data object.
+
+        Returns:
+            True if the data object exists and is of the expected type, False otherwise.
```
eakmanrq: My instinct is that this should return a bool indicating if the drop was performed or not.

sqlmesh/core/plan/stages.py
```diff
@@ -437,11 +438,18 @@ def _get_virtual_layer_update_stage(
         demoted_environment_naming_info: t.Optional[EnvironmentNamingInfo],
         all_snapshots: t.Dict[SnapshotId, Snapshot],
         deployability_index: DeployabilityIndex,
+        is_dev: bool,
     ) -> t.Optional[VirtualLayerUpdateStage]:
-        promoted_snapshots = {s for s in promoted_snapshots if s.is_model and not s.is_symbolic}
-        demoted_snapshots = {s for s in demoted_snapshots if s.is_model and not s.is_symbolic}
+        def _should_update_virtual_layer(snapshot: SnapshotTableInfo) -> bool:
+            # Skip virtual layer update for snapshots with virtual environment support disabled
+            virtual_environment_enabled = is_dev or snapshot.virtual_environment_mode.is_full
```
eakmanrq, line 447: The assumption here is that dev is always a virtual environment right? Wondering if there should be logic like "if is_dev and virtual_environment_mode.includes_dev" and let `virtual_environment_mode.includes_dev` determine if dev is virtual.

Thinking about it more, it seems like the next step of this would be to make specific environments VDE or not. If that was the case then it could be more useful here to do something like `def is_virtual(environment_name: str)` and then right now it would just check if is_full or if dev only then confirm the name isn't prod. Then it would be easily expanded to include support for pinned environment or custom environments or whatever in the future.

Primary concern is an assumption being backed in here that is missed in later iterations.

eakmanrq: Hmm reading this more and I think this isn't the only place that the assumption that dev is always vde is being made. Yeah so I think this can be ignored since the implementation seems strictly focused on the current capability of `VirtualEnvironmentMode` and changing it later would require a refactor in many places.

sqlmesh/core/config/root.py
izeigerman: Actually, since dev-only VDE always assumes a forward-only plan, this check is no longer relevant. Will remove it, thanks

sqlmesh/core/engine_adapter/base.py
izeigerman: Not sure I follow what you mean. If the drop wasn't permitted the operation would fail, right?

sqlmesh/core/plan/stages.py
izeigerman: > Thinking about it more, it seems like the next step of this would be to make specific environments VDE or not. If that was the case then it could be more useful here to do something like def is_virtual(environment_name: str) and then right now it would just check if is_full or if dev only then confirm the name isn't prod. Then it would be easily expanded to include support for pinned environment or custom environments or whatever in the future.

Unfortunately, supporting pinned environments as non-virtual is a lot more work than just doing this check. I'd rather not try to future-proof this API in any way.

sqlmesh/core/engine_adapter/base.py
eakmanrq: I'm not sure what you mean by permitted. 

To help explain my point, the method is called `drop_data_object_on_type_mismatch` and returns a bool. I read the method as conditional and saying it is going to perform some action (drop) on some condition (type mismatch). Therefore I would expect the bool to indicate if the condition was true and the action was performed. Right now though it returns true if the condition is not true and the action was not performed which doesn't match my intuition.

sqlmesh/core/plan/stages.py
eakmanrq: Ok sounds good - I figured that as I read more of the PR. Thanks!

sqlmesh/core/plan/builder.py
```diff
@@ -801,27 +800,6 @@ def _ensure_valid_date_range(self) -> None:
                     f"Plan end date: '{time_like_to_str(end)}' cannot be in the future (execution time: '{time_like_to_str(self.execution_time)}')"
                 )
 
-    def _ensure_no_forward_only_revert(self) -> None:
```
eakmanrq, line 804: Why does allowing users to not have a VDE prod mean that we no longer need this check?

sqlmesh/core/config/root.py
izeigerman: I thought about this more and concluded that indeed this configuration is meaningless when used with dev-only VDEs since there are no previous table versions to go back to.

sqlmesh/core/snapshot/definition.py
```diff
@@ -499,6 +509,10 @@ class SnapshotTableInfo(PydanticModel, SnapshotInfoMixin, frozen=True):
     dev_table_suffix: str
     model_gateway: t.Optional[str] = None
     forward_only: bool = False
+    table_naming_convention: TableNamingConvention = TableNamingConvention.default
```
erindru, line 512: Interesting, I struggled to get this to work. mypy wouldnt let me have it both here and on `SnapshotInfoMixin`

sqlmesh/core/snapshot/definition.py
```diff
@@ -1535,14 +1567,20 @@ def create(
...
                 continue
-            # Make sure that the node is deployable according to all its parents
-            this_deployable = all(
-                children_deployability_mapping[p_id]
-                for p_id in snapshots[node].parents
-                if p_id in children_deployability_mapping
-            )
+            snapshot = snapshots[node]
+
+            if not snapshot.virtual_environment_mode.is_full:
```
erindru, line 1568: I see a lot of `if not snapshot.virtual_environment_mode.is_full:`

rather than `if snapshot.virtual_environment_mode.is_dev_only:`

I'm guessing this is defensive in case you were planning on adding other variants to that enum but I do think it makes the code harder to follow

tests/core/engine_adapter/test_base.py
```diff
@@ -2713,6 +2730,27 @@ def test_replace_query(make_mocked_engine_adapter: t.Callable, mocker: MockerFix
...
+        adapter,
+        "_get_data_objects",
+        return_value=[DataObject(schema="", name="test_table", type="view")],
+    )
+
+    adapter.replace_query(
+        "test_table", parse_one("SELECT a FROM tbl"), {"a": exp.DataType.build("INT")}
+    )
+
+    # TODO: Shouldn't we enforce that `a` is casted to an int?
```
erindru: Isnt that what the query is doing? Rather than cast it twice, it's just casting it at the final step?

tests/core/test_test.py
```diff
@@ -2153,44 +2153,6 @@ def test_test_generation_with_timestamp(tmp_path: Path) -> None:
     }
 
 
-def test_test_generation_with_decimal(tmp_path: Path, mocker: MockerFixture) -> None:
```
erindru, line 2156: Why did we lose this test?

tests/core/test_plan.py
```diff
@@ -1097,53 +1096,6 @@ def test_end_validation(make_snapshot, mocker: MockerFixture):
     assert restatement_prod_plan_builder.build().end == "2022-01-04"
 
 
-def test_forward_only_revert_not_allowed(make_snapshot, mocker: MockerFixture):
```
erindru, line 1100: Is this test no longer valid even when operating in "normal" mode on a native project?

erindru (approved):

sqlmesh/core/engine_adapter/base.py
izeigerman: The idea here was to save the `table_exists` call downstream since we get the data objects already. That's why it returns what it returns. Given this context how do you suggest to update this method?

sqlmesh/core/plan/builder.py
izeigerman: This check has been obsolete for a long time now. Users should never really reach this point, since there's nothing a user can do to recover from this easily. The suggestion in the error message wouldn't help to resolve the issue. 

In any case, the reverting issue is addressed upstream with the "unrestorable" mechanics and so this check is largely useless. The reason I'm removing it here is because the introduction of this feature causes false-positive failures for this check under some circumstances. So I had to either update this check or get rid of it.

sqlmesh/core/snapshot/definition.py
izeigerman: it is indeed defensive, since many of these conditions apply no matter what option it's going to be unless it's full

tests/core/test_plan.py
izeigerman: Users can never end up in this situation due to the "unrestorable" mechanics that we have in place: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/context.py#L2724. So the test was largely obsolete. More context: https://github.com/TobikoData/sqlmesh/pull/5087#discussion_r2268139652

tests/core/engine_adapter/test_base.py
izeigerman: A copy paste from a different test. Thanks

tests/core/test_test.py
izeigerman: Because it's a badly formed test that assumes that `fetchdf` is only called once to return the values for the input query + it feeds the dataframe directly, regardless of what’s in the actual query. So I'm questioning the value of this.

docs/guides/configuration.md
```diff
@@ -538,6 +538,44 @@ sqlmesh_md5__d3b07384d113edec49eaa6238ad5ff00__dev
...
+
+- Integration with third-party tools and platforms, such as data catalogs, may not work well with the virtual view layer that SQLMesh imposes by default
+- A desire to rely on time travel features provided by cloud data warehouses such as BigQuery, Snowflake, and Databricks
```
sungchun12, line 548:
```suggestion
- A desire to rely on time travel features provided by cloud data warehouses such as BigQuery, Snowflake, and Databricks. This is mainly for hot fixes in production.
```

docs/guides/configuration.md
```diff
@@ -538,6 +538,44 @@ sqlmesh_md5__d3b07384d113edec49eaa6238ad5ff00__dev
...
+
+    ```python linenums="1"
+    from sqlmesh.core.config import Config
+
+    config = Config(
+        virtual_environment_mode="dev_only",
+    )
+    ```
+
+As the name suggests, 'dev-only' mode means that VDE is applied only in development environments, while in production, model tables and views are updated directly, bypassing the virtual layer. This also means that physical tables in production will be created using the original, unversioned model names. Users will still benefit from VDE and data reuse across development environments.
```
sungchun12:
```suggestion
'dev-only' mode means that VDE is applied only in development environments. While in production, model tables and views are updated directly and bypass the virtual layer. This also means that physical tables in production will be created using the original, **unversioned** model names. Users will still benefit from VDE and data reuse across development environments.
```

docs/guides/configuration.md
```diff
@@ -538,6 +538,44 @@ sqlmesh_md5__d3b07384d113edec49eaa6238ad5ff00__dev
...
+
+Please note the following tradeoffs when enabling this mode:
+
+- All data inserted in development environments is used only for [preview](../concepts/plans.md#data-preview-for-forward-only-changes) and will **not** be reused in production
+- Reverting a model to a previous version will be applied going forward and may require an explicit data restatement
+
+!!! warning
+    Switching the mode for an existing project will result in a complete rebuild of all models in the project. Refer to the [Table Migration Guide](./table_migration.md) to migrate existing tables without rebuilding them from scratch.
```
sungchun12:
```suggestion
    Switching the mode for an existing project will result in a **complete rebuild** of all models in the project. Refer to the [Table Migration Guide](./table_migration.md) to migrate existing tables without rebuilding them from scratch.
```

sungchun12 (approved):
Quick docs suggestions.

codykonior-hsi:
> Switching the mode for an existing project will result in a complete rebuild of all models in the project.

For anyone finding this at a later stage, that rebuild is not automatic.

You must destroy and plan the environment from scratch when changing from full to dev_only or vice-versa.

PR #5091, opened 2025-08-04
Builtin Linting Rule: cronintervalalignment

sungchun12:
This checks whether cron schedules will behave as expected. This prevents misalignment in scheduling expectations. 

For example, if I have a DAG like this: step_1(`@weekly`) -> step_2(`@daily`) -> step_3(`*/5 * * * *`)

step_2 and step_3 will only process empty intervals because they're anchored to the intervals processed in step_1. This rule prevents confusion on why step_2 and step_3 are wastefully backfilling empty intervals. The fix is to align the schedules where a downstream model's cron is the same or has a longer interval than an upstream model's.

This rule allows the below DAG for step_2 as it's a valid use case. step_2 should still run daily and not emit a rule violation because it's processing backfilled intervals from step_1b. step_3 will still violate this linting rule.

[step_1a(`@weekly`). step_1b(`@hourly`)] -> step_2(`@daily`) -> step_3(`*/5 * * * *`)

In use:
image

image

### DIFF
--- a/sqlmesh/core/linter/rules/builtin.py
+++ b/sqlmesh/core/linter/rules/builtin.py
@@ -7,7 +7,6 @@ import typing as t
 from sqlglot.expressions import Star
 from sqlglot.helper import subclasses
 
-from sqlmesh.core.node import IntervalUnit
 from sqlmesh.core.constants import EXTERNAL_MODELS_YAML
 from sqlmesh.core.dialect import normalize_model_name
 from sqlmesh.core.linter.helpers import (
@@ -279,9 +278,7 @@ class CronIntervalAlignment(Rule):
     """Upstream model has a cron expression with longer intervals than downstream model."""
 
     def check_model(self, model: Model) -> t.Optional[t.List[RuleViolation]]:
-        placeholder_start_date = "2020-01-01 10:00:00"
-
-        this_model_cron_next = model.cron_next(placeholder_start_date)
+        placeholder_start_date = "2020-01-01 00:25:00"
 
         violations = []
         for upstream_model_name in model.depends_on:
@@ -296,22 +293,23 @@ class CronIntervalAlignment(Rule):
             if upstream_model.kind.name in skip_kinds:
                 continue
 
-            upstream_model_cron_next = upstream_model.cron_next(placeholder_start_date)
+            for _ in range(12):
+                this_next = model.cron_next(placeholder_start_date)
+                upstream_next = upstream_model.cron_next(placeholder_start_date)
+                current_time = this_next
 
-            upstream_cron_interval_unit = IntervalUnit.from_cron(upstream_model.cron)
-            this_cron_interval_unit = IntervalUnit.from_cron(model.cron)
-            rule_violation = RuleViolation(
-                rule=self,
-                violation_msg=f"Upstream model {upstream_model_name} has longer cron interval ({upstream_model.cron}) "
-                f"than this model ({model.cron})",
-            )
+                # Find the first iteration pair where upstream runs after downstream
+                if upstream_next > this_next:
+                    violations.append(
+                        RuleViolation(
+                            rule=self,
+                            violation_msg=f"Upstream model {upstream_model_name} runs less frequently ({upstream_model.cron}) than this model ({model.cron})",
+                        )
+                    )
+                    break  # Found violation, stop checking this upstream model
+                else:
+                    return None
 
-            if upstream_model_cron_next > this_model_cron_next:
-                violations.append(rule_violation)
-            elif upstream_cron_interval_unit.seconds > this_cron_interval_unit.seconds:
-                violations.append(rule_violation)
-            elif upstream_model_cron_next <= this_model_cron_next:
-                return None
         return violations
 
 
--- a/tests/core/linter/test_builtin.py
+++ b/tests/core/linter/test_builtin.py
@@ -187,14 +187,19 @@ def test_no_missing_external_models_with_existing_file_not_ending_in_newline(
             "@weekly",
             "@daily",
             1,
-            'Upstream model "memory"."sushi"."step_1" has longer cron interval (@weekly) than this model (@daily)',
+            'Upstream model "memory"."sushi"."step_1" runs less frequently (@weekly) than this model (@daily)',
         ),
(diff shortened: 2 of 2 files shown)


sqlmesh/core/linter/rules/builtin.py
```diff
@@ -274,4 +274,34 @@ def create_fix(self, model_name: str) -> t.Optional[Fix]:
...
+
+        for upstream_model_name in model.depends_on:
+            upstream_model = self.context.get_model(upstream_model_name)
+
+            # Skip if upstream model doesn't exist
+            if upstream_model is None:
+                continue
+
+            # Skip model kinds since they don't run on cron schedules
+            skip_kinds = ["EXTERNAL", "EMBEDDED", "SEED"]
```
izeigerman: I suggest using enum values instead of hardcoded strings (`ModelKindName`)

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -274,4 +274,34 @@ def create_fix(self, model_name: str) -> t.Optional[Fix]:
         )
 
 
+class CronValidator(Rule):
```
izeigerman: Should the name be more specific? I feel like there can be multiple approaches to validating cron

sungchun12: What about this? `CronAlignmentValidator`

sungchun12: Or `CronIntervalAlignment`

izeigerman: the last one is good, thanks

sqlmesh/core/linter/rules/builtin.py
sungchun12: updated

sqlmesh/core/linter/rules/builtin.py
sungchun12: updated

tests/core/linter/test_builtin.py
```diff
@@ -172,3 +177,111 @@ def test_no_missing_external_models_with_existing_file_not_ending_in_newline(
...
+        d.parse("MODEL (name memory.sushi.step_a, cron '@weekly'); SELECT * FROM (SELECT 1)")
+    )
+
+    upstream_model_b = load_sql_based_model(
+        d.parse("MODEL (name memory.sushi.step_b, cron '@hourly'); SELECT * FROM (SELECT 1)")
+    )
+
+    downstream_model = load_sql_based_model(
+        d.parse(
+            "MODEL (name memory.sushi.step_c, cron '@daily', depends_on ['memory.sushi.step_1', 'memory.sushi.step_b']); SELECT * FROM (SELECT 1)"
```
sungchun12: ah shoot, I need to update `step_1` to `step_a`

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -274,4 +274,40 @@ def create_fix(self, model_name: str) -> t.Optional[Fix]:
...
+                continue
+
+            # Skip model kinds since they don't run on cron schedules
+            skip_kinds = {ModelKindName.EXTERNAL, ModelKindName.EMBEDDED, ModelKindName.SEED}
+            if upstream_model.kind.name in skip_kinds:
+                continue
+
+            upstream_model_cron_next = upstream_model.cron_next(placeholder_start_date)
+
+            if upstream_model_cron_next > this_model_cron_next:
```
izeigerman: I thought about this more and I don't think this is the right way to handle this. Imagine graph A (upstream) <- B (downstream) with crons
- A: `5 * * * *` (run at 5th minute of every hour)
- B: `0 * * * *` (run every hour)

Both models are hourly, so technically the upstream's cron interval is not "longer" as the violation suggests. However, B will run **before** A which might not be the desired behavior.

But if it was this instead:
- A: `15 10 * * *` (run at 10:15AM of every day)
- B: `0 * * * *` (run every hour)

Your check will return false given the current placeholder start date, because `cron_next` for A will be `10:15` but for B it will be `11:00`. In this case upstream is daily, and its interval is indeed longer than Bs (hourly). You probably want your check to fail but it won't.

So if the length of the interval is what you're trying to capture here then you want to do the following instead:
```
upstream_cron_interval_unit = IntervalUnit.from_cron(upstream_model.cron)
this_cron_interval_unit = IntervalUnit.from_cron(model.cron)

if upstream_cron_interval_unit.seconds > this_cron_interval_unit.seconds:
    # violation
```
and move away from the placehoder date.

sungchun12: This was actually my first approach, but it won't work in isolation because the `IntervalUnit` is limited: https://github.com/TobikoData/sqlmesh/blob/5861dbaac291113b504380456cc1b38a897611cc/sqlmesh/core/node.py#L39-L45

For example, `@weekly` and `@daily` will be interpreted as being equal in interval unit seconds because they are both categorized as `DAY`.

```python
upstream_cron_interval_unit = IntervalUnit.from_cron(upstream_model.cron)
this_cron_interval_unit = IntervalUnit.from_cron(model.cron)

if upstream_cron_interval_unit.seconds > this_cron_interval_unit.seconds:
    # violation
```

**New Approach:** Use an OR condition where if upstream model's next cron is greater OR its interval unit in seconds is greater, return a violation.

Example in action for your first scenario: You'll notice the placeholder date implementation works as expected. This is NOT a rule violation because model B runs AFTER model A. 

A: 5 * * * * (run at 5th minute of every hour)
B: 0 * * * * (run every hour)

```shell
(Pdb) upstream_cron_interval_unit.seconds
3600
(Pdb) this_cron_interval_unit.seconds
3600
(Pdb) upstream_model_cron_next
datetime.datetime(2020, 1, 1, 10, 5, tzinfo=datetime.timezone.utc)
(Pdb) this_model_cron_next
datetime.datetime(2020, 1, 1, 11, 0, tzinfo=datetime.timezone.utc)
```

sungchun12: Note: it's important WHEN models are linted because **when** it's to deployed to prod determines the next run. For nuanced cron schedules, a model applied at a specific time can emit violation vs. not.

sungchun12:
Closing this to prioritize other work

PR #5093, opened 2025-08-05
Chore: Introduce Claude subagents to the repo

izeigerman:
Introduces the following subagents to this repo:
* `developer` - to implement features and fix issues
* `code-reviewer` - to review the code produced by the `developer` agent
* `technical-writer` - to update the docs


.claude/agents/code-reviewer.md
```diff
@@ -0,0 +1,73 @@
...
+tools: Glob, Grep, LS, Read, NotebookRead, WebFetch, TodoWrite, WebSearch, Bash
+model: sonnet
```
eakmanrq, line 5: This comment is for all agents - should we not include tools/model and let the main thread determine what is allowed? I don't think we are ready to be opinionated on a detail like that for users.

.claude/agents/developer.md
```diff
@@ -0,0 +1,110 @@
...
+color: red
+---
+
+You are an expert software engineer with deep expertise in Python, SQL, data engineering, and modern software development practices. You specialize in working with complex codebases like SQLMesh, understanding architectural patterns, and implementing robust, well-tested solutions.
+
+Your core responsibilities:
+
+# Project-Specific Expertise
+
+- Understand SQLMesh's core concepts: virtual environments, fingerprinting, snapshots, plans. You can find documentation at https://sqlmesh.readthedocs.io/
```
eakmanrq: Do you know if linking to the website is better for claude vs. linking it to the repo's document files? Wondering if you have tested this.

eakmanrq (approved):

.claude/agents/developer.md
izeigerman: Yeah, I think your approach makes more sense. Not sure how to test this

.claude/agents/code-reviewer.md
izeigerman: This header is a part of the Claude agent definition format. This is metadata I believe and is not part of the prompt. Just like `color`.

sungchun12:
The heuristic is enough for me. We'll let people try it out in the wild and iterate from there!

.claude/agents/code-reviewer.md
eakmanrq: This is in the docs:
```
You have two options for configuring tools:

Omit the tools field to inherit all tools from the main thread (default), including MCP tools
Specify individual tools as a comma-separated list for more granular control (can be edited manually or via /agents)
```

I read that as the tools provided here override what I have already approved in my main thread.

izeigerman: Ah, I see. The reason tools are outlined here is because some of the agents are read-only (eg. the code-reviewer one) and this is what captured here.

Issue #4767, opened 2025-06-19
Janitor fails to delete object with dependencies in Postgres


petrikoro:
When running the Janitor against a Postgres database, objects that have dependent objects (e.g. views depending on tables) are not dropped in the first place, resulting in a failure when attempting to drop the table with views on top.

<details>
  <summary>Traceback</summary>

    ```
    2025-06-19 15:41:06,011 - MainThread - sqlmesh.cli - ERROR - Unhandled exception (__init__.py:53)
    Traceback (most recent call last):
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/utils/concurrency.py", line 69, in _process_node
        self.fn(node)
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/utils/concurrency.py", line 172, in <lambda>
        lambda s_id: fn(snapshots_by_id[s_id]),
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 474, in <lambda>
        lambda s: self._cleanup_snapshot(
                ^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 1038, in _cleanup_snapshot
        evaluation_strategy.delete(
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/snapshot/evaluator.py", line 1535, in delete
        self.adapter.drop_table(name)
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/shared.py", line 334, in internal_wrapper
        return func(*list_args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 931, in drop_table
        self._drop_object(name=table_name, exists=exists)
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 959, in _drop_object
        self.execute(exp.Drop(this=exp.to_table(name), kind=kind, exists=exists, **drop_args))
    File "/usr/local/lib/python3.12/site-packages/sqlmesh/core/engine_adapter/base.py", line 2166, in execute
        self._execute(sql, **kwargs)
[...]

PR #5098, opened 2025-08-06
Fix: Include unexpired downstream views when cleaning up expired tables

### LINKED ISSUE #5083: Unexpected behaviour with FULL model kinds (DuckDB+Postgres)
SQLMesh is attempting to delete a physical layer table, which fails in Postgres due to the virtual layer view still pointing to it.  Error will look something like: `cannot drop table ... because other objects depend on it`

I expect SQLMesh to behave like:
- on new interval, create new physical table
- recreate view to point to new physical table
- (either now or later by janitor) delete old physical table

Right now it seems to be trying to delete the old/existing physical table first.

SQLMesh version 0.192.0
Model config looks like:

```
MODEL (
  kind FULL,
  name postgresdb.schema.table,
  gateway duckdb,
  dialect duckdb,
  cron '@daily',
  start '2025-07-28'
);
```

With a project config like:
```
config = Config(
    gateways={
        "athena": GatewayConfig(
            connection=AthenaConnectionConfig(XXXX),
            state_connection=PostgresConnectionConfig(XX),
            state_schema="X",
        ),
        "duckdb": GatewayConfig(
            connection=DuckDBConnectionConfig(
                catalogs={
                    "postgresdb": DuckDBAttachOptions(
                        type="postgres",
                        path=f"XXX"
                    ),
                },
                extensions=["iceberg", "aws", "postgres", "avro", "httpfs"],
            )
        ),
    },
    default_gateway="athena",
    disable_anonymized_analytics=True,
    gateway_managed_virtual_layer=True,
)
```

erindru:
Fixes #5083 #4767

Prior to this, it was possible for the janitor to drop a snapshot table while another snapshot view still depended on it.

The reason is not related to DAG order - it turns out the janitor always drops in DAG order.

Rather, the reason was that it's possible to get into a situation where a snapshot table has expired (based on `updated_ts + ttl_ms` being earlier than the current time) but a downstream view that depends on it had *not* expired, because it was created later.

A naive implementation might simply just call `DROP... CASCADE` to automatically drop downstream objects but the problem with this is that while the physical objects are gone there will still be references in SQLMesh state.

So this PR updates the code that detects expired objects to also consider unexpired views downstream of expired tables.

Open question: Is it ever valid to drop an expired upstream dependency and not cascade that to all downstream dependencies? Should we be dropping all downstream unexpired tables too?

### DIFF
--- a/sqlmesh/core/state_sync/db/snapshot.py
+++ b/sqlmesh/core/state_sync/db/snapshot.py
@@ -230,7 +230,8 @@ class SnapshotState:
             )
 
             if result := fetchone(
-                self.engine_adapter, exp.select("count(*)").from_(expired_query.subquery())
+                self.engine_adapter,
+                exp.select("count(*)").from_(expired_query.subquery().as_("expired_snapshots")),
             ):
                 expired_record_count = result[0]
 


sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -253,37 +287,39 @@ def _is_snapshot_used(snapshot: SharedVersionSnapshot) -> bool:
...
-        expired_snapshot_ids = set()
+        expired_sv_snapshot_ids = set()
         for versions_batch in version_batches:
-            snapshots = self._get_snapshots_with_same_version(versions_batch)
+            sv_snapshots = self._get_snapshots_with_same_version(versions_batch)
 
             snapshots_by_version = defaultdict(set)
             snapshots_by_dev_version = defaultdict(set)
-            for s in snapshots:
+            for s in sv_snapshots:
```
erindru, line 307: These variable renames were just to placate mypy, because `snapshot` is used above to refer to objects of type `Snapshot` which does not match the objects of type `SharedVersionSnapshot` used here

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -224,18 +225,51 @@ def _get_expired_snapshots(
...
+            expired_candidates = candidates
+        else:
+            # Fetch full snapshots because we need to build a dependency tree
+            full_candidates = self.get_snapshots(candidates.keys())
+
+            # Build DAG so we can check if any views that are not expired depend on a table snapshot that is expired
+            dag = snapshots_to_dag(full_candidates.values())
+
+            # remove any non-expired views that dont depend on expired tables
+            for snapshot_id in dag.reversed:
```
erindru: It occurs to me that a better way to do this might be to just look at `dag.roots` and then capture any downstream unexpired views, rather than checking every object and all upstreams

erindru: Actually, thinking about this more, since for any given unexpired view we only only need to check immediate parents so we don't need a DAG at all.

The operations still occur in the right order at drop time because a DAG is created to carry that out

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -224,18 +224,54 @@ def _get_expired_snapshots(
 
         if not ignore_ttl:
             expired_query = expired_query.where(
-                (exp.column("updated_ts") + exp.column("ttl_ms")) <= current_ts
+                ((exp.column("updated_ts") + exp.column("ttl_ms")) <= current_ts)
```
themisvaltinos: I’m wondering if it might make sense to break this into two steps, first querying only with the initial condition
`(exp.column("updated_ts") + exp.column("ttl_ms")) <= current_ts`
then only if that query returns results, we proceed with a second query for the views. the downside is that it adds an extra query in cases where there are expired tables.but I’m thinking about the scenario where there are no expired tables, where currently we still end up querying snapshots via `full_candidates = self.get_snapshots(candidates.keys())` for all views and going through the full logic, even though there won’t be any results in the end. that said im not sure since both have tradeoffs

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -2799,3 +2802,180 @@ def test_identifier_length_limit(ctx: TestContext):
         match=re.escape(match),
     ):
         adapter.create_table(long_table_name, {"col": exp.DataType.build("int")})
+
+
+def test_janitor_drops_downstream_unexpired_hard_dependencies(
```
themisvaltinos, line 2806: could you also add a test for a transitive dependency? For example:
**Table A** is expired ← **View B** (not expired) ← **View C** (not expired)
I believe this case was handled with the reversed dag before and that **View C** got picked up, but I’m not entirely certain without the dag construction now, so it’d be good to have a test to confirm

erindru: Yeah you're right, the drops weren't cascading through transitive dependencies. I've re implemented this to use a DAG again which simplified the implementation because it can be traversed in topological order and for any given node we can:

- check if its expired (gets put directly in the expired list)
- if it's a view, check if any of its parents are in the expired list. if they are, it should be expired too

I also updated the tests to test transitive dependencies too

sqlmesh/core/state_sync/db/snapshot.py
erindru: Yeah I understand the concern. I've broken it into the following steps (which only apply if `ignore_ttl` is not set):

- Query a count of expired snapshots
- If that count is >0, add in the OR clause to the query to fetch all view snapshots as well
- If that count is 0, use it to short-circuit and return

themisvaltinos (approved):
looks good to me, thanks for addressing comments and adding the extra tests

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -221,21 +222,65 @@ def _get_expired_snapshots(
...
+            for snapshot_id in dag:
+                snapshot = full_candidates.get(snapshot_id, None)
+
+                if not snapshot:
+                    continue
+
+                if snapshot.expiration_ts <= current_ts:
+                    # All expired snapshots should be included regardless
+                    expired_candidates[snapshot.snapshot_id] = snapshot.name_version
+                elif snapshot.model_kind_name == ModelKindName.VIEW:
```
izeigerman, line 277: Can this be filtered by a database rather than the application layer?

erindru: How do you mean? Line 242 adjusts the query to the database to include the views:

```
            if expired_record_count > 0:
                expired_query = t.cast(
                    exp.Select, expired_query.or_(exp.column("kind_name").eq(ModelKindName.VIEW))
                )
```

The DAG can only be built in the application layer, right? Since we have no way of knowing at the database level what views are affected, don't we have to select all views and filter them at the application layer?

We can't filter to expired views in the DB query because that's what caused this problem to begin with.

Issue #4916, opened 2025-07-07
Modifying model metadata often causes plan to fail if they haven't been run recently


codykonior:
1. Create an INCREMENTAL_BY_UNIQUE_KEY model.
2. Run sqlmesh plan.
3. Wait a few days.
4. Modify the model metadata in some way, for example add:


```
MODEL (
...
physical_properties ('mssql_merge_exists' = True)
)
```

5. Run sqlmesh plan. It will fail:

```
---
                                                                                                                                                                                                   
+++                                                                                                                                                                                                
                                                                                                                                                                                                   
@@ -1,6 +1,7 @@
                                                                                                                                                                                                   
 MODEL (                                                                                                                                                                                           
   name [xxx].[xxx],                                                                                                                                                             
-  dialect [tsql,normalization_strategy=case_sensitive]
+  dialect [tsql,normalization_strategy=case_sensitive],
+  physical_properties ('mssql_merge_exists' = 1)
 )                                                                                                                                                                                                 
 SELECT
   w.[ClientPrefix],                                                                                                                                                                               

Directly Modified: xxx.xxx (Non-breaking)
Error: Start date / time (2025-07-06 00:00:00+00:00) can't be greater than end date / time (1751299200000)
```

I get this a lot when making changes to my templates. It forces me to run `sqlmesh run` to reprocess everything I have, before I can do a plan again. But as you can imagine this kind of model doesn't use timestamps/intervals and so I don't think this kind of logic check should apply.

izeigerman:
Can you please clarify whether you setup `start` attribute in the project config.

codykonior:
> Can you please clarify whether you setup `start` attribute in the project config.

I’m not. Should I be?

VaggelisD:
Hey @codykonior,

Yes, ideally `start` should be defined at either at a project config or at a `MODEL` level; With no `start` defined, the default value is set to "yesterday" while the end time will be initialized as end time from the model's last evaluated intervals, yielding the error you're observing.

Could you let us know if that fixes your issue?

codykonior:
@themisvaltinos Thanks! Sorry I forgot to come back. After adding a start date a few weeks ago on your advice I didn’t see the problem recur. 

I see that when sqlmesh creates its configs it has one in there. I probably removed it because I’m using unique key models everywhere and so there’s no intervals etc relevant to my data. Now I just keep a dummy date in there.  

Glad others won’t encounter this though!

PR #5100, opened 2025-08-06
Chore: Validate start and end dates during plan build time

themisvaltinos:
Update to validate start is less than end date/time early during plan build time in the event that the user hasn't provided a start date in the config or in all models explicitly, fixes: #4916

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1621,7 +1621,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                                 or to_timestamp(snapshot.node.start) > end_ts
                             ):
                                 raise SQLMeshError(
-                                    f"Start date / time ({to_datetime(start_ts)}) can't be greater than end date / time ({to_datetime(end_ts)}).\n"
+                                    f"Model '{model_name}': Start date / time ({to_datetime(start_ts)}) can't be greater than end date / time ({to_datetime(end_ts)}).\n"
                                     f"Set the `start` attribute in your project config model defaults to avoid this issue."
                                 )
 
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1603,28 +1603,6 @@ class GenericContext(BaseContext, t.Generic[C]):
                 execution_time or now(),
             )
 
-            # Validate that the start is not greater than the end date / time
-            effective_start = start or default_start
-            effective_end = end or default_end
-            if effective_start is not None and effective_end is not None:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/context.py
```diff
@@ -1603,6 +1603,28 @@ def plan_builder(
...
+                        | set(modified_model_names)
+                        | set(max_interval_end_per_model)
+                    ):
+                        if snapshot := snapshots.get(model_name):
+                            if (
+                                snapshot.node.start is None
+                                or to_timestamp(snapshot.node.start) > end_ts
+                            ):
+                                raise SQLMeshError(
+                                    f"Start date / time ({to_datetime(start_ts)}) can't be greater than end date / time ({to_datetime(end_ts)}).\n"
```
erindru: Do we want to indicate which model triggered this error as well?

erindru (approved):

sqlmesh/core/context.py
themisvaltinos: yes good point added it

sqlmesh/core/context.py
```diff
@@ -1603,6 +1603,28 @@ def plan_builder(
                 execution_time or now(),
             )
 
+            # Validate that the start is not greater than the end date / time
+            effective_start = start or default_start
+            effective_end = end or default_end
+            if effective_start is not None and effective_end is not None:
```
izeigerman: Let's have a separate function for this block

sqlmesh/core/context.py
```diff
@@ -1603,6 +1603,28 @@ def plan_builder(
...
+                        set(backfill_models or {})
+                        | set(modified_model_names)
+                        | set(max_interval_end_per_model)
+                    ):
+                        if snapshot := snapshots.get(model_name):
+                            if (
+                                snapshot.node.start is None
+                                or to_timestamp(snapshot.node.start) > end_ts
+                            ):
+                                raise SQLMeshError(
```
izeigerman: I don't think this is the right place to perform validation. With `PlanBuilder` users can set the `start` and `end` dynamically which means that currently they can pass this initial check and then set an invalid value later. 

This should be moved into the `PlanBuilder` itself and the check needs to be performed every time `build` is called just like with other `_ensure_*` methods in the builder. Also, this should be `PlanError`.

themisvaltinos: refactored and moved it to the `_ensure_valid_date_range` method which had similar logic and gets called in build

PR #5106, opened 2025-08-07
Fix!: Add model default audits in the model preserving their args

themisvaltinos:
This update ensures that audit arguments from default_audits are no longer dropped when model audits are present.
Previously, if a model defined its own audits, only those were retained and the arguments from default_audits were silently dropped (since we used the audit_definitions with empty args). This adds both in the model's audits, preserving them along with their arguments.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2538,7 +2538,22 @@ def _create_model(
         for jinja_macro in jinja_macros.root_macros.values():
             used_variables.update(extract_macro_references_and_variables(jinja_macro.definition)[1])
 
-    default_audits = defaults.get("audits", None) if kwargs.get("audits") else None
+    # Merge model-specific audits with default audits
+    model_audits = kwargs.pop("audits", [])
+    default_audits = defaults.pop("audits", [])
+
+    if isinstance(model_audits, (exp.Tuple, exp.Array)):
+        model_audits_list = [d.extract_func_call(i) for i in model_audits.expressions]
+    elif isinstance(model_audits, exp.Paren):
+        model_audits_list = [d.extract_func_call(model_audits.this)]
+    elif isinstance(model_audits, exp.Expression):
+        model_audits_list = [d.extract_func_call(model_audits)]
+    elif isinstance(model_audits, list):
+        model_audits_list = model_audits
+    else:
+        model_audits_list = []
+    merged_audits = default_audits + model_audits_list
+    kwargs["audits"] = merged_audits
 
     model = klass(
         name=name,
@@ -2556,9 +2571,6 @@ def _create_model(
         **(audit_definitions or {}),
         **(inline_audits or {}),
     }
-
-    if default_audits:
-        model = model.copy(update={"audits": default_audits + model.audits})
 
     used_audits: t.Set[str] = {audit_name for audit_name, _ in model.audits}
 
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2539,21 +2539,8 @@ def _create_model(
             used_variables.update(extract_macro_references_and_variables(jinja_macro.definition)[1])
 
     # Merge model-specific audits with default audits
-    model_audits = kwargs.pop("audits", [])
-    default_audits = defaults.pop("audits", [])
-
-    if isinstance(model_audits, (exp.Tuple, exp.Array)):
-        model_audits_list = [d.extract_func_call(i) for i in model_audits.expressions]
-    elif isinstance(model_audits, exp.Paren):
-        model_audits_list = [d.extract_func_call(model_audits.this)]
-    elif isinstance(model_audits, exp.Expression):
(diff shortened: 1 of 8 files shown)


sqlmesh/migrations/v0088_include_default_audits_in_model.py
```diff
@@ -0,0 +1,5 @@
+"""Include the model defaults audits along with their args in the model."""
```
izeigerman: Why does this require migration?

sqlmesh/core/model/definition.py
```diff
@@ -2558,12 +2557,10 @@ def _create_model(
...
     }
 
-    # TODO: default_audits needs to be merged with model.audits; the former's arguments
-    # are silently dropped today because we add them in audit_definitions. We also need
-    # to check for duplicates when we implement this merging logic.
-    used_audits: t.Set[str] = set()
-    used_audits.update(audit_name for audit_name, _ in default_audits or [])
-    used_audits.update(audit_name for audit_name, _ in model.audits)
+    if default_audits:
+        model = model.copy(update={"audits": default_audits + model.audits})
```
izeigerman: Why can't we merge upstream and pass them into `klass` directly? Why do copy after the fact?

themisvaltinos: that was the alternative approach but because `default_audits` is already a list of `FunctionCall`s while the model audits can be an `exp.Paren` expression, a `exp.Tuple`, an `exp.Array` it becomes a bit tricky. so I tried initially merging them first, by changing the logic to not extract the model_default audits as function calls in the validator so now we'd have a list of expressions, but then it became a bit complicated because of these possibilities that the `kwargs` audits can be at this point. so it requires a more complex logic but I can go that route instead if you prefer and maybe merge them all into one tuple where every audit is an expression of this tuple

sqlmesh/migrations/v0088_include_default_audits_in_model.py
themisvaltinos: because if a project has an audit in the model defaults and also have a model with an audit, this `model.audits` would contain only the latter. so the migration would add the default audit in the model, so the next time the user runs plan this is included as well

sqlmesh/core/model/definition.py
izeigerman: > but then it became a bit complicated because of these possibilities that the kwargs audits can be at this point

not sure why exactly does it get more complicated

sqlmesh/migrations/v0088_include_default_audits_in_model.py
izeigerman: i'm sorry, I'm confused. Migrations don't affect what will be loaded the next time the user runs plan. Do fingerprints of snapshots that are already in state change in any way? If no, why do we need a migration?

themisvaltinos: yes the fingerprint changes since the model `audits` might be modified if default audits have been defined in the config(previously we only had these in the audit_definitions of the model) so the hash will change

izeigerman: the hash will change locally, I get that. But for the hash to change your new logic should apply which only happens at load time, right? Why will the hash change for existing snapshots that are already in the state?

sqlmesh/core/model/definition.py
themisvaltinos: because there would be a need to either push this logic back for audits:
https://github.com/TobikoData/sqlmesh/blob/49b5574732cb83fe2b0525e6ccbedefc85e48549/sqlmesh/core/model/meta.py#L97-L103
so that the default audits (which are a list) can be merged before validation or pass each audit from the list into a new expression's expression (for example tuple) but that still would need some detection of what the audits are before validation

themisvaltinos: so the part that makes it tricky is that audits from `defaults` in this place are already extracted so: `t.List[FunctionCall]` while audits from `kwargs` are `exp.Tuple | exp.Paren | exp.Array | list`

sqlmesh/migrations/v0088_include_default_audits_in_model.py
themisvaltinos: oh right I see, now I understand what you're saying

sqlmesh/core/model/definition.py
```diff
@@ -2541,6 +2538,23 @@ def _create_model(
         for jinja_macro in jinja_macros.root_macros.values():
             used_variables.update(extract_macro_references_and_variables(jinja_macro.definition)[1])
 
+    # Merge model-specific audits with default audits
+    model_audits = kwargs.pop("audits", [])
+    default_audits = defaults.pop("audits", [])
+
+    if isinstance(model_audits, (exp.Tuple, exp.Array)):
```
izeigerman: This block looks like something that could be an utility function and doesn't have to be a part of `_create_model`

izeigerman (approved):
Housekeeping comment, LGTM otherwise

PR #5110, opened 2025-08-07
Feat!: Decouple forward-only from change categorization

izeigerman:
This PR decouples the `forward-only` behavior of a snapshot from its change category. Thus, we end up with two independent dimensions:

- **Forward-only:** determines whether the change needs to be applied "going forward" to an existing production shared-version physical table or rebuilt from scratch
- **Change category:** determines what needs to be *rebuilt* if the change is **not** forward-only or *previewed* in development if the change is forward-only

Doing this allows us to apply forward-only changes without losing the benefits of impact analysis and change categorization, particularly for non-breaking changes. So even if either the model or the plan is forward-only, SQLMesh will still categorize changes as `BREAKING` and `NON_BREAKING` and will select what needs to be previewed in dev accordingly. For example, previews for `INDIRECT_NON_BREAKING` changes can be safely skipped.

This approach is also conceptually more logical, as “forward-only” refers to how a change is applied in production, rather than to the nature of the change itself, unlike other categories.

Additionally, users are never exposed to the `FORWARD_ONLY` category in the plan output which was quite confusing. Forward-only now only affects whether changes built in dev can be reused in production, as well as whether a full rebuild of a model is required when deploying to production.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -527,6 +527,9 @@ class PlanBuilder:
 
     def _check_destructive_changes(self, directly_modified: t.Set[SnapshotId]) -> None:
         for s_id in sorted(directly_modified):
+            if s_id.name not in self._context_diff.modified_snapshots:
+                continue
+
             snapshot = self._context_diff.snapshots[s_id]
             # should we raise/warn if this snapshot has/inherits a destructive change?
             should_raise_or_warn = (


sqlmesh/core/plan/builder.py
```diff
@@ -395,6 +393,9 @@ def _build_restatements(
         for s_id in dag:
             snapshot = self._context_diff.snapshots[s_id]
 
+            if is_preview and snapshot.is_no_preview:
```
tobymao, line 396: why couldn't this be if snapshot.is_no_preview?

sqlmesh/core/plan/builder.py
```diff
@@ -583,38 +584,35 @@ def _categorize_snapshots(
...
+        indirectly_modified: SnapshotMapping,
     ) -> None:
         s_id = snapshot.snapshot_id
 
         if self._context_diff.directly_modified(s_id.name):
-            if self._auto_categorization_enabled:
+            new, old = self._context_diff.modified_snapshots[s_id.name]
+            if _is_breaking_kind_change(old, new):
+                snapshot.categorize_as(SnapshotChangeCategory.BREAKING, False)
+            elif self._auto_categorization_enabled:
```
eakmanrq: Why can you check if `_is_breaking_kind_change` and categorize even if auto categorization is not enabled?

sqlmesh/core/plan/builder.py
izeigerman: because if it's a legit restatement and not a preview we still want to include the model

sqlmesh/core/plan/builder.py
izeigerman: because the kind change is unconditionally breaking? Is this too aggressive?

eakmanrq: If someone didn't enable auto-categorization then they want to be in control of the change category and not defer to us even if we think it should be breaking. Or am I not understanding this part? I might not be understanding the full nuance you intend with "unconditionally breaking".

izeigerman: Yeah, I agree. I think we should force this to be non-forward-only, but the actual category is up to a user

izeigerman: Fixed

eakmanrq (approved):

PR #5111, opened 2025-08-08
feat!: add support for `on_additive_change`

eakmanrq:
Breaking explanation: This change is breaking if you currently use the dbt adapter and have models using `on_schema_change` where we used to not correctly mimic the behavior in the SQLMesh runtime. This is non-breaking for SQLMesh native users. 

In main, SQLMesh currently just detects destructive changes and allows users 3 different behavior options (ERROR, WARN, ALLOW). The problem is that dbt has schema change options that, in SQLMesh terms, differentiates destructive and additive changes. For example `append_new_columns` allows additive changes but ignores destructive changes. Furthermore they have an `ignore` option which means that no schema changes are performed on tables. 

Therefore two fundamental changes are made in this PR:
* `on_additive_change` incremental property is added that behaves just like `on_destructive_change` except applies to additive changes
* `ignore` is added as an option to `on_destructive_change`. This should never really be used outside of some workaround edge-cases and documentation properly reflects this. dbt just says it will fail next time you run if you ignore (which is the default behavior)...

The dbt adapter is updated to use these changes. See table below for a mapping of behavior from dbt -> SQLMesh (the docs also contain this):

| `on_schema_change` | SQLMesh `on_destructive_change` | SQLMesh `on_additive_change` |
|--------------------|---------------------------------|------------------------------|
| ignore             | ignore                          | ignore                       |
| fail               | error                           | error                        |
| append_new_columns | ignore                           | allow                       |
| sync_all_columns   | allow                           | allow                        |

Two follow up PRs are required to reach `on_schema_change` parity:
* `append_new_columns` ignores alters even if the alters are additive. This PR would consider that additive and allow the change through. So need a way for dbt adapter to configure schema differ to consider alters destructive so it can be ignored. 
* `on_schema_change` only tracks top-level column changes while, on some engines, SQLMesh tracks nested column changes.  Therefore there needs to be a way to let dbt adapter to configure SQLMesh so it doesn't do nested changes.

dbt documentation: https://docs.getdbt.com/docs/build/incremental-models#what-if-the-columns-of-my-incremental-model-change

TODO: 
* Manual testing
* Run docs locally and validate it renders properly


sqlmesh/core/model/meta.py
```diff
@@ -418,6 +420,31 @@ def _root_validator(self) -> Self:
...
+        if isinstance(kind, _Incremental):
+            destructive = kind.on_destructive_change
+            additive = kind.on_additive_change
+
+            # Unusual combination: ALLOW destructive but ERROR on additive
+            # (very restrictive on additions, permissive on removals)
+            if destructive.is_allow and additive.is_error:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
```
izeigerman, line 433: Not sure what is the point of this warning. If it was intentional the warning would just be a noise that can't be dismissed

sqlmesh/core/model/meta.py
```diff
@@ -418,6 +420,31 @@ def _root_validator(self) -> Self:
...
+                    "This allows destructive changes (like dropping columns) but blocks additive changes (like adding columns)."
+                )
+
+            # Unusual combination: ERROR on destructive but IGNORE on additive
+            # (very permissive on additions, strict on removals - this is actually reasonable)
+            elif destructive.is_error and additive.is_ignore:
+                from sqlmesh.core.console import get_console
+
+                get_console().log_warning(
+                    f"Model {self.name} has on_destructive_change=ERROR and on_additive_change=IGNORE. "
```
izeigerman, line 444: Same here

sqlmesh/core/model/meta.py
eakmanrq: Ok I'm good with removing it. Just my way of trying to address the main negative with the approach of having two ways of controlling changes.

sqlmesh/core/model/kind.py
```diff
@@ -1004,6 +1053,15 @@ def create_model_kind(v: t.Any, dialect: str, defaults: t.Dict[str, t.Any]) -> M
         ):
             props["on_destructive_change"] = defaults.get("on_destructive_change")
 
+        # only pass the on_additive_change user default to models inheriting from _Incremental
+        # that don't explicitly set it in the model definition
+        if (
```
izeigerman, line 1058: Any reason we need to duplicate this entire block instead of doing something like:
```
if issubclass(kind_type, _Incremental):
    for on_change_property in ("on_additive_change", "on_desctructive_change"):
        ...
```

sqlmesh/core/model/meta.py
izeigerman: Well, we should either forbid this and fail or allow it and say nothing.

sqlmesh/core/plan/builder.py
```diff
@@ -535,37 +590,62 @@ def _check_destructive_changes(self, directly_modified: t.Set[SnapshotId]) -> No
...
+
+            if not should_raise_or_warn or not snapshot.is_model:
+                continue
+
+            # Skip if model doesn't have additive change handling (non-incremental models)
+            # Only incremental model kinds have on_additive_change attribute
+            if not isinstance(snapshot.model.kind, _Incremental):
+                continue
+
+            schema_diff = self._get_schema_changes_for_snapshot(snapshot)
```
izeigerman, line 630: Why compute this twice?

sqlmesh/core/plan/builder.py
```diff
@@ -524,6 +537,48 @@ def _adjust_new_snapshot_intervals(self) -> None:
...
+            return None
+
+        # Generate all schema changes first
+        alter_expressions = get_schema_differ(snapshot.model.dialect).compare_columns(
+            new.name,
+            old_columns_to_types,
+            new_columns_to_types,
+        )
+
+        # Apply IGNORE filtering BEFORE returning expressions
```
izeigerman, line 569: This feels redundant. I think if ignore is set we shouldn't even iterate over the list of alterations. I don't think this method should be concerned with this.

sqlmesh/core/schema_diff.py
```diff
@@ -256,22 +259,27 @@ def expression(
         self, table_name: t.Union[str, exp.Table], array_element_selector: str
     ) -> exp.Alter:
         if self.is_alter_type:
+            alter_column = exp.AlterColumn(
+                this=self.column(array_element_selector),
+                dtype=self.column_type,
+            )
+            # Add current type metadata for additive classification
+            if self.current_type:
+                alter_column.meta["current_type"] = self.current_type
```
izeigerman, line 268: Do we use this anywhere?

sqlmesh/core/schema_diff.py
```diff
@@ -256,22 +259,27 @@ def expression(
...
             )
         if self.is_add:
             alter_table = exp.Alter(this=exp.to_table(table_name), kind="TABLE")
             column = self.column_def(array_element_selector)
             alter_table.set("actions", [column])
             if self.add_position:
                 column.set("position", self.add_position.column_position_node)
+            # Set metadata if this ADD operation supports destructive changes
+            if self.is_part_of_destructive_change:
+                column.meta["sqlmesh_destructive"] = True
```
izeigerman, line 282: I don't understand why does it matter. There's will be a corresponding drop alteration which will trigger the check. Why complicate this?

izeigerman: Can we just check whether the same column shows up for both drop and add alteration in the generated list? This will make it more explicit.

sqlmesh/core/plan/builder.py
izeigerman: I think I understand this now. Disregard

sqlmesh/core/schema_diff.py
eakmanrq: Conceptually the reason why something like this is needed is because we can't assume ADD commands are additive - if they are ADD commands that are only needed due to a DROP command then we don't want that add either.

Doing it based on column name also works and I agree is cleaner since it wouldn't require creating metadata like this.

izeigerman:
I believe we still need an empty migration script to recompute the metadata hash

eakmanrq:
Deprecated by: https://github.com/TobikoData/sqlmesh/pull/5193

All the feedback from this PR was address in the new PR.

PR #5112, opened 2025-08-08
Feat: Re-introduce merge for updating auto restatements

vchan:
#5016 was reverted with #5108 because we didn't properly handle fetching server version with pg8000. This PR re-introduces merge for updating auto restatements but with added support for pg8000.

### DIFF
--- a/sqlmesh/core/engine_adapter/postgres.py
+++ b/sqlmesh/core/engine_adapter/postgres.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import re
 import typing as t
-from functools import partial
+from functools import cached_property, partial
 from sqlglot import exp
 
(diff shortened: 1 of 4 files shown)


sqlmesh/core/engine_adapter/postgres.py
```diff
@@ -125,3 +123,23 @@ def merge(
             when_matched=when_matched,
             merge_filter=merge_filter,
         )
+
+    def get_server_version(self) -> t.Tuple[int, int]:
```
erindru: this seems fairly hacky vs just running something like `select current_setting('server_version_num')` when the EngineAdapter is constructed and storing the result (or even just making this method run the query and use `@cached_property` to cache the result)

izeigerman (approved):

PR #5115, opened 2025-08-08
Fix!: sqlmesh.dbt.adapter.RuntimeAdapter.get_columns_in_relation()

### LINKED ISSUE #1874: dbt get_column_in_relation should return proper subclass
get_column_in_relation currently returns a base column class 

https://github.com/TobikoData/sqlmesh/pull/1873/files

we had to revert the change to the engine specific subclasses because each subclass has a different init signature. from_description crashes on bigquery because bigquery is not expecting the same initializer.

newtonapple:
returns columns in the proper Column subclass type.

fixes #1874

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -272,6 +272,9 @@ filterwarnings = [
 ]
 retry_delay = 10
 
+[tool.ruff]
+line-length = 100
+
 [tool.ruff.lint]
 select = [
     "F401",
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -266,6 +266,13 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
                 raise
             logger.warning("Failed to create schema '%s': %s", schema_name, e)
 
+    def get_bq_schema(self, table_name: TableName) -> t.List[bigquery.SchemaField]:
+        table = exp.to_table(table_name)
+        if len(table.parts) == 3 and "." in table.name:
+            self.execute(exp.select("*").from_(table).limit(0))
+            return self._query_job._query_results.schema
+        return self._get_table(table).schema
+
     def columns(
         self, table_name: TableName, include_pseudo_columns: bool = False
     ) -> t.Dict[str, exp.DataType]:
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -291,11 +291,31 @@ class RuntimeAdapter(BaseAdapter):
         return relations
 
     def get_columns_in_relation(self, relation: BaseRelation) -> t.List[Column]:
-        from dbt.adapters.base.column import Column
-
         mapped_table = self._map_table_name(self._normalize(self._relation_to_table(relation)))
+
+        if self.project_dialect == "bigquery":
+            # dbt.adapters.bigquery.column.BigQueryColumn has a different constructor signature
+            # We need to use BigQueryColumn.create_from_field() to create the column instead
+            if (
+                hasattr(self.column_type, "create_from_field")
+                and callable(getattr(self.column_type, "create_from_field"))
+                and hasattr(self.engine_adapter, "get_bq_schema")
+                and callable(getattr(self.engine_adapter, "get_bq_schema"))
+            ):
+                return [
+                    self.column_type.create_from_field(field)  # type: ignore
+                    for field in self.engine_adapter.get_bq_schema(mapped_table)  # type: ignore
+                ]
+            from dbt.adapters.base.column import Column
+
+            return [
+                Column.from_description(
+                    name=name, raw_data_type=dtype.sql(dialect=self.project_dialect)
+                )
+                for name, dtype in self.engine_adapter.columns(table_name=mapped_table).items()
+            ]
         return [
-            Column.from_description(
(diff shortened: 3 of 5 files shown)


sqlmesh/dbt/adapter.py
```diff
@@ -291,14 +291,36 @@ def list_relations_without_caching(self, schema_relation: BaseRelation) -> t.Lis
...
 
     def get_columns_in_relation(self, relation: BaseRelation) -> t.List[Column]:
-        from dbt.adapters.base.column import Column
-
         mapped_table = self._map_table_name(self._normalize(self._relation_to_table(relation)))
+
+        columns = self.engine_adapter.columns(table_name=mapped_table).items()
+
+        if self.project_dialect == "bigquery":
+            # dbt.adapters.bigquery.column.BigQueryColumn has a different constructor signature
```
newtonapple, line 297: ## Reference:
- [dbt.adapters.bigquery.column.BigQueryColumn](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-bigquery/src/dbt/adapters/bigquery/column.py#L24-L29)
- [BigQueryColumn.create_from_field](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-bigquery/src/dbt/adapters/bigquery/column.py#L44-L50)

pyproject.toml
```diff
@@ -272,6 +272,9 @@ filterwarnings = [
 ]
 retry_delay = 10
 
+[tool.ruff]
+line-length = 100
```
izeigerman, line 276: Why has this config changed?

newtonapple: Some auto-formatters (like my vim's auto-formatter) use `pyproject.toml`'s config for formatting.  This just made it the same as our [pre-commit hook](https://github.com/TobikoData/sqlmesh/blob/main/.pre-commit-config.yaml#L13).

sqlmesh/dbt/adapter.py
izeigerman: I don't think we should be querying the database just to construct the `Column` class. If you look at the reference you provided: https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-bigquery/src/dbt/adapters/bigquery/column.py#L128 we can construct `SchemaField` manually without making any queries.

Additionally, I think we can totally use the `BigQueryColumn` constructor directly and only provide `fields` when it's a struct column. 

If you want to return `SchemaField` from the BQ client directly, that's fine, but you should replace the `.columns` call with the variant that returns a list of SchemaField instances, rather than performing two calls. Perhaps this would be a better approach.

izeigerman: Actually, I don't think the latter approach would work when the project is in BQ dialect but is running against a different data warehouse.

newtonapple: > I don't think we should be querying the database just to construct the Column class.

But `self.engine_adapter.columns()` is already [querying the db](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/bigquery.py#L317-L324) to fetch the column now.  `get_bq_schema()` is basically doing the same thing but without [parsing the schema](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/bigquery.py#L322) into sqlgot `DataType`.

newtonapple: > Additionally, I think we can totally use the BigQueryColumn constructor directly and only provide fields when it's a struct column.

I didn't reference this, but this is exactly how dbt's implementation is using `create_from_field()` to construct the column:
https://github.com/dbt-labs/dbt-adapters/blob/3325afe0b2cd9cede991736f872d7b6739a47038/dbt-bigquery/src/dbt/adapters/bigquery/impl.py#L443-L454

izeigerman: Yes, you’re right. I didn’t read this correctly, sorry

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -266,6 +266,13 @@ def create_schema(
                 raise
             logger.warning("Failed to create schema '%s': %s", schema_name, e)
 
+    def get_bq_schema(self, table_name: TableName) -> t.List[bigquery.SchemaField]:
+        table = exp.to_table(table_name)
+        if len(table.parts) == 3 and "." in table.name:
+            self.execute(exp.select("*").from_(table).limit(0))
```
izeigerman, line 272: Why are we doing this here?

newtonapple: This logic was ported from `BigQueryEngineAdapter.columns()`: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/bigquery.py#L318-L325.

izeigerman:
Thanks for clarifying. This looks good. How do we test this?

newtonapple:
> How do we test this?



> Thanks for clarifying. This looks good. How do we test this?



> > Thanks for clarifying. This looks good. How do we test this?
> 
> I added an [integration test](https://github.com/TobikoData/sqlmesh/pull/5115/commits/dc724205eb9284ff47d159235feee64a8bfc120d) for `get_bq_schema()`. But, I'm not sure where is the best place to test the dbt adapter. Should I create a test for `RuntimeAdapter.get_columns_in_relation()` in the bigquey adapter integration?

Actually, I think I can mock it out.

izeigerman (approved):

PR #5117, opened 2025-08-09
feat!: add ignore destructive support

eakmanrq:
BREAKING: `columns_to_types` has been renamed to `target_columns_to_types` for engine adapter methods. 

This PR adds support for "ignoring" a destructive change using the `on_destructive_change` property. As documentation says, this property should not be used in most cases and is really just being added for dbt support. It could be useful though for an "escape-hatch" situation.

The concept of ignoring a destructive change is pushed down to the SchemaDiffer class level because the SchemaDiffer needs to be aware of this property is being used due to positional changes. If the engine supports positional changes and the SchemaDiffer wasn't aware then it could generate alter commands that involve changing columns assuming a column order that wouldn't really happen.

Prior to this PR, the engine adapter class assumed that the Query or DataFrame had the same columns as `columns_to_types`. `columns_to_types` has been renamed to `target_columns_to_types` and now represents the target table columns and can now drift out of sync of the "source" query or DataFrame. To account for this, engine adapter methods which take a `QueryOrDF` now also take a `source_columns` which tells the adapter what columns are in that source. If needed, it will pad out the columns in the source to match the target. 

Note: I am intentionally not changing dbt adapter implementation since I plan on updating it once `on_additive_change` support is added in follow up PR. That will isolate the breaking changes related to `on_schema_change` support to a single PR and it is just breaking for dbt adapter.

### DIFF
--- a/docs/guides/incremental_time.md
+++ b/docs/guides/incremental_time.md
@@ -171,7 +171,12 @@ The check is performed at plan time based on the model definition. SQLMesh may n
 
 A model's `on_destructive_change` [configuration setting](../reference/model_configuration.md#incremental-models) determines what happens when SQLMesh detects a destructive change.
 
-By default, SQLMesh will error so no data is lost. You can set `on_destructive_change` to `warn` or `allow` in the model's `MODEL` block to allow destructive changes.
+By default, SQLMesh will error so no data is lost. You can set `on_destructive_change` to `warn` or `allow` in the model's `MODEL` block to allow destructive changes. 
+`ignore` can be used to not perform the schema change and allow the table's definition to diverge from the model definition.
+
+!!! warning "Ignore is Dangerous"
+
+    `ignore` is dangerous since it can result in error or data loss. It likely should never be used but could be useful as an "escape-hatch" or a way to workaround unexpected behavior.
 
 This example configures a model to silently `allow` destructive changes:
 
--- a/sqlmesh/core/engine_adapter/_typing.py
+++ b/sqlmesh/core/engine_adapter/_typing.py
@@ -13,7 +13,7 @@ if t.TYPE_CHECKING:
 
     snowpark = optional_import("snowflake.snowpark")
 
-    Query = t.Union[exp.Query, exp.DerivedTable]
+    Query = exp.Query
     PySparkSession = t.Union[pyspark.sql.SparkSession, pyspark.sql.connect.dataframe.SparkSession]
     PySparkDataFrame = t.Union[pyspark.sql.DataFrame, pyspark.sql.connect.dataframe.DataFrame]
 
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -206,21 +206,23 @@ class EngineAdapter:
     @classmethod
     def _casted_columns(
         cls,
-        columns_to_types: t.Dict[str, exp.DataType],
+        target_columns_to_types: t.Dict[str, exp.DataType],
         source_columns: t.Optional[t.List[str]] = None,
     ) -> t.List[exp.Alias]:
-        source_columns = source_columns or list(columns_to_types)
+        source_columns_lookup = set(source_columns or target_columns_to_types)
         return [
             exp.alias_(
                 exp.cast(
-                    exp.column(column, quoted=True) if column in source_columns else exp.Null(),
+                    exp.column(column, quoted=True)
+                    if column in source_columns_lookup
+                    else exp.Null(),
                     to=kind,
                 ),
                 column,
                 copy=False,
                 quoted=True,
             )
-            for column, kind in columns_to_types.items()
+            for column, kind in target_columns_to_types.items()
         ]
 
     @property
@@ -241,7 +243,7 @@ class EngineAdapter:
     def _get_source_queries(
         self,
         query_or_df: QueryOrDF,
-        columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
+        target_columns_to_types: t.Optional[t.Dict[str, exp.DataType]],
         target_table: TableName,
         *,
         batch_size: t.Optional[int] = None,
@@ -253,16 +255,17 @@ class EngineAdapter:
         if isinstance(query_or_df, exp.Query):
             query_factory = lambda: query_or_df
             if source_columns:
-                if not columns_to_types:
+                source_columns_lookup = set(source_columns)
+                if not target_columns_to_types:
                     raise SQLMeshError("columns_to_types must be set if source_columns is set")
-                if not set(columns_to_types).issubset(set(source_columns)):
+                if not set(target_columns_to_types).issubset(source_columns_lookup):
                     select_columns = [
                         exp.column(c, quoted=True)
-                        if c in source_columns
-                        else exp.cast(exp.Null(), columns_to_types[c], copy=False).as_(
+                        if c in source_columns_lookup
+                        else exp.cast(exp.Null(), target_columns_to_types[c], copy=False).as_(
                             c, copy=False, quoted=True
                         )
-                        for c in columns_to_types
+                        for c in target_columns_to_types
                     ]
                     query_factory = (
                         lambda: exp.Select()
@@ -271,7 +274,7 @@ class EngineAdapter:
                     )
             return [SourceQuery(query_factory=query_factory)]  # type: ignore
 
-        if not columns_to_types:
+        if not target_columns_to_types:
             raise SQLMeshError(
                 "It is expected that if a DataFrame is passed in then columns_to_types is set"
             )
@@ -285,7 +288,7 @@ class EngineAdapter:
 
         return self._df_to_source_queries(
             query_or_df,
-            columns_to_types,
+            target_columns_to_types,
             batch_size,
             target_table=target_table,
             source_columns=source_columns,
@@ -294,7 +297,7 @@ class EngineAdapter:
     def _df_to_source_queries(
         self,
         df: DF,
-        columns_to_types: t.Dict[str, exp.DataType],
+        target_columns_to_types: t.Dict[str, exp.DataType],
         batch_size: int,
         target_table: TableName,
         source_columns: t.Optional[t.List[str]] = None,
@@ -307,7 +310,7 @@ class EngineAdapter:
 
         # we need to ensure that the order of the columns in columns_to_types columns matches the order of the values
         # they can differ if a user specifies columns() on a python model in a different order than what's in the DataFrame's emitted by that model
-        df = df[list(source_columns or columns_to_types)]
+        df = df[list(source_columns or target_columns_to_types)]
         values = list(df.itertuples(index=False, name=None))
 
         return [
@@ -315,7 +318,7 @@ class EngineAdapter:
                 query_factory=partial(
                     self._values_to_sql,
                     values=values,  # type: ignore
-                    columns_to_types=columns_to_types,
+                    target_columns_to_types=target_columns_to_types,
                     batch_start=i,
                     batch_end=min(i + batch_size, num_rows),
                     source_columns=source_columns,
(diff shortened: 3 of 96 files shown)


docs/guides/incremental_time.md
```diff
@@ -171,7 +171,12 @@ The check is performed at plan time based on the model definition. SQLMesh may n
 
 A model's `on_destructive_change` [configuration setting](../reference/model_configuration.md#incremental-models) determines what happens when SQLMesh detects a destructive change.
 
-By default, SQLMesh will error so no data is lost. You can set `on_destructive_change` to `warn` or `allow` in the model's `MODEL` block to allow destructive changes.
+By default, SQLMesh will error so no data is lost. You can set `on_destructive_change` to `warn` or `allow` in the model's `MODEL` block to allow destructive changes. 
+`ignore` can be used to not perform the schema change and allow the table's definition to diverge from the model definition.
+
+!!! warning "Ignore is Dangerous"
+
+    `ignore` is dangerous since it can result in error or data loss. It likely should never be used but could be useful as an "escape-hatch" or a way to workaround unexpected behavior.
```
izeigerman, line 179: Why would it result in data loss?

izeigerman:
>  wasn't aware then it could generate alter commands that involve changing columns assuming a column order that wouldn't really happen.

I'm somewhat confused. Does changing the position result in a destructive change?

docs/guides/incremental_time.md
izeigerman: Also how can it result in error?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1594,62 +1617,86 @@ def delete(self, name: str, **kwargs: t.Any) -> None:
         logger.info("Dropped table '%s'", name)
 
 
-class IncrementalByPartitionStrategy(MaterializableStrategy):
+class IncrementalMaterializableStrategy(MaterializableStrategy, abc.ABC):
+    def _ensure_user_query_columns_to_types(
+        self, table_name: str, model: Model, render_kwargs: t.Dict[str, t.Any]
+    ) -> t.Dict[str, t.Any]:
```
izeigerman: Why `t.Any` in the return value?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1865,10 +1928,12 @@ def append(
         table_name: str,
         query_or_df: QueryOrDF,
         model: Model,
+        render_kwargs: t.Dict[str, t.Any],
         **kwargs: t.Any,
     ) -> None:
-        # Source columns from the underlying table to prevent unintentional table schema changes during the insert.
-        columns_to_types = self.adapter.columns(table_name)
+        columns_to_types, extra_columns_to_types = self._get_extra_columns(
```
izeigerman: Do we use `extra_columns_to_types` here?

docs/guides/incremental_time.md
eakmanrq: >Also how can it result in error?

If you made a data type change to a new type that isn't coerceable. The next time it runs it would error trying to insert the rows (since the underlying table wasn't actually changed).

>Why would it result in data loss?

Not certain yet. Putting this warning for now as we do more testing with this fully implemented in the dbt adapter. After a bit if we don't identify any issues I can remove this part of the warning. For now though I would rather overly caution native users about this feature.

sqlmesh/core/engine_adapter/_typing.py
```diff
@@ -13,7 +13,7 @@
 
     snowpark = optional_import("snowflake.snowpark")
 
-    Query = t.Union[exp.Query, exp.DerivedTable]
+    Query = exp.Query
```
eakmanrq, line 16: Discussed internally and we don't see a reason why `exp.DerivedTable` was expected before and that code is from version `0.0.2` so assuming it is no longer needed.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -199,9 +204,18 @@ def catalog_support(self) -> CatalogSupport:
         return CatalogSupport.UNSUPPORTED
 
     @classmethod
-    def _casted_columns(cls, columns_to_types: t.Dict[str, exp.DataType]) -> t.List[exp.Alias]:
+    def _casted_columns(
+        cls,
+        columns_to_types: t.Dict[str, exp.DataType],
```
izeigerman: do we want to be more specific with this name. i.e. `target_columns_to_types`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -199,9 +204,18 @@ def catalog_support(self) -> CatalogSupport:
...
+        source_columns: t.Optional[t.List[str]] = None,
+    ) -> t.List[exp.Alias]:
+        source_columns = source_columns or list(columns_to_types)
```
izeigerman: should this be a set? I see that each column gets looked up

sqlmesh/core/engine_adapter/base.py
```diff
@@ -227,12 +241,31 @@ def _get_source_queries(
...
 
         batch_size = self.DEFAULT_BATCH_SIZE if batch_size is None else batch_size
-        if isinstance(query_or_df, (exp.Query, exp.DerivedTable)):
-            return [SourceQuery(query_factory=lambda: query_or_df)]  # type: ignore
+        if isinstance(query_or_df, exp.Query):
+            query_factory = lambda: query_or_df
+            if source_columns:
+                if not columns_to_types:
+                    raise SQLMeshError("columns_to_types must be set if source_columns is set")
+                if not set(columns_to_types).issubset(set(source_columns)):
```
izeigerman: Perhaps this is just a syntax issue but I have a hard time interpreting why would `columns_to_types` ever be a subset of `source_columns`.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -227,12 +241,31 @@ def _get_source_queries(
...
+                    select_columns = [
+                        exp.column(c, quoted=True)
+                        if c in source_columns
```
izeigerman: ditto: should `source_columns` be a set?

sqlmesh/core/engine_adapter/base.py
eakmanrq: I originally used a set but it creates a bit of a trap for developers: It can be very tempting, if needing to select columns from a user query/df, to use `source_columns`. If you do then you loose the order. So I switch to list to avoid this potential bug.

sqlmesh/core/engine_adapter/base.py
eakmanrq: https://github.com/TobikoData/sqlmesh/pull/5117#discussion_r2277627588

sqlmesh/core/engine_adapter/base.py
eakmanrq: That would be the common case where they are the same. If source columns is less then you would evaluate this logic.

sqlmesh/core/engine_adapter/base.py
eakmanrq: Currently the engine adapter interface as an implicit understanding that `columns_to_types` is the `target_columns_to_types`. I would hoping the existence of source columns makes that clear without having to rename everything and extending the length of that variable.

sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -164,22 +164,21 @@ def _df_to_source_queries(
...
-        df = self._ensure_pyspark_df(df, columns_to_types)
+        pyspark_df = self._ensure_pyspark_df(df, columns_to_types, source_columns=source_columns)
 
         def query_factory() -> Query:
             temp_table = self._get_temp_table(target_table or "spark", table_only=True)
-            df.createOrReplaceTempView(temp_table.sql(dialect=self.dialect))
+            pyspark_df.createOrReplaceTempView(temp_table.sql(dialect=self.dialect))
             self._connection_pool.set_attribute("use_spark_engine_adapter", True)
-            return exp.select(*self._casted_columns(columns_to_types)).from_(temp_table)
+            return exp.select(*self._select_columns(columns_to_types)).from_(temp_table)
```
izeigerman: we don't need to cast anymore?

eakmanrq: Correct `_ensure_pyspark_df` does the cast now so we just need to select. That is also why we don't need to include source columns since it has already been padded.

sqlmesh/core/engine_adapter/spark.py
```diff
@@ -240,61 +240,91 @@ def try_get_pandas_df(cls, value: t.Any) -> t.Optional[pd.DataFrame]:
...
     def _ensure_pyspark_df(
-        self, generic_df: DF, columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
+        self,
+        generic_df: DF,
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        source_columns: t.Optional[t.List[str]] = None,
     ) -> PySparkDataFrame:
-        pyspark_df = self.try_get_pyspark_df(generic_df)
-        if pyspark_df:
+        def _get_pyspark_df() -> PySparkDataFrame:
```
izeigerman: What's the point of having this inner function here? It doesn't look like it gets called more than once. IMHO it doesn't make the code more readable either since it's part of another function and only makes it more nested.

sqlmesh/core/engine_adapter/base.py
izeigerman: I don't follow how this is relevant. Doesn't this lead to `O(n^2)` though?

sqlmesh/core/engine_adapter/base.py
izeigerman: FWIW, if there's any such understanding at all today, it's very implicit even to me. If you asked me about what `columns_to_types` was before this PR I'd tell you it's the model's `columns_to_types`.

sqlmesh/core/engine_adapter/spark.py
eakmanrq: The alternative is a lot of logic indented because of this:
```
pyspark_df = self.try_get_pyspark_df(generic_df)
if pyspark_df:
    return pyspark_df
```
Otherwise it would be something like this:
```
pyspark_df = self.try_get_pyspark_df(generic_df)
if not pyspark_df:
   ....
```

I felt like that was easier to read since I could return and separates out getting the pyspark df and then selecting the columns (regardless of how we got it).

sqlmesh/core/engine_adapter/base.py
eakmanrq: You would prefer then to have it renamed within the engine adapter class?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1606,6 +1605,66 @@ def delete(self, name: str, **kwargs: t.Any) -> None:
...
+    def _get_target_and_source_columns(
+        self,
+        model: Model,
+        table_name: str,
+        render_kwargs: t.Dict[str, t.Any],
+        columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+    ) -> t.Tuple[t.Dict[str, exp.DataType], t.Optional[t.List[str]]]:
+        if not columns_to_types:
+            columns_to_types = (
+                model.columns_to_types if model.annotated else self.adapter.columns(table_name)
```
izeigerman: I'm a bit confused. I thought we should only use `model.columns_to_types` for `source_columns` and never for actual `columns_to_types`.

Btw, this is another reason why we should be consistent about the `columns_to_types` name throughout the code base.

izeigerman: At the very least we should source `columns_to_types` from the table **always** if `model.on_destructive_change.is_ignore` is true

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1606,6 +1605,66 @@ def delete(self, name: str, **kwargs: t.Any) -> None:
...
+        **kwargs: t.Any,
+    ) -> None:
+        """Replaces the table for the given model.
+
+        Args:
+            model: The target model.
+            name: The name of the target table.
+            query_or_df: The query or DataFrame to replace the target table with.
+        """
+        if (model.is_seed or model.kind.is_full) and model.annotated:
```
izeigerman, line 1606: isn't `not model.is_incremental` a more accurate criteria here?

eakmanrq: Original source for this logic: https://github.com/TobikoData/sqlmesh/commit/012f851dea0ecfca3871fd024cac72c338f8ab57

sqlmesh/core/snapshot/evaluator.py
izeigerman: I don't think we should rely on the caller to provide the correct `columns_to_types`. I think this method should make a decision instead

eakmanrq: I think `model.columns_to_types` if fine but I should add a condition to this which is `if model.annotated not model.on_destructive_change.is_ignore`. Does that make sense?

sqlmesh/core/engine_adapter/base.py
izeigerman: sure

sqlmesh/core/engine_adapter/base.py
izeigerman: Wouldn't it be cleaner to test for set equality then? If this is an intention

sqlmesh/core/snapshot/evaluator.py
izeigerman: yes, this matches my previous comment in this thread

izeigerman: the problem with this is that this method assumes that the caller does the right thing, which is a dangerous assumption. If you **always** check `model.on_destructive_change.is_ignore` and run `adapter.columns` then you risk running `adapter.columns` more than once like in case of SCD2

sqlmesh/core/engine_adapter/base.py
eakmanrq: Well this logic doesn't handle the case currently when source columns is more (which will come with ignore additive  changes). Its strictly written to deal with when `columns_to_types` is not a subset.

sqlmesh/core/engine_adapter/spark.py
izeigerman: IMHO the fact that early returns are easier to read / follow is a very subjective statement. Personally, I really hate early returns because they break the flow unexpectedly and debugging them is a nightmare. I'm ok with this block being a separate function but any chance it can become top-level?

eakmanrq: I don't want anything else to use it besides this method so if I'm choosing between those option I will do the indented option.

izeigerman (approved):

PR #5119, opened 2025-08-11
feat: function to get range of key/value in model block

benfdking:
# Stages of the changes

- **refactor: refactor dialect parsing code to be reused**
- **feat: add the ability to return range for key and value**

# Actual changes

- improved the parsing logic and made it more robust
- added the ability to return the value for the value range as well

### DIFF
--- a/sqlmesh/core/linter/helpers.py
+++ b/sqlmesh/core/linter/helpers.py
@@ -150,33 +150,21 @@ def get_start_and_end_of_model_block(
     except StopIteration:
         return None
 
-    # 3) Find the matching closing parenthesis for that list by tracking depth
-    depth = 0
-    rparen_idx: t.Optional[int] = None
-    for i in range(lparen_idx, len(tokens)):
-        tt = tokens[i].token_type
-        if tt is TokenType.L_PAREN:
-            depth += 1
-        elif tt is TokenType.R_PAREN:
-            depth -= 1
-            if depth == 0:
-                rparen_idx = i
-                break
-
-    if rparen_idx is None:
-        # Fallback: stop at the first semicolon after MODEL
-        try:
-            rparen_idx = next(
-                i
-                for i in range(lparen_idx + 1, len(tokens))
-                if tokens[i].token_type is TokenType.SEMICOLON
-            )
-        except StopIteration:
-            return None
-    return (
-        lparen_idx,
-        rparen_idx,
-    )
+    # 3) Find the matching closing parenthesis by looking for the first semicolon after
+    # the opening parenthesis and assuming the MODEL block ends there.
+    try:
+        closing_semicolon = next(
+            i
+            for i in range(lparen_idx + 1, len(tokens))
+            if tokens[i].token_type is TokenType.SEMICOLON
+        )
+        # If we find a semicolon, we can assume the MODEL block ends there
+        rparen_idx = closing_semicolon - 1
+        if tokens[rparen_idx].token_type is TokenType.R_PAREN:
+            return (lparen_idx, rparen_idx)
+        return None
+    except StopIteration:
+        return None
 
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/linter/helpers.py
```diff
@@ -156,57 +156,193 @@ def get_range_of_model_block(
...
+    rparen_idx: t.Optional[int] = None
+    for i in range(lparen_idx, len(tokens)):
+        tt = tokens[i].token_type
+        if tt is TokenType.L_PAREN:
+            depth += 1
+        elif tt is TokenType.R_PAREN:
+            depth -= 1
+            if depth == 0:
+                rparen_idx = i
+                break
```
georgesittas: Should we just match the first semicolon and get the token before that for the closing parenthesis? Tracking depth seems a bit much.

sqlmesh/core/linter/helpers.py
```diff
@@ -156,57 +156,193 @@ def get_range_of_model_block(
...
+
+    if rparen_idx is None:
+        # Fallback: stop at the first semicolon after MODEL
+        try:
+            rparen_idx = next(
+                i
+                for i in range(lparen_idx + 1, len(tokens))
+                if tokens[i].token_type is TokenType.SEMICOLON
+            )
```
georgesittas: Ah, we already do this, great. Why not just use this instead of having it as a fallback?

sqlmesh/core/linter/helpers.py
```diff
@@ -156,57 +156,193 @@ def get_range_of_model_block(
...
+            if depth <= 0:
+                break
+            continue
+
+        if depth == 1 and tt is TokenType.VAR and tok.text.upper() == key.upper():
+            # Validate key position: it should immediately follow '(' or ',' at top level
+            prev_idx = i - 1
+            # Skip comments
+            while prev_idx >= 0 and tokens[prev_idx].token_type in (TokenType.COMMENT,):
+                prev_idx -= 1
```
georgesittas: What case is this expected to handle? Comments aren't represented using tokens.

sqlmesh/core/linter/helpers.py
```diff
@@ -156,57 +156,193 @@ def get_range_of_model_block(
...
+                continue
+
+            # Key range
+            lines = sql.splitlines()
+            key_start = TokenPositionDetails(
+                line=tok.line, col=tok.col, start=tok.start, end=tok.end
+            )
+            key_range = key_start.to_range(lines)
+
+            # Find value start: the next non-comment token after the key
```
georgesittas: Ditto about comment here.

sqlmesh/core/linter/helpers.py
```diff
@@ -156,57 +156,193 @@ def get_range_of_model_block(
...
+                return t in (TokenType.L_PAREN, TokenType.L_BRACE, TokenType.L_BRACKET)
+
+            def is_close(t: TokenType) -> bool:
+                return t in (TokenType.R_PAREN, TokenType.R_BRACE, TokenType.R_BRACKET)
+
+            while j < rparen_idx:
+                ttype = tokens[j].token_type
+                if ttype is TokenType.COMMENT:
+                    j += 1
+                    continue
```
georgesittas: Ditto

sqlmesh/core/linter/helpers.py
```diff
@@ -153,60 +193,124 @@ def get_range_of_model_block(
...
+                continue
+
+            # Key range
+            lines = sql.splitlines()
+            key_start = TokenPositionDetails(
+                line=tok.line, col=tok.col, start=tok.start, end=tok.end
+            )
+            key_range = key_start.to_range(lines)
+
+            # Find value start: the next non-comment token after the key
```
```suggestion
```

PR #5122, opened 2025-08-11
Chore: Info instead of warn for conditional properties assignment

themisvaltinos:
Use info in the logs instead of warning for cases where a property is conditionally assigned.


georgesittas (approved):
LGTM, thank you.

Just to share some context, the rationale behind this change was that `None` and `NULL` are valid values that are used for conditional property rendering, so warning about them is too noisy and there is no clear action for the user to silence that warning, despite these values being intentionally used on their end.

PR #5127, opened 2025-08-11
chore: adding shared ui setup

benfdking:


web/shared_ui/eslint.config.mjs
```diff
@@ -0,0 +1,26 @@
+import eslint from '@eslint/js'
+import tseslint from 'typescript-eslint'
+
+export default tseslint.config(
+  {
+    ignores: [
+      'dist/**/*',
+      'node_modules/**',
+      'src/utils/tbk-components.js',
```
mykalmax: i think this is not relevant for shared package

web/shared_ui/eslint.config.mjs
```diff
@@ -0,0 +1,26 @@
...
+      '**/*.cjs',
+      '**/*.mjs',
+      'src/api/client.ts',
```
mykalmax: i think this is not relevant for shared package

mykalmax (approved):

PR #5132, opened 2025-08-12
Feat(dbt_cli): Use sqlmesh.yaml for config instead of trying to inject dbt_project.yml

erindru:


sqlmesh_dbt/operations.py
```diff
@@ -65,21 +65,26 @@ def create(
...
         configure_logging(force_debug=debug)
         set_console(DbtCliConsole())
 
         progress.update(load_task_id, description="Loading project", total=None)
 
-        # inject default start date if one is not specified to prevent the user from having to do anything
-        _inject_default_start_date(project_dir)
+        cli_config = get_or_create_sqlmesh_config(project_dir)
+        # todo: we will need to build this out when we start storing more than model_defaults
+        model_defaults = (
```
izeigerman, line 77: I don't like this alternative config here. It should be normal SQLMesh config.

Please note that despite having a separate CLI command I still want other `sqlmesh` commands to work as expected. The current direction doesn't ensure that

erindru: The problem with normal SQLMesh config is that it has fields and validations designed for normal SQLMesh projects, which includes a bunch of redundancy that is already specified in `dbt_project.yml` or `profiles.yml` in a dbt project (like the entire connection config).

It's also called `config.yaml` which is kind of generic next to `dbt_project.yml` instead of something more obvious like `sqlmesh.yaml`.

But i'll investigate how it can be re-used instead of defining a working subset tuned down to just what's required for the dbt-compatible CLI.

>  I still want other sqlmesh commands to work as expected

Understood, that helps a lot

PR #5133, opened 2025-08-12
Fix: Use drop cascade in janitor

erindru:
Second attempt at https://github.com/TobikoData/sqlmesh/pull/5098

Fixes https://github.com/TobikoData/sqlmesh/issues/5083 https://github.com/TobikoData/sqlmesh/issues/4767

The solution in #5098 unfortunately needed to read in all the view snapshot records in order to work out the dependency graph of what should be dropped, because the way our state database is currently structured means that this operation can't easily be pushed down to the database level.

Fetching a large amount of snapshots from state sync can cause memory problems in other state sync implementations that buffer instead of stream so this approach was considered a non-starter.

In addition, it turns out that having dangling pointers in state is not as bad as I originally thought it was due to "create if not exists"-style logic elsewhere.

So this PR does the following:
 - Implement `DROP CASCADE` in the SnapshotEvaluator cleanup method. This leaves dangling pointers in state but that is considered acceptable
 - Check that it works by producing a scenario on Postgres that was previously failing


sqlmesh/core/engine_adapter/athena.py
```diff
@@ -314,13 +314,13 @@ def _build_table_properties_exp(
 
         return None
 
-    def drop_table(self, table_name: TableName, exists: bool = True) -> None:
+    def drop_table(self, table_name: TableName, exists: bool = True, **kwargs: t.Any) -> None:
```
erindru, line 318: Since `drop_view()` already had `**kwargs` I figured it was ok to add it to `drop_table()` as well

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1590,7 +1594,7 @@ def migrate(
 
     def delete(self, name: str, **kwargs: t.Any) -> None:
         _check_table_db_is_physical_schema(name, kwargs["physical_schema"])
-        self.adapter.drop_table(name)
+        self.adapter.drop_table(name, cascade=kwargs.pop("cascade", False))
```
erindru, line 1733: I deliberately just pop off the `cascade` argument because `delete()` is called with a bunch of other arguments that arent relevant for `drop_table()` but end up making their way to the `exp.Drop` AST node in the EngineAdapter if they aren't filtered out here

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1083,6 +1083,10 @@ def _cleanup_snapshot(
                     table_name,
                     is_table_deployable=is_table_deployable,
                     physical_schema=snapshot.physical_schema,
+                    # we need to set cascade=true or we will get a 'cant drop because other objects depend on it'-style
+                    # error on engines that enforce referential integrity, such as Postgres
+                    # this situation can happen when a snapshot expires but downstream view snapshots that reference it have not yet expired
+                    cascade=True,
```
themisvaltinos, line 1213: should we control this with a flag which is set if the engine supports cascade or not (maybe from the schema differ)? unless im doing something wrong I tried with a BigQuery project for example to run the janitor which stops when it tries to drop a table with the error `Syntax error: Expected end of input but got keyword CASCADE at`

relevant docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#drop_table_statement it seems cascade is supported for schema but not table

georgesittas: > should we control this with a flag which is set if the engine supports cascade or not (maybe from the schema differ)?

Imo, this should happen further downstream, e.g., in the adapter itself.

erindru: Oh, nice catch, the downside of not running this test across all engines.

I naively thought SQLGlot would not generate the `CASCADE` output if it's unsupported (even if cascade=true on the AST node) but I guess the fact it doesn't is why its not considered a validator.

I'll improve the coverage and make sure this works on all engines

.circleci/continue_config.yml
```diff
@@ -305,10 +305,10 @@ workflows:
                 - clickhouse-cloud
                 - athena
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru: TODO: revert prior to merge

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -3548,3 +3545,137 @@ def test_identifier_length_limit(ctx: TestContext):
...
+
+@pytest.mark.parametrize(
+    "environment_suffix_target",
+    [
+        EnvironmentSuffixTarget.TABLE,
+        EnvironmentSuffixTarget.SCHEMA,
+        EnvironmentSuffixTarget.CATALOG,
+    ],
+)
+def test_janitor(
```
erindru, line 3567: This test inits a project, creates a dev env, invalidates it, runs the janitor to clean it up and checks it was cleaned up
 - Across every engine we support
 - For every EnvironmentSuffixTarget we support

themisvaltinos (approved):
lgtm, nice the supported cascade list in the adapters is a much better solution to cover every operation than the flag i was suggesting for only this use case and thanks for adding the integration test!

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1063,8 +1065,13 @@ def _drop_object(
             name: The name of the table to drop.
             exists: If exists, defaults to True.
             kind: What kind of object to drop. Defaults to TABLE
+            cascade: Whether or not to DROP ... CASCADE.
+                Note that this is ignored for :kind's that are not present in self.SUPPORTED_DROP_CASCADE_OBJECT_KINDS
             **drop_args: Any extra arguments to set on the Drop expression
         """
+        if cascade and kind.upper() in self.SUPPORTED_DROP_CASCADE_OBJECT_KINDS:
+            drop_args["cascade"] = cascade
```
georgesittas, line 1080: Since we're expecting `CASCADE` semantics, should this warn if cascade is unsupported?

georgesittas (approved):
Did a quick pass– looks reasonable.

sqlmesh/core/engine_adapter/base.py
izeigerman: I think we don't really "expect" the `CASCADE` semantics per se. We just want to delete our stuff without error. We're **forced** to do CASCADE because otherwise the engine won't let us delete.

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -24,6 +24,7 @@ class BasePostgresEngineAdapter(EngineAdapter):
     DEFAULT_BATCH_SIZE = 400
     COMMENT_CREATION_TABLE = CommentCreationTable.COMMENT_COMMAND_ONLY
     COMMENT_CREATION_VIEW = CommentCreationView.COMMENT_COMMAND_ONLY
+    SUPPORTED_DROP_CASCADE_OBJECT_KINDS = ["SCHEMA", "TABLE", "VIEW"]
```
izeigerman, line 27: IMHO I don't know if we should ever apply this for tables. I realized that this was added for completeness sake, but I'd rather make a conscious choice for supporting this when the need arises.

izeigerman (approved):

sqlmesh/core/engine_adapter/base.py
georgesittas: That's fair; I was thinking that a warning could simply surface the fact that some objects were left orphan because we couldn't cascade. Although, if we don't know _which_ objects depend on the removed thing, then its value is probably questionable anyway.

erindru: Yeah I did consider warning but then I noticed we silently do nothing [elsewhere](https://github.com/TobikoData/sqlmesh/blob/014fe6ae850780fd206341c8cb4835b4d6276f4e/sqlmesh/core/engine_adapter/mysql.py#L78) so decided to keep with that strategy

PR #5140, opened 2025-08-12
Fix(dbt): Use the info to control logging for dbt log builtin

themisvaltinos:
Previously when using in dbt `log` the info parameter was ignored and we only logged in the logs if the debug mode was enabled. This updates the log builtin to use the info flag to control the levelling level to match dbt's handling ( https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/context/base.py#L570)

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -170,7 +170,13 @@ def env_var(name: str, default: t.Optional[str] = None) -> t.Optional[str]:
 
 
 def log(msg: str, info: bool = False) -> str:
-    logger.debug(msg)
+    if info:
+        # Write to both log file and stdout
+        logger.info(msg)
+        print(msg)
+    else:
+        logger.debug(msg)
+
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -170,7 +170,13 @@ def env_var(name: str, default: t.Optional[str] = None) -> t.Optional[str]:
 
 
 def log(msg: str, info: bool = False) -> str:
-    logger.debug(msg)
+    if info:
+        # Write to both log file and stdout
+        logger.info(msg)
+        print(msg)
```
georgesittas: Is this `print` here faithful to dbt's behavior?

themisvaltinos: yes it is in the cli as well: https://docs.getdbt.com/reference/dbt-jinja-functions/log

erindru (approved):

PR #5141, opened 2025-08-12
Fix: support auto_restatement_cron in python models

treysp:
For python models we currently attempt to render `auto_restatement_cron`, which incorrectly treats values like `@hourly` as SQLMesh macro variables.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2866,6 +2866,9 @@ def render_meta_fields(
             for key, value in field_value.items():
                 if key in RUNTIME_RENDERED_MODEL_FIELDS:
                     rendered_dict[key] = parse_strings_with_macro_refs(value, dialect)
+                elif key == "auto_restatement_cron":
+                    # Don't parse auto_restatement_cron="@..." kwarg (e.g. @daily) into MacroVar
+                    rendered_dict[key] = value
                 elif (rendered := render_field_value(value)) is not None:
                     rendered_dict[key] = rendered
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -2817,6 +2817,9 @@ def render_field_value(value: t.Any) -> t.Any:
             for key, value in field_value.items():
                 if key in RUNTIME_RENDERED_MODEL_FIELDS:
                     rendered_dict[key] = parse_strings_with_macro_refs(value, dialect)
+                elif key == "auto_restatement_cron":
```
erindru: I may be missing some context but doesn't this mean that people cant use macros at all? I believe we support macros in normal `cron` with a heuristic on line 2805, would it make sense to just extend that to include `auto_restatement_cron` as well?

treysp: D'oh - totally missed that...

treysp: Thank you!

tests/core/test_model.py
```diff
@@ -2881,7 +2881,15 @@ def my_model_2(context):
     # no warning with valid kind dict
     with patch.object(get_console(), "log_warning") as mock_logger:
 
-        @model("kind_valid_dict", kind=dict(name=ModelKindName.FULL), columns={'"COL"': "int"})
```
erindru, line 2901: Should testing the behaviour of `auto_restatement_cron` be its own entry in this test vs modifying an existing entry that was testing something else?

PR #5144, opened 2025-08-13
fix: properly load dbt relation type for get_relation() and related functions

### LINKED ISSUE #1987: Set the correct dbt Relation type
Currently SQLMesh always sets the `type` value to None when creating a relation instance returned by the `adapter.get_relation` and similar methods.

newtonapple:
fixes #1987


sqlmesh/dbt/adapter.py
```diff
@@ -259,6 +260,9 @@ def load_relation(self, relation: BaseRelation) -> t.Optional[BaseRelation]:
         if not self.engine_adapter.table_exists(mapped_table):
             return None
 
+        if do := self.engine_adapter.get_data_object(mapped_table):
+            return self._data_object_to_relation(do)
+
         return self._table_to_relation(mapped_table)
```
newtonapple: We can technically return `None` here, but I'm keep the existing behavior.

sqlmesh/dbt/adapter.py
```diff
@@ -259,6 +260,9 @@ def load_relation(self, relation: BaseRelation) -> t.Optional[BaseRelation]:
         if not self.engine_adapter.table_exists(mapped_table):
```
newtonapple: This is technically redundant, but can be a bit faster without caching if the table doesn't exist.

sqlmesh/dbt/adapter.py
```diff
@@ -259,6 +260,10 @@ def load_relation(self, relation: BaseRelation) -> t.Optional[BaseRelation]:
...
             return None
 
+        data_object = self.engine_adapter.get_data_object(mapped_table)
```
newtonapple, line 261: This can be problematic as `get_data_object()` calls `get_data_objects()` which loads all the relations in memory for every call to `get_data_object()`. DBT has a caching mechanism to cache the global relations. I wonder if we should do the same.

izeigerman: > This can be problematic as get_data_object() calls get_data_objects() which loads all the relations in memory for every call to get_data_object()

what all relations? Doesn't it filter?

newtonapple: Since we actually push down the filter to actual query in our `get_data_object()` implementation, we concluded cache is not essential here.

PR #5147, opened 2025-08-13
Chore!: Optimize snapshot unpausing

izeigerman:
With introduction of the dev-only VDE mode, all snapshots that share the name also share the version. This means that fetching all snapshots with the same version may no longer be feasible, as it could result in retrieving the entire state in extreme cases.

This update optimizes the snapshot unpausing operation eliminating the need to fetch snapshots with shared versions. There were a few changes in the past that allow us to simplify the current implementation substantially:

- Deprecation of the Airflow support means that we no longer need to align the unpaused timestamp by the interval unit
- Interval compaction for garbage-collected snapshots means we no longer track intervals for individual snapshots, so there is no need to accurately reflect the effective-from timestamp in the state. Please note the effective-from timestamp is still considered when merging intervals for individual snapshots.

### DIFF
--- a/sqlmesh/core/state_sync/db/snapshot.py
+++ b/sqlmesh/core/state_sync/db/snapshot.py
@@ -10,6 +10,7 @@ from pydantic import Field
 
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.state_sync.db.utils import (
+    snapshot_name_filter,
     snapshot_name_version_filter,
     snapshot_id_filter,
     fetchone,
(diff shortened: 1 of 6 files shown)


sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -112,84 +112,48 @@ def unpause_snapshots(
...
-                            )
-                        ]
-                    )
-
-                if snapshot.unpaused_ts:
-                    logger.info("Pausing snapshot %s", snapshot.snapshot_id)
-                    snapshot.set_unpaused_ts(None)
-                    paused_snapshots.append(snapshot.snapshot_id)
+        for snapshot in snapshots:
+            # We need to mark all other snapshots that have opposite forward only status as unrestorable
```
tobymao: what is opposite forward only status, you mean non forward only?

izeigerman: It means that if the target is forward-only, then all non-forward-only snapshots should become unrestorable and vice-versa.

erindru (approved):

PR #5150, opened 2025-08-13
Fix: Skip cleanup of missing physical tables

izeigerman:


georgesittas (approved):

georgesittas:
Should we warn if the object doesn't exist?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1073,11 +1073,20 @@ def _cleanup_snapshot(
...
-            evaluation_strategy.delete(
-                table_name,
-                is_table_deployable=is_table_deployable,
-                physical_schema=snapshot.physical_schema,
-            )
+            # Use `get_data_object` to check if the table exists instead of `table_exists` since the former
+            # is based on `INFORMATION_SCHEMA` and avoids touching the table directly.
+            # This is important when the table name is malformed for some reason and running any statement
+            # that touches the table would result in an error.
+            if adapter.get_data_object(table_name) is not None:
```
erindru: Will this cause every deletion to be at minimum 2 queries? That could get quite slow

izeigerman: Correct, but this only happens in the janitor so I'm not too concerned. Do you have alternative suggestions?

erindru: Mostly just wrap it in a try/catch and if an exception is thrown only then do the `exists()` check.

If the `exists()` check indicates that the table does not exist then we can continue, if it indicates that the table still exists after the drop attempt then we can reraise the exception because something else is going on

This would cut down on the number of queries which I think does matter when there are 1000 objects to clean up

izeigerman: ah, that's a valid suggestion. Thanks

izeigerman: fixed

erindru (approved):

PR #5153, opened 2025-08-14
fix: BigQueryEngineAdapter.get_table_schema() mypy error

newtonapple:
This follows the same logic from [BigQueryEngineAdapter.columns()](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/bigquery.py#L351-L357) to ensure `query_job` is not None before using it.

### DIFF
--- a/sqlmesh/core/engine_adapter/bigquery.py
+++ b/sqlmesh/core/engine_adapter/bigquery.py
@@ -296,7 +296,9 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
         table = exp.to_table(table_name)
         if len(table.parts) == 3 and "." in table.name:
             self.execute(exp.select("*").from_(table).limit(0))
-            return self._query_job._query_results.schema
+            query_job = self._query_job
+            assert query_job is not None
+            return query_job._query_results.schema
         return self._get_table(table).schema
 
     def columns(


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -296,7 +296,9 @@ def get_bq_schema(self, table_name: TableName) -> t.List[bigquery.SchemaField]:
         table = exp.to_table(table_name)
         if len(table.parts) == 3 and "." in table.name:
             self.execute(exp.select("*").from_(table).limit(0))
-            return self._query_job._query_results.schema
+            query_job = self._query_job
+            assert query_job is not None
```
erindru, line 300: Since asserts only work when running python in debug mode, it's generally better to do something like:

```
if query_job is None:
  raise ...
```

mypy will still pick up on that and it makes the check explicit

georgesittas: I think we have quite a few `assert`s in the codebase. Is this something we want to be consistent about? Not sure if we've been thinking about them the way you described, [but you're right](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement):

> In the current implementation, the built-in variable `__debug__` is `True` under normal circumstances, `False` when optimization is requested (command line option [-O](https://docs.python.org/3/using/cmdline.html#cmdoption-O)). The current code generator emits no code for an [assert](https://docs.python.org/3/reference/simple_stmts.html#assert) statement when optimization is requested at compile time

erindru: Yeah I realised that we have been using `assert` incorrectly outside of tests to make the type checker happy.

I learned that validating the assumptions using actual runtime checks:
 - Still keeps the type checker happy in most cases
 - Forces us to handle the case when something that theoretically shouldn't happen, happens anyway

newtonapple: Sure, I can change it to raise a `RuntimeError` or an explicit `AssertionError`.  But like I said in the comment, I'm simply porting the original logic over.  Do you prefer to change the [BigQueryEngineAdapter.columns()](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/engine_adapter/bigquery.py#L351-L357) implementation to raise an exception explicitly as well?

georgesittas:
@newtonapple @erindru I'll go ahead and merge this PR for now to unblock CI in main / other PRs. Let's address the `assert` vs `raise` in a followup.

sqlmesh/core/engine_adapter/bigquery.py
georgesittas: Perhaps we can one-shot this more generally in the codebase by asking Claude to do it. Like, "change assertions aiming to satisfy mypy with `raise` statements" or something.

Or... just do it for the BigQuery adapter and leave the bigger refactor for later :)

PR #5154, opened 2025-08-14
Se/improve athena get data objs

sean-eyre:
The current query used to get data objects in Athena becomes very slow as more models are included.

This PR uses a "CTE of VALUES with a join" strategy to speed up the query time significantly.  From ~4 mins to <10 seconds when querying for 412 objects in my case.


erindru:
Nice, thanks!

Can you please sign the CLA and also enable the Athena integration tests to run for this PR, which you can do by adjusting the `engine_tests_cloud` section in `.circleci/continue_config.yml` to something like:

```
          matrix:
            parameters:
              engine:
#                - snowflake
#                - databricks
#                - redshift
#                - bigquery
#                - clickhouse-cloud
                - athena
#                - gcp-postgres
#          filters:
#            branches:
#              only:
#                - main
```

(that is, comment out the branch filter and the other databases to get the athena tests to run)

PR #5156, opened 2025-08-14
Feat: yaml config for dbt projects

erindru:
Second attempt at https://github.com/TobikoData/sqlmesh/pull/5132

This PR enables SQLMesh to load DBT projects using just yaml config, rather than forcing users to create a `config.py` and use python config.

The following channges have been made:
 - `sqlmesh.yaml` is now a valid config file name. This is because it gets written next to the existing `dbt_project.yml` file so I wanted it to be clear that this config is relevant to SQLMesh only, which the existing `config.yaml` name is not clear about
 - `sqlmesh init -t dbt` now produces a small yaml file instead of a python config file
 - A check has been added to `load_configs` to return a `Config` configured with the `DbtLoader` if a `dbt_project.yml` file is detected. This essentially transparently automates what the users had to do manually before
 - The `sqlmesh_dbt` CLI automatically runs `sqlmesh init -t dbt` if it doesn't detect a sqlmesh config file before proceeding normally. This enables a user to get started with dbt on the SQLMesh engine with zero manual steps

The existing config mechanism is preserved - that is, the python base config is overlayed with all the yaml configs to produce a final config

Note: if this approach is considered acceptable, i'll update the docs in a followup PR


sqlmesh/core/config/loader.py
```diff
@@ -51,10 +56,7 @@ def load_configs(
         return {path: config for path in absolute_paths}
 
     config_env_vars = None
-    personal_paths = [
-        sqlmesh_path / "config.yml",
-        sqlmesh_path / "config.yaml",
-    ]
+    personal_paths = [sqlmesh_path / name for name in YAML_CONFIG_FILENAMES]
```
erindru, line 59: I noticed that python config wasnt previously allowed in personal paths, so that's why there is both `YAML_CONFIG_FILENAMES` and `ALL_CONFIG_FILENAMES`

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -296,6 +296,8 @@ def get_bq_schema(self, table_name: TableName) -> t.List[bigquery.SchemaField]:
         table = exp.to_table(table_name)
         if len(table.parts) == 3 and "." in table.name:
             self.execute(exp.select("*").from_(table).limit(0))
+            if not self._query_job:
```
erindru: This is potentially covered by https://github.com/TobikoData/sqlmesh/pull/5153 but I wanted to unblock the build on this PR

sqlmesh/core/config/loader.py
izeigerman: There is a reason for this. We can't correctly merge configs if the one in the home folder is a python config.

sqlmesh_dbt/operations.py
```diff
@@ -109,25 +89,20 @@ def create(
...
-        loaded_project_file = load_yaml(project_yaml_path)
-        start_date_keys = ("start", "+start")
-        if "models" in loaded_project_file and all(
-            k not in loaded_project_file["models"] for k in start_date_keys
-        ):
-            loaded_project_file["models"]["+start"] = yesterday_ds()
-            # todo: this may format the file differently, is that acceptable?
-            with project_yaml_path.open("w") as f:
-                dump(loaded_project_file, f)
+    from sqlmesh.cli.project_init import init_example_project, ProjectTemplate
```
izeigerman, line 102: Shouldn't all these imports be under the check of the file existence? Otherwise doesn't this defeat the purpose?

izeigerman (approved):

sqlmesh_dbt/operations.py
erindru: How do you mean? The `sqlmesh` module has already been imported by the time `init_project_if_required()` is called, so these imports are "free"

PR #5158, opened 2025-08-14
Chore!: deprecate support for mixing Jinja & SQLMesh macros

georgesittas:
Mixing jinja and sqlmesh macros is discouraged today and it probably doesn't quite work in many cases. This PR deprecates any support that existed for mixing the two systems, in favour of reducing the overhead involved in rendering jinja-based queries (e.g., that come from dbt projects). Now, sqlmesh macros are only rendered for non-jinja-based models.

### DIFF
--- a/docs/concepts/macros/jinja_macros.md
+++ b/docs/concepts/macros/jinja_macros.md
@@ -321,6 +321,6 @@ Some SQL dialects interpret double and single quotes differently. We could repla
 
 ## Mixing macro systems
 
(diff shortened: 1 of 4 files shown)


georgesittas:
Discussed more and for now we'll avoid doing this after all, as there's a risk of breaking projects that rely on builtin datetime sqlmesh macro variables.

For future reference, an example like this doesn't work today, because we don't have an AST at load time to extract the referenced variables and populate the `python_env`:

```
MODEL (
  name test_model,
  kind FULL
);

JINJA_QUERY_BEGIN;
SELECT 1 AS
   c1, @foo AS c2
JINJA_END;
```

However, if instead of `@foo` we had `@start_date`, that'd work because it's a builtin. Users may be relying on this behavior, which is exactly why we're not moving forward with this now.

PR #5159, opened 2025-08-14
Feat(dbt): Add dbt graph context variable support

themisvaltinos:
This adds support for the dbt `graph` context variable which contains information about the nodes in a dbt project. 

Docs: https://docs.getdbt.com/reference/dbt-jinja-functions/graph

### DIFF
--- a/sqlmesh/dbt/context.py
+++ b/sqlmesh/dbt/context.py
@@ -11,6 +11,7 @@ from sqlmesh.dbt.builtin import _relation_info_to_relation
 from sqlmesh.dbt.manifest import ManifestHelper
 from sqlmesh.dbt.target import TargetConfig
 from sqlmesh.utils import AttributeDict
+from sqlmesh.utils.conversions import serializable
 from sqlmesh.utils.errors import ConfigError, SQLMeshError
 from sqlmesh.utils.jinja import (
     JinjaGlobalAttribute,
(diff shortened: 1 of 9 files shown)


sqlmesh/dbt/context.py
```diff
@@ -195,6 +196,49 @@ def refs(self) -> t.Dict[str, t.Union[ModelConfig, SeedConfig]]:
...
+                "groups": {},
+                "metrics": {},
+                "nodes": {},
+                "sources": {},
+                "semantic_models": {},
+                "saved_queries": {},
+            }
+
+        manifest = self._manifest._manifest
+        return {
```
erindru: Would it be worth building this structure within `ManifestHelper`? Since you're poking into its private member `_manifest` to pull out info

themisvaltinos: yes good point moved the structure there and changed to not access the private member

sqlmesh/dbt/manifest.py
```diff
@@ -155,6 +157,38 @@ def all_macros(self) -> t.Dict[str, t.Dict[str, MacroInfo]]:
                 result[package_name][macro_name] = macro_config.info
         return result
 
+    @property
+    def flat_graph(self) -> t.Dict[str, t.Any]:
+        return {
+            "exposures": {
+                k: serializable(v.to_dict(omit_none=False))
+                for k, v in getattr(self._manifest, "exposures", {}).items()
```
erindru, line 165: FMI: Is this a shortcut (along with `serializable()`) to just reach into the Manifest object returned by the dbt core lib and pull out fields, vs wrapping them in `*Config` objects and exposing them like we have done for `self.models`, `self.seeds` etc?

If so, this will probably need to be refactored in future if we want to do something else with this info

erindru (approved):

sqlmesh/utils/conversions.py
```diff
@@ -19,3 +20,13 @@ def try_str_to_bool(val: str) -> t.Union[str, bool]:
         return maybe_bool == "true"
 
     return val
+
+
+def serializable(obj: t.Any) -> t.Any:
```
georgesittas: Let's use a verb so it's clear it _serializes_? Right now it reads like a property check, i.e. "is this serializable?"

sqlmesh/dbt/manifest.py
themisvaltinos: I was thinking of this originally as we already convert the node objects for example into Config ones (models, seeds, tests, hooks etc.) but we then add extra sqlmesh-specific fields and map dbt terms to native ones in them. If we used those objects to construct the flat graph,  we’d need to strip out the SQLMesh-specific attributes or the result wouldn't match what dbt returns. So since this is a user facing dictionary to be used in macros and not something we internally use I thought it better not to overcomplicate it and to construct the flat graph similar to dbt and use the already available dictionaries. That way when a user accesses it in a macro they get exactly what they would in dbt

erindru: Yeah that makes sense.

I dont think we will ever get rid of the DBT manifest and the dependency on dbt-core :)

PR #5160, opened 2025-08-14
Feat(dbt): Add dbt debug macro

themisvaltinos:
This adds support for dbt's [debug macro](https://docs.getdbt.com/reference/dbt-jinja-functions/debug-method) in a dbt project which opens an iPython debugger in the context of a macro

```python
{% macro my_macro() %}
  {% set something_complex = my_complicated_macro() %}
  {{ debug() }}
{% endmacro %}
```

This is enabled both by using the dbt `DBT_MACRO_DEBUGGING` environment variable or the sqlmesh native: `SQLMESH_DEBUG` or running with `--debug` flag.

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -21,7 +21,7 @@ from sqlmesh.dbt.adapter import BaseAdapter, ParsetimeAdapter, RuntimeAdapter
 from sqlmesh.dbt.relation import Policy
 from sqlmesh.dbt.target import TARGET_TYPE_TO_CONFIG_CLASS
 from sqlmesh.dbt.util import DBT_VERSION
-from sqlmesh.utils import AttributeDict, yaml
+from sqlmesh.utils import AttributeDict, debug_mode_enabled, yaml
 from sqlmesh.utils.date import now
 from sqlmesh.utils.errors import ConfigError, MacroEvalError
 from sqlmesh.utils.jinja import JinjaMacroRegistry, MacroReference, MacroReturnVal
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -329,6 +338,10 @@ def _try_literal_eval(value: str) -> t.Any:
     "zip_strict": lambda *args: list(zip(*args)),
 }
 
+# Add debug function conditionally both with dbt or sqlmesh equivalent flag
+if os.environ.get("DBT_MACRO_DEBUGGING") or debug_mode_enabled():
+    BUILTIN_GLOBALS["debug"] = debug
```
erindru, line 350: What does dbt do with a `debug()` call where `DBT_MACRO_DEBUGGING` is not set?

Does it just no-op, or does it throw an error? Because if it no-op's then wouldnt we need to always include it in the builtin globals and move the "is enabled?" check to within the implementation of `debug()`?

erindru (approved):

sqlmesh/dbt/builtin.py
themisvaltinos: yes the latter, they have it under a conditional as well so that it throws an error (maybe since they warn "Do not deploy code to production that uses the debug macro" to ensure this is only for debugging. This is why I wanted to exactly mirror their approach to get the same result. 

so running dbt when the flag is not set you get an error it is undefined:
```bash
06:44:48  Encountered an error:
Compilation Error in operation jaffle_shop-on-run-end-0 (./dbt_project.yml)
  'debug' is undefined
```

and for sqlmesh you'd get similarly the undefined error:
```bash
Error: An error occurred during rendering of the 'after_all' statements:
Could not render or parse jinja at 'None'.
'debug' is undefined
```

PR #5162, opened 2025-08-14
feat: track rows processed during model evaluation

treysp:
Captures the number of rows processed (and potentially other information) during a model evaluation for display to the user.

Example running sushi on Bigquery (includes bytes processed for each model batch):

Screenshot 2025-08-18 at 4 21 44 PM

**Context**
- Python client libraries conformant to the DB-API 2.0 spec provide a `cursor.row_count` property that contains the number of rows affected by a DML command execution (e.g., `INSERT`, `CTAS`)
- Variations
  - Some clients provide the property, but they don't actually track rows so always return -1
  - Snowflake conforms other than CTAS
  - Bigquery python API necessary to handle all scenarios

**Implementation context**
- SQLMesh makes many execute calls that should not be tracked
- In most cases, SQLMesh DML actions take place via the engine adapter `_execute` method, so we read the row_count property there immediately after `cursor.execute`
- Snapshot evaluation may take place concurrently, where a single snapshot evaluation occurs within one thread

**Determining what to track**
- We should only track specific operations, so execute does NOT track rows by default
- Some engine adapter methods should always track, so they pass `track_row_count=True` in the execute call within them
- Other engine methods vary in their need to track so expose a track_row_count arg
  - Example: `engine_adapter.insert_append`
    - Should almost always be tracked during snapshot evaluation
    - Should never be tracked when called from state sync

**Tracker implementation**
- Row count values are collected in the `QueryExecutionContext` dataclass
  - Accumulates row counts if called multiple times
  - Ignores row counts of -1 (default value when cursor.row_count not populated)
- Trackers implemented as context manager
- Engine adapter `_execute` calls single `record_execution` method, which determines which tracker is active and should be used (if any)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1754,20 +1755,21 @@ def create(
...
-                    if index == 0:
-                        self._replace_query_for_model(model, table_name, df, **kwargs)
-                    else:
-                        self.adapter.insert_append(
-                            table_name, df, columns_to_types=model.columns_to_types
-                        )
-            except Exception:
-                self.adapter.drop_table(table_name)
-                raise
+        with SeedExecutionTracker.track_execution(model.name):
```
erindru: Is it worth shifting this to `SnapshotEvaluator._create_snapshot` (where the snapshot object is available) so we can track seeds by snapshot_id to keep things uniform and remove the need for a special `SeedExecutionTracker`?

And if we tracked evaluations in `SnapshotEvaluator._evaluate_snapshot` in a similar way, keyed by (snapshot_id, batch_index), would we be able to pull the information out in the scheduler based on snapshot_id regardless of what thread placed it there?

sqlmesh/core/execution_tracker.py
```diff
@@ -0,0 +1,165 @@
...
+    rows processed.
+    """
+
+    _thread_contexts: t.Dict[int, QueryExecutionContext] = {}
+    _contexts_lock = Lock()
+
+    @classmethod
+    def get_execution_context(cls) -> t.Optional[QueryExecutionContext]:
+        thread_id = get_ident()
+        with cls._contexts_lock:
```
erindru: Is there a reason you went with this `_contexts_lock` approach vs just [using](https://docs.python.org/3/library/threading.html#thread-local-data) `threading.local()`?

erindru (commented):
Nice start, I think I follow the logic. I've left some comments for your consideration.

I'm not sure about the `track_row_count` parameter on the engine adapter methods. I'd like to understand more about the undesirable queries that were getting picked up - perhaps that indicates we were being too coarse with the statistics gathering / wrapping in the wrong place

sqlmesh/core/execution_tracker.py
treysp: Uhh...ignorance 😭

sqlmesh/core/execution_tracker.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Context manager for tracking snapshot execution statistics.
+        """
+        if not condition:
+            yield None
+            return
+
+        context = QueryExecutionContext(id=snapshot_id_batch)
+        cls._thread_local.context = context
+        cls._contexts[snapshot_id_batch] = context
```
erindru: I believe this only works (the `_contexts` dict being threadsafe) because of Python's GIL. I think this is ok for now because it's simple and usable GIL-less Python is a while off

sqlmesh/core/execution_tracker.py
```diff
@@ -0,0 +1,110 @@
+from __future__ import annotations
```
izeigerman: Shouldn't this go into `snapshot/`?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -1400,30 +1410,42 @@ def insert_append(
...
+                        track_execution_stats=track_execution_stats,
+                    )
 
     def _insert_append_query(
         self,
         table_name: TableName,
         query: Query,
         target_columns_to_types: t.Dict[str, exp.DataType],
         order_projections: bool = True,
+        track_execution_stats: bool = True,
```
izeigerman: Until we know what other execution stats we need to track, can we maybe be more specific with the name?

sqlmesh/core/console.py
```diff
@@ -4248,13 +4286,52 @@ def _calculate_audit_str_len(snapshot: Snapshot, audit_padding: int = 0) -> int:
...
-            + _calculate_audit_str_len(snapshot, audit_padding),
+            + _calculate_audit_str_len(snapshot, audit_padding)
+            + execution_stats_len,
         )
     return annotation_str_len
+
+
+# Convert number of bytes to a human-readable string
+# https://github.com/dbt-labs/dbt-adapters/blob/34fd178539dcb6f82e18e738adc03de7784c032f/dbt-bigquery/src/dbt/adapters/bigquery/connections.py#L165
+def _format_bytes(num_bytes: t.Optional[int]) -> str:
```
erindru: Did we misappropriate these implementations to match dbt's output exactly?

The reason I ask is because SQLMesh already has a transitive dependency on [humanize](https://pypi.org/project/humanize/) which contains [similar functions](https://humanize.readthedocs.io/en/stable/filesize/)

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2420,9 +2445,29 @@ def _log_sql(
...
+
+    def _execute(self, sql: str, track_rows_processed: bool = False, **kwargs: t.Any) -> None:
         self.cursor.execute(sql, **kwargs)
 
+        if (
+            self.SUPPORTS_QUERY_EXECUTION_TRACKING
+            and track_rows_processed
+            and QueryExecutionTracker.is_tracking()
+        ):
+            rowcount_raw = getattr(self.cursor, "rowcount", None)
```
erindru: Nit: This... seems slightly awkward.

```
if (rowcount := getattr(self.cursor, "rowcount", None)) and rowcount is not None:
  try:
    self._record_execution_stats(sql, int(rowcount))
  except (TypeError, ValueError):
    return
```

But maybe I just like walruses

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -662,3 +667,41 @@ def close(self) -> t.Any:
...
+          - "Table <table_name> successfully created."
+          - "<table_name> already exists, statement succeeded."
+
+        We do not want to record the incorrect row count of 1, so we check whether that row contains the table
+        successfully created string. If so, we return early and do not record the row count.
+
+        Ref: https://github.com/snowflakedb/snowflake-connector-python/issues/645
+        """
+        if rowcount == 1:
+            results = self.cursor.fetchone()
```
erindru: Doesn't this advance the resultset?

So anything calling `execute(..., track_rows_processed=True)` followed by some kind of data read on the cursor would report 0 rows if there was 1 row in the resultset, because we consumed the row for this check?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -169,19 +171,20 @@ def evaluate(
...
-            snapshot=snapshot,
-            snapshots=snapshots,
-            allow_destructive_snapshots=allow_destructive_snapshots or set(),
-            allow_additive_snapshots=allow_additive_snapshots or set(),
-            deployability_index=deployability_index,
-            batch_index=batch_index,
-            target_table_exists=target_table_exists,
-            **kwargs,
-        )
+        with self.execution_tracker.track_execution(f"{snapshot.snapshot_id}_{batch_index}"):
```
erindru: nit: This relies on the string convention being correctly set up everywhere it's needed.

Did you consider wrapping it in a dataclass, eg `SnapshotIdBatch` that takes fields `snapshot_id: SnapshotId` and `batch_index: int` and then override its `__repr__` implementation to always generate the correct string key?

sqlmesh/core/snapshot/execution_tracker.py
```diff
@@ -0,0 +1,97 @@
...
+                if self.stats.total_bytes_processed is None:
+                    self.stats.total_bytes_processed = bytes_processed
+                else:
+                    self.stats.total_bytes_processed += bytes_processed
+
+    def get_execution_stats(self) -> QueryExecutionStats:
+        return self.stats
+
+
+class QueryExecutionTracker:
```
erindru, line 56: Nice, this is a much simpler implementation

erindru (commented):
Nice work!

sqlmesh/core/console.py
treysp: good catch - i didn't realize we already had that dependency!

georgesittas: > [...] has a transitive dependency

If we end up relying on `humanize`, let's make that dependency explicit.

sqlmesh/core/engine_adapter/snowflake.py
treysp: Hmm...i see what you mean. If this is a "regular" query, we just consumed the result someone else will be looking for down the line. 

I guess we could parse the query to determine if it's CTAS without consuming the row?

erindru: > If this is a "regular" query, we just consumed the result someone else will be looking for down the line.

Yes, exactly. I think the key distinction is `SELECT` vs anything else. If it's a SELECT, we cant safely peek into the resultset without advancing it.

I was thinking that in many cases we don't need to parse the query because it was already parsed further up the call chain. Perhaps `EngineAdapter.execute()` could pass the original query as a hint to `EngineAdapter._execute()` (alongside the generated SQL) to avoid parsing each query twice in the case where it's already parsed

izeigerman (approved):

PR #5168, opened 2025-08-15
fix(web_common): install deps

mykalmax:


benfdking:
https://github.com/TobikoData/sqlmesh/pull/5170 This is the fix

PR #5174, opened 2025-08-18
Feat(dbt_cli): Add support for '--profile' and '--target'

erindru:
Enables the usage of the `--profile` flag (to select a specific DBT profile) and the `--target` flag (to select a target within that profile) for the `sqlmesh_dbt` CLI.

Also adds a global error handler so when the target or profile doesnt match, we dont print a stack trace.

### DIFF
--- a/sqlmesh_dbt/error.py
+++ b/sqlmesh_dbt/error.py
@@ -16,7 +16,7 @@ def cli_global_error_handler(
             return func(*args, **kwargs)
         except Exception as ex:
             # these imports are deliberately deferred to avoid the penalty of importing the `sqlmesh`
-            # package for every CLI command
+            # package up front for every CLI command
             from sqlmesh.utils.errors import SQLMeshError
             from sqlglot.errors import SqlglotError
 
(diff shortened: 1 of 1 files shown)


sqlmesh/core/config/loader.py
```diff
@@ -168,7 +171,11 @@ def load_config_from_paths(
         if dbt_project_file:
             from sqlmesh.dbt.loader import sqlmesh_config
 
-            dbt_python_config = sqlmesh_config(project_root=dbt_project_file.parent)
+            dbt_python_config = sqlmesh_config(
```
erindru, line 174: The `config_loader_kwargs` parameter was added to `Context` so it could flow through to here.

I first attempted to make these arguments to the `DbtLoader`, but theyre needed to work out the `ConnectionConfig` and by the time `DbtLoader` is invoked it's too late

sqlmesh/dbt/profile.py
```diff
@@ -101,8 +101,10 @@ def _read_profile(
             target_name = context.render(project_data.get("target"))
 
         if target_name not in outputs:
+            target_names = "\n".join(f"- {name}" for name in outputs)
             raise ConfigError(
-                f"Target '{target_name}' not specified in profiles for '{context.profile_name}'."
+                f"Target '{target_name}' not specified in profiles for '{context.profile_name}'. "
+                f"The valid target names for this profile are:\n{target_names}"
```
erindru, line 107: I noticed that the dbt cli lists valid targets if you get the profile correct but the target wrong, so I updated this to do the same

PR #5177, opened 2025-08-18
Feat(dbt): Add support for selected resources context variable

themisvaltinos:
This aims to add support for [selected resources](https://docs.getdbt.com/reference/dbt-jinja-functions/selected_resources) in dbt projects. When accessed in jinja this would return they keys which correspond to the dbt nodes objects (example format: `model.project_name.model_name`), but that have been selected for the current plan or run using the sqlmesh [selectors](https://sqlmesh.readthedocs.io/en/latest/guides/model_selection/?h=model+sele).

### DIFF
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -199,7 +199,7 @@ class _Node(PydanticModel):
     interval_unit_: t.Optional[IntervalUnit] = Field(alias="interval_unit", default=None)
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
-    node_name: t.Optional[str] = None  # dbt node name
+    dbt_name: t.Optional[str] = None  # dbt node name
     _path: t.Optional[Path] = None
     _data_hash: t.Optional[str] = None
     _metadata_hash: t.Optional[str] = None
--- a/sqlmesh/core/node.py
+++ b/sqlmesh/core/node.py
@@ -199,6 +199,7 @@ class _Node(PydanticModel):
     interval_unit_: t.Optional[IntervalUnit] = Field(alias="interval_unit", default=None)
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
+    node_name: t.Optional[str] = None  # dbt node name
     _path: t.Optional[Path] = None
     _data_hash: t.Optional[str] = None
     _metadata_hash: t.Optional[str] = None
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -70,6 +70,8 @@ class Plan(PydanticModel, frozen=True):
     execution_time_: t.Optional[TimeLike] = Field(default=None, alias="execution_time")
 
     user_provided_flags: t.Optional[t.Dict[str, UserProvidedFlags]] = None
+    selected_models: t.Optional[t.Set[str]] = None
+    """Models that have been selected for this plan (used for dbt selected_resouces)"""
 
     @cached_property
     def start(self) -> TimeLike:
@@ -282,6 +284,7 @@ class Plan(PydanticModel, frozen=True):
             },
             environment_statements=self.context_diff.environment_statements,
             user_provided_flags=self.user_provided_flags,
+            selected_models=self.selected_models,
         )
 
     @cached_property
@@ -319,6 +322,7 @@ class EvaluatablePlan(PydanticModel):
     disabled_restatement_models: t.Set[str]
     environment_statements: t.Optional[t.List[EnvironmentStatements]] = None
     user_provided_flags: t.Optional[t.Dict[str, UserProvidedFlags]] = None
+    selected_models: t.Optional[t.Set[str]] = None
 
     def is_selected_for_backfill(self, model_fqn: str) -> bool:
         return self.models_to_backfill is None or model_fqn in self.models_to_backfill
--- a/sqlmesh/core/plan/definition.py
+++ b/sqlmesh/core/plan/definition.py
@@ -71,7 +71,7 @@ class Plan(PydanticModel, frozen=True):
 
     user_provided_flags: t.Optional[t.Dict[str, UserProvidedFlags]] = None
     selected_models: t.Optional[t.Set[str]] = None
-    """Models that have been selected for this plan (used for dbt selected_resouces)"""
+    """Models that have been selected for this plan (used for dbt selected_resources)"""
 
     @cached_property
     def start(self) -> TimeLike:
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -507,5 +507,6 @@ class RuntimeAdapter(BaseAdapter):
         return normalized_table
 
     def _dbt_model_id(self, sqlmesh_model_name: str) -> str:
+        # Model prefix is needed to correspond to the key in the nodes within the dbt context variable
         parts = [part.strip('"') for part in sqlmesh_model_name.split(".")]
         return f"model.{parts[0]}.{parts[-1]}"
--- a/sqlmesh/dbt/adapter.py
+++ b/sqlmesh/dbt/adapter.py
@@ -181,10 +181,6 @@ class BaseAdapter(abc.ABC):
             }
         )
 
-    @property
-    def selected_resources(self) -> t.List[str]:
-        return []
-
 
 class ParsetimeAdapter(BaseAdapter):
     def get_relation(self, database: str, schema: str, identifier: str) -> t.Optional[BaseRelation]:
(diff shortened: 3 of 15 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -478,3 +479,33 @@ def _relation_info_to_relation(
         }
     )
     return relation_type.create(**relation_info, quote_policy=quote_policy)
+
+
+_selected_resources: t.List[str] = []
```
erindru: I'm not sure populating a global variable from within `Context.plan` is the way to go here.

The "builtin globals" get amended with extra objects / functions at various points (generally whenever there is some kind of render call), would we not be better to make the selected model list a first class property of the Plan?

themisvaltinos: yes you're right, I tried to follow what dbt was doing but it doesn't seem like the best idea maybe a propery of plan would be better

sqlmesh/core/plan/definition.py
```diff
@@ -70,6 +70,8 @@ class Plan(PydanticModel, frozen=True):
     execution_time_: t.Optional[TimeLike] = Field(default=None, alias="execution_time")
 
     user_provided_flags: t.Optional[t.Dict[str, UserProvidedFlags]] = None
+    selected_models: t.Optional[t.Set[str]] = None
+    """Models that have been selected for this plan (used for dbt selected_resouces)"""
```
erindru: nit: `resouces`

sqlmesh/dbt/adapter.py
```diff
@@ -504,3 +515,7 @@ def _normalize(self, input_table: exp.Table) -> exp.Table:
             normalized_table.set("db", normalized_table.this)
             normalized_table.set("this", None)
         return normalized_table
+
+    def _dbt_model_id(self, sqlmesh_model_name: str) -> str:
+        parts = [part.strip('"') for part in sqlmesh_model_name.split(".")]
+        return f"model.{parts[0]}.{parts[-1]}"
```
erindru: I had to check the [dbt docs](https://docs.getdbt.com/reference/dbt-jinja-functions/selected_resources) to understand why the hardcoded `model.` prefix is necessary, might be worth leaving a comment.

Also, `sqlmesh_model_name` is coming through as a quoted fqn, right? And this is on the `RuntimeAdapter` which has the snapshots available on `self.snapshots`?

Rather than string munging, is it possible to look up the matching snapshot and grab the original model name, something like:
`f"model.{self.snapshots[sqlmesh_model_name].node.name}"` ?

themisvaltinos:
> Also I guess the seed. and test. nodes are out of scope for this PR? 

yes good point so mainly the goal of this pr was to match the `dbt run` for models and there were a few things to discuss on how to tackle this topic. 

So dbt if im not mistaken never runs tests as part of the run but as part of [dbt test](https://docs.getdbt.com/reference/commands/test). 

Also for [dbt seed](https://docs.getdbt.com/reference/commands/seed) as well, while for sqlmesh seed models run as part of the sqlmesh plan and run commands and not sqlmesh_dbt run as well. 

So since they are models they would appear in the selected resources during the run command but as models rather than seeds:
```
model.jaffle_shop.customers
model.jaffle_shop.orders
model.jaffle_shop.raw_customers
model.jaffle_shop.raw_orders
model.jaffle_shop.raw_payments
model.jaffle_shop.stg_customers
model.jaffle_shop.stg_orders
model.jaffle_shop.stg_payments
```
This I feel provides information to the user that these have been selected to be run. 
In dbt you would have to first run dbt seed:
```
07:57:40  seed.jaffle_shop.raw_customers
07:57:40  seed.jaffle_shop.raw_orders
07:57:40  seed.jaffle_shop.raw_payments
```
and then dbt run:
```
08:25:42  model.jaffle_shop.stg_payments
08:25:42  model.jaffle_shop.customers
08:25:42  model.jaffle_shop.orders
08:25:42  model.jaffle_shop.stg_customers
08:25:42  model.jaffle_shop.stg_orders
```
So it is a bit tough to map this in exactly the same 1-1 way with dbt. I guess I could map these instead of model as seed to keep the same keys though that would be a bit inconsistent with sqlmesh.

Now regarding dbt tests, these are our audits which do run as part of the sqlmesh plan, my thinking was for now not include them to match more the dbt output when they use selectors and see the models. But in hindsight and to your point I wonder if it makes sense to include and have in the list of models and tests (mapped back from our audits) that have been run. Does that feel more intuitive to you?

sqlmesh/dbt/adapter.py
themisvaltinos: yes didn't like the string munging either don't know if there was a cleaner way to do this. we don't keep snapshots in the RuntimeAdapter just use it for the table mapping but we could save it or use it during init similar to table mapping and keep a mapping of snapshot names to dbt ones. but the issue is this `{self.snapshots[sqlmesh_model_name].node.name}` wouldn't work in the way that it would return `main.orders` for example rather than `jaffle_shop.orders` for `'"jaffle_shop"."main"."orders"'`. I feel you encountered a similar issue with the selectors. the reason I want to have the dbt way and not sqlmesh btw is they might be used in macros as a key to the graph which would contain the dbt way in the nodes dictionary so `model - package - model name` rather than schema

erindru: >  we don't keep snapshots in the RuntimeAdapter just use it for the table mapping

Ah, gotcha, this trips me up all the time. Yes you're right, it appears the snapshots currently limited to whatever is referenced by the model query, not all the snapshots available in the environment.

> the reason I want to have the dbt way and not sqlmesh btw is they might be used in macros as a key to the graph 

yes, absolutely, this needs to return the dbt model name and not the sqlmesh model name. The thing is, I don't know if we can accurately determine the dbt model name based on munging the SQLMesh fqn.

Although I guess all the dbt models do live in a global namespace and schemas are an implementation detail, so maybe this isn't actually the problem I think it is. In the absence of a clear better solution, let's run with this and see how far it gets us :+1:

erindru (approved):
> So dbt if im not mistaken never runs tests as part of the run but as part of dbt test
> Also for dbt seed as well

Oh, sorry, of course :facepalm: . It didnt occur to me that `dbt list` would have everything but `dbt run`, `dbt test` and `dbt seed` would just have their own relevant subsets, but it's very obvious when you mention it.

Ok, this LGTM then :)

sqlmesh/dbt/adapter.py
themisvaltinos: > The thing is, I don't know if we can accurately determine the dbt model name based on munging the SQLMesh fqn. Although I guess all the dbt models do live in a global namespace and schemas are an implementation detail, so maybe this isn't actually the problem I think it is.

actually it is the problem you think it is as I run more examples this solution falls short for example in Postgres where the package might not be the db name.. I'm looking for a better way to map the sqlmesh model names to the dbt ones

themisvaltinos: @erindru  so I tried different ways but all got quite complicated and without having perfect robustness. so settled in getting it for the snapshot itself (which does simplify things a bunch as all these helper methods are removed) as you originally suggested:

> Rather than string munging, is it possible to look up the matching snapshot and grab the original model name

but since we don't have the dbt name in there, had to add it in the node. I remember you also needed this dbt model name somewhere for a mapping, unless I'm mistaken, so if you can take a final look to see how this looks

sqlmesh/core/node.py
```diff
@@ -199,6 +199,7 @@ class _Node(PydanticModel):
     interval_unit_: t.Optional[IntervalUnit] = Field(alias="interval_unit", default=None)
     tags: t.List[str] = []
     stamp: t.Optional[str] = None
+    node_name: t.Optional[str] = None  # dbt node name
```
erindru: Should we just call this `dbt_node_name` instead? So it's clear to all internal code that this relates to dbt, is only populated in dbt mode and should not be confused with `Snapshot.node.name`.

Or even just `dbt_name` next to `name` since this is already on `_Node`

sqlmesh/dbt/adapter.py
erindru: Nice work, I didnt realise we already had a property `BaseModelConfig.node_name` to produce this value.

Yes, adding it to the main `_Node` abstraction is probably the way to go here because it populates it early on and  makes it accessible from everywhere.

I think this will provide a nice way to leverage this value in the selector engine too

sqlmesh/core/node.py
themisvaltinos: good point this will be much clearer revised it to `dbt_name`

erindru: LGTM, feel free to merge when ready

PR #5188, opened 2025-08-19
Feat(dbt): Add support for adapter.rename_relation

vchan:


sqlmesh/dbt/adapter.py
```diff
@@ -349,6 +355,18 @@ def drop_relation(self, relation: BaseRelation) -> None:
...
+        old_table_name = None
+        if from_relation.schema is not None and from_relation.identifier is not None:
+            old_table_name = self._normalize(self._relation_to_table(from_relation))
+
+        new_table_name = None
+        if to_relation.schema is not None and to_relation.identifier is not None:
+            new_table_name = self._normalize(self._relation_to_table(to_relation))
+
+        if old_table_name and new_table_name:
+            self.engine_adapter.rename_table(old_table_name, new_table_name)
```
erindru: The Copilot suggestion although horrendous was right in principle - how does the user know one of their inputs isnt right? This looks like it would silently do nothing if the user supplied two tables without qualifying them with schemas, rather than warn that the `rename_relation` operation didn't do anything

If that's the dbt behaviour as well then no problem I guess

vchan: I agree with Copilot and you. Was just about to update the PR.

PR #5189, opened 2025-08-19
Feat!: Create physical tables as part of evaluation

izeigerman:
This update eliminates the need for a separate "creating physical tables" step in the plan application. The tables are now created as part of the model evaluation DAG.

This has the following benefits:
- Fewer queries submitted to the target engine, since the table creation is coupled with model evaluation in a single `CREATE OR REPLACE` call
- Better alignment with the execution model of dbt projects which assumes that for each model in the DAG all upstream model both created tables **and** inserted data into them
- Only the necessary tables are created. Previously, SQLMesh would preemptively create both the dev and non-dev tables for certain model snapshots. These tables are now created lazily, as needed

Note, that the "creating physical tables" step will still show up if the plan was applied with either the `--empty-backfill` or  the `---skip-backfill` flag, since there are no evaluations to perform. 

Breaking changes:
- Macros relying on `@IF(@runtime_stage = 'creating', ...)` will run every time a table is created for the first time even if data is inserted as part of the creation process (i.e. `CREATE OR REPLACE TABLE...`). Similarly, macros that rely on the `evaluating` runtime stage will no longer run as part of the first evaluation that also creates the table for the first time. Previously, there was a clear distinction between empty table creation (`creating` stage) and data insertion (`evaluating` stage). Since the two are now coupled, we had to relax the definition of the `creating` runtime stage to preserve some level of backwards compatibility.

### DIFF
--- a/examples/sushi/config.py
+++ b/examples/sushi/config.py
@@ -1,6 +1,6 @@
 import os
 
-from sqlmesh.core.config.common import VirtualEnvironmentMode
+from sqlmesh.core.config.common import VirtualEnvironmentMode, TableNamingConvention
 from sqlmesh.core.config import (
     AutoCategorizationMode,
     BigQueryConnectionConfig,
@@ -193,6 +193,10 @@ environment_catalog_mapping_config = Config(
     before_all=before_all,
 )
 
-hash_md5_naming_config = config.copy(update={"physical_table_naming_convention": "hash_md5"})
+hash_md5_naming_config = config.copy(
+    update={"physical_table_naming_convention": TableNamingConvention.HASH_MD5}
+)
 
-table_only_naming_config = config.copy(update={"physical_table_naming_convention": "table_only"})
+table_only_naming_config = config.copy(
+    update={"physical_table_naming_convention": TableNamingConvention.TABLE_ONLY}
+)


examples/sushi/config.py
```diff
@@ -177,4 +190,9 @@
         "^prod$": "prod_catalog",
         ".*": "dev_catalog",
     },
+    before_all=before_all,
 )
+
+hash_md5_naming_config = config.copy(update={"physical_table_naming_convention": "hash_md5"})
```
erindru: Nit: should these be using the enum members?

sqlmesh/core/engine_adapter/base.py
```diff
@@ -472,6 +478,7 @@ def replace_query(
                 **kwargs,
             )
         if self_referencing:
+            assert target_columns_to_types is not None
```
erindru, line 481: Should this be an explicit check?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -923,9 +1012,10 @@ def _migrate_snapshot(
         deployability_index = DeployabilityIndex.all_deployable()
         render_kwargs: t.Dict[str, t.Any] = dict(
             engine_adapter=adapter,
-            snapshots=parent_snapshots_by_name(snapshot, snapshots),
+            snapshots=snapshots,
             runtime_stage=RuntimeStage.CREATING,
             deployability_index=deployability_index,
+            execution_time=now(),
```
erindru: Should this be the plan execution time?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1296,19 +1302,30 @@ def _create_catalogs(
...
+        gateway_table_pairs: t.Iterable[t.Tuple[t.Optional[str], t.Union[exp.Table, str]]],
     ) -> None:
-        table_exprs = [exp.to_table(t) for t in tables]
-        unique_schemas = {(t.args["db"], t.args.get("catalog")) for t in table_exprs if t and t.db}
-        # Create schemas sequentially, since some engines (eg. Postgres) may not support concurrent creation
-        # of schemas with the same name.
-        for schema_name, catalog in unique_schemas:
+        table_exprs = [(gateway, exp.to_table(t)) for gateway, t in gateway_table_pairs]
+        unique_schemas = {
+            (gateway, t.args["db"], t.args.get("catalog"))
```
erindru, line 1313: Nit: aren't these available as `t.db` and `t.catalog`?

sqlmesh/core/scheduler.py
```diff
@@ -55,9 +55,46 @@
...
+    def __lt__(self, other: BaseNode) -> bool:
+        return (self.__class__.__name__, self.snapshot_name) < (
+            other.__class__.__name__,
+            other.snapshot_name,
+        )
+
+
+@dataclass(frozen=True)
+class EvaluateNode(BaseNode):
+    snapshot_name: str
```
eakmanrq, line 73: This is already defined in `BaseNode`. Same applies for `CreateNode` and `Dummy Node`.

sqlmesh/core/scheduler.py
```diff
@@ -55,9 +55,46 @@
...
+class CreateNode(BaseNode):
+    snapshot_name: str
+
+
+@dataclass(frozen=True)
+class DummyNode(BaseNode):
+    snapshot_name: str
+
+
+SchedulingUnit = t.Union[EvaluateNode, CreateNode, DummyNode]
```
eakmanrq: Why not refer to `BaseNode` instead of creating this new `SchedulingUnit` type?

izeigerman: fixed

sqlmesh/core/scheduler.py
```diff
@@ -522,46 +601,72 @@ def _dag(self, batches: SnapshotToIntervals) -> DAG[SchedulingUnit]:
...
                 if batch_concurrency and i >= batch_concurrency:
                     batch_idx_to_wait_for = i - batch_concurrency
                     dag.add(
                         node,
                         [
-                            (
-                                snapshot.name,
-                                (intervals[batch_idx_to_wait_for], batch_idx_to_wait_for),
-                            )
+                            EvaluateNode(
```
erindru, line 665: Making these implicit data structures explicit really helps when trying to understand what is going on

erindru (approved):
Looks good to me! I guess we will find out soon how many projects were relying on create first -> then evaluate

sqlmesh/core/engine_adapter/base.py
georgesittas: Just sharing [this](https://github.com/TobikoData/sqlmesh/pull/5153#discussion_r2275251240) thread for some more context.

sqlmesh/core/snapshot/evaluator.py
georgesittas: They are, but the property accessors return strings, whereas looking them up through `args` results in `Expression` instances. Given we used to do this:

```python
        unique_schemas = {(t.args["db"], t.args.get("catalog")) for t in table_exprs if t and t.db}
```

I'm assuming the `Expression` instances are necessary to preserve identifier quotes, etc.

sqlmesh/core/engine_adapter/base.py
izeigerman: We already have an explicit check for this on line 455. This is our 2nd entry into the `self_referencing` check so at this point we indeed assert that `target_columns_to_types` is not None.

sqlmesh/core/snapshot/evaluator.py
izeigerman: IMO it doesn't really matter here since we use to create a temporary empty table to source the schema from to perform the migration. The actual value doesn't really matter here.

izeigerman: Actually, there's no point in providing the value explicitly here. The renderer will automatically insert an appropriate default value.

sqlmesh/core/scheduler.py
```diff
@@ -55,9 +56,43 @@
 
 logger = logging.getLogger(__name__)
 SnapshotToIntervals = t.Dict[Snapshot, Intervals]
-# we store snapshot name instead of snapshots/snapshotids because pydantic
-# is extremely slow to hash. snapshot names should be unique within a dag run
-SchedulingUnit = t.Tuple[str, t.Tuple[Interval, int]]
+
+
+class SchedulingUnit(abc.ABC):
+    snapshot_name: str
```
eakmanrq, line 62: If you make this a dataclass then you don't have to redefine `snapshot_name` in the classes that inherit from this.

eakmanrq: I think when we talked you said you wanted this to be more of an interface but I don't really get the advantage we get from this framing. Either way works though.

izeigerman: I believe that a dataclass is a product type and we should be able to instantiate it. This directly conflicts with ABC, which is supposed to provide an interface and can't be instantiated. That's why I'm against mixing the two.

PR #5198, opened 2025-08-20
Fix snowflake int tests

izeigerman:


tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -2847,16 +2854,12 @@ def _run_plan(sqlmesh_context: Context, environment: str = None) -> PlanResults:
     assert plan_1.snapshot_for(model_a).model.view_name in plan_1.schema_metadata.views
     assert plan_1.snapshot_for(model_b).model.view_name in plan_1.schema_metadata.views
 
-    assert len(plan_1.internal_schema_metadata.tables) == 3
+    assert len(plan_1.internal_schema_metadata.tables) == 1
 
     assert plan_1.table_name_for(model_a) in plan_1.internal_schema_metadata.tables
-    assert plan_1.dev_table_name_for(model_a) in plan_1.internal_schema_metadata.tables
```
erindru, line 2889: I think this is correct if dev previews are no longer created "just in case" and are only created when needed

erindru (approved):

PR #5203, opened 2025-08-21
feat: mimic dbt nuanced on_schema_change behavior

eakmanrq:
Addresses the gaps outlined in this PR: https://github.com/TobikoData/sqlmesh/pull/5193

Key changes:
* Allowed ConnectionConfig to be able to override the schema differ behavior. The dbt adapter then uses this to configure the schema differ to behave in a way that is compatible with dbt runtime. 
* SchemaDiffer can now be configured to ignore nested operations in order to mimic dbt behavior
* SchemaDiffer can now be configured to configure alter operations to change column data type as "destructive" instead of "additive" in order to mimic dbt behavior

All of these changes are to support dbt adapter - there is not a good use case for them for native users. 

The biggest gap remaining to support dbt `on_schema_change` is robust integration tests comparing dbt to dbt adapter to ensure they behave the same. Since this is runtime related, and we have loading errors to address, will prioritize this later as needed.


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -68,8 +68,8 @@ class BigQueryEngineAdapter(InsertOverwriteWithMergeMixin, ClusteredByMixin, Row
     MAX_COLUMN_COMMENT_LENGTH = 1024
     SUPPORTED_DROP_CASCADE_OBJECT_KINDS = ["SCHEMA"]
 
-    SCHEMA_DIFFER = SchemaDiffer(
-        compatible_types={
+    SCHEMA_DIFFER_KWARGS = {
+        "compatible_types": {
```
erindru, line 72: nit (and I guess this is also a matter of opinion): did you consider using the `dict()` constructor so you can keep treating the kwarg names as identifiers instead of strings?

sqlmesh/core/schema_diff.py
```diff
@@ -278,6 +281,29 @@ def column_position_node(self) -> t.Optional[exp.ColumnPosition]:
         return exp.ColumnPosition(this=column, position=position)
 
 
+class NestedSupport(str, Enum):
+    ALL = "ALL"
```
erindru, line 286: Is there any chance you can document what each of these enum variants is intended to do?

for example, like what we have for [CatalogSupport](https://github.com/TobikoData/sqlmesh/blob/86140b2a1a81c78ab9dac080b3b48f8f4669c0da/sqlmesh/core/engine_adapter/shared.py#L182) and [CommentCreationView](https://github.com/TobikoData/sqlmesh/blob/86140b2a1a81c78ab9dac080b3b48f8f4669c0da/sqlmesh/core/engine_adapter/shared.py#L109)

erindru (approved):

sqlmesh/core/engine_adapter/bigquery.py
eakmanrq: I don't have a preference either way. Just leaving how it is to save some time.

PR #5205, opened 2025-08-21
Feat(dbt_cli): Add support for `--vars`

erindru:
Add support for the `--vars '<yaml blob>'` syntax of the dbt core cli.

It passes these in `config_loader_kwargs` so the variables are available before the project is loaded


tests/dbt/cli/test_list.py
```diff
@@ -15,3 +15,17 @@ def test_list(jaffle_shop_duckdb: Path, invoke_cli: t.Callable[..., Result]):
...
     assert "main.stg_payments" in result.output
+
+
+def test_list_with_vars(jaffle_shop_duckdb: Path, invoke_cli: t.Callable[..., Result]):
+    (jaffle_shop_duckdb / "models" / "aliased_model.sql").write_text("""
+    {{ config(alias='model_' + var('foo')) }}                                                              
+    select 1
+    """)
+
+    result = invoke_cli(["list", "--vars", "foo: bar"])
```
izeigerman, line 57: Do we need to test more variances like:
```
$ dbt run --vars '{"key": "value", "date": 20180101}'
$ dbt run --vars '{key: value, date: 20180101}'
```
or are we confident that the YAML parser takes care of it?

sqlmesh_dbt/cli.py
```diff
@@ -4,12 +4,31 @@
...
 
 
-def _get_dbt_operations(ctx: click.Context) -> DbtOperations:
-    if not isinstance(ctx.obj, DbtOperations):
+def _get_dbt_operations(ctx: click.Context, vars: t.Optional[t.Dict[str, t.Any]]) -> DbtOperations:
+    if not isinstance(ctx.obj, functools.partial):
         raise ValueError(f"Unexpected click context object: {type(ctx.obj)}")
-    return ctx.obj
+
+    dbt_operations = ctx.obj(vars=vars)
```
izeigerman, line 15: Why do we handle this differently from `profile` / target`?

erindru: `--profile` / `--target` are top level dbt options - they show in `dbt --help` and apply to every subcommand.

`--vars` only applies to certain subcommands. eg it shows in `dbt list --help` and `dbt run --help` but not `dbt source --help`

tests/dbt/cli/test_list.py
erindru: The YAML parser does indeed take care of it, but I added a test `test_yaml_param_type()` to prove it

PR #5207, opened 2025-08-22
feat: dbt adapter allow invalid ref for tests

eakmanrq:
Dbt ignores with a warning tests that have refs to models that don't exist. This mimics that behavior. 

dbt function that explicitly does not fail if test: https://github.com/dbt-labs/dbt-core/blob/0836095a572532fc22b1c0981346eebe56f7d1fb/core/dbt/parser/manifest.py#L1324-L1332

I noticed in testing that constraints that make invalid refs also do not affect execution so I included that in the test although I didn't have to make any code changes at this time to support it.

### DIFF
--- a/sqlmesh/dbt/context.py
+++ b/sqlmesh/dbt/context.py
@@ -265,8 +265,7 @@ class DbtContext:
                 else:
                     models[ref] = t.cast(ModelConfig, model)
             else:
-                exception = MissingModelError(f"Model '{ref}' was not found.")
-                exception.model_name = ref
+                exception = MissingModelError(ref)
                 raise exception
 
         for source in dependencies.sources:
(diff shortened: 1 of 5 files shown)


sqlmesh/dbt/context.py
```diff
@@ -265,7 +265,9 @@ def context_for_dependencies(self, dependencies: Dependencies) -> DbtContext:
                 else:
                     models[ref] = t.cast(ModelConfig, model)
             else:
-                raise ConfigError(f"Model '{ref}' was not found.")
+                exception = MissingModelError(f"Model '{ref}' was not found.")
```
erindru: Any reason you didnt follow the existing pattern with `ApiError`, `NodeAuditsErrors` etc and make `model_name` part of the constructor?

eakmanrq: I just didn't think about of it for some reason and yeah that is a better approach. Updating.

izeigerman (approved):

PR #5208, opened 2025-08-22
Feat(dbt): Default to `virtual_environment_mode: dev_only` on init

erindru:


sqlmesh_dbt/operations.py
```diff
@@ -34,7 +34,6 @@ def run(self, select: t.Optional[str] = None, full_refresh: bool = False) -> Non
 
         self.context.plan(
             select_models=select_models,
-            no_auto_categorization=True,  # everything is breaking / foward-only
```
erindru, line 37: This was causing a test to fail / throw up a `Enter the effective date (eg. '1 year', '2020-01-01') to apply forward-only changes retroactively` message, even though `no_prompts=True`.

The `sqlmesh_dbt` plan args arent quite right anyway, so ive removed it for now

izeigerman (approved):

PR #5209, opened 2025-08-22
Fix(tests): Address some test flakiness

erindru:
I notice the Fabric tests failing on main with:
```
FAILED tests/core/engine_adapter/integration/test_integration.py::test_janitor[fabric-catalog] -
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.fabric.microsoft.com/v1/workspaces/**/warehouses
```

The cause was essentially the create warehouse API calls failing if the warehouse already existed.

SQLMesh code generally expects `CREATE IF NOT EXISTS` / `DROP IF EXISTS` semantics, so I implemented these.

In addition, some other tests are starting to become more flaky with issues like:
```
teardown failed on attempt 2! Exiting immediately!
        KeyError: <_pytest.stash.StashKey object at 0x7f1666072320>
```

and

```
duckdb.duckdb.IOException: IO Error: Could not set lock on file "testing.duckdb": Conflicting lock is held in /home/circleci/.pyenv/versions/3.12.11/bin/python3.12 (PID 3249).
```

so I had a go at addressing these too. The general theme is we run tests concurrently for speed but aren't always good at ensuring each test gets its own unique copy of things and can work in isolation from other tests

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -310,10 +310,10 @@ workflows:
                 - athena                
                 - fabric
                 - gcp-postgres
-          #filters:
-          #  branches:
-          #    only:
-          #      - main
+          filters:
+            branches:
+              only:
+                - main
       - ui_style
       - ui_test
       - vscode_test


.circleci/continue_config.yml
```diff
@@ -302,18 +302,18 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-                - athena                
+                #- snowflake
```
erindru: TODO: revert prior to merge

tests/core/engine_adapter/integration/config.yaml
```diff
@@ -5,7 +5,7 @@ gateways:
       type: duckdb
       catalogs:
         memory: ':memory:'
-        testing: 'testing.duckdb'
+        testing: "{{ var('tmp_path') }}/testing.duckdb"
```
erindru, line 8: This was causing flakiness in `style_and_cicd_tests`, [example](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/23572/workflows/daa350c4-201b-4a41-9a1c-28b908ac9816/jobs/266017)

```
duckdb.duckdb.IOException: IO Error: Could not set lock on file "testing.duckdb": Conflicting lock is held in /home/circleci/.pyenv/versions/3.12.11/bin/python3.12 (PID 3249). See also https://duckdb.org/docs/stable/connect/concurrency
```

The `duckdb` integration tests get run as part of that and without prefixing the path, `testing.duckdb` gets created in the sqlmesh root dir rather than the unique dir for each test.

This causes it to be re-used between tests and potentially accessed in parallel from multiple workers although there is an `xdist_group` that forces most of the tests to run sequentially

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -1936,49 +1934,16 @@ def test_transaction(ctx: TestContext):
...
             "Sushi end-to-end tests only need to run once for Athena because sushi needs a hybrid of both Hive and Iceberg"
         )
 
-    tmp_path = tmp_path_factory.mktemp(f"sushi_{ctx.test_id}")
-
     sushi_test_schema = ctx.add_test_suffix("sushi")
     sushi_state_schema = ctx.add_test_suffix("sushi_state")
     raw_test_schema = ctx.add_test_suffix("raw")
 
-    config = load_config_from_paths(
```
erindru, line 1951: This was duplicating logic already in `TestContext`, since it pre-dated `TestContext.create_context()`

tests/core/engine_adapter/integration/test_integration.py
```diff
@@ -2383,33 +2362,15 @@ def _normalize_snowflake(name: str, prefix_regex: str = "(sqlmesh__)(.*)"):
 
     init_example_project(tmp_path, ctx.engine_type, schema_name=schema_name)
 
-    config = load_config_from_paths(
```
erindru, line 2386: This was duplicating logic already in `TestContext`, since it pre-dated `TestContext.create_context()`

tests/core/engine_adapter/integration/conftest.py
```diff
@@ -27,15 +27,15 @@
 logger = logging.getLogger(__name__)
 
 
-@pytest.fixture(scope="session")
```
erindru, line 30: This was session-scoped because its predecessor was session scoped and the session scope got retained through an earlier refactor.

However, I was encountering some very hard-to-pin-down issues that I have seen a bunch of times regarding `StashKey`, [example](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/23572/workflows/daa350c4-201b-4a41-9a1c-28b908ac9816/jobs/266033):

```
teardown failed on attempt 2! Exiting immediately!
          File "/home/********/.pyenv/versions/3.12.11/lib/python3.12/site-packages/_pytest/runner.py", line 344, in from_call
            result: TResult | None = func()
                                     ^^^^^^
        KeyError: <_pytest.stash.StashKey object at 0x7f1666072320>
```

That looks like a concurrency issue and I had a theory it was related to session and function scoped fixtures being mixed.

So i've made this function scoped because I couldnt see a strong reason for it to be session scoped

tests/conftest.py
```diff
@@ -212,6 +212,38 @@ def pytest_collection_modifyitems(items, *args, **kwargs):
             item.add_marker("fast")
 
 
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
+def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo):
```
erindru, line 216: This seems to be the only hook that can catch the `StashKey` errors. Fixtures that `yield` their values (like `tmp_path`) appear to hit different codepaths than other types of fixtures.

I tried both `pytest_fixture_post_finalizer` and `pytest_runtest_teardown` before resorting to this

.circleci/manage-test-db.sh
```diff
@@ -51,7 +51,7 @@ databricks_init() {
 
     # Note: the cluster doesnt need to be running to create / drop catalogs, but it does need to be running to run the integration tests
     echo "Ensuring cluster is running"
-    databricks clusters start $CLUSTER_ID || true
+    databricks clusters start $CLUSTER_ID
```
erindru, line 54: This is to make Databricks fail early with the cluster start error rather than time out for 40mins trying to run tests

georgesittas (approved):
Nice work 👍

PR #5226, opened 2025-08-26
Feat(dbt_cli): Set proper plan flags and also allow --empty and --environment

erindru:
This PR:
 - adds the `--empty` dbt option which maps to SQLMesh `--skip-backfill` (creates tables but doesnt write data to them).
 - adds the `--environment` option to allow creating development VDE's. This is not a native dbt option but is a value-add of the SQLMesh engine.

In addition, this PR tightens up the plan flags to make SQLMesh more closely mimic dbt behaviour:

- All plans have `--run` and `--ignore_cron` set - except dev plans with no `--select`.
  This mimics the dbt behaviour of "always run up until right now" while falling back to the default SQLMesh behaviour for dev plans where the user has no interest in a particular set of models
- Dev plans set `include_unmodifed=False` to create dev environments with just the changes, `create_from=prod,always_recreate_environment=True` to ensure that dev environments are always based off the latest prod at the time the command is run and `enable_preview=None` to use default behaviour for dev previews.

Caveats:
 -  `enable_preview=True` is deliberately set even though this will cause changed dbt incremental models to be fully refreshed in *dev* environments. The assumption is that if the user is opting in to VDE, and this is a problem, then they may also consider opting in to SQLMesh incremental models
 -  `--empty` in dbt replaces existing tables with empty ones, but `--skip-backfill` just skips the backfill step and wont touch a table if it didnt already need to be backfilled. Therefore, this is a partial implementation of `--empty` until we have everything in place to enable `--full-refresh` semantics

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -260,9 +260,7 @@ def test_plan_skip_backfill(runner, tmp_path, flag):
     # plan for `prod` errors if `--skip-backfill` is passed without --no-gaps
     result = runner.invoke(cli, ["--log-file-dir", tmp_path, "--paths", tmp_path, "plan", flag])
     assert result.exit_code == 1
-    assert (
-        "Skipping the backfill stage for production can lead to unexpected results" in result.output
-    )
+    assert "Skipping the backfill stage for production can lead to unexpected" in result.output
 
     # plan executes virtual update without executing model batches
     # Input: `y` to perform virtual update


sqlmesh_dbt/cli.py
```diff
@@ -81,7 +81,15 @@ def dbt(
     "--full-refresh",
     help="If specified, dbt will drop incremental models and fully-recalculate the incremental table from the model definition.",
 )
+@click.option(
+    "--environment",
```
izeigerman, line 97: should we add `--env` shortcut? or maybe make it an argument instead of an option? like we do for a plan? feel intuitive. Plus it doesn't look like `dbt run` has any arguments.

sqlmesh_dbt/operations.py
```diff
@@ -66,6 +66,90 @@ def _selected_models(
...
+        if empty:
+            # dbt --empty adds LIMIT 0 to the queries, resulting in empty tables
+            # this lines up with --skip-backfill in SQLMesh
+            options["skip_backfill"] = True
+
+            if is_prod:
+                # to prevent the following error:
+                # > ConfigError: When targeting the production environment either the backfill should not be skipped or
+                # > the lack of data gaps should be enforced (--no-gaps flag).
+                options["no_gaps"] = True
```
izeigerman: I don't know if we should even do this.

erindru: I couldnt get `--skip-backfill` to work without it in my example project?

sqlmesh_dbt/cli.py
erindru: I think it would feel intuitive for a sqlmesh user, but people used to running `dbt run` are used to specifying a boatload of options, right?

`dbt run --help` reveals that the original command has no arguments, so the established practice appears to be to keep adding options

erindru: I did add `--env` as an alias

sqlmesh_dbt/operations.py
erindru: After internal discussion, i've changed the `ConfigError` to a warning and stopped setting `no_gaps` to get around it.

PR #5231, opened 2025-08-26
Fix!: stop treating dbt schema data types as columns_to_types

georgesittas:
After changing `sushi_dbt` as I did in this PR, the following commands run successfully and I can see `unused_column` in `local.duckdb`:

```
$ dbt -t duckdb seed
$ dbt -t duckdb run
```

Whereas SQLMesh fails to execute the project, because it currently treats the columns in `schema.yaml` as the source of truth, which does not match dbt's behavior:

```
$ sqlmesh plan  # Error: In model 'sushi.top_waiters', a description is provided for column 'waiter_id' but it is not a column in the model.
```

I verified locally (using `dbt-duckdb`) that setting `data_type` in `schema.yml` does not determine the actual DDL type. For example, for a simple model like `SELECT 1 AS c`, the following entry does not result in a `TEXT` type for `c`:

```yaml
models:
  - name: simple_model
    description: "A simple model with a single column"
    columns:
      - name: c
        description: "A constant value"
        data_type: text
```

The `data_type` appears to mostly be relevant when `contract` is `enforced`, in which case it acts as an audit. For example, if I add the following config to the above model:

```yaml
    config:
      contract:
        enforced: true
```

I see this error at runtime (`dbt compile` ran fine):

```
$ dbt run
...
22:52:33    Compilation Error in model simple_model (models/simple_model.sql)
  This model has an enforced contract that failed.
  Please ensure the name, data_type, and number of columns in your contract match the columns in your model's definition.

  | column_name | definition_type | contract_type | mismatch_reason    |
  | ----------- | --------------- | ------------- | ------------------ |
  | c           | INTEGER         | VARCHAR       | data type mismatch |
```

### DIFF
--- a/tests/core/test_integration.py
+++ b/tests/core/test_integration.py
@@ -42,7 +42,7 @@ from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context import Context
 from sqlmesh.core.config.categorizer import CategorizerConfig
 from sqlmesh.core.config.plan import PlanConfig
-from sqlmesh.core.engine_adapter import EngineAdapter
+from sqlmesh.core.engine_adapter import EngineAdapter, DuckDBEngineAdapter
 from sqlmesh.core.environment import EnvironmentNamingInfo
 from sqlmesh.core.macros import macro
 from sqlmesh.core.model import (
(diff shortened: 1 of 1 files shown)


sqlmesh/dbt/basemodel.py
```diff
@@ -349,6 +346,21 @@ def sqlmesh_model_kwargs(
...
+        # Only the `columns_types` config of seed models is actually respected. We don't set the
+        # columns attribute to self.columns intentionally in all other cases, as that could result
+        # in unfaithful types when models are materialized.
+        #
+        # See:
+        # - https://docs.getdbt.com/reference/resource-properties/columns
+        # - https://docs.getdbt.com/reference/resource-configs/contract
+        # - https://docs.getdbt.com/reference/resource-configs/column_types
+        if column_types_override:
+            model_kwargs["columns"] = column_types_to_sqlmesh(column_types_override)
```
izeigerman: shouldn't we pass the dialect? like we used to previously

georgesittas: yep, great catch thanks!

PR #5237, opened 2025-08-27
Fix!: render `{% raw %}...{% endraw %}` correctly

georgesittas:
Docs for reference: https://jinja.palletsprojects.com/en/stable/templates/#escaping

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -12,7 +12,6 @@ from string import Template
 from datetime import datetime, date
 
 import sqlglot
-from jinja2 import Environment
 from sqlglot import Generator, exp, parse_one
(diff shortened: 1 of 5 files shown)


sqlmesh/core/macros.py
```diff
@@ -284,7 +286,7 @@ def evaluate_macros(
                     return exp.to_identifier(text, quoted=node.quoted or None)
             if node.is_string:
                 text = node.this
-                if has_jinja(text):
+                if render_jinja and has_jinja(text):
```
georgesittas: I'm investigating why this is needed. Something looks off– I don't think we should be rendering jinja in the macro evaluator intuitively. I'm guessing some test(s) are using unguarded jinja code, i.e. without `JINJA_*` block tags.

PR #5238, opened 2025-08-27
fix: find variable node if nested

eakmanrq:
See test for example where the variable is nested and we need to find the actual variable reference.

### DIFF
--- a/sqlmesh/utils/jinja.py
+++ b/sqlmesh/utils/jinja.py
@@ -200,6 +200,14 @@ def extract_dbt_adapter_dispatch_targets(jinja_str: str) -> t.List[t.Tuple[str, 
     return extracted
 
 
+def is_variable_node(n: nodes.Node) -> bool:
+    return (
+        isinstance(n, nodes.Call)
+        and isinstance(n.node, nodes.Name)
+        and n.node.name in (c.VAR, c.BLUEPRINT_VAR)
+    )
+
+
 def extract_macro_references_and_variables(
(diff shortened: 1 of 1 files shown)


sqlmesh/utils/jinja.py
```diff
@@ -228,9 +228,24 @@ def extract_macro_references_and_variables(
                         )
                     )
 
+        def is_variable_node(n: nodes.Node) -> bool:
+            return (
+                isinstance(n, nodes.Call)
+                and isinstance(n.node, nodes.Name)
+                and n.node.name in (c.VAR, c.BLUEPRINT_VAR)
+            )
```
georgesittas: I'd extract this helper out instead of making it a closure to reduce the complexity of this function.

sqlmesh/utils/jinja.py
```diff
@@ -228,9 +228,24 @@ def extract_macro_references_and_variables(
...
             if call_name[0] in (c.VAR, c.BLUEPRINT_VAR):
-                assert isinstance(node, nodes.Call)
+                if not is_variable_node(node):
+                    # Find the variable node which could be nested
+                    for n in node.find_all(nodes.Call):
+                        if is_variable_node(n):
+                            node = n
+                            break
+                    else:
+                        raise ValueError(f"Could not find variable name in {jinja_str}")
```
georgesittas, line 248: Should we just push that logic into the helper and call it like `extract_variable_node` or something?

georgesittas (commented):
Just a couple of comments but LGTM otherwise.

PR #5239, opened 2025-08-27
Fix(dbt_cli): Add global error handling group for dbt subcommands

themisvaltinos:
Currently the traceback is printed for the sqlmesh_dbt subcommands, this introduces a custom click group subclass that automatically wraps all subcommand callbacks with `cli_global_error_handler` ensuring error handling across the cli commands like `run`, `list` and in case more are added in the future.

### DIFF
--- a/tests/dbt/cli/test_global_flags.py
+++ b/tests/dbt/cli/test_global_flags.py
@@ -74,7 +74,7 @@ def test_run_error_handler(
     assert "Error: List command error" in result.output
     assert "Traceback" not in result.output
 
-    # test SQLMeshError in main command wuthout subcommand
+    # test SQLMeshError in main command without subcommand
     mock_create = mocker.patch("sqlmesh_dbt.cli.create")
     mock_create.side_effect = SQLMeshError("Failed to load project")
     result = invoke_cli(["--profile", "jaffle_shop"])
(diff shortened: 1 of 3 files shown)


erindru (approved):
This is better, i'll adjust my other PR

tests/dbt/cli/test_global_flags.py
```diff
@@ -28,3 +31,61 @@ def test_profile_and_target(jaffle_shop_duckdb: Path, invoke_cli: t.Callable[...
     result = invoke_cli(["--profile", "jaffle_shop", "--target", "dev"])
     assert result.exit_code == 0
     assert "No command specified" in result.output
+
+
+def test_run_error_handler(jaffle_shop_duckdb: Path, invoke_cli: t.Callable[..., Result]):
+    with patch("sqlmesh_dbt.operations.DbtOperations.run") as mock_run:
```
erindru: oh, actually, shouldnt this use pytest's MockerFixture?

themisvaltinos: right revised to use mockerfixture instead

erindru: thanks. arguably it's less pretty but MockerFixture plays nicely with pytest and helps prevent mocks from clobbering each other between tests

PR #5241, opened 2025-08-27
Chore: Add integration test to verify before all execution order

themisvaltinos:
This adds a test to verify before all and after all order of execution

### DIFF
--- a/tests/core/test_integration.py
+++ b/tests/core/test_integration.py
@@ -42,7 +42,7 @@ from sqlmesh.core.console import Console, get_console
 from sqlmesh.core.context import Context
 from sqlmesh.core.config.categorizer import CategorizerConfig
 from sqlmesh.core.config.plan import PlanConfig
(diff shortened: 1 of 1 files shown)


erindru (approved):

tests/core/test_integration.py
```diff
@@ -6427,6 +6427,75 @@ def test_environment_statements_error_handling(tmp_path: Path):
...
+        state_tables = [
+            "_snapshots",
+            "_environments",
+            "_versions",
+            "_intervals",
+            "_auto_restatements",
+            "_environment_statements",
+            "_plan_dags",
+        ]
+
```
georgesittas, line 6475: Why do we need to ignore those? So that the before all shows up as the first statement?

georgesittas (approved):

tests/core/test_integration.py
erindru: I would assume so, rather than trying to identify it as the 86th statement after all the migration queries have run

PR #5243, opened 2025-08-27
chore(web_common): add more shared components

mykalmax:


web/common/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+import './Button.css'
+
+const DEFAULT_BUTTON_TYPE = 'button'
+
+export interface ButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
+    VariantProps<typeof buttonVariants> {
+  asChild?: boolean
+  type?: ButtonType
+}
```
benfdking: I have been looking at this `Button` and I worry that the structure of it makes it quite difficult to figure out how it works. I think we can do better to make these components easier to consume for each other. 

Here is my thought process. When you use a component, the way you want to understand it is: 
1. What are the props?
2. How does it function?

This is why in past lives I have always structured a component file like. 

```ts
// Imports
import type { ButtonType } from './help'

// Prop of component, named Props so you know it's the props for that component
export interface Props {}

// Component
export const Component = () => {}

// Anything else that makes up the Component
const InternalComponent = () => {}
```

It's not a thing but figuring out how this Button component works requires quite a bit of jumping around files and while this is small, I think the complexity is pretty signficant to figure out for something so small.

web/common/src/components/Button/help.ts
```diff
@@ -0,0 +1,58 @@
+import { cva } from 'class-variance-authority'
+
+import { EnumSize } from '@/types/enums'
+
+export const EnumButtonType = {
+  Button: 'button',
+  Submit: 'submit',
+  Reset: 'reset',
+} as const
```
benfdking: I am going to keep slowly pushing against enums 👅 

https://bluepnume.medium.com/nine-terrible-ways-to-use-typescript-enums-and-one-good-way-f9c7ec68bf15 or even subtypes. 

type ButtonType = "button" | "submit" | "reset" 

verifying switches are exhaustive 

and the usage with just <Button type="button" /> to me is a nicer experience.

web/common/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,157 @@
...
+      control: { type: 'select' },
+      options: Object.values(EnumSize),
+    },
+    type: {
+      control: { type: 'select' },
+      options: ['button', 'submit', 'reset'],
+    },
+    disabled: {
+      control: 'boolean',
+    },
```
benfdking: Do we have to write these? We should be able to set it up so an enum can be inferred?

web/common/src/components/ClipboardCopy/ClipboardCopy.stories.tsx
```diff
@@ -0,0 +1,244 @@
...
+    size: {
+      control: { type: 'select' },
+      options: Object.values(EnumSize),
+    },
+    disabled: {
+      control: 'boolean',
+    },
+    title: {
+      control: 'text',
+    },
```
benfdking: This should all be inferred with a default?

web/common/src/components/ClipboardCopy/ClipboardCopy.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export interface ClipboardCopyProps extends Omit<ButtonProps, 'children'> {
+  text: string
+  delay?: number
+  children: (copied: boolean) => React.ReactNode
+}
+
+export const ClipboardCopy = React.forwardRef<
+  HTMLButtonElement,
+  ClipboardCopyProps
```
benfdking: Nit: Would a shared hook not be a more flexible option for this?

web/common/src/utils/index.ts
```diff
@@ -12,3 +12,45 @@ export function isNil(value: unknown): value is Nil {
...
+  return isString(value) && value.length === 0
+}
+export function isNilOrEmptyString(
+  value: unknown,
+): value is Maybe<EmptyString> {
+  return isNil(value) || isEmptyString(value)
+}
+export function nonEmptyString(value: unknown): value is string {
+  return isString(value) && value.length > 0
+}
```
benfdking: This scares me a little because it implies:
1. that we pass around quite a lot of known. 
2. It ties code together which doesn't need to be tied together, to me this is an unnecessary, it means that for the most simple things like testing a string is empty `s.length === 0` we know have tied our code to this very far reaching `utils` part of the codebase.

An example of this is that if I want to move code around which previously had no tie to anything else because it's just checking whether a string is empty, I now need to think about this very big utils package which I really wouldn't want to do.

web/common/src/components/ClipboardCopy/ClipboardCopy.tsx
mykalmax: so we dont have a use case other than having button , (and i renamed component to be CopyButton) . once we have need to copy text from other interaction i can move to custom hook

PR #5251, opened 2025-08-28
chore: create make command install dbt version

eakmanrq:
Add make command to make it easier to create a local venv for a given dbt version.

Ex: `make install-dev-dbt-1.7` to install dbt `1.7.X` where `X` is the latest patch release. Add `UV=1` before to have it use UV.


themisvaltinos (approved):
running `pnpm -w run fmt` inside the `.github/workflows` folder should fix the formatting issues

PR #5254, opened 2025-08-28
Chore: Allow duplicate keys in dbt project yaml files

vchan:
dbt allows duplicate keys, although newer versions emit a deprecation warning. This PR updates yaml loading of dbt projects to also allow duplicate keys and to keep the last value of duplicate keys, matching pyyaml's behavior.

### DIFF
--- a/sqlmesh/dbt/common.py
+++ b/sqlmesh/dbt/common.py
@@ -36,7 +36,9 @@ JINJA_ONLY = {
 
 def load_yaml(source: str | Path) -> t.Dict:
     try:
-        return load(source, render_jinja=False)
+        return load(
+            source, render_jinja=False, allow_duplicate_keys=True, keep_last_duplicate_key=True
+        )
     except DuplicateKeyError as ex:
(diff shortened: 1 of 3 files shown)


sqlmesh/dbt/common.py
```diff
@@ -34,9 +34,9 @@
 }
 
 
-def load_yaml(source: str | Path) -> t.Dict:
+def load_yaml(source: str | Path, allow_duplicate_keys: bool = False) -> t.Dict:
     try:
-        return load(source, render_jinja=False)
+        return load(source, render_jinja=False, allow_duplicate_keys=allow_duplicate_keys)
```
georgesittas: Do we actually need a flag in `load_yaml` itself here? Is there any call site where we actually don't want to allow duplicates? Curious if we'd match dbt's behavior in those cases.

vchan: I just saw this. I made some larger changes for yaml loading to match dbt/pyyaml's behavior of keeping the last occurrence of duplicate keys. I think my changes also address your concerns though.

eakmanrq (approved):

PR #5256, opened 2025-08-28
Fix!: Avoid using rendered query when computing the data hash

izeigerman:
This update eliminates the need for a rendered query when computing the model’s data hash. This means that the categorizer can now decide that the snapshot represents a metadata change even though the data hash has changed.

Doing this has the following benefits:
- Faster fingerprint calculation
- Eliminates the dependency on SQLGlot optimizer when calculating fingerprints which brings us one step closer to eliminating snapshot record migration as a result of the product upgrade

### DIFF
--- a/sqlmesh/core/audit/definition.py
+++ b/sqlmesh/core/audit/definition.py
@@ -15,11 +15,10 @@ from sqlmesh.core.model.common import (
     bool_validator,
     default_catalog_validator,
     depends_on_validator,
-    expression_validator,
     sort_python_env,
     sorted_python_env_payloads,
 )
-from sqlmesh.core.model.common import make_python_env, single_value_or_tuple
+from sqlmesh.core.model.common import make_python_env, single_value_or_tuple, ParsableSql
 from sqlmesh.core.node import _Node
 from sqlmesh.core.renderer import QueryRenderer
 from sqlmesh.utils.date import TimeLike
@@ -67,15 +66,21 @@ class AuditMixin(AuditCommonMetaMixin):
         jinja_macros: A registry of jinja macros to use when rendering the audit query.
     """
 
-    query: t.Union[exp.Query, d.JinjaQuery]
+    query_: ParsableSql
     defaults: t.Dict[str, exp.Expression]
-    expressions_: t.Optional[t.List[exp.Expression]]
+    expressions_: t.Optional[t.List[ParsableSql]]
     jinja_macros: JinjaMacroRegistry
     formatting: t.Optional[bool]
 
     @property
+    def query(self) -> t.Union[exp.Query, d.JinjaQuery]:
+        return t.cast(t.Union[exp.Query, d.JinjaQuery], self.query_.parse(self.dialect))
+
+    @property
     def expressions(self) -> t.List[exp.Expression]:
-        return self.expressions_ or []
+        if self.expressions_:
+            return [e.parse(self.dialect) for e in self.expressions_]
+        return []
 
     @property
     def macro_definitions(self) -> t.List[d.MacroDef]:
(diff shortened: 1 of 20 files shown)


sqlmesh/core/model/definition.py
```diff
@@ -1519,17 +1542,22 @@ def _query_renderer(self) -> QueryRenderer:
...
 
-        query = self.render_query() or self.query
-        data.append(gen(query))
-        data.extend(self.jinja_macros.data_hash_values)
-        return data
+    @property
+    def _data_hash_values(self) -> t.List[str]:
+        return [
+            *self._data_hash_values_no_query,
+            gen(self.query, comments=False),
```
tobymao: can you get rid of this one to and just hash the raw sql?

erindru: If we hash raw sql, doesnt that make it sensitive to whitespace changes?

izeigerman: yes, that's why we now also check `is_metadata_only_change` when data hashes don't match

sqlmesh/core/audit/definition.py
```diff
@@ -67,15 +66,26 @@ class AuditMixin(AuditCommonMetaMixin):
...
+    query_: ParsableSql
     defaults: t.Dict[str, exp.Expression]
-    expressions_: t.Optional[t.List[exp.Expression]]
+    expressions_: t.Optional[t.List[ParsableSql]]
     jinja_macros: JinjaMacroRegistry
     formatting: t.Optional[bool]
 
+    @property
+    def query(self) -> t.Union[exp.Query, d.JinjaQuery]:
+        return t.cast(t.Union[exp.Query, d.JinjaQuery], self.query_.parse(self.dialect))
```
tobymao, line 77: is this called multiple times? should we cache this somehow?

tobymao: i see, you use a class?

sqlmesh/core/model/common.py
```diff
@@ -663,3 +658,65 @@ def parse_strings_with_macro_refs(value: t.Any, dialect: DialectType) -> t.Any:
...
 )(depends_on)
+
+
+class ParsableSql(PydanticModel):
+    sql: str
+
+    _parsed: t.Optional[exp.Expression] = None
+    _parsed_dialect: t.Optional[str] = None
+
+    def parse(self, dialect: str) -> exp.Expression:
```
tobymao, line 669: does the dialect ever change for a model outside of a test?

sqlmesh/core/model/common.py
```diff
@@ -663,3 +658,65 @@ def parse_strings_with_macro_refs(value: t.Any, dialect: DialectType) -> t.Any:
...
+            self._parsed = d.parse_one(self.sql, dialect=dialect)
+            self._parsed_dialect = dialect
+        return self._parsed
+
+    @classmethod
+    def from_parsed_expression(
+        cls, parsed_expression: exp.Expression, dialect: str, use_meta_sql: bool = False
+    ) -> ParsableSql:
+        sql = (
+            parsed_expression.meta.get("sql") or parsed_expression.sql(dialect=dialect)
```
tobymao, line 680: what are the situation where we wouldn't want to use meta sql?

sqlmesh/core/model/common.py
```diff
@@ -663,3 +658,65 @@ def parse_strings_with_macro_refs(value: t.Any, dialect: DialectType) -> t.Any:
...
+    @classmethod
+    def validator(cls) -> classmethod:
+        def _validate_parsable_sql(
+            v: t.Any, info: ValidationInfo
+        ) -> t.Optional[t.Union[ParsableSql, t.List[ParsableSql]]]:
+            if v is None:
+                return v
+            if isinstance(v, str):
+                return ParsableSql(sql=v)
+            if isinstance(v, exp.Expression):
```
tobymao, line 698: when is this branch called

sqlmesh/core/audit/definition.py
izeigerman: yes, it's cached inside `ParsableQuery`

sqlmesh/core/model/common.py
izeigerman: It shouldn't, why? I'd rather implement this correctly in case circumstances change.

sqlmesh/core/model/common.py
izeigerman: when I'm using a custom loader and do `create_sql_model` directly with a query. I don't think we can trust the correctness of the meta sql in that case.

sqlmesh/core/model/common.py
izeigerman:
```
SqlModel(
    name="some_name",
    query=some_other_model.query.select(...),
    ...
 )
 ```

tobymao (approved):

PR #5258, opened 2025-08-28
Chore: Address more integration test flakiness

erindru:
This addresses:

 - An oversight in tearing down the database for Redshift that caused [this failure](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/23761/workflows/3630bb10-cb60-4a40-8a46-ceac25108e44/jobs/270112)
 - The failures caused by `teardown failed on attempt N! Exiting immediately!` when there is a `StashKey` error
    - eg [here for trino](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/23774/workflows/da3cc0fb-02fc-41d7-8bcd-0d89309731bf/jobs/270410)
    - and [here for fabric](https://app.circleci.com/pipelines/github/TobikoData/sqlmesh/23756/workflows/541b8d2c-da3d-4f18-94db-43871d1dec05/jobs/270034)

I had a handler for capturing the `StashKey` errors and it worked - if `pytest-retry` wasnt in the mix. So it was working for `style_and_cicd_tests` but not the engine adapter integration tests.

`pytest-retry` registers its own handler, which runs before ours and bombs out before we even have a chance to handle the error.

So this PR adjusts the hook call order at runtime to call ours first before the one from `pytest-retry`

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -310,10 +310,10 @@ workflows:
                 - athena
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
       - ui_style
       - ui_test
       - vscode_test
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -310,10 +310,10 @@ workflows:
                 - athena
                 - fabric
                 - gcp-postgres
-          #filters:
-          #  branches:
-          #    only:
-          #      - main
+          filters:
+            branches:
+              only:
+                - main
       - ui_style
       - ui_test
       - vscode_test
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -245,6 +245,51 @@ def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo):
             )
 
 
+def pytest_configure(config: pytest.Config):
+    # we need to adjust the hook order if pytest-retry is present because it:
+    # - also declares a `pytest_runtest_makereport` with `hookwrapper=True, tryfirst=True`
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -245,6 +245,51 @@ def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo):
...
+    # but, we need our one to run first because it's capturing and ignoring certain errors that cause pytest-retry to fail
+    # and not retry. so we need to adjust the order the hooks are called which pytest does NOT make easy.
+    #
+    # we can't just unload the pytest-retry plugin, load our hook and reload the pytest-retry plugin either.
+    # this causes an error:
+    # > Hook 'pytest_set_excluded_exceptions' is already registered within namespace
+    # because unregister() apparently doesnt unregister plugins cleanly in such a way they can be re-registered
+    #
+    # so what we end up doing below is a small monkey-patch to adjust the call order of the hooks
+    pm = config.pluginmanager
```
erindru, line 262: I tried quite a few things, including trying to register `tests` as a pytest plugin in `pyproject.toml` (which has its own set of problems) before arriving at this implementation as the least horrendous option

.circleci/continue_config.yml
```diff
@@ -310,10 +310,10 @@ workflows:
...
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
+          #  branches:
+          #    only:
+          #      - main
```
eakmanrq: Reminder to revert this

eakmanrq (approved):

.circleci/continue_config.yml
erindru: thanks :)

PR #5266, opened 2025-08-29
Fix: Match dbt handling of cluster_by config for view materializations

crericha:
dbt disregards cluster_by config when creating tables for view models.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -503,6 +503,7 @@ class ModelConfig(BaseModelConfig):
         """Converts the dbt model into a SQLMesh model."""
         model_dialect = self.dialect(context)
         query = d.jinja_query(self.sql_no_config)
+        kind = self.model_kind(context)
 
         optional_kwargs: t.Dict[str, t.Any] = {}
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/model.py
```diff
@@ -467,7 +468,7 @@ def to_sqlmesh(
                 partitioned_by.append(self._big_query_partition_by_expr(context))
             optional_kwargs["partitioned_by"] = partitioned_by
 
-        if self.cluster_by:
+        if self.cluster_by and not isinstance(kind, ViewKind):
```
georgesittas: Don't we want to push this further down into sqlmesh's model validator? I thought it wasn't possible to cluster normal views in general.

crericha: It *can* be pushed down into sqlmesh; however, will it match how sqlmesh wants to handle misconfiguration? I'm under the impression that sqlmesh wants to error, not ignore. wdyt?

georgesittas: I see, that makes sense to me. Do we already throw an error today for this? (Just curious, fine if we don't for now.)

PR #5269, opened 2025-09-01
Fix!: depend on all attributes of dbt `model` when passed to a macro

georgesittas:
When we pass the dbt `model` keyword to a macro (e.g., `{{ foo(model) }}`), we can't easily track the attributes that are actually used, because the argument may be aliased and hence tracking actual uses of `model` requires a proper data flow analysis.

This PR handles this case by conservatively including all of `model`'s supported attributes if a "standalone" reference is detected. We decided to exclude `raw_code` to reduce the snapshot's payload size, since we can access it through the `JinjaQuery` node directly.

P.S.: may need to revisit before merging to see if we can migrate this properly.

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -594,8 +594,7 @@ class _Model(ModelMeta, frozen=True):
                 python_env=self.python_env,
                 only_execution_time=False,
                 default_catalog=self.default_catalog,
-                model_fqn=self.fqn,
-                raw_code=self._raw_code,
+                model=self,
             )
         return self._statement_renderer_cache[expression_key]
 
@@ -1306,10 +1305,6 @@ class _Model(ModelMeta, frozen=True):
     def violated_rules_for_query(self) -> t.Dict[type[Rule], t.Any]:
         return {}
 
-    @property
-    def _raw_code(self) -> t.Optional[str]:
-        return None
-
 
 class SqlModel(_Model):
     """The model definition which relies on a SQL query to fetch the data.
@@ -1578,7 +1573,6 @@ class SqlModel(_Model):
             self.dialect,
             self.macro_definitions,
             schema=self.mapping_schema,
-            model_fqn=self.fqn,
             path=self._path,
             jinja_macro_registry=self.jinja_macros,
             python_env=self.python_env,
@@ -1586,7 +1580,7 @@ class SqlModel(_Model):
             default_catalog=self.default_catalog,
             quote_identifiers=not no_quote_identifiers,
             optimize_query=self.optimize_query,
-            raw_code=self._raw_code,
+            model=self,
         )
 
     @property
@@ -1611,11 +1605,6 @@ class SqlModel(_Model):
     def violated_rules_for_query(self) -> t.Dict[type[Rule], t.Any]:
         self.render_query()
         return self._query_renderer._violated_rules
-
-    @property
-    def _raw_code(self) -> t.Optional[str]:
-        query = self.query
-        return query.name if isinstance(query, d.JinjaQuery) else None
 
 
 class SeedModel(_Model):
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -31,6 +31,7 @@ if t.TYPE_CHECKING:
     from sqlglot.dialects.dialect import DialectType
 
     from sqlmesh.core.linter.rule import Rule
+    from sqlmesh.core.model.definition import _Model
     from sqlmesh.core.snapshot import DeployabilityIndex, Snapshot
 
 
@@ -50,10 +51,9 @@ class BaseExpressionRenderer:
         schema: t.Optional[t.Dict[str, t.Any]] = None,
         default_catalog: t.Optional[str] = None,
         quote_identifiers: bool = True,
-        model_fqn: t.Optional[str] = None,
         normalize_identifiers: bool = True,
         optimize_query: t.Optional[bool] = True,
-        raw_code: t.Optional[str] = None,
+        model: t.Optional[_Model] = None,
     ):
         self._expression = expression
         self._dialect = dialect
@@ -67,9 +67,9 @@ class BaseExpressionRenderer:
         self._quote_identifiers = quote_identifiers
         self.update_schema({} if schema is None else schema)
         self._cache: t.List[t.Optional[exp.Expression]] = []
-        self._model_fqn = model_fqn
+        self._model_fqn = model.fqn if model else None
         self._optimize_query_flag = optimize_query is not False
-        self._raw_code = raw_code
+        self._model = model
 
     def update_schema(self, schema: t.Dict[str, t.Any]) -> None:
         self.schema = d.normalize_mapping_schema(schema, dialect=self._dialect)
@@ -206,7 +206,7 @@ class BaseExpressionRenderer:
                     "default_catalog": self._default_catalog,
                     "runtime_stage": runtime_stage.value,
                     "resolve_table": _resolve_table,
-                    "raw_code": self._raw_code,
+                    "model_instance": self._model,
                 }
 
                 if this_model:
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -16,6 +16,7 @@ from sqlglot import Dialect
 
 from sqlmesh.core.console import get_console
 from sqlmesh.core.engine_adapter import EngineAdapter
+from sqlmesh.core.model.definition import SqlModel
 from sqlmesh.core.snapshot.definition import DeployabilityIndex
 from sqlmesh.dbt.adapter import BaseAdapter, ParsetimeAdapter, RuntimeAdapter
 from sqlmesh.dbt.common import RAW_CODE_KEY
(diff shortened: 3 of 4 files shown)


examples/sushi_dbt/macros/check_model_is_table.sql
erindru, line 1: Adding stuff directly to `sushi_dbt` means that every single test that uses `sushi_dbt` pays the tax, even if they're testing something completely unrelated.

Did you consider using `sushi_dbt` as a base and creating this macro file + a model to call it within the relevant test code?

izeigerman: +1. We have this project in `tests/fixtures`. Can we use that one instead?

sqlmesh/core/renderer.py
```diff
@@ -188,30 +188,32 @@ def _resolve_table(table: str | exp.Table) -> str:
...
+                        **_prepare_python_env_for_jinja(macro_evaluator, self._python_env),
+                        **variables,
+                    },
+                    "snapshots": snapshots or {},
+                    "table_mapping": table_mapping,
+                    "deployability_index": deployability_index,
+                    "default_catalog": self._default_catalog,
+                    "runtime_stage": runtime_stage.value,
+                    "resolve_table": _resolve_table,
+                    "raw_code": self._expression.name,
```
izeigerman: I don't think this is accurate. If we render a pre-hook / post-hook and refer to `raw_code` it will include the raw code of the hook instead of the model.

sqlmesh/dbt/common.py
```diff
@@ -19,6 +19,7 @@
 T = t.TypeVar("T", bound="GeneralConfig")
 
 PROJECT_FILENAME = DBT_PROJECT_FILENAME
+DBT_ALL_MODEL_ATTRS = "__DBT_ALL_MODEL_ATTRS__"
```
izeigerman: I don't like the idea of bespoke placeholders. Can we use a union type or a dataclass instead? 
```
@dataclass
class AllModelAttrs:
    pass

model_attrs: t.Union[t.Set[str], AllModelAttrs]
```

or

```
@dataclass
class ModelAttrs:
    attrs: t.Set[str]
    all_attrs: bool = False

model_attrs: ModelAttrs
```

examples/sushi_dbt/macros/check_model_is_table.sql
georgesittas: Weird, I thought I pushed this change already yesterday when I discussed with Ryan. Yeah I agree, will amend.

sqlmesh/dbt/basemodel.py
```diff
@@ -375,15 +377,23 @@ def to_sqlmesh(
...
+        if context._manifest and self.node_name in context._manifest._manifest.nodes:
+            attributes = context._manifest._manifest.nodes[self.node_name].to_dict()
+            if DBT_ALL_MODEL_ATTRS in dependencies.model_attrs:
+                model_node: AttributeDict[str, t.Any] = AttributeDict(attributes)
+            else:
+                model_node = AttributeDict(
+                    filter(lambda kv: kv[0] in dependencies.model_attrs, attributes.items())
+                )
+
+            raw_code_key = "raw_code" if DBT_VERSION >= (1, 3, 0) else "raw_sql"  # type: ignore
```
izeigerman: This is something that I think should belong in `common.py` as a named constant

sqlmesh/dbt/builtin.py
```diff
@@ -249,6 +256,21 @@ def source(package: str, name: str) -> t.Optional[BaseRelation]:
     return source
 
 
+def generate_model(model: AttributeDict, raw_code: str) -> Model:
+    class Model:
+        def __init__(self, model: AttributeDict) -> None:
+            self._model = model
+            self._raw_code_key = "raw_code" if DBT_VERSION >= (1, 3, 0) else "raw_sql"  # type: ignore
```
izeigerman: ditto

sqlmesh/core/renderer.py
georgesittas: Ah, I see. Good catch, I clearly missed this case :)

sqlmesh/dbt/builtin.py
```diff
@@ -249,6 +256,21 @@ def source(package: str, name: str) -> t.Optional[BaseRelation]:
...
+
+        def __getattr__(self, key: str) -> t.Any:
+            if key == self._raw_code_key:
```
izeigerman: I don't think this is necessary if you just do:
```
self._model = AttributeDict({
    **model,
    raw_code_key: raw_code,
})
```
in __init__. Actually if you don't really need the `Model` class at all if you just construct this dict and return it in `generate_model`.

sqlmesh/dbt/common.py
georgesittas: Sure, it's a small refactor. Could you elaborate a bit on why you dislike it? I considered using a `Union` too but it felt like it would make the implementation a bit more involved without providing an obvious advantage.

examples/sushi_dbt/macros/check_model_is_table.sql
georgesittas: I [moved the relevant logic](https://github.com/TobikoData/sqlmesh/pull/5269/commits/59e91de6f8b9dc2c78f89e9a079cc7db02105557) to `tests/fixtures/dbt`, thanks folks.

sqlmesh/dbt/common.py
georgesittas: Refactored: https://github.com/TobikoData/sqlmesh/pull/5269/commits/eea374718d4d343f7791dbdd2f80ab29fa3d3e41

sqlmesh/dbt/builtin.py
georgesittas: Good feedback, thank you! Addressed here: https://github.com/TobikoData/sqlmesh/pull/5269/commits/f8bd60b58deb457f7cfe48f85881bb945c2c277f.

sqlmesh/dbt/basemodel.py
georgesittas: Addressed here: https://github.com/TobikoData/sqlmesh/pull/5269/commits/f8bd60b58deb457f7cfe48f85881bb945c2c277f

sqlmesh/core/renderer.py
georgesittas: @izeigerman I played around with this a bit and it turns out we can't use the model's raw query SQL... dbt executes the following successfully and the result indicates that the whole file is taken into account for `raw_code`:

```
((.venv) ) ➜  cat models/test_model.sql
{{ config(pre_hook=['CREATE TABLE t AS SELECT \'Length is {{ model.raw_code|length }}\' AS length_col']) }}

SELECT 1 AS c
((.venv) ) ➜  duckdb -c "select * from t" local.duckdb
┌───────────────┐
│  length_col   │
│    varchar    │
├───────────────┤
│ Length is 122 │
└───────────────┘
((.venv) ) ➜  wc models/test_model.sql
       3      18     123 models/test_model.sql
```

This means that relying solely on the query or pre/post-statements of a model is not enough, unfortunately.. Perhaps we need to keep the `raw_code` in state after all?

georgesittas: Discussed internally and arrived at this solution: https://github.com/TobikoData/sqlmesh/pull/5269/commits/8919c487ac662314c7d1ca55468293586267b1d1

sqlmesh/core/model/definition.py
```diff
@@ -595,6 +595,7 @@ def _statement_renderer(self, expression: exp.Expression) -> ExpressionRenderer:
                 only_execution_time=False,
                 default_catalog=self.default_catalog,
                 model_fqn=self.fqn,
+                raw_code=self._raw_code,
```
izeigerman: I don't know if we need this interface. Can we just pass `self` instead of both `model_fqn` and `raw_code`? The builtins module can just use the model reference directly to extract the raw query

sqlmesh/dbt/builtin.py
```diff
@@ -26,6 +27,13 @@
 from sqlmesh.utils.errors import ConfigError, MacroEvalError
 from sqlmesh.utils.jinja import JinjaMacroRegistry, MacroReference, MacroReturnVal
 
+if t.TYPE_CHECKING:
+    from typing import Protocol
+
+    class Model(Protocol):
```
izeigerman: Is this used for anything?

sqlmesh/migrations/v0095_warn_about_dbt_raw_sql_diff.py
```diff
@@ -0,0 +1,47 @@
...
+    if schema:
+        snapshots_table = f"{schema}.{snapshots_table}"
+
+    warning = (
+        "SQLMesh now includes dbt's {{ config(...) }} blocks in the model's raw SQL when "
+        "processing dbt models. This change ensures that all model attributes referenced "
+        "in macros are properly tracked for fingerprinting. As a result, you may see diffs "
+        "for existing dbt models even though the actual SQL logic hasn't changed. This is "
+        "a one-time diff that will be resolved after applying a plan. Run 'sqlmesh diff prod' "
+        "to review any changes, then apply a plan if the diffs look expected."
```
izeigerman: We should mention `sqlmesh plan --forward-only` to avoid backfills.

sqlmesh/core/model/definition.py
```diff
@@ -1606,6 +1612,11 @@ def violated_rules_for_query(self) -> t.Dict[type[Rule], t.Any]:
         self.render_query()
         return self._query_renderer._violated_rules
 
+    @property
+    def _raw_code(self) -> t.Optional[str]:
+        query = self.query
+        return query.name if isinstance(query, d.JinjaQuery) else None
```
izeigerman: btw, you can now do `model.query_.sql` to access the original raw sql of any model

izeigerman: ah, no, nvm. In this case it will have `JINJA_*` block which is undesirable

sqlmesh/dbt/builtin.py
georgesittas: Nope, leftover from previous iteration. Thanks!

sqlmesh/core/model/definition.py
georgesittas: Done

PR #5271, opened 2025-09-01
Feat(dbt): Add config object to provide methods aligned with dbt

themisvaltinos:
This adds support for the `config` variable and its methods. This exists to handle end-user configuration for custom materialisations and this a first part in supporting that, but also in general to align with dbt's [config](https://docs.getdbt.com/reference/dbt-jinja-functions/config).

  ```jinja
{% materialization incremental, default -%}
  -- Example w/ no default. unique_key will be None if the user does not provide this configuration
  {%- set unique_key = config.get('unique_key') -%}

  -- Example w/ alternate value. Use alternative of 'id' if 'unique_key' config is provided, but it is None
  {%- set unique_key = config.get('unique_key') or 'id' -%}

  -- Example w/ default value. Default to 'id' if the 'unique_key' config does not exist
  {%- set unique_key = config.get('unique_key', default='id') -%}

  -- Example of a custom config nested under `meta` as required in v1.10 and higher.
  {% set my_custom_config = config.get('meta').custom_config_key %}
  ...

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -164,16 +164,16 @@ class Var:
         return name in self.variables
 
 
-class ConfigObject:
+class Config:
     def __init__(self, config_dict: t.Dict[str, t.Any]) -> None:
-        self._config = config_dict or {}
+        self._config = config_dict
 
     def __call__(self, **kwargs: t.Any) -> str:
         self._config.update(**kwargs)
         return ""
 
     def set(self, name: str, value: t.Any) -> str:
-        self.__call__(**{name: value})
+        self._config.update({name: value})
         return ""
 
     def _validate(self, name: str, validator: t.Callable, value: t.Optional[t.Any] = None) -> None:
@@ -447,7 +447,7 @@ def create_builtin_globals(
     if variables is not None:
         builtin_globals["var"] = Var(variables)
 
-    builtin_globals["config"] = ConfigObject(jinja_globals.pop("config", {}))
+    builtin_globals["config"] = Config(jinja_globals.pop("config", {}))
 
     deployability_index = (
         jinja_globals.get("deployability_index") or DeployabilityIndex.all_deployable()


tests/dbt/test_transformation.py
```diff
@@ -946,7 +946,7 @@ def test_schema_jinja(sushi_test_project: Project, assert_exp_eq):
 
 @pytest.mark.xdist_group("dbt_manifest")
 def test_config_jinja(sushi_test_project: Project):
-    hook = "{{ config(alias='bar') }} {{ config.alias }}"
+    hook = "{{ config(alias='bar') }} {{ config.get('alias') }}"
```
georgesittas, line 949: Why did this change?

sqlmesh/dbt/builtin.py
```diff
@@ -164,6 +164,58 @@ def has_var(self, name: str) -> bool:
         return name in self.variables
 
 
+class ConfigObject:
```
georgesittas: Nit: let's rename to just `Config` for brevity & consistency w/ others in `builtin.py`.

sqlmesh/dbt/builtin.py
```diff
@@ -164,6 +164,58 @@ def has_var(self, name: str) -> bool:
...
+    def __init__(self, config_dict: t.Dict[str, t.Any]) -> None:
+        self._config = config_dict or {}
```
georgesittas: The `or {}` term looks redundant given `config_dict`'s type hint.

sqlmesh/dbt/builtin.py
```diff
@@ -164,6 +164,58 @@ def has_var(self, name: str) -> bool:
...
+
+    def __call__(self, **kwargs: t.Any) -> str:
+        self._config.update(**kwargs)
+        return ""
+
+    def set(self, name: str, value: t.Any) -> str:
+        self.__call__(**{name: value})
```
georgesittas: Why delegate to `__call__` instead of doing a simple dictionary update?

georgesittas (commented):
Does this PR affect the snapshot payload, i.e., can it result in post-migration diffs?

tests/dbt/test_transformation.py
themisvaltinos: because before it was an attributedict with dot-attribute access, while now wrapping it in the `Config` class it holds still the dict internally but uses the get methods similar to dbt

sqlmesh/dbt/builtin.py
themisvaltinos: good point revised it to simply do update

tests/dbt/test_transformation.py
georgesittas: Does dbt support the attribute syntax though? I'm worried about whether this introduces a regression.

themisvaltinos: no the attribute dict does work for model.config so not sure if that was the confusion and why we had it like that in the past for config. but from looking at the docs and the dbt codebase the attribute syntax wouldn't work. so if you take a model with a simple config like this:
```sql
{{
    config(
        materialized='table',
        unique_key='order_id',
        pre_hook="{{ log_unique_key() }}"
    )
}}
```
and the macro using the config:
```sql
{% macro log_unique_key() %}
  {% set unique_key = config.unique_key %}
  {{ log("key: " + unique_key, info="true")}}
{% endmacro %}
```
would raise the error:
```bash
07:08:20  Encountered an error:
Compilation Error in model customers (models/customers.sql)
  'dbt.context.providers.ParseConfigObject object' has no attribute 'unique_key'
```
while using the getter method (`config.get("unique_key")` ) logs it as expected:
```bash
07:12:20  key: order_id
```

georgesittas: Nice, thank you for verifying.

PR #5273, opened 2025-09-01
Feat(state_sync): Add the ability to fetch all versions of a snapshot by name

erindru:
This is the first in a series of PR's looking at adjusting and clarifying how **restatements** work.

Currently, when a model is restated in `prod`, the following occurs:
 1. Intervals are cleared from state for that model
 2. Intervals are cleared from state for all versions of that model present in dev environments

The purpose of that last step is to prevent stale / old data from getting promoted to `prod` if someone decides to deploy a dev environment containing a version of the model that got built before the prod restatement occurred.

However, the current implementation has a gap. It only clears intervals for snapshots that are *promoted* in dev environments, because it's limited by what `StateSync` methods are available and it currently iterates over all environments from `StateSync.get_environments_summary()`.

What about snapshots that arent currently promoted in any environment? There is still a risk that someone makes a change in dev that matches one of these snapshots, and it becomes promoted in dev, and then gets deployed to prod.

So, this PR adds a method to `StateSync` that allows fetching all versions of a snapshot (based on name) regardless of if it's currently being used in an environment or not.

This will allow the `prod` restatement "clear intervals" code to properly clear all relevant intervals across all snapshots of the model, rather than skipping intervals for snapshots that don't currently belong to any environment like it does today.

### DIFF
--- a/sqlmesh/core/state_sync/db/snapshot.py
+++ b/sqlmesh/core/state_sync/db/snapshot.py
@@ -26,7 +26,7 @@ from sqlmesh.core.snapshot import (
     SnapshotNameVersion,
     SnapshotInfoLike,
     Snapshot,
-    MinimalSnapshot,
+    SnapshotIdAndVersion,
     SnapshotId,
     SnapshotFingerprint,
 )
@@ -214,7 +214,7 @@ class SnapshotState:
             for snapshot in environment.snapshots
         }
 
-        def _is_snapshot_used(snapshot: MinimalSnapshot) -> bool:
+        def _is_snapshot_used(snapshot: SnapshotIdAndVersion) -> bool:
             return (
                 snapshot.snapshot_id in promoted_snapshot_ids
                 or snapshot.snapshot_id not in expired_candidates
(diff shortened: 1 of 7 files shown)


georgesittas (approved):

izeigerman (approved):

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -308,6 +307,52 @@ def get_snapshots(
...
+        else:
+            unexpired_expr = None
+
+        return {
+            MinimalSnapshot(
+                name=name,
+                identifier=identifier,
+                version=version,
+                dev_version=dev_version,
+                fingerprint=SnapshotFingerprint.parse_raw(fingerprint),
```
izeigerman: I think this can still be too much parsing. Can we parse it lazily instead? Similar to how we do this with `snapshots` in the `Environment` model.

erindru: I did wonder about that. I've updated it to be lazily parsed on first access, although it still needs to be provided in string form to construct the object to begin with

PR #5274, opened 2025-09-02
Fix: When restating prod, clear intervals across all related snapshots, not just promoted ones

erindru:
This builds on `StateSync.get_snapshot_ids_by_names` introduced in #5273 

The current implementation of prod restatements only prevents dev data from getting promoted to prod for snapshots that are currently "active" in a dev environment.

However, it's possible for snapshots to exist that aren't currently promoted in any dev environment (since they represent an older version of a model) but may get promoted if a user reverts to that older version.

If that revert happens and then the environment is subsequently deployed, the old data may still make its way to prod.

So this PR ensures that all versions of affected snapshots have their intervals cleared, regardless of if they are promoted in an environment or not

### DIFF
--- a/sqlmesh/core/plan/common.py
+++ b/sqlmesh/core/plan/common.py
@@ -1,6 +1,15 @@
 from __future__ import annotations
+import typing as t
+import logging
+from dataclasses import dataclass, field
 
-from sqlmesh.core.snapshot import Snapshot
+from sqlmesh.core.state_sync import StateReader
+from sqlmesh.core.snapshot import Snapshot, SnapshotId, SnapshotTableInfo, SnapshotNameVersion
+from sqlmesh.core.snapshot.definition import Interval
+from sqlmesh.utils.dag import DAG
+from sqlmesh.utils.date import now_timestamp
+
+logger = logging.getLogger(__name__)
 
 
 def should_force_rebuild(old: Snapshot, new: Snapshot) -> bool:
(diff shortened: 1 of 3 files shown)


sqlmesh/core/plan/common.py
```diff
@@ -24,3 +33,173 @@ def is_breaking_kind_change(old: Snapshot, new: Snapshot) -> bool:
...
+    @property
+    def snapshot_id(self) -> SnapshotId:
+        return self.table_info.snapshot_id
+
+    @property
+    def sorted_environment_names(self) -> t.List[str]:
+        return list(sorted(self.environment_names))
+
+
+def identify_restatement_intervals_across_snapshot_versions(
```
erindru, line 63: This got moved to `common` because I plan to call it outside the evaluator in an upcoming PR that improves the console output as well as the `--explain` output around restatements

sqlmesh/core/plan/common.py
```diff
@@ -27,3 +36,182 @@ def is_breaking_kind_change(old: Snapshot, new: Snapshot) -> bool:
...
+        for s in state_reader.get_snapshots_by_names(
+            snapshot_names=unique_snapshot_names, current_ts=current_ts, exclude_expired=True
+        )
+        # Don't clear intervals for a snapshot if it shares the same physical version with prod.
+        # Otherwise, prod will be affected by what should be a dev operation
+        if s.name_version not in prod_name_versions
+    }
+
+    # identify the ones that we havent picked up yet, which are the ones that dont exist in any environment
+    if remaining_snapshot_ids := set(all_matching_non_prod_snapshots).difference(
```
izeigerman, line 148: I guess it's worth noting that there's still a slight risk of missing a relevant dependency. For example, if there's a snapshot A' that is not promoted anywhere which has a downstream dependency D which is a model that has been removed from all existing environments, we won't drop intervals for D like we're suppose to because its name won't show up here.

sqlmesh/core/plan/common.py
```diff
@@ -27,3 +36,182 @@ def is_breaking_kind_change(old: Snapshot, new: Snapshot) -> bool:
...
+            next_start = min(current_start, next_start)
+            next_end = max(current_end, next_end)
+
+            snapshot_name_to_widest_interval[s_id.name] = (next_start, next_end)
+
+        # we need to fetch full Snapshot's to get access to the SnapshotTableInfo objects
+        # required by StateSync.remove_intervals()
+        # but at this point we have minimized the list by excluding the ones that are already present in prod
+        # and also excluding the ones we have already matched earlier while traversing the environment DAGs
+        remaining_snapshots = state_reader.get_snapshots(snapshot_ids=remaining_snapshot_ids)
```
izeigerman: Why are we still fetching full snapshots for all remaining snapshots and not just the ones that are `full_history_restatement_only`?

sqlmesh/core/plan/common.py
erindru: Yes, that's true. We only determine the dependency tree based on promoted snapshots.

I couldn't think of a sane way to push dependency resolution to the db layer, and reading all snapshots to figure this out is a non-starter, so you're right in that this PR improves rather than fully eliminates the current situation.

sqlmesh/core/plan/common.py
erindru: The handling of `full_history_restatement_only` happens next on line 183.

This step is just for snapshots that are not promoted in any environment. They may or may not be `full_history_restatement_only`, at this point we are just identifying them and setting the intervals to clear based on the requested restatements.

Note that Environments only have `SnapshotTableInfo` available, not full `Snapshot`, so everything within this method is based on `SnapshotTableInfo`. The other reason for needing `SnapshotTableInfo` is that ultimately these get passed to `StateSync.remove_intervals()`, which requires `SnapshotTableInfo` at a minimum.

So the order currently goes:

 - Iterate through each environment, pick up any snapshots _and downstream dependencies_ matching the requested restatements and add the `SnapshotId`/`SnapshotTableInfo` `snapshot_intervals_to_clear`. 
 - Then, based on `name`, pick up any remaining snapshots that weren't promoted in any environment, and add the `SnapshotId`/`SnapshotTableInfo` to `snapshot_intervals_to_clear`. In order to get `SnapshotTableInfo` for them, we have to look up the full `Snapshot` record. Note that we extend `loaded_snapshots` with these so we can use them in the next step.
 - Next, for everything we have identified in `snapshot_intervals_to_clear`, we check if any are `full_history_restatement_only` and if they are selectively widen intervals for those ones. To selectively widen, we need full snapshots to pass to `Snapshot.get_removal_interval`, so we fetch full `Snapshot`s here.
 
Snapshot-fetching optimizations exist in the following places:
 - `remaining_snapshot_ids` are just the ones we haven't already fetched. The ones we have already fetched are in `snapshot_intervals_to_clear` so we do a set difference to limit to just the additional ones we need and put those in `remaining_snapshot_ids`. 
 - After fetching full snapshots for `remaining_snapshot_ids` to get their `SnapshotTableInfo`'s, we add them to the list of `loaded_snapshots` so they dont need to be looked up again,
 - When handling `full_history_restatement_only`, we only fetch full `Snapshot`s for the ones that don't already exist in `loaded_snapshots`
 
 And of course if any of the affected snapshot id's had ever been loaded at a different time, they will already exist in the disk cache

izeigerman: > In order to get SnapshotTableInfo for them, we have to look up the full Snapshot record. Note that we extend loaded_snapshots with these so we can use them in the next step.

Didn't we introduce `SnapshotIdAndVersion` precisely to avoid fetching full snapshots here? 

> The other reason for needing SnapshotTableInfo is that ultimately these get passed to StateSync.remove_intervals(), which requires SnapshotTableInfo at a minimum.

I'm pretty sure we can update that interface to accept `SnapshotIdAndVersion`, no?

erindru: Ok, i've:
 - added `kind_name` to `SnapshotIdAndVersion` (so we can determine `full_history_restatement_only` without fetching full `Snapshot`s) 
 - Created a `SnapshotIdAndVersionLike` type that encompasses `SnapshotIdAndVersion` plus its supersets
 - reduced the `StateSync.remove_intervals` api to only require `SnapshotIdAndVersionLike` rather than `SnapshotInfoLike`

sqlmesh/core/snapshot/definition.py
```diff
@@ -595,6 +606,7 @@ class SnapshotIdAndVersion(PydanticModel):
 
     name: str
     version: str
+    kind_name: t.Optional[ModelKindName] = None
```
erindru: Note: this is optional because it's defined as optional on `SnapshotTableInfo` and I wanted to keep treating `SnapshotTableInfo` as a superset

sqlmesh/core/plan/common.py
```diff
@@ -27,3 +36,176 @@ def is_breaking_kind_change(old: Snapshot, new: Snapshot) -> bool:
...
+    # include the whole time range for that snapshot. This requires a call to state to load the full snapshot record,
+    # so we only do it if necessary
+    full_history_restatement_snapshot_ids = [
+        # FIXME: full_history_restatement_only is just one indicator that the snapshot can only be fully refreshed, the other one is Model.depends_on_self
+        # however, to figure out depends_on_self, we have to render all the model queries which, alongside having to fetch full snapshots from state,
+        # is problematic in secure environments that are deliberately isolated from arbitrary user code (since rendering a query may require user macros to be present)
+        # So for now, these are not considered
+        s_id
+        for s_id, s in snapshot_intervals_to_clear.items()
+        if s.snapshot.kind_name and s.snapshot.kind_name.full_history_restatement_only
```
izeigerman: I'm pretty sure `SnapshotIdAndVersion` can implement `ModelKindMixin` and then you can continue doing just `snapshot.full_history_restatement_only` .

erindru: Good idea, that didnt occur to me. Done

izeigerman (approved):

PR #5275, opened 2025-09-02
feat(experimental): add official support for model grants

### LINKED ISSUE #569: Support dbt grants config
- Depends on #568
- grants config docs: https://docs.getdbt.com/reference/resource-configs/grants

newtonapple:
# Native grants for SQLMesh models
SQLMesh now supports configuring model-level grants directly in model definitions. Use the new `grants` and `grants_target_layer` properties to assign privileges to specific roles.

- `grants`: maps each privilege to the list of grantees.
- `grants_target_layer`: selects the data layer object that receives the grant (`virtual` default view, `physical` table, or `all`).

> Note: When [`virtual_environment_mode`](https://sqlmesh.readthedocs.io/en/stable/guides/configuration/#virtual-data-environment-modes) is `dev_only`, `grants_target_layer` is ignored for the `prod` environment. Per-environment plans (e.g. `sqlmesh plan staging`) still honor the configured target layer.

This PR comes with grants support for the following engine adapters:
- bigquery
- databricks
- postgres
- redshift
- snowflake

## Model configuration example
```sql
MODEL (
    name prod.user_ids,
    kind FULL,
    grants (
        'select' = ['readonly_user', 'dashboard_user']
    ),
    grants_target_layer virtual,
);
SELECT 1 AS id;
```

See `tests/core/test_model.py` for additional scenarios.

Fixes #569.


sqlmesh/core/engine_adapter/base.py
```diff
@@ -2345,6 +2350,193 @@ def wap_publish(self, table_name: TableName, wap_id: str) -> None:
         """
         raise NotImplementedError(f"Engine does not support WAP: {type(self)}")
 
+    def _get_current_grants_config(self, table: exp.Table) -> GrantsConfig:
```
izeigerman: [Nit] Let's try and keep private methods at the end of the class definition.

sqlmesh/core/model/meta.py
```diff
@@ -284,6 +327,14 @@ def _refs_validator(cls, vs: t.Any, info: ValidationInfo) -> t.List[exp.Expressi
     def ignored_rules_validator(cls, vs: t.Any) -> t.Any:
         return LinterConfig._validate_rules(vs)
 
+    @field_validator("grants_target_layer", mode="before")
+    def _grants_target_layer_validator(cls, v: t.Any) -> t.Any:
+        return cls._validate_str_enum_value(v)
+
+    @field_validator("virtual_environment_mode", mode="before")
+    def _virtual_environment_mode_validator(cls, v: t.Any) -> t.Any:
```
izeigerman: The 2 fields can share the same validator. There's no point in having 2 and a separate `_validate_str_enum_value` class method
```
@field_validator("virtual_environment_mode", "grants_target_layer", mode="before")
```

sqlmesh/core/model/meta.py
```diff
@@ -462,6 +520,35 @@ def custom_materialization_properties(self) -> CustomMaterializationProperties:
...
+                return e.this.strip()
+            if isinstance(e, exp.Identifier):
+                return e.name
+            return e.sql(dialect=self.dialect).strip()
+
+        grants_dict = {}
+        for eq_expr in self.grants_.expressions:
+            permission_name = parse_exp_to_str(eq_expr.this)  # left hand side
+            grantees_expr = eq_expr.expression  # right hand side
+            if isinstance(grantees_expr, exp.Array):
```
izeigerman: If it's not array we just silently ignore it? Do we validate anywhere that the right-hand side is an array?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1713,6 +1763,9 @@ def promote(
             view_properties=model.render_virtual_properties(**render_kwargs),
         )
 
+        # Apply grants to the virtual layer view
+        self._apply_grants(model, view_name, GrantsTargetLayer.VIRTUAL)
```
izeigerman: If all evaluation strategies share this, I suggest calling this in the `SnapshotEvaluator` itself rather than in individual strategies.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1609,6 +1612,53 @@ def demote(self, view_name: str, **kwargs: t.Any) -> None:
...
+        if not (model_grants_target_layer.is_all or model_grants_target_layer == target_layer):
+            logger.debug(
+                f"Skipping grants application for model {model.name} in {target_layer} layer"
+            )
+            return
+
+        logger.info(f"Applying grants for model {model.name} to table {table_name}")
+
+        try:
+            self.adapter._sync_grants_config(
```
izeigerman: Why are we calling private engine adapter methods here?

sqlmesh/core/model/meta.py
```diff
@@ -86,6 +119,8 @@ class ModelMeta(_Node):
     )
     formatting: t.Optional[bool] = Field(default=None, exclude=True)
     virtual_environment_mode: VirtualEnvironmentMode = VirtualEnvironmentMode.default
+    grants_: t.Optional[exp.Tuple] = Field(default=None, alias="grants")
+    grants_target_layer: GrantsTargetLayer = GrantsTargetLayer.default
```
izeigerman, line 122: Do we really need this? Why not just apply it in both places unconditionally?

sqlmesh/core/model/meta.py
```diff
@@ -284,6 +327,14 @@ def _refs_validator(cls, vs: t.Any, info: ValidationInfo) -> t.List[exp.Expressi
     def ignored_rules_validator(cls, vs: t.Any) -> t.Any:
         return LinterConfig._validate_rules(vs)
 
+    @field_validator("grants_target_layer", mode="before")
+    def _grants_target_layer_validator(cls, v: t.Any) -> t.Any:
+        return cls._validate_str_enum_value(v)
+
+    @field_validator("virtual_environment_mode", mode="before")
```
izeigerman: `virtual_environment_mode` is not supposed to be set in the model definition, so we should have a validator for it.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2345,6 +2350,193 @@ def wap_publish(self, table_name: TableName, wap_id: str) -> None:
...
+        current_grants = self._get_current_grants_config(table)
+        new_grants, revoked_grants = self._diff_grants_configs(grants_config, current_grants)
+        revoke_exprs = self._revoke_grants_config_expr(table, revoked_grants, table_type)
+        grant_exprs = self._apply_grants_config_expr(table, new_grants, table_type)
+        dcl_exprs = revoke_exprs + grant_exprs
+
+        if dcl_exprs:
+            self.execute(dcl_exprs)
+
+    def _apply_grants_config(
```
izeigerman: Is this method used anywhere outside tests? If no, then I don't think we need it

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2345,6 +2350,193 @@ def wap_publish(self, table_name: TableName, wap_id: str) -> None:
...
+            table_type: The type of database object (TABLE, VIEW, MATERIALIZED_VIEW).
+
+        Raises:
+            NotImplementedError: If the engine does not support grants.
+        """
+
+        if grants := self._apply_grants_config_expr(table, grants_config, table_type):
+            self.execute(grants)
+
+    def _revoke_grants_config(
```
izeigerman: Same here: if we don't use it outside tests, let's remove it

sqlmesh/core/snapshot/evaluator.py
newtonapple: I was going to keep the adapter grants API (`_sync_grants_config`,` _apply_grants_config`, `_revoke_grants_config`) private in this implementation and eventually making them public once they're stabilized. 
`_apply_grants_config` & `_revoke_grants_config` were implemented for completeness before I implemented `_sync_grants_config` thinking I will use them in sync, but I ended up just batching all the expressions in one list and executing them in sync.  I can get rid of `_apply_grants_config` & `_revoke_grants_config` since they're not being used right now and make `sync_grants_config` public.

sqlmesh/core/model/meta.py
newtonapple: Do you mean the `grants_target_layer`?  I thought we wanted to give the user the ability to choose which layer(s) grants should be applied to from our previous discussion. But, I can sync up with you on this.

sqlmesh/core/model/meta.py
newtonapple: ~~Right now, it'll ignore it.  But I'll raise an exception if it's not an Array.  As far as the right-hand side expression, we don't do any special parsing / validation on them other than checking and making sure it's an Array.~~

I have actually expanded the grants_config [parsing logic](https://github.com/TobikoData/sqlmesh/pull/5275/commits/773c71f811aa0ddc3a08193cf035205df8c0bcc1) to be a more flexible.  Let me know what you think @izeigerman

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1871,6 +1928,10 @@ def _replace_query_for_model(
             source_columns=source_columns,
         )
 
+        # Apply grants after table replacement (unless explicitly skipped by caller)
+        if not kwargs.get("skip_grants", False):
```
izeigerman: Can we make this a top-level argument?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2139,16 +2208,34 @@ def create(
...
 
-        super().create(table_name, model, is_table_deployable, render_kwargs, **kwargs)
+        super().create(
+            table_name,
+            model,
+            is_table_deployable,
+            render_kwargs,
+            skip_grants=True,  # Skip grants; they're applied after data insertion
+            **kwargs,
+        )
```
eakmanrq, line 2403: I wonder if this create is even needed? We do the create a few lines down with the replace query. I wonder if we did this before to maintain the contract of creating empty table first but we don't do that anymore. I bring it up since I then think you could just have it grant and not have to do the skip grants here.

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2922,6 +2954,127 @@ def _check_identifier_length(self, expression: exp.Expression) -> None:
...
+    def _revoke_grants_config_expr(
+        self,
+        table: exp.Table,
+        grant_config: GrantsConfig,
+        table_type: DataObjectType = DataObjectType.TABLE,
+    ) -> t.List[exp.Expression]:
+        """Returns SQLGlot expressions to revoke grants from a table.
+
+        Note: SQLGlot doesn't yet have a Revoke expression type, so implementations
+        may return other expression types or handle revokes as strings.
```
eakmanrq: I believe this comment is no longer correct.

sqlmesh/core/model/meta.py
```diff
@@ -126,6 +161,14 @@ def _normalize(value: t.Any) -> t.Any:
 
         return v
 
+    @classmethod
+    def _validate_str_enum_value(cls, v: t.Any) -> t.Any:
```
izeigerman: dead code?

sqlmesh/core/model/meta.py
```diff
@@ -465,6 +520,67 @@ def custom_materialization_properties(self) -> CustomMaterializationProperties:
             return self.kind.materialization_properties
         return {}
 
+    @cached_property
+    def grants(self) -> t.Optional[GrantsConfig]:
```
izeigerman, line 516: Seems a little weird to go through the trouble of converting `t.Dict[str, t.List[str]]` into `exp.Tuple` just to convert it back again here.

sqlmesh/core/model/meta.py
```diff
@@ -465,6 +520,67 @@ def custom_materialization_properties(self) -> CustomMaterializationProperties:
...
+            if isinstance(expr, exp.Identifier):
+                return expr.name
+            if isinstance(expr, exp.Column):
+                return expr.name
+            return expr.sql(dialect=self.dialect).strip()
+
+        def normalize_to_string_list(value_expr: exp.Expression) -> t.List[str]:
+            result = []
+
+            def process_expression(expr: exp.Expression) -> None:
```
izeigerman: 3 level of nested function definitions is probably a bit too much. Any opportunity to move `exp_to_string` and `normalize_to_string_list` out?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1673,6 +1686,59 @@ def demote(self, view_name: str, **kwargs: t.Any) -> None:
...
+        if not self.adapter.SUPPORTS_GRANTS:
+            logger.warning(
+                f"Engine {self.adapter.__class__.__name__} does not support grants. "
+                f"Skipping grants application for model {model.name}"
+            )
+            return
+
+        model_grants_target_layer = model.grants_target_layer
+
+        is_prod_and_dev_only = is_snapshot_deployable and model.virtual_environment_mode.is_dev_only
```
izeigerman: the name is a bit misleading. Without reading the expression itself it reads as "Is Prod and Dev only?"

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1673,6 +1686,59 @@ def demote(self, view_name: str, **kwargs: t.Any) -> None:
...
+
+        if not (
```
izeigerman: To avoid a somewhat confusing inversion can we just put the grant application under the `if` and put the skip debug log message in `else`?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1777,6 +1844,16 @@ def promote(
             view_properties=model.render_virtual_properties(**render_kwargs),
         )
 
+        snapshot = kwargs["snapshot"]
+        deployability_index = kwargs["deployability_index"]
+        is_snapshot_deployable = deployability_index.is_deployable(snapshot)
+
+        # Apply grants to the physical layer (referenced table / view) after promotion
+        self._apply_grants(model, table_name, GrantsTargetLayer.PHYSICAL, is_snapshot_deployable)
```
izeigerman: Why are we applying the physical grants as part of promotion?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2239,9 +2362,17 @@ def create(
                 model,
                 is_table_deployable,
                 render_kwargs,
+                skip_grants,
                 **kwargs,
             )
 
+        if not skip_grants:
+            # Apply grants after SCD Type 2 table creation
```
izeigerman, line 2509: Don't we recreate a table for SCD 2 every time? Shouldn't we add grants on insert then too?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2381,12 +2516,21 @@ def create(
...
     ) -> None:
+        is_snapshot_deployable: bool = kwargs["is_snapshot_deployable"]
+
         if self.adapter.table_exists(table_name):
             # Make sure we don't recreate the view to prevent deletion of downstream views in engines with no late
             # binding support (because of DROP CASCADE).
             logger.info("View '%s' already exists", table_name)
+
+            if not skip_grants:
+                # Always apply grants when present, even if view exists, to handle grants updates
```
izeigerman, line 2673: Why isn't migration taking care of that?

sqlmesh/core/snapshot/evaluator.py
newtonapple: Added _apply_grants to SCD2 insert: https://github.com/TobikoData/sqlmesh/pull/5275/commits/90a6818e44774a0aa80b0ab7806277c61ca3ec45

sqlmesh/core/snapshot/evaluator.py
newtonapple: This is for when a plan apply fails for some external reasons.  I tested the case where I was applying grants without first creating a role, I then created the role in the db and re-applied thepaln, the view would exist, but not the grants.  As a result, grants were skipped.

sqlmesh/core/snapshot/evaluator.py
newtonapple: I think this was needed before this [fix](https://github.com/TobikoData/sqlmesh/pull/5389).  But, it didn't fully address the issue when VDE is `dev_only`.  This can be removed now.

sqlmesh/core/model/meta.py
newtonapple: I agree, it is a bit of a loop.  But, this similar to how `physical_properties`, `virtual_properties`, and `session_properties` work as well.  As far as I can tell, in order for `grants` to work both as a hash (set directly via python model: `model.grants = {'select': ['readonly_user']}`) or in a SQL model file (parsed as sqlglot expression), this is the only way to handle to it?

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -26,6 +26,7 @@ class BasePostgresEngineAdapter(EngineAdapter):
     COMMENT_CREATION_VIEW = CommentCreationView.COMMENT_COMMAND_ONLY
     SUPPORTS_QUERY_EXECUTION_TRACKING = True
     SUPPORTED_DROP_CASCADE_OBJECT_KINDS = ["SCHEMA", "TABLE", "VIEW"]
+    CURRENT_SCHEMA_EXPRESSION = exp.func("current_schema")
```
VaggelisD: Is this needed anywhere else? Could we inline it directly into `get_current_schema`?

sqlmesh/core/engine_adapter/base_postgres.py
```diff
@@ -188,3 +190,10 @@ def _get_data_objects(
             )
             for row in df.itertuples()
         ]
+
+    def get_current_schema(self) -> str:
+        """Returns the current default schema for the connection."""
+        result = self.fetchone(exp.select(self.CURRENT_SCHEMA_EXPRESSION))
+        if result and result[0]:
```
VaggelisD, line 204: [Nit] Since `fetchone` returns a tuple, `if result` should suffice to check that there's at least one element present, right?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -127,6 +135,118 @@ def snowpark(self) -> t.Optional[SnowparkSession]:
...
 
+    @staticmethod
+    def _grant_object_kind(table_type: DataObjectType) -> str:
+        if table_type == DataObjectType.VIEW:
+            return "VIEW"
+        if table_type == DataObjectType.MATERIALIZED_VIEW:
+            return "MATERIALIZED VIEW"
+        if table_type == DataObjectType.MANAGED_TABLE:
+            return "DYNAMIC TABLE"
+        return "TABLE"
```
VaggelisD, line 150: This function is repeated across DBX, Snowflake and Redshift afaict, should we move it to the base adapter?

sqlmesh/core/engine_adapter/snowflake.py
```diff
@@ -127,6 +135,118 @@ def snowpark(self) -> t.Optional[SnowparkSession]:
...
+
+    def _get_current_schema(self) -> str:
+        """Returns the current default schema for the connection."""
+        result = self.fetchone("SELECT CURRENT_SCHEMA()")
+        if not result or not result[0]:
```
VaggelisD, line 155: [Nit] Ditto about `if not result` covering both cases

sqlmesh/core/engine_adapter/base_postgres.py
newtonapple: It is possible for `SELECT current_schema` to return NULL according to the [documentation](https://www.postgresql.org/docs/current/functions-info.html). Specifically:

```
current_schema(): Returns the name of the schema that is first in the search path (or a null value if the search path is empty)
```

sqlmesh/core/engine_adapter/base_postgres.py
newtonapple: will inline this.

sqlmesh/core/engine_adapter/snowflake.py
newtonapple: This method is inherited from the `GrantsFromInfoSchemaMixin`.  Their implementation is not always the same as different engine DCL requires different mappings here.

sqlmesh/core/engine_adapter/snowflake.py
newtonapple: Same as previous comment, I think snowflake's `CURRENT_SCHEMA()` can return `NULL`.

sqlmesh/core/model/definition.py
```diff
@@ -1823,6 +1844,12 @@ def _data_hash_values_no_sql(self) -> t.List[str]:
         for column_name, column_hash in self.column_hashes.items():
             data.append(column_name)
             data.append(column_hash)
+
+        # Include grants in data hash for seed models to force recreation on grant changes
+        # since seed models don't support migration
+        data.append(json.dumps(self.grants, sort_keys=True) if self.grants else "")
```
izeigerman, line 1850: Should we do this only if grans are non empty? Otherwise this is a breaking change that requires migration.

izeigerman: which is fine, but we need to mark this as such

izeigerman: ah, I found the migration script

PR #5277, opened 2025-09-02
Feat!: Skip model evaluation if upstream external model(s) have not changed

VaggelisD:
This PR enables SQLMesh to skip re-evaluating models that depend on external model(s) which have not been updated since the last run.

This is done by:
1. Adding the `last_altered_ts` field in the `_intervals` table which tracks the last modification to the physical table of that snapshot
2. Querying the `INFORMATION_SCHEMA` for the external model's "last altered" timestamp wherever applicable (BQ & Snowflake for the time being)

Edit n.1: The freshness check was refactored into a built-in signal to (1) make the feature opt-in and (2) matches conceptually and the rest of the UX is given to us "for free".

### DIFF
--- a/sqlmesh/core/state_sync/db/facade.py
+++ b/sqlmesh/core/state_sync/db/facade.py
@@ -381,8 +381,9 @@ class EngineAdapterStateSync(StateSync):
         start: TimeLike,
         end: TimeLike,
         is_dev: bool = False,
+        last_altered_ts: t.Optional[int] = None,
     ) -> None:
-        super().add_interval(snapshot, start, end, is_dev)
+        super().add_interval(snapshot, start, end, is_dev, last_altered_ts)
 
     @transactional()
     def add_snapshots_intervals(self, snapshots_intervals: t.Sequence[SnapshotIntervals]) -> None:


sqlmesh/core/snapshot/definition.py
```diff
@@ -690,6 +694,12 @@ def hydrate_with_intervals_by_version(
             )
             for interval in snapshot_intervals:
                 snapshot.merge_intervals(interval)
+
+                if interval.last_altered_ts:
```
izeigerman: I don't think this is correct. I don't see us distinguishing between dev and non-dev intervals. For example, adding an interval for a dev table should not have impact on `last_altered_ts` used for the non-dev / prod table.

VaggelisD: Discussed a bit async, the function here hydrates snapshots by `(name, version)`. This is not entirely correct because the `version` may be reused e.g between a prod vs a forward-only dev instance. 

Thus, I think the solution here is to extend this branch to also filter by `dev_version`, as we do in `merge_intervals` for `dev_intervals`. 

I've also added a test case to validate that scenario.

.circleci/continue_config.yml
```diff
@@ -310,10 +310,10 @@ workflows:
...
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
+          #   branches:
+          #     only:
+          #       - main
```
VaggelisD: TODO: Revert

sqlmesh/core/engine_adapter/base.py
```diff
@@ -2927,6 +2928,9 @@ def _check_identifier_length(self, expression: exp.Expression) -> None:
                     f"Identifier name '{name}' (length {name_length}) exceeds {self.dialect.capitalize()}'s max identifier limit of {self.MAX_IDENTIFIER_LENGTH} characters"
                 )
 
+    def get_external_model_freshness(self, table_names: t.List[TableName]) -> t.List[int]:
```
izeigerman: I don't think the engine adapter should be aware of high-level concepts like "model" and "freshness". This should just be `get_table_last_modified_ts`. I'd also suggest returning a dict as to not implicitly rely on the order of the list.

sqlmesh/core/scheduler.py
```diff
@@ -188,6 +190,46 @@ def merged_missing_intervals(
             }
         return snapshots_to_intervals
 
+    def can_skip_evaluation(self, snapshot: Snapshot, snapshots: t.Dict[str, Snapshot]) -> bool:
```
izeigerman: Is this dead code?

sqlmesh/core/scheduler.py
```diff
@@ -934,13 +987,27 @@ def _check_ready_intervals(
         if not (signals and signals.signals_to_kwargs):
             return intervals
 
+        signal_names = signals.signals_to_kwargs.keys()
+
+        if (
```
izeigerman: This check is quite brittle. What if there are multiple signals one of which is `freshness`?

izeigerman: Can we somehow determine this based on the target interval start in the signal itself?

sqlmesh/core/scheduler.py
```diff
@@ -913,6 +965,7 @@ def _check_ready_intervals(
         intervals: Intervals,
         context: ExecutionContext,
         environment_naming_info: EnvironmentNamingInfo,
+        is_restatement_plan: bool = False,
```
izeigerman: The scheduler should not be aware of plans

sqlmesh/core/scheduler.py
```diff
@@ -934,13 +987,27 @@ def _check_ready_intervals(
...
+            return intervals
+
         self.console.start_signal_progress(
             snapshot,
             self.default_catalog,
             environment_naming_info or EnvironmentNamingInfo(),
         )
 
         for signal_idx, (signal_name, kwargs) in enumerate(signals.signals_to_kwargs.items()):
+            if is_restatement_plan and signal_name == "freshness":
```
izeigerman: ah, I see that. I'd prefer finding a way to determine this in the signal itself

sqlmesh/core/engine_adapter/base.py
```diff
@@ -119,6 +119,7 @@ class EngineAdapter:
     MAX_IDENTIFIER_LENGTH: t.Optional[int] = None
     ATTACH_CORRELATION_ID = True
     SUPPORTS_QUERY_EXECUTION_TRACKING = False
+    SUPPORTS_EXTERNAL_MODEL_FRESHNESS = False
```
izeigerman: Ditto: the name should not refer to neither "models" nor "freshess".

sqlmesh/core/scheduler.py
izeigerman: Maybe we can somehow extend `ExecutionContext` with the restatement information?

sqlmesh/core/snapshot/definition.py
```diff
@@ -205,6 +207,12 @@ def add_dev_interval(self, start: int, end: int) -> None:
     def add_pending_restatement_interval(self, start: int, end: int) -> None:
         self._add_interval(start, end, "pending_restatement_intervals")
 
+    def add_last_altered_ts(self, last_altered_ts: t.Optional[int]) -> None:
```
izeigerman: `add` assumes adding to a collection. As far as I can tell we're setting and not adding

sqlmesh/core/snapshot/definition.py
```diff
@@ -957,12 +977,20 @@ def merge_intervals(self, other: t.Union[Snapshot, SnapshotIntervals]) -> None:
             if not apply_effective_from or end <= effective_from_ts:
                 self.add_interval(start, end)
 
+        if other.last_altered_ts:
+            self.last_altered_ts = max(self.last_altered_ts or -1, other.last_altered_ts)
```
izeigerman: `-1` seems like a weird choice for a default here. SHouldn't this be `0`?

sqlmesh/core/signal.py
```diff
@@ -33,3 +38,39 @@ class signal(registry_decorator):
 
 
 SignalRegistry = UniqueKeyDict[str, signal]
+
+
+@signal()
+def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionContext) -> bool:
```
izeigerman, line 45: How do we expect users to use this? Do they need to manually opt in? Can they opt-in globally via a config?

sqlmesh/core/engine_adapter/base.py
VaggelisD: > I don't think the engine adapter should be aware of high-level concepts like "model" and "freshness". This should just be get_table_last_modif

Sounds good, I renamed:
- The adapter flag `SUPPORTS_EXTERNAL_MODEL_FRESHNESS` --> `SUPPORTS_METADATA_TABLE_LAST_MODIFIED_TS`
- The adapter function `get_external_model_freshness ` --> `get_table_last_modified_ts`

sqlmesh/core/scheduler.py
VaggelisD: Yes, I thought I had removed it in the transition towards signals..

sqlmesh/core/engine_adapter/base.py
VaggelisD: Renamed to `SUPPORTS_METADATA_TABLE_LAST_MODIFIED_TS` as per https://github.com/TobikoData/sqlmesh/pull/5277#discussion_r2378354656

sqlmesh/core/snapshot/definition.py
VaggelisD: Renamed to `update_last_altered_ts`

sqlmesh/core/snapshot/definition.py
VaggelisD: Yeah can do, thought technically `0` is a valid ts (epoch start) whereas a negative number is not, but it doesn't make much sense since `other.last_altered_ts` is expected to be a valid ts.

sqlmesh/core/signal.py
VaggelisD: I placed it in `signal.py` so it's loaded in the signal registry along with the module. For this PR, the signal is opt-in and must be defined at each `MODEL` block, e.g:

```Python3
MODEL (
  name ...,
  signal (
    freshness(), ...,
  )
)

....
```

Once the feature is ready i.e freshness works for mixed workloads (sources & models), I intend to work on its UX such as:
- Adding more knobs to enable/disable globally
- Reworking its `Console` output since the "default" experience for signals produces too much noise with interval batches, and that is orthogonal to freshness.

sqlmesh/core/scheduler.py
VaggelisD: I performed this filtering at this point to avoid introducing noise to the output; Moving that check to the signal itself is semantically equivalent, but now the plans are riddled with:

```
Checking signals for TEST_SCHEMA_JV8GVW2J__DEV.NEW_MODEL
└── [1/1] freshness 0.00s
    ├── Check: 2025-09-24 - 2025-09-24
    └── All ready: 2025-09-24 - 2025-09-24
```

Maybe this is fine for the time being and I can fix that on a subsequent PR that reworks freshness's UX?

sqlmesh/core/scheduler.py
VaggelisD: > Maybe we can somehow extend ExecutionContext with the restatement information?

Went with that approach. Sadly afaict `Scheduler` needs to propagate that boolean since it constructs `ExecutionContext` internally.

sqlmesh/core/engine_adapter/base.py
VaggelisD: > I'd also suggest returning a dict as to not implicitly rely on the order of the list.

I see your point, the order is non-deterministic, but the reason I didn't enforce it is because freshness doesn't really need it so it simplified the implementation. What do you think, should we still implement the mapping?

sqlmesh/core/scheduler.py
VaggelisD: Replied in https://github.com/TobikoData/sqlmesh/pull/5277#discussion_r2378428707 as well; The `Scheduler` actually creates `ExecutionContext`, checks the signals etc so I think it needs the restatement information.

izeigerman: ok, looks like we already pass `restatements` into the scheduler. Can we at least call this `is_restatement` without `plan`. Here and everywhere

izeigerman: btw, I don't believe this argument is needed in `_check_ready_intervals` any more

sqlmesh/core/scheduler.py
```diff
@@ -200,6 +200,7 @@ def evaluate(
         allow_destructive_snapshots: t.Optional[t.Set[str]] = None,
         allow_additive_snapshots: t.Optional[t.Set[str]] = None,
         target_table_exists: t.Optional[bool] = None,
+        is_restatement_plan: bool = False,
```
izeigerman: not needed here

sqlmesh/core/scheduler.py
```diff
@@ -367,13 +371,15 @@ def batch_intervals(
...
                 default_catalog=self.default_catalog,
+                is_restatement_plan=is_restatement_plan,
             )
 
             intervals = self._check_ready_intervals(
                 snapshot,
                 intervals,
                 context,
                 environment_naming_info,
+                is_restatement_plan=is_restatement_plan,
```
izeigerman: not needed

sqlmesh/core/scheduler.py
```diff
@@ -542,6 +552,7 @@ def run_node(node: SchedulingUnit) -> None:
                             allow_additive_snapshots=allow_additive_snapshots,
                             target_table_exists=snapshot.snapshot_id not in snapshots_to_create,
                             selected_models=selected_models,
+                            is_restatement_plan=is_restatement_plan,
```
izeigerman: ditto

sqlmesh/core/scheduler.py
```diff
@@ -934,6 +946,8 @@ def _check_ready_intervals(
         if not (signals and signals.signals_to_kwargs):
             return intervals
 
+        signal_names = signals.signals_to_kwargs.keys()
```
izeigerman: dead code?

sqlmesh/core/signal.py
```diff
@@ -33,3 +38,42 @@ class signal(registry_decorator):
...
+    if context.is_restatement_plan:
+        return True
+
+    deployability_index = context.deployability_index
+    adapter = context.engine_adapter
+
+    if not deployability_index or not adapter.SUPPORTS_METADATA_TABLE_LAST_MODIFIED_TS:
```
izeigerman: should just be `deployability_index = context.deployability_index or DeployabilityIndex.all_deployable()`

sqlmesh/core/signal.py
```diff
@@ -33,3 +38,42 @@ class signal(registry_decorator):
...
+    last_altered_ts = (
+        snapshot.last_altered_ts
+        if deployability_index.is_deployable(snapshot)
+        else snapshot.dev_last_altered_ts
+    )
+    if not last_altered_ts:
+        return True
+
+    parent_snapshots = {context.snapshots[p.name] for p in snapshot.parents}
+    if len(parent_snapshots) != len(snapshot.node.depends_on) or not all(
```
izeigerman, line 61: I don't follow what this condition is for

sqlmesh/core/signal.py
```diff
@@ -33,3 +38,42 @@ class signal(registry_decorator):
...
+    ):
+        # The mismatch can happen if e.g an external model is not registered in the project
+        return True
+
+    # Finding new data means that the upstream depedencies have been altered
+    # since the last time the model was evaluated
+    upstream_dep_has_new_data = any(
+        upstream_last_altered_ts > last_altered_ts
+        for upstream_last_altered_ts in adapter.get_table_last_modified_ts(
+            [p.name for p in parent_snapshots]
```
izeigerman, line 72: This looks wrong:
1. Why do we check tables of non-external model snapshots? Shouldn't we rely on their `last_altered_ts` instead?
2. Why are we not using the physical table name?

sqlmesh/core/signal.py
VaggelisD: I think I had placed it here to denote that all external models that this model depends on should be registered in `external_models.yaml`; That was to ensure that their gateway is set properly and so on.

Should we allow _any_ external model here?

sqlmesh/core/signal.py
VaggelisD: This PR enables `freshness` only if **all** the parent models are external, as per the last condition. In the following PR I'll enable mixed external & managed models, so `upstream_last_altered_ts` would come from either:
- External models: Will query their table name directly
- Managed models: Will use their `last_altered_ts`

sqlmesh/core/signal.py
izeigerman: I believe we shouldn't assume that external models will be registered at all. A more typical scenario is no registered external models

izeigerman: Nvm, your comment below provided more context

sqlmesh/core/signal.py
izeigerman: Ok, I see. Thanks

PR #5279, opened 2025-09-02
feat: trino support replace table iceberg and delta

eakmanrq:
Prior to this PR we assumed all Trino catalogs didn't support replace table but this isn't true for Iceberg and Delta Lake. Therefore this PR updates Trino to use replace table for those catalogs.

Some other key changes:
* Users may have a forked custom connector instead of the open source connector and if they do this then it will have a new connector name. This PR updates the catalog name check to check if "iceberg" or "delta_lake" are in the name. That isn't perfect but will help cover more cases.
* The engine adapter methods did `current_catalog_type` but this just looks at the current default catalog for the connection. We should be looking at the target table of the operation and it's catalog if it exists and then fallback to the connection catalog if it does not. 
   * `current_catalog_type` still exists just to support truncating comment methods. They should also be updated to use the target table/catalog but that change was proving to be a larger refactor and determined to be out of scope of this PR. 
* Tests were using a catalog type of `delta` when it should be `delta_lake`. I base this on documentation and our own integration tests
   * https://trino.io/docs/current/connector/delta-lake.html#general-configuration
   * https://github.com/TobikoData/sqlmesh/blob/e8576e9aab71d3237a80b360d8ae1ed417849540/tests/core/engine_adapter/integration/docker/trino/catalog/datalake_delta.properties#L1

### DIFF
--- a/sqlmesh/core/engine_adapter/trino.py
+++ b/sqlmesh/core/engine_adapter/trino.py
@@ -34,6 +34,8 @@ if t.TYPE_CHECKING:
     from sqlmesh.core._typing import SchemaName, SessionProperties, TableName
     from sqlmesh.core.engine_adapter._typing import DF, QueryOrDF
 
+CATALOG_TYPES_SUPPORTING_REPLACE_TABLE = {"iceberg", "delta_lake"}
+
 
 @set_catalog()
 class TrinoEngineAdapter(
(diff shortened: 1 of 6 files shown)


treysp (approved):

tests/core/engine_adapter/test_trino.py
```diff
@@ -64,13 +64,14 @@ def test_get_catalog_type(
...
-    assert adapter.get_catalog_type("datalake_delta") == "delta"
+    assert adapter.get_catalog_type("datalake_delta_lake") == "delta_lake"
 
     mocker.patch(
         "sqlmesh.core.engine_adapter.trino.TrinoEngineAdapter.get_current_catalog",
         return_value=f"system_{storage_type}",
     )
-    assert adapter.current_catalog_type == storage_type
+    expected_current_type = storage_type
+    assert adapter.current_catalog_type == expected_current_type
```
erindru, line 74: Does this change do anything?

tests/core/engine_adapter/test_trino.py
```diff
@@ -113,7 +114,8 @@ def test_partitioned_by_hive_delta(
         "sqlmesh.core.engine_adapter.trino.TrinoEngineAdapter.get_current_catalog",
         return_value=f"datalake_{storage_type}",
     )
-    assert adapter.get_catalog_type(f"datalake_{storage_type}") == storage_type
+    expected_type = storage_type
+    assert adapter.get_catalog_type(f"datalake_{storage_type}") == expected_type
```
erindru, line 118: Same with this one?

sqlmesh/core/engine_adapter/trino.py
```diff
@@ -115,6 +117,37 @@ def session(self, properties: SessionProperties) -> t.Iterator[None]:
...
+        table_name: TableName,
+        query_or_df: QueryOrDF,
+        target_columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        source_columns: t.Optional[t.List[str]] = None,
+        supports_replace_table_override: t.Optional[bool] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        catalog_type = self.get_catalog_type(self.get_catalog_type_from_table(table_name))
```
erindru, line 131: Doesn't `self.get_catalog_type_from_table(table_name)` already call `self.get_catalog_type()`?

So this would do:

`self.get_catalog_type_from_table(table_name)` -> may return `"delta_lake"`

Then it would call:

`self.get_catalog_type("delta_lake")`

There is probably no catalog called `delta_lake` as it's a trino _*connector*_ type (which is what we call "catalog type" here), so this would fall back on the connector type of the default catalog and not that of the table.

I think this "worked" in the test you wrote because for unit testing purposes we infer the catalog type from the catalog name, but this doesn't happen on a running cluster.

The catalog can be named anything you want and the catalog _type_ comes from whatever was configured for the `connector.name` property for that catalog in the Trino config

PR #5280, opened 2025-09-02
Fix: Ignore bigquery partition_by config when the target isn't bigquery

crericha:
dbt ignores bigquery partition_by config when not using the bigquery plugin. Have sqlmesh mimic this behavior.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -529,7 +529,9 @@ class ModelConfig(BaseModelConfig):
                     partitioned_by.append(self._big_query_partition_by_expr(context))
                 else:
                     logger.warning(
-                        "Ignoring partition_by config for model '%s' targeting %s; it is only supported for BigQuery.", self.name, context.target.dialect
+                        "Ignoring partition_by config for model '%s' targeting %s. The format of the config field is only supported for BigQuery.",
+                        self.name,
+                        context.target.dialect,
                     )
 
             if partitioned_by:
@@ -538,7 +540,8 @@ class ModelConfig(BaseModelConfig):
         if self.cluster_by:
             if isinstance(kind, ViewKind):
                 logger.warning(
-                    f"Ignoring cluster_by config for model '{self.name}'; cluster_by is not supported for views."
+                    "Ignoring cluster_by config for model '%s'; cluster_by is not supported for views.",
+                    self.name,
                 )
             else:
                 clustered_by = []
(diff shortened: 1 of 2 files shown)


erindru:
Can you add a test to validate this behaviour?

sqlmesh/dbt/model.py
```diff
@@ -517,9 +521,16 @@ def to_sqlmesh(
...
                         ) from e
-            else:
-                partitioned_by.append(self._big_query_partition_by_expr(context))
-            optional_kwargs["partitioned_by"] = partitioned_by
+            elif isinstance(self.partition_by, dict):
+                if context.target.dialect == "bigquery":
+                    partitioned_by.append(self._big_query_partition_by_expr(context))
+                else:
+                    logger.warning(
+                        f"Ignoring partition_by config for model '{self.name}' targeting {context.target.dialect}; it is only supported for bigquery."
```
izeigerman:
```suggestion
                        "Ignoring partition_by config for model '%s' with target %s; it is only supported for BigQuery.", self.name, context.target.dialect
```
please don't use string interpolation when logging

izeigerman: actually, the message is confusing. Partitioned by is not support for BigQuery only. This specific format is only support for BigQuery

sqlmesh/dbt/model.py
```diff
@@ -517,9 +521,16 @@ def to_sqlmesh(
                         raise ConfigError(
                             f"Failed to parse model '{self.canonical_name(context)}' partition_by field '{p}' in '{self.path}': {e}"
                         ) from e
-            else:
-                partitioned_by.append(self._big_query_partition_by_expr(context))
-            optional_kwargs["partitioned_by"] = partitioned_by
+            elif isinstance(self.partition_by, dict):
```
izeigerman, line 527: what if neither dict nor list?

sqlmesh/dbt/model.py
```diff
@@ -517,9 +521,16 @@ def to_sqlmesh(
...
+                if context.target.dialect == "bigquery":
+                    partitioned_by.append(self._big_query_partition_by_expr(context))
+                else:
+                    logger.warning(
+                        "Ignoring partition_by config for model '%s' targeting %s; it is only supported for BigQuery.", self.name, context.target.dialect
```
izeigerman: In case you missed my follow-up comment:
> actually, the message is confusing. Partitioned by is not support for BigQuery only. This specific format is only support for BigQuery

sqlmesh/dbt/model.py
crericha: The validator ensures its one or the other. 
https://github.com/TobikoData/sqlmesh/blob/d3079815cff6c73fe99f76c719967e83394d4258/sqlmesh/dbt/model.py#L168

PR #5285, opened 2025-09-03
Feat: prevent other processes seeing missing intervals during restatement

erindru:
This PR builds on #5273 and #5274

Currently, the `RestatementStage` clears intervals from state _before_ the `BackfillStage` populates data.

This means that other processes that look at state while the restatement is running will see missing intervals and try to fill them, competing with the restatement process.

This PR adjusts the plan evaluation order to:
 - move `RestatementStage` to after `BackfillStage`
     - note that despite the name, `RestatementStage` doesn't actually perform restatement, it's just responsible for clearing intervals from state
 - adjust `RestatementStage` to clear intervals from all environments _except_ prod

The result of this means that:
 - Intervals are never cleared from `prod` so no other processes see missing intervals and compete with the current plan to try and fill them
 - Intervals are only cleared from dev environments if they were successfully restated in prod

There is a new failure mode:
 - If the plan fails during restatement, `prod` will be in an inconsistent state. It's expected that the user will run the plan again until it succeeds. Once it succeeds, `prod` will be back in a consistent state.
 
**STILL TODO:**
  - There needs to be a final check in `RestatementStage` that verifies new versions of the prod snapshots being restated have not been deployed to prod while the restatement plan was running


sqlmesh/core/plan/stages.py
```diff
@@ -219,15 +228,17 @@ class PlanStagesBuilder:
...
 
     def __init__(
-        self,
-        state_reader: StateReader,
-        default_catalog: t.Optional[str],
+        self, state_reader: StateReader, default_catalog: t.Optional[str], explain: bool = False
     ):
         self.state_reader = state_reader
         self.default_catalog = default_catalog
+        self.explain = explain
```
izeigerman: oh no, let's not do this. Stages should be completely independent from how they are interpreted downstream. That's the whole point. We don't want to have diverging paths between the explainer and the actual evaluation.

sqlmesh/core/plan/stages.py
```diff
@@ -98,16 +101,22 @@ class AuditOnlyRunStage:
...
+            expaining exactly which physical tables are affected. Note that these snapshots are from
+            other environments that are not present in this plan, so we cant use the plan DeployabilityIndex
         all_snapshots: All snapshots in the plan by name.
     """
 
-    snapshot_intervals: t.Dict[SnapshotTableInfo, Interval]
     all_snapshots: t.Dict[str, Snapshot]
 
+    # Only used for --explain so may not be populated
+    snapshot_intervals_to_clear: t.Optional[t.Dict[str, SnapshotIntervalClearRequest]]
```
izeigerman: If this is only used for explainer, it might as well use `plan.restatements` / `plan.deployability_index` directly

sqlmesh/core/plan/stages.py
erindru: Well, I just wanted to attach some extra metadata for `--explain` that isnt needed by the evaluator because the evaluator re-calculates it at runtime in order to capture any drift that may have occurred while the restatements were happening,

If the metadata is not conditionally attached then we incur the overhead of computing it on every plan, even if it's not needed, right?

The goal is to **enable** CLI output like the following (in an upcoming PR) when someone plans with `--explain`:
Screenshot From 2025-09-04 08-56-44

because right now, the restatements CLI output is a bit confusing and misleading:
Screenshot From 2025-09-04 09-05-34

sqlmesh/core/plan/stages.py
erindru: It contains snapshots that aren't part the plan though (since they exist in other environments and are just getting intervals cleared / not evaluated), so `plan.restatements` isn't the full picture and thus neither is `plan.deployability_index` because these both only include snapshots that are being specifically evaluated in the plan.

I was attaching this in the stage builder and not in the explainer because the explainer console doesnt have a reference to the state sync or the current plan, perhaps I should pass these to the explainer console so it can look up extra data it needs to explain something?

sqlmesh/core/plan/stages.py
izeigerman: Then we should have a common logic between explainer and the evaluator to compute this information, and then call it in the explainer itself. But it shouldn't happen in the stage builder.

erindru: Ok, i've updated the explainer to take a `RestatementStage` and turn it into an `ExplainableRestatementStage` which contains the fields needed for the extended console output

sqlmesh/core/plan/stages.py
izeigerman: > I was attaching this in the stage builder and not in the explainer because the explainer console doesnt have a reference to the state sync or the current plan, perhaps I should pass these to the explainer console so it can look up extra data it needs to explain something?

Yeah, that's fine. Alternatively you can fetch in the explainer itself and then pass the result into the console directly.

erindru: Ok, PR has been updated to make the explainer turn `RestatementStage` into `ExplainableRestatementStage` (which has the extra fields) and then the console can utilize them

sqlmesh/core/plan/evaluator.py
```diff
@@ -284,32 +284,62 @@ def visit_audit_only_run_stage(
...
+            current_ts=to_timestamp(plan.execution_time or now()),
         )
 
+        if not intervals_to_clear:
+            # Nothing to do
+            return
+
         self.state_sync.remove_intervals(
-            snapshot_intervals=list(snapshot_intervals_to_restate),
+            snapshot_intervals=[(s.table_info, s.interval) for s in intervals_to_clear.values()],
```
izeigerman: I believe we should **not** nuke intervals for snapshots that have changed while the restatement was running. As per our discussion earlier:

> There’s a chance that another snapshot sneaks into prod while restatement was running but before we nuked the intervals.  This snapshot’s intervals will be nuked, but now we have an intermittent period during which the prod is in incorrect state. The subsequent run will, of course, catch up the missing intervals. However, this is confusing to the user since their next run will now backfill models it wasn’t suppose to. To prevent this, the restatement plan will now check whether the snapshots for restated models changed in prod while this plan was running. If some snapshots were changed, the intervals for those won’t be removed and instead the plan will return an error stating that the restatement for these models wasn’t successful. The user will then have a choice whether to reissue restatement for these models or accept that these models haven’t been restated properly and move on.

Specifically
>  If some snapshots were changed, the intervals for those won’t be removed

sqlmesh/core/plan/evaluator.py
```diff
@@ -284,32 +284,62 @@ def visit_audit_only_run_stage(
...
+                version = snapshot.table_info.version
+                if (
+                    prod_snapshot := promoted_snapshots_by_name.get(name)
+                ) and prod_snapshot.version != version:
+                    deployed_during_restatement.append(
+                        (snapshot.table_info, prod_snapshot.table_info)
+                    )
+
+            if deployed_during_restatement:
+                self.console.log_models_updated_during_restatement(
```
izeigerman: Shouldn't we fail the plan at this point instead of just logging?

sqlmesh/core/plan/evaluator.py
```diff
@@ -284,32 +284,62 @@ def visit_audit_only_run_stage(
...
+                    deployed_during_restatement,
+                    deployed_env.summary,
+                    plan.environment.naming_info,
+                    self.default_catalog,
+                )
+                # note: the plan will automatically fail at the promotion stage with a ConflictingPlanError because the environment was changed by another plan
```
izeigerman: I see. Please note that this is not always going to be the case since we plan to support concurrent plan applications soon. So I'd suggest we fail explicitly and not rely on a failure downstream.

sqlmesh/core/plan/explainer.py
```diff
@@ -54,6 +71,61 @@ def explain(self, stages: t.List[stages.PlanStage]) -> None:
...
+        if all_restatement_intervals:
+            snapshot_intervals_to_clear = {
+                s_id.name: r for s_id, r in all_restatement_intervals.items()
+            }
+
+            # creating a deployability index over the "snapshot intervals to clear"
+            # allows us to print the physical names of the tables affected in the console output
+            # note that we can't use the DeployabilityIndex on the plan because it only includes
+            # snapshots for the current environment, not across all environments
+            deployability_index = DeployabilityIndex.create(
```
izeigerman: We should only do this if the target environment is not prod. 

Actually, I believe we should always use `DeployabilityIndex.all_deployable()` since "restatement" doesn't apply to "dev" tables. For "dev" tables it's always a preview.

sqlmesh/core/plan/evaluator.py
izeigerman: Additionally, we won't need `log_models_updated_during_restatement` since we can just put the entire message into exception.

sqlmesh/core/plan/explainer.py
erindru: `RestatementStage` isnt even produced if the target is not prod, because as of this PR it only refers to clearing intervals any environment that _isnt_ prod, which is only necessary if the target _is_ prod.

I think I see what you're saying - a prod restatement would never care about dev previews because that data has no chance of ever being deployed. So there is no need to build a deployability index, and in the console output we should just reference the deployable version of the model's physical table, which we can do with `DeployabilityIndex.all_deployable()`

sqlmesh/core/plan/evaluator.py
erindru: Ah, good point, I glossed over that when I was focusing on identifying the affected snapshots to begin with.

I've updated it to only clear intervals for snapshots whose names are not in the "deployed during restatement" list

sqlmesh/core/plan/evaluator.py
erindru: I think we still need `log_models_updated_during_restatement` to have useful formatted console output about the affected models rather than trying to jam all this info into an exception message.

However, i've moved the "what to do about it" part into the exception message and now raise an exception here instead of relying on the finalize stage throwing the error.

I've updated the screenshot in the PR description to give an indication of what it looks like

sqlmesh/core/plan/evaluator.py
erindru: See below

sqlmesh/core/snapshot/definition.py
```diff
@@ -1789,7 +1789,19 @@ def display_name(
...
     if snapshot_info_like.is_audit:
         return snapshot_info_like.name
-    view_name = exp.to_table(snapshot_info_like.name)
+
+    return model_display_name(
+        snapshot_info_like.name, environment_naming_info, default_catalog, dialect
+    )
+
+
+def model_display_name(
```
erindru, line 1807: I added this because I wanted to output model display names based on `SnapshotIdAndVersion` and the existing `display_name` function required an _entire_ `SnapshotTableInfo` object just to read the node type and name.

If we already know it's a model then all we need to produce a display name is the snapshot `name` string, so `model_display_name` reduces the requirement to just this.

`SnapshotIdLike` would work as well in order to help narrow the inputs to still be related to `Snapshots` rather than any random `str`; let me know if this is preferred

sqlmesh/core/plan/evaluator.py
```diff
@@ -287,34 +287,78 @@ def visit_audit_only_run_stage(
...
+
+        if filtered_intervals_to_clear:
+            # We still clear intervals in other envs for models that were successfully restated without having new versions promoted during restatement
+            self.state_sync.remove_intervals(
+                snapshot_intervals=filtered_intervals_to_clear,
+                remove_shared_versions=plan.is_prod,
+            )
+
+        if deployed_env and deployed_during_restatement:
+            self.console.log_models_updated_during_restatement(
```
izeigerman, line 352: Why do we still needs this? Can't we include this in the error message instead?

izeigerman: ok, I saw your reasoning in the other thread

sqlmesh/core/plan/evaluator.py
```diff
@@ -287,34 +287,78 @@ def visit_audit_only_run_stage(
...
+
+        if deployed_env := self.state_sync.get_environment(plan.environment.name):
+            promoted_snapshots_by_name = {s.name: s for s in deployed_env.snapshots}
+
+            for name in plan.restatements:
+                snapshot = stage.all_snapshots[name]
+                version = snapshot.table_info.version
+                if (
+                    prod_snapshot := promoted_snapshots_by_name.get(name)
+                ) and prod_snapshot.version != version:
```
izeigerman, line 328: Isn't it sufficient to just check the version? I'm sure the version will match if the snapshots match

sqlmesh/core/snapshot/definition.py
izeigerman: Should we just add `display_name` to `SnapshotIdAndVersion` for consistency? Then you no longer need to do `isinstance` check

sqlmesh/core/plan/explainer.py
```diff
@@ -265,15 +323,22 @@ def visit_finalize_environment_stage(
...
+            default_catalog=self.default_catalog
+            if self.verbosity < Verbosity.VERY_VERBOSE
+            else None,
             dialect=self.dialect,
         )
 
+        if isinstance(snapshot, SnapshotInfoMixin):
+            return snapshot.display_name(**naming_kwargs)
+
+        return model_display_name(node_name=snapshot.name, **naming_kwargs)
```
izeigerman: See https://github.com/TobikoData/sqlmesh/pull/5285/files#r2342240695

sqlmesh/core/plan/evaluator.py
erindru: Resolved internally, this is checking the version, `:=` just got misread as `=`

sqlmesh/core/snapshot/definition.py
erindru: I didnt do that initially because `SnapshotIdAndVersion` didnt have enough information to implement  `SnapshotInfoMixin` and I didn't want to duplicate the method signature for `display_name`.

However, i've added it and I guess mypy will keep the method signatures in sync

PR #5292, opened 2025-09-03
feat: dbt microbatch ref filter support

eakmanrq:
One behavior difference between dbt microbatch and SQLMesh incremental by time range is that dbt microbatch, by default, includes the date filter for you. You can opt-out of this by adding a `.render()` after the ref. This PR adds that functionality.

Some tricky parts of this:
* To add the filter we need to know the event column and the start/end dates. All of that is known at load time for dbt but we only know event column at load time in SQLMesh. Therefore preparing the `EventTimeFilter` happens in two stages on SQLMesh - one when we load to record the event column and then at execution we add the start/end
* dbt uses end as exclusive and therefore doesn't use `BETWEEN` in the expression. Therefore the end is made exclusive. An alternative approach is to patch the filter render method to have it use between which you can see that implementation in this commit: https://github.com/TobikoData/sqlmesh/pull/5292/commits/a5957914931123b7b69275b1baa5249e1545d733

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -214,6 +214,13 @@ class BaseExpressionRenderer:
                         dialect=self._dialect, identify=True, comments=False
                     )
 
+                all_refs = list(self._jinja_macro_registry.global_objs.get("sources", {}).values()) + list(  # type: ignore
+                    self._jinja_macro_registry.global_objs.get("refs", {}).values()  # type: ignore
+                )
+                for ref in all_refs:
+                    if ref.event_time_filter:
+                        ref.event_time_filter["start"] = render_kwargs["start_tstz"]
+                        ref.event_time_filter["end"] = render_kwargs["end_tstz"]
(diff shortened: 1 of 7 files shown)


sqlmesh/core/renderer.py
```diff
@@ -207,6 +214,15 @@ def _resolve_table(table: str | exp.Table) -> str:
                 dialect=self._dialect, identify=True, comments=False
             )
 
+        all_refs = list(self._jinja_macro_registry.global_objs.get("sources", {}).values()) + list(  # type: ignore
+            self._jinja_macro_registry.global_objs.get("refs", {}).values()  # type: ignore
+        )
+        for ref in all_refs:
+            if ref.event_time_filter:
+                ref.event_time_filter["start"] = render_kwargs["start_tstz"]
```
eakmanrq: I use `tstz` since my testing with Postgres showed that dbt rendered a tstz. It might end up being adapter specific so this could change after more testing.

erindru (approved):

PR #5293, opened 2025-09-03
Chore: improve test stability and address some warnings

erindru:
After spending an ungodly amount of time trying to work around `pytest-retry`'s design choices, I decided to revisit why it was chosen in the first place given its not even mentioned in the [pytest docs](https://docs.pytest.org/en/stable/explanation/flaky.html#plugins).

I recall that it was chosen because it supported Python 3.8 and `pytest-rerunfailures` was Python 3.9+. However, since then, SQLMesh no longer supports Python 3.8, so in this PR i've switched to use `pytest-rerunfailures` and everything suddenly started working correctly.

I also cleaned up a bunch of warnings like the following:

> PytestCollectionWarning: cannot collect test class 'TestConfig' because it has a __init__ constructor (from: tests/dbt/test_model.py)

and

> PytestUnknownMarkWarning: Unknown pytest.mark.registry_isolation - is this a typo?  You can register custom marks to avoid this warning


.circleci/continue_config.yml
```diff
@@ -310,10 +310,10 @@ workflows:
                 - athena
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          #filters:
```
erindru: TODO: revert

georgesittas (approved):

PR #5294, opened 2025-09-04
Adding in NoMissingUnitTest linter rule

cmgoffena13:
I'm having trouble fleshing out the test due to the path having to point to the example project. Since I have to trigger the pytest from the root it is trying to find the /tests directory in the root rather than the example project root. Any help on resolving the path issue would be appreciated so I can finish the test. (EDIT: Figured out how to switch the contexts, let me know if there's anything that needs adjustment!)

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -40,13 +40,13 @@ import sys
 import time
 import traceback
 import typing as t
-from datetime import datetime
 from functools import cached_property
 from io import StringIO
 from itertools import chain
 from pathlib import Path
 from shutil import rmtree
 from types import MappingProxyType
+from datetime import datetime
 
 from sqlglot import Dialect, exp
 from sqlglot.helper import first
@@ -63,7 +63,6 @@ from sqlmesh.core.config import (
 )
 from sqlmesh.core.config.connection import ConnectionConfig
 from sqlmesh.core.config.loader import C
-from sqlmesh.core.config.model import ModelDefaultsConfig
 from sqlmesh.core.config.root import RegexKeyDict
 from sqlmesh.core.console import get_console
 from sqlmesh.core.context_diff import ContextDiff
@@ -77,24 +76,24 @@ from sqlmesh.core.dialect import (
 )
 from sqlmesh.core.engine_adapter import EngineAdapter
 from sqlmesh.core.environment import Environment, EnvironmentNamingInfo, EnvironmentStatements
-from sqlmesh.core.janitor import cleanup_expired_views, delete_expired_snapshots
+from sqlmesh.core.loader import Loader
 from sqlmesh.core.linter.definition import AnnotatedRuleViolation, Linter
 from sqlmesh.core.linter.rules import BUILTIN_RULES
-from sqlmesh.core.loader import Loader
 from sqlmesh.core.macros import ExecutableOrMacro, macro
 from sqlmesh.core.metric import Metric, rewrite
 from sqlmesh.core.model import Model, update_model_schemas
+from sqlmesh.core.config.model import ModelDefaultsConfig
 from sqlmesh.core.notification_target import (
     NotificationEvent,
     NotificationTarget,
     NotificationTargetManager,
 )
-from sqlmesh.core.plan import Plan, PlanBuilder, PlanExplainer, SnapshotIntervals
+from sqlmesh.core.plan import Plan, PlanBuilder, SnapshotIntervals, PlanExplainer
 from sqlmesh.core.plan.definition import UserProvidedFlags
 from sqlmesh.core.reference import ReferenceGraph
-from sqlmesh.core.scheduler import CompletionStatus, Scheduler
+from sqlmesh.core.scheduler import Scheduler, CompletionStatus
 from sqlmesh.core.schema_loader import create_external_models_file
-from sqlmesh.core.selector import NativeSelector, Selector
+from sqlmesh.core.selector import Selector, NativeSelector
 from sqlmesh.core.snapshot import (
     DeployabilityIndex,
     Snapshot,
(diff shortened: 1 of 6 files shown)


sqlmesh/core/linter/rules/builtin.py
```diff
@@ -129,6 +130,39 @@ def check_model(self, model: Model) -> t.Optional[RuleViolation]:
...
+        test_dir = self.context.path / "tests"
+        found_test = False
+
+        yaml_parser = YAML(typ="safe")
+        for test_file in test_dir.rglob("*.yaml"):
+            try:
+                test_data = yaml_parser.load(test_file) or {}
+            except Exception:
+                # Skip files with Jinja templating or other parse errors
+                continue
```
georgesittas: I'm afraid this is going to be very expensive. For every model, we're rglobing the test directory and then loading every test, just to see if the model's name appears in one of them. I don't think this overhead will be acceptable in practice.

We should be smarter about this. For example, one idea is to register the models that have tests associated with them at load time, so that this done in a single pass– the existing one done to load tests. Then, implementing a rule like this becomes a simple lookup in a set.

cmgoffena13: Hey, thanks for checking this out, I agree, my knowledge of the repo is lacking here, couple questions:
1. What do you mean by load time, what triggers that? 
2a. Is there a certain place in the codebase where storing this would be best? I can start looking around.
2b. Best approach for storing it? Is there something similar already happening or would this be an entirely new initiative

georgesittas: No problem!

1. The loading stage corresponds to the code path where SQLMesh loads a project's assets (models, tests, etc). You can find the relevant logic within `sqlmesh/core/loader.py`– check out `GenericContext.load` too.
2a. One idea is to load tests eagerly, similar to other assets, so that you know all tests that exist as early as possible and can extract the information of what models have tests at linting time.
2b. We'd want fast lookup, so either dict or set

@VaggelisD why don't we eagerly load all tests at load time? It appears that we only do it in `GenericContext.test` today, lazily.

cmgoffena13: After an initial scan here's my code plan:
1. sqlmesh.core.test.discovery.py has ModelTestMetadata - implement a method to pull the model name out of the test body
2. sqlmesh.core.loader.py has Loader.load_model_tests - doesn't look like its implemented yet, so flesh that out and then pass the ModelTestMetadata classes to the project. I can then implement another method that creates a Set of the model names by looping through once.
3. Make sure sqlmesh.core.context.py GenericContext.load has the Set implementation as well since that's new
4. From there I can access the create Set method  from context in the linter rule to easily compare model names against.

Idea is to load the metadata classes (not sure about ModelTest, I assume ModelTestMetadata is the way to go here) into the project and then I can create a method that returns a set of the model names from the Metadata classes that I can use to compare in the linter from the context

Thoughts?

VaggelisD: > why don't we eagerly load all tests at load time? It appears that we only do it in GenericContext.test today, lazily.

@georgesittas Not really sure, I think that always the case i.e even before my refactor. I don't think there's anything blocking us from moving `load_model_tests` to load time.

@cmgoffena13 Regarding your plan:

- For bullet point (2), that is the base class implementation that's left empty on purpose. The actual function is `SqlMeshLoader::load_model_tests()`
- The `ModelTestMetadata` is indeed the class of interest, and more specifically, its `body` dict which contains the `model` attr. `ModelTest` is the class that implements the runnable test i.e it combines the metadata class with the engine adapter etc
- Bear in mind, each model is linted with a new instance of `Rule`, so we'd want to avoid repeated computations in `check_model`

In my opinion, we'd ideally:
- Move `load_model_tests()` at load time i.e `Context::load()`, as is the case for all other project related information. That will then be accessed/reused in both `Context::test()` and in the linter
- Do a single pass over the pre-loaded `ModelTestMetadata` and add `metadata.body.model` in a dict/set. That would then be reused in your `check_model` for a quick check, thus avoiding the need to reparse the YAML files or rebuild the model set from scratch.

cmgoffena13: Hey @georgesittas @VaggelisD -- let me know if the above implementation works and I can look into adjusting the tests. (Side Note: not sure what's going on with `tests/integrations/github/cicd/test_integration.py::test_pr_add_model`)

sqlmesh/core/test/discovery.py
```diff
@@ -20,6 +20,10 @@ class ModelTestMetadata(PydanticModel):
     def fully_qualified_test_name(self) -> str:
         return f"{self.path}::{self.test_name}"
 
+    @property
+    def model_name(self) -> str:
+        return self.body["model"]
```
georgesittas: We should be a bit more conservative here with the lookup. IIRC the body is validated later at runtime, when the tests are actually ran, by which time it'll be too late if the test is missing its name. We don't wanna throw a `KeyError` in that case.

```suggestion
    @property
    def model_name(self) -> str:
        return self.body.get("model", "")
```

sqlmesh/core/loader.py
```diff
@@ -64,6 +64,8 @@ class LoadedProject:
     excluded_requirements: t.Set[str]
     environment_statements: t.List[EnvironmentStatements]
     user_rules: RuleSet
+    model_test_metadata: t.List[ModelTestMetadata]
+    models_with_tests: t.Set[str]
```
georgesittas: Why is this an attribute of the `LoadedProject`? We just need the metadata right? Then you other consumers (e.g., the context, or linter) can extract this info and use it as needed.

sqlmesh/core/context.py
```diff
@@ -2782,6 +2796,33 @@ def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
             raise SQLMeshError(f"Gateway '{gateway}' not found in the available engine adapters.")
         return self.engine_adapter
 
+    def _filter_preloaded_tests(
```
georgesittas: What motivated this addition? Seems like it's something that doesn't need to be included in the scope of this PR.

cmgoffena13: If I remember correctly, the test() method was setup to still lazily load the tests, so I had to implement this to ensure it maintained filtering functionality and utilized the eagerly loaded tests instead.

sqlmesh/core/loader.py
cmgoffena13: Yeah, good catch, I can move this to the linter and keep the core code clean. Or actually, the context might be a little better, but I'm open to both

sqlmesh/core/test/discovery.py
cmgoffena13: Gotcha, quick fix, I'll get it done

sqlmesh/core/loader.py
georgesittas: Make sure this is computed once and is kept in the context so that we don't recompute it for every model. It is the same value for all models, anyway.

georgesittas (commented):
Btw, let's make sure that we track calls to `load_model_tests` and refactor as needed. For example, in `magics.py`, I think we should be able to skip that call since we already have the loaded context which contains all tests. Another example is in `lsp/context.py`.

sqlmesh/core/context.py
georgesittas: Ah, makes sense. I suggest renaming this to `_select_tests` instead, then.

sqlmesh/core/context.py
```diff
@@ -2782,6 +2796,33 @@ def _get_engine_adapter(self, gateway: t.Optional[str] = None) -> EngineAdapter:
...
+            for test in tests:
+                if "::" in test:
+                    filename, test_name = test.split("::", maxsplit=1)
+                    test_path = Path(filename)
+                    filtered_tests.extend(
+                        [t for t in test_meta if t.path == test_path and t.test_name == test_name]
+                    )
+                else:
+                    test_path = Path(test)
+                    filtered_tests.extend([t for t in test_meta if t.path == test_path])
```
georgesittas: These iterations over `test_meta` look costly. Can we store a mapping `test_name -> test_meta` instead of keeping everything in a list and having to walk it every time?

cmgoffena13: Just pushed, I added in two dictionaries, one for the fully qualified names and another for the Paths so they're just lookups. Calculated once in the context.

georgesittas:
Thanks for addressing the comments @cmgoffena13. I'll take another look soon.

cmgoffena13:
@georgesittas  - bumping this now that we're past the first round of holidays

georgesittas (approved):
Thanks @cmgoffena13, I'll get this in and see if I can slightly refactor by removing / compacting some of the new context attributes.

PR #5299, opened 2025-09-04
Fix: Add support for args syntax in dbt config

themisvaltinos:
SQLMesh's dbt integration only supported keyword argument syntax for the `config()` function:
  ```sql
  {{ config(materialized='table', alias='my_table') }}
```
  However dbt also supports dictionary syntax:
  ```
{{ config({"materialized": "table", "alias": "my_table"}) }}
```
and since this: https://github.com/TobikoData/sqlmesh/pull/5269 would make us actually render this jinja config block it would fail without supporting the args


georgesittas (approved):
Nice!

PR #5305, opened 2025-09-05
Fix: disable query validations for dbt models

georgesittas:
This PR makes the SQL model definition validation opt-in, via a linter rule. It is enabled by default, but only for native SQLMesh projects. This is done to be more lenient when loading dbt projects, since we shouldn't reject models that have, e.g., duplicate projections– they run fine in dbt.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -81,7 +81,7 @@ from sqlmesh.core.linter.definition import AnnotatedRuleViolation, Linter
 from sqlmesh.core.linter.rules import BUILTIN_RULES
 from sqlmesh.core.macros import ExecutableOrMacro, macro
 from sqlmesh.core.metric import Metric, rewrite
-from sqlmesh.core.model import Model, update_model_schemas
+from sqlmesh.core.model import Model, SqlModel, update_model_schemas
 from sqlmesh.core.config.model import ModelDefaultsConfig
 from sqlmesh.core.notification_target import (
     NotificationEvent,
@@ -680,10 +680,13 @@ class GenericContext(BaseContext, t.Generic[C]):
                 cache_dir=self.cache_dir,
             )
 
-            models = self.models.values()
-            for model in models:
-                # The model definition can be validated correctly only after the schema is set.
-                model.validate_definition()
+            # The model definition can be validated correctly only after the schema is set.
+            for model in self.models.values():
+                if isinstance(model, SqlModel):
+                    # Validates only the model metadata; SQL validations are handled by the linter
+                    super(SqlModel, model).validate_definition()  # type: ignore
+                else:
+                    model.validate_definition()
 
         duplicates = set(self._models) & set(self._standalone_audits)
         if duplicates:
--- a/sqlmesh/core/linter/rules/builtin.py
+++ b/sqlmesh/core/linter/rules/builtin.py
@@ -25,6 +25,7 @@ from sqlmesh.core.linter.rule import (
 )
 from sqlmesh.core.linter.definition import RuleSet
 from sqlmesh.core.model import Model, SqlModel, ExternalModel
+from sqlmesh.utils.errors import ConfigError
 from sqlmesh.utils.lineage import extract_references_from_query, ExternalModelReference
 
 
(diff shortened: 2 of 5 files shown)


sqlmesh/core/context.py
```diff
@@ -680,10 +680,13 @@ def load(self, update_schemas: bool = True) -> GenericContext[C]:
...
 
-            models = self.models.values()
-            for model in models:
-                # The model definition can be validated correctly only after the schema is set.
-                model.validate_definition()
+            # The model definition can be validated correctly only after the schema is set.
+            for model in self.models.values():
+                if isinstance(model, SqlModel):
+                    # Validates only the model metadata; SQL validations are handled by the linter
+                    super(SqlModel, model).validate_definition()  # type: ignore
```
georgesittas: For whatever reason, mypy wasn't happy without this `# type: ignore`:

```
sqlmesh/core/context.py:687: error: "validate_definition" undefined in superclass  [misc]
Found 1 error in 1 file (checked 515 source files)
```

Seems like its own limitation though, since `_Model` clearly defines `validate_definition`.

tobymao: i think you should keep the old version, we still call validate_definition() for somethings, but just move some of the logic out

tobymao: validation_definition should be defined, some change of your must have messed with the types?

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -274,4 +275,22 @@ def create_fix(self, model_name: str) -> t.Optional[Fix]:
...
+    Checks whether a model satisfies certain properties, such as (but not limited to):
+
+    - If SQL-based, it contains at least one projection & projection names are unique
+    - Its kind is configured correctly (e.g., the VIEW kind is not supported for Python models)
+    - Other metadata properties are well-formed (e.g., incremental-by-time models require a time column)
+    """
+
+    def check_model(self, model: Model) -> t.Optional[RuleViolation]:
+        try:
+            model.validate_definition()
```
tobymao: let's extract the logic out here, and not have it in validate_definition,

let's change validate definition lint into a different name, like duplicate names and inferrable projections etc

sqlmesh/core/linter/rules/builtin.py
```diff
@@ -274,4 +275,22 @@ def create_fix(self, model_name: str) -> t.Optional[Fix]:
         )
 
 
+class ValidateModelDefinition(Rule):
```
tobymao: this linter needs to be enabled by default for sqlmesh, disabled by default in dbt, but then be opt in or opt out for both

sqlmesh/core/context.py
georgesittas: > validation_definition should be defined, some change of your must have messed with the types?

This code path is definitely hit at runtime, so this should work. Don't see how my changes could affect types.

> I think you should keep the old version, we still call validate_definition() for somethings, but just move some of the logic out

Ah, so you prefer just picking out the duplicate projection / aliasing validations and moving them here?

sqlmesh/core/linter/rules/builtin.py
georgesittas: Sounds good

sqlmesh/core/context.py
georgesittas: Ok based on https://github.com/TobikoData/sqlmesh/pull/5305#discussion_r2326063423 it seems like that's it. I'll give it a whirl.

sqlmesh/core/linter/rules/builtin.py
georgesittas: I thought this was already the case, but I'll double-check.

sqlmesh/core/linter/rules/builtin.py
izeigerman: +1 to this. Ideally we want some granular rules here. One per validation

georgesittas: Took a stab at this here: https://github.com/TobikoData/sqlmesh/pull/5305/commits/82982228ebad683aedcc876bceeee189cd9f2047.

> Ideally we want some granular rules here. One per validation

I merged the duplicate projection & ambiguous name rules into one rule because both are essentially checking for ambiguity and computing them at the same time felt natural.

sqlmesh/core/context.py
```diff
@@ -3131,7 +3131,9 @@ def lint_models(
         found_error = False
 
         model_list = (
-            list(self.get_model(model) for model in models) if models else self.models.values()
+            list(self.get_model(model, raise_if_missing=True) for model in models)
+            if models
+            else self.models.values()
```
georgesittas, line 3135: @VaggelisD do you remember if the "lax" `get_model` call was intentional here?

PR #5306, opened 2025-09-05
feat: add dedicated `incremental_by_time_range` strategy

eakmanrq:
Prior to this PR, users could opt-in to using the incremental by time range SQLMesh behavior by adding a `time_column` property to their incremental config and then setting `{% if sqlmesh_incremental is defined %}` jinja block. 

This PR deprecates this behavior (but still supports it for now) and add a new incremental strategy called `incremental_by_time_range`. This gives SQLMesh it's own API for defining it's specific fields and removes the need for defining a jinja code block to activate it.

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -31,6 +31,7 @@ from sqlmesh.core.model.kind import (
     OnAdditiveChange,
     on_destructive_change_validator,
     on_additive_change_validator,
+    TimeColumn,
 )
 from sqlmesh.dbt.basemodel import BaseModelConfig, Materialization, SnapshotStrategy
 from sqlmesh.dbt.common import SqlStr, sql_str_validator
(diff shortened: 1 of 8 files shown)


sqlmesh/dbt/model.py
```diff
@@ -313,16 +331,21 @@ def model_kind(self, context: DbtContext) -> ModelKind:
...
 
-                if strategy not in INCREMENTAL_BY_TIME_STRATEGIES:
+                if strategy not in INCREMENTAL_BY_TIME_RANGE_STRATEGIES:
                     get_console().log_warning(
                         f"SQLMesh incremental by time strategy is not compatible with '{strategy}' incremental strategy in model '{self.canonical_name(context)}'. "
-                        f"Supported strategies include {collection_to_str(INCREMENTAL_BY_TIME_STRATEGIES)}."
+                        f"Supported strategies include {collection_to_str(INCREMENTAL_BY_TIME_RANGE_STRATEGIES)}."
+                    )
+
+                if self.time_column and strategy not in {"incremental_by_time_range", "microbatch"}:
```
izeigerman: Why do we include `microbatch` here?

PR #5307, opened 2025-09-05
Chore: Split migration script implementations into DDL and DML

izeigerman:
The split allows us to skip unnecessary DML operations during the initial state setup, which significantly speeds up initialization when using engines like BigQuery.

In my test the init time went from ~1m30s down to ~25s

### DIFF
--- a/sqlmesh/core/state_sync/db/environment.py
+++ b/sqlmesh/core/state_sync/db/environment.py
@@ -285,11 +285,13 @@ class EnvironmentState:
         return []
 
     def _environment_from_row(self, row: t.Tuple[str, ...]) -> Environment:
-        return Environment(**{field: row[i] for i, field in enumerate(Environment.all_fields())})
+        return Environment(
+            **{field: row[i] for i, field in enumerate(sorted(Environment.all_fields()))}
+        )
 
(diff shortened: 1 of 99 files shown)


sqlmesh/core/state_sync/db/environment.py
```diff
@@ -285,11 +285,13 @@ def get_environment_statements(self, environment: str) -> t.List[EnvironmentStat
...
     def _environment_from_row(self, row: t.Tuple[str, ...]) -> Environment:
-        return Environment(**{field: row[i] for i, field in enumerate(Environment.all_fields())})
+        return Environment(
+            **{field: row[i] for i, field in enumerate(sorted(Environment.all_fields()))}
+        )
 
     def _environment_summmary_from_row(self, row: t.Tuple[str, ...]) -> EnvironmentSummary:
         return EnvironmentSummary(
-            **{field: row[i] for i, field in enumerate(EnvironmentSummary.all_fields())}
+            **{field: row[i] for i, field in enumerate(sorted(EnvironmentSummary.all_fields()))}
```
eakmanrq, line 294: Why is sorting needed?

izeigerman: Because `all_fields()` returns a set

eakmanrq (approved):

PR #5308, opened 2025-09-06
Fix: make sure builtin dbt macros can be referenced with the 'context.' prefix

treysp:
No explicit documentation of this usage, but seen in the wild

Ref: https://github.com/dbt-labs/dbt-core/blob/a9dae5cac16607558d318dd20e0aab383a58d1e2/core/dbt/context/base.py#L217

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -545,7 +545,7 @@ def create_builtin_globals(
 
     builtin_globals["run_started_at"] = jinja_globals.get("execution_dt") or now()
     builtin_globals["dbt"] = AttributeDict(builtin_globals)
-    builtin_globals["context"] = AttributeDict(builtin_globals)
+    builtin_globals["context"] = builtin_globals["dbt"]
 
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -545,6 +545,7 @@ def create_builtin_globals(
 
     builtin_globals["run_started_at"] = jinja_globals.get("execution_dt") or now()
     builtin_globals["dbt"] = AttributeDict(builtin_globals)
+    builtin_globals["context"] = AttributeDict(builtin_globals)
```
izeigerman: I don't know if we need another instance of the same dict. Can we do
```
builtin_globals["context"] = builtin_globals["dbt"]
```
or define a variable and share between the 2?

izeigerman (approved):
One suggestion, LGTM otherwise.

sqlmesh/dbt/builtin.py
treysp: Ah, right - my bad

PR #5317, opened 2025-09-08
fix: scd type 2 support table properties

eakmanrq:
Slack context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1757327355130579

Prior to this PR, users could define a partition by on their SCD type 2 model and it wouldn't be implemented properly on follow-up inserts/appends that replace the table. This PR fixes that.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -2241,6 +2241,11 @@ class SCDType2Strategy(IncrementalStrategy):
                 column_descriptions=model.column_descriptions,
                 truncate=is_first_insert,
                 source_columns=source_columns,
+                storage_format=model.storage_format,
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2273,51 +2283,14 @@ def append(
...
-            model,
+        return self.insert(
             table_name,
+            query_or_df,
+            model,
+            is_first_insert=False,
             render_kwargs=render_kwargs,
-            force_get_columns_from_target=True,
+            **kwargs,
         )
```
eakmanrq, line 2293: An append in this context seems to just be an `insert` where `is_first_insert` is `False` so updated it to reflect that and remove duplicate code.

PR #5318, opened 2025-09-08
Chore: Provide additional info in jinja rendering errors

themisvaltinos:
This PR adds a best effort frame analysis to identify the specific macro invocation that triggered the jinja error. 

Currently, when the macro rendering fails, the error message only references the model containing the macro, but not which macro led to this error:
Screenshot 2025-09-09 at 12 35 55

With this change the error message is augmented with the originating macro to improve debugging:
Screenshot 2025-09-09 at 12 42 31

### DIFF
--- a/sqlmesh/core/renderer.py
+++ b/sqlmesh/core/renderer.py
@@ -242,10 +242,9 @@ class BaseExpressionRenderer:
             except ParsetimeAdapterCallError:
                 raise
             except Exception as ex:
-                error_msg = f"Could not render jinja for '{self._path}'.\n" + extract_error_details(
-                    ex
-                )
-                raise ConfigError(error_msg) from ex
+                raise ConfigError(
+                    f"Could not render jinja for '{self._path}'.\n" + extract_error_details(ex)
+                ) from ex
 
             if rendered_expression.strip():
                 try:


sqlmesh/core/renderer.py
```diff
@@ -247,9 +252,45 @@ def _resolve_table(table: str | exp.Table) -> str:
...
                 raise
             except Exception as ex:
-                raise ConfigError(
-                    f"Could not render or parse jinja at '{self._path}'.\n{ex}"
-                ) from ex
+                error_details: t.List[str] = []
+                if isinstance(ex, UndefinedError):
+                    try:
+                        _, _, exc_traceback = exc_info()
+                        for frame, _ in walk_tb(exc_traceback):
```
erindru: I guess this is necessary because Jinja gets turned into a Python script and run.

Are you able to add a test that exercises this codepath and shows the expected results?

themisvaltinos: yes added tests for native and dbt project for this

georgesittas:
Quick heads up about https://github.com/TobikoData/sqlmesh/pull/5330– it'll cause conflicts with this one.

sqlmesh/core/renderer.py
```diff
@@ -247,9 +247,11 @@ def _resolve_table(table: str | exp.Table) -> str:
             except ParsetimeAdapterCallError:
                 raise
             except Exception as ex:
-                raise ConfigError(
-                    f"Could not render or parse jinja at '{self._path}'.\n{ex}"
-                ) from ex
+                error_msg = f"Could not render or parse jinja for '{self._path}'.\n" + (
+                    extract_error_details(ex) or f"{ex}"
+                )
```
georgesittas: Nit: can we just fallback to `str(ex)` within `extract_error_details` instead of needing an `or` here?

georgesittas (approved):

sqlmesh/core/renderer.py
themisvaltinos: sure good point, added the fallback within the method

PR #5325, opened 2025-09-09
tiny comment typo builder.py

fivetran-ezralee:
Corrected typo in comment.

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -409,7 +409,7 @@ class PlanBuilder:
                 continue
 
             # Since we are traversing the graph in topological order and the largest interval range is pushed down
-            # the graph we just have to check our immediate parents in the graph and not the whole upstream graph.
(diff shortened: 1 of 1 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -409,7 +409,7 @@ def _build_restatements(
                 continue
 
             # Since we are traversing the graph in topological order and the largest interval range is pushed down
-            # the graph we just have to check our immediate parents in the graph and not the whole upstream graph.
+            # the graph we just have to check are immediate parents in the graph and not the whole upstream graph.
```
erindru, line 412: I don't think this is a typo?

"our" refers to "us" - the current node of the graph. We check **our** immediate parents and not all upstream nodes

fivetran-ezralee: ok i see, there is an implicit pause (comma) after "the graph".  Thanks

PR #5327, opened 2025-09-09
Fix: Set dbt execute variable for parse time and run time accordingly

themisvaltinos:
This update sets the dbt variable `execute` to False during parse-time and true for run time. Dbt models and macros have conditional logic which is guarded with this flag for runtime execution:

```jinja
{% if execute %}
  {% set result = run_query("SELECT COUNT(*) FROM my_table") %}
  SELECT {{ result.columns[0][0] }} as count
{% endif %}
```

These jinja expressions might depend on the data warehouse or to objects (like `graph`) not available at parse time, so the `execute` ensure these only should run at the runtime stage. 

dbt docs: https://docs.getdbt.com/reference/dbt-jinja-functions/execute

### DIFF
--- a/sqlmesh/dbt/builtin.py
+++ b/sqlmesh/dbt/builtin.py
@@ -533,7 +533,7 @@ def create_builtin_globals(
     builtin_globals.update(
         {
             "adapter": adapter,
-            "execute": True,
+            "execute": isinstance(adapter, RuntimeAdapter),
             "load_relation": adapter.load_relation,
             "store_result": sql_execution.store_result,
             "load_result": sql_execution.load_result,
--- a/tests/core/test_integration.py
+++ b/tests/core/test_integration.py
@@ -2068,6 +2068,7 @@ def test_dbt_is_incremental_table_is_missing(sushi_test_dbt_context: Context):
     assert context.engine_adapter.table_exists(snapshot.table_name())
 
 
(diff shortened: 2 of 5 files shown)


sqlmesh/dbt/builtin.py
```diff
@@ -533,7 +533,7 @@ def create_builtin_globals(
     builtin_globals.update(
         {
             "adapter": adapter,
-            "execute": True,
+            "execute": isinstance(adapter, RuntimeAdapter),
```
izeigerman, line 536: No, we should not do this. `execute` is set unconditionally to True for a reason. Since we used the rendered query for change detection and categorization we want to follow the execution path during rendering. If we happen to encounter the adapter call we fail and return None which indicates that the query cannot be rendered and categorized.

izeigerman (changes requested):
Why isn't `graph` available at load time?

sqlmesh/dbt/builtin.py
themisvaltinos: > Since we used the rendered query for change detection and categorization we want to follow the execution path during rendering. If we happen to encounter the adapter call we fail and return None which indicates that the query cannot be rendered and categorized.

I see was wondering about this, thanks that makes sense. and I guess now that we can fail and still proceed later to runtime this can't lead to any issues

PR #5330, opened 2025-09-09
Chore: Separate jinja rendering and parsing try blocks for more precise error output

crericha:


sqlmesh/core/renderer.py
```diff
@@ -239,17 +239,19 @@ def _resolve_table(table: str | exp.Table) -> str:
...
                     if not expressions:
                         raise ConfigError(f"Failed to parse an expression:\n{self._expression}")
-            except ParsetimeAdapterCallError:
-                raise
-            except Exception as ex:
-                raise ConfigError(
-                    f"Could not render or parse jinja at '{self._path}'.\n{ex}"
-                ) from ex
+                except Exception as ex:
+                    raise ConfigError(f"Could not parse jinja at '{self._path}'.\n{ex}") from ex
```
georgesittas: Since we've rendered the jinja, it's just SQL now right?
```suggestion
                    raise ConfigError(f"Could not parse the rendered jinja at '{self._path}'.\n{ex}") from ex
```

PR #5338, opened 2025-09-09
fix: dbt handle indirect cycles

eakmanrq:
Expands on this PR to handle cycles that are not created by direct dependencies: https://github.com/TobikoData/sqlmesh/pull/5253

Since the logic to detect a cycle had to be added, also updated the graph class to use this logic in order to improve the cycle error message to provide the exact cycle causing the issue instead of candidates (which in some cases was hard to determine the actual cycle).


sqlmesh/utils/dag.py
```diff
@@ -99,6 +99,56 @@ def upstream(self, node: T) -> t.Set[T]:
...
+        if not nodes_in_cycle:
+            return None
+
+        # Use DFS to find a cycle path
+        visited: t.Set[T] = set()
+        rec_stack: t.Set[T] = set()
+        path: t.List[T] = []
+
+        def dfs(node: T) -> t.Optional[t.List[T]]:
+            if node in rec_stack:
```
izeigerman: In this case I don't expect `path` to be very long which means we can likely check it directly and get rid of `rec_stack`.

sqlmesh/utils/dag.py
```diff
@@ -125,23 +175,28 @@ def sorted(self) -> t.List[T]:
...
-                    if last_processed_nodes:
-                        last_processed_msg = "\nLast nodes added to the DAG: " + ", ".join(
-                            str(node) for node in last_processed_nodes
-                        )
+                    last_processed_msg = ""
+                    if cycle_path:
+                        cycle_msg = f"\nCycle: {' -> '.join(str(node) for node in cycle_path)} -> {cycle_path[0]}"
                     else:
-                        last_processed_msg = ""
+                        # Fallback message in case a cycle can't be found
```
crericha, line 185: Are there scenarios where a cycle won't be found? I'm wondering if we can remove the else.

tests/utils/test_dag.py
```diff
@@ -57,8 +57,7 @@ def test_sorted_with_cycles():
 
     expected_error_message = (
         "Detected a cycle in the DAG. Please make sure there are no circular references between nodes.\n"
-        "Last nodes added to the DAG: c\n"
-        "Possible candidates to check for circular references: d, e"
+        "Cycle: d -> e -> d"
```
crericha: Is the output format of the error usable when real model names are in play? For example, if you have a 5 model cycle and you have the FQN for each model, is it easy to read with word wrap? Or do we need some newlines to make it more understandable?

sqlmesh/utils/dag.py
eakmanrq: No known scenarios at this time but it seems like a safe fallback to have in place for now.

tests/utils/test_dag.py
eakmanrq: Updated: https://github.com/TobikoData/sqlmesh/pull/5338/commits/b8507e345d2f4696db1c7318a75a418f433b7a64

sqlmesh/utils/dag.py
crericha: Would it be better to not have a fallback to get user feedback on what scenario hits it more quickly?

crericha (approved):

PR #5349, opened 2025-09-11
Fix: Include root package in search candidates when resolving dbt macros

vchan:
dbt package macros can be overridden in the root package. This PR includes the root package when searching  packages during macro dispatch resolution.


georgesittas (approved):
So, to try and summarize: when there's a macro namespace, we now always check the root package first for any possible overrides. And this is bound to change when we take `dispatch` configs into account.

PR #5351, opened 2025-09-11
Fix(databricks): Materialized view drop failure due to incorrect type

### LINKED ISSUE #5249: SQLMesh v0.211.2 - Databricks Materialised View cannot be dropped
I just upgraded from v0.209 to v0.211.2. After I made some changes and `sqlmesh plan dev`, the CLI says `ServerOperationError: Cannot drop a materialized view with DROP VIEW. Please use DROP MATERIALISED VIEW instead`

This never happens on v0.209.

themisvaltinos:
This fixes #5249 and correctly maps `MATERIALIZED_VIEW` to `materialized_view` instead of `view` in the case statement of `_get_data_objects`.

Previously the Databricks adapter was incorrectly identifying materialized views as regular views. This caused `DROP VIEW` to be issued instead of `DROP MATERIALIZED VIEW` resulting in the error:

`  Cannot drop a materialized view with DROP VIEW. Please use DROP MATERIALIZED VIEW instead
`

---

#### failure scenatio

So this wasn't exercised before #5087 but now since `drop_data_object_on_type_mismatch` is called like this inside `create_view`:
```
self.drop_data_object_on_type_mismatch(
      self.get_data_object(view_name),
      DataObjectType.VIEW if not materialized else DataObjectType.MATERIALIZED_VIEW,
)
```
- `get_data_object` internally would return a materialised view's type as view while the conditional would correctly map as materialised view
- this incorrect mismatch would result in the `drop_data_object` operation inside `drop_data_object_on_type_mismatch` 
- the drop would use the type returned from the query which would be `view` thus resulting in the error

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -302,18 +302,18 @@ workflows:
           matrix:
             parameters:
               engine:
-                # - snowflake
+                - snowflake
                 - databricks
-                # - redshift
-                # - bigquery
-                # - clickhouse-cloud
-                # - athena
-                # - fabric
-                # - gcp-postgres
-          # filters:
-          #   branches:
-          #     only:
-          #       - main
+                - redshift
+                - bigquery
+                - clickhouse-cloud
+                - athena
+                - fabric
+                - gcp-postgres
+          filters:
+            branches:
+              only:
+                - main
       - ui_style
       - ui_test
       - vscode_test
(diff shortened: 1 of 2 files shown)


.circleci/continue_config.yml
```diff
@@ -303,7 +303,7 @@ workflows:
             parameters:
               engine:
                 - snowflake
-                - databricks
+                # - databricks
```
themisvaltinos: todo: reinstate

PR #5352, opened 2025-09-11
Fix: treat all instances of macro variables as case-insensitive

georgesittas:
What the title says. We now treat all macro references in SQL as case-insensitive. The rationale is that we already normalize `variables` and `blueprint_variables`, so doing it for all variables feels like a natural extension of that. The motivating problem was the following:

```sql
MODEL (
    name foo.bar_@{X},
    blueprints ((X := 'bla')),
);

SELECT
  1 AS c
```

The `@{X}` variable is not rendered, because we don't do case-insensitive lookups for `@{...}`-type variables:

```python
>>> from sqlmesh import Context
>>> ctx = Context()
>>> ctx.models
mappingproxy({'"db"."foo"."bar_@{x}"': SqlModel<foo.bar_@{X}: test.sql>})
```

### DIFF
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -9377,9 +9377,9 @@ def test_model_blueprinting(tmp_path: Path) -> None:
         model_defaults=ModelDefaultsConfig(dialect="duckdb"),
     )
 
-    blueprint_sql = tmp_path / "macros" / "identity_macro.py"
-    blueprint_sql.parent.mkdir(parents=True, exist_ok=True)
-    blueprint_sql.write_text(
+    identity_macro = tmp_path / "macros" / "identity_macro.py"
+    identity_macro.parent.mkdir(parents=True, exist_ok=True)
+    identity_macro.write_text(
         """from sqlmesh import macro
 
 @macro()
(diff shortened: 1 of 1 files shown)


sqlmesh/core/macros.py
```diff
@@ -256,14 +267,18 @@ def evaluate_macros(
                 changed = True
                 variables = self.variables
 
-                if node.name not in self.locals and node.name.lower() not in variables:
+                # This makes all variables case-insensitive, e.g. @X is the same as @x. We do this
+                # for consistency, since `variables` and `blueprint_variables` are normalized.
+                var_name = node.name.lower()
+
+                if var_name not in self.locals and var_name not in variables:
```
georgesittas, line 274: This change has a subtle side-effect: today, despite being counter-intuitive, one can actually reference a variable defined in Python within a model:

```
# model.sql
MODEL (name test, kind full);
SELECT @x_plus_one() AS xp1, @X AS x

# macros.py
from sqlmesh import macro

X = 1

@macro()
def x_plus_one(evaluator):
    return X + 1
```

The above configuration is correct and the model renders into:

```sql
SELECT
  2 AS "xp1",
  1 AS "x"
```

After this PR, the reference `@X` will be invalid because we'll look up the lowercase `"x"` in `locals`, which won't match the Python uppercase key `"X"`, resulting in:

> Macro variable 'X' is undefined.

I think this is fine. I doubt anyone's relying on Python variables being exposed to SQL models today. I don't think we've even documented this anywhere, it's just a weird side-effect of populating `locals` with `python_env` items.

sqlmesh/core/macros.py
```diff
@@ -313,11 +328,11 @@ def template(self, text: t.Any, local_variables: t.Dict[str, t.Any]) -> str:
...
         # We try to convert all variables into sqlglot expressions because they're going to be converted
         # into strings; in sql we don't convert strings because that would result in adding quotes
-        mapping = {
-            k: convert_sql(v, self.dialect)
+        base_mapping = {
+            k.lower(): convert_sql(v, self.dialect)
             for k, v in chain(self.variables.items(), self.locals.items(), local_variables.items())
         }
-        return MacroStrTemplate(str(text)).safe_substitute(mapping)
+        return MacroStrTemplate(str(text)).safe_substitute(CaseInsensitiveMapping(base_mapping))
```
georgesittas, line 335: The idea here is to make variable lookups in SQL case-insensitive; the `template` method is only used for substituting variables in SQL, as far as I can tell.

sqlmesh/core/macros.py
```diff
@@ -327,7 +342,9 @@ def evaluate(self, node: MacroFunc) -> exp.Expression | t.List[exp.Expression] |
                     args[0] if len(args) == 1 else exp.Tuple(expressions=list(args))
                 )
             else:
-                self.locals[node.name] = self.transform(node.expression)
+                # Make variables defined through `@DEF` case-insensitive
+                self.locals[node.name.lower()] = self.transform(node.expression)
```
georgesittas, line 346: Again, this should just affect SQL models that have `@DEF`– now the defined variables are automatically stored as lowercase.

georgesittas:
> Do we need to update docs anywhere?

Maybe, I'll take a look and update if needed. 👍

PR #5357, opened 2025-09-11
Fix!: don't consume connected tokens when parsing quoted identifiers

georgesittas:
This PR addresses two bugs:

1. In SQLMesh's override of `_parse_id_var`, once we parse an identifier we [check](https://github.com/TobikoData/sqlmesh/pull/1835/files#diff-32e7379e6d699794516fd766eee46a0a093be9bb23c3fdb4f8ea20117c2a44c3R141-R148) whether the next token in the stream is "connected" to the identifier, i.e., there's no whitespace between the two tokens. This is done in order to support our custom macro syntax ([example](https://github.com/TobikoData/sqlmesh/pull/1835/files#diff-4e33c889891763da6cb9bc1366a3c4b6401ae9c800a7d35b00ec4865e5f03e9cR133)). Unfortunately, the condition was too lax, resulting in consuming both tokens in cases like `"foo"bar`, when we needed to consume only one.
2. There is a [subtle bug](https://github.com/TobikoData/sqlmesh/pull/5189/files#diff-9112c1d87cfc89a10e62a210d1c88414ce406ee2eb42be336acc3e84a0104a21R1070), where `_schema_tmp` can be appended to a quoted identifier like `"a"."b"."c"`, resulting in `"a"."b"."c"_schema_tmp`. Due to the parser being lax, we were somehow able to parse this.. we've succeeded by failing in this case :)

### DIFF
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1128,7 +1128,7 @@ def haversine_distance(
 def pivot(
     evaluator: MacroEvaluator,
     column: SQL,
-    values: t.List[SQL],
+    values: t.List[exp.Expression],
     alias: bool = True,
     agg: exp.Expression = exp.Literal.string("SUM"),
     cmp: exp.Expression = exp.Literal.string("="),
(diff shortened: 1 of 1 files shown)


sqlmesh/core/macros.py
```diff
@@ -1146,23 +1146,23 @@ def pivot(
         >>> from sqlmesh.core.macros import MacroEvaluator
         >>> sql = "SELECT date_day, @PIVOT(status, ['cancelled', 'completed']) FROM rides GROUP BY 1"
         >>> MacroEvaluator().transform(parse_one(sql)).sql()
-        'SELECT date_day, SUM(CASE WHEN status = \\'cancelled\\' THEN 1 ELSE 0 END) AS "\\'cancelled\\'", SUM(CASE WHEN status = \\'completed\\' THEN 1 ELSE 0 END) AS "\\'completed\\'" FROM rides GROUP BY 1'
+        'SELECT date_day, SUM(CASE WHEN status = \\'cancelled\\' THEN 1 ELSE 0 END) AS "cancelled", SUM(CASE WHEN status = \\'completed\\' THEN 1 ELSE 0 END) AS "completed" FROM rides GROUP BY 1'
```
georgesittas, line 1149: Not sure what led to this decision, i.e. including the SQL-representation of the value as-is in the alias.

tests/core/engine_adapter/test_clickhouse.py
```diff
@@ -640,7 +640,7 @@ def test_scd_type_2_by_time(
     "test_valid_from",
     "test_valid_to",
     TRUE AS "_exists"
-  FROM ""__temp_target_efgh""
+  FROM "__temp_target_efgh"
```
georgesittas, line 643: This test was incorrect, the quotes here are off.

tests/core/test_format.py
```diff
@@ -28,7 +28,7 @@ def test_format_files(tmp_path: pathlib.Path, mocker: MockerFixture):
     f3 = create_temp_file(
         tmp_path,
         pathlib.Path(audits_dir, "audit_1.sql"),
-        "AUDIT(name assert_positive_id, dialect 'duckdb'); SELECT  * FROM @this_model WHERE  \"CaseSensitive\"_item_id < 0;",
+        "AUDIT(name assert_positive_id, dialect 'duckdb'); SELECT  * FROM @this_model WHERE  \"CaseSensitive_item_id\" < 0;",
```
georgesittas, line 31: This test was also incorrect, it's like trying to do `SELECT * FROM t WHERE "foo"_bar < 0`.

PR #5368, opened 2025-09-12
Fix: load dbt model containing only a comment and non-sql jinja

treysp:


sqlmesh/core/renderer.py
```diff
@@ -249,15 +249,22 @@ def _resolve_table(table: str | exp.Table) -> str:
...
-                        raise ConfigError(f"Failed to parse an expression:\n{self._expression}")
-                except Exception as ex:
-                    raise ConfigError(
-                        f"Could not parse the rendered jinja at '{self._path}'.\n{ex}"
-                    ) from ex
+                # ensure there is actual SQL and not just comments and non-SQL jinja
+                dialect = Dialect.get_or_raise(self._dialect)
+                tokens = dialect.tokenize(rendered_expression)
+
+                if tokens:
```
izeigerman, line 256: Since we already tokenized can we avoid tokenizing again as part of `parse` and just use existing tokens?

treysp: Updated

georgesittas (approved):

Issue #5365, opened 2025-09-12
Avoid MV recreation in prod: implement instantaneous version switching


xardasos:
Consider the following DAG: A (materialized view) <- B (materialized view).

Suppose these models are deployed in `prod` and got assigned physical tables `A__1` and `B__1`, respectively. If we make a `Non-Breaking` change to `A`, a new table `A__2` will be assigned to it. Additionally, view `B` must be recreated (otherwise it would continue pointing at the old `A__1` table). This means that `B__1` gets deleted and then created again, this time pointing at `A__2`.

The problem with this solution is that the data from `B` is not available in prod during the recreation time (which can be very long in the case of bigger MVs). Moreover, rollbacks also require another change in the physical layer.

A possible improvement would be to instead create a new version of the physical table `B__2` first and then do an almost instantaneous switch in the virtual layer, avoiding the long downtime.

This solution requires changing the way snapshot versions (e.g. `__1` and `__2`) are calculated. We need to calculate the MV version based on both parent data hash and data hash, similarly to the current dev versions.

Perhaps this new way of calculating the versions of MVs could be optional/configurable?

Feel free to share your thoughts and objections - I'd love to hear your perspective.
/cc @izeigerman

PR #5374, opened 2025-09-15
Feat!: Categorize indirect MV changes as breaking for seamless version switching

xardasos:
Addresses issue: https://github.com/TobikoData/sqlmesh/issues/5365

We now categorize indirect changes to MVs as breaking to allow for instantaneous switches in a virtual layer. This addresses the issue of potentially long downtimes during MVs recreation. This improvement is especially relevant for RisingWave, where chaining MVs is a common use case, as described in https://risingwave.com/understanding-materialized-views/.


sqlmesh/core/plan/builder.py
```diff
@@ -674,6 +674,12 @@ def _categorize_snapshot(
                     if mode == AutoCategorizationMode.FULL:
                         snapshot.categorize_as(SnapshotChangeCategory.BREAKING, forward_only)
         elif self._context_diff.indirectly_modified(snapshot.name):
+            if snapshot.is_materialized_view:
+                # We categorize changes as breaking to allow for instantaneous switches in a virtual layer.
+                # Otherwise, there might be a potentially long downtime during MVs recreation.
+                snapshot.categorize_as(SnapshotChangeCategory.INDIRECT_BREAKING, forward_only)
```
izeigerman, line 688: I think this looks good, but there's still a question what should be the correct behavior when `forward_only` is True, which will lead to the version being reused despite the `INDIRECT_BREAKING` category.

The `forward_only` flag can be set due to:
1. Running `sqlmesh plan --forward-only`
2. The `virtual_environment_mode` is set to `dev_only` in the project config, indicating that no virtual layer should be used in production

In both cases, it is the user's explicit intent to continue using the same (existing) table version. How should we handle these scenarios?

xardasos: Good question. I'm not sure, but maybe we should keep the old behavior in case of forward-only changes?

xardasos: I added a check to preserve the original behavior for forward-only changes.

xardasos:
Hey @izeigerman,

please let me know if there’s anything else that should be changed or improved before moving forward.

sqlmesh/core/plan/builder.py
```diff
@@ -680,6 +680,14 @@ def _categorize_snapshot(
                     if mode == AutoCategorizationMode.FULL:
                         snapshot.categorize_as(SnapshotChangeCategory.BREAKING, forward_only)
         elif self._context_diff.indirectly_modified(snapshot.name):
+            if snapshot.is_materialized_view and not forward_only:
+                # We categorize changes as breaking to allow for instantaneous switches in a virtual layer.
+                # Otherwise, there might be a potentially long downtime during MVs recreation.
```
georgesittas, line 685: I'm curious to understand a bit more about this. What would cause this long downtime? Can you give an example?

georgesittas (approved):
This looks good to me, thanks @xardasos!

izeigerman (approved):

sqlmesh/core/plan/builder.py
xardasos: RisingWave doesn't support `create or replace` nor transactional DDL (although it can alter MV `A_1` and swap it with another existing MV `A_2`). Previously, sqlmesh would [drop and recreate](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/snapshot/evaluator.py#L2358) the MV in case of indirect changes, making the data unavailable until the MV was rebuilt (and this can take a lot of time in case of larger MVs).

Even if an engine supports `create or replace` of some sort, I think having side-by-side versions of an MV allows for instantaneous rollbacks in a virtual layer.

PR #5379, opened 2025-09-15
feat: Use merge for delta and iceberg trino catalogs

eakmanrq:
Prior to this PR, Trino engine against Iceberg and Delta Lake catalogs would use DELETE+INSERT without a transaction for insert/overwrite instead of MERGE. MERGE allows the operation to be atomic which is preferred.

Also did some cleanup by deprecating `InsertOverwriteWithMergeMixin` and instead rely solely on `INSERT_OVERWRITE_STRATEGY`. Basically before we had two ways of doing the same thing so this consolidates into a single way.

### DIFF
--- a/tests/core/engine_adapter/test_mssql.py
+++ b/tests/core/engine_adapter/test_mssql.py
@@ -16,7 +16,6 @@ from sqlmesh.core import dialect as d
 from sqlmesh.core.engine_adapter.shared import (
     DataObject,
     DataObjectType,
-    InsertOverwriteStrategy,
 )
 from sqlmesh.utils.date import to_ds
 from tests.core.engine_adapter import to_sql_calls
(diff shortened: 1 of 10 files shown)


tests/core/engine_adapter/test_mssql.py
```diff
@@ -342,46 +341,6 @@ def test_insert_overwrite_by_time_partition_supports_insert_overwrite_pandas_exi
...
-def test_insert_overwrite_by_time_partition_replace_where_pandas(
-    make_mocked_engine_adapter: t.Callable, mocker: MockerFixture, make_temp_table_name: t.Callable
-):
-    mocker.patch(
-        "sqlmesh.core.engine_adapter.mssql.MSSQLEngineAdapter.table_exists",
-        return_value=False,
-    )
-
-    adapter = make_mocked_engine_adapter(MSSQLEngineAdapter)
-    adapter.INSERT_OVERWRITE_STRATEGY = InsertOverwriteStrategy.REPLACE_WHERE
```
eakmanrq, line 354: This test didn't make sense. It was changing the MSSQL engine adapter to use `REPLACE_WHERE` and then showing that it did nothing. It would now do something but that is expected. I think it will likely a copy/paste mistake from other tests without understanding the intent.

PR #5380, opened 2025-09-15
chore: refactor to remove merge mixin

eakmanrq:
Deprecating `InsertOverwriteWithMergeMixin` and instead rely solely on `INSERT_OVERWRITE_STRATEGY`. Basically before we had two ways of doing the same thing so this consolidates into a single way.

### DIFF
--- a/tests/core/engine_adapter/test_mssql.py
+++ b/tests/core/engine_adapter/test_mssql.py
@@ -16,7 +16,6 @@ from sqlmesh.core import dialect as d
 from sqlmesh.core.engine_adapter.shared import (
     DataObject,
     DataObjectType,
-    InsertOverwriteStrategy,
 )
 from sqlmesh.utils.date import to_ds
 from tests.core.engine_adapter import to_sql_calls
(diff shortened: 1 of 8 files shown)


tests/core/engine_adapter/test_mssql.py
```diff
@@ -342,46 +341,6 @@ def test_insert_overwrite_by_time_partition_supports_insert_overwrite_pandas_exi
...
-    )
-    adapter._connection_pool.get().bulk_copy.assert_called_with(
-        f"__temp_test_table_{temp_table_id}", [(1, "2022-01-01"), (2, "2022-01-02")]
-    )
-
-    assert to_sql_calls(adapter) == [
-        f"""IF NOT EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = '__temp_test_table_{temp_table_id}') EXEC('CREATE TABLE [__temp_test_table_{temp_table_id}] ([a] INTEGER, [ds] VARCHAR(MAX))');""",
-        f"""MERGE INTO [test_table] AS [__MERGE_TARGET__] USING (SELECT [a] AS [a], [ds] AS [ds] FROM (SELECT CAST([a] AS INTEGER) AS [a], CAST([ds] AS VARCHAR(MAX)) AS [ds] FROM [__temp_test_table_{temp_table_id}]) AS [_subquery] WHERE [ds] BETWEEN '2022-01-01' AND '2022-01-02') AS [__MERGE_SOURCE__] ON (1 = 0) WHEN NOT MATCHED BY SOURCE AND [ds] BETWEEN '2022-01-01' AND '2022-01-02' THEN DELETE WHEN NOT MATCHED THEN INSERT ([a], [ds]) VALUES ([a], [ds]);""",
-        f"DROP TABLE IF EXISTS [__temp_test_table_{temp_table_id}];",
-    ]
```
eakmanrq, line 382: This test didn't make sense. It was changing the MSSQL engine adapter to use REPLACE_WHERE and then showing that it did nothing. It would now do something but that is expected. I think it will likely a copy/paste mistake from other tests without understanding the intent.

PR #5384, opened 2025-09-15
fix: VIEW model with metadata-only changes in VDE=dev_only should always migrate

newtonapple:


sqlmesh/core/plan/stages.py
```diff
@@ -364,6 +363,19 @@ def build(self, plan: EvaluatablePlan) -> t.List[PlanStage]:
             )
         )
 
+        if snapshots_with_schema_migration:
+            missing_ids_before = {s.snapshot_id for s in missing_intervals_before_promote}
+            missing_ids_after = {s.snapshot_id for s in missing_intervals_after_promote}
+            # Only skip migrate for VIEWs that are already scheduled for recreation in Backfill.
```
izeigerman: No, let's not do this. The reason is that some snapshots may rely on view schema for their own migrations. So we absolutely must migrate those views. Additionally, there's already logic in place to skip view recreation during evaluation when appropriate.

tests/core/test_plan_stages.py
```diff
@@ -1936,6 +1936,76 @@ def test_build_plan_stages_virtual_environment_mode_no_updates(
     assert len(virtual_stages) == 0
 
 
+def test_build_plan_stages_virtual_environment_mode_metadata_only_changes(
```
izeigerman, line 1940: Let's also add an integration test which involves a forward-only change to a view and check that the schema changes get reflected even though there are no evaluations.

PR #5385, opened 2025-09-16
feat(web_common): add components for lineage

mykalmax:


web/common/src/components/Lineage/layout/dagreLayout.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    function handler(
+      event: MessageEvent<LayoutedGraph<TNodeData> & { error: ErrorEvent }>,
+    ) {
+      cleanup()
+
+      if (event.data.error) return errorHandler(event.data.error)
+
+      resolve(event.data as LayoutedGraph<TNodeData, TEdgeData>)
```
benfdking: Why do we need this as here, if you have typed the input?

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
```diff
@@ -0,0 +1,82 @@
...
+) => {
+  try {
+    const { edges, nodesMap } = event.data
+    const nodes = Object.values(nodesMap)
+    const nodeCount = nodes.length
+    const edgeCount = edges.length
+    const maxCount = Math.max(nodeCount, edgeCount)
+
+    if (nodeCount === 0)
+      return self.postMessage({
```
benfdking: Can we type this post message?

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
```diff
@@ -0,0 +1,82 @@
...
+      const halfHeight = height >> 1
+
+      nodesMap[nodeId] = {
+        ...node,
+        position: {
+          x: nodeWithPosition.x - halfWidth,
+          y: nodeWithPosition.y - halfHeight,
+        },
+      }
+    }
```
benfdking: Can we make this slightly more functional looking, or is it, in fact, for performance make a note of it?

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
```diff
@@ -0,0 +1,82 @@
...
+        g.setEdge(edges[i].source, edges[i].target)
+      }
+      if (i < nodeCount) {
+        const node = nodes[i]
+        g.setNode(node.id, {
+          width: node.width || DEFAULT_NODE_WIDTH,
+          height: node.height || 0,
+        })
+      }
+    }
```
benfdking: Can you make this into two separate loops and eliminate the concept of maxCount?

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    dagre.layout(g)
+
+    for (let i = 0; i < nodeCount; i++) {
+      const node = nodes[i]
+      const width = node.width || DEFAULT_NODE_WIDTH
+      const height = node.height || 0
+      const nodeId = node.id as NodeId
+      const nodeWithPosition = g.node(nodeId)
+      const halfWidth = width >> 1
```
benfdking: I appreciate the bitwise operator for performance, but can we either note that we use this code for performance, or make a note of it

web/common/src/components/Lineage/LineageColumnLevel/ColumnLevelLineageContext.ts
```diff
@@ -0,0 +1,55 @@
...
+  adjacencyListColumnLevel: ColumnLevelLineageAdjacencyList
+  selectedColumns: Set<PortId>
+  columnLevelLineage: Map<PortId, ColumnLevelLineageAdjacencyList>
+  setColumnLevelLineage: React.Dispatch<
+    React.SetStateAction<Map<PortId, ColumnLevelLineageAdjacencyList>>
+  >
+  showColumns: boolean
+  setShowColumns: React.Dispatch<React.SetStateAction<boolean>>
+  fetchingColumns: Set<PortId>
+  setFetchingColumns: React.Dispatch<React.SetStateAction<Set<PortId>>>
```
benfdking: I have always wondered about this, wouldn't the interface be much better if you made it the much more generic feeling 

```
  setFetchingColumns: (portId: PortId) => void
```

That way you can put in what ever you want?

web/common/src/components/Lineage/edge/FactoryEdgeWithGradient.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const gradientId = useId()
+    const startColor = data?.startColor || 'var(--color-lineage-edge)'
+    const endColor = data?.endColor || 'var(--color-lineage-edge)'
+    const pathType = data?.pathType || 'bezier'
+    const strokeWidth = data?.strokeWidth || 4
+    const edgePath = getEdgePath(pathType)
+    const isActive = selectedEdges.has(edgeId)
+
+    function getEdgePath(pathType: PathType) {
```
benfdking: nit: this function just structurally reads a little weird because you call it before it's defined, and you have inputs to it from above the function call.

web/common/src/components/Lineage/LineageColumnLevel/help.ts
```diff
@@ -0,0 +1,205 @@
...
+        sourceColumns.forEach(sourceColumnName => {
+          connectedColumns.add(toPortID(targetModelName, sourceColumnName))
+        })
+      }
+    }
+  }
+  return connectedColumns
+}
+
+export function calculateNodeColumnsCount(columnsCount: number = 0) {
```
benfdking, line 187: nit: might be worth specifying some of the outputs of these functiosn to make them easier to get/refactor, etc.

web/common/src/components/Lineage/LineageColumnLevel/useColumns.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface Column {
+  data_type: string
+  description?: string | null
+}
+
+export function useColumns(
+  selectedPorts: Set<PortId>,
+  adjacencyListKey: AdjacencyListKey,
+  rawColumns: Record<AdjacencyListColumnKey, Column> = {},
+) {
```
benfdking: It might be nice to have the output as well so you don't have to read the implementation to understand what it returns.

web/common/src/components/Lineage/node/base-node.tsx
```diff
@@ -0,0 +1,17 @@
...
+>(({ className, ...props }, ref) => (
+  <div
+    ref={ref}
+    className={cn('relative rounded-md border-2 p-0', className)}
+    tabIndex={0}
+    {...props}
+  />
+))
+
+BaseNode.displayName = 'BaseNode'
```
benfdking, line 17: nit: I wonder if using an inline helper might make this prettier and avoid forgetting the display name 

```
function namedForwardRef<T, P>(name: string, render: React.ForwardRefRenderFunction<T, P>) {
  const component = forwardRef(render)
  component.displayName = name
  return component
}

export const BaseNode = namedForwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement> & { selected?: boolean }>(
  'BaseNode',
  ({ className, ...props }, ref) => (
    <div
      ref={ref}
      className={cn('relative rounded-md border-2 p-0', className)}
      tabIndex={0}
      {...props}
    />
  )
)
```

web/common/src/components/Lineage/node/NodeBase.tsx
```diff
@@ -0,0 +1,32 @@
+import { type NodeProps } from '@xyflow/react'
+import React from 'react'
+
+import { BaseNode } from '@/components/Lineage/node/base-node'
+import { cn } from '@/utils'
+
+export interface NodeBaseProps extends NodeProps {
+  className?: string
+  children?: React.ReactNode
+  style?: React.CSSProperties
```
benfdking: I don't think you do use `style`?

web/common/src/components/Lineage/node/useNodeMetadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { type LineageNode, type LineageNodeData, type NodeId } from '../utils'
+
+export function useNodeMetadata<
+  TNodeData extends LineageNodeData = LineageNodeData,
+>(
+  nodeId: NodeId,
+  currentNode: LineageNode<TNodeData> | null,
+  selectedNodeId: NodeId | null,
+  selectedNodes: Set<NodeId>,
+) {
```
benfdking, line 20: nit: again here I would add the output so you know what it returns.

web/common/src/components/Lineage/help.ts
```diff
@@ -0,0 +1,185 @@
...
+  for (let i = 0; i < nodesCount; i++) {
+    const nodeId = adjacencyListKeys[i]
+    const encodedNodeId = toNodeID(nodeId)
+    nodesMap[encodedNodeId] = transformNode(
+      encodedNodeId,
+      lineageDetails[nodeId],
+    )
+  }
+
+  return nodesMap
```
benfdking, line 55: nit: same here but make more functional to be able to or state that it is for performance

web/common/src/components/Lineage/help.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  return edges
+}
+
+export function createNode<TNodeData extends LineageNodeData = LineageNodeData>(
+  nodeId: NodeId,
+  type: string = 'node',
+  data: TNodeData,
+) {
```
benfdking, line 109: I really think we should be defining output types.

web/common/src/components/Lineage/help.ts
```diff
@@ -0,0 +1,185 @@
...
+  return encodeURI(toID(...args)) as NodeId
+}
+
+export function toEdgeID(...args: string[]) {
+  return encodeURI(toID(...args)) as EdgeId
+}
+
+export function toPortID(...args: string[]) {
+  return encodeURI(toID(...args)) as PortId
+}
```
benfdking: I think this isn't easy to untangle. There really shouldn't be the idea of encoded things in the lineage? Encoding is a concept we've jused to go over the network. 

Can you have a look at what I have done in the vscode lineage to type these things more strictly? We should be differentiating between the different ID types and typing which ids can be mapped to what. 

We should also be adding return types to functions. I wonder if we should make that a lint rule: https://typescript-eslint.io/rules/explicit-function-return-type/

web/common/src/components/Lineage/edge/FactoryEdgeWithGradient.tsx
mykalmax: i tend to move all implementations out of the main flow. in my mind it does not matter where `getEdgePath` defined, if i need to see details of `getEdgePath` i can just `click and go` to implementation

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
mykalmax: will make a note

web/common/src/components/Lineage/layout/dagreLayout.worker.ts
mykalmax: should get rid of this

benfdking (approved):

PR #5389, opened 2025-09-16
Fix: Regression that caused view snapshos not to be migrated

izeigerman:
Previously, we were incorrectly skipping the migration stage for view snapshots.

### DIFF
--- a/sqlmesh/core/plan/stages.py
+++ b/sqlmesh/core/plan/stages.py
@@ -281,9 +281,7 @@ class PlanStagesBuilder:
             snapshots_with_schema_migration = [
                 snapshots[s_id]
                 for s_id in dag.subdag(*snapshot_ids_with_schema_migration)
-                if snapshots[s_id].is_paused
-                and snapshots[s_id].is_model
-                and not snapshots[s_id].is_symbolic
+                if snapshots[s_id].supports_schema_migration_in_prod
             ]
 
         snapshots_to_intervals = self._missing_intervals(
--- a/sqlmesh/core/plan/stages.py
+++ b/sqlmesh/core/plan/stages.py
@@ -12,6 +12,7 @@ from sqlmesh.core.snapshot.definition import (
     Snapshot,
     SnapshotTableInfo,
     SnapshotId,
+    snapshots_to_dag,
 )
 
 
@@ -248,6 +249,7 @@ class PlanStagesBuilder:
         stored_snapshots = self.state_reader.get_snapshots(plan.environment.snapshots)
         snapshots = {**new_snapshots, **stored_snapshots}
         snapshots_by_name = {s.name: s for s in snapshots.values()}
+        dag = snapshots_to_dag(snapshots.values())
 
         all_selected_for_backfill_snapshots = {
             s.snapshot_id for s in snapshots.values() if plan.is_selected_for_backfill(s.name)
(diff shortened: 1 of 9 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1081,20 +1080,32 @@ def _migrate_snapshot(
...
                 )
+            else:
+                self._execute_create(
+                    snapshot=snapshot,
+                    table_name=snapshot.table_name(is_deployable=True),
+                    is_table_deployable=True,
+                    deployability_index=deployability_index,
+                    create_render_kwargs=render_kwargs,
+                    rendered_physical_properties=rendered_physical_properties,
+                    dry_run=True,
```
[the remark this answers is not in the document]
izeigerman: this is just a bad comment all around

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -489,15 +489,14 @@ def migrate(
...
         with self.concurrent_context():
             # Only migrate snapshots for which there's an existing data object
             concurrent_apply_to_snapshots(
-                snapshots_with_data_objects,
+                snapshots_by_name.values(),
                 lambda s: self._migrate_snapshot(
                     s,
                     snapshots_by_name,
-                    target_data_objects[s.snapshot_id],
+                    target_data_objects.get(s.snapshot_id),
```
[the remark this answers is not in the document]
izeigerman: the downstream method indicates that it supports an optional `target_data_object` in its signature

sqlmesh/core/plan/stages.py
```diff
@@ -248,6 +249,7 @@ def build(self, plan: EvaluatablePlan) -> t.List[PlanStage]:
         stored_snapshots = self.state_reader.get_snapshots(plan.environment.snapshots)
         snapshots = {**new_snapshots, **stored_snapshots}
         snapshots_by_name = {s.name: s for s in snapshots.values()}
+        dag = snapshots_to_dag(snapshots.values())
```
eakmanrq, line 252: What is the reasoning for having this logic outside of EvaluatablePlan? For example should it have properties `evaluatable_snapshots` and `dag` which would return this info? Reading this over and I was surprised I couldn't consult the plan itself to get this information.

izeigerman: Because `EvaluatablePlan` instance only contains data which can be (and will be) serialized. We could consider making it a property method of the `EvaluatblePlan` but currently there isn't really any other consumer for this. Plus we need access to the StateReader to fetch all snapshots which `EvalutablePlan` doesn't (and shouldn't) posses.

sqlmesh/core/plan/stages.py
```diff
@@ -271,8 +273,17 @@ def build(self, plan: EvaluatablePlan) -> t.List[PlanStage]:
...
+            ]
+            # Include all upstream dependencies of snapshots that require schema migration to make sure
+            # the upstream tables are created before the schema updates are applied
             snapshots_with_schema_migration = [
-                s for s in snapshots.values() if s.requires_schema_migration_in_prod
+                snapshots[s_id]
+                for s_id in dag.subdag(*snapshot_ids_with_schema_migration)
+                if snapshots[s_id].is_paused
+                and snapshots[s_id].is_model
+                and not snapshots[s_id].is_symbolic
```
eakmanrq: Just to help me learn: Why couldn't these if conditions be applied when creating `snapshot_ids_with_schema_migration`?

eakmanrq: Oh I see it is part of `requires_schema_migration_in_prod`. This is making sure the returned snapshots in subdag also have these properties.

sqlmesh/core/plan/stages.py
eakmanrq: Ah yes the state reader. Thanks.

sqlmesh/core/plan/stages.py
izeigerman: correct. So the snapshots in the subdag may not return True for `requires_schema_migration_in_prod` and it's fine, but we still need to perform basic checks.

eakmanrq (approved):

PR #5394, opened 2025-09-17
fix: remove need for uvicorn

benfdking:
- removes an old pin for uvicorn for the lsp
- was copied as depends on web but uvicorn is not an actual python dep


benfdking:
My reasoning is because two lines above I say we copied from web and so want to keep fact, we've copied as such.

PR #5397, opened 2025-09-17
Fix: Sort nested AttributeDict's to prevent visual diff

erindru:
Prior to this, on dbt projects, if a model had a change, often the text diff would show a diff for `refs` even though those didn't change at all. In this example, all I did was add a `stamp`, but the text diff shows a diff for `refs`:

Screenshot From 2025-09-18 09-58-54

The cause is that while the _keys_ ("refs", "sources", "vars") were sorted - the values weren't. So calling `model.render_definition()` on the Snapshot in state, calling `model.render_definition()` the local Snapshot and comparing the output could result in a text diff, even though nothing had actually changed, because the values are output in a non-deterministic order.

After this PR, we no longer see the extra "change" and the text diff focuses on what actually got changed:

Screenshot From 2025-09-18 10-03-06

### DIFF
--- a/sqlmesh/utils/jinja.py
+++ b/sqlmesh/utils/jinja.py
@@ -206,6 +206,20 @@ def extract_macro_references_and_variables(
     return macro_references, variables
 
 
+def sort_dict_recursive(
+    item: t.Dict[str, t.Any],
+) -> t.Dict[str, t.Any]:
+    sorted_dict: t.Dict[str, t.Any] = {}
+    for k, v in sorted(item.items()):
+        if isinstance(v, list):
+            sorted_dict[k] = sorted(v)
+        elif isinstance(v, dict):
+            sorted_dict[k] = sort_dict_recursive(v)
+        else:
+            sorted_dict[k] = v
+    return sorted_dict
+
+
 JinjaGlobalAttribute = t.Union[str, int, float, bool, AttributeDict]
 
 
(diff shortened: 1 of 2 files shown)


sqlmesh/utils/jinja.py
```diff
@@ -206,6 +206,20 @@ def extract_macro_references_and_variables(
     return macro_references, variables
 
 
+def sort_dict_recursive(
```
erindru, line 209: I couldnt find an existing util function that would recurse into values and sort those too, let me know if it exists and i'm just blind

georgesittas: @eakmanrq did we end up removing the utility you implemented for `python_env` sorting?

erindru:
> why is the inner dictionary order non-deterministic

We [like to use sets](https://github.com/TobikoData/sqlmesh/blob/7cfddd87584079710f80130f7070f1d417daa874/sqlmesh/dbt/manifest.py#L687). It's "deterministic" until the cache is busted and the project is re-loaded from disk.

I figured that it was easier to fix it once on output rather than trying to fix every call site now and also again in future when we inevitably make this mistake again and again

sqlmesh/utils/jinja.py
themisvaltinos: do you mean `_dict_sort` or `sort_python_env`? i'm not sure either of these would achieve the same, unless there is another one

eakmanrq: Yeah sorry I'm bad with Github notification. I originally added a recrusive dict sort but rolled that back in this PR and just sort the top level: https://github.com/TobikoData/sqlmesh/pull/4984

PR #5398, opened 2025-09-18
Fix: Dont backfill seeds if the seed data didnt actually change

erindru:
Prior to this, a simple metadata change on a seed would trigger it to be recreated / data reloaded:

Screenshot From 2025-09-18 13-23-31

This PR updates `should_force_rebuild` to return `False` on a seed snapshot if only its metadata changed. This works because snapshots with metadata changes share the physical table with the previous snapshot, meaning:
 - intervals are inherited from the previous snapshot
 - this causes `missing_intervals()` to return no missing intervals
 - causing no backfill!
 
Screenshot From 2025-09-18 14-20-58

### DIFF
--- a/sqlmesh/core/plan/builder.py
+++ b/sqlmesh/core/plan/builder.py
@@ -712,6 +712,12 @@ class PlanBuilder:
                 snapshot.categorize_as(SnapshotChangeCategory.INDIRECT_NON_BREAKING, forward_only)
         else:
             # Metadata updated.
+            if not forward_only and snapshot.is_seed:
+                # seeds with metadata updates should be categorized as forward_only to prevent unnecessary backfill
+                # backfill should only happen if the actual seed data changes, in which case it will be classed as
+                # directly modified / breaking and not forward only as this code path will not be hit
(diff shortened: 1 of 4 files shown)


sqlmesh/core/plan/builder.py
```diff
@@ -712,6 +712,12 @@ def _categorize_snapshot(
                 snapshot.categorize_as(SnapshotChangeCategory.INDIRECT_NON_BREAKING, forward_only)
         else:
             # Metadata updated.
+            if not forward_only and snapshot.is_seed:
+                # seeds with metadata updates should be categorized as forward_only to prevent unnecessary backfill
+                # backfill should only happen if the actual seed data changes, in which case it will be classed as
+                # directly modified / breaking and not forward only as this code path will not be hit
+                forward_only = True
```
izeigerman: Actually, I've realized this might not be necessary. `METADATA` category already reuses the version

erindru: Oh, you're right. I've simplified this PR

PR #5401, opened 2025-09-18
Chore: Pin cryptography instead of leaving it to the resolver

themisvaltinos:
Pin cryptography due to this change in 46.0.1: https://github.com/pyca/cryptography/commit/2726efdb6d67f1c90cf9c6062d9fe051965586f8 which leads to conflict on cffi since snowflake requires <2.0.0.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -302,18 +302,18 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
-                - databricks
-                - redshift
-                - bigquery
-                - clickhouse-cloud
-                - athena
-                - fabric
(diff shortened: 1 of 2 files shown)


.circleci/continue_config.yml
```diff
@@ -302,18 +302,18 @@ workflows:
...
-                - athena
-                - fabric
+                # - snowflake
+                # - databricks
+                # - redshift
+                # - bigquery
+                # - clickhouse-cloud
+                # - athena
+                # - fabric
                 - gcp-postgres
```
themisvaltinos: todo: reinstate prior to merging

benfdking (approved):

PR #5409, opened 2025-09-18
fix: ignore cluster_by config for embedded model kind

toriwei:
dbt ignores cluster_by for ephemeral materializations, updating to match that behavior.


sqlmesh/dbt/model.py
```diff
@@ -544,6 +544,11 @@ def to_sqlmesh(
                     "Ignoring cluster_by config for model '%s'; cluster_by is not supported for views.",
                     self.name,
                 )
+            elif isinstance(kind, EmbeddedKind):
```
themisvaltinos: nit: maybe combine with above to avoid code duplication. so something like:
```
if isinstance(kind, (ViewKind, EmbeddedKind)):
  logger.warning(
    "Ignoring cluster_by config for model '%s'; cluster_by is not supported for %s.",
     self.name,
     "views" if isinstance(kind, ViewKind) else "embedded models",
)
```

themisvaltinos (approved):
lgtm! one nit though not important. and thinking in the warning message referring to it as `ephemeral` rather than `embedded` to avoid confusion for the dbt users, though should be fine either way

sqlmesh/dbt/model.py
toriwei: will update with this fix + changing warning message to `ephemeral`. thanks for catching!

PR #5412, opened 2025-09-19
Feat!: bring dbt node information through to SQLMesh

erindru:
This PR does the following:
 - Remove `dbt_name` from `_Node` and replace it with `dbt_node_info`  (as opposed to adding more top-level fields directly to `_Node`)
 - Since this gets saved to state, I also added `dbt_node_info` to the model metadata values to help ensure that state stays in sync with reality
 - Added a migration to migrate `dbt_name` -> `dbt_node_info`
    - This is a lossy migration because `dbt_name` only represents the `unique_id` and not the other fields
    - However, since it's part of the metadata hash, after migration, on the next plan, the user will get something like:    
      Screenshot From 2025-09-19 11-27-45
       (that is, a low impact virtual update to synchronise state with reality)
  - Exposes the dbt fqn's in `sqlmesh_dbt list` to show that theyre working
    Screenshot From 2025-09-19 11-05-05

Also, some housekeeping in the test suite:
 - I noticed @eakmanrq and myself had independently implemented `create_empty_project` fixtures, so I combined them and moved them to a common location
 - I also moved the `jaffle_shop_duckdb` fixture out of the sqlmesh_dbt cli conftest and into the main dbt conftest so I could use it in `test_intregration` to verify the dbt names matched the output of `dbt list`


sqlmesh/core/audit/definition.py
```diff
@@ -150,6 +151,10 @@ def __str__(self) -> str:
         path = f": {self._path.name}" if self._path else ""
         return f"{self.__class__.__name__}<{self.name}{path}>"
 
+    @property
+    def dbt_node_info(self) -> t.Optional[DbtNodeInfo]:
+        return self.dbt_node_info_
```
izeigerman, line 156: Looks like the property method doesn't add anything extra. Why not use `dbt_node_info` as the attribute name directly?

sqlmesh/core/model/definition.py
```diff
@@ -1197,6 +1197,9 @@ def metadata_hash(self) -> str:
                 for k, v in sorted(args.items()):
                     metadata.append(f"{k}:{gen(v)}")
 
+            if self.dbt_node_info:
+                metadata.append(self.dbt_node_info.json(sort_keys=True))
```
izeigerman, line 1201: I think we need to extend the `./.circleci/test_migration.sh` test to also run on `sushi_dbt` project and not just sushi, since we now have dbt-specific logic when hashing.

sqlmesh/core/audit/definition.py
izeigerman: ah, the mixin. nvm

sqlmesh/core/model/definition.py
erindru: I've added this and I think it's failing in the way you expected it to:

```
**Metadata Updated:**
- `sushi.customer_revenue_by_day`
- `sushi.customers`
- `sushi.top_waiters`
- `sushi.waiter_as_customer_by_day`
- `sushi.waiter_revenue_by_day_v1`
- `sushi.waiter_revenue_by_day_v2`
- `sushi.waiters`
- `sushi_raw.items`
- `sushi_raw.order_items`
- `sushi_raw.orders`
- `sushi_raw.waiter_names`

Exited with code exit status 1
```

This is correct because after the migration in this PR runs and a new plan is made, the extra fields get read from disk and show as a metadata change because they go from being empty in state to being populated in state

PR #5420, opened 2025-09-21
Feat(sqlmesh_dbt): Select based on dbt name, not sqlmesh name

erindru:
This builds on the dbt node info exposed in #5412 to enable its usage in the selector engine.

Previously, to use `sqlmesh_dbt --select`, you had to use the sqlmesh-native model names.

This PR:
 - Adds `dbt_mode` flag to the `Context` to pass to the `Selector` engine
 - Sets this flag in the `sqlmesh_dbt` CLI when constructing the `Context`
 - Uses it to indicate that dbt fqn's should be matched against instead of the sqlmesh model name

The selectors still _resolve_ to SQLMesh fqn's so everything continues to match up downstream


sqlmesh/core/context.py
```diff
@@ -368,6 +370,7 @@ def __init__(
         load: bool = True,
         users: t.Optional[t.List[User]] = None,
         config_loader_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
+        dbt_mode: bool = False,
```
erindru: I called this generically `dbt_mode` instead of something more specific to selectors incase we discovered other behaviour that we wanted to attach to it

themisvaltinos: I wonder if we need the extra flag instead of using the project_type, for when self._project_type == c.DBT

sqlmesh/core/selector.py
```diff
@@ -194,10 +197,9 @@ def evaluate(node: exp.Expression) -> t.Set[str]:
                     return {
                         fqn
                         for fqn, model in all_models.items()
-                        if fnmatch.fnmatchcase(model.name, node.this)
+                        if fnmatch.fnmatchcase(self._model_name(model), node.this)
```
themisvaltinos, line 199: I see the wildcard matching happens outside `_pattern_to_model_fqns` which seems straightforward and should work, but could you add a test for this as well to ensure nothing breaks in the future if someone modifies this? unless if I missed it and there is a wildcard test for dbt model names already

sqlmesh/core/context.py
themisvaltinos: actually never mind I realise this is for when someone runs `sqlmesh_dbt` rather than a dbt project with the native sqlmesh commands so this is indeed needed

izeigerman: I don't like this flag at all. Can we instead have a base Selector class and 2 implementations and then we can configure the implementation used when creating a Context instance.

sqlmesh/core/selector.py
erindru: Well spotted, I added some tests to exercise this codepath based on the [dbt docs](https://docs.getdbt.com/reference/node-selection/methods).

Note that the patterns currently need a `*` in them to trigger this branch (even though `?`, `[` and `]` could be used as well) so this will need more work in future if we find people using these in the wild

sqlmesh/core/context.py
erindru: Sure, i've refactored `Selector` to be an abstract base class and then implemented it as `NativeSelector` and `DbtSelector`

izeigerman (approved):
Thanks!

PR #5428, opened 2025-09-23
fix: bigquery snowflake source columns support

eakmanrq:
Dataframe tests were accidentally disabled in this PR: https://github.com/TobikoData/sqlmesh/pull/5117

This was then fixed in this PR: https://github.com/TobikoData/sqlmesh/pull/5423

Which then exposed this issue. Therefore this PR fixes support for ignored changes for BQ and Snowflake using Pandas.

Confirmed CI checks for Snowflake/BQ pass.

### DIFF
--- a/tests/core/engine_adapter/test_bigquery.py
+++ b/tests/core/engine_adapter/test_bigquery.py
@@ -487,7 +487,13 @@ def test_merge_pandas(make_mocked_engine_adapter: t.Callable, mocker: MockerFixt
     retry_resp_call.errors = None
     retry_mock.return_value = retry_resp
     db_call_mock.return_value = AttributeDict({"errors": None})
-    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    df = pd.DataFrame(
+        {
+            "id": [1, 2, 3],
+            "ts": ["2025-01-01 00:00:00", "2025-01-01 00:00:00", "2025-01-01 00:00:00"],
+            "val": [7, 8, 9],
+        }
+    )
     adapter.merge(
         target_table="target",
         source_table=df,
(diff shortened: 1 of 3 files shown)


tests/core/engine_adapter/integration/config.yaml
```diff
@@ -154,9 +154,6 @@ gateways:
   inttest_bigquery:
     connection:
       type: bigquery
-      method: service-account
-      keyfile: {{ env_var('BIGQUERY_KEYFILE') }}
-      check_import: false
```
georgesittas, line 159: This likely needs to be re-instated. I'm assuming it was removed to locally run BigQuery tests but CI will fail w/out it right?

eakmanrq: yes thanks will revert it back

tests/core/engine_adapter/test_bigquery.py
```diff
@@ -487,7 +487,13 @@ def temp_table_exists(table: exp.Table) -> bool:
...
     retry_mock.return_value = retry_resp
     db_call_mock.return_value = AttributeDict({"errors": None})
-    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    df = pd.DataFrame(
+        {
+            "id": [1, 2, 3],
+            "ts": ["2025-01-01 00:00:00", "2025-01-01 00:00:00", "2025-01-01 00:00:00"],
+            "val": [7, 8, 9],
+        }
+    )
```
eakmanrq, line 496: The actual values in the dataframe didn't matter before so they just had placeholder values but due to the ordering check they are now updated to match what we claim the structure is in the merge call.

PR #5431, opened 2025-09-23
fix: handle duplicate audit names

toriwei:
create unique `canonical_name` to prevent duplicate audit names


sqlmesh/dbt/test.py
```diff
@@ -151,6 +151,9 @@ def to_sqlmesh(self, context: DbtContext) -> Audit:
         blocking = self.severity == Severity.ERROR
 
         audit: Audit
+        if self.package_name != context.project_name:
+            self.name = f"{self.package_name}.{self.name}"
```
erindru: This subtly mutates `self.name` and makes it different before / after calling `to_sqlmesh()`.

Have you considered using a local variable for this instead and passing it for the `name=` parameter in the `StandaloneAudit`/`ModelAudit` constructors?

crericha (approved):

PR #5432, opened 2025-09-23
Chore: Break up the core integration tests

izeigerman:


tests/core/integration/test_change_scenarios.py
```diff
@@ -0,0 +1,1939 @@
+from __future__ import annotations
```
izeigerman, line 1: This will likely become the next candidate for breakup due to the vagueness of its scope. I’m looking for feedback on how to improve it

PR #5435, opened 2025-09-24
Feat(dbt): Add support for dbt custom materializations

themisvaltinos:
This update adds support for dbt custom materializations which allows users to define their own strategies via jinja macros. 

Docs:  https://docs.getdbt.com/guides/create-new-materializations?step=2

### DIFF
--- a/sqlmesh/core/model/kind.py
+++ b/sqlmesh/core/model/kind.py
@@ -120,6 +120,10 @@ class ModelKindMixin:
         return self.model_kind_name == ModelKindName.MANAGED
 
     @property
+    def is_dbt_custom(self) -> bool:
+        return self.model_kind_name == ModelKindName.DBT_CUSTOM
+
+    @property
     def is_symbolic(self) -> bool:
         """A symbolic model is one that doesn't execute at all."""
         return self.model_kind_name in (ModelKindName.EMBEDDED, ModelKindName.EXTERNAL)
@@ -170,6 +174,7 @@ class ModelKindName(str, ModelKindMixin, Enum):
     EXTERNAL = "EXTERNAL"
     CUSTOM = "CUSTOM"
     MANAGED = "MANAGED"
+    DBT_CUSTOM = "DBT_CUSTOM"
 
     @property
     def model_kind_name(self) -> t.Optional[ModelKindName]:
@@ -887,6 +892,52 @@ class ManagedKind(_ModelKind):
         return False
 
 
+class DbtCustomKind(_ModelKind):
+    name: t.Literal[ModelKindName.DBT_CUSTOM] = ModelKindName.DBT_CUSTOM
+    materialization: str
+    adapter: str = "default"
+    definition: str
+    dialect: t.Optional[str] = Field(None, validate_default=True)
+
+    _dialect_validator = kind_dialect_validator
+
+    @field_validator("materialization", "adapter", "definition", mode="before")
+    @classmethod
+    def _validate_fields(cls, v: t.Any) -> str:
+        return validate_string(v)
+
+    @property
+    def data_hash_values(self) -> t.List[t.Optional[str]]:
+        return [
+            *super().data_hash_values,
+            self.materialization,
+            self.definition,
+            self.adapter,
+            self.dialect,
+        ]
+
+    @property
+    def metadata_hash_values(self) -> t.List[t.Optional[str]]:
+        return [
+            *super().metadata_hash_values,
+        ]
+
+    def to_expression(
+        self, expressions: t.Optional[t.List[exp.Expression]] = None, **kwargs: t.Any
+    ) -> d.ModelKind:
+        return super().to_expression(
+            expressions=[
+                *(expressions or []),
+                *_properties(
+                    {
+                        "materialization": exp.Literal.string(self.materialization),
+                        "adapter": exp.Literal.string(self.adapter),
+                    }
+                ),
+            ],
+        )
+
+
 class EmbeddedKind(_ModelKind):
     name: t.Literal[ModelKindName.EMBEDDED] = ModelKindName.EMBEDDED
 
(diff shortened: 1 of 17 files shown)


sqlmesh/core/model/kind.py
```diff
@@ -887,6 +892,52 @@ def supports_python_models(self) -> bool:
         return False
 
 
+class DbtCustomKind(_ModelKind):
+    name: t.Literal[ModelKindName.DBT_CUSTOM] = ModelKindName.DBT_CUSTOM
+    materialization: str
+    adapter: str = "default"
```
izeigerman, line 898: what is this adapter?

sqlmesh/core/model/kind.py
```diff
@@ -887,6 +892,52 @@ def supports_python_models(self) -> bool:
...
+            self.materialization,
+            self.definition,
+            self.adapter,
+            self.dialect,
+        ]
+
+    @property
+    def metadata_hash_values(self) -> t.List[t.Optional[str]]:
+        return [
+            *super().metadata_hash_values,
```
izeigerman: What is the point of this override?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -747,7 +748,8 @@ def _evaluate_snapshot(
             adapter.transaction(),
             adapter.session(snapshot.model.render_session_properties(**render_statements_kwargs)),
         ):
-            adapter.execute(model.render_pre_statements(**render_statements_kwargs))
+            if not snapshot.is_dbt_custom:
```
izeigerman: Does the custom materialization strategy calls the pre-hooks?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2593,6 +2610,139 @@ def get_custom_materialization_type_or_raise(
...
+        query_or_df: QueryOrDF,
+        model: Model,
+        is_first_insert: bool,
+        render_kwargs: t.Dict[str, t.Any],
+        create_only: bool = False,
+        **kwargs: t.Any,
+    ) -> None:
+        from sqlmesh.dbt.builtin import create_builtin_globals
+
+        jinja_macros = getattr(model, "jinja_macros", JinjaMacroRegistry())
```
izeigerman: why do we need to use `getattr` instead of accessing the attribute directly?

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2593,6 +2610,139 @@ def get_custom_materialization_type_or_raise(
...
+            **existing_globals,
+            "this": relation_info,
+            "database": parts.catalog,
+            "schema": parts.db,
+            "identifier": parts.name,
+            "target": existing_globals.get("target", {"type": self.adapter.dialect}),
+            "execution_dt": kwargs.get("execution_time"),
+        }
+
+        context = create_builtin_globals(
```
izeigerman: Why do we manually construct the context instead of relying on `JinjaMacros` to do it for us?

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: yes internally using `run_hooks` so it's up to the user when they should call them or if the specific materialisation doesn't call them at all: https://docs.getdbt.com/guides/create-new-materializations?step=2#run-pre-hooks

sqlmesh/core/model/kind.py
themisvaltinos: it's to allow users to define different materialisation tied to different adapters and for dispatch then to appropriately match the correct one used
```
{% materialization my_materialization_name, default %}
 -- cross-adapter materialization... assume Redshift is not supported
{% endmaterialization %}

{% materialization my_materialization_name, adapter='redshift' %}
-- override the materialization for Redshift
{% endmaterialization %}
```

test for this example: `test_adapter_specific_materialization_override`

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: yes good point I forgot all our model classes have jinja_macros

sqlmesh/core/model/kind.py
themisvaltinos: thank you good catch. I had originally the definition there and then realised it should be in the data hash and moved it but forgot to remove the `metadata_hash_values` entirely since it's not needed will do

sqlmesh/core/snapshot/evaluator.py
themisvaltinos: so that I didn't use its private method, but I revised it to use `jinjamacros` instead if this is what you meant. if you meant to do it through `build_environment` that got quite complicated and it updates internally the macros that were needed which lead to failures. if you mean why this entire logic is needed in the first place it is because we need to inject runtime values specific to the materialization, like the actual sql of the query being materialized, the pre/post hooks, the physical table name parts if it’s a vde etc.

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -2593,6 +2610,137 @@ def get_custom_materialization_type_or_raise(
     raise SQLMeshError(f"Custom materialization '{name}' not present in the Python environment")
 
 
+class DbtCustomMaterialization(MaterializableStrategy):
```
izeigerman: Let's follow the existing naming convention and add `Strategy` at the end of the name

sqlmesh/core/snapshot/evaluator.py
izeigerman: ok, can we instead extend the materialization strategy with:
```
    def run_pre_statements(...)
    
    def run_post_statements(...)
```
and in the default implementation we do
```
adapter.execute(model.render_pre_statements(**render_statements_kwargs))
```
while for `DbtCustomMaterializationStrategy` it's a noop

themisvaltinos: yes that seems better, revised it with run methods instead in the materialisation strategy

izeigerman (approved):

PR #5436, opened 2025-09-24
feat: Databricks grants

eakmanrq:


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -149,6 +151,109 @@ def spark(self) -> PySparkSession:
...
+                    "table_privileges",
+                    db="information_schema",
+                    catalog=catalog_name,
+                )
+            )
+            .where(
+                exp.and_(
+                    exp.column("table_catalog").eq(exp.Literal.string(catalog_name.lower())),
+                    exp.column("table_schema").eq(exp.Literal.string(schema_name.lower())),
+                    exp.column("table_name").eq(exp.Literal.string(table_name.lower())),
```
newtonapple, line 229: I assume the reason we lower case everything is because we are assuming Databricks users will be using Unity catalog? Is that always the case? I know some deployments still use hive metastore. If I remember correctly, it table names etc. can be case sensitive there.

eakmanrq: The engine adapter itself assumes Unity Catalog. So we don't need to worry about hive metastore compatibility.

PR #5443, opened 2025-09-26
Feat(sqlmesh_dbt): Implement --model and --resource-type

erindru:
This PR adds the `-m/--model/--models` argument to the `sqlmesh_dbt` CLI.

It does this by implementing support for the `resource_type` [selector method](https://docs.getdbt.com/reference/node-selection/methods#resource_type) in the SQLMesh selector engine and then if `--models` is supplied, injecting `resource_type:model` into the selector expression alongside the user-supplied values.

Note that `--models` is different to `--select`:
 - `--models` just includes model nodes - so excludes seeds, sources and tests
 - `--select` includes all nodes


sqlmesh_dbt/selectors.py
```diff
@@ -4,7 +4,45 @@
 logger = logging.getLogger(__name__)
 
 
-def to_sqlmesh(dbt_select: t.Collection[str], dbt_exclude: t.Collection[str]) -> t.Optional[str]:
+def consolidate(
```
izeigerman, line 7: Why even expose this outside this module? Why not make it a part of a single unified `to_sqlmesh` interface?

izeigerman (approved):

sqlmesh_dbt/cli.py
```diff
@@ -33,15 +33,39 @@ def _cleanup() -> None:
...
+    "saved_query",
+    "source",
+    "analysis",
+    "model",
+    "test",
+    "unit_test",
+    "exposure",
+    "snapshot",
+    "seed",
+    "default",
```
themisvaltinos, line 62: small question on the last two, what will `all` and `default` be in the sqlmesh context? I suppose this is a question for later on when multiple `resource types` support is added, but will these two mirror dbt as well?

erindru: Currently, unimplemented :) but at a guess i'd say they would be aliases for the same thing because SQLMesh includes everything by default

sqlmesh_dbt/selectors.py
erindru: I thought about that but I wanted to keep the surface of `to_sqlmesh()` simple and take the minimum required parameters.

`combine()` still operates in dbt-land and produces consolidated selectors in dbt format. The goal is to not have to always pass down `select`, `exclude`, `resource_type`, `models` etc parameters through every layer when really all of these parameters can be described in terms of `select`/`exclude`.

So they get consolidated early to reduce the number of parameters being passed around and then the lower layers then just call `to_sqlmesh()` on just `select` / `exclude` and dont have to worry about or consider any of the other parameters. 

You can see this in `DbtOperations._plan_builder` which deliberately only needs select/exclude, not all the other options that may be passed.

PR #5444, opened 2025-09-26
feat: add grants for BigQuery

newtonapple:


sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -1297,6 +1303,115 @@ def _session_id(self) -> t.Any:
     def _session_id(self, value: t.Any) -> None:
         self._connection_pool.set_attribute("session_id", value)
 
+    def _get_bq_dataset_location(self, project: str, dataset: str) -> str:
+        return self._db_call(self.client.get_dataset, dataset_ref=f"{project}.{dataset}").location
```
eakmanrq, line 1310: Do you know if `dataset_ref` needs to be properly quoted? Or are you assuming the strings passed in will have quotes if needed?

sqlmesh/core/engine_adapter/bigquery.py
```diff
@@ -1297,6 +1303,115 @@ def _session_id(self) -> t.Any:
...
+
+    def _get_current_grants_config(self, table: exp.Table) -> GrantsConfig:
```
eakmanrq: Can you override `_get_grant_expression` instead?

sqlmesh/core/engine_adapter/bigquery.py
newtonapple: Yeah, I checked their [API](https://cloud.google.com/bigquery/docs/samples/bigquery-get-dataset#bigquery_get_dataset-python). I don't think we need to quote it.

PR #5447, opened 2025-09-26
chore: grant mixin and normalize

eakmanrq:
Creates a mixin for grants that are using the information schema in order to share common logic.

Also properly normalizes the Grant Principle which is why there are quotes around the identifiers now.


tests/core/engine_adapter/test_postgres.py
```diff
@@ -202,10 +202,10 @@ def test_sync_grants_config(make_mocked_engine_adapter: t.Callable, mocker: Mock
     sql_calls = to_sql_calls(adapter)
     assert len(sql_calls) == 4
 
-    assert 'GRANT SELECT ON "test_schema"."test_table" TO user1, user2' in sql_calls
-    assert 'GRANT INSERT ON "test_schema"."test_table" TO user3' in sql_calls
-    assert 'REVOKE SELECT ON "test_schema"."test_table" FROM old_user' in sql_calls
-    assert 'REVOKE UPDATE ON "test_schema"."test_table" FROM admin_user' in sql_calls
+    assert 'GRANT SELECT ON "test_schema"."test_table" TO "user1", "user2"' in sql_calls
```
erindru, line 205: Nice, the previous implementation that didn't quote was asking for trouble

PR #5450, opened 2025-09-27
Fix: Support dbt macro dispatch search order

vchan:


sqlmesh/dbt/adapter.py
```diff
@@ -122,48 +122,30 @@ def dispatch(
...
-                    (macro_package, macro_name): macro_callable
-                    for macro_name, macro_callable in macros.items()
-                    if macro_name.endswith(macro_suffix)
-                }
-            )
 
-        if candidates:
-            sorted_candidates = sorted(candidates, key=_relevance)
-            return candidates[sorted_candidates[0]]
+            for prefix in (f"{target_type}__", "default__", ""):
```
themisvaltinos, line 146: why is the empty string needed? I might be wrong but also looking at the [docs](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch#a-simple-example)  when dbt doesn't find either the adapter or namespace in dispatch it always falls back to `default__` as prefix

PR #5454, opened 2025-09-29
fix: ignore partition_by field for ephemeral models

toriwei:
`partition_by` is not applicable for ephemeral models. dbt ignores this config if set


erindru (approved):
Nice, thanks for also updating the test!

PR #5462, opened 2025-09-30
feat!: support unicode characters

eakmanrq:
BREAKING: If users currently have unicode characters then their table name will update to actually include the unicode name. Still thinking through how to migrate this. 

This PR affects `santize_name` function would before would replace unicode characters with underscores. This is used for populating the file cache and creating table names. Therefore someone could have had a SQLMesh project that would run but it would not have correct names and they could get conflicts when reading from cache or writing. 

Therefore this PR updates `sanitize_name` to maintain the unicode characters while still removing things like double quotes from the resulting name.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -300,8 +300,8 @@ workflows:
           name: cloud_engine_<< matrix.engine >>
           context:
             - sqlmesh_cloud_database_integration
-          requires:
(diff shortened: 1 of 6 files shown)


eakmanrq:
Views don't get santiized. Therefore just the underlying tables get their unicode characters replaced which is still not optimal but not worth making this breaking change for users at this time.

PR #5469, opened 2025-10-01
feat: support batching of janitor cleanup

eakmanrq:
Prior to this PR, janitor would always read all expired snapshots into memory. This is fine during normal operations but if for some reason janitor has not run for a while this can result in a lot of snapshots being read into memory. 

This PR allows users to configure the janitor to batch up what is being processed. It does this by adjusting the "current_ts" (`batch_start`) and then incrementing this by the provided `batch_size_seconds`. So for example a user could provide the following to process the last 7 days of expired snapshots 1 day at a time:

```
sqlmesh janitor --batch-start "7 days ago" --batch-seconds 86400
```

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -855,12 +855,42 @@ class GenericContext(BaseContext, t.Generic[C]):
         return completion_status
 
     @python_api_analytics
-    def run_janitor(self, ignore_ttl: bool) -> bool:
+    def run_janitor(
+        self,
+        ignore_ttl: bool,
+        batch_start: t.Optional[TimeLike] = None,
+        batch_size_seconds: t.Optional[int] = None,
+    ) -> bool:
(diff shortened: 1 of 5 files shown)


sqlmesh/core/context.py
```diff
@@ -855,12 +855,41 @@ def _has_environment_changed() -> bool:
...
+                        )
+
+                current_ts = now_timestamp()
+
+                if batch_start is not None:
+                    batch_start = to_timestamp(batch_start)
+
+                batch_delta_ms = None if batch_size_seconds is None else batch_size_seconds * 1000
+
+                next_batch_ts = batch_start if batch_start is not None else current_ts
```
newtonapple: We could merge this if condition w/ the previous `if batch_start is not None` in line 878.

newtonapple (approved):

PR #5475, opened 2025-10-02
fix: parse column from cast expression for ScdType2 models

toriwei:


sqlmesh/dbt/model.py
```diff
@@ -172,6 +172,17 @@ def _validate_check_cols(cls, v: t.Union[str, t.List[str]]) -> t.Union[str, t.Li
             return "*"
         return ensure_list(v)
 
+    @field_validator("updated_at", mode="before")
+    @classmethod
+    def _validate_updated_at(cls, v: t.Optional[str]) -> t.Optional[str]:
+        if v is None:
+            return None
+        parsed = d.parse_one(v)
+        if isinstance(parsed, exp.Cast) and isinstance(parsed.this, exp.Column):
```
eakmanrq, line 186: Can you add a comment explaining that both scd type 2 expects a column and the logic for it does a cast anyways so this shouldn't be needed? Could be an issue later so it would at least explaining the reasoning at this time.

toriwei: updated with:
```

        """
        Extract column name if updated_at contains a cast.

        SCDType2ByTimeKind and SCDType2ByColumnKind expect a column, and the casting is done later.
        """
```

PR #5478, opened 2025-10-03
fix: cluster_by as single string with multiple columns

geooo109:
This PR aims to add support for more `cluster_by` configs.
For example:
```
cluster_by='col1,col2'
cluster_by=['to_date(col1),col2']
cluster_by='"col1,col2"'
```

**DOCS**
[DBT cluster_by config](https://docs.getdbt.com/reference/resource-configs/snowflake-configs#using-cluster_by)

### DIFF
--- a/sqlmesh/dbt/model.py
+++ b/sqlmesh/dbt/model.py
@@ -157,18 +157,12 @@ class ModelConfig(BaseModelConfig):
 
     @field_validator(
         "unique_key",
+        "cluster_by",
         "tags",
         mode="before",
     )
     @classmethod
     def _validate_list(cls, v: t.Union[str, t.List[str]]) -> t.List[str]:
-        return ensure_list(v)
-
-    @field_validator("cluster_by", mode="before")
-    @classmethod
-    def _validate_cluster_by(cls, v: t.Union[str, t.List[str]]) -> t.Union[str, t.List[str]]:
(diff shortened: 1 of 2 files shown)


sqlmesh/dbt/model.py
```diff
@@ -157,14 +157,20 @@ class ModelConfig(BaseModelConfig):
...
     )
     @classmethod
     def _validate_list(cls, v: t.Union[str, t.List[str]]) -> t.List[str]:
         return ensure_list(v)
 
+    @field_validator("cluster_by", mode="before")
+    @classmethod
+    def _validate_cluster_by(cls, v: t.Union[str, t.List[str]]) -> t.Union[str, t.List[str]]:
+        if isinstance(v, str):
+            return [c.strip() for c in v.split(",")]
```
georgesittas: What happens when a column name has literal commas in it? How does dbt handle that?

geooo109: You have to wrap it like `cluster_by='"col1,col2"'`, so needs extra handling.

georgesittas: Note: we discussed offline and we shouldn't do this. Instead, @geooo109 will leverage SQLGlot's parser to properly recognize the comma-separated expressions in the `cluster_by` config.

PR #5480, opened 2025-10-03
Feat(dbt): Add support for transaction in dbt pre and post hooks

themisvaltinos:
This adds support for `transaction` in dbt pre and post hooks. By default hooks are executed inside of the same transaction as the model being created. Hooks that need to run outside a transaction are for example when executing a VACUUM in Redshift or inserting into an audit table at the start of a run where the insert should persist even if the model fails. Running the hook outside a transaction ensures it won’t be rolled back.

These are set in a dbt project either with the `before_begin` and `after_commit` macro:
```
{{
  config(
    pre_hook=before_begin("SQL-statement"),
    post_hook=after_commit("SQL-statement")
  )
}}
```
or by setting the `transaction` flag in a dict:
```
{{
  config(
    pre_hook={
      "sql": "SQL-statement",
      "transaction": False
    },
    post_hook={
      "sql": "SQL-statement",
      "transaction": False
    }
  )
}}
```
By default and without setting the transaction the hooks are executed inside the transaction.

Docs: https://docs.getdbt.com/reference/resource-configs/pre-hook-post-hook#transaction-behavior

### DIFF
--- a/tests/dbt/test_transformation.py
+++ b/tests/dbt/test_transformation.py
@@ -2849,7 +2849,7 @@ def test_dbt_hooks_with_transaction_flag_execution(sushi_test_dbt_context: Conte
     assert shared_table["hook_name"][2] == "after_commit"
     assert shared_table["execution_order"][2] == 3
 
-    # the timestamps also should be monotonically increasing for teh same reason
+    # the timestamps also should be monotonically increasing for the same reason
     for i in range(len(shared_table) - 1):
         assert shared_table["created_at"][i] <= shared_table["created_at"][i + 1]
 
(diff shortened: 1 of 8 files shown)


tests/dbt/test_transformation.py
```diff
@@ -2678,3 +2678,180 @@ def test_selected_resources_context_variable(
...
+
+    assert shared_table["id"][1] == 2
+    assert shared_table["hook_name"][1] == "inside_post"
+    assert shared_table["execution_order"][1] == 2
+
+    assert shared_table["id"][2] == 3
+    assert shared_table["hook_name"][2] == "after_commit"
+    assert shared_table["execution_order"][2] == 3
+
+    # the timestamps also should be monotonically increasing for teh same reason
```
erindru: typo: teh

erindru (approved):

PR #5481, opened 2025-10-03
feat: add `batch_concurrency` to `ModelDefaultsConfig`

etonlels:
This change allows users to set a default `batch_concurrency` value in their
project configuration that will be applied to incremental models.

- Added `batch_concurrency` field to `ModelDefaultsConfig`
- Updated `create_model_kind()` to apply `batch_concurrency` default to models
  inheriting from `_IncrementalBy` when not explicitly set
- Skips applying default for subclasses with hardcoded `batch_concurrency`
  (e.g., `IncrementalByUniqueKeyKind` which has `batch_concurrency: Literal[1]`)
- Added comprehensive tests covering default behavior and edge cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -45,6 +45,7 @@ class ModelDefaultsConfig(BaseConfig):
         allow_partials: Whether the models can process partial (incomplete) data intervals.
         enabled: Whether the models are enabled.
         interval_unit: The temporal granularity of the models data intervals. By default computed from cron.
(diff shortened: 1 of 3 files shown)


erindru:
Does it make sense to add `batch_concurrency` without also adding `batch_size`?

erindru:
It's just `batch_size` because `batch_size` and `batch_concurrency` are related and work together

PR #5485, opened 2025-10-06
Feat: Add config flag to infer the state schema per dbt target

erindru:
Currently, when storing state in the warehouse, by default SQLMesh assumes that you only have a single project accessing the warehouse and creates a schema called `sqlmesh` to store state in. This is because it _also_ assumes that you want to use Virtual Data Environments.

However, this creates a problem in dbt projects with existing workflows that are not built around VDE's. A common pattern is to use different _targets_ to point to the same warehouse and just override the default schema.

For example, a `dev` target may populate models into a `dev` schema while a `prod` target may put them in a `prod` schema, but these schemas exist side by side in the same warehouse. Creating models using `--target dev` allows analysts to test things out and then deploy by running the models against `--target prod`.

This creates a problem for SQLMesh because it assumes a single state schema but the state for these targets should not overlap.

So this PR:
 - Adds a new section to the root config called `dbt` with, for now, a single flag `infer_state_schema_name`
 - Uses this flag to, if set, infer a default state schema name based on the dbt profile name and target schema
 - Enables this flag by default for `sqlmesh init -t dbt` so that dbt projects by default have isolated state between targets

Note that users can still override the inferred schema by setting the state schema manually in `sqlmesh.yaml` as [documented here](https://sqlmesh.readthedocs.io/en/stable/integrations/dbt/?h=dbt#selecting-a-different-state-connection). Existing projects will not have this new property set so will continue to behave as usual.

### DIFF
--- a/sqlmesh/dbt/loader.py
+++ b/sqlmesh/dbt/loader.py
@@ -73,16 +73,28 @@ def sqlmesh_config(
     if threads is not None:
         # the to_sqlmesh() function on TargetConfig maps self.threads -> concurrent_tasks
         profile.target.threads = threads
-        
+
     gateway_kwargs = {}
     if infer_state_schema_name:
         profile_name = context.profile_name
+
         # Note: we deliberately isolate state based on the target *schema* and not the target name.
         # It is assumed that the project will define a target, eg 'dev', and then in each users own ~/.dbt/profiles.yml the schema
         # for the 'dev' target is overriden to something user-specific, rather than making the target name itself user-specific.
         # This means that the schema name is the indicator of isolated state, not the target name which may be re-used across multiple schemas.
         target_schema = profile.target.schema_
-        gateway_kwargs["state_schema"] = f"sqlmesh_state_{profile_name}_{target_schema}"
+
+        # dbt-core doesnt allow schema to be undefined, but it does allow an empty string, and then just
+        # fails at runtime when `CREATE SCHEMA ""` doesnt work
+        if not target_schema:
+            raise ConfigError(
+                f"Target '{profile.target_name}' does not specify a schema.\n"
+                "A schema is required in order to infer where to store SQLMesh state"
+            )
+
+        inferred_state_schema_name = f"sqlmesh_state_{profile_name}_{target_schema}"
+        logger.info("Inferring state schema: %s", inferred_state_schema_name)
+        gateway_kwargs["state_schema"] = inferred_state_schema_name
 
     return Config(
         loader=loader,


sqlmesh/core/config/root.py
```diff
@@ -137,6 +143,7 @@ class Config(BaseConfig):
     )
     default_scheduler: SchedulerConfig = BuiltInSchedulerConfig()
     default_gateway: str = ""
+    state_schema_naming_pattern: t.Optional[str] = None
```
erindru: As always, not sold on the name. I wanted to communicate that it was a pattern that could contain dynamic substitutions rather than a string literal

sqlmesh/cli/project_init.py
```diff
@@ -116,7 +116,13 @@ def _gen_config(
     - invalidselectstarexpansion
     - noambiguousprojections
 """,
-        ProjectTemplate.DBT: f"""# --- Virtual Data Environment Mode ---
+        ProjectTemplate.DBT: f"""# --- State ---
+# This default configuration ensures that each dbt target gets its own isolated state.
+# If this is undesirable, you may configure the state connection manually.
+# https://sqlmesh.readthedocs.io/en/stable/integrations/dbt/?h=dbt#selecting-a-different-state-connection
+state_schema_naming_pattern: sqlmesh_state_@{{dbt_profile_name}}_@{{dbt_target_name}}
```
erindru: The `{{` is to just escape `{` in a f-string, this is not jinja.

I wasnt sure what syntax to use so I used what we used for eg [@resolve_template](https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#resolve_template) and [schema_location_mapping](https://sqlmesh.readthedocs.io/en/stable/integrations/engines/trino/?h=trino#schemas)

sqlmesh/dbt/loader.py
```diff
@@ -67,10 +68,19 @@ def sqlmesh_config(
...
+        # Note: Profile.load() mutates `context` and will have already raised an exception if profile_name is not set,
+        # but mypy doesnt know this because the field is defined as t.Optional[str]
+        raise ConfigError(f"profile name must be set")
+
     return Config(
         loader=loader,
         model_defaults=model_defaults,
         variables=variables or {},
+        dbt_config_info=DbtConfigInfo(
+            profile_name=dbt_profile_name or context.profile_name,
```
izeigerman: If we check `context.profile_name` above why not use it always since it's the source of truth?

sqlmesh/core/config/root.py
erindru: I've renamed this to `infer_state_schema_name`, refactored this to hardcode the pattern and nested it under a `dbt` node in the root config

sqlmesh/dbt/loader.py
erindru: I've refactored this to do essentially that

sqlmesh/dbt/loader.py
```diff
@@ -67,16 +69,28 @@ def sqlmesh_config(
...
 
+    gateway_kwargs = {}
+    if infer_state_schema_name:
+        profile_name = context.profile_name
+        # Note: we deliberately isolate state based on the target *schema* and not the target name.
+        # It is assumed that the project will define a target, eg 'dev', and then in each users own ~/.dbt/profiles.yml the schema
+        # for the 'dev' target is overriden to something user-specific, rather than making the target name itself user-specific.
+        # This means that the schema name is the indicator of isolated state, not the target name which may be re-used across multiple schemas.
+        target_schema = profile.target.schema_
+        gateway_kwargs["state_schema"] = f"sqlmesh_state_{profile_name}_{target_schema}"
```
izeigerman: Should we check whether `target_schema` is empty?

erindru: Man, this was a rabbit hole.

So in dbt's `Credentials` object that defines this field, `schema` is required. If you don't define it, the base validator fails with something like:

```
Runtime Error
  Credentials in profile "jaffle_shop", target "postgres" invalid: 'schema' is a required property
```

(and some adapters, like the duckdb adapter, have a [hardcoded default](https://github.com/duckdb/dbt-duckdb/blob/0346c00583daa650529cb859b4ece4b3661a4d12/dbt/adapters/duckdb/credentials.py#L161) to pass this check).

dbt will, however, happily allow you to specify an empty string for the schema. If you do this, you push the failure to runtime:

```
23:11:33  Encountered an error:
Database Error
  zero-length delimited identifier at or near """"
  LINE 2: create schema if not exists ""
```

So i've changed our side to raise an exception if the schema is an empty string. If someone encounters this and their project works fine on dbt core then hopefully theyre willing to work with us to understand the correct behaviour here (and the workaround is of course to define the state schema name manually)

PR #5486, opened 2025-10-06
feat: batch expired snapshots

eakmanrq:
Batches the deletion of expired snapshots by paging over the candidates using the `updated_at`, `name`, and `identifier` as stable fields to do the paging. Simple viz explaining what is going on (just shows `updated_at` but know that it includes all 3 columns):

Screenshot 2025-10-06 at 9 46 31 AM

In that example, 3 batches are processed. Getting expired snapshots a limited since it would be unbounded otherwise. Instead of passing though snapshots directly into delete instead the "boundary" is passed into the delete. That becomes the "upper boundary" for the delete to then look backwards to perform the deletes. Limit is not needed since this is bounded. There is an underlying assumption that these upper boundaries are provided in paged order (on other words, the last batch isn't provided first since that would remove the benefit of batching). 

This approach compared to passing in the deleted snapshot ids directly has the following benefits:
* In a client/server architecture this places less trust on the input from the client by asking for just the boundary instead of the explicit list of snapshot ids to delete
* Also in client/server architecture, less information needs to be passed to perform the delete
* If something were to change between the calculation of the batch and the eventual deletion then this approach will correctly not delete the snapshot while if the explicit list if provided then the delete would still happen


sqlmesh/core/state_sync/common.py
```diff
@@ -215,3 +217,94 @@ def __iter__(self) -> t.Iterator[StateStreamContents]:
...
+        batch = state_reader.get_expired_snapshots(
+            current_ts=current_ts,
+            ignore_ttl=ignore_ttl,
+            batch_boundary=batch_boundary,
+        )
+
+        if not batch:
+            return num_expired_snapshots
+
+        callback(batch)
```
erindru: Instead of using a callback that just passes the batch, should this just be a generator that yields the batch?

sqlmesh/core/state_sync/base.py
```diff
@@ -315,15 +359,23 @@ def export(self, environment_names: t.Optional[t.List[str]] = None) -> StateStre
...
     def get_expired_snapshots(
-        self, current_ts: t.Optional[int] = None, ignore_ttl: bool = False
-    ) -> t.List[SnapshotTableCleanupTask]:
-        """Aggregates the id's of the expired snapshots and creates a list of table cleanup tasks.
+        self,
+        *,
+        batch_boundary: BatchBoundary,
+        current_ts: t.Optional[int] = None,
+        ignore_ttl: bool = False,
+    ) -> t.Optional[ExpiredSnapshotBatch]:
```
erindru, line 311: Was there a reason to make this `t.Optional` vs just returning an empty list if there are no expired snapshots like it did before?

erindru (commented):
I don't quite grok `UpperBatchBoundary` vs `LowerBatchBoundary` and when to use one vs the other. I can see the problem that it solves though

sqlmesh/core/state_sync/base.py
eakmanrq: Yes that is correct

sqlmesh/core/state_sync/common.py
eakmanrq: If I am thinking about this correctly, yielding implies there is a state being maintained which wouldn't be true in a client/server architecture.

sqlmesh/core/state_sync/base.py
```diff
@@ -72,6 +72,50 @@ def _schema_version_validator(cls, v: t.Any) -> int:
 SCHEMA_VERSION: int = MIN_SCHEMA_VERSION + len(MIGRATIONS) - 1
 
 
+class BatchBoundary(PydanticModel):
```
izeigerman: Should these classes be in `common` instead?

sqlmesh/core/state_sync/common.py
```diff
@@ -215,3 +217,94 @@ def __iter__(self) -> t.Iterator[StateStreamContents]:
...
+
+
+def for_each_expired_snapshot_batch(
+    state_reader: StateReader,
+    *,
+    current_ts: int,
+    ignore_ttl: bool = False,
+    callback: t.Callable[[ExpiredSnapshotBatch], None],
+    batch_size: t.Optional[int] = None,
+) -> int:
```
izeigerman: This says the return type is int but it doesn't appear to return anything. +1 to Erin's comment about the generator.

sqlmesh/core/state_sync/base.py
eakmanrq: Would you want promotion result to be in common? https://github.com/TobikoData/sqlmesh/blob/afe4d1e42288fdaf6e2326858a83dae9450a8de4/sqlmesh/core/state_sync/base.py#L119-L130

sqlmesh/core/state_sync/common.py
eakmanrq: Nevermind I get what you are saying now. Will update.

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -261,26 +263,39 @@ def invalidate_environment(self, name: str, protect_prod: bool = True) -> None:
...
-        for expired_snapshot_ids, cleanup_targets in self.snapshot_state._get_expired_snapshots(
-            self.environment_state.get_environments(), ignore_ttl=ignore_ttl, current_ts=current_ts
-        ):
-            self.snapshot_state.delete_snapshots(expired_snapshot_ids)
-            self.interval_state.cleanup_intervals(cleanup_targets, expired_snapshot_ids)
+        upper_batch_boundary = upper_batch_boundary or UpperBatchBoundary.include_all_boundary()
+        batch = self.get_expired_snapshots(
+            ignore_ttl=ignore_ttl,
+            current_ts=current_ts,
+            batch_boundary=upper_batch_boundary,
```
izeigerman: I don't quite understand why do we only need the upper boundary. I thought we always need both in order to paginate correctly.

sqlmesh/core/state_sync/common.py
eakmanrq: It would return the number of expired snapshots: https://github.com/TobikoData/sqlmesh/blob/afe4d1e42288fdaf6e2326858a83dae9450a8de4/sqlmesh/core/state_sync/common.py#L252-L253

This will shift to the caller though once updated to yield.

sqlmesh/core/state_sync/db/facade.py
eakmanrq: When getting batches to delete, you are using an upper boundary. This code you commented on is in `delete_expired_snapshots`

sqlmesh/core/state_sync/common.py
eakmanrq: updated

sqlmesh/core/state_sync/common.py
```diff
@@ -215,3 +225,165 @@ def __iter__(self) -> t.Iterator[StateStreamContents]:
...
+        )
+        batch_range = ExpiredBatchRange(
+            start=batch.batch_range.end,
+            end=LimitBoundary(batch_size=batch_size),
+        )
+
+
+def delete_expired_snapshots(
+    state_sync: StateSync,
+    snapshot_evaluator: SnapshotEvaluator,
```
izeigerman, line 436: IMHO, I don't think this belongs here. I don't believe State Sync should have anything to do with SnapshotEvaluator. Plus this creates a circular dependency between modules. 

The logical relationship should be SnapshotEvaluator -> StateSync.

Anticipating your question - yes, I think putting `cleanup_expired_views` in here was a mistake.

sqlmesh/core/state_sync/common.py
```diff
@@ -215,3 +225,165 @@ def __iter__(self) -> t.Iterator[StateStreamContents]:
                 yield EnvironmentsChunk(environments)
 
         return _StateStream()
+
+
+class ExpiredBatchRange(PydanticModel):
```
izeigerman, line 230: This is great 👍

sqlmesh/core/state_sync/db/snapshot.py
```diff
@@ -166,111 +168,158 @@ def get_expired_snapshots(
...
+            exp.column("name"),
+            exp.column("identifier"),
+        ]
+        start_values = [
+            exp.Literal.number(batch_range.start.updated_ts),
+            exp.Literal.string(batch_range.start.name),
+            exp.Literal.string(batch_range.start.identifier),
+        ]
+
+        start_condition = _expanded_tuple_comparison(columns, start_values, exp.GT)
```
izeigerman: This seems like something that could've been a part of `RowBoundary` itself. Or at the very least returning a list of literal expressions

sqlmesh/core/state_sync/db/facade.py
```diff
@@ -261,26 +262,39 @@ def invalidate_environment(self, name: str, protect_prod: bool = True) -> None:
...
     def get_expired_environments(self, current_ts: int) -> t.List[EnvironmentSummary]:
         return self.environment_state.get_expired_environments(current_ts=current_ts)
 
     @transactional()
     def delete_expired_snapshots(
-        self, ignore_ttl: bool = False, current_ts: t.Optional[int] = None
+        self,
+        ignore_ttl: bool = False,
+        current_ts: t.Optional[int] = None,
+        batch_range: t.Optional[ExpiredBatchRange] = None,
```
izeigerman: Why shouldn't we make this argument required?

sqlmesh/core/state_sync/db/snapshot.py
eakmanrq: Sure I felt that way too but thought it would get pushback. will update.

sqlmesh/core/state_sync/common.py
eakmanrq: Where do you want it located then?

sqlmesh/core/state_sync/db/facade.py
eakmanrq: I was debating that and open either way. Will update.

sqlmesh/core/state_sync/common.py
izeigerman: perhaps it's time to create the `core/janitor.py` module?

eakmanrq: Can we consider this out of scope for this PR? I'm open to following up if you think it is important but it doesn't seem like this should be required to get this change in.

izeigerman: sure. But I do think it's important

izeigerman (approved):

PR #5492, opened 2025-10-06
Fix: Add workaround to ignore source dependency if fqn matches model

crericha:
An idiom used in dbt projects is to define a model as both a model and a source to circumvent circular dependencies. 

For example:

- Model A is a cache of load times from an external source
- Model B is the loaded values form the external source

Model A uses model B to figure out the previous load time by the max timestamp
Model B uses Model A to know what timestamp it needs to fetch from without having to redo that expensive calculation

In this idiom, Model B can be defined also as a source. Model A can then call it using the `source()` method, circumventing the circular dependency. 

However, it does also mean the developer has to make sure the table for Model B is created before Model A is called. Putting the `source()` call in an `is_incremental()` block is one way to solve that. There are many more examples.

The purpose of this fix is to not preclude developers from using the idiom, while also making them responsible for resolving the dependency prior to runtime. If the runtime behavior is not resolved, new environments will error when trying to run with a missing table.

If the source/model fqn collision was unintentional, we will no longer catch the circular dependency at load time. It will instead error at runt ime.

### DIFF
--- a/sqlmesh/dbt/context.py
+++ b/sqlmesh/dbt/context.py
@@ -51,6 +51,7 @@ class DbtContext:
     _project_name: t.Optional[str] = None
     _variables: t.Dict[str, t.Any] = field(default_factory=dict)
     _models: t.Dict[str, ModelConfig] = field(default_factory=dict)
+    _models_by_fqn: t.Dict[str, ModelConfig] = field(default_factory=dict)
     _seeds: t.Dict[str, SeedConfig] = field(default_factory=dict)
     _sources: t.Dict[str, SourceConfig] = field(default_factory=dict)
     _refs: t.Dict[str, t.Union[ModelConfig, SeedConfig]] = field(default_factory=dict)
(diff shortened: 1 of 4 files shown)


sqlmesh/dbt/context.py
```diff
@@ -51,6 +51,7 @@ class DbtContext:
     _project_name: t.Optional[str] = None
     _variables: t.Dict[str, t.Any] = field(default_factory=dict)
     _models: t.Dict[str, ModelConfig] = field(default_factory=dict)
+    _models_by_fqn: t.Dict[str, ModelConfig] = field(default_factory=dict)
```
izeigerman: Doesn't look like we need this at all

sqlmesh/dbt/context.py
```diff
@@ -144,13 +145,20 @@ def models(self) -> t.Dict[str, ModelConfig]:
...
+        self._models_by_fqn = {}
         self.add_models(models)
 
     def add_models(self, models: t.Dict[str, ModelConfig]) -> None:
         self._refs = {}
         self._models.update(models)
         self._jinja_environment = None
 
+    @property
+    def models_by_fqn(self) -> t.Dict[str, ModelConfig]:
```
izeigerman: Looks like this should just be `@cached_property` which returns a set of FQNs. No need for this to be a dict

crericha: Sure, it can be a set instead. I really only need the FQN.

izeigerman (approved):

PR #5495, opened 2025-10-07
feat(lineage): vscode updated lineage

mykalmax:


vscode/react/src/pages/help.ts
```diff
@@ -0,0 +1,37 @@
+import { type NodeType } from './ModelLineageContext'
+
+export function getNodeTypeColorVar(nodeType: NodeType) {
```
benfdking: nit:  hoisting the const avoids the per call allocations

const NODE_TYPE_COLOR_VAR = {
  sql: 'var(--color-lineage-node-type-background-sql)',
  python: 'var(--color-lineage-node-type-background-python)',
  'cte/subquery': 'var(--color-lineage-node-type-background-cte-subquery)',
  source: 'var(--color-lineage-node-type-background-source)',
} as const satisfies Record<NodeType, string>;

export function getNodeTypeColorVar(nodeType: NodeType): string {
  return NODE_TYPE_COLOR_VAR[nodeType]; 
}

vscode/react/src/pages/help.ts
```diff
@@ -0,0 +1,37 @@
...
+    sql: 'var(--color-lineage-node-type-background-sql)',
+    python: 'var(--color-lineage-node-type-background-python)',
+    'cte/subquery': 'var(--color-lineage-node-type-background-cte-subquery)',
+    source: 'var(--color-lineage-node-type-background-source)',
+  }[nodeType]
+}
+
+export function getNodeTypeColor(nodeType: NodeType) {
+  return {
+    sql: 'bg-lineage-node-type-background-sql',
```
benfdking: nit: it might be good for the library to expose these strings so that they are tied together correctly. If you change the library you'll break these downstream without any link.

vscode/react/src/pages/lineage.tsx
benfdking: Can we type this model string.

vscode/react/package.json
```diff
@@ -18,22 +18,24 @@
...
     "elkjs": "^0.8.2",
+    "lodash": "4.17.21",
+    "lucide-react": "0.542.0",
     "orval": "^7.10.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-router": "^7.7.0",
     "reactflow": "^11.11.4",
-    "tailwindcss": "^4.1.11",
+    "tailwindcss": "3.4.17",
```
benfdking, line 35: We should really not downgrade tailwind, that feels like a huge step backwards? Getting this up to date took a lot of time and we're just putting ourselves back in that hole.

vscode/react/package.json
```diff
@@ -18,22 +18,24 @@
...
     "@tanstack/react-router": "^1.129.8",
     "@tanstack/react-router-devtools": "^1.131.26",
     "@tanstack/react-virtual": "^3.13.12",
     "@tanstack/router-plugin": "^1.129.8",
+    "@tobikodata/sqlmesh-common": "file:../../web/common",
     "apache-arrow": "^19.0.1",
     "clsx": "^2.1.1",
+    "cronstrue": "3.3.0",
     "elkjs": "^0.8.2",
+    "lodash": "4.17.21",
```
benfdking: I really think we should avoid a library that hasn't been touched in 5 years.

vscode/react/src/pages/ModelNodeColumn.tsx
```diff
@@ -0,0 +1,122 @@
...
+        >
+
+      if (columnLineageData == null) {
+        setTimeout(() => {
+          setFetchingColumns(prev => new Set(prev.add(id)))
+        })
+
+        const { data } = await getColumnLineage()
+
+        columnLevelLineage = data as ColumnLevelLineageAdjacencyList<
```
benfdking: There are so many `as` around here that there really is little guarantee of typesafety in this codebase. If it's pulled in from the context it should already be correctly typed. The only place where `as` may be appropriate is at the top level where we map from the api into the local type.

vscode/react/src/pages/ModelNodeColumn.tsx
```diff
@@ -0,0 +1,122 @@
...
+  } = useModelLineage()
+
+  const isSelectedColumn = selectedColumns.has(id)
+
+  const {
+    data: columnLineageData,
+    refetch: getColumnLineage,
+    isFetching: isColumnLineageFetching,
+    error: columnLineageError,
+  } = useApiColumnLineage(nodeId, name, { models_only: true })
```
benfdking, line 55: These api calls should not leak into the different parts and should be passed down with the correct abstractions.

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+  cn,
+  HorizontalContainer,
+  ModelName,
+  Tooltip,
+  VerticalContainer,
+} from '@tobikodata/sqlmesh-common'
+import { ModelNodeColumn } from './ModelNodeColumn'
+import type { ModelName as ModelNameType } from '@/domain/models'
+
+export const ModelNode = React.memo(function ModelNode({
```
benfdking, line 46: Have we tried the react compiler? I think it should do a much better job than us implementing memos by hand?

benfdking (changes requested):
Quite nervous about these changes because they undo a lot of work that went into making the lineage more up to date and safer with better type safety.

vscode/react/src/pages/help.ts
mykalmax: those colors (for ModelType) are specific to this particular (vscode ext) version of a lineage

vscode/bus/src/brand.ts
```diff
@@ -17,3 +17,26 @@ type Brand<B> = { [__brand]: B }
...
+ *
+ * type FQNMap = BrandedRecord<ModelFQN, string>
+ * type NameMap = BrandedRecord<ModelName, string>
+ *
+ * const fqnMap: FQNMap = {}
+ * const nameMap: NameMap = fqnMap // TypeScript error!
+ */
+export type BrandedRecord<K extends BrandedString, V> = Record<K, V> & {
+  readonly __recordKeyBrand: K
+}
```
benfdking: nit: For me, I am less convinced about these types. They are an additional thing to learn that actually doesn't stop you from writing any less code. Every time you write one of these, it's another thing for someone new to the codebase to jump to and understand. 

BrandedString<"ThingA">
versus 
Branded<String, "ThingA">

 BrandedRecord<"ThingA", int>
versus 
Record<Branded<"ThingA", string>, int>

vscode/react/src/pages/lineage.tsx
```diff
@@ -86,16 +95,16 @@ function Lineage() {
   } = useApiModels()
   const rpc = useRpc()
   React.useEffect(() => {
-    const fetchFirstTimeModelIfNotSet = async (
+    const fetchFirstTimeModelIfNotSet = async <T extends BrandedString>(
```
benfdking: Why the indirection of the generic? In this case, we always want this to return just a model fqn.

vscode/react/src/pages/help.ts
```diff
@@ -0,0 +1,36 @@
+import type { ModelType } from '@/api/client'
+
+type NodeType = ModelType | 'cte/subquery'
```
benfdking: If you use `cte/subquery`, then you need to invent a new concept for ModelFQN. A subquery's id is not a model fqn. You need to define a new concept for it.

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+    selectedNodeId,
+    selectedNodes,
+    showColumns,
+    fetchingColumns,
+  } = useModelLineage()
+
+  const [showNodeColumns, setShowNodeColumns] = React.useState(showColumns)
+  const [isHovered, setIsHovered] = React.useState(false)
+
+  const nodeId = id as ModelNodeId
```
benfdking, line 64: nit: To me, this is the first thing you should and we should make a note why we do it.

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+  ...props
+}: NodeProps<NodeData>) {
+  const {
+    selectedColumns,
+    zoom,
+    currentNode,
+    selectedNodeId,
+    selectedNodes,
+    showColumns,
+    fetchingColumns,
```
benfdking, line 58: I know that we built this dynamically in the past, but I took this away. There are two big problems with this:
1. The context changes over time; to make sure we don't ever show bad graphs, I reset the graph on every context reload. 
2. The performance of this dynamic fetching makes the lineage unusable if you are using a remote machine, which is true for anyone using Codespaces

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+export const ModelNode = React.memo(function ModelNode({
+  id,
+  data,
+  ...props
+}: NodeProps<NodeData>) {
+  const {
+    selectedColumns,
+    zoom,
+    currentNode,
```
benfdking: What is currentNode?

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const nodeId = id as ModelNodeId
+
+  const {
+    leftId,
+    rightId,
+    isCurrent,
+    isSelected, // if selected from inside the lineage and node is selcted
+    isActive, // if selected from inside the lineage and node is not selected but in path
+  } = useNodeMetadata(nodeId, currentNode, selectedNodeId, selectedNodes)
```
benfdking: Do we not store the info so that we could just pass the `useNodeMetadata` without having to pass the rest?

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const {
+    columns,
+    selectedColumns: modelSelectedColumns,
+    columnNames,
+  } = useColumns<ModelFQN, ModelColumnName, Column, ModelColumnID>(
+    selectedColumns,
+    data.name,
+    data.columns,
+  )
```
benfdking, line 82: Why does this need to be used and again cannot be just gotten from state?

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const hasSelectedColumns = selectedColumns.intersection(columnNames).size > 0
+  const hasFetchingColumns = fetchingColumns.intersection(columnNames).size > 0
```
benfdking, line 85: Similar to my comments above, I don't think we should have dynamic data graphs. 

The graph should either be loading fully or not.

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+          ? 'opacity-100'
+          : 'opacity-10',
+      )}
+      style={{
+        height: `${nodeHeight}px`,
+      }}
+    >
+      <NodeAppendix
+        position="top"
+        className="bg-lineage-node-appendix-background"
```
benfdking, line 141: I have really no problem with these but if they are coming from the common it might be worth just making them constants.

vscode/react/src/pages/ModelNode.tsx
```diff
@@ -0,0 +1,293 @@
...
+          'ring-offset-2 z-10',
+          isSelected
+            ? 'ring-2 ring-lineage-node-selected-border ring-offset-lineage-node-background'
+            : 'hover:ring-2 hover:ring-lineage-node-border-hover',
+        )}
+      >
+        <VerticalContainer
+          className="overflow-visible"
+          onMouseEnter={() => setIsHovered(true)}
+          onMouseLeave={() => setIsHovered(false)}
```
benfdking, line 198: I am always going to fight to not have visual stuff be JS dependent because it's slow 😛 but I don't have a good view of whether it is possible. Might be a question for Claude 😛

vscode/react/src/pages/ModelNodeColumn.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <ModelColumn
+      id={id}
+      nodeId={nodeId}
+      modelName={modelName}
+      name={name}
+      type={type}
+      description={description}
+      className={cn(
+        'ModelNodeColumn',
```
benfdking: don't we use kebab casing for css `ModelNodeColumn`?

vscode/react/src/pages/ModelNode.tsx
mykalmax: we passing useLineage hook dynamically , so `useNodeMetadata` needs to accept all args directly

vscode/react/src/pages/ModelNode.tsx
mykalmax: wdym, gotten from state ?

it is calculating for each node

vscode/react/src/pages/ModelNode.tsx
mykalmax: it is fetching data for CLL, we cant make send it upfront, yet

vscode/react/src/pages/ModelNodeColumn.tsx
mykalmax: we do, but for component class i prefer component name as css class to find it faster if needed

PR #5499, opened 2025-10-07
Feat: Check freshness for mixed external & sqlmesh models

VaggelisD:
Follow up PR on model freshness ([Previous PR](https://github.com/TobikoData/sqlmesh/pull/5277))

This PR enables mixed SQLMesh & external (source) model freshness, e.g:

```
A: External model 
B: SELECT * FROM A
C: SELECT * FROM A, B
``` 

Model `C` will be evaluated if its upstream models are fresh, which is determined by:
- `C`'s `last_altered_ts` compared to external model's (`A`) `INFORMATION_SCHEMA` metadata
- The existence (or not) of parent intervals for SQLMesh models (`B`), signalling if any model will be evaluated as part of this run

A new freshness-specific test file is added to break up the original test in various scenarios.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -313,10 +313,10 @@ workflows:
                 - athena
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
+          #   branches:
+          #     only:
+          #       - main
       - ui_style
       - ui_test
       - vscode_test
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -373,11 +373,19 @@ class Scheduler:
                 is_restatement=is_restatement,
             )
 
+            parent_intervals = []
+            for parent in snapshot.parents:
+                if parent.snapshot_id not in snapshot_intervals:
+                    continue
+                _, p_intervals = snapshot_intervals[parent.snapshot_id]
+                parent_intervals.append(p_intervals)
+
             intervals = self._check_ready_intervals(
                 snapshot,
                 intervals,
                 context,
                 environment_naming_info,
+                parent_intervals=parent_intervals,
             )
             unready -= set(intervals)
 
@@ -923,6 +931,7 @@ class Scheduler:
         intervals: Intervals,
         context: ExecutionContext,
         environment_naming_info: EnvironmentNamingInfo,
+        parent_intervals: t.Optional[t.List[Intervals]] = None,
     ) -> Intervals:
         """Checks if the intervals are ready for evaluation for the given snapshot.
 
@@ -965,6 +974,7 @@ class Scheduler:
                     dialect=snapshot.model.dialect,
                     path=snapshot.model._path,
                     snapshot=snapshot,
+                    parent_intervals=parent_intervals,
                     kwargs=kwargs,
                 )
             except SQLMeshError as e:
--- a/sqlmesh/core/signal.py
+++ b/sqlmesh/core/signal.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 import typing as t
 from sqlmesh.utils import UniqueKeyDict, registry_decorator
+from sqlmesh.utils.errors import MissingSourceError
 
 if t.TYPE_CHECKING:
     from sqlmesh.core.context import ExecutionContext
     from sqlmesh.core.snapshot.definition import Snapshot
     from sqlmesh.utils.date import DatetimeRanges
     from sqlmesh.core.snapshot.definition import DeployabilityIndex
+    from sqlmesh.core.snapshot.definition import Intervals
 
 
 class signal(registry_decorator):
@@ -42,7 +44,17 @@ SignalRegistry = UniqueKeyDict[str, signal]
 
 
 @signal()
-def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionContext) -> bool:
+def freshness(
+    batch: DatetimeRanges,
+    snapshot: Snapshot,
+    context: ExecutionContext,
+    parent_intervals: t.Optional[t.List[Intervals]] = None,
+) -> bool:
+    """
+    Implements model freshness as a signal, i.e it considers this model to be fresh if:
+    - Any upstream SQLMesh model has available intervals to compute i.e is fresh
+    - Any upstream external model has been altered since the last time the model was evaluated
+    """
     adapter = context.engine_adapter
     if context.is_restatement or not adapter.SUPPORTS_METADATA_TABLE_LAST_MODIFIED_TS:
         return True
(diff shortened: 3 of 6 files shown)


.circleci/continue_config.yml
```diff
@@ -313,10 +313,10 @@ workflows:
...
                 - fabric
                 - gcp-postgres
-          filters:
-            branches:
-              only:
-                - main
+          # filters:
+          #   branches:
+          #     only:
+          #       - main
```
VaggelisD: TODO: Reinstate before merging

sqlmesh/core/signal.py
```diff
@@ -42,7 +44,17 @@ class signal(registry_decorator):
 
 
 @signal()
-def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionContext) -> bool:
+def freshness(
+    batch: DatetimeRanges,
+    snapshot: Snapshot,
+    context: ExecutionContext,
+    parent_intervals: t.Optional[t.List[Intervals]] = None,
```
izeigerman: Can we somehow make this a part of `ExecutionContext` rather than adding another argument?

sqlmesh/core/signal.py
```diff
@@ -54,24 +66,38 @@ def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionConte
...
     )
+
     if not last_altered_ts:
         return True
 
     parent_snapshots = {context.snapshots[p.name] for p in snapshot.parents}
-    if len(parent_snapshots) != len(snapshot.node.depends_on) or not all(
-        p.is_external for p in parent_snapshots
-    ):
+    if len(parent_snapshots) != len(snapshot.node.depends_on):
```
izeigerman: I'm still confused. If the 2 sets don't match that means the model references external tables. Why don't we treat them just like external models?

sqlmesh/core/signal.py
VaggelisD: Agreed, was hesitant to keep loading `ExecutionContext` but it's a much cleaner approach.

sqlmesh/core/signal.py
VaggelisD: Removed this condition in the latest commit. To account for any edge cases e.g an external model not being present in the `INFORMATION_SCHEMA`, I added a raise condition (and a test for it) instead.

sqlmesh/core/signal.py
```diff
@@ -54,24 +64,35 @@ def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionConte
...
     parent_snapshots = {context.snapshots[p.name] for p in snapshot.parents}
-    if len(parent_snapshots) != len(snapshot.node.depends_on) or not all(
-        p.is_external for p in parent_snapshots
-    ):
-        # The mismatch can happen if e.g an external model is not registered in the project
+
+    upstream_parent_snapshots = {p for p in parent_snapshots if not p.is_external}
+    external_parents = snapshot.node.depends_on - {p.name for p in upstream_parent_snapshots}
+
+    if context.parent_intervals:
```
izeigerman, line 76: We should exclude external models from this check. There could be intervals for external models to, for example, run audits, but it doesn't mean that the table is actually fresh.

sqlmesh/core/scheduler.py
```diff
@@ -364,13 +364,21 @@ def batch_intervals(
 
             adapter = self.snapshot_evaluator.get_adapter(snapshot.model_gateway)
 
+            parent_intervals = []
+            for parent in snapshot.parents:
+                if parent.snapshot_id not in snapshot_intervals:
+                    continue
+                _, p_intervals = snapshot_intervals[parent.snapshot_id]
+                parent_intervals.append(p_intervals)
```
izeigerman: I don't think this is the right place to construct this. The missing intervals will change as we apply signal checks to snapshots. If you set them once here, the parent signals will not be reflected.

sqlmesh/core/signal.py
VaggelisD: Right. I guess I can exclude them at the time of `parent_intervals` construction, since that's a list so we've lost the snapshot mapping of each `Interval`.

sqlmesh/core/scheduler.py
VaggelisD: Good call.

 I guess our best choice would be to check `snapshot_batches` then (same loop) which applies the signal to filter out the unready intervals. 

So, for each snapshot in the DAG, its `ExecutionContext::parent_intervals` list will be extended by the `snapshot_batches[parent]`, provided the `parent` snapshot is NOT external.

izeigerman: yes, this sounds correct, assuming I understood the idea correctly

VaggelisD: Implemented in https://github.com/TobikoData/sqlmesh/pull/5499/commits/9a8af2e6fe9bacf1c97178681a785804b77230a8

sqlmesh/core/signal.py
VaggelisD: Implemented in https://github.com/TobikoData/sqlmesh/pull/5499/commits/9a8af2e6fe9bacf1c97178681a785804b77230a8, external snapshots are excluded from `parent_intervals`.

sqlmesh/core/scheduler.py
```diff
@@ -364,13 +364,22 @@ def batch_intervals(
 
             adapter = self.snapshot_evaluator.get_adapter(snapshot.model_gateway)
 
+            parent_intervals: Intervals = []
+            for parent_id in snapshot.parents:
+                parent_snapshot, _ = snapshot_intervals.get(parent_id, (None, []))
```
izeigerman, line 369: Are you sure that `snapshot_intervals` reflects updated intervals after signals were applied to parents. If I'm reading the code correctly, `snapshot_intervals` remain unchanged and only `snapshot_batches` reflect signal outcomes.

sqlmesh/core/signal.py
```diff
@@ -54,24 +64,35 @@ def freshness(batch: DatetimeRanges, snapshot: Snapshot, context: ExecutionConte
...
         return True
 
     parent_snapshots = {context.snapshots[p.name] for p in snapshot.parents}
-    if len(parent_snapshots) != len(snapshot.node.depends_on) or not all(
-        p.is_external for p in parent_snapshots
-    ):
-        # The mismatch can happen if e.g an external model is not registered in the project
+
+    upstream_parent_snapshots = {p for p in parent_snapshots if not p.is_external}
+    external_parents = snapshot.node.depends_on - {p.name for p in upstream_parent_snapshots}
```
izeigerman, line 74: Sorry man, I'm still confused. `external_parents` contains external tables that were **not** registered as external models. Both `upstream_parent_snapshots` **and** `context.parent_intervals` exclude registered external models. So at which point do we check the last modified timestamp for tables that are registered external models?

izeigerman: I believe `external_models` should include both:
```
snapshot.node.depends_on - {p.name for p in upstream_parent_snapshots}
```
AND
```
{p.name for p in parent_snapshots if p.is_external}
```
am I missing something?

sqlmesh/core/scheduler.py
VaggelisD: We only use `snapshot_interval` here to retrieve the parent snapshots. 

Since the latest commit, the new `parent_intervals` are only based on `snapshot_batches`, see L373. I'll add a test that verifies this given a parent model that has a custom signal suddenly evaluating to `False`.

sqlmesh/core/signal.py
VaggelisD: My thinking was the following: For every model, its parents (i.e the entirety of `model.depends_on` set) can be:
1. SQLMesh models
2. Registered external models (we have a `Snapshot` for them)
3. Unregistered external models

So, given all of the parent **snapshots** i.e (1) + (2):
- `upstream_parent_snapshots` computes (1)
- `external_parents` gathers both (2) and (3) by doing a set difference on `model.depends_on - (1)`

I guess the gotcha here is that we don't really use `upstream_parent_snapshots` elsewhere, but I wanted to break the computation of `external_parents` in steps.

So, the actual freshness computation comes from:
- The `context.parent_intervals`, which indicates whether **any** model from (1) has **any** ready intervals
- The model's `last_altered_ts` comparison with the `external_models` metadata i.e (2) + (3)


Does this make more sense? I hope I'm not butchering any set operation here

sqlmesh/core/scheduler.py
VaggelisD: I added a new test here that should exercise this scenario:

https://github.com/TobikoData/sqlmesh/commit/d4e3fbe91d4c58e06bca5e3ec869708f78184a87#diff-4dff5ecd31963149685166599deae8171b2cda7a1e8bcc99476a173b04925c33

sqlmesh/core/signal.py
VaggelisD: Added a new test to validate this behavior: https://github.com/TobikoData/sqlmesh/pull/5499/commits/de4630620a51c7ba115f102d34bc54e486a01925

Issue #5399, opened 2025-09-18
Fabric Warehouse - Alter table not available


fresioAS:
Alter table does not exist in Fabric Warehouse (!)

```2025-09-18 10:14:56,752 - MainThread - sqlmesh.core.snapshot.evaluator - INFO - Dry running model 'timeplan.planset_all' (evaluator.py:1804)
2025-09-18 10:14:56,756 - MainThread - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: /* SQLMESH_PLAN: c4be7319bc834acf8c5a20d0e4019d21 */ WITH [cte_planset_union] AS (<REDACTED>); (base.py:2498)
2025-09-18 10:14:56,875 - MainThread - sqlmesh.core.snapshot.evaluator - INFO - Migrating table schema from 'my_warehouse.sqlmesh__timeplan.timeplan__planset_all__267513998_schema_tmp' to 'my_warehouse.sqlmesh__timeplan.timeplan__planset_all__267513998' (evaluator.py:1141)
2025-09-18 10:14:56,877 - MainThread - sqlmesh.core.snapshot.evaluator - INFO - Altering table 'my_warehouse.sqlmesh__timeplan.timeplan__planset_all__267513998' (evaluator.py:1831)
2025-09-18 10:14:56,877 - MainThread - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: /* SQLMESH_PLAN: c4be7319bc834acf8c5a20d0e4019d21 */ SELECT DB_NAME(); (base.py:2498)
2025-09-18 10:14:56,910 - MainThread - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: /* SQLMESH_PLAN: c4be7319bc834acf8c5a20d0e4019d21 */ SELECT DB_NAME(); (base.py:2498)
2025-09-18 10:14:56,948 - MainThread - sqlmesh.core.engine_adapter.base - INFO - Executing SQL: /* SQLMESH_PLAN: c4be7319bc834acf8c5a20d0e4019d21 */ SELECT [COLUMN_NAME], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [NUMERIC_PRECISION], [NUMERIC_SCALE] FROM [INFORMATION_SCHEMA].[COLUMNS] WHERE [TABLE_NAME] = 'timeplan__planset_all__267513998' AND [TABLE_SCHEMA] = 'sqlmesh__timeplan'; (base.py:2498)
[...]

PR #5511, opened 2025-10-08
fix(fabric): Alter table workaround

fresioAS:
Fix for issue described in #5399 - ALTER TABLE ALTER COLUMN is not supported for Fabric Warehouse

1. Add new empty column
2. Copy data from old column to new column
3. Drop old column
4. Rename new column to old column name


mattiasthalen:
@georgesittas, can we get traction on this? It's causing some issues ☺️

erindru (approved):

PR #5515, opened 2025-10-09
fix: ensure _get_data_objects respects snapshots with different schem…

z3z1ma:
…a or catalog.

```sql
MODEL (name foo.bar);
SELECT 1;

MODEL (name baz.bar);
SELECT 1;
```

this is enough for get data objects in the snapshot evaluator to think a table doesn't exist, which causes cascading issues (as you'd expect) with, for example, skipping `SnapshotEvaluator.migrate` in situations where the table did in fact exist. This causes issues with downstream tables which expect a correct schema when doing CTAS and such. Its a hard to diagnose breakage without a _lot_ of fishing.

### DIFF
--- a/sqlmesh/core/snapshot/evaluator.py
+++ b/sqlmesh/core/snapshot/evaluator.py
@@ -1593,14 +1593,14 @@ class SnapshotEvaluator:
         tables_by_gateway_and_schema: t.Dict[t.Union[str, None], t.Dict[exp.Table, set[str]]] = (
             defaultdict(lambda: defaultdict(set))
         )
-        snapshots_by_table_name: t.Dict[t.Tuple[str, str, str], Snapshot] = {}
+        snapshots_by_table_name: t.Dict[exp.Table, t.Dict[str, Snapshot]] = defaultdict(dict)
         for snapshot in target_snapshots:
             if not snapshot.is_model or snapshot.is_symbolic:
                 continue
             table = table_name_callable(snapshot)
             table_schema = d.schema_(table.db, catalog=table.catalog)
             tables_by_gateway_and_schema[snapshot.model_gateway][table_schema].add(table.name)
-            snapshots_by_table_name[(table.catalog, table.db, table.name)] = snapshot
+            snapshots_by_table_name[table_schema][table.name] = snapshot
 
         def _get_data_objects_in_schema(
             schema: exp.Table,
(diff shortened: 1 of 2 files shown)


sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1593,14 +1593,14 @@ def _get_data_objects(
         tables_by_gateway_and_schema: t.Dict[t.Union[str, None], t.Dict[exp.Table, set[str]]] = (
             defaultdict(lambda: defaultdict(set))
         )
-        snapshots_by_table_name: t.Dict[str, Snapshot] = {}
+        snapshots_by_table_name: t.Dict[t.Tuple[t.Optional[str], str, str], Snapshot] = {}
```
themisvaltinos: nit but could you key by `table` instead so have:
```suggestion
        snapshots_by_table_name: t.Dict[exp.Table], Snapshot] = {}
```
to make it a bit more consistent with the `tables_by_gateway_and_schema` dictionary as well and since `exp.Table` can contain the catalog and db info as well. and then using `obj.to_table` to create the key when creating the set. Though this will be a bit trickier due to quotation

sqlmesh/core/snapshot/evaluator.py
```diff
@@ -1629,7 +1629,10 @@ def _get_data_objects_in_schema(
                 ]
                 existing_objects.extend(objs_for_gateway)
 
-        return {snapshots_by_table_name[obj.name].snapshot_id: obj for obj in existing_objects}
+        return {
+            snapshots_by_table_name[(obj.catalog, obj.schema_name, obj.name)].snapshot_id: obj
```
themisvaltinos:
```suggestion
            snapshots_by_table_name[(obj.catalog or "", obj.schema_name, obj.name)].snapshot_id: obj
```

if you end up sticking with this solution this is needed since catalog can be `None` here. this also should fix most of the failing tests, the rest unfortunately need to have their `get_data_objects` mock adapted

sqlmesh/core/snapshot/evaluator.py
z3z1ma: I tried and equality check failed because one side was quoted=True in the ast and the other wasn't (IIRC). So rather than deal with that, I opted to unpack into a 3-tuple for simplicity

themisvaltinos: yes it's trickier because DataObject makes it always quoted, so I see why you did that

PR #5522, opened 2025-10-09
chore: ignore partition_by config if dialect is snowflake

toriwei:
snowflake does not support partitioning. log a warning and ignore if the config set.

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -1,10 +1,8 @@
 import json
-import logging
 import os
 import pytest
 import string
 import time_machine
-from contextlib import contextmanager
 from os import getcwd, path, remove
 from pathlib import Path
 from shutil import rmtree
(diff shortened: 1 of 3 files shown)


sqlmesh/dbt/model.py
```diff
@@ -569,7 +569,13 @@ def to_sqlmesh(
                 )
             else:
                 partitioned_by = []
-                if isinstance(self.partition_by, list):
+                if context.target.dialect == "snowflake":
+                    logger.warning(
+                        "Ignoring partition_by config for model '%s' targeting %s. The partition_by config is not supported for Snowflake.",
+                        self.name,
+                        context.target.dialect,
+                    )
```
eakmanrq: Would it make sense to move this up to the outer if clause since it seems to be responsible for checking if the partition by should be ignored? So it would check for view and then elif check if snowflake and then else would remain the same.

tests/dbt/test_transformation.py
```diff
@@ -1880,6 +1880,11 @@ def test_partition_by(sushi_test_project: Project):
     context.target = DuckDbConfig(name="target", schema="foo")
     assert model_config.to_sqlmesh(context).partitioned_by == []
 
+    context.target = SnowflakeConfig(
+        name="target", schema="test", database="test", account="foo", user="bar", password="baz"
+    )
+    assert model_config.to_sqlmesh(context).partitioned_by == []
```
georgesittas, line 1934: Nit: does this pollute the terminal with a warning that we should be catching instead?

sqlmesh/dbt/model.py
toriwei: that makes sense to me, will adjust

tests/dbt/test_transformation.py
toriwei: it doesn't show in the terminal, but added extra assert to check we are outputting the correct warning

PR #5528, opened 2025-10-10
Fix: Mask credentials in duckdb attach for postgres, mysql logging

themisvaltinos:
Add masking in `password` credentials when logging attach information in the logs, similar to the mother duck token.

### DIFF
--- a/sqlmesh/core/config/connection.py
+++ b/sqlmesh/core/config/connection.py
@@ -509,12 +509,12 @@ class BaseDuckDBConnectionConfig(ConnectionConfig):
         return None
 
     def _mask_sensitive_data(self, string: str) -> str:
-        # Mask MotherDuck tokens
+        # Mask MotherDuck tokens with fixed number of asterisks
         result = MOTHERDUCK_TOKEN_REGEX.sub(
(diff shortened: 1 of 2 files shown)


sqlmesh/core/config/connection.py
```diff
@@ -507,10 +508,14 @@ def get_catalog(self) -> t.Optional[str]:
...
-    def _mask_motherduck_token(self, string: str) -> str:
-        return MOTHERDUCK_TOKEN_REGEX.sub(
+    def _mask_sensitive_data(self, string: str) -> str:
+        # Mask MotherDuck tokens
+        result = MOTHERDUCK_TOKEN_REGEX.sub(
             lambda m: f"{m.group(1)}{m.group(2)}{'*' * len(m.group(3))}", string
         )
+        # Mask PostgreSQL and MySQL passwords
+        result = PASSWORD_REGEX.sub(lambda m: f"{m.group(1)}{'*' * len(m.group(2))}", result)
+        return result
```
georgesittas, line 518: Can we write a fixed number of asterisks? There's no reason to emit the same number as the field's length. In fact, we shouldn't do this, as it provides non-trivial information about the field.

PR #5529, opened 2025-10-10
Fix!: Allow expressions in scd type 2 models columns, check_cols in dbt

themisvaltinos:
This adds support for expressions in the list of columns in SCD type 2 models in order to support dbt snapshots that have `check_cols` with casts for example:
```sql
{{
    config(
      target_schema='snapshots',
      unique_key='id',
      strategy='check',
      check_cols=['ds::DATE'],
    )
}}
```
Quite similar to: https://github.com/TobikoData/sqlmesh/pull/2164

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -18,7 +18,7 @@ from functools import cached_property, partial
 
 from sqlglot import Dialect, exp
 from sqlglot.errors import ErrorLevel
-from sqlglot.helper import ensure_list
+from sqlglot.helper import ensure_list, seq_get
 from sqlglot.optimizer.qualify_columns import quote_identifiers
 
 from sqlmesh.core.dialect import (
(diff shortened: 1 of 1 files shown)


sqlmesh/core/engine_adapter/base.py
```diff
@@ -1885,8 +1885,14 @@ def remove_managed_columns(
         # they are equal or not, the extra check is not a problem and we gain simplified logic here.
         # If we want to change this, then we just need to check the expressions in unique_key and pull out the
         # column names and then remove them from the unmanaged_columns
-        if check_columns and check_columns == exp.Star():
-            check_columns = [exp.column(col) for col in unmanaged_columns_to_types]
+        if check_columns:
+            # Handle both Star directly and [Star()] (which can happen during serialization/deserialization)
+            if isinstance(check_columns, exp.Star) or (
```
erindru: I think this can be represented more succinctly with the following helpers:

```
from sqlglot.helper import ensure_list, seq_get

if isinstance(seq_get(ensure_list(check_columns), 0), exp.Star):
 ...
```

but it's six in one, half a dozen the other

themisvaltinos: nice that's much cleaner with seq_get thanks

PR #5534, opened 2025-10-13
Fix(sqlmesh_dbt): Allow global options to be specified in both top-level and subcommand positions

erindru:
the dbt cli allows a certain set of "global" options to be defined both at top level (before a subcommand) and at the subcommand level.

This means that the following are equivalent and valid:
```
$ dbt --profile jaffle_shop list
$ dbt list --profile jaffle_shop
```

This PR refactors how options are captured by the `sqlmesh_dbt` CLI to ensure that `sqlmesh_dbt` supports this too

### DIFF
--- a/sqlmesh/__init__.py
+++ b/sqlmesh/__init__.py
@@ -232,15 +232,20 @@ def configure_logging(
         logger.addHandler(file_handler)
 
     if debug:
-        import faulthandler
+        import faulthandler, io
 
         enable_debug_mode()
 
         # Enable threadumps.
-        faulthandler.enable()
+        try:
+            faulthandler.enable()
 
-        # Windows doesn't support register so we check for it here
-        if hasattr(faulthandler, "register"):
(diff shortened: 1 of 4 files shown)


sqlmesh/__init__.py
```diff
@@ -232,15 +232,20 @@ def configure_logging(
...
 
     if debug:
-        import faulthandler
+        import faulthandler, io
 
         enable_debug_mode()
 
         # Enable threadumps.
-        faulthandler.enable()
+        try:
```
erindru, line 240: One of the tests for `--debug` was failing because `faulthandler.enable()` raises an Exception if `sys.stderr.fileno()` fails.

`sys.stderr.fileno()` can fail if something, for example the Click TestRunner (or even pytest) has messed with `sys.stderr` and it's no longer a file.

So rather than bomb out, I changed this to just log a warning

PR #5544, opened 2025-10-15
Fix: Respect mssql_merge_exists flag and use merge as a strategy

themisvaltinos:
A user reported that scd type 2 in python models don't respect `mssql_merge_exists` flag, but this is a more general issue. This is because this pr: https://github.com/TobikoData/sqlmesh/pull/4531 added `DELETE+INSERT` strategy to be used instead of `MERGE` in `_insert_overwrite_by_condition` for mssql. This update respects the flag if the user has opted in to use merge strategy for a particular model and uses `merge` instead.

slack thread: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1758704337222309

### DIFF
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -438,7 +438,7 @@ class MSSQLEngineAdapter(
                 **kwargs,
             )
 
-        # For conditional overwrites or when mssql_merge_exists is set, use MERGE from InsertOverwriteWithMergeMixin
+        # For conditional overwrites or when mssql_merge_exists is set use MERGE
         return super()._insert_overwrite_by_condition(
             table_name=table_name,
             source_queries=source_queries,


sqlmesh/core/engine_adapter/mssql.py
```diff
@@ -436,7 +438,7 @@ def _insert_overwrite_by_condition(
                 **kwargs,
             )
 
-        # For actual conditional overwrites, use MERGE from InsertOverwriteWithMergeMixin
+        # For conditional overwrites or when mssql_merge_exists is set, use MERGE from InsertOverwriteWithMergeMixin
```
erindru: Actually, `InsertOverwriteWithMergeMixin` was removed entirely in #5380

erindru (approved):

sqlmesh/core/engine_adapter/mssql.py
themisvaltinos: yes good point I'll adapt the comment now that it's consolidated

PR #5551, opened 2025-10-17
Fix: Unexpected backfill of a parent when an interval outside the parent's range is restated for a child

izeigerman:
The issue might occur when the parent's start is inferred to be later (greater) than the child's and the restated interval is outside the parent's range. In this case a parent is backfilled unexpectedly.

### DIFF
--- a/sqlmesh/core/snapshot/definition.py
+++ b/sqlmesh/core/snapshot/definition.py
@@ -2081,16 +2081,20 @@ def missing_intervals(
                 continue
             snapshot_end_date = existing_interval_end
 
+        snapshot_start_date = max(
+            to_datetime(snapshot_start_date),
+            to_datetime(start_date(snapshot, snapshots, cache, relative_to=snapshot_end_date)),
+        )
+        if snapshot_start_date > to_datetime(snapshot_end_date):
+            continue
+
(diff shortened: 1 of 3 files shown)


sqlmesh/core/snapshot/definition.py
```diff
@@ -2295,14 +2299,16 @@ def start_date(
...
-            earliest = min(
-                earliest,
-                start_date(snapshots[parent], snapshots, cache=cache, relative_to=relative_to),
-            )
+    parent_starts = [
+        start_date(snapshots[parent], snapshots, cache=cache, relative_to=relative_to)
+        for parent in snapshot.parents
+        if parent in snapshots
+    ]
+    earliest = (
```
themisvaltinos, line 2307: in the case where a snapshot has parents it never considers its own start, shouldn't it still consider it here similar to before so have something like:
```
  cron_start = snapshot.node.cron_prev(snapshot.node.cron_floor(relative_to or now()))
  earliest = min([cron_start] + parent_starts) if parent_starts else cron_start
```

themisvaltinos: actually I see now this is needed to constrain the data of the child to avoid the backfill of parents

PR #5559, opened 2025-10-22
Fix: create_test error with query and CTEs included

### LINKED ISSUE #5550: TypeError when using --include-ctes with DATE, DATETIME, or TIMESTAMP columns in sqlmesh create_test
Hi team! 

When creating unit tests, if a column has a data type of DATE, DATETIME, or TIMESTAMP, the command
 `sqlmesh create_test {model} --include-ctes --query {}` throws the following error:

`TypeError: list indices must be integers or slices, not str`

Without the `--include-ctes `option, it works perfectly (I believe there are no structural issues or incorrect column types). However, when using `--include-ctes`, it only works if those columns are cast to STRING.  (I’m using the BigQuery engine with the default DuckDB test engine.)

Here is a simple code example based on the SQLMesh documentation.

model_unit_test.sql
```
MODEL (
  name sqlmesh_example.filtered_cte_view,
  kind VIEW,
  audits (
    NOT_NULL(columns = [id])
  ),
  ignored_rules ['ambiguousorinvalidcolumn']
);

WITH filtered_seed_cte AS (
  SELECT
    id,
    item_id,
    event_date
  WHERE
    item_id = 1
)
SELECT
  item_id,
  COUNT(DISTINCT id) AS num_orders
FROM filtered_seed_cte
GROUP BY
  item_id
```
seed_model.sql
```
MODEL (
  kind SEED (
    path '../seeds/seed_data.csv'
  ),
  audits (
    unique_values(columns := (
      id
    ))
  )
)

```
Command for creating the unit test:
```
sqlmesh create_test "sqlmesh_example.filtered_cte_view" \
  --include-ctes \
  --query "sqlmesh_example.seed_model" \
  "SELECT
      4 AS id,
      1 AS item_id
   UNION ALL
   SELECT
      5,
      1 
   UNION ALL
   SELECT
      7,
      2 
   UNION ALL
   SELECT
      8,
      2 " \
  --overwrite
```

Could someone from the SQLMesh team take a look at this issue?

Context: https://tobiko-data.slack.com/archives/C044BRE5W4S/p1760719851382139

treysp:
closes #5550

### DIFF
--- a/sqlmesh/core/test/definition.py
+++ b/sqlmesh/core/test/definition.py
@@ -925,8 +925,7 @@ def generate_test(
                 cte_output = test._execute(cte_query)
                 ctes[cte.alias] = (
                     pandas_timestamp_to_pydatetime(
-                        cte_output.apply(lambda col: col.map(_normalize_df_value)),
-                        cte_query.named_selects,
+                        df=cte_output.apply(lambda col: col.map(_normalize_df_value)),
                     )
                     .replace({np.nan: None})
                     .to_dict(orient="records")
--- a/sqlmesh/utils/date.py
+++ b/sqlmesh/utils/date.py
@@ -444,7 +444,7 @@ def to_time_column(
 
 
 def pandas_timestamp_to_pydatetime(
-    df: pd.DataFrame, columns_to_types: t.Optional[t.Dict[str, exp.DataType]]
+    df: pd.DataFrame, columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None
 ) -> pd.DataFrame:
     import pandas as pd
     from pandas.api.types import is_datetime64_any_dtype  # type: ignore


sqlmesh/core/test/definition.py
```diff
@@ -925,8 +925,7 @@ def generate_test(
                 cte_output = test._execute(cte_query)
                 ctes[cte.alias] = (
                     pandas_timestamp_to_pydatetime(
-                        cte_output.apply(lambda col: col.map(_normalize_df_value)),
-                        cte_query.named_selects,
+                        df=cte_output.apply(lambda col: col.map(_normalize_df_value)),
```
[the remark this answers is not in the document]
treysp: `cte_query.named_selects` is a list containing column names - not a columns_to_types dict as expected by pandas_timestamp_to_pydatetime

georgesittas, line 928: Seems like I [introduced](https://github.com/tobikodata/sqlmesh/commit/3cd3985debade1d4481dbdf688dd9450da7fa99c) the erroneous logic some time ago. Thanks for fixing it.

Curious whether we should do a best-effort inference of the CTE's columns-to-types mapping. How bad is it if we don't hit this branch, within `pandas_timestamp_to_pydatetime`?

```python
            if columns_to_types and columns_to_types[column].this in (
                exp.DataType.Type.DATE,
                exp.DataType.Type.DATE32,
            ):
                # Sometimes `to_pydatetime()` has already converted to date, so we only extract from datetime objects.
                df[column] = df[column].map(
                    lambda x: x.date() if type(x) is datetime and not pd.isna(x) else x
                )
```

georgesittas (approved):

sqlmesh/core/test/definition.py
treysp: yeah, i was going to ask you if it was worth it to try type inference for this reason - how complex to implement?

if we don't hit that branch, we'll write the time component to the test file even for the `DATE` types. 

However, we pass the inline YAML values to `select_from_values_for_batch_range`, which then CASTs to the correct known type. Since casting datetime/timestamp to date is supported by every engine, I think the impact here is limited to it just "looking wrong" in the YAML file.

georgesittas: I see, thanks for clarifying– that's fine then. Let's keep what you have to avoid increasing the complexity of the solution.

Issue #5562, opened 2025-10-24
--execution-time param of CLI's audit command isn't working


makaroni4:
## Problem

When one runs an audit via sqlmesh CLI like so:

```bash
sqlmesh audit \
  --execution-time "2025-10-24 19:18:09" \
  --model XXX.YYY
```

execution macros like `@execution_ds` are set to `1970-01-01`. Here's an audit query from the previous command

```sql
WITH `date_filtered` AS (
  SELECT * 
  FROM `XXX`.`YYY`.`ZZZ` AS `ZZZ` 
  WHERE                   
    SAFE_CAST(`publisher_date` AS DATE) > (DATE('1970-01-01')) -- DATE(@execution_ds) in the audit definition
) 

SELECT * 
FROM `date_filtered` AS `date_filtered`                            
```

PR #5563, opened 2025-10-24
fix: added missing execution_time param for CLI audit command

makaroni4:
This PR fixes https://github.com/TobikoData/sqlmesh/issues/5562.

## Problem

Atm `execition_time` CLI param isn't working, since it's not passed down to the `SnapshotEvaluator`.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2256,8 +2256,8 @@ class GenericContext(BaseContext, t.Generic[C]):
         Args:
             start: The start of the interval to audit.
             end: The end of the interval to audit.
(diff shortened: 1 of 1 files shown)


sqlmesh/core/context.py
```diff
@@ -2247,17 +2247,17 @@ def audit(
         self,
         start: TimeLike,
         end: TimeLike,
+        execution_time: t.Optional[TimeLike] = None,
```
izeigerman: why was this moved?

makaroni4: Good point, let's keep `execution_time` keyword-only 👍

izeigerman (approved):

PR #5578, opened 2025-11-05
Fix: Include column types in Databricks materialized views with comments

afredojala:
## Summary
  - Fixes an issue where Databricks materialized views with column comments were missing column
  type definitions
  - Databricks requires column types to be explicitly specified in CREATE MATERIALIZED VIEW
  statements when column comments are present

  ## Technical Details
  The fix overrides the `is_view` parameter to `False` when building column definitions for
  materialized views that have column comments. This ensures the base class includes column types
   in the schema definition, which Databricks requires for this scenario.

Let me know if we should take do another approach for this.

### DIFF
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -405,8 +405,6 @@ class DatabricksEngineAdapter(SparkEngineAdapter, GrantsFromInfoSchemaMixin):
         # in CREATE MATERIALIZED VIEW statements. Override is_view to False to force
         # column types to be included when comments are present.
         if is_view and column_descriptions:
-            engine_supports_schema_comments = self.COMMENT_CREATION_VIEW.supports_schema_def
-            if engine_supports_schema_comments and self.comments_enabled:
-                is_view = False
+            is_view = False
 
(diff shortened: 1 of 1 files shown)


sqlmesh/core/engine_adapter/databricks.py
```diff
@@ -394,3 +394,19 @@ def _build_table_properties_exp(
...
+        self,
+        target_columns_to_types: t.Dict[str, exp.DataType],
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        is_view: bool = False,
+    ) -> t.List[exp.ColumnDef]:
+        # Databricks requires column types to be specified when adding column comments
+        # in CREATE MATERIALIZED VIEW statements. Override is_view to False to force
+        # column types to be included when comments are present.
+        if is_view and column_descriptions:
+            engine_supports_schema_comments = self.COMMENT_CREATION_VIEW.supports_schema_def
```
izeigerman: Do we even need this check here provided that we know that this will always be true for Databricks?

afredojala: Hmm, yeah you are correct. I've removed the inner condition

PR #5597, opened 2025-11-17
Fix: Avoid unnecessary view recreation for metadata changes during migration

xardasos:
This PR optimizes views migration. In the case of metadata changes, it is not always necessary to recreate views if the engine doesn't support view comments. This optimization is especially relevant for materialized views, which can take more time to be rebuilt.

### DIFF
--- a/sqlmesh/core/plan/evaluator.py
+++ b/sqlmesh/core/plan/evaluator.py
@@ -15,7 +15,6 @@ Refer to `sqlmesh.core.plan`.
 """
 
 import abc
-import itertools
 import logging
 import typing as t
 from sqlmesh.core import analytics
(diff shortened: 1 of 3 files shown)


xardasos:
Hi @izeigerman, @erindru and @eakmanrq,  I'd appreciate your review of this PR.

The changes aim to prevent unnecessary view (especially MV) recreation when deploying metadata snapshots to production on engines that don't support view comments. In previous versions of SQLMesh (before 0.183.0, related to https://github.com/TobikoData/sqlmesh/pull/4414), metadata changes didn't trigger view recreation. This PR aims to restore this behavior while still allowing view recreation in scenarios where it's necessary.

Note: For engines that support comments in post-creation commands (like COMMENT or ALTER), we might be able to avoid recreation and only update the comments. This optimization is not included in this PR. However, it's unlikely to be feasible for engines that only allow comments in CREATE VIEW schema definitions.

I'm open to any suggestions you may have. Thanks for your time!

jukiewiczm:
> Hi @tobymao, I noticed that the PR was closed and I just wanted to ask if there's something I could adjust. I realize now that I should have also mentioned earlier in the PR description that one of my goals is to address a bug that causes data to be unavailable in production for a long time and makes rollback difficult. I've created a related issue (#5653). I'd be happy to update or resubmit if this PR change is useful. Thanks for maintaining this project.

This is actually a pretty crucial bug. It makes sense to close it if it doesn't fit for whatever reason, but doing so without a word of explanation doesn't seem like a good move.

PR #5617, opened 2025-12-03
Fix(databricks): Add column type only along materialized view comments

themisvaltinos:
Fix to only add types in Databricks materialised views and not regular views (introduced here: https://github.com/TobikoData/sqlmesh/pull/5578)

DOCS:
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-materialized-view#parameters
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-view

### DIFF
--- a/sqlmesh/core/engine_adapter/base.py
+++ b/sqlmesh/core/engine_adapter/base.py
@@ -811,7 +811,7 @@ class EngineAdapter:
         column_descriptions: t.Optional[t.Dict[str, str]] = None,
         expressions: t.Optional[t.List[exp.PrimaryKey]] = None,
         is_view: bool = False,
-        is_materialized_view: bool = False,
+        materialized: bool = False,
     ) -> exp.Schema:
         """
         Build a schema expression for a table, columns, column comments, and additional schema properties.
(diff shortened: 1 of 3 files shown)


.circleci/continue_config.yml
```diff
@@ -309,18 +309,18 @@ workflows:
           matrix:
             parameters:
               engine:
-                - snowflake
+          #       - snowflake
```
themisvaltinos: todo: reinstate prior to mergig

sqlmesh/core/engine_adapter/base.py
```diff
@@ -811,6 +811,7 @@ def _build_schema_exp(
         column_descriptions: t.Optional[t.Dict[str, str]] = None,
         expressions: t.Optional[t.List[exp.PrimaryKey]] = None,
         is_view: bool = False,
+        is_materialized_view: bool = False,
```
georgesittas: Should this be `is_materialized`? And then we'd use it in combination with `is_view`.

themisvaltinos: good point will pass the flag as is to avoid confusion

PR #5634, opened 2025-12-18
fix: set `ExternalModel`'s default `kind` to `ExternalKind()`

etonlels:


themisvaltinos (approved):
lgtm @etonlels thank you! you can go ahead and move it out of draft

PR #5651, opened 2026-01-09
Fix: ignore non-key "dialect" in MODEL/AUDIT block

treysp:
Every model/audit definition begins with a header block containing configuration key-value pairs. One optional pair is the SQL dialect, specified as `dialect {dialect}`, `dialect '{dialect}'`, or `dialect "{dialect}"`.

We must detect a model's dialect BEFORE we parse the header because the header may contain syntax the default sqlglot dialect cannot parse. We use a regex to detect the dialect prior to parsing.

The regex currently detects `dialect` anywhere in the header block, so it will attempt to parse it as a key-value even if it's inside a description, etc.

This PR makes the regex stricter, such that:
1. The `dialect` key must come after the initial paren in `MODEL(` or after a comma
2. The dialect value must be followed by a comma or closing paren
3. Whitespace/comments are allowed in between (1) and the dialect key, or between the dialect key and (2)

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -806,7 +806,7 @@ def text_diff(
 WS_OR_COMMENT = r"(?:\s|--[^\n]*\n|/\*.*?\*/)"
 HEADER = r"\b(?:model|audit)\b(?=\s*\()"
 KEY_BOUNDARY = r"(?:\(|,)"  # key is preceded by either '(' or ','
-DIALECT_VALUE = r"(?:'(?P<d_quoted>[a-z_][a-z0-9_]*)?'|(?P<d_unquoted>[a-z_][a-z0-9_]*))"  # value is single-quoted (maybe empty) or unquoted
+DIALECT_VALUE = r"['\"]?(?P<dialect>[a-z][a-z0-9]*)['\"]?"
 VALUE_BOUNDARY = r"(?=,|\))"  # value is followed by comma or closing paren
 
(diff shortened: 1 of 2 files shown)


sqlmesh/core/dialect.py
```diff
@@ -803,8 +803,15 @@ def text_diff(
     return "\n".join(unified_diff(a_sql, b_sql))
 
 
+WS_OR_COMMENT = r"(?:\s|--[^\n]*\n|/\*.*?\*/)"
+HEADER = r"\b(?:model|audit)\b(?=\s*\()"
+KEY_BOUNDARY = r"(?:\(|,)"  # key is preceded by either '(' or ','
+DIALECT_VALUE = r"(?:'(?P<d_quoted>[a-z_][a-z0-9_]*)?'|(?P<d_unquoted>[a-z_][a-z0-9_]*))"  # value is single-quoted (maybe empty) or unquoted
```
georgesittas: Are `dialect foo` and `dialect 'foo'` the only allowed variants? In BigQuery, for example, `"` can also be used as a string quote delimiter. So, `dialect "foo"` would also make sense there in theory.

Is this quoted/non-quoted check necessary if we introduce these "boundary" checks?

treysp: > In BigQuery, for example, " can also be used as a string quote delimiter

I did not know that! Agreed that the quote check isn't necessary, and that will simplify the implementation.

PR #5662, opened 2026-01-14
feat: add cron_tz to model_defaults

### LINKED ISSUE #5661: Support cron_tz in model_defaults for project-wide timezone configuration
### Feature Request
Add support for cron_tz in the model_defaults section of config.yaml to allow setting a default timezone for all models in a project.

### Use Case
In projects where all models use the same timezone for cron scheduling, it's repetitive and error-prone to specify the timezone in every individual model definition. A project-wide default would significantly reduce configuration duplication and maintenance overhead.

### Current Behavior
Currently, cron_tz must be specified in each individual model's MODEL() definition:

```
MODEL (
  name my_schema.my_model,
  cron: '0 * * * *',
  cron_tz 'Europe/London',
);
```
When attempting to set it in model_defaults, SQLMesh returns an error:

> Error: 'cron_tz' is not a valid model default configuration key. 
Please remove it from the `model_defaults` specification in your config file.

### Desired Behavior
Allow `cron_tz` to be specified in model_defaults:

```
model_defaults:
  dialect: trino
  start: '2025-01-01'
  kind: VIEW
  cron: '0 * * * *'
  cron_tz: 'Europe/London'  # Should be valid here
  interval_unit: 'hour'
```

lafirm:
Should fix https://github.com/TobikoData/sqlmesh/issues/5661

### Summary
Adds support for `cron_tz` in `model_defaults`, allowing users to set a default timezone for cron schedules at the project level and gateway level.

### Changes
- Added `cron_tz` field to `ModelDefaultsConfig` with proper validation
- Updated documentation to include `cron_tz` in the list of supported model_defaults options
- Added tests for both project-level and gateway-specific `cron_tz` defaults

### Benefits
- **DRY**: No need to specify `cron_tz` on every model when using a standard timezone
- **Gateway flexibility**: Different environments can use different default timezones
- **Consistency**: Uses the same validation logic as model-level `cron_tz`
- **Backward compatible**: Existing configs without `cron_tz` continue to work (defaults to UTC)

### Example Usage
```yaml
# Project-level default
model_defaults:
  dialect: snowflake
  cron: '@daily'
  cron_tz: 'America/Los_Angeles'

# Gateway-specific override
gateways:
  production:
    connection:
      type: snowflake
    model_defaults:
      cron_tz: 'America/New_York'

### DIFF
--- a/docs/reference/model_configuration.md
+++ b/docs/reference/model_configuration.md
@@ -178,6 +178,7 @@ The SQLMesh project-level `model_defaults` key supports the following options, d
 - kind
 - dialect
 - cron
+- cron_tz
 - owner
 - start
 - table_format
--- a/sqlmesh/core/config/model.py
+++ b/sqlmesh/core/config/model.py
@@ -14,7 +14,7 @@ from sqlmesh.core.model.kind import (
     OnAdditiveChange,
 )
 from sqlmesh.core.model.meta import FunctionCall
-from sqlmesh.core.node import IntervalUnit
+from sqlmesh.core.node import IntervalUnit, cron_tz_validator
 from sqlmesh.utils.date import TimeLike
 from sqlmesh.utils.pydantic import field_validator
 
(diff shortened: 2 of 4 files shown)


StuffbyYuki:
@lafirm can you add DCO checks on your commit?

lafirm:
> @lafirm can you add DCO checks on your commit?

done :)

PR #5670, opened 2026-01-16
fix bug with quoted identifiers in ALTER SESSION SET

### LINKED ISSUE #5666: quoting parameters in ALTER SESSION SET for Snowflake is illegal syntax.
sqlmesh quotes parameters in `ALTER SESSION SET` for **Snowflake** is **illegal syntax**.
What’s worse, it is even used in auto-generated queries. This is what I found in the logs:

```sql
2026-01-15 16:35:07,526 - ThreadPoolExecutor-30_0 - sqlmesh.core.engine_adapter.base - INFO - Executing SQL:
 /* SQLMESH_PLAN: 1f393dd1a70d447e851739fbfeddd4b2 */
 ALTER SESSION SET "TIMEZONE" = 'UTC'
```

`ALTER SESSION SET "TIMEZONE" = 'UTC'` is **invalid SQL syntax** for Snowflake.

korbash:
Fixes #5666
This fix looks pretty bad because it moves Snowflake-specific logic out of `engine_adapter/snowflake.py`. But anyway, it highlighted the problem, and it works. I hope it doesn't have side effects, but I'm not sure.

### DIFF
--- a/sqlmesh/core/engine_adapter/snowflake.py
+++ b/sqlmesh/core/engine_adapter/snowflake.py
@@ -622,6 +622,12 @@ class SnowflakeEngineAdapter(
         return expression
 
     def _to_sql(self, expression: exp.Expression, quote: bool = True, **kwargs: t.Any) -> str:
+        # Snowflake doesn't accept quoted identifiers in ALTER SESSION SET statements
+        # e.g., ALTER SESSION SET "TIMEZONE" = 'UTC' is invalid
+        # We need to disable quoting for these statements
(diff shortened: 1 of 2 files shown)


StuffbyYuki:
@korbash can you add tests?

StuffbyYuki:
@korbash Following up on my prev message

korbash:
@StuffbyYuki Sorry for the late response. I’m not really like my solution because I moved dialect-specific logic out of `engine_adapter/snowflake.py`. I don’t think it should be merged.

PR #5688, opened 2026-02-02
Fix!: Allow init to be walked to track its dependencies

themisvaltinos:
Fixes an issue in python model's serialization where class dependencies referenced inside `__init__` were not added to the serialised env, for example:

```python
class ReferencedClass:
    def __init__(self, value: int):
        self.value = value

    def get_value(self) -> int:
        return self.value

class ClassThatReferencesAnother:
    def __init__(self, x: int):
        self.helper = ReferencedClass(x * 2)
```

This is because in metaprogramming the `build_env()` function was skipping all dunder methods including `__init__` so the dependencies were never added to the serialized environment.

### DIFF
--- a/tests/utils/test_metaprogramming.py
+++ b/tests/utils/test_metaprogramming.py
@@ -83,7 +83,18 @@ class DataClass:
     x: int
 
 
+class ReferencedClass:
+    def __init__(self, value: int):
+        self.value = value
+
+    def get_value(self) -> int:
+        return self.value
+
+
 class MyClass:
+    def __init__(self, x: int):
+        self.helper = ReferencedClass(x * 2)
+
     @staticmethod
     def foo():
         return KLASS_X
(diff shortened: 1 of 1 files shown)


tests/utils/test_metaprogramming.py
georgesittas, line 1: @themisvaltinos can we augment the existing tests we have instead of defining new ones? No need to repeat the boilerplate logic in `test_serialize_env_with_class_referencing_another_class`.

georgesittas:
There may be an edge case where this is breaking, FYI. If `__init__` references a global variable, for example, which was previously treated as metadata-only because it appeared elsewhere, now it will be picked up and should not be treated as metadata-only.

Not sure if we can do much about this, just mentioning so that we can defensively mark the PR accordingly.

tests/utils/test_metaprogramming.py
themisvaltinos: yes removed the `test_serialize_env_with_class_referencing_another_class` test and refactored to be part of the existing serialise test

PR #5720, opened 2026-03-05
Fix: Exclude seed models from default plan end date calculation

themisvaltinos:
Seed models are evaluated when first loaded and their interval end is set at that point. This might become a problem when a pr introduces a new model that depends on an existing seed. in that case the seed is the only model with intervals in prod so its stale interval end becomes the default plan end date. Now if `default_pr_start` is configured (say "6 days ago" or a specific hardcoded value) it can resolve to a timestamp newer than the seeds frozen interval end causing the plan to fail with:

```
PlanError: Plan end date: '2025-03-03 00:00:00' must be after the plan start date: '2026-01-04 00:00:00'
```

This excludes the seed models intervals from the default end calculation to avoid this, the test shows an example of how this can happen.

### DIFF
--- a/tests/core/test_context.py
+++ b/tests/core/test_context.py
@@ -1208,13 +1208,14 @@ def test_plan_seed_model_excluded_from_default_end(copy_to_temp_path: t.Callable
 
         # the plan start date 2025-01-01 is after the seeds end date but shouldnt cause the plan to fail
         plan = context.plan(
-            "dev",
-            start="2025-01-01",
-            no_prompts=True,
-            select_models=["*waiter_summary"]
+            "dev", start="2025-01-01", no_prompts=True, select_models=["*waiter_summary"]
         )
 
         # the end should fall back to execution_time rather than seeds end
+        assert plan.models_to_backfill == {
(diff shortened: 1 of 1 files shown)


tests/core/test_context.py
```diff
@@ -1157,6 +1157,64 @@ def test_plan_start_ahead_of_end(copy_to_temp_path):
...
+        )
+
+        # the seed's interval end would still be 2024-06-01
+        max_ends = context.state_sync.max_interval_end_per_model("prod")
+        seed_fqns = [k for k in max_ends if "waiter_names" in k]
+        assert len(seed_fqns) == 1
+        assert max_ends[seed_fqns[0]] == to_timestamp("2024-06-01")
+
+        # the plan start date 2025-01-01 is after the seeds end date but shouldnt cause the plan to fail
+        context.plan(
```
erindru: This test could be improved by inspecting the `Plan` object returned from `context.plan()` and verifying what dates it used

themisvaltinos: good point added to check the dates from the plan

PR #5730, opened 2026-03-13
Add `--skip-audits` flag to plan and run commands

airhorns:
## Description

Allow users to skip audit execution during plan application and run operations. This is useful for large-scale backfills or development scenarios where audit overhead is not desired, or where you want to run audits out of band.

## Test Plan

 - added new unit tests
 - tophatted locally in my project

## Checklist

- [X] I have run `make style` and fixed any issues
- [X] I have added tests for my changes (if applicable)
- [X] All existing tests pass (`make fast-test`)
- [X] My commits are signed off (`git commit -s`) per the [DCO](DCO)

PR #5731, opened 2026-03-13
Run audits concurrently using concurrent_tasks setting

### LINKED ISSUE #5468: Concurrency in tests and audits
https://tobiko-data.slack.com/archives/C044BRE5W4S/p1758688241595459

Hi, 

Opening an issue as discussed in the slack thread above. My sqlmesh project has py spark models and uses databricks connect to be able to plan using the cli. We arent able to use duck db for local tests as the tests for these python based models cannot execute on duck, and arent able to use a local spark cluster as we use databricks connect. We hence provision a small cluster in databricks for tests to run here and  use sychronized deployments with the CI/CD bot. 

I need help with two issues
1. Each time the bot is run, our tests  take more than 5 mins to succeed as they are executed one after the other. Same case with audits run from local. This doesnt improve if i provision a medium cluster or any other for that matter.
2.I also would like to understand , if its possible to switch to a local spark cluster(as its more closer to the databricks dialect) for testing while we plan against a prod databricks, as it saves us cost. Would this be the right usage of disable-connect flag? I havent tried this option yet

TIA,
Sid

airhorns:
## Description

Before, audits were run one at a time after each model, and block the execution of downstream models. If you are a heavy user of audits, this trades correctness for a lot of speed, where the run will only move forward if all the audits pass. But, if you want to get the run done and then audit after the fact, it really slows things down. `dbt` takes a different approach, where you can do a run, and then later run tests. I wouldn't say that's better, but in some circumstances, especially with non-blocking audits or warnings, you don't need to know their state before moving on and building the next model.

So, in an effort to facilitate running audits out of band as fast as possible, or faster when run in band, this changes audits to execute concurrently as part of the DAG! Paired with #5730, this lets folks run the models without any audits, and later run audits with a high degree of paralleism.

#### Implementation

When audit_only=True, all audit tasks across all snapshots are flattened into a single thread pool instead of following DAG ordering. Since audits are read-only SELECT queries with no side effects, DAG dependencies are irrelevant and all concurrent_tasks worker slots stay filled.

Per-model audit concurrency is also plumbed through SnapshotEvaluator.audit() via a new audit_concurrent_tasks parameter (defaults to sequential). The cross-model path hardcodes this to 1 to avoid nested thread pool multiplication.

The SnapshotEvaluator parameter ddl_concurrent_tasks is renamed to concurrent_tasks to reflect its broader scope.

## Checklist

- [ ] My commits are signed off (`git commit -s`) per the [DCO](DCO)

Closes #5468

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -492,7 +492,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                     gateway: adapter.with_settings(execute_log_level=logging.INFO)
                     for gateway, adapter in self.engine_adapters.items()
                 },
-                ddl_concurrent_tasks=self.concurrent_tasks,
+                concurrent_tasks=self.concurrent_tasks,
                 selected_gateway=self.selected_gateway,
             )
         return self._snapshot_evaluator
--- a/sqlmesh/core/scheduler.py
+++ b/sqlmesh/core/scheduler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 import abc
 import logging
+import threading
 import typing as t
 import time
 from datetime import datetime
@@ -37,7 +38,11 @@ from sqlmesh.core.snapshot.definition import (
 )
 from sqlmesh.core.state_sync import StateSync
 from sqlmesh.utils import CompletionStatus
-from sqlmesh.utils.concurrency import concurrent_apply_to_dag, NodeExecutionFailedError
+from sqlmesh.utils.concurrency import (
+    concurrent_apply_to_dag,
+    concurrent_apply_to_values,
+    NodeExecutionFailedError,
+)
 from sqlmesh.utils.dag import DAG
 from sqlmesh.utils.date import (
     TimeLike,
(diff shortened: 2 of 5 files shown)


StuffbyYuki:
This is an interesting one. I'm also new to the codebase in general. I wonder somebody from the former Tobiko could jump in, like @izeigerman , @eakmanrq , or @tobymao

mday-io:
(also commenting regarding https://github.com/SQLMesh/sqlmesh/pull/5730) When you say "audit after", when do you mean? What if the backfill is done and then the following audit fails - do you expect that audit failure to be caught on the prod plan application? Can you share an example of a case where you'd want to do this? How long does it take to run your audits to where it's a blocker for you?

If the execution time for audits was improved, is there any scenario you can see that one would need to --skip-audits?

cmgoffena13:
Just putting my 2 cents in here, I think we have three possible changes here and it probably makes sense to break them out so we can review this better:

1. Adding a `--skip-audits` flag for `plan` and `run` commands (Low-ish Complexity)
2. Upgrading the `sqlmesh audit` command to use `concurrent_tasks`, right now its serial (Low-Medium Complexity)
3. Adding in audit parallelism for audits per model, so plans can be faster (Medium Complexity)

2 & 3 can share a helper: take a list of audit jobs, run them with `concurrent_tasks`. Plan/run pass one model’s audits after evaluate (to run audits for a model in parallel). `sqlmesh audit` will concatenate all audits into that list (flat pool), to greatly increase the effectiveness of the command.

I think these changes make sense to provide more flexibility for users, especially if they are coming from dbt. Let's just simplify how we add this in.

One initial piece of feedback: we need to respect the config's concurrency limit. We can't introduce more threads than what is specified in the config. Never more in-flight queries than `concurrent_tasks`.

PR #5732, opened 2026-03-14
Fix: Surface destructive schema changes on forward-only models during dev plans (#5719)

### LINKED ISSUE #5719: Destructive schema changes on forward-only models not surfaced until prod deploy
### Describe the bug

  When a forward-only model has columns removed from its definition, the plan/apply cycle succeeds
  in dev and PR environments but fails with a `DestructiveChangeError` when deploying to production.
  This leaves the production environment in an unfinalized state with downstream models falling behind.

  The failure is deterministic — the change cannot be applied to prod — but the user receives no
  warning until the final prod deploy step.

  ### To Reproduce

  1. Have any forward-only model in production with columns `A`, `B`, `C` — this includes:
     - `INCREMENTAL_BY_PARTITION` (always forward-only by design)
     - Any model with `forward_only = true` in its config
     - Models categorized as forward-only at plan time
  2. Remove column `C` from the model definition
  3. Run `sqlmesh plan dev` and apply — succeeds ✓
  4. Open a PR; CI checks pass — succeeds ✓
  5. Attempt to deploy to production — **fails** with:

  Plan requires destructive change to forward-only model ''s schema that drops columns 'C'.
  To allow the destructive change, set the model's on_destructive_change setting to warn, allow,
  or ignore or include the model in the plan's --allow-destructive-model option.

  ### Expected behavior

  The user should be warned — at minimum during the PR environment plan, ideally during `plan dev` —
  that this change will fail at prod deploy. The outcome is fully predictable from:

  - the model's `forward_only` status
  - `on_destructive_change` defaulting to `ERROR`
  - the column diff between the previous snapshot schema and the new model definition

  There is no reason to let the user reach prod before surfacing this.

  ### Root cause (identified)

  The destructive schema change check (`_check_destructive_schema_change` in
  `snapshot/evaluator.py`) only runs as part of `MigrateSchemasStage`, which is explicitly skipped
  for dev plans:

```python
  if plan.is_dev:
      snapshots_with_schema_migration = []  # always empty in dev
```

  Additionally, dev tables are created fresh with a versioned suffix, so no ALTER TABLE is
  generated — the check would have nothing to compare even if it ran.

  All the information needed to predict the failure is available at plan-build time: the previous
  snapshot schema, the new model schema, whether the model is forward-only, and the
  on_destructive_change setting. The check simply isn't performed proactively.

  ### Impact

  - CI/CD bot deployments fail at the last step, leaving the prod environment in an unfinalized state
  - Downstream models fall behind until an operator manually intervenes
  - No actionable signal is given to the user before the failure occurs

  ### Suggested fix

  During plan construction, for any forward-only model where on_destructive_change = ERROR, diff
  the previous snapshot schema against the new model schema. If a column drop is detected, raise (or
  warn) at plan time — regardless of target environment. This would surface the issue during
  plan dev, as a failed PR check, or at any earlier stage in the pipeline before the user reaches
  prod.

  ### Environment

  - SQLMesh version: 0.230.1

brucearctor:
## Summary

Fixes #5719. Destructive schema changes on forward-only models are now surfaced at plan-build time during dev plans (`sqlmesh plan dev`), not just prod plans.

## Problem

When a forward-only model has columns removed or types narrowed, `sqlmesh plan dev` and PR checks would succeed, but the production deployment would fail with a `DestructiveChangeError` during `MigrateSchemasStage`. This left the production environment in an unfinalized state.

## Root Cause

`PlanBuilder._check_destructive_additive_changes` already checked for destructive schema changes at plan-build time (regardless of `is_dev`), but only for **directly modified** snapshots. **Indirectly modified** forward-only models (e.g., downstream models inheriting schema changes from an upstream dependency) were not checked. The evaluator-level check in `MigrateSchemasStage` catches these for prod, but that stage is skipped for dev plans.

### `sqlmesh/core/plan/builder.py`
- Extract schema diffing into reusable `_check_schema_change` helper method
- Pass `indirectly_modified` mapping to `_check_destructive_additive_changes`
- Add a second loop to check indirectly modified forward-only snapshots for destructive/additive schema changes

### `tests/core/test_plan.py`
Add 5 new tests for `is_dev=True` scenarios:
- `test_forward_only_destructive_change_dev_plan` — ERROR mode raises `PlanError`
- `test_forward_only_destructive_change_dev_plan_warn` — WARN mode logs warning
- `test_forward_only_destructive_change_dev_plan_allow` — ALLOW mode passes silently
- `test_forward_only_indirect_destructive_change_dev_plan` — indirect modification path exercised
- `test_forward_only_flag_destructive_change_dev_plan` — `--forward-only` flag with dev plan

## Test Results

All 27 forward-only/destructive tests pass (5 new + 22 existing), zero regressions.


StuffbyYuki:
@brucearctor it looks like the direct-modification case already raises PlanError on current main. Could you maybe clarify what version/scenario you reproduced, and update the PR description to state that the net-new behavior is the indirect-modification loop (plus the _check_schema_change refactor).

brucearctor:
This PR was opened months ago.  

Way too long has passed for me to recall.

Too stale for me to worry about - it was solving someone else's reported bug -- i was trying to get involved with supporting this project.

brucearctor:
Closing, as i thats too long to maintain context

StuffbyYuki:
@brucearctor Apologize we didn't pick it up sooner. The TSC was formed and starting to be active in the last couple of weeks after sqlmesh had been donated to Linux Foundation. We'll be more active going forward though

brucearctor:
All good.  Apologies not needed.  I just cant recall the status as was someone else's issue.  And given 3 months has passed since opening that PR, it seemed reasonable for me to consider sufficiently stale.

PR #5736, opened 2026-03-17
Chore!!: Migrate sqlglot to v30

VaggelisD:
## Summary of Changes

  | File | Change | Why |
  |------|--------|-----|
  | `pyproject.toml` | `sqlglot[rs]~=28.10.1` → `sqlglot~=30.0.1` | Version bump; `[rs]` extra removed (replaced by `sqlglot`) |
  | 60+ files across `sqlmesh/`, `tests/`, `web/` | `exp.Expression` → `exp.Expr` in all type hints, isinstance checks, and type casts. Class definitions remain as `exp.Expression` subclasses | sqlglot v30 introduced `exp.Expr` as the new base class; `Expression` is now a subclass. Functions like `exp.alias_()`, `exp.func()`, `parse_one()` return `Expr`, causing mypy errors when assigned to `Expression`-typed variables. Custom expression classes must still inherit `Expression` since `Expr` doesn't support constructor arguments |
  | `sqlmesh/core/schema_diff.py`, `tests/core/integration/utils.py` | `exp.DataType.Type` → `exp.DType` in type annotations | `DataType.Type` enum is no longer recognized as a valid type by mypy; `exp.DType` is the v30 alias |
  | `sqlmesh/core/dialect.py` | New import path for `AthenaTrinoParser`, updated `self.expression()` calls | sqlglot API: Athena parser renamed, `expression()` now takes pre-constructed objects |
  | `sqlmesh/utils/jinja.py` | Import `Expression` from `sqlglot.expressions` instead of `sqlglot` | `sqlglot.expressions` is now a package; top-level re-export changed |
  | `sqlmesh/core/model/kind.py` | Clear `meta["sql"]` in `_time_data_type_validator` | `DataType.build` now preserves the parsed expression directly. Our `_parse_types` extension sets `meta["sql"]`, which the pydantic encoder prioritizes over dialect-aware rendering |
  | `sqlmesh/core/model/meta.py` | Normalize column types through dialect roundtrip during deserialization | BigQuery `INT` no longer auto-maps to `BIGINT` during `DataType.build`; roundtripping through dialect SQL restores canonical form for stable data hashes |
  | `sqlmesh/utils/metaprogramming.py` | Added `_resolve_import_module()` to walk module hierarchy for re-exports | `to_table.__module__` is now `sqlglot.expressions.builders` but it's re-exported from `sqlglot.expressions`; generated import statements must use the public module |
  | `sqlmesh/core/context_diff.py` | Added `sqlglotc` to `IGNORED_PACKAGES` | New sqlglot conditionally imports `sqlglotc` (C extension); dependency detection was picking it up as a user requirement |
  | `tests/core/test_macros.py` | Simplified expected Snowflake `ARRAY_GENERATE_RANGE` arithmetic | sqlglot now simplifies `(DATEDIFF(...) + 1 - 1) + 1` to `DATEDIFF(...) + 1` |
  | `tests/core/test_config.py` | Relaxed regex for error message module path | `Column` class now reports `__module__` as `sqlglot.expressions.core` instead of `sqlglot.expressions` |

### DIFF
--- a/tests/utils/test_metaprogramming.py
+++ b/tests/utils/test_metaprogramming.py
@@ -23,6 +23,7 @@ from sqlmesh.utils.metaprogramming import (
     Executable,
     ExecutableKind,
     _dict_sort,
+    _resolve_import_module,
     build_env,
     func_globals,
     normalize_source,
@@ -49,7 +50,7 @@ def test_print_exception(mocker: MockerFixture):
     except Exception as ex:
         print_exception(ex, test_env, out_mock)
 
-    expected_message = r"""  File ".*?.tests.utils.test_metaprogramming\.py", line 48, in test_print_exception
+    expected_message = r"""  File ".*?.tests.utils.test_metaprogramming\.py", line 49, in test_print_exception
     eval\("test_fun\(\)", env\).*
 
   File '/test/path.py' \(or imported file\), line 2, in test_fun
@@ -638,3 +639,18 @@ def test_dict_sort_executable_integration():
     # non-deterministic repr should not change the payload
     exec3 = Executable.value(variables1)
     assert exec3.payload == "{'env': 'dev', 'debug': True, 'timeout': 30}"
+
+
+def test_resolve_import_module():
+    """Test that _resolve_import_module finds the shallowest public re-exporting module."""
+    # to_table lives in sqlglot.expressions.builders but is re-exported from sqlglot.expressions
+    assert _resolve_import_module(to_table, "to_table") == "sqlglot.expressions"
+
+    # Objects whose __module__ is already the public module should be returned as-is
+    assert _resolve_import_module(exp.Column, "Column") == "sqlglot.expressions"
+
+    # Objects not re-exported by any parent should return the original module
+    class _Local:
+        __module__ = "some.deep.internal.module"
+
+    assert _resolve_import_module(_Local, "_Local") == "some.deep.internal.module"


sqlmesh/core/model/kind.py
```diff
@@ -756,13 +756,16 @@ class _SCDType2Kind(_Incremental):
...
+    def _time_data_type_validator(cls, v: t.Union[str, exp.Expr], values: t.Any) -> exp.Expr:
+        if isinstance(v, exp.Expr) and not isinstance(v, exp.DataType):
             v = v.name
         dialect = get_dialect(values)
         data_type = exp.DataType.build(v, dialect=dialect)
+        # Clear meta["sql"] (set by our parser extension) so the pydantic encoder
+        # uses dialect-aware rendering: e.sql(dialect=meta["dialect"]). Without this,
+        # the raw SQL text takes priority, which can be wrong for dialect-normalized
+        # types (e.g., default "TIMESTAMP" should render as "DATETIME" in BigQuery).
+        data_type.meta.pop("sql", None)
```
georgesittas, line 768: Is there a test for this? Why did this work before the bump? Do we risk messing up the serialized version of this field by popping `sql`?

sqlmesh/lsp/main.py
```diff
@@ -374,7 +374,7 @@ def _custom_api(
                     table_diffs = context.context.table_diff(
                         source=source,
                         target=target,
-                        on=exp.condition(on) if on else None,
+                        on=t.cast(exp.Condition, exp.condition(on)) if on else None,
```
georgesittas: Why do we need a `cast` here?

sqlmesh/utils/metaprogramming.py
```diff
@@ -444,6 +444,42 @@ def value(
...
+    # package, stopping at the first ancestor that also exports the object.
+    # We stop before the top-level package to avoid over-normalizing
+    # (e.g. ``sqlglot`` re-exports everything, but callers expect the more
+    # specific ``sqlglot.expressions`` when that is where the object lives).
+    for i in range(len(parts) - 1, 1, -1):
+        parent = ".".join(parts[:i])
+        try:
+            parent_module = sys.modules.get(parent) or importlib.import_module(parent)
+            if getattr(parent_module, name, None) is obj:
+                return parent
```
georgesittas, line 475: Doesn't this only work for 1 parent right now? For example, if the new module became `sqlglot.expressions.foo.bar`, would you stop at `foo`?

sqlmesh/core/model/kind.py
VaggelisD: I think it was because of this commit https://github.com/tobymao/sqlglot/pull/7092.

On **v28**, we'd:
1. Parse the string text into  `exp.DataType` with `parse_one` (setting `meta["sql"]` in the process) 
2. Reconstruct a fresh object again before returning, effectively discarding `meta`

On **v30** we only do (1) now after the optimization PR.

We should have the same serialization and data hashes, otherwise e.g we'd notice BigQuery parsing `"TIMESTAMP"` into `exp.DType.DATETIME` but roundtripping it to `"TIMESTAMP"` again due to `meta` taking priority as raw text verbatim.

sqlmesh/utils/metaprogramming.py
VaggelisD: This is under a for loop walking up to the root module, this should work I believe

sqlmesh/lsp/main.py
VaggelisD: Ah, on v30 that was widened to `exp.Expr`, I'll widen the type hints for all the call chains

sqlmesh/core/model/kind.py
georgesittas: Ok, so the current behavior is coincidentally consistent with what we used to do before the migration, because of the buggy AST node reconstruction in v28?

sqlmesh/lsp/main.py
georgesittas: Got it, thanks.

sqlmesh/utils/metaprogramming.py
georgesittas: I meant that if `foo` re-exports `bar` in the above example, don't we stop at `foo`? Which means that we'll still see a diff in the serialized import due to a different path.

sqlmesh/core/model/kind.py
VaggelisD: Right, that's my understanding as well

sqlmesh/utils/metaprogramming.py
VaggelisD: Ah I think that makes sense, I'll flip the for loop order to descend into nested modules starting from the most public ones

sqlmesh/core/model/kind.py
georgesittas: Ok, if we've verified this then I'm good with this change. Feel free to resolve.

sqlmesh/utils/metaprogramming.py
georgesittas: Can we add a test for this and verify the serialized payload structure?

PR #5745, opened 2026-03-26
Fix: create_external_models misclassifies cross-repo models as external

### LINKED ISSUE #5326: In a multi repo the create_external_models command has problems with models being defined in one project and selected from in another project
# Problem statement

In this [multi repo setup](https://github.com/TobikoData/sqlmesh/blob/main/examples/multi) the model `silver.c` from `repo_2` selects from the model `bronze.a` which is defined in `repo_1`.

It seems that `sqlmesh create_external_models` treats `bronze.a` as an external model for `repo_2`  and fails to discover that the model is in fact defined in `repo_1`.

**SQLMesh version**: 0.216.0

# Reproducible example

Using forementioned multi repo setup:

```sh
cd  sqlmesh/examples/multi
```

Trying to create external models:

```sh
sqlmesh -p repo_1 -p repo_2 create_external_models
```

Output:

```
Initializing new project state...
[WARNING] Unable to get schema for '"db"."bronze"."a"': 'Catalog Error: Table with name a does not exist!
Did you mean "information_schema.tables"?

LINE 1: DESCRIBE "db"."bronze"."a"
```

Despite including both projects SQLMesh doesn't infer that `bronze.a` is defined within the multi-repo.

Planning first:

```sh
sqlmesh -p repo_1 -p repo_2 plan
```

Output:

```
[WARNING] Linter warnings for /home/statsdb/projects/sqlmesh/examples/multi/repo_1/models/b.sql:
 - nomissingdescription: All models should be documented.
 - Line 1: nomissingaudits - Model `audits` must be configured to test data quality.
 - Line 7: noselectstar - Query should not contain SELECT * on its outer most projections, even if it can be expanded.
[WARNING] Linter warnings for /home/statsdb/projects/sqlmesh/examples/multi/repo_1/models/a.sql:
 - nomissingdescription: All models should be documented.
 - Line 1: nomissingaudits - Model `audits` must be configured to test data quality.

`prod` environment will be initialized

Environment statements:

before_all:                                                                                                                                                                                   
  + CREATE TABLE IF NOT EXISTS before_1 AS select @one()                                                                                                                                      
  + CREATE TABLE IF NOT EXISTS before_2 AS select @two()                                                                                                                                      
                                                                                                                                                                                              
after_all:                                                                                                                                                                                    
  + CREATE TABLE IF NOT EXISTS after_1 AS select @dup()                                                                                                                                       
  + CREATE TABLE IF NOT EXISTS after_2 AS select @dup()                                                                                                                                       
                                                                                                                                                                                              
Models:
??? Added:
    ??? bronze.a
    ??? bronze.b
    ??? silver.c
    ??? silver.d
    ??? silver.e
Models needing backfill:
??? bronze.a: [full refresh]
??? bronze.b: [full refresh]
??? silver.c: [full refresh]
??? silver.d: [full refresh]
??? silver.e: [full refresh]
[1/1] bronze.a   [full refresh]                            0.10s   
[1/1] bronze.b   [full refresh]                            0.10s   
[1/1] silver.c   [full refresh]                            0.10s   
[1/1] silver.e   [full refresh]                            0.11s   
[1/1] silver.d   [full refresh]                            0.10s   
Executing model batches ???????????????????????????????????????? 100.0% ? 5/5 ? 0:00:00                                                                                                       
                                                                                                                                                                                              
? Model batches executed

Updating virtual layer  ???????????????????????????????????????? 100.0% ? 5/5 ? 0:00:00

? Virtual layer updated
```

Now creating external models:

```sh
sqlmesh -p repo_1 -p repo_2 create_external_models
```

Output:

```sh
[WARNING] The following models already exist and can't be converted to external: "db"."bronze"."a". Perhaps these models have been removed, while downstream models that reference them 
weren't updated accordingly.
```

Again - despite including both projects SQLMesh doesn't infer that `bronze.a` is defined within the multi-repo.

Creating external models for `repo_2` only:

```sh
sqlmesh -p repo_2 create_external_models
```

There is no terminal output. The file `external_models.yaml` just contains `[]`. So in this case SQLMesh correctly decides that `bronze.a` is not an external model, presumably because `bronze.a` is present in the state of the `prod` environment - but shouldn't the command that includes both projects be able to do the same?

gabepesco:
Resolves #5326 

## Description

In a multi-repo setup, `create_external_models` partitions models per-config before passing them to `create_external_models_file`. This caused dependencies that belong to another repo (e.g. `bronze.a` referenced by `silver.c` in repo_2) to appear missing from the current config's model set and be incorrectly classified as external.

The fix adds an `all_models` parameter to `create_external_models_file` that represents the full set of loaded models across all repos. Dependency resolution uses this complete set, so cross-repo internal models are never mistaken for external ones. The per-config `models` argument is still used to scope which models' dependencies are written to each repo's `external_models.yaml`.

## Test Plan

Added `test_multi_repo_create_external_models` in `tests/core/integration/test_multi_repo.py` using the existing `examples/multi` fixture. The test verifies that after `create_external_models()`, no "Unable to get schema" warnings are emitted for cross-repo models, and that `plan` still resolves all models as internal (no new snapshots treated as external).

Confirmed the test fails without the fix and passes with it. Ran full regression on `tests/core/test_schema_loader.py` and `tests/core/integration/test_multi_repo.py`, all passing. I also monkey patched it into my own pipeline, and the warning was gone and no other issues occurred.

## Checklist

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2488,6 +2488,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 gateway=external_models_gateway,
                 max_workers=self.concurrent_tasks,
                 strict=strict,
+                all_models=self._models,
             )
 
     @python_api_analytics
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -2495,6 +2495,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 gateway=external_models_gateway,
                 max_workers=self.concurrent_tasks,
                 strict=strict,
+                all_models=self._models,
             )
 
     @python_api_analytics
--- a/sqlmesh/core/schema_loader.py
+++ b/sqlmesh/core/schema_loader.py
@@ -24,26 +24,33 @@ def create_external_models_file(
     gateway: t.Optional[str] = None,
     max_workers: int = 1,
     strict: bool = False,
+    all_models: t.Optional[t.Dict[str, Model]] = None,
 ) -> None:
     """Create or replace a YAML file with column and types of all columns in all external models.
 
     Args:
         path: The path to store the YAML file.
-        models: FQN to model
+        models: FQN to model for the current repo/config being processed.
         adapter: The engine adapter.
         state_reader: The state reader.
         dialect: The dialect to serialize the schema as.
         gateway: If the model should be associated with a specific gateway; the gateway key
         max_workers: The max concurrent workers to fetch columns.
         strict: If True, raise an error if the external model is missing in the database.
+        all_models: FQN to model across all loaded repos. When provided, a dependency is only
+            classified as external if it is absent from this full set. This prevents cross-repo
+            internal models from being misclassified as external in multi-repo setups.
     """
+    known_models: t.Union[UniqueKeyDict[str, Model], t.Dict[str, Model]] = (
+        all_models if all_models is not None else models
+    )
     external_model_fqns = set()
 
     for fqn, model in models.items():
         if model.kind.is_external:
             external_model_fqns.add(fqn)
         for dep in model.depends_on:
-            if dep not in models:
+            if dep not in known_models:
                 external_model_fqns.add(dep)
 
     # Make sure we don't convert internal models into external ones.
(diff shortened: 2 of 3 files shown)


StuffbyYuki:
@gabepesco Hey it looks like a few tests including style ci jobs failed. Could you look into those?

blecourt-private:
Thanks for fixing this @gabepesco!

PR #5747, opened 2026-03-28
fix(lsp): read self.context_state instead of stale local variable in _context_get_or_load

jthurlburt:
## Summary

- `_context_get_or_load` captures `self.context_state` into local variable `state` at line 1065
- After `_create_lsp_context()` or `_ensure_context_for_document()` successfully updates `self.context_state` to `ContextLoaded`, the final check at line 1077 reads the stale local `state` (still `NoContext`) instead of `self.context_state`
- This causes the method to always raise `RuntimeError("Context failed to load")` on the first `didOpen`, even when context loading succeeds

## Impact

The LSP logs "Loaded SQLMesh Context" then immediately crashes on the first `textDocument/didOpen`. In VS Code, this puts the extension client into a permanent "Client got disposed and can't be restarted" state. Affects both the `specified_paths` code path (via `initializationOptions.project_paths`) and the `_ensure_context_for_document` fallback path.

## Fix

Read `self.context_state` instead of the stale local `state` variable on lines 1077-1078.

## Regression test

Added two tests to `tests/lsp/test_context.py` that call `_context_get_or_load` starting from `NoContext`, covering both affected code paths:

- `test_context_get_or_load_from_no_context_with_specified_paths` — the `specified_paths` path
- `test_context_get_or_load_from_no_context_via_workspace_folder` — the `_ensure_context_for_document` fallback path

Both fail with `RuntimeError("Context failed to load")` on the unpatched code and pass with the fix.

## Test Plan

- [x] Verified fix locally — LSP hover, go-to-definition, references, and diagnostics all work after patch
- [x] `make style` passes (ruff, ruff-format, mypy, valid migrations)
- [x] Confirmed bug is present on upstream main
- [x] Added regression tests that fail without the fix and pass with it
- [x] Rebased onto latest main and re-verified

### DIFF
--- a/sqlmesh/lsp/main.py
+++ b/sqlmesh/lsp/main.py
@@ -1074,8 +1074,8 @@ class SQLMeshLanguageServer:
                     loaded_sqlmesh_message(self.server)
             else:
                 self._ensure_context_for_document(document_uri)
-        if isinstance(state, ContextLoaded):
-            return state.lsp_context
+        if isinstance(self.context_state, ContextLoaded):
+            return self.context_state.lsp_context
         raise RuntimeError("Context failed to load")
 
     def _ensure_context_for_document(
--- a/tests/lsp/test_context.py
+++ b/tests/lsp/test_context.py
@@ -2,6 +2,7 @@ from pathlib import Path
 
 from sqlmesh.core.context import Context
 from sqlmesh.lsp.context import LSPContext, ModelTarget
+from sqlmesh.lsp.main import ContextLoaded, NoContext, SQLMeshLanguageServer
 from sqlmesh.lsp.uri import URI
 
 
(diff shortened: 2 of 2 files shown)


StuffbyYuki:
@jthurlburt Thanks for the PR! 

Approve with one optional ask question - do you think it'd be beneficial to add a small regression test in tests/lsp/ that calls _context_get_or_load starting from NoContext with specified_paths set? That would prevent this from coming back.

StuffbyYuki:
@jthurlburt thanks for the additional changes! It looks like DCO checks are needed

jthurlburt:
> @jthurlburt thanks for the additional changes! It looks like DCO checks are needed

Thanks for all the feedback! I revised the commit with the DCO so I think we are all good now?

StuffbyYuki:
@jthurlburt Thanks so much! Just merged it

PR #5750, opened 2026-03-30
fix: prevent default catalog leak into catalog-unsupported gateways

### LINKED ISSUE #5748: Bug: Default gateway catalog leaks into catalog-unsupported secondary gateways
## Describe the bug

When a multi-gateway SQLMesh project uses a default gateway that has a default catalog set
(e.g., Trino with `catalog: example_catalog`), that catalog is silently prepended to model
names targeting secondary gateways that do not support catalogs (e.g., ClickHouse).

At evaluation time, the secondary gateway's engine adapter rejects the catalog:

```
UnsupportedCatalogOperationError:
  clickhouse does not support catalogs and a catalog was provided: example_catalog
```

The failure is deterministic. No ClickHouse model can be evaluated in a multi-gateway setup
where the default gateway has a default catalog set.

## To Reproduce

1. Configure a multi-gateway SQLMesh project with a catalog-supporting default gateway and
   a ClickHouse secondary gateway
2. Create a model targeting the ClickHouse gateway:
   ```sql
   MODEL (
       name my_schema.my_model,
       kind VIEW,
       gateway 'clickhouse_gw',
       dialect 'clickhouse',
   );
   SELECT 1 AS id
   ```
3. Run `sqlmesh plan`. The model is loaded and its name is normalized to
   `"example_catalog"."my_schema"."my_model"` (three-part name with leaked catalog)
4. Attempt to evaluate or apply. Fails with `UnsupportedCatalogOperationError`

Minimal reproduction with Python:

```python
from sqlglot import parse
from sqlmesh.core.model import load_sql_based_model

expressions = parse("""
    MODEL (name my_schema.my_model, kind FULL, gateway clickhouse_gw, dialect clickhouse);
    SELECT 1 AS id
""", read="clickhouse")

model = load_sql_based_model(expressions, default_catalog="example_catalog", dialect="clickhouse")

print(model.fqn)      # "example_catalog"."my_schema"."my_model"  <-- catalog leaked
print(model.catalog)   # example_catalog
```

## Expected behavior

The model name should remain a two-part name (`"my_schema"."my_model"`) when targeting a
gateway whose engine does not support catalogs. The default gateway's catalog should only apply
to models that target catalog-supporting gateways.

All the information needed to prevent the leak is available at model-loading time:
- Each adapter exposes `catalog_support` (returns `CatalogSupport.UNSUPPORTED` for ClickHouse)
- The per-gateway catalog dict is already built by the scheduler
- The model loader already has per-gateway override logic

The override logic simply doesn't account for gateways that are absent from the dict.

## Root cause (identified)

The per-gateway catalog dict (`default_catalog_per_gateway`) is built by the scheduler in
`core/config/scheduler.py`:

```python
def get_default_catalog_per_gateway(self, context):
    default_catalogs_per_gateway = {}
    for gateway, adapter in context.engine_adapters.items():
        if catalog := adapter.default_catalog:
            default_catalogs_per_gateway[gateway] = catalog
    return default_catalogs_per_gateway
```

Catalog-unsupported adapters return `None` for `default_catalog` (because the base adapter
short-circuits: `if self.catalog_support.is_unsupported: return None`). They are never added
to the dict.

The model loader in `core/model/definition.py` then checks the dict:

```python
if (
    default_catalog_per_gateway
    and gateway_name
    and (catalog := default_catalog_per_gateway.get(gateway_name)) is not None
):
    loader_kwargs["default_catalog"] = catalog
```

When a gateway is **not in the dict**, this condition is `False`, and the global
`default_catalog` (from the default gateway) remains in `loader_kwargs` unchanged. It gets
prepended to the model name during normalization.

The loader cannot distinguish "this gateway has no catalog" from "this gateway wasn't checked"
because both cases result in the gateway being absent from the dict.

## Impact

- ClickHouse (and potentially other `UNSUPPORTED` engines) cannot be used as a secondary
  gateway in any project where the default gateway has a default catalog set
- The plan step succeeds without warning. The error only surfaces at evaluation/apply time,
  after the user has already committed to the change
- Workarounds exist but are disruptive: removing the catalog from the default gateway
  connection, swapping which gateway is the default, or fully qualifying all model names

## Suggested fix

Two possible fix locations:

**Option A: In the scheduler (`core/config/scheduler.py`)**

Explicitly register catalog-unsupported gateways with empty string in the per-gateway dict:

```python
def get_default_catalog_per_gateway(self, context):
    default_catalogs_per_gateway = {}
    for gateway, adapter in context.engine_adapters.items():
        if catalog := adapter.default_catalog:
            default_catalogs_per_gateway[gateway] = catalog
        elif adapter.catalog_support.is_unsupported:
            default_catalogs_per_gateway[gateway] = ""
    return default_catalogs_per_gateway
```

This works because the model loader checks `is not None`. Empty string passes the check and
overrides `default_catalog` to `""`, which does not get prepended to the model name. The
empty string value represents "this gateway was checked and has no catalog" as distinct from
`None` which means "this gateway was not found in the dict."

**Option B: In the model loader (`core/model/definition.py`)**

Set `loader_kwargs["default_catalog"] = None` when a gateway is known but absent from the dict:

```python
if default_catalog_per_gateway and gateway_name:
    catalog = default_catalog_per_gateway.get(gateway_name)
    if catalog is not None:
        loader_kwargs["default_catalog"] = catalog
    else:
        loader_kwargs["default_catalog"] = None
```

## Test verification

| Test | Input | Model FQN | Status |
|------|-------|-----------|--------|
| Bug reproduction | `default_catalog="example_catalog"` | `"example_catalog"."my_schema"."my_model"` | Catalog leaked |
| Fix verification | `default_catalog=""` | `"my_schema"."my_model"` | Correct |
| Integration | Scheduler with Option A applied | `{'default_gw': 'example_catalog', 'clickhouse_gw': ''}` | Correct |

Regression testing with Option A applied against SQLMesh test suite:
- `test_multi_virtual_layer`: passed (multi-gateway integration)
- `test_multiple_gateways`: passed (multi-gateway context)
- 353 core tests passed; 14 failures + 53 errors match the results on unmodified `main`

## Environment

- SQLMesh version: 0.230.1

brucearctor:
## Description

Fixes #5748.

When a multi-gateway SQLMesh project uses a default gateway with a catalog (e.g., Trino with `catalog: example_catalog`), that catalog is silently prepended to model names targeting secondary gateways that do not support catalogs (e.g., ClickHouse), causing `UnsupportedCatalogOperationError` at evaluation time.

**Root cause:** The per-gateway catalog dict (`default_catalog_per_gateway`) is built in `BuiltInSchedulerConfig.get_default_catalog_per_gateway()`. Catalog-unsupported adapters return `None` for `default_catalog` and are never added to the dict. The model loader in `create_models_from_blueprints()` then cannot distinguish "this gateway has no catalog" from "this gateway was not checked", so the global `default_catalog` leaks through.

**Fix:** Explicitly register catalog-unsupported gateways with empty string (`""`) in the per-gateway dict. The model loader's existing `is not None` check picks up the empty string and overrides `default_catalog` to `""`, which does not get prepended to the model name.

### Changes

| File | Change |
|------|--------|
| `sqlmesh/core/config/scheduler.py` | Added `elif` branch to register catalog-unsupported gateways with `""` |
| `tests/core/test_model.py` | Added regression test `test_default_catalog_not_leaked_to_unsupported_gateway` |

## Test Plan

- Added `test_default_catalog_not_leaked_to_unsupported_gateway` regression test that verifies:
  - With `""` in `default_catalog_per_gateway` → `example_catalog` does **not** leak into the model FQN
  - Without the entry for the gateway → `example_catalog` **does** leak (confirming the bug behavior)
- All existing gateway-related tests pass: `test_gateway_macro`, `test_gateway_macro_jinja`, `test_gateway_python_model`, `test_gateway_specific_render`, `test_model_cache_gateway`
- `ruff` and `ruff-format` pass
- `mypy` passes on changed file (pre-existing unrelated error in `gateway.py`)

## Checklist

### DIFF
--- a/sqlmesh/core/config/scheduler.py
+++ b/sqlmesh/core/config/scheduler.py
@@ -141,6 +141,8 @@ class BuiltInSchedulerConfig(_EngineAdapterStateSyncSchedulerConfig, BaseConfig)
         for gateway, adapter in context.engine_adapters.items():
             if catalog := adapter.default_catalog:
                 default_catalogs_per_gateway[gateway] = catalog
+            elif adapter.catalog_support.is_unsupported:
+                default_catalogs_per_gateway[gateway] = ""
         return default_catalogs_per_gateway
 
 
--- a/tests/core/test_model.py
+++ b/tests/core/test_model.py
@@ -7186,6 +7186,70 @@ def test_gateway_macro_jinja() -> None:
     assert model.render_query_or_raise().sql() == "SELECT 'in_memory' AS \"gateway_jinja\""
 
 
+def test_default_catalog_not_leaked_to_unsupported_gateway() -> None:
+    """Regression test for https://github.com/SQLMesh/sqlmesh/issues/5748.
+
+    When a model targets a gateway that does not support catalogs (e.g. ClickHouse),
+    the default gateway's catalog must not leak into the model's FQN. The scheduler
+    registers such gateways with empty string in default_catalog_per_gateway, and
+    the model loader should use that to override the global default_catalog.
+    """
+    expressions = d.parse(
+        """
+        MODEL (
+            name my_schema.my_model,
+            kind VIEW,
+            gateway clickhouse_gw,
+            dialect clickhouse,
+        );
+        SELECT 1 AS id
+        """
+    )
+
+    # Simulate a multi-gateway setup: default gateway has catalog "example_catalog",
+    # but clickhouse_gw is catalog-unsupported (registered with empty string).
+    models = load_sql_based_models(
+        expressions,
+        get_variables=lambda gw: {},
+        dialect="clickhouse",
+        default_catalog="example_catalog",
+        default_catalog_per_gateway={
+            "default_gw": "example_catalog",
+            "clickhouse_gw": "",
+        },
+    )
+
+    assert len(models) == 1
+    model_result = models[0]
+    # The catalog should be empty — not the leaked "example_catalog"
+    assert model_result.catalog != "example_catalog"
+    # The FQN should be a two-part name, not three-part with the leaked catalog
+    assert "example_catalog" not in model_result.fqn
+
+    # Verify the positive case: without the per-gateway override, catalog leaks
+    models_leaked = load_sql_based_models(
+        d.parse(
+            """
+            MODEL (
+                name my_schema.my_model2,
+                kind VIEW,
+                gateway clickhouse_gw,
+                dialect clickhouse,
+            );
+            SELECT 1 AS id
+            """
+        ),
+        get_variables=lambda gw: {},
+        dialect="clickhouse",
+        default_catalog="example_catalog",
+        # No entry for clickhouse_gw — catalog will leak
+        default_catalog_per_gateway={"default_gw": "example_catalog"},
+    )
+    assert len(models_leaked) == 1
+    # Without the fix, example_catalog leaks into the model name
+    assert "example_catalog" in models_leaked[0].fqn
+
+
 def test_gateway_python_model(mocker: MockerFixture) -> None:
     @model(
         "test_gateway_python_model",


sqlmesh/core/config/scheduler.py
```diff
@@ -141,6 +141,8 @@ def get_default_catalog_per_gateway(self, context: GenericContext) -> t.Dict[str
         for gateway, adapter in context.engine_adapters.items():
             if catalog := adapter.default_catalog:
                 default_catalogs_per_gateway[gateway] = catalog
+            elif adapter.catalog_support.is_unsupported:
```
themisvaltinos, line 144: this will leak the empty string "" into downstream operations as actual catalog value. `None` is already the representation for no catalog in most code paths, so the fix should be in `definition.py` instead, so take a look and adapt it to be similar to the option b suggested in the original issue

themisvaltinos:
fixed here: https://github.com/SQLMesh/sqlmesh/pull/5752

PR #5752, opened 2026-03-30
fix: prevent default catalog leak into catalog-unsupported gateways

mday-io:
**Description:**

When a model's gateway is not in `default_catalog_per_gateway`, explicitly sets `default_catalog` to `None` instead of letting the global default leak through. Implements the fix in `definition.py` as suggested by @themisvaltinos in #5750 (comment).

**Test Plan:**

- 3 new regression tests in `tests/core/test_catalog_leak.py`
- Existing gateway tests pass: `test_gateway_macro`, `test_gateway_macro_jinja`, `test_gateway_python_model`, `test_gateway_specific_render`, `test_model_cache_gateway`
- Multi-gateway integration tests pass: `test_multi_virtual_layer`, `test_multiple_gateways`
- `ruff check` and `ruff format` pass

**Checklist:**

- ran `make style`
- added tests for my changes
- commits are signed off per the DCO
- all existing test pass: haven't yet run the full suite of tests via `make fast-test` (and it has pre-existing failures).

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -2065,7 +2065,9 @@ def create_models_from_blueprints(
     **loader_kwargs: t.Any,
 ) -> t.List[Model]:
     model_blueprints: t.List[Model] = []
+    original_default_catalog = loader_kwargs.get("default_catalog")
     for blueprint in _extract_blueprints(blueprints, path):
+        loader_kwargs["default_catalog"] = original_default_catalog
         blueprint_variables = _extract_blueprint_variables(blueprint, path)
 
         if gateway:
@@ -2083,12 +2085,15 @@ def create_models_from_blueprints(
         else:
             gateway_name = None
 
-        if (
-            default_catalog_per_gateway
-            and gateway_name
-            and (catalog := default_catalog_per_gateway.get(gateway_name)) is not None
-        ):
-            loader_kwargs["default_catalog"] = catalog
+        if default_catalog_per_gateway and gateway_name:
+            catalog = default_catalog_per_gateway.get(gateway_name)
+            if catalog is not None:
+                loader_kwargs["default_catalog"] = catalog
+            else:
+                # Gateway exists but has no entry in the dict (e.g., catalog-unsupported
+                # engines like ClickHouse). Clear the default catalog so the global
+                # default from the primary gateway doesn't leak into this model's name.
+                loader_kwargs["default_catalog"] = None
 
         model_blueprints.append(
             loader(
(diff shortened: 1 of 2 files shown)


tests/core/test_catalog_leak.py
```diff
@@ -0,0 +1,122 @@
...
+support catalogs (e.g., ClickHouse).
+"""
+
+from sqlglot import parse
+
+from sqlmesh.core.model import load_sql_based_model
+from sqlmesh.core.model.definition import load_sql_based_models
+
+
+def test_default_catalog_not_leaked_to_unsupported_gateway():
```
themisvaltinos: nit: if you could move all these tests under `core/test_model.py` instead of having them in a separate file

sqlmesh/core/model/definition.py
```diff
@@ -2083,12 +2083,15 @@ def create_models_from_blueprints(
...
+            if catalog is not None:
+                loader_kwargs["default_catalog"] = catalog
+            else:
+                # Gateway exists but has no entry in the dict (e.g., catalog-unsupported
+                # engines like ClickHouse). Clear the default catalog so the global
+                # default from the primary gateway doesn't leak into this model's name.
+                loader_kwargs["default_catalog"] = None
 
         model_blueprints.append(
             loader(
```
themisvaltinos, line 2099: we also need to adapt in line 2068 to keep the original catalog between blueprint iterations since otherwise it might get overwritten and permanently lost:
```python
    original_default_catalog = loader_kwargs.get("default_catalog")
    for blueprint in _extract_blueprints(blueprints, path):
        loader_kwargs["default_catalog"] = original_default_catalog
```
so keeping the catalog and restore it at the start of each iteration to prevent cross contamination between blueprints in the case that a model uses the gateway controlled by blueprint:
```
MODEL (
  name @{blueprint}.my_model,
  kind FULL,
  gateway @{gw},
  blueprints (
    (blueprint := my_schema, gw := clickhouse_gw),
    (blueprint := other_schema, gw := default_gw)
),
);

SELECT 1 AS id
```

themisvaltinos (commented):
thank you one minor adjustment needed in the code otherwise lgtm. and if possible add also an extra test for the blueprint case I explained

tests/core/test_catalog_leak.py
mday-io: fix applied!
https://github.com/SQLMesh/sqlmesh/compare/90a3566963ac634a0d15507996d245efec6da525..f6188e7ab4238e3fa9c050f7554cfa5b9aaac717

sqlmesh/core/model/definition.py
mday-io: good catch. applied the update:
https://github.com/SQLMesh/sqlmesh/compare/90a3566963ac634a0d15507996d245efec6da525..f6188e7ab4238e3fa9c050f7554cfa5b9aaac717

themisvaltinos (approved):

PR #5759, opened 2026-04-05
Fix: allow --select-model to plan a model deletion

viniciusnunest:
## Description

Fixes #5741

Right now if you delete a model file locally and try to plan its removal with
`--select-model=<deleted-model>`, you get:

```
PlanError: Selector did not return any models.
```

The problem is that `expand_model_selections` only looks at local models. Since
the file is gone, it finds nothing, `backfill_models` ends up empty, and
`_plan()` raises. The workaround was adding some unrelated active model to the
selection, which works but it is not great.

The thing is, the underlying machinery already handles deletions fine. When
`backfill_models` has at least one entry, deleted models show up correctly in
`context_diff.removed_snapshots`. So the fix is about making the selector and
plan validation aware that "no local matches" can still be valid when matching
models pending deletion in the deployed environment.

Changes:
- Extracted `_load_env_models()` from `select_models()` to avoid duplicating
  the environment loading logic
- Added `expand_model_selections_with_env()` that expands selections against
  both local and environment models
- In `_plan()`, when `backfill_models` is empty, it now checks whether the
  selection matched models in the environment that were deleted locally. If so,
  `PlanError` is suppressed and the deletions surface via `removed_snapshots`
- Selecting a model that exists neither locally nor in the environment still
  raises `PlanError` as before

## Test Plan

- `test_expand_model_selections_with_env`: deleted models found in env, local
  models still work, wildcards match, non-existent returns empty
- `test_expand_model_selections_with_env_fallback`: fallback environment is used
  when target env is missing
- `test_expand_model_selections_with_env_expired`: expired environments ignored
- `test_plan_select_model_deleted_model`: end-to-end with the sushi example,
  deploys to prod, deletes model file, reloads, runs plan and checks that the
  model appears in `removed_snapshots`
- `test_plan_select_model_deleted_model_still_rejects_nonexistent`: negative
  case, makes sure garbage input still fails

## Checklist

### DIFF
--- a/sqlmesh/core/selector.py
+++ b/sqlmesh/core/selector.py
@@ -78,27 +78,9 @@ class Selector(abc.ABC):
         Returns:
             A dictionary of models.
         """
-        target_env = self._state_reader.get_environment(Environment.sanitize_name(target_env_name))
-        if target_env and target_env.expired:
-            target_env = None
-
-        if not target_env and fallback_env_name:
-            target_env = self._state_reader.get_environment(
-                Environment.sanitize_name(fallback_env_name)
-            )
-
-        env_models: t.Dict[str, Model] = {}
-        if target_env:
-            environment_snapshot_infos = (
-                target_env.snapshots
-                if not ensure_finalized_snapshots
-                else target_env.finalized_or_current_snapshots
-            )
-            env_models = {
-                s.name: s.model
-                for s in self._state_reader.get_snapshots(environment_snapshot_infos).values()
-                if s.is_model
-            }
+        env_models = self._load_env_models(
+            target_env_name, fallback_env_name, ensure_finalized_snapshots
+        )
 
         all_selected_models = self.expand_model_selections(
             model_selections, models={**env_models, **self._models}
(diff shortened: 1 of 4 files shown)


eakmanrq:
Thanks for your submission @viniciusnunest. Does this solution assume that the selector only contains deleted models or not deleted models? What if there was a mix?

viniciusnunest:
@eakmanrq Good question. It handles both cases.

The new code path (`if not backfill_models`) only kicks in when the selection matched zero local models, so when every selected model is a deletion. In that scenario, `expand_model_selections_with_env` confirms the models exist in the deployed environment, and we let the plan proceed so the deletions surface through `context_diff.removed_snapshots`.

When there is a mix (some active + some deleted), `expand_model_selections` still finds the active ones and `backfill_models` is not empty, so the new block is skipped entirely. The deleted models show up in `removed_snapshots` naturally because they exist in the environment but not in the local project. That part of the behavior was already working before this PR, the issue was only when the selection contained nothing but deletions.

sqlmesh/core/selector.py
eakmanrq, line 87: Is this what you need since it uses the env models? I'm wondering if you could return this alongside the current UniqueKeyDict and then use that and remove the need for `expand_model_selections_with_env`? Ex:
```
models_override, selected_fqns = model_selector.select_models(...)
if not backfill_models:
    selected_deletion_fqns = selected_fqns - set(self._models)
```

eakmanrq:
> @eakmanrq Good question. It handles both cases.
> 
> The new code path (`if not backfill_models`) only kicks in when the selection matched zero local models, so when every selected model is a deletion. In that scenario, `expand_model_selections_with_env` confirms the models exist in the deployed environment, and we let the plan proceed so the deletions surface through `context_diff.removed_snapshots`.
> 
> When there is a mix (some active + some deleted), `expand_model_selections` still finds the active ones and `backfill_models` is not empty, so the new block is skipped entirely. The deleted models show up in `removed_snapshots` naturally because they exist in the environment but not in the local project. That part of the behavior was already working before this PR, the issue was only when the selection contained nothing but deletions.

Thanks I understand now. Can you make sure tests cover this case?

viniciusnunest:
@eakmanrq  Good call, refactored. `select_models` now returns the matched FQNs alongside the model dict, so `context._plan` can figure out which selections are deletions without loading env models a second time. Removed `expand_model_selections_with_env` entirely.

Also added test coverage for the mixed case (active + deleted in the same selection). The mixed scenario was already working before but now it is explicitly tested, `test_select_models_returns_selected_fqns` covers deleted-only, local-only, mixed, wildcard, and nonexistent.

viniciusnunest:
Hey @eakmanrq ,  I pushed a fix addressing the linting failures from the last run. Could you re-trigger the CI checks when you get a chance?

viniciusnunest:
Hey @eakmanrq, I took a look at the latest CI run and I believe the two failures are not related to my changes:

1. **style-and-cicd-tests** - `test_audit_formatting_flag_serde` fails with `AttributeError: 'NoneType' object has no attribute 'get'` in `sqlmesh/utils/pydantic.py:55`. I didn't touch `test_audit.py` or `pydantic.py`. I ran it locally (including with xdist parallelism) and it passes consistently. Looks like a flaky test, probably some race condition when running the full suite in parallel.

2. **cicd-tests-windows** - SSL error during `pip install`. Infrastructure issue on GitHub Actions side.

All linting (ruff, ruff-format, mypy) passes across all Python versions in this run, and all 2974 other tests passed too.

Should I look into fixing the flaky test as part of this PR, or would you prefer to handle it separately?

eakmanrq (approved):
Thanks @viniciusnunest!

PR #5783, opened 2026-05-01
Fix(databricks): Get correct datatypes from information_schema table

MisterWheatley:
## Description
Updated `DatabricksEngineAdapter` to query `system.information_schema.columns` (documented [here](https://docs.databricks.com/gcp/en/sql/language-manual/information-schema/columns)) over running `DESCRIBE <table_fqn>` as there are problems particularly with respect to long and nested struct columns (see issue: https://github.com/SQLMesh/sqlmesh/issues/5781).

## Test Plan
A manual test with a script was created on a sample table. Minimal script (assuming that adapter and connection is set up elsewhere and passed as argument)

```python
def get_dbx_columns(
    adapter: EngineAdapter, table: str, schema: str, catalog: str
) -> t.Optional[dict[str, exp.DataType]]:
    info_schema = exp.to_table("system.information_schema.columns")
    query = (
        exp.select("column_name", "full_data_type")
        .from_(info_schema)
        .where(
            exp.and_(
                exp.column("table_catalog", table="columns").eq(catalog),
                exp.column("table_schema", table="columns").eq(schema),
                exp.column("table_name", table="columns").eq(table),
            )
        )
        .order_by("ordinal_position ASC")
    )

    result = adapter.fetchall(query)
    parsed_columns = {
        row.column_name: exp.DataType.build(row.full_data_type, dialect="databricks")
        for row in result
    }
    return parsed_columns
```
Output was then manually checked over to correspond with definition from UI

Additionally a unit test has been added to `tests/core/engine_adapter/test_databricks.py`

NB: When running `make fast-test` some `mssql` and `fabric` tests were observed to be failing with a `cannot open shared object file` error. I assume these are unrelated to changes to Databricks adapter. 

In addition `tests/core/test_context.py::test_wildcard` failed with a `SQLMesh project config could not be found error` again assumed to be unrelated to this change.

## Checklist


StuffbyYuki:
@MisterWheatley does `make fast-test` pass without your changes?

MisterWheatley:
> @MisterWheatley does `make fast-test` pass without your changes?

Hi @StuffbyYuki, sorry for the late reply, it was a holiday weekend.

I just cloned the main branch, set up a new venv and ran `make install-dev`, then ran `make fast-test` on my machine and still got the error.

A quick google search lead me to believe that my machine was missing the `unixodbc` and `unixodbc-dev` as it was missing a `libodbc.so.2` binary, which lead me to believe that it is not declared/installed correctly as part of `make install-dev`.

I installed those two packages and re-ran `make fast-test` that then succeeded (with warnings), confirming that the original errors reported were unrelated to my changes but simply due to me missing a dependency. 

Let me know if you need anything else added/tested regarding my PR and I will be happy to oblige. If it requires more/new integration tests for Databricks, I will probably need a bit of help getting started as I don't fully grasp how the specific engine integration tests are loaded

MisterWheatley:
@StuffbyYuki 

I can see integration test fails. It seems I made a mistake in implementing the DB API by mixing up Adapter implementation and adapter use.

The correct code should be in `databricks_adapter.py`
```python
self.cursor.execute(query)
result = self.cursor.fetchall()
```
edit: The unit test might need to be updated as well to check the call signature

StuffbyYuki:
@MisterWheatley Can you maybe open a followup PR to fix?

MisterWheatley:
> @MisterWheatley Can you maybe open a followup PR to fix?

Certainly. I didn't do it right away as I was uncertain how to. I'll get right on it.

MisterWheatley:
@StuffbyYuki Follow up PR is open here: https://github.com/SQLMesh/sqlmesh/pull/5816, again apologies.

It seems I had local changes that weren't included as part of my original commit, which is why I didn't catch it myself.

StuffbyYuki:
@MisterWheatley Thanks! I just kicked off the workflows on that PR. Let's continue the discussion there.

PR #5797, opened 2026-05-18
Perf: batch column comment ALTER statements in Databricks adapter

callum-mcdata:
Instead of executing one ALTER TABLE ... ALTER COLUMN ... COMMENT query per column (~1s each on Databricks), batch all column comments into a single statement using Databricks' supported multi-column ALTER syntax.

## Description

  - Overrides _create_column_comments in DatabricksEngineAdapter to batch all column comments into a single ALTER TABLE ... ALTER COLUMN col1 COMMENT '...', COLUMN col2 COMMENT '...' statement
  - Previously each column comment was a separate query (~1s per query on Databricks), making it very slow for tables with many columns
  - Adds a unit test verifying the batched SQL output

## Test Plan

Followed the testing below. 

## Checklist

### DIFF
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -411,3 +411,34 @@ class DatabricksEngineAdapter(SparkEngineAdapter, GrantsFromInfoSchemaMixin):
         return super()._build_column_defs(
             target_columns_to_types, column_descriptions, is_view, materialized
         )
+
+    def _create_column_comments(
+        self,
+        table_name: TableName,
+        column_comments: t.Dict[str, str],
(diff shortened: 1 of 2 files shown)


StuffbyYuki:
@callum-mcdata Would you mind taking a look at resolving conflicts?

gabepesco:
Hey @callum-mcdata, it looks like this is [unsupported syntax prior to DBR 16.3](https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/sql-ref-syntax-ddl-alter-table-manage-column#alter-column-clause). Do we have/need a mechanism to tell when this can be used safely for older LTS DBRs?

PR #5802, opened 2026-05-21
Fix: Make the janitor best effort

themisvaltinos:
This updates the janitor instead of stopping on the first drop failure, to attempt every cleanup operation and aggregates all failures into a single error raised at the end. The `warn_on_delete_failure` is also respected so in that case it is set to true it would warn rather than raise.

It also introduces the force delete option. If the janitor fails to drop a view, schema or snapshot table it retains the corresponding state records so the next run can retry. This is correct for transient failures, but if the failure is permanent the janitor will retry forever and the affected environments and snapshots will never be cleaned up.

  In that case,  with the `--force-delete` flag:

  ```bash
  sqlmesh janitor --force-delete
  ```

purges the state records for expired environments and snapshots even when their physical objects could not be dropped. But any objects that were not successfully dropped becomes orphaned and must be removed manually from the data warehouse.

### DIFF
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -887,12 +887,12 @@ class GenericContext(BaseContext, t.Generic[C]):
         return completion_status
 
     @python_api_analytics
-    def run_janitor(self, ignore_ttl: bool) -> bool:
+    def run_janitor(self, ignore_ttl: bool, force_delete: bool = False) -> bool:
         success = False
 
         if self.console.start_cleanup(ignore_ttl):
             try:
-                self._run_janitor(ignore_ttl)
+                self._run_janitor(ignore_ttl, force_delete=force_delete)
                 success = True
             finally:
                 self.console.stop_cleanup(success=success)
@@ -2896,12 +2896,14 @@ class GenericContext(BaseContext, t.Generic[C]):
 
         return True
 
-    def _run_janitor(self, ignore_ttl: bool = False) -> None:
+    def _run_janitor(self, ignore_ttl: bool = False, force_delete: bool = False) -> None:
         current_ts = now_timestamp()
         failures: t.List[str] = []
 
         # Clean up expired environments by removing their views and schemas
-        failures.extend(self._cleanup_environments(current_ts=current_ts))
+        failures.extend(
+            self._cleanup_environments(current_ts=current_ts, force_delete=force_delete)
+        )
 
         failures.extend(
             delete_expired_snapshots(
@@ -2909,6 +2911,7 @@ class GenericContext(BaseContext, t.Generic[C]):
                 self.snapshot_evaluator,
                 current_ts=current_ts,
                 ignore_ttl=ignore_ttl,
+                force_delete=force_delete,
                 console=self.console,
                 batch_size=self.config.janitor.expired_snapshots_batch_size,
             )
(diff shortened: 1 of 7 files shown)


sqlmesh/core/janitor.py
```diff
@@ -85,10 +85,8 @@ def get_adapter(gateway_managed: bool, gateway: t.Optional[str] = None) -> Engin
                 console.update_cleanup_progress(expired_view)
         except Exception as e:
             message = f"Failed to drop the expired environment view '{expired_view}': {e}"
-            if warn_on_delete_failure:
-                logger.warning(message)
-            else:
-                raise SQLMeshError(message) from e
+            logger.warning(message)
```
georgesittas, line 88: Do we need both this warning and the one emitted in context.py:L2922? Similarly for L103 and L116.

sqlmesh/core/janitor.py
```diff
@@ -165,17 +162,23 @@ def delete_expired_snapshots(
...
+                    start=RowBoundary.lowest_boundary(),
+                    end=batch.batch_range.end,
+                ),
+                ignore_ttl=ignore_ttl,
+            )
+            logger.info("Cleaned up expired snapshots batch")
+            num_expired_snapshots += len(batch.expired_snapshot_ids)
+        except Exception as e:
+            message = f"Failed to clean up an expired snapshots batch: {e}"
+            logger.warning(message)
```
georgesittas, line 175: Ditto.

tobymao (approved):

sqlmesh/core/context.py
```diff
@@ -2923,15 +2935,19 @@ def _cleanup_environments(self, current_ts: t.Optional[int] = None) -> None:
...
+                        engine_adapters=self.engine_adapters,
+                        environments=[expired_env],
+                        console=self.console,
+                    )
                 )
 
-        self.state_sync.delete_expired_environments(current_ts=current_ts)
+        # we want to retry on the next janitor pass if drops failed
+        if not failures:
+            self.state_sync.delete_expired_environments(current_ts=current_ts)
```
crericha, line 2957: In the case where a schema was permanently deleted by the user, prior to the janitor running, won't this cause the environment to never expire? I feel like we need an escape hatch.

sqlmesh/core/janitor.py
themisvaltinos: they're complimentary rather than duplicate. these are per drop and added to the `logs` file immediately so that we can get it in detail and when it happened. the one in context is the aggregated one which surfaces it at job completion and warns if the warn flag is set either wise raises. 

so my thinking is that if someone is tailing the logs they would want to see each failure as and when it happened. in terms of the cli it is only the error that is raised the warnings dont show up there:

Screenshot 2026-05-21 at 18 23 28

georgesittas: Gotcha. Ok, seems reasonable.

sqlmesh/core/context.py
themisvaltinos: so this pr unblocks the janitor so it doesn't accumulate stale snapshots in the case of one failure and it can continue with the rest of the operations

this guard to not delete the expired environments in this case of a failure is matching the current behaviour we have of retrying the next time that the janitor runs rather than deleting the expired environments and leaving orphans in the db. the test: `test_janitor_cleanup_order` is an example of this behaviour that we currently want

I see the value of an escape hatch for the cases where drops persistently fail, but this would require a more deliberate mechanism. I can't think of a simple way to do it as part of this pr without changing the janitor design substantially unless you have a suggestion of a simple way

vchan (approved):

Issue #5182, opened 2025-08-19
sqlmesh format requires DB connection


lucargir:
When using the sqlmesh CLI to format the project, it tries to connect to a DB. Given that I am using it as part of PR tests to make sure the code is properly formatted, there's no need for external connections. Would it be possible to drop this requirement?

j-hartshorn:
We have encountered something similar too; format is failing in CI because it's trying to authenticate with GCP. Formatting shouldn't need to connect or auth at all.

j-hartshorn:
PR #5804 makes sqlmesh `format` runnable without a reachable state backend.

Unfortunately I couldn't make the `lint` command follow suit because it needs the remote state to resolve reference in the m multi-repo case. Details are in the PR discussion. some possible follow ups if anyone wants to take a look:
- Attempt the state connection and fall back, documenting that this might cause linter errors.
- Allow an explicit flag for `lint` to make it ignore state, documenting that this may cause extra linting errors to appear in the multi-repo case.

PR #5804, opened 2026-05-21
feat: skip state-sync for the `format` command

j-hartshorn:
## Description

`sqlmesh format` previously required a reachable state-sync backend on every invocation, because `Context.load()` unconditionally merges remote-prod state into the local model registry whenever a config declares a `project:` name. That made the command unrunnable in CI jobs that have no credentials or network access to the state backend, even though formatting is a pure local operation - it parses each local file and re-prints the AST, never resolving references.

This PR adds a `load_state: bool = True` parameter to `GenericContext` that gates the two `if any(self._projects):` state-merge blocks in `Context.load()`, and adds a `LOCAL_ONLY_COMMANDS = ("format",)` tuple in `sqlmesh/cli/main.py` that flips it to `False` for that subcommand. The default behaviour for every other command (`plan`, `run`, `lint`, `diff`, etc.) is unchanged.

**Scope note:** this PR originally covered `lint` as well, but was narrowed after review (see discussion below). In multi-repo projects, lint results depend on models merged from prod state at load time - models owned by other repos aren't on disk, and skipping the merge flipped lint from pass to fail even with state credentials available (reproduced with `examples/multi`, where repo_2's `silver.c` selects from repo_1's `bronze.a`). `lint` therefore still loads state, locked in by a guard test, and the lint case in #5182 remains open.

Verified end-to-end with a real `config.yaml` declaring a Postgres state connection via `{{ env_var() }}` with no env vars set: pre-PR, `sqlmesh format` exits 1 with `psycopg2.OperationalError: could not translate host name`. Post-PR, exit 0.

One known caveat: Snowflake state connections under default authentication validate `user`/`password` at config-load time via `SnowflakeConnectionConfig._validate_authenticator` (`sqlmesh/core/config/connection.py:633-638`), which runs before this PR's gate. That pre-existing validator means a Snowflake-default-auth state config still requires placeholder user/password values to be set. Postgres, MySQL, MSSQL, and the other backends pass validation with placeholder values.

## Test Plan

Seven tests across two files:

- `tests/core/test_format.py::test_format_without_state_load` (fast) - `context.format(check=True)` with `load_state=False` and a patched `EngineAdapterStateSync.get_versions` that raises. Asserts the patched method is never called.
- `tests/cli/test_cli.py::test_format_runs_without_state` (slow) - end-to-end via the CLI runner.
- `tests/cli/test_cli.py::test_local_only_commands_skip_state_multiple_paths` (slow) - same, invoked with two `--paths`, locking in that the gate applies regardless of invocation shape.
- `tests/cli/test_cli.py::test_format_runs_without_state_multi_repo_partial` (slow) - formats one repo of `examples/multi` whose upstream models live only in prod state; the shape that motivated narrowing the PR.
- `tests/cli/test_cli.py::test_lint_still_loads_state` (slow) - guard-rail for the narrowing. Spies on `Context.__init__`, asserts `load_state=True` is explicitly passed for `lint`, and asserts state sync is actually reached.
- `tests/cli/test_cli.py::test_plan_still_loads_state` (slow) - same guard for `plan`. Both verified non-vacuous by regression modes (deleting the kwarg from cli/main.py, and re-adding the command to `LOCAL_ONLY_COMMANDS`).
- `tests/cli/test_cli.py::test_format_does_not_open_state_connection` (slow) - most realistic CI scenario: a `config.yaml` declaring a Postgres state connection via `{{ env_var() }}` with all env vars unset. Verifies the gate prevents the connection.

Commands run:
- `make py-style` - clean (ruff, ruff-format, mypy, valid migrations).
- Full test selection above - all passing.


j-hartshorn:
Addresses [Issue #5182](https://github.com/SQLMesh/sqlmesh/issues/5182)

StuffbyYuki:
@j-hartshorn  Thanks for this PR! If this PR intends local-only behavior for format/lint regardless of invocation shape, could we perhaps add a multi---paths regression test to lock that in? Let me know if I'm missing anything!

j-hartshorn:
@StuffbyYuki Thanks for the review! It _is_ the intent to have local-only behaviour regardless of invocation shape.

I added the test you recommended (parametrised over `format` and `lint`), and generalised the test helper to support it.

While looking into the multi-paths case in more detail I found a problem with this PR's approach for `lint`, and I'm not sure how to proceed. In a multi-repo project, models owned by other repos aren't on disk - `Context.load()` reconstructs them from the prod environment's snapshots in the state backend (the two blocks this PR gates). Forcing `lint` to never load state means those models become unknown names, which changes lint results.

I reproduced this with `examples/multi`, where repo_2's `silver.c` selects from repo_1's `bronze.a`. After applying both repos to prod, linting repo_2 alone - with the state backend reachable:

```
nomissingexternalmodels     load_state=True: clean    load_state=False: LinterError
invalidselectstarexpansion  load_state=True: clean    load_state=False: LinterError
```

So under this PR, `sqlmesh lint` in one repo of a multi-repo project flips from pass to fail, even when state credentials are available. (`format` is unaffected - it only re-prints local files and never resolves references.)

Options as I see them:

- Lock in this failure mode with a test and document it. I don't like this - it breaks legitimate multi-repo use.
- _Try_ the state connection and fall back to local-only if it's unreachable, perhaps with a warning. Keeps current behaviour when state is available while still fixing the no-credentials CI case.
- Narrow the PR: keep `format` local-only, leave `lint` loading state as it does today.

What do you think?

StuffbyYuki:
I want to say we'd go with option 3 "Narrow the PR: keep format local-only, leave lint loading state as it does today" which is probably the safest option and let's open/keep the issue up so we know it still needs to be worked on.

Thoughts?

j-hartshorn:
Sounds like the most prudent option, I'll make that change.

StuffbyYuki:
@j-hartshorn thanks for making the changes! It looks like DCO checks are needed

j-hartshorn:
@StuffbyYuki Thanks again for your time on this. I'll leave a comment on the issue so the `lint` side of this doesn't get lost.

Issue #5438, opened 2025-09-25
SQLMesh generates DROP TABLE for PostgreSQL views when using DuckDB engine


juqkai:
**Describe the bug**

I am using SQLMesh with DuckDB as an intermediate engine to transfer data from a MySQL source to a PostgreSQL destination. This setup is necessary because SQLMesh does not directly support cross-engine reads and writes within a single model.

The issue stems from a known bug in DuckDB's postgres_scanner extension, which incorrectly identifies PostgreSQL views as BASE TABLE instead of VIEW. This is documented in the DuckDB repository here: https://github.com/duckdb/duckdb-postgres/issues/269

This incorrect type information is then passed to SQLMesh. When SQLMesh needs to manage database objects (for example, when applying a plan that replaces an existing model), it relies on the object type reported by the engine to generate the correct DROP statement.

Because DuckDB reports my PostgreSQL view as a TABLE, SQLMesh generates a DROP TABLE IF EXISTS "dwd__dev"."scm_contract" command. This command subsequently fails when executed against PostgreSQL because scm_contract is actually a VIEW, leading to an exception that halts the process.

**To Reproduce**

1. Set up a SQLMesh project using DuckDB as the default engine.
2. Within DuckDB, attach a PostgreSQL database that contains a view.
3. Create a SQLMesh model that is materialized as a view in the target PostgreSQL database.
4. Run sqlmesh plan to create the view. This will succeed.
5. Make any change to the model's SQL definition.
6. Run sqlmesh plan again. When the plan attempts to apply the changes, it will try to drop the old version of the view in PostgreSQL.
7. The operation fails because SQLMesh issues a DROP TABLE command against the view.

**Expected behavior**

SQLMesh should ideally generate the correct DDL statement (DROP VIEW) for the object in the target database, perhaps by querying the target database's catalog directly before dropping, rather than relying solely on the intermediate engine's (DuckDB's) description of the object.

**Actual behavior**

SQLMesh generates an incorrect DROP TABLE statement for a PostgreSQL view, causing the plan application to fail with an error because the object types do not match.

Environment (please complete the following information):

    SQLMesh version: [0.217.0]

    Database engines: PostgreSQL [14.1]

    OS: [windwos 10]

```
[...]

PR #5809, opened 2026-05-22
Fix: Fallback to DROP VIEW if DROP TABLE fails in DuckDB

alessio-giuliano-pix4d:
Related issue: #5438

## Description

DuckDB has trouble detecting the type of relations when interacting with Postgres through `duckdb-postgres`: Views are treated like tables. (see linked issue for more detail)

The proposed fix is simply to add a fallback to drop the relation with `DROP VIEW` if `DROP TABLE` fails. The code is inspired from another PR attempting a fix: https://github.com/SQLMesh/sqlmesh/pull/5695.  

## Test Plan

I tested this locally on a project that was impacted by issue #5438, and verified that the problem was indeed gone. 

Note that since I don't know this codebase at all, I'm not 100% convinced about the robustness of this change. 

## Checklist

    - I don't really see how to test this in any useful way


StuffbyYuki:
@alessio-giuliano-pix4d Can you add tests?

PR #5810, opened 2026-05-22
feat: add --environment flag to sqlmesh janitor for scoped cleanup

mday-io:
## Description

Add an `--environment` / `-e` flag to `sqlmesh janitor` that scopes cleanup to a single named environment, and fix `sqlmesh invalidate <env> --sync` to only clean up the named environment rather than all currently-expired environments.

**New flag**: `sqlmesh janitor --environment <name>` removes views, schemas, and state records for the named environment only. If the environment is not expired or does not exist, a warning is emitted. Global snapshot expiration and interval compaction are skipped when this flag is set — those operations are cross-environment and cannot be safely scoped to a single one.

**Bug fix**: `sqlmesh invalidate <env> --sync` previously called `_cleanup_environments()` with no filter after marking the target environment expired, which caused all other currently-expired environments to be deleted as a side effect. It now scopes the cleanup to the named environment only.

## Test Plan

- Tested manually against the sushi example project using `--gateway duckdb_persistent` with two environments (`dev1`, `dev2`). Invalidating `dev1` and running `sqlmesh janitor --environment dev1` cleaned up `dev1` while leaving `dev2` untouched.
- Verified the warning message fires when targeting an environment that is not expired or does not exist.
- Added state sync tests for name-filtered expiration and deletion (`test_state_sync.py`).
- Added Context-level tests verifying `invalidate_environment(..., sync=True)` passes `name=` to `_cleanup_environments` and that `sync=False` skips cleanup entirely (`test_context.py`).

## Checklist

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -307,9 +307,17 @@ Usage: sqlmesh janitor [OPTIONS]
   The janitor cleans up old environments and expired snapshots.
 
 Options:
-  --ignore-ttl  Cleanup snapshots that are not referenced in any environment,
-                regardless of when they're set to expire
-  --help        Show this message and exit.
+  --ignore-ttl      Cleanup snapshots that are not referenced in any
+                    environment, regardless of when they're set to expire. Has
+                    no effect when --environment is specified.
+  --force-delete    Delete expired environment and snapshot state records even
+                    when the physical table or view drops fail. Any objects
+                    that could not be dropped become orphaned and must be
+                    removed manually.
+  -e, --environment TEXT
+                    Scope cleanup to a single expired environment. Global
+                    snapshot and interval compaction are skipped.
+  --help            Show this message and exit.
 ```
 
 ## migrate
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -629,7 +629,7 @@ def invalidate(ctx: click.Context, environment: str, **kwargs: t.Any) -> None:
 @click.option(
     "--ignore-ttl",
     is_flag=True,
-    help="Cleanup snapshots that are not referenced in any environment, regardless of when they're set to expire",
+    help="Cleanup snapshots that are not referenced in any environment, regardless of when they're set to expire. Has no effect when --environment is specified.",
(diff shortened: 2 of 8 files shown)


StuffbyYuki:
@mday-io Not a blocker for merge, but a few things I noticed: 
- Notebook magic doesn’t expose `--environment`. I don' think it's too big of a deal, but it would be nice that notebook magic is aligned as well
- `get_expired_environments` and `delete_expired_environments` have dup logic, where it might be able to be put in to `_create_expiration_filter_expr` instead. The dup part:
```
where: exp.Expr = self._create_expiration_filter_expr(current_ts)
        if name is not None:
            where = exp.and_(t.cast(exp.Condition, where), exp.column("name").eq(name))
```
- Should we add a Context test that `invalidate_environment(..., sync=True)` calls `_cleanup_environments(name=...)`. The state sync test covers targeted deletion, but not the Context wiring where the `--sync` bug lived.

StuffbyYuki (approved):

mday-io:
> @mday-io Not a blocker for merge, but a few things I noticed:
> 
> * Notebook magic doesn’t expose `--environment`. I don' think it's too big of a deal, but it would be nice that notebook magic is aligned as well
> * `get_expired_environments` and `delete_expired_environments` have dup logic, where it might be able to be put in to `_create_expiration_filter_expr` instead. The dup part:
> 
> ```
> where: exp.Expr = self._create_expiration_filter_expr(current_ts)
>         if name is not None:
>             where = exp.and_(t.cast(exp.Condition, where), exp.column("name").eq(name))
> ```
> 
> * Should we add a Context test that `invalidate_environment(..., sync=True)` calls `_cleanup_environments(name=...)`. The state sync test covers targeted deletion, but not the Context wiring where the `--sync` bug lived.
> 
> Let me know if I'm missing anything!

Not tackling the notebooks wrapper for now, but created a [new commit](https://github.com/SQLMesh/sqlmesh/pull/5810/changes/e3e0b07fbfacf9d9d3f938f5c2193bda7944734e) for the other comments. Thanks for the review!

PR #5826, opened 2026-06-03
fix(clickhouse): support multi-gateway projects with catalog-aware engines

mday-io:
## Description

Adding a ClickHouse gateway to a project that already has a catalog-aware gateway (e.g. Trino) causes `sqlmesh plan` to blow up:

```
Error: Failed to update model schemas

Table "default"."connection_test" must match the schema's nesting level: 3.
```

The issue is that sqlglot's `MappingSchema` expects all models to use the same FQN depth. ClickHouse doesn't have a catalog concept so its models are 2-level (`db.table`), while Trino models are 3-level (`catalog.db.table`). SQLMesh doesn't handle the mix — registering models from both gateways causes a nesting mismatch.

Found this by accident: plan worked fine after manually adding a fake third tier to ClickHouse model references, which confirmed the mismatch was the culprit.

The fix assigns a **virtual catalog** to ClickHouse models internally so they match the 3-level depth required by `MappingSchema`. The virtual catalog is stripped before anything hits the wire — ClickHouse never sees it.

The default virtual catalog name is `__{gateway_name}__` (e.g. `__clickhouse__` for a gateway named `clickhouse`), which is visually distinct from real catalog names. It can be overridden via a new `virtual_catalog` field in the ClickHouse connection config.

### Stripping coverage

The virtual catalog is stripped at every point where SQL is emitted to ClickHouse. A `_strip_virtual_catalog()` helper on `ClickhouseEngineAdapter` is called in all overridden public methods that receive table/schema names:

- `create_schema` — schema DDL
- `delete_from` — DELETE statements
- `insert_overwrite_by_partition` — INSERT OVERWRITE paths
- `alter_table` — ALTER TABLE statements (stripped from the Alter expression's target table)

The `@set_catalog()` class decorator handles inherited (non-overridden) methods automatically via the `SINGLE_CATALOG_ONLY` branch.

### Re-materialization warning

When a catalog-aware gateway is added to an existing ClickHouse-only project, every ClickHouse model's FQN changes from 2-level to 3-level — SQLMesh treats these as new snapshots and schedules re-materialization. A plan-time warning is now emitted listing the affected models and explaining the cost (FULL models recreated once; INCREMENTAL_BY_TIME_RANGE models require a full historical backfill).

**Note**: Projects that previously used ClickHouse as the only gateway and now add a catalog-aware gateway will see ClickHouse models re-materialized once — their FQNs change from 2-level to 3-level, which changes their fingerprints. This is a one-time cost at the transition point.

## Checklist

- [x] I have added tests for my changes
  - Unit tests: stripping verified for `create_schema`, `delete_from`, `insert_overwrite_by_partition`, `alter_table`
  - Integration test: virtual catalog injection → FQN uniformity → DDL stripping end-to-end
  - Unit tests: plan-time warning fires when/only when expected


StuffbyYuki:
@mday-io  A few tests failed:
- DCO checks
- Are dbt tests related to this PR? 

Not a blocker, just a thought:
- An integration test via Context + multi-gateway config
- Perhaps release notes for one-time CH re-materialization when adding a catalog-aware gateway
- Maybe document that only ClickHouse is supported today (other UNSUPPORTED engines still break with catalog-aware peers)

PR #5827, opened 2026-06-04
fix(dbt): preserve source identifier quoting when it contains dots or spaces

### LINKED ISSUE #5762: SourceConfig.canonical_name() fails with ValueError: too many values to unpack for source identifiers containing dots
SQLMesh version: 0.231.1                                                                                                                                                                                
Adapter: dbt-snowflake 1.11.3
                                                                                                                                                                                                          
## Description                                                                                                                                                                                             
                                                                                                                                                                                                          
  When a dbt source has a table identifier that contains a dot or space (common with Airbyte-ingested CSV files stored as Snowflake tables, e.g. "my_csv.csv"), loading the project fails 
  with:
                                                                                                                                                                                                          
```
  Error: 1 validation error for SqlModel                                                                                                                                                                
  depends_on                                                                                                                                                                                              
    Value error, too many values to unpack (expected 3)
```
                                                                                                                                                                                                          
## Root cause                                                                                                                                                                                            
                                                                                                                                                                                                          
  SourceConfig.canonical_name() in sqlmesh/dbt/source.py unconditionally strips all identifier quoting before calling relation.render(). This produces a string like                                      
  database.my_schema.my_csv_file.CSV, which sqlglot cannot parse as a table expression. The fallback in exp.to_table calls split_num_words(sql_path, ".", 3), which raises ValueError: too many 
  values to unpack (expected 3) because the string splits into 4 parts.                                                                                                                                   
                                                                                                                                                                                                        
  Minimal reproduction

```
  Source YAML:
  sources:
    - name: my_source
      database: RAW_DEV                                                                                                                                                                                   
      tables:          
        - name: my_table                                                                                                                                                                                  
          identifier: "FILENAME.CSV"                                                                                                                                                                    
          quoting:                  
            identifier: true     
```                                                                                                                                                                         
                            
## Fix  
(will put a PR in soon)                                                                                                                                                                                                
                                                                                                                                                                                                          
  In sqlmesh/dbt/source.py, preserve quoting when the identifier contains dots or spaces:
                                                                                                                                                                                                          
  needs_identifier_quoting = bool(                                                                                                                                                                      
      self.table_name                                                                                                                                                                                     
      and ("." in self.table_name or " " in self.table_name)
  )                                                                                                                                                                                                       
  relation = relation.quote(                                                                                                                                                                            
      database=False,
      schema=False,
      identifier=needs_identifier_quoting,
  )                                                                                                                                                                                                       
   
  ---

jagannalla:
## Problem
When a dbt source table identifier contains a dot (e.g., `"FILENAME.CSV"`) or a space, `SourceConfig.canonical_name()` unconditionally strips the quoting. This renders the canonical name as `database.schema.FILENAME.CSV`. 

When `sqlglot` tries to parse this unquoted string back into an AST, it splits the path by `.` into 4 components, causing a `ValueError: too many values to unpack (expected 3)` and failing to load the project.

## Solution
This PR checks if the resolved table name contains a dot (`.`) or a space (` `), and if so, preserves the quoting for the identifier component during canonicalization. Standard table names remain unquoted and normalized.

## Test Plan
- Added `test_source_canonical_name_with_dots_and_spaces` to `tests/dbt/test_config.py` validating correct quoting behavior for dots, spaces, target database omission, and standard identifiers.
- Ran all dbt config tests successfully (`64 passed`).
- Verified style, formatting, and types with `pre-commit` (`ruff`, `ruff-format`, and `mypy` all passed).
Fixes #5762

### DIFF
--- a/sqlmesh/dbt/source.py
+++ b/sqlmesh/dbt/source.py
@@ -82,10 +82,13 @@ class SourceConfig(GeneralConfig):
                     f"'source' macro failed for '{self.config_name}' with exception '{e}'."
                 )
 
+            needs_identifier_quoting = bool(
+                self.table_name and ("." in self.table_name or " " in self.table_name)
+            )
             relation = relation.quote(
                 database=False,
                 schema=False,
-                identifier=False,
+                identifier=needs_identifier_quoting,
             )
             if relation.database == context.target.database:
                 relation = relation.include(database=False)
--- a/sqlmesh/dbt/source.py
+++ b/sqlmesh/dbt/source.py
@@ -82,9 +82,8 @@ class SourceConfig(GeneralConfig):
                     f"'source' macro failed for '{self.config_name}' with exception '{e}'."
                 )
 
-            needs_identifier_quoting = bool(
-                self.table_name and ("." in self.table_name or " " in self.table_name)
-            )
+            identifier = relation.identifier or ""
+            needs_identifier_quoting = "." in identifier or " " in identifier
             relation = relation.quote(
                 database=False,
                 schema=False,
--- a/tests/dbt/test_config.py
+++ b/tests/dbt/test_config.py
@@ -521,6 +521,71 @@ def test_quoting():
     assert str(BaseRelation.create(**source.relation_info)) == 'foo."bar"'
 
 
+def test_source_canonical_name_with_dots_and_spaces(mocker):
+    from dbt.adapters.base import BaseRelation
+
+    mock_context = mocker.Mock()
+    mock_context.target.database = "target_db"
+
+    def mock_source_macro(source_name, table_name):
+        if table_name == "my_table_dot":
+            identifier = "FILENAME.CSV"
+        elif table_name == "my_table_space":
+            identifier = "my table space"
+        else:
+            identifier = "my_table_std"
+        return BaseRelation.create(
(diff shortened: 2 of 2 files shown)


StuffbyYuki:
@jagannalla Does it make more sense to have logic like:
```
identifier = relation.identifier or ""
needs_identifier_quoting = "." in identifier or " " in identifier
```

than 

```
needs_identifier_quoting = bool(
    self.table_name and ("." in self.table_name or " " in self.table_name)
)
```

?

The former approach might be slightly safer because it checks the final, dbt-resolved name, not just what's in the config file. 

jagannalla:
Thanks for the reviews, @StuffbyYuki and @noezhiya-dot! 
I have updated the PR to address the feedback:
* Changed the check to use the resolved `relation.identifier` instead of `self.table_name` to ensure we are always checking the final, dbt-resolved name.
* Verified that all tests are passing. 
Ready for merge!

StuffbyYuki:
@jagannalla Can you make sure all commits have DCO checks? Thanks!

jagannalla:
@StuffbyYuki All commits have been signed off with the correct email (`jnalla987@gmail.com`) and force-pushed. The DCO checks should be passing now!

PR #5828, opened 2026-06-04
fix: Swap parser override order for macro IF under T-SQL (Issue #5823)

### LINKED ISSUE #5823: Required keyword: 'true' missing for <class 'sqlglot.expressions.functions.If'>
Version: 0.234.1

I am trying to add primary keys to MSSQL models within a post-statement @IF.  I can't get beyond this error:

`Required keyword: 'true' missing for <class 'sqlglot.expressions.functions.If'>`

I have copied the example here but still get the same error:
https://sqlmesh.readthedocs.io/en/stable/concepts/macros/sqlmesh_macros/#prepost-statements

My research took me to this issue/comment and it looks to be very close to what I need but I can't get beyond the issue with any and all @IF usage.
https://github.com/SQLMesh/sqlmesh/issues/4568#issuecomment-2918575750

jagannalla:
### Summary
This PR fixes a bug where conditional macro statements (e.g. `@IF(cond, statement)`) using the T-SQL / MSSQL dialect would fail validation with:
`Required keyword: 'true' missing for <class 'sqlglot.expressions.functions.If'>`

### Root Cause
In `extend_sqlglot()` within `sqlmesh/core/dialect.py`, the overrides were registered in the wrong order:
```python
_override(TSQL.Parser, Parser._parse_if)
_override(Parser, _parse_if)

```

### Solution
1. Swapped the registration lines in `sqlmesh/core/dialect.py` so the custom parser wrapper is registered on the base `Parser` first.
2. Added unit tests under `tests/core/test_dialect.py` for T-SQL-specific statements parsed within `@IF`.

### Link to Issue
Closes #5823

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -556,6 +556,10 @@ def _parse_if(self: Parser) -> t.Optional[exp.Expr]:
     # to parse a statement / command to support the macro @IF(condition, statement)
     index = self._index
     try:
+        if self.dialect == "tsql":
+            if not (self._index >= 2 and self._tokens[self._index - 2].text == "@"):
+                return self.__parse_if()  # type: ignore
+            return Parser.__parse_if(self)  # type: ignore
         return self.__parse_if()  # type: ignore
     except ParseError:
         self._retreat(index)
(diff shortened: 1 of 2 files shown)


jagannalla:
Thanks for the review, @noezhiya-dot!

I have updated the PR to address all comments and successfully force-pushed:
* **Test Case**: Added an explicit regression test to verify that the original error case from issue #5823 (the `Required keyword: 'true' missing` error) no longer occurs. It now correctly parses `@IF` into an `Anonymous` expression rather than `exp.If`.
* **Comments**: Added an explanatory comment in the test suite regarding SQLGlot's fallback `Command` parser stripping string literal quotes from unparsed commands like `PRINT`.

Ready for review/merge!

PR #5846, opened 2026-06-17
feat: support `CLUSTER BY [AUTO, NONE]`  for Databricks

EhabEasee:
## Description

Databricks supports two keyword forms of liquid clustering that don't take column arguments:

- `CLUSTER BY AUTO` — lets Databricks automatically select clustering columns
- `CLUSTER BY NONE` — disables liquid clustering on a table

Previously, SQLMesh had no way to express these in a model definition. This PR adds support for both.

**`constants.py`:** Adds `LIQUID_CLUSTERING_KEYWORDS = frozenset({"AUTO", "NONE"})` as a shared constant used across the parser, validator, and adapter.

**Parsing (`dialect.py`):** The `clustered_by` property parser now recognises bare `AUTO` and `NONE` tokens (unquoted `VAR` tokens) as liquid clustering keywords rather than column references. Backtick-quoted \`auto\` / \`none\` are still treated as regular column names, preserving backwards compatibility for columns that happen to share those names.

**Validation (`meta.py`):** A single string passed to `clustered_by` is normalised to a list before processing. The validator then skips the column-count check for `exp.Var(AUTO|NONE)`, but only when the field is `clustered_by` and the dialect is `databricks`. On deserialisation from JSON, keyword strings are restored to `exp.Var` sentinels before `list_of_fields_validator` can normalise them into quoted columns.

**Validation (`definition.py`):** The `validate_definition` column-existence check skips keyword sentinels for the same `clustered_by` + `databricks` scope.

**Code generation (`databricks.py`):** `_build_table_properties_exp` detects a single `exp.Var` in `clustered_by` (guarded by a `ValueError` if the Var holds an unexpected value), and emits `CLUSTER BY AUTO` / `CLUSTER BY NONE` without wrapping in a tuple. Multi-column paths are unchanged.

**Usage:**

```sql
-- In a SQLMesh model definition
MODEL (
  name my_catalog.my_schema.my_table,
  kind FULL,
  dialect databricks,
  clustered_by AUTO
);

MODEL (
  name my_catalog.my_schema.my_table,
  kind FULL,
  dialect databricks,
  clustered_by NONE
);
```

Via the Python API, both a plain string and `exp.Var` are accepted:

```python
create_sql_model(..., dialect="databricks", clustered_by="AUTO")
create_sql_model(..., dialect="databricks", clustered_by=exp.Var(this="AUTO"))
```

Columns with the names `auto` or `none` are still supported via backtick quoting:

```sql
MODEL (
  name my_catalog.my_schema.my_table,
  kind FULL,
  dialect databricks,
  clustered_by (`auto`, `none`)
);
```

## Test Plan

- `tests/core/test_dialect.py` — parser round-trips: `AUTO`/`NONE` keywords, backtick-quoted columns, paren-wrapped single columns, multi-column lists, mixed list `(a, AUTO)`, non-Databricks dialect
- `tests/core/test_model.py` — model DDL; Python API with both `exp.Var` and plain string; backtick-quoted column names; `render_definition` output; JSON serialisation round-trip; non-Databricks dialect rejection; mixed-list column treatment
- `tests/core/engine_adapter/test_databricks.py` — adapter emits `CLUSTER BY AUTO` / `CLUSTER BY NONE` without column parens

## Checklist


StuffbyYuki:
@EhabEasee Thanks for this PR! 

Not trying to be nit-picky, but here's a few items:
- Docs: Add a note in model docs that Databricks supports clustered_by AUTO / NONE, and that backticks are needed for real columns named auto/none.
- Test: test_clustered_by_keyword_non_databricks_dialect: perhaps use pytest.raises(ConfigError) instead of (ConfigError, Exception).

EhabEasee:
@StuffbyYuki both comments make sense and I've made the updates. However, the comment in the docs feels misplaced and easy to miss. 

I was considering adding it in the Databricks engine docs but couldn't find a reasonable place to add it. Do you have any suggestions on a more relevant place to add that note? The [StarRocks](https://github.com/SQLMesh/sqlmesh/blob/main/docs/integrations/engines/starrocks.md#configuration-matrix) docs seem to have something similar so I could imitate that?

StuffbyYuki:
@EhabEasee thanks! Yeah I don't think it has to be that big block like starrocks docs do, but I just figured adding something somewhere in the docs might be helpful! I'll let you decide where and how to put it on the docs

EhabEasee:
@StuffbyYuki I added a new section to the databricks integration docs. Let me know if you have any more feedback

StuffbyYuki:
@EhabEasee It looks like your commits need DCO checks!

EhabEasee:
@StuffbyYuki Done

StuffbyYuki (approved):

PR #5847, opened 2026-06-18
fix: Add alias_tokens parameter to _parse_join() for sqlglot 30.9.0 compactibilty 

### LINKED ISSUE #5840: Upgrade to sqlglot 30.9.0 to support external dialect plugins
## Summary

While working on a DB2 adapter for SQLMesh, I encountered a compatibility issue with `sqlglot 30.9.0`.

The DB2 implementation relies on an external DB2 dialect plugin, which currently requires `sqlglot==30.9.0`. However, SQLMesh is currently pinned to `sqlglot~=30.8.0`. Upgrading to `sqlglot 30.9.0` exposes an API incompatibility in SQLMesh's parser overrides.

## Observed Error

```text
TypeError: _parse_join() got an unexpected keyword argument 'alias_tokens'
```

## Root Cause

The signature of `Parser._parse_join()` changed between `sqlglot 30.8.0` and `30.9.0`.

### sqlglot 30.8.0

```python
def _parse_join(
    self,
    skip_join_token: bool = False,
    parse_bracket: bool = False,
)
```

### sqlglot 30.9.0

```python
def _parse_join(
    self,
    skip_join_token: bool = False,
    parse_bracket: bool = False,
    alias_tokens: t.Collection[TokenType] | None = None,
)
```

SQLMesh overrides this method in `dialect.py`, but the override does not currently accept the new `alias_tokens` parameter.

As a result, when sqlglot 30.9.0 internally invokes:

```python
_parse_join(alias_tokens=...)
```

SQLMesh raises the exception shown above.

## Impact

* Prevents SQLMesh from working with `sqlglot 30.9.0`
* Blocks integration with external dialect plugins that depend on newer sqlglot versions
* Affects DB2 adapter development, where the DB2 dialect plugin currently requires `sqlglot 30.9.0`

## Possible Resolution

Either:

1. Upgrade SQLMesh to support `sqlglot 30.9.0`, or
2. Make the `_parse_join()` override forward-compatible with newer sqlglot parser signatures.

## Additional Context

I discovered this issue while implementing a DB2 adapter, but the underlying problem appears to be a general compatibility issue with newer sqlglot releases rather than something specific to DB2.

## Proposed Contribution

I believe I have a fix for this issue and would be happy to contribute it as a PR if maintaining compatibility with `sqlglot 30.9.0` is something the project would like to support.

iawanish:
## Description

Fixes #5840
This PR fixes compatibility with `sqlglot` 30.9.0 by updating the overridden `_parse_join()` method in `sqlmesh/core/dialect.py` to match the upstream parser signature.

### Problem

With `sqlglot` 30.9.0+, SQLMesh raises:

```text
TypeError: _parse_join() got an unexpected keyword argument 'alias_tokens'
```

This happens because `sqlglot` introduced the `alias_tokens` parameter in `Parser._parse_join()`, while SQLMesh's override was still using the older method signature.

### Changes

- Added the optional `alias_tokens` parameter to `_parse_join()`.
- Forwarded `alias_tokens` to the internal `__parse_join()` calls.
- Preserved backward compatibility by making the parameter optional.

This change allows SQLMesh to work correctly with `sqlglot` 30.9.0+ without affecting existing behavior.

## Test Plan

- Tested with `sqlglot==30.9.0`.
- Verified that the previous `TypeError` no longer occurs.
- Tested the DB2 adapter using the external DB2 sqlglot dialect.
- Confirmed that SQL parsing behaves as expected after the change.

## Checklist

- [] I have run `make style` and fixed any issues
- [x] I have added tests for my changes (if applicable) - **No new tests needed; existing tests verify the fix**


StuffbyYuki:
@iawanish Thanks for the PR! 

A few nits:
- perhaps fix the title typo, correct the “tests added” checklist item, and confirm CI passes. 
- If the goal is official sqlglot 30.9 support (not just compatibility when users override the pin), consider a follow-up to bump sqlglot~=30.9.0 and run the full test suite.

Also it looks like a bunch of ci tests failed. Prob due to sqlglot version changes in this pr

PR #5848, opened 2026-06-18
fix: Databricks column type identification

### LINKED ISSUE #5841: Databricks runs fail to execute on 0.235.4 with thrift error
# Summary

When running with SQLMesh 0.235.4 and Databricks SQL Connector, existing environments fail to run in Databricks with the error:

`ERROR:databricks.sql.thrift_backend:ThriftBackend.attempt_request: Exception: encoding without a string argument`.

Downgrading to 0.235.3 resolves this issue.

## Root Cause

The underlying model SQL I can see being executed successfully against a Databricks Warehouse, the failure occurs when acquiring the built tables' column types.

I believe that this has been caused by PRs #5783 and #5816, as when I remove the `DatabricksEngineAdapter.columns` method added by those PRs to `sqlmesh/core/engine_adapter/databricks.py` this error does not occur.

Potentially the fix would be to change line 434 from:

```py
self.cursor.execute(query)
```

to:

```py
self.cursor.execute(query.sql(dialect="databricks"))
```

## Environment Tested

Environments:
1. Type: Databricks Cluster 17.3 LTS
   Connection mode: Databricks SQL Connector
   Python: 3.12.3
1. Type: Windows 11
   Connection mode: Databricks SQL Connector
   Python: 3.12.3

The Databricks SQL Connector and thrift versions do not make a difference in success, it's purely whether version 0.235.3 or 0.235.4 is running.

davem-bis:
## Description

I have updated the call to the execution cursor in `DatabricksEngineAdapter.column` to have rendered SQL passed to it as opposed to the SQLGlot AST it was previously being supplied.

This should resolve #5841.

## Test Plan

I have run this change against an existing environment in Databricks using:

- databricks-sql-connector==4.0.5
- thrift==0.20.0

This was run against a classic SQL Warehouse against Unity Catalog.

I have updated the unit test for this method to expect rendered SQL to be passed.

## Checklist

  > Note: This is currently failing on main, however I don't believe this has introduced any new issues. `ruff` and `ruff-format` pass, whilst `mypy` fails on main.

  > Note: There are failing tests on main. Tests on `DatabricksEngineAdapter.column` pass.

### DIFF
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -431,7 +431,7 @@ class DatabricksEngineAdapter(SparkEngineAdapter, GrantsFromInfoSchemaMixin):
             .order_by("ordinal_position ASC")
         )
 
-        self.cursor.execute(query)
+        self.cursor.execute(query.sql(dialect="databricks"))
         result = self.cursor.fetchall()
 
         return {row[0]: exp.DataType.build(row[1], dialect=self.dialect) for row in result}
--- a/sqlmesh/core/engine_adapter/databricks.py
+++ b/sqlmesh/core/engine_adapter/databricks.py
@@ -431,7 +431,7 @@ class DatabricksEngineAdapter(SparkEngineAdapter, GrantsFromInfoSchemaMixin):
             .order_by("ordinal_position ASC")
         )
 
-        self.cursor.execute(query.sql(dialect="databricks"))
+        self.execute(query.sql(dialect=self.dialect))
         result = self.cursor.fetchall()
 
         return {row[0]: exp.DataType.build(row[1], dialect=self.dialect) for row in result}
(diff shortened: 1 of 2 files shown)


StuffbyYuki:
@davem-bis Thanks for the PR! 

Two small nits (non-blocking):

- Use `query.sql(dialect=self.dialect)` instead of hardcoded "`databricks`" for consistency with the rest of this file.
- Consider `self.execute(query)` + `self.cursor.fetchall()` instead of `self.cursor.execute(...)` — matches other columns() implementations and reuses the adapter’s normal SQL rendering/logging path.

Those two failed dbt tests would pass if you rebase your branch btw

davem-bis:
Hi @StuffbyYuki - I've pushed those changes.

I've rebased to the latest version `main` but I'm still seeing test failures for other parts of the codebase locally, it's possible my local environment isn't setup correctly. The unit tests for the modified code and the ruff formatting all pass though.

PR #5851, opened 2026-06-19
fix: non-breaking detection for redundant cast projections

albertosuman-1k5:
### Summary
- Fixes a false-positive Breaking categorization when adding a mid-list projection with a redundant same-type cast, e.g. `new_col::STRING` above an existing `::STRING` projection.
- Adds a conservative projection-additivity fallback for `SqlModel.is_breaking_change()` when SQLGlot’s AST diff emits spurious `Move` / `Update` edits.
- Preserves conservative behavior for real changes such as projection rewrites, removals, reorders, filter/group changes, and cardinality-changing UDTFs.

### Context
SQLGlot’s tree diff can cross-match structurally similar cast nodes, especially identical `DataType` leaves such as `STRING`. This can make a purely additive projection change look like a non-`Insert` diff, causing SQLMesh to return `None` from `is_breaking_change()`. Under `AutoCategorizationMode.FULL`, that becomes `BREAKING`.

The new fallback verifies directly that:
- both rendered queries are top-level `SELECT`s,
- the query skeleton is unchanged except for the projection list,
- previous projections are preserved in order,
- and added projections do not include cardinality-changing UDTFs.

## Test plan
- `python -m pytest tests/core/test_snapshot.py -k "categorize_change_sql or categorize_change_seed" -q`
- `python -m pytest cast_breaking_change_demo/test_cast_breaking_change.py -q`
- `python -m pytest tests/core/test_snapshot.py tests/core/test_model.py::test_is_breaking_change tests/core/test_plan.py -q`
- `pre-commit run --files sqlmesh/core/model/definition.py tests/core/test_snapshot.py cast_breaking_change_demo/test_cast_breaking_change.py`

## Checklist

### DIFF
--- a/sqlmesh/core/model/definition.py
+++ b/sqlmesh/core/model/definition.py
@@ -1588,7 +1588,7 @@ class SqlModel(_Model):
 
         for edit in edits:
             if not isinstance(edit, Insert):
-                return None
+                return _additive_projection_change(previous_query, this_query, self.dialect)
 
             expr = edit.expression
             if isinstance(expr, exp.UDTF):
@@ -1602,7 +1602,7 @@ class SqlModel(_Model):
                 expr = parent
 
             if not _is_projection(expr) and expr.parent not in inserted_expressions:
-                return None
+                return _additive_projection_change(previous_query, this_query, self.dialect)
 
         return False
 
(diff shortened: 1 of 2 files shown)


albertosuman-1k5:
Can we trigger again the workflows? Pretty sure the first run failed because of the issue described in #5852

mday-io:
> Can we trigger again the workflows? Pretty sure the first run failed because of the issue described in #5852

rerunning...

sqlmesh/core/model/definition.py
```diff
@@ -2907,6 +2907,75 @@ def _is_projection(expr: exp.Expr) -> bool:
...
+    # Adding a UDTF projection (e.g. EXPLODE / UNNEST) can change row cardinality, so such a
+    # change is not safely non-breaking even when it appears as an extra SELECT item.
+    for projection in this_projections:
+        if isinstance(projection, exp.UDTF) and not projection.find_ancestor(exp.Subquery):
+            return None
+
+    # Everything other than the projection list must be structurally identical. Replacing each
+    # SELECT list with the same dummy literal lets the expression equality check focus on the
+    # FROM / WHERE / GROUP BY / ORDER BY / etc. parts of the query.
+    previous_skeleton = previous_query.copy()
```
mday-io, line 2965: This fallback can incorrectly treat projection inserts as non-breaking when the query uses positional references outside the SELECT list.

For example, this is currently categorized as NON_BREAKING on this branch:

`SELECT a::DATE, s::TEXT FROM t ORDER BY 2`
->
`SELECT a::DATE, x::TEXT, s::TEXT FROM t ORDER BY 2`

But `ORDER BY 2` changes from sorting by `s` to sorting by `x`. The same issue applies to `GROUP BY 2`. Replacing both select lists with a dummy literal makes the skeleton comparison pass even though ordinal references still depend on the original projection positions.

sqlmesh/core/model/definition.py
```diff
@@ -2907,6 +2907,75 @@ def _is_projection(expr: exp.Expr) -> bool:
...
+    previous_projections = previous_query.expressions
+    this_projections = this_query.expressions
+    # If the new query has not gained any projections, this cannot be an additive projection-only
+    # change, so there is nothing for this fallback to prove.
+    if len(this_projections) <= len(previous_projections):
+        return None
+
+    # Adding a UDTF projection (e.g. EXPLODE / UNNEST) can change row cardinality, so such a
+    # change is not safely non-breaking even when it appears as an extra SELECT item.
+    for projection in this_projections:
```
mday-io, line 2957: This UDTF guard misses aliased UDTF projections because `EXPLODE(y) AS y` is an `exp.Alias`, not an `exp.UDTF`.

When the fallback is triggered by same-type casts, this branch can categorize a change like this as NON_BREAKING:

`SELECT a::DATE AS a, s::TEXT AS s FROM t`
->
`SELECT a::DATE AS a, x::TEXT AS x, EXPLODE(y) AS y, s::TEXT AS s FROM t`

Upstream returns `None` for this case. The check probably needs to inspect newly added projections for nested UDTFs, while still allowing UDTFs inside scalar subqueries as the existing tests expect.

mday-io:
Could you add tests for queries that reference SELECT-list positions from other clauses, such as ordinal `ORDER BY` / `GROUP BY`, to ensure those stay undetermined instead of being auto-categorized as non-breaking?

It would also be good to add coverage for aliased UDTF projections, since those may not appear as a top-level UDTF expression in the projection list but should still remain conservative.

albertosuman-1k5:
@mday-io added logic to avoid report as non breaking when order/group by uses positional arguments whose reference is past the introduced expression and to correctly mark aliased UDTF

PR #5856, opened 2026-06-21
feat(janitor): add scoped snapshot cleanup for invalidated environments

### LINKED ISSUE #5844: Add scoped snapshot cleanup to `invalidate` command
## Problem

Currently, when invalidating an environment with `sqlmesh invalidate <env> -s`, the command only removes the environment record and virtual layer (views/schemas), but does not clean up the physical snapshot tables. 

To clean up snapshots, users must run `sqlmesh janitor --ignore-ttl` separately. However, this operates globally and will delete **all** unreferenced snapshots across the project, including those from other previously invalidated environments that haven't been cleaned up yet.

This creates a problem when you want to clean up snapshots for a specific environment without affecting other environments' snapshots.

## Proposed Solution

Add a `--cleanup-snapshots` flag to the `invalidate` command that triggers scoped snapshot cleanup for only the target environment. This would:

1. Capture the snapshot IDs from the environment being invalidated (before deletion)
2. Identify which of those snapshots are not referenced by any other environment
3. Delete only those exclusively-owned snapshots

## Implementation Approach

The feature requires changes across multiple layers:

### 1. CLI Layer (`sqlmesh/cli/main.py`)
Add a new flag to the `invalidate` command:
```python
@click.option(
    "--cleanup-snapshots",
    is_flag=True,
    help="Clean up snapshots exclusively used by this environment after invalidation",
)
```

### 2. Context Layer (`sqlmesh/core/context.py`)
Modify `invalidate_environment` to capture snapshot IDs before deletion and pass them to a scoped cleanup function when the flag is set. [1](#4-0) 

### 3. State Sync Layer (`sqlmesh/core/state_sync/db/facade.py`)
Add a method to get expired snapshots scoped to a specific set of snapshot IDs, while still respecting the protection of snapshots referenced by other environments.

### 4. Snapshot State Layer (`sqlmesh/core/state_sync/db/snapshot.py`)
Modify `get_expired_snapshots` to accept an optional `target_snapshot_ids` parameter that restricts the query to only those IDs, while still excluding any that appear in other environments' `promoted_snapshot_ids`. 

### 5. Janitor Layer (`sqlmesh/core/janitor.py`)
Add support for scoped snapshot deletion in `delete_expired_snapshots` and `iter_expired_snapshot_batches`. 

## Example Usage

```bash
# Invalidate environment and clean up its exclusively-owned snapshots
sqlmesh invalidate dev_mday_picnic_ibuk_test -s --cleanup-snapshots
```

This would be equivalent to running:
```bash
sqlmesh invalidate dev_mday_picnic_ibuk_test -s
sqlmesh janitor --ignore-ttl  # but scoped to only this environment's snapshots
```

## Notes

- The implementation must preserve the existing protection mechanism: snapshots shared with other environments (e.g., prod) must never be deleted
- The `--cleanup-snapshots` flag should work with or without the `-s` flag
- This feature would be particularly useful for CI/CD workflows where PR environments are frequently created and invalidated

Wiki pages you might want to explore:
- [Overview (SQLMesh/sqlmesh)](/wiki/SQLMesh/sqlmesh#1)

### Citations

**File:** docs/integrations/github.md (L289-292)
```markdown
| Option                                | Description                                                                                                                                                                                                                                                                                                                                                                                               |  Type  | Required |
|---------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:------:|:--------:|
| `invalidate_environment_after_deploy` | Indicates if the PR environment created should be automatically invalidated after changes are deployed. Invalidated environments are cleaned up automatically by the Janitor. Default: `True`                                                                                                                                                                                                             |  bool  |    N     |
| `merge_method`                        | The merge method to use when automatically merging a PR after deploying to prod. Defaults to `None` meaning automatic merge is not done. Options: `merge`, `squash`, `rebase`                                                                                                                                                                                                                             | string |    N     |
```

mday-io:
## Summary

Closes #5844

Adds scoped snapshot cleanup for invalidated environments.

`sqlmesh invalidate --cleanup-snapshots` now invalidates the target environment and then runs scoped janitor cleanup for only unreferenced snapshots formerly referenced by that environment. This avoids the previous workaround of running `sqlmesh janitor --ignore-ttl`, which performs project-wide cleanup of all unreferenced snapshots.

The change also makes the scoped cleanup behavior available directly through:

```bash
sqlmesh janitor --environment dev_my_pr_env --ignore-ttl
```

This is particularly useful for CI/CD workflows where PR environments are frequently created and invalidated, but cleanup should not remove unrelated unreferenced snapshots from other invalidated environments.

## How it works

1. Captures the expired/invalidated environment's snapshot IDs before deleting the environment state.
2. Uses `get_expired_snapshots(..., target_snapshot_ids=...)` to restrict candidate snapshots to that captured set.
3. Still excludes any snapshot referenced by another active/finalized environment, such as `prod`.
4. Emits a warning before scoped physical snapshot table cleanup.
5. Drops physical tables through `SnapshotEvaluator.cleanup()`.
6. Deletes snapshot state through the normal janitor/state-sync expired snapshot path, so interval cleanup still runs with the same semantics as regular janitor cleanup.

## Examples

```bash
sqlmesh invalidate dev_my_pr_env --cleanup-snapshots
```

```bash
sqlmesh janitor --environment dev_my_pr_env --ignore-ttl
```

## Checklist

- [x] I have added tests for my changes
- [x] My commits are signed off (`git commit -s`) per the DCO

## Additional validation

- [x] `pytest tests/core/state_sync/test_state_sync.py -q` — 92 passed
- [x] `pytest tests/core/integration/test_aux_commands.py -q` — 13 passed


StuffbyYuki:
@mday-io Might be a good follow-up: GitHub CICD bot (`controller.py`) still calls `invalidate_environment()` without `cleanup_snapshots=True`.  Can be a separate follow-up if we want PR env teardown to auto-clean snapshot tables.

StuffbyYuki:
@mday-io following up on my prev comment, when you come back from your vacation!

mday-io:
@StuffbyYuki so I explored setting "cleanup_snapshots=True" after every PR deploy but that would override the TTL behavior for snapshots (1 week). I meant to include this as a "clean the house" option for any stale snapshot issues - like ones I've had personally.

StuffbyYuki:
@mday-io Great! Just one leftover nit: docs/reference/cli.md still has a stray “Has” in the --ignore-ttl help (“Has When --environment…”). Refer to my pending review above...

StuffbyYuki:
@mday-io Following up one last time! Or I could also make that small change I suppose

StuffbyYuki:
@mday-io This is merge ready yea? I guess conflicts need to be resolved

mday-io:
@StuffbyYuki yes, all good here :)

PR #5859, opened 2026-06-24
fix: Set DatabricksConnectionConfig to only use a shared_connection with U2M OAuth

### LINKED ISSUE #5858: bug(Databricks): databricks-sql-connector executes queries sequentially regardless of concurrent_tasks
# Summary

When using the Databricks SQL Connector for SQL models and setting `concurrent_tasks > 1`, SQLMesh will spawn as many threads as concurrent_tasks, but they can only execute commands sequentially due to the lock in the `ThreadLocalSharedConnectionPool`.

## Root Cause

I believe this has been caused by setting `shared_connection = True` in #5746, whilst this has resolved the issue of OAuth race conditions, it has done so by forcing everything to be sequential.

# Proposed Solutions

There is an interim solution and a longer term solution that I believe should be implemented:

## Interim Solution

Set `shared_connection` to be based on whether an OAuth authentication method is being used. If an access token is being used, then a non-shared connection is suitable.

In `core/config/connection.py:DatabricksConnectionConfig`:
```py
shared_connection: t.ClassVar[bool] = auth_type is not None
```

This issue will remain for OAuth authentication flows, but this is a quick fix for those using access tokens.

## 'Proper' Solution

For OAuth, change the authentication flow to use a single connection, and then use a non-shared pool after that point.

# Environments Tested On

- Databricks Cluster
  DBR: 17.3 LTS (Python 3.12.3)
  SQLMesh Version: 0.253.3
- Windows 11
  Python: 3.12.3
  SQLMesh Version: 0.253.3

davem-bis:
## Description

The DatabricksConnectionConfig was set to use a shared connection pool for all authentication types to resolve a race condition with U2M OAuth, however this has the side effect that SQL execution is sequential regardless of the `concurrent_tasks` config setting.

This change updates:
- The `shared_connection` variable to be a computed field for all `ConnectionConfig` classes.
- The `shared_connection` variable to return `True` only for U2M OAuth authentication methods for the `DatabricksConnectionConfig` class.

Partially resolves #5858.

## Test Plan

- Unit testing for all 3 situations in Databricks.
- Tested on Python 3.12.3, SQLMesh 0.253.3,:
  - databricks-sql-connector 4.0.5 with Access Token and U2M OAuth.
  - duckdb 1.5.3 with a local database.

## Checklist


StuffbyYuki:
@davem-bis Thanks for the PR! Could you look into the merge conflict and failed ci tests?

StuffbyYuki:
@davem-bis Could you look into failed ci tests?

davem-bis:
Hi @StuffbyYuki - will do, I'm still working on the branch, didn't mean to push the changes yet!

davem-bis:
Hi @StuffbyYuki - I've should have resolved the test failure from the CI.

I haven't been able to get all of the tests to pass locally, but I have at least managed to get the number of failed tests to match that on main! FTAOD, the tests I have failing locally all look to be passing in the CI.

Once I've got a dev container that can run all of the tests locally I'll submit another PR with that in.

StuffbyYuki:
I reviewed the pr. Not a blocker, but maybe align U2M detection with _static_connection_kwargs using not oauth_client_secret instead of oauth_client_id is None? Let me know what you think

davem-bis:
That makes sense to me @StuffbyYuki - I've updated the logic and unit test.

PR #5861, opened 2026-06-25
detect new models in the UI when the project path is a symlink

### LINKED ISSUE #5860: Cannot create new model in SQLMesh Browser UI if the project is a symlinked directory
Hi, thanks for making SQLMesh!

I'm running into a issue with creating new models when launching the UI if the project is a symlinked directory. There are a couple of problems:

- files created in the UI are not picked up unless the browser is refreshed
- new models are not picked up even if the browser is refreshed
- changes to models are not picked up unless the browser is refreshed

Making a change to an existing model then refreshing the browser will enable the UI to pick up new models added.

### To reproduce

```bash
# Create a SLQMesh project
rm -rf ~/sqlmesh-repro
mkdir -p ~/sqlmesh-repro/real
cd ~/sqlmesh-repro/real
uv venv --python 3.13
source .venv/bin/activate
uv pip install "sqlmesh[web]"
sqlmesh init duckdb

# Symlink to another directory and launch the UI
ln -s ~/sqlmesh-repro/real ~/sqlmesh-repro/linked
sqlmesh -p ~/sqlmesh-repro/linked ui
```

Then try to create a new model in the UI. It will only be picked up if an existing model is modified and the browser is then refreshed.

### Cause

I think part of the problem is in `web/server/watcher.py`. The watcher watches the project at the path you launched with (i.e. the symlinked path) so the change events coming from `watchfiles.awatch` carry symlinked paths. But the list of model directories it checks each event against is built with `.resolve()`, which expands the symlink back to the real location:

https://github.com/SQLMesh/sqlmesh/blob/192fbe9e15e36283e8d364b5f49660ee703f710a/web/server/watcher.py#L24-L30

This means the following check is always `False` for symlinked files:

https://github.com/SQLMesh/sqlmesh/blob/192fbe9e15e36283e8d364b5f49660ee703f710a/web/server/watcher.py#L79

And so changes are not picked up by the watcher. I'm not sure why refreshing the browser would make a difference, nor why modifying an existing model is needed for new models to be picked up?

I appreciate the browser is deprecated, but I'm finding it quite useful for now, and I'm happy to make a PR to fix this.

-----

Environment:

sqlmesh: 0.235.4
python: 3.13
os: macOS (Darwin 24.5.0, x86_64)

-----

Here's a video showing the broken behaviour:

https://github.com/user-attachments/assets/0635e154-261e-4d53-bb34-e5d011e7592b

p-j-smith:
## Description

Fixes #5860

- `awatch` can report resolved paths even for a symlinked root, which broke the watcher's path handling. Now, compare paths in resolved space and re-root under the project path so it works either way
- Resolve the path in the model-dir (`in_paths`) check too, so new files under a symlinked project get tracked.
- Currently there a check in `watch_project` that is always truthy (`should_reset_mtime = change == Change.added or is_modified_new_file`). Fix this so that `should_reset_mtime` is not always `True`

## Test Plan

Added a new file `tests/web/test_watcher.py` that:

- adds a fixture that creates a new project and symlink to to it
- add a test that creates a `Context` with the symlinked project. Add a new file and check it is picked up by the watcher

## Checklist

### DIFF
--- a/tests/web/test_watcher.py
+++ b/tests/web/test_watcher.py
@@ -6,6 +6,7 @@ from pathlib import Path
 import pytest
 from watchfiles import Change
 
+from sqlmesh.cli.project_init import init_example_project
 from sqlmesh.core.context import Context
 from web.server import watcher as watcher_module
 from web.server.settings import Settings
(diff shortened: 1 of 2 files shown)


p-j-smith:
Closing as the UI is deprecated and I've switched to the VSCode extension

PR #5863, opened 2026-06-29
feat(jinja): add b64decode/b64encode builtin filters

### LINKED ISSUE #5754: Request: Support base64 decode as a Jinja plugin (for configuration yaml files)
Context:

I run SQLMesh on a shared system that connects to BigQuery. BigQuery authentication keys are stored as long JSON files that (in my setup, at least) don't work nicely as environment variables. The "workaround" that seems effective (...in my setup, at least) is to base64 encode the JSON file contents and store that as an environment variable. I can then use a script to decode the environment variable, write the blob to a file, and then have SQLMesh configured to read the key from that file. It works, albeit somewhat clunkily.

Request:

It would be a nice enhancement to have a Jinja2 filter built in for this. Ansible packages something like this with their Jinja2 filter; the code they use is here:
https://github.com/ansible/ansible/blob/ed7623ecdec8585282ce91f1534d02e6a38c22a4/lib/ansible/plugins/filter/core.py#L377

This would let me just run SQLmesh directly from the command line and pull my credentials from the environment, rather than needing to explicitly decode the secret beforehand.

anxkhn:
## Description

Core SQLMesh ships no builtin Jinja filters, so there is no in-tree way to decode a base64-encoded value inside a config YAML. As issue #5754 notes, this is awkward for secrets that are commonly stored base64-encoded in environment variables (for example a GCP/BigQuery service-account key), since they cannot be decoded where the connection config is rendered.

This adds two builtin filters, `b64decode` and `b64encode` (matching ansible's filter names referenced in the issue), and registers them on the shared Jinja `environment()` factory in `sqlmesh/utils/jinja.py`. Because `ENVIRONMENT` is built from that factory and config files are rendered through it, the filters work in config YAML:

```yaml
env_vars:
  BIGQUERY_KEY: "{{ env_var('BIGQUERY_KEY_B64') | b64decode }}"
```

They are also available in models, since the model `JinjaMacroRegistry` builds its environment from the same factory. Both helpers are str/bytes tolerant and return `str`. This mirrors the existing dbt builtin-filter pattern (`create_builtin_filters` in `sqlmesh/dbt/builtin.py`).

Closes #5754

## Test Plan

- Added `test_builtin_base64_filters` in `tests/utils/test_jinja.py`: renders `b64decode`/`b64encode` (and a round-trip) via a built environment, and asserts the filter is usable in config YAML through `yaml.load`.
- Confirmed the new test fails on `main` (`No filter named 'b64decode'`) and passes on the branch.
- `pytest tests/utils/test_jinja.py` -> 14 passed; dbt jinja/filter tests + `tests/utils/test_yaml.py` pass.
- `ruff check`, `ruff format --check`, `mypy sqlmesh/utils/jinja.py` clean.

## Checklist

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -170,6 +170,16 @@ The examples specify a Snowflake connection whose password is stored in an envir
           account: <account>
     ```
 
+    !!! tip "Base64-encoded secrets"
+
+        If a secret is distributed base64-encoded in a single environment variable (for example a BigQuery service-account key), pipe the variable through the built-in `b64decode` filter to decode it to text inline:
+
+        ```yaml
+        keyfile_json: {{ env_var('BIGQUERY_KEY_B64') | b64decode }}
+        ```
+
+        A matching `b64encode` filter is also available. Both return UTF-8 text, so they are intended for string/JSON secrets rather than arbitrary binary data.
+
 === "Python"
 
     Python accesses environment variables via the `os` library's `environ` dictionary.
--- a/sqlmesh/utils/jinja.py
+++ b/sqlmesh/utils/jinja.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import importlib
 import json
 import re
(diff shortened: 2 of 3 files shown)


StuffbyYuki (approved):
@anxkhn Thanks for your work on this! 

Optional (non-blocking):

- Add a short note in `docs/guides/configuration.md` (near the env_var examples) showing `{{ env_var('BIGQUERY_KEY_B64') | b64decode }}` for base64-encoded secrets.
- Consider a test that uses env_var + b64decode together (mock/set env) to match the real-world use case.
- A brief docstring on b64decode noting it returns UTF-8 text (intended for JSON/string secrets, not arbitrary binary) would help future users.

JohnSchlerf:
This is amazing. Thank you!

PR #5864, opened 2026-06-29
fix(format): keep MODEL/AUDIT/METRIC header dialect-agnostic (#5773)

### LINKED ISSUE #5773: TSQL dialect formatting corrupts boolean true in MODEL header to (1=1)
### Description

When formatting a TSQL model using `sqlmesh format`, boolean values (`true`) in the `MODEL (...)` header are transpiled to `(1=1)` — the T-SQL representation of boolean true. Once corrupted, SQLMesh's parser cannot re-parse `(1=1)` as a boolean in the MODEL DDL context, breaking the model file.

### Steps to Reproduce

1. Create a TSQL model with a boolean property in the header, e.g.:

```sql
MODEL (
  name my_schema.my_model,
  kind FULL,
  standalone true
);

SELECT 1 AS col
```

2. Run `sqlmesh format` on the project.

3. The header is rewritten as:

```sql
MODEL (
  name my_schema.my_model,
  kind FULL,
  standalone (1=1)
);
```

4. SQLMesh can no longer parse the model file.

### Root Cause

In `sqlmesh/core/dialect.py`, `format_model_expressions` renders **all** expressions — including the `MODEL (...)` / `AUDIT (...)` header — using the target dialect:

```python
# sqlmesh/core/dialect.py ~line 781
return ";\n\n".join(
    expression.sql(pretty=True, dialect=dialect, **kwargs) for expression in expressions
).strip()
```

The `MODEL (...)` header is SQLMesh-specific DDL, not standard SQL. When `dialect="tsql"` is used, SQLGlot's TSQL generator converts `exp.Boolean(this=True)` → `(1=1)` because T-SQL has no native `TRUE` keyword.

The same issue exists in the early-return path for single-expression files (~line 750):

```python
if len(expressions) == 1 and is_meta_expression(expressions[0]):
    return expressions[0].sql(pretty=True, dialect=dialect)
```

### Affected Properties

Any boolean property set to `true` in a `MODEL (...)` or `AUDIT (...)` header, including:
- `standalone true`
- `formatting true`
- Any other boolean model/audit property

### Suggested Fix

Meta expressions (the `MODEL`/`AUDIT` header) should be rendered with `dialect=None` since they are SQLMesh-specific syntax and must not be transpiled. Only the actual SQL query expressions should use the target dialect.

```python
return ";\n\n".join(
    expression.sql(
        pretty=True,
        dialect=None if is_meta_expression(expression) else dialect,
        **kwargs
    )
    for expression in expressions
).strip()
```

And similarly for the early-return path:

```python
if len(expressions) == 1 and is_meta_expression(expressions[0]):
    return expressions[0].sql(pretty=True, dialect=None)
```

### Environment

- Dialect: `tsql`
- Command: `sqlmesh format`
- File: `sqlmesh/core/dialect.py` — `format_model_expressions`
```

anxkhn:
## Description

`sqlmesh format` corrupts boolean properties in the `MODEL (...)` header when the model dialect has no native `TRUE` keyword. On a T-SQL model, a property such as `allow_partials true` is rewritten to `allow_partials (1 = 1)`, and SQLMesh can no longer re-parse that as a boolean, so the model file breaks. Fixes #5773.

Root cause: `format_model_expressions` (`sqlmesh/core/dialect.py`) rendered *every* parsed expression with the target dialect, including the SQLMesh-specific `MODEL`/`AUDIT`/`METRIC` header. That header is SQLMesh DDL, not standard SQL, so SQLGlot's tsql generator transpiles a boolean to `(1 = 1)`.

This renders meta expressions (`is_meta_expression` -> `Audit`/`Metric`/`Model`) with `dialect=None` on both the single-meta early return and the multi-expression join path, so the header stays exactly as written, while the actual query/statement body still transpiles to the target dialect.

## Test Plan

- Added a tsql case to `tests/core/test_dialect.py::test_format_model_expressions` asserting the MODEL header keeps `allow_partials TRUE` (not `(1 = 1)`) while the query body still transpiles (`SELECT TRUE` -> `1`, `CAST(x AS INT)` -> `x::INTEGER`).
- Confirmed the new assertion fails on `main` and passes with this change.
- `pytest tests/core/test_dialect.py tests/core/test_format.py` green; `make style` (ruff + mypy) clean on both files.

`AUDIT`/`METRIC` headers share the identical `is_meta_expression` code path and behave the same; happy to add an AUDIT assertion too if that would be useful.

## Checklist

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -783,7 +783,10 @@ def format_model_expressions(
         A string representing the formatted model.
     """
     if len(expressions) == 1 and is_meta_expression(expressions[0]):
-        return expressions[0].sql(pretty=True, dialect=dialect)
+        # Meta expressions (MODEL/AUDIT/METRIC) are SQLMesh DDL, not standard SQL,
+        # so they must never be transpiled to the target dialect (e.g. tsql would
+        # rewrite a boolean property like `allow_partials TRUE` to `(1 = 1)`).
+        return expressions[0].sql(pretty=True, dialect=None)
 
     if rewrite_casts:
 
@@ -815,7 +818,14 @@ def format_model_expressions(
         expressions = new_expressions
 
     return ";\n\n".join(
-        expression.sql(pretty=True, dialect=dialect, **kwargs) for expression in expressions
+        # Meta expressions (MODEL/AUDIT/METRIC) are SQLMesh DDL and must stay
+        # dialect-agnostic; only the actual query/statement expressions transpile.
+        expression.sql(
+            pretty=True,
+            dialect=None if is_meta_expression(expression) else dialect,
+            **kwargs,
+        )
+        for expression in expressions
     ).strip()
 
 
--- a/tests/core/test_dialect.py
+++ b/tests/core/test_dialect.py
@@ -233,6 +233,60 @@ SELECT
     x = format_model_expressions(
         parse(
             """
+            MODEL(name a.b, kind FULL, dialect tsql, allow_partials true);
+            SELECT TRUE AS col, CAST(x AS INT) AS y FROM t
+            """
+        ),
+        dialect="tsql",
+    )
+    # The MODEL header is SQLMesh DDL and must not be transpiled: a boolean property
+    # such as `allow_partials true` must stay `TRUE`, not become tsql's `(1 = 1)`.
+    # The query body must still transpile to the target dialect.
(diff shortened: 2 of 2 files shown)


StuffbyYuki (approved):
@anxkhn Thanks for this PR! 

Optional: AUDIT/false boolean test cases; text_diff may deserve the same treatment in a follow-up.

PR #5870, opened 2026-06-30
Fix seed boolean null handling

### LINKED ISSUE #5589: Databricks CSV seed load coercing NULL booleans to False
Observed in sqlmesh[databricks] v0.224. 
Seed models with Boolean columns having NULL values are coerced to False in the INSERT query. The ideal behavior would be to leave NULLs as is without manipulating data. I was unable to change this behavior through csv_settings.keep_default_na or csv_settings.na_values. 

Example snippets to reproduce issue:
test_seed.csv
```
id,test_ind
1,null
2,false
3,true
4,null
``` 

test_seed.sql
```
MODEL (
  name test_lookup.test_seed
  kind SEED (
    path '$root/seeds/test_seed.csv',
    -- csv_settings (
    --   keep_default_na = false,
    --   na_values = []
    -- )
  ),
  columns (
    id string,
    test_ind boolean
  ),
  tags ( 'SEED' )
);
```

Generated INSERT Query 
```
INSERT INTO `sqlmesh_testing_dev`.`test_lookup`.test_seed
  REPLACE WHERE
    TRUE
  SELECT
    CAST(`id` AS STRING) AS `id`,
    CAST(`test_ind` AS BOOLEAN) AS test_ind
  FROM
    VALUES ('1', FALSE), ('2', FALSE), ('3', TRUE), ('4', FALSE) AS `t` (`id`, `test_ind`)
```

nkwork9999:
## Summary

Fixes #5589.

This preserves NULL values in seed boolean columns instead of coercing them to `False`.

The issue was caused by seed boolean conversion calling `str_to_bool(str(value))` for every value. Pandas reads CSV `null` values as `NaN`, so the seed renderer converted `NaN` to the string `"nan"`, and `str_to_bool("nan")` returned `False`.

## Changes

- Preserve pandas NA values as `None` when rendering seed boolean columns.
- Add a regression test for a seed CSV containing `null`, `false`, `true`, and `null` in a boolean column.

## Databricks validation

Verified on Databricks Free Edition / SQL Warehouse that Databricks preserves boolean NULL values with the expected SQL shape:

```sql
CREATE SCHEMA IF NOT EXISTS sqlmesh_smoke;

CREATE OR REPLACE TABLE sqlmesh_smoke.seed_bool_null_smoke (
  id STRING,
  test_ind BOOLEAN
);

INSERT INTO sqlmesh_smoke.seed_bool_null_smoke
SELECT
  CAST(id AS STRING) AS id,
  CAST(test_ind AS BOOLEAN) AS test_ind
FROM VALUES
  ('1', NULL),
  ('2', FALSE),
  ('3', TRUE),
  ('4', NULL) AS t(id, test_ind);

SELECT
  id,
  test_ind,
  test_ind IS NULL AS is_null
FROM sqlmesh_smoke.seed_bool_null_smoke
ORDER BY id;
```

Result:

```text
1  null   true
2  false  false
3  true   false
4  null   true
```

Also verified locally that SQLMesh now generates Databricks SQL preserving NULL values:

```sql
FROM VALUES
  ('1', CAST(NULL AS BOOLEAN)),
  ('2', FALSE),
  ('3', TRUE),
  ('4', NULL) AS t(id, test_ind)
```

## Tests

```text
uv run pytest tests/dbt/test_transformation.py::test_seed_boolean_nulls_are_preserved -q
1 passed

uv run pytest tests/dbt/test_transformation.py::test_seed_boolean_nulls_are_preserved tests/dbt/test_transformation.py::test_seed_single_whitespace_is_na tests/dbt/test_transformation.py::test_seed_partial_column_inference -q
3 passed

uv run ruff check sqlmesh/core/model/definition.py tests/dbt/test_transformation.py
All checks passed!
```


StuffbyYuki (commented):
Thanks for this PR! A few things: 
- Test lives under tests/dbt/ but fix is in core render_seed(). A native SEED model test in tests/core/ would better match scope.
- Test asserts DataFrame only, a quick check that rendered SQL contains NULL/CAST(NULL AS BOOLEAN) would lock in the reported INSERT bug.

nkwork9999:
Thanks for the review! Updated this to address both points:

- Moved the regression coverage from the dbt transformation test into a native seed model test under `tests/core/test_model.py`, directly exercising `SeedModel.render_seed()`.
- Added a Databricks rendered SQL assertion so the generated seed SQL now explicitly locks in `CAST(NULL AS BOOLEAN)` / `NULL` handling for the VALUES insert shape.

Locally re-ran:

```text
ruff check tests/core/test_model.py tests/dbt/test_transformation.py sqlmesh/core/model/definition.py
pytest tests/core/test_model.py::test_seed_model_boolean_nulls_are_preserved tests/dbt/test_transformation.py::test_seed_single_whitespace_is_na tests/dbt/test_transformation.py::test_seed_partial_column_inference -q -o cache_dir=/private/tmp/sqlmesh_pytest_cache
```

Both passed.

PR #5882, opened 2026-07-05
fix(cicd): chunk bot messages on character boundaries

anxkhn:
`GithubController._chunk_up_api_message` splits a message into chunks that fit
GitHub's per-field byte limit (`MAX_BYTE_LENGTH`, 65535). It did this by UTF-8
encoding the whole message, slicing the resulting bytes at fixed
`MAX_BYTE_LENGTH` offsets, and decoding each slice with `errors="ignore"`:

```python
message_encoded = message.encode("utf-8")
return [
    message_encoded[i : i + self.MAX_BYTE_LENGTH].decode("utf-8", "ignore")
    for i in range(0, len(message_encoded), self.MAX_BYTE_LENGTH)
]
```

When a multibyte character (accented letter, CJK, emoji, etc.) straddles a chunk
boundary, its UTF-8 bytes are split across two slices. `decode("utf-8", "ignore")`
then discards the incomplete byte sequence on both sides, so that character is
silently dropped from the bot's PR comment or check-run output.

This change accumulates characters into a chunk while the running UTF-8 byte
length stays within `MAX_BYTE_LENGTH`, so a boundary is only ever placed between
whole characters and no character is bisected. Behavior for the common case is
unchanged: ASCII-only messages chunk into the exact same pieces as before, and an
empty message still yields no chunks. Both existing callers rely only on the
number of chunks (comment edit unpacks `body, *truncated`; check-run output
unpacks `summary, text, *truncated`), and that contract is preserved.

## Test Plan

- Added `test_chunk_up_api_message_preserves_multibyte_chars`, which builds a
  message just over `MAX_BYTE_LENGTH` with a multibyte character (`é`) placed on
  the byte boundary between two chunks, and asserts the chunks reassemble into the
  original message and each chunk stays within `MAX_BYTE_LENGTH` bytes. This test
  fails on the previous byte-slicing implementation (the boundary character is
  dropped) and passes with the fix.
- `pytest tests/integrations/github/cicd/ -m "not slow and not docker"` -> 55
  passed, 28 deselected.
- `make style` (ruff check + ruff format) clean on the changed files.

## Checklist

### DIFF
--- a/sqlmesh/integrations/github/cicd/controller.py
+++ b/sqlmesh/integrations/github/cicd/controller.py
@@ -1182,13 +1182,26 @@ class GithubController:
 
     def _chunk_up_api_message(self, message: str) -> t.List[str]:
         """
-        Chunks up the message into `MAX_BYTE_LENGTH` byte chunks
+        Chunks up the message into chunks of at most `MAX_BYTE_LENGTH` bytes when
+        UTF-8 encoded.
+
+        Chunk boundaries are placed between characters rather than raw bytes so a
+        multibyte character that lands on a boundary is never split and dropped.
         """
-        message_encoded = message.encode("utf-8")
-        return [
-            message_encoded[i : i + self.MAX_BYTE_LENGTH].decode("utf-8", "ignore")
-            for i in range(0, len(message_encoded), self.MAX_BYTE_LENGTH)
-        ]
+        chunks: t.List[str] = []
+        current = ""
+        current_length = 0
+        for char in message:
+            char_length = len(char.encode("utf-8"))
+            if current and current_length + char_length > self.MAX_BYTE_LENGTH:
+                chunks.append(current)
+                current = ""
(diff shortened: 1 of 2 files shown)


StuffbyYuki (approved):
@anxkhn Thanks for the PR! Not a blocker but here's a few things:
- Add a small regression test for empty string / pure-ASCII chunking to lock in the “unchanged behavior” claim.
- Consider boundary cases with emoji/CJK, not just é.

StuffbyYuki:
@anxkhn Following up on my prev message!

StuffbyYuki:
@anxkhn Another follow up

anxkhn:
@stuffbyyuki thanks for the follow-up. i added focused regression coverage for empty and exact-boundary ascii chunking, plus emoji and cjk characters crossing the byte boundary. the implementation did not need further changes. could you please take another look?

PR #5885, opened 2026-07-07
fix(date): import pandas locally in make_ts_exclusive

anxkhn:
`make_ts_exclusive` uses `pd.Timedelta` in its `dialect == "tsql"` branch, but
`pandas` is only imported under `TYPE_CHECKING` in `sqlmesh/utils/date.py` (it is
banned as a module-level import by the ruff `flake8-tidy-imports` config). Unlike
the sibling helpers `make_inclusive_end` and `to_utc_timestamp`, which do
`import pandas as pd` at runtime, `make_ts_exclusive` never imports pandas, so:

```python
from sqlmesh.utils.date import make_ts_exclusive
make_ts_exclusive("2020-01-01 12:00:00", dialect="tsql")
# NameError: name 'pd' is not defined
```

This is reachable in normal use: rendering an `INCREMENTAL_BY_TIME_RANGE` model
that has a jinja source/ref event-time filter on the `tsql` dialect calls
`make_ts_exclusive(..., dialect="tsql")` in `sqlmesh/core/renderer.py`. The
`NameError` is caught there and re-raised as a `ConfigError` reporting that the
model could not be rendered.

The fix adds a function-local `import pandas as pd` as the first line of
`make_ts_exclusive`, mirroring the existing runtime-pandas helpers in the same
module. The non-tsql path is unaffected (it did not reference `pd`).

## Test Plan

- Added `test_make_ts_exclusive` in `tests/utils/test_date.py` covering both
  paths: the `tsql` dialect returns `to_utc_timestamp(ts) - 1ns` (and no longer
  raises `NameError`), and a non-tsql dialect (`duckdb`) returns `ts + 1us`.
- Verified the test is a true regression test: with the runtime import removed it
  fails with `NameError: name 'pd' is not defined`; with the fix it passes.
- `pytest tests/utils/test_date.py` -> 61 passed.
- `pytest tests/utils/ -m "not slow and not docker"` -> 236 passed, 1 skipped.
- `ruff check` + `ruff format --check` + `mypy` on the changed files -> clean.

## Checklist

### DIFF
--- a/sqlmesh/utils/date.py
+++ b/sqlmesh/utils/date.py
@@ -344,6 +344,8 @@ def make_exclusive(time: TimeLike) -> datetime:
 
 
 def make_ts_exclusive(time: TimeLike, dialect: DialectType) -> datetime:
+    import pandas as pd
+
     ts = to_datetime(time)
     if dialect == "tsql":
         return to_utc_timestamp(ts) - pd.Timedelta(1, unit="ns")
--- a/tests/utils/test_date.py
+++ b/tests/utils/test_date.py
@@ -14,11 +14,13 @@ from sqlmesh.utils.date import (
     is_categorical_relative_expression,
     is_relative,
     make_inclusive,
+    make_ts_exclusive,
     to_datetime,
     to_time_column,
     to_timestamp,
     to_ts,
     to_tstz,
+    to_utc_timestamp,
 )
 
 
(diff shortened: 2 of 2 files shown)


anxkhn:
The only failing check here is `style-and-cicd-tests (3.9)`, and the single failure is unrelated to this change:

```
FAILED tests/core/test_test.py::test_freeze_time_concurrent - sqlmesh.utils.errors.TestError: Failed to create test test_2019_py_model_2 (...) list index out of range
```

`test_freeze_time_concurrent` runs the model tests concurrently (`concurrent_tasks=8`), so this looks like an intermittent concurrency race rather than anything this PR touches. Supporting evidence:

- This PR only adds a function-local `import pandas as pd` inside `make_ts_exclusive` (`sqlmesh/utils/date.py`) plus a unit test in `tests/utils/test_date.py`; it does not touch the model-testing code path.
- The same job passes on Python 3.10, 3.11, 3.12 and 3.13, on the Windows job, and across all `engine-tests-docker` (including `mssql`, the T-SQL engine this fix concerns).
- `tests/core/test_test.py::test_freeze_time_concurrent` passes consistently for me locally.

Could a maintainer re-run the `style-and-cicd-tests (3.9)` job? I don't have permission to re-run CI on this repo. Happy to rebase if that's preferred.

PR #5892, opened 2026-07-08
fix: allow an explicit --execution-time to extend the plan end past the prod frontier

### LINKED ISSUE #5640: `--execution-time` flag is ignored in `sqlmesh plan` only in prod (not in dev)
**Actual behavior:**

Given a model

```
MODEL (
    name sqlmesh_example.demo,
        time_column day,
        batch_size 1
    ),
    start '2025-12-20'
); 
SELECT @start_ds AS day;
```

When running `sqlmesh plan --execution-time '2025-12-26'`, the intervals between the model's start date and the execution time are filled (in this case 2025-12-20 until 2025-12-25) are created as expected.

However, subsequently running `sqlmesh plan --execution-time '2025-12-28'`, sqlmesh does nothing, just stating:

> No changes to plan: project files match the prod environment

This happens **only in prod**. If I do the same in dev environments, then the intervals 2025-12-26 and 2025-12-27 _are_ filled as expected.

**Expected behavior:**

Since I explicitly pass `--execution-time '2025-12-28'`, I expect the intervals 2025-12-26 and 2025-12-27 to be filled.

If this is somehow intended behavior, then at least I expect:

* dev and prod to behave the same way
* the documentation should mention this gotcha

**Self-contained reproduction script:**

```
# create temp dir for sqlmesh project
cd $(mktemp -d)
python -m venv .venv
source .venv/bin/activate
pip install sqlmesh

# create empty sqlmesh project with all defaults (duckdb)
yes 1 | sqlmesh init -t EMPTY

# create model
echo "\
MODEL (
    name sqlmesh_example.demo,
        time_column day,
        batch_size 1
    ),
    start '2025-12-20'
); 
SELECT @start_ds AS day;
" > models/demo.sql

echo "Running in dev with --execution-time '2025-12-22'"
sqlmesh plan dev --execution-time '2025-12-22' --auto-apply

echo "Looking at the data:"
sqlmesh fetchdf "select * from sqlmesh_example__dev.demo"
# Note: this prints (as expected):
#           day
# 0  2025-12-20
# 1  2025-12-21

echo "Running in dev with --execution-time '2025-12-24'"
sqlmesh plan dev --execution-time '2025-12-24' --auto-apply

echo "Looking at the data:"
sqlmesh fetchdf "select * from sqlmesh_example__dev.demo"
# Note: this prints (as expected):
#           day
# 0  2025-12-20
# 1  2025-12-21
# 0  2025-12-22
# 1  2025-12-23

echo "Running in prod with --execution-time '2025-12-26'"
sqlmesh plan --execution-time '2025-12-26' --auto-apply

echo "Looking at the data:"
sqlmesh fetchdf "select * from sqlmesh_example.demo"
# Note: this prints (as expected):
#           day
# 0  2025-12-20
# 1  2025-12-21
# 0  2025-12-22
# 1  2025-12-23
# 0  2025-12-24
# 1  2025-12-25

echo "Running in prod with --execution-time '2025-12-28'. Should update the model (but doesn't)!"
sqlmesh plan --execution-time '2025-12-28' --auto-apply

echo "Note that the above incorrectly prints:"
echo "  No changes to plan: project files match the prod environment"
echo ""
echo "Notice the model still shows the only the days 20 to 25:"

sqlmesh fetchdf "select * from sqlmesh_example.demo"
```

nnqtruong:
## Description

**Why.** `--execution-time` tells a plan "pretend it is now X". Today that only works in one direction: if prod's recorded intervals are *newer* than X, the plan end is clamped down to X — but if X is *beyond* the last interval recorded in prod, the plan end stays capped at the prod frontier and the flag is silently ignored. That's the behavior reported in #5640: with no model changes, `sqlmesh plan --execution-time '2025-12-28' --auto-apply` reports "No changes to plan" and fills nothing, even though 2025-12-26/27 are due. The confusing part for users is that the same command *appears* to work in dev and on a first deploy — only because no prod frontier exists yet to cap against — and that `plan --run` (which skips the caps) fills the intervals fine.

**What.** In `Context.plan_builder()`, when `execution_time` is explicitly provided, no explicit `end` is given, and the execution time is past the computed default end (the prod frontier), the default end and the per-model interval-end caps (`end_override_per_model`) are raised up to the execution time. This mirrors the existing precedent for explicit `--end` (which drops those caps entirely via `PlanBuilder.override_end`), and is bounded by the already-correct `plan --run` behavior: raising caps toward the explicit execution time cannot schedule anything the cap-free `--run` path wouldn't. No behavior changes when `--execution-time` isn't passed, when `--end` is given, on first deploys, or when the execution time is behind the frontier (the downward clamp is preserved).

Closes #5640

## Test Plan

- New: `test_plan_execution_time_ahead_of_prod_frontier` — the issue scenario end-to-end on the sushi example (backfill required, plan end = execution time, intervals advance after apply; downward clamp still asserted).
- New: `test_plan_execution_time_ahead_of_prod_frontier_matches_run_for_all_models` — a daily+weekly project with different per-model frontiers; asserts plain `plan --execution-time X` reports byte-identical missing intervals to `plan --run --execution-time X`.
- New: `test_plan_execution_time_ahead_of_prod_frontier_with_explicit_end` — explicit `--end` + far-future execution time: backfill stops at the explicit end.
- Updated: `test_plan_min_intervals` — raw `plan.end` now equals the explicit execution time (1s past the old day-aligned frontier); comment explains why no interval computation changes.
- The new first test fails without the fix (`assert plan.requires_backfill` → `False`).
- Ran: targeted 9-test set, `tests/core/test_context.py` and `tests/core/test_plan.py` (`-m "not slow and not docker"`), the `make fast-test` fast component (2544 passed; 6 failures are local-only `pyodbc`/unixODBC driver import errors in `test_connection_config.py`, a module this PR doesn't touch), and the CLI repro from the issue before/after.

## Checklist

### DIFF
--- a/docs/concepts/plans.md
+++ b/docs/concepts/plans.md
@@ -109,6 +109,12 @@ For context, every model has a start date. The start can be specified in [the mo
 
 Because the prod environment supports business operations, prod plans ensure every model is backfilled from its start date until the most recent completed time interval. Due to that restriction, the `plan` command's `--start` and `--end` options are not supported for regular plans against prod. The options are supported for [restatement plans](#restatement-plans) against prod to allow re-processing a subset of existing data.
 
+!!! note "Explicit execution time"
+
+    "The most recent completed time interval" is measured relative to the plan's *execution time*, which defaults to now. If you pass an explicit `--execution-time` that is later than the intervals already loaded in prod, the plan extends its end date up to that time and backfills the intervals in between.
+
+    For example, if a daily model in prod is loaded through 2025-12-25, running `sqlmesh plan --execution-time '2025-12-28'` backfills the missing 2025-12-26 and 2025-12-27 intervals.
+
 Non-prod plans are typically used for development, so their models can optionally be backfilled for any date range with the `--start` and `--end` options. Limiting the date range makes backfills faster and development more efficient, especially for incremental models using large tables.
 
 #### Model kind limitations
--- a/sqlmesh/core/context.py
+++ b/sqlmesh/core/context.py
@@ -1728,21 +1728,11 @@ class GenericContext(BaseContext, t.Generic[C]):
                 and default_end is not None
                 and execution_time_ts > default_end
             ):
-                # An explicitly provided execution time acts as the plan's effective "now", so the
-                # default end is allowed to extend past the recorded prod frontier instead of being
-                # capped by it. This mirrors how an explicit `end` causes the per-model interval end
-                # caps to be dropped entirely in PlanBuilder.build() (see `self.override_end`,
-                # sqlmesh/core/plan/builder.py line 206). Raising (rather than dropping) the caps is
-                # safe here because `plan --run` already runs with no caps at all.
-                #
-                # Note this raises every entry already present in max_interval_end_per_model (i.e.
-                # every model in this dict, which _get_max_interval_end_per_model above has already
-                # scoped down to backfill_models/its ancestors when a selection is in effect) - not
-                # just modified or explicitly selected models within that scope. That's intentional,
-                # not an oversight. It's what makes a plain, unscoped `sqlmesh plan --execution-time X`
-                # in prod report the same missing intervals as `sqlmesh plan --run --execution-time X`
-                # would at the same simulated time. Narrowing this further would reintroduce a `plan`
-                # vs `plan --run` mismatch for models within that scope.
+                # An explicit execution time is the plan's effective "now", so the default end may
+                # extend past the recorded prod frontier (as an explicit `end` already does via
+                # PlanBuilder.override_end). Raising every per-model cap to it keeps a plain
+                # `plan --execution-time X` in step with `plan --run --execution-time X`, which
+                # already runs with no caps.
                 default_end = execution_time_ts
                 execution_time_dt = to_datetime(execution_time_ts)
                 max_interval_end_per_model = {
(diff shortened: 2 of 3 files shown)


StuffbyYuki:
@nnqtruong Thanks for your work on this PR!

No blockers, but one thing, it might be worth a short docs note that explicit `--execution-time` past the prod frontier extends the plan end / fills intervals (the gotcha called out in #5640). 

Let me know what you think!

nnqtruong:
@StuffbyYuki Thanks appreciate the review! Great call — that's exactly the gap the issue flagged. I've added a short note to the Start and end dates section of docs/concepts/plans.md: it explains that an explicit --execution-time past the intervals already loaded in prod extends the plan's end and backfills the intervening intervals, with a small example. Happy to adjust the wording or move it if you'd prefer it somewhere else.

StuffbyYuki:
@nnqtruong Thanks for the PR! It feels the inline comment in `plan_builder()` is quite long (and cites a hard-coded `builder.py` line number that will drift). Do you think it's worth shortening to the “why” (explicit execution time = effective now, raise caps for `--run` parity) and drop the line reference

nnqtruong:
@StuffbyYuki Agreed on both counts! the block was doing too much, and that line number would definitely drift. I've trimmed it to the rationale (an explicit execution time is the plan's effective "now", so the per-model caps get raised to it to keep plan in step with plan --run) and dropped the builder.py line reference. It is now just names PlanBuilder.override_end. Thanks for the catch!

PR #5893, opened 2026-07-09
fix: adding in run_id injection

cmgoffena13:
## Description

Adding in run_id injection to the run command, similar to how plan_id is injected into queries. Identifies SQLMesh triggered queries.

## Test Plan

Added in test similar to the plan_id injection test ("test_plan_evaluator_correlation_id")

## Checklist


mday-io:
Looks good to me. The new `ctx.run()` integration test covers the main SQL comment behavior.

Optional follow-up: add BigQuery-specific coverage too, since BigQuery also exposes correlation IDs as job labels and there is already similar coverage for plan_id labels.

cmgoffena13:
@mday-io good call out, I added in another test for the BigQuery integration specifically for the run_id to show up in the labels

PR #5894, opened 2026-07-09
feat: Add Development Container and Windows specific development pre-requisites

davem-bis:
## Description

Adds a development container for those working on Windows as the repository relies on Python functionality only available on UNIX systems.

Expands the `development.md` documentation to include caveats with developing on Windows.

## Test Plan

Dev container was built from scratch and `make fast-test` run successfully.

## Checklist

- ~[ ] I have added tests for my changes (if applicable)~ Not Applicable


StuffbyYuki:
@davem-bis Thanks for this PR! 

**Blocker**
- `post-create-command.sh` installs `msodbcsql18` without adding the Microsoft apt repo. That package isn’t in Ubuntu’s default repos (see `.circleci/install-prerequisites.sh` fabric path), so a clean create should fail here. Easiest fix: drop `msodbcsql18` from the default install (it’s only needed for Fabric/ODBC), or you can also restore the Microsoft repo setup.

**Optional**
- Add `set -e` so apt failures don’t continue into `make install-dev`.
- Docs: how to open the container, and that Windows host clones need `core.symlinks=true` before bind-mount.
- Soften the `SIGUSR1` rationale, that path is already gated via `hasattr(faulthandler, "register")`; symlinks are the clearer Windows caveat.

davem-bis:
Hi @StuffbyYuki, thanks for the review.

Regarding the `msodbcsql18` install - because this is using a Microsoft hosted version of the Ubuntu container, this already has the Microsoft apt repository available by default.

I'll implement the rest though. Regarding the SIGUSR1 rational, one of the tests fails on that import on Windows - I'll try to track down which one it is and see if I can tweak it.

davem-bis:
Hi @StuffbyYuki - sorry for the delay. Those changes have been implemented.

StuffbyYuki:
Thanks! The `set -e` and docs updates look good.

On `msodbcsql18`: I think this is mixing up two different “Microsoft-hosted Ubuntu” setups. GitHub Actions / Azure DevOps agents often already have the Microsoft apt repo, but this PR uses `mcr.microsoft.com/devcontainers/base:ubuntu-24.04`, a Docker image on MCR, not a runner. MCR hosting doesn’t mean the repo is configured inside the image.

Microsoft’s Ubuntu install steps still require adding `packages-microsoft-prod.deb` first ([docs](https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server?view=sql-server-ver17&tabs=ubuntu18-install)). The image is a plain Ubuntu base ([Dockerfile](https://github.com/devcontainers/images/blob/main/src/base-ubuntu/.devcontainer/Dockerfile)); I checked it and there’s no `packages.microsoft.com` source, so `msodbcsql18` isn’t available.

If you agree, let's either drop `msodbcsql18` from the default install (I think only needed for Fabric/ODBC), or add the Microsoft repo setup like `.circleci/install-prerequisites.sh` does for fabric.

davem-bis:
Hi @StuffbyYuki - I had a deeper dig into this, I was mistaken before, the Microsoft package repository was being added by the Java and Docker-in-Docker features, not the base docker image.

I've added an extra step to the install script to add the repository if it's not available, in case those features are removed from the development container configuration in the future.

PR #5897, opened 2026-07-10
Fix duplicate external model generation from directory files

### LINKED ISSUE #5880: Bug: create_external_models causes duplicate model ConfigError when using external_models/ directory
## Steps to reproduce

1. Create a SQLMesh project with a model referencing an external table:
```sql
MODEL (name db.my_model, kind FULL);
SELECT * FROM raw.source_table;
```

2. Run `sqlmesh create_external_models` — the external model is written to `external_models.yaml`.

3. Move the entry into the `external_models/` directory as recommended in the docs (e.g., `external_models/source_table.yaml`), then delete `external_models.yaml`.

4. Run `sqlmesh create_external_models` again — the entry is written back into `external_models.yaml`. The same model now exists in both `external_models.yaml` and `external_models/source_table.yaml`.

5. Run `sqlmesh plan` or any command that loads models — crashes with:

```
ConfigError: Failed to load model from file external_models/source_table.yaml:
  Duplicate external model name: memory.raw.source_table.
```

## Why this happens

`create_external_models_file` writes all fetched external models into `external_models.yaml`, regardless of whether they already exist in `external_models/*.yaml` files. The loader then raises a `ConfigError` when it finds the same model name in two different files.

## Workaround

Delete `external_models.yaml` and re-run `create_external_models`, or avoid using the `external_models/` directory together with `create_external_models`.

reachsrinivaschennupati-tech:
## Summary

Fixes duplicate external model generation when external models are already defined in the `external_models` directory.

## Problem

`create_external_models_file` considered every external model in the loaded model set, including models that originated from the configured external models directory. That could cause SQLMesh to regenerate an external model entry that was already explicitly defined by the project.

## Solution

- Only include external models whose source path matches the generated external models file path.
- Preserve externally defined models from the `external_models` directory without writing duplicate generated entries.
- Add regression coverage for a project that contains both a SQL model dependency and an existing external model YAML definition.

## Validation

- Local tests passed.

Fixes #5880

### DIFF
--- a/sqlmesh/core/schema_loader.py
+++ b/sqlmesh/core/schema_loader.py
@@ -45,7 +45,7 @@ def create_external_models_file(
     external_model_fqns = set()
 
     for fqn, model in models.items():
-        if model.kind.is_external:
+        if model.kind.is_external and getattr(model, "_path", None) == path:
             external_model_fqns.add(fqn)
         for dep in model.depends_on:
             if dep not in known_models:
--- a/sqlmesh/core/schema_loader.py
+++ b/sqlmesh/core/schema_loader.py
@@ -45,7 +45,7 @@ def create_external_models_file(
     external_model_fqns = set()
 
     for fqn, model in models.items():
-        if model.kind.is_external and getattr(model, "_path", None) == path:
+        if model.kind.is_external and model._path == path:
             external_model_fqns.add(fqn)
         for dep in model.depends_on:
             if dep not in known_models:
--- a/tests/core/test_schema_loader.py
+++ b/tests/core/test_schema_loader.py
@@ -305,6 +305,53 @@ def test_create_external_models_no_duplicates(tmp_path: Path):
     assert len(_load_external_models()) == 1
 
 
+def test_create_external_models_skips_models_from_external_models_directory(tmp_path: Path):
+    config = Config(gateways={"": GatewayConfig(connection=DuckDBConnectionConfig())})
+
+    model_dir = tmp_path / c.MODELS
+    model_dir.mkdir()
+
+    with open(model_dir / "table.sql", "w", encoding="utf8") as fd:
+        fd.write(
+            """
+        MODEL (
+            name lake.table,
+            kind FULL,
+        );
+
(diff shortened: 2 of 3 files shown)


reachsrinivaschennupati-tech:
Hi maintainers, this PR is ready for review when you have a chance.

I kept the change focused on preventing generated external model entries from duplicating models that are already explicitly defined under the external models directory. The PR includes regression coverage for the mixed case where a project has both a SQL model dependency and an existing external model YAML definition.

Validation noted in the PR:
- Local tests passed.

Happy to make any adjustments if you prefer a different boundary for the generated-vs-explicit external model handling.

StuffbyYuki:
@reachsrinivaschennupati-tech Thanks for this PR! One consideration - it might be preferable to have `model._path == path` over `getattr(model, "_path", None) == path` — `_path` is a normal model attribute elsewhere. `getattr` is likely unnecessary.

reachsrinivaschennupati-tech:
@StuffbyYuki  Thank you for your kind review and suggestion. I agree `_path` is a normal model attribute here, so the extra `getattr` fallback is unnecessary. I updated the condition to use `model._path == path` directly.

Please review.

StuffbyYuki:
Can you add DCO checks?

reachsrinivaschennupati-tech:
@StuffbyYuki Thanks for pointing that out. I added the DCO sign-off to the follow-up commit as well.

StuffbyYuki:
@reachsrinivaschennupati-tech Hey some of the style related ci tests didn't pass. Could you check?

reachsrinivaschennupati-tech:
@StuffbyYuki I checked the failing style-and-cicd job and it was failing in the Jupyter create_external_models test. I updated the test expectation to match the new behavior where external models already registered under the external_models directory are skipped when rebuilding the root external_models.yaml file.\n\nLocal focused checks passed:\n\n- uv run pytest tests/integrations/jupyter/test_magics.py::test_create_external_models tests/core/test_schema_loader.py::test_create_external_models_skips_models_from_external_models_directory\n\nPlease review when you have a chance.

PR #5899, opened 2026-07-14
fix(migrations): keep still-live dev intervals in v0071

anxkhn:
**Note:** This only helps installs that have **not** already run v0071. Users who already migrated still have the dropped intervals and need a re-backfill / manual repair. There is no recovery path in this PR.

The v0071 intervals migration drops every still-live dev interval on upgrade
because of a type mismatch in a set-membership check.

In `_migrate_intervals`, `used_dev_versions` is a `Set[Tuple[str, str]]` and its
only insertion adds a `(name, dev_version)` 2-tuple:

```python
used_dev_versions.add((name, dev_version))
```

But the guard that decides whether to drop a dev interval compares a bare
`dev_version` string against that set:

```python
dev_version = snapshot_ids_to_dev_versions.get((name, identifier))
if dev_version not in used_dev_versions and is_dev:
    # If the interval's dev version is no longer used and this is a dev interval, we can safely delete it
    continue
```

A `str` is never an element of a set of 2-tuples, so `dev_version not in
used_dev_versions` is always `True` and the condition collapses to
`if is_dev: continue`. Every dev interval that had already passed the preceding
`used_versions` check (i.e. is still live) is discarded, forcing a re-backfill of
dev environments after the upgrade.

This changes the check to compare the correctly-typed `(name, dev_version)` key,
matching the tuples stored in `used_dev_versions`, so only genuinely unused dev
intervals are removed.

## Test Plan

- Added `test_migrate_v0071_keeps_still_live_dev_intervals` in
  `tests/core/state_sync/test_state_sync.py`. It builds an in-memory DuckDB state
  at schema v70 (intervals table without the `dev_version` column) with a live
  prod snapshot/interval and a live dev snapshot/interval, runs the v0071
  `migrate_schemas` + `migrate_rows`, and asserts both intervals survive.
- Confirmed the test fails without the fix (the dev interval is dropped:
  `{'prod_int'} != {'dev_int', 'prod_int'}`) and passes with it.
- Also confirmed a genuinely-stale dev interval (its dev version no longer mapped
  to a live snapshot) is still correctly removed, so the fix is not over-broad.
- `make style` (ruff + ruff-format + mypy) passes on the changed files; the
  `tests/core/state_sync/test_state_sync.py` migration/interval subset passes.

## Checklist

### DIFF
--- a/sqlmesh/migrations/v0071_add_dev_version_to_intervals.py
+++ b/sqlmesh/migrations/v0071_add_dev_version_to_intervals.py
@@ -106,7 +106,7 @@ def _migrate_intervals(
             continue
 
         dev_version = snapshot_ids_to_dev_versions.get((name, identifier))
-        if dev_version not in used_dev_versions and is_dev:
+        if (name, dev_version) not in used_dev_versions and is_dev:
             # If the interval's dev version is no longer used and this is a dev interval, we can safely delete it
             continue
 
--- a/tests/core/state_sync/test_state_sync.py
+++ b/tests/core/state_sync/test_state_sync.py
@@ -2937,6 +2937,141 @@ def test_migrate_rows(state_sync: EngineAdapterStateSync, mocker: MockerFixture)
     )
 
 
+def test_migrate_v0071_keeps_still_live_dev_intervals() -> None:
+    from sqlmesh.migrations import v0071_add_dev_version_to_intervals as migration
+
+    engine_adapter = create_engine_adapter(duckdb.connect, "duckdb")
+    engine_adapter.create_schema(c.SQLMESH)
+
+    engine_adapter.create_table(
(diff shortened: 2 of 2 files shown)


StuffbyYuki:
@anxkhn Thanks for the PR! Do you think it's good to call out in the PR / release notes that this only helps installs that have **not** already run v0071, users who already migrated already lost those intervals and still need a re-backfill (no recovery path in this PR).

StuffbyYuki:
@anxkhn Wanted to follow up on my prev comment!

StuffbyYuki:
@anxkhn Bump

anxkhn:
good call. called that out in the pr description. this only helps installs that have not already run v0071; already-migrated dbs still need a manual repair.

PR #5913, opened 2026-07-23
Perf: add opt-in project-index loading for lint

### LINKED ISSUE #5907: Feature Request:  `sqlmesh lint` loads the full project context even when scoped to specific models with `--model`
### Summary

`sqlmesh lint --model <mymodel>` loads the entire project context, parses all model files, resolves all schemas, and validates all model definitions before a single lint rule runs. `--model` only scopes the *linting loop* — not the context initialization that precedes it. Cost scales with total project size, not with the number of models being linted.

### Where this happens

**1. `lint` always triggers a full context load**

`lint` is not in `SKIP_LOAD_COMMANDS`, so `GenericContext.__init__` always calls `self.load()` before any linting happens:
`sqlmesh/cli/main.py` ~L31–42:
```python
SKIP_LOAD_COMMANDS = (
    "clean", "create_external_models", "destroy", "environments",
    "invalidate", "janitor", "migrate", "rollback", "run", "table_name",
)
# "lint" is not here
```
`sqlmesh/core/context.py` ~L473:
```python
if load:
    self.load()
```

**2. `load()` parses and validates the entire project**

`load()` parses every model, macro, audit, test, and metric, then runs schema resolution and definition validation on every model:
`sqlmesh/core/context.py` ~L719–728:
```python
update_model_schemas(
    self.dag,
    models=self._models,
    cache_dir=self.cache_dir,
)
models = self.models.values()
for model in models:
    model.validate_definition()
```
`update_model_schemas()` traverses the full project DAG to propagate column types/schema info — needed for rules like `ambiguousorinvalidcolumn`, but run for all models regardless of what's being linted.

**3. `lint_models()` scopes the loop, but not the context**

The CLI `lint` command calls `obj.lint_models(models)` on the already fully-loaded context. The loop itself is scoped:
`sqlmesh/core/context.py` ~L3209–3219:
```python
model_list = (
    list(self.get_model(model, raise_if_missing=True) for model in models)
    if models
    else self.models.values()
)
for model in model_list:
    if linter := self._linters.get(model.project):
        lint_violation, violations = (
            linter.lint_model(model, self, console=self.console) or found_error
        )
```
...but the full context (`self`) is passed into every rule invocation.

**4. Every rule receives the full context**

`RuleSet.check_model()` instantiates each rule with the full context, regardless of whether the rule needs it:
`sqlmesh/core/linter/definition.py` ~L106–116:
```python
def check_model(self, model: Model, context: GenericContext) -> t.List[RuleViolation]:
    for rule in self._underlying.values():
        violation = rule(context).check_model(model)
        ...
```

**5. Some rules genuinely need cross-model context — but not the whole project**

- `NoMissingUnitTest` checks `self.context.models_with_tests` (full test index):
  `sqlmesh/core/linter/rules/builtin.py` ~L140.
- `NoMissingExternalModels` calls `self.context.get_model(depends_on_model)` per dependency:
  `sqlmesh/core/linter/rules/builtin.py` ~L160–163.

Both only need the target model's *dependencies*, not the entire project.

**6. Single-model rules pay the same cost as cross-model rules**

Rules that only inspect the model itself (`NoSelectStar`, `NoMissingAudits`, `NoAmbiguousProjections`) still receive the full context even though they never use it — there's no mechanism to distinguish rules that need cross-model context from those that don't.

### What `--model` actually does (and doesn't do)

It limits which models are iterated in the linting loop. It does **not** prevent: loading all model files from disk, running `update_model_schemas()` on the full project DAG, calling `model.validate_definition()` on all models, or passing the full context to every rule.

### Impact

- **Large projects**: `sqlmesh lint --model my_model` in a project with thousands of models still parses every file and resolves every schema. Overhead scales with project size, not selection size.
- **Iteration loop**: `sqlmesh lint` is meant to be faster than `sqlmesh plan` for quick iteration; a full context load erodes that advantage.
- **Schema resolution is full-project**: `update_model_schemas()` traverses the entire DAG even for single-model rules like `ambiguousorinvalidcolumn`, when only the target model's upstream subgraph is needed.
- **Cross-model rules need only a bounded subgraph**: e.g. `NoMissingExternalModels` only needs the target model's direct dependencies, not the full model registry.

### Suggested direction

1. **Lazy/partial context loading for lint** — when `--model` is specified, load only the target model(s) and their transitive upstream dependencies, deferring parsing of unrelated models.
2. **Scoped schema resolution** — allow `update_model_schemas()` to run on a subgraph instead of always the full project DAG.
3. **Rule context interface** — distinguish rules that need full cross-model context from those that only need the model itself, so the latter can run without a full context load.
4. **Scoped cross-model rules** — satisfy rules like `NoMissingExternalModels`/`NoMissingUnitTest` with a partial context bounded to the target model's dependencies/tests, rather than the full project load.

### References

- CLI `lint` command and `SKIP_LOAD_COMMANDS`: `sqlmesh/cli/main.py` ~L31–42, ~L1168–1183
- Unconditional `self.load()`: `sqlmesh/core/context.py` ~L473–474
- `load()` full parse, schema update, validation: `sqlmesh/core/context.py` ~L629–728
- `lint_models()` scoped loop, full context passed: `sqlmesh/core/context.py` ~L3209–3230
- `RuleSet.check_model()` full context injected into every rule: `sqlmesh/core/linter/definition.py` ~L106–116
- `NoMissingUnitTest` uses `context.models_with_tests`: `sqlmesh/core/linter/rules/builtin.py` ~L140
- `NoMissingExternalModels` uses `context.get_model()` per dependency: `sqlmesh/core/linter/rules/builtin.py` ~L160–163

fresioAS:
## Description
Speeds up specific model linting. General use case would be for pre-commit - to not have to loop through the full project for specific models.

Add --use-project-index to targeted and full lint. The default path preserves the existing full context load. With both --use-project-index and --model, SQLMesh uses a persistent model-to-file dependency index to load, resolve, and validate only the selected models and their transitive upstream dependencies.

A full indexed lint still loads and lints every model, but creates or refreshes the index for later targeted commands. Missing or expanded dependencies safely retry with a full indexed load.

To mitigate long lint runs - https://github.com/SQLMesh/sqlmesh/issues/5907

## Test Plan

Tested on live project (395 models)

=== Timing summary ===
sqlmesh lint --local                                                     23.46s  ok
sqlmesh lint --local --model sap.exposed.project                         18.31s  ok
sqlmesh lint --local --model sap.exposed.person                          16.59s  ok
sqlmesh lint --local --model sap.exposed.calendar                        16.59s  ok
sqlmesh lint --local --use-project-index                                 19.92s  ok
sqlmesh lint --local --use-project-index --model sap.exposed.project      3.68s  ok
sqlmesh lint --local --use-project-index --model sap.exposed.person       4.20s  ok
sqlmesh lint --local --use-project-index --model sap.exposed.calendar     3.97s  ok

## Checklist

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -329,6 +329,16 @@ By default, the SQLMesh cache is stored in a `.cache` directory within your proj
 
 The cache directory is automatically created if it doesn't exist. You can clear the cache using the `sqlmesh clean` command.
 
+#### Project index
+
+The `--use-project-index` option on supported commands maintains a persistent model dependency index in the cache directory. Each project writes a file named `<project>_<hash>_model_index.json`.
+
+A full project load with the option enabled creates or refreshes the index. SQLMesh invalidates it when relevant configuration, gateway, macro, audit, or signal metadata changes, or when the set of model files changes. If the index is missing, invalid, or stale, SQLMesh safely falls back to a full project load and rebuilds it.
+
+For operations targeting selected models, the index allows SQLMesh to load only those models and their upstream dependencies.
+
+In multi-repository projects, dependencies that cross project boundaries may not be represented by an individual project's index. SQLMesh detects incomplete scoped loads and falls back to loading the full configured project set.
+
 ### Table/view storage locations
 
 SQLMesh creates schemas, physical tables, and views in the data warehouse/engine. Learn more about why and how SQLMesh creates schema in the ["Why does SQLMesh create schemas?" FAQ](../faq/faq.md#schema-question).
--- a/docs/guides/linter.md
+++ b/docs/guides/linter.md
@@ -135,6 +135,21 @@ $ sqlmesh lint --local
 
 This can make linting faster in repositories where all referenced models are loaded from local files. In multi-repository setups, or when linting only a subset of projects, `--local` may cause additional linting errors because SQLMesh will not resolve references or schemas from models that exist only in remote state.
 
+For faster targeted linting, enable the persistent project index with `--use-project-index`. When
+models are selected with `--model`, SQLMesh loads, resolves, and validates only those models and
+their transitive upstream dependencies. The same behavior can be enabled by default for the
+Python API and CLI with the `linter.use_project_index` configuration option:
+
+```yaml
+linter:
+  enabled: true
+  use_project_index: true
+```
+
+`Context.lint_models` uses this configuration value when `use_project_index` is omitted. Passing
+`use_project_index=False` explicitly disables it for that call. If a context was already loaded,
+an indexed lint of selected models reloads the context so the requested scope is applied.
+
 
 ## Applying linting rules
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -650,6 +650,9 @@ Usage: sqlmesh lint [OPTIONS]
 
 Options:
   --model TEXT           A model to lint. Multiple models can be linted.  If no models are specified, every model will be linted.
+  --use-project-index    Use the persistent project index. With --model, only the selected models and their upstream dependencies
+                         are loaded, resolved, and validated, so errors in unrelated models are not reported. Without --model,
+                         every model is still loaded and linted.
   --local                Lint using only locally loaded project files without loading state. In multi-repository setups, or when
                          linting only a subset of projects, this may cause additional linting errors because SQLMesh will not resolve
                          references or schemas from models that exist only in remote state.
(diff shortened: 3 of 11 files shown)


fresioAS:
@StuffbyYuki What is your take on this? The command is opt-in now, would it be better to default the functionality? or perhaps add it as a config setting?

StuffbyYuki:
@fresioAS I’d keep the CLI opt-in for now. Turning this on by default changes `lint --model` semantics (unrelated model errors/validation are skipped), which is the right tradeoff for pre-commit but can be surprising as a silent default. 

I think a config setting (e.g. under `linter`) would be a good next step though

StuffbyYuki (approved):
Not a blocker but these might be good follow up:

1. Python API footgun: `use_project_index=True` is a no-op if the Context was already fully loaded (`load=True`). Perhaps document that on `lint_models` / `load`, or force a scoped reload when the flag is set.
2. Config knob for the pre-commit UX (same as above)

PR #5921, opened 2026-07-25
fix(dialect): make extend_sqlglot idempotent

### LINKED ISSUE #5908: extend_sqlglot() is not idempotent — a second call makes Parser._parse_types recurse infinitely (RecursionError)
`extend_sqlglot()` (`sqlmesh/core/dialect.py`) patches sqlglot's global `Tokenizer`/`Parser`/`Generator` classes via `_override`:

```python
def _override(klass: t.Type[Tokenizer | Parser], func: t.Callable) -> None:
    name = func.__name__
    setattr(klass, f"_{name}", getattr(klass, name))
    setattr(klass, name, func)
```

`_override` saves *whatever is currently installed* under `_<name>` — so applying it twice saves the **wrapper itself** as "the original". For `_parse_types` (saved as `__parse_types`, which the module-level wrapper reads back un-mangled at `dialect.py:487`), the second application makes the wrapper call itself: the next type/CAST parse anywhere in the process raises `RecursionError`, and sqlglot's original method is lost (overwritten), so the process cannot recover without re-importing sqlglot.

### Minimal repro

```python
import sqlglot
import sqlmesh                                  # 1st extend_sqlglot() — sqlmesh/__init__.py:19

from sqlmesh.core.dialect import extend_sqlglot
extend_sqlglot()                                # 2nd application

sqlglot.parse_one("SELECT CAST(1 AS INT)")      # RecursionError: maximum recursion depth exceeded
```

### How this happens without anyone calling it twice on purpose

`extend_sqlglot()` is the **first statement** of `sqlmesh/__init__.py` (line 19), before the rest of the import cascade (analytics, context, model kinds, croniter, ...). If anything later in that first `import sqlmesh` raises, CPython evicts the half-initialised `sqlmesh` package from `sys.modules` — but the fully-imported `sqlmesh.core.dialect` (and the patched global `Parser`) survive. Any *retry* of `import sqlmesh` then re-runs `extend_sqlglot()` on the already-patched classes and arms the recursion.

We hit this in a real test suite two ways: a sandboxed/monkeypatched environment made croniter's import-time `platform.architecture()` subprocess probe raise mid-import; and a deep ambient call stack made a class-body type-parse overflow during the model-kind imports. Both left the parser patched, the package evicted, and the next import poisoned the whole process (`RecursionError` at `dialect.py:487` on every subsequent parse).

### Suggested fix

Make the patch idempotent — e.g. in `_override`:

```python
def _override(klass: t.Type[Tokenizer | Parser], func: t.Callable) -> None:
    name = func.__name__
    if getattr(klass, name, None) is func:      # already installed — don't re-save
        return
    setattr(klass, f"_{name}", getattr(klass, name))
    setattr(klass, name, func)
```

or a module-level ran-once flag at the top of `extend_sqlglot()`. Note the function already has partial re-entry guards (the generator `TRANSFORMS`/`WITH_SEPARATED_COMMENTS` membership checks) — but not on the `_override` path, and `generator.UNWRAPPED_INTERVAL_VALUES` also grows on every call.

### Environment

- sqlmesh 0.236.0 (also reproduced by inspection on main @ `97d60cb`), sqlglot 30.8.0, Python 3.13, macOS/Linux.

### References

- `_override`: https://github.com/SQLMesh/sqlmesh/blob/97d60cb45661f968e202b2d318e8d1d7dbf0cf06/sqlmesh/core/dialect.py#L783-L786
- wrapper / recursion site: https://github.com/SQLMesh/sqlmesh/blob/97d60cb45661f968e202b2d318e8d1d7dbf0cf06/sqlmesh/core/dialect.py#L479-L497
- call site: https://github.com/SQLMesh/sqlmesh/blob/97d60cb45661f968e202b2d318e8d1d7dbf0cf06/sqlmesh/__init__.py#L17-L19
- croniter import-time subprocess (one realistic mid-import failure): https://github.com/pallets-eco/croniter/blob/28ec077e57dd41d96cd9ae664cce31cdc58d07fe/src/croniter/croniter.py#L52-L58

Sanjays2402:
Closes #5908

## Description

`_override` saved whatever was currently installed under `_<name>`, so a second `extend_sqlglot()` call saved the override itself as the "original". The wrapper then called itself and every subsequent type/CAST parse raised `RecursionError`, with sqlglot's original method lost. This is reachable without calling it twice on purpose: `extend_sqlglot()` is the first statement of `sqlmesh/__init__.py`, so if anything later in that import cascade raises, the package is evicted from `sys.modules` while the patched globals survive, and the retry re-applies the patch.

`_override` now returns early when the override is already installed.

## Test Plan

Added `tests/core/test_dialect.py::test_extend_sqlglot_is_idempotent`, which fails with `RecursionError` without the fix and passes with it. `tests/core/test_dialect.py` is green (160 passed).

## Checklist

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -782,6 +782,10 @@ def _parse_interval_span(self: Parser, this: exp.Expr) -> exp.Interval:
 
 def _override(klass: t.Type[Tokenizer | Parser], func: t.Callable) -> None:
     name = func.__name__
+    if getattr(klass, name, None) is func:
+        # Already overridden. Re-applying would save the override itself as the
+        # "original", making the wrapper call itself and recurse infinitely.
+        return
     setattr(klass, f"_{name}", getattr(klass, name))
     setattr(klass, name, func)
 
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -1174,11 +1174,12 @@ def extend_sqlglot() -> None:
                 MacroDef,
             )
 
-        generator.UNWRAPPED_INTERVAL_VALUES = (
-            *generator.UNWRAPPED_INTERVAL_VALUES,
-            MacroStrReplace,
-            MacroVar,
-        )
+        if MacroVar not in generator.UNWRAPPED_INTERVAL_VALUES:
+            generator.UNWRAPPED_INTERVAL_VALUES = (
+                *generator.UNWRAPPED_INTERVAL_VALUES,
+                MacroStrReplace,
+                MacroVar,
+            )
 
     _override(Parser, _parse_select)
     _override(Parser, _parse_statement)
--- a/tests/core/test_dialect.py
+++ b/tests/core/test_dialect.py
@@ -1190,6 +1190,12 @@ def test_pipe_syntax():
 def test_extend_sqlglot_is_idempotent():
     # extend_sqlglot() runs at import time; calling it again must not re-wrap the
     # already-installed overrides, otherwise they call themselves (RecursionError).
+    from sqlglot.generator import Generator
+
+    before = Generator.UNWRAPPED_INTERVAL_VALUES
+
     d.extend_sqlglot()
 
     assert parse_one("SELECT CAST(1 AS INT)").sql() == "SELECT CAST(1 AS INT)"
+    # The class-level registries must not grow on repeated calls.
+    assert Generator.UNWRAPPED_INTERVAL_VALUES == before
--- a/tests/core/test_dialect.py
+++ b/tests/core/test_dialect.py
@@ -1185,3 +1185,11 @@ def test_pipe_syntax():
         ast.sql("bigquery")
         == "SELECT * FROM (WITH __tmp1 AS (SELECT id FROM t2) SELECT * FROM __tmp1)"
     )
+
+
+def test_extend_sqlglot_is_idempotent():
+    # extend_sqlglot() runs at import time; calling it again must not re-wrap the
+    # already-installed overrides, otherwise they call themselves (RecursionError).
+    d.extend_sqlglot()
+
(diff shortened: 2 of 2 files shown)


StuffbyYuki (approved):
@Sanjays2402 Thanks for the PR! Here are a few things (not blockers): 

- Also guard `generator.UNWRAPPED_INTERVAL_VALUES` (still appends every call; TRANSFORMS / WITH_SEPARATED_COMMENTS already have membership guards — called out in #5908).
- Optional assert that a second call doesn’t grow `UNWRAPPED_INTERVAL_VALUES` (or that `__parse_types` still points at sqlglot’s original).

Merge ready once ci tests pass

Sanjays2402:
Thanks @StuffbyYuki — pushed 5f2c7b9.

Added a `MacroVar not in generator.UNWRAPPED_INTERVAL_VALUES` guard so it matches the TRANSFORMS / WITH_SEPARATED_COMMENTS pattern, and the idempotence test now snapshots `Generator.UNWRAPPED_INTERVAL_VALUES` before the second `extend_sqlglot()` call and asserts it's unchanged.

On the 3.9 style job: the 12 failures are all `TypeError: unsupported operand type(s) for |: 'type' and 'type'` in tests/dbt (bigquery relation/column classes), which is PEP 604 syntax at runtime on 3.9 — unrelated to this diff. tests/core/test_dialect.py is green locally (160 passed).

StuffbyYuki:
@Sanjays2402 Can you add DCO checks?

Sanjays2402:
Done — signed off the second commit and force-pushed, head is now fb992b7. `10c1b46` already had the trailer; `132f16b` (the `UNWRAPPED_INTERVAL_VALUES` guard) was missing it, which is what the Commits Check was failing on.

Amended with `--trailer` only, so the tree is byte-identical to 132f16b (`git diff 132f16b HEAD --stat` is empty). CI is re-queuing now.

PR #5925, opened 2026-07-28
fix(plan): add explainer visit method for PhysicalLayerSchemaCreationStage

### LINKED ISSUE #5619: Explainer doesn't have visit method for PhysicalLayerSchemaCreationStage
When explaining a plan for which the PhysicalLayerSchemaCreationStage is created the following error is thrown in the logs:

```
sqlmesh.core.plan.explainer - ERROR - Unexpected stage: PhysicalLayerSchemaCreationStage (explainer.py:134)
```

Link to code: https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/plan/explainer.py#L134

It seems that the visit method `visit_physical_layer_schema_creation_stage()` for this stage is missing from explainer.py.

The corresponding method is defined in evaluator.py: [visit_physical_layer_schema_creation_stage()](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/plan/evaluator.py#L209).

nkwork9999:
Fixes #5619

## Problem

`RichExplainerConsole` has no `visit_physical_layer_schema_creation_stage`, so when a plan contains `PhysicalLayerSchemaCreationStage`, `explain()` falls into the `hasattr` guard at [`explainer.py:134`](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/plan/explainer.py#L134):

```
sqlmesh.core.plan.explainer - ERROR - Unexpected stage: PhysicalLayerSchemaCreationStage
```

The stage is then silently dropped from the explained plan, even though `PlanEvaluator` handles it in [`evaluator.py:209`](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/plan/evaluator.py#L209).

## Change

Add the missing visit method. It mirrors the schema derivation in `SnapshotEvaluator.create_physical_schemas` / `_create_schemas` — non-symbolic models only, table name resolved through the stage's deployability index, deduplicated by `(db, catalog)` — and lists the distinct physical schemas that will be created. Returns `None` when there is nothing to create, consistent with the other `Optional[Tree]` visit methods.

Placed directly before `visit_physical_layer_update_stage` to match the stage ordering produced in `stages.py`.

Output on `examples/sushi`:

```
├── Create physical schemas if they do not exist
│   └── memory.sqlmesh__sushi
├── Backfill models by running their queries and run standalone audits
│   ├── sushi.waiter_revenue_by_day -> ...
```

## Tests

`test_plan_explain` previously only asserted that explaining does not raise, which is why this went unnoticed. It now asserts that no stage is reported as unexpected, which covers this stage and guards against the same omission for stages added later.

Verified the test fails on `main` with the fix reverted:

```
E  AssertionError: assert 'Unexpected stage' not in '...'
E    'Unexpected stage' is contained here:
E      er.py:134 Unexpected stage: PhysicalLayerSchemaCreationStage
```

`tests/core/test_plan.py`, `tests/core/test_plan_stages.py`, `tests/core/test_context.py` and `tests/core/integration/test_plan_options.py` pass (207 tests), as do `ruff format`, `ruff check` and `mypy` on the changed files.

### DIFF
--- a/sqlmesh/core/plan/explainer.py
+++ b/sqlmesh/core/plan/explainer.py
@@ -8,9 +8,11 @@ from collections import defaultdict
 
 from rich.console import Console as RichConsole
 from rich.tree import Tree
+from sqlglot import exp
 from sqlglot.dialects.dialect import DialectType
 from sqlmesh.core import constants as c
 from sqlmesh.core.console import Console, TerminalConsole, get_console
+from sqlmesh.core.dialect import schema_
 from sqlmesh.core.environment import EnvironmentNamingInfo
 from sqlmesh.core.plan.common import (
     SnapshotIntervalClearRequest,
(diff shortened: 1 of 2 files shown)


StuffbyYuki:
@nkwork9999 Thanks for this pr! 

No blocker, but do you think it's worth asserting the explained output includes the schema-creation node (e.g. “Create physical schemas…” / `sqlmesh__sushi`), not only that “Unexpected stage” is absent. Today’s guard catches missing handlers but not wrong/empty rendering.

StuffbyYuki:
@nkwork9999 Just wanted to follow up on this!

StuffbyYuki:
@nkwork9999 Bump

PR #5931, opened 2026-07-30
fix(macros): resolve_template fails in audits when this_model is a subquery

### LINKED ISSUE #5927: Bug: @resolve_template fails in audits on INCREMENTAL_BY_TIME_RANGE models (this_model is a Subquery)
## Summary

`@resolve_template` raises `AttributeError: 'Subquery' object has no attribute 'catalog'` when used in an audit attached to an `INCREMENTAL_BY_TIME_RANGE` model. The root cause is not in the audit itself but in how `this_model` is passed to the macro evaluator during the audit rendering phase for incremental models.

## Root Cause

In `sqlmesh/core/model/definition.py` ([`render_audit_query`](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/model/definition.py)), when the model has a `time_column`, `this_model` is set to a `Subquery`:

```python
"this_model": exp.select("*").from_(quoted_model_name).where(where).subquery()
if where is not None
else quoted_model_name,
```

This means for any model with a `time_column` (i.e. all `INCREMENTAL_BY_TIME_RANGE` models), `this_model` in `evaluator.locals` is always a `Subquery`, not a `Table`.

Then in `sqlmesh/core/macros.py` ([`resolve_template`](https://github.com/TobikoData/sqlmesh/blob/main/sqlmesh/core/macros.py)), the macro assumes `this_model` is always convertible to a `Table`:

```python
this_model = exp.to_table(evaluator.locals["this_model"], dialect=evaluator.dialect)
template_str: str = template.this
result = (
    template_str.replace("@{catalog_name}", this_model.catalog)
    .replace("@{schema_name}", this_model.db)
    .replace("@{table_name}", this_model.name)
)
```

`exp.to_table()` on a `Subquery` does not produce an object with `.catalog`/`.db`/`.name` attributes, so it raises:

```
AttributeError: 'Subquery' object has no attribute 'catalog'
```

## How to Reproduce

1. Create an `INCREMENTAL_BY_TIME_RANGE` model with a `time_column`
2. Attach an audit that uses `@resolve_template` with `mode := 'table'`

```sql
-- Model
MODEL (
  name my_schema.my_model,
    time_column (event_date, '%Y-%m-%d')
  ),
  audits (my_audit),
);

SELECT event_date, app_name, user_id FROM upstream.source
WHERE event_date BETWEEN @start_date AND @end_date;
```

```sql
-- Audit
AUDIT (name my_audit, dialect snowflake);

SELECT app_name, COUNT(*) as row_count
FROM @resolve_template('@{catalog_name}.@{schema_name}.@{table_name}', mode := 'table')
WHERE event_date >= DATEADD('day', -30, @end_date::date)
GROUP BY app_name
HAVING row_count = 0;
```

3. Run `sqlmesh run` — the audit fails with the `AttributeError`.

## Use Case

`@resolve_template(..., mode := 'table')` is useful in audits that need to query a broader time range than the current batch. For example, an audit that checks "no app has gone completely silent in the last 30 days" needs access to the full materialized table — not just the current 7-day batch that `@this_model` provides. The `mode := 'table'` parameter was designed for exactly this purpose, but it doesn't work in the audit phase for incremental models.

## Suggested Fix

`@resolve_template` should handle the case where `this_model` is a `Subquery` by extracting the underlying `Table` from it. For example:

```python
this_model_expr = evaluator.locals["this_model"]
if isinstance(this_model_expr, exp.Subquery):
    # Extract the Table from: (SELECT * FROM <table> WHERE ...) AS _q_0
    table = this_model_expr.find(exp.Table)
else:
    table = exp.to_table(this_model_expr, dialect=evaluator.dialect)
```

Alternatively, `render_audit_query` could pass the raw table name as a separate key (e.g., `this_model_table`) alongside the subquery version so that macros like `@resolve_template` can access it without needing to unwrap the subquery.

## Related

- #3990 / PR #3991 fixed the `where is None` branch (when there's no `time_column`), but the `where is not None` path still produces the Subquery that breaks `@resolve_template`.

## Environment

- SQLMesh version: 0.236.0
- Dialect: Snowflake

sravankumarkunadi:
## Description

`@resolve_template` raises `AttributeError: 'Subquery' object has no attribute 'catalog'` in an audit attached to a model with a `time_column` (so, any `INCREMENTAL_BY_TIME_RANGE` model).

The macro assumes `@this_model` is a table and calls `.catalog` / `.db` / `.name` on it. But `Model.render_audit_query` only passes a table when there's no time filter — otherwise it passes a subquery restricting the physical table to the audited range:

```python
"this_model": exp.select("*").from_(quoted_model_name).where(where).subquery()
if where is not None
else quoted_model_name,
```

`exp.to_table()` returns non-`Table` expressions untouched, so the `Subquery` reaches `.catalog` and raises. #3991 fixed the `where is None` branch; the time-column branch still produces the subquery.

This unwraps the subquery to the table it selects from before resolving the template. `@this_model` is unchanged — audits still get the time-filtered subquery — only the catalog/schema/table lookup reaches through it, which is what those placeholders mean. That makes `mode := 'table'` usable in audits that need the full materialized table rather than the current batch.

Closes #5927

## Test Plan

- `test_resolve_template_subquery` (`test_macros.py`) — `mode := 'table'` with a subquery `this_model`; fails on `main` with the reported `AttributeError`.
- `test_resolve_template` (`test_audit.py`) — end to end through `render_audit_query` on an incremental model, both modes; the literal case also asserts `@this_model` still renders as the time-filtered subquery.
- `pytest tests/core/test_macros.py tests/core/test_audit.py tests/core/test_model.py -m "not slow and not docker"` → 487 passed vs 485 on `main` (the 2 new tests). Both runs hit the same 27 errors, all `ConfigError: Materialization strategy ... was not found` — test fixture entry points missing from my venv, none in the two files touched here.
- `make style` clean on the changed files.

## Checklist

- [ ] All existing tests pass (`make fast-test`) — ran the three files above rather than the full suite, see the pre-existing local errors noted there

### DIFF
--- a/sqlmesh/core/console.py
+++ b/sqlmesh/core/console.py
@@ -2920,7 +2920,10 @@ class NotebookMagicConsole(TerminalConsole):
 
         super().__init__(console, **kwargs)
 
-        self.display = display or get_ipython().user_ns.get("display", ipython_display)
+        ipython = get_ipython()
+        self.display = display or (
+            ipython.user_ns.get("display", ipython_display) if ipython else ipython_display
+        )
         self.missing_dates_output = widgets.Output()
         self.dynamic_options_after_categorization_output = widgets.VBox()
 
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1402,7 +1402,13 @@ def resolve_template(
         "'s3://data-bucket/prod/test_catalog/sqlmesh__test/test__test_model__2517971505'"
     """
     if "this_model" in evaluator.locals:
-        this_model = exp.to_table(evaluator.locals["this_model"], dialect=evaluator.dialect)
+        this_model_expr = evaluator.locals["this_model"]
+        if isinstance(this_model_expr, exp.Subquery):
+            # Audits on models with a time column render @this_model as a subquery that filters the
+            # physical table on the audited time range, so extract the table it selects from
+            this_model_expr = this_model_expr.find(exp.Table) or this_model_expr
+
+        this_model = exp.to_table(this_model_expr, dialect=evaluator.dialect)
         template_str: str = template.this
         result = (
             template_str.replace("@{catalog_name}", this_model.catalog)
--- a/sqlmesh/core/macros.py
+++ b/sqlmesh/core/macros.py
@@ -1379,15 +1379,17 @@ def resolve_template(
     """
     Generates either a String literal or an exp.Table representing a physical table location, based on rendering the provided template String literal.
 
-    Note: It relies on the @this_model variable being available in the evaluation context (@this_model resolves to an exp.Table object
-    representing the current physical table).
+    Note: It relies on the @this_model variable being available in the evaluation context. @this_model usually resolves to an
+    exp.Table object representing the current physical table, but in an audit on a model with a time column it resolves to a
+    subquery that selects from that table and filters it down to the audited time range. In that case the placeholders below
+    are resolved against the physical table the subquery selects from.
     Therefore, the @resolve_template macro must be used at creation or evaluation time and not at load time.
 
     Args:
         template: Template string literal. Can contain the following placeholders:
-            @{catalog_name} -> replaced with the catalog of the exp.Table returned from @this_model
-            @{schema_name} -> replaced with the schema of the exp.Table returned from @this_model
-            @{table_name} -> replaced with the name of the exp.Table returned from @this_model
+            @{catalog_name} -> replaced with the catalog of the physical table @this_model refers to
+            @{schema_name} -> replaced with the schema of the physical table @this_model refers to
+            @{table_name} -> replaced with the name of the physical table @this_model refers to
         mode: What to return.
             'literal' -> return an exp.Literal string
             'table' -> return an exp.Table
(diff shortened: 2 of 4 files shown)


sravankumarkunadi:
@StuffbyYuki can I please get a review on this one?

StuffbyYuki (approved):
No blocker but a few things: 
- Update the `@resolve_template` docstring, it still says `@this_model` is always an `exp.Table`.
- Prefer extracting the table from the subquery’s `FROM` rather than `.find(exp.Table)`, so nested/extra table refs can’t surprise us later.

Thoughts? Let me know if I'm missing anything! Omce ci passes I can merge

sravankumarkunadi:
Both done, thanks.

Reads the FROM clause now, plus a test with a second table in the WHERE. (`.find()` happened to pick the right one there anyway.)

The failing style jobs are unrelated — that's the `console.py` mypy error from IPython 9.16, red on main too, fix is in #5933. Can merge main in once that lands.

StuffbyYuki:
@sravankumarkunadi hey can you check the failed ci tests? It looks like a few style tests have failed

sravankumarkunadi:
@StuffbyYuki fixes the style issue please review and lmk if anything should be adjusted thanks

PR #5937, opened 2026-08-04
fix(tsql): ALTER COLUMN with a nullability clause is parsed as a Command

### LINKED ISSUE #5932: Bug: Syntax for MSSQL ALTER COLUMN in post-statement is not recognised and causes an error
Using a post-statement this SQL is executed correctly:

@IF(
  @runtime_stage = 'creating',
  ALTER TABLE @SQL('@this_model')
  ADD CONSTRAINT @resolve_template('PK_@{table_name}') PRIMARY KEY CLUSTERED (id) 
)

The syntax words are all highlighted correctly.

This similar post-statement causes an error:

@IF(
  @runtime_stage = 'creating',
  ALTER TABLE @SQL('@this_model')
  ALTER COLUMN id INT NOT NULL
)

Tellingly the world COLUMN in ALTER COLUMN is not highlighted.

Executing plan first fires this warning:
[WARNING] 'ALTER TABLE @SQL('@this_model')
  ALTER COLUMN id INT NOT NULL' could not be semantically understood as it contains unsupported syntax, SQLMesh will treat the command as is. Note that any references to the model's underlying physical table can't be resolved in this case, consider using Jinja as explained here 
https://sqlmesh.readthedocs.io/en/stable/concepts/macros/macro_variables/#audit-only-variables

Before the actual error:
ProgrammingError:
      (102, b"Incorrect syntax near '@SQL'.DB-Lib error message 20018, severity 15:\nGeneral SQL Server error: Check messages from the SQL Server\nDB-Lib error message 20018, 
severity 15:\nGeneral SQL Server error: Check messages from the SQL Server\n")

As an aside, this is itself a hacky workaround as the model creates the column as nullable and I do not see any option to override that.  If there's a better way to define a column as non-nullable then I would really appreciate the advice.

sravankumarkunadi:
## Description

`ALTER TABLE t ALTER COLUMN c INT NOT NULL` doesn't parse in T-SQL and falls back to a `Command`. Macros aren't resolved inside a `Command`, so `@SQL('@this_model')` reaches SQL Server verbatim and fails with `Incorrect syntax near '@SQL'`.

sqlglot's `_parse_alter_table_alter` returns as soon as it has the type, so T-SQL's trailing `NOT NULL` is left unconsumed and parsing raises. The `ADD CONSTRAINT` variant in the issue works because it parses cleanly.

Adds a T-SQL-scoped parser override that consumes the trailing `NULL` / `NOT NULL` into `allow_null`, plus a generator override to emit it — sqlglot's `altercolumn_sql` returns at the `dtype` branch before it reaches `allow_null`. Both only engage when a `dtype` is present, so `SET NOT NULL` / `DROP NOT NULL` keep their existing path.

Post-statement from the issue, `creating` stage:

```diff
- ALTER TABLE @SQL('@this_model') ALTER COLUMN id INT NOT NULL   # + "could not be semantically understood" warning
+ ALTER TABLE [cat].[sqlmesh__test].[test__demo__3571846471] ALTER COLUMN [id] INTEGER NOT NULL
```

Doesn't address the reporter's aside about declaring a column `NOT NULL` without a post-statement.

Closes #5932

## Test Plan

- `test_tsql_alter_column_nullability` (`test_dialect.py`) — parse/round-trip for `INT NOT NULL`, `INT NULL`, bare `INT`, the `@IF` form from the issue, and `DROP NOT NULL`. Fails on `main`.
- `test_tsql_alter_column_post_statement` (`test_model.py`) — `render_post_statements` resolves `@this_model` at the `creating` stage.
- `pytest tests/core/test_dialect.py tests/core/test_model.py tests/core/test_macros.py` → 609 passed vs 607 on `main`; both runs hit the same 26 pre-existing `sushi_context` fixture errors (`Materialization strategy ... was not found`).

## Checklist

- [x] I have run `make style` and fixed any issues — `ruff format`/`ruff check`/`mypy` on the changed files, not `pre-commit run --all-files`
- [ ] All existing tests pass (`make fast-test`) — ran the three files above, see the pre-existing local errors noted there

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -551,6 +551,36 @@ def _parse_table_parts(
     return table
 
 
+# Only needed for T-SQL: it spells a column's nullability right after its type, e.g.
+# ALTER TABLE t ALTER COLUMN c INT NOT NULL. Without this the trailing clause is left
+# over, so the whole statement falls back to a Command and any macros it contains (such as
+# @this_model) are no longer resolved, which means they reach the engine verbatim.
+#
+# See: https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql
+def _parse_alter_table_alter(self: Parser) -> t.Optional[exp.Expr]:
+    alter_column = self.__parse_alter_table_alter()  # type: ignore
+
+    if isinstance(alter_column, exp.AlterColumn) and alter_column.args.get("dtype"):
+        if self._match_pair(TokenType.NOT, TokenType.NULL):
+            alter_column.set("allow_null", False)
+        elif self._match(TokenType.NULL):
+            alter_column.set("allow_null", True)
+
+    return alter_column
+
+
+def altercolumn_sql(self: Generator, expression: exp.AlterColumn) -> str:
+    sql = self._altercolumn_sql(expression)  # type: ignore
+
+    # sqlglot's generator returns as soon as it renders the type, so the nullability parsed
+    # above has to be appended here
+    allow_null = expression.args.get("allow_null")
+    if expression.args.get("dtype") and allow_null is not None:
+        sql = f"{sql} NULL" if allow_null else f"{sql} NOT NULL"
+
+    return sql
+
+
 def _parse_if(self: Parser) -> t.Optional[exp.Expr]:
     # If we fail to parse an IF function with expressions as arguments, we then try
     # to parse a statement / command to support the macro @IF(condition, statement)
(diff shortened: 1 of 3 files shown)


sravankumarkunadi:
@StuffbyYuki can I please get a review on this one? thanks.

StuffbyYuki (approved):
Is it worth considering upstreaming the parser/generator nullability handling to sqlglot long-term? This is really a dialect gap there, the SQLMesh override is a reasonable short-term fix though

sravankumarkunadi:
Agreed, it should be fixed in sqlglot. I'll open a PR there and link it.

This one doesn't need to wait though, we're pinned to 30.8.0 so it'd be a while before the fix reaches us. Happy to remove the override later.

sravankumarkunadi:
Opened it upstream: https://github.com/tobymao/sqlglot/pull/8053

Went with a `SUPPORTS_ALTER_COLUMN_NULLABILITY` flag there instead of a tsql-only generator override, so other dialects warn rather than silently dropping the `NOT NULL`.

PR #5940, opened 2026-08-05
fix(clickhouse): strip virtual catalog from view sources

### LINKED ISSUE #5938: ClickHouse virtual catalog is not stripped from CREATE VIEW source queries
## Describe the bug

In a multi-gateway project using ClickHouse's virtual-catalog support from
#5826, SQLMesh strips the synthetic catalog from the target of a virtual-layer
`CREATE VIEW`, but leaves it on the physical view referenced by the `SELECT`.

ClickHouse receives an invalid three-part source name:

```sql
CREATE OR REPLACE VIEW "my_db"."connection_test__dev" ON CLUSTER "my_cluster"
AS SELECT *
FROM "__clickhouse_gw__"."my_db"."my_db__connection_test__1234567890"
```

and rejects it because ClickHouse supports only `database.table` names.

## To reproduce

1. Configure a project with a catalog-aware default gateway and a ClickHouse
   secondary gateway.
2. Enable `gateway_managed_virtual_layer`.
3. Define a ClickHouse `VIEW` model using `gateway clickhouse_gw`.
4. Run `sqlmesh plan dev --auto-apply`.

The physical view is created successfully. Virtual-layer promotion then fails
with ClickHouse error code 62 when SQLMesh executes the three-part source
reference shown above.

## Expected behavior

SQLMesh should remove the configured synthetic virtual catalog from every
ClickHouse table reference before sending view DDL to ClickHouse. The generated
query should be equivalent to:

```sql
CREATE OR REPLACE VIEW "my_db"."connection_test__dev" ON CLUSTER "my_cluster"
AS SELECT *
FROM "my_db"."my_db__connection_test__1234567890"
```

## Likely cause

The `@set_catalog()` handling on the inherited `create_view()` method normalizes
the `view_name` argument, but not tables nested inside the `query_or_df`
expression. The snapshot evaluator supplies the physical source using its
three-level virtual FQN, so the synthetic catalog reaches ClickHouse.

A local adapter override that copied the query expression and removed only the
adapter's configured virtual catalog from nested `exp.Table` nodes allowed the
same plan to promote successfully. An unexpected catalog was still rejected.

## Impact

Gateway-managed virtual-layer promotion cannot complete for ClickHouse models
in affected multi-gateway projects. Physical objects may be created before the
promotion failure, leaving a partially applied plan to clean up.

## Environment

- SQLMesh: 0.236.0
- ClickHouse: 25.8
- ClickHouse adapter in cluster mode

## Related work

- #5748
- #5826

mday-io:
## Summary

- strip ClickHouse's configured virtual catalog from physical source tables in `CREATE VIEW` queries
- preserve the caller's query expression and reject unexpected source catalogs

## Why

The inherited catalog handling normalized the virtual view target but left synthetic three-part source names in the query. ClickHouse only accepts two-part `database.table` names, so gateway-managed virtual-layer promotion failed after creating the physical object.

## Validation

- `make style`
- `make fast-test`
- `pytest tests/core/engine_adapter/test_clickhouse.py`

Closes #5938

### DIFF
--- a/sqlmesh/core/engine_adapter/clickhouse.py
+++ b/sqlmesh/core/engine_adapter/clickhouse.py
@@ -595,6 +595,48 @@ class ClickhouseEngineAdapter(EngineAdapterWithIndexSupport, LogicalMergeMixin):
                 target_columns_to_types or self.columns(table_name),
             )
 
+    def create_view(
+        self,
+        view_name: TableName,
+        query_or_df: QueryOrDF,
+        target_columns_to_types: t.Optional[t.Dict[str, exp.DataType]] = None,
+        replace: bool = True,
+        materialized: bool = False,
+        materialized_properties: t.Optional[t.Dict[str, t.Any]] = None,
+        table_description: t.Optional[str] = None,
+        column_descriptions: t.Optional[t.Dict[str, str]] = None,
+        view_properties: t.Optional[t.Dict[str, exp.Expr]] = None,
+        source_columns: t.Optional[t.List[str]] = None,
+        **create_kwargs: t.Any,
+    ) -> None:
+        if self._default_catalog and isinstance(query_or_df, exp.Query):
+            from sqlmesh.utils.errors import SQLMeshError
+
+            query_or_df = query_or_df.copy()
+            for table in query_or_df.find_all(exp.Table):
+                if not table.catalog:
+                    continue
+                if table.catalog != self._default_catalog:
+                    raise SQLMeshError(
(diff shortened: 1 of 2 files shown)


StuffbyYuki (approved):
Should we use `_strip_virtual_catalog()` instead of inlining `table.set("catalog", None)`

mday-io:
That helper creates a separate table object, so it wouldn’t update the table inside the view query. We also need to catch unexpected catalogs here, which is why this code handles it directly

PR #5944, opened 2026-08-06
feat(format): add transpile_meta opt-in for dialect-aware MODEL headers

albertosuman-1k5:
## Summary
- Adds opt-in `format.transpile_meta` so `sqlmesh format` can render `MODEL`/`AUDIT`/`METRIC` headers with the model dialect again.
- Default remains dialect-agnostic headers (fix for #5773 / #5864): SQLMesh boolean properties stay as `TRUE`/`FALSE` instead of dialect rewrites like T-SQL `(1 = 1)`.
- Projects that author headers in warehouse dialect (column types, audits, labels, macros, etc.) can set `format.transpile_meta: true` to preserve those values.

## Context
#5864 stopped transpiling meta expressions to avoid corrupting SQLMesh DDL booleans. That also rewrote dialect-specific header values (for example `DATETIME2(6)` → `TIMESTAMP(6)`). A surgical per-property approach (see #5926 for columns) would need similar handling for audits, labels, macros, and other header properties; this flag restores the previous whole-header behavior as an explicit opt-in instead.

**Trade-off:** enabling the flag reopens the T-SQL boolean rewrite from #5773 for SQLMesh properties in the header. Documented in `docs/reference/configuration.md`.

## Test plan
- [x] Unit tests for `format_model_expressions(..., transpile_meta=True)` (multi-expression and single-meta paths)
- [x] Config test that `transpile_meta` is excluded from `generator_options` (does not leak into SQLGlot `Generator`)
- [x] End-to-end `Context.format()` with `FormatConfig(transpile_meta=True)` preserves `DATETIME2(6)`
- [ ] Manual: project with T-SQL headers containing `columns` / `audits` / macros — format with and without the flag

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -119,6 +119,7 @@ Formatting settings for the `sqlmesh format` command and UI.
 | `max_text_width`      | The maximum text width in a segment before creating new lines (Default: 80)                    |   int   |    N     |
 | `append_newline`      | Whether to append a newline to the end of the file (Default: False)                            | boolean |    N     |
 | `no_rewrite_casts`    | Preserve the existing casts, without rewriting them to use the :: syntax. (Default: False)                  | boolean |    N     |
+| `transpile_meta`      | Whether to render the `MODEL`/`AUDIT`/`METRIC` header with the model's dialect instead of keeping it dialect-agnostic. Headers are dialect-agnostic by default because SQLMesh properties are not warehouse SQL, but projects that author headers in their warehouse dialect can enable this to preserve dialect-specific values such as column types. Note that enabling it also transpiles SQLMesh's own properties, so a boolean such as `allow_partials TRUE` is rendered as `(1 = 1)` in T-SQL. (Default: False) | boolean |    N     |
 
 
 ## Janitor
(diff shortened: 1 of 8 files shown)


fresioAS:
@albertosuman-1k5 
I'm not sure I understand the rationale here? We changed all our T-SQL models over to dialect agnostic headers - the main reason being that (1=1) was not respected and we got into issues with everything from disable_restatement (1=0) to formatting (1=0) etc. In addition we had fields like description being [<desc>] instead of '<desc>'. 

Based on this I don't see who format.transpile_meta: true would be useful for?

On the other side - it is a pain now when the formatter wants to change our columns to something that does not exist - hence the proposed PR #5926

mday-io:
@albertosuman-1k5 thank you very much for proposing this. I do worry that this all-or-nothing approach will continue to cause people headaches. I believe #5926 by @fresioAS (with a few additions) might get us a quick-fix and set a foundation for a more scaleable solution.

albertosuman-1k5:
@mday-io We got headaches because of the sudden change in how headers are parsed, change that was not communicated and without an opt-out flag 😅 

Why would an opt-out feature that can be removed once everyone had the chance to prepare their projects (let's say in 6 months) would give everyone headers? We're dealing with 2000+ models, roughly half of them change when using the agnostic formatter. At the moment we don't have the capacity to tackle this but we'd like to upgrade to make use of a feature I contributed to

PR #5949, opened 2026-08-10
fix(format): preserve dialect-specific SQL in MODEL/AUDIT/METRIC headers

mday-io:
## Description

`#5864` fixed formatting corrupting SQLMesh's own header properties (on tsql, `allow_partials TRUE` was becoming the unparseable `(1 = 1)`) by no longer transpiling MODEL/AUDIT/METRIC headers at all.

That was too broad a fix: it also stopped transpiling the properties that *are* real warehouse SQL, like `columns`, `audits`, and `physical_properties`. So `columns (ts DATETIME2(6))` would silently flatten to generic `TIMESTAMP(6)`. On tsql this compounds across repeated formatting — `DATETIME2` → `TIMESTAMP` → reparsed as `ROWVERSION` → `VARBINARY` — so two `sqlmesh format` passes could turn a datetime column into a binary one with no error.

This PR renders header properties individually instead of applying one blanket rule. Properties holding real warehouse SQL (`columns`, `audits`, `physical_properties`, macro properties, and expression fields nested inside `kind` like `time_data_type`/`unique_key`/`time_column`) render with the model's dialect. Properties that are SQLMesh's own semantics (`allow_partials`, `description`, scalar `kind` properties like `forward_only`) stay dialect-agnostic. Which bucket a property falls into is derived automatically from its field declaration, so the split stays correct as new properties are added, and a missed property fails safe — a keyword goes uncanonicalized rather than a user's SQL being corrupted.

Covers MODEL, AUDIT, and METRIC headers.

## Test Plan

- `test_format_model_expressions_meta_render_policy` (parametrized over tsql/fabric) and `test_format_audit_expressions_meta_render_policy`, asserting `columns`, `audits`, `physical_properties`, and `kind`'s nested expression properties keep their dialect-specific spelling while SQLMesh's own scalar properties stay dialect-agnostic.
- `test_format_model_expressions_is_idempotent`, parametrized over `columns`, `audits`, `kind`, `physical_properties`, SQLMesh-owned scalars, and macro properties, asserting a second `sqlmesh format` pass is a no-op.
- `test_format_model_expressions_time_column_dialect` — `time_column`'s identifier quoting (e.g. tsql `[end]`) survives formatting instead of falling back to ANSI quoting.
- `test_format_model_expressions_macro_property_comments_preserved_with_dialect` — comments inside a macro header-property's arguments survive formatting on a dialect-bearing model, plus an idempotency check.
- `test_format_model_expressions_kind_scalar_sibling_dialect` — a `kind` block combining an expression property (`time_column`) with a boolean scalar sibling (`forward_only`) on tsql round-trips both correctly; asserts on the formatted string *and* round-trips through `load_sql_based_model` to confirm `model.kind.forward_only is True` survives (not just the string).
- `pytest tests/core/test_dialect.py` — 173 passed.
- `ruff check` on the changed files — clean.

## Release Note

**Fix: `sqlmesh format` corrupting warehouse-specific types in MODEL/AUDIT/METRIC headers**

`sqlmesh format` on models with an explicit `dialect` (tsql, fabric, and others with dialect-specific type spellings) could silently rewrite warehouse-specific types in `columns`, `audits`, `physical_properties`, and `kind`-nested properties (e.g. `time_data_type`, `time_column`) to a generic spelling. For tsql specifically, this was compounding: `DATETIME2` → `TIMESTAMP` → `VARBINARY` across two format runs — for an SCD Type 2 model's `time_data_type`, that silently turned the physical type of the `valid_from`/`valid_to` columns from a datetime into a binary type.

Header properties that hold SQLMesh's own semantics (`allow_partials`, `description`, `kind` name, boolean kind properties like `forward_only`, etc.) are unaffected and continue to render dialect-agnostically, so this only changes output for the subset of header properties that carry actual warehouse SQL.

**Action for affected users**: if you run `sqlmesh format` on tsql/fabric models with warehouse-specific types in `columns`, `audits`, `physical_properties`, or SCD `kind` blocks, re-run `sqlmesh format` after upgrading and review the diff — previously-corrupted types will be restored to their correct dialect-specific spelling.

## Checklist

### DIFF
--- a/sqlmesh/core/dialect.py
+++ b/sqlmesh/core/dialect.py
@@ -734,6 +734,52 @@ PARSERS = {
 }
 
 
+_SQLMESH_META_DIALECT = "sqlmesh_meta_dialect"
+
+
+def _holds_expression(annotation: t.Any) -> bool:
+    """Whether a declared field type bottoms out in a SQLGlot expression.
+
+    Covers List[exp.Expr], Optional[Dict[str, exp.DataType]], Optional[exp.Tuple] and
+    the nested Tuple[str, Dict[str, exp.Expr]] shape used by audits/signals.
+    """
+    if isinstance(annotation, type) and issubclass(annotation, exp.Expr):
+        return True
+    return any(_holds_expression(arg) for arg in t.get_args(annotation))
+
+
+@functools.lru_cache(maxsize=1)
+def _meta_render_policy() -> t.Dict[str, bool]:
+    """Map header property name -> whether its value is warehouse SQL.
+
+    Derived from the field declarations themselves, so it stays correct as properties
+    are added: expression-typed values (columns, audits, physical_properties, ...) are
+    the user's warehouse SQL and must render in the model's dialect, while scalar-typed
+    values (allow_partials, description, kind, ...) are SQLMesh's own semantics and must
+    stay dialect-agnostic -- transpiling those is what corrupts `allow_partials TRUE`
+    into tsql's unparseable `(1 = 1)`.
+    """
+    import inspect
+
+    from sqlmesh.core.audit.definition import ModelAudit
+    from sqlmesh.core.metric.definition import MetricMeta
+    from sqlmesh.core.model import kind as kind_module
+    from sqlmesh.core.model.meta import ModelMeta
+
+    sources: t.List[t.Any] = [ModelMeta, ModelAudit, MetricMeta]
+    sources.extend(
+        obj
+        for name, obj in vars(kind_module).items()
+        if inspect.isclass(obj) and hasattr(obj, "model_fields") and name.endswith("Kind")
+    )
+
+    policy: t.Dict[str, bool] = {}
+    for source in sources:
+        for name, field in source.model_fields.items():
+            policy.setdefault((field.alias or name).lower(), _holds_expression(field.annotation))
+    return policy
+
+
 def _props_sql(self: Generator, expressions: t.List[exp.Expr]) -> str:
     props = []
     size = len(expressions)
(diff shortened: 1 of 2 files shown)


albertosuman-1k5:
Hey @mday-io 
I'm planning to test the fix by the end of the week

albertosuman-1k5:
I just tested this branch against our project (~2k models, BigQuery dialect). 

`sqlmesh format` rewrites tags ['C1'] to tags ARRAY('C1') in MODEL headers, which then fails to load: Required keyword: 'value' missing for Property — the sqlmesh dialect parser only accepts bracket array literals. This is the issue that also @cmgoffena13 mentioned

It also rewrote a string literal inside a query body: `JSON_VALUE(action, '$."28d_click"')` → `'$.28d_click'`, which is invalid BigQuery JSONPath (keys starting with digits need quoting). Both suggest header/property expressions are still being rendered with a non-BigQuery/non-sqlmesh dialect instead of preserved.

mday-io:
Thank you very much. Looking into this ASAP.

albertosuman-1k5:
Hi @mday-io 
Any news on the fix?

mday-io:
@albertosuman-1k5 thanks for testing on your models. I pushed a fix for the tags ['C1'] → ARRAY('C1') problem: list properties like tags, ignored_rules and grain now stay [...] on BigQuery, and the formatted models load again. There's a new test that runs sqlmesh format on a BigQuery model. Could you re-run the branch on your models?

The JSON path problem ('$."28d_click"' → '$.28d_click') comes from sqlglot, not this PR. That will need to be handled separately. 

@cmgoffena13 I was wrong earlier to call the ignored_rules → ARRAY(...) rewrite cosmetic: with more than one value it broke loading on BigQuery. That's fixed now, along with tags.

cmgoffena13:
@mday-io -- is this ready for review? I can take a look

mday-io:
@cmgoffena13 plz

PR #5966, opened 2026-08-21
feature: adding smarter unit test execution

### LINKED ISSUE #5965: Smarter Unit Test Execution
Right now all unit tests run during a plan unless the `--skip-tests` flag is used.

We already have the plan which shows us which models are affected by the changes. It makes sense to trim the unit tests that run during a plan to only the models that are affected by applying the plan.

I am not sure what behavior should happen for the `--skip-backfill` / `--dry-run` flags, but my initial thought is to skip all tests for this flag as well since it's considered a no-op operation.

To have backwards compatibility, we can add in an `--all-tests` flag to the plan command. I can see this being useful for CI dry-runs as well as edge cases that may show up.

cmgoffena13:
## Description

- Add in `--test-changed-only` flag to only run tests for changed models in the plan.
- Default behavior remains running all tests for all models
- Add `--select-model` flag to the `test` command for improved functionality

Addresses #5965

## Test Plan
- [X] Plan runs all unit tests by default (including no-change plans)
- [X] Plan with model changes runs the full suite unless --test-changed-only is set
- [X] --test-changed-only runs only changed/restated models’ unit tests
- [X] Plan with no changes and --test-changed-only skips unit tests
- [X] --skip-tests still skips; incompatible with --test-changed-only
- [X] Docs cover run-all default and --test-changed-only

## Checklist

### DIFF
--- a/sqlmesh/cli/main.py
+++ b/sqlmesh/cli/main.py
@@ -424,6 +424,12 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     default=None,
 )
 @click.option(
+    "--all-tests",
+    is_flag=True,
+    help="Run all unit tests instead of only tests for models included in the plan.",
+    default=None,
+)
+@click.option(
     "--skip-linter",
     is_flag=True,
     help="Skip linting prior to generating the plan if the linter is enabled.",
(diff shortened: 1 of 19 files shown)


sqlmesh/cli/main.py
```diff
@@ -423,6 +423,12 @@ def diff(ctx: click.Context, environment: t.Optional[str] = None) -> None:
     help="Skip tests prior to generating the plan if they are defined.",
     default=None,
 )
+@click.option(
+    "--all-tests",
```
mday-io: Most of the other flags are `verb_*`. should this be `--include_all_tests`, or something similar?

mday-io (approved):

PR #6009, opened 2026-09-01
fix(test): support out-of-nanosecond-range timestamps in unit test comparisons

### LINKED ISSUE #5929: Bug: Failed to convert expected value into datetime for out-of-bounds nanosecond timestamp 0001-01-01 00:00:00 in unit tests
**Bug Description:**
When writing a SQLMesh unit test that includes a TIMESTAMP column with the value 0001-01-01 00:00:00, the test runner throws the following warnings and fails to compare the expected and actual values correctly.

**[WARNING] Failed to convert expected value for timestamp_field 
into datetime for unit test 'test_model'
(tests\test_model.yaml)'. 
Out of bounds nanosecond timestamp: 0001-01-01 00:00:00, at position 0.**

The value 0001-01-01 00:00:00 is a valid SQL TIMESTAMP but falls outside the range that pandas can handle with nanosecond precision (which starts at 1677-09-21).

**Steps to Reproduce:**
1. Create a model with a TIMESTAMP column that can contain '0001-01-01 00:00:00'.
2. Write a unit test with that value in the expected output
3. Run SQLMesh test

**Expected Behaviour:**
The test should either:
Successfully compare the timestamp values, or
Provide a clear error message with a suggested workaround

**Actual Behaviour:**
The test throws a 'Failed to convert expected value into datetime' warning, and the comparison fails.

**Attempted Workarounds:**
The following workarounds were attempted but did not resolve the issue:
- Specifying the column type as VARCHAR in the columns attribute of the test
- Using a SQL query format for the expected output with explicit CAST
- Using explicit CAST('0001-01-01 00:00:00' AS VARCHAR) in the output query
- Adding column_definitions to the model.

**Environment:**
SQLMesh version: 0.227.1
Database dialect: Redshift

mokashang:
## Description

Fixes #5929.

When a database engine returns a TIMESTAMP column as an object-dtype series of python `datetime.datetime` instances (Redshift is the reporter; any engine whose adapter returns object-dtype behaves the same), the unit-test `assert_equal` path parses the YAML expected values through `pd.to_datetime`, which defaults to nanosecond resolution. That resolution is bounded to 1677-09-21..2262-04-11 and overflows on values SQL TIMESTAMP fully supports — the reporter's example `0001-01-01 00:00:00` is a common sentinel value.

The overflow was caught by a broad `except Exception` inside `ModelTest.assert_equal` that logged a `Failed to convert expected value into datetime` warning and left `expected[col]` as the original strings. Two things happened next:
- For values whose repr survives str-coercion (e.g. `datetime.datetime(1, 1, 1)` → `"0001-01-01 00:00:00"`), the downstream `_to_hashable` normalization stringified both sides and the mismatch was **silent** — the test appeared to pass while emitting a warning.
- For values whose repr did not match, `pd.testing.assert_frame_equal` raised, producing a confusing failure with the same warning above it.

This PR extracts the parsing into `_parse_expected_datetime_column` and falls back to `datetime64[us]` resolution on `pd.errors.OutOfBoundsDatetime`. Microsecond resolution covers year 1 through year 294246, matching SQL TIMESTAMP's supported range, and it round-trips cleanly through `.dt.date` / `.dt.time` / `.dt.to_pydatetime()` for all three sentinel value types the caller already dispatches on. The fast path (`pd.to_datetime`) is preserved verbatim for in-range values, so nothing about existing tests changes.

## Test Plan

- Added `test_out_of_bounds_nanosecond_timestamp_comparison` in `tests/core/test_test.py`. It constructs an object-dtype `actual` DataFrame containing `datetime.datetime(1, 1, 1)` and `datetime.datetime(9999, 12, 31, 23, 59, 59)` — one below and one above pandas' default nanosecond range — and asserts (a) `assert_equal` succeeds and (b) no `Failed to convert expected value` warning is emitted. Both conditions are needed because the pre-fix code path silently mismatched on these particular values.
- Verified the test fails on unpatched `main` (the assertion on `log_warning` fires) and passes on this branch.
- `pytest tests/core/test_test.py` — 76 passed, 1 deselected (`test_pyspark_python_model`; environment lacks a Java runtime, unrelated to this change).
- `ruff check` and `ruff format --check` clean on both changed files. `mypy sqlmesh/core/test/definition.py` clean.

## Checklist

### DIFF
--- a/tests/core/test_test.py
+++ b/tests/core/test_test.py
@@ -2931,6 +2931,46 @@ def test_timestamp_normalization() -> None:
     )
 
 
+def test_out_of_bounds_nanosecond_timestamp_comparison(mocker: MockerFixture) -> None:
+    # https://github.com/TobikoData/sqlmesh/issues/5929
+    # Engines like Redshift may return a TIMESTAMP column as an object-dtype
+    # series of python `datetime.datetime` instances. Values outside pandas'
+    # default `datetime64[ns]` range (1677-09-21..2262-04-11) - which SQL
+    # `TIMESTAMP` fully supports - previously raised `OutOfBoundsDatetime`
+    # while parsing the expected values, producing a "Failed to convert
+    # expected value into `datetime`" warning and a false mismatch on values
+    # whose repr survives str-coercion (the values below happen to compare
+    # equal via `str()`, so the mismatch was silent).
+    test = _create_test(
+        body=load_yaml(
+            """
+test_foo:
+  model: sushi.foo
+  outputs:
+    query:
+      - ts_col: "0001-01-01 00:00:00"
+      - ts_col: "9999-12-31 23:59:59"
+            """
+        ),
+        test_name="test_foo",
+        model=_create_model("SELECT ts_col FROM raw"),
+        context=Context(config=Config(model_defaults=ModelDefaultsConfig(dialect="duckdb"))),
+    )
+    actual = pd.DataFrame(
+        {
+            "ts_col": pd.Series(
+                [datetime.datetime(1, 1, 1), datetime.datetime(9999, 12, 31, 23, 59, 59)],
+                dtype=object,
+            )
+        }
+    )
+    expected = pd.DataFrame({"ts_col": ["0001-01-01 00:00:00", "9999-12-31 23:59:59"]})
+    log_warning = mocker.spy(get_console(), "log_warning")
+    test.assert_equal(expected=expected, actual=actual, sort=False)
+    for call_args in log_warning.call_args_list:
+        assert "Failed to convert expected value" not in call_args.args[0]
+
+
 @use_terminal_console
 def test_disable_test_logging_if_no_tests_found(mocker: MockerFixture, tmp_path: Path) -> None:
     init_example_project(tmp_path, engine_type="duckdb")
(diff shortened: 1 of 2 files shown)


tests/core/test_test.py
```diff
@@ -2931,6 +2931,46 @@ def test_timestamp_normalization() -> None:
...
+    )
+    actual = pd.DataFrame(
+        {
+            "ts_col": pd.Series(
+                [datetime.datetime(1, 1, 1), datetime.datetime(9999, 12, 31, 23, 59, 59)],
+                dtype=object,
+            )
+        }
+    )
+    expected = pd.DataFrame({"ts_col": ["0001-01-01 00:00:00", "9999-12-31 23:59:59"]})
```
cmgoffena13: For these, just so the values do not exactly match and we can prove the code ran, could you put:
`0001-01-01T00:00:00` and `9999-12-31T23:59:59`

mokashang: Done in bba1caa — switched the expected column to `["0001-01-01T00:00:00", "9999-12-31T23:59:59"]` and added a short comment explaining why. If the datetime parse ever regresses back to string-only comparison, that mismatch against `str(datetime.datetime(1, 1, 1)) == "0001-01-01 00:00:00"` will now surface through `assert_equal` in addition to the existing `log_warning` check. The `T` form still parses to the same year-1/year-9999 instances via the `us` fallback, and the test passes locally.

PR #6011, opened 2026-09-02
fix: write file cache entries atomically to avoid read races

### LINKED ISSUE #6010: CI Fix - Windows Tests Race Condition
AI Notes

## Failing Test
`tests/lsp/test_reference_macro_find_all.py::test_multi_repo_macro_references`

## What's Wrong

The on-disk model cache is not written atomically. `FileCache.put` opens the cache file with `"wb"`, which truncates it, then pickle/gzip writes the contents.

With xdist, two workers can share the same cache file. A reader can open the file after truncate and before the write finishes. `pickle.load` then fails with “Ran out of input”. In this run, `silver.d` never made it into the LSP map.

This is a cache race, not a bad assertion in the LSP test.

## Possible Fix

In `sqlmesh/utils/cache.py` `FileCache.put`, stop writing the cache file in place (`"wb"` truncates it first).

1. Write the gzip/pickle to a temp file in the same directory.
2. `os.replace` that temp file onto the real cache path.

That way another pytest-xdist worker never reads a half-written `models__d` entry.

Optionally retry `FileCache.get` once if pickle hits EOF / “Ran out of input”, then load from SQL.

Don’t xfail the LSP test. The test is fine; the cache write is the bug.

7487:
## Description

Fixes #6010

`FileCache.put` opened the cache entry with `"wb"`, which truncates the file in place before the gzip/pickle write completes. A concurrent reader (e.g. another pytest-xdist worker sharing the cache directory) can open the file inside that window and hit `pickle.load` EOF ("Ran out of input"), which is what intermittently broke `tests/lsp/test_reference_macro_find_all.py::test_multi_repo_macro_references` on Windows CI.

This change writes the payload to a temp file in the same cache directory and `os.replace`s it onto the entry path, so readers only ever observe complete entries. Details:

- The temp file is prefixed with the cache version so the stale-file cleanup in `FileCache.__init__` doesn't unlink it mid-write (fresh atime keeps it alive; orphaned temp files from a crashed process still get cleaned up by the atime threshold).
- The store is best-effort: on Windows `os.replace` fails with a sharing violation if a reader still has the target open. In that case we log a warning and keep the existing intact entry rather than raising (mirroring how `get` handles unreadable entries).

I didn't add the optional `get` retry from the issue — `get` already swallows unpickling errors and `get_or_load` falls back to the loader, and with atomic writes a partial read can no longer happen.

## Test Plan

- Added `test_file_cache_put_is_atomic`, which verifies that a failed `os.replace` leaves the existing entry untouched and cleans up the temp file.
- `pytest tests/utils/test_cache.py` passes.
- Stress-checked with 2 readers looping `get` against a writer looping `put` on the same entry: on `main` this produced ~2900 failed reads (partial file -> unpickle EOF); with this change, 0.

## Checklist

- [x] I have run `make style` and fixed any issues (ruff check, ruff format and mypy are clean on the touched files)
- [x] All existing tests pass (ran `tests/utils`; remaining `fast-test` failures in my env are missing optional engine deps, unrelated to this change)
- [x] My commits are signed off (`git commit -s`) per the [DCO](https://github.com/SQLMesh/sqlmesh/blob/main/DCO)

### DIFF
--- a/sqlmesh/utils/cache.py
+++ b/sqlmesh/utils/cache.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import gzip
 import logging
+import os
 import pickle
 import shutil
+import tempfile
 import typing as t
 from pathlib import Path
 
@@ -125,8 +127,25 @@ class FileCache(t.Generic[T]):
         if not self._path.is_dir():
             raise SQLMeshError(f"Cache path '{self._path}' is not a directory.")
 
-        with gzip.open(self._cache_entry_path(name, entry_id), "wb", compresslevel=1) as fd:
-            pickle.dump(value, fd)
+        # Write to a temporary file and then atomically move it into place. Writing the
+        # entry in place ("wb" truncates the target file first) means that a concurrent
+        # reader (e.g. another pytest-xdist worker) could observe a partially written
+        # entry and fail to unpickle it.
+        tmp_fd, tmp_name = tempfile.mkstemp(dir=self._path, prefix=f"{self._cache_version}__tmp")
+        try:
+            with os.fdopen(tmp_fd, "wb") as raw_fd:
+                with gzip.open(raw_fd, "wb", compresslevel=1) as fd:
+                    pickle.dump(value, fd)
+            os.replace(tmp_name, self._cache_entry_path(name, entry_id))
+        except OSError as ex:
+            # Storing an entry is best-effort; e.g. on Windows os.replace fails if a
+            # concurrent reader still has the target file open.
+            logger.warning("Failed to store a cache entry '%s': %s", name, ex)
+        finally:
+            try:
+                os.unlink(tmp_name)
+            except OSError:
+                pass
 
     def exists(self, name: str, entry_id: str = "") -> bool:
         """Returns true if the cache entry with the given name and ID exists, false otherwise.
--- a/sqlmesh/utils/cache.py
+++ b/sqlmesh/utils/cache.py
@@ -136,11 +136,11 @@ class FileCache(t.Generic[T]):
             with os.fdopen(tmp_fd, "wb") as raw_fd:
                 with gzip.open(raw_fd, "wb", compresslevel=1) as fd:
                     pickle.dump(value, fd)
-            os.replace(tmp_name, self._cache_entry_path(name, entry_id))
-        except OSError as ex:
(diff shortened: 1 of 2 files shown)


sqlmesh/utils/cache.py
```diff
@@ -125,8 +127,25 @@ def put(self, name: str, entry_id: str = "", *, value: T) -> None:
...
+        # Write to a temporary file and then atomically move it into place. Writing the
+        # entry in place ("wb" truncates the target file first) means that a concurrent
+        # reader (e.g. another pytest-xdist worker) could observe a partially written
+        # entry and fail to unpickle it.
+        tmp_fd, tmp_name = tempfile.mkstemp(dir=self._path, prefix=f"{self._cache_version}__tmp")
+        try:
+            with os.fdopen(tmp_fd, "wb") as raw_fd:
+                with gzip.open(raw_fd, "wb", compresslevel=1) as fd:
+                    pickle.dump(value, fd)
+            os.replace(tmp_name, self._cache_entry_path(name, entry_id))
```
cmgoffena13: Your `try` wraps a lot here. I would keep the `try/finally` to unlink, but add in a `try/except` for the os.replace specifically, like this:
```
try:
    with...
         ....
    try:
         os.replace
     except OSError as ex:
          warning...
finally:
    try:
    ....
```

sqlmesh/utils/cache.py
```diff
@@ -125,8 +127,25 @@ def put(self, name: str, entry_id: str = "", *, value: T) -> None:
...
+        except OSError as ex:
+            # Storing an entry is best-effort; e.g. on Windows os.replace fails if a
```
cmgoffena13: I would just cut this down to: "Windows os.replace fails if a concurrent reader still has the target file open."

sqlmesh/utils/cache.py
7487: Done in 21588c5 — kept the try/finally for the temp-file unlink and moved the except OSError to wrap just the os.replace. One side effect worth noting: a failure while writing the temp file (e.g. disk full) now propagates instead of being logged, which I think is the right behavior anyway since nothing was stored.

sqlmesh/utils/cache.py
7487: Done in 21588c5.

PR #6012, opened 2026-09-02
fix: Simplify Java Installation in DevContainer

davem-bis:
## Description

An apparent breaking change in the SDKMAN! API caused the Java DevContainer feature to fail to install.

This change removes that feature and instead installs Java via the `msopenjdk-21` package.

An alternate solution would have been to update the DevContainer feature, however this method appeared more robust against future changes.

## Test Plan

Dev Container successfully builds and all tests run by `make fast-test` succeed.

A caveat - the test `tests/core/test_context.py::test_python_model_empty_df_raises` fails when run in parallel, but passes when run sequentially. I believe this is unrelated to this change and I am fixing in PR #6013.

## Checklist

- [ ] I have added tests for my changes (if applicable) (N/A)

### DIFF
--- a/.devcontainer/post-create-command.sh
+++ b/.devcontainer/post-create-command.sh
@@ -20,7 +20,7 @@ else
     rm packages-microsoft-prod.deb
 fi
 
-ALL_DEPENDENCIES="libpq-dev netcat-traditional unixodbc-dev default-jdk msodbcsql18"
+ALL_DEPENDENCIES="libpq-dev netcat-traditional unixodbc-dev default-jdk msodbcsql18 msopenjdk-21"
 sudo apt-get clean && sudo apt-get -y update && sudo ACCEPT_EULA='Y' apt-get -y install $ALL_DEPENDENCIES
 
 # Install Python dependencies
--- a/.devcontainer/post-create-command.sh
+++ b/.devcontainer/post-create-command.sh
@@ -20,7 +20,7 @@ else
     rm packages-microsoft-prod.deb
 fi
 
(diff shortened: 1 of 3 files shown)


.devcontainer/post-create-command.sh
```diff
@@ -20,7 +20,7 @@ else
     rm packages-microsoft-prod.deb
 fi
 
-ALL_DEPENDENCIES="libpq-dev netcat-traditional unixodbc-dev default-jdk msodbcsql18"
+ALL_DEPENDENCIES="libpq-dev netcat-traditional unixodbc-dev default-jdk msodbcsql18 msopenjdk-21"
```
cmgoffena13, line 23: I noticed default-jdk is still here, I think we need to remove it right?

davem-bis: Aha. Right I can see what has happened here.

- In the initial setup of the DevContainer I originally used the java feature to install java.
- I didn't realise at the time but that was adding the Microsoft apt repository which was allowing `msodbcsql18` to be installed here without manually adding the repository.
- I then added the rest of the requisites from `.github/scripts/install-prerequisites.sh` here, meaning a Java install was added here too.
- The Java DevContainer feature breaks, hence this PR. I completely missed the installation of `default-jdk` here, which turns out is sufficient for the dev container.
- `msopenjdk-21` could replace it, but I think staying as consistent to the `install-prerequisites.sh` script is preferable, so I've the MS one.

davem-bis: In summary, had 3 Java installs at one point! I've reduced it to 1.

PR #6026, opened 2026-09-04
fix(mssql): skip dbo on drop schema

### LINKED ISSUE #4833: mssql: Exclude dbo schema from drop schema statements
When using the CLI "destroy" command to expire an environment, exclude the MSSQL built in 'dbo' schema from DROP SCHEMA statements.

Error:
```
Error: Failed to drop the expired environment schema 'Database.dbo': (15150, b"Cannot drop the schema 'dbo'.DB-Lib error message 20018, severity 16:\nGeneral SQL Server error: Check messages from the SQL Server\n")
```

cmgoffena13:
## Description

Closes https://github.com/SQLMesh/sqlmesh/issues/4833

SQL Server does not allow the dropping of the `dbo` schema. Skipping for drop_schema method.

## Test Plan

New Test: `test_drop_schema_skips_dbo`

## Checklist

### DIFF
--- a/sqlmesh/core/engine_adapter/mssql.py
+++ b/sqlmesh/core/engine_adapter/mssql.py
@@ -180,7 +180,10 @@ class MSSQLEngineAdapter(
         **drop_args: t.Dict[str, exp.Expr],
     ) -> None:
         """
-        MsSql doesn't support CASCADE clause and drops schemas unconditionally.
+        MsSql doesn't support CASCADE clause so objects are dropped individually.
+
+        SQL Server also forbids dropping the built-in ``dbo`` schema (error 15150).
+        Objects inside it are still dropped when cascade=True, but the schema itself is left in place.
         """
         if cascade:
             objects = self._get_data_objects(schema_name)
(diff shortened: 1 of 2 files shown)


mday-io:
The `.db` extraction has an edge case that lets `dbo` slip through:

```python
schema = schema_name.db if isinstance(schema_name, exp.Table) else schema_name
if schema.lower() == "dbo":
    return
```

`exp.Table.db` is only populated when the object was built via `schema_()` (which explicitly sets `db=<schema>`). If it's built with `exp.to_table("dbo")` instead — which is a supported way to call `drop_schema`, per the existing `test_drop_schema_with_special_identifiers` test right above the new one — the schema name ends up in `.name`, not `.db`, and `.db` is just `""`. So the check silently fails to match, and it falls through to `DROP SCHEMA dbo` again.

Repro:
```python
schema_name = exp.to_table("dbo", dialect="tsql")
adapter.drop_schema(schema_name, cascade=False)
# still emits: DROP SCHEMA IF EXISTS [dbo];
```

Suggest widening the extraction to check both fields:
```python
schema = (schema_name.db or schema_name.name) if isinstance(schema_name, exp.Table) else schema_name
```

That covers both ways an `exp.Table` schema name shows up in this codebase.

cmgoffena13:
@mday-io -- good catch, addressed

PR #6040, opened 2026-09-09
fix(test): create ModelTests before concurrent pool workers

### LINKED ISSUE #6039: Concurrent model tests crash when `execution_time` is set
`sqlmesh test` with `concurrent_tasks > 1` can fail if tests set `vars.execution_time`:
```
TestError: Failed to create test ...
IndexError: list index out of range
```

If a unit test sets `execution_time`, we freeze the clock so `datetime.now()` and similar calls match that timestamp. The freeze (`time_machine`) changes `time.time()` for every thread in the process.

We take a lock around that freeze when rendering SQL (`_concurrent_render_context` in `sqlmesh/core/test/definition.py`), but we still build each test on a worker thread with no lock (`ModelTest.create_test()` in `sqlmesh/core/test/runner.py`).

Building a test calls `to_datetime()` (`ModelTest.__init__` in `definition.py`), which goes through `ttl_cache` (`sqlmesh/utils/__init__.py`) and hits `time.time()`. If another worker is starting or stopping the freeze at the same time, that call crashes.

Caught by `tests/core/test_test.py::test_freeze_time_concurrent`. Do not skip or weaken this test.

## Fix
Create the tests before we submit to the thread pool.

`sqlmesh/core/test/runner.py`
Function: `run_tests()`

Example
```python
tests = []  # <----NEW
for metadata, engine_adapter in metadata_to_adapter.items():  # <----NEW
    test = ModelTest.create_test(
        body=metadata.body,
        test_name=metadata.test_name,
        models=models,
        engine_adapter=engine_adapter,
        dialect=dialect,
        path=metadata.path,
        preserve_fixtures=preserve_fixtures,
        concurrency=num_workers > 1,
        verbosity=verbosity,
    )
    if test:  # <----NEW
        tests.append(test)  # <----NEW

def _run_single_test(test: ModelTest) -> ModelTextTestResult:  # <----NEW (signature)
    result = t.cast(
        ModelTextTestResult,
        ModelTextTestRunner().run(t.cast(unittest.TestCase, test)),
    )

    with lock:
        combined_results.merge(result)

    return result

test_results = []

start_time = time.perf_counter()
try:
    with ThreadPoolExecutor(max_workers=num_workers) as pool:
        futures = [pool.submit(_run_single_test, test) for test in tests]  # <----NEW

        for future in concurrent.futures.as_completed(futures):
            test_results.append(future.result())
```

## Validation
`pytest tests/core/test_test.py::test_freeze_time_concurrent -v` a few times. This is a race, so a single pass is not enough.

_Disclaimer: AI Notes_

devtechedge:
## Summary

When `sqlmesh test` runs with `concurrent_tasks > 1` and a unit test sets `vars.execution_time`, `ModelTest.create_test()` was running on worker threads. That path can call `to_datetime()` / `ttl_cache` (`time.time()`) while another worker is starting or stopping a `time_machine` freeze, which produces `IndexError: list index out of range` (or similar) during test creation.

This moves `ModelTest.create_test()` onto the calling thread before submitting work to the pool. Workers only run the already-built tests. Creation stays inside the same try/finally that closes engine adapters so an invalid `create_test` still cleans up connections.

Fixes #6039

## Test plan

- [x] `pytest tests/core/test_test.py::test_freeze_time_concurrent -v` passed 20 consecutive times locally (race; single pass is not enough)
- [ ] CI green

## Checklist

### DIFF
--- a/sqlmesh/core/test/runner.py
+++ b/sqlmesh/core/test/runner.py
@@ -125,9 +125,11 @@ def run_tests(
     # Ensure workers are not greater than the number of tests
     num_workers = min(len(model_test_metadata) or 1, default_test_connection.concurrent_tasks)
 
-    def _run_single_test(
-        metadata: ModelTestMetadata, engine_adapter: EngineAdapter
-    ) -> t.Optional[ModelTextTestResult]:
+    # Build ModelTest instances on the calling thread before workers start. create_test()
+    # can call to_datetime() / ttl_cache (time.time()), which races with another worker's
+    # time_machine freeze when execution_time is set under concurrent_tasks > 1.
+    tests: list[ModelTest] = []
+    for metadata, engine_adapter in metadata_to_adapter.items():
         test = ModelTest.create_test(
             body=metadata.body,
             test_name=metadata.test_name,
@@ -140,10 +142,10 @@ def run_tests(
             concurrency=num_workers > 1,
             verbosity=verbosity,
         )
+        if test:
+            tests.append(test)
 
-        if not test:
-            return None
-
+    def _run_single_test(test: ModelTest) -> ModelTextTestResult:
         result = t.cast(
             ModelTextTestResult,
             ModelTextTestRunner().run(t.cast(unittest.TestCase, test)),
@@ -159,10 +161,7 @@ def run_tests(
     start_time = time.perf_counter()
     try:
         with ThreadPoolExecutor(max_workers=num_workers) as pool:
-            futures = [
-                pool.submit(_run_single_test, metadata=metadata, engine_adapter=engine_adapter)
-                for metadata, engine_adapter in metadata_to_adapter.items()
-            ]
+            futures = [pool.submit(_run_single_test, test) for test in tests]
 
             for future in concurrent.futures.as_completed(futures):
                 test_results.append(future.result())
--- a/sqlmesh/core/test/runner.py
+++ b/sqlmesh/core/test/runner.py
@@ -125,26 +125,6 @@ def run_tests(
     # Ensure workers are not greater than the number of tests
     num_workers = min(len(model_test_metadata) or 1, default_test_connection.concurrent_tasks)
 
-    # Build ModelTest instances on the calling thread before workers start. create_test()
-    # can call to_datetime() / ttl_cache (time.time()), which races with another worker's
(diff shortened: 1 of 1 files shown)


sqlmesh/core/test/runner.py
```diff
@@ -159,10 +161,7 @@ def _run_single_test(
     start_time = time.perf_counter()
     try:
```
cmgoffena13, line 142: The `create_test` can hit an invalid test and leave connections open as well. So we should move the test creation into this try loop so we properly close connections

sqlmesh/core/test/runner.py
```diff
@@ -125,9 +125,11 @@ def run_tests(
     # Ensure workers are not greater than the number of tests
     num_workers = min(len(model_test_metadata) or 1, default_test_connection.concurrent_tasks)
 
-    def _run_single_test(
-        metadata: ModelTestMetadata, engine_adapter: EngineAdapter
-    ) -> t.Optional[ModelTextTestResult]:
+    # Build ModelTest instances on the calling thread before workers start. create_test()
+    # can call to_datetime() / ttl_cache (time.time()), which races with another worker's
+    # time_machine freeze when execution_time is set under concurrent_tasks > 1.
```
cmgoffena13: I would also add a `NOTE:` here detailing that we can run create_tests in a separate parallel stage for a future optimization.

We just can't overlap runs/creations

cmgoffena13 (changes requested):
Looks pretty close, I didn't notice connections being created. So right after connections are created we should make sure all relevant code is in a try statement to properly cleanup connections. Thanks for picking this up so quickly!

P.S. Don't forget our PR checklist. I added it to your description.

sqlmesh/core/test/runner.py
devtechedge: Good catch - thanks. I moved the calling-thread `create_test` loop into the same try/finally that closes the engine adapters, so an invalid create still cleans up connections. Also added a NOTE that create could be a separate parallel stage later, as long as it does not overlap with runs.

sqlmesh/core/test/runner.py
devtechedge: Added the NOTE about a future parallel create stage (no overlap with runs).

cmgoffena13 (approved):

